PR #46, opened 2015-01-08
Fixes building on a Mac using Homebrew as package manager

utzig:
Correctly finds and uses Homebrew's installation of libreadline/libffi/bison.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,9 @@ ifeq (Darwin,$(findstring Darwin,$(shell uname)))
 	# add macports/homebrew include and library path to search directories, don't use '-rdynamic' and '-lrt':
 	CXXFLAGS += -I/opt/local/include -I/usr/local/opt/readline/include
 	LDFLAGS += -L/opt/local/lib -L/usr/local/opt/readline/lib
+	# add homebrew's libffi include and library path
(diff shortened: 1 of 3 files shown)


utzig:
Thanks Clifford!

I suggested adding more developer friendly options for Homebrew eg "brew pkgconfig libffi", "brew path bison". I'll implement them and see if they get accepted upstream.

Also I had to clear ENABLE_COVER to build here. I suppose it's for code coverage so what tool should I have installed to use this? Any heads up?

utzig:
Oh the error is:

#define YS_ATTRIBUTE(...) __attribute__((__VA_ARGS__))
                                           ^
kernel/rtlil.cc:2325:2: error: argument to 'section' attribute is not valid for this target: mach-o section specifier requires a segment and section separated by a comma.

Btw, the arrow points to __VA_ARGS__. Any clues?

cliffordwolf:
No external tool is required. It is all built in to Yosys (see "help cover"). But if your compiler does not support `__attribute__((section(...)))` (or something equivalent) then I don't see how this feature can be ported. But I don't see it as a problem if this feature is only available on Linux (my main dev platform). So I have now simply disabled code coverage counters for all non-linux systems (see 95f1eb9b87064719c17d45f24775c3e97497b6df).

PR #48, opened 2015-02-01
Fixed typos found by lintian

rubund:


rubund:
When building the debian package, these typos caused some lintian warnings. Nice to get rid of the warnings..

PR #62, opened 2015-05-07
Added support for $mem cells in the verilog backend.

wluker:
This adds support for $mem cells to the verilog backend, but I do have some open questions:

1. The $mem cell parameter "\\WR_EN" seems to have width number of bits for every writer and all of the bits for each port appear to be the same, at least for my simple trials. For example here is the parameter for 2 write ports : *connect \WR_EN { 8'11111111 \wr_en \wr_en \wr_en \wr_en \wr_en \wr_en \wr_en \wr_en }*. Will these enable bits ever be different? 
I've added some code to check if they are different and fail if they are, but I'm not sure if this is necessary.

2. The "RD_TRANSPARENT" parameter appears to be unused. Is it ever used and what does it indicate?

Suggestion : Does it make sense to add an init member to RTLIL::Memory?

### DIFF
--- a/backends/verilog/verilog_backend.cc
+++ b/backends/verilog/verilog_backend.cc
@@ -788,8 +788,127 @@ bool dump_cell_expr(std::ostream &f, std::string indent, RTLIL::Cell *cell)
 		return true;
 	}
 
+	if (cell->type == "$mem")
+	{
+		std::ostringstream os;
+		RTLIL::IdString memid = cell->parameters["\\MEMID"].decode_string();
+		std::string mem_id = id( cell->parameters["\\MEMID"].decode_string() );
+		int abits = cell->parameters["\\ABITS"].as_int();
+		int size = cell->parameters["\\SIZE"].as_int();
+		int width = cell->parameters["\\WIDTH"].as_int();
+		int offset = cell->parameters["\\OFFSET"].as_int();
+		bool use_init = !(RTLIL::SigSpec( cell->parameters["\\INIT"] ).is_fully_undef());
+
+		// for memory block make something like:
+		//  reg [7:0] memid [3:0];
+		//  initial begin
+		//    memid[0] <= ...
+		//  end
+		int mem_val;
+		RTLIL::Memory memory;
+		memory.name = memid;
+		memory.width = width;
+		memory.start_offset = offset;
+		memory.size = size;
+		dump_memory(os, indent.c_str(), &memory);
+		if (use_init)
+		{
+			os << stringf("%s" "initial begin\n", indent.c_str());
+			for (int i=0; i<size; i++)
+			{
+				mem_val = cell->parameters["\\INIT"].extract(i*width, width).as_int();
+				os << stringf("%s" "  %s[%d] <= %d'd%d;\n", indent.c_str(), mem_id.c_str(), i, width, mem_val);
+			}
+			os << stringf("%s" "end\n", indent.c_str());
+		}
+
+
+		int nread_ports = cell->parameters["\\RD_PORTS"].as_int();
+		RTLIL::SigSpec sig_rd_clk, sig_rd_data, sig_rd_addr;
+		bool use_rd_clk, rd_clk_posedge;
+		// read ports
+		for (int i=0; i < nread_ports; i++)
+		{
+			sig_rd_clk = cell->getPort("\\RD_CLK").extract(i);
+			sig_rd_data = cell->getPort("\\RD_DATA").extract(i*width, width);
+			sig_rd_addr = cell->getPort("\\RD_ADDR").extract(i*abits, abits);
+			use_rd_clk = cell->parameters["\\RD_CLK_ENABLE"].extract(i).as_bool();
+			rd_clk_posedge = cell->parameters["\\RD_CLK_POLARITY"].extract(i).as_bool();
+			if (use_rd_clk)
+			{
+				// for clocked read ports make something like:
+				//   always @(posedge clk)
+				//      r_data <= array_reg[r_addr];
+				os << stringf("%s" "always @(%sedge ", indent.c_str(), rd_clk_posedge ? "pos" : "neg");
+				dump_sigspec(os, sig_rd_clk);
+				os << stringf(")\n");
+				os << stringf("%s" "  ", indent.c_str());
+				dump_sigspec(os, sig_rd_data);
+				os << stringf(" <= %s[", mem_id.c_str());
+				dump_sigspec(os, sig_rd_addr);
+				os << stringf("];\n");
+			}else{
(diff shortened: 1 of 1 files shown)


cliffordwolf:
Re. your two questions: the following module will use RD_TRANSPARENT and smaller-than-width WR_EN groups:

	module test (
		input clk,
		input [3:0] addr, di,
		input [1:0] we,
		output [3:0] do
	);
		reg [3:0] mem [0:15];
		reg [3:0] raddr;
		always @(posedge clk) begin
			if (we[0]) mem[addr][1:0] <= di[1:0];
			if (we[1]) mem[addr][3:2] <= di[3:2];
			raddr <= addr;
		end
		assign do = mem[raddr];
	endmodule

please extend your patch accordingly. otherwise your patch looks good. (there seems to be a minor issue with blank vs. space indenting)

Re. init member to RTLIL::Memory: instead of, or in addition to $meminit cells? I cannot see instead of working because some initial blocks yield logic in the verilog front-end that must be const-folded by subsequent optimizations. for this I need wires that can be driven by logic. In addition to $meminit cells would be working. I have an opt_memory pass on my long todo list. This could also pack the const $meminit cells into RTLIL::Memory (ordered by priority, stopping at the first non-const cell). This would be something for the yosys-0.7 roadmap.

wluker:
The new commit should now handle your example above.

On another note, I have a question about RTLIL::SigSpec::is_wire(). If I use is_wire() on a SigSpec that is something like we[0], a single bit in a larger signal, should it return true or false? It seems to be returning false, which surprises me, but maybe I'm misunderstanding this.

cliffordwolf:
> The new commit should now handle your example above.

excellent.

just a few more issues before I merge this:

can you explain the rationale behind bit_check_equal()? why not just using `sigmap(current_bit) == sigmap(last_bit)`?

Also: you might want to consider using RTLIL::SigBit for current_bit and last_bit, and RTLIL::SigSpec for lof_wen, using lof_wen.append_bit() instead of lof_wen.push_back(). (Most of the code in the verilog backend predates "modern" Yosys APIs like RTLIL::SigBit. So it maybe is not the best example code to look at..)

Another thing: Whenever possible use dict<> instead of std::map<> (and pool<> instead of std::set<>). See CodingReadme for details.

> RTLIL::SigSpec::is_wire()

This is true if the sigspec is exactly equal to a wire. So the sigspec we[0] only is_wire() if the width of we is 1, i. e. if we[0] is the same as we.

The function RTLIL::SigSpec::as_wire() does log_assert() that RTLIL::SigSpec::is_wire() is true.

wluker:
> can you explain the rationale behind bit_check_equal()? why not just using sigmap(current_bit) == sigmap(last_bit)?

I also wanted to group together sequences of 0's and 1's, so I felt that I needed an additional check for constants before using sigmap to check for wires. I had thought sigmap had thrown an error when I passed in a constant, but I can re-check that. I'll also look at using RTLIL::SigBit.

> This is true if the sigspec is exactly equal to a wire. ...

Ok, that makes sense now. Thanks for the explanation.

This has been a learning experience for me, so thanks for taking the time to provide examples and advise on the coding. I'll get back to you on the recommended changes.

cliffordwolf:
> I had thought sigmap had thrown an error when I passed in a constant, but I can re-check that.

please do. i'm sure it works..

> This has been a learning experience for me, so thanks for taking the time to provide examples and advise on the coding.

I welcome every chance to force some knowledge about yosys internals on people.. An open pull request, waiting to be merged, is such an occasion. ;)

However, when you do not have an open pull request, don't hesitate to ask API questions on the yosys subreddit. I'll be happy to answer them.

wluker:
Ok, I've made the changes you recommended.
Thanks for your help.

cliffordwolf:
I've merged your changes, but now I've seen that your code segfaults in "make test". So I've disabled your code again in commit 9e56739634bc8aef9b0b342a47d4b01eeb116e36. Please fix and open a new pull request with your fix.

You can reproduce the problem by running

    ./yosys -p "proc; opt; memory -nomap; opt -fast -full; dump" -b verilog test.v

On the following input:

    module bram_00_01(B1ADDR, A1ADDR, B2ADDR, B2EN, B1EN, B1DATA, A1DATA, CLK1, CLK3, B2DATA);
      (* nomem2reg *) reg [4:0] memory [0:3];
      `ifndef SYNTHESIS
        event UPDATE_A1;
      `endif
      input CLK3;
      input [1:0] A1ADDR;
      output reg [4:0] A1DATA;
      `ifndef SYNTHESIS
        event UPDATE_B1;
      `endif
      input CLK1;
      input [1:0] B1ADDR;
      input [4:0] B1DATA;
      input [0:0] B1EN;
      parameter CLKPOL3 = 0;
      wire CLK1_CLKPOL3 = CLK1 == CLKPOL3;
      `ifndef SYNTHESIS
        event UPDATE_B2;
      `endif
      input [1:0] B2ADDR;
      input [4:0] B2DATA;
      input [0:0] B2EN;
      always @(posedge CLK3) begin
        `ifndef SYNTHESIS
          -> UPDATE_A1;
        `endif
        A1DATA <= memory[A1ADDR];
      end
      always @(posedge CLK1_CLKPOL3) begin
        `ifndef SYNTHESIS
          #2; -> UPDATE_B1;
        `endif
        if (B1EN[0]) memory[B1ADDR][4:0] = B1DATA[4:0];
        `ifndef SYNTHESIS
          #1; -> UPDATE_B2;
        `endif
        if (B2EN[0]) memory[B2ADDR][4:0] = B2DATA[4:0];
      end
    endmodule

wluker:
Please see the new pull request. It was a simple fix, but something I should have caught earlier. 
Thanks for pointing out the failing example.

cliffordwolf:
Still fails "make test". Simplified example:

    ./yosys -p 'synth -run coarse' -o synth.v test.v
    iverilog synth.v

Will result in:

    synth.v:17: error: B1DATA is not a valid l-value in bram_01_00.
    synth.v:12:      : B1DATA is declared here as wire.
    Elaboration failed

For the following input:

    module bram_01_00(B1ADDR, A1ADDR, B1DATA, A1DATA, CLK1);
      input CLK1;
      input [6:0] A1ADDR;
      input [7:0] A1DATA;
      input [6:0] B1ADDR;
      output reg [7:0] B1DATA;
      reg [7:0] memory [0:127];
      always @(posedge CLK1) begin
        memory[A1ADDR] <= A1DATA;
        B1DATA <= memory[B1ADDR];
      end
    endmodule

Typing is a bit of a challenge for me at the moment:
- https://twitter.com/oe1cxw/status/596636063134818304
- https://twitter.com/oe1cxw/status/597275275232161792

So I would appreciate it if you could make sure on your end that your PR passes a "make test"..

PR #64, opened 2015-05-11
In test scripts changed python to python2.

wluker:
On my system (archlinux), python is a symlink to python3 by default. The yosys test scripts used in "make test" are written in python2 syntax and error out when called from python3, so this just make python2 explicit.


cliffordwolf:
Commit dae00e1d8301aae25f3432ca12995c9a3b380679 should fix this issue.

PR #66, opened 2015-06-04
Bug fix in $mem verilog backend + changed tests/bram flow of make test.

wluker:
Hi Clifford,
This should *finally* complete the $mem cell portion of the verilog backend. This was a pain in the butt, but mostly because of the simulation differences in tests/bram.
I've tried to use your advice given on my reddit question (at least the way I understood it) and had to change the command sequence in tests/bram/run_single.sh

The flow of test/bram is now:
1. attempt to map to brams (as it was before)
2. any memories that were not matched to a bram fall through and the output verilog is written to file (using the new $mem code)
3. read the verilog back into yosys, combine into $mem cells, write the verilog back to file, but this time don't use the new $mem cell code; this allows the $mem cells to be defined by simlib.v and should also test the generated code

Using this flow I have been able to get "make test" to pass, where is was not passing before due to the race conditions.

### DIFF
--- a/backends/verilog/verilog_backend.cc
+++ b/backends/verilog/verilog_backend.cc
@@ -38,7 +38,7 @@
 USING_YOSYS_NAMESPACE
 PRIVATE_NAMESPACE_BEGIN
 
-bool norename, noattr, attr2comment, noexpr, nomem;
+bool norename, noattr, attr2comment, noexpr;
 int auto_name_counter, auto_name_offset, auto_name_digits;
 std::map<RTLIL::IdString, int> auto_name_map;
 std::set<RTLIL::IdString> reg_wires, reg_ct;
@@ -791,7 +791,7 @@ bool dump_cell_expr(std::ostream &f, std::string indent, RTLIL::Cell *cell)
 		return true;
 	}
 
-	if (cell->type == "$mem" && nomem == false)
+	if (cell->type == "$mem")
 	{
 		RTLIL::IdString memid = cell->parameters["\\MEMID"].decode_string();
 		std::string mem_id = id(cell->parameters["\\MEMID"].decode_string());
@@ -818,6 +818,13 @@ bool dump_cell_expr(std::ostream &f, std::string indent, RTLIL::Cell *cell)
 			f << stringf("%s" "end\n", indent.c_str());
 		}
 
+		// create a map : "edge clk" -> expressions within that clock domain
+		dict<std::string, std::vector<std::string>> clk_to_lof_body;
+		clk_to_lof_body[""] = std::vector<std::string>();
+		std::string clk_domain_str;
+		// create a list of reg declarations
+		std::vector<std::string> lof_reg_declarations;
+
 		int nread_ports = cell->parameters["\\RD_PORTS"].as_int();
 		RTLIL::SigSpec sig_rd_clk, sig_rd_data, sig_rd_addr;
 		bool use_rd_clk, rd_clk_posedge, rd_transparent;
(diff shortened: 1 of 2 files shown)


cliffordwolf:
Nope. You have not fixed it, just broken the test cases capability of detecting that the code generated is depending on unspecified behavior. Here is how the generated code should be fixed instead:

    always @(posedge CLK2_CLKPOL3)
      if (A1EN)
        memory[A1ADDR] <= A1DATA;
    always @(posedge CLK2_CLKPOL3)
      if (A2EN)
        memory[A2ADDR] <= A2DATA;
    always @(posedge CLK2_CLKPOL3)
      if (B1EN)
        memory[B1ADDR] <= B1DATA;
    always @(posedge CLK2_CLKPOL3)
      if (B2EN)
        memory[B2ADDR] <= B2DATA;
    always @(negedge CLK1)
      if (C1EN)
        memory[C1ADDR] <= C1DATA;

This is from a test case that fails. See the 3 always blocks for CLK2_CLKPOL3? The verilog standard allows them to be "evaluated" in arbitrary order, i.e. it is unspecified which port takes precedence in case of a write address conflict.

Yosys will "evaluate" them in-order. That's why you don't see a problem when you run it through yosys a 2nd time. But other tools will behave differently!! (And are still complying to the Verilog standard!)

The solution is of course to consolidate the three write ports for the one clock into a single always block like so:

    always @(posedge CLK3_CLKPOL2)
      _3_ <= memory[D1ADDR];
    assign D1DATA = _3_;
    always @(posedge CLK2_CLKPOL3) begin
      if (A1EN)
        memory[A1ADDR] <= A1DATA;
      if (A2EN)
        memory[A2ADDR] <= A2DATA;
      if (B1EN)
        memory[B1ADDR] <= B1DATA;
      if (B2EN)
        memory[B2ADDR] <= B2DATA;
    end
    always @(negedge CLK1)
      if (C1EN)
        memory[C1ADDR] <= C1DATA;

With this manual change the generated Verilog for this test case passed the test.

wluker:
Right, I was just trying to use simlib.v to get around the unspecified order for iverilog used in tests/bram. I thought that was the point of your reddit response. 

I can certainly output the code in the form suggested above, but due to the use of non-blocking assignments (<= vs =) it still looks like a race condition to me. I'm still learning the nuances of verilog I guess.

I'll get back to you with the new form...

wluker:
See the latest commit for grouping the writes together (as you suggested above).

Unfortunately "make test" still does not pass in tests/bram. When I pull out the failing tests there appears to still be race conditions and a simulation mismatch. Using a testbench generated from  test_autotb, the only way I can get rid of the simulation mismatch is to manually add the matching "`ifndef SYNTHESIS #?; `endif" to my generated code. To me, this doesn't seem like a reasonable thing to have to do.

You can use this PR if you desire, or not, but unfortunatly I don't have the time to chase any more simulation issues. I personally don't expect determinate behavior when simulating write collisions anyway, so I don't see the value in spending such a large effort on this. I do however appreciate your attention to detail and your desire for full compliance with the verilog standard, so it won't bother me in the least if you choose to not use this PR.

cliffordwolf:
Thanks for sticking with this issue for so long.

I have now merged your code and [fixed the remaining issue](https://github.com/cliffordwolf/yosys/commit/b57cb4a7fea01b1e5b3242d3ee4bd90077bac960) of `@(nededge clk)` and `@(posedge !clk)` not being merged correctly into one single clock for $mem. Now "make test" passes fine.

PR #70, opened 2015-08-10
Bugfix

gaomy3832:
One bug fix to add blackbox modules into the known celltype in opt_clean to avoid delete them.

And another minor bug fix for USE_CELL_HASH_CACHE option.

### DIFF
--- a/passes/opt/opt_clean.cc
+++ b/passes/opt/opt_clean.cc
@@ -350,6 +350,11 @@ struct OptCleanPass : public Pass {
 		ct.setup_stdcells();
 		ct.setup_stdcells_mem();
 
+		for (auto module : design->modules()) {
+			if (module->get_bool_attribute("\\blackbox"))
+				ct.setup_module(module);
+		}
+
 		ct_reg.setup_internals_mem();
 		ct_reg.setup_stdcells_mem();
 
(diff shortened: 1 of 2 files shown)


cliffordwolf:
Can you give an example of the problem fixed by the "Avoid deleting blackbox modules in opt_clean." commit?

cliffordwolf:
Btw: I did now merge the "Bugfix for cell hash cache option in opt_share." commit (8c4c62f3e1f6d0e6128c4b88f764cc58a744aee8).

gaomy3832:
I am sorry that the commit message was incorrect. It was a bugfix some time ago when I was busy doing a project. I don't have time to submit it until recent. So I didn't remember the intention clearly.

The commit was actually indent to be able to remove unused blackbox modules, rather than keeping them. Consider the following case:
```
// foo.v
// a blackbox
module foo(a, b, y);
input a, b;
output y;
endmodule

// bar.v
// top design
module bar(a, b, y);
input a, b;
output y;
wire t;
// i0 is good
foo i0(.a(a), .b(b), .y(y));
// i1 is unused and should be removed by opt_clean
foo i1(.a(a), .b(b), .y(t));
endmodule
```

The previous version could not remove the unused module i1. Run commands
```
read_verilog -lib foo.v
read_verilog bar.v
hierarchy -check -top bar;
proc; opt; flatten; opt;
opt -full -purge;
stat
```
gives
```
=== design hierarchy ===

   bar                               1

   Number of wires:                  4
   Number of wire bits:              4
   Number of public wires:           4
   Number of public wire bits:       4
     foo                             2
```

This commit fixes the problem.
```
=== design hierarchy ===

   bar                               1

   Number of wires:                  3
   Number of wire bits:              3
     foo                             1
```

I've changed the commit message to reflect the correct intention.

cliffordwolf:
Ah, ok. Yes. This makes much more sense..

PR #77, opened 2015-08-24
Microsoft Visual C++ 2013 fixes, 64-bit build.

Andrei-Errapart:
These are C++ source code changes only; no build scripts, no project files.

Only 64-bit build tested.

There are about 10000 compiler warnings about possible loss of data when converting size_t to unsigned int, etc.

### DIFF
--- a/kernel/bitpattern.h
+++ b/kernel/bitpattern.h
@@ -30,7 +30,7 @@ struct BitPatternPool
 	int width;
 	struct bits_t {
 		std::vector<RTLIL::State> bitdata;
-		unsigned int cached_hash;
+		mutable unsigned int cached_hash;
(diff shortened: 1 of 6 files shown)


cliffordwolf:
I did not apply the BigInteger part of the patch, because with clang this would give me the following build error:

    libs/bigint/BigInteger.cc:56:49: error: call to constructor of 'BigUnsigned' is ambiguous
    BigInteger::BigInteger(unsigned long long  x) : mag(x) { sign = mag.isZero() ? zero : positive; }
                                                    ^   ~
    libs/bigint/BigUnsigned.hh:57:2: note: candidate constructor
            BigUnsigned(unsigned long  x);
    ...
 
I've tested the other commits with clang and gcc-4.6 and they looked good. So I merged the other three commits now.

Issue #114, opened 2016-01-30
read_verilog is very slow when module has many for iterations in an always block


kc8apf:
read_verilog seems to take exponentially longer reading the following module depending on the number of iterations in the for loop:

    module TestCase(
      input wire clock,

    );

    localparam COUNT = 1600;

    reg buffer[0:COUNT-1];

    always @(posedge clock) begin : GENERATE
      integer ii;

      for(ii = 0; ii < COUNT; ii = ii + 1) begin
        buffer[ii] <= 1'b0;
      end
    end

    endmodule

Time spent by 'yosys -p read_verilog test.v':
* 100 iterations -> 0s
* 200 iterations -> 2s
* 400 iterations -> 8s
* 800 iterations -> 36s
* 1600 iterations -> 160s

The OS X performance tools ([Sample of yosys](https://github.com/cliffordwolf/yosys/files/110653/Sample.of.yosys.txt)) seem to implicate Yosys::AST_INTERNAL::ProcessGenerator::removeSignalFromCaseTree(const std::set<RTLIL::SigBit> &pattern, RTLIL::CaseRule *cs) which is calling Yosys::RTLIL::SigSpec::remove2(const RTLIL::SigSpec &pattern, RTLIL::SigSpec *other).  This form of remove2() calls to_sigbit_pool() to get a pool<RTLIL::SigBit>.  Inside to_sigbit_pool(), it creates a pool<RTLIL::SigBit> and then inserts each each bit of each chunk.  This insert seems to trigger a memory allocation via pool<>::insert calling pool<>::do_insert which does a push_back() on a new vector.  It seems that creating a pool on each call to remove2() is excessive in this case.

cliffordwolf:
For the record: Fixed by PR #115.

PR #115, opened 2016-01-30
RTLIL::SigSpec::remove2() speedups

kc8apf:
As described in #114, remove2() as used by Yosys::AST_INTERNAL::ProcessGenerator::removeSignalFromCaseTree() was doing a lot of memory allocations and causing a performance bottleneck.  These changes mostly remove unnecessary copies.  With these changes, read_verilog of the test module from #114 with COUNT=1600 improves from 160s to 4s.


cliffordwolf:
Well done! Thanks!

PR #116, opened 2016-01-31
More RTLIL performance improvements

kc8apf:
More changes to avoid unnecessary copies and memory allocations.  I've been using the following module and synthesizing to Xilinx as the test case:
    module test(
      input wire clock
    );

    parameter ITERATIONS = 1600;

    reg buffer[0:ITERATIONS-1];

    always @(posedge clock) begin : GENERATE
      integer ii;
      for (ii = 0; ii < ITERATIONS; ii = ii + 1) begin
        buffer[ii] <= 1'b0;
      end
    end

    endmodule

Building at commit 5462399c880c4736d0382bf5ba294e97637b393d, read_verilog+synth_xilinx took 79s as reported by yosys.  With these changes, it now takes 22s.

### DIFF
--- a/frontends/ast/genrtlil.cc
+++ b/frontends/ast/genrtlil.cc
@@ -379,7 +379,7 @@ struct AST_INTERNAL::ProcessGenerator
 	// e.g. when the last statement in the code "a = 23; if (b) a = 42; a = 0;" is processed this
 	// function is called to clean up the first two assignments as they are overwritten by
 	// the third assignment.
-	void removeSignalFromCaseTree(const std::set<RTLIL::SigBit> &pattern, RTLIL::CaseRule *cs)
+	void removeSignalFromCaseTree(const RTLIL::SigSpec &pattern, RTLIL::CaseRule *cs)
 	{
 		for (auto it = cs->actions.begin(); it != cs->actions.end(); it++)
 			it->first.remove2(pattern, &it->second);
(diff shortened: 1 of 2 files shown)


cliffordwolf:
Fyi: I've [now added reserve() methods](https://github.com/cliffordwolf/yosys/commit/13e15a24a206f56d9daa15665aba4948fd5dfef5) to the hashlib classes that can be used to avoid successive rehashing when creating a dict<> or pool<>.

kc8apf:
That should help for cases where the dict<> or pool<> is kept for a while.
SigMap shows up in profiles a fair amount but I'm having a hard time
understanding how mfp<> works. I think there are a few passes that are
using them as temporary objects and loading them with many SigSpec objects
each time.

cliffordwolf:
mfp<> is a [merge–find set](https://en.wikipedia.org/wiki/Disjoint-set_data_structure). The name stands for merge-find-promote. (The promote() method makes an element the representative of its set.)

It's primary user is SigMap. This example for what SigMap does is from the CodingReadme file:

        SigBit a = module->addWire(NEW_ID);
        SigBit b = module->addWire(NEW_ID);
        module->connect(a, b);

        log("%d\n", a == b); // will print 0

        SigMap sigmap(module);
        log("%d\n", sigmap(a) == sigmap(b)); // will print 1

Most passes are structured into an analysis phase in which is decided what to do, followed by an action phase in which it is done. In most cases a SigMap object is created for the module at the beginning of the analyses phase after the action phase.

I have now committed some performance improvements [in mfp<>](https://github.com/cliffordwolf/yosys/commit/ea492abcf01ee588a9756f8ae1f632ff418cd30c) and [in SigMap](https://github.com/cliffordwolf/yosys/commit/01bcc5663fa7c65ddbb6c55abebcfc60af8d200b). In my benchmark (`yosys -Sv2 picorv32.v`) the total runtime improved by ~1.9% (from 17.32 seconds to 16.99 seconds, median of 3 runs for each case).

PR #180, opened 2016-06-18
Added support for SystemVerilog packages with localparam definitions

rubund:
Hi Clifford,

I've put together some code which enables having packages in separate files with localparams. This is a pretty common thing to do in SystemVerilog. The behaviour seems to be correct after the changes. I don't expect that you accept the pull request right away, but I'm looking for feedback.

At first I made a global vector with all packages, but this did not fit well with the "Design" concept, so therefore I had to add a package-vector as a member of the Design struct. This requires including "ast.h" in rtlil.h

It is not very efficient to always copy all members of every package to every module, but to change this, I wanted to (in simplify) check if the identifier-name contains "::", and in that case include all members of the corresponding packages in current_scope. This would require defining some global variable for having access to the package-vector in simplify, or I would have to add another argument to the function argument list of simplify. I guess you would prefer global variables?

Best regards,
Ruben

### DIFF
--- a/frontends/ast/ast.cc
+++ b/frontends/ast/ast.cc
@@ -997,7 +997,7 @@ void AST::process(RTLIL::Design *design, AstNode *ast, bool dump_ast1, bool dump
 			for (auto n : global_decls)
 				(*it)->children.push_back(n->clone());
 
-			for (auto n : design->packages){
+			for (auto n : design->verilog_packages){
 				for (auto o : n->children) {
 					AstNode *cloned_node = o->clone();
 					cloned_node->str = n->str + std::string("::") + cloned_node->str.substr(1);
@@ -1023,7 +1023,7 @@ void AST::process(RTLIL::Design *design, AstNode *ast, bool dump_ast1, bool dump
 			design->add(process_module(*it, defer));
 		}
 		else if ((*it)->type == AST_PACKAGE){
-			design->packages.push_back((*it)->clone());
+			design->verilog_packages.push_back((*it)->clone());
 		}
 		else
 			global_decls.push_back(*it);
--- a/frontends/ast/ast.cc
+++ b/frontends/ast/ast.cc
@@ -151,6 +151,7 @@ std::string AST::type2str(AstNodeType type)
 	X(AST_POSEDGE)
 	X(AST_NEGEDGE)
 	X(AST_EDGE)
+	X(AST_PACKAGE)
 #undef X
 	default:
 		log_abort();
@@ -996,6 +997,14 @@ void AST::process(RTLIL::Design *design, AstNode *ast, bool dump_ast1, bool dump
 			for (auto n : global_decls)
 				(*it)->children.push_back(n->clone());
 
+			for (auto n : design->packages){
+				for (auto o : n->children) {
+					AstNode *cloned_node = o->clone();
+					cloned_node->str = n->str + std::string("::") + cloned_node->str.substr(1);
+					(*it)->children.push_back(cloned_node);
+				}
+			}
+
 			if (flag_icells && (*it)->str.substr(0, 2) == "\\$")
 				(*it)->str = (*it)->str.substr(1);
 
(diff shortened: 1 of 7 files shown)


frontends/ast/ast.cc
```diff
@@ -996,6 +997,14 @@ void AST::process(RTLIL::Design *design, AstNode *ast, bool dump_ast1, bool dump
 			for (auto n : global_decls)
 				(*it)->children.push_back(n->clone());
 
+			for (auto n : design->packages){
+				for (auto o : n->children) {
+					AstNode *cloned_node = o->clone();
+					cloned_node->str = n->str + std::string("::") + cloned_node->str.substr(1);
```
rubund: maybe it's not correct to operate directly on the "str" member?

kernel/rtlil.h
```diff
@@ -792,6 +793,7 @@ struct RTLIL::Design
 
 	int refcount_modules_;
 	dict<RTLIL::IdString, RTLIL::Module*> modules_;
+	std::vector<AST::AstNode*> packages;
```
rubund: Added to as a member of this class since the group of packages belong to the Design context.

cliffordwolf:
Looks good. Some minor change requests and remarks:

- Rename Design::packages to Design::verilog_packages.

- The `#include "frontends/ast/ast.h"` in `kernel/rtlil.h` shouldn't be needed. The header `kernel/yosys.h` already contains `namespace AST { struct AstNode; }`, which should be sufficient to have AstNode pointers.

- Add a simple example `.sv` file to `tests/simple/` that is using packages. (I will add .sv support to the autotest framework.)

- What happens if a package is re-declared, for example because a .svh containing packages is included from multiple .sv files? Is it going to produce an error? Is it going to have O(n^2) performance (for n files) because the list of packages in Design::verilog_packages is constantly growing and each new file is importing a larger list of (redundant) packages?

rubund:
Thanks for the quick answer.

> - Rename Design::packages to Design::verilog_packages.

will do 

> - The `#include "frontends/ast/ast.h"` in `kernel/rtlil.h` shouldn't be needed. The header `kernel/yosys.h` already contains `namespace AST { struct AstNode; }`, which should be sufficient to have AstNode pointers.

aah. First I added it before I found the correct namespace.. I will remove it
again.
 
> - Add a simple example `.sv` file to `tests/simple/` that is using packages. (I will add .sv support to the autotest framework.)

Cool. I will also need to an example test where the package is in a separate
file.

> - What happens if a package is re-declared, for example because a .svh containing packages is included from multiple .sv files? Is it going to produce an error? Is it going to have O(n^2) performance (for n files) because the list of packages in Design::verilog_packages is constantly growing and each new file is importing a larger list of (redundant) packages?

I think the variables should either be overwritten or ignored. I need to check
what other commercial tools do or if this is written in the spec somewhere.
Maybe we also should print out a warning telling that "Redefinition is
ignored/overwritten".

Cheers,
Ruben

rubund:
Btw.
Does the autotest framework depend on iverilog being capable of understanding SV packages ?

cliffordwolf:
> Does the autotest framework depend on iverilog being capable of understanding SV packages ?

Yes. (You need to run iverilog with `-g2005-sv`, `-g2009`, or `-g2012` to enable SystemVerilog features.)

> Cool. I will also need to an example test where the package is in a separate file.

Not easy to do with the current "make test" framework. I'd say this is something that should be tested by adding larger SystemVerilog designs to [yosys-bigsim](https://github.com/cliffordwolf/yosys-bigsim).

rubund:
I wasn't aware of the SV capabilities of iverilog. It seems like it now handles both packages and interfaces well with the -g2012 option! The Debian package is lagging behind though, but I hope it will be updated soon (https://bugs.debian.org/826615).

Does it mean that if we want to add some more SV functionality to yosys which is not yet in iverilog, we should first send a patch with the support to iverilog?

cliffordwolf:
> [..] we should first send a patch with the support to iverilog?

Not necessarily. But it will be much harder to test stuff that isn't supported by iverilog yet..

PR #181, opened 2016-06-20
Allow defining input ports as "input logic" in SystemVerilog

rubund:
Hi Clifford,

A lot of SV code uses the following syntax in the module port definition list:

   input logic signal1,
   etc.

I removed the check that makes read_verilog fail when it encounters this for SV. Do you think this has any negative implications? I have not completely been able to grasp how you treat wire and reg differently in yosys, and what implications this will have for "logic".

Cheers,
Ruben


cliffordwolf:
Generally Yosys does not care if something is declared as wire or reg. (There are a few exceptions.) Since your patch only disables an error message in sv_mode there is no way that it can break any cases that work now, and I care much more about processing valid Verilog code than good error reporting for invalid code..

PR #184, opened 2016-07-01
Added 'assert-limit' option for 'select' command

eshellko:
For resource limited designs such as FPGA it can be useful to specify limit of specific resources available on board. So user can check if he should change RTL as early as mapping done.


cliffordwolf:
Good idea. I've replaced `-assert-limit` with `-assert-max` and `-assert-min` in comit 6ed6b3cb6d1f1735201861d30cd70736b76e5221.

PR #201, opened 2016-08-15
Added graphviz dot backend in order to be able to use the show command from programs...

forflo:
... linking to the yosys library


forflo:
FYI: I verified it to be working

cliffordwolf:
1) This contains a verbatim copy of ShowWorker from show.cc. Clearly this is not the way to maintain a clean and maintainable codebase.

2) Why not simply run `show -prefix schematic -format dot` to write the current design to `schematic.dot`?

forflo:
I'm not ready to include Yodl as command for Yosys. I need the ShowWorker functionality available in libyosys. This is because Yodl is based on sources that are licensed under the GPL.

Regarding your first point: Of course, it's a verbatim copy. I didn't change any of the existing code. This PR was intended as a refactoring suggestion, because IMHO the ShowWorker should clearly be exposed as backend. Could you please reconsider your decision?

I can, of course, also provide for the appropriate simplifications of the show command. Since ShowWorker is not needed there anymore. I thought this would be trivial though, because this step simply requires you to delete the ShowWorker definition from show.cc.

cliffordwolf:
You did not address my second point: Why not simply use the existing functionality of the `show` command to write a `.dot` file?

forflo:
What do you mean by `use the [ ... ] show command`. Is the functionality already in the shared lib? From your use of `PRIVATE_NAMESPACE_BEGIN` at the beginning of `show.cc` I concluded that it's not intended to be used that way. Moreover, there are no comments whatsoever regarding this issue...

The only other option I can think of is to dump the rtlil code with yosys's ilang backend, manually start another yosys process with your suggested command as input in order to convert the ilang file into dot code. I don't know if I'm interpreting your suggestion right, but if I did: Sorry, I'm certainly not going to do this. I dislike temporary files with a passion and moreover, the show command generates a *hidden* file in my home directory! I don't want my testing to litter my home directory.

cliffordwolf:
Simpy run e.g. `Yosys::Pass::call(design, "show -prefix schematic -format dot");` from your C++ code with the design you want to dump.

forflo:
Ok, that's nicer. Thanks

PR #216, opened 2016-08-28
cmd 'show', help, add info to generate 'dot' file

nononame:
I discover [here](https://www.reddit.com/r/yosys/comments/2w9uvn/newbie_howto_yosys_on_windows/#thing_t1_cop8q1p) that there is a proper way to generate dot graph description instead of image (the graph->image process can take a long time). Before found this I used `-viewer echo` but the right way is `-format dot`. I propose to add this information in the help message of the `show` command

### DIFF
--- a/passes/cmds/show.cc
+++ b/passes/cmds/show.cc
@@ -584,6 +584,7 @@ struct ShowPass : public Pass {
 		log("    -format <format>\n");
 		log("        Generate a graphics file in the specified format.\n");
 		log("        Usually <format> is 'svg' or 'ps'.\n");
+		log("        To generate graph description instead of image, 'dot' can be used.\n");
 		log("\n");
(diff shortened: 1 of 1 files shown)


cliffordwolf:
tbh I find the change you proposed to make the help message more confusing, not less. Also, there is already the following paragraph at the end of the "show" help message, that already explains that 'dot' is a valid option for -format, and how -format, -viewer, and -prefix interact.

> When no \<format\> is specified, 'dot' is used. When no \<format\> and \<viewer\> is
specified, 'xdot' is used to display the schematic.

> The generated output files are '~/.yosys_show.dot' and '~/.yosys_show.\<format\>',
unless another prefix is specified using -prefix \<prefix\>.

I have now improved the "show" help message a bit in commit 66582964bc11aadf3d0783a346706d801451a13f.

nononame:
perfect.

PR #223, opened 2016-09-02
Let class SmtModInfo take responsibility of its functionality.

Kmanfi:


backends/smt2/smtbmc.py
```diff
@@ -326,14 +326,8 @@ def write_vlogtb_trace(steps_start, steps_stop, index):
         print("  reg [4095:0] vcdfile;", file=f)
         print("  reg clock = 0, genclock = 1;", file=f)
 
-        primary_inputs = list()
-        clock_inputs = set()
-
-        for name in smt.modinfo[topmod].inputs:
-            if name in ["clk", "clock", "CLK", "CLOCK"]:
```
cliffordwolf: This list, `clk clock CLK CLOCK` is a feature (some might call it hack :) of smtbmc.py, other driver programs might want to use a different list, or use `smtio.py` in a different way altogether.

backends/smt2/smtio.py
```diff
@@ -45,6 +45,46 @@ def __init__(self):
...
+        for name in self.inputs:
+            yield (name, self.wsize[name])
+
+    def clock_inputs(self):
+        for name in self.inputs:
+            if name in ["clk", "clock", "CLK", "CLOCK"]:
+                yield name
+
+    def update(self, stmt):
+        fields = stmt.split()
```
cliffordwolf: The intent behind SmtInfo is that it is a "struct", not a "class". SmtIo does all the SMT-LIB stuff and parsing of yosys-smt2- comments. Imo moving this code from SmtIo to SmtInfo just fractures the code without adding anything to it. It's not like you could use SmtInfo stand-alone without SmtInfo and do anything useful with it.

If anything, SmtInfo should actually be something like a collections.namedtuple. But I don't think that this would actually make the code any cleaner, it would just prevent the next guy from adding methods to it. :)

Kmanfi:
For me it looked like a struct. I found myself quite often polluting my structs too. Let's keep struct as structs ;)

PR #231, opened 2016-09-15
Add optional SEED=n command line option to Makefile, and -s n command…

brouhaha:
… line option to test scripts, for deterministic regression tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -404,16 +404,22 @@ endif
 yosys-abc$(EXE): abc/abc-$(ABCREV)$(EXE)
 	$(P) cp abc/abc-$(ABCREV)$(EXE) yosys-abc$(EXE)
 
+ifneq ($(SEED),)
+SEEDOPT="-S $(SEED)"
+else
+SEEDOPT=""
+endif
+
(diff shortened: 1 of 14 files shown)


cliffordwolf:
Oh, nice! That was quick.

The autotest.sh script (tests/tools/autotest.sh) also has a -S option for a rng seed. (It is however not used for testcase generation but for test pattern generation.) For "make test" this script is run via tests/tools/autotest.mk, see for example tests/simple/run-test.sh. I think it might be a good idea to also include that in "make SEED=1234 test". Would you like to also add support for that?

I think if the autotest.sh seed is also managed by the top-level makefile, then two successive runs of "make SEED=1234 test" should generate exactly the same files in the tests/ directory.

brouhaha:
Yes, I'll add that. I probably won't get around to it for a week or more, though.  Perhaps I should change the -s option I added to the other scripts to -S for consistency.

brouhaha:
The pull request has been updated to pass the seed to autotest, and to use "-S" option rather than "-s".

PR #244, opened 2016-10-13
Added  a new configuration variable GIT_REV_WHERE

set-soft:
It determines from where we get the gits SHA1 value. By default is HEAD,
suitable for Clifford, but for Debian we can define it as upstream/master

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -411,7 +411,7 @@ else
 SEEDOPT=""
 endif
 
-test-all: $(TARGETS) $(EXTRA_TARGETS)
+test: $(TARGETS) $(EXTRA_TARGETS)
 	+cd tests/simple && bash run-test.sh $(SEEDOPT)
 	+cd tests/hana && bash run-test.sh $(SEEDOPT)
 	+cd tests/asicworld && bash run-test.sh $(SEEDOPT)
@@ -558,6 +558,6 @@ echo-git-rev:
 -include kernel/*.d
 -include techlibs/*/*.d
 
-.PHONY: all top-all abc test-all install install-abc manual clean mrproper qtcreator
+.PHONY: all top-all abc test install install-abc manual clean mrproper qtcreator
 .PHONY: config-clean config-clang config-gcc config-gcc-4.8 config-gprof config-sudo
 
--- a/Makefile
+++ b/Makefile
@@ -72,8 +72,9 @@ else
 	LDLIBS += -lrt
 endif
 
+GIT_REV_WHERE ?= HEAD
 YOSYS_VER := 0.6+$(shell test -e .git && { git log --author=clifford@clifford.at --oneline 5869d26da021.. | wc -l; })
-GIT_REV := $(shell cd $(YOSYS_SRC) && git rev-parse --short HEAD 2> /dev/null || echo UNKNOWN)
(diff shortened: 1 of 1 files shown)


cliffordwolf:
I'm not sure if I understand the need for this correctly. If `HEAD` is different from `upstream/master`, why would you use the git rev from `upstream/master`? The whole idea of having a git rev in the version string in the first place is to be exact about which version of the source code was used.

If a person has a Yosys build from the debian sources, and another one from the upstream sources, I would expect them to report different git sha1 checksums.

set-soft:
The Debian packages are named VERSIONgitSHA1-PACKAGE_VERSION
Here the SHA1 is the SHA1 of the original project, not the SHA1 of the Debian repository.
The PACKAGE_VERSION indicates changes in the packaging stuff, but the SHA1 remains the same.
This helps people to reproduce the package without the need to access the Debian repository. A user can just checkout from GitHub using the indicated SHA1 and then add the packaging files (available from the Debian packages interface).
Ideally the Debian package stuff won't need to touch the code, but could need to apply some patches. And I think this is what confuses you: a binary could claim to be created using certain SHA1, but in fact have additional patches. This is normal in Debian, the patches are available. The SHA1 reported by the binary should help to know which code was used as base for the package. That's why the SHA1 must refer to your repository at GitHub, not a SHA1 from Debian's repo.
I hope my explanation is clear, english isn't my native language ;-)

brouhaha:
I have pretty much the same issue with the Fedora yosys package. I chose to leave the git sha1 reported by the program itself as UNKNOWN. If someone has the package installed, and needs to know what upstream git snapshot it was built from, they can use "rpm -q yosys", as the date and first seven hex digits of the sha1 hash of the git snapshot are in the package release.

Given that the Fedora packaging may patch the source (presently only patching yosys-config), I thought it could be misleading to have the program display its version as the upstream hash, since it might not exactly match. The user can get the SRPM (source RPM) corresponding to the binary RPM they have installed, and the SRPM contains the exact upstream snapshot and all patches used to build the binary RPM.

If I were going to change the yosys version message specifically for Fedora, I think I'd want it to contain the Fedora RPM name (e.g., "yosys-0.6.0-1.20160923git8f5bf6d"), which provides traceability to the Fedora build system.

cliffordwolf:
I would suggest setting `GIT_REV` directly via `make` command line in this cases. I have now also added a `YOSYS_VER_STR` Makefile variable that can be used to set the entire version string manually.

set-soft:
I can use GIT_REV, but in this case: Can you change := by ?= is easier to define it from the environment. In this way one definition affects all the calls to make.
What I'm trying to avoid is a nasty patch that currently patches the Makefile to define GIT_REV with a hardcoded value.

cliffordwolf:
> Can you change := by ?=

No. `?=` has the semantics of `=`, not `:=`, so the entire shell command would be re-evaluated whenever the variable is used!

Can you explain why it is so difficult to simply pass `GIT_REV="..."` as argument to `make` for this? Why does it need to be defined via the environment?

set-soft:
Is just easier to export the variable, instead of catching every call and overwriting it. But I can do it.

PR #245, opened 2016-10-13
Modified test target name (to test-all)

set-soft:
As this target depends on external tools, and packagers run "make test",
I think the name should be less generic.


cliffordwolf:
Packers do run `make test`. The external tools are build dependencies. Tests that aren't suitable in this context are not included in `make test` (i.e. `make vgtest` and `make vloghtb`).

PR #251, opened 2016-10-20
Preparations for wand and wor

smoe:
Hello,

A few bits are missing, i.e. code generation in particular and I was not certain about what the json representation or wand/wor should look alike, but the parsers should be fine and also the ilang and verilog backends. Also I thought that you may possibly accept the extra netType parameter for the wire class to distinguish the three types.

Of course I do not expect this patch to be accepted at this stage of development and just before a release. But you could possibly give me a quick feedback with an estimate if the way I was approaching the implementation is acceptable for you in principle.

I learned that wand and wor wire-types are not fashionable among professionals. Somehow I had immediately stumbled across them and use them to assign a variable number of wires (~ 30) to a single wor in a generate loop. I would not know how to perform that auto-generation with the | operator, so here I go and hope to be of some help.

Many thanks

Steffen


cliffordwolf:
You are not going to like this: I really don't think this is the best way to do that! By adding wor/wand to the wire type you'd have to modify virtually everything that is using RTLIL, starting by patching each and every single user of SigMap, because now you suddenly have a situation where a wire could both be driven by 0 and/or 1 and/or a non-const signal, and depending on the "wire type" this would have one or the other meaning.

Instead, the Verilog front-end should convert a `wor/wand` into a dedicated cell that is essentially a `$reduce_or` or `$reduce_and` cell (for a single bit wire). Certainly the way this is implemented must not require massive changes to the rest of the framework to make sure a single `wor/wand` in the design does not suddenly break everything!

Start by adding `$wor` and `$wand` cell types with something like the following simlib model:

```
module \$wor (A, Y);

parameter WIDTH = 0;
parameter DEPTH = 0;

input [WIDTH*DEPTH-1:0] A;
output reg [WIDTH-1:0] Y;

integer i;

always @* begin
        Y = 0;
        for (i = 0; i < DEPTH; i = i+1)
                Y = Y | A[WIDTH*i +: WIDTH];
end

endmodule
```

(And similar for `$wand`. See `CodingReadme` for how to add cell types.)

PR #274, opened 2016-11-22
Added optional flag for linking curses with readline.

oldtopman:
On some Linux systems, -lcurses is required in addition to -lreadline. LINK_CURSES adds this (optionally) both to yosys and abc. This is a problem that has been recognized by ABC as well on their [bitbucket page](https://bitbucket.org/alanmi/abc).

I'm not sure if this is the best way to go about adding it, but I think it's a good start.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,7 @@ ENABLE_LIBYOSYS := 0
 # other configuration flags
 ENABLE_GPROF := 0
 ENABLE_NDEBUG := 0
+LINK_CURSES := 0
 
 # clang sanitizers
 SANITIZER =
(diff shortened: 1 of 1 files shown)


cliffordwolf:
This is a bug in the distributions readline package and should be fixed there. Each .so should contain its own runtime linker dependencies.

> I'm not sure if this is the best way to go about adding it, but I think it's a good start.

It should be fixed by correcting the distributions readline package, not patching every user of it.

oldtopman:
According to [this](https://bugzilla.redhat.com/show_bug.cgi?id=499837) RedHat post, the reason readline isn't linked against curses is not because it's a misconfigured package, but because readline can be linked with either termcap or curses at compile-time. Picking one in the .so would break binaries that require the other backend. Essentially, it's not a bug, it's a feature. :)

Regardless, does this mean that distributions with 'bugged' readline packages are completely unsupported by this project? I would really like to have support for this added, or at the very least, have some information about this known issue added to yosys.

The reason I chose to write this into the makefile is because my current build method is:

1. make config-gcc,
2. make,
3. edit the abc makefile to add lcurses,
4. edit the yosys makefile to add lcurses,
5. edit the yosys makefile to change ABCREV=default,
6. make,

which is repeated on every clean build, when it should be:

1. make config-gcc
2. make

Would it be possible to expose variables that could be set from the environment that would add lcurses to both abc and yosys builds?

cliffordwolf:
> because readline can be linked with either termcap or curses at compile-time [..], it's not a bug, it's a feature. :)

Interesting. Thanks for pointing that out to me.

PR #281, opened 2016-12-04
Unit test structure

rodrigosiqueira:
In this Pull Request we concentrate to add the basic structure for unit tests. We tried to make it without insert a lot of extra code in the main Makefile. Finally, we tried to make the processes of adding new unit test easy.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.o
 *.d
 .*.swp
+*.gch
 /.cproject
 /.project
 /.settings
@@ -27,3 +28,5 @@
 /yosys-win32-vcxsrc-*
 /yosysjs-*
 /libyosys.so
+/tests/unit/bintest/
+/tests/unit/objtest/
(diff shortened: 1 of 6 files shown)


cliffordwolf:
Request for some changes:

- At least the version of gtest that I am using here requires the tests to be built with `-pthreads`
- I would prefer if `make unit-test` would also run the tests. I don't think a separate target `make run-all-unitest` is necessary.
- The dependencies in `tests/unit/Makefile` are all messed up. The target for `%Test.o` builds a .o file with a different name (other directory), and also the binary. If you want to run both commands (compile and link) in one make rule, make sure the make rule is actually for the final output product. Running `make unit-test` a second time should not rebuild the tests unless necessary. (A dependency on `../../libyosys.so` might help making sure that the test are rebuilt if yosys sources have changed.)

cliffordwolf:
Merged with minor changes (see 71c47f1 and 5c96982).

PR #290, opened 2017-01-10
Allow overriding ABC's revision and mode (C/C++)

udif:
As of ABC commit 63268717e48b, my changes for allowing ABC to be compiled in C mode were accepted.
the following commit allows external override of the C/C++ compile mode selection.
You can either accept this (perhaps updating the ABC commit ID further), decide to adopt the C compile mode, or keep it in.
My hope was that compiling the code as C would improve runtime and/or size, but I have no concrete data to back this yet.
In any case, I think that allowing external override over ABCREV is useful, even if ABCCPP is not accepted.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,7 +86,8 @@ OBJS = kernel/version_$(GIT_REV).o
 # is just a symlink to your actual ABC working directory, as 'make mrproper'
 # will remove the 'abc' directory and you do not want to accidentally
 # delete your work on ABC..
-ABCREV = 38b26a543f1d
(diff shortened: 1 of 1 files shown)


udif:
Please see https://github.com/udif/cae-win-tools
I have forked 3 repositories only due to Win32 cross compile issues (icestorm, yosis, icarus), and if this pull request is accepted I could remove the yosys fork.

cliffordwolf:
Updated ABC and simply removed the now unnecessary `-x c++` option in 8953a55.

Issue #269, opened 2016-11-14
a < 0 can be optimized if a is signed


AlexDaniel:
Here is the shortest module to reproduce the issue:
```verilog
module isneg (input signed [7:0] a, output logic o);
   assign o = a < 0;
   //assign o = a[7];
endmodule
```

Given that signed numbers are in two's complement, you can use MSB to see if the number is negative or not (``assign o = a[7]``). Synthesis result in this case should be obvious.

However, for ``a < 0`` yosys produces this:
![“show” diagram after synthesis](https://files.progarm.org/isneg.svg)
(complete command to reproduce exactly the same graph is ``yosys -p 'synth_xilinx; show -colors 67 -format svg -prefix isneg' isneg.sv``)

Sure, I can always optimize it manually, but I find it less readable this way.

cliffordwolf:
I'll add a proper peephole optimizer with syntax for a rules files and a code generator that takes rules files and generates the optimizer pass, in addition to or as a replacement for opt_expr. I'll make sure to add this optimization as one of the first rules when this feature becomes available.

I'll update this issue when there is any progress. I hope I will get to it this year, but there is not much year left..

cliffordwolf:
This optimization has been implemented by #292, so I'm now closing this issue.
(The generic poophole optimizer stays on my todo list. Not sure when I will get to it though.)

AlexDaniel:
Great! Thanks.

PR #292, opened 2017-01-15
Fix issue #269, optimize signed and unsigned comparisons

C-Elegans:
add opt_compare pass and add it to opt
for a < 0:
    if a is signed, replace with a[max_bit-1]
for a >= 0:
    if a is signed, replace with ~a[max_bit-1]


cliffordwolf:
As said in #269, the plan is to add a generic pass for peephole optimizations.

If you really want a hand-written peephole optimizer, then it should be merged with `opt_expr`, where all the other hand-written peephole optimizers are.

Also: Just comparison of signed signals with zero is imo a too special case. For example consider the following cases (assuming `a` is declared as `wire [7:0] a`):

`a >= 16` is equivalent to `|a[7:4]`.

`a < 16` is equivalent to `!a[7:4]`

`a < 48` is equivalent to `a[7:4] < 3`

I think `opt_expr -fine` would be the right place to add all this kind of stuff.

C-Elegans:
I added it to opt_expr -fine and added your suggestions, but I wasn't sure whether to squash the commits together.

cliffordwolf:
There is no normalization in Yosys to `$lt` and `$ge`. There is also a `$gt` and `$le`, and if one port is constant, there is no guarantee that the constant port is `B`.

Also: you must only convert an `RTLIL::Const` to an `int` with `RTLIL::Const::as_int()` if you know that an `int` is actually large enough to hold the value in the Const.

cliffordwolf:
Your code still has assumptions about which port is the constant one. Consider the following example:

```
module test(input [7:0] A, output [7:0] Y);
  assign Y = {
    A <  8,
    A <= 8,
    A >= 8,
    A >  8,
    8 <  A,
    8 <= A,
    8 >= A,
    8 >  A
  };
endmodule
```

Only 4 of the 8 comparison operators are transformed by your version of `opt_expr`:

image

(Output of `yosys -p 'opt_expr -fine; show' test.v`.)

C-Elegans:
Right. The optimizations discussed above only work on < and >= when x is on the left. Let's take the example of x >= 8 vs x > 8 when x is 8 bits long.
x >= 8 -> min(x) = 0b00001000, so bits [7:3] can be ORed together to replace a comparison
x > 8   -> min(x) = 0b00001001, so it cannot be replaced with a simple or. Instead, it needs a special case for x == 8, likely requiring a full equals comparison of x. This would considerably complicate the optimization, which is why I didn't include it.

cliffordwolf:
merged. (with many indenting fixes in commit ffbe8d4 :)

C-Elegans:
Oh shoot, I forgot to check for that.

PR #296, opened 2017-01-17
Use cp -p in Makefile to preserve last modified date

C-Elegans:
Replaces all invocations of `cp` with `cp -p` to preserve the last modified date when installing the yosys header files. This means that after `make install` a subsequent call to `make` will only compile the files that have been modified, instead of a full recompile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -252,14 +252,14 @@ define add_share_file
 EXTRA_TARGETS += $(subst //,/,$(1)/$(notdir $(2)))
 $(subst //,/,$(1)/$(notdir $(2))): $(2)
 	$$(P) mkdir -p $(1)
-	$$(Q) cp "$(YOSYS_SRC)"/$(2) $(subst //,/,$(1)/$(notdir $(2)))
+	$$(Q) cp -p "$(YOSYS_SRC)"/$(2) $(subst //,/,$(1)/$(notdir $(2)))
 endef
 
 define add_gen_share_file
 EXTRA_TARGETS += $(subst //,/,$(1)/$(notdir $(2)))
 $(subst //,/,$(1)/$(notdir $(2))): $(2)
 	$$(P) mkdir -p $(1)
-	$$(Q) cp $(2) $(subst //,/,$(1)/$(notdir $(2)))
+	$$(Q) cp -p $(2) $(subst //,/,$(1)/$(notdir $(2)))
 endef
 
 define add_include_file
@@ -462,7 +462,7 @@ install: $(TARGETS) $(EXTRA_TARGETS)
 	$(INSTALL_SUDO) mkdir -p $(DESTDIR)$(BINDIR)
 	$(INSTALL_SUDO) install $(TARGETS) $(DESTDIR)$(BINDIR)
 	$(INSTALL_SUDO) mkdir -p $(DESTDIR)$(DATDIR)
-	$(INSTALL_SUDO) cp -r share/. $(DESTDIR)$(DATDIR)/.
+	$(INSTALL_SUDO) cp -p -r share/. $(DESTDIR)$(DATDIR)/.
 ifeq ($(ENABLE_LIBYOSYS),1)
 	$(INSTALL_SUDO) cp libyosys.so $(DESTDIR)$(LIBDIR)
 	$(INSTALL_SUDO) ldconfig


cliffordwolf:
This would be problematic in situations where one has two different yosys source directories and uses `make install` in one of them to install the files over the previously installed files from the other directory in an alternating pattern. Something like this is not uncommon during development. In this situation you always want to make sure that every file that has been changed by `make install` also has the current time as modification time.

A possible solution would be to e.g. use `rsync -c` to install. This will update the timestamp on the target if the target is different from the source, and leave the target untouched otherwise. But I'm hesitant to add a dependency to rsync for this.

> This means that after make install a subsequent call to make will only compile the files that have been modified, instead of a full recompile.

I don't really understand what you mean here. Calling `make install && make` will not trigger a full recompile:

```
clifford@euklid:~/Work/yosys$ make install
[Makefile.conf] INSTALL_SUDO := sudo
sudo mkdir -p /usr/local/bin
sudo install yosys yosys-config yosys-abc yosys-filterlib yosys-smtbmc /usr/local/bin
sudo mkdir -p /usr/local/share/yosys
sudo cp -r share/. /usr/local/share/yosys/.

clifford@euklid:~/Work/yosys$ make
[Makefile.conf] INSTALL_SUDO := sudo

  Build successful.

clifford@euklid:~/Work/yosys$ 
```

Are you talking about running `make` somewhere else? Like a Yosys plugin?

C-Elegans:
It triggers a full recompile on my system because nearly every source file depends on the headers installed in /usr/local/share/yosys

cliffordwolf:
> It triggers a full recompile on my system because nearly every source file depends on the headers installed in /usr/local/share/yosys

None of the yosys source files depend on anything in /usr/local/share/yosys. Can you give a concrete example?

C-Elegans:
Sure. After running 
`cp share/include/kernel/yosys.h /usr/local/share/yosys/include/kernel/yosys.h`
make --debug outputs this:
```
Reading makefiles...
Updating goal targets....
 File `all' does not exist.
   File `top-all' does not exist.
       Prerequisite `/usr/local/share/yosys/include/kernel/yosys.h' is newer than target `libs/subcircuit/subcircuit.o'.
      Must remake target `libs/subcircuit/subcircuit.o'.
[  8%] Building libs/subcircuit/subcircuit.o
      Successfully remade target file `libs/subcircuit/subcircuit.o'.
       Prerequisite `/usr/local/share/yosys/include/kernel/yosys.h' is newer than target `frontends/ast/ast.o'.
      Must remake target `frontends/ast/ast.o'.
[ 12%] Building frontends/ast/ast.o
      Successfully remade target file `frontends/ast/ast.o'.
       Prerequisite `/usr/local/share/yosys/include/kernel/yosys.h' is newer than target `frontends/ast/genrtlil.o'.
      Must remake target `frontends/ast/genrtlil.o'.
```

cliffordwolf:
Can you tell where this comes from? E.g. why does `libs/subcircuit/subcircuit.o` depend on `/usr/local/share/yosys/include/kernel/yosys.h` on your machine? Is it listed in `libs/subcircuit/subcircuit.d`? Why? Does your `libs/subcircuit/subcircuit.d` list the correct dependency to `kernel/yosys.h`?

E.g. what output do you get for this grep command:

```
clifford@euklid:~/Work/yosys$ grep yosys.h libs/subcircuit/subcircuit.d 
  /usr/include/assert.h kernel/yosys.h \
```

What command are you using to build `subcircuit.o`? I.e. what output do you get for this:

```
clifford@euklid:~/Work/yosys$ rm -f libs/subcircuit/subcircuit.o; make PRETTY=0 libs/subcircuit/subcircuit.o
[Makefile.conf] INSTALL_SUDO := sudo
mkdir -p libs/subcircuit/
clang -o libs/subcircuit/subcircuit.o -c  -Wall -Wextra -ggdb -I. -I"./" -MD -D_YOSYS_ -fPIC -I/usr/local/include -std=c++11 -Os -DYOSYS_ENABLE_READLINE -DYOSYS_ENABLE_PLUGINS  -I/usr/include/tcl8.6 -DYOSYS_ENABLE_TCL -DYOSYS_ENABLE_ABC -DYOSYS_ENABLE_COVER libs/subcircuit/subcircuit.cc
```

Last question for this post: Do you have any bogus symlinks from `/usr/include/...` to `/usr/local/share/yosys/include/...`?

C-Elegans:
`grep yosys.h libs/subcircuit/subcircuit.d` outputs
  `/usr/include/sys/_types/_mode_t.h kernel/yosys.h \`
and 
`make PRETTY=0 libs/subcircuit/subcircuit.o` outputs
```
mkdir -p libs/subcircuit/
clang -o libs/subcircuit/subcircuit.o -c  -Wall -Wextra -ggdb -I. -I"./" -MD -D_YOSYS_ -fPIC -I/usr/local/include -I/opt/local/include -I/usr/local/opt/readline/include -I/usr/local/Cellar/libffi/3.0.13/lib/libffi-3.0.13/include -std=c++11 -Os -DYOSYS_ENABLE_READLINE -DYOSYS_ENABLE_PLUGINS  -I/usr/include/tcl8.5 -DYOSYS_ENABLE_TCL -DYOSYS_ENABLE_ABC -DYOSYS_ENABLE_COVER libs/subcircuit/subcircuit.cc
```
(I am on a mac if that helps)

cliffordwolf:
> grep yosys.h libs/subcircuit/subcircuit.d outputs
/usr/include/sys/_types/_mode_t.h kernel/yosys.h \

Interesting. If `libs/subcircuit/subcircuit.d` does not list `/usr/local/share/yosys/include/kernel/yosys.h`, why does `subcircuit.o` depend on it? (I know, this are hard questions. But since I cannot reproduce the issue here, asking hard questions is pretty much the only thing I can do to figure out whats going on.)

> (I am on a mac if that helps)

I don't have a mac and never had one. So if there is some mac-specific thing that would be useful to know about to understand what's going on here: I would not know about it.. :(

C-Elegans:
I just updated my make to 4.2.1 and it still does it. 
Oh well, if this isn't a problem for anybody else, I'll just keep it on my local copy.

PR #304, opened 2017-02-01
Use gsed vs sed on Darwin.

esden:
BSD sed does not support -r parameter.

This is a quick and dirty solution. Better suggestions welcome.


joshhead:
BSD and GNU sed both support -E, which from the GNU sed man page seems to be equivalent to -r. Changing r to E might be a simpler fix.

esden:
@joshhead Yes of course you are right. I have changed the patch accordingly. I have tested it on Mac OS and on Linux.

PR #313, opened 2017-02-14
More progress on Firrtl backend.

azidar:
Chisel -> Firrtl -> Verilog -> Firrtl -> Verilog is successful for a
simple rocket-chip design.

### DIFF
--- a/backends/firrtl/test.sh
+++ b/backends/firrtl/test.sh
@@ -1,16 +1,20 @@
 #!/bin/bash
 set -ex
 
-../../yosys -p 'prep -nordff; write_firrtl test.fir' test.v
+cd ../../
+make
+cd backends/firrtl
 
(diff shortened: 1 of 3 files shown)


backends/firrtl/test.sh
```diff
@@ -1,22 +1,25 @@
 #!/bin/bash
 set -ex
 
-../../yosys -p 'prep -nordff; write_firrtl test.fir' test.v
+cd ../../
+make
+cd backends/firrtl
```
cliffordwolf: Maybe `make -C ../..` would be better? But tbh I personally am not a big fan of this either way. I'd prefer just using something like `make -C ../.. -j$(nproc) && bash test.sh` when calling it, iff I want to re-run make before running the test. But you have to use this script far more frequently than I, so do whatever you like :), I just could not resist to comment..

PR #318, opened 2017-02-23
Fix ambiguous conversion

Jesus89:
Using the *x86_64-w64-mingw32-gcc* compiler, Yosys build crashes:

> kernel/calc.cc: In function ‘Yosys::RTLIL::Const Yosys::const_shift_worker(const Yosys::RTLIL::Const&, const Yosys::RTLIL::Const&, bool, int, int)’:
kernel/calc.cc:294:34: error: conversion from ‘std::vector<Yosys::RTLIL::State>::size_type {aka long long unsigned int}’ to ‘const BigInteger’ is ambiguous
   else if (pos >= arg1.bits.size())

> kernel/calc.cc: In function ‘Yosys::RTLIL::Const Yosys::const_shift_shiftx(const Yosys::RTLIL::Const&, const Yosys::RTLIL::Const&, bool, bool, int, Yosys::RTLIL::State)’:
kernel/calc.cc:345:40: error: conversion from ‘std::vector<Yosys::RTLIL::State>::size_type {aka long long unsigned int}’ to ‘const BigInteger’ is ambiguous
   if (pos < 0 || pos >= arg1.bits.size())

This pull request fixes the ambiguity of conversion.

### DIFF
--- a/kernel/calc.cc
+++ b/kernel/calc.cc
@@ -291,7 +291,7 @@ static RTLIL::Const const_shift_worker(const RTLIL::Const &arg1, const RTLIL::Co
 		BigInteger pos = BigInteger(i) + offset;
 		if (pos < 0)
 			result.bits[i] = RTLIL::State::S0;
-		else if (pos >= arg1.bits.size())
+		else if (pos.toUnsignedLong() >= arg1.bits.size())
 			result.bits[i] = sign_ext ? arg1.bits.back() : RTLIL::State::S0;
 		else
 			result.bits[i] = arg1.bits[pos.toInt()];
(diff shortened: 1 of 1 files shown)


cliffordwolf:
This introduces a bug (pos might very well be a huge number, i.e. > 2**64, and only has its MSB set).

Proper fix in commit 0822b21, but I can't test it. Please let me know if this fixes the issue.

Jesus89:
I tested your solution before, but it throws these errors:

> kernel/calc.cc: In function ‘Yosys::RTLIL::Const Yosys::const_shift_worker(const Yosys::RTLIL::Const&, const Yosys::RTLIL::Const&, bool, int, int)’:
kernel/calc.cc:294:46: error: call of overloaded ‘BigInteger(std::vector<Yosys::RTLIL::State>::size_type)’ is ambiguous
   else if (pos >= BigInteger(arg1.bits.size()))

> kernel/calc.cc: In function ‘Yosys::RTLIL::Const Yosys::const_shift_shiftx(const Yosys::RTLIL::Const&, const Yosys::RTLIL::Const&, bool, bool, int, Yosys::RTLIL::State)’:
kernel/calc.cc:345:52: error: call of overloaded ‘BigInteger(std::vector<Yosys::RTLIL::State>::size_type)’ is ambiguous
   if (pos < 0 || pos >= BigInteger(arg1.bits.size()))

cliffordwolf:
Hmm.. Does c6d8d70 fix it?

Jesus89:
Yes. Fixed.

PR #328, opened 2017-02-25
change bit order

klammerj:
things were mildly reversed...

### DIFF
--- a/backends/edif/edif.cc
+++ b/backends/edif/edif.cc
@@ -364,7 +364,7 @@ struct EdifBackend : public Backend {
 							auto m = design->module(cell->type);
 							if (m) {
 								auto w = m->wire(p.first);
-								if (w && !w->upto)
+								if (w)
 									member_idx = GetSize(w)-i-1;
 							}
 							net_join_db[sig[i]].insert(stringf("(portRef (member %s %d) (instanceRef %s))",
--- a/backends/edif/edif.cc
+++ b/backends/edif/edif.cc
@@ -317,10 +317,7 @@ struct EdifBackend : public Backend {
 					*f << stringf("          (port (array %s %d) (direction %s))\n", EDIF_DEFR(wire->name, port_rename, b[0], b[1]), wire->width, dir);
 					for (int i = 0; i < wire->width; i++) {
 						RTLIL::SigSpec sig = sigmap(RTLIL::SigSpec(wire, i));
-						if (wire->upto)
-							net_join_db[sig].insert(stringf("(portRef (member %s %d))", EDIF_REF(wire->name), i));
-						else
-							net_join_db[sig].insert(stringf("(portRef (member %s %d))", EDIF_REF(wire->name), GetSize(wire)-i-1));
+						net_join_db[sig].insert(stringf("(portRef (member %s %d))", EDIF_REF(wire->name), GetSize(wire)-i-1));
 					}
 				}
 			}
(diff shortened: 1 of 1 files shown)


cliffordwolf:
Can you post a mvce and point out the error in the output?

Using the following example it looks all correct to me:

```
module top(output [1:0] downto, output [0:1] upto);
  sub sub(downto, upto);
endmodule

module sub(output [2:0] downto, output [0:2] upto);
  assign downto = 1, upto = 1;
endmodule
```

(Run with e.g. `yosys -p 'prep; write_edif -pvector par' test.v`.)

klammerj:
Your mcve will do just fine.
On my side of the fork I'm getting the following connections in cell top:
``
            (portRef (member upto 0) (instanceRef sub))
            (portRef (member upto 0))
``
sub.upto(0) <-> top.upto(0)
``
            (portRef (member upto 1) (instanceRef sub))
            (portRef (member upto 1))
``
sub.upto(1) <-> top.upto(1)
``
            (portRef (member downto 0))
            (portRef (member downto 1) (instanceRef sub))
``
sub.downto(1) <-> top.downto(1)
``
            (portRef (member downto 1))
            (portRef (member downto 2) (instanceRef sub))
``
sub.downto(0) top.downto(0)

I've noted down the way IDS interprets those below the quotes. 
That is: the member 0 resolves to the leftmost index of the vector. 
1 and 2 for downto 0 and 0 for upto(in top and sub respectively)

Two wires end up dangling...

What do you get? 

Also, if I change the test as follows:
``
module top(output [1:0] downto, output [0:1] upto);
  sub sub(downto, upto);
endmodule

module sub(output [1:0] downto, output [0:1] upto);
  assign downto[0] = 1, upto[0] = 1;
  assign downto[1] = 0, upto[1] = 0;
endmodule
``

I'm getting wrong results, now.

``
          (net VCC_NET (joined
            (portRef (member downto 1))
            (portRef (member upto 1))
            (portRef P (instanceRef VCC))
          ))
          (net GND_NET (joined
            (portRef (member downto 0))
            (portRef (member upto 0))
            (portRef G (instanceRef GND))
          ))

``
which AFAIK resolves to:
downto(0)=1;
upto(1)=1;
downto(1)=0;
upto(0)=0;

(and that's not what I wrote...)
So something's still fishy..

cliffordwolf:
Did you close the issue on purpose? I'm reopening it..

klammerj:
Yes. Dont merge this(yet).

klammerj:
To get out of the off-by-one territory I tried a different testcase.

``
module top(output [10:0] downto, output [0:10] upto);
  sub sub(upto, downto);
endmodule

module sub(output [10:0] downto, output [0:10] upto);
  assign downto[0:3] = 1;
  assign downto[3:10] = 0;
  assign upto[0:4] = 1;
  assign upto[4:10] = 0;
endmodule
``

klammerj:
This is still broken

klammerj:
Now the output looks ok..
That is, the ports get connected in a bitswapped manner
upto(8)<->downto(2)
and the vcc/gnd connections look ok.
The net names are still somewhat off...

klammerj:
One thing I just noticed, is that if I omit the prep, I'd get wrong port names for the submodules inside the join(). not sure if significant...

cliffordwolf:
Should now be fixed in 0ac72e7. (Commit 850f829 is almost identical to your patch, but without the `b[2] = {0, port_it.second-1}` change and with better handling of cells with no matching blackbox.)

Please see `backends/edif/runtest.py` for a test-case generator (and runner) using Xilinx Vivado to parse EDIF and converting it back to Verilog. Future problems like that should either relate to missing coverage of cases in this test generator, or to tools that handle EDIF differently than Xilinx Vivado. In the former case the test case generator should be extended in addition to fixing the bug, in the latter case an option to `write_edif` must be added to generate both flavors of EDIF.

PR #352, opened 2017-06-26
Initial Coolrunner-II support

rqou:
The PAR functionality isn't done yet, but I wanted to create this initial PR to see if I did it right.

### DIFF
--- a/techlibs/coolrunner2/synth_coolrunner2.cpp
+++ b/techlibs/coolrunner2/synth_coolrunner2.cpp
@@ -145,6 +145,7 @@ struct SynthCoolrunner2Pass : public ScriptPass
 		{
 			run("opt -fast -full");
 			run("techmap");
+			run("techmap -map +/coolrunner2/cells_latch.v");
 			run("dfflibmap -prepare -liberty +/coolrunner2/xc2_dff.lib");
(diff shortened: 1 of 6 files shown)


rqou:
cc @azonenberg

techlibs/coolrunner2/synth_coolrunner2.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+	virtual void script() YS_OVERRIDE
+	{
+		if (check_label("begin"))
+		{
+			run("read_verilog -lib +/coolrunner2/cells_sim.v");
+			run(stringf("hierarchy -check %s", help_mode ? "-top <top>" : top_opt.c_str()));
+		}
+
+		if (flatten && check_label("flatten", "(unless -noflatten)"))
```
cliffordwolf: I would change the order of the two conditions:

    if (check_label("flatten", "(unless -noflatten)") && flatten)

This way `flatten` is still a valid label even if the `-noflatten` option is used.

(Yes, I know that this is borderline nitpicking. :)

cliffordwolf:
Merged with minor changes. See 5c1c126 and 621787a.

PR #364, opened 2017-07-26
Peep hole opt for $logic_not with constant 0 in input A.

Kmanfi:
Initial commit for comments.


cliffordwolf:
1. Can you please rebase this to be one clean commit ontop of my current master?

2. If I read this code correctly you replace a $logic_not cell cell with a constant zero driver if you find any constant 0 bits in the input port. But that's of course bogus! You'd want to set the output to constant zero if you find any constant 1 bits in the input port.

3. This is a PR from your master branch. You are aware that pull requests track the branch they are from? That's why the usual approach is to create a separate branch for your PR, not use master.

Kmanfi:
Close this and make new try.

PR #367, opened 2017-07-31
Add CMake buildsystem

swallat:
I'm not sure if there is any interest in this, but I needed CMake for our development environment. So I transferred the Makefile based system to CMake. Nevertheless support for Emscripten and Verific is still missing as I don't have any experience with both. I haven't modified the test yet, as they currently only support in source builds.

Otherwise it should be complete compared to the Makefiles. I would complete the transition if there is a demand for this, otherwise I would maintain them just for myself.


cliffordwolf:
> I'm not sure if there is any interest in this [..]

I'm actually very happy with the Makefile based build system we have now. So I'm afraid you'll have to maintain the CMake build system in your separate branch.

swallat:
Ok no problem.

PR #379, opened 2017-08-11
Updates to GreenPAK technology library, added some coarse-grained synthesis and optimization passes

azonenberg:


cliffordwolf:
Could we break this down into a few smaller pull requests?

azonenberg:
See #380, #381 . More on the way.

azonenberg:
Replacing with individual feature PRs.

PR #380, opened 2017-08-14
Added opt_rmports pass

azonenberg:
This is a special-use optimization pass (should NOT be called by default when the "opt" command is run) that deletes all top-level ports in the design that have no cells connected to them.

### DIFF
--- a/passes/opt/Makefile.inc
+++ b/passes/opt/Makefile.inc
@@ -6,7 +6,7 @@ OBJS += passes/opt/opt_reduce.o
 OBJS += passes/opt/opt_rmdff.o
 OBJS += passes/opt/opt_clean.o
 OBJS += passes/opt/opt_expr.o
-OBJS += passes/opt/opt_rmports.o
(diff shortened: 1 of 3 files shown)


cliffordwolf:
> should NOT be called by default when the "opt" command is run

I don't even think it is an "optimization". Let's not call it `opt_*` then. Just `rmports` maybe?

Also: `design->top_module()` will return a null pointer if there is no top module. The code does not look like it handles this case correctly.

I also think its a little bit limiting that it only operates on the top module. I think it should operate on all selected module ports. And when it removes a port it should also remove the port reference from all instances of that module.

And finally: Why is ProcessModule() a virtual method?

azonenberg:
Should it still be filed under the opt directory, or should I move it to passes?

azonenberg:
I think this is everything you asked for. Look good now?

cliffordwolf:
Thanks, merged. See 88983f5 for some minor changes I made ontop of your PR.

PR #385, opened 2017-08-14
Added "recover_adder" pass to find adders in gate-level netlists

azonenberg:


cliffordwolf:
First: The capabilities of the `extract` pass are very limited with larger haystack circuits. (But it is extremely good with large needles.) For applications like this we actually should write another `extract` pass that works well with large haystacks and small needles.

But more importantly: This pass looks to me like a pretty large hack. The kind of hack that would be OK in a synthesis script, but should not really be cast into a Yosys command. It should be possible to identify candidates for half and full adders by generating a list of all small combinatorical sub-circuits that have two outputs and two or three inputs, and then simply use ConstEval to check which of them match the lookup table of an adder (or subtractor).

cliffordwolf:
Here is a proof-of-concept implementation of what I mean. Ironically it does not work yet. (I have a bug somewhere in the handling of full adders, and I have not completed support for full subtractors yet.)

https://github.com/cliffordwolf/yosys/blob/adders_test/passes/techmap/adders.cc

rqou:
I consider these "reverse engineering" passes to be only at the level of quality of a synthesis script rather than a core command. I only put these commands in `passes/techmap/` because there currently isn't really a better place to put it. Perhaps we can create a `re/` directory to house these commands? In the longer term, @azonenberg and I want to use Yosys as a common intermediate platform for implementing all types of netlist operations (synthesis, reverse engineering, simulating, and formal verification), and I want to have all of our passes upstream so that we don't have to "chase upstream." In the longer term, it would be useful to have a place to put "crappier" passes other than in `techlibs/`.

As for the actual implementation of `recover_adders`, the way I see it, the three different implementations are all about the same:
* The Petkovska et. al. implementation searches for adders by writing a custom pass for ABC that "allows detecting full adders as pairs of nodes having two cuts with the same three inputs and with logic functions equal to XOR3 and Majority."
* Your proposed approach does the same thing by writing a custom pass for Yosys rather than ABC.
* My proposed approach does the same thing by tricking ABC using the techmapping functionality. This requires writing a minimal amount of new code but does feel somewhat inelegant.

Other than the part where `extract` is not very efficient, is there anything "wrong" with my "write the least amount of new code" approach?

cliffordwolf:
> The Petkovska et. al. implementation searches for adders by writing a custom pass for ABC that "allows detecting full adders as pairs of nodes having two cuts with the same three inputs and with logic functions equal to XOR3 and Majority."

Relevant references (just for the record):

- https://people.eecs.berkeley.edu/~alanmi/publications/2015/fpt15_carry.pdf
- https://people.eecs.berkeley.edu/~alanmi/publications/2017/tech17_arith.pdf

> Other than the part where extract is not very efficient, is there anything "wrong" with my "write the least amount of new code" approach?

No, but I don't think it is necessary to create a ScriptPass for this. Just keeping the data files around should be sufficient. (Like we do with e.g. `techlibs/common/adff2dff.v`: There is no special ScriptPass to wrap the techmap call.)

> I consider these "reverse engineering" passes to be only at the level of quality of a synthesis script rather than a core command.

That distinction is not going to be clear to a user discovering the command tho. :)

Give me a few more days.. I like to see if I can make my implementation of the half/full adder extractor work..

cliffordwolf:
Either way.. I think `recover_adder_core` (or whatever it will ultimately be called) should operate on full and half adders only (not subtractors), because the only difference is the inverter and I think it would be best to first recover the chain and then look at inverters and XORs in the data path leading into the adder chain.

cliffordwolf:
In Yosys git head there is now a command called `extract_fa` that extracts full and half adders as `$fa` cells. Can your `recover_adder_core` be modified to use the netlists produced by `extract_fa`?

rqou:
I just tested `extract_fa` on [this netlist](https://gist.github.com/rqou/d9441c8b388d15832688b0ff4c008ecf) and it seems to hang forever without outputting anything, but I didn't investigate why. The hacky `recover_adder` script works correctly though. Do you want to continue to work on `extract_fa`, or can we maybe just remove the `recover_adder` driver script and keep the `.lib` and `.v` files (keeping the hacky approach without "advertising" it to users)?

cliffordwolf:
> I just tested extract_fa on this netlist and it seems to hang forever without outputting anything

Please test again with latest git head.

rqou:
I finally got around to testing this again, and it unfortunately still doesn't work as well as my hack. I noticed two immediate issues:

* The pass will extract `$fa` cells that have the X output unused. In many of these cases, the C input is also unused. In other words, the pass managed to extract a single AND gate into a `$fa`, which then interferes with the extraction of a larger gate that might use this AND gate.
* The pass seems to only look at inputs/outputs that exist as wires in the design. This means that if the intermediate carry nodes aren't accessible as a wire (e.g. because it's part of some much larger gate) the pass won't find it. This can be worsened by running `abc` (no args) beforehand as the AOI/OAI cells  seem to often trigger this problem. This can be somewhat improved by breaking the design into an AIG with `abc -g AND`, but this still runs into the first issue. Also, at this point, you've on the path of just reinventing abc :)

And finally, `extract_fa` is actually slower than tricking abc into techmapping. A very unscientific benchmark:
```
$ ../yosys/yosys -p "read_json math_for_clifford.json" -p "extract_fa"
<snip>
End of script. Logfile hash: d58a61e0d8
CPU: user 1.47s system 0.00s, MEM: 49.21 MB total, 17.39 MB resident
Yosys 0.7+300 (git sha1 a85c39ea, clang 3.8.1-24 -fPIC -O0)
Time spent: 99% 1x extract_fa (1 sec), 0% 2x read_json (0 sec)
```
```
$ ../yosys/yosys -p "read_json math_for_clifford.json" -p "recover_adder"
<snip>
End of script. Logfile hash: 439da55f1b
CPU: user 0.07s system 0.01s, MEM: 49.23 MB total, 18.15 MB resident
Yosys 0.7+300 (git sha1 a85c39ea, clang 3.8.1-24 -fPIC -O0)
Time spent: 19% 2x read_json (0 sec), 19% 1x extract (0 sec), ...
```

I still think a better approach is to combine tricking abc with a new optimized-for-small-needles subcircuit extraction pass.

rqou:
Just for reference, `math_for_clifford.json` contains a single 8+8->8 bit adder (as well as one each of a loose AND, OR, and XOR gate)

cliffordwolf:
> And finally, extract_fa is actually slower than tricking abc into techmapping. A very unscientific benchmark:

The problem is not absolute performance on small benchmarks. The problem is algorithmic complexity in large circuits.

The "extract" command has best case algorithmic complexity of O(N^2) for a haystack of size N. (Worst case is exponential.)

cliffordwolf:
> The pass will extract $fa cells that have the X output unused. In many of these cases, the C input is also unused. In other words, the pass managed to extract a single AND gate into a $fa, which then interferes with the extraction of a larger gate that might use this AND gate.

Not exactly what happens but I can see what you mean. :) That's now fixed in current git head.

However, there is still a bug. (The new circuit after extract_fa isn't equivalent to the original circuit.) I'll keep investigating.

> The pass seems to only look at inputs/outputs that exist as wires in the design.

Not true. It also looks at the inverse of all signals.

> This can be worsened by running abc (no args) beforehand as the AOI/OAI cells seem to often trigger this problem. This can be somewhat improved by breaking the design into an AIG with abc -g AND, but this still runs into the first issue.

Yes, of course. I should have posted my script: I'm testing with `abc -g AND; opt -fast` and `abc -g AND,XOR; opt -fast` before calling `extract_fa`.

> Also, at this point, you've on the path of just reinventing abc :)

No, I'm not! ABC can only map XOR3 and MAJ **independently of each other**. It might find a good map for the entire circuit with only XOR3 and MAJ, but in a way so that those cells never align, and might miss a different possible graph cover (that might require more cells) what has the XOR3 and MAJ cells all aligned! Given the associativity of the XOR operator this is not an unlikely scenario at all! The whole idea of my `extract_fa` pass is to *only* map if both the XOR3 and the MAJ align so that they can be converted into a `$fa` cell.

cliffordwolf:
PR superseded by `extract_fa`. Please report `extract_fa` issues if not working as intended.

PR #387, opened 2017-08-15
Added "recover_reduce" pass to extract reduce_* cells from a primitive cell-level netlist

azonenberg:


cliffordwolf:
Thanks for the patience guys. There are some things that need discussing with this PR imo:

- Let's start with the simple things: First Yosys C++ files are called .cc, not .cpp.

- Next: Coding style. I might need to write a style guide to clarify the details of the coding style used in Yosys, but the most obvious thing is this: I indent with tabs, and continuations of previous lines are indented with double tabs. I'm not so strict about that in stuff in `techlibs/<....>/` because I consider that code more loosely connected to the rest of Yosys and kind-of isolated projects. But for everything in the "general" source directories I'd like to keep a consistent coding style.

- Now the real issue: I think the "recover_reduce" as proposed in this patch should be removed and "recover_reduce_core" should be renamed to "recover_reduce". The documentation to that command should mention that it operates on `$_AND_, $_OR_, and $_XOR_` cells and might contain a hint that more reduce cells can be extracted by first mapping all logic to this cell types using `abc -g AND,OR,XOR`.

rqou:
Rebased, and updated coding style.

Sorry for blindly hitting "new file" in my text editor without checking the formatting settings :P

cliffordwolf:
Merged with changes. Among other things I've renamed it to `extract_reduce`.

Btw jfyi regarding this:

```
virtual void execute(std::vector<std::string> args, RTLIL::Design *design)
{
    (void)args;
```

Two problems with this: (1) You should of course use at least `extra_args(...)` on args so that the pass can be used with a selection and (2) if you really have unused parameters in C++ code, simply omit the parameter name in the function header like this:

```
virtual void execute(std::vector<std::string>, RTLIL::Design *design)
```

PR #398, opened 2017-08-28
Added extract_tff pass to find toggle flipflops in a netlist

azonenberg:


cliffordwolf:
Like with the adder stuff, I don't think we should add "commands" that are just ABC+extract scripts, especially since extract is very inefficient for the large haystack + small needle scenario.

I'll be adding a small framework that will simply matching small subcircuits. Give me a few weeks to get to that..

nakengelhardt:
This looks to be superseded by #521 (and I think the "small framework" may have ended up being pmgen?) so I'll close this. If I'm wrong leave a comment!

PR #401, opened 2017-08-29
Initial coarse cell support.

Kmanfi:
Tested with 200k+ design. 60k @or cells converted to 6k @reduce_or cells. Might be useful if you can ask from gate.is_single_bit().


cliffordwolf:
@rqou Can you review this PR please?

passes/techmap/extract_reduce.cc
```diff
@@ -115,9 +115,21 @@ struct ExtractReducePass : public Pass
...
+					gt = GateType::Or;
+				else if (cell->type == "$xor")
+					gt = GateType::Xor;
 				else
 					continue;
 
+				int a_width = cell->getParam("\\A_WIDTH").as_int();
+				int b_width = cell->getParam("\\B_WIDTH").as_int();
+				int y_width = cell->getParam("\\Y_WIDTH").as_int();
+				if ((a_width != 1) || (b_width != 1)|| (y_width != 1))
```
rqou: Nit: The second set of `||` is missing a space

passes/techmap/extract_reduce.cc
```diff
@@ -143,6 +164,11 @@ struct ExtractReducePass : public Pass
 				}
 
 				log("  Head cell is %s\n", head_cell->name.c_str());
+				int ha_width = head_cell->getParam("\\A_WIDTH").as_int();
```
rqou: This seems (extra) redundant. The previous loop to find the sink node checks for the widths, and the subsequent merging operation also checks the widths. Is this actually necessary?

passes/techmap/extract_reduce.cc
```diff
@@ -154,36 +180,54 @@ struct ExtractReducePass : public Pass
 					cur_supercell.insert(x);
 
 					auto a = sigmap(x->getPort("\\A"));
-					log_assert(a.size() == 1);
-					// Must have only one sink
-					// XXX: Check that it is indeed this node?
-					if (sig_to_sink[a[0]].size() + port_sigs.count(a[0]) == 1)
+					if (a.size() == 1)
```
rqou: The way the code used to work was that any cells that were in the BFS queue have already been checked to ensure that they are appropriate. Therefore, I don't think check should be added here. The code previously had `log_assert` statements to ensure that cells had the expected number of inputs, but your new code will just fail silently if something unexpected happens.

passes/techmap/extract_reduce.cc
```diff
@@ -154,36 +180,54 @@ struct ExtractReducePass : public Pass
...
+							// XXX: Check that it is indeed this node?
+							if (sig_to_sink[a[0]].size() + port_sigs.count(a[0]) == 1)
+							{
+								Cell* cell_a = sig_to_driver[a[0]];
+								if (((cell_a->type == "$_AND_" && gt == GateType::And) ||
+									 (cell_a->type == "$_OR_" && gt == GateType::Or) ||
+									 (cell_a->type == "$_XOR_" && gt == GateType::Xor) ||
+									 (cell_a->type == "$and" && gt == GateType::And) ||
+									 (cell_a->type == "$or" && gt == GateType::Or) ||
+									 (cell_a->type == "$xor" && gt == GateType::Xor)
```
rqou: I think you should move a copy of the width checking from line 148 to here instead.

passes/techmap/extract_reduce.cc
Kmanfi: I removed the guard and it will give assert. These lines will add more than single bit cells to BFS queue.
cell_a may be any width.
```
								Cell* cell_a = sig_to_driver[a[0]];
								if (((cell_a->type == "$_AND_" && gt == GateType::And) ||
									 (cell_a->type == "$_OR_" && gt == GateType::Or) ||
									 (cell_a->type == "$_XOR_" && gt == GateType::Xor) ||
									 (cell_a->type == "$and" && gt == GateType::And) ||
									 (cell_a->type == "$or" && gt == GateType::Or) ||
									 (cell_a->type == "$xor" && gt == GateType::Xor)
									 ))
								{
									// The cell here is the correct type, and it's definitely driving only
									// this current cell.
									bfs_queue.push_back(cell_a);

```

Kmanfi: Please try with this @rqou .
```
read_verilog this.v
prep
extract_reduce
```
```
module and_reduce_get_bit (a, c, d, y);
input [3:0] a, c, d;
output y;
wire [3:0] tmp;
    assign tmp = (c & d);
    assign y = (a[3] & a[2] & a[1] & a[0] & tmp[3]);
endmodule
```

passes/techmap/extract_reduce.cc
Kmanfi: Refactored.

Kmanfi:
Latest version doesn't assert and seems to find reduce_* cells from big design, but from mini example I posted here it doesn't find.
Please could you @rqou help with this.

rqou:
+1

Looks like it works, and even fixed a bug that I had.

passes/techmap/extract_reduce.cc
```diff
@@ -177,18 +216,19 @@ struct ExtractReducePass : public Pass
...
+								// The cell here is the correct type, and it's definitely driving only
+								// this current cell.
+								bfs_queue.push_back(cell_b);
+							}
 						}
 					}
 				}
 
-				log("  Cells:\n");
+				log("  Cells: count:%zu\n", cur_supercell.size());
```
cliffordwolf: Please use `log("  Cells: count:%d\n", GetSize(cur_supercell));` instead. The `z` type specifier has portability issues.

passes/techmap/extract_reduce.cc
```diff
@@ -50,6 +50,35 @@ struct ExtractReducePass : public Pass
...
+		return res;
+	}
+
+	bool is_single_bit(Cell *cell, SigMap sigmap)
+	{
+		bool res = true;
+		auto a = sigmap(cell->getPort("\\A"));
+		auto b = sigmap(cell->getPort("\\B"));
+		auto y = sigmap(cell->getPort("\\Y"));
+		if ((a.size() != 1) || (b.size() != 1) ||(y.size() != 1))
```
cliffordwolf: `((GetSize(a) != 1) || (GetSize(b) != 1) || (GetSize(y) != 1))` would be idiomatic for Yosys code.

Kmanfi:
Fixed as requested.

cliffordwolf:
Thanks. The PR currently has merge conflicts with Yosys master. Can you pull (merge) master and resolve the conflicts? Thanks.

Kmanfi:
Wondering... did I make this pull(merge), resolve as it supposed to be. Learning to use git...

Kmanfi:
Current head version of extract_reduce is quite different so it's easier to make new PR based on it.

PR #403, opened 2017-08-29
Add support for source line tracking through synthesis phase

jlowder:
These are some minor edits to help retain source line tracking information on cells through some of the synthesis passes, i.e. propagate "src" attributes from original cells to the derived cells they are being replaced with.

### DIFF
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -1620,18 +1620,19 @@ RTLIL::Cell *RTLIL::Module::addCell(RTLIL::IdString name, const RTLIL::Cell *oth
 }
 
 #define DEF_METHOD(_func, _y_size, _type) \
-	RTLIL::Cell* RTLIL::Module::add ## _func(RTLIL::IdString name, RTLIL::SigSpec sig_a, RTLIL::SigSpec sig_y, bool is_signed) { \
+		RTLIL::Cell* RTLIL::Module::add ## _func(RTLIL::IdString name, RTLIL::SigSpec sig_a, RTLIL::SigSpec sig_y, bool is_signed, std::string src) { \
 		RTLIL::Cell *cell = addCell(name, _type);           \
 		cell->parameters["\\A_SIGNED"] = is_signed;         \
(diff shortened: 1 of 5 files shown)


kernel/rtlil.cc
```diff
@@ -1620,18 +1620,19 @@ RTLIL::Cell *RTLIL::Module::addCell(RTLIL::IdString name, const RTLIL::Cell *oth
 }
 
 #define DEF_METHOD(_func, _y_size, _type) \
-	RTLIL::Cell* RTLIL::Module::add ## _func(RTLIL::IdString name, RTLIL::SigSpec sig_a, RTLIL::SigSpec sig_y, bool is_signed) { \
+		RTLIL::Cell* RTLIL::Module::add ## _func(RTLIL::IdString name, RTLIL::SigSpec sig_a, RTLIL::SigSpec sig_y, bool is_signed, std::string src) { \
```
cliffordwolf: Why the indent change? Imo it was correct before.

passes/techmap/alumacc.cc
```diff
@@ -55,27 +55,27 @@ struct AlumaccWorker
 
 		RTLIL::SigSpec get_gt() {
 			if (GetSize(cached_gt) == 0)
-				cached_gt = alu_cell->module->Not(NEW_ID, alu_cell->module->Or(NEW_ID, get_lt(), get_eq()));
+				cached_gt = alu_cell->module->Not(NEW_ID, alu_cell->module->Or(NEW_ID, get_lt(), get_eq()), false, alu_cell->attributes["\\src"].decode_string());
```
cliffordwolf: Commit 8530333 adds `get_src_attribute()` and `set_src_attribute()`. Using `get_src_attribute()` has the advantage that it does not create an empty src attribute on objects that do not have one yet. Please update your PR to use this two functions.

jlowder: Sounds good. I have updated the PR to use the new functions - please let me know if anything else is needed.

cliffordwolf:
I've merged the PR and changed the remaining places where the new functions should be used in the new code in commit 8a66bd3.

PR #410, opened 2017-09-13
Added "opt_demorgan" pass (fixes #408)

### LINKED ISSUE #408: Feature request: extract_reduce should push inverters to output if DeMorgan representation is smaller
If more than half of the inputs to a $reduce_and are complemented, the circuit will be smaller in the DeMorgan alternative (inverted-input $reduce_or plus output inverter)

azonenberg:
This pass replaces "reduce_and" or "reduce_or" cells in which >50% of the inputs are driven by NOT gates with their DeMorgan equivalents, to reduce area consumed by inverters.

### DIFF
--- a/passes/opt/Makefile.inc
+++ b/passes/opt/Makefile.inc
@@ -6,11 +6,11 @@ OBJS += passes/opt/opt_reduce.o
 OBJS += passes/opt/opt_rmdff.o
 OBJS += passes/opt/opt_clean.o
 OBJS += passes/opt/opt_expr.o
-OBJS += passes/opt/opt_demorgan.o
-OBJS += passes/opt/rmports.o
 
 ifneq ($(SMALL),1)
 OBJS += passes/opt/share.o
 OBJS += passes/opt/wreduce.o
+OBJS += passes/opt/opt_demorgan.o
+OBJS += passes/opt/rmports.o
 endif
 
--- a/passes/opt/Makefile.inc
+++ b/passes/opt/Makefile.inc
@@ -6,6 +6,7 @@ OBJS += passes/opt/opt_reduce.o
 OBJS += passes/opt/opt_rmdff.o
 OBJS += passes/opt/opt_clean.o
 OBJS += passes/opt/opt_expr.o
+OBJS += passes/opt/opt_demorgan.o
 OBJS += passes/opt/rmports.o
 
 ifneq ($(SMALL),1)
--- a/passes/opt/opt_demorgan.cc
+++ b/passes/opt/opt_demorgan.cc
@@ -1,0 +1,202 @@
+/*
+ *  yosys -- Yosys Open SYnthesis Suite
+ *
+ *  Copyright (C) 2017  Clifford Wolf <clifford@clifford.at>
+ *
+ *  Permission to use, copy, modify, and/or distribute this software for any
+ *  purpose with or without fee is hereby granted, provided that the above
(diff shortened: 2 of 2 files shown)


passes/opt/opt_demorgan.cc
```diff
@@ -0,0 +1,202 @@
...
+		//Don't automatically delete the inverter since other stuff might still use it
+		else
+			insig[i] = srcinv->getPort("\\A");
+	}
+
+	//Cosmetic fixup: If our input is just a scrambled version of one bus, rearrange it
+	//Reductions are all commutative, so there's no point in having them in a weird order
+	bool same_signal = true;
+	RTLIL::Wire* srcwire = insig[0].wire;
+	std::map<int, int> seen_bits;
```
cliffordwolf: You might want to use `dict<int, int>` instead. It is much more efficient.

passes/opt/opt_demorgan.cc
```diff
@@ -0,0 +1,202 @@
...
+		log("This pass pushes inverters through $reduce_* cells if this will reduce the\n");
+		log("overall gate count of the circuit\n");
+		log("\n");
+	}
+	virtual void execute(std::vector<std::string> /*args*/, RTLIL::Design *design)
+	{
+		log_header(design, "Executing OPT_DEMORGAN pass (push inverters through $reduce_* cells).\n");
+
+		//int argidx = 0;
+		//extra_args(args, argidx, design);
```
cliffordwolf: Without `extra_args()` the `[selection]` in the help message is a lie. :)
Please call `extra_args()`.

passes/opt/opt_demorgan.cc
```diff
@@ -0,0 +1,202 @@
...
+	auto m = cell->module;
+
+	//TODO: Add support for reduce_xor
+	//DeMorgan of XOR is either XOR (if even number of inputs) or XNOR (if odd number)
+
+	if( (cell->type != "$reduce_and") && (cell->type != "$reduce_or") )
+		return;
+
+	auto insig = sigmap(cell->getPort("\\A"));
+	log("Inspecting %s cell %s (%d inputs)\n", log_id(cell->type), log_id(cell->name), insig.size());
```
cliffordwolf: Please use `GetSize(insig)` instead. This returns an `int` which will match `%d` on all platforms.
Same in `log()` below.

passes/opt/opt_demorgan.cc
```diff
@@ -0,0 +1,202 @@
...
+	int num_inverted = 0;
+	for(int i=0; i<insig.size(); i++)
```
cliffordwolf: `GetSize(insig)` is idiomatic in Yosys code. It just casts the unsigned size_t into a signed int. There are a few instances of `.size()` that should be changed.

passes/opt/Makefile.inc
```diff
@@ -6,6 +6,7 @@ OBJS += passes/opt/opt_reduce.o
 OBJS += passes/opt/opt_rmdff.o
 OBJS += passes/opt/opt_clean.o
 OBJS += passes/opt/opt_expr.o
+OBJS += passes/opt/opt_demorgan.o
```
cliffordwolf: Should be moved down to the non-small section. Same thing for `rmports` actually.

azonenberg:
I think this fixes everything you wanted?

cliffordwolf:
Looks good. Thanks!

PR #413, opened 2017-09-15
Added support for off-chain loads in extract_reduce

azonenberg:
Added an optional, non-default argument that makes "extract_reduce" infer a reduction operation even if intermediate gates in the reduction chain drive other logic. This can lead to needless replication of logic and an increase in design area, but is a useful intermediate in coarse-grained synthesis and reverse engineering flows.

### DIFF
--- a/passes/techmap/extract_reduce.cc
+++ b/passes/techmap/extract_reduce.cc
@@ -37,7 +37,7 @@ struct ExtractReducePass : public Pass
 	{
 		//   |---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|
 		log("\n");
-		log("    extract_reduce [selection]\n");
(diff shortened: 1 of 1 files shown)


passes/techmap/extract_reduce.cc
```diff
@@ -56,13 +68,14 @@ struct ExtractReducePass : public Pass
...
-			// if (args[argidx] == "-v") {
-			// 	verbose = true;
-			// 	continue;
-			// }
-			break;
+			if (args[argidx] == "-allow-off-chain")
+			{
+				allow_off_chain = true;
+				continue;
+			}
```
cliffordwolf: There is a `break` missing and the end of this loop.

azonenberg:
Fixed

cliffordwolf:
Thanks!

PR #415, opened 2017-09-21
Adding support for string macros and macros with arguments after include

combinatorylogic:
It's a bit clumsy, but illustrates the problem.


cliffordwolf:
It passes the big test bench, so it must be okay.. :) Thanks for the patch.

PR #430, opened 2017-10-02
Resolved classical Bison IF/THEN/ELSE shift/reduce conflict using the…

udif:
… textbook solution

(Oreilly 'Flex & Bison' page 188-189)

The current solution simply allows the Bison output to be used if it has exactly 2 conflicts.
This is less robust than this solution which explicitly solves these two if/then/else conflicts (one for normal if/then/else, the other within generate blocks)

patch passed 'make test'.


cliffordwolf:
I'm pretty sure I read in the dragon book that shift/reduce conflicts for ELSE are fine. So I claim that the old code was the textbook solution!  :)

Just kidding, thanks for the patch. That's obviously an improvement.

PR #433, opened 2017-10-09
Support for single bit coarse cells. Option for debug prints.

Kmanfi:


cliffordwolf:
@rqou Can you review this PR please? Thanks.

rqou:
Apologies, I missed this. I will review this soon.

whitequark:
It would be really helpful if this PR added a debug option and a new feature in separate commits.

cliffordwolf:
> Apologies, I missed this. I will review this soon.

Are we there yet? @rqou     :D

passes/techmap/extract_reduce.cc
```diff
@@ -52,14 +52,30 @@ struct ExtractReducePass : public Pass
...
 		return (cell->type == "$_AND_" && gt == GateType::And) ||
 				(cell->type == "$_OR_" && gt == GateType::Or) ||
-				(cell->type == "$_XOR_" && gt == GateType::Xor);
+				(cell->type == "$_XOR_" && gt == GateType::Xor) ||
+				(cell->type == "$and" && gt == GateType::And) ||
+				(cell->type == "$or" && gt == GateType::Or) ||
+				(cell->type == "$xor" && gt == GateType::Xor);
+	}
+
+	bool is_single_bit(Cell *cell, SigMap sigmap)
```
rqou: I think it would be a good idea to call this function from the `IsRightType` function (rather than modifying all of the call sites). This helps prevent accidentally forgetting to call `is_single_bit` in a critical location.

passes/techmap/extract_reduce.cc
```diff
@@ -144,9 +187,17 @@ struct ExtractReducePass : public Pass
...
 							break;
+						if (!is_single_bit(x, sigmap))
+							break;
 
 						head_cell = x;
 
+						if (debug_verbose)
+						{
+							log("  Chain head cell is: %s\n", head_cell->name.c_str());
+							log("  Chain head cell Y width: %i\n", GetSize(head_cell->getPort("\\Y")));
```
rqou: Can any of these Y width debug prints ever print a value other than 1? If not, there's no value in having these particular print statements (the other verbose logging is still useful).

PR #435, opened 2017-10-17
Reduce vector fix

Kmanfi:
New try. Infinite loop should be now fixed.


whitequark:
Is there a testcase available for the bug that is fixed by this PR?

cliffordwolf:
> Is there a testcase available for the bug that is fixed by this PR?

see #434 @whitequark

PR #454, opened 2017-11-15
Add option to statically link abc; emscripten fixes

rqou:


rqou:
This depends on this following abc PR: https://bitbucket.org/alanmi/abc/pull-requests/85/add-an-option-to-use-c99-stdinth/diff

Makefile
```diff
@@ -453,6 +474,29 @@ endif
 yosys-abc$(EXE): abc/abc-$(ABCREV)$(EXE)
 	$(P) cp abc/abc-$(ABCREV)$(EXE) yosys-abc$(EXE)
 
+abc/libabc-$(ABCREV).a:
```
cliffordwolf: This seems to be only different from `abc/abc-$(ABCREV)$(EXE)` above in the make target passed to ABC. I think this rule can be merged with the rule for `abc/abc-$(ABCREV)$(EXE)`.

Something like the following might work (warning: untested nontrivial Makefile code ahead :).

```
abc/abc-$(ABCREV)$(EXE) abc/libabc-$(ABCREV).a:
    ....
    $(Q) cd abc && $(MAKE) ... $(if $(filter %.a,$@),libabc-$(ABCREV).a)
```

rqou:
Finally updated. What do you think should be done about `ABC_MEMALIGN`? The workaround I proposed or something more correct?

passes/techmap/abc.cc
```diff
@@ -943,8 +949,24 @@ void abc_module(RTLIL::Design *design, RTLIL::Module *current_module, std::strin
 		buffer = stringf("%s -s -f %s/abc.script 2>&1", exe_file.c_str(), tempdir_name.c_str());
 		log("Running ABC command: %s\n", replace_tempdir(buffer, tempdir_name, show_tempdir).c_str());
 
+#ifndef YOSYS_ENABLE_ABC_STATIC
 		abc_output_filter filt(tempdir_name, show_tempdir);
 		int ret = run_command(buffer, std::bind(&abc_output_filter::next_line, filt, std::placeholders::_1));
+#else
+		// FIXME: This needs to be mutable because abc is terrible
+		char *abc_argv[5];
```
cliffordwolf: I think every program that uses getopt needs argv to be mutable.

kernel/driver.cc
```diff
@@ -80,20 +80,37 @@ USING_YOSYS_NAMESPACE
...
+#  include <emscripten.h>
 
-extern "C" int main(int, char**);
+extern "C" int main(int, const char**);
 extern "C" void run(const char*);
 extern "C" const char *errmsg();
 extern "C" const char *prompt();
 
-int main(int, char**)
+int main(int argc, const char**argv)
```
cliffordwolf: That's an odd. Why is the `const` needed? According to 5.1.2.2.1 of ISO C99, `main()` must be declared as `int main(void)` or `int main(int argc, char *argv[])` or equivalent. (GNU libc adds `int main (int argc, char *argv[], char *envp[])`.)

rqou:
Ping. Sorry for repeatedly dropping this.

Regarding disabling threads in ABC: Thread support in the emscripten environment was originally experimental and is now disabled due to Spectre. Just to be on the safe side, we disable it.

Regarding ABC_MEMALIGN, there is a description in the commit message that I've copied below:

Apparently abc has a memory pool implementation that by default returns memory that is unaligned. There is a workaround in the abc makefile that uses uname to look for "arm" specifically and then sets the alignment. However, ARM is not the only platform that requires proper alignment (e.g. emscripten does too). For now, pessimistically force the alignment for 8 bytes all the time (somehow 4 wasn't enough for fixing emscripten despite being approximately a 32-bit platform).

kernel/driver.cc
```diff
@@ -85,20 +85,37 @@ USING_YOSYS_NAMESPACE
...
 extern "C" const char *prompt();
 
-int main(int, char**)
+int main(int argc, char **argv)
 {
+	EM_ASM(
+		if (ENVIRONMENT_IS_NODE)
+		{
+			FS.mkdir('/hostcwd');
+			FS.mount(NODEFS, { root: '.' }, '/hostcwd');
```
yuyichao, line 101: This causes the code to fail if NODEFS isn't linked when running on node. It should be up to the user to decide if the host file system should be mounted...

PR #455, opened 2017-11-18
Add UltraPlus specific cells to ice40 techlib

daveshah1:
This adds `SB_MAC16`, `SB_SPRAM256KA`, `SB_HFOSC`, `SB_LFOSC` and `SB_RGBA_DRV` to the ice40 `cells_sim.v`.

It does not yet contain simulation models or any support for inference for any of these new primitives.


cliffordwolf:
Why the `keep` attributes? All those blocks seem safe to optimize away if their outputs are unused.

daveshah1:
Yes, you're right - thanks for pointing that out. I've removed them in the latest commit.

PR #464, opened 2017-12-03
On travis download ABC as part of setup

mithro:
This makes travis mark the build as broken rather than failing if ABC fails to download.

This also allows you to download ABC separately from doing the compile, useful for working offline :-)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ cache:
 env:
   global:
     - MAKEFLAGS="-j 2"
+    - ABCINSECURE=1
 
 matrix:
   include:
--- a/.travis/setup.sh
+++ b/.travis/setup.sh
@@ -95,3 +95,18 @@ fi
 )
 
 ##########################################################################
+
+# Downloading abc
+(
+	if [ ! -d abc ]; then
+		echo
+		echo 'Downloading abc...' && echo -en 'travis_fold:start:before_install.abc\\r'
+		echo
+		make abc
+		echo
+		echo -en 'travis_fold:end:before_install.abc\\r'
+		echo
+	fi
+)
+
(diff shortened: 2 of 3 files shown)


cliffordwolf:
1. I don't see how this follows ABC updates, i.e. makes sure it always tests with the correct version of ABC.

2. Afaict `hg` changed their default behavior from not checking certs to checking certs and Travis just has a newer version of `hg` now. So I think using ` --insecure` with hg should be the new default behavior and I don't see any need to make it configurable by adding `ABCINSECURE` and `ABCHGARGS`.

mithro:
This change (assuming I have it correct) just splits downloading abc from compiling abc. Behaviour for the user should be the same unless they specifically call `make abc` first.

I'm afraid I don't understand what you mean by (1) -- It still uses the ABC_REV like previously?

Checking certificates is the right behaviour and should be encouraged.

cliffordwolf:
> I'm afraid I don't understand what you mean by (1) -- It still uses the ABC_REV like previously?

Afaict it wouldn't because after updating the source base `if [ ! -d abc ]; then` would fail because the abc directory already exists and so it wouldn't rebuild ABC, even when ABCREV has changed. Or am I missing something?

> Checking certificates is the right behaviour and should be encouraged.

Sure, and once bitbucket.org gets their certificates straight we would remove `--insecure` again. But right now there are two groups of user: (1) those with old hg that simply don't check certificates because its the old default behavior and (2) those with new hg for which it fails unless `--insecure` is used.

I don't see how not using `--insecure` is helping either group. It's a no-op for the old version and for users of the new version the build is just broken without it.

cliffordwolf:
ABC is now on github so this is obsolete now.

PR #468, opened 2017-12-10
Fix SB_IO_OD module

grahamedgecombe:
This PR fixes two mistakes that were preventing Icarus from being able to parse the `SB_IO_OD` module:

* Removes the trailing comma at the end of the port list. Although Yosys supports this syntax, Icarus doesn't seem to and it's inconsistent with all the other modules in `ice40/cells_sim.sv`.
* Fixes port names inside the body of the module - the port names are declared without underscores (and although this is inconsistent with the similar `SB_IO` module, it is correct according to the technology library PDF and existing code on the Internet) but the body attempted to refer to the ports with underscores in the names.

### DIFF
--- a/techlibs/ice40/cells_sim.v
+++ b/techlibs/ice40/cells_sim.v
@@ -1205,44 +1205,44 @@ module SB_IO_OD (
 	reg outena_q;
 
 	generate if (!NEG_TRIGGER) begin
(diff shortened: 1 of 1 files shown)


cliffordwolf:
cc @daveshah1

PR #469, opened 2017-12-13
Use quote includes for yosys.h

kkiningh:
I'm building Yosys using Bazel, and I noticed that you're using <> includes for some non-system headers. Changing this makes it easier to build since Bazel can be picky about how files get included.

### DIFF
--- a/kernel/celltypes.h
+++ b/kernel/celltypes.h
@@ -20,7 +20,7 @@
 #ifndef CELLTYPES_H
 #define CELLTYPES_H
 
-#include <kernel/yosys.h>
+#include "kernel/yosys.h"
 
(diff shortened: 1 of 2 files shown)


mithro:
Hi @kkiningh,

Why are you using Yosys with Bazel? I have BUILD rules for Yosys but didn't think they would be useful for anyone else..

Tim 'mithro' Ansell

kkiningh:
I'm working on a (soon to be) open source project to do verification + full ASIC/FPGA flow using Bazel. It's a super early WIP, but ultimately I'd like to setup a service to do RTL DV/CI on AWS using Bazel as the build system.

mithro:
Send me an email at me@mith.ro and we can chat some more....

PR #471, opened 2017-12-15
Add output_directory arg for bram_init

kkiningh:
This change adds a single argument to specify the output directory for xilinx/bram_init.py, defaulting to the script directory (which was the previous behavior). This makes it easier for (non-default) build systems to perform out-of-source builds.

### DIFF
--- a/techlibs/xilinx/brams_init.py
+++ b/techlibs/xilinx/brams_init.py
@@ -1,6 +1,18 @@
 #!/usr/bin/env python3
 
-with open("techlibs/xilinx/brams_init_18.vh", "w") as f:
+from __future__ import absolute_import
+from __future__ import division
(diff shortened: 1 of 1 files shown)


techlibs/xilinx/brams_init.py
```diff
@@ -1,6 +1,18 @@
 #!/usr/bin/env python3
 
-with open("techlibs/xilinx/brams_init_18.vh", "w") as f:
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
```
cliffordwolf: This is a python 3 script. Those `from __future__ import` lines are unnecessary.

cliffordwolf:
Closing because OP did not respond to requested changes.

Issue #87, opened 2015-09-30
Performance issues with memory initialization


gaomy3832:
I have two performance issues with the memory initialization. I am not sure if there are good and easy solutions, or even it is worth it to improve (while I do think these are moderate size applications), but just want to report these issues to motivate some discussion.

First, the verilog frontend runs really slow if there is an `initial` block with a moderate-sized `$readmemh` function (e.g., 4096 * 32). I attach an example here

https://github.com/gaomy3832/yosys_tests/blob/master/rom.v
https://github.com/gaomy3832/yosys_tests/blob/master/rom_ram.dat

If I run `./yosys -p "read_verilog rom.v;"`, it takes about 15 min.

```
CPU: user 875.76s system 0.06s, MEM: 110.15 MB total, 86.50 MB resident
Yosys 0.5+332 (git sha1 ddcfc99, clang 3.4-1ubuntu3 -fPIC -Os)
Time spent: 100% 2x read_verilog (875 sec)
```

Another issue is a little complex. What I want to do is to techmap the BRAM into my custom 1-bit cells (i.e., split the columns). For example, mapping a 64 * 8 BRAM into eight 64 * 1 BRAMs. If I want to preserve the initial values, I need to extract each column from the `INIT` parameter, which I assume is stored in row major format. So I try to use a constant function to do the transpose, see the example code here

https://github.com/gaomy3832/yosys_tests/blob/master/transpose.v

If I run `./yosys -p "read_verilog transpose.v"`, it doesn't finish after 30 min (I have to kill it).

I am not sure if using constant function is the best solution, so please let me know if yosys supports a better way to do the work.

A final note for my environment, I ran these tests on a server-class CPU (Xeon E5-2630) so my system should have enough processing power. I've also disabled tcl, abc, and readline when compiling, but I don't think it has any significant effect.

cliffordwolf:
Easy things first: I have now fixed the `$readmem` performance issue in commit 9caeadf79713534a1fd5f51fb458af245c34dad9.

Regarding block ram mapping:

You might want to consider using the `memory_bram` pass to map the memory to the type of block rams you have, and then use techmap to implement a 1:1 mapping of the cells created by `memory_bram` to the cells you have in your architecture. For example with your `rom.v` test case:

First you need to create a rules file for `memory_bram`:

    $ cat my_brams.txt 
    bram MY_64x1_BRAM
      init 1
      abits 6
      dbits 1
      groups 2
      ports  1 1
[...]

PR #479, opened 2018-01-03
Some standard cell libraries include a latch with only set/reset.

Fatsie:
I am using a standard cell library with latches that only contain set/reset. This patch adds command line option that allows to ignore these cells.

### DIFF
--- a/frontends/liberty/liberty.cc
+++ b/frontends/liberty/liberty.cc
@@ -290,7 +290,7 @@ static void create_ff(RTLIL::Module *module, LibertyAst *node)
 	log_assert(!cell->type.empty());
 }
 
-static void create_latch(RTLIL::Module *module, LibertyAst *node)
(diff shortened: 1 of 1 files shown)


Fatsie:
I am not familiar with travis and CI but I don't understand why job #87.1 fails in .travis/setup.sh and the others don't.

cliffordwolf:
This CI fail has obviously nothing to do with your patch. @mithro thinks this was a temporary issue on the travis side. I'm re-running the failed build now. Let's see if that fixes it.

cliffordwolf:
Nope. @mithro can you have a look at this? https://travis-ci.org/cliffordwolf/yosys/jobs/324763316

cliffordwolf:
Clearly that CI result has nothing to do with the actual code changes, so I'm now merging this.

PR #480, opened 2018-01-03
Value of properties can be expression.

Fatsie:
Patch to fix expressions used in value for properties in the liberty file. I will need to push similar patch to abc for full support.


Fatsie:
FYI, similar patch has been merged in abc: https://bitbucket.org/alanmi/abc/commits/6e3c24b3308aeba7325a9dbfb5d5d94f84ea5d17

PR #494, opened 2018-01-22
firrtl backend demote errors cell type not supported, write port

edcote:
FIRRTL backend is not fully implemented.  This PR is to throw warnings instead of errors "cell type not supported" and unsupported memory write port configurations.

### DIFF
--- a/backends/firrtl/firrtl.cc
+++ b/backends/firrtl/firrtl.cc
@@ -373,10 +373,10 @@ struct FirrtlWorker
 				for (int i = 0; i < wr_ports; i++)
 				{
 					if (wr_clk_enable[i] != State::S1)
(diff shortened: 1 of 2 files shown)


cliffordwolf:
I think it is a bad idea to change the default behavior to this. After all: The generated FIRRTL file is pretty much guaranteed to not work in any of the cases where those errors are generated.

Please add an option to `write_firrtl` that can be used to turn those errors to warnings instead.

edcote:
Ack. ETA next week.

edcote:
I have deeper knowledge of the backend and will close this PR.  It should not be necessary to demote this to error.  I don't have the exact solution, but I will fix once I hit this again.

PR #505, opened 2018-02-11
FreeBSD build

thefallenidealist:
Changes (only for FreeBSD):
- ABC revision changed to revision from Feb 05 which supports FreeBSD
- removed <code>-ldl</code> - FreeBSD's <code>dlopen()</code> and similar functions are in libc
- paths: changed TCL include and library path
- tests: don't use <code>/bin/bash</code> directly, and use gmake for running test on FreeBSD

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,6 +31,7 @@ SANITIZER =
 # SANITIZER = cfi
 
 
+OS := $(shell uname -s)
 PREFIX ?= /usr/local
 INSTALL_SUDO :=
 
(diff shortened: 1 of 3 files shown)


cliffordwolf:
This is too much in one PR. Would you rather I pick the pieces I don't think need discussing an then you rebase and create small PRs for the rest, or should I tell you what may go in the "large uncontroversial PR"?

thefallenidealist:
Sounds OK, pick the appropriate pieces, and I'll rebase the rest and create small PRs for them.

cliffordwolf:
I have merged the `#ifdef __FreeBSD__` stuff.

Re `#!/usr/bin/env bash`: I don't see how this is necessary. The Makefile already runs those scripts with `bash <filename>`, so the `#!` should be unused actually.

Same with the `MAKE=gmake` stuff. The Makefile calls those scripts with `+ ...`, so `$MAKE` should already be set appropriately. Moreover: Setting `MAKE=gmake` breaks `gmake -jN`.

thefallenidealist:
Changes are rebased and split into 3 commits.

> Re #!/usr/bin/env bash: I don't see how this is necessary. The Makefile already runs those scripts with bash <filename>, so the #! should be unused actually.

You are correct. But /usr/bin/env bash was needed for autotest.sh (because it is run with "../tools/autotest.sh").

Issue #512, opened 2018-03-07
Assigning to a wire on the clock should be an error


ZipCPU:
The following code should generate an error:

```
module test(i_clk, o_reg);
   input wire i_clk;
   output wire o_reg;

  initial o_reg = 1'b0;
   always @(posedge i_clk)
      o_reg <= !o_reg;
endmodule
```

At issue is the fact that o_reg is a wire and not a register, therefore you should not be able to give it a value as though it were a register within an always block,

Dan

udif:
I took a look to see what's required to fix it and I've noticed that a simple fix to enforce this will break the SystemVerilog "logic" keyword which at the moment is simply aliased to "reg", and so it works as both reg/wire only because wire/reg definitions are not enforced.

~~~
module test(i_clk, o_reg, o_wire);
   input wire i_clk;
   output wire o_reg;
   output reg o_wire;

  initial o_reg = 1'b0;
   always @(posedge i_clk)
      o_reg <= !o_reg;
  assign  o_wire = !o_reg;
endmodule
~~~

I took a look at the relevant code, and it doesn't look too hard to fix. I'll try coming up with something.

udif:
PR has been updated, now passes all tests

daveshah1:
#513 is now merged, this code now produces the following warnings (that can be made into errors using `-e` if desired)
```
Warning: wire '\o_reg' is assigned in a block at wirereg.v:5.
Warning: wire '\o_reg' is assigned in a block at wirereg.v:7.
```

PR #513, opened 2018-03-09
Add error checking for reg/wire/logic misuse - PR now passes 'make test' (plus a new test)

udif:
In response to Issue #512, here is a PR that starts addressing this issue (but has issues of
It correctly detects reg/wire mix and incorrect use on blocking, nonblocking assignments within blocks and assign statements.
It seems that within Yosys, wire/reg/logic were all synonyms until now, and that actual behavior was determined by usage. (assign stmt vs (non)blocking assignment within an always block).

You can test it with tests/simple/reg_wire_error.v (look inside for the comments to enable/disable specific lines)

What it DOES'T do (yet):
Detect registers connected to output ports of instances.

***Where it FAILS (do not commit this PR as is!):***
1. memorty nonblocking assignments causes spurious (I assume??) errors on yosys-generated "_ADDR", "_DATA", "EN" signals, as these appear in the AST dump without any "reg" attribute, yet they are driven within an AST_BLOCK by an AST_ASSIGN_LE (tests/techmap/mem_simple_4x1_uut.v)
2. running 'make test' fails with a segmentation fault on tests/simple/mem2reg .

so @cliffordwolf, before I proceed to do additional work, I would like to know first your position on 2 issues:
1. Do you really _want_ support for more strict error checking, or is the current situation really what you intended to be (where no error checking is done)?
2. Distinguishing between 1364-1995, 1364-2001, 1364-2005, 1800-2009, 1800-2012 ?
I remember from a past PR you actually _removed_ a check that enabled a SV feature only when the SV flag was used. What you wrote at the time was that as long as any SV extension did not have the potential to misinterpret legacy Verilog code, you enable it even when no -sv flag was used.
Personally, I think we should add separate flags for the above, and I don't mind contibuting PRs towards this goal. At  least that's what other tools  do (Verilator & gcc '-std=' flag).

I'll leave my specific questions for debugging the memory issue for a separate comment.


udif:
Taking this simple testcase:
```
module t(input clk, input d, output o);
reg a[0:0];

always @(posedge clk)
  a[0] <= d;

assign o = a[0];
endmodule
```

Producing the following AST2 output:
```
Dumping Verilog AST after simplification:
    AST_MODULE <t.v:1> [0xd65f60] str='\t'
      AST_WIRE <t.v:1> [0xd66060] str='\clk' input port=1 range=[0:0]
      AST_WIRE <t.v:1> [0xd66160] str='\d' input port=2 range=[0:0]
      AST_WIRE <t.v:1> [0xd66280] str='\o' output port=3 range=[0:0]
      AST_MEMORY <t.v:2> [0xd66870] str='\a' reg
        AST_RANGE <t.v:2> [0xd66970] range=[0:0]
          AST_CONSTANT <t.v:2> [0xd66a70] bits='00000000000000000000000000000000'(32) signed range=[31:0]
          AST_CONSTANT <t.v:2> [0xd66b90] bits='00000000000000000000000000000000'(32) signed range=[31:0]
        AST_RANGE <t.v:2> [0xd66770] range=[0:0]
          AST_CONSTANT <t.v:2> [0xd66510] bits='00000000000000000000000000000000'(32) signed range=[31:0]
          AST_CONSTANT <t.v:2> [0xd66640] bits='00000000000000000000000000000000'(32) signed range=[31:0]
      AST_ALWAYS <t.v:4> [0xd663b0]
        AST_POSEDGE <t.v:4> [0xd66db0]
          AST_IDENTIFIER <t.v:4> [0xd66cb0 -> 0xd66060] str='\clk'
        AST_BLOCK <t.v:4> [0xd66eb0]
          AST_BLOCK <t.v:5> [0xd6bef0]
            AST_ASSIGN_LE <t.v:5> [0xd6acc0]
              AST_IDENTIFIER <t.v:5> [0xd6aa90 -> 0xd69f00] str='$memwr$\a$t.v:5$1_ADDR'
              AST_CONSTANT <t.v:5> [0xd6ab90] bits='xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx'(32) range=[31:0]
            AST_ASSIGN_LE <t.v:5> [0xd6b030]
              AST_IDENTIFIER <t.v:5> [0xd6ae30 -> 0xd6a390] str='$memwr$\a$t.v:5$1_DATA'
              AST_CONSTANT <t.v:5> [0xd6af30] bits='x'(1) range=[0:0]
            AST_ASSIGN_LE <t.v:5> [0xd6b3a0]
              AST_IDENTIFIER <t.v:5> [0xd6b1a0 -> 0xd6a840] str='$memwr$\a$t.v:5$1_EN'
              AST_CONSTANT <t.v:5> [0xd6b2a0] bits='0'(1) range=[0:0]
          AST_BLOCK <t.v:5> [0xd6b8f0]
            AST_ASSIGN_LE <t.v:5> [0xd69a30]
              AST_IDENTIFIER <t.v:5> [0xd6cfa0 -> 0xd69f00] str='$memwr$\a$t.v:5$1_ADDR'
              AST_CONSTANT <t.v:5> [0xd6d0a0] bits='00000000000000000000000000000000'(32) signed range=[31:0]
            AST_ASSIGN_LE <t.v:5> [0xd6d1a0]
              AST_IDENTIFIER <t.v:5> [0xd6d2a0 -> 0xd6a390] str='$memwr$\a$t.v:5$1_DATA'
              AST_IDENTIFIER <t.v:5> [0xd6d3a0 -> 0xd66160] str='\d'
            AST_ASSIGN_LE <t.v:5> [0xd6d4a0]
              AST_IDENTIFIER <t.v:5> [0xd6d5a0 -> 0xd6a840] str='$memwr$\a$t.v:5$1_EN'
              AST_CONSTANT <t.v:5> [0xd6d6a0] bits='1'(1) range=[0:0] int=1
      AST_ASSIGN <t.v:7> [0xd6bdf0]
        AST_IDENTIFIER <t.v:7> [0xd6b9f0 -> 0xd66280] str='\o'
        AST_MEMRD <t.v:7> [0xd6bcf0 -> 0xd66870] str='\a'
          AST_CONSTANT <t.v:7> [0xd6ddd0] bits='00000000000000000000000000000000'(32) signed range=[31:0]
      AST_WIRE <t.v:5> [0xd69f00] str='$memwr$\a$t.v:5$1_ADDR' range=[31:0]
        AST_RANGE <t.v:5> [0xd69e00] range=[31:0]
          AST_CONSTANT <t.v:5> [0xd69ba0] bits='00000000000000000000000000011111'(32) signed range=[31:0] int=31
          AST_CONSTANT <t.v:5> [0xd69cd0] bits='00000000000000000000000000000000'(32) signed range=[31:0]
      AST_WIRE <t.v:5> [0xd6a390] str='$memwr$\a$t.v:5$1_DATA' range=[0:0]
        AST_RANGE <t.v:5> [0xd6a290] range=[0:0]
          AST_CONSTANT <t.v:5> [0xd6a030] bits='00000000000000000000000000000000'(32) signed range=[31:0]
          AST_CONSTANT <t.v:5> [0xd6a160] bits='00000000000000000000000000000000'(32) signed range=[31:0]
      AST_WIRE <t.v:5> [0xd6a840] str='$memwr$\a$t.v:5$1_EN' range=[0:0]
        AST_RANGE <t.v:5> [0xd6a740] range=[0:0]
          AST_CONSTANT <t.v:5> [0xd6a4c0] bits='00000000000000000000000000000000'(32) signed range=[31:0]
          AST_CONSTANT <t.v:5> [0xd6a610] bits='00000000000000000000000000000000'(32) signed range=[31:0]
      AST_MEMWR <t.v:5> [0xd6cae0 -> 0xd66870] str='\a'
        AST_IDENTIFIER <t.v:5> [0xd6cbe0 -> 0xd69f00] str='$memwr$\a$t.v:5$1_ADDR'
        AST_IDENTIFIER <t.v:5> [0xd6cce0 -> 0xd6a390] str='$memwr$\a$t.v:5$1_DATA'
        AST_IDENTIFIER <t.v:5> [0xd6cde0 -> 0xd6a840] str='$memwr$\a$t.v:5$1_EN'
--- END OF AST DUMP ---
```

The '$memwr$\a$t.v:5$1_ADDR' signal is driven by an AST_ASSIGN_LE within an AST_BLOCK, yet it is not marked as 'reg'.
Are these really registers? just ports for the memory? what should be the criteria for disabling the check on those _ADDR, _DATA, _EN signals? (I don't want to look at the signal names)

ZipCPU:
In this case, the address in a[0] (i.e. the zero) is a constant.  An inferred wire will work just as well to represent it.  Indeed, I might argue that any inferred constant value should be represented by a "wire".

I think the question you want to answer first is whether or not your patch handles reg [0:0] a; properly, and only then start thinking about handling reg a [0:0];

Dan

cliffordwolf:
First: Besides the problem that it does not work in the cases you pointed out, the code looks good and I'd be happy to merge something like that.

> Do you really want support for more strict error checking, or is the current situation really what you intended to be (where no error checking is done)?

I'm generally not a big fan of adding a lot of complexity to the Verilog front-end for the sake of checking language errors. Simply because it's already a huge mess and checks usually rely on passing information between different components of the front-end and this either ends up being a hack or would require some refactoring. So if someone want's to invest a lot of time then I'd be in favor of refactoring a big chunk of the codebase first.

> Distinguishing between 1364-1995, 1364-2001, 1364-2005, 1800-2009, 1800-2012 ?
I remember from a past PR you actually removed a check that enabled a SV feature only when the SV flag was used. What you wrote at the time was that as long as any SV extension did not have the potential to misinterpret legacy Verilog code, you enable it even when no -sv flag was used.
Personally, I think we should add separate flags for the above, and I don't mind contibuting PRs towards this goal. At least that's what other tools do (Verilator & gcc '-std=' flag).

What would the different behavior triggered by those flags be, besides enabling/disabling keywords (and maybe the subtle $readmem difference between 1364-2001 and 1364-2005)? If it's just that then this sounds simple and unintrusive enough. (Maybe we could then also switch to full support for all language features as new default. Adding -sv all the time is getting old.. :)

mithro:
I would like to add support for having `-std=xxxx` type modes for Yosys which enforced a particular standard in the same way that gcc does. It would be useful to be able to explicitly state which version of Verilog you are using in a project. GCC also has `-std=gnu11` for when you want C++11 plus GNU extensions and similar. You can also have `-std=latest` as default if you want everything supported by default.

udif:
I have finally took the time to update this PR.
The code should now handle all reg/wire/logic uses as found in the tests, plus an additional SystemVerilog test was added.
Also I've turned the original error into warning (at least for the time being). The intention is to warn, but not break existing code, even if the code itself is broken.

The way I solved my original issues was to add another boolean field to the AstNode class, named 'was checked'. The purpose of this field is to run the reg/wire tests only on the original AST , before any memory and assertion changes are applied on it. Any new node created by those code paths is marked with 'was_checked', and so they are ignored.
Unfortunately, this has resulted in driving the 'was_checked' field in many places. I would have preferred a more centralized solution, but so far this is what I came up with.

udif:
@cliffordwolf any issues with this PR?

cliffordwolf:
I like that it is only a warning now. This drastically reduces the risks involved with making such a change.

PR #518, opened 2018-03-27
passes/hierarchy: Reduce code duplication in expand_module

xerpi:
This also makes it easier to add new file extensions support.

Signed-off-by: Sergi Granell <xerpi.g.12@gmail.com>


cliffordwolf:
Jfyi: changed `static const std::map<std::string, std::string> extensions_map` to `static const vector<pair<string, string>> extensions_list` in ee3c12d.

PR #522, opened 2018-04-01
Fixed broken Quartus backend on dffeas init value, and other updates.

c60k28:
## Why this pull request?
As per the recent changes in **__synth_intel.cc__** regarding the power up initialization of DFFEAS primitive, the missing parameter in the techlibs was making Quartus mapper fails. The generated netlist was incorrect.
I did corrected all the dffeas primitives in all the Intel families, but __tested only on Cyclone V__ since is the only hardware I have from such FPGA vendor. But also tried to compile the same working design after and before the changes and all of them are passing fine Quartus map.

## What else the commit does contain?
* Achronix Speedster organization. Should work fine now.
* VPR implementation on Intel devices. For experiments and hopefully someone else could use it as well.
* Fixed Cyclone V lut functions map. The Yosys netlist can generate a working SOF file fur such device now.

This is the first commit from myself. Please do let me know if there is anything else needed in order to add this commit to the head. Also please take into account that, without Intel changes, Yosys-Intel implementation is broken right now.

---
Signed by: Viacheslav T <c60k28 [at] yandex [dot] com

### DIFF
--- a/techlibs/achronix/Makefile.inc
+++ b/techlibs/achronix/Makefile.inc
@@ -1,5 +1,5 @@
 
-OBJS += techlibs/achronix/synth_speedster.o
+OBJS += techlibs/achronix/synth_achronix.o
 
 $(eval $(call add_share_file,share/achronix/speedster22i/,techlibs/achronix/speedster22i/cells_sim.v))
 $(eval $(call add_share_file,share/achronix/speedster22i/,techlibs/achronix/speedster22i/cells_map.v))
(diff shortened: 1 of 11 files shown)


cliffordwolf:
Wow, that's a bit much in one PR. Next time please try splitting it in smaller PRs that can be reviewed (and discussed, if necessary) separately.

That being said: Thanks for looking into this! The intel fpga support and achronix support is currently un-maintained. Do you plan on spending more time on this? Bram inference is currently broken for Intel FPGAs (in a pretty obvious manner) and afaict never actually worked. I'd be happy to answer questions if you want to spend time looking into those issues.

PR #531, opened 2018-04-08
Enable parallel jobs by default

mithro:


mithro:
I found this when doing a clean up.

cliffordwolf:
You should run make with something like `make -j$(nproc)` if you want to utilize all cores. See for example [icestorm build instructions](http://www.clifford.at/icestorm/):

image

PR #534, opened 2018-04-16
support for multiple -liberty opt in stat cmd

edcote:
I don't expect this PR to be merged AS-IS, but I can help make modifications if necessary.

The stat command can only take a single .lib file for purpose of generating chip area #.

Does your API for parsing options work for varargs options?

Can you suggest a better implementation than the hack I have here?

### DIFF
--- a/passes/cmds/stat.cc
+++ b/passes/cmds/stat.cc
@@ -253,6 +253,19 @@ struct StatPass : public Pass {
 				read_liberty_cellarea(cell_area, liberty_file);
 				continue;
 			}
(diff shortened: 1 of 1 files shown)


cliffordwolf:
What's wrong with simply using `-liberty <file1> -liberty <file2> ..`?

edcote:
Great question.  I think I tried it quick and it didn't work?  I will try again later today.

cliffordwolf:
So I guess `-liberty <file1> -liberty <file2> ..` worked. :) Closing this issue.

PR #549, opened 2018-05-16
Move strip -S option to STRIPFLAGS

doctaweeks:
Symbol stripping could be disabled by calling make STRIP=, except the
dangling "-S" option causes make errors. This moves the option to a
variable that can also be overridden, allowing stripping to be disabled
without errors.


doctaweeks:
> simply use something like STRIP=true

Ah, of course! I hadn't thought of this. Thank for the tip.

PR #560, opened 2018-06-04
Support SV implicit named ports and C-type array sizes.

PeterCrozier:
Two trivial SystemVerilog syntax enhancements

* implicit named port connections in a module instantiation where the names are the same.
* C-style array sizes e.g. `reg [7:0] mem_array[1024]`

There is a simple test but this required changes to `autotest.sh` to support SV files with names ending in `.sv`.

### DIFF
--- a/frontends/verilog/verilog_parser.y
+++ b/frontends/verilog/verilog_parser.y
@@ -121,6 +121,7 @@ static void free_attr(std::map<std::string, AstNode*> *al)
 
 %type <ast> range range_or_multirange  non_opt_range non_opt_multirange range_or_signed_int
 %type <ast> wire_type expr basic_expr concat_list rvalue lvalue lvalue_concat_list
+%type <ast> named_port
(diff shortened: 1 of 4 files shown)


whitequark:
Do you think you can split this PR into two, one per feature? This would greatly simplify review and integration.

tux3:
I'm also interested in named implicit port connections, so I reused this code (tux3@d6fed8f).

@PeterCrozier @whitequark Since this PR doesn't seem to be moving forward, is there any objection if I send one with just the named ports feature?

whitequark:
@tux3 Sure, your commit looks self-contained to me, so that's good.

towoe:
@PeterCrozier @whitequark I reused the rest of this PR in https://github.com/YosysHQ/yosys/pull/1107

cliffordwolf:
Thanks @towoe. Closing this one now.

PR #562, opened 2018-06-06
Detect illegal port declaration, e.g input/output/inout keyword must …

udif:
This PR is tightening the verilog parser around port definition.

Currently the following invalid code is accepted:
```
module x(wire input i);
endmodule
```
or
```
module y(i);
wire input i;
endmodule
```

These code snippets should be illegal, and in addition I hope will help future parser enhancements (e.g. for typedefs).
From my past experience any effort to touch the wire/reg declaration rules for typedefs, where the reg/wire keyword is replaced by a simple TOK_ID triggers ~15 shift/reduce or reduce/reduce conflicts.

### DIFF
--- a/frontends/verilog/verilog_parser.y
+++ b/frontends/verilog/verilog_parser.y
@@ -376,9 +376,10 @@ wire_type:
 	};
 
 wire_type_token_list:
-	wire_type_token | wire_type_token_list wire_type_token;
+	wire_type_token | wire_type_token_list wire_type_token |
+	wire_type_token_io ;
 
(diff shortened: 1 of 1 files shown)


udif:
@cliffordwolf any comments on this ?

cliffordwolf:
This looks like `input wire` or `output integer` would be also illegal after this patch is applied. However, those are perfectly legal port declarations. From IEEE Std 1364-2005:

image

cliffordwolf:
scratch that. I think I misread the patch..

PR #573, opened 2018-06-26
Add support for 64-bit builds using msys2 environment.

cr1901:
At long last a cascade of problems has forced me to dump the hybrid 32/64-bit `msys2` environment I have and go completely 64-bit. Since I never compiled `yosys` for 64-bit on Windows before today, I added a corresponding config.

`yosys` and `abc` both build fine with the changes I've made: `Yosys 0.7+606 (git sha1 b4c9dcc2, x86_64-w64-mingw32-g++ 7.3.0 -Os)`


cr1901:
Please do not accept yet. Upon further testing, I managed to break `abc`/pass it the wrong options. I'm getting 32-bit/64-bit dll mismatch. I need to investigate why this is (`pthread` related).

cr1901:
Okay, I've fixed compilation, and expanded the scope of the PR. To clarify:

* The previous `msys2` builds needed a user to copy `pthreadVC2.dll` to the binary directory where `yosys` lives (only once though). Instead, we now use the `msys2`-provided `libpthread` which is more up to date anyway than the copy `abc` provides.
* Add a 64-bit `msys2` build, so a user has the option to choose which version of `yosys` they want (depending on what compilers are installed).

I've personally verified this on Windows 7 (all default `yosys` options enabled) and Ubuntu cross-compiling to Windows (`ENABLE_ABC` and `ENABLE_COVER` only). Which means, I compiled and then ran `yosys` without arguments and `yosys-abc -h` and there were no complaints of missing DLLs :).

PR #578, opened 2018-07-07
show: Add -nobg argument.

mithro:
Makes yosys wait for the viewer command to finish before continuing.


mwkmwkmwk:
Had to do non-trivial conflict resolution because of modifications in the meantime.

PR #584, opened 2018-07-19
Provide source-location logging.

hzeller:
Provide a standardized warning/error logging which makes it much easier to develop verilog as it allows to step through the messages with IDEs, emacs or vi as they all know the `<filename>:<line>:<msg>` format that is also standardized in compilers such as gcc or clang.

I intend to survey all the places where we have a filename and line-number available in future pull requests, but for now just the base-work and a proof-of-concept place in genrtlil.

o Provide log_file_warning() and log_file_error() that prefix the log
  message with `<filename>:<lineno>:` to be easily picked up by IDEs that
  need to step through errors.
o Simplify some duplicate logging code in kernel/log.cc
o Use the new log functions in genrtlil.


cliffordwolf:
Looks good to me. If I merge this, are you going to provide further PRs for switching all the other messages in the verilog/ast front-ends to this new API?

hzeller:
Yes, I already have a change pending for that.

PR #591, opened 2018-07-21
Consistent use of 'override' for virtual methods in derived classes.

hzeller:
- Not all derived methods were marked 'override', but it is a great
  feature of C++11 that we should make use of.
 - While at it: touched header files got a `-*- c++ -*-` for emacs to
  provide support for that language.
 - use YS_OVERRIDE for all override keywords (though we should probably
  use the plain keyword going forward now that C++11 is established)

Mostly a suggestion, but I think it would make working with the code easier as it provides automatic typo-detection in derived classes.

Note, this does not contain any functional changes, just mechanical text changes.

I think we should use the plain `override` keyword and remove the `#define YS_OVERRIDE` (and YS_FINAL); it looks like these defines were necessary in the transition to full c++11 support in compilers, but now that these compilers are established, simplifying that would probably be good. Your call, I am happy to do a follow-up pull-request that converts all `YS_OVERRIDE` -> `override` and remove the `#define`.

This change does not touch any overrides in `libs/` as these seem to be mostly upstream-imported from somewhere.

### DIFF
--- a/CodingReadme
+++ b/CodingReadme
@@ -21,7 +21,7 @@ Here is a the C++ code for a "hello_world" Yosys command (hello.cc):
 
 	struct HelloWorldPass : public Pass {
 		HelloWorldPass() : Pass("hello_world") { }
-		virtual void execute(vector<string>, Design*) {
+		void execute(vector<string>, Design*) override {
 			log("Hello World!\n");
 		}
 	} HelloWorldPass;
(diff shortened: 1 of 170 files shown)


cliffordwolf:
> I think we should use the plain override keyword and remove the #define YS_OVERRIDE

SGTM, but only *after* the next release. (You wouldn't believe how old compilers are on the typical "EDA workstation". RHEL 6 is probably still the most wide-spread distribution used for those machines.)

PR #593, opened 2018-07-23
Improve debugging Yosys with pretty printers and better default settings.

mithro:
Adding GDB pretty printer for IdString. Now when you print a structure (such as a `cell`) in gdb you get something much more readable.

 * Script to pretty print the IdString primitive, making it *much* easier to use GDB with Yosys.

 * Install the gdb scripts to the correct location so that they can be autoloaded.

 * Add script entries so the pretty printers are automatically loaded.

Example usage;
```shell
make ENABLE_DEBUG=1
gdb \
 -iex "add-auto-load-safe-path $PWD/misc/gdb" \
 --args ./yosys
```
(If Yosys is installed, the `add-auto-load-safe-path` is not needed.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -318,7 +318,11 @@ ifeq ($(ENABLE_DEBUG),1)
 ifeq ($(CONFIG),clang)
 CXXFLAGS := -O0 -DDEBUG $(filter-out -Os,$(CXXFLAGS))
 else
+ifeq ($(CONFIG),gcc)
+CXXFLAGS := -O0 -ggdb3 -fvar-tracking -fvar-tracking-assignments -DDEBUG $(filter-out -Os -ggdb,$(CXXFLAGS))
+else
 CXXFLAGS := -Og -DDEBUG $(filter-out -Os,$(CXXFLAGS))
+endif
 endif
 endif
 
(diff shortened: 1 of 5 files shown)


daveshah1:
Seems to work well for me. Thanks!

boqwxp:
What's blocking this?  An agreement on where exactly some code should go?  I think this is really useful for Yosys development/debugging and should get merged sooner rather than later regardless of where the code goes, exactly.

whitequark:
@boqwxp I agree with you that the kind of functionality introduced by this PR is quite useful. I also agree with @cliffordwolf that this PR has a lot more moving parts than it needs to, and seems quite fragile.

I think the way forward would be someone simplifying this PR and testing it on various configurations (Linux gcc/clang, MinGW, MSVC). I'll be happy to review it or assist with the GDB bits. I don't expect that I'll have time to do it myself any time soon, though I might get around to it at some point.

mithro:
I'm afraid I won't have any time to work on the pull request so would be very supportive of someone taking it over and getting it landed.

nakengelhardt:
Anyone not a maintainer taking this over would have to create a new branch anyway so I think we can just close this for now and open a new PR when someone gets around to rewriting it...

Issue #565, opened 2018-06-08
Yosys isn't able to import it's own eblif output format?


mithro:
## Steps to reproduce the issue

 * Generate an eblif with `.cname` or `.conn` values in the output
 * Try and import the file into yosys with `read_blif`
 * Get `Syntax Error` message.

## Expected behavior

 * Be able to read the eblif file into yosys (for simulation / equivalence checking).

litghost:
This can be closed with #596 pulled.

mithro:
BTW Do we have any tests for "round tripping via blif"?

mithro:
Happy to close this issue and open a seperate one for that....

PR #596, opened 2018-08-02
#565 Add BLIF parsing support for .conn and .cname

litghost:
Implements .cname and .conn per http://vtr-verilog-to-routing.readthedocs.io/en/latest/vpr/file_formats.html#conn

### DIFF
--- a/frontends/blif/blifparse.cc
+++ b/frontends/blif/blifparse.cc
@@ -83,6 +83,7 @@ void parse_blif(RTLIL::Design *design, std::istream &f, std::string dff_name, bo
 	RTLIL::Module *module = nullptr;
 	RTLIL::Const *lutptr = NULL;
 	RTLIL::Cell *sopcell = NULL;
+	RTLIL::Cell *lastcell = nullptr;
 	RTLIL::State lut_default_state = RTLIL::State::Sx;
 	int blif_maxnum = 0, sopmode = -1;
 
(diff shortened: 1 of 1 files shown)


frontends/blif/blifparse.cc
```diff
@@ -281,12 +300,16 @@ void parse_blif(RTLIL::Design *design, std::istream &f, std::string dff_name, bo
 						const_v.bits[i] = v[n-i-1] != '0' ? State::S1 : State::S0;
 				}
 				if (!strcmp(cmd, ".attr")) {
-					if (obj_attributes == nullptr)
+					if (obj_attributes == nullptr) {
+						log_error("No object to attach .attr too.\n");
 						goto error;
```
cliffordwolf: This doesn't make a lot of sense. `log_error` does not return, so the `goto` is never reached and the user is never told in which line the error occurred. (same issue with parameters below)

litghost: Indeed, switched to log_warning prior to jumping to the common error message.

cliffordwolf:
But it's not a warning! It is part of an error. A correct solution would be something like the following.

```
    const char *err_reason = nullptr;
...
    if (...) {
        err_reason = "No object to attach .attr too.";
        goto error_with_reason;
    }
...
error:
    log_error("Syntax error in line %d!\n", line_count);
error_with_reason:
    log_error("Syntax error in line %d: %s\n", line_count, err_reason);
```

cliffordwolf:
jfyi: 93efbd5

PR #634, opened 2018-09-19
Fix for ilang_lexer

lvcargnini:
Just found an error compiling Yosys, trying to fix a previous error on SystemVerilog source code.
So found this error on ilang_lexer, fixed and pushing the fix.


cliffordwolf:
As you can see from Travis CI, this would actually break the build. The correct file name is `.hh` since commit ddc1761f1. Maybe you have some old files in your build envirenmoent? I'd recommend trying to re-build with a fresh rebuild.

Issue #639, opened 2018-09-26
Problem about equivalence check


tklam:
Hi, I have two Verilog files:
### old.v
``` verilog
module top (
    input a,
    input b,
    output [1:0] z);
wire n1, n2;

middle M1(.a(a), .b(b), .c(n1));
assign n2 = ~n1;
assign z[0] = n2;

assign z[1] = a & b;
endmodule

module middle (
    input a,
    input b,
    output c);
assign c = a | b;
endmodule
```
### new.v
``` verilog
module top (
    input a,
    input b,
    output [1:0] z);
wire n1, n2;

middle M1(.a(a), .b(b), .c(n1));
assign n2 = n1;  // z[0] in new.v is different from that in old.v due to this line
assign z[0] = n2;

assign z[1] = a | b;
endmodule

module middle (
    input a,
    input b,
    output c);
assign c = a | b;
endmodule
```
Then, I used the following script to check their equivalence. The netlists were not flattened on purpose.
```
read_verilog -ignore_redef -nooverwrite  old.v
hierarchy -check -top top
rename top netlist_v1
hierarchy -check -top netlist_v1
prep
splitnets -ports
design -save netlist_v1

read_verilog -ignore_redef -overwrite  new.v
hierarchy -check -top top
rename top netlist_v2
hierarchy -check -top netlist_v2
prep
splitnets -ports
design -save netlist_v2

design -copy-from netlist_v1 -as netlist_old netlist_v1
design -copy-from netlist_v2 -as netlist_new netlist_v2

equiv_make -inames netlist_old netlist_new miter_netlist
equiv_simple -undef -seq 10
equiv_induct -undef -seq 10
equiv_status
```

## Expected behavior

z[0] in old.v and new.v should be different.

## Actual behavior

But equiv_simple reported that they are equivalent.

I would like to know whether this is a bug. Thanks a lot.

tklam:
Just created a pull request to fix this issue. I hope the fix is useful. Thanks.

tklam:
In the function check_signal_in_fanout(), I did not check the cell type before going into the next level of recursion. Should I check it to avoid flip flop loops?

PR #640, opened 2018-09-26
Fix issue #639

tklam:
I found that Yosys will create a combinational loop when running equiv_make given my example files. Some functions have been added to check whether a net is in the fanout cone of another net before committing the net substitution. Many thanks.

### DIFF
--- a/passes/equiv/equiv_make.cc
+++ b/passes/equiv/equiv_make.cc
@@ -40,6 +40,8 @@ struct EquivMakeWorker
 	pool<SigBit> undriven_bits;
 	SigMap assign_map;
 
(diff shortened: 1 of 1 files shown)


tklam:
Found a bug. Let me close this and open another. My apologies.

cliffordwolf:
Sure. :) Is the bug the FFs thing? Jfyi, you might want to use `CellTypes` to limit the fanout to internal combinatorical-only cells:

```
// once during initialization
CellTypes comb_ct;
comb_ct.setup_internals();
comb_ct.setup_stdcells();

// when checking a cell
bool is_comb = comb_ct.cell_known(cell->type);
```

PR #641, opened 2018-09-26
Fix issue #639

tklam:
I found that Yosys will create a combinational loop when running equiv_make given my example files. Some functions have been added to check whether a net is in the fanout cone of another net before committing the net substitution. Many thanks.


cliffordwolf:
Ah, that was the bug. :) Should we stop looking further at FFs? Otherwise every signal will be in every others signal fanout cone for nontrivial designs.

tklam:
Hi. I am sorry for late response.
Ya, that was the bug caused by one char :p
Now check_signal_in_fanout() checks whether a cell is comb or not.

PR #648, opened 2018-10-02
Evaluate testcases in parallel

msgctl:
Signed-off-by: Mateusz Zalega <mateusz@appliedsourcery.com>

### DIFF
--- a/tests/asicworld/run-test.sh
+++ b/tests/asicworld/run-test.sh
@@ -11,4 +11,4 @@ do
 done
 shift "$((OPTIND-1))"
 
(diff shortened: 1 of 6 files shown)


cliffordwolf:
If you run `make -jN test` it should already run all those "sub-makes" with `-jN` via `MAKEFLAGS`. There shouldn't be any need to pass `-jN` explicitly to the sub-makes.

Actually it might even hurt to have that in there because we don't want to overwrite the number of jobs the user has specified on the top-level make. (I'm not sure how specifying an explicit `-j` option in a sub-make affects the make job-server mechanism.)

msgctl:
@cliffordwolf OK, I didn't see that 'test' target in the top-level makefile.

Issue #370, opened 2017-08-03
single bit for loop increments


heiner-bauer:
While elaborating the Verilog module below, Yosys rapidly hogs memory (multiple GB) and does not seem to finish. If the counter increment is changed to the examples in the comment, everything works fine. Although it can be easily avoided, I assume `i = i + 1'b1` is still valid syntax and hence should be supported by Yosys.
``` Verilog
module yosys_loop_increment(trigger_i, data_o);

   input trigger_i;
   output [7:0] data_o;

    parameter DEPTH = 5;
   integer i;

   always @(*) begin
      data_o = 8'b0;
      if(trigger_i) begin
         // working: i = i + 1 or i = i + 2'b1 or i = i + 32'b1 or i = i + 'b1
         for(i = 0; i < DEPTH; i = i + 1'b1) begin
            data_o = data_o + i;
         end
      end
   end
          
endmodule
```

cliffordwolf:
The problem here is that for some reason 1'b1 is sign extended to a 32 bit -1, and then the whole loop is counting down instead of up. I have no idea why this happens. I'll investigate..

cliffordwolf:
Afaict #952 fixes this issue.

PR #649, opened 2018-10-03
Don't apply result sign attribute to operands.

tomverbeure:
This is potential fix for issue #370.

When expanding the operands of the `i = i + 1'b` addition, the code uses the sign hint of the result instead of the sign of the operand. Since `i` is a signed integer, it will expand `1'b` as a signed number as well, resulting in -1 instead of +1.

The fix uses the sign of the operands as sign hint for the bitsAsConst method. 

I also use the sign of the 2 operands as a parameters of `const_func`. So the only place were the `sign_hint` is still used, is in the final `mkconst_bits`.

Tom


cliffordwolf:
> the code uses the sign hint of the result [..]

But that's not what `sign_hint` should be. It should be the sign of the expression.

Can you provide an MCVE that's not using a loop but just e.g. localparams?

cliffordwolf:
I think #952 is the correct fix for the issue.

cliffordwolf:
Merged #952, closing this one.

PR #651, opened 2018-10-04
Fix IdString M in setup_stdcells()

ARandomOWL:
Only affects `$_MUX16_`


cliffordwolf:
Hahaha, what a stupid bug. Thanks for fixing it!

PR #659, opened 2018-10-11
Support for SystemVerilog interfaces and modports

rubund:
Hi Clifford,

This time doing the changes mostly in AST before RTLIL generation.

It seems to work correctly for several test cases. Also the normal yosys test suite passes.

It is a bit hard to add test cases in the normal way since iverilog does not support this yet.

### DIFF
--- a/frontends/ast/ast.cc
+++ b/frontends/ast/ast.cc
@@ -905,7 +905,7 @@ RTLIL::Const AstNode::realAsConst(int width)
 // create a new AstModule from an AST_MODULE AST node
 static AstModule* process_module(AstNode *ast, bool defer)
 {
-	log_assert(ast->type == AST_MODULE);
+	log_assert(ast->type == AST_MODULE || ast->type == AST_INTERFACE);
 
 	if (defer)
 		log("Storing AST representation for module `%s'.\n", ast->str.c_str());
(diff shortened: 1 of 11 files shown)


frontends/ast/ast.cc
```diff
@@ -1083,8 +1086,122 @@ AstModule::~AstModule()
...
+		for(auto &intf : interfaces) {
+			RTLIL::Module * intfmodule = intf.second;
+			std::string intfname = log_id(intf.first);
+			intfname = "\\" + intfname;
+			for (auto &wire_it : intfmodule->wires_){
+				AstNode *wire = new AstNode(AST_WIRE, new AstNode(AST_RANGE, AstNode::mkconst_int(wire_it.second->width -1, true), AstNode::mkconst_int(0, true)));
+				std::string newname = log_id(wire_it.first);
+				newname = intfname + "." + newname;
+				wire->str = newname;
+				wire->is_input = true;
```
rubund: This will change later when support for modports is added. Then the modport will decide whether it is input or output.

rubund:
Also modport support has been added now to this branch. It works surprisingly well. I am going to write a few more test cases.

didjcodt:
Just tested on my laptop, seems to be working perfectly on my use case :)
I will do other tests in a few days.

rubund:
@didjcodt Great to hear! It is perfect to get some assistance when it comes to testing.
Note one thing. There is still not support for having the interface as a **port** in the **top module**. I will probably add support for that soon, but it can come in a later pull request.

Let me know if you see any problems.

PR #664, opened 2018-10-16
Ignore protect endprotect

tklam:
Modify verilog_lexer.l just to ignore the {protect, endprotect} markers.

### DIFF
--- a/frontends/verilog/verilog_lexer.l
+++ b/frontends/verilog/verilog_lexer.l
@@ -135,6 +135,9 @@ YOSYS_NAMESPACE_END
 		frontend_verilog_yyerror("Unsupported default nettype: %s", p);
 }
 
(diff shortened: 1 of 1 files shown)


cliffordwolf:
Where are those two compiler directives defined? (Which standard?) I only know of ``` `pragma protect ...``` (which is also not handled by Yosys in any way right now.)

tklam:
I found them in Cadence's documentation. For example on page 329 of this [PDF](http://mspic.ee.nchu.edu.tw/class_course/university/97_VLSI-design/handout/2_2%20Cadence%20Verilog%20Languaje%20and%20Simulation%20Course.pdf).

cliffordwolf:
Interesting and good to know. Doesn't seem to be a Verilog/SystemVerilog std feature. But I see no harm in ignoring those two directives.

PR #674, opened 2018-10-20
Support for SystemVerilog interfaces as ports in the top level module + test case

rubund:


cliffordwolf:
This PR introduced a whole bunch of additional attributes. Can we have another PR that adds short descriptions of those attributes to the attributes section in the README file? @rubund Thanks!

Also: How hard would it be to extend this to also support SV structs?

Issue #675, opened 2018-10-20
equiv_make breaking for multibit in 0.8


rafaeltp:
## Steps to reproduce the issue

input.v
```
module graphtest(A, Y, Z);
  input [3:0] A;
  output [2:0] Y;
  output [1:0] Z;
  assign Y = { A[2:0] & 3'h2 };
  assign Z = A[2:1];
endmodule
```

script:
```
yosys -p "read_verilog graphtest.v; rename -top gold; 
design -stash gold; 
read_verilog graphtest.v; 
rename -top gate; design -stash gate; 
design -copy-from gold -as gold gold; 
design -copy-from gate -as gate gate; 
equiv_make gold gate equiv"
```

## Expected behavior

Equiv circuit generated.

## Actual behavior

Assertion ./kernel/rtlil.h:1307 fails.

log bellow:

```

 /----------------------------------------------------------------------------\
 |                                                                            |
 |  yosys -- Yosys Open SYnthesis Suite                                       |
 |                                                                            |
 |  Copyright (C) 2012 - 2018  Clifford Wolf <clifford@clifford.at>           |
 |                                                                            |
 |  Permission to use, copy, modify, and/or distribute this software for any  |
 |  purpose with or without fee is hereby granted, provided that the above    |
 |  copyright notice and this permission notice appear in all copies.         |
 |                                                                            |
 |  THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES  |
 |  WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF          |
 |  MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR   |
 |  ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES    |
 |  WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN     |
 |  ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF   |
 |  OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.            |
 |                                                                            |
 \----------------------------------------------------------------------------/

 Yosys 0.8+17 (git sha1 11c8a9eb, clang 7.0.0 -fPIC -Os)


[...]

PR #676, opened 2018-10-21
Splits SigSpec into bits before calling check_signal_in_fanout (solves #675)

rafaeltp:
Splits SigSpec into bits before calling check_signal_in_fanout in the equiv_make pass.

### DIFF
--- a/passes/equiv/equiv_make.cc
+++ b/passes/equiv/equiv_make.cc
@@ -298,20 +298,21 @@ struct EquivMakeWorker
 				SigSpec new_sig = rd_signal_map(old_sig);
 
 				if(old_sig != new_sig) {
-					for (auto &old_bit : old_sig.bits()) {
-						SigBit new_bit = new_sig.bits()[old_bit.offset];
+					SigSpec tmp_sig = old_sig;
+					for (int i = 0; i < GetSize(old_sig); i++) {
+						SigBit old_bit = old_sig[i], new_bit = new_sig[i];
 
 						visited_cells.clear();
-						if (old_bit != new_bit) {
-							if (check_signal_in_fanout(visited_cells, old_bit, new_bit))
-								continue;
+						if (check_signal_in_fanout(visited_cells, old_bit, new_bit))
+							continue;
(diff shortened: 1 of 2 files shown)


passes/equiv/equiv_make.cc
```diff
@@ -290,22 +290,28 @@ struct EquivMakeWorker
...
+
+						visited_cells.clear();
+						if (old_bit != new_bit) {
+							if (check_signal_in_fanout(visited_cells, old_bit, new_bit))
+								continue;
+
+							log("Changing input %s of cell %s (%s): %s -> %s\n",
+									log_id(conn.first), log_id(c), log_id(c->type),
+									log_signal(old_bit), log_signal(new_bit));
+							c->setPort(conn.first, new_bit);
```
cliffordwolf: This also doesn't look right. Shouldn't this only update a single bit in that cell port?

passes/equiv/equiv_make.cc
```diff
@@ -290,22 +290,28 @@ struct EquivMakeWorker
...
-					if (check_signal_in_fanout(visited_cells, old_sig, new_sig)) 
-						continue;
 
-					log("Changing input %s of cell %s (%s): %s -> %s\n",
-							log_id(conn.first), log_id(c), log_id(c->type),
-							log_signal(old_sig), log_signal(new_sig));
-					c->setPort(conn.first, new_sig);
+				if(old_sig != new_sig) {
+					for (auto &old_bit : old_sig.bits()) {
+						SigBit new_bit = new_sig.bits()[old_bit.offset];
```
cliffordwolf: This doesn't look right. There is no guarantee that the N'th bit olf old_sig has `old_bit.offset=N`.

I think you want something like

```
for (int i = 0; i < GetSize(old_sig); i++) {
    SigBit old_bit = old_sig[i], new_bit = new_sig[i];
```

rafaeltp:
Thanks for the comments, I have addressed both, let me know if there's still something you'd think should be changed.

PR #679, opened 2018-10-24
More meaningful SystemVerilog/Verilog parser error messages

udif:
Currently, the Verilog/SystemVerilog parser has about 14 cryptic "Syntax error" messages.
This PR renames thosee generic "Syntax error" message from into unique, meaningful info on the error.
Also add 13 compilation examples that triggers each of these messages.

### DIFF
--- a/frontends/verilog/verilog_parser.y
+++ b/frontends/verilog/verilog_parser.y
@@ -305,7 +305,7 @@ module_arg_opt_assignment:
 			else
 				ast_stack.back()->children.push_back(new AstNode(AST_ASSIGN, wire, $2));
 		} else
-			frontend_verilog_yyerror("Syntax error.");
+			frontend_verilog_yyerror("SystemVerilog interface in module port list cannot have a default value.");
 	} |
 	/* empty */;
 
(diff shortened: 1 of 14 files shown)


frontends/verilog/verilog_parser.y
```diff
@@ -1068,15 +1068,15 @@ wire_name_and_opt_assign:
 wire_name:
 	TOK_ID range_or_multirange {
 		if (astbuf1 == nullptr)
-			frontend_verilog_yyerror("Syntax error.");
+			frontend_verilog_yyerror("Internal error - should not happen - no AST_WIRE node.");
```
udif: I was not able to trigger this error :-(

frontends/verilog/verilog_parser.y
```diff
@@ -883,7 +883,7 @@ param_signed:
 param_integer:
 	TOK_INTEGER {
 		if (astbuf1->children.size() != 1)
-			frontend_verilog_yyerror("Syntax error.");
+			frontend_verilog_yyerror("Internal error in param_integer - should not happen?");
```
udif: I was not able to trigger this error as well, and I believe it should not happen.
I believe it was created as a copy of the equivalent error message on line 886.
As much as these look symmetrical, they are NOT, due to that param_integer is specified before param_real in the param_decl and localparam_decl rules.
As a result, TOK_INTEGER will be parsed *after* TOK_REAL, preventing the condition on line 885.

frontends/verilog/verilog_parser.y
```diff
@@ -1478,7 +1478,7 @@ behavioral_stmt:
 			node->str = *$3;
 	} behavioral_stmt_list TOK_END opt_label {
 		if ($3 != NULL && $7 != NULL && *$3 != *$7)
-			frontend_verilog_yyerror("Syntax error.");
+			frontend_verilog_yyerror("Begin label (%s) and end label (%s) doesn't match.", $3->c_str()+1, $7->c_str()+1);
```
udif: should be "don't"

cliffordwolf: Fixed in 5ab58d4

PR #690, opened 2018-11-03
Change to non-blocking assignments in ice40 SB_RAM40_4K model

olofk:
Verilator complains about INITIALDLY for the memory initialization.
Unlikely to affect anything in practice, but verilator does have a point

### DIFF
--- a/techlibs/ice40/cells_sim.v
+++ b/techlibs/ice40/cells_sim.v
@@ -409,22 +409,22 @@ module SB_RAM40_4K (
 
 	initial begin
 		for (i=0; i<16; i=i+1) begin
-			memory[ 0*16 + i] <= INIT_0[16*i +: 16];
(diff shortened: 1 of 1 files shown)


cliffordwolf:
I get that but with this change Yosys will produce an warning because it can't (yet) be sure that it doesn't miss an interdependency between those assignments. (There is none, but yosys doesn't check that yet and thus produces a warning saying it would ignore any such interdependency.) I will only merge that if you also extend the Yosys Verilog front-end in a way that solve the issue that warning is about.

olofk:
I see. I won't be able to help out with fixing the root cause any time soon :/

Would you be fine with an `ifdef verilator just to shut up the linter or is that too much of a hack?

cliffordwolf:
Yes, ifdef verilator would be fine with me.

olofk:
Superseded by #725

PR #693, opened 2018-11-06
improve rlimit handling in smtio.py

### LINKED ISSUE #692: yosys-smtbmc: smtio.py ->"smtio_stacksize" requirement too high for WSL setups
yosys-smtbmc does not start in my "Windows Subsystem for Linux" environment, because it does not pass the "smtio_stacksize" check in "smtio.py" (line 35). 

The script requires a "smtio_stacksize" of 134217728.  In the WSL environment of my computer (16GB RAM), "resource.getrlimit(resource.RLIMIT_STACK)[0]" only has a value of 8388608. 
(My Linux Computer with 8GB of RAM does not run into this resource limit.)

When I remove the RAM check:
```
#if resource.getrlimit(resource.RLIMIT_STACK)[0] < smtio_stacksize:
#    resource.setrlimit(resource.RLIMIT_STACK, (smtio_stacksize, -1))
```
yosys-smybmc seems to run fine. I have not seen any crashes so far.

Is it possible to reduce the "smtio_stacksize" requirement, so that yosys can be installed on a WSL system without hand edits? Or is this a true memory requirement?

cliffordwolf:
This should fix #692 and cleanup #691.

Please test @hotwolf @arjenroodselaar @mmicko.

### DIFF
--- a/backends/smt2/smtio.py
+++ b/backends/smt2/smtio.py
@@ -31,16 +31,16 @@ from threading import Thread
 # does not run out of stack frames when parsing large expressions
 if os.name == "posix":
     smtio_reclimit = 64 * 1024
-    smtio_stacksize = 128 * 1024 * 1024
-    smtio_stacklimit = resource.RLIM_INFINITY
-    if os.uname().sysname == "Darwin":
-        # MacOS has rather conservative stack limits
-        smtio_stacksize = 16 * 1024 * 1024
-        smtio_stacklimit = resource.getrlimit(resource.RLIMIT_STACK)[1]
     if sys.getrecursionlimit() < smtio_reclimit:
         sys.setrecursionlimit(smtio_reclimit)
-    if resource.getrlimit(resource.RLIMIT_STACK)[0] < smtio_stacksize:
-        resource.setrlimit(resource.RLIMIT_STACK, (smtio_stacksize, smtio_stacklimit))
+
+    smtio_stacksize = 128 * 1024 * 1024
+    current_rlimit_stack = resource.getrlimit(resource.RLIMIT_STACK)
+    if current_rlimit_stack[0] != resource.RLIM_INFINITY:
+        if current_rlimit_stack[1] != resource.RLIM_INFINITY:
+            smtio_stacksize = min(smtio_stacksize, smtio_stacklimit)
+        if current_rlimit_stack[0] < smtio_stacksize:
+            resource.setrlimit(resource.RLIMIT_STACK, (smtio_stacksize, current_rlimit_stack[1]))
 
 
 # currently running solvers (so we can kill them)
--- a/backends/smt2/smtio.py
+++ b/backends/smt2/smtio.py
@@ -34,9 +34,12 @@ if os.name == "posix":
     if sys.getrecursionlimit() < smtio_reclimit:
         sys.setrecursionlimit(smtio_reclimit)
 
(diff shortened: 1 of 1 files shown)


mmicko:
Works under WSL

backends/smt2/smtio.py
```diff
@@ -31,16 +31,16 @@
...
-    if resource.getrlimit(resource.RLIMIT_STACK)[0] < smtio_stacksize:
-        resource.setrlimit(resource.RLIMIT_STACK, (smtio_stacksize, smtio_stacklimit))
+
+    smtio_stacksize = 128 * 1024 * 1024
+    current_rlimit_stack = resource.getrlimit(resource.RLIMIT_STACK)
+    if current_rlimit_stack[0] != resource.RLIM_INFINITY:
+        if current_rlimit_stack[1] != resource.RLIM_INFINITY:
+            smtio_stacksize = min(smtio_stacksize, smtio_stacklimit)
+        if current_rlimit_stack[0] < smtio_stacksize:
+            resource.setrlimit(resource.RLIMIT_STACK, (smtio_stacksize, current_rlimit_stack[1]))
```
mmicko: this does not work on macOS. 
Should not it be:
resource.setrlimit(resource.RLIMIT_STACK, (current_rlimit_stack[0], smtio_stacksize))

cliffordwolf: Hmm.. I don't think so. The 2nd tuple element is the hard limit. You get a `ValueError`?

arjenroodselaar: setrlimit(...) throws a ValueError for anything over 16MB of MacOS. One can run successfully run "ulimit -s $(ulimit -Hs)" though, which puts the stack at 64MB (ulimit -a shows the expected limit). I am not sure where this falls apart, I have not had a chance to look at Python interpreter source so see if this exception is some artificial limit.

arjenroodselaar: I poked at this for a bit and using getrlimit(...)/setrlimit(...) in a C++ program works as expected. You can set the soft limit to the max and afterwards allocate a std::array on the next stack frame with a size close to the hard limit. The CPython does pretty much the same thing (https://github.com/python/cpython/blob/d2b11af91560eaaeb499f702f4b0f244ec756280/Modules/resource.c#L216), but for some reason seems to run into an EPERM errno. Not sure what's going on here.

mmicko: Well others as well have same issue on  macOS with python3 https://github.com/sagemath/cysignals/issues/71  but there is no actual fix for this issue there.

arjenroodselaar: I tried this with py2.7 on MacOS and it works as expected. Thanks Apple :/

cliffordwolf: @mmicko That sage issue is about the message for EINVAL, but @arjenroodselaar says in our case we get EPERM.

Anyway, I've now pushed another commit that limits the stack size we attempt to set to 16 MB on Darwin. Grrr. This is not very satisfying..

mmicko:
Will try out tonight to confirm, but looks there is some issue how it is
handled in python 3.x  that causes trouble

mmicko:
Confirming working fine on macOS

hotwolf:
Sorry, I wasn't fast enough to try out the "rlimit" branch before it was removed. But,I can confirm that my WSL problem is resolved in the current "master" branch.

PR #704, opened 2018-11-19
Using awk rather than gawk

webhat:
Correct me if I'm wrong, and I don't see that the use of this tool is GNU awk specific. Also on linux systems awk is usually a symbolic link to gawk.

*I'm propose this change as it will allow me to build this on OSX with minimum of dependencies*


whitequark:
LGTM.

PR #714, opened 2018-12-04
Proof-of-concept: preserve naming through ABC using dress

daveshah1:
This likely doesn't work for all cases, nor am I sure if it is the correct approach, just trying to work out what is possible.

### DIFF
--- a/passes/techmap/abc.cc
+++ b/passes/techmap/abc.cc
@@ -329,6 +329,18 @@ void extract_cell(RTLIL::Cell *cell, bool keepff)
 
 std::string remap_name(RTLIL::IdString abc_name)
 {
+	std::string abc_sname = abc_name.substr(1);
+	if (abc_sname.substr(0, 5) == "ys__n") {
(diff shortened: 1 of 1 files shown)


cliffordwolf:
lgtm. should we add `-nodress` so the user can disable it if they want?

daveshah1:
Sure, it looks like one of the regression tests is also failing so I need to sort that out too (I think dress might not always be legal to use).

cliffordwolf:
Maybe then let's do `-dress` for now so we can test it more easily without braking anything yet? Later we can use `-dress` or `-nodress` to force one behavior or the other and use some kind of auto-detect by default.

daveshah1:
Seems like something is still broken with `dress` optional probably due to the other renaming stuff I did to track what was coming from Yosys and what was coming from abc. I'll investigate further.

daveshah1:
Nevermind, this is fine now. I simply forgot to actually push the changes that made `dress` optional before...

PR #717, opened 2018-12-05
Add a new opt_lut pass, which combines inefficiently packed LUTs

whitequark:
On iCE40, techmapping `$alu` always results in a `SB_LUT4` cell with only 3 inputs used. This is inefficient and nothing later in the synthesis pipeline is currently able to take advantage of this unused LUT input. This LUT input could be very useful for certain ALU designs, as using it to fold a multiplexer LUT into an ALU LUT in an expression such as `c ? a : a + b` or `c ? 0 : a - b` could result, in extreme cases, a 2x reduction in LUT count.

The new `opt_lut` pass is able to combine *any* _k_-LUT whose output is connected solely to an _n_-LUT where both of them share _m_ inputs, given that at least one of these LUTs has a width of _k_ + _n_ - _m_ - 1. (Attributes can be used to restrict LUT combining, such as to preserve instantiated carry logic.)

This pass is very generic and operates on any `$lut` cells, even ones of different size. To work with iCE40's `SB_LUT4`, the `ice40_unlut` pass (#716) is used; normal techmapping takes care of mapping the resulting `$lut` cells back, as `opt_lut` takes care to not disturb existing inputs or unnecessarily resize LUT cells.

This pass is tested by using `equiv_induct` and a behavioral model of `SB_CARRY` to check equivalence of several ALU elements similar to real-world ones before and after `opt_lut`.

This concludes my work on #715.

PR #718, opened 2018-12-05
 gate2lut: new techlib, for converting Yosys gates to FPGA LUTs

whitequark:
This techlib is a straightforward mapping from `$_NOT_`, `$_AND_`, `$_OR_`, `$_XOR_`, and `$_MUX_` to a _k_-LUT, with configurable _k_. It is useful for experimenting with replacement of `abc`.

Combined with #717, this extremely naïve techmapper produces, in my non-rigorous tests, similar LUT density to `abc`, at a miniscule fraction of the complexity and effort. For example, the design below is synthesized as follows:
  * `abc` as invoked by `synth_ice40`: 17 `SB_LUT4`;
  * LUT techmapping: 51 `SB_LUT4`;
  * LUT techmapping followed by LUT packing via `opt_lut`: 18 `SB_LUT4`.

In addition, while right now `simplemap` and `techmap` do not seem to preserve wire names (or maybe I'm doing something wrong; haven't looked in detail yet), it seems far simpler to adjust them to do so. In that case, the naive techmapper could be greatly useful for design optimization even if it produces substantially worse logic, so long as the logic it produces is *uniformly* worse than the logic produces by `abc`.

<details>

```verilog
module top(
  input [15:0] s_a,
  input [15:0] s_b,
  output reg [16:0] s_o,
  input c_or,
  input c_and,
  input c_xor
);

always @(*) begin
  x_sel <= 2'd0;
  if (c_or)  x_sel <= 1'd1;
  if (c_and) x_sel <= 2'd2;
  if (c_xor) x_sel <= 2'd3;
end

always @(*) begin
  s_o <= 17'd0;
  case (x_sel)
    1'd0: s_o <= s_a;
    1'd1: s_o <= (s_a | s_b);
    2'd2: s_o <= (s_a & s_b);
    2'd3: s_o <= (s_a ^ s_b);
  endcase
end

endmodule
```
</details>


whitequark:
According to @daveshah1, picorv32 on HX8K with `-noabc` now achieves Fmax of 54 MHz with the very latest nextpnr, which is comparable to picorv32 with `abc` one month ago.

> Note that the numbers reported above are for _multiple_ invocations of `opt_lut` pass; a single invocation of that pass will show significantly worse results. I think I'm not updating the worklist correctly somewhere. I'll take care of this in the other PR.

Also this was fixed.

cliffordwolf:
This now has some (trivial) conflicts in synth_icxe40. I'll merge it as soon as those conflicts are resolved.

whitequark:
Done!

PR #719, opened 2018-12-05
Fix Travis on OSX

q3k:
I can't easily debug this (it's 2018 and I'm still not allowed to run OSX on non-Apple hardware), so I'll just try changing some random things until it works ¯\\\_(ツ)\_/¯.


q3k:
And it works. Thanks for the help, @justMaku!

PR #723, opened 2018-12-06
synth_ice40: split `map_gates` off `fine`

whitequark:
I am currently looking at some optimizations to be done on iCE40 that would run on the coarse-grain cells. In this case it is very useful to me to be able to stop synthesis before all arch-dependent transformations, but after all arch-independent optimizations have already been done. I.e. see and affect the state just before techmapping, and restart afterwards.

Feel free to close this PR if this is not something desirable, or tell me to also implement this for other `synth_*` commands.

### DIFF
--- a/techlibs/ice40/synth_ice40.cc
+++ b/techlibs/ice40/synth_ice40.cc
@@ -240,6 +240,10 @@ struct SynthIce40Pass : public ScriptPass
 			run("opt -fast -mux_undef -undriven -fine");
 			run("memory_map");
 			run("opt -undriven -fine");
+		}
+
+		if (check_label("map_gates"))
+		{
 			if (nocarry)
 				run("techmap");
 			else


cliffordwolf:
> Feel free to close this PR if this is not something desirable, or tell me to also implement this for other synth_* commands.

LGTM.

The only thing I'm unsure about is re keeping the name `fine` for the last pre-map block. For the generic `synth` command it makes sense to have a `corase` and `fine` block. But for the others most of `fine` has been converted into various `map_*` blocks and now for them `fine` is just just the things run between the coarse synthesis stuff an mapping.

Now the name `fine` is essentially a copy&paste artifact, not a reasonable name for most `synth_*` commands.

What do you think about renaming `fine` to something like `premap`, or maybe simply `map` as it is the start of the mapping phase?

whitequark:
> What do you think about renaming `fine` to something like `premap`, or maybe simply `map` as it is the start of the mapping phase?

I think `map` sounds good! I assume you'll make that change?

cliffordwolf:
Renamed `synch_ice40` fine to map in 5fa5dbb.

PR #724, opened 2018-12-06
equiv_opt: new command, for verifying optimization passes

whitequark:
When writing optimization passes I heavily rely on `equiv` commands to verify correctness. I have a few scripts for that locally, but Yosys scripts are not very flexible, and I feel this is in general a very useful shortcut command to have. This is also quite useful for tests.

### DIFF
--- a/passes/equiv/Makefile.inc
+++ b/passes/equiv/Makefile.inc
@@ -9,4 +9,4 @@ OBJS += passes/equiv/equiv_induct.o
 OBJS += passes/equiv/equiv_struct.o
 OBJS += passes/equiv/equiv_purge.o
 OBJS += passes/equiv/equiv_mark.o
-
+OBJS += passes/equiv/equiv_opt.o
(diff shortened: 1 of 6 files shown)


mithro:
Yes please! This is awesome.

whitequark:
> What do you prefer for easy-to fix indenting stuff like this: Shall I just merge it and fix indenting ontop of it, potentially (but unlikely) causing merge conflicts don't the road for you, or should I complain in a change request like this and let you fix it, potentially causing a bit of overhead work?

Whatever results in less work for you. I am (unfortunately) proficient with Git so any merge conflicts upstream add maybe a few seconds required to type the right rebase command. Not an issue for me at all.

cliffordwolf:
In that case I'll simply merge and then commit indenting fixes ontop, and post a note like this one:

Merged with indenting fixes, see f53e19c.

whitequark:
Thanks!

PR #726, opened 2018-12-07
write_verilog: inline internal cells that are used exactly once

whitequark:
This has the effect of leaving *much* nicer Verilog code, and making Yosys quite usable for converting other HDLs to human-readable Verilog (which is what I'm experimenting with).

For example, a simple ALU:

<details>
<summary>Before (or with -noinline)</summary>

```verilog
module \$1 ();
  wire [15:0] _00_;
  wire [15:0] _01_;
  wire [16:0] _02_;
  wire [16:0] _03_;
  wire [16:0] _04_;
  wire [16:0] _05_;
  wire _06_;
  wire [17:0] _07_;
  wire _08_;
  wire _09_;
  wire [15:0] _10_;
  wire [15:0] a;
  wire [15:0] \a$next ;
  wire [15:0] b;
  wire [15:0] \b$next ;
  reg co;
  wire \co$next ;
  reg [15:0] o;
  wire [15:0] \o$next ;
  wire [1:0] sel;
  wire [1:0] \sel$next ;
  assign _00_ = a & b;
  assign _01_ = a ^ b;
  assign _03_ = + a;
  assign _05_ = - b;
  assign _04_ = + $signed(_05_);
  assign _07_ = $signed(_03_) + $signed(_04_);
  assign _02_ = + $signed(_07_);
  assign _06_ = sel == 0'b0;
  assign _08_ = sel == 1'h1;
  assign _09_ = sel == 2'h2;
  assign _10_ = a | b;
  always @* begin
    o = 16'h0000;
    co = 1'h0;
    casez ({ 1'h1, _09_, _08_, _06_ })
      4'bzzz1:
          o = _10_;
      4'bzz1z:
          o = _00_;
      4'bz1zz:
          o = _01_;
      4'b1zzz:
          { co, o } = _02_;
    endcase
  end
  always @* begin
      o <= \o$next ;
      co <= \co$next ;
  end
endmodule
```
</details>

<details>
<summary>After</summary>

```verilog
module \$1 ();
  wire [15:0] a;
  wire [15:0] \a$next ;
  wire [15:0] b;
  wire [15:0] \b$next ;
  reg co;
  wire \co$next ;
  reg [15:0] o;
  wire [15:0] \o$next ;
  wire [1:0] sel;
  wire [1:0] \sel$next ;
  always @* begin
    o = 16'h0000;
    co = 1'h0;
    casez ({ 1'h1, sel == 2'h2, sel == 1'h1, sel == 0'b0 })
      4'bzzz1:
          o = a | b;
      4'bzz1z:
          o = a & b;
      4'bz1zz:
          o = a ^ b;
      4'b1zzz:
          { co, o } = + $signed($signed(+ a) + $signed(+ $signed(- b)));
    endcase
  end
  always @* begin
      o <= \o$next ;
      co <= \co$next ;
  end
endmodule
```
</details>

This is rather close to what one might write by hand.

This pass is conservative; only wires that are driven by exactly one cell and are consumed exactly once by another cell or exactly once in a process are eliminated. Everything else, including unused wires, is kept as-is. In some cases, e.g. if a wire is used exactly once but it is sliced, such a wire is also left as-is as an arbitrary expression is not legal in Verilog in such context, only an identifier.

### DIFF
--- a/backends/verilog/verilog_backend.cc
+++ b/backends/verilog/verilog_backend.cc
@@ -25,6 +25,7 @@
 #include "kernel/celltypes.h"
 #include "kernel/log.h"
 #include "kernel/sigtools.h"
+#include "kernel/modtools.h"
 #include <string>
 #include <sstream>
 #include <set>
@@ -33,15 +34,17 @@
 USING_YOSYS_NAMESPACE
 PRIVATE_NAMESPACE_BEGIN
 
-bool verbose, norename, noattr, attr2comment, noexpr, nodec, nohex, nostr, extmem, defparam, decimal, siminit;
+bool verbose, norename, noattr, attr2comment, noexpr, noinline, nodec, nohex, nostr, extmem, defparam, decimal, siminit;
 int auto_name_counter, auto_name_offset, auto_name_digits, extmem_counter;
 std::map<RTLIL::IdString, int> auto_name_map;
 std::set<RTLIL::IdString> reg_wires, reg_ct;
+std::map<RTLIL::Wire*, int> proc_consumed_wires;
 std::string auto_prefix, extmem_prefix;
 
 RTLIL::Module *active_module;
 dict<RTLIL::SigBit, RTLIL::State> active_initdata;
 SigMap active_sigmap;
+ModIndex active_modindex;
 
 void reset_auto_counter_id(RTLIL::IdString id, bool may_rename)
 {
@@ -181,6 +184,72 @@ bool is_reg_wire(RTLIL::SigSpec sig, std::string &reg_name)
 	}
 
 	return true;
+}
+
+bool can_inline_cell_expr(RTLIL::Cell *cell)
+{
+	static pool<IdString> inlinable_cells = {
+		"$not", "$pos", "$neg",
+		"$and", "$or", "$xor", "$xnor",
+		"$reduce_and", "$reduce_or", "$reduce_xor", "$reduce_xnor", "$reduce_bool",
+		"$shl", "$shr", "$sshl", "$sshr",
+		"$lt", "$le", "$eq", "$ne", "$eqx", "$nex", "$ge", "$gt",
+		"$add", "$sub", "$mul", "$div", "$mod", "$pow",
+		"$logic_not", "$logic_and", "$logic_or",
+		"$mux"
+	};
+
+	if (noinline)
+		return false;
+
+	if (!inlinable_cells.count(cell->type))
+		return false;
+
+	const RTLIL::SigSpec &output = cell->getPort(ID::Y);
+	if (!output.is_wire())
+		return false;
+
+	RTLIL::Wire *output_wire = output.as_wire();
+	if (output_wire->port_id || (!noattr && output_wire->attributes.size()))
+		return false;
+
+	pool<ModIndex::PortInfo> ports = active_modindex.query_ports(output[0]);
+	if (proc_consumed_wires[output_wire] == 1 && ports.size() == 1)
+		return true;
+
+	if (proc_consumed_wires[output_wire] == 0 && ports.size() == 2)
+	{
+		auto port1 = ports.pop();
+		auto port2 = ports.pop();
+		return port1.cell->getPort(port1.port).size() ==
+		       port2.cell->getPort(port2.port).size();
+	}
+
+	return false;
+}
+
+bool can_inline_wire(RTLIL::Wire *wire)
+{
+	if (noinline)
+		return false;
+
+	RTLIL::SigSpec wire_spec = RTLIL::SigSpec(wire);
+	if (wire_spec.empty())
+		return false;
+
+	pool<ModIndex::PortInfo> ports = active_modindex.query_ports(wire_spec[0]);
+	if (ports.size() > 2)
+		return false;
+
+	for (auto &port_info : ports)
+	{
+		if (port_info.cell->name[0] == '$' && port_info.port == "\\Y")
+		{
+			if (can_inline_cell_expr(port_info.cell))
+				return true;
+		}
+	}
+	return false;
 }
 
 void dump_const(std::ostream &f, const RTLIL::Const &data, int width = -1, int offset = 0, bool no_decimal = false, bool escape_comment = false)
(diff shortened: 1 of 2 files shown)


cliffordwolf:
Travis CI seems to think this is breaking things.

In general, my gut feeling regarding this is that it might be something that's extremely hard to get right, considering the LRM spec details of resolving expression size and signedness in Verilog.

whitequark:
> In general, my gut feeling regarding this is that it might be something that's extremely hard to get right, considering the LRM spec details of resolving expression size and signedness in Verilog.

Right, this is why I'm making the change very conservatively: only when the entire expression result (i.e. every bit in the wire) is used exactly as-is, it is inlined. Do you think this can still cause problems?

I can switch this to be behind a flag `-inline` default-off, rather than `-noinline` default-on.

> Travis CI seems to think this is breaking things.

Yeah. I wanted to see general feedback before digging down to resolve that failure. Will look at it soon.

whitequark:
> Maybe we should go with default-off for now, and merge it as soon as we think it "looks good" from simple tests. Later, when we have more confidence in the feature we can switch it to default-on.

nMigen heavily relies on this feature to produce Verilog. (I've designed nMigen's IR to be a very close match for RTLIL, and rely on write_verilog to produce Verilog from that). So, I am confident that it produces correct Verilog in many cases.

At some point I will add expand nMigen's testsuite to check that all signals in the design match cycle-for-cycle between nMigen's internal simulator, iverilog on pre-`proc` write_verilog, and iverilog on post-`proc` write_verilog, which should give some additional assurances.

whitequark:
@cliffordwolf I've fixed the test failure. One line patch. `-noexpr` should obviously imply `-noinline`...

whitequark:
@cliffordwolf Is there some explanation on how to run VlogHammer?

cliffordwolf:
> Is there some explanation on how to run VlogHammer?

- Clone the repo
  - https://github.com/YosysHQ/VlogHammer
- Edit `Makefile`:
  - Set `SYN_LIST` to `yosys`
  - Set `SIM_LIST` to `icarus yosim`
  - Set `YOSYS_MODE` to `inline`
- Edit `scripts/syn_yosys.sh`:
  - Add `inline` mode
- Generate test cases
  - `make generate`
- Synthesize test cases
  - `make -j$(nproc) syn`
- Check synthesis results
  - `make -j$(nproc) check`
- Generate Reports
  - `make -j$(nproc) report`
  - Open `report.html` in web browser

For testing purposes you might want to remove all except a few files from `rtl/` after `make generate` so `syn check report` completes quickly. If that works you can run it on a full `rtl/` directory.

In the past I've found bugs with VlogHammer that are only triggered by one in 100k test cases. To find crap like that you'd need to write a small shell script that runs the whole thing in a loop until it generates a non-empty report.

whitequark:
Unfortunately, I cannot do this, because it appears that upstream Yosys does not pass this test either. There is a number of crashes, e.g. with `issue_057`, where the Verilog backend generates this:
```verilog
module issue_057(a, y);
  wire [31:0] _00_;
  wire [31:0] _01_;
  wire [31:0] _02_;
  wire [31:0] _03_;
  wire [4:0] _04_;
  wire [31:0] _05_;
  input [2:0] a;
  output [3:0] y;
  assign _00_ = 32'd15 + a;
  assign _01_ = _00_ + 32'd1;
  assign _02_ = _05_ - 32'd5;
  assign _03_ = $signed(32'd6) - $signed(_02_);
  assign _04_ = 11'h1ac[$signed(_03_) +: 5];
  assign _05_ = _01_ - 32'd5;
  assign y = _04_[3:0];
endmodule
```
I'm frankly not sure what to do here. Can you take a look?

cliffordwolf:
> I'm frankly not sure what to do here. Can you take a look?

That's funny. I ran it after my previous post and all I found was one new Vivado bug. I'll have a closer look.

cliffordwolf:
Fixed upstream Yosys issue in 6c5049f, added new `minimal` yosys synthesis mode to vloghammer.

whitequark:
@cliffordwolf You are right and vloghammer found at least one bug. However, I am confused as to its source. The cause is that in the following code:
```verilog
module top(a, b, c);
  input a;
  input [2:0] b;
  input [3:0] c;
  wire [3:0] _0_;
  wire [3:0] _1_;
  wire [3:0] _2_;
  assign _0_ = + $signed(b);
  assign _1_ = a ? _0_ : c;
  assign _2_ = a ? (+ $signed(b)) : c;
endmodule
```
the values `_1_` and `_2_` are not equivalent. I'm not really sure why is that--do you know the reason offhand?

whitequark:
That is, my understanding was that the result is signed in Verilog only if both operands are signed, and `c` is not signed. So I'm a bit mystified.

cliffordwolf:
Without having actually played with the code you posted, I would assume this is the issue:

image

Thus, `a ? (+ $signed(b)) : c;` should be the same as `a ? b : c`, which is different from `a ? _0_ : c` because `_0_` is sign-extended and not zero-extended.

whitequark:
@cliffordwolf Incredible. I have learned more about Verilog arithmetic rules than I ever wanted to know, and I'm not actually sure I'm persistent enough to make this work. That does not happen often... and yes, I do see why you want 100k fuzzer testcases as a validation.

nakengelhardt:
@whitequark do you want to assign this one to yourself to mark that you're still interested in finishing it someday?

whitequark:
Yup.

ZirconiumX:
For the record, I've rebased this branch to master and fixed a bug where it attempts to inline the left-hand-side of an `assign`.

whitequark:
> For the record, I've rebased this branch to master in my Yosys fork and fixed a bug where it attempts to inline the left-hand-side of an `assign`.

Thanks! If you'd like we can try to bring it to a usable state now.

ZirconiumX:
Here's a bug that @mwkmwkmwk found and I forgot about:

```
module \top
  wire input 1 \i
  wire output 2 \o
  wire \t
  wire output 2 \o2
  cell $not \not1
    parameter \A_SIGNED 0
    parameter \A_WIDTH 1
    parameter \Y_WIDTH 1
    connect \A \i
    connect \Y \t
  end
  connect \o \t
  cell $not \not2
    parameter \A_SIGNED 0
    parameter \A_WIDTH 1
    parameter \Y_WIDTH 1
    connect \A \t
    connect \Y \o2
  end
end
```

Here, the inlining heuristic decides that `assign t = i;` is inlineable and elides it, but considers assignments to `o` and `o2` to not be inlineable (presumably because the left-hand-side is an output port), resulting in references to `t` when `t` is not assigned a value.

```verilog
module top(i, o, o2);
  input i;
  output o;
  output o2;
  wire t;
  assign o2 = ~ t;
  assign o = t;
endmodule
```

Here, either `t` needs to be considered inlineable as a cell input port  (resulting in `assign o = i; assign o2 = ~ (~ i));`) or the `assign t = i;` line needs to be considered not inlineable (resulting in `assign t = i; assign o2 = ~ t; assign o = t;`).

whitequark:
Rebased on top of master, incluing @ZirconiumX's fixes (thanks!)

whitequark:
Revisiting this: I will definitely not be finishing this PR.

PR #735, opened 2018-12-12
deminout fixes

daveshah1:
Fixes for issues raised in YosysHQ/nextpnr#176

 - Consider constant drivers in `deminout` so constant driven signals aren't demoted to inputs
 - Treat word-level `$tribuf` the same as `$_TBUF_` in `deminout`, as `synth_ice40` runs `deminout` on the word-level netlist


whitequark:
LGTM.

PR #736, opened 2018-12-13
select: print selection if a -assert-* flag causes an error

whitequark:
This is useful for sanity checks in batch scripts. This functionality cannot be replaced with `select -list` because `-list` does not output anything if `-q` is passed to `yosys`.

### DIFF
--- a/passes/cmds/select.cc
+++ b/passes/cmds/select.cc
@@ -896,6 +896,29 @@ static void select_stmt(RTLIL::Design *design, std::string arg)
 	select_filter_active_mod(design, work_stack.back());
 }
 
+static std::string describe_selection_for_assert(RTLIL::Design *design, RTLIL::Selection *sel)
+{
+	std::string desc = "Selection contains:\n";
+	for (auto mod_it : design->modules_)
+	{
+		if (sel->selected_module(mod_it.first)) {
+			for (auto &it : mod_it.second->wires_)
+				if (sel->selected_member(mod_it.first, it.first))
+					desc += stringf("%s/%s\n", id2cstr(mod_it.first), id2cstr(it.first));
+			for (auto &it : mod_it.second->memories)
+				if (sel->selected_member(mod_it.first, it.first))
+					desc += stringf("%s/%s\n", id2cstr(mod_it.first), id2cstr(it.first));
+			for (auto &it : mod_it.second->cells_)
+				if (sel->selected_member(mod_it.first, it.first))
+					desc += stringf("%s/%s\n", id2cstr(mod_it.first), id2cstr(it.first));
+			for (auto &it : mod_it.second->processes)
+				if (sel->selected_member(mod_it.first, it.first))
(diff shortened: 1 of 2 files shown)


whitequark:
To give an example of such sanity check, here is what nMigen does before emitting Verilog:
```
# Make sure there are no undriven wires in generated RTLIL.
proc
select -assert-none w:* i:* %a %d o:* %a %ci* %d c:* %co* %a %d n:$* %d
```

passes/cmds/select.cc
```diff
@@ -896,6 +896,28 @@ static void select_stmt(RTLIL::Design *design, std::string arg)
...
+					fprintf(stderr, "%s/%s\n", id2cstr(mod_it.first), id2cstr(it.first));
+			for (auto &it : mod_it.second->memories)
+				if (sel->selected_member(mod_it.first, it.first))
+					fprintf(stderr, "%s/%s\n", id2cstr(mod_it.first), id2cstr(it.first));
+			for (auto &it : mod_it.second->cells_)
+				if (sel->selected_member(mod_it.first, it.first))
+					fprintf(stderr, "%s/%s\n", id2cstr(mod_it.first), id2cstr(it.first));
+			for (auto &it : mod_it.second->processes)
+				if (sel->selected_member(mod_it.first, it.first))
+					fprintf(stderr, "%s/%s\n", id2cstr(mod_it.first), id2cstr(it.first));
```
cliffordwolf: This shouldn't simply print to stderr.

These `fprintf(stderr, ...)` should either be replaced by `log()` or `log_warning()`, or `print_selection_for_assert` should be replaced by a function that returns a multi-line string that can be included in the error message printed with `log_error` in the callers below.

passes/cmds/select.cc
```diff
@@ -1394,7 +1417,12 @@ struct SelectPass : public Pass {
 				log_cmd_error("No selection to check.\n");
 			work_stack.back().optimize(design);
 			if (!work_stack.back().empty())
-				log_error("Assertion failed: selection is not empty:%s\n", sel_str.c_str());
+			{
+				RTLIL::Selection *sel = &work_stack.back();
+				sel->optimize(design);
+				std::string desc = describe_selection_for_assert(design, sel);
+				log_error("%s\nAssertion failed: selection is not empty:%s\n", desc.c_str(), sel_str.c_str());
```
cliffordwolf: Would the following format be also be okay for you in your application?

```
log_error("Assertion failed: selection is not empty:%s\n%s", sel_str.c_str(), desc.c_str());
```

I think that output format would work better with the fact that `log_error()` appends something in front of the message, thus would modifies the first line. It is the better format imo for cases where the selection is empty. (Obviously not relevant for this error but for some of the errors below.)

whitequark: Makes no difference for me as nMigen should never generate any RTL that triggers the assertion. I've changed the format to your proposed one.

PR #738, opened 2018-12-14
verilog_parser: Properly handle recursion when processing attributes

smunaut:
Instead of having a single attr_list object, we create a new one at the beginning of each option list.
If there is an existing one, then we put it on top of a stack and at the end of the opt list, we check the
stack, pop one if there is one.

Signed-off-by: Sylvain Munaut <tnt@246tNt.com>

### DIFF
--- a/frontends/verilog/verilog_parser.y
+++ b/frontends/verilog/verilog_parser.y
@@ -35,6 +35,7 @@
 
 %{
 #include <list>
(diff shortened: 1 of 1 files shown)


daveshah1:
Looking at the CI failure on the OS X build, seems like a `#include <stack>` might be missing?

PR #740, opened 2018-12-16
Improve ABC netname preservation 

daveshah1:
- Better recovered netnames, don't use log_signal as this was adding spaces
 - Copy `src` attribute for use with `rename -src`
 - Use `abc -dress` by default in the basic part of `synth_ice40` (not when retiming, as this doesn't seem to work)

I'd like to test this a bit more before merging.

### DIFF
--- a/passes/techmap/abc.cc
+++ b/passes/techmap/abc.cc
@@ -329,6 +329,18 @@ void extract_cell(RTLIL::Cell *cell, bool keepff)
 
 std::string remap_name(RTLIL::IdString abc_name)
 {
+	std::string abc_sname = abc_name.substr(1);
+	if (abc_sname.substr(0, 5) == "ys__n") {
+		int sid = std::stoi(abc_sname.substr(5));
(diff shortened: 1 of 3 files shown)


eddiehung:
Just to summarise, this PR:
- Adds a new ABC option: `-dress`
- Enables this by default (with no recourse) for the `synth_ice40` command, but only for the last ABC invocation during the `map_luts` stage -- it does not get called during the prior invocations of ABC when `-retime`, or when `-abc2` options are given

I've merged current master 266511b into @daveshah1 's branch to get [eddiehung:improve_dress](https://github.com/eddiehung/yosys/tree/improve_dress) on which I've verified that [yosys-bench](https://github.com/YosysHQ/yosys-bench) produces the same result before and after (1466 Verilog benchmarks)

daveshah1:
I am happy with this now.

eddiehung:
@cliffordwolf Happy to merge?

PR #742, opened 2018-12-16
Update CHANGELOG to mention my improvements

whitequark:
I haven't realized before that it exists...


cliffordwolf:
jfyi: no need to add stuff to changelog. (but doesn't hurt either of course.) before a release I usually review the entire git history since the last release and write the changelog. now that I write this down I realize that's probably not a very scaleable solution.. well, we need to do more frequent releases (= smaller changelogs) anyway.  :)

whitequark:
Right, just thought I'd decrease your workload. Because I don't know about you but I really hate manually writing CHANGELOG on my projects, and only do it when essential...

PR #746, opened 2018-12-17
Support for DRAM inferring on Anlogic FPGAs

Icenowy:


mmicko:
LGTM. Will probably split physical and logical black box elements in future

PR #748, opened 2018-12-17
Add btor ops for $mul, $div, $mod and $concat

makaimann:
This simply adds four btor ops to the btor backend.

Here is the very simple Verilog and script I used to test this: 
[ops_test.zip](https://github.com/YosysHQ/yosys/files/2687906/ops_test.zip)

Perhaps the only non-trivial part to this is how to handle $mod. BTOR2 has support for signed and unsigned remainder, and signed mod. I believe this follows the SMT-LIB semantics, where the only difference between srem and smod is whether it takes the sign of the dividend (srem) or the divisor (smod). Here's that reference: [QF_BV](http://smtlib.cs.uiowa.edu/logics-all.shtml#QF_BV). According to these [slides](http://web.engr.oregonstate.edu/~traylor/ece474/beamer_lectures/verilog_operators.pdf), the verilog mod operator takes the sign of the dividend, so I used srem.


makaimann:
By the way, I tried to match the coding style as best I could, but I'm happy to make stylistic as well as functional changes as needed.

PR #749, opened 2018-12-18
anlogic: fix dbits of Anlogic Eagle DRAM16X4

Icenowy:
The dbits of DRAM16X4 is wrong set to 2, which leads to waste of DRAM
bits.

Fix the dbits number in the RAM configuration.

Signed-off-by: Icenowy Zheng <icenowy@aosc.io>

### DIFF
--- a/techlibs/anlogic/drams.txt
+++ b/techlibs/anlogic/drams.txt
@@ -1,7 +1,7 @@
 bram $__ANLOGIC_DRAM16X4
   init 0
   abits 4
-  dbits 2
+  dbits 4
   groups 2
   ports  1 1
   wrmode 0 1


mmicko:
Looks almost same like one from ECP5 now.  @daveshah1 can you point what is best way to confirm if clkpol  should be 0 1 or 0 2 (like on ECP5) in this case ?

Icenowy:
@mmicko Please note that the EG_LOGIC_DRAM16X4 cell doesn't have any way to configure clock polarity, so the clkpol shouldn't have 2. The write clock is posedge, so it's 1.

mmicko:
Great, thanks for confirming you already checked. Please check email. :)

PR #750, opened 2018-12-18
Initialization of Anlogic DFFs

Icenowy:


mmicko:
Thanks. Please note that probably mappings are also wrong for some types. _DFF_P_ and _DFF_N_  are those that were covered by testing

Icenowy:
@mmicko I found a problem -- Inside a AL_MAP_SEQ The SR will set the FF's value to the value indicated by INIT, which is not a expected behavior by any DFFs in Yosys, specially when initial values are inserted to the system.

mmicko:
@Icenowy Busy right now, but can investigate this. What confused me before was that there is also parameter INI that is used in simulation model for AL_MAP_SEQ but it is not available when you try synthesis so we can say for DFFs init is always 0, but need to check logic for SRs.

Icenowy:
BTW investigations show that internally Tang Dynasty doesn't generate AL_MAP_SEQ.

If run `write_verilog` after `map` in Tang Dynasty, in the generated code there will be "reg_ar_as_w1" cells for DFFs, and interestingly this cell is not a blacbox -- TD will also output its implementation, as well as two cells AL_MUX and AL_DFF used by it.

Icenowy:
After let TD to generate PHY level cells (MSLICE & LSLICE), it's discovered that internally the SLICEs of Eagle only supports async SR (reset to the initial value), do not support clear/set.

So if the initial value of the register is not equal to the async reset value, it's just not implemetable on Eagle.

Maybe we should catch this with a custom duplicate of `dffinit`, and report an error when the behavior is not implementable?

mmicko:
@Icenowy problem with usage of those custom DFF elements is that they are just represented like that in their output verilog, there are no primitives to support them. Was looking what is best to use, and at the end it was AL_MAP_SEQ as only able to synth. Note that best would be for start to remove DFF types that are not able to be synth with TD, and then to see if there is a need for additional step. For Ice40 there were custom steps for FFs but would check with @cliffordwolf  what would be best approach.

Icenowy:
@mmicko If we remove DFF types that are not *FULLY* implementable on Eagle FPGAs, we will need to remove all $_DFF_{P,N}{P,N}{0,1}_ cells, because some code similar to the below is not implementable on Eagle FPGAs:
```
reg dff_pp0 = 1;
always @(posedge clk or posedge reset)
  if (reset)
    dff_pp0 <= 0;
  else
    dff_pp0 <= ~dff_pp0;
```

In the slices, the DFFs are implemented with "SR" signals, which means async set to the inital value, like what's present in AL_MAP_SEQ. However, in most case the async reset value will be equal to the initial value, so the design will work.

Considering this, I think we should insert custom DFF initialzation value injection step, which will detect async reset DFFs, and report an error when the async reset value is not equal to the DFF initialization value.

BTW for async reset DFFs, should the SRMODE be set to "ASYNC" instead of "SYNC"?

Icenowy:
Force-pushed the branch with some code that checks conflicts when initializing DFFs.

daveshah1:
Yes, the ECP5 hardware is identical here - FFs init only to their reset value. Although I want to double check how Diamond handles things with sync and async FFs, once merged I should easily be able to duplicate this for ECP5.

PR #757, opened 2018-12-20
manual: document $meminit cell and memory_* passes

whitequark:
See inline comments for a few things that I've had to infer from code, and not sure whether my guess is the same as the authoritative definition.

Semi-related, can we get rid of non-free fonts in the Yosys manual? Not only they are non-free but the installation process, as is typical with LaTeX, is prone to obscure failures. I have eventually realized that I had to run the commands from README exactly as stated, with sudo, because it looks like installing to `$HOME` is broken, but that took a while...

### DIFF
--- a/manual/CHAPTER_CellLib.tex
+++ b/manual/CHAPTER_CellLib.tex
@@ -254,7 +254,7 @@ enable bit for each data bit), an address input \B{ADDR} and a data input
 
 \begin{itemize}
 \item \B{MEMID} \\
-The name of the RTLIL::Memory object that is associated with this read port.
+The name of the RTLIL::Memory object that is associated with this write port.
 
 \item \B{ABITS} \\
 The number of address bits (width of the \B{ADDR} input port).
@@ -263,7 +263,7 @@ The number of address bits (width of the \B{ADDR} input port).
 The number of data bits (width of the \B{DATA} output port).
 
 \item \B{CLK\_ENABLE} \\
-When this parameter is non-zero, the clock is used. Otherwise this read port is asynchronous and
+When this parameter is non-zero, the clock is used. Otherwise this write port is asynchronous and
 the \B{CLK} input is not used.
 
 \item \B{CLK\_POLARITY} \\
--- a/manual/CHAPTER_CellLib.tex
+++ b/manual/CHAPTER_CellLib.tex
@@ -211,14 +211,15 @@ Add information about {\tt \$sr} cells (set-reset flip-flops) and d-type latches
 \subsection{Memories}
 \label{sec:memcells}
 
-Memories are either represented using RTLIL::Memory objects and {\tt \$memrd} and {\tt \$memwr} cells
-or simply by using {\tt \$mem} cells.
+Memories are either represented using RTLIL::Memory objects, {\tt \$memrd}, {\tt \$memwr}, and {\tt \$meminit}
+cells, or by {\tt \$mem} cells alone.
 
 In the first alternative the RTLIL::Memory objects hold the general metadata for the memory (bit width,
 size in number of words, etc.) and for each port a {\tt \$memrd} (read port) or {\tt \$memwr} (write port)
 cell is created. Having individual cells for read and write ports has the advantage that they can be
 consolidated using resource sharing passes. In some cases this drastically reduces the number of required
-ports on the memory cell.
+ports on the memory cell. In this alternative, memory initialization data is represented by {\tt \$meminit} cells,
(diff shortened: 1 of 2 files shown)


manual/CHAPTER_CellLib.tex
```diff
@@ -211,14 +211,15 @@ \subsection{Registers}
...
+Memories are either represented using RTLIL::Memory objects, {\tt \$memrd}, {\tt \$memwr}, and {\tt \$meminit}
+cells, or by {\tt \$mem} cells alone.
 
 In the first alternative the RTLIL::Memory objects hold the general metadata for the memory (bit width,
 size in number of words, etc.) and for each port a {\tt \$memrd} (read port) or {\tt \$memwr} (write port)
 cell is created. Having individual cells for read and write ports has the advantage that they can be
 consolidated using resource sharing passes. In some cases this drastically reduces the number of required
-ports on the memory cell.
+ports on the memory cell. In this alternative, memory initialization data is represented by {\tt \$meminit} cells,
+which allow delaying constant folding for initialization addresses and data until after the frontend finishes.
```
whitequark: Or at least, I assume that's why `ADDR` and `DATA` are ports and not parameters.

manual/CHAPTER_CellLib.tex
```diff
@@ -273,6 +274,15 @@ \subsection{Memories}
 The cell with the higher integer value in this parameter wins a write conflict.
 \end{itemize}
 
+The {\tt \$meminit} cells have an address input \B{ADDR} and a data input \B{DATA}. Both of the inputs
+must resolve to a constant for synthesis to succeed. If two {\tt \$meminit} cells have the same
+\B{MEMID} parameter and \B{ADDR} input, the contents of that memory at that address is undefined.
```
whitequark: Just documenting the current behavior of `memory_collect`. It would be prudent to check for duplicate addresses, I think, if this is not explictily desired to make this undefined.

manual/CHAPTER_CellLib.tex
```diff
@@ -345,9 +358,11 @@ \subsection{Memories}
 This input is \B{WR\_PORTS}*\B{WIDTH} bits wide, containing all data signals for the write ports.
 \end{itemize}
 
-The {\tt techmap} pass can be used to manually map {\tt \$mem} cells to
-specialized memory cells on the target architecture, such as block ram resources
-on an FPGA.
```
whitequark: I've removed this passage because I *think* this isn't actually done by `techmap`? Or else what would be all the custom steps in `synth_*` for.

manual/CHAPTER_CellLib.tex
whitequark: Actually, now that I look closer there is a `PRIORITY` parameter... that promptly gets ignored by `memory_collect`.

manual/CHAPTER_CellLib.tex
cliffordwolf: yes, that's correct.

manual/CHAPTER_CellLib.tex
cliffordwolf: No, `memory_collect` uses `PRIORITY`. See `memcells_cmp()`. The `$meminit` cell with higher priority overwrites the values from cells with lower priority.

manual/CHAPTER_CellLib.tex
cliffordwolf: You could do it with `techmap`. But then I wrote the bram mapper which is better for the job in almost all practical cases. (That paragraph is from before the bram mapper existed. I think it's a good choice to remove it.)

manual/CHAPTER_CellLib.tex
```diff
@@ -273,6 +274,29 @@ \subsection{Memories}
 The cell with the higher integer value in this parameter wins a write conflict.
 \end{itemize}
 
+The {\tt \$meminit} cells have an address input \B{ADDR} and a data input \B{DATA}, with the width
+of the \B{DATA} port equal to \B{WIDTH} parameter times \B{WORDS} parameter. Both of the inputs
+must resolve to a constant for synthesis to succeed. If two {\tt \$meminit} cells have the same
+\B{MEMID} parameter and \B{ADDR} input, the contents of that memory at that address is undefined;
+the \B{PRIORITY} parameter is currently ignored by the {\tt memory\_collect} pass.
```
cliffordwolf: Redundant comment to force github into "changes requested" mode: `memory_collect` _does_ use PRIORITY to resolve write conflicts.

whitequark: Ah, I misread--fixed now!

PR #761, opened 2018-12-22
proc_clean: remove any empty cases, if possible to do safely

whitequark:
Previously, only completely empty switches were removed.

This is a seemingly simple change with surprisingly far reaching consequences and motivation, so bear with me for a bit.

The immediate effect of this seemingly simple change, using Boneless CPU as a benchmark, is that the LUT count goes up: from 475 to 511 on one specific instantiation. This points to a larger inefficiency in mux tree optimization.

However! The motivation for adding this (and, later, I'd like to add further simplifications) to `proc_clean` is quite complicated, and is mostly because of the behavior of Verilog simulators. Specifically, nMigen would emit all of its logic into one giant RTLIL process, which ends up as a giant `always @*` block after `write_verilog`, plus a few `always @(posedge)` blocks that actually update the register values. This is entirely fine for synthesis, but every single Verilog simulator in existence, proprietary, FOSS, whichever, appears to choke on this pattern because of false combinatorial loops, once the Migen/nMigen design reaches a large enough size.

As such, I've changed nMigen to do the following transformation. For every signal, an RTLIL process individual for that specific process is emitted, which is essentially the same as the original RTLIL process that would be emitted, but "filtered" to only include assignments to that signal. This means that the resulting IL file (and Verilog output) will contain a very large amount of empty cases and processes, many of which are not cleaned up by the current `proc_clean` pass.

There is a worse consequence. I've mentioned that with this `proc_clean` change, LUT count goes up from 475 to 511. However, with the *nMigen* change, which is completely mechanical and produces equivalent code, the LUT count rises from 343 to 475!

My conclusions is that the current mux tree optimizations in Yosys are unable to cope with *this* `proc_clean` change, but are *also* unable to cope with other changes in input, in more severe ways. Therefore, it is my opinion that `proc_clean` should be improved to remove empty cases (such as done in this PR), and the mux optimizations should separately be improved to cope with redundant muxes.

### DIFF
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -3793,6 +3793,11 @@ RTLIL::CaseRule::~CaseRule()
 		delete *it;
 }
 
+bool RTLIL::CaseRule::empty() const
+{
+	return actions.empty() && switches.empty();
+}
+
 RTLIL::CaseRule *RTLIL::CaseRule::clone() const
 {
 	RTLIL::CaseRule *new_caserule = new RTLIL::CaseRule;
@@ -3807,6 +3812,11 @@ RTLIL::SwitchRule::~SwitchRule()
 {
 	for (auto it = cases.begin(); it != cases.end(); it++)
 		delete *it;
+}
+
+bool RTLIL::SwitchRule::empty() const
+{
+	return cases.empty();
 }
 
 RTLIL::SwitchRule *RTLIL::SwitchRule::clone() const
--- a/kernel/rtlil.h
+++ b/kernel/rtlil.h
@@ -1227,6 +1227,8 @@ struct RTLIL::CaseRule
 	~CaseRule();
 	void optimize();
 
+	bool empty() const;
+
 	template<typename T> void rewrite_sigspecs(T &functor);
 	RTLIL::CaseRule *clone() const;
 };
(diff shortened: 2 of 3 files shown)


whitequark:
> My conclusions is that the current mux tree optimizations in Yosys are unable to cope with _this_ `proc_clean` change, but are _also_ unable to cope with other changes in input, in more severe ways. Therefore, it is my opinion that `proc_clean` should be improved to remove empty cases (such as done in this PR), and the mux optimizations should separately be improved to cope with redundant muxes.

Interestingly, it looks like Synplify also cannot cope well with this style of coding (with many empty cases), though it seems marginally less inefficient than Yosys on a cursory check.

whitequark:
Interestingly#2, Vivado appears to have no problems, and infers the exact same amount of LUT regardless of coding style. This implies that Yosys should be able to do this too.

whitequark:
Not yet, since there are other bugs in nMigen that are higher priority than poor optimization. But it seems likely that I will end up making such an example at some point.

whitequark:
To confirm: do you generally agree that this change is desirable? If yes, I want to extend it so that it also removes empty cases in the middle, for switches that have no z/x in them (and so you can do this without reencoding the entire switch). I suspect that this will make the deficiencies in `opt_muxtree` easier to notice, but I don't want to bother if this change is not desirable in the end.

cliffordwolf:
Yes, I do agree this change is desirable. But I'm reluctant to merge it before I have a better understanding for why this has a negative impact on quality of results.

whitequark:
Sure, not merging until the cause for pessimization is clear is fine by me. I will improve `proc_clean` a bit more then.

whitequark:
I've added more code to clean empty cases if every case's compare is fully defined, which I think is the extent of cleanup `proc_clean` can do safely. I've also tried to torture the entire synthesis pipeline a bit with different nMigen options and cleanup options, and it looks like there's no QoS degradation!

PR #769, opened 2019-01-02
Fix typographical and grammatical errors and inconsistencies

whitequark:
The initial list of hits was generated with the codespell command
below, and each hit was evaluated and fixed manually while taking
context into consideration.

    DIRS="kernel/ frontends/ backends/ passes/ techlibs/""
    codespell $DIRS -S *.o -L upto,iff,thru,synopsys,uint

More hits were found by looking through comments and strings manually.

### DIFF
--- a/backends/table/table.cc
+++ b/backends/table/table.cc
@@ -109,7 +109,7 @@ struct TableBackend : public Backend {
 				else if (cell->output(conn.first))
 					*f << "out" << "\t";
 				else
-					*f << "unkown" << "\t";
+					*f << "unknown" << "\t";
 
 				*f << log_signal(sigmap(conn.second)) << "\n";
 			}
--- a/frontends/verific/verificsva.cc
+++ b/frontends/verific/verificsva.cc
@@ -827,9 +827,9 @@ struct SvaFsm
 
 			for (auto &it : nodes[i].edges) {
 				if (it.second != State::S1)
-					log("          egde %s -> %d\n", log_signal(it.second), it.first);
+					log("          edge %s -> %d\n", log_signal(it.second), it.first);
 				else
-					log("          egde -> %d\n", it.first);
+					log("          edge -> %d\n", it.first);
 			}
 
 			for (auto &it : nodes[i].links) {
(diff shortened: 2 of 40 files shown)


whitequark:
Also, there are a number of typos in ezsat and subcircuit libraries, but I wasn't able to find the upstream source for these. Do you have a repository somewhere I should add a PR against, or should I just fix them here?

backends/table/table.cc
```diff
@@ -109,7 +109,7 @@ struct TableBackend : public Backend {
 				else if (cell->output(conn.first))
 					*f << "out" << "\t";
 				else
-					*f << "unkown" << "\t";
+					*f << "unknown" << "\t";
```
whitequark: This seems like a bug.

frontends/verific/verificsva.cc
```diff
@@ -827,9 +827,9 @@ struct SvaFsm
 
 			for (auto &it : nodes[i].edges) {
 				if (it.second != State::S1)
-					log("          egde %s -> %d\n", log_signal(it.second), it.first);
+					log("          edge %s -> %d\n", log_signal(it.second), it.first);
```
whitequark: These four seem like a bug.

passes/cmds/setundef.cc
```diff
@@ -137,7 +137,7 @@ struct SetundefPass : public Pass {
 		log("        replace with $anyconst drivers (for formal)\n");
 		log("\n");
 		log("    -random <seed>\n");
-		log("        replace with random bits using the specified integer als seed\n");
+		log("        replace with random bits using the specified integer seed\n");
```
whitequark: Not sure what "als" means here, I assumed a copy-paste error.

cliffordwolf: Oops, yes. Should have been "as".  :)

cliffordwolf:
> Also, there are a number of typos in ezsat and subcircuit libraries, but I wasn't able to find the upstream source for these. Do you have a repository somewhere I should add a PR against, or should I just fix them here?

There is no other repo for those two libraries. This is already upstream. They are just in `libs/` because they are written in a manner that resembles a self-contained library.

whitequark:
All fixed.

passes/cmds/select.cc
```diff
@@ -987,7 +987,7 @@ struct SelectPass : public Pass {
 		log("list of selected objects.\n");
 		log("\n");
 		log("Note that many commands support an optional [selection] argument that can be\n");
-		log("used to YS_OVERRIDE the global selection for the command. The syntax of this\n");
+		log("used to override the global selection for the command. The syntax of this\n");
```
cliffordwolf: Hahaha!  Looks like this was collateral damage when we switched all users of the C++ keyword `overwrite` to the `YS_OVERWRITE` macro.

whitequark: Yes. I was greatly amused by this one.

daveshah1: https://www.newscientist.com/letter/mg16221857-700-ocralomania/

PR #770, opened 2019-01-02
opt_expr: refactor and improve simplification of comparisons

whitequark:
See individual commit messages. Only the first and the last commit intend to introduce any functional change; the rest is just refactoring, although I think it would incidentally avoid some aborts the original code warns about.

### DIFF
--- a/passes/opt/opt_expr.cc
+++ b/passes/opt/opt_expr.cc
@@ -1344,6 +1344,75 @@ void replace_const_cells(RTLIL::Design *design, RTLIL::Module *module, bool cons
 			}
 		}
 
+		// simplify comparisons
+		// currently, only replaces comparisons with an extreme of the signal range with a constant
+		// TODO: since, unlike the following optimization, this one does not assume that both inputs to the cell are constant,
+		// it is more robust; the following optimization should be folded into this one at some point
+		if (do_fine && (cell->type == "$lt" || cell->type == "$ge" || cell->type == "$gt" || cell->type == "$le"))
+		{
+			IdString cmp_type = cell->type;
+			SigSpec var_sig = cell->getPort("\\A");
+			SigSpec const_sig = cell->getPort("\\B");
+			int var_width = cell->parameters["\\A_WIDTH"].as_int();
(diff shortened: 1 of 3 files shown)


passes/opt/opt_expr.cc
```diff
@@ -259,6 +259,20 @@ bool is_one_or_minus_one(const Const &value, bool is_signed, bool &is_negative)
 	return last_bit_one;
 }
 
+int get_highest_hot_index(RTLIL::SigSpec signal)
+{
+	for (int i = GetSize(signal) - 1; i >= 0; i--)
+	{
+		if (signal[i] == RTLIL::State::S0)
+			continue;
```
cliffordwolf: Wouldn't the function do exactly the same thing without those two lines, since `(signal[i] == RTLIL::State::S0)` would imply `(signal[i] != RTLIL::State::S1)`.

whitequark: What about Sx?

cliffordwolf: Sx would also imply `(signal[i] != RTLIL::State::S1)`, thus we'd reach the end of the loop body, which effectively is the same as `continue`. Same for non-constant signal bits.

whitequark: hm, you are right. What is the proper way to deal with Sx here anyhow? I think I tried to do something but failed.

passes/opt/opt_expr.cc
```diff
@@ -259,6 +259,20 @@ bool is_one_or_minus_one(const Const &value, bool is_signed, bool &is_negative)
...
+
+		if (signal[i] == RTLIL::State::S1)
+			return i;
+	}
```
whitequark: I think I wanted a `break` here or something.

cliffordwolf: Oh, yeah, that would make sense then. (I have to admit I haven't had a close look at the users of `get_highest_hot_index()`. So I don't really know which behavior makes sense, just noticed that the first `if ()` statement didn't contribute to the behavior of the function.)

whitequark: Fixed. The handling of x should be conservative...

PR #771, opened 2019-01-02
cmp2lut: new techmap pass

whitequark:
Certain arithmetic operations between a signal of width n and a constant can be directly mapped to a single k-LUT (where n <= k). This is preferable to normal alumacc techmapping process because for many targets, arithmetic techmapping creates hard logic (such as carry cells) which often cannot be optimized further.

For example, a naively written (`{ w>0, w>1, w>2, ... }`) expansion of mask width (4 bit signal) to mask (16 bit signal) currently expands to 34 SB_CARRY cells and 25 SB_LUT4 cells in Yosys. But, it obviously should only expand to 16 SB_LUT4 cells.

Currently, only comparisons with 1-bit output are mapped. Potentially, all arithmetic cells with n <= k inputs should be techmapped in this way, because this shortens the critical path from n to 1 by avoiding carry chains. However, this is not currently done, since I have not yet thought of every potential ramification.


whitequark:
See [this attachment](https://github.com/YosysHQ/yosys/files/2720202/gateware.zip) for an example design. Interestingly, (after #772), both `synth -lut 4`, which purely uses abc to minimize logic, and `synth_ice40`, which uses this mapper before `alumacc`, arrive at the same 14 LUT result.

The core of the issue here is that abc does not understand carry logic (or LUTs, I think?). Do you have any suggestions to address this, apart from me continuing to write somewhat ad-hoc passes that replace parts of abc in sometimes nicer and sometimes less nice ways? I don't mind but I feel like we're missing some kind of structural opportunity here.

whitequark:
OK, I have this idea for a pass: mine the *coarse* circuit for any cells with four inputs (or less), then convert them directly to LUTs with `ConstEval`.

There are some problems with it though:
  * Right now this will only ever bring any benefit for cases like ice40 techmapping, that map directly to entities that abc doesn't understand. (On the other hand the benefit here can be considerable, given the potential to turn an n-hop critical path to 1-hop.)
  * As far as I can tell abc (or at least the abc pass) doesn't understand LUTs either, so any logic pre-mapped like that would not be available for further mapping. So, this means that only *exactly* k input subcircuits should be mapped.

So it's something like this: so long as we're using abc in a way similar to the way we currently are, a pass like this brings specific, potentially significant benefits and should be merged. If we are *not* using abc, then we can make a *more* generic version of this pass and rely on `opt_lut` to handle the cases where it generates inefficient logic. But before we can really use that more generic version, `opt_lut` needs to gain a logic minimization routine and probably functional understanding of carry logic too (fortunately, we can just rely on `ConstEval` for that.)

whitequark:
I think I will implement a pass that uses the [FlowMap algorithm](http://cadlab.cs.ucla.edu/~cong/papers/iccad92.pdf) and see where that brings us.

PR #783, opened 2019-01-07
bugpoint: new pass

whitequark:
A typical use of `bugpoint` would involve a script with a pass under test, e.g.:

    # flowmap.ys
    flowmap -relax -optarea 100

and would be invoked as:

    bugpoint -yosys ./yosys -script flowmap.ys -clean -cells

This replaces the current design with the minimal design that still crashes the `flowmap.ys` script.

`bugpoint` can also be used to perform generic design minimization using `select`, e.g. the following script:

    select i:* %x t:$_MUX_ %i -assert-max 0

would remove all parts of the design except for an unbroken path from an input to an output port that goes through exactly one `$_MUX_` cell. (The condition is inverted.)

### DIFF
--- a/passes/cmds/Makefile.inc
+++ b/passes/cmds/Makefile.inc
@@ -29,4 +29,4 @@ OBJS += passes/cmds/chformal.o
 OBJS += passes/cmds/chtype.o
 OBJS += passes/cmds/blackbox.o
 OBJS += passes/cmds/ltp.o
-
+OBJS += passes/cmds/bugpoint.o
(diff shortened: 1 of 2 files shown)


cliffordwolf:
JFYI: Switched it from `system()` to `run_command()` in b5f6e78. (In this case `run_command()` is just a wrapper for `system()`, so it's really not important.)

whitequark:
Ah, great, I haven't realized that function exists.

PR #786, opened 2019-01-13
Pattern Matcher Generator and iCE40 DSP Mapper

cliffordwolf:
Initial work on a code generator for structural subgraph pattern matchers, aimed at simplifying custom technology mappers for coarse non-memory blocks such as "DSP" cells in FPGA architectures.


cliffordwolf:
Simple example:

```
module test (
	input clk,
	input [7:0] A, B, C,
	output reg [15:0] X, Y,
	output [15:0] Z
);
	reg [7:0] AB;
	reg [7:0] AQ, BQ;
	always @(posedge clk) begin
		AQ <= A;
		BQ <= B;
		AB <= A + B;
		X <= AB * C;
		Y <= A * B;
	end
	assign Z = AQ * BQ;
endmodule
```

Running `yosys -p 'prep; show; ice40_dsp' test.v` will produce the following three matches:

```
mul: $mul$test.v:13$3
ffA: $procdff$8
ffB: --
ffY: $procdff$7

mul: $mul$test.v:14$4
ffA: --
ffB: --
ffY: $procdff$6

mul: $mul$test.v:16$5
ffA: $procdff$9
ffB: $procdff$10
ffY: --
```

And the following circuit diagram:

image

cliffordwolf:
Updated example:

```
module test (
	input clk, rst,
	input [7:0] A, B, C,
	output [15:0] W,
	output reg [15:0] X, Y, Z
);
	reg [7:0] AB;
	reg [7:0] AQ, BQ;
	always @(posedge clk) begin
		AQ <= A;
		BQ <= B;
		AB <= A + B;
		X <= AB * C;
		Y <= A * B;
		Z <= B * C + Z;

		if (rst)
			Z <= 0;
	end
	assign W = AQ * BQ;
endmodule
```

image

```
ffA:   $procdff$13
ffB:   --
mul:   $mul$test.v:13$3
ffY:   $procdff$11
addAB: --
muxAB: --
ffS:   --

ffA:   --
ffB:   --
mul:   $mul$test.v:14$4
ffY:   $procdff$10
addAB: --
muxAB: --
ffS:   --

ffA:   --
ffB:   --
mul:   $mul$test.v:15$5
ffY:   --
addAB: $add$test.v:15$6
muxAB: $procmux$8
ffS:   $procdff$12

ffA:   $procdff$14
ffB:   $procdff$15
mul:   $mul$test.v:20$7
ffY:   --
addAB: --
muxAB: --
ffS:   --
```

cliffordwolf:
With this latest version, `synth_ice40 -dsp` will map stuff to `SB_MAC16` cells.

passes/pmgen/README.md
```diff
@@ -0,0 +1,224 @@
...
+There is no automatic management of udata variables. For this reason it is
+recommended that the user-supplied matcher code treats them as read-only
+variables.
+
+They are declared like state variables, just using the `udata` statement:
+
+    udata <int> min_data_width max_data_width
+    udata <IdString> data_port_name
+
+They are atomatically initialzed to the default constructed value of their type
```
daveshah1: automatically initialized

passes/pmgen/ice40_dsp.pmg
```diff
@@ -0,0 +1,160 @@
...
+
+match mul
+	select mul->type.in($mul)
+	select GetSize(mul->getPort(\A)) + GetSize(mul->getPort(\B)) > 10
+	select GetSize(mul->getPort(\Y)) > 10
+endmatch
+
+match ffA
+	select ffA->type.in($dff)
+	// select nusers(port(ffA, \Q)) == 2
```
daveshah1: Is there a reason for this check being commented out?

passes/pmgen/ice40_dsp.cc
```diff
@@ -0,0 +1,237 @@
...
+			log(" ffY:%s", log_id(pm.st.ffY));
+
+		if (pm.st.ffS)
+			log(" ffS:%s", log_id(pm.st.ffS));
+
+		log("\n");
+	}
+	else
+	{
+		cell->setPort("\\CLK", State::S0);
```
daveshah1: icecube2 convention is unused clock to be disconnected rather than 1'b0, otherwise it will explicitly route 0 even though its safe to be floating, although we could fix this in nextpnr

passes/pmgen/ice40_dsp.cc
```diff
@@ -0,0 +1,237 @@
...
+		cell->setPort("\\CE", State::S0);
+		cell->setParam("\\NEG_TRIGGER", State::S0);
+	}
+
+	// SB_MAC16 Cascade Interface
+
+	cell->setPort("\\SIGNEXTIN", State::Sx);
+	cell->setPort("\\SIGNEXTOUT", pm.module->addWire(NEW_ID));
+
+	cell->setPort("\\CI", State::Sx);
```
daveshah1: More of an FYI, but important to note that CI/CO are regular signals available to the fabric without any dedicated routing whereas the SIGNEXT and ACCUMCI/O signals use the dedicated cascade routing only.

passes/pmgen/ice40_dsp.pmg
cliffordwolf: Yes. Even if the FF has multiple users we can still use it here because it's an input FF, so we can just in essence duplicate it. But we can't do that with the `D` input of `ffY`, so we need the check there.

passes/pmgen/ice40_dsp.cc
cliffordwolf: Maybe we should just automatically disconnect constant drivers if they drive the "default" value? This way we would also improve behavior for manual instances that simply set CLK to zero. We could do that either in a dedicated Yosys pass or in nextpnr.

passes/pmgen/ice40_dsp.cc
cliffordwolf: Thanks. Good to know.

passes/pmgen/ice40_dsp.cc
daveshah1: This is special-cased in nextpnr at the moment: https://github.com/YosysHQ/nextpnr/blob/master/ice40/pack.cc#L303-L329

Clock is not dropped when tied to zero because IIRC the vendor tools do this too, although Lattice do have examples where it is floating.

PR #789, opened 2019-01-17
WIP: Add IOLogic simulation models to cells_sim for ecp5

C-Elegans:


techlibs/ecp5/cells_sim.v
```diff
@@ -547,3 +547,286 @@ module DP16KD(
...
+	sr = !(gsr && pur);
+      else
+	sr = !pur;
+    wire  rst_internal = RST || sr;
+
+    reg   last_clock;
+    always @(SCLK)
+      last_clock <= SCLK;
+    
+    always @(SCLK or rst_internal) begin
```
daveshah1: I think it should be possible to write this using more normal `always @(posedge SCLK or posedge rst_internal)` and `always @(negedge SCLK or posedge rst_internal)` blocks - this is probably needed for correct simulation in Verilator

nakengelhardt:
@daveshah1 I don't think OP is coming back to complete this - what do you think, should we just close this PR or is there something here we should keep?

daveshah1:
Will close for now, a few parts may be useful but I plan a slightly different approach overall.

PR #794, opened 2019-01-21
ECP5 Improvements

daveshah1:
- Increase threshold for mapping ALUs (resolving a QoR issue with TMDS encoders)
 - Adding a few more blackboxes
 - Adding flipflop PRLD support
 - Adding flipflop init support (where hardware allows)


daveshah1:
Something seems to be up with flipflop initialisation, hold merging for now.

daveshah1:
All seems to be working well, this also adds some more blackboxes and a Migen fix.

techlibs/ecp5/cells_sim.v
```diff
@@ -556,3 +556,20 @@ module DP16KD(
   parameter INITVAL_3E = 320'h00000000000000000000000000000000000000000000000000000000000000000000000000000000;
   parameter INITVAL_3F = 320'h00000000000000000000000000000000000000000000000000000000000000000000000000000000;
 endmodule
+
+// For Migen compatibility
+module FD1S3BX(input PD, D, CK, output Q);
```
whitequark: Shouldn't this be fixed in Migen? It seems odd to support a primitive *just* for Migen.

daveshah1: It is a primitive that Diamond supports, so it seems fair to support in Yosys (really we should support all the Diamond flipflop primitives)

whitequark: So it's more of a Diamond compatibility then...

daveshah1: Changed comment to that effect and added a FIXME

PR #801, opened 2019-01-28
Don't strip executable files during `make install` (fixes #795)

### LINKED ISSUE #795: MacOS: dyld: malformed mach-o image: dylib load command #12 string extends beyond end of load command
Hello,

I'm using yosys on MacOS Mojave 10.14.2 installed using homebrew:

```
$ brew install yosys
...
$ brew info yosys
yosys: stable 0.8 (bottled), HEAD
Framework for Verilog RTL synthesis
http://www.clifford.at/yosys/
/usr/local/Cellar/yosys/0.8_1 (97 files, 19.2MB)
  Poured from bottle on 2019-01-22 at 12:57:47
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/yosys.rb
==> Dependencies
Build: bison ✘, pkg-config ✔
Required: libffi ✔, python ✔, readline ✔
==> Options
--HEAD
	Install HEAD version
==> Analytics
install: 77 (30 days), 181 (90 days), 474 (365 days)
install_on_request: 77 (30 days), 181 (90 days), 474 (365 days)
build_error: 0 (30 days)
```

When I try to execute yosys I get an error:

```
$ yosys
dyld: malformed mach-o image: dylib load command #12 string extends beyond end of load command
Abort trap: 6
```

I'm new to MacOS thus I'm not sure what exactly causes the issue. A little piece of advice would be much appreciated.

afiskon:
Don't execute `strip` on installed binaries. Strip may damage executable files on some systems, e.g. MacOS 10.14.2.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,7 +71,6 @@ PLUGIN_LDFLAGS :=
 PKG_CONFIG ?= pkg-config
 SED ?= sed
 BISON ?= bison
-STRIP ?= strip
 AWK ?= awk
 
 ifeq ($(OS), Darwin)
@@ -617,20 +616,10 @@ clean-unit-test:
 install: $(TARGETS) $(EXTRA_TARGETS)
 	$(INSTALL_SUDO) mkdir -p $(DESTDIR)$(BINDIR)
 	$(INSTALL_SUDO) cp $(TARGETS) $(DESTDIR)$(BINDIR)
-ifneq ($(filter yosys,$(TARGETS)),)
-	$(INSTALL_SUDO) $(STRIP) -S $(DESTDIR)$(BINDIR)/yosys
-endif
-ifneq ($(filter yosys-abc,$(TARGETS)),)
-	$(INSTALL_SUDO) $(STRIP) $(DESTDIR)$(BINDIR)/yosys-abc
-endif
-ifneq ($(filter yosys-filterlib,$(TARGETS)),)
-	$(INSTALL_SUDO) $(STRIP) $(DESTDIR)$(BINDIR)/yosys-filterlib
-endif
 	$(INSTALL_SUDO) mkdir -p $(DESTDIR)$(DATDIR)
 	$(INSTALL_SUDO) cp -r share/. $(DESTDIR)$(DATDIR)/.
 ifeq ($(ENABLE_LIBYOSYS),1)
 	$(INSTALL_SUDO) cp libyosys.so $(DESTDIR)$(LIBDIR)
-	$(INSTALL_SUDO) $(STRIP) -S $(DESTDIR)$(LIBDIR)/libyosys.so
 	$(INSTALL_SUDO) ldconfig
 endif
 


afiskon:
The original change was introduced in 57e02b662. The comment says `Strip debug symbols from binaries on install`. Using strip to remove debug symbols was not the right approach from the beginning. To exclude debug symbols corresponding compiler arguments should be used.

As a side note personally I don't think that excluding debug symbols is necessarily. The size of executable files is not a problem on modern systems, even on Raspberry Pi. In the same time debug symbols may help users to debug some bug.

cliffordwolf:
> This is definitely not the right way to solve this.

Agreed 100%.

> The comment says Strip debug symbols from binaries on install. Using strip to remove debug symbols was not the right approach from the beginning.

Unfortunately you don't include a rationale for this conclusion. I don't agree.

The rationale for using strip is that you can debug the built binaries easily without having to make changes to the Makefile to include debug symbols and rebuilding everything.

> The size of executable files is not a problem on modern systems, even on Raspberry Pi.

We literally added strip because users on Raspberry Pi complained about the size of the executable files.

afiskon:
> Unfortunately you don't include a rationale for this conclusion.

Building yosys without `-g` doesn't break Yosys on MacOS. Using `strip` does. I don't see what else rationale is necessary. Currently Yosys is literally broken on MacOS. Working software > any optimization.

cliffordwolf:
As whitequark said, macOS issues should not affect people who do not use macOS.

I'll be more than happy to merge a proper fix for the issue, but just breaking the intended behavior on all platforms because you can't provide a clean fix for macOS isn't an option.

afiskon:
I wold like to point out that the "intended behavior" caused the issue in the first place. Proposed patch just rolls back this change. As I compromise I suggest to skip execution of `strip` if the Makefile see that the system is MacOS.

cliffordwolf:
The intended behavior is that installed binaries are stripped of debug symbols, but the binaries in the build directory aren't. The implied claim that stripping binaries of debug symbols inherently produces broken binaries is not plausible imo.

And it is not a "Raspberry Pi issue". Raspberry Pi users simply are the user group that are hurt most by not stripping debug symbols. But we don't want debug symbols in the installed binaries on any system.

justMaku:
Thanks @q3k for nerd-sniping me into this. I was hoping to find some bizarre bug, but unfortunately I have to report that it just works fine for me without any changes required, tested on both 14.2.2 and 14.2.3 with building from master and installing from homebrew.

umenthum:
I think afiskon is trying to suggest that the build process be parameterized for debug (-g or -Og) and release (-O3 for performance or -Os for binary size minimization), eliminating the need for strip.

afiskon:
Never mind, apparently there is a workaround https://github.com/YosysHQ/yosys/issues/795#issuecomment-463260424

sur5r:
Another option would be to use `--strip-debug` or `--strip-unneeded` to prevent important information from being removed. This should be fine across all platforms.

q3k:
`strip` for yosys is already being run with `--strip-debug` (`-S` is a shorthand)

PR #805, opened 2019-02-06
write_verilog to write initial statement for initial flop state

eddiehung:
Otherwise testcase tests/simple/dff_init.v fails

Also added tests/simple_defparam/run-test.sh which copies all `*.v` from tests/simple and runs them with `write_verilog -defparam` to test that path

### DIFF
--- a/backends/verilog/verilog_backend.cc
+++ b/backends/verilog/verilog_backend.cc
@@ -1065,43 +1065,46 @@ bool dump_cell_expr(std::ostream &f, std::string indent, RTLIL::Cell *cell)
 			use_rd_clk = cell->parameters["\\RD_CLK_ENABLE"].extract(i).as_bool();
 			rd_clk_posedge = cell->parameters["\\RD_CLK_POLARITY"].extract(i).as_bool();
 			rd_transparent = cell->parameters["\\RD_TRANSPARENT"].extract(i).as_bool();
+			if (use_rd_clk)
 			{
-				std::ostringstream os;
-				dump_sigspec(os, sig_rd_clk);
-				clk_domain_str = stringf("%sedge %s", rd_clk_posedge ? "pos" : "neg", os.str().c_str());
-				if( clk_to_lof_body.count(clk_domain_str) == 0 )
-					clk_to_lof_body[clk_domain_str] = std::vector<std::string>();
-			}
-			if (use_rd_clk && !rd_transparent)
-			{
-				// for clocked read ports make something like:
-				//   reg [..] temp_id;
-				//   always @(posedge clk)
-				//      if (rd_en) temp_id <= array_reg[r_addr];
-				//   assign r_data = temp_id;
-				std::string temp_id = next_auto_id();
-				lof_reg_declarations.push_back( stringf("reg [%d:0] %s;\n", sig_rd_data.size() - 1, temp_id.c_str()) );
 				{
 					std::ostringstream os;
-					if (sig_rd_en != RTLIL::SigBit(true))
+					dump_sigspec(os, sig_rd_clk);
+					clk_domain_str = stringf("%sedge %s", rd_clk_posedge ? "pos" : "neg", os.str().c_str());
+					if( clk_to_lof_body.count(clk_domain_str) == 0 )
+						clk_to_lof_body[clk_domain_str] = std::vector<std::string>();
+				}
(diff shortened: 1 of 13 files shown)


eddiehung:
Calling @daveshah1 to review, SVP

backends/verilog/verilog_backend.cc
```diff
@@ -1247,7 +1247,14 @@ void dump_cell(std::ostream &f, std::string indent, RTLIL::Cell *cell)
 	dump_attributes(f, indent, cell->attributes);
 	f << stringf("%s" "%s", indent.c_str(), id(cell->type, false).c_str());
 
-	if (!defparam && cell->parameters.size() > 0) {
+	std::string init;
+	if (cell->name[0] == '$' && reg_ct.count(cell->type) && cell->hasPort("\\Q")) {
```
daveshah1: What is the reason for `cell->name[0] == '$' `?

eddiehung: I pilfered it from https://github.com/YosysHQ/yosys/blob/e112d2fbf5a31f00ef19e6d05f28fecc1e9c56b9/backends/verilog/verilog_backend.cc#L301

 I imagine it is to save a hash lookup since all internal primitives begin with `$`?

daveshah1: I think this should be `cell->type`, otherwise this is checking if cells have an internal name. Although not sure about that other usage, need to look further.

daveshah1: Yes, looks like the other usage is to do with renaming internal cell names, hence checking if name is internal rather than an optimisation.

eddiehung: Yes indeed, let me fix that up! Wonder why it works at all... [git blame](https://github.com/YosysHQ/yosys/blame/e112d2fbf5a31f00ef19e6d05f28fecc1e9c56b9/backends/verilog/verilog_backend.cc#L433) shows @cliffordwolf

daveshah1: No, the existing place with that check is correct I think, because it is part of the logic to prevent internally named signals when `-norename` isn't specified. But that doesn't apply to the usage when looking for init here.

eddiehung: Roger that. Removed that first clause entirely now.

cliffordwolf:
I don't think that's the right way to solve this issue as I don't think having two competing ways of representing FF initialization is an improvement.

Alternative suggestion: Add something like `initial <ff_name>.Q = <value>;` next to the module instantiation when dealing with initialized FFs in `-noexpr` mode, when the new option `-ffinit` is present. Also ugly, but at least it's a hack that's much more limited in scope.

eddiehung:
> Alternative suggestion: Add something like `initial <ff_name>.Q = <value>;`
Possible problem with that, though, is that some synthesis tools may not support (even trivial) hierarchical references?

cliffordwolf:
> Possible problem with that, though, is that some synthesis tools may not support (even trivial) hierarchical references?

What kind of tool would process Yosys Verilog netlists generated with `-noexpr`?

You'd always either produce a behavioral verilog output (i.e. without `-noexpr`) or map your circuit to a technology library that has initialized FFs (see for example `ice40_ffinit` for the latter).

eddiehung:
> What kind of tool would process Yosys Verilog netlists generated with `-noexpr`?
> 
> You'd always either produce a behavioral verilog output (i.e. without `-noexpr`) or map your circuit to a technology library that has initialized FFs (see for example `ice40_ffinit` for the latter).

Fair point. My particular point of concern was that the `iverilog` that we use in tests may not support it, but it looks like that issue was unfounded.

eddiehung:
> I don't think having two competing ways of representing FF initialization is an improvement.

For my understanding: were you referring to a second competing way for power on init than using parameters (my first attempt) and initial-hier-ref (current attempt)?

cliffordwolf:
> > I don't think having two competing ways of representing FF initialization is an improvement.
>
> For my understanding: were you referring to a second competing way for power on init than using parameters (my first attempt) and initial-hier-ref (current attempt)?

I mean internal in Yosys we don't need a second way. The canonical way is `init` attributes on the initialized wires.

eddiehung:
@cliffordwolf Okay to merge what we have now then?

PR #812, opened 2019-02-15
Define basic_cell_type() function and use it to derive the cell type …

ucbjrl:
…for array references (instead of duplicating the code).

### DIFF
--- a/passes/hierarchy/hierarchy.cc
+++ b/passes/hierarchy/hierarchy.cc
@@ -523,15 +523,15 @@ int find_top_mod_score(Design *design, Module *module, dict<Module*, int> &db)
 			// Is this an array instance
 			if (celltype.substr(0, 7) == "$array:") {
 				celltype = basic_cell_type(celltype);
-				// Is this cell is a module instance?
-				if (celltype[0] != '$') {
-					auto instModule = design->module(celltype);
-					// If there is no instance for this, issue a warning.
-					if (instModule == NULL) {
-						log_warning("find_top_mod_score: no instance for %s.%s\n", celltype.c_str(), cell->name.c_str());
-					}
-					if (instModule != NULL)
-						score = max(score, find_top_mod_score(design, instModule, db) + 1);
+			}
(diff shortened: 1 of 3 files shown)


passes/hierarchy/hierarchy.cc
```diff
@@ -502,9 +518,23 @@ int find_top_mod_score(Design *design, Module *module, dict<Module*, int> &db)
...
+				if (celltype[0] != '$') {
+					auto instModule = design->module(celltype);
+					// If there is no instance for this, issue a warning.
+					if (instModule == NULL) {
+						log_warning("find_top_mod_score: no instance for %s.%s\n", celltype.c_str(), cell->name.c_str());
+					}
+					if (instModule != NULL)
+						score = max(score, find_top_mod_score(design, instModule, db) + 1);
+				}
+			}
```
cliffordwolf: Afaict this code sets the score for non-array types to constant zero. That doesn't look right at all.

ucbjrl: Good catch @cliffordwolf. The surgery removed an excessive amount of viable tissue.
Will fix.

passes/hierarchy/hierarchy.cc
```diff
@@ -502,9 +518,23 @@ int find_top_mod_score(Design *design, Module *module, dict<Module*, int> &db)
...
-				score = max(score, find_top_mod_score(design, design->module(cell->type), db) + 1);
+		for (auto cell : module->cells()) {
+			std::string celltype = cell->type.str();
+			// Is this an array instance
+			if (celltype.substr(0, 7) == "$array:") {
+				celltype = basic_cell_type(celltype);
+			}
+			// Is this cell a module instance?
+			if (celltype[0] != '$') {
+				auto instModule = design->module(celltype);
```
cliffordwolf: It should really just check if `design->module(celltype)` is non-nullptr, like the old code. There are various scenarios where a module instance is of a type that starts with `$`, for example `$paramod$...` modules.

PR #819, opened 2019-02-21
Rename "yosys -D" to "yosys -U", add "yosys -D" with expected behavior

cliffordwolf:

### DIFF
--- a/kernel/driver.cc
+++ b/kernel/driver.cc
@@ -272,7 +272,7 @@ int main(int argc, char **argv)
 		printf("    -D <macro>[=<value>]\n");
 		printf("        set the specified Verilog define (via \"read -define\")\n");
 		printf("\n");
-		printf("    -U <header_id>[:<filename>]\n");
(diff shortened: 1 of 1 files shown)


kernel/driver.cc
```diff
@@ -268,7 +269,10 @@ int main(int argc, char **argv)
 		printf("    -A\n");
 		printf("        will call abort() at the end of the script. for debugging\n");
 		printf("\n");
-		printf("    -D <header_id>[:<filename>]\n");
+		printf("    -D <macro>[=<value>]\n");
+		printf("        set the specified Verilog define (via \"read -define\")\n");
+		printf("\n");
+		printf("    -U <header_id>[:<filename>]\n");
```
whitequark: `-U` seems like the inverse of `-D`. What about `-P` instead?

cliffordwolf: SGTM. Changed it accordingly.

PR #820, opened 2019-02-21
Fix #810 and fix #814

### LINKED ISSUE #810: SegFault in genrtlil
I traced a segfault down to the current_ast->dumpAst(*) line below at line 1568 in genrtlil.cc.  As far as I can tell, this was created from calling $past() on an object of zero width.  (Yes, this is a bug, but Yosys shouldn't segfault on this bug.  Fixing the width in the source file eliminated the segfault too.)

    // everything should have been handled above -> print error if not.
    default:
    	for (auto f : log_files)
    		current_ast->dumpAst(f, "verilog-ast> ");
    	type_name = type2str(type);
    	log_file_error(filename, linenum, "Don't know how to generate RTLIL code for %s node!\n", type_name.c_str());

## Expected behavior

An error message indicating the line of code with a problem.  (The simple test cases I've created to try to replicate this behavior have failed to do so.)

It seems as though under certain conditions (I'm not quite sure which), current_ast is never getting set.  A further look into the source, suggests this call should be with respect to this and not current_ast--at least in those cases where current_ast is NULL.  Changing current_ast to "this" fixed the problem and produced a valuable error message.

### LINKED ISSUE #814: Yosys assertion error, rtlil.cc:3277

Using the attached files, run "read -formal test.v".  Yosys then crashes before a next command can be issued.

## Expected behavior

1. Minimum: I would like to see an error message indicating what line of my file is broken
2. Because of the `default_nettype none line, I would expect to see an error before yosys attempts to elaborate using an unknown identifier.

## Actual behavior

    yosys> read -formal test.v
    1. Executing Verilog-2005 frontend.
    Parsing formal SystemVerilog input from `test.v' to AST representation.
    ERROR: Assert `chunk.width >= 0' failed in kernel/rtlil.cc:3277.

This appears to be related to [this other issue](https://github.com/YosysHQ/yosys/issues/579), but the example test looks quite different.

Here's the test.v file.

[test.zip](https://github.com/YosysHQ/yosys/files/2881326/test.zip)

Dan

cliffordwolf:


cliffordwolf:
Sorry for throwing the Travis fix mess in here. But I had it breaking in this branch so I decided to have a stab at it.

ZipCPU:
@cliffordwolf : Approved.  Thank you!  I'm especially pleased with the #814 fix that now properly creates a default-nettype-none error.

PR #822, opened 2019-02-21
Add -params mode to force undef parameters in selected cells.

litghost:
This enables synthesis output to tools that require that all parameter values have a particular value, but is more generic than something like https://github.com/YosysHQ/yosys/blame/master/techlibs/anlogic/anlogic_determine_init.cc#L26.  I was considering adding a parameters filter, but I'm not convinced it is required beyond the existing "select" mechanism.

### DIFF
--- a/passes/cmds/setundef.cc
+++ b/passes/cmds/setundef.cc
@@ -143,6 +143,9 @@ struct SetundefPass : public Pass {
 		log("    -init\n");
 		log("        also create/update init values for flip-flops\n");
 		log("\n");
+		log("    -params\n");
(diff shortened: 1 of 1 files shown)


litghost:
So sorry, there was a bug in this PR.  I'll submit a follow up to fix it.

litghost:
> So sorry, there was a bug in this PR. I'll submit a follow up to fix it. The review was so fast! I expected some feedback!

Looks like this was already fixed and made to conform with existing functions in d55790909c3b4244889d092c8eae630c7efd1aee.  Thanks!

PR #829, opened 2019-02-27
Dockerfile for Yosys

abdelrahmanhosny:
This pull request proposes to dockerize Yosys suite. To build the Docker image: `docker build -t yosys .`. To use the image, mount the directory where you have your RTL and library files to `/data` inside the image. There are two options to run yosys:

- Use the interactive shell: `docker run -it -v <data-directory>:/data yosys`. Read files from `/data` directory. For example: `read_verilog /data/design.v`
- Use the yosys commands from a file: `docker run -v <data-directory>:/data yosys /data/synth.ys`, where `synth.ys` should be located in the host data directory as well.

To try without building the image locally, use image name `openroad/yosys` on Docker Hub: https://hub.docker.com/r/openroad/yosys

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,47 @@
+FROM ubuntu:16.04 as builder
+LABEL author="Abdelrahman Hosny <abdelrahman.hosny@hotmail.com>"
+
+RUN apt-get update && apt-get install -y build-essential \
+    clang \
+    bison \
+    flex \
+    libreadline-dev \
+    gawk \
+    tcl-dev \
+    libffi-dev \
+    git \
+    graphviz \
+    xdot \
+    pkg-config \
+    python3
+
+COPY . /
+RUN make && \
+    make install
+
+
+FROM ubuntu:16.04
+RUN apt-get update && apt-get install -y clang \
+    bison \
+    flex \
+    libreadline-dev \
+    gawk \
+    tcl-dev \
+    libffi-dev \
+    git \
+    graphviz \
+    xdot \
+    pkg-config \
+    python3
+COPY --from=builder /yosys /build/yosys
+COPY --from=builder /yosys-abc /build/yosys-abc
+COPY --from=builder /yosys-config /build/yosys-config
+COPY --from=builder /yosys-filterlib /build/yosys-filterlib
+COPY --from=builder /yosys-smtbmc /build/yosys-smtbmc
+
+ENV PATH /build:$PATH
+
+RUN mkdir /data
+WORKDIR /data
+
+ENTRYPOINT ["yosys"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
-FROM ubuntu:16.04 as builder
+FROM ubuntu:18.04 as builder
 LABEL author="Abdelrahman Hosny <abdelrahman.hosny@hotmail.com>"
-
+ENV DEBIAN_FRONTEND=noninteractive
 RUN apt-get update && apt-get install -y build-essential \
     clang \
     bison \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,47 @@
+FROM ubuntu:16.04 as builder
```
q3k: Can we bump this to 18.04? That's the latest LTS.

Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+    clang \
+    bison \
+    flex \
+    libreadline-dev \
+    gawk \
+    tcl-dev \
+    libffi-dev \
+    git \
+    graphviz \
+    xdot \
```
q3k: Do we need graphviz/xdot when building?

Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+    pkg-config \
+    python3
+
+COPY . /
+RUN make && \
+    make install
+
+
+FROM ubuntu:16.04
+RUN apt-get update && apt-get install -y clang \
```
q3k: Do we need clang, bison, flex, libreadline-dev, tcl-dev, libffi-dev, git, pkg-config at runtime?

Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+    flex \
+    libreadline-dev \
+    gawk \
+    tcl-dev \
+    libffi-dev \
+    git \
+    graphviz \
+    xdot \
+    pkg-config \
+    python3
```
q3k:
```
    python && \
    rm -rf /var/lib/apt/lists
```

Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+COPY --from=builder /yosys /build/yosys
+COPY --from=builder /yosys-abc /build/yosys-abc
+COPY --from=builder /yosys-config /build/yosys-config
+COPY --from=builder /yosys-filterlib /build/yosys-filterlib
+COPY --from=builder /yosys-smtbmc /build/yosys-smtbmc
+
+ENV PATH /build:$PATH
+
+RUN mkdir /data
```
q3k: This should be up to the end-user of the Dockerfile, no?

Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+WORKDIR /data
+
+ENTRYPOINT ["yosys"]
```
q3k:
```
RUN useradd -m yosys
USER yosys
```

Dockerfile
q3k: (this likely pulls in X11, which we don't want)

Dockerfile
q3k: I mean, even for `-v foo:/data`, you don't need to create this directory first in the Dockerfile.

q3k:
Thanks! Let's sort out the above nitpicks, and I'd love to merge it. I'll also set up a yosys orga on dockerhub.

Dockerfile
q3k: (also the runtime container below)

abdelrahmanhosny: fixed

Dockerfile
abdelrahmanhosny: removed.

Dockerfile
abdelrahmanhosny: I only kept `libreadline-dev` and `tcl-dev` and removed the rest. They give the following errors when removed:
`error while loading shared libraries`

Dockerfile
abdelrahmanhosny: fixed.

Dockerfile
abdelrahmanhosny: You are right. I removed it.

Dockerfile
abdelrahmanhosny: added.

abdelrahmanhosny:
Thank you Serge for pointing out these improvements. I addressed them accordingly. The image size is now `553MB`. Let me know if you see more room for improvement.
I tested the resulting image on a sample design and it worked as expected.

XVilka:
Have you considered to use Alpine linux as container base instead? It will help to reduce image size substantially. See for example https://github.com/gliderlabs/docker-alpine

abdelrahmanhosny:
I tried `alpine` and the resulting image is `420MB` in size. I can push the Dockerfile if you think it is better and would prefer it over the `ubuntu` image.

q3k:
Let's get this merged as is now. We can then consider moving to an Alpine-based container.

PR #830, opened 2019-02-27
Reversed array indexing in EDIF backend

mkurc-ant:
I was trying to implement designs using Yosys and Vivado and stumbled upon a problem with signal array ordering in EDIF files. It seemed that all signal arrays were indexed in reverse order with respect to what Vivado expects.

In this PR I've simply reversed the array indexing order in EDIF backend. And now it is possible to pass EDIFs from Yosys to Vivado and everything works.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -16,6 +16,7 @@ Yosys 0.8 .. Yosys 0.8-dev
     - Added "gate2lut.v" techmap rule
     - Added "rename -src"
     - Added "equiv_opt" pass
+    - Reversed array inexing order in EDIF backend.
 
 
 Yosys 0.7 .. Yosys 0.8
--- a/backends/edif/edif.cc
+++ b/backends/edif/edif.cc
@@ -333,7 +333,7 @@ struct EdifBackend : public Backend {
 					*f << stringf("          (port (array %s %d) (direction %s))\n", EDIF_DEFR(wire->name, port_rename, b[0], b[1]), wire->width, dir);
 					for (int i = 0; i < wire->width; i++) {
 						RTLIL::SigSpec sig = sigmap(RTLIL::SigSpec(wire, i));
-						net_join_db[sig].insert(stringf("(portRef (member %s %d))", EDIF_REF(wire->name), GetSize(wire)-i-1));
+						net_join_db[sig].insert(stringf("(portRef (member %s %d))", EDIF_REF(wire->name), i));
 					}
 				}
 			}
@@ -385,12 +385,12 @@ struct EdifBackend : public Backend {
 						else if (sig.size() == 1)
 							net_join_db[sig[i]].insert(stringf("(portRef %s (instanceRef %s))", EDIF_REF(p.first), EDIF_REF(cell->name)));
 						else {
-							int member_idx = GetSize(sig)-i-1;
+							int member_idx = i;
 							auto m = design->module(cell->type);
 							if (m) {
 								auto w = m->wire(p.first);
 								if (w)
-									member_idx = GetSize(w)-i-1;
+									member_idx = i;
 							}
 							net_join_db[sig[i]].insert(stringf("(portRef (member %s %d) (instanceRef %s))",
 									EDIF_REF(p.first), member_idx, EDIF_REF(cell->name)));


eddiehung:
@mkurc-ant Thanks for the report and fix! Since you're the most qualified, can you drop a testcase in here that didn't work before, and now works? Thanks!

mkurc-ant:
@eddiehung Yes, I can do that. But the problem manifests itself only when running a design on an actual hardware (Yosys -> Vivado) flow. I can make an example design that demonstrates this problem but I think it will not be possible to test it automatically.

mkurc-ant:
I've created a simple test design for Bassy3 board. It is a free running counter which drives 15 LEDs. LED pins are defined as array.

But with the current master version of Yosys I cannot get the design to implement (I used to work on a different version). Vivado complains that it does not know about `GND` and `VCC` primitives:

> ERROR: [DRC INBB-1] Black Box Instances: Cell 'GND' of type 'GND/GND' has undefined contents and is considered a black box. This design will fail to generate a bitstream unless the partial reconfiguration feature is used.
ERROR: [DRC INBB-1] Black Box Instances: Cell 'VCC' of type 'VCC/VCC' has undefined contents and is considered a black box. This design will fail to generate a bitstream unless the partial reconfiguration feature is used.
ERROR: [DRC RTSTAT-1] Unrouted nets: 2 net(s) are unrouted. The problem bus(es) and/or net(s) are GND_NET, and VCC_NET.
WARNING: [DRC RTSTAT-10] No routable loads: 27 net(s) have no routable loads. The problem bus(es) and/or net(s) are $auto$alumacc.cc:474:replace_alu$6.C[1], $auto$alumacc.cc:474:replace_alu$6.C[2], $auto$alumacc.cc:474:replace_alu$6.C[3], $auto$alumacc.cc:474:replace_alu$6.C[5], $auto$alumacc.cc:474:replace_alu$6.C[6], $auto$alumacc.cc:474:replace_alu$6.C[7], $auto$alumacc.cc:474:replace_alu$6.C[9], $auto$alumacc.cc:474:replace_alu$6.C[10], $auto$alumacc.cc:474:replace_alu$6.C[11], $auto$alumacc.cc:474:replace_alu$6.C[13], $auto$alumacc.cc:474:replace_alu$6.C[14], $auto$alumacc.cc:474:replace_alu$6.C[15], $auto$alumacc.cc:474:replace_alu$6.C[17], $auto$alumacc.cc:474:replace_alu$6.C[18], $auto$alumacc.cc:474:replace_alu$6.C[19]... and (the first 15 of 27 listed).

There is another problem: When I export the EDIF using `write_blif -nogndvcc` command then it generates a malformed EDIF file. Like it was cut in half.

I can attach the test case to this PR. Can you point me out in which folder should I place it ?

eddiehung:
> @eddiehung Yes, I can do that. But the problem manifests itself only when running a design on an actual hardware (Yosys -> Vivado) flow. I can make an example design that demonstrates this problem but I think it will not be possible to test it automatically.

Not a problem. Just drop it in here for now. I just want to cross-validate (by hand) that it is indeed fixed  before merging the PR. While this verification step is by hand for now, I'd still like to collect all the data so that it could be made automatic in the future.

> But with the current master version of Yosys I cannot get the design to implement (I used to work on a different version). Vivado complains that it does not know about `GND` and `VCC` primitives

Sorry that's the case. Perhaps it is best that you file this as another issue, with the old version's hash, if you have it? The reason I'm pushing to collect all these test cases is to start a regression suite such that things that were known to be working before (like this!) don't suddenly break (again) as we see here.

> There is another problem: When I export the EDIF using `write_blif -nogndvcc` command then it generates a malformed EDIF file. Like it was cut in half.
>
> I can attach the test case to this PR. Can you point me out in which folder should I place it ?

Another issue report, please! Testcases can be attached to the issue for now, we're still working on an easy system for people to submit test cases.

Appreciate the cycles you spend on reporting these issues. Only together can we make Yosys better!

cliffordwolf:
See also 850f8299a9374ae4b9caeca8b6cd888be7c0a406 and #328.

The nice thing about EDIF is that there are different flavors.. :)

So afaict there are three different behaviors:
- Bit 0 is always LSB
- Bit 0 is always MSB
- Bit 0 is LSB for upto busses and MSB for downto busses
- Bit 0 is MSB for upto busses and LSB for downto busses

I think PRs that change back and forth between those behaviors are pointless. Instead, we should add an option that can be used to select the behavior the user wants.

Then we can add something like `-flavor vivado` as an alias for a certain set of options.

mkurc-ant:
Thanks. Yes, I was afraid that there are different EDIF flavors and that the patch will work only for Vivado.

I'think it would be good to have something like `-order msb` which will explicitly set bit ordering as well as different `-flavor` options suitable for particular toolchains/IDEs.

I can take on this feature. Can you provide me with some examples of FPGA toolchains or IDEs that expect different flavors so I will have something to test on ? I currently have only experience with Vivado.

eddiehung:
@mkurc-ant It would great to get your contribution! Let me know if you get stuck anywhere, and I'll try and help. Perhaps the scope of this PR would be to target:
* Default (everything as we have before this PR)
* `-flavour vivado` as an alias just for what you need to work in Vivado -- others can add their whatever flavours they need later

cliffordwolf:
According to the following experiment the current array behavior of `write_edif` is identical to the Vivado behavior (tested with current Yosys github master and Vivado 2018.2): http://svn.clifford.at/handicraft/2019/vivadoedif

After running `test.sh` one can find in `test_yosys.edn`:

```
          (net VCC_NET (joined
            (portRef (member o_downto 0))
            (portRef (member o_upto 0))
            (portRef P (instanceRef VCC))
          ))
```

And in `test_vivado.edn`:

```
         (net (rename &_const1_ "<const1>") (joined
          (portref P (instanceref VCC))
          (portref (member o_downto 0))
          (portref (member o_upto 0))
          )
         )
```

I.e. it looks like Vivado is always using the array element 0 for the MSB bit.

For completeness, this is the design (`test.v`):

```
module test (
	output reg [7:0] o_downto,
	output reg [0:7] o_upto
);
	always @* begin
		o_downto = 0;
		o_upto = 0;

		// set the MSB bit (index 0 in edif array)
		o_downto[7] = 1;
		o_upto[0] = 1;
	end
endmodule
```

mkurc-ant:
Thanks, I'll think how to test it thoroughly once I have some time.

Here is my design in which I've observed the bit reversal. It is a counter with output to LEDs. Works on Basys3 board.

counter.v:
```
module top (
    input  wire         clk,
    input  wire         rst,

    output wire [15:0] leds,
);

    localparam BITS  = 16;
    localparam DELAY = 20;

    reg [BITS+DELAY-1:0] counter = 0;

        if (rst) counter <= 0;
        else     counter <= counter + 1;
    end

    assign leds = counter >> DELAY;

endmodule
```

The Yosys script:
```
echo on

# Synth for Xilinx
synth_xilinx -flatten

# Write EDIF file
write_edif counter.edif
```

And the XDC file for pin mapping (for Vivado):
```
create_clock -period 10.000 -name clk [get_ports clk]

set_property PACKAGE_PIN W5  [get_ports clk]
set_property PACKAGE_PIN U18 [get_ports rst]

set_property PACKAGE_PIN U16 [get_ports leds[ 0]]
set_property PACKAGE_PIN E19 [get_ports leds[ 1]]
set_property PACKAGE_PIN U19 [get_ports leds[ 2]]
set_property PACKAGE_PIN V19 [get_ports leds[ 3]]
set_property PACKAGE_PIN W18 [get_ports leds[ 4]]
set_property PACKAGE_PIN U15 [get_ports leds[ 5]]
set_property PACKAGE_PIN U14 [get_ports leds[ 6]]
set_property PACKAGE_PIN V14 [get_ports leds[ 7]]
set_property PACKAGE_PIN V13 [get_ports leds[ 8]]
set_property PACKAGE_PIN V3  [get_ports leds[ 9]]
set_property PACKAGE_PIN W3  [get_ports leds[10]]
set_property PACKAGE_PIN U3  [get_ports leds[11]]
set_property PACKAGE_PIN P3  [get_ports leds[12]]
set_property PACKAGE_PIN N3  [get_ports leds[13]]
set_property PACKAGE_PIN P1  [get_ports leds[14]]
set_property PACKAGE_PIN L1  [get_ports leds[15]]

foreach port [get_ports] {
    set_property IOSTANDARD LVCMOS33 $port
}
```

In the design the LSB of the counter should go to LED[0] (pin U16) but in fact it goes to the LED[15] (pin L1)

cliffordwolf:
So I had a closer look at the EDIF files vivado generates and I think your problem has nothing to do with array indexing per se  but with the way that vivado renames nets. In an out-of context build:

```
      (interface
        (port (array (rename o_downto "o_downto[7:0]") 8) (direction OUTPUT))
        (port (array (rename o_upto "o_upto[0:7]") 8) (direction OUTPUT))
       )
```

And in a non-out-of-context build:

```
         (net (rename o_downto_0_ "o_downto[0]") (joined
          (portref O (instanceref o_downto_OBUF_0__inst))
          (portref (member o_downto 7))
          )
         )
         (net (rename o_upto_0_ "o_upto[0]") (joined
          (portref O (instanceref o_upto_OBUF_0__inst))
          (portref (member o_upto 0))
          )
         )
```

I've changes the Xilinx edif config in yosys to match this behavior in 13844c7.

cliffordwolf:
> I've changes the Xilinx edif config in yosys to match this behavior in 13844c7.

@mkurc-ant does this fix the issue you are seeing?

eddiehung:
Pinging @mkurc-ant -- does this resolve your issue?

mkurc-ant:
@cliffordwolf @eddiehung Sorry for late answer, I've been busy. I'm going to check that tomorrow (hopefully).

mkurc-ant:
@cliffordwolf @eddiehung Hello, I finally found some time to test that. And yes, https://github.com/YosysHQ/yosys/commit/13844c765818b5d8c0d16d62dcc530f688d2c28a solves the issue. Thanks.

Issue #831, opened 2019-02-27
smt2 support for partially initialized memory words is missing


cliffordwolf:
## Steps to reproduce the issue

test.v:
```
module test(input [1:0] addr, output [7:0] o);
	reg [7:0] rom [0:3];
	initial begin
		rom[0] <= 8'b 0100100x;
		rom[1] <= 8'b 00100x01;
		rom[2] <= 8'b 010x0010;
		rom[3] <= 8'b 0x010100;
	end
	assign o = rom[addr];
endmodule
```

test.ys:
```
read_verilog test.v
prep
write_smt2 test.smt2
```

## Expected behavior

Create an SMT2 file with partially constrained array members.

## Actual behavior

It creates an SMT2 file with 'x' characters in SMT2 constants, which of course isn't valid SMT2 code.

PR #833, opened 2019-02-28
Fix smt2 code generation for partially initialized memory words, fixe…

cliffordwolf:
…s #831


ZipCPU:
Thank you!  This fixes the original issue.

PR #839, opened 2019-03-01
Reduced error to warning in dfflibmap

trcwm:
Reduced error to warning in dfflibmap so cells with next_state expressions are ignored, instead of yosys bailing.

Next_state expressions are used in scanable/DFT flip flops, which are present in most ASIC tech libs.
Generating an error & exiting instead of warning and ignoring the cell makes Yosys unusable on all but the most basic techlibs.


trcwm:
PIBKAC.

PR #842, opened 2019-03-01
Changes required for VPR place and route in synth_xilinx

litghost:


manual/command-reference-manual.tex
```diff
@@ -4244,9 +4244,15 @@ \section{synth\_xilinx -- synthesis for Xilinx FPGAs}
         is omitted if this parameter is not specified.
 
     -vpr
-        generate an output netlist (and BLIF file) suitable for VPR
+        generate an output netlist (and BLIF file) suitable for VPR.
         (this feature is experimental and incomplete)
 
+    -no-brams
```
daveshah1: (nitpick) Tradition in Yosys is to use -nobram, -nodram, etc rather than -no-brams

litghost: Fixed

techlibs/xilinx/brams_map.v
```diff
@@ -30,8 +30,8 @@ module \$__XILINX_RAMB36_SDP (CLK2, CLK3, A1ADDR, A1DATA, A1EN, B1ADDR, B1DATA,
 		.RAM_MODE("SDP"),
 		.READ_WIDTH_A(72),
 		.WRITE_WIDTH_B(72),
-		.WRITE_MODE_A("READ_FIRST"),
-		.WRITE_MODE_B("READ_FIRST"),
+		.WRITE_MODE_A("WRITE_FIRST"),
+		.WRITE_MODE_B("WRITE_FIRST"),
```
cliffordwolf: why?

techlibs/xilinx/synth_xilinx.cc
```diff
@@ -137,6 +144,8 @@ struct SynthXilinxPass : public Pass
 		bool flatten = false;
 		bool retime = false;
 		bool vpr = false;
+		bool noBrams = false;
+		bool noDrams = false;
```
cliffordwolf: using camel case here in this form is inconsistent with the coding style of similar variables in other synth_ passes.

techlibs/xilinx/cells_sim.v
```diff
@@ -104,56 +122,75 @@ module CARRY4(output [3:0] CO, O, input CI, CYINIT, input [3:0] DI, S);
...
+  assign O = S ^ CI;
+endmodule
+
+`endif
+
 module FDRE (output reg Q, input C, CE, D, R);
   parameter [0:0] INIT = 1'b0;
-  parameter [0:0] IS_C_INVERTED = 1'b0;
-  parameter [0:0] IS_D_INVERTED = 1'b0;
-  parameter [0:0] IS_R_INVERTED = 1'b0;
```
cliffordwolf: This would break user code that sets those parameters. Those parameters are part of the cell descriptions provided and maintained by Xilinx. Removing them or rendering them ineffective is not an option.

cliffordwolf:
This PR is attempting to change a lot of things all in one big blop, some of them clearly independent of the others. Smaller PRs that are easy to review have a much better chance of being processed quickly.

techlibs/xilinx/brams_map.v
litghost: Default BRAM behavior per https://www.xilinx.com/support/documentation/sw_manuals/xilinx2017_2/ug953-vivado-7series-libraries.pdf is WRITE_FIRST.  It also specifically recommends WRITE_FIRST in some situations.

manual/command-reference-manual.tex
```diff
@@ -4244,9 +4244,15 @@ \section{synth\_xilinx -- synthesis for Xilinx FPGAs}
         is omitted if this parameter is not specified.
 
     -vpr
-        generate an output netlist (and BLIF file) suitable for VPR
+        generate an output netlist (and BLIF file) suitable for VPR.
```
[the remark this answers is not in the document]
litghost: Sure, will revert.

techlibs/xilinx/cells_sim.v
litghost: Fair enough.  The D and R inverts are actually interesting because they can not be used in all FF's.  In fact, the D invert can only be used on I/O FF's.  I'll revert the sim change, but I think the FF mapping is more correct, as most DFF's are not doing to be I/O FF's.

techlibs/xilinx/synth_xilinx.cc
```diff
@@ -63,6 +63,12 @@ struct SynthXilinxPass : public Pass
 		log("        generate an output netlist (and BLIF file) suitable for VPR\n");
 		log("        (this feature is experimental and incomplete)\n");
 		log("\n");
+		log("    -nobrams\n");
+		log("        disable infering of block rams\n");
+		log("\n");
+		log("    -nodrams\n");
+		log("        disable infering of distributed rams\n");
+		log("\n");
```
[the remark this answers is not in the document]
litghost: Done

techlibs/xilinx/synth_xilinx.cc
litghost: Done

techlibs/xilinx/cells_sim.v
litghost: Done

techlibs/xilinx/brams_map.v
cliffordwolf: That's irrelevant. What's relevant is if READ_FIRST or WRITE_FIRST produces the correct result for the given input verilog that we are mapping to a BRAM cell there. So why do you think READ_FIRST produces the wrong results and WRITE_FIRST produces the correct results?

litghost: That's valid.  However I think this has exposed a problem with the current BRAM synthesis in the case of asynchronous clocks.  READ_FIRST can only be used reliably if both clocks are synchronous (https://www.xilinx.com/support/documentation/user_guides/ug473_7Series_Memory_Resources.pdf section "Conflict Avoidance").  I believe the current brams.txt are listed as:
 - Independent clocks
 - Non-transparent reads (e.g. READ_FIRST)

The current inference configuration is not correct in the general case, specifically if the clocks are async and a read/write conflict occurs on separate ports.  The solution would appear to be to expose the BRAMs in more configurations:

- Synchronous clocks, transparency is configurable (e.g. READ_FIRST or WRITE_FIRST based on config) -> This is generally correct.
- Asynchronous clocks, transparent non-configurable -> WRITE_FIRST

The "make_transp" part in the brams.txt also appears to be incorrect, as that will emit logic for transparent reads, where-as the tech mapping should just map a transparent read to the WRITE_FIRST mode, rather than adding logic.

Thoughts?  If the above plan seems in the right direction, I'll isolate the BRAM's inference changes as they are fairly independent.

cliffordwolf: I think that just means that for independent clocks READ_FIRST doesn't give a guarantee about collision resolution, which is fine, because collisions are undefined anyways in Verilog for independent clocks. Afaict this means READ_FIRST is the safe setting to use here for all cases.

If you think otherwise I'd ask you to create a separate PR for this that includes a test case in the description that demonstrates that the current behavior produces incorrect results and your proposed behavior produces correct results.

litghost: The most recent tech library (Feb 5, 2019) has a table confirms what you said https://www.xilinx.com/support/documentation/user_guides/ug473_7Series_Memory_Resources.pdf Table 1-5, which is that READ_FIRST is safe for all non-transparent configurations.

There is a power advantage to exposing NO_CHANGE (~15%), but that is a optimization rather than correctness change.  Because of the power advantage, probably worth creating an issue for NO_CHANGE synthesis.  This is most straight forward in the SDP case.

litghost:
I believe this PR is ready if there are no further comments.

PR #843, opened 2019-03-01
Use mem2reg on memories that only have constant-index write ports

cliffordwolf:


eddiehung:
Original testcase that prompted this: [top.v.gz](https://github.com/YosysHQ/yosys/files/2921517/top.v.gz)

PR #846, opened 2019-03-02
[WIP/RFC] ecp5: 'Fix' transparent DRAMs by putting register on address

daveshah1:
I'm not convinced that this is the best way to solve the problem, but I'm working on getting liteeth working with Yosys on the ECP5.

Migen generates a lot of "write through" memories thus:

```verilog
reg [41:0] storage_4[0:63];
reg [5:0] memadr_5;
reg [5:0] memadr_6;
always @(posedge eth_rx_clk) begin
	if (ethmac_rx_cdc_wrport_we)
		storage_4[ethmac_rx_cdc_wrport_adr] <= ethmac_rx_cdc_wrport_dat_w;
	memadr_5 <= ethmac_rx_cdc_wrport_adr;
end

always @(posedge sys_clk) begin
	memadr_6 <= ethmac_rx_cdc_rdport_adr;
end

assign ethmac_rx_cdc_wrport_dat_r = storage_4[memadr_5];
assign ethmac_rx_cdc_rdport_dat_r = storage_4[memadr_6];
```

This is mapped to DRAM because it is too shallow/wide to be economically mapped to BRAM. 

The liteeth core fails to work with Yosys unless one of the following is carried out
 - The threshold is changed such BRAM is used instead
 - DRAM is disabled altogether and logic is used instead
 - memory_bram.cc is patched to not generate the transparent "bypass" logic even if a port is supposed to be transparent

Indeed this small case 
[outreg_transp.zip](https://github.com/YosysHQ/yosys/files/2922317/outreg_transp.zip) shows some synthesis/simulation mismatches without this patch.

master:
screenshot from 2019-03-02 16-58-07

patched:
screenshot from 2019-03-02 16-59-22

This patch creates a new bram type and techmap to deal with this specifically for the ECP5. I'm not sure if this should really be looked at in a more generic way, or whether I'm barking up the wrong tree altogether here. 

Incidentally, this patch gives significantly improved Fmax - from 95MHz to 108MHz.

### DIFF
--- a/techlibs/ecp5/dram.txt
+++ b/techlibs/ecp5/dram.txt
@@ -1,3 +1,20 @@
+bram $__TRELLIS_DPR16X4_REG
+  init 1
+  abits 4
+  dbits 4
+  groups 2
+  ports  1 1
+  wrmode 0 1
+  enable 1 1
+  transp 2 0
+  clocks 2 1
+  clkpol 3 2
+endbram
+
+match $__TRELLIS_DPR16X4_REG
+  min wports 1
+endmatch
+
 bram $__TRELLIS_DPR16X4
   init 1
   abits 4
@@ -12,6 +29,5 @@ bram $__TRELLIS_DPR16X4
 endbram
 
 match $__TRELLIS_DPR16X4
-  make_outreg
   min wports 1
 endmatch
--- a/techlibs/ecp5/drams_map.v
+++ b/techlibs/ecp5/drams_map.v
@@ -26,3 +26,61 @@ module \$__TRELLIS_DPR16X4 (CLK1, A1ADDR, A1DATA, B1ADDR, B1DATA, B1EN);
 		.WRE(B1EN)
 	);
 endmodule
+
+module \$__TRELLIS_DPR16X4_REG (CLK1, CLK2, A1ADDR, A1DATA, A1EN, B1ADDR, B1DATA, B1EN);
+	parameter [63:0] INIT = 64'bx;
+	parameter CLKPOL2 = 1;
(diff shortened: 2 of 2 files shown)


cliffordwolf:
The PR lgtm but..

> Indeed this small case outreg_transp.zip shows some synthesis/simulation mismatches without this patch.

without having looked at the test case, shouldn't it just be less efficient without this patch? Why do we get a syn/sim mismatch?

daveshah1:
The circuit after mapping is:
screenshot from 2019-03-03 11-01-16

Creating this trace:

screenshot from 2019-03-03 11-02-08


There seem to be a few problems here:
 - The new rather than old read address (in terms of the read clock domain) is used, meaning the first collision (read data should go from 9 to E) is not detected
 - The collision is detected for one write clock cycle too many, causing the spurious "8" to be read
 - Once the collision is over, the read value reverts to the previously read data in the output data rather than the new data, giving the "2"

cliffordwolf:
Ah, now I get it! There are two different clocks. Sorry, bit slow today. :man_facepalming:

So I think that's a bug in memory_bram. I think we should not infer that bypass logic unless all the memory ports are driven by the same clock. Does this sound right?

daveshah1:
Yes, that makes sense.

It seems like in the case of both make_outreg and make_transp, putting the register on the address should always be safe and can be done generically in memory_bram too.

cliffordwolf:
Yes. Do you think you can patch `memory_bram` accordingly? (I know, the pass is a mess.. ;)

daveshah1:
Superseded by #896

PR #859, opened 2019-03-07
iCE40 BRAM primitives init from file

smunaut:
This adds a INIT_FILE attribute to the SB_RAM40_4K blocks that will
initialize content from a hex file. Same behavior is imlemented in the
simulation model and in a new pass for actual synthesis

### DIFF
--- a/techlibs/ice40/Makefile.inc
+++ b/techlibs/ice40/Makefile.inc
@@ -1,5 +1,6 @@
 
 OBJS += techlibs/ice40/synth_ice40.o
+OBJS += techlibs/ice40/ice40_braminit.o
 OBJS += techlibs/ice40/ice40_ffssr.o
 OBJS += techlibs/ice40/ice40_ffinit.o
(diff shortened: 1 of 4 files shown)


cliffordwolf:
The code looks fine but I'm not sure I understand what this is for. Afaik the vendor tools do not support this cell parameter, so it's not for compatibility, and I don't see in what situation a user ever would want to use this instead of simply declaring an array, initializing that with `$memread[hb]`, and letting the tool infer the BRAM from that.

smunaut:
There are several case where I used instanciation rather than inferrence :

(1) Inference makes it hard to lock down the actual RAM to a precise BEL

(2) I couldn't find a way to infer a ram with different Write/Read clock and that has different read/write width _without_ any additional logic (i.e. using the READ/WRITE MODE appropriately) because I can't afford any added LUT/MUX in that path.

cliffordwolf:
Fair enough.

PR #869, opened 2019-03-12
Install launcher executable when running yosys-smtbmc on Windows.

cr1901:
This is the complement PR to the [one](https://github.com/YosysHQ/SymbiYosys/pull/32) I made for SymbiYosys. Now `yosys-smtbmc` can be used from either a msys2 `bash` prompt or from `cmd.exe` or `powershell.exe` :).

In principle, I can also convert `yosys-config`, but seeing as you already need a `bash` prompt to _compile_ `yosys` natively on Windows (without using the MSVC fork), I don't see much point in making compiling plugins usable from `cmd.exe`/`powershell.exe`.

### DIFF
--- a/backends/smt2/Makefile.inc
+++ b/backends/smt2/Makefile.inc
@@ -3,14 +3,30 @@ OBJS += backends/smt2/smt2.o
 
 ifneq ($(CONFIG),mxe)
 ifneq ($(CONFIG),emcc)
+
+# MSYS targets support yosys-smtbmc, but require a launcher script
+ifeq ($(CONFIG),$(filter $(CONFIG),msys2 msys2-64))
+TARGETS += yosys-smtbmc.exe yosys-smtbmc-script.py
+# Needed to find the Python interpreter for yosys-smtbmc scripts.
+# Override if necessary, it is only used for msys2 targets.
+PYTHON := $(shell cygpath -w -m $(PREFIX)/bin/python3)
+
+yosys-smtbmc-script.py: backends/smt2/smtbmc.py
+	$(P) sed -e 's|##yosys-sys-path##|sys.path += [os.path.dirname(os.path.realpath(__file__)) + p for p in ["/share/python3", "/../share/yosys/python3"]]|;' \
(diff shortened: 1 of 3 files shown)


backends/smt2/Makefile.inc
```diff
@@ -3,14 +3,30 @@ OBJS += backends/smt2/smt2.o
 
 ifneq ($(CONFIG),mxe)
 ifneq ($(CONFIG),emcc)
+
+# MSYS targets support yosys-smtbmc, but require a launcher script
+ifeq ($(CONFIG),$(filter $(CONFIG),msys2 msys2-64))
```
cliffordwolf: There is no `msys2-64` config in the Yosys build system.

misc/launcher.c
```diff
@@ -0,0 +1,335 @@
+/*  Setuptools Script Launcher for Windows
```
cliffordwolf: Origin? License?

backends/smt2/Makefile.inc
mmicko: Actually there is ( https://github.com/YosysHQ/yosys/blob/master/Makefile#L245 ) he added it time ago

mmicko: Actually there is ( https://github.com/YosysHQ/yosys/blob/master/Makefile#L245 ) he added it time ago

misc/launcher.c
mmicko: Seams this is part of python setuptools https://svn.python.org/projects/sandbox/branches/setuptools-0.6/launcher.c

cr1901: @mmicko I made an equivalent commit to symbiyosys and needed to add a license.

Guess who `wget` the file verbatim into yosys and forgot to add the license/source :)

cr1901: @cliffordwolf Added the license and origin. Whoops :D!

backends/smt2/Makefile.inc
cliffordwolf: weird... oh! my Yosys win32 working copy is on an old branch. Sorry, my fault.

PR #884, opened 2019-03-19
fix local name resolution in prefix constructs

zachjs:
This addresses an issue where Yosys would fail to resolve items declared within a generate block loop's scope that were used within an `AST_PREFIX` node in that scope.

I have very limited experience with Yosys and HDLs in general, so feedback is greatly appreciated!


eddiehung:
Thanks @zachjs! Just want to give you an extra thumbs up for a PR with a test. We all need to be more like you!

PR #889, opened 2019-03-21
Fix mem2reg handling of memories with upto data ports

### LINKED ISSUE #888: LSB first registers (e.g. [0:N-1]) leads to incorrect logic

```
module top (CLK, CE, SEL, SI, DO);
parameter SELWIDTH = 1;
parameter DATAWIDTH = 2;
input    CLK, CE;
input [DATAWIDTH-1:0] SI;
input [SELWIDTH-1:0] SEL;
output [DATAWIDTH-1:0] DO;
localparam DATADEPTH = 2**SELWIDTH;
reg [0:DATADEPTH-1] data1 [DATAWIDTH-1:0];
reg [DATADEPTH-1:0] data2 [DATAWIDTH-1:0];

generate
    genvar i;
    for (i = 0; i < DATAWIDTH; i=i+1) begin
        always @(posedge CLK)
        begin
            if (CE == 1'b1) begin
`ifdef BROKEN
                data1[i] <= {SI[i], data1[i][0:DATADEPTH-2]};
`else
                data2[i] <= {data2[i][DATADEPTH-2:0], SI[i]};
`endif
            end
        end
`ifdef BROKEN
        assign DO[i] = data1[i][SEL];
`else
        assign DO[i] = data2[i][SEL];
`endif
    end
endgenerate
endmodule
```

`yosys -p "read_verilog -DBROKEN top.v; synth_xilinx"` gives:
```
   Number of cells:                  5
     FDRE                            3
     LUT3                            2
```

## Expected behavior

`yosys -p "read_verilog top.v; synth_xilinx"` gives:

```
   Number of cells:                  6
     FDRE                            4
     LUT3                            2
```

## Actual behavior

`yosys -p "read_verilog -DBROKEN top.v; synth_xilinx"` on 8c0740bc gives:
```
   Number of cells:                  5
     FDRE                            3
     LUT3                            2
```

Same command on e112d2fb gave:
```
   Number of cells:                  8
     FDRE                            4
     LUT3                            4
```

cliffordwolf:

### DIFF
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -138,9 +138,15 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
 				int mem_width, mem_size, addr_bits;
 				node->meminfo(mem_width, mem_size, addr_bits);
 
+				int data_range_left = node->children[0]->range_left;
+				int data_range_right = node->children[0]->range_right;
+
+				if (node->children[0]->range_swapped)
+					std::swap(data_range_left, data_range_right);
+
 				for (int i = 0; i < mem_size; i++) {
 					AstNode *reg = new AstNode(AST_WIRE, new AstNode(AST_RANGE,
-							mkconst_int(mem_width-1, true), mkconst_int(0, true)));
+							mkconst_int(data_range_left, true), mkconst_int(data_range_right, true)));
 					reg->str = stringf("%s[%d]", node->str.c_str(), i);
 					reg->is_reg = true;
 					reg->is_signed = node->is_signed;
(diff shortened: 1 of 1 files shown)


cliffordwolf:
> whereas I would expect it to be 2xFDRE and 1xLUT3?

I think that's a separate issue that comes from the fact that we use `(WIDTH-1)-sel` to accomplish the bit select on upto wires. (And `synth_xilinx` seems to map those one-way to an adder chain.)

image

An obvious way to optimize this for power-of-two widths would be to replace `const - var` with `~var` in `opt_expr` when `const` is the all-ones pattern. (This would probably be a reasonable optimization either way.)

Another way to fix this would be to bit-reflect the input and output of the `$shiftx` cell instead of using arithmetic to calculate the shift amount.

Whitebox support in ABC integration would probably also fix this. (You get the expected output with just `synth`, where we don't have special mapping rules for arithmetic operations.)

PR #895, opened 2019-03-23
RFC: Add a pmux-to-shiftx optimisation to proc_mux

eddiehung:
I'd like opinions on this `$pmux` to `$shiftx` optimisation inside `proc_mux`
Doing so transforms the following circuit:
```
module top (CLK, SEL, SI, DO);
parameter SELWIDTH = 2;
parameter DATAWIDTH = 1;
input    CLK;
input [DATAWIDTH-1:0] SI;
input [SELWIDTH-1:0] SEL;
output [DATAWIDTH-1:0] DO;
localparam DATADEPTH = 2**SELWIDTH-1; // Note minus 1 so isn't a complete case
reg [DATAWIDTH-1:0] data [DATADEPTH-1:0];

generate
    genvar i;
    for (i = 0; i < DATADEPTH; i=i+1) begin
        always @(posedge CLK)
        begin
            if (i == 0)
                data[i] <= SI[i];
            else
                data[i] <= data[i-1];
        end
    end
    assign DO = data[SEL];
endgenerate
endmodule
```

from creating this on master @ ccfa2fe:
image

to creating this:
image

Are there any side effects to this?
The advantage is that this makes analysis much easier downstream...

### DIFF
--- a/passes/proc/proc_mux.cc
+++ b/passes/proc/proc_mux.cc
@@ -340,6 +340,7 @@ RTLIL::SigSpec signal_to_mux_tree(RTLIL::Module *mod, SnippetSwCache &swcache, d
 		// evaluate in reverse order to give the first entry the top priority
 		RTLIL::SigSpec initial_val = result;
 		RTLIL::Cell *last_mux_cell = NULL;
+		bool shiftx = true;
 		for (size_t i = 0; i < sw->cases.size(); i++) {
 			int case_idx = sw->cases.size() - i - 1;
 			RTLIL::CaseRule *cs2 = sw->cases[case_idx];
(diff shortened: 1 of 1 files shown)


cliffordwolf:
Looks like it is breaking `make test` (see Travis CI):

```
Test: aes_kexp128 -> ok
Test: always02 -> ok
Test: always03 -> ok
Test: arraycells -> ok
Test: arrays01 -> ok
Test: constmuldivmod -> ERROR!
```

> The advantage is that this makes analysis much easier downstream...


I'm not sure if `proc_mux` is the right place for this optimization. Maybe a generic `opt_*` pass would be better. For example, consider a design that contains something like the following:

```
  wire [1:0] sel = ...;
  assign out = (sel == 0) ? ... : (sel == 1) ? ... : (sel == 2) ? ... : ...;
```

eddiehung:
Yeah, I noticed that breakage just after I pushed. Assuming I fix it, what's your verdict (taking into account my arguments below):

Not quite following your reasoning, but the reason I've done it in `proc_mux` is that I can combine pmux-to-shiftx analysis inside `signal_to_mux_tree` which is where the `$pmux` is created. Doing it outside means having to perform analysis on its upstream `$eq` cells...

For example, are you arguing that this is an optimisation that won't capture other cases? To that, I agree, and this would be the topic of further study.
Right now, this particular case is blocking shift register inference because it certain `assign out = foo[bar]` indexing is causing `$pmux`+`$eq`-s (which is hard to analyse) and not a `$shiftx` as in other cases.

PR #896, opened 2019-03-24
memory_bram: Fix multiclock make_transp

daveshah1:
See discussion in #846 

 - Will not make transparent logic if read & write clock domains differ
 - When both make_outreg and make_transp are in affect, register is put on the address port (and data left unregistered), and transparent bypass logic is not created.


daveshah1:
yosys-tests PR will be arriving shortly

Issue #894, opened 2019-03-23
Relaxing of newline results in strict ';' need


Fatsie:
In 04cd179 newline handling was relaxed but it results in more strict need of ';' usage.
I have a .lib file from TSMC where not each line ends in ; and this now fails in yosys, e.g.

    cell (XXX) {
      ...
      area : 87.5
      cell_footprint : "XXX"
      pin (A1) {
      ...
    }

Reverting 04cd179 makes the TSMC .lib work again.

trcwm:
TSMC is using a incorrect .lib syntax 😃 Will fix asap.

Fatsie:
BTW, culprit is likely Library Compiler of Synopsys itself. The std. cell lib also includes 'compiled' liberty files with .db extension. This means they went through the Synopsys Library Compiler without problem.

Fatsie:
I can confirm #897 works with the TSMC liberty file.

Issue #825, opened 2019-02-22
read_library bus pin support


mvphilip:
Hello, I have a bus pin that read_liberty seems not handle:

  type( IO_bus_3_to_0 ) {
     base_type : array ;
     data_type : bit ;
     bit_width : 4;
     bit_from : 3 ;
     bit_to : 0 ;
     downto : true ;
  }

		bus(I1)  {
			bus_type : IO_bus_3_to_0 ;
			direction : input;
			pin (I1[3:0]) {

The error is on the last line:
ERROR: Syntax error in liberty file on line 28.

## Steps to reproduce the issue

[sram.lib.gz](https://github.com/YosysHQ/yosys/files/2895623/sram.lib.gz)

yosys
read_liberty -lib -ignore_miss_func -ignore_miss_data_latch -ignore_miss_dir -setattr blackbox sram.lib

## Expected behavior

The original library is valid and is accepted by commercial tools.

## Actual behavior

An error is reported on line 28:

Thank you,

-MP

mguthaus:
Adding a comment to follow developments. I've run into this error with OpenRAM support.

mvphilip:
Easy work around is to simply expand the pin members within the bus statement. It will still be recognized as a bus. It will make liberty much larger as arcs are now duplicated however. For example:

	bus(I1)  {
		bus_type : IO_bus_3_to_0 ;
		direction : input;
		pin (I1[3]) { .. }
                pin(I1[2]) { .. }
		pin (I1[1]) { .. }
                pin(I1[0]) { .. }
      }

Thanks,

-mp

eddiehung:
Looks like this was fixed by #897 and did not get closed.
Confirmed on 7dece795 to give this output:
```
 Yosys 0.9+932 (git sha1 a7e0cca4, clang 9.0.0-2 -fPIC -Os)


-- Running command `read_liberty -lib -ignore_miss_func -ignore_miss_data_latch -ignore_miss_dir -setattr blackbox sram.lib; dump' --

1. Executing Liberty frontend.
Imported 1 cell types from liberty file.

autoidx 1

attribute \blackbox 1
module \SRAM

  wire width 4 input 2 \I1

  wire input 1 \CE1
end
```

PR #897, opened 2019-03-24
Liberty parser: Accept ranges [A:B], and ignore missing ';'.

trcwm:
* Updated the liberty parser to accept [A:B] ranges (AST has not been updated).
* Liberty parser now also accepts key : value pair lines that do not end in ';'.
* Added tests.

Prepares for fixing https://github.com/YosysHQ/yosys/issues/825.
Should fix reading of non-standard liberty files: https://github.com/YosysHQ/yosys/issues/894.

### DIFF
--- a/passes/techmap/libparse.cc
+++ b/passes/techmap/libparse.cc
@@ -202,11 +202,10 @@ LibertyAst *LibertyParser::parse()
 	{
 		tok = lexer(str);
 
-		if (tok == ';')
+        // allow both ';' and new lines to 
+        // terminate a statement.
+		if ((tok == ';') || (tok == 'n'))
 			break;
-
-		if (tok == 'n')
-			continue;
 
 		if (tok == ':' && ast->value.empty()) {
 			tok = lexer(ast->value);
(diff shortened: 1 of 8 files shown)


passes/techmap/libparse.cc
```diff
@@ -86,15 +87,17 @@ int LibertyParser::lexer(std::string &str)
...
 	int c;
 
+    // eat whitespace
 	do {
 		c = f.get();
 	} while (c == ' ' || c == '\t' || c == '\r');
 
-	if (('a' <= c && c <= 'z') || ('A' <= c && c <= 'Z') || ('0' <= c && c <= '9') || c == '_' || c == '-' || c == '+' || c == '.' || c == '[' || c == ']') {
+    // search for identifiers, numbers, plus or minus.
+	if (('a' <= c && c <= 'z') || ('A' <= c && c <= 'Z') || ('0' <= c && c <= '9') || c == '_' || c == '-' || c == '+' || c == '.') {
```
cliffordwolf: Looks like some of your lines are indented with blanks (and a tabstop of 4), but the Yosys code bases uses tabs. Can you correct the indenting in this PR? Thanks. (Otherwise LGTM.)

trcwm: Converted spaces to tabs.

PR #905, opened 2019-03-28
Feature/python bindings

christian-krieg:
Python bindings to interact with the Yosys C++ API Yosys from within Python. This feature enables to call Yosys data structures and functions from a Python script, and to call passes written in Python from the Yosys shell.


py_wrap_generator.py
```diff
@@ -0,0 +1,2101 @@
...
+						debug("\t\tFound constructor of class \"" + class_[0].name + "\" in namespace " + concat_namespace(namespaces),2)
+						class_[0].found_constrs.append(candidate)
+						continue
+		if candidate != None:
+			while i < j:
+				i = i + 1
+				line = source_text[i].replace("YOSYS_NAMESPACE_BEGIN", "                    namespace Yosys{").replace("YOSYS_NAMESPACE_END","                    }")
+				ugly_line = unpretty_string(line)
+				if len(namespaces) != 0:
+					namespaces[-1] = (namespaces[-1][0], namespaces[-1][1] + ugly_line.count("{
```
daveshah1: I think this should be USING_YOSYS_NAMESPACE in case the Yosys namespace is changed.

kernel/yosys.cc
```diff
@@ -57,6 +57,16 @@
 #  include <sys/sysctl.h>
 #endif
 
+#ifdef WITH_PYTHON
+#if PY_MAJOR_VERSION >= 3
+#   define INIT_MODULE PyInit_libyosys
+    extern "C" PyObject* INIT_MODULE();
+#else
+#   define INIT_MODULE initlibyosys
+	    extern "C" void INIT_MODULE();
```
daveshah1: (nitpick) indentation here looks inconsistent

passes/cmds/plugin.cc
```diff
@@ -36,14 +44,48 @@ void load_plugin(std::string filename, std::vector<std::string> aliases)
...
+		{
+			int last_slash = filename.find_last_of('/');
+			std::string path = filename.substr(0, last_slash);
+			filename = filename.substr(last_slash+1, filename.size());
+			filename = filename.substr(0,filename.size()-3);
+			PyRun_SimpleString(("sys.path.insert(0,\""+path+"\")").c_str()); 
+			PyObject *filename_p = PyUnicode_FromString(filename.c_str());
+			if(filename_p == NULL)
+			{
+				log_cmd_error("Issues converting `%s' to Python\n", filename.c_str());
```
daveshah1: Is it possible to get better error information through exceptions etc either here or in the PyImport_Import returning null case?

passes/cmds/plugin.cc
```diff
@@ -36,14 +44,48 @@ void load_plugin(std::string filename, std::vector<std::string> aliases)
...
+	#ifdef WITH_PYTHON
+	if (!loaded_plugins.count(filename) && !loaded_python_plugins.count(filename)) {
+	#else
 	if (!loaded_plugins.count(filename)) {
+	#endif
+
+		#ifdef WITH_PYTHON
+		if(boost::algorithm::ends_with(filename, ".py"))
+		{
+			int last_slash = filename.find_last_of('/');
```
daveshah1: Should this consider Windows where '\' might be used? I can see that the other plugin related code doesn't consider this, but it also seems to rely on dlopen so won't work on Windows in any case - whereas I guess the Python plugin interface should be cross-platform.

kernel/rtlil.cc
```diff
@@ -639,7 +665,18 @@ RTLIL::Module::~Module()
 		delete it->second;
 	for (auto it = processes.begin(); it != processes.end(); ++it)
 		delete it->second;
+#ifdef WITH_PYTHON
+	RTLIL::Module::get_all_modules()->erase(hashidx_);
+#endif
+}
+
+#ifdef WITH_PYTHON
+static std::map<unsigned int, RTLIL::Module*> *all_modules = new std::map<unsigned int, RTLIL::Module*>();
```
daveshah1: Is there any reason why this is a pointer allocated by calling new at startup, rather than a simple static object?

py_wrap_generator.py
```diff
@@ -0,0 +1,2101 @@
...
+			return None
+
+		parts = split_list(str_def.strip(), " ")
+
+		prefix = ""
+		i = 0
+		for part in parts:
+			if part in ["unsigned", "long", "short"]:
+				prefix += part + " "
+				i = i + 1
```
daveshah1: (nitpick) i += 1

py_wrap_generator.py
```diff
@@ -0,0 +1,2101 @@
...
+				text += "\n\t\t\t" + long_name + "* ret = " + long_name + "::get_all_" + self.name.lower() + "s()->at(this->" + self.id_.varname + ");"
+				text += "\n\t\t\tif(ret != NULL && ret == this->ref_obj)"
+				text += "\n\t\t\t\treturn ret;"
+				text += "\n\t\t\tthrow std::runtime_error(\"" + self.name + "'s c++ object does not exist anymore.\");"
+				text += "\n\t\t\treturn NULL;"
+				text += "\n\t\t}\n"
+
+			#if self.link_type != link_types.pointer:
+			text += "\n\t\tstatic " + self.name + "* get_py_obj(" + long_name + "* ref)\n\t\t{"
+			text += "\n\t\t\t" + self.name + "* ret = (" + self.name + "*)malloc(sizeof(" + self.name + "));"
```
daveshah1: Could new be used instead of malloc & cast?

py_wrap_generator.py
```diff
@@ -0,0 +1,2101 @@
...
+				if self.needs_clone:
+					text += "\n\t\t\tret->ref_obj = ref.clone();"
+				else:
+					text += "\n\t\t\tret->ref_obj = new "+long_name+"(ref);"
+				text += "\n\t\t\treturn ret;"
+				text += "\n\t\t}\n"
+
+			if self.link_type != link_types.global_list:
+				text += "\n\t\t~" + self.name + "()\n\t\t{"
+				text += "\n\t\t\t//delete(this->ref_obj);\n\t\t}\n"
```
daveshah1: This looks like it is generating commented out code?

daveshah1:
Overall this is very cool, and I like the approach it has taken! Just a few small comments left.

Could you update the installed packages in .travis.yml so this builds on CI, and also add the new dependencies to the README?

py_wrap_generator.py
btut: USING_YOSYS_NAMESPACE will not do, since I only need the name of the namespace. I introduced a new define in kernel/yosys.h that only contains the namespace name and used that one. It is near the other defines that need to be changed in case the namespace changes.
0774a500d4a03104787e6514c69ff4995633a65f

kernel/yosys.cc
btut: Fixed
072c9393805d59b988f93b184356c54731237a5a

passes/cmds/plugin.cc
btut: I found out that Boost.Python actually provides a function that prints the error to stderr and used that one bbfb43006d2a7d67d06ee151a1b8ad05ec5b1750.
It would be better to write it to Yosys's error log though, but I could not find how to get to the error message itself. The only way I could think of would be to temporarily redirect stderr to a buffer and then write that to the Yosys log, do you have a better idea perhaps?

passes/cmds/plugin.cc
btut: Solved using the experimental c++ std library filesystem. fd7fb1377d4d30d692c78eb55173198339fea17d

kernel/rtlil.cc
btut: No :) 827a96d3a38afe025c9efbd182069a9c9adee267

py_wrap_generator.py
btut: 827a96d3a38afe025c9efbd182069a9c9adee267

py_wrap_generator.py
btut: Actually no, since new would call the constructor and for classes that do not contain a default constructor this will not work.

py_wrap_generator.py
btut: It did!
827a96d3a38afe025c9efbd182069a9c9adee267

passes/cmds/plugin.cc
daveshah1: Does this help? https://stackoverflow.com/questions/24386818/how-can-i-get-output-of-pyerr-print-in-windows-or-save-it-as-a-string

btut: That's what I found, too. But this solution relies on a function throwing an actual exception. PyUnicode_FromString and PyImport_Import do not throw any exceptions, in case of an error NULL is returned and the user may call PyErr_Print(). All other solutions that I found suggest redirecting stdout and stderr.

daveshah1: Ah, I see. This seems fine then.

Makefile
```diff
@@ -261,6 +268,19 @@ ifeq ($(ENABLE_LIBYOSYS),1)
 TARGETS += libyosys.so
 endif
 
+ifeq ($(ENABLE_PYOSYS),1)
+	ifeq ($(PYTHON_MAJOR_VERSION),3)
+		LDLIBS += -lpython$(PYTHON_VERSION)m -lboost_python-py$(subst .,,$(PYTHON_VERSION)) -lboost_system -lboost_filesystem
+	else
+		LDLIBS += -lpython$(PYTHON_VERSION) -lboost_python-py$(subst .,,$(PYTHON_VERSION)) -lboost_system -lboost_filesystem
+	endif
+CXXFLAGS += -I/usr/include/python$(PYTHON_VERSION) -D WITH_PYTHON
```
daveshah1: Trying to build on Arch Linux and I see this fail because the path is `/usr/include/python3.7m`. This should probably be using `python-config --includes` rather than hardcoded paths.

Makefile
```diff
@@ -261,6 +268,19 @@ ifeq ($(ENABLE_LIBYOSYS),1)
 TARGETS += libyosys.so
 endif
 
+ifeq ($(ENABLE_PYOSYS),1)
+	ifeq ($(PYTHON_MAJOR_VERSION),3)
+		LDLIBS += -lpython$(PYTHON_VERSION)m -lboost_python-py$(subst .,,$(PYTHON_VERSION)) -lboost_system -lboost_filesystem
```
daveshah1: Arch Linux calls the Boost Python library boost_python37 rather than boost_python-py37. There is significant inter-distro variation here, see all the different cases that we ended up having to detect in nextpnr: https://github.com/YosysHQ/nextpnr/blob/master/CMakeLists.txt#L132-L186 - I'm not sure if there is a better way than bruteforcing all the possibilities here.

Makefile
```diff
@@ -16,9 +16,16 @@ ENABLE_READLINE := 1
...
-ENABLE_LIBYOSYS := 0
+ENABLE_LIBYOSYS := 1
 ENABLE_PROTOBUF := 0
 
+# python wrappers
+ENABLE_PYOSYS := 1
+PYTHON_VERSION_TESTCODE := "import sys;t='{v[0]}.{v[1]}'.format(v=list(sys.version_info[:2]));print(t)"
+PYTHON_VERSION := $(shell if python3 -c ""; then python3 -c ""$(PYTHON_VERSION_TESTCODE)""; else python -c ""$(PYTHON_VERSION_TESTCODE)""; fi)
+PYTHON_MAJOR_VERSION := $(shell echo $(PYTHON_VERSION) | cut -f1 -d.)
+PYTHON_DESTDIR := /usr/local/lib/python$(PYTHON_VERSION)/dist-packages
```
daveshah1: `python-config --prefix` rather than hardcoded `/usr/local`

Makefile
btut: I was unaware of python-config. Looks pretty userful!
cc270ea81bc7209e9e1673e6c3dbd5db6d3ceee8

Makefile
btut: Thanks for the heads-up. I implemented 4 versions now:

1. boost_python-py<version>
2. boost_python-py<major_version>
3. boost_python<version>
4. boost_python<major_version>

If none of those work, the user is prompted to define BOOST_PYTHON_LIB manually.
cc270ea81bc7209e9e1673e6c3dbd5db6d3ceee8

Makefile
btut: Done
cc270ea81bc7209e9e1673e6c3dbd5db6d3ceee8

Makefile
daveshah1: Thanks, this works well for me now. One small comment is the auto-detection prints
```
/usr/bin/ld: cannot find -lboost_python-py37
collect2: error: ld returned 1 exit status
/usr/bin/ld: cannot find -lboost_python-py3
collect2: error: ld returned 1 exit status
/usr/bin/ld: cannot find -lboost_python-py37
collect2: error: ld returned 1 exit status
/usr/bin/ld: cannot find -lboost_python-py3
collect2: error: ld returned 1 exit status
```
at the moment, which could be confusing as it looks like an error. Perhaps you should redirect the output to /dev/null or similar?

btut: Suppressed error output in e19981ab6111765baa5b2ab7a16c92278130fd8b

daveshah1:
Thanks for the reminder @cliffordwolf !

Generally very happy with this now. My last remaining issue is that as far as I can tell, this is enabled by default for all Yosys targets. It would be good to confirm that it works for all of them (e.g. mxe, msys2, Emscripten, Visual Studio) and if not, either fix it or disable it by default if it isn't practicable to get it to work (I suspect this may be the case for Emscripten at least).

cliffordwolf:
I've now merged it, but also set `ENABLE_PYOSYS=0` as default value.

cliffordwolf:
I'd like to keep the root dir as clean as possible. Would it be possible to move `__init__.py` and `py_wrap_generator.py` to `misc/`, maybe make `__init__.py` a make target and only place it in the top-level dir if we are actually performing a build with python enabled?

btut:
> I'd like to keep the root dir as clean as possible. Would it be possible to move `__init__.py` and `py_wrap_generator.py` to `misc/`, maybe make `__init__.py` a make target and only place it in the top-level dir if we are actually performing a build with python enabled?

This is a simple fix. No need for a build-target for __init__.py, as it is not needed when importing the .so file from the yosys directory, just when it is installed, so copying it on install (from the misc dir) is enough.
I am not a github pro, is there any way to update this pull-request with a new commit?

> My last remaining issue is that as far as I can tell, this is enabled by default for all Yosys targets. It would be good to confirm that it works for all of them

I need to setup a lot of virtual machines to test this out, unfortunately this will take some time.

cliffordwolf:
> I am not a github pro, is there any way to update this pull-request with a new commit?

No, but you can just push to your existing branch and create another PR. But usually one would create a new feature branch, forked from current master, and create a new PR. (This sounds like a hassle but it's really just a few seconds once you are used to git.)

> I need to setup a lot of virtual machines to test this out, unfortunately this will take some time.

I did turn it off by default for now. So no need to get wild with testing on all platforms yet.

btut:
> But usually one would create a new feature branch, forked from current master, and create a new PR.

I stuck to the same feature-branch for now since I already pushed to it before you commented, but will keep your suggestion in mind for next time. Thanks!

PR #909, opened 2019-04-01
support repeat loops with constant repeat counts outside of constant functions

zachjs:
I changed the `sat/counters` test to use repeat instead of for, and it seems to work. I noticed that the test ran much slower (about ~0.2s vs. ~1.3s), so I imagine I'm missing something important here.

Feedback is greatly appreciated!


cliffordwolf:
> I noticed that the test ran much slower (about ~0.2s vs. ~1.3s), so I imagine I'm missing something important here.

That's fine. The issue is that in the for-loop case the values of `i` are resolved by the front-end when unrolling the loop, whereas in the repeat case the front-end doesn't know what the values of `i` are for the individual loop iterations. Therefore the front-end generates a much more complicated circuit and is takes a second or so to optimize to what the for-loop case generated right away.

The diff looks good to me. Merging it now.

PR #911, opened 2019-04-02
Make nobram false by default for gowin

mmicko:
Make nobram to false, to make it same as for other architectures.
Right not it is not possible to make yosys actually use bram due to this.
Also makes testing of parts of code for bram instantiation possible.


dh73:
I'm sorry @mmicko nobram was disabled in GoWin due lack of techmap model. I have it incomplete (only DRAM). I will push that change also today so you can continue.

PR #912, opened 2019-04-02
memory_bram: Consider read enable for address expansion register

daveshah1:
The register for the address signals used in the expansion mux wasn't considering the read enable previously.

Thanks @smunaut for finding


whitequark:
This has been probably *the* worst bug I have ever encountered.

PR #914, opened 2019-04-03
synth_xilinx to now infer SRL16E/SRLC32E

eddiehung:
* Added `shregmap -tech xilinx` which will infer all variable-length shift registers into `$__XILINX_SHREG_` internal primitives
* `synth_xilinx` will now call `shregmap -tech xilinx -minlen 3` (before the first `techmap`) by default unless `-nosrl` is specified. This first call infers variable-length shift registers. 
* `synth_xilinx` will now call `shregmap -minlen 3 -init -params -enpol any_or_none` (after `abc`) by default, unless `-nosrl` specified. This second call infers fixed-length shift registers after ABC has done its bit, in case retiming was performed -- otherwise, inferring fixed-length shift registers before ABC will cause them to not be available for retiming.
* These `$__XILINX_SHREG_` and `$__SHREG_` (from second call) internal primitives can then be technology mapped into concrete SRL16E/SRLC32E blocks
* Many (exhaustive) tests are currently queued up inside [yosys-tests](https://github.com/YosysHQ/yosys-tests/compare/xc7srl)

~WIP until ~#916~ ~#917~ ~#927~ #944 are merged.~


techlibs/xilinx/synth_xilinx.cc
```diff
@@ -64,10 +64,13 @@ struct SynthXilinxPass : public Pass
...
 		log("\n");
 		log("    -nobram\n");
-		log("        disable infering of block rams\n");
+		log("        disable inference of block rams\n");
 		log("\n");
 		log("    -nodram\n");
-		log("        disable infering of distributed rams\n");
+		log("        disable inference of distributed rams\n");
+		log("\n");
+		log("    -nobram\n");
```
daveshah1: -nosrl not -nobram

techlibs/xilinx/cells_map.v
```diff
@@ -17,4 +17,131 @@
...
+        MUXF7 fpga_mux_1 (.O(T8), .I0(T4), .I1(T6), .S(L[5]));
+        MUXF8 fpga_mux_2 (.O(Q), .I0(T7), .I1(T8), .S(L[6]));
+      end
+    end
+    else if (DEPTH <= 129 && ~&_TECHMAP_CONSTMSK_L_) begin
+      // Handle cases where depth is just 1 over a convenient value,
+      // For variable length, bump up to the next length
+      \$__XILINX_SHREG_ #(.DEPTH(DEPTH+1), .INIT({INIT,1'b0}), .CLKPOL(CLKPOL), .ENPOL(ENPOL)) _TECHMAP_REPLACE_ (.C(C), .D(D), .L(L), .E(E), .Q(Q));
+    end
+    else /*if (DEPTH > 128)*/ begin
```
daveshah1: As I read this, this comment isn't correct as this also handles the case of one over a convenient value and L _is_ all constant (but it looks like it will handle this case correct, so just the comment needs fixing)

techlibs/xilinx/synth_xilinx.cc
```diff
@@ -269,16 +281,36 @@ struct SynthXilinxPass : public Pass
...
 		if (check_label(active, run_from, run_to, "map_cells"))
 		{
+			if (!nosrl) {
+				// shregmap operates on bit-level flops, not word-level,
+				//   so break those down here
+				Pass::call(design, "simplemap t:$dff t:$dffe");
+				// shregmap -tech xilinx can cope with $shiftx and $mux
+				//   cells for identifiying variable-length shift registers,
+				//   so attempt to convert $pmux-es to the former
+				Pass::call(design, "pmux2shiftx");
```
cliffordwolf: That's very late for running `pmux2shiftx`. Look at all the fine-grain stuff that happened already at this point. Are there even any $eq cells left at this stage? This should be run at about the same time as `wreduce`, or `alumacc`, during the coarse-grain phase. Any word-level structures that are still preserved at this point are just preserved accidentally, not because there'd be any guarantee.

techlibs/xilinx/synth_xilinx.cc
```diff
@@ -251,16 +264,15 @@ struct SynthXilinxPass : public Pass
 
 		if (check_label(active, run_from, run_to, "fine"))
 		{
-			Pass::call(design, "opt -fast -full");
+			Pass::call(design, "opt -fast");
 			Pass::call(design, "memory_map");
 			Pass::call(design, "dffsr2dff");
 			Pass::call(design, "dff2dffe");
-			Pass::call(design, "opt -full");
```
cliffordwolf: I guess `-full` is removed here in an attempt to make `pmux2shiftx` below work. But the correct solution would be to move `pmux2shiftx` up to the coarse-grain phase where it belongs.

techlibs/xilinx/synth_xilinx.cc
eddiehung: I've made a number of (big) changes to the flow such that coarse grained/word-level should be preserved right up to `abc` being called. I'm hesitant to move it up to the coarse-grained level since (currently) `shregmap` can only analyse bit-level DFFs, but I'm open to discussing the pros and cons of each.

techlibs/xilinx/synth_xilinx.cc
eddiehung: Yes, and the moving of `techmap -map +/techmap.v` down to just before `abc`.

techlibs/xilinx/synth_xilinx.cc
cliffordwolf: Fair enough. Approving it. You will be the one who has to deal with the fallout if this isn't working as expected anyway.   ;)

PR #915, opened 2019-04-04
Use 'bvashr' SMT-LIB function instead of bvLshr.

Edgeworth:
According to http://smtlib.cs.uiowa.edu/logics-all.shtml I think this is meant to be bvashr. Without this change, the following assertion fails:

wire signed [3:0] f_tmp = 4'b1111;
assert (($signed(f_tmp) >>> 1) == 4'b1111);

I want to add a regression test for this but it would be great to get a pointer to where the tests are for smt2 related code.


cliffordwolf:
```assert (($signed(f_tmp) >>> 1) == 4'b1111);```

According to Verilog rules this is not a signed right shift. The following code would be:

```assert (($signed(f_tmp) >>> 1) == 4'sb1111);```

And with the latter version the generated SMT2 code also contains an arithmetic right shift.

You patch would introduce a bug as it would use the arithmetic right shift unconditionally, even in cases where the verilog rules tell us to interpret `>>>` as a logic right shift.

Edgeworth:
Thank you for the explanation and sorry for the noise.

PR #916, opened 2019-04-04
synth_xilinx to map_cells before map_luts

eddiehung:
Since `cells_map.v` may require some soft-logic, it makes sense to technology map them before `abc` is called in map_luts


eddiehung:
Held pending #917

PR #917, opened 2019-04-05
Retime by default when abc -dff

eddiehung:
* Calling `abc -dff` now implies the addition of `-D 1` (delay target of 1ps) if `-D` argument is not specified -- without this, ABC's `retime` command will not get called
This is particularly useful for `synth_* -retime` calls in which you can't specify `-D` easily
* `synth_xilinx` to now perform `techmap +/xilinx/ff_map.v` after `abc -dff` call

EDIT:
* Also fixed a nasty S0 -> S1 typo when storing initial flop states!!

On this circuit:
```
module top(input clk, input [7:0] a, output z);
reg [3:0] ff = 4'b0;
always @(posedge clk)
    ff <= {ff[2:0], ^a};
assign z = ff[3];
endmodule
```

Without these two fixes this is what we get: [before.pdf](https://github.com/YosysHQ/yosys/files/3049763/before.pdf)
After: [after.pdf](https://github.com/YosysHQ/yosys/files/3049764/after.pdf)

### DIFF
--- a/passes/techmap/abc.cc
+++ b/passes/techmap/abc.cc
@@ -29,17 +29,17 @@
 // Kahn, Arthur B. (1962), "Topological sorting of large networks", Communications of the ACM 5 (11): 558-562, doi:10.1145/368996.369025
 // http://en.wikipedia.org/wiki/Topological_sorting
 
-#define ABC_COMMAND_LIB "strash; ifraig; scorr; dc2; dretime; strash; &get -n; &dch -f; &nf {D}; &put"
-#define ABC_COMMAND_CTR "strash; ifraig; scorr; dc2; dretime; strash; &get -n; &dch -f; &nf {D}; &put; buffer; upsize {D}; dnsize {D}; stime -p"
-#define ABC_COMMAND_LUT "strash; ifraig; scorr; dc2; dretime; strash; dch -f; if; mfs2"
-#define ABC_COMMAND_SOP "strash; ifraig; scorr; dc2; dretime; strash; dch -f; cover {I} {P}"
-#define ABC_COMMAND_DFL "strash; ifraig; scorr; dc2; dretime; strash; &get -n; &dch -f; &nf {D}; &put"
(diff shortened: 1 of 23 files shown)


daveshah1:
ff_map.v contains inversion:

https://github.com/YosysHQ/yosys/blob/master/techlibs/xilinx/ff_map.v#L31

Therefore I don't think mapping after `abc -lut` will work. Instead I think a three-stage approach (`abc -dff`, then FF techmapping, then `abc -lut`) like other synth_ commands use is the best way forward.

eddiehung:
> ff_map.v contains inversion:
> 
> https://github.com/YosysHQ/yosys/blob/master/techlibs/xilinx/ff_map.v#L31
> 
> Therefore I don't think mapping after `abc -lut` will work. Instead I think a three-stage approach (`abc -dff`, then FF techmapping, then `abc -lut`) like other synth_ commands use is the best way forward.

Disagree with calling ABC twice. Instead, I have (at the risk of making things less tidy) added a new `techmap` call that transforms just those dffs that need inversion before `abc`.

passes/techmap/abc.cc
```diff
@@ -1674,6 +1674,8 @@ struct AbcPass : public Pass {
 			}
 			if (arg == "-dff") {
 				dff_mode = true;
+				if (delay_target.empty())
+					delay_target = "-D 1";
```
cliffordwolf: Why not just call `abc` with `-D 1` in `synth_xilinx`? Forcing this default value seems to remove flexibility. And `1 ps` really doesn't seem like a good default value, especially for things like `upsize` and `dnsize`.

eddiehung: The intention was to make it work across the board (for all `synth_* -retime`) and also for `abc -dff` when a timing target was not specified.
I assume ASIC users would normally be expected to provide a delay target, no?

eddiehung:
Consider running `retime` even if no `-D` specified?

eddiehung:
This now means that calling `abc -dff` will always retime the circuit...

PR #918, opened 2019-04-06
Add command line flag -shiftx to proc_mux

aman-goel:
Add command line flag for pmux-to-shiftx optimisation to proc_mux. 

Creation of $shiftx cells isn't always preferred when dealing with case statements. Its best to control their generation using a flag.


eddiehung:
Hi @aman-goel, thanks for the PR! The intention of #895 (which added the `$pmux`-to-`$shiftx` is to enable further optimisations downstream -- with the reasoning being that analysing a `$shiftx` is much easier than analysing a `$pmux` fed by `$cmp` cells. Can you describe the regression you are seeing so we can decide the best way forward?

aman-goel:
Thanks @eddiehung for your comment. 

For going from a given Verilog RTL to a simple flat RTL (or Boolean netlist), you might me correct in that the analysis required in this procedure may be simpler with $shiftx cells. However for performing tasks like model checking on a simple RTL, it is much better to avoid cells like $shiftx (unless originally present in the original Verilog RTL) since SMT solver reasoning involving shift with a variable amount is much much more complex and expensive compared to a simple nested mux tree. Moreover interesting techniques of formal verification like data abstraction may get affected when introducing data operations like $shiftx for control logic like $pmux.

I made the above observation when performing regression tests for my formal verification tool [AVR](https://github.com/aman-goel/avr). AVR performs model checking on Verilog RTL using SMT solvers, and uses Yosys as the frontend to go from Verilog RTL to simple flat RTL. 

My request with this PR is to simply control the $pmux-to-$shiftx optimization based on user flag, instead of forcing it unconditionally for all applications.

Please let me know if any further detail is required from my side. Thanks.

--
Aman

eddiehung:
Great post @aman-goel, thanks for taking the time to educate me. You make an excellent point that I never considered, and thus I would like to take a step back to consider undoing the pmux-to-shiftx optimisation (#895) entirely and taking a different approach for downstream optimisations... let me try a few things and get back to you.

eddiehung:
Allow me to propose #927 as the solution, thus undoing #895 as it is likely a little too aggressive and could have wider-ranging effects. Would this work for you?

eddiehung:
Original cause of problem in #895 now reverted by #939

Issue #878, opened 2019-03-16
Don't care optimisation for $shiftx


eddiehung:
## Steps to reproduce the issue

`yosys -p "read_verilog -icells shiftx.v; synth_xilinx; stat"`

```
module top(input A, input [1:0] B, output [5:0] Y);
// Containing $shiftx instantiations
no_opt no_opt(A, B, Y[0]);
no_opt2 no_opt2(A, B, Y[1]);
opt opt(A, B, Y[2]);

// Containing $shiftx inferred from
// assign Y = A[B];
no_opt3 no_opt3(C3, B, Y[3]);
opt3 opt3(C4, B, Y[4]);
endmodule

module no_opt(input A, input [1:0] B, output Y);
$shiftx #(.A_WIDTH(3), .B_WIDTH(2), .Y_WIDTH(1), .A_SIGNED(0), .B_SIGNED(0))
    s (.A({A,A,A}), .B(B), .Y(Y));
endmodule

module no_opt2(input A, input [1:0] B, output Y);
$shiftx #(.A_WIDTH(4), .B_WIDTH(2), .Y_WIDTH(1), .A_SIGNED(0), .B_SIGNED(0))
    s (.A({1'bx,A,A,A}), .B(B), .Y(Y));
endmodule

module no_opt3(input A, input [1:0] B, output Y);
wire [2:0] A3 = {A,A,A};
assign Y = A3[B];
endmodule

module opt(input A, input [1:0] B, output Y);
$shiftx #(.A_WIDTH(4), .B_WIDTH(2), .Y_WIDTH(1), .A_SIGNED(0), .B_SIGNED(0))
    s (.A({A,A,A,A}), .B(B), .Y(Y));
endmodule

module opt3(input A, input [1:0] B, output Y);
wire [3:0] A4 = {A,A,A,A};
assign Y = A4[B];
endmodule
```

## Expected behavior

`no_opt*` circuits should be optimised away since all its inputs are the same.
Difference between those and `opt*` is that the latter have A_WIDTH == 2**B_WIDTH exactly.

## Actual behavior

`no_opt*` circuits are not optimised away.

eddiehung:
Discussion with @cliffordwolf indicates that the problem could be that 1'bx in mux cells are not passed to ABC as 1'bx, but instead set to 1'b0...

eddiehung:
[...]

PR #927, opened 2019-04-08
Undo #895 by instead setting an attribute

eddiehung:


cliffordwolf:
I would actually prefer #918, or completely undoing #895. I don't think the `shiftx_compatible` attribute would be very useful, because a later pass using it would need to re-check if the pmux is still "shiftx compatible" anyways as other passes may have messed with the control logic in the mean time. So there is little advantage compared to a generic opt pass that just detects and optimizes mux/pmux+eq circuits.

I could write such an opt pass if that's what's needed to resolve this.

eddiehung:
Thanks for the review. You make a good point. I was also having my doubts and started working on a new `pmux2shiftx` command [here](https://github.com/YosysHQ/yosys/compare/pmux2shiftx) to perform that transformation separately. This transformation is only necessary just before we starting doing some downstream inference (e.g. SRL or MUX) as it should have net zero effect on synthesis quality. 
Currently, this `pmux2shiftx` takes the one-hot `S` input of the original `$pmux`, and then uses this to creates a new decoder (using another `$pmux`) to generate a binary value which is then used for the `B` input of a new `$shiftx` that replaces the original `$pmux`. The problem with this approach (since the one-hot encoder and the new decoder should all get optimised away by `abc`) is that it doesn't work very well with non-pow-2 inputs, and in particular, we can't take advantage of don't cares (similar to #878?). Thoughts on this also welcome.

I don't believe it should go into `opt` because `opt` gets called many times -- and it would be painful to have to specify any extra argument every single time you call it.

If you've already got ideas about how to optimise mux/pmux+eq circuits, would love to hear them as I'm still wrapping my head around how to do this efficiently. Of course, if it's easier to go ahead and build it, then please do so, but I know you're a busy man.

eddiehung:
Superceded by #939

PR #929, opened 2019-04-10
Use include directories also for top-level read_verilog in a script.

olegendo:
The first top-level read_verilog in a script does not use include
directories which could have been specified before via verilog_defaults.

This patch allows the top-level verilog file to be located outside the
current working directory.  This was needed for me to integrate yosys
into a cmake + ninja build system with properly working dependencies and
rebuilds when using yosys script files.

### DIFF
--- a/frontends/verilog/verilog_frontend.cc
+++ b/frontends/verilog/verilog_frontend.cc
@@ -393,7 +393,7 @@ struct VerilogFrontend : public Frontend {
 			}
 			break;
 		}
-		extra_args(f, filename, args, argidx);
+		extra_args(f, filename, args, argidx, include_dirs);
 
 		log("Parsing %s%s input from `%s' to AST representation.\n",
 				formal_mode ? "formal " : "", sv_mode ? "SystemVerilog" : "Verilog", filename.c_str());
--- a/kernel/driver.cc
+++ b/kernel/driver.cc
@@ -181,6 +181,7 @@ int main(int argc, char **argv)
 	std::string backend_command = "auto";
 	std::vector<std::string> vlog_defines;
 	std::vector<std::string> passes_commands;
+	std::vector<std::string> passes_commands_before;
 	std::vector<std::string> plugin_filenames;
 	std::string output_filename = "";
 	std::string scriptfile = "";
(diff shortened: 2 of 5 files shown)


eddiehung:
Hi @olegendo. I'm not quite following what the issue that this PR fixes-- can you provide a testcase to help me understand? Thanks!

olegendo:
Hi Eddie,

[yosys_incdir_test.zip](https://github.com/YosysHQ/yosys/files/3069048/yosys_incdir_test.zip)

This is a small test case which demonstrates that 
    `verilog_defaults -add -I${src_dir}`

is ignored and not effective for the first occurrence of 'read_verilog' in the script file.  This is because the top-level 'read_verilog' is handled in a different way than the 'read_verilog' once the frontend is actually processing a verilog file.  It also doesn't matter whether the verilog_defaults is called via the '-p' argument or whether it is written in the script file itself -- the additional include directories are never used at that point.

To run this example, just run the build.sh script in the same directory as the script.  With the unmodified yosys it will throw

> ERROR: Can't open input file `top.v' for reading: No such file or directory


The point of this simplified example is realizing out-of-tree builds, i.e. having the build output data (synthesized results) stored in some arbitrary location unrelated to the source tree.  I've made it as a subdirectory of the source just to simplify the case.  out-of-tree builds are a common thing to do in software builds and are much easier to deal with than in-tree builds.

The way yosys handles the script file and the first top-level verilog file makes it difficult to implement out-of-tree builds with some software build systems, because yosys effectively demands that the script file and the top-level verilog file are in the same directory and that yosys must be invoked within that directoy.  This is what the first commit in this PR addresses.

The second commit deals with the fact that it's currently impossible to specify any prefix commands before the script file is processed, if the script file is specified as a regular input file and not via the '-p' command.

I hope this explains my intentions a little bit better than the comments in the commits.

eddiehung:
Thanks for explaining @olegendo! I understand what you're trying to do, but I'm not sure I agree with this approach. 
`-I` describes include directories which are for `include` constructs, and not for top level files, and I'm not sure other EDA tools out there support this approach either.

There are two workarounds I can think of right now:
- Mirror the source tree inside your build directory
- Use Yosys' Tcl front end which supports variables (here's a recent example of its usage: https://github.com/YosysHQ/nextpnr/blob/659c9325591845806143865563081a6644a3b900/generic/synth/synth_generic.tcl)

cliffordwolf:
What @eddiehung said.

olegendo:
Do other EDA tools also support this kind of script interface with that same semantics and Tcl front end?

eddiehung:
@olegendo Not sure if this is exactly what you're asking, but Tcl (rightly or wrongly) is a very prevalent language in EDA that's often used as the scripting interface across Cadence/Synopsys/Mentor products as well as FPGA vendors'.

olegendo:
@eddiehung Yeah, kind of what I was asking :)  Thanks!

PR #941, opened 2019-04-18
ice40 cells_sim.v: update clock enable behaviour based on hardware experiments

Wren6991:
This is as a result of @whitequark and I abusing some iCE40s for a while. [Useful part of thread starts here](https://twitter.com/whitequark/status/1117916172442316800). Here is a testbench containing the 5th and 6th test from that thread:

```
module tb;

reg e, clk;
wire q;

SB_IO #(
	.PIN_TYPE (6'b01_00_00)
	//            |  |
	//            |  \-------- DDR output
	//            \----------- Permanent output enable
) inst_SB_IO (
	.CLOCK_ENABLE      (e),
	.OUTPUT_CLK        (clk),
	.D_OUT_0           (1'b0),
	.D_OUT_1           (1'b1),
	.PACKAGE_PIN       (q)
);

localparam DELAY = 1000;

initial begin: stimulus
	integer i;
	{e, clk} = 2'b00;
	for (i = 0; i < 5; i = i + 1) begin
		#DELAY; {e, clk} = 2'b00;
		#DELAY; {e, clk} = 2'b01;
		#DELAY; {e, clk} = 2'b11;
		#DELAY; {e, clk} = 2'b10;
	end
	#DELAY;
	{e, clk} = 2'b00;
	#(4 * DELAY);
	for (i = 0; i < 5; i = i + 1) begin
		#DELAY; {e, clk} = 2'b00;
		#DELAY; {e, clk} = 2'b10;
		#DELAY; {e, clk} = 2'b11;
		#DELAY; {e, clk} = 2'b01;
	end
	$finish;
end

endmodule
```

The behaviour of the actual hardware is to generate no transitions (constant 0) for the first of these, and a full DDR set of transitions for the second. Using the patched simlib from this branch:

image

Using the current simlib:
image

(the hardware seems like it might clear the DOUT0/DOUT1 flops during power-on-reset, although I'm not sure how to positively check this)

The patched simlib contains our best guess as to the actual circuit, which is (ignoring effects of `NEG_TRIGGER`):
- DOUT0 is registered with a posedge DFFE, enabled by CLOCK_ENABLE
- CLOCK_ENABLE is also reregistered with a posedge DFF
- DOUT1 is registered with a negedge DFFE, enabled by reregistered version of CLOCK_ENABLE

We didn't find any tests that disproved this model. There is one other change in this patch, which is to add a behavioural pullup to CLOCK_ENABLE, since the IO tile seems to generate a constant 1 internally if the bitstream does not connect an external net.

To be more certain, I added a different test which dynamically changes the DOUT0 and DOUT1 inputs to ensure both flops are genuinely enabled.

Old simlib:
image
(We think the Xs will be 0s in real life)

New simlib:
image

Capture from logic analyser:
image

### DIFF
--- a/techlibs/ice40/cells_sim.v
+++ b/techlibs/ice40/cells_sim.v
@@ -27,18 +27,27 @@ module SB_IO (
 	reg dout_q_0, dout_q_1;
 	reg outena_q;
 
+	// IO tile generates a constant 1'b1 internally if global_cen is not connected
+	wire clken_pulled = CLOCK_ENABLE || CLOCK_ENABLE === 1'bz;
+	reg  clken_pulled_ri;
+	reg  clken_pulled_ro;
+
 	generate if (!NEG_TRIGGER) begin
-		always @(posedge INPUT_CLK)  if (CLOCK_ENABLE) din_q_0  <= PACKAGE_PIN;
-		always @(negedge INPUT_CLK)  if (CLOCK_ENABLE) din_q_1  <= PACKAGE_PIN;
-		always @(posedge OUTPUT_CLK) if (CLOCK_ENABLE) dout_q_0 <= D_OUT_0;
-		always @(negedge OUTPUT_CLK) if (CLOCK_ENABLE) dout_q_1 <= D_OUT_1;
-		always @(posedge OUTPUT_CLK) if (CLOCK_ENABLE) outena_q <= OUTPUT_ENABLE;
+		always @(posedge INPUT_CLK)                       clken_pulled_ri <= clken_pulled;
+		always @(posedge INPUT_CLK)  if (clken_pulled)    din_q_0         <= PACKAGE_PIN;
+		always @(negedge INPUT_CLK)  if (clken_pulled_ri) din_q_1         <= PACKAGE_PIN;
+		always @(posedge OUTPUT_CLK)                      clken_pulled_ro <= clken_pulled;
+		always @(posedge OUTPUT_CLK) if (clken_pulled)    dout_q_0        <= D_OUT_0;
+		always @(negedge OUTPUT_CLK) if (clken_pulled_ro) dout_q_1        <= D_OUT_1;
+		always @(posedge OUTPUT_CLK) if (clken_pulled)    outena_q        <= OUTPUT_ENABLE;
 	end else begin
-		always @(negedge INPUT_CLK)  if (CLOCK_ENABLE) din_q_0  <= PACKAGE_PIN;
-		always @(posedge INPUT_CLK)  if (CLOCK_ENABLE) din_q_1  <= PACKAGE_PIN;
-		always @(negedge OUTPUT_CLK) if (CLOCK_ENABLE) dout_q_0 <= D_OUT_0;
-		always @(posedge OUTPUT_CLK) if (CLOCK_ENABLE) dout_q_1 <= D_OUT_1;
-		always @(negedge OUTPUT_CLK) if (CLOCK_ENABLE) outena_q <= OUTPUT_ENABLE;
+		always @(negedge INPUT_CLK)                       clken_pulled_ri <= clken_pulled;
+		always @(negedge INPUT_CLK)  if (clken_pulled)    din_q_0         <= PACKAGE_PIN;
+		always @(posedge INPUT_CLK)  if (clken_pulled_ri) din_q_1         <= PACKAGE_PIN;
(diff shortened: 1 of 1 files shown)


whitequark:
:+1: I looked at the SB_IO sim model and I thought it matched hardware but evidently I missed the negedge flop behavior.

Wren6991:
Realised that last test is problematic as data edges are aligned with clock negedge. I think this ends up deterministically using the updated value, as it's a single NBA in one process updating clock and data simultaneously, and SB_IO contains a single clocked process which will be enqueued after the NBA update step... dammit Verilog. Has implications for the hardware too, so I'll move those data transitions back by 90° and try again.

Should also investigate whether the DIN1 input flop uses the reregistered CLOCK_ENABLE; I haven't experimented with this, so left the simlib unchanged. Maybe DIN changes ought to be in a different PR.

Wren6991:
Same results. Sorry for delay.

image

image

Wren6991:
It seems as though there is also a DFF which reregisters CLOCK_ENABLE to INPUT_CLOCK for use by D_IN_1.

Here's the test:

```
module din_test (
	input wire clk_osc,

	output wire mirror_clk,
	output wire mirror_e,
	output wire q0,
	output wire q1,
	output wire ddr_out, // wire loopback ─┐
	input  wire ddr_in   // wire loopback <┘  (must be on separate IO tile due to clke)
);

(* keep = 1'b1 *) reg clk = 0, e = 0, d = 0;
assign {mirror_clk, mirror_e, ddr_out} = {clk, e, d};

reg  [1:0] ctr = 0;
reg  [2:0] lfsr = 0;
wire [2:0] lfsr_next = {~^lfsr[1:0], lfsr[2:1]};

always @ (posedge clk_osc) begin
	ctr <= ctr + 1'b1;
	case (ctr)
		2'h0: begin {clk, e} <= 2'b00;                                  end
		2'h1: begin {clk, e} <= 2'b01; d <= lfsr[0]; lfsr <= lfsr_next; end
		2'h2: begin {clk, e} <= 2'b11;                                  end
		2'h3: begin {clk, e} <= 2'b10; d <= lfsr[0]; lfsr <= lfsr_next; end
	endcase
end

SB_IO #(
	.PIN_TYPE (6'b00_01_00)
	//            │  │  └───── Set DIN0 registered input (+ DIN1 always registered.)
	//            │  └──────── Output registered, oclk tied.
	//            └─────────── Permanent output disable
) dut (
	.CLOCK_ENABLE      (e),
	.INPUT_CLK         (clk),
	.PACKAGE_PIN       (ddr_in),
	.D_IN_0            (q0),
	.D_IN_1            (q1)
);

endmodule
```

```# For HX8k-EVN board

# on-board oscillator
set_io clk_osc J3

# ddr_in needs to be on its own tile due to clke. Others don't matter.
set_io ddr_in  N10 # ─┐
                   #  │ external wire loopback
set_io ddr_out N12 # <┘

set_io mirror_clk T13
set_io mirror_e   T14
set_io q0         T15
set_io q1         T16
```

Simulating with current simlib:
image

Simulating with my most recent version:
image

Capture from real hardware:
image

I've updated my branch to reflect this.

techlibs/ice40/cells_sim.v
```diff
@@ -27,18 +27,27 @@ module SB_IO (
 	reg dout_q_0, dout_q_1;
 	reg outena_q;
 
+	// IO tile generates a constant 1'b1 internally if global_cen is not connected
+	wire clken_pulled = CLOCK_ENABLE === 1'b1 || CLOCK_ENABLE === 1'bz;
```
cliffordwolf: I think `CLOCK_ENABLE == 1'b1 || CLOCK_ENABLE === 1'bz` would be slightly better here (using `==` for the first comparison). This will convert a `1'bz` to `1'b1`, but it will still preserve `1'bx` (because `1'bx || 1'b1 === 1'b1` but `1'bx || 1'b0 === 1'bx`).

Wren6991: Oh, great point, thanks!

The following gives the same X-preserving behaviour, and perhaps requires less brainpower to read:
```suggestion
	wire clken_pulled = CLOCK_ENABLE || CLOCK_ENABLE === 1'bz;
```

cliffordwolf: Yes, that change would also be fine with me. (Not sure which version needs less brain power to read, but no strong opinions either way either. ;)

Wren6991: Ok thanks! Have updated my branch.

PR #943, opened 2019-04-18
[WIP] Add "whitebox" attribute, add "read_verilog -wb"

cliffordwolf:


eddiehung:
Made two small changes -- one for `flatten -wb` and another for disabling the contents of blackboxes (black + white) from being written out in `write_json`

eddiehung:
Another potential talking point: `techmap` will reach into whiteboxes (which is actually useful for my use case)

daveshah1:
The problem seems to be the equivalence check failing because it is skipping the content of whitebox cells - it is running 'read_verilog' without -lib so previously would have seen the content of this cells. The equivalence checking and SAT stuff should probably be looking inside whitebox cells.

cliffordwolf:
> Another potential talking point: techmap will reach into whiteboxes (which is actually useful for my use case)

Yeah, that's a bug. I've now fixed this and added a `wbflip` command that can be used to switch back and forth between synthesis/logic optimization for whiteboxes and the actual design.

cliffordwolf:
> The equivalence checking and SAT stuff should probably be looking inside whitebox cells.

f3ad8d6 should do that now

eddiehung:
Previous `SB_CARRY` issue now resolved.

PR #945, opened 2019-04-20
New behavior for read_verilog handling of whiteboxes

cliffordwolf:
```
    -noblackbox
        do not automatically add a (* blackbox *) attribute to an
        empty module.

    -lib
        only create empty blackbox modules. This implies -DBLACKBOX.
        modules with the (* whitebox *) attribute will be preserved.
        (* lib_whitebox *) will be treated like (* whitebox *).

    -nowb
        delete (* whitebox *) and (* lib_whitebox *) attributes from
        all modules.
```

Without `-lib` it automatically deletes `(* lib_whitebox *)` attributes. You can even do for example `(* whitebox, lib_whitebox = 0 *)`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -310,7 +310,12 @@ Verilog Attributes and non-standard features
   that have the same ports as the real thing but do not contain information
   on the internal configuration. This modules are only used by the synthesis
   passes to identify input and output ports of cells. The Verilog backend
-  also does not output blackbox modules on default.
+  also does not output blackbox modules on default. ``read_verilog``, unless
+  called with ``-noblackbox`` will automatically set the blackbox attribute
+  on any empty module it reads.
+
+- The ``noblackbox`` attribute set on an empty module prevents ``read_verilog``
+  from automatically setting the blackbox attribute on the module.
 
 - The ``whitebox`` attribute on modules triggers the same behavior as
   ``blackbox``, but is for whitebox modules, i.e. library modules that
--- a/README.md
+++ b/README.md
@@ -316,6 +316,9 @@ Verilog Attributes and non-standard features
   ``blackbox``, but is for whitebox modules, i.e. library modules that
   contain a behavioral model of the cell type.
 
+- The ``lib_whitebox`` attribute overwrites ``whitebox`` when ``read_verilog``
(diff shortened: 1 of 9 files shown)


eddiehung:
Can you explain in what situation the new `(* lib_whitebox *)` would be beneficial?

cliffordwolf:
> Can you explain in what situation the new (* lib_whitebox *) would be beneficial, over the capabilities we get from (* whitebox *) alone?

I think you'd actually use `whitebox` (or `blackbox`) very seldom in practice. In most cases you'd use `lib_whitebox`. For example in `techlibs/ice40/cells_sim.v` you'd tag `SB_LUT4` and `SB_CARRY` with `(* lib_whitebox *)`. Then, when you read it with `read_verilog` it's as if the attribute wasn't there, but if you read it with `read_verilog -lib` then you get a whitebox module.

> EDIT: Oh this takes out the -wb flag from read_verilog added in #943

Yes, that's the point. Users would just say "read that lib" by running `read_verilog -lib` without having to care/know about whiteboxes. And users would just say "read those modules" by running `read_verilog` without having to worry that some might be whiteboxes and that this may screw up their flow.

> EDIT2: I have to say I'm getting confused with the lib_whitebox and whitebox options, and also with techmap -wb and flatten -wb compared to the new wbflip. Is there a need for all of them, or will they only serve to confuse?

There definitely is a need for both of them. `wbflip` allows you to operate only on whiteboxes or only on non-whiteboxes, whereas `-wb` allows you to work on both of them at the same time.

You'd use `flatten -wb` if you wanted to flatten your whiteboxes into the modules instantiating them, but you'd run `wbflip; proc; opt -fast; techmap; opt -fast; wbflip` if you wanted to do some logic synthesis and optimization on your whitebox modules only.

cliffordwolf:
Haha. It's because those are now detected as black boxes:

```
(* techmap_simplemap *)
(* techmap_celltype = "$sr $ff $dff $dffe $adff $dffsr $dlatch" *)
module _90_simplemap_registers;
endmodule
```

Fix is on the way..

cliffordwolf:
Fixed it.

eddiehung:
Confirm fix! Cheers.

PR #953, opened 2019-04-22
Add support for zero-width signals to Verilog back-end

### LINKED ISSUE #948: Valid RTLIL produces invalid Verilog, which Yosys then cannot read back
From [an nMigen issue](https://github.com/m-labs/nmigen/issues/41).

bug.rtlil:
```
module \top
  wire width 1 input 0 $signal
  wire width 1 output 1 \a
  wire width 1 $next\a
  wire width 1 $1
  cell $reduce_bool $2
    parameter \A_SIGNED 0
    parameter \A_WIDTH 0
    parameter \Y_WIDTH 1
    connect \A {}
    connect \Y $1
  end
  process $group_0
    assign $next\a 1'0
    assign $next\a $1
    sync init
    sync always
      update \a $next\a
  end
end
```

On latest master (0e7901e4), run the following:

```
yosys -p "read_ilang bug.rtlil; write_verilog bug.v; delete; read_verilog bug.v"
```

Yosys gives a parse error when reading back the Verilog it has just generated. The cause is a zero-item concatenation list, which is *syntactically* illegal (1364-2005 A.8.1).

However, the following succeeds:

```
yosys -p "read_ilang bug.rtlil; synth; write_verilog bug.v; delete; read_verilog bug.v"
```
This produces a `1'h0` assignment to `a`.

## Expected behavior

Zero-item concatenations are explicitly special-cased in `ilang_parser.y`, and zero-width wires are not disallowed in the RTLIL documentation. As far as I have read through the source code (admittedly not very) Yosys copes with these in its in-memory representation.

OTOH the Verilog standard explicitly forbids zero-item concatenations, forbids zero-length `wire`s in a roundabout way, and is surprisingly woolly on zero-length signals produced by expressions (with surprising edge cases, e.g. being allowed inside of concatenations).

As:
- Zero-length signals are not disallowed by RTLIL, but are disallowed by Verilog
- The or reduction of a zero-width signal is logically zero (empty sum convention)
- The current synthesis behaviour is to produce zero for the or reduction

I would expect `read_ilang bug.rtlil; write_verilog bug.v` to produce a zero assignment to `a`. If not, some other valid Verilog, based on some documented behaviour. Note that the above is a minimal example to provoke the problematic behaviour, but it was isolated from real code.

## Actual behavior

Yosys produces invalid Verilog, which it cannot then read back. Paste of `write_verilog` output:

```
/* Generated by Yosys 0.8+354 (git sha1 0e7901e4, gcc 8.2.0-7ubuntu1 -fPIC -O2) */

module top(a, _2_);
  wire _0_;
  reg _1_;
  input _2_;
  output a;
  reg a;
  assign _0_ = | {  };  // syntax error here
  always @* begin
    _1_ = 1'h0;
    _1_ = _0_;
  end
  initial begin
  end
  always @* begin
      a <= _1_;
  end
endmodule
```

cliffordwolf:
Fixes #948

### DIFF
--- a/backends/verilog/verilog_backend.cc
+++ b/backends/verilog/verilog_backend.cc
@@ -187,6 +187,10 @@ void dump_const(std::ostream &f, const RTLIL::Const &data, int width = -1, int o
 {
 	if (width < 0)
 		width = data.bits.size() - offset;
+	if (width == 0) {
+		f << "\"\"";
+		return;
+	}
 	if (nostr)
 		goto dump_hex;
 	if ((data.flags & RTLIL::CONST_FLAG_STRING) == 0 || width != (int)data.bits.size()) {
@@ -340,6 +344,10 @@ void dump_sigchunk(std::ostream &f, const RTLIL::SigChunk &chunk, bool no_decima
 
 void dump_sigspec(std::ostream &f, const RTLIL::SigSpec &sig)
 {
+	if (GetSize(sig) == 0) {
+		f << "\"\"";
+		return;
+	}
 	if (sig.is_chunk()) {
 		dump_sigchunk(f, sig.as_chunk());
 	} else {


whitequark:
@cliffordwolf Is this actually legal? I mean, I wish it would be, because it would save me a lot of trouble, but 1364-2005 is not being very clear on this topic:
Screenshot_20190714_024603
The above seems to imply to me that an empty string should be the same as `8'h00`.

whitequark:
I think there is a way to do this that is unambiguously legal and defined to produce the expected result: `{0{1'0}}`.
Screenshot_20190714_141449

udif:
This is true for IEEE1364-2005 and IEEE1800. Unfortunately IEEE1364-2001 and earlier didn't support this.

whitequark:
@udif Thanks for clarification. Do you happen to know what bit width does an empty string have?

udif:
Sorry, no. I rarely write DV code, only synthesizable RTL, so I had no need for strings.
(Yes, I know they are synthesizable, but in practice I haven't ran into any practical usecase for synthesizable ASCII strings).

cliffordwolf:
I didn't know that about `""`. I always thought that `""` is just zero-extended, but would be zero bits long in a self-sized context. But I just tried with Icarus Verilog and `$bits("")` indeed returns 8. I agree that `{0{1'b0}}` is probably the only remaining option.

PR #958, opened 2019-04-26
[WIP] Add "split shiftx" to split multi-bit $shiftx cells into single-bit ones

eddiehung:
Taking into account any `$macc` cells -- configured a constant multiplier -- that might be driving the `$shiftx`'s `B` input.

### DIFF
--- a/passes/pmgen/split_shiftx.pmg
+++ b/passes/pmgen/split_shiftx.pmg
@@ -1,0 +1,56 @@
+state <SigSpec> shiftxB
+
+match shiftx
+	select shiftx->type == $shiftx
+	select param(shiftx, \Y_WIDTH).as_int() > 1
+endmatch
+
+code shiftxB
+	shiftxB = port(shiftx, \B);
+	const int b_width = param(shiftx, \B_WIDTH).as_int();
+	if (param(shiftx, \B_SIGNED) != 0 && shiftxB[b_width-1] == RTLIL::S0)
+		shiftxB = shiftxB.extract(0, b_width-1);
+endcode
+
+match macc
+	select macc->type == $macc
+	select param(macc, \B_WIDTH).as_int() == 0
+	index <SigSpec> port(macc, \Y) === shiftxB
+	optional
+endmatch
+
+code shiftxB
+	if (macc) {
+		Const config = param(macc, \CONFIG);
+		const int config_width = param(macc, \CONFIG_WIDTH).as_int();
+		const int num_bits = config.extract(0, 4).as_int();
+		const int num_ports = (config_width - 4) / (2 + 2*num_bits);
+		if (num_ports != 1) {
+			shiftxB = nullptr;
+			reject;
+		}
+		// IS_SIGNED?
+		if (config[4] == 1) {
+			shiftxB = nullptr;
+			reject;
+		}
+		// DO_SUBTRACT?
+		if (config[5] == 1) {
+			shiftxB = nullptr;
+			reject;
+		}
+		const int port_size_A = config.extract(6, num_bits).as_int();
+		const int port_size_B = config.extract(6 + num_bits, num_bits).as_int();
+		const SigSpec port_B = port(macc, \A).extract(port_size_A, port_size_B);
+		if (!port_B.is_fully_const()) {
+			shiftxB = nullptr;
+			reject;
+		}
+		const int multiply_factor = port_B.as_int();
+		if (multiply_factor != param(shiftx, \Y_WIDTH).as_int()) {
+			shiftxB = nullptr;
+			reject;
+		}
+		shiftxB = port(macc, \A).extract(0, port_size_A);
+	}
+endcode
--- a/passes/pmgen/split_shiftx.pmg
+++ b/passes/pmgen/split_shiftx.pmg
@@ -5,50 +5,51 @@ match shiftx
 	select param(shiftx, \Y_WIDTH).as_int() > 1
 endmatch
 
-code shiftxB
-	shiftxB = port(shiftx, \B);
-	const int b_width = param(shiftx, \B_WIDTH).as_int();
-	if (param(shiftx, \B_SIGNED) != 0 && shiftxB[b_width-1] == RTLIL::S0)
-		shiftxB = shiftxB.extract(0, b_width-1);
-endcode
-
 match macc
 	select macc->type == $macc
 	select param(macc, \B_WIDTH).as_int() == 0
-	index <SigSpec> port(macc, \Y) === shiftxB
 	optional
 endmatch
 
 code shiftxB
 	if (macc) {
+		shiftxB = port(shiftx, \B);
+		const int b_width = param(shiftx, \B_WIDTH).as_int();
+		if (param(shiftx, \B_SIGNED) != 0 && shiftxB[b_width-1] == RTLIL::S0)
+			shiftxB = shiftxB.extract(0, b_width-1);
+		if (port(macc, \Y) != shiftxB) {
+			blacklist(shiftx);
+			reject;
+		}
+
 		Const config = param(macc, \CONFIG);
 		const int config_width = param(macc, \CONFIG_WIDTH).as_int();
 		const int num_bits = config.extract(0, 4).as_int();
 		const int num_ports = (config_width - 4) / (2 + 2*num_bits);
 		if (num_ports != 1) {
-			shiftxB = nullptr;
+			blacklist(shiftx);
 			reject;
 		}
 		// IS_SIGNED?
 		if (config[4] == 1) {
-			shiftxB = nullptr;
+			blacklist(shiftx);
 			reject;
 		}
 		// DO_SUBTRACT?
 		if (config[5] == 1) {
-			shiftxB = nullptr;
+			blacklist(shiftx);
 			reject;
 		}
 		const int port_size_A = config.extract(6, num_bits).as_int();
 		const int port_size_B = config.extract(6 + num_bits, num_bits).as_int();
 		const SigSpec port_B = port(macc, \A).extract(port_size_A, port_size_B);
 		if (!port_B.is_fully_const()) {
-			shiftxB = nullptr;
+			blacklist(shiftx);
 			reject;
 		}
 		const int multiply_factor = port_B.as_int();
 		if (multiply_factor != param(shiftx, \Y_WIDTH).as_int()) {
-			shiftxB = nullptr;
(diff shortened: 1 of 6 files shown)


eddiehung:
@cliffordwolf Please cast your gaze over this and let me know if there's a more efficient way of using it. In particular, I am struggling with how to say to tell the pattern matcher to backtrack/rollback past `$macc`'s `optional` block.

The story is thus: I am trying to identify `$macc`'s driving `$shiftx`.`B`. Of course, a `$macc` is optional in the sense that not all `$shiftx` will be driven by one. However, I would like to express that: when `$macc` is present, then I need it to fulfil *these* conditions otherwise the whole pattern is void. Right now, if I reject in my last code block, it will simply backtrack to the `optional` statement, and treat the `$macc` as not being there. 

*Perhaps* this isn't the best way of using the pattern matcher, but *perhaps* the pattern matcher could be improved? (c.f. peephole optimisations)

cliffordwolf:
I think you are trying to make this translation too late in the flow. I think this should be done before alumacc is run.

cliffordwolf:
Also, instead of splitting it, it might be better to just "pad" the word size to the next power of two.

passes/pmgen/split_shiftx.pmg
```diff
@@ -0,0 +1,59 @@
...
+
+match shiftx
+	select shiftx->type == $shiftx
+	select param(shiftx, \Y_WIDTH).as_int() > 1
+endmatch
+
+match macc
+	select macc->type == $macc
+	select param(macc, \B_WIDTH).as_int() == 0
+	optional
```
cliffordwolf: Why is this optional? Below you only seem to do something when you found a macc cell.

Also this will run the code block below for all combinations of shiftx and macc cells inside the module. I.e. if you have 100 shiftx cells in a module and 100 macc cells, then the `code..endcode` block below is going to be executed 10000 for all the possible combinations, even if not a single macc is even connected to any shiftx cell.

So this should absolutely use something like `index <SigSpec> port(macc, \Y) === port(shiftx, \Y)` or `index <SigSpec> port(macc, \Y) === shiftxB`.

passes/pmgen/split_shiftx.pmg
```diff
@@ -0,0 +1,59 @@
...
+endmatch
+
+code shiftxB
+	shiftxB = port(shiftx, \B);
+
+	if (macc) {
+		const int b_width = param(shiftx, \B_WIDTH).as_int();
+		if (param(shiftx, \B_SIGNED) != 0 && shiftxB[b_width-1] == RTLIL::S0)
+			shiftxB = shiftxB.extract(0, b_width-1);
```
cliffordwolf: None of this depends on macc so it should probably be done right after matching shiftx.

passes/pmgen/split_shiftx.pmg
```diff
@@ -0,0 +1,59 @@
...
+
+		if (port(macc, \Y) != shiftxB) {
+			blacklist(shiftx);
+			reject;
```
cliffordwolf: I don't understand what you are trying to do here. You are blacklisting this shiftx when it happens to not match the one macc cell you just tested? What if it matches another macc cell in the module? You'll never know, because this will backtrack and just select the next shiftx.

passes/pmgen/split_shiftx.pmg
```diff
@@ -0,0 +1,59 @@
...
+		}
+
+		Const config = param(macc, \CONFIG);
+		const int config_width = param(macc, \CONFIG_WIDTH).as_int();
+		const int num_bits = config.extract(0, 4).as_int();
+		const int num_ports = (config_width - 4) / (2 + 2*num_bits);
+		if (num_ports != 1) {
+			blacklist(shiftx);
+			reject;
```
cliffordwolf: `num_ports != 1` doesn't depend on shiftx or any other state, just on the macc cell. So this should be a `select` statement inside of `match macc` so it is only done once when the pattern matcher initializes.

I also don't understand why you are blacklisting shiftx here.

passes/pmgen/split_shiftx.pmg
```diff
@@ -0,0 +1,59 @@
...
+		}
+		// IS_SIGNED?
+		if (config[4] == 1) {
+			blacklist(shiftx);
+			reject;
```
cliffordwolf: Same as above.

passes/pmgen/split_shiftx.pmg
```diff
@@ -0,0 +1,59 @@
...
+		}
+		// DO_SUBTRACT?
+		if (config[5] == 1) {
+			blacklist(shiftx);
+			reject;
```
cliffordwolf: Same as above.

passes/pmgen/split_shiftx.pmg
```diff
@@ -0,0 +1,59 @@
...
+		}
+		const int port_size_A = config.extract(6, num_bits).as_int();
+		const int port_size_B = config.extract(6 + num_bits, num_bits).as_int();
+		const SigSpec port_B = port(macc, \A).extract(port_size_A, port_size_B);
+		if (!port_B.is_fully_const()) {
+			blacklist(shiftx);
+			reject;
```
cliffordwolf: Same as above.

passes/pmgen/split_shiftx.pmg
```diff
@@ -0,0 +1,59 @@
...
+		}
+		const int multiply_factor = port_B.as_int();
+		if (multiply_factor != param(shiftx, \Y_WIDTH).as_int()) {
+			blacklist(shiftx);
+			reject;
```
cliffordwolf: Same as above. And I think you want `multiply_factor > param(shiftx, \Y_WIDTH).as_int()`.

passes/pmgen/split_shiftx.pmg
eddiehung: > Why is this optional? Below you only seem to do something when you found a macc cell.

Not the full picture. I want to consider doing something for any multi-bit `$shiftx` where `Y_WIDTH > 1`, where:
1. If there is a `$macc` attached to B, check that it must conform to some conditions, otherwise ignore this `$shiftx` (i.e. let it get blown to soft logic). Accepted pattern: `$macc` `->` `$shiftx`.
2. If there is no `$macc` then accept the pattern `$shiftx`
3. Potentially other patterns down the line?

> Also this will run the code block below for all combinations of shiftx and macc cells inside the module. I.e. if you have 100 shiftx cells in a module and 100 macc cells, then the code..endcode block below is going to be executed 10000 for all the possible combinations, even if not a single macc is even connected to any shiftx cell.

Yup. I'm trying to nudge you towards thinking about (a) whether pmgen is the right tool for the job *ever*; (b) whether pmgen could be made into the right tool for the job. I think it would be a great little general purpose layer to have, without having to dive into writing your own custom pass to merge two cells. 

> So this should absolutely use something like index <SigSpec> port(macc, \Y) === port(shiftx, \Y) or index <SigSpec> port(macc, \Y) === shiftxB.

`index` doesn't currently let me examine subsets, right?
I've observed that sometimes `B_SIGNED` is set, and the MSB of `B` is `1'b0` -- thus checking if `macc.Y` === `shiftx.Y` will fail because of that extra sign bit. Thus, I used a `code` block to trim that sign bit off if the opportunity arises.

passes/pmgen/split_shiftx.pmg
eddiehung: See above re: trimming constant-0 sign bit.

passes/pmgen/split_shiftx.pmg
eddiehung: This is my hacky way of communicating to my pass that we should ignore this entire pattern, this pattern being that:
1. We have a `$shiftx`
2. That `$shiftx` is being driven by a `$macc`.
3. But that `$macc` doesn't fulfill all conditions (in this case, Y does not drive a potentially-sign-bit-less `B`)
As I said in the description, I'm not sure how I can express "give up on this `$shiftx` pattern", if that makes more sense now.

eddiehung:
For context, this design:
```
module mux_index_3_3 #(parameter N=3, parameter W=3) (input [N*W-1:0] i, input [$clog2(N)-1:0] s, output [W-1:0] o);
assign o = i[s*W+:W];
endmodule
```

Generates this dump:
```
module \mux_index_3_3

  parameter \N
  parameter \W

  wire width 4 $add$mux_index_3_3.v:3$2_Y

  attribute \src "mux_index_3_3.v:2"
  wire width 9 input 1 \i

  attribute \src "mux_index_3_3.v:2"
  wire width 3 output 3 \o

  attribute \src "mux_index_3_3.v:2"
  wire width 2 input 2 \s

  attribute \src "mux_index_3_3.v:3"
  cell $macc $auto$alumacc.cc:354:replace_macc$9
    parameter \A_WIDTH 4
    parameter \B_WIDTH 0
    parameter \CONFIG 10'1010000010
    parameter \CONFIG_WIDTH 10
    parameter \Y_WIDTH 4
    connect \A { 2'11 \s }
    connect \B { }
    connect \Y $add$mux_index_3_3.v:3$2_Y
  end

  attribute \src "mux_index_3_3.v:3"
  cell $shiftx $shiftx$mux_index_3_3.v:3$3
    parameter \A_SIGNED 0
    parameter \A_WIDTH 9
    parameter \B_SIGNED 1
    parameter \B_WIDTH 5
    parameter \Y_WIDTH 3
    connect \A \i
    connect \B { 1'0 $add$mux_index_3_3.v:3$2_Y }
    connect \Y \o
  end
end
```

passes/pmgen/split_shiftx.pmg
eddiehung: My reasoning for this WIP is to have a stab at expressing what patterns I want to find in the design, and identify where the efficiencies are and how we can solve them. I think one of the things that all patterns would be able to benefit from would be the ability to walk connectivity (sing it with me) 

*The `$shiftx` cells is connected to the.... `$macc` cell.....*

eddiehung:
Okay, I'm closing this PR because I'm giving up on a pmgen approach in favour of a `muxcover` one for my particular problem. But I do want to continue this discussion about how to make pmgen more efficient, perhaps over a different channel.

cliffordwolf:
> make pmgen more efficient

Just want to add here, I think pmgen already is very efficient. You just used it in a very inefficient way. But that's not pmgen's fault. With the changes I had suggested your pmgen matcher would be blazing fast.

PR #963, opened 2019-04-26
Revert synth_xilinx 'fine' label more to how it used to be...

eddiehung:
I think this is much more like how it used to be before I started mucking about with it for the `xc7srl` branch. Sorry.


eddiehung:
Sorry, just realised `shregmap` after `techmap` won't work since `$shiftx`-es will be blown away...

cliffordwolf:
Just btw, since there's now more activity in `synth_xilinx`, maybe we should finally convert it to `ScriptPass`?

eddiehung:
> Just btw, since there's now more activity in `synth_xilinx`, maybe we should finally convert it to `ScriptPass`?

Done in #962

eddiehung:
@cliffordwolf Okay back to passing state. Are you happy with where the `pmux2shiftx` call sits (inside `fine`)?

techlibs/xilinx/synth_xilinx.cc
```diff
@@ -205,45 +205,41 @@ struct SynthXilinxPass : public ScriptPass
 		}
 
 		if (check_label("fine")) {
-			run("opt -fast");
+			// shregmap -tech xilinx can cope with $shiftx and $mux
+			//   cells for identifiying variable-length shift registers,
+			//   so attempt to convert $pmux-es to the former
+			if (!nosrl || help_mode)
+				run("pmux2shiftx", "(skip if '-nosrl')");
```
cliffordwolf: Fine for now. In the long run we probably want to replace `synth -run coarse` above with an explicit coarse grain script and have `pmux2shiftx` there. We can call it a 2nd time here if that helps with shift register inference.

eddiehung: In terms of soft-logic `pmux2shiftx` is a net-zero transformation pass, as ABC _should_ figure it all out. However, for hard-blocks like shift registers and wide multiplexers, this transformation makes the analysis much easier. Thus, I would argue that in the absence of inferring those blocks, we shouldn't be calling `pmux2shiftx`.

PR #969, opened 2019-04-30
Improve pmgen, Add "peepopt" pass with shift-mul pattern

cliffordwolf:
Add `peepopt` pass with a shift-mul pattern, addressing the same issue as #958.

Converts this

```
module mux_index_3_3 #(parameter N=3, parameter W=3) (input [N*W-1:0] i, input [$clog2(N)-1:0] s, output [W-1:0] o);
assign o = i[s*W+:W];
endmodule
```

into that

image

(with ``yosys -p 'synth -run coarse; show -width' test.v``)


passes/pmgen/peepopt.cc
```diff
@@ -0,0 +1,68 @@
...
+		//   |---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|
+		log("\n");
+		log("    peepopt [options] [selection]\n");
+		log("\n");
+		log("This pass applies a collection of peephole optimizers to the current design.\n");
+		log("\n");
+	}
+	void execute(std::vector<std::string> args, RTLIL::Design *design) YS_OVERRIDE
+	{
+		log_header(design, "Executing PEEOPOPT pass (run peephole optimizers).\n");
```
eddiehung: `PEEOPOPT` -> `PEEPOPT`

passes/pmgen/ice40_dsp.pmg
```diff
@@ -1,3 +1,5 @@
+pattern ice40_dsp
```
eddiehung: What does this new `pattern` keyword do?
(Does it need to go into the doc?)

passes/opt/wreduce.cc
```diff
@@ -529,6 +529,42 @@ struct WreducePass : public Pass {
 						module->connect(sig, Const(0, GetSize(sig)));
 					}
 				}
+
```
eddiehung: This isn't a peepopt!

passes/pmgen/peepopt.cc
cliffordwolf: Fixed

passes/pmgen/ice40_dsp.pmg
cliffordwolf: Yes, added to the doc now. I'm also planning on adding `subpattern` statements as well as `namespace` and `endnamespace`. I'll update the doc accordingly as soon as I add them.

passes/opt/wreduce.cc
cliffordwolf: No, but it simplifies pattern matching involving `$div/$mod/$pow` cells if we do at least a bit of wreduction on those cells as well.

cliffordwolf:
@eddiehung Does this solve #958 for you? (This should be run after `wreduce` and before `alumacc` and `share`.)

eddiehung:
> @eddiehung Does this solve #958 for you? (This should be run after `wreduce` and before `alumacc` and `share`.)

I merged this into my xc7mux branch and it works great, muchos thank-you-ias. It looks like all my mux inference tests are coming back with the expected (minimal) area now. 

However, I currently do it after `synth -coarse`, and it seems to work. Will reconsider that whole piece when revisiting DSP inference.

Happy for you to merge after removing WIP in title.

PR #971, opened 2019-04-30
synth_xilinx: Add -nocarry and -nomux options.

koriakin:


eddiehung:
Can I start a discussion on the need for blocking `LUT7`/`LUT8` via `MUXF[78]`?
There are a number of factors at play:
- I understand that on architectures like Spartan6, there are 50% fewer MUXF7/8 primitives. In that case, we shouldn't be blocking their use, but increasing the cost of their use.
- I am working on tighter integration with ABC, and as part of that work I am tuning the area/delay costs of all LUTs, and combined with providing blackbox delay information to ABC, I am seeing far fewer LUT7/8s being used. This is still a few weeks away from being ready though, but something to think about moving forward.
- Lastly, I am already using `-nomux` as an option to block wide multiplexer support (i.e. inferring an 8:1 mux more efficiently using 2xLUT6s + MUXF7), so perhaps we should consider naming both options in a less confusing fashion -- perhaps `-nolut78`  and `-nowidemux`?
Cheers!

koriakin:
> Goddamn it I deleted my comment when I meant to move into review. Let me try again (sorry if it's brief)
> 
> Let's have a discussion on why we're blocking `LUT[78]`s with `-nomux`:
> 
> * I understand that Spartan6 has 50% fewer MUXF7/8s, in which case we should be increasing the cost of their use, not banning them outright, no?

Yes.  This is supposed to be a debug option only, to rule out MUXF7/MUXF8 support issues in later parts of the toolchain.

> * I'm actually working on tighter integration with ABC. As part of this work, I am tuning the area/delay costs of all LUTs, and in combination with communicating blackbox delay/connectivities to ABC, I am seeing far fewer LUT7/8s being used. For example, right now it thinks that a LUT6 and a LUT7 have the same delay (but double the area) -- which is patently untrue. This is still some weeks away from being merge-able, but still something to be thinking of moving forward.
> * Lastly, I am already using `-nomux` as an option to stop wide mux inference (i.e. 8:1 mux implemented efficiently as 2xLUT6+MUXF7), so perhaps we should reconsider the naming of both options to avoid confusion. I'll start by suggesting `-nolut78` and `-nowidemux`?

ECP5 already has an option to disable wide muxes and calls it `-nomux`, I think we should be consistent there.

> 
> Cheers!

eddiehung:
> ECP5 already has an option to disable wide muxes and calls it -nomux, I think we should be consistent there.

Good spot! Consistency could also mean we change the same option in `synth_ecp5` as I would like to maintain that `-nomux` is not an accurate description of what the flag is doing -- @daveshah1 do you want to cast a deciding vote?

Let me change my suggestions to: 
* `-nowideluts`?
* `-nolargeluts`?

daveshah1:
I think that there should be a new option to disable inference of hard multiplexers from wide muxes not called `-nomux`. If it is kept then `-nomux` should disable all usage of hard muxes either from large LUTs or inference, imo. Another option could disable the wide LUTs only.

eddiehung:
> I think that there should be a new option to disable inference of hard multiplexers from wide muxes not called `-nomux`. If it is kept then `-nomux` should disable all usage of hard muxes either from large LUTs or inference, imo. Another option could disable the wide LUTs only.

I think in this case, @koriakin is just asking for an option to disable wide LUTs. We can figure out the wide mux option later.

Is `-nowidelut` a good replacement for the proposed `-nomux` option for `synth_xilinx` here, and for `synth_ecp5` too?

daveshah1:
Sorry, looks like I forgot to reply to this one.

I am happy with `-nowidelut` going forward in the future. I think we should keep `-nomux` in `synth_ecp5` as well however to avoid breaking existing scripts.

eddiehung:
Do you see many people using `-nomux` in ECP5?

If you really want to keep it, can we add `-nowidelut` to `synth_xilinx` and `synth_ecp5`, and alias this to `-nomux` for `synth_ecp5`?

daveshah1:
At least right now it can give a significant area saving so at least a few people are using it, enough that I'd rather not break it.

cliffordwolf:
FWIW I think that `-nomux` should probably disable all inference of hard MUXes, be it from larger LUTs or wide multiplexers. To me that seems like the most sensible definition from the user point of view.

There may be other options for more fine-grain control over what sources of MUX inference to allow, but I would guess that's more something that we would use internally for testing/development, not necessarily something that users might want to use.

eddiehung:
I'm not sure I agree. From a user perspective, why would they want to turn off either source of muxes? One is to reduce LUT size, since our use of LUT7/8s may not be the most efficient considering soft costs like reduced placement and routing flexibility. The other is to (supposedly) increase area efficiency.

In any case, for my wide mux work I'm moving away from a Boolean flag to an integer one to control the threshold for hard mux inference, so my original protest is less relevant now, but I maintain that `-nomux` has a potential for confusion.

Issue #968, opened 2019-04-29
Loop variables don't persist following the loop but within the same block


ZipCPU:
## Steps to reproduce the issue

Consider the following code:

```verilog
module mcve1(i_value, o_value);
	input	wire	[3:0]	i_value;
	output	reg	[7:0]	o_value;

	integer k;
	always @(*)
	begin
		for(k=0; k<4; k=k+1)
			o_value[k] = i_value[k];
		o_value[4] = i_value[k];
		o_value[5] = i_value[k];
		o_value[6] = i_value[k];
		o_value[7] = i_value[k];
	end
endmodule
```

and process it with
```
yosys -p 'read -sv mcve1.v; synth; write_verilog mcve1y.v'
```

## Expected behavior

I'd expect the resulting Verilog code to look something like,

```verilog
    assign o_value = { {(5){i_value[3]}}, i_value[2:0] };
```

## Actual behavior

Instead, I get

```verilog
    assign o_value = { 4'hx, i_value };
```

The difference between the two is whether or not the loop variable maintains its last value after exiting the context of the for loop.

cliffordwolf:
> I'd expect the resulting Verilog code to look something like, ``assign o_value = { {(5){i_value[3]}}, i_value[2:0] };``

Actually, `k` is 4 after the for-loop is completed, not 3. Updated test case with wider `i_value`:

```
module mcve1(i_value, o_value);
	input	wire	[4:0]	i_value;
	output	reg	[7:0]	o_value;

	integer k;
	always @(*)
	begin
		for(k=0; k<4; k=k+1)
			o_value[k] = i_value[k];
		o_value[4] = i_value[k];
		o_value[5] = i_value[k];
		o_value[6] = i_value[k];
		o_value[7] = i_value[k];
	end
endmodule
```

And this one is actually broken. Fixed in #972.

(Verific results are looking fine with the updated test case.)

ZipCPU:
I just tested this patch within a much larger design, and so I'm now getting driver-driver conflicts based upon the loop variables.

cliffordwolf:
@ZipCPU can you send me that design?

ZipCPU:
You should have a copy now.

With this patch, yosys also generates the same warnings when verifying [this design](https://github.com/ZipCPU/wb2axip/blob/master/rtl/wbxbar.v).  See [this sby file for reference](https://github.com/ZipCPU/wb2axip/blob/master/bench/formal/wbxbar.v)

[...]

PR #972, opened 2019-04-30
Add final loop variable assignment when unrolling for-loops

cliffordwolf:
Fixes this test case (per #968):

```
module test(i_value, o_value);
	input	wire	[4:0]	i_value;
	output	reg	[7:0]	o_value;

	integer k;
	always @(*)
	begin
		for(k=0; k<4; k=k+1)
			o_value[k] = i_value[k];
		o_value[4] = i_value[k];
		o_value[5] = i_value[k];
		o_value[6] = i_value[k];
		o_value[7] = i_value[k];
	end
endmodule
```


ZipCPU:
Recommend reverting this patch.  I'm now getting driver-driver conflicts on loop variables in other code which I wasn't getting before.

PR #975, opened 2019-05-01
Re-enable "final loop assignment" feature and fix opt_clean warnings

### LINKED ISSUE #968: Loop variables don't persist following the loop but within the same block

cliffordwolf:
Second attempt at fixing #968.


ZipCPU:
This passes all of my current tests.  Sorry for the late review.

PR #979, opened 2019-05-02
fail svinterfaces testcases on yosys error exit

jakobwenzel:
Hi!

The svinterfaces testcases did not fail when yosys exited with a nonzero exit code. Instead, they continued and compared possibly out-of-date outputs.

Regards,
Jakob


cliffordwolf:
Fixed another obvious bug in that file in d2aa123226f39fb6d076b9a0add2ad4f0e596166.

PR #980, opened 2019-05-02
Extend AST's dumpVlog to output actual verilog code

jakobwenzel:
Hi!

I am planning to use Yosys for some AST level code transformations. Therefore, I need to be able to export the AST as working Verilog code.

I extended AstNode's dumpVlog function to no longer dump Verilog *Pseudo* Code, but rather working Verilog code. It can be parsed again and produce an identical AST. To that end, I made some changes in the Verilog parser, hopefully without breaking stuff.

I am not very familiar with Yosys' code base, so please have a close look at my work. :-)

Kind Regards,
Jakob

### DIFF
--- a/frontends/ast/ast.cc
+++ b/frontends/ast/ast.cc
@@ -262,16 +262,19 @@ AstNode::~AstNode()
 
 // create a nice text representation of the node
 // (traverse tree by recursion, use 'other' pointer for diffing two AST trees)
-void AstNode::dumpAst(FILE *f, std::string indent) const
+void AstNode::dumpAst(FILE *f, std::string indent, bool dumpFileLine) const
 {
 	if (f == NULL) {
 		for (auto f : log_files)
-			dumpAst(f, indent);
+			dumpAst(f, indent, dumpFileLine);
 		return;
 	}
 
 	std::string type_name = type2str(type);
-	fprintf(f, "%s%s <%s:%d>", indent.c_str(), type_name.c_str(), filename.c_str(), linenum);
+    	fprintf(f, "%s%s", indent.c_str(), type_name.c_str());
+	if (dumpFileLine) {
+	    fprintf(f, " <%s:%d>", filename.c_str(), linenum);
+	}
 
 	if (!flag_no_dump_ptr) {
 		if (id2ast)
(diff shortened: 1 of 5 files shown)


udif:
I built your PR out of curiocity to see how it expands generate blocks (one of the most frustrating issues working with generate/endgenerate is to visualize what your code actually looks like flat), and noticed the following (probably incorrect) translation in tests/simple/generate.v:

In the code below, i is a genvar:
```
                casez ({a[i], b[i], carry[i]})
                        3'b?11, 3'b1?1, 3'b11?:
                                carry[i+1] = 1;
                        default:
                                carry[i+1] = 0;
                endcase
```

Got turned into (this is for i=0, 

```
            case ({carry[0:0], b[0:0], a[0:0]})
              3'b -11, 3'b 1-1, 3'b 11-:
                carry[1:1] = 1;
              default:
                carry[1:1] = 0;
            endcase
```
Notice that casez->case and '?' in case items got turned into '-'.

jakobwenzel:
Dear udif,

I only ever tested dumping code with the `-dump_vlog1` switch. Apparently, you used the `-dump_vlog2` switch (or some other way to dump after simplify?). For the testcase you posted, `dumpVlog` actually crashed. The output you posted seems to be from before the PR. Probably Yosys crashed unnoticed and did not update your output file.

The simplify process leaves some stray `if`, `for` and `case` statements without children, which my code could not handle before. I have now pushed a new commit that fixes this issue. 

Kind Regards,
Jakob

udif:
Thanks for the quick response. I recompiled and ran with both dump_vlog1 and dump_vlog2 and will look at this later today or tomorrow.
I must confess that apparently I have earlier ran 'yosys' by mistake instead of './yosys' causing me to run the locally installed yosys which is older, and has only 'dump_vlog', not vlog1/2.
The output I gave is apparently from there.

udif:
@oe1cxw, I wonder if we could add a regression that takes a few existing tests, generates verilog out of them, and compares the output against the original source thorough symbiyosys.

nakengelhardt:
Hi, sorry for leaving this lying around for so long. While this would be nice to have, I'm afraid the verilog frontend is a bit of a superfund site and we have to be careful introducing changes there lest we accidentally break things in subtle ways. After discussion we decided not to merge this PR. I hope you were nevertheless able to accomplish what you needed without having this feature upstreamed!

jakobwenzel:
Hi, I can Unserstand your concerns. I am also not sure that I did not break the parser.

How would you feel about the mergeability of a subset of this patch that only changes the dumpVlog function? Nodes thar were dumped and then read in would no longer be equals but still functionally identical.

nakengelhardt:
I'll put it on the agenda to be discussed next time! It might be a bit until I have an answer for you though, as Claire, our expert for the frontend, is not in office for the next three weeks I believe.

cliffordwolf:
> How would you feel about the mergeability of a subset of this patch that only changes the dumpVlog function?

Yes, something that's limited to only dumpVlog would be much more mergable.

Issue #982, opened 2019-05-02
INIT attribute not set on FDRE when shift register inference


eddiehung:
## Steps to reproduce the issue

`yosys -p "synth_xilinx; dump t:FDRE" lfsr_24.v`

```
module lfsr_24 (input clk, output dout);
  reg [24:1] state = 24'b0;
  always @(posedge clk)
    state <= { state[24-1:1], state[24] ~^ state[23] ~^ state[22] ~^ state[17] };
  assign dout = state[24];
endmodule
```

## Expected behavior

`FDRE` to have `INIT` parameter set to `1'b0`

## Actual behavior

`FDRE` has `INIT` parameter set to `1'bx`
```
<<SNIP>>
  cell \FDRE $auto$simplemap.cc:420:simplemap_dff$18.fpga_srl_1
    parameter \INIT 1'x
    connect \C \clk
    connect \CE 1'1
    connect \D $auto$simplemap.cc:420:simplemap_dff$18.T0
    connect \Q \state [16]
    connect \R 1'0
  end
<<SNIP>>
```
Works when shift register inference disabled (`synth_xilinx -nosrl`).
Could this be related to #956 (though there are no warning messages as that issue has)

But I did notice this:
```
2.27. Executing DFFINIT pass (set INIT param on FF cells).
Setting lfsr_24.$auto$simplemap.cc:420:simplemap_dff$18.fpga_srl_1.INIT (port=Q, net=\state [16]) to 1'x.
Setting lfsr_24.$auto$simplemap.cc:420:simplemap_dff$40.INIT (port=Q, net=\state [22]) to 1'0.
Setting lfsr_24.$auto$simplemap.cc:420:simplemap_dff$41.INIT (port=Q, net=\state [23]) to 1'0.
Removing init attribute from wire lfsr_24.state.
Removed 0 unused cells and 47 unused wires.
```

eddiehung:
Issue appears to be that `shregmap` will set away the `(* init *)` attribute on signals that it infers into a shift register to `1'bx`. However, our tech mapping rules can sometimes blow these inferred shift registers back into flops, and when it does that, a subsequent `dffinit` will init those registers at `1'bx`... (even if I have instantiated a tech specific cell (e.g. `FDRE`))

Number of solutions:
- `shregmap` to now blow away `(* init *)` attribute
- `dffinit` to not overwrite cells with `INIT` already set with new `1'bx` value.

Am thinking latter...

cliffordwolf:
[...]

PR #984, opened 2019-05-03
dffinit to do nothing when (* init *) value is 1'bx

eddiehung:
Fix for #982 

@cliffordwolf If this approach is acceptable, perhaps we should do this even when `-noreinit` is not specified?

### DIFF
--- a/passes/techmap/dffinit.cc
+++ b/passes/techmap/dffinit.cc
@@ -102,7 +102,8 @@ struct DffinitPass : public Pass {
 				if (wire->attributes.count("\\init")) {
 					Const value = wire->attributes.at("\\init");
 					for (int i = 0; i < min(GetSize(value), GetSize(wire)); i++)
-						init_bits[sigmap(SigBit(wire, i))] = value[i];
+						if (value[i] != State::Sx)
+							init_bits[sigmap(SigBit(wire, i))] = value[i];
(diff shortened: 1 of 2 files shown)


passes/techmap/dffinit.cc
```diff
@@ -135,10 +135,18 @@ struct DffinitPass : public Pass {
...
+						if (noreinit && value.bits[i] != State::Sx && value.bits[i] != init_bits.at(sig[i])) {
+							if (init_bits.at(sig[i]) != State::Sx) {
+								log_error("Trying to assign a different init value for %s.%s.%s which technically "
+										"have a conflicted init value.\n",
+										log_id(module), log_id(cell), log_id(it.second));
+							}
+							else {
+								// Trying to overwrite an existing INIT value with 1'bx,
+								//   silently ignore?
+								continue;
```
cliffordwolf: More consistent with other passes would be to add `if (value[i] != State::Sx)` in the code quoted below so that X values are not even added to `init_bits`. (I think the proper behavior would be to not act on X bits in init attributes at all, independent of `-noreinit`.)

https://github.com/YosysHQ/yosys/blob/1f62dc9081feb4852b1848d01951f631853edb38/passes/techmap/dffinit.cc#L103-L105

PR #989, opened 2019-05-04
ABC name recovery fixes

daveshah1:
- Fix accidental n² bug when finding original name
 - Add handling for `new_` prefix and postfixes other than `inv`

Marking [WIP] because it could do with a bit of testing with things like retiming and other odd modes to make sure there aren't any issues like duplicate names.

Also to consider: should we be copying recovered netnames onto the driving cell too with a post/prefix like `driver_` (instead of using an arbitrary name) - this would result in a nicer critical path report.

### DIFF
--- a/passes/techmap/abc.cc
+++ b/passes/techmap/abc.cc
@@ -330,18 +330,31 @@ void extract_cell(RTLIL::Cell *cell, bool keepff)
 std::string remap_name(RTLIL::IdString abc_name, RTLIL::Wire **orig_wire = nullptr)
 {
 	std::string abc_sname = abc_name.substr(1);
-	if (abc_sname.substr(0, 5) == "ys__n") {
+	bool isnew = false;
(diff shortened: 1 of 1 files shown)


cliffordwolf:
> Also to consider: should we be copying recovered netnames onto the driving cell too with a post/prefix like driver_ (instead of using an arbitrary name) - this would result in a nicer critical path report.

That would be an extra pass, right? In what way would that be different from `rename -wire`?

daveshah1:
I was thinking about doing it as part of abc, given that abc (by its very nature) is otherwise unable to produce any kind of meaningful name when it creates cells (this causes the recovered net names to get "lost" among a sea of meaningless cell names)

daveshah1:
Removing [WIP] now as `make test` passes and no new yosys-tests failures are occurring either, and manual tests all seem fine. I'm happy to leave the cell naming issue for now as I don't have the time to implement it right now in any case.

PR #991, opened 2019-05-05
Fix all warnings that occurred when compiling with gcc9

kristofferkoch:
The most annoying warning was that the implicit copy assignment was not defined.

### DIFF
--- a/libs/minisat/Vec.h
+++ b/libs/minisat/Vec.h
@@ -94,6 +94,7 @@ public:
 };
 
 
+
 template<class T, class _Size>
 void vec<T,_Size>::capacity(Size min_cap) {
     if (cap >= min_cap) return;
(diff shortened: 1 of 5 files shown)


libs/minisat/Vec.h
```diff
@@ -94,6 +94,8 @@ class vec {
 };
 
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wclass-memaccess"
```
cliffordwolf: This needs some kind of ifdef guard. Otherwise for example with GCC 7.4.0:

```
test.cc:3:32: warning: unknown option after ‘#pragma GCC diagnostic’ kind [-Wpragmas]
 #pragma GCC diagnostic ignored "-Wclass-memaccess"
                                ^~~~~~~~~~~~~~~~~~~
```

or with clang 6.0.0:

```
test.cc:3:32: warning: unknown warning group '-Wclass-memaccess', ignored [-Wunknown-warning-option]
#pragma GCC diagnostic ignored "-Wclass-memaccess"
```

kristofferkoch: Oh, that's very ugly. Maybe it's better to just have the warning, and then someone can think hard about writing this realloc stuff in a way that is more pleasing both for humans and gcc-9?

For now, I'll just remove the pragmas, then?

cliffordwolf:
Merged. JFYI: caad497

btw, upstream for minisat is https://github.com/niklasso/minisat, but it hasn't seen a commit since 2013.

PR #996, opened 2019-05-06
Optimize ceil_log2 function

mdaiter:
Hey all,
Just was scrolling through `yosys` and this caught my eye. Do you have any thoughts on how to properly incorporate this compiler-specific function? Would really like to land this push, as it leads to a speedup for your `ceil_log2` function. 

Proof:
```
int ceil_log2(int x) {
  return (x > 0) * (32 - __builtin_clz(x-1) - (x == 1));
}

int ceil_orig_log2(int x) {
        if (x <= 0)
                return 0;
        for (int i = 0; i < 32; i++)
                if (((x-1) >> i) == 0)
                        return i;
}
```

compiles to:
```
ceil_log2(int):
  lea edx, [rdi-1]
  mov eax, 32
  bsr edx, edx
  xor edx, 31
  sub eax, edx
  xor edx, edx
  cmp edi, 1
  sete dl
  sub eax, edx
  xor edx, edx
  test edi, edi
  setg dl
  imul eax, edx
  ret
```
and
```
ceil_orig_log2(int):
        push    rbp
        mov     rbp, rsp
        mov     DWORD PTR [rbp-20], edi
        cmp     DWORD PTR [rbp-20], 0
        jg      .L4
        mov     eax, 0
        jmp     .L3
.L4:
        mov     DWORD PTR [rbp-4], 0
.L8:
        cmp     DWORD PTR [rbp-4], 31
        jg      .L6
        mov     eax, DWORD PTR [rbp-20]
        lea     edx, [rax-1]
        mov     eax, DWORD PTR [rbp-4]
        mov     ecx, eax
        sar     edx, cl
        mov     eax, edx
        test    eax, eax
        jne     .L7
        mov     eax, DWORD PTR [rbp-4]
        jmp     .L3
.L7:
        add     DWORD PTR [rbp-4], 1
        jmp     .L8
.L6:
.L3:
        pop     rbp
        ret
```

### DIFF
--- a/kernel/yosys.cc
+++ b/kernel/yosys.cc
@@ -151,14 +151,16 @@ void yosys_banner()
 
 int ceil_log2(int x)
 {
+#if defined(__GNUC__)
+        return x > 1 ? (8*sizeof(int)) - __builtin_clz(x-1) : 0;
+#else
 	if (x <= 0)
 		return 0;
-
 	for (int i = 0; i < 32; i++)
 		if (((x-1) >> i) == 0)
 			return i;
-
 	log_abort();
+#endif
 }
 
 std::string stringf(const char *fmt, ...)
--- a/kernel/yosys.h
+++ b/kernel/yosys.h
@@ -244,7 +244,7 @@ extern bool memhasher_active;
 inline void memhasher() { if (memhasher_active) memhasher_do(); }
 
 void yosys_banner();
-int ceil_log2(int x);
+int ceil_log2(int x) YS_ATTRIBUTE(const);
 std::string stringf(const char *fmt, ...) YS_ATTRIBUTE(format(printf, 1, 2));
 std::string vstringf(const char *fmt, va_list ap);
 int readsome(std::istream &f, char *s, int n);


cliffordwolf:
Note that `__builtin_clz(0)` has _undefined behavior_. Because of this you function doesn't only return a wrong value when called with x=1, it has undefined behavior.

> bsr edx, edx

BSR is undefined when the argument is zero. See for example [here](https://c9x.me/x86/html/file_module_x86_id_20.html).

ABM adds the x86 instruction LZCNT that does not have this property. But it isn't available on all x86 implementations.

> imul eax, edx

Not sure why you felt the need to discuss this at the assembler code level, but here we are. `imul` is usually slower than a branch on modern x86.

In this particular case I think `(x > 0) * y` should be replaced by `x > 0 ? y : 0` if your objective is faster code.

> Proof:

I don't think this proves anything.

Here is an actual proof with cbmc that shows the issue with x=1. Notice that I created a `my__builtin_clz()` that tries to capture the undefinedness of clz(0), but afaik you can only have undefined values in cbmc, not undefined behavior.

```
int my__builtin_clz(int x) {
        for (int i = 0; i < 32; i++)
                if ((x << i) >> 31)
                        return i;
}

int ceil_log2(int x) {
  return (x > 0) * (32 - my__builtin_clz(x-1) - (x == 1));
}

int ceil_orig_log2(int x) {
        if (x <= 0)
                return 0;
        for (int i = 0; i < 32; i++)
                if (((x-1) >> i) == 0)
                        return i;
}

void proof(int x) {
#if 0
        if (x != 1)
#endif
                assert(ceil_log2(x) == ceil_orig_log2(x));
}
```

(Run with `cbmc --function proof --trace test.c`.)

Also, your code assumes that `sizeof(int)==4`, which is not generally true. (The current code just assumes that the value can be represented in a signed 32 bit, but the code still works if that value is stored in a 64-bit integer.)

The following would be one correct implementation of clog2, that's avoiding to call ctz/clz with a zero argument, and is branch-free for positive integer arguments. 

```
int ceil_log2(int x) {
        return x > 1 ? (8*sizeof(int)) - my__builtin_clz(x-1) : 0;
}
```

And the x86 assembler code for this:

```
0000000000000076 <ceil_log2>:
  76:	31 c0                	xor    %eax,%eax
  78:	83 ff 01             	cmp    $0x1,%edi
  7b:	7e 0f                	jle    8c <ceil_log2+0x16>
  7d:	ff cf                	dec    %edi
  7f:	b8 20 00 00 00       	mov    $0x20,%eax
  84:	0f bd ff             	bsr    %edi,%edi
  87:	83 f7 1f             	xor    $0x1f,%edi
  8a:	29 f8                	sub    %edi,%eax
  8c:	c3                   	retq   
```

And for completeness, here's a version using ctz:

```
int ceil_log2(int x) {
        if (x <= 0)
                return 0;
        x -= 1;
        x |= x >> 1;
        x |= x >> 2;
        x |= x >> 4;
        x |= x >> 8;
        x |= x >> 16;
        x += 1;
        return __builtin_ctz(x);
}
```

Both of those implementations are _actually_ proven to be correct with `cbmc`.

kernel/yosys.cc
```diff
@@ -149,16 +149,18 @@ void yosys_banner()
 	log("\n");
 }
 
-int ceil_log2(int x)
+int ceil_log2(const int x)
 {
+#if defined(__GNUC__)
+        return (x > 0) * (32 - __builtin_clz(x-1) - (x == 1));
```
cliffordwolf: ``return x > 1 ? (8*sizeof(int)) - my__builtin_clz(x-1) : 0;``

mdaiter: Did you mean to put "my__builtin_clz"?

mdaiter: Done!

cliffordwolf: No, I did not. That was just a left-over from the correctness proof.  ;)

kernel/yosys.cc
```diff
@@ -149,16 +149,18 @@ void yosys_banner()
 	log("\n");
 }
 
-int ceil_log2(int x)
+int ceil_log2(const int x)
```
cliffordwolf: I didn't look into the rest of the PR before: What are you hoping to achieve with this `const`?

kernel/yosys.h
```diff
@@ -244,7 +244,7 @@ extern bool memhasher_active;
 inline void memhasher() { if (memhasher_active) memhasher_do(); }
 
 void yosys_banner();
-int ceil_log2(int x);
+int ceil_log2(const int x) YS_ATTRIBUTE(const) YS_ATTRIBUTE(pure);
```
cliffordwolf: From the [GCC manual](https://embarc.org/man-pages/gcc/Common-Function-Attributes.html):
> Decorating the same function with both the const and the pure attribute is diagnosed.

So a function is either `pure` or `const`, but it can't have both attributes. (This function is `const` because it is not reading any global memory. If it would read global memory it would only be `pure`.)

mdaiter: Removed the `pure` annotation

kernel/yosys.cc
mdaiter: Just adds clarity to the code. I can remove it if you'd like.

cliffordwolf: Not sure what it clarifies tho.. Please remove it.

mdaiter: Done

PR #999, opened 2019-05-09
initialize more registers in setundef -init

jakobwenzel:
Currently, setundef -init misses some registers that are also assigned to something:

Running the following Verilog code:

    module register #(
    	parameter DEPTH = 10
    ) (
    	input wire clk,
    	input wire in,
    	output wire out
    );
    	reg [DEPTH-1:0] regs;
    	always @(posedge clk) regs <= { regs[DEPTH-2:0], in};
    	assign out = regs[DEPTH-1];
    endmodule

through this script:

    read_verilog register.v
    prep
    setundef -zero -init

## Expected behavior

The register `regs` should be initialized to zero (attribute init set to 0).

    attribute \src "register.v:6"
    wire output 3 \out
    attribute \init 10'0000000000
    attribute \src "register.v:8"
    wire width 10 \regs

## Actual behavior

`out` is initialized to zero.

    attribute \init 1'0
    attribute \src "register.v:6"
    wire output 3 \out
    attribute \src "register.v:8"
    wire width 10 \regs

## Fix
This PR tries to fix the issue by waiting to erase from `ffbits` until all wires have been looked at.

With this PR applied, both `regs` and `out` get the `init` attribute. I am not sure if having the attribute on `out` as well is problematic. Comments?

### DIFF
--- a/passes/cmds/setundef.cc
+++ b/passes/cmds/setundef.cc
@@ -404,22 +404,29 @@ struct SetundefPass : public Pass {
 					initwires.insert(wire);
 				}
 
-				for (int wire_types = 0; wire_types < 2; wire_types++)
-					for (auto wire : module->wires())
-					{
-						if (wire->name[0] == (wire_types ? '\\' : '$'))
-					next_wire:
-							continue;
+				for (int wire_types = 0; wire_types < 2; wire_types++) {
+                                        pool<SigBit> ffbitsToErase;
+                                        for (auto wire : module->wires()) {
+                                                if (wire->name[0] == (wire_types ? '\\' : '$')) {
+                                                        next_wire:
+                                                        continue;
+                                                }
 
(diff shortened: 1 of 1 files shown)


cliffordwolf:
LGTM. Merging.

>  I am not sure if having the attribute on out as well is problematic. Comments?

As long as it's consistent it's fine to set init on multiple wires like this.

cliffordwolf:
Please note the "Coding Style" section in `CodingReadme`. Especially note that Yosys code is indented with tabs, not blanks. I've now fixed that in 6cc60ff.

cliffordwolf:
JFYI, I had a 2nd look and this PR isn't really fixing the issue, just making it less worse. I think I've now completely fixed it in b33176d. (And there's an assert now that will trigger if for some reason some bits would stay un-initialized.)

PR #1011, opened 2019-05-15
Fix two instances of integer-assignment to string.

hzeller:
o In cover.cc, the int-result of mkstemps() was assigned to a string
  and silently interpreted as a single-character filename with a funny
  value. Fix with the intent: assign the filename.
o in libparse.cc, an int was assigned to a string, but depending on
  visible constructors, this is ambiguous. Explicitly cast this to
  a char.


cliffordwolf:
jfyi, because I just saw it: f67ec1b

Issue #1020, opened 2019-05-20
python: AttrObject member functions not available in derived classes


pumbor:
## Preface
In yosys's C++ class hierarchy, several classes derive from AttrObject so that they have access to a common set of attribute getters/setters. A python wrapper is generated for AttrObject, but the wrappers for the derived classes don't include the inheritance information to enable the derived classes to use the AttrObject functions.

## Steps to reproduce the issue

```
cd yosys/examples/python-api
python3 -i script.py
design.top_module().get_src_attribute()
```

## Expected behavior

In python, wrapped derived classes can use AttrObject methods to get to attributes.

## Actual behavior

In python, wrapped derived classes of AttrObject cannot access AttrObject methods:

```
>>> design.top_module().get_src_attribute()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'Module' object has no attribute 'get_src_attribute'
```

cliffordwolf:
cc @btut

mmicko:
This works fine with latest pyosys and new wrappers using pybind11

Issue #995, opened 2019-05-06
Python bindings, OS X


Kmanfi:
## Steps to reproduce the issue
FYI.
To get Python bindings working on **my** OS X machine I made following modifications to Makefile.

ENABLE_LIBYOSYS := 0
ENABLE_LIBYOSYS := **1**
__
PYTHON_DESTDIR := `$(PYTHON_EXECUTABLE)-config --prefix`/lib/python$(PYTHON_VERSION)/dist-packages

PYTHON_DESTDIR := `$(PYTHON_EXECUTABLE)-config --prefix`/lib/python$(PYTHON_VERSION)/**site**-packages
__
ENABLE_NDEBUG := 1
ENABLE_NDEBUG := **0**
__
ifneq ($(shell which brew),)
BREW_PREFIX := $(shell brew --prefix)/opt
$(info $$BREW_PREFIX is [${BREW_PREFIX}])
**CXXFLAGS += -I$(BREW_PREFIX)/boost/include/boost
LDFLAGS += -L$(BREW_PREFIX)/boost/lib**
__
All..
$(PYTHON_EXECUTABLE)-config --libs
$(PYTHON_EXECUTABLE)-config --**ldflags**
__
$(P) $(LD) -o libyosys.so -shared -Wl,-soname,libyosys.so $(LDFLAGS) $^ $(LDLIBS)
$(P) $(LD) -o libyosys.so -shared -Wl,-install_name,libyosys.so $(LDFLAGS) $^ $(LDLIBS)
__
$(INSTALL_SUDO) ldconfig
**#** $(INSTALL_SUDO) ldconfig

## Expected behavior

Python bindings work also on OS X.
## Actual behavior

wip

cliffordwolf:
CC @btut @christian-krieg

Kmanfi:
If I only change
  ENABLE_NDEBUG := 1
then
I get many of these:

```
In file included from kernel/rtlil.cc:20:
In file included from ./kernel/yosys.h:270:
./kernel/log.h:93:33: warning: unused parameter 'n' [-Wunused-parameter]
static inline bool ys_debug(int n = 0) { return false; }
```

btut:
Hi @Kmanfi,
Most of the changes you applied to the makefile can be safely applied without breaking my linux build. I guess it would be safe to apply those to the repo.

This is not true for the following line:
[...]

PR #1027, opened 2019-05-22
Python-api tests

Kmanfi:
**My motivation**: Tests would work also as a reference how to use Yosys with the api.
Here is some initial work for testing python-api. One contains simple sanity tests. Another test is implementation of **JSON writer** by using python-api. Currently JSON writer is wip state but can make output that matches with fixed reference JSON. Goal is to make it match with Yosys JSON output.
There are lines marked with TODO: = need some attention like one already mentioned #1020 and #995 
I tried to add C code behind #ifdef WITH_PYTHON but somehow it's not working. Any **help** is appreciated.
Found that backticks mentioned #1029 disturbed compile flags and WITH_PYHTON. Proposed fix for this in PR #1030

### DIFF
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -3725,6 +3725,7 @@ RTLIL::SigBit RTLIL::SigSpec::as_bit() const
 		return bits_[0];
 }
 
+#ifdef WITH_PYTHON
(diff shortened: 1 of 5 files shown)


Kmanfi:
Let's put this PR on hold until I have explored pybind11.

Issue #478, opened 2018-01-03
PREFIX can't be set in Makefile.conf


Fatsie:
In Makefile the PREFIX variable is used before Makefile.conf is included this means the PREFIX can't be properly set in Makefile.conf.

mmicko:
closed with #1057

PR #1030, opened 2019-05-22
OS X related Makefile fixes.

Kmanfi:
Cleaned my OS X Makefile and put needed if(Darwin..) #1029 fix is also included in this.
This PR fixes also  -DWITH_PYTHON mentioned in #1027

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ PYTHON_VERSION_TESTCODE := "import sys;t='{v[0]}.{v[1]}'.format(v=list(sys.versi
 PYTHON_EXECUTABLE := $(shell if python3 -c ""; then echo "python3"; else echo "python"; fi)
 PYTHON_VERSION := $(shell $(PYTHON_EXECUTABLE) -c ""$(PYTHON_VERSION_TESTCODE)"")
 PYTHON_MAJOR_VERSION := $(shell echo $(PYTHON_VERSION) | cut -f1 -d.)
-PYTHON_PREFIX := `$(PYTHON_EXECUTABLE)-config --prefix`
+PYTHON_PREFIX := $(shell $(PYTHON_EXECUTABLE)-config --prefix)
 PYTHON_DESTDIR := $(PYTHON_PREFIX)/lib/python$(PYTHON_VERSION)/site-packages
 
 # other configuration flags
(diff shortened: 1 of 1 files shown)


cliffordwolf:
I definitely approve of using `$(shell .. )` in PYTHON_PREFIX, however, it would make sense to put the entire block of `PYTHON_* := ` assignments in an `ifeq ($(ENABLE_PYOSYS),1)` block so that those shell commands aren't executed when `ENABLE_PYOSYS` is not active.

cc @btut @mmicko please comment on the pyosys and MacOS aspects of the PR respectively.

mmicko:
Can confirm that without change it does not find boost python at all, and that with changes it compiles and runs fine on macOS.

thasti:
From what I can see, this MR broke building yosys+libyosys+pyosys with having 
```
ENABLE_LIBYOSYS=1
ENABLE_PYOSYS=1
```
in Makefile.conf (on Linux). Any chance of having a second look?


Currently failing with:
```
[Makefile.conf] CONFIG := gcc
[Makefile.conf] ENABLE_LIBYOSYS=1
[Makefile.conf] ENABLE_PYOSYS=1
/bin/sh: -config: command not found
/bin/sh: -config: command not found
/bin/sh: -config: command not found
/bin/sh: -config: command not found
```

mmicko:
Issue here is that ENABLE_PYOSYS is checked before .conf is included. In order to work fine, block setting variables needs to be moved below include Makefile.conf, but then overriding those variables (PYTHON_EXECUTABLE for example) could be possible only from command line not from conf as well.

cliffordwolf:
I think the solution is maybe to include `Makefile.conf` multiple times? Idk..

mmicko:
That will work, anyway it can be only loaded one more time inside ENABLE_PYOSYS block after all variables are assigned, that could be cleanest way of doing it

cliffordwolf:
Including `Makefile.conf` multiple times may also be a good solution for #478.

mmicko:
Option is also to include Makefile.conf at very beginning of file and then do conditional assignment (?=) for all variables, that would changed them only if they are not yet assigned. We can discuss this offline.

thasti:
Thanks for fixing this so quickly, my builds are back to normal!

PR #1032, opened 2019-05-23
WIP: Add support for wand / wor net type.

thasti:
I'm working on synthesis of generated verilog code, which makes quite some use of the wor-nettype in conjunction with generate loops. As there's no easy way for me to change the code generator, I felt like it could be worth trying to add some support to yosys.

Basically, I came up with a possible implementation that solves my use case and wanted to get some feedback, on possible improvements or objections anyone might have.

The gist:
- The Verilog frontend adds a flag for nets that are wand/wor to their AstNodes when it encounters them.
- The genrtlil-pass only creates simple wires, which are tagged with a corresponding wand/wor attribute.
- The `hierarchy` pass
  - transforms these special wires into a `$reduce_or`/`$reduce_and` cell and connects its output to the wire created before
  - adds one input to the `$reduce_*`-cell for assignments to the nets in a module
  - adds one input to the `$reduce_*`-cell for connections to other cells in a module, when they are outputs

Current limitations:
- I only added support for 1-bit wide wires of type wand/wor

Thank you for any feedback.


cliffordwolf:
This currently breaks `make test`. I don't thing it makes much sense to discuss it before this is fixed.

thasti:
Sorry for that - fixed the issue!

frontends/ast/genrtlil.cc
```diff
@@ -1511,14 +1573,38 @@ RTLIL::SigSpec AstNode::genRTLIL(int width_hint, bool sign_hint)
 				}
 				if (child->type == AST_ARGUMENT) {
 					RTLIL::SigSpec sig;
-					if (child->children.size() > 0)
+					RTLIL::SigSpec new_sig;
+					if (child->children.size() > 0) {
 						sig = child->children[0]->genRTLIL();
+						for (int i = 0; i < GetSize(sig); i++) {
+							if (sig[i].wire) {
```
cliffordwolf: This assumes that all module ports are outputs (drivers), which of course isn't true. This raises the question: Maybe it would be better to simply tag all wor/wand wires with a `wor`/`wand` attribute, and insert the OR/AND cells in `hierarchy`, where we have the information which cell port is an input or output.

thasti: Correct. Am I correct to assume that at this step (AST->RTLIL) there's no way to get the direction information? It seemed like that to me and that's why I came up with this simplifying assumption in the first place.

I'd be happy to move it into `hierarchy`, in case it would be an acceptable departure from the separation of 'verilog implementation details' from the RTLIL design.

cliffordwolf: Yes, it's impossible to make the input/output information available at this step because it's possible that the verilog code for the module instantiated is being read later.

It's acceptable to have stuff like that in `hierarchy` when it's necessary to put it there. We already have similar things in `hierarchy` to deal with positional arguments in module instantiations (the cell ports are initially named $1, $2, $3, ... and so forth and then renamed in `hierarchy`) and array instantiations (for which we need to know the width of cell ports, which also isn't available in AST world).

thasti:
As suggested, I moved the resolution of wand/wor nets to the hierarchy pass and simply add an attribute in the RTLIL generation stage.

passes/hierarchy/hierarchy.cc
```diff
@@ -941,6 +941,61 @@ struct HierarchyPass : public Pass {
...
+						reduce_in.append(new_reduce_input);
+						reduce->setPort("\\A", reduce_in);
+						reduce->fixup_parameters();
+						sig[i] = new_reduce_input;
+					}
+				}
+				new_connections.push_back(SigSig(sig, conn.second));
+			}
+			module->new_connections(new_connections);
+		}
```
thasti: Is there a better way of doing this than replacing the full connections vector? Couldn't come up with one...

cliffordwolf: Yes, that's the best way to do this.

cliffordwolf:
I've refactored this a bit in #1050. Please check if that solves your use-case. (Note that #1050 also works with wand/wor signals that are wider than 1 bit.) Closing here now.

PR #1035, opened 2019-05-23
opt_rmdff to work on $dffe and $_DFFE_*

### LINKED ISSUE #1034: Constant unpacked reg creates constant-value flop in synth_xilinx

Following on from #1033:

`./yosys -p "synth_xilinx -nodram" dram.v -DUNPACKED`

```
module dram(input [3:0] i, output [4:0] o);
`ifdef UNPACKED
reg val [0:0];
`else
reg [0:0] val;
`endif
initial val[0] = 1'h1;
assign o = i + val[0];
endmodule
```

## Expected behavior

```
   Number of cells:                 10
     LUT1                            1
     MUXCY                           4
     XORCY                           5
```
as is the case without `-DUNPACKED`

## Actual behavior

```
   Number of cells:                 12
     FDRE                            1
     LUT2                            2
     MUXCY                           4
     XORCY                           5
```
where:
```
  FDRE #(
    .INIT(1'h1)
  ) \$auto$simplemap.cc:442:simplemap_dffe$34  (
    .C(1'h0),
    .CE(1'h0),
    .D(\val[0] ),
    .Q(\val[0] ),
    .R(1'h0)
  );
```

eddiehung:
Fixes #1034.

In particular, flops with constant enables that are never high (according to enable polarity) will get optimised away to their INIT value (or if one doesn't exist, to nothing at all).

### DIFF
--- a/passes/opt/opt_rmdff.cc
+++ b/passes/opt/opt_rmdff.cc
@@ -364,39 +364,60 @@ bool handle_dff(RTLIL::Module *mod, RTLIL::Cell *dff)
 		}
 	}
 
+	// If clock is driven by a constant and (i) no reset signal
+	//                                      (ii) Q has no initial value
+	//                                      (iii) initial value is same as reset value
 	if (!sig_c.empty() && sig_c.is_fully_const() && (!sig_r.size() || !has_init || val_init == val_rv)) {
 		if (val_rv.bits.size() == 0)
 			val_rv = val_init;
+		// Q is permanently reset value or initial value
 		mod->connect(sig_q, val_rv);
 		goto delete_dff;
 	}
 
+	// If D is fully undefined and reset signal present and (i) Q has no initial value
+	//                                                     (ii) initial value is same as reset value
 	if (sig_d.is_fully_undef() && sig_r.size() && (!has_init || val_init == val_rv)) {
+		// Q is permanently reset value
 		mod->connect(sig_q, val_rv);
 		goto delete_dff;
 	}
 
+	// If D is fully undefined and no reset signal and Q has an initial value
 	if (sig_d.is_fully_undef() && !sig_r.size() && has_init) {
+		// Q is permanently initial value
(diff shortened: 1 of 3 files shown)


cliffordwolf:
For the most part LGTM.

> or if one doesn't exist, to nothing at all

I think it should still set the output to `val_init`, even if `has_init` is false. Otherwise one might get warnings for undriven used signals.

> In particular, flops with constant enables that are never high [..]

Arguably that's not far enough. This PR is certainly an improvement over the current state, but the following obvious optimization are still left on the table:

- flops with constant enables that are always high should be "downgraded" to flops without enable

- many of the optimizations in handle_dff are still valid for a FF with enable. For example, if D = initval = resetval then it's save to to simply set Q to that value, even when it's a FF with enable.

I'm equally fine with just fixing the `val_init` thing and leaving the other improvements for a later PR, or further improving this PR, whatever you prefer.

passes/opt/opt_rmdff.cc
```diff
@@ -320,6 +338,16 @@ bool handle_dff(RTLIL::Module *mod, RTLIL::Cell *dff)
 		val_init.bits.push_back(bit.wire == NULL ? bit.data : RTLIL::State::Sx);
 	}
 
+	if (sig_e.size()) {
+		if (!sig_e.is_fully_const())
+			return false;
+		if (sig_e != val_ep) {
+			if (has_init)
+				mod->connect(sig_q, val_init);
```
cliffordwolf: It should even set the output to `val_init` if `has_init` is false. Otherwise one might get warnings for undriven used signals.

eddiehung:
> many of the optimizations in handle_dff are still valid for a FF with enable. For example, if D = initval = resetval then it's save to to simply set Q to that value, even when it's a FF with enable.

Ugh. Now I have to use my brain to figure out what optimisations are still relevant...

cliffordwolf:
LGTM. Merging.

PR #1038, opened 2019-05-24
Link sdf component when building Verific

arjenroodselaar:
I received new eval binaries from Verific, which seem to introduce a new dependency between the verilog and sdf component. Link error here for good measure: https://gist.github.com/arjenroodselaar/aaf03d3f2577c273a91ba104a5086035

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -400,7 +400,7 @@ endif
 
 ifeq ($(ENABLE_VERIFIC),1)
 VERIFIC_DIR ?= /usr/local/src/verific_lib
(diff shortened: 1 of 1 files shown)


cliffordwolf:
We will soon add an `#ifdef #error` block to `verific.cc` that makes sure the version of verific is the one we actually support. This might fix the build with the eval binary of verific you might have, but it will break the build for everyone else who is using the version of verific we actually support.

Please contact http://symbioticeda.com if you need pre-built Yosys binaries with Verific, or support for Yosys+Verific.

arjenroodselaar:
Totally fair. I will contact http://symbioticeda.com for binaries.

PR #1046, opened 2019-05-28
Optimizing DFFs whose initial value prevents their value from changing

bogdanvuk:
I'm pretty new to the Yosys codebase so this is only a proof of concept implementation which invokes SAT solver to optimize-away DFFs in the cases that boil down:

````
module top(
    input  clk,
    input  a,
    output b
);
  
    reg b_reg;
    initial begin
        b_reg <= 0;
    end

    assign b = b_reg;
    always @(posedge clk) begin
        b_reg <= a && b_reg;
    end
endmodule
````

Here, the b_reg initial value of 0 is preventing the register value from ever changing. I used SAT solver to check for these situations since expressions for the DFF next value can be more complex and end up in this simple form only after minimization.

I invoked the SAT solver using string command via Pass::call, since it seemed that I would need to either copy or refactor a lot of the code from the ``passes/sat/sat.cc`` module. The generated SAT problem for the example above is:

 ``sat -ignore_unknown_cells -prove $0\b_reg[0:0][0] 0 -set b_reg[0] 0``

So for the multi-bit registers, the above check would be run for each of the individual bits. If the SAT solver proves that a portion of the register cannot change its value, that portion of the register is replaced by a constant.

Again, I'm not familiar with the codebase so I'll be willing to put in some more work into the implementation if this approach makes sense and I could be given some guidelines on how to fit it naturally into Yosys.

### CODE REVIEW

**Comment by cliffordwolf** [comment]
This breaks `make test`. Please investigate and fix.

**Comment by cliffordwolf** [comment]
Hey. Had a quick look now.

> I'm pretty new to the Yosys codebase

Welcome!

> so this is only a proof of concept implementation

Simply calling the `sat` command is an okay thing to do in a proof-of concept. However, for this to be something that can be upstreamed you should actually create an instance of SatGen that you can reuse for muliple bits, and that only contains the part of the module that is relevant to the question you are asking.

**Comment by bogdanvuk** [comment]
OK, the code now creates an instance of SatGen. For each register, it only includes the cells that affect the register input combinatorially (i.e. its combinatorial cone). I refactored some of the procedures for traversing the netlist into passes/opt/netlist.h file, since they might be useful in different contexts.

I'm only not sure how to reuse SatGen for different bits of the multi-bit register. Even though the bits have the same combinatorial cone, they produce different SAT assumptions (to test whether their values can change). These assumptions seem to linger in SatGen instance after running solve. Is there a way to remove only these assumptions, and reuse the combinatorial cone?

**Comment by bogdanvuk** [comment]
Also, what is a procedure to create an automated test that would check (given a Verilog test file) whether the registers are indeed correctly removed by this optimization?

**Comment by cliffordwolf** [comment]
> Also, what is a procedure to create an automated test that would check (given a Verilog test file) whether the registers are indeed correctly removed by this optimization?

See for example `tests/various/pmux2shiftx.{ys,v}`. We use `select -assert-count <N> t:<T>` to make sure the results contain exactly `<N>` cells of type `<T>`.

File: passes/opt/opt_rmdff.cc [hunk 1]
```diff
@@ -452,6 +459,71 @@ bool handle_dff(RTLIL::Module *mod, RTLIL::Cell *dff)
...
+		for (int position = 0; position < GetSize(sig_d); position += 1) {
+			RTLIL::SigBit q_sigbit = sig_q[position];
+			RTLIL::SigBit d_sigbit = sig_d[position];
+
+			if ((!q_sigbit.wire) || (!d_sigbit.wire)) {
+				continue;
+			}
+
+			ezSatPtr ez;
+			SatGen satgen(ez.get(), &net.sigmap);
```
cliffordwolf: This creates a new SAT instance for each bit of the FF. Generally it is better to re-use SAT instances, at least for the cases where we have reason to assume that the input cone will have large overlap with the previous cone. Therefore we should at least re-use the same sat instance for all bits of the same FF.

cliffordwolf:
I've now refactored this according to my comments above. See 2454ad99bf49afe752d6fd1c1567f59ee9e26736.

You should be able to pull my changes into your branch with this:

```
git pull https://github.com/YosysHQ/yosys.git clifford/ext1046
```

bogdanvuk:
I've pulled the changes you made.

Should I add a sentence about what "-sat" option does in "opt_rmdff" command?

Should I make "-sat" option available via "opt" command?

I have a slightly larger example where this procedure should detect and remove FFs. It consists of 3 different modules with lots of logic that should be optimized away, which makes it more convoluted. Should I add it to the test suite?

cliffordwolf:
> Should I add a sentence about what "-sat" option does in "opt_rmdff" command?
> Should I make "-sat" option available via "opt" command?

Yes, that would be great.

> I have a slightly larger example where this procedure should detect and remove FFs. It consists of 3 different modules with lots of logic that should be optimized away, which makes it more convoluted. Should I add it to the test suite?

Yes, but don't add it to `tests/` within Yosys, add it to https://github.com/YosysHQ/yosys-tests.

bogdanvuk:
I created a pull request for the larger test here: https://github.com/YosysHQ/yosys-tests/pull/41

Using this test I discovered that the code had an issue when there are multiple DFF cells in the design. The test occasionally produced an exception, so I checked it with Valgrind, which in turn helped me find out that there were dangling references to the removed DFF cells left in bit2driver dict. This commit removes the issue: https://github.com/YosysHQ/yosys/pull/1046/commits/35fa7b30574244e4f99373f2a790f004b4a1dbbb

I also added the help text and possibility to pass "-sat" switch via "opt" command

PR #1050, opened 2019-05-28
Refactored wand/wor support

cliffordwolf:
Supersedes #1032.


thasti:
Thank you for refactoring this, looks quite a bit cleaner! It also works like a charm in my application.

PR #1051, opened 2019-05-28
Make PREFIX used after .conf include

mmicko:
Proposed solution for issue #478

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,10 +46,6 @@ OS := $(shell uname -s)
 PREFIX ?= /usr/local
 INSTALL_SUDO :=
 
-BINDIR := $(PREFIX)/bin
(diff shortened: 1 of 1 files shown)


cliffordwolf:
That wouldn't work afaict (haven't actually tested it), bacaus of this:  
https://github.com/YosysHQ/yosys/pull/1051/files#diff-b67911656ef5d18c4ae36cb6741b7965R68

```
LDFLAGS := $(LDFLAGS) -L$(LIBDIR)
```

You are moving LIBDIR after this line, and because its using `:=` it means `$(LIBDIR)` is being evaluated right there.

mmicko:
Ah you are right, missed that one, will investigate other options

Issue #1058, opened 2019-05-31
Add support for attributes 


mithro:
## Steps to reproduce the issue

The Verilog-2005 standard says;
> An attribute_instance can appear in the Verilog description as a prefix attached to a declaration, a module item, a statement, or a port connection. It can appear as a suffix to an operator or a Verilog function name in an expression.

This means for Yosys to be compatible with Verilog-2005 it should support attributes on the following (ones marked with `x` are already known to work);
- [x] Module declaration: `(* optimize *) module control ( ... );`
- [x] Port declaration (input/output/inout): `(* port_attr *) input clk;`
- [x] Parameter: `(* param_attr *) parameter WIDTH = 10;`
- [x] Local parameter: `(* localparam_attr *) localparam MAGIC = 8'hFE;`
- [x] Net/Variable (wire/reg): `(* state_variable *) reg [7:0] S;`
- [ ] Port connection: `dff il (q, d, (* clock_line *) ckl, rst);`
- [x] Suffix to an operator: `sum = a + (* ripple_adder *) b;`
- [x] Suffix to a function call: `assign next_state = fsm_func(* meally_fsm *) (state, a, b, c);`
- [x] Module instantiation: `(* inst_attr *) my_module my_instance (...);`

These examples are from [Verilog — 2001 - A Guide to the New Features of the Verilog® Hardware Description Language](https://link.springer.com/book/10.1007/978-1-4615-1713-9) Section [`25. Attributes`](https://link.springer.com/chapter/10.1007%2F978-1-4615-1713-9_27),

> ```verilog
> always @(state) //state machine with one-hot encoding
>   (* full_case, parallel_case *) case (state)
>   3'b001: next_state = 3'b010;
>   3'b010: next_state = 3'b100;
>   3'b100: next_state = 3'b001;
> endcase
> ```
>
> The attribute is specified as a prefix to a declaration, a module item, a statement or
> a port connection. Three examples are:
> ```verilog
> (* optimize *) module control ( ... );
>
> (* state_variable *) reg [7:0] S;
>
> dff il (q, d, (* clock_line *) ckl, rst};
> ```
>
> > The attribute is specified as a suffix to an operator or the function name in a function call.
> ```verilog
> sum = a + (* ripple_adder *) b;
>
> assign next_state = fsm_func(* meally_fsm *) (state, a, b, c);
> ```

[...]

Issue #1009, opened 2019-05-14
Attributes on parameters / localparams are not parsed correctly per Verilog 2005 spec


litghost:
## Steps to reproduce the issue

1. Run "yosys [example.v](https://gist.github.com/litghost/2f4201d0983c0d2736de5c2cd5e47e72)"

## Expected behavior

Yosys parses the file per the [Verilog 2005 spec](http://staff.ustc.edu.cn/~songch/download/IEEE.1364-2005.pdf):
```
module_or_generate_item ::=
 { attribute_instance } module_or_generate_item_declaration
| { attribute_instance } local_parameter_declaration ;
| { attribute_instance } parameter_override
| { attribute_instance } continuous_assign
| { attribute_instance } gate_instantiation
| { attribute_instance } udp_instantiation
| { attribute_instance } module_instantiation
| { attribute_instance } initial_construct
| { attribute_instance } always_construct
| { attribute_instance } loop_generate_construct
| { attribute_instance } conditional_generate_construct
non_port_module_item ::=
 module_or_generate_item
| generate_region
| specify_block
| { attribute_instance } parameter_declaration ;
| { attribute_instance } specparam_declaration
```

The Verilog 2001 specification also indicates that attributes are supported on parameter declarations:
```
non_port_module_item ::=
 { attribute_instance } generated_instantiation
| { attribute_instance } local_parameter_declaration
| { attribute_instance } module_or_generate_item
| { attribute_instance } parameter_declaration
| { attribute_instance } specify_block
| { attribute_instance } specparam_declaration 
```

## Actual behavior

```
1. Executing Verilog-2005 frontend.
Parsing Verilog input from `example.v' to AST representation.
example.v:3: ERROR: syntax error, unexpected TOK_PARAMETER
```

mkurc-ant:
I'd like to work on that issue as it is very relevant to what I am doing. I need support for parameters with attributes.

mkurc-ant:
I think that this issue depends to #1001.

cliffordwolf:
> I need support for parameters with attributes.

Do you just need the parser to accept the parameter with attribute syntax, or do you actually want to see the parameter in some way in the RTLIL output? Because the former is easy but the latter might be a bit of a challenge..

mkurc-ant:
Hi, I need the latter. I am using Yosys to convert Verilog to JSON representation. And I need to have both parameters on modules and their attributes exported to JSON.

[...]

PR #1060, opened 2019-05-31
Added support for parsing attributes on port connections.

mkurc-ant:
This PR adds support for attributes on port connections to the Verilog frontend. Attributes are allowed but discarded.

### DIFF
--- a/frontends/verilog/verilog_parser.y
+++ b/frontends/verilog/verilog_parser.y
@@ -1535,11 +1535,13 @@ cell_port:
 	attr {
 		AstNode *node = new AstNode(AST_ARGUMENT);
 		astbuf2->children.push_back(node);
+		free_attr($1);
 	} |
 	attr expr {
 		AstNode *node = new AstNode(AST_ARGUMENT);
 		astbuf2->children.push_back(node);
 		node->children.push_back($2);
+		free_attr($1);
 	} |
 	attr '.' TOK_ID '(' expr ')' {
 		AstNode *node = new AstNode(AST_ARGUMENT);
@@ -1547,12 +1549,14 @@ cell_port:
 		astbuf2->children.push_back(node);
 		node->children.push_back($5);
 		delete $3;
+		free_attr($1);
 	} |
 	attr '.' TOK_ID '(' ')' {
 		AstNode *node = new AstNode(AST_ARGUMENT);
 		node->str = *$3;
 		astbuf2->children.push_back(node);
 		delete $3;
+		free_attr($1);
 	};
 
 always_stmt:
--- a/frontends/verilog/verilog_parser.y
+++ b/frontends/verilog/verilog_parser.y
@@ -1532,27 +1532,27 @@ cell_port_list_rules:
 	cell_port | cell_port_list_rules ',' cell_port;
 
 cell_port:
-	/* empty */ {
+	attr {
 		AstNode *node = new AstNode(AST_ARGUMENT);
 		astbuf2->children.push_back(node);
 	} |
-	expr {
+	attr expr {
 		AstNode *node = new AstNode(AST_ARGUMENT);
 		astbuf2->children.push_back(node);
-		node->children.push_back($1);
+		node->children.push_back($2);
 	} |
-	'.' TOK_ID '(' expr ')' {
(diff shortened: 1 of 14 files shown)


eddiehung:
I don't see any issue with attributes on ports, since a port is an `RTLIL::Wire*`, and has an attributes dictionary already so it just needs populating.
The problem with #1009 was that parameters are not representing in the RTLIL, however, we now have an idea of how to resolve that.

Can you add some tests too, please?

mkurc-ant:
But I do not mean attributes on port declarations which indeed are wires. I mean on port connections, like that:

```
cell cell_instance (clk, rst, (* this_is_an_attribute *) inp, out);
```

Which of the test folders would be best suited for a Verilog parsing test case ?

eddiehung:
> But I do not mean attributes on port declarations which indeed are wires. I mean on port connections, like that:
Oops sorry. I misread your issue (again). I will change my answer to: then this would require a deeper rethink.

> Which of the test folders would be best suited for a Verilog parsing test case ?

`/tests/simple` might be a good place to add a test for this as you did in #1013

mithro:
@mkurc-ant Please do add tests. The examples from https://github.com/YosysHQ/yosys/issues/1058

mkurc-ant:
I added tests for attributes in various places to `/tests/simple`, but I encountered a problem. It seems that Icarus Verilog, which is used in those tests as well, does not support attributes on port connections and on function calls. For example for attribute on port connection it throws a syntax error:

```
+ false
+ compile_and_run attrib05_port_conn_tb_ref attrib05_port_conn_out_ref attrib05_port_conn_tb.v attrib05_port_conn_ref.v /home/mkurc/Repos/google-prjxray-yosys/tests/tools/../../techlibs/common/simlib.v /home/mkurc/Repos/google-prjxray-yosys/tests/tools/../../techlibs/common/simcells.v
+ exe=attrib05_port_conn_tb_ref
+ output=attrib05_port_conn_out_ref
+ shift 2
+ false
+ false
+ iverilog -g2005 '-Doutfile="attrib05_port_conn_out_ref"' -s testbench -o attrib05_port_conn_tb_ref attrib05_port_conn_tb.v attrib05_port_conn_ref.v /home/mkurc/Repos/google-prjxray-yosys/tests/tools/../../techlibs/common/simlib.v /home/mkurc/Repos/google-prjxray-yosys/tests/tools/../../techlibs/common/simcells.v
attrib05_port_conn_ref.v:19: syntax error
attrib05_port_conn_ref.v:19: error: Syntax error in instance port expression(s).
```

So for now I disabled `attrib05_port_conn.v` and `attrib07_func_call.v` for now. But now this is pointless as nothing gets tested.

Does it make sense to additional "test suite" for example `/tests/parse`? It would hold test cases to be used for parsing by Yosys only.

eddiehung:
In that case, I would suggest you add all tests/failing tests to `tests/various` -- where `tests/simple` generates a tb that and then checks that the pre- and post- synth designs behaves identically, `tests/various` runs a `*.ys` script, in which you can choose to read the design, and/or check that those attributes are indeed captured using a `select` command.

mithro:
@mkurc-ant We should log an issue with Icarus and look at adding support in that tool too.

mkurc-ant:
Ok, I'll modify tests and see to logging the issue upstream to the Icarus Verilog dev(s).

mkurc-ant:
I moved those tests which currently fail with Icarus to `/tests/various` folder. I added Yosys scripts which just parse the input Verilog. Since this PR is only about parsing there is no point of any further check.

frontends/verilog/verilog_parser.y
```diff
@@ -1532,27 +1532,27 @@ cell_port_list_rules:
 	cell_port | cell_port_list_rules ',' cell_port;
 
 cell_port:
-	/* empty */ {
+	attr {
```
cliffordwolf: You are leaking the `std::map` pointer returned by attr. Please add `free_attr($1);` to all the cases where add an instance of `attr` that you just want ignored.

mkurc-ant: You're right. I pushed a fix.

PR #1068, opened 2019-06-05
SystemVerilog support for implicit named port connections

tux3:
This code was split out and adapted from #560.

### DIFF
--- a/frontends/verilog/verilog_parser.y
+++ b/frontends/verilog/verilog_parser.y
@@ -154,7 +154,7 @@ struct specify_rise_fall {
 %token TOK_INCREMENT TOK_DECREMENT TOK_UNIQUE TOK_PRIORITY
 
 %type <ast> range range_or_multirange  non_opt_range non_opt_multirange range_or_signed_int
-%type <ast> wire_type expr basic_expr concat_list rvalue lvalue lvalue_concat_list
+%type <ast> wire_type expr basic_expr concat_list rvalue lvalue lvalue_concat_list named_port
 %type <string> opt_label opt_sva_label tok_prim_wrapper hierarchical_id
 %type <boolean> opt_signed opt_property unique_case_attr
 %type <al> attr case_attr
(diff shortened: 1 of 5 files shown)


tests/simple/implicit_ports.sv
```diff
@@ -0,0 +1,17 @@
...
+
+module named_ports;
+	wire [7:0] alu_out;
+	wire [7:0] ain, bin;
+	wire [2:0] opcode;
+
+	alu alu (
+		.alu_out(alu_out),
+		.zero(),	// named empty ports are explicitely unconnected
+		.ain,		// implicit connection is equivalent to .ain(ain)
```
cliffordwolf: This test case isn't testing anything. Imagine the code would be handled like `.zero, .ain()` instead, the exact opposite of what should happen. This test would still pass.

tux3: Yes, I think it was meant as just a syntax-level test. I'll try to write a test that actually checks the connections are made correctly.

tux3: @cliffordwolf I couldn't find some other test that checks connections, so I made something up with the sat solver, hope that makes sense.

cliffordwolf: I don't quite understand why use the sat solver here. Shouldn't autotest work with this exact same test case just as well?

tux3: Sorry, I'm new to this, not very familiar with how the different test systems work. I used SAT only because I could figure out how to make it work.
What does autotest do exactly, does it expect SystemVerilog assertions? Is it some automated magic where it compares outputs with iverilog?

I assumed autotest is what runs the cases in the simple/ folder, so I tried is putting this exact same test case there. That doesn't work however, if I intentionally mess up the connection in verilog_parser.y the test case still passes.

What am I missing, is there some reference to writing auto tests I should be reading?

cliffordwolf:
I've now merged this with some of my own cleanups ontop.

> What does autotest do exactly, does it expect SystemVerilog assertions?

autotest simply simulates the original verilog code and the synthesis results with random inputs, and checks if the output is the same.

tux3:
Alright. Either I misunderstood what you where trying to test, or I wasn't clear.
With the code commented out like this:

```
attr '.' TOK_ID {
	AstNode *node = new AstNode(AST_ARGUMENT);
	node->str = *$3;
	astbuf2->children.push_back(node);
	//node->children.push_back(new AstNode(AST_IDENTIFIER));
	//node->children.back()->str = *$3;
	delete $3;
	free_attr($1);
};
```

The tests still pass. If that's intended, then that's fine, but I don't know what we're testing here.

cliffordwolf:
> With the code commented out like this: [..] The tests still pass.

I'll have a look.

cliffordwolf:
The problem is that 3f87cf86ccefe6e66f768fbf19c34db97cf7246d broke autotest.sh for `.sv` files. I would assume to make it work with SV constructs that iverilog doesn't support. But it's not actually testing the SV stuff the way it behaves at the moment. As quick work-around I've now renamed that one test from `.sv` to `.v`.

tux3:
Good to know, thanks for taking a look.

PR #1075, opened 2019-06-06
Add new "muxpack" command for packing chains of $mux cells

eddiehung:
Derived from `shregmap`

~TODO: Packing chains of `$pmux` cells~


passes/opt/muxpack.cc
```diff
@@ -0,0 +1,270 @@
...
+
+			if (cases < 2) {
+				cursor++;
+				continue;
+			}
+
+			Cell *last_cell = chain[cursor+cases-1];
+
+			log("Converting %s.%s ... %s.%s to a pmux with %d cases.\n",
+				log_id(module), log_id(first_cell), log_id(module), log_id(last_cell), cases);
```
cliffordwolf: There's no check anywhere here yet that makes sure the control inputs are exclusive. As it is, this pass incorrectly converts the following test case into a `$pmux`:

```
module test ( 
        input wire x, y, z,
        input wire a, b, c, d,
        output reg o
);
        always @* begin
                o = a;
                if (x) o = b;
                if (y) o = c;
                if (z) o = d;
        end
endmodule
```

passes/opt/muxpack.cc
```diff
@@ -0,0 +1,270 @@
...
+			{
+				for (auto bit : a_sig.bits())
+					if (sigbit_with_non_chain_users.count(bit))
+						goto start_cell;
+
+				Cell *c1 = sig_chain_prev.at(a_sig);
+				Cell *c2 = cell;
+
+				if (c1->getParam("\\WIDTH") != c2->getParam("\\WIDTH"))
+					goto start_cell;
```
cliffordwolf: Shouldn't this already be guaranteed by the fact that c1.Y == c2.A? However, if this is run after freduce then you want to strip zero extension from A/B/Y before creating the database entries and before the lookups. Here's a test case (use with `proc; opt; freduce; opt`):

```
module test ( 
        input wire [1:0] s,
        input wire a, b, c, d,
        output reg [3:0] o
);
        always @* begin
                o = {4{a}};
                if (s == 0) o = {3{b}};
                if (s == 1) o = {2{c}};
                if (s == 2) o = d;
        end
endmodule
```

passes/opt/muxpack.cc
eddiehung: Yes, you're right. I was thinking of a `$pmux` more as a priority mux and not as a one-hot only.
It is concerning that my testing didn't pick this up though:
https://github.com/YosysHQ/yosys/blob/887df8914c64220b9f306b7d21f199fa247224fd/tests/various/muxpack.ys#L1-L15

passes/opt/muxpack.cc
eddiehung: Ooh `freduce` is a funky one. Basically, it replaces redundant wires on the driving cell.`Y` with dummy wires, and on the sink cell's input, it un-redundant-ises those previous bits to get the original signal back.

For example, driving.Y might have been `{a,a,a,a}` which `freduces` transforms into `{dummy1,dummy2,dummy3,a}`. sink.A would then, instead of just receiving `Y`, would recreate it by doing `{Y[0],Y[0],Y[0],Y[0]}`.

Given that we rely on driving.Y == sink.A in order to follow the chain of muxes, wouldn't using `freduce` break this flow?

passes/opt/muxpack.cc
eddiehung: Okay, I've had a stab at creating an `ExclusiveDatabase` that looks through `$eq` cells that compare against a constant, and `$logic_not` cells (which are `$eq` with zero).
For chains of `$pmux` cells it currently suffers from O(N^2) since it has to compare every bit of `S` in one `$pmux` with every bit of `S` on the other...

Currently doesn't handle `$ne` cells for `if (s != 0) if (s != 1) else else` type structures (thus `mux_if_unbal_5_3_invert` test is commented out right now).

eddiehung: More explicitly: it will only handle `$eq` cells -- not `$ne` nor `$ge` or any of those...

cliffordwolf:
LGTM

passes/opt/muxpack.cc
cliffordwolf: Do you have a MCVE that I can use to reproduce the case where freduce prevents muxpack from working as desired?

eddiehung: I used your exact testcase here: https://github.com/YosysHQ/yosys/blob/4ddc0354c1cc61f2e2b3f15cc341fd277c710e89/tests/various/muxpack.v#L170-L181 on script (currently commented out): https://github.com/YosysHQ/yosys/blob/4ddc0354c1cc61f2e2b3f15cc341fd277c710e89/tests/various/muxpack.ys#L184-L199
which gives:
before
which seems to be quite messy to analyse, no?

cliffordwolf: Thanks. I'll have a closer look.

PR #1085, opened 2019-06-10
Improve shregmap to handle case where first flop is common to two chains

eddiehung:
e.g.
```
module shregmap_test(input i, clk, output [1:0] q);
reg head = 1'b0;
reg [3:0] shift1 = 4'b0000;
reg [3:0] shift2 = 4'b0000;

always @(posedge clk) begin
    head <= i;
    shift1 <= {shift1[2:0], head};
    shift2 <= {shift2[2:0], head};
end

assign q = {shift2[3], shift1[3]};
endmodule
```

Previously, `shregmap` would create one `SHREG` only since the key for `sigbit_chain_next` would be clobbered; now I create a new temporary wire (which isn't updated by `SigMap`) to circumvent this clobbering.... a bit hacky -- is there a better way to do this?

EDIT: The alternative would be to change the data structure (multimap) or a new one (vector) capable of containing duplicates, because I really don't like the fact it could break on a sigmap update...

### DIFF
--- a/passes/techmap/shregmap.cc
+++ b/passes/techmap/shregmap.cc
@@ -295,7 +295,18 @@ struct ShregmapWorker
 				{
 					auto r = sigbit_chain_next.insert(std::make_pair(d_bit, cell));
 					if (!r.second) {
+						// Insertion not successful means that d_bit is already
+						// connected to another register, thus mark it as a
+						// non chain user ...
 						sigbit_with_non_chain_users.insert(d_bit);
+						// ... and clone d_bit into another wire, and use that
+						// wire as a different key in the d_bit-to-cell dictionary
+						// so that it can be identified as another chain
+						// (omitting this common flop)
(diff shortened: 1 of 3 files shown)


eddiehung:
Reason for this enhancement: https://github.com/alexforencich/verilog-cam/blob/master/rtl/cam_srl.v

cliffordwolf:
> a bit hacky -- is there a better way to do this?

maybe adding a comment would improve things a bit?

eddiehung:
Actually @cliffordwolf, there might not be any harm to updating sigmap, let's see if e63324f5ef2ebb14fa0cc88544f577406f95b223 passes

cliffordwolf:
It did pass. Not sure what that means in terms of how much testing shregmap got from this tho. ;) If it works, LGTM. Feel free to merge it.

PR #1086, opened 2019-06-11
Fixed broken $error()/$info/$warning() on non-generate blocks (within always/initial blocks)

### LINKED ISSUE #1084: Fails to parse `initial if (0) $error`

`./yosys system.v` on a91ea661; regression from f01a61f (before PR #1077)

```
module top;
initial
   if (0) $error("ASDF");
endmodule
```

## Expected behavior

No parser error, as on f01a61f.

## Actual behavior

```
-- Parsing `system.v' using frontend `verilog' --

1. Executing Verilog-2005 frontend: system.v
Parsing Verilog input from `system.v' to AST representation.
system.v:3: ERROR: syntax error, unexpected TOK_ELAB_TASK
```

udif:
This seems to fix #1084.

It seems I've broken regular $info/$error/$warning tasks, since the lexer always returns TOK_ELAB_TASK for those, so I made the parser accept it for normal tasks as well (and rename it TOK_MSG_TASKS since it no longer applies to elaboration time tasks only).

### DIFF
--- a/frontends/verilog/verilog_lexer.l
+++ b/frontends/verilog/verilog_lexer.l
@@ -313,7 +313,7 @@ supply1 { return TOK_SUPPLY1; }
 
 "$"(info|warning|error|fatal) {
 	frontend_verilog_yylval.string = new std::string(yytext);
-	return TOK_ELAB_TASK;
+	return TOK_MSG_TASKS;
 }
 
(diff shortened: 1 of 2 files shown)


daveshah1:
FYI, the ethernet yosys-bench is broken without this PR (as it uses `$error` in initial). @cliffordwolf can this be merged?

PR #1098, opened 2019-06-14
"abc9" pass for timing-aware techmapping (experimental, FPGA only, no FFs)

eddiehung:


techlibs/ice40/synth_ice40.cc
```diff
@@ -311,7 +331,10 @@ struct SynthIce40Pass : public ScriptPass
 				run("techmap -map +/gate2lut.v -D LUT_WIDTH=4", "(only if -noabc)");
 			}
 			if (!noabc) {
-				run("abc -dress -lut 4", "(skip if -noabc)");
+				if (abc == "abc9")
+					run(abc + stringf(" -dress -lut +/ice40/abc_%s.lut -box +/ice40/abc_%s.box", device_opt.c_str(), device_opt.c_str()), "(skip if -noabc)");
```
daveshah1: `-dress` has been removed from `abc9` now, so this is failing. Might also be worth adding a wire delay here (subject to experimentation)

techlibs/ice40/abc_u.box
```diff
@@ -0,0 +1,113 @@
...
+
+# Inputs: C D R
+# Outputs: Q
+SB_DFFSR 3 0 3 1
+- - -
+
+# Inputs: C D R
+# Outputs: Q
+SB_DFFR 4 0 3 1
+- - -
```
daveshah1: The Xilinx equivalent, FDCE, has a path from async clear to out, whereas this doesn't - is there a reason for the difference?

backends/aiger/xaiger.cc
```diff
@@ -0,0 +1,1015 @@
...
+				aiger_encode(f, delta1);
+			}
+		}
+
+		f << "c";
+
+		if (!box_list.empty() || !ff_bits.empty()) {
+			std::stringstream h_buffer;
+			auto write_h_buffer = [&h_buffer](int i32) {
+				// TODO: Don't assume we're on little endian
```
daveshah1: As well as big endian support (for all those Yosys users on OpenRISC :smile: ), we should also have a soft fallback if neither of those compiler intrinsics are available

backends/aiger/xaiger.cc
```diff
@@ -0,0 +1,1015 @@
...
+#ifdef _WIN32
+				int i32_be = _byteswap_ulong(i32);
```
daveshah1: Shouldn't this be int32_t (otherwise this will end up writing 8 bytes later on if on some hypothetical ILP64 operating system)

frontends/aiger/aigerparse.cc
```diff
@@ -131,36 +269,199 @@ void AigerReader::parse_aiger()
...
-	design->add(module);
+	post_process();
+}
+
+static uint32_t parse_xaiger_literal(std::istream &f)
+{
+	uint32_t l;
+	f.read(reinterpret_cast<char*>(&l), sizeof(l));
+	if (f.gcount() != sizeof(l))
+		log_error("Offset %ld: unable to read literal!\n", static_cast<int64_t>(f.tellg()));
```
daveshah1: I think this won't work on a typical 32-bit system which has 32-bit longs (so `%ld` will be expecting a 32-bit value)

frontends/aiger/aigerparse.cc
```diff
@@ -131,36 +269,199 @@ void AigerReader::parse_aiger()
...
+	RTLIL::Wire* n0 = module->wire("\\__0__");
+	if (n0)
+		module->connect(n0, RTLIL::S0);
+
+	dict<int,IdString> box_lookup;
+	for (auto m : design->modules()) {
+		auto it = m->attributes.find("\\abc_box_id");
+		if (it == m->attributes.end())
+			continue;
+		if (m->name[0] == '$') continue;
```
daveshah1: I'm guessing this is an attempt to avoid using a `$paramod` accidentally? Might be better to use `$paramod` rather than just `$` here, in case we ever want a flow with some Yosys-internal cells as (white|black)boxes

passes/techmap/abc9.cc
```diff
@@ -0,0 +1,1173 @@
...
+		bool show_tempdir = false, sop_mode = false;
+		vector<int> lut_costs;
+		markgroups = false;
+
+#if 0
+		cleanup = false;
+		show_tempdir = true;
+#endif
+
+		map_mux4 = false;
```
daveshah1: Might as well remove these variables as they seem to be unused in ABC9

passes/techmap/abc9.cc
```diff
@@ -0,0 +1,1173 @@
...
+// http://www.eecs.berkeley.edu/~alanmi/abc/
+
+#if 0
+// Based on &flow3 - better QoR but more experimental
+#define ABC_COMMAND_LUT "&st; &ps -l; "/*"&sweep -v;"*/" &scorr; " \
+						"&st; &if {W}; &save; &st; &syn2; &if {W}; &save; &load; "\
+						"&st; &if -g -K 6; &dch -f; &if {W}; &save; &load; "\
+						"&st; &if -g -K 6; &synch2; &if {W}; &save; &load"
+#else
+#define ABC_COMMAND_LUT "&st; &sweep; &scorr; "/*"&dc2; "*/"&retime; &dch -f; &ps -l; &if {W} -v; "/*"&mfs; "*/"&ps -l"
```
daveshah1: For consideration: should we be passing the delay target, if given, to `&if`, to allow users to explore delay/area tradeoffs?

techlibs/ice40/abc_u.box
eddiehung: No intentional difference. I guess there wasn't an async clear that delay handy in https://github.com/cliffordwolf/icestorm/blob/be0bca0230d6fe1102e0a360b953fbb0d273a39f/icefuzz/timings_hx1k.txt

Also, not sure we/ABC can model async clears on flops...

techlibs/ice40/synth_ice40.cc
eddiehung: Good spot!

techlibs/ice40/abc_u.box
daveshah1: If you want the timing, it's the sr to lcout IOPATH.

I know abc9 has some internal modelling of set/resets, but I think that's for sync set/reset using the complex flop pattern - I can't see how you'd model an async one this way, using a whitebox and an AIG flipflop

backends/aiger/xaiger.cc
eddiehung: Just added big endian support, and by using https://stackoverflow.com/a/46137633 I've covered Windows/Apple/Linux which should be good enough for now

techlibs/ice40/abc_u.box
eddiehung: My understanding is that it can't be modelled using ABC9 either. I've removed this async timing from the Xilinx box.

frontends/aiger/aigerparse.cc
```diff
@@ -22,19 +22,183 @@
 // Armin Biere. The AIGER And-Inverter Graph (AIG) Format Version 20071012. Technical Report 07/1, October 2011, FMV Reports Series, Institute for Formal Models and Verification, Johannes Kepler University, Altenbergerstr. 69, 4040 Linz, Austria.
 // http://fmv.jku.at/papers/Biere-FMV-TR-07-1.pdf
 
-#ifndef _WIN32
+#ifdef _WIN32
 #include <libgen.h>
```
cliffordwolf: As discussed in call: ?

kernel/rtlil.cc
```diff
@@ -1565,13 +1565,21 @@ void RTLIL::Module::remove(const pool<RTLIL::Wire*> &wires)
 
 void RTLIL::Module::remove(RTLIL::Cell *cell)
 {
+	auto it = cells_.find(cell->name);
+	log_assert(it != cells_.end());
+	remove(it);
+}
+
+dict<RTLIL::IdString, RTLIL::Cell*>::iterator RTLIL::Module::remove(dict<RTLIL::IdString, RTLIL::Cell*>::iterator it)
```
eddiehung: @cliffordwolf says `dict` and `pool` containers allow erasure during iteration meaning we don't need this...

backends/aiger/xaiger.cc
```diff
@@ -0,0 +1,849 @@
...
+				RTLIL::IdString wire_name = wire->name.str() + "$inout.out";
+				RTLIL::Wire *new_wire = module->wire(wire_name);
+				if (!new_wire)
+					new_wire = module->addWire(wire_name, GetSize(wire));
+				SigBit new_bit(new_wire, bit.offset);
+				module->connect(new_bit, bit);
+				if (not_map.count(bit))
+					not_map[new_bit] = not_map.at(bit);
+				else if (and_map.count(bit)) {
+				    //and_map[new_bit] = and_map.at(bit); // Breaks gcc-4.8
```
daveshah1: We determined this was UB rather than a bug in older, right? Probably best to remove it and rely only on the below, working implementation using insert

backends/aiger/xaiger.cc
```diff
@@ -0,0 +1,849 @@
...
+				    and_map.insert(std::make_pair(new_bit, and_map.at(bit)));
+				}
+				else if (alias_map.count(bit))
+					alias_map[new_bit] = alias_map.at(bit);
```
daveshah1: Will this have the same problem as above?

cliffordwolf: Yes.

backends/aiger/xaiger.cc
```diff
@@ -0,0 +1,849 @@
...
+ *  MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ *  ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ *  WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ *  ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ *  OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *
+ */
+
+// https://stackoverflow.com/a/46137633
+#ifdef _MSC_VER
```
daveshah1: Can we have a generic fallback to ensure it doesn't break the build if someone wants to build Yosys on a really obscure platform? I can't imagine endian swapping is going to make or break performance

techlibs/ice40/synth_ice40.cc
```diff
@@ -37,6 +37,10 @@ struct SynthIce40Pass : public ScriptPass
 		log("\n");
 		log("This command runs synthesis for iCE40 FPGAs.\n");
 		log("\n");
+		log("    -device < hx | lp | u >\n");
```
daveshah1: Perhaps we should mention this is only for abc9 flows? Otherwise the lack of a difference with the default flow might be confusing.

gsomlo:
I wanted to test drive this by applying it on top of current master (74945dd at the time of this writing), but trying to rebase against master failed with a merge conflict that I'm not qualified to resolve. Any chance at a rebase and force-push against master? -- Thanks!

techlibs/ecp5/abc_5g.box
```diff
@@ -0,0 +1,43 @@
...
+# name  ID   w/b   ins    outs
+CCU2C   1      1   9      3
+
+#A0   A1   B0   B1   C0    C1  D0   D1   CIN
+379  -    379  -    275   -    141  -    257
+630  379  630  379  526   275  392  141  273
+516  516  516  516  412   412  278  278  43
+
+# Box 2 : TRELLIS_DPR16X4 (16x4 dist ram)
+# Outputs: DO0, DO1, DO2, DO3, DO4
```
eddiehung: TODO: Remove DO4

techlibs/ecp5/abc_5g.box
```diff
@@ -0,0 +1,43 @@
...
+# name            ID  w/b   ins   outs
+TRELLIS_DPR16X4   2     0   14    4
+
+#DI0   DI1   DI2   DI3   RAD0   RAD1   RAD2   RAD3   WAD0    WAD1   WAD2   WAD3  WCK   WRE
+-      -     -     -     141    379    275    379    -       -      -      -     -     -
```
eddiehung: @daveshah1 Is there a reason these are not in ascending or descending order?

techlibs/ecp5/abc_5g.box
eddiehung: TODO: Add "WAD" to `abc_scc_break`

daveshah1: Oops, DO4 Is just a typo in the comment

techlibs/ecp5/abc_5g.box
daveshah1: Yes, the ECP5 distributed RAM has some slightly crazy wiring: https://github.com/YosysHQ/nextpnr/blob/master/ecp5/cells.cc#L355-L370

PR #1100, opened 2019-06-17
Support ~ in filename parsing

bwidawsk:
Untested on non-Linux platforms...

Signed-off-by: Ben Widawsky <ben@bwidawsk.net>

### DIFF
--- a/kernel/yosys.cc
+++ b/kernel/yosys.cc
@@ -651,6 +651,10 @@ void rewrite_filename(std::string &filename)
 		filename = filename.substr(1, GetSize(filename)-2);
 	if (filename.substr(0, 2) == "+/")
 		filename = proc_share_dirname() + filename.substr(2);
+#ifndef _WIN32
+	if (filename.substr(0, 2) == "~/")
+		filename = filename.replace(0, 1, getenv("HOME"));
+#endif
 }
 
 #ifdef YOSYS_ENABLE_TCL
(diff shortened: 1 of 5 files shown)


kernel/yosys.cc
```diff
@@ -651,6 +651,8 @@ void rewrite_filename(std::string &filename)
 		filename = filename.substr(1, GetSize(filename)-2);
 	if (filename.substr(0, 2) == "+/")
 		filename = proc_share_dirname() + filename.substr(2);
+	if (filename.substr(0, 1) == "~")
+		filename = filename.replace(0, 1, getenv("HOME"));
```
cliffordwolf: The convention is that `~/` expands to `$HOME` and `~<user>/` expands to the home directory of the given user. Either you support only the first case, then the match should be for `~/`, not just `~`, or you support the latter case as well but then you'd have to actoally parse whatever is between the `~` and the first `/` and look up that users home directory (using `getpwnam()` or `getpwnam_r()`).

eddiehung: Also it might be worth wrapping it around `#ifdef _WIN32`

bwidawsk: @cliffordwolf Let's start with the first case... and then maybe someone else who cares about the second case can do it.

@eddiehung Good idea (but I think you mean ifndef).

I'll fix it up a bit later.

udif:
I spent 15 minutes of my life to see how ~username is done in bash.

It boils down to:

#include <sys/types.h>
#include <pwd.h>
struct passwd *user_entry;
user_entry = getpwnam (username);
// use user_entry->pw_dir to get user's home dir.

PR #1104, opened 2019-06-19
Clarify switch/case semantics in RTLIL

whitequark:


kernel/rtlil.cc
```diff
@@ -1381,7 +1381,34 @@ void RTLIL::Module::check()
 	for (auto &it : processes) {
 		log_assert(it.first == it.second->name);
 		log_assert(!it.first.empty());
-		// FIXME: More checks here..
```
whitequark: I think I check every RTLIL process related invariant, so I've removed the FIXME.

PR #1106, opened 2019-06-19
Parameters as wires

mkurc-ant:
This PR allows to store attributes of a Verilog module parameter in a wire object. For each parameter a mocking wire is created with the same name. Those mocking wires are connected to constants which correspond to default values of parameters. Parameters with their default values are outputted via Ilang, Verilog and JSON backends.

The PR also adds a small test suite which compares various backend outputs with a golden reference.


kernel/rtlil.cc
```diff
@@ -2373,6 +2382,24 @@ std::map<unsigned int, RTLIL::Wire*> *RTLIL::Wire::get_all_wires(void)
...
+bool RTLIL::Wire::is_localparam() const {
+	if (attributes.count("\\is_localparam")) {
+		return attributes.at("\\is_localparam").as_bool();
+	}
+	return false;
+}
+
+bool RTLIL::Wire::is_mockup() const {
+	return is_parameter() || is_localparam();
+}
```
cliffordwolf: There is no need to add any of those three functions or any of the many uses you add for them. I have no idea why you think you need to add hundreds of lines of code all over the place for this. The reason we have suggested the use of an attribute is so that this can be accomplished with a small non-intrusive change.

cliffordwolf:
See #1109, which adds <50 lines instead of >1100.

mithro:
```
# git diff upstream/master --stat
...
 34 files changed, 1154 insertions(+), 211 deletions(-)

# git diff upstream/master --stat tests
...
 14 files changed, 791 insertions(+), 43 deletions(-)
```
Calculations;
```
In [1]: (1154-791)
Out[1]: 363

In [2]: (211-43)
Out[2]: 168

In [3]: 363-168
Out[3]: 195
```
Excluding tests, this PR adds 195 lines of code.

cliffordwolf:
> Excluding tests, this PR adds 195 lines of code.

Excluding the infrastructure for making this an optional feature, which your patch is lacking, and excluding handling of the case of attributes on Verilog-2001-style parameters, which your patch is lacking as well, and excluding the added documentation for the new parameters, which is also missing in your patch, my patch adds about 20 lines.

mkurc-ant:
@cliffordwolf I appreciate that you were able to implement almost the same functionality within ~20 lines of code. But your patch targets only the Verilog frontend. What about backends?

With your patch those wires that correspond to parameters will appear as wires and there is no way to distinguish them from actual wires of a module. The change I proposed also affects backends such as Verilog, Ilang an JSON so module parameters appear in their output as parameters. That's why the code is longer.

If you want the feature of having those wires to be optional it can be easily added as well as support for attributes on Verilog 2001 style parameters.

cliffordwolf:
> With your patch those wires that correspond to parameters will appear as wires and there is no way to distinguish them from actual wires of a module.

No. That's why we have the attribute. That's the whole point for it.

> The change I proposed also affects backends such as Verilog, Ilang an JSON so module parameters appear in their output as parameters.

We proposed the attribute solution specifically because we did not want to have intrusive changes all over the place for an obscure feature that besides you nobody is using. And we've been very clear about that context.

I you don't like it I can also remove the feature altogether again.

PR #1107, opened 2019-06-19
Unpacked array declaration using size

towoe:
The test needs checks.
Manually I was able to see the output of `stat` to see that there are 2 memories and 16 memory bits as opposed to 9 memory bits in the same test without the changes to `verilog_parser.y` but I did not find a way to check for this.
I would be happy to extend this checks if I could get a pointer of how to do it.


cliffordwolf:
Merged and added test that actually checks something in 6a6dd5e.

towoe:
Thanks a lot Clifford for adding a test!

PR #1108, opened 2019-06-19
Add support for partial matches to muxcover

### LINKED ISSUE #1091: muxcover refuses to create MUX4 for 3:1 mux

`yosys -p "proc; opt; wreduce; techmap; muxcover; stat"`

```
module mux_if_bal_3_1 #(parameter N=3, parameter W=1) (input [N*W-1:0] i, input [$clog2(N)-1:0] s, output reg [W-1:0] o);
always @* begin
    o <= {{W{{1'bx}}}};
    if (s[0] == 1'b0)
     if (s[1] == 1'b0)
      o <= i[0*W+:W];
     else
      o <= i[1*W+:W];
    else
     if (s[1] == 1'b0)
      o <= i[2*W+:W];
end
endmodule
```

## Expected behavior

```
     $_MUX4_                         1
```
as is the case when `wreduce` is removed (which optimises 1/3 `$_MUX_` cells away)

## Actual behavior

```
     $_MUX_                          2
```
Even when changing the mux4 cost by specifying the new `-mux4=0` argument to `muxcover`.

Same behaviour (and perhaps more a more compelling case) is the same behaviour occurs for 7:1 mux should create a MUX8, and 15:1 mux should create a MUX16.

In fact, I want to influence the algorithm in such a way to always "round up": 
* \>= 3:1 mux into MUX4
* \>=5:1 mux into MUX8
* \>=9:1 mux into MUX16
which currently changing the costs does not do.

cliffordwolf:
Fixes #1091

### DIFF
--- a/passes/techmap/muxcover.cc
+++ b/passes/techmap/muxcover.cc
@@ -57,6 +57,7 @@ struct MuxcoverWorker
 	bool use_mux8;
 	bool use_mux16;
 	bool nodecode;
+	bool freedecode;
 	bool nopartial;
 
 	int cost_mux2;
@@ -70,6 +71,7 @@ struct MuxcoverWorker
 		use_mux8 = false;
 		use_mux16 = false;
 		nodecode = false;
+		freedecode = false;
 		nopartial = false;
 		cost_mux2 = COST_MUX2;
 		cost_mux4 = COST_MUX4;
@@ -176,6 +178,9 @@ struct MuxcoverWorker
 			return 0;
 
 		if (A == State::Sx || B == State::Sx)
+			return 0;
+
+		if (freedecode)
 			return 0;
 
 		return std::max((cost_mux2 / GetSize(std::get<1>(entry))) - 1, 1);
@@ -620,6 +625,9 @@ struct MuxcoverPass : public Pass {
 		log("        substitutions, but guarantees that the resulting circuit is not\n");
 		log("        less efficient than the original circuit.\n");
 		log("\n");
+		log("    -freedecode\n");
+		log("        Do not count cost for generated decode logic\n");
+		log("\n");
 		log("    -nopartial\n");
 		log("        Do not consider mappings that use $_MUX<N>_ to select from less\n");
 		log("        than <N> different signals.\n");
@@ -633,6 +641,7 @@ struct MuxcoverPass : public Pass {
 		bool use_mux8 = false;
 		bool use_mux16 = false;
 		bool nodecode = false;
+		bool freedecode = false;
 		bool nopartial = false;
 		int cost_mux4 = COST_MUX4;
 		int cost_mux8 = COST_MUX8;
@@ -670,6 +679,10 @@ struct MuxcoverPass : public Pass {
 				nodecode = true;
 				continue;
 			}
+			if (arg == "-freedecode") {
+				freedecode = true;
+				continue;
+			}
 			if (arg == "-nopartial") {
 				nopartial = true;
 				continue;
@@ -694,6 +707,7 @@ struct MuxcoverPass : public Pass {
 			worker.cost_mux8 = cost_mux8;
 			worker.cost_mux16 = cost_mux16;
 			worker.nodecode = nodecode;
+			worker.freedecode = freedecode;
 			worker.nopartial = nopartial;
 			worker.run();
 		}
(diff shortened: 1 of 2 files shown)


eddiehung:
So it seems that `opt_expr -full` restructures a 5:1 mux (design below) from [before.pdf](https://github.com/YosysHQ/yosys/files/3307291/before.pdf) to [after.pdf](https://github.com/YosysHQ/yosys/files/3307294/after.pdf) which causes MUX8 covering to fail in the sense that it is being used a MUX4:

```
  cell $_MUX8_ $auto$muxcover.cc:509:implement_best_cover$76
    connect \A \i [0]
    connect \B \i [0]
    connect \C \i [1]
    connect \D \i [1]
    connect \E \i [2]
    connect \F \i [2]
    connect \G \i [3]
    connect \H \i [3]
    connect \S 1'x
    connect \T \s [2]
    connect \U \s [1]
    connect \Y $procmux$20_Y
```

Reproduce with: `./yosys -p "proc; opt -full; techmap; muxcover -mux8=200" mux_if_bal_5_1.v`
on:
```
module mux_if_bal_5_1 #(parameter N=5, parameter W=1) (input [N*W-1:0] i, input [$clog2(N)-1:0] s, output reg [W-1:0] o);
always @* begin
    o <= {{W{{1'bx}}}};
    if (s[0] == 1'b0)
     if (s[1] == 1'b0)
      if (s[2] == 1'b0)
       o <= i[0*W+:W];
      else
       o <= i[1*W+:W];
     else
      if (s[2] == 1'b0)
       o <= i[2*W+:W];
      else
       o <= i[3*W+:W];
    else
     if (s[1] == 1'b0)
      if (s[2] == 1'b0)
       o <= i[4*W+:W];
end
endmodule
```

cliffordwolf:
> causes MUX8 covering to "fail" in the sense that it is being used as a MUX4.

Since you have disabled MUX4 it is sensible to use MUX8 as MUX4.

The solution also makes sense from a cost perspective. In both the "correct" and the "incorrect" case a MUX8+MUX2 is needed. In one case the MUX2 is just tagged at the end to implement the 5th input, in the other case the MUX2 is used for the decoder for the select pins. I've now changed the cost function slightly so that decoder MUXes are ever so slightly cheaper than datapath MUXes.

Now it prefers the following solution, which I think is the solution one would expect:

image

And `muxcover -mux4=198 -mux8=200` will produce the following:

image

Note that because of the changed cost function, `-mux4=199 -mux8=200` will now prefer (seemingly incorrectly) the MUX8 solution because of the slightly decreased mux2 cost for decoder logic.

I also added some debug log output. `debug muxcover [...]` will now print the costs for all steps in that dynamic programming algorithm.

eddiehung:
Okay, my original testcase now passes. However, this one has the same problem for MUX16:
mux_if_bal_9_1

`./yosys -p "proc; opt -full; techmap; muxcover -mux16=200" mux_if_bal_9_1.v`
```
module mux_if_bal_9_1 #(parameter N=9, parameter W=1) (input [N*W-1:0] i, input [$clog2(N)-1:0] s, output reg [W-1:0] o);
always @* begin
    o <= {{W{{1'bx}}}};
    if (s[0] == 1'b0)
     if (s[1] == 1'b0)
      if (s[2] == 1'b0)
       if (s[3] == 1'b0)
        o <= i[0*W+:W];
       else
        o <= i[1*W+:W];
      else
       if (s[3] == 1'b0)
        o <= i[2*W+:W];
       else
        o <= i[3*W+:W];
     else
      if (s[2] == 1'b0)
       if (s[3] == 1'b0)
        o <= i[4*W+:W];
       else
        o <= i[5*W+:W];
      else
       if (s[3] == 1'b0)
        o <= i[6*W+:W];
       else
        o <= i[7*W+:W];
    else
     if (s[1] == 1'b0)
      if (s[2] == 1'b0)
       if (s[3] == 1'b0)
        o <= i[8*W+:W];
end
endmodule
```

cliffordwolf:
JFYI: Please don't force-push to other peoples branches without checking with thte other person. Resolving the resulting merge conflicts is extremely frustrating. (Pushing is fine but force-pushing is an issue imo.)

> However, this one has the same problem for MUX16

I don't quite understand what result you would expect here. This is the most cost-efficient implementation. I've now added a `-freedecode` option where it doesn't count the cost for the extra decode logic, but I don't think this makes a lot of sense in practice.

q3k:

eddiehung:
> JFYI: Please don't force-push to other peoples branches without checking with thte other person. Resolving the resulting merge conflicts is extremely frustrating. (Pushing is fine but force-pushing is an issue imo.)

You're right, I'm sorry. The PR had a some diff noise due to one of my commits so I rebased it onto master. I'll check with you in the future.

Issue #298, opened 2017-01-19
typedef support


OrDavidi:
Hi,
I encounter some code with the typedef syntax.
Any idea of working around it ? maybe some preprocessor tricks?  or maybe you have a future plan to support it ?

for example : 
typedef reg [127:0] data_t; 

Thanks

cliffordwolf:
Continuing discussion from #1110 @towoe:

> If you would be open to suggest a different approach I would be happy to implement it. Should we discuss this in the issue instead?

The biggest issue here I think is being able to parse variable declarations that simply are `<some_identifier> <some_identifier>;` without knowing already in the lexer that the 1st identifier refers to a type. Currently this would result in reduce/reduce conflicts (or at least shift/reduce conflicts) with the rules for parsing instances.

Spending some time on figuring out how to deal with this mess would be a good first step I think.

jjj11x:
I think https://github.com/YosysHQ/yosys/pull/1389 mostly fixes this issue, could close it. The typedef'd typenames still have to be enclosed in parentheses when used (even the tests do so), but other than that everything seems to work for me.

zachjs:
The outstanding portion of this issue was resolved in #1803.

PR #1110, opened 2019-06-19
Typedef support

towoe:
Based on the work from https://github.com/PeterCrozier/yosys/commit/881833aa738e7404987646ea8076284e911fce3f
this allows the usage of typedef inside a module.

Should satisfy https://github.com/YosysHQ/yosys/issues/298


cliffordwolf:
I think resolving types in the lexer means it will be virtually impossible to have typedefs in packages, because package content is injected into the module at AST-level and thus the lexer will be unaware of it. (Similar issues for things like types local to a context.)

towoe:
Thanks Clifford for the feedback! Yeah, would be better to have typedef also for packages.
My idea was to start with typedef and then add support for packages in a second step, but as you said it makes no sense to implement it this way, I need to do it differently. If you would be open to suggest a different approach I would be happy to implement it. Should we discuss this in the issue instead? Thanks, Tobias

cliffordwolf:
> Should we discuss this in the issue instead?

See post in #298. Closing here.

PR #1111, opened 2019-06-19
Fixed the help summary line for a few commands

acw1251:
When you run the "help" command in Yosys without an argument, you get a list of commands and a summary of each command. I noticed a few summaries were incorrect since they described other commands. I went through and fixed all the incorrect summaries that appeared to be for other commands.

### DIFF
--- a/techlibs/ice40/ice40_unlut.cc
+++ b/techlibs/ice40/ice40_unlut.cc
@@ -74,7 +74,7 @@ static void run_ice40_unlut(Module *module)
 }
 
 struct Ice40UnlutPass : public Pass {
-	Ice40UnlutPass() : Pass("ice40_unlut", "iCE40: perform simple optimizations") { }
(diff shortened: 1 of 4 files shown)


techlibs/ice40/ice40_unlut.cc
```diff
@@ -74,7 +74,7 @@ static void run_ice40_unlut(Module *module)
 }
 
 struct Ice40UnlutPass : public Pass {
-	Ice40UnlutPass() : Pass("ice40_unlut", "iCE40: perform simple optimizations") { }
+	Ice40UnlutPass() : Pass("ice40_unlut", "iCE40: transform SBLUT4 cells to $lut cells") { }
```
eddiehung: Really small nitpick: should be `SB_LUT4`

acw1251: I just pushed another commit to fix it. Thanks!

PR #1112, opened 2019-06-19
Fixed pyosys commands returning RTLIL::SigSig

acw1251:
In the Python API for Yosys (pyosys), when I try to access `connections_` from a `Module`, I get the following error in python:
```
TypeError: No to_python (by-value) converter found for C++ type: Yosys::RTLIL::SigSpec
```

I have no previous experience with Boost Python, but it appears from looking at the rest of the C++ wrapper that the C++ code for creating a tuple is missing a call to `get_py_obj` when necessary. I looked at the code in `misc/py_wrap_generator.py` for other containers and modeled my solution after those. In addition to fixing the bug, I also removed a significant amount of dead code in the function.

A possible improvement for this function (beyond my bugfix) is to add logic for the case where the types are in `known_containers`. This case doesn't happen with the current Yosys API, but if any function is added that returns a tuple of one or more containers, then a bug will reappear. This simplest way to protect against this is to raise an exception if either type is a container.

Here's a python script to reproduce the bug I fixed:

```python
try:
except:
    import sys
    sys.path.append('/usr/lib/python3.5/site-packages')

with open('test.v', 'w') as f:
    f.write('''
module test_module(
        input x,
        output y
    );
    assign y = x;
endmodule''')

design = ys.Design()
ys.run_pass("read_verilog test.v", design);

module = design.selected_whole_modules_warn()[0]
connections = module.connections_
print(str(connections))
```

### DIFF
--- a/misc/py_wrap_generator.py
+++ b/misc/py_wrap_generator.py
@@ -508,23 +508,17 @@ class TupleTranslator(PythonDictTranslator):
 	#Generate c++ code to translate to a boost::python::tuple
 	@classmethod
 	def translate_cpp(c, varname, types, prefix, ref):
-		text  = prefix + TupleTranslator.typename + " " + varname + "___tmp = boost::python::make_tuple(" + varname + ".first, " + varname + ".second);"
-		return text
-		tmp_name = "tmp_" + str(Translator.tmp_cntr)
-		Translator.tmp_cntr = Translator.tmp_cntr + 1
-		if ref:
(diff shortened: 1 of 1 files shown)


cliffordwolf:
cc @btut

q3k:

cliffordwolf:
ping! ping! ping! @btut

btut:
Sorry, didn't realize this needed my approval or that I was even allowed to approve commits in this repo.
I will take care of the known containers scenario.

PR #1122, opened 2019-06-21
Added JSON upto and offset

cliffordwolf:


q3k:

PR #1130, opened 2019-06-25
memory_dff: walk through more than one mux for computing read enable

### LINKED ISSUE #710: Yosys fails to infer BRAM with complex enable logic

Run `synth_ice40` on the below design:

```
module top(input clk, input we, re, input [7:0] addr, wdata, output reg [7:0] rdata);

reg [7:0] bram[0:255];
(* keep *) reg dummy;

always @(posedge clk)
	if (reset)
		dummy <= 1'b0;
	else if (re)
		rdata <= bram[addr];
	else if (we)
		bram[addr] <= wdata;
endmodule

```

## Expected behavior

BRAM is inferred, with both read and write enables ANDed with !reset

## Actual behavior

Yosys fails to infer BRAM with the following message:

```
    Mapping to bram type $__ICE40_RAM4K_M0 (variant 1):
      Write port #0 is in clock domain \clk.
        Mapped to bram port B1.
      Read port #0 is in clock domain !~async~.
        Bram port A1.1 has incompatible clock type.
        Failed to map read port #0.
    Mapping to bram type $__ICE40_RAM4K_M0 failed.
```
This is because no flipflop is merged into the read port:
```
Checking cell `$memrd$\bram$read_enable.v:10$6' in module `\top': no (compatible) $dff found.
```

eddiehung:
Fixes #710

### DIFF
--- a/passes/memory/memory_dff.cc
+++ b/passes/memory/memory_dff.cc
@@ -193,10 +193,9 @@ struct MemoryDffWorker
 				en.append(enable_invert ? module->LogicNot(NEW_ID, mux->getPort("\\S")) : mux->getPort("\\S"));
 			} while (mux_cells_a.count(sig_data) || mux_cells_b.count(sig_data));
 
-                        for (auto bit : sig_data)
-                                if (sigbit_users_count[bit] > 1) {
-                                        goto skip_ff_after_read_merging;
-                                }
+			for (auto bit : sig_data)
+				if (sigbit_users_count[bit] > 1)
+					goto skip_ff_after_read_merging;
 
 			if (find_sig_before_dff(sig_data, clk_data, clk_polarity, true) && clk_data != RTLIL::SigSpec(RTLIL::State::Sx) && sig_data == check_q)
 			{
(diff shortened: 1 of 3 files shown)


passes/memory/memory_dff.cc
```diff
@@ -182,20 +182,28 @@ struct MemoryDffWorker
...
-			for (auto bit : sig_data)
-				if (sigbit_users_count[bit] > 1)
-					goto skip_ff_after_read_merging;
+			do {
+				bool enable_invert = mux_cells_a.count(sig_data) != 0;
+				Cell *mux = enable_invert ? mux_cells_a.at(sig_data) : mux_cells_b.at(sig_data);
+				check_q = sigmap(mux->getPort(enable_invert ? "\\B" : "\\A"));
+
+				sig_data = sigmap(mux->getPort("\\Y"));
+				for (auto bit : sig_data)
```
eddiehung: TODO: Move this check (ensure that the one and only consumer is this read port) outside of loop so that it only operates on the last `sig_data`

PR #1073, opened 2019-06-06
ECP5: implement most Diamond I/O buffer primitives

whitequark:
I'm not sure what to do with the differential ones (or how are they supposed to be used in the first place) so those are stubbed for now. But the list includes every primitive according to Lattice's [FPGA Libraries Reference Guide](http://www.latticesemi.com/~/media/LatticeSemi/Documents/UserManuals/EI/FPGALibrariesReferenceGuide32.pdf?document_id=50687).

PR #1133, opened 2019-06-25
Add more ECP5 Diamond flip-flops.

whitequark:
This includes all I/O registers, and a few more regular FFs where it was convenient.

Also, it looks like I've missed the I/O buffers in `cells_sim.v` in #1073.


whitequark:
I think I didn't commit a part of this PR; force-pushed to update.

PR #1138, opened 2019-06-26
synth_xilinx: Add -nocarry and -nowidelut options

eddiehung:
Based on and superseding #971 by @koriakin.

Instead of `-nomux` for `synth_xilinx`, it is now `-nowidelut`.
On `synth_ecp5` I've also renamed the flag and the doc, but it still accepts the old `-nomux`.

### DIFF
--- a/techlibs/ecp5/synth_ecp5.cc
+++ b/techlibs/ecp5/synth_ecp5.cc
@@ -76,7 +76,7 @@ struct SynthEcp5Pass : public ScriptPass
 		log("    -nodram\n");
 		log("        do not use distributed RAM cells in output netlist\n");
 		log("\n");
-		log("    -nomux\n");
+		log("    -nowidelut\n");
 		log("        do not use PFU muxes to implement LUTs larger than LUT4s\n");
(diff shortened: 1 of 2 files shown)


techlibs/xilinx/synth_xilinx.cc
```diff
@@ -237,10 +253,12 @@ struct SynthXilinxPass : public ScriptPass
...
 			}
 
-			if (!vpr || help_mode)
+			if (help_mode)
+				run("techmap -map +/techmap.v -map +/xilinx/arith_map.v", "(skip if '-nocarry')");
+			else if (!vpr)
 				run("techmap -map +/techmap.v -map +/xilinx/arith_map.v");
 			else
-				run("techmap -map +/techmap.v +/xilinx/arith_map.v -D _EXPLICIT_CARRY");
+				run("techmap -map +/techmap.v -map +/xilinx/arith_map.v -D _EXPLICIT_CARRY");
```
koriakin: Shouldn't nocarry option be actually used somewhere around here?

PR #1139, opened 2019-06-26
tests: Check that Icarus can parse arch sim models

daveshah1:
Should avoid problems like those fixed in #1137 in the future.

### DIFF
--- a/tests/arch/run-test.sh
+++ b/tests/arch/run-test.sh
@@ -4,5 +4,15 @@ set -e
 
 echo "Running syntax check on arch sim models"
 for arch in ../../techlibs/*; do
-	find $arch -name cells_sim.v -print0 | xargs -0 -n1 -r iverilog -t null -I$arch
+	find $arch -name cells_sim.v | while read path; do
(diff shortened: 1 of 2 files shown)


tests/arch/run-test.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+set -e
+
+echo "Running syntax check on arch sim models"
+iverilog -t null ../../techlibs/achronix/speedster22i/cells_sim.v
```
eddiehung: Can we wildcard this? It would mean we don't need to remember to update this file when we add new arches

daveshah1: Yes, I started this not sure how different arches specified these files, but then it seems like they are fairly consistent so just looking for cells_sim in root or subfolders should work

daveshah1: Changed to use find and xargs

eddiehung:
We might as well also chuck techlibs/common/sim{cells,lib}.v in here as well.

Issue #1132, opened 2019-06-25
wreduce causes muxcover to not identify covering opportunity


eddiehung:
## Steps to reproduce the issue

` ./yosys -p "proc; opt; wreduce; simplemap; muxcover; dump" mux_if_bal_3_1.v`

```
module mux_if_bal_3_1 #(parameter N=3, parameter W=1) (input [N*W-1:0] i, input [$clog2(N)-1:0] s, output reg [W-1:0] o);
always @* begin
    o <= {{W{{1'bx}}}};
    if (s[0] == 1'b0)
     if (s[1] == 1'b0)
      o <= i[0*W+:W];
     else
      o <= i[1*W+:W];
    else
     if (s[1] == 1'b0)
      o <= i[2*W+:W];
end
endmodule
```

## Expected behavior

Pack into `$_MUX4_`

## Actual behavior

No packeroonies.

Before `wreduce`:
![before](https://user-images.githubusercontent.com/6740044/60113502-c0296300-9726-11e9-96a6-7610f84c343f.png)

After `wreduce`:
![after](https://user-images.githubusercontent.com/6740044/60113518-c61f4400-9726-11e9-9092-db58672b76e4.png)

eddiehung:
@cliffordwolf As discussed, you mentioned that `wreduce` had some special handling for mux cells that we could make optional, so that these opportunities are not lost...

cliffordwolf:
Thanks for this example. I think there's no need to do anything to wreduce here. It's just that the default costs in muxcover are set to values that don't incentivize muxcover to use a MUX4 to replace two MUX. From `help muxcover`:

```
    -mux4[=cost], -mux8[=cost], -mux16[=cost]
        Use the specified types of MUXes (with optional integer costs). If none
        of these options are given, the effect is the same as if all of them are.
        Default costs: $_MUX_ = 100, $_MUX4_ = 220,
                       $_MUX8_ = 460, $_MUX16_ = 940
```

And after wreduce has eliminated one of the muxes, muxcover has the choice to replace two MUX2 with one MUX4, but doesn't do it because two MUX2 have lower cost. (Before wreduce muxcover thinks it replaces three MUX2 with one MUX4, which it happily does.)

Setting the MUX4 cost to under 200 results in the desired output:

```
yosys -p "proc; opt; wreduce; simplemap; muxcover -mux4=150;; show" mux_if_bal_3_1.v
```

[...]

PR #1140, opened 2019-06-26
[WIP] muxcover decoder generation to be undef aware

eddiehung:
Fixes an issue uncovered in https://github.com/YosysHQ/yosys/issues/1132#issuecomment-506008086

This fixes issues with MUXes not being able to be covered by MUX8 or MUX16 due to generation of a redundant decoder, by making decoder generation undef (`1'bx`) aware.

~EDIT: Not sure this is working; even though it passes my micro benchmarks, it (specifically, flows that employ `muxcover`) fails when used to synth the picorv32 and then checking the result with the testbench...~

Now passes picorv32 tb, hopefully not just by luck.

### DIFF
--- a/passes/techmap/muxcover.cc
+++ b/passes/techmap/muxcover.cc
@@ -324,7 +324,13 @@ struct MuxcoverWorker
 			ok = ok && follow_muxtree(G, tree, bit, "BBA");
 			ok = ok && follow_muxtree(H, tree, bit, "BBB");
 
-			ok = ok && follow_muxtree(U1, tree, bit, "S");
+			ok = ok && follow_muxtree(S1, tree, bit, "AAS");
+			ok = ok && follow_muxtree(S2, tree, bit, "ABS");
+			ok = ok && follow_muxtree(S3, tree, bit, "BAS");
+			ok = ok && follow_muxtree(S4, tree, bit, "BBS");
+
+			if (nodecode)
+				ok = ok && S1 == S2 && S2 == S3 && S3 == S4;
 
 			ok = ok && follow_muxtree(T1, tree, bit, "AS");
 			ok = ok && follow_muxtree(T2, tree, bit, "BS");
@@ -332,14 +338,7 @@ struct MuxcoverWorker
 			if (nodecode)
 				ok = ok && T1 == T2;
 
-			ok = ok && follow_muxtree(S1, tree, bit, "AAS");
-			ok = ok && follow_muxtree(S2, tree, bit, "ABS");
-			ok = ok && follow_muxtree(S3, tree, bit, "BAS");
-			ok = ok && follow_muxtree(S4, tree, bit, "BBS");
-
-			if (nodecode)
-				ok = ok && S1 == S2 && /*S2 == S3 &&*/ S3 == S4;
-						       // ^^ Should already be checked by T1 == T2 above
+			ok = ok && follow_muxtree(U1, tree, bit, "S");
 
 			if (ok)
 			{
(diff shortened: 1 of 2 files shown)


passes/techmap/muxcover.cc
```diff
@@ -324,21 +324,22 @@ struct MuxcoverWorker
...
 
-			ok = ok && follow_muxtree(U1, tree, bit, "S");
+			ok = ok && follow_muxtree(S1, tree, bit, "AAS");
+			ok = ok && follow_muxtree(S2, tree, bit, "ABS");
+			ok = ok && follow_muxtree(S3, tree, bit, "BAS");
+			ok = ok && follow_muxtree(S4, tree, bit, "BBS");
+
+			if (nodecode)
+				ok = ok && S1 == S2 && /*S2 == S3 &&*/ S3 == S4;
+						       // ^^ Should already be checked by T1 == T2 above
```
eddiehung: NB This comment

passes/techmap/muxcover.cc
```diff
@@ -407,21 +410,29 @@ struct MuxcoverWorker
...
 
 			ok = ok && follow_muxtree(T1, tree, bit, "AAS");
 			ok = ok && follow_muxtree(T2, tree, bit, "ABS");
 			ok = ok && follow_muxtree(T3, tree, bit, "BAS");
 			ok = ok && follow_muxtree(T4, tree, bit, "BBS");
 
 			if (nodecode)
-				ok = ok && T1 == T2 && T2 == T3 && T3 == T4;
+				ok = ok && (T1 == T2 || T1 == State::Sx || T2 == State::Sx) &&
+					/*T2 == T3 &&*/
```
eddiehung: My understanding is that this check is overly strict since `U1 == U2` is checked downstream?

passes/techmap/muxcover.cc
```diff
@@ -446,19 +457,26 @@ struct MuxcoverWorker
...
+				mux.cost += prepare_decode_mux(U1, U2, V1, bit);
 
 				mux.cost += prepare_decode_mux(T1, T2, U1, bit);
 				mux.cost += prepare_decode_mux(T3, T4, U2, bit);
-				mux.cost += prepare_decode_mux(T1, T3, V1, bit);
+				if (U1 != U2 && U1 != State::Sx && U2 != State::Sx) {
+					mux.cost += prepare_decode_mux(T1, T3, V1, bit);
 
-				mux.cost += prepare_decode_mux(U1, U2, V1, bit);
+					mux.cost += prepare_decode_mux(S1, S2, T1, bit);
```
eddiehung: TODO: Perhaps this shouldn't be nested inside `if (U1 != U2 && U1 != State::Sx && U2 != State::Sx)`

passes/techmap/muxcover.cc
```diff
@@ -330,13 +330,15 @@ struct MuxcoverWorker
 			ok = ok && follow_muxtree(S4, tree, bit, "BBS");
 
 			if (nodecode)
-				ok = ok && S1 == S2 && S2 == S3 && S3 == S4;
+				ok = ok && (S1 == S2 || S1 == State::Sx || S2 == State::Sx) &&
+					/*S2 == S3 && */
+					(S3 == S4 || S3 == State::Sx || S4 == State::Sx);
```
cliffordwolf: That doesn't look right at all. The original code checks that S1, S2, S3, and S4 are all the same. The new code would be satisfied if only S1==S2 and S3==S4, but the link between S1 and S4 is broken. I'll prepare an alternative PR.

cliffordwolf:
See #1142 for my proposed fix.

passes/techmap/muxcover.cc
eddiehung: My line of thinking was that we only check if S1==S2 and S3==S4 here, because we check T1==T2 later, and if T1 or T2 is don't care then there is no need for S2 == S3?

eddiehung:
Superseded by #1142

PR #1141, opened 2019-06-26
Also allow unique/priority case keywords on if statements

tux3:
This naively extends 00dba4c197b7e3b6c1d1f7b90ae7b1e6172b1e5f to also allow the keywords on if statements.
I'm wholeheartedly hoping the `parallel_case` attribute works just as well on if statements.


eddiehung:
Thanks for the contribution @tux3. Would you be able to build a test for this please?

cliffordwolf:
I don't think that actually does what it's supposed to to, because unique/unique0/priority on an if-else-if-else-if-...-else cascade should act on the whole cascade as if it's one big case statement, turning it into one big $pmux in case of unique/unique0, but what actually happens in this patch is that it only acts on the top-level mux afaict.

tux3:
Ok, that sounds like something for later then =]

PR #1147, opened 2019-06-28
Improve specify dummy parser

### LINKED ISSUE #1144: Fails to parse edge sensitive path in specify

```
module top(input clk, d, output q);
specify
  // Fails:
  (posedge clk => (q +: d)) = (3,1);
  (/*posedge*/ clk => (q +: d)) = (3,1);
  (posedge clk *> (q +: d)) = (3,1);
  (/*posedge*/ clk *> (q +: d)) = (3,1);

  // Works:
  (/*posedge*/ clk => q) = (3,1);
  (/*posedge*/ clk *> q) = (3,1);
endspecify
endmodule
```

From: http://verilog.renerta.com/source/vrg00014.htm

cliffordwolf:
Fixes #1144

### DIFF
--- a/frontends/verilog/verilog_parser.y
+++ b/frontends/verilog/verilog_parser.y
@@ -1068,11 +1068,18 @@ list_of_path_delay_extra_expressions :
 	  path_delay_expression ',' path_delay_expression ',' path_delay_expression
 	;
 
+specify_edge_identifier :
+	TOK_POSEDGE | TOK_NEGEDGE ;
+
 parallel_path_description :
-	'(' specify_input_terminal_descriptor opt_polarity_operator '=' '>' specify_output_terminal_descriptor ')' ;
+	'(' specify_input_terminal_descriptor opt_polarity_operator '=' '>' specify_output_terminal_descriptor ')' |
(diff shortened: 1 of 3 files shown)


cliffordwolf:
> I've just added the test from #1144 [...] Is this something you want to resolve too?

As I wrote in https://github.com/YosysHQ/yosys/issues/1144#issuecomment-506649738, two of the lines in that code are not valid specify syntax. Did you comment them out before testing?

eddiehung:
Sorry I wasn't clear. It is currently failing at line 10 of the _existing_ specify.v test, when read without the `-specify` option:

https://github.com/YosysHQ/yosys/blob/dc677c791de438a493ad5e0101987da29c6a6d0f/tests/various/specify.v#L10

which is before the #1144 testcase I added in line 32. I didn't even get to verify that the two invalid lines were indeed rejected because it failed on line 10.

cliffordwolf:
In order to properly support such constructs in the default "ignore specify" parser, that parser essentially needs to be re-written, and frankly I don't have the time to do that right now. So I've fixed the easy-to-fix stuff and then added an ifdef-guard that comments out the more advanced constructs in non`-specify`-mode in 1f17321.

PR #1148, opened 2019-06-28
synth_xilinx to infer wide multiplexers using new '-widemux <min>' option

eddiehung:
From doc:

```
    -widemux <int>
        enable inference of hard multiplexer resources (MuxFx) for muxes at or
        above this number of inputs (minimum value 5).
        default: 0 (no inference)
```


techlibs/xilinx/synth_xilinx.cc
```diff
@@ -186,6 +193,10 @@ struct SynthXilinxPass : public ScriptPass
 				nosrl = true;
 				continue;
 			}
+			if (args[argidx] == "-widemux" && argidx+1 < args.size()) {
+				widemux = atoi(args[++argidx].c_str());
```
daveshah1: Might as well use `std::stoi` here?

techlibs/xilinx/synth_xilinx.cc
```diff
@@ -260,32 +274,65 @@ struct SynthXilinxPass : public ScriptPass
...
+				run("muxcover <internal options>, ('-widemux' only)");
+			}
+			else if (widemux > 0) {
+				run("simplemap t:$mux");
+				std::string muxcover_args = " -nodecode";
+				switch (widemux) {
+					// NB: Cost of mux2 is 100; mux8 should cost between 3 and 4
+					//     of those so that 4:1 muxes and below are implemented
+					//     out of mux2s
+					case  5: muxcover_args += " -mux8=350 -mux16=400"; break;
```
daveshah1: Wouldn't it be neater to calculate these rather than have a hard-to-maintain enumeration of possibilities here?

eddiehung:
> LGTM (except the `synth -keepdc` part, see [#1166 (comment)](https://github.com/YosysHQ/yosys/pull/1166#issuecomment-509559477))

Reverted in #1171

PR #1149, opened 2019-06-28
Make abc9 pass aware of optional ABCEXTERNAL override

gsomlo:
This is a fixup for the P.R. #1098 merge.

### DIFF
--- a/passes/techmap/Makefile.inc
+++ b/passes/techmap/Makefile.inc
@@ -10,6 +10,7 @@ OBJS += passes/techmap/abc.o
 OBJS += passes/techmap/abc9.o
 ifneq ($(ABCEXTERNAL),)
 passes/techmap/abc.o: CXXFLAGS += -DABCEXTERNAL='"$(ABCEXTERNAL)"'
+passes/techmap/abc9.o: CXXFLAGS += -DABCEXTERNAL='"$(ABCEXTERNAL)"'
 endif
 endif
 


eddiehung:
Thanks @gsomlo! I don't use an external ABC, so this fell under my radar.

Do reach out if you have run into any ABC9 issues!

gsomlo:
On Fri, Jun 28, 2019 at 03:02:55PM -0700, Eddie Hung wrote:
> Thanks @gsomlo! I don't use an external ABC, so this fell under my radar.

Thanks for merging -- I have a vested interest in yosys being easy
and straightforward to package in Fedora (who strongly discourage
bundling, so the RPM depends on a separately packaged ABC at both
build- and run-time).
 
> Do reach out if you have run into any ABC9 issues!

No actual issues, I can still build the same things I could build
before without errors.

As far as "user feedback" (and @daveshah1 may also find this useful),
on ECP5, I am getting *slightly* worse area (approx. 1% more
TRELLIS_SLICEs used), and *significantly* better timing from nextpnr
when it takes over the output from yosys.

The two things I'm building are github.com/gsomlo/yoloRISC
(an rv64imac Rocket based blinky) and LiteX+RocketRV64imac SoC.

Slice utilization stayed about the same for yoloRISC, but timing
went from barely meeting 20MHZ to comfortably exceeding 50MHz.

For LiteX+Rocket64bit, slice utilization went from 97% to 99%, and
timing went from barely meeting 50MHz to a bit over 60MHz.

To be precise, this is aggregate across yosys+trellis+nextpnr,
utilizing builds from mid-May (yosys #c907899, trellis #5eb0ad8, and
nextpnr #c9ba65e) vs. yesterday and today (trellis #3311e6d,
nextpnr #ff95883, and yosys #b5f1bd0 + the fixup you just merged).

Thanks much to you all for the awesome tool chain!

--Gabriel

daveshah1:
Thanks for your testing! It would be good to know how much is from ABC9 and how much is from other improvements. I'd be curious what the difference is between now without -abc9 (so just looking at general improvements), and now with -abc9.

eddiehung:
Thanks for your detailed first impressions @gsomlo! I've been focussed on `synth_xilinx -abc9` where I have been seeing some good area/delay benefits, but with @daveshah1's help it's been brought to my attention that the benefits are not as big on the ecp5 front. That's something I'll be investigating soon, so please keep feeding me any information you come across. There are many suspicious factors such as whether the timing numbers are right, through to better ways we can use ABC.

gsomlo:
On Fri, Jun 28, 2019 at 04:01:47PM -0700, David Shah wrote:
> Thanks for your testing! It would be good to know how much is from
> ABC9 and how much is from other improvements. I'd be curious what
> the difference is between now without -abc9 (so just looking at
> general improvements), and now with -abc9.

Oh, so in that case the numbers above are *without* "-abc9", as I
simply re-ran my unmodified makefiles against the updated toolchain
components. As such, the improved timing and approximately similar
utilization is most likely attributable to nextpnr...

After adding explicit "-abc9" arguments to my yosys "synth_ecp5"
commands, I ended up with the following:

- yoloRISC blinky: slices used: 74% (down from 85%);
Max. frequency: 62.12MHz (slightly up from 60.07MHz);

- LiteX+Rocket64bit: slices used: 83% (down from 99%);
Max. frequency: 57.67MHz (within spitting distance from the
earlier 60.0MHz);

These numbers are extra super awesome from my perspective, and
hopefully much closer to what you were expecting from "-abc9" :)

Thanks again,
--Gabriel

eddiehung:
Thanks @gsomlo those are much closer to the numbers I'm expecting. Basically, abc9 allows us to communicate the existence and delay of hard blocks (rather than making them primary inputs and outputs), so that it can better understand what the critical path is. As such, it should "unflatten" its slack histogram and in doing so, can perform better delay relaxation and use more area efficient implementations of the parts of the circuit that are less critical. This is not an exact science, so it may be doing too much of that and causing a performance drop. Tuning all these interactions is very much ongoing work.

gsomlo:
At @daveshah1's suggestion I tried adding "-nowidelut" to my synth_ecp5 command:
`yosys -p "synth_ecp5 -abc9 -nowidelut -json foo.json -top foo.top" foo.v bar.v ...`
but when I do that, I get something that looks like a permission error:
`ERROR: Can't open ABC output file '/tmp/yosys-abc-RxU45f/output.aig'.`
It seems to be generated here: https://github.com/YosysHQ/yosys/blob/master/passes/techmap/abc9.cc#L517 and so far I've only managed to trigger it when I added "-nowidelut" to the command line; if I leave that out, synth_ecp5 with -abc9 works great as per the results I posted earlier. Am I doing something wrong, or tickling some weird corner case?

eddiehung:
Good catch @gsomlo -- just fixed this is in dd8d264b!

gsomlo:
Thanks @eddiehung, that took care of it! New numbers with `synth_ecp5 -nowidelut -abc9` (vs. just `-abc9`) are:
- yoloRISC blinky: 71% slice util. (further down from 74%); 61.9MHz Fmax (slightly down from 62.12);
- LiteX+Rocket64bit: 81% slice util. (down from 83%); 58.39MHz Fmax (actually up from 57.67);

PR #1150, opened 2019-06-28
Add "script -select [selection]" to allow commands to be taken from wires

eddiehung:
Inspired by `_TECHMAP_DO_*`, with the intention for it to be used in testing circuits procedurally generated outside of Yosys so that we don't need to generate two files -- a Verilog and a YS file (and without using less elegant solutions).

### DIFF
--- a/kernel/yosys.cc
+++ b/kernel/yosys.cc
@@ -1259,7 +1259,7 @@ struct ScriptCmdPass : public Pass {
 		//   |---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|
 		log("\n");
 		log("    script <filename> [<from_label>:<to_label>]\n");
-		log("    script -select [selection]\n");
+		log("    script -scriptwire [selection]\n");
 		log("\n");
 		log("This command executes the yosys commands in the specified file (default\n");
 		log("behaviour), or commands embedded in the constant text value connected to the\n");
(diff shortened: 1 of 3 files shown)


kernel/yosys.cc
```diff
@@ -1254,24 +1254,55 @@ struct HistoryPass : public Pass {
 #endif
 
 struct ScriptCmdPass : public Pass {
-	ScriptCmdPass() : Pass("script", "execute commands from script file") { }
+	ScriptCmdPass() : Pass("script", "execute commands from file or wire") { }
 	void help() YS_OVERRIDE {
+		//   |---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|
 		log("\n");
 		log("    script <filename> [<from_label>:<to_label>]\n");
+		log("    script -select [selection]\n");
```
cliffordwolf: I'd prefer a different name than `-select`. Maybe `-scriptwire`? Or just `-wire`?

kernel/yosys.cc
```diff
@@ -1254,24 +1254,55 @@ struct HistoryPass : public Pass {
...
+		size_t argidx;
+		for (argidx = 1; argidx < args.size(); argidx++) {
+			if (args[argidx] == "-select") {
+				select_mode = true;
+				continue;
+			}
+			break;
+		}
+		if (select_mode) {
+			extra_args(args, argidx, design);
```
cliffordwolf: This will make the script wire current selection. But I think the script should not be run just on the wire that contains it.

kernel/yosys.cc
```diff
@@ -1254,24 +1254,55 @@ struct HistoryPass : public Pass {
...
+				for (auto &c : mod->connections()) {
+					if (!c.first.is_wire())
+						continue;
+					auto w = c.first.as_wire();
+					if (!mod->selected(w))
+						continue;
+					if (!c.second.is_fully_const())
+						log_error("RHS of selected wire %s.%s is not constant.\n", log_id(mod), log_id(w));
+					auto v = c.second.as_const();
+					Pass::call(design, v.decode_string());
```
cliffordwolf: I would have expected the script to be called for the module that contains it. Maybe add a special tweak to run on the entire design instead.

cliffordwolf: Hint: Use `Pass:call_on_module()`.

PR #1152, opened 2019-07-01
Dockerfile

1138-4EB:
This PR enhances the Dockerfile:

- Environment variable DEBIAN_FRONTEND should not be permanent.
- The number of COPY layers is reduced.
- ARG IMAGE is used, to make using other base images easier.
- Three stages are used, to better differentiate the dependencies in each stage.


1138-4EB:
*ping*

q3k:
Thanks!

PR #1153, opened 2019-07-02
memory_dff: Fix checking of feedback mux input when more than one mux

daveshah1:
Fixes a regression from #1130

cc @gsomlo, this was found in https://github.com/gsomlo/yoloRISC

### DIFF
--- a/passes/memory/memory_dff.cc
+++ b/passes/memory/memory_dff.cc
@@ -17,6 +17,7 @@
  *
  */
 
+#include <algorithm>
 #include "kernel/yosys.h"
 #include "kernel/sigtools.h"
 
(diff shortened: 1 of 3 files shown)


daveshah1:
Assuming we are happy with this, I think this should be added to the 0.9 RC as it fixes a relatively serious regression. Not sure how we want to approach that (either cherry-pick, or submit this PR straight to yosys-0.9-rc instead?)

gsomlo:
Tested (with and without "-nowidelut -abc9") on both yoloRISC and LiteX+RocketChip (https://github.com/enjoy-digital/litex); fixes regression on both -- thanks for the quick turnaround!

eddiehung:
Looks good. Sorry this was something I introduced.
The only minor concern I have is that when it comes to support reset-able read ports, this needs to be revisited.

passes/memory/memory_dff.cc
```diff
@@ -197,7 +198,8 @@ struct MemoryDffWorker
 				if (sigbit_users_count[bit] > 1)
 					goto skip_ff_after_read_merging;
 
-			if (find_sig_before_dff(sig_data, clk_data, clk_polarity, true) && clk_data != RTLIL::SigSpec(RTLIL::State::Sx) && sig_data == check_q)
+			if (find_sig_before_dff(sig_data, clk_data, clk_polarity, true) && clk_data != RTLIL::SigSpec(RTLIL::State::Sx) &&
+					std::all_of(check_q.begin(), check_q.end(), [&](const SigSpec &cq) {return cq == sig_data; }))
```
eddiehung: Wait, this checks that all entries in `check_q` are equal to `sig_data`?

daveshah1: Yes, at this point `sig_data` is actually the FF output because of `find_sig_before_dff`'s behaviour

eddiehung: Ahh never mind, I see what it's doing. Merge when ready.

PR #1158, opened 2019-07-04
memory: Add memory_dff -zeroinit and use in Xilinx and ECP5 flows

daveshah1:
This allows mapping of memories with zero-initialised read data registers.

Ultimately for Xilinx flows we want to support configurable initialization too, but this would require more work on the `$mem` cell.


cliffordwolf:
What happens if the resulting `$mem` ends up not being mapped by `memory_bram` but instead is transformed back to logic and FFs by `memory_map`? Afaict the init value would be lost in this case.

daveshah1:
Closing, will work on adding an `INIT` parameter to `$mem` instead as discussed out-of-band.

PR #1160, opened 2019-07-07
synth_intel: Warn about untested Quartus backend

ZirconiumX:
My first PR; since there doesn't seem to be any testing framework and a history of breakage between Quartus and Yosys, I thought it would be best to warn on trying to emit a VQM.

cc @daveshah1 @dh73

### DIFF
--- a/techlibs/intel/synth_intel.cc
+++ b/techlibs/intel/synth_intel.cc
@@ -48,6 +48,8 @@ struct SynthIntelPass : public ScriptPass {
 		log("    -vqm <file>\n");
 		log("        write the design to the specified Verilog Quartus Mapping File. Writing of an\n");
 		log("        output file is omitted if this parameter is not specified.\n");
+		log("        Note that this backend has not been tested and is likely incompatible\n");
+		log("        with recent versions of Quartus.\n");
 		log("\n");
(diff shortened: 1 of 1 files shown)


techlibs/intel/synth_intel.cc
```diff
@@ -108,6 +110,7 @@ struct SynthIntelPass : public ScriptPass {
 			}
 			if (args[argidx] == "-vqm" && argidx + 1 < args.size()) {
 				vout_file = args[++argidx];
+				log_warning("The Quartus backend has not been tested recently and is likely incompatible with modern versions of Quartus.");
```
daveshah1: `log_warning` needs a \n at the end of the string

ZirconiumX: Fixed.

dh73:
JFYI: The -vqm argument is just the write_verilog Yosys backend with [these arguments](https://github.com/YosysHQ/yosys/blob/030483ffb909ab38e10d437d09ec922cb0ad2ce8/techlibs/intel/synth_intel.cc#L246).

ZirconiumX:
The netlist could have broken while the Verilog backend is known to work. That would cause incompatibilities.

PR #1163, opened 2019-07-08
More support for case rule attributes

whitequark:
This is a follow-up PR to #1162 that depends on it and adds inessential (but handy) additional support for attributes on case rules.


whitequark:
(The tests actually pass, GitHub seems to have glitched out.)

PR #1164, opened 2019-07-08
Add muxcover -mux2=cost option

eddiehung:


passes/techmap/muxcover.cc
```diff
@@ -631,7 +631,7 @@ struct MuxcoverPass : public Pass {
 		log("\n");
 		log("Cover trees of $_MUX_ cells with $_MUX{4,8,16}_ cells\n");
```
ZirconiumX: Shouldn't this be `$_MUX{2,4,8,16}_`?

eddiehung: Good spot! `$_MUX_` cells are actually 2:1 multiplexers. I'm adding the ability to change its cost base (rather than be hardcoded to 100) for use in decisions about covering with the 4/8/16:1 cells.

ZirconiumX: That makes sense, but I thought the documentation was a little confusing here.

eddiehung: Fair point! I've updated the doc -- is it clearer now?

ZirconiumX: Yes it is, thank you!

cliffordwolf:
Why? All mux costs are relative to each other. Making the mux2 cost configurable doesn't add anything.

eddiehung:
The reason behind this was that, instead of exporting the default costs from muxcover:
https://github.com/YosysHQ/yosys/blob/6951e3207063f66814e765d7d72e553b8b8eace1/passes/techmap/muxcover.cc#L27

I decided to let the user determine this base cost so that it can compute new relative costs:
https://github.com/YosysHQ/yosys/blob/6951e3207063f66814e765d7d72e553b8b8eace1/techlibs/xilinx/synth_xilinx.cc#L287-L304

I can revert and go the former route, if you think that's better.

cliffordwolf:
> I can revert and go the former route, if you think that's better.

No, it's fine. I was just curious about the reason for this change.

Issue #1165, opened 2019-07-09
Why does wreduce remove $mux cells with 1'bx as input?


eddiehung:
## Steps to reproduce the issue

`./yosys -p "read_verilog -icells mux.v; hierarchy; proc; wreduce` on:
```
module top(input [1:0] i, input [1:0] s, output reg [1:0] o);
assign o[0] = s[1] ? 1'bx : s[0] ? i[1] : i[0];
always @* begin
  o[1] <= 1'bx;
  case (s)
    0: o[1] <= i[0];
    1: o[1] <= i[1];
  endcase
end
endmodule
```

## Expected behavior

`wreduce` to not remove `$mux` cells, but instead leave it to `opt_expr -mux_undef` to optimise them? I understand that there's a `wreduce -keepdc` but is that too big of a hammer, when all I want to do is to preserve `1'bx` inputs of `$mux`-es (such that when they later undergo `muxpack` and `pmux2shiftx` I don't end up with having a really big `$shiftx`...)

## Actual behavior

`wreduce` causes one of the two `$mux` cells driving `o[0]` to be removed, whilst not changing anything in the `$pmux` that drives `o[1]`.
![show](https://user-images.githubusercontent.com/6740044/60853966-344a1900-a1b3-11e9-8022-84c9e4966d19.png)

eddiehung:
From #1019 it looks like `-keepdc` only affects muxes.

PR #1166, opened 2019-07-09
Add "synth -keepdc" option

eddiehung:
Propagated through to `wreduce` call.

Also clarify doc for `wreduce -keepdc` to indicate use only for `$mux` cells.

Response to #1165

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -12,6 +12,7 @@ Yosys 0.9 .. Yosys 0.9-dev
     - Added "synth_xilinx -abc9" (experimental)
     - Added "synth_ice40 -abc9" (experimental)
     - Added "synth -abc9" (experimental)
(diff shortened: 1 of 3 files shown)


cliffordwolf:
A `synth -keepdc` command would need to, well, keep don't cares. This is what the option does everywhere else we support it. But `synth -keepdc` obviously isn't doing that.

At the very least, the option should be renamed to something more sensible.

eddiehung:
I used `-keepdc` due to it being the option that `wreduce` takes as of #1019. Happy to rename the option, but should consider renaming then both.

eddiehung:
As discussed, reverted this particular PR since I'll be calling the commands directly from `synth_xilinx` and won't need this anymore. However, the naming of `wreduce -keepdc` still needs to be reconsidered.

PR #1172, opened 2019-07-09
write_verilog: write RTLIL::Sa aka - as Verilog ?

whitequark:
Currently, the only ways (determined by grepping for regex \bSa\b) to
end up with RTLIL::Sa in a netlist is by reading a Verilog constant
with ? in it as a part of case, or by running certain FSM passes.
Both of these cases should be round-tripped back to ? in Verilog.

I'm not sure if there was some reason it was written as z. That code
goes back to the very first Yosys commit.

### DIFF
--- a/backends/verilog/verilog_backend.cc
+++ b/backends/verilog/verilog_backend.cc
@@ -222,7 +222,7 @@ void dump_const(std::ostream &f, const RTLIL::Const &data, int width = -1, int o
 				case RTLIL::S1: bin_digits.push_back('1'); break;
 				case RTLIL::Sx: bin_digits.push_back('x'); break;
 				case RTLIL::Sz: bin_digits.push_back('z'); break;
-				case RTLIL::Sa: bin_digits.push_back('z'); break;
+				case RTLIL::Sa: bin_digits.push_back('?'); break;
 				case RTLIL::Sm: log_error("Found marker state in final netlist.");
 				}
 			}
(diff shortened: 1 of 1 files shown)


cliffordwolf:
`?` and `z` is the same thing in Verilog. `?` is just syntactic sugar, only allowed in certain places. From IEEE Std 1364-2005:

image

...

image

whitequark:
@cliffordwolf I know that very well. And I think, given that we have `z` and `?` in Verilog map to `z` and `-` in RTLIL, there's no good reason for them to not roundtrip back to `z` and `?` in Verilog. (My immediate motivation is improving nMigen's Verilog output, but this would still be my view if I never worked with anything but Verilog.)

If you want me to make sure `?` is never emitted other than as a part of casez patterns, I can do that as well, it's a pretty minimal addition.

cliffordwolf:
> given that we have z and ? in Verilog map to z and - in RTLIL

That's not correct. See the following example. the `z` and the `?` are both mapped to `-`.

```
module test (input [1:0] a, output reg x, y);
  always @* begin
    x = 0;
    y = 0;
    casez (a)
      2'bz1: x = 1;
      2'b1?: y = 1;
    endcase
  end
endmodule
```

cliffordwolf:
> If you want me to make sure ? is never emitted other than as a part of casez patterns, I can do that as well, it's a pretty minimal addition.

Short clarification: I'm not saying I'm categorically against the change (although we must absolutely make sure `?` is only used in casez, otherwise we'd produce illegal Verilog), I'm trying to understand what the advantage is this change provides, considering that everything that reads the produced Verilog must treat `?` and `z` the same.

whitequark:
> That's not correct. See the following example. the `z` and the `?` are both mapped to `-`.

Hm, you are right. I misread the code in const2ast. (What semantics for `z` in switches does RTLIL define?)

> I'm trying to understand what the advantage is this change provides, considering that everything that reads the produced Verilog must treat `?` and `z` the same.

Essentially, readability. Most Verilog materials I've seen recommend using exclusively `?` in `casez` because `z` is more overloaded, and I agree with that. Since `?` and `z` must be treated exactly the same, there is no change for downstream tooling. That's actually good, because it means we can make things a little nicer for human readers without regressing anything.

whitequark:
Thanks!

PR #1181, opened 2019-07-11
"abc9 -dff" option for sequential synthesis

eddiehung:
Enabled by `abc9 -dff`, the advantage of providing seq logic to ABC9 is so that it can do sequential optimisations.
The only flow currently supported is `synth_xilinx`.

Currently, this behaviour is not so similar to `abc -dff` ~(but `abc9` does not do any retiming, to be made less ambigious by #1601)~ since it requires a lot of massaging (see [abc9_map.v](https://github.com/YosysHQ/yosys/blob/d1fccd5a2d63b265c1866cb4d54aba8f2c9d225c/techlibs/xilinx/abc9_map.v)), after which `abc9` is able to understand the synchronous behaviour of the flop (e.g. clock enable, sync resets, etc.) and also to infer the clock domains that each flop belongs to, thus communicating that to ABC.

Since the default behaviour of `abc9` is to find the fastest mapping possible (and secondly, to minimise area under that fastest constraint), the presence of multiple clock domains with differing clock periods will cause delay relaxation to happen optimistically, since it has no understand of which clocks are fast, and which clocks are slow, thus it will relax all excessively fast paths until it is as slow as the slowest paths in the name of saving area.

Obviously this is not desirable, but it is no worse than current behaviour without the `-dff` option where the entire circuit is treated as a combinatorial cloud. This would be a topic for a future PR.


techlibs/xilinx/abc_map.v
```diff
@@ -18,8 +18,143 @@
...
+//   a format that it understands.
+//
+// For example, (complex) flip-flops are expected to be described as an
+//   combinatorial box (containing all control logic such as clock enable
+//   or synchronous resets) followed by a basic D-Q flop.
+
 // ============================================================================
 
+// The purpose of the following FD* rules are to wrap the flop (which, when
+//   called with the `_ABC' macro set captures contains only its combinatorial
```
eddiehung: ~contains~

techlibs/xilinx/abc_map.v
```diff
@@ -18,8 +18,143 @@
...
+//   behaviour) with:
+// (a) a special $__ABC_FF_ in front of the FD*'s output, indicating to abc9
+//     the location of its basic D-Q flop
```
eddiehung: location -> connectivity

techlibs/xilinx/abc_map.v
```diff
@@ -18,8 +18,143 @@
...
+// (b) a special TECHMAP_REPLACE_.$currQwire that will be used for feedback
```
eddiehung: $currQ<space>wire

PR #1182, opened 2019-07-11
synth_xilinx: Initial Spartan 6 block RAM inference support.

koriakin:
Notes:

- XST leaves a couple more wires unconnected (low bits of address and unused data input bits)
- there are no IS_*_INVERTED parameters on the blackbox (matching ISE), an explicit inverter is emitted instead if needed (it will be folded into the primitive by the map step in ISE)


mithro:
FYI @HackerFoo

PR #1183, opened 2019-07-11
synth_ice40: switch -relut to be always on

whitequark:
As far as I know this switch either does nothing or makes the design faster. People like @smunaut have been using it extensively and it appears robust.

### DIFF
--- a/techlibs/ice40/synth_ice40.cc
+++ b/techlibs/ice40/synth_ice40.cc
@@ -67,9 +67,6 @@ struct SynthIce40Pass : public ScriptPass
 		log("    -retime\n");
 		log("        run 'abc' with -dff option\n");
 		log("\n");
-		log("    -relut\n");
-		log("        combine LUTs after synthesis\n");
-		log("\n");
 		log("    -nocarry\n");
 		log("        do not use SB_CARRY cells in output netlist\n");
 		log("\n");
@@ -106,7 +103,7 @@ struct SynthIce40Pass : public ScriptPass
 	}
 
 	string top_opt, blif_file, edif_file, json_file, abc, device_opt;
-	bool nocarry, nodffe, nobram, dsp, flatten, retime, relut, noabc, abc2, vpr;
+	bool nocarry, nodffe, nobram, dsp, flatten, retime, noabc, abc2, vpr;
 	int min_ce_use;
 
 	void clear_flags() YS_OVERRIDE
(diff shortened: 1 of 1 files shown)


smunaut:
As I mentionned on IRC, wouldn't it be better to ignore the "-relut" option and not remove it ? Just to not break all existing scripts that have it.

Else you need in a Makefile to test yosys's version to know if you need to explicitely add that option or leave it out :/

whitequark:
I think Yosys has never done that for any other options, did it?

whitequark:
Note: with `-abc9`, `-relut` doesn't seem to do much of anything on the designs I tested it on. Which is strange, because it should, in principle, on a techmapped adder followed by a mux with a constant or one of the operands. Nevertheless it doesn't hurt to enable it there as well.

smunaut:
I couldn't find any instance in synth_ice40 of any option being removed.

But in synth_xilinx for instance, I could at least find one example ( 6c256b8cda66e2ba128d5fa3ba344fe4717711f8 ) where `-arch` got replace by `-family` but the code still has `-arch` support as compatibility.
But I can also find examples of non-compatible changes as well ( 36e6da53964b406ad379a60fc289aa3af9beb8a9 ).

So not sure what the policy is here. But `-relut` is definitely a pretty popular option currently ...

whitequark:
@smunaut It is now parsed and ignored.

eddiehung:
> But I can also find examples of non-compatible changes as well ( [36e6da5](https://github.com/YosysHQ/yosys/commit/36e6da53964b406ad379a60fc289aa3af9beb8a9) ).
> 
> So not sure what the policy is here. But `-relut` is definitely a pretty popular option currently ...
This incompatible example was a new option made on a branch and changed on that same branch before it was merged into master. Generally, I would say we preserve them as much as possible, even silently. 
I've knowingly broken this once I think, for an option that was wrong to begin with and was there only for a short time....

janrinze:
Could we please have a -no-relut option now. My design actually runs around 30% slower with -relut.
Not sure if there are other designs that suffer the same fate but I have been tracking this as the cause of a regression where the design used to run at 44 MHz but now only does 33 MHz.
I'm a little surprised that this went from optional to always on without the means to opt-out.

eddiehung:
@janrinze Please see #1187

whitequark:
> I'm a little surprised that this went from optional to always on without the means to opt-out.

This regression was caused by commit 437fec0d88b4a2ad172edf0d1a861a38845f3b1d, which changed the semantics of `SB_LUT4` without considering all in-tree users of `SB_LUT4`, which unfortunately sometimes happens and leads to undesirable results. There is nothing wrong with `-relut` itself though.

PR #1184, opened 2019-07-11
synth_{ice40,ecp5}: more sensible pass label naming

whitequark:
I've always found the labels in synth scripts frustratingly uninformative as to their purpose. This is more of an RFC than a PR, but this change makes it more clear, in my opinion. I'm also open to fixing the labels in any other synth passes people are interested in, I'm not entirely sure which ones are actively maintained so I didn't spend effort to fix all of them upfront.


eddiehung:
While we're on this topic, my RFC would be to make the step that calls `abc` on its own label. That way we can do `synth_* -run :abc; <do whatever>; synth_* -run post_abc:` to override the default `abc` options (or indeed the techmapper completely)

whitequark:
@eddiehung Isn't that just `map_luts`? At least, I don't see any way to cleanly split it off, given that the sequence of commands in `map_luts` depends on which abc mode is selected (or if abc is used at all).

ZirconiumX:
If this gets merged, I'll make a PR to rename synth_intel.

eddiehung:
RFC: `map_dram` -> `map_lutram` (equivalently with supporting files) across the board.

whitequark:
> RFC: `map_dram` -> `map_lutram`

Agreed, will update PR.

daveshah1:
LGTM bar the outstanding change to LUTRAM

To consider: if we change the pass label (which I support), do we also change the `-nodram` option name?

eddiehung:
> To consider: if we change the pass label (which I support), do we also change the `-nodram` option name?

Can we have the old labels alias to the new labels, and the `-nodram` alias to the new `-nolutram` (and only document the new labels) to preserve backwards compat?

ZirconiumX:
Since I'm still kinda new to Yosys; how does one create an alias to a label?

eddiehung:
My guess (without trying) would be something like this:
```
-		if (!nodram && check_label("bram", "(skip if -nobram)"))
+		if (!nolutram && (check_label("map_lutram", "(skip if -nolutram)") || check_label("map_bram")))
```
And maybe we'd need to suppress the second doc string....? (RFC)

daveshah1:
I am less worried about aliasing the pass label, these are really only used for debugging. But aliasing the option flag would be a good idea.

eddiehung:
I expect some scripts would skip a stage by running up-to certain labels and then from certain+1 labels. IMO there's no harm in supporting them if it isn't a big effort, at the cost of some code clutter (we can deprecate them entirely in a few releases time).

whitequark:
@eddiehung I updated the PR with the dram→lutram change.

PR #1186, opened 2019-07-12
abc9/ice40: encapsulate SB_CARRY+SB_LUT4 into one box

eddiehung:
Otherwise ABC warns about internal fanouts and may not preserve carry chains correctly

### DIFF
--- a/passes/techmap/abc.cc
+++ b/passes/techmap/abc.cc
@@ -1172,8 +1172,8 @@ void abc_module(RTLIL::Design *design, RTLIL::Module *current_module, std::strin
 					continue;
 				}
 			}
-
-			cell_stats[RTLIL::unescape_id(c->type)]++;
+			else
+				cell_stats[RTLIL::unescape_id(c->type)]++;
 
 			if (c->type == "\\_const0_" || c->type == "\\_const1_") {
 				RTLIL::SigSig conn;
--- a/techlibs/ice40/abc_hx.box
+++ b/techlibs/ice40/abc_hx.box
@@ -3,15 +3,11 @@
 # NB: Inputs/Outputs must be ordered alphabetically
 #     (with exceptions for carry in/out)
 
-# Inputs: I0 I1 CI
-# Outputs: CO
+# Inputs: A B CI
+# Outputs: O CO
 #   (NB: carry chain input/output must be last
 #        input/output and have been moved there
 #        overriding the alphabetical ordering)
-SB_CARRY 1 1 3 1
+$__ICE40_CARRY_LUT4 1 1 3 2
+400 379 316
 259 231 126
-
-# Inputs: I0 I1 I2 I3
-# Outputs: O
-SB_LUT4 2 1 4 1
-449 400 379 316
--- a/techlibs/ice40/abc_hx.box
+++ b/techlibs/ice40/abc_hx.box
@@ -8,6 +8,6 @@
 #   (NB: carry chain input/output must be last
 #        input/output and have been moved there
 #        overriding the alphabetical ordering)
-$__ICE40_CARRY_LUT4 1 1 3 2
+$__ICE40_FULL_ADDER 1 1 3 2
 400 379 316
 259 231 126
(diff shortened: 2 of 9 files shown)


whitequark:
@eddiehung My personal view is that it would be best to make this not ABC-specific: use pmgen to extract the box from any input, including user input that explicitly instantiates `SB_CARRY` and `SB_LUT4`, and then work on a single box until the very end of the pipeline. Unless something important is blocked on this or you have an objection, I would prefer if you followed this approach as it would lead to further simplifications downstream, such as in `opt_lut` and `ice40_opt`, and it also would make iCE40's `arith_map` not have different code paths for ABC and no-ABC.

whitequark:
Actually, the extract pass is not necessary as long as not including the extract pass will produce results comparable to the existing ABC flow. (I am not entirely sure if ABC9 pessimizes that or not.)

eddiehung:
@whitequark Thanks for your feedback! I think you make some very valid points, and I agree with them but would like to address those in a different PR if that is acceptable?

I would say that `synth_ice40 -abc9` is broken right now due to this issue, as exposed by YosysHQ/yosys-tests#51 so I would like to merge this in and open a new PR to unify both paths.

whitequark:
@eddiehung That's entirely reasonable and I have no objection to merging this if it fixes an acute issue.

Bikeshedding regarding the cell naming: the underscores in `$__` are totally redundant because without the `-icells` option to the Verilog frontend you cannot instantiate any cell starting with `$` at all.

Separately from that, I would prefer it to be called `$SB_CARRY_LUT4` to keep with the general naming sense of the primitives and make it a bit easier to read and understand the IL. It does not require us to commit to having that name forever because it is still an internal cell that can't be easily instantiated. 

Going forward I think it makes sense to have `SB_CARRY_LUT4` as an user-exposed primitive, but I understand the hesitation of doing it right away, so I won't insist on that right now.

whitequark:
Actually, looking again / more closely at it, the new cell is not a combination of CARRY and LUT4, is it? Is there a reason for it to not have all 4 inputs as well as the LUT exposed? Right now it is more of a `$__ICE40_FULL_ADDER`, since its function is hardcoded.

eddiehung:
> Bikeshedding regarding the cell naming: the underscores in `$__` are totally redundant because without the `-icells` option to the Verilog frontend you cannot instantiate any cell starting with `$` at all.

That statement is only true if you do not escape the identifier.
My understand is that there's a Yosys convention that indicates the `$__` are known to be internal (transient) cells that are not meant to be instantiated by the user (and have no simulation description). See for example: https://github.com/YosysHQ/yosys/blob/2a7198db516800720e6e63d5676b68b8add81a5c/techlibs/ice40/brams_map.v#L2
which is generated by `memory_bram` and then techmapped to a concrete primitive.

I am also against `SB_CARRY_LUT4` being a user instantiated primitive. I disagree with naming something with a prefix that /could/ imply it was an original primitive, and with that, confusion and loss of interoperability between Yosys and vendor tools.

whitequark:
> My understand is that there's a Yosys convention that indicates the `$__` are known to be internal (transient) cells that are not meant to be instantiated by the user (and have no simulation description).

Ah, I did not realize there is such a convention, sorry. Then ignore all my other comments about the name prefix here, since they are not relevant. But I still think `$__ICE40_FULL_ADDER` is a more descriptive, arguably less misleading name for this cell.

whitequark:
> I am also against `SB_CARRY_LUT4` being a user instantiated primitive. I disagree with naming something with a prefix that /could/ imply it was an original primitive, and with that, confusion and loss of interoperability between Yosys and vendor tools.

OK. I agree with this argument.

eddiehung:
> But I still think $__ICE40_FULL_ADDER is a more descriptive, arguably less misleading name for this cell.
I agree. Thanks for the suggestion: change made.

eddiehung:
And just for completeness, the addition of `$__ICE40_FULL_ADDER` into `cells_sim.v` (thereby giving it a simulation model even though one isn't expected as it should never be user-facing) is so that we can describe the contents of this block to ABC for it to do whitebox optimisations.

PR #1189, opened 2019-07-13
Error out if enable > dbits in memory_bram file

### LINKED ISSUE #1151: SIGFPE in memory_bram

Create a BRAM rules file where there are more enable bits than data bits and then invoke `memory_bram` with it.

## Expected behavior

Yosys catches this as an error and quits.

## Actual behavior

https://github.com/YosysHQ/yosys/blob/c5563b67c849e084015040c8b0d25d5f9489be43/passes/memory/memory_bram.cc#L592-L601

Here, `(bram.dbits / pi.enable)` evaluates to zero, resulting in `i % (bram.dbits / pi.enable)` performing division by zero, which crashes.

eddiehung:
Fixes #1151


ZirconiumX:
This looks good to me. Haven't tested it though.

PR #1191, opened 2019-07-13
Make opt_lut less chatty

whitequark:
The opt_lut pass, now enabled by default on iCE40, prints thousands of lines of debug messages. I think it predates the suppression of debug messages in e158ea209 so it didn't use that facility. Use it now.

### DIFF
--- a/passes/opt/opt_lut.cc
+++ b/passes/opt/opt_lut.cc
@@ -105,7 +105,7 @@ struct OptLutWorker
 				SigSpec lut_input = cell->getPort("\\A");
 				int lut_arity = 0;
 
-				log("Found $lut\\WIDTH=%d cell %s.%s.\n", lut_width, log_id(module), log_id(cell));
+				log_debug("Found $lut\\WIDTH=%d cell %s.%s.\n", lut_width, log_id(module), log_id(cell));
 				luts.insert(cell);
 
 				// First, find all dedicated logic we're connected to. This results in an overapproximation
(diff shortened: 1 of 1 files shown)


eddiehung:
I was just about to do something similar! +1

One gentle and tangential suggestion I have is for `opt_lut` (in fact, all passes should do this) to print a single line stating how many things it actually optimised.

whitequark:
But it does that:
```
Eliminated 1 LUTs.
Combined 14 LUTs.
```

eddiehung:
Oops. Sorry. I must have missed that! As you were.

passes/opt/opt_lut.cc
```diff
@@ -147,23 +147,23 @@ struct OptLutWorker
 					{
 						if (lut_width <= dlogic_conn.first)
 						{
-							log("  LUT has illegal connection to %s cell %s.%s.\n", lut_dlogic->type.c_str(), log_id(module), log_id(lut_dlogic));
-							log("    LUT input A[%d] not present.\n", dlogic_conn.first);
+							log_debug("  LUT has illegal connection to %s cell %s.%s.\n", lut_dlogic->type.c_str(), log_id(module), log_id(lut_dlogic));
```
eddiehung: Should something as serious sounding as an "illegal connection" be promoted to be a warning message?

whitequark: That just means it doesn't know which `SB_LUT4` is the `SB_CARRY` connected to. Ideally this entire chunk of logic would be removed in favor of `SB_CARRY_LUT4` cell; it's a hack.

PR #747, opened 2018-12-17
[wip] Travis -> Cirrus

q3k:
We replace Travis with Cirrus as Travis keeps giving us trouble, and Cirrus is generally a better alternative (notably, supports running in Docker containers for Linux, and supports Windows in addition to macOS).

PR #1194, opened 2019-07-14
Fix missing semicolon in Windows-specific code in aigerparse.cc.

cr1901:
Fixes a compile error on Windows, tested using `msys2`. No other platforms affected.

### DIFF
--- a/frontends/aiger/aigerparse.cc
+++ b/frontends/aiger/aigerparse.cc
@@ -1062,8 +1062,8 @@ struct AigerFrontend : public Frontend {
 		if (module_name.empty()) {
 #ifdef _WIN32
 			char fname[_MAX_FNAME];
-			_splitpath(filename.c_str(), NULL /* drive */, NULL /* dir */, fname, NULL /* ext */)
-				module_name = fname;
+			_splitpath(filename.c_str(), NULL /* drive */, NULL /* dir */, fname, NULL /* ext */);
+			module_name = fname;
 #else
 			char* bn = strdup(filename.c_str());
 			module_name = RTLIL::escape_id(bn);


frontends/aiger/aigerparse.cc
```diff
@@ -1062,7 +1062,7 @@ struct AigerFrontend : public Frontend {
 		if (module_name.empty()) {
 #ifdef _WIN32
 			char fname[_MAX_FNAME];
-			_splitpath(filename.c_str(), NULL /* drive */, NULL /* dir */, fname, NULL /* ext */)
+			_splitpath(filename.c_str(), NULL /* drive */, NULL /* dir */, fname, NULL /* ext */);
 				module_name = fname;
```
eddiehung: Oops. Looks like I did an extra tab here. Would you be able fix that too whilst you're in the region? Thanks!

eddiehung:
Thanks. Will merge when CI goes green, not that CI even builds on Windows anymore (I think it was having issues in the past; we should really fix the underlying cause not by disabling stuff!)

cr1901:
>(I think it was having issues in the past; we should really fix the underlying cause not by disabling stuff!)

I would fix it myself, but Travis currently [doesn't play nice](https://travis-ci.community/t/how-do-i-use-mingw-and-msys2-shells/1787) w/ `msys2`, and IME that is the only way to really get a `yosys` on Windows that supports everything (as opposed to mxe).

Maybe PR #747 would give me an opportunity to test this? On my radar is _also_ to make official packages for `msys2` for `yosys-git`, `nextpnr-git`, and `yices2` (`z3` is already there!).

eddiehung:
> Maybe PR #747 would give me an opportunity to test this?

My impression is that PR#747 doesn't has fallen off the radar and no-one is driving it forward anymore. Would you like to adopt it this poor orphan? We use Cirrus in [nextpnr](http://github.com/YosysHQ/nextpnr)...

PR #1198, opened 2019-07-16
proc_group: new pass

whitequark:
The proc_group pass canonicalizes left-hand side of assignments in
processes, ensuring that for any two assign actions, the left-hand
side either has same bits in same order, or no common bits at all.
This allows the following passes to consider assigns the smallest
trackable unit; for example it improves the precision of proc_prune.

The actual motivation for this pass is my work on replacement for
proc_mux where I would like to assume that each assignment LHS
can be directly turned into an output of a `$pmux` cell, but it seems
useful on its own as well.

### DIFF
--- a/passes/proc/Makefile.inc
+++ b/passes/proc/Makefile.inc
@@ -1,5 +1,6 @@
 
 OBJS += passes/proc/proc.o
+OBJS += passes/proc/proc_group.o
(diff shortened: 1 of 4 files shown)


whitequark:
I believe this pass is not necessary, as my tentative replacement for `proc_mux` now uses a flow analysis that is both more elegant and more powerful than this pass.

PR #1208, opened 2019-07-18
Assorted synth_intel cleanups from @bwidawsk

ZirconiumX:
I figured it'd be best to get these into the code base, as they're non-functional changes that are easy to review.

### DIFF
--- a/techlibs/intel/synth_intel.cc
+++ b/techlibs/intel/synth_intel.cc
@@ -166,20 +166,12 @@ struct SynthIntelPass : public ScriptPass {
 	void script() YS_OVERRIDE
 	{
 		if (check_label("begin")) {
-			if (check_label("family") && family_opt == "max10")
-				run("read_verilog -sv -lib +/intel/max10/cells_sim.v");
-			else if (check_label("family") && family_opt == "a10gx")
-				run("read_verilog -sv -lib +/intel/a10gx/cells_sim.v");
-			else if (check_label("family") && family_opt == "cyclonev")
-				run("read_verilog -sv -lib +/intel/cyclonev/cells_sim.v");
-			else if (check_label("family") && family_opt == "cyclone10")
-				run("read_verilog -sv -lib +/intel/cyclone10/cells_sim.v");
-			else if (check_label("family") && family_opt == "cycloneiv")
-				run("read_verilog -sv -lib +/intel/cycloneiv/cells_sim.v");
-			else if (check_label("family") && family_opt == "cycloneive")
-				run("read_verilog -sv -lib +/intel/cycloneive/cells_sim.v");
-			else
-				log_cmd_error("Invalid or not family specified: '%s'\n", family_opt.c_str());
+			string cmd = "read_verilog -sv -lib +/intel/FAMILY/cells_sim.v";
+			cmd.replace(cmd.find("FAMILY"), 6, family_opt);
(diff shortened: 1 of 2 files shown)


techlibs/intel/synth_intel.cc
```diff
@@ -147,8 +147,12 @@ struct SynthIntelPass : public ScriptPass {
...
 			log_cmd_error("This command only operates on fully selected designs!\n");
-		if (family_opt != "max10" && family_opt != "a10gx" && family_opt != "cyclonev" && family_opt != "cycloneiv" &&
-		    family_opt != "cycloneive" && family_opt != "cyclone10")
+		if (family_opt != "max10" &&
+		    family_opt != "a10gx" &&
+		    family_opt != "cyclonev" &&
+		    family_opt != "cycloneiv" &&
+		    family_opt != "cycloneive" &&
+		    family_opt != "cyclone10")
 			log_cmd_error("Invalid or not family specified: '%s'\n", family_opt.c_str());
```
daveshah1: Nitpick - while you are touching this code, can you fix "not family" to be "no family" here?

ZirconiumX: I found another instance of the same typo, which I also fixed.

techlibs/intel/synth_intel.cc
```diff
@@ -222,18 +220,10 @@ struct SynthIntelPass : public ScriptPass {
...
-			else if (family_opt == "cyclonev")
-				run("techmap -map +/intel/cyclonev/cells_map.v");
-			else if (family_opt == "cyclone10")
-				run("techmap -map +/intel/cyclone10/cells_map.v");
-			else if (family_opt == "cycloneiv")
-				run("techmap -map +/intel/cycloneiv/cells_map.v");
-			else
-				run("techmap -map +/intel/cycloneive/cells_map.v");
+			string cmd = "techmap -map +/intel/FAMILY/cells_map.v";
+			cmd.replace(cmd.find("FAMILY"), 6, family_opt);
```
daveshah1: It would be more Yosys-idiomatic to use `stringf` here, see things like https://github.com/YosysHQ/yosys/blob/master/techlibs/ice40/synth_ice40.cc#L340

ZirconiumX: Fixed.

techlibs/intel/synth_intel.cc
```diff
@@ -162,18 +166,12 @@ struct SynthIntelPass : public ScriptPass {
...
-			else if (check_label("family") && family_opt == "cyclonev")
-				run("read_verilog -sv -lib +/intel/cyclonev/cells_sim.v");
-			else if (check_label("family") && family_opt == "cyclone10")
-				run("read_verilog -sv -lib +/intel/cyclone10/cells_sim.v");
-			else if (check_label("family") && family_opt == "cycloneiv")
-				run("read_verilog -sv -lib +/intel/cycloneiv/cells_sim.v");
-			else
-				run("read_verilog -sv -lib +/intel/cycloneive/cells_sim.v");
+			string cmd = "read_verilog -sv -lib +/intel/FAMILY/cells_sim.v";
+			cmd.replace(cmd.find("FAMILY"), 6, family_opt);
```
daveshah1: Missed this the first time round, could you also use stringf here?

PR #1209, opened 2019-07-18
[WIP] synth xilinx renaming, as per #1184

eddiehung:
`synth_xilinx` is a spaghetti of intertwined features which are necessary for supporting some of the newer features like shift register and mux inference, which will take a little more thinking (help welcome) to separate into distinct labels.

Currently, and as a departure `synth_ecp5` and the like, flop mapping is performed after ABC so that it can retime flops if necessary. However, on the `abc9` roadmap is to communicate the existence of (and their concrete types), and so I'll be moving it back (plus `abc9` doesn't appear to support whitebox-aware retiming...).


CHANGELOG
```diff
@@ -14,6 +14,9 @@ Yosys 0.9 .. Yosys 0.9-dev
     - Added "synth -abc9" (experimental)
     - Added "script -scriptwire
     - "synth_xilinx" to now infer wide multiplexers (-widemux <min> to enable)
+    - Renamed labels/options in synth_ice40 (e.g. map_dram -> map_lutram; -nodram -> -nolutram)
+    - Renamed labels/options in synth_ecp5 (e.g. map_dram -> map_lutram; -nodram -> -nolutram)
+    - Renamed labels/options in synth_xilinx (e.g. map_dram -> map_lutram; -nodram -> -nolutram)
```
ZirconiumX: Ah, I didn't add notes for synth_intel, which got a similar update in #1207. Would you mind adding that?

eddiehung: Thanks for noticing! Added in 280288a6

ZirconiumX:
Bump; does this need somebody to review before merging?

eddiehung:
Sorry, from my PoV this is not quite finished yet as I think we should make the labels consistent with the other `synth_*` passes, but due to its spaghetti nature it's a little harder to divide things up nicely. Feel free to propose an arrangement.

PR #1213, opened 2019-07-19
wreduce/opt_expr: improve width reduction for $add and $sub cells

eddiehung:
`wreduce` now trims the MSBs of `$sub`:
`assign o = (j[7:0] >> 4) - i[3:0]` now `assign o[3:0] = j[7:4] - i[3:0], o[7:4] = {4{o[3]}}`

`opt_expr -fine` now trims the LSBs of `$add`/`$sub`:
`assign o = (i[3:0] << 4) + j[7:0]` now `assign o[7:4] = i[3:0] + j[7:4], o[3:0] = j[3:0]`
`assign o = j[7:0] - (i[3:0] << 4)` now `assign o[7:4] = j[7:4] - i[3:0], o[3:0] = j[3:0]`


kernel/rtlil.cc
```diff
@@ -3353,7 +3353,7 @@ RTLIL::SigSpec RTLIL::SigSpec::extract(int offset, int length) const
 {
 	unpack();
 	cover("kernel.rtlil.sigspec.extract_pos");
-	return std::vector<RTLIL::SigBit>(bits_.begin() + offset, bits_.begin() + offset + length);
+	return std::vector<RTLIL::SigBit>(bits_.begin() + offset, length >= 0 ? bits_.begin() + offset + length : bits_.end() + length + 1);
```
eddiehung: Consider `extract_end()` or similiar.

passes/opt/wreduce.cc
```diff
@@ -342,17 +342,17 @@ struct WreduceWorker
 			}
 		}
 
-		if (cell->type.in("$pos", "$add", "$mul", "$and", "$or", "$xor"))
+		if (cell->type.in("$pos", "$add", "$mul", "$and", "$or", "$xor", "$sub"))
 		{
-			bool is_signed = cell->getParam("\\A_SIGNED").as_bool();
+			bool is_signed = cell->getParam("\\A_SIGNED").as_bool() || cell->type == "$sub";
```
eddiehung: Why?

passes/opt/wreduce.cc
```diff
@@ -365,14 +365,36 @@ struct WreduceWorker
 			}
 		}
 
+		if (cell->type.in("$add", "$sub")) {
```
eddiehung: Move to `opt_expr`?

passes/opt/wreduce.cc
```diff
@@ -365,14 +365,36 @@ struct WreduceWorker
...
+		}
+
 		if (GetSize(sig) == 0) {
 			log("Removed cell %s.%s (%s).\n", log_id(module), log_id(cell), log_id(cell->type));
 			module->remove(cell);
 			return;
 		}
 
 		if (bits_removed) {
-			log("Removed top %d bits (of %d) from port Y of cell %s.%s (%s).\n",
```
eddiehung: Restore

passes/opt/wreduce.cc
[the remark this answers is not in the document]
eddiehung: Without this, borrow extension doesn't occur correctly since: https://github.com/YosysHQ/yosys/blob/bfc7164af7bf64cb2fe5d00e87bbfead841a4dc2/passes/opt/wreduce.cc#L362

Otherwise proof fails with:
```
a = 4'd8;
b = 8'd0;
assign o = (b >> 4) - a;
```
Pre-wreduce: o = 9'd111111000
Post-wreduce (without `|| cell->type == "sub"`): o = 9'b000011000 where the 4 MSBs that were trimmed were fixed to zero instead of extending.

whitequark:
Looks good (and very useful) to me, but I didn't look at opt_expr a lot. Do you have any before/after LUT counts?

eddiehung:
> Looks good (and very useful) to me, but I didn't look at opt_expr a lot. Do you have any before/after LUT counts?

Here's an example from the test-case:
```
module opt_expr_add_test(input [3:0] i, input [7:0] j, output [8:0] o);
    assign o = (i << 4) + j;
endmodule
```
Prior to this PR: `./yosys -p "proc; opt -full" test.v -b ilang` gives:
```
  cell $add $add$test.v:2$2
    parameter \A_SIGNED 0
    parameter \A_WIDTH 9
    parameter \B_SIGNED 0
    parameter \B_WIDTH 8
    parameter \Y_WIDTH 9
    connect \A { 1'0 \i 4'0000 }
    connect \B \j
    connect \Y \o
  end
```
which doesn't take the opportunity to optimise into a 4-bit `$add`.
`synth_ice40` gives:
```
   Number of cells:                  8
     SB_CARRY                        4
     SB_LUT4                         4
```
likely due to `ice40_opt`.
On the other hand, `synth_xilinx` (which doesn't have an equivalent `xilinx_opt`) gives:
```
   Number of cells:                 25
     LUT2                            8
     MUXCY                           8
     XORCY                           9
```

The original motivation of this PR was to trim word-level adders early in the flow (even if they get optimised downstream by whitebox passes or by the techmapper) in order to aid DSP inference.

PR #1219, opened 2019-07-24
made ObjectIterator comply with Iterator Interface

jakobwenzel:
this makes it possible to use std algorithms on them

### DIFF
--- a/kernel/rtlil.h
+++ b/kernel/rtlil.h
@@ -420,7 +420,11 @@ namespace RTLIL
 	// It maintains a reference counter that is used to make sure that the container is not modified while being iterated over.
 
 	template<typename T>
-	struct ObjIterator
+	struct ObjIterator : public std::iterator<std::forward_iterator_tag,
+		T,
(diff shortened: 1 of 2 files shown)


ZirconiumX:
I don't think iterators work this way. The C++ standard library (before concepts, anyway) does not require iterators to inherit from std::iterator, only that they comply with the interface that the standard requires.

Additionally, std::iterator was deprecated in C++17, [this article](https://www.fluentcpp.com/2018/05/08/std-iterator-deprecated/) explains why.

jakobwenzel:
>The C++ standard library (before concepts, anyway) does not require iterators to inherit from std::iterator, only that they comply with the interface that the standard requires.
True, but ObjectIterator did not.

> Additionally, std::iterator was deprecated in C++17, [this article](https://www.fluentcpp.com/2018/05/08/std-iterator-deprecated/) explains why.
Thanks, I was not aware of that deprecation. I updated the PR to remove std::iterator.

PR #1222, opened 2019-07-24
Add a simple example for Spartan 6

koriakin:


examples/mimas2/run_yosys.ys
```diff
@@ -0,0 +1,4 @@
+read_verilog example.v
+synth_xilinx -top example -family xc6s
+iopadmap -bits -outpad OBUF I:O -inpad IBUF O:I
+write_edif -pvector bra example.ndf
```
eddiehung: Why `.ndf` file extension and not `.edif`?

koriakin: Good point, fixed.

examples/mimas2/run_yosys.ys
```diff
@@ -0,0 +1,4 @@
+read_verilog example.v
+synth_xilinx -top example -family xc6s
+iopadmap -bits -outpad OBUF I:O -inpad IBUF O:I
```
eddiehung: Perhaps I was too hasty to merge it in. Would it be useful to have this `iopadmap` command be part of `synth_xilinx -family xc6s`?

koriakin: To be discussed in #1235

PR #1226, opened 2019-07-26
Add support for gzip'd input files

daveshah1:
As discussed with @eddiehung, this should make adding larger benchmarks to yosys-bench neater.

Things I would like to check
 - This enables decompression based on gzip magic. I could change this to be based on filename if you prefer.
 - This is currently enabled by default. I think zlib is pretty much omnipresent and unproblematic, but I can change this to disabled-by-default if this is a problem.


mithro:
Any chance you are you able to add support for writing compressed files out too?

daveshah1:
Yes, I could definitely add that too. That couldn't auto detect based on magic, so I suspect the best approach would be to enable gzip based on output filename extension, otherwise we'd need a manual gzip flag for each backend. Unless you have any other thoughts how to enable this?

udif:
Just adding my 2 cents, processing gzip'ed files by the tool is standard behavior, at least for the Cadence toolset. I don't know if it is merely based on the file extension or by reading any magic bytes from the beginning of the file, but if you feed the tool with a proper gzip'ed file, it will accept it and uncompress it on-the-fly. This is really useful for large netlist files.

kernel/register.cc
```diff
@@ -25,6 +25,26 @@
...
+PRIVATE_NAMESPACE_BEGIN
+#define GZ_BUFFER_SIZE 8192
+void decompress_gzip(const std::string &filename, std::stringstream &out)
+{
+	char buffer[GZ_BUFFER_SIZE];
+	int bytes_read;
+	gzFile gzf = gzopen(filename.c_str(), "rb");
+	while(!gzeof(gzf)) {
+		bytes_read = gzread(gzf, reinterpret_cast<void *>(buffer), GZ_BUFFER_SIZE);
+		out.write(buffer, bytes_read);
```
eddiehung: Would prefer to not have to copy the entire file like this, but I had a quick look and couldn't see other way other than using boost or rolling your own implementation.

eddiehung:
> Yes, I could definitely add that too. That couldn't auto detect based on magic, so I suspect the best approach would be to enable gzip based on output filename extension, otherwise we'd need a manual gzip flag for each backend. Unless you have any other thoughts how to enable this?

I agree that's the most straightforward way -- if a backend ends in `*.gz` then gzip it.

kernel/register.cc
daveshah1: Yes, as far as I can tell this would require much more work for a memory gain likely to be fairly small in terms of total peak memory throughout runtime. Even with Boost, I think some hacking would be needed to manage destruction of both the decompresser and the underlying file correctly. We can look at this in the future again if it proves to be an issue.

Issue #299, opened 2017-01-21
Adder-Substractor inference


drom:
Adder-Subtractor inference is one of common transformations done by synthesis tools.

https://en.wikipedia.org/wiki/Adder%E2%80%93subtractor

From this Verilog:
```verilog
assign res = sel
  ? (a - b)
  : (a + b);
```
Full version: https://github.com/drom/icicles/blob/master/src/add_sub.v

Synthesizer can infer something like this:

```verilog
assign sum = {a, sel} + {(sel ? ~b : b), 1'b1};
assign res = sum[W:1];
```
Full version: https://github.com/drom/icicles/blob/master/src/add_sub_opt.v

Second version requires ~2x less LUTs and faster too.

Does such optimizations belongs to Yosys project?

cliffordwolf:
The following command will map your `add_sub.v` to 15 `SB_CARRY` and 32 `SB_LUT4` cells:

```
yosys -p 'proc; opt; alumacc; share -aggressive; synth_ice40' add_sub.v
```

To do this out-of-the-box I will need to write a resource sharing pass that is less universal (but faster) than the `share` pass. That new pass would only be looking at cells driving the same mux (tree). This is on my todo list (proposed name: `opt_share`, this is the reason why I renamed the old `opt_share` to `opt_merge` in March 2016).

I'll keep this issue open until I'll find the time to write that new pass.

drom:
LGTM

akashlevy:
This appears to be resolved by https://github.com/YosysHQ/yosys/pull/1303

widlarizer:
Yes, alumacc and opt_share are doing a good job already

PR #1227, opened 2019-07-26
Implement opt_share

bogdanvuk:
Hi, I needed the arithmetic operator sharing optimization, and stumbled upon https://github.com/YosysHQ/yosys/issues/299, so I thought I'd give it a shot.

This pass identifies arithmetic operators that share an operand and whose 
results are used in mutually exclusive cases controlled by a multiplexer, and
merges them together by multiplexing the other operands

Following features have been implemented

- only $mux is considered for mergers, hence "pmuxtree" command is needed before "opt_share" to handle $pmux
- only $add and $sub are handled.
- the pass checks whether operator outputs are exclusively used by the same $mux. If they are also connected elsewhere it does not perform the merge.
- if operator outputs are concatenated with other signals before connected to $mux, "opt_share" can extract them from the concatenation and still merge them.
- "opt_share" takes care of which signals are connected to the negative port of the $sub operator, and inverts them when merging with with $add
- operands that are multiplexed for the merged operator are sign extended when they are of different sizes


cliffordwolf:
> Reimplement opt_share to work on $alu and $pmux

Much better. Thank you! Is there any reason why it would be hard to make it work with just any combinatorial cell, not just $alu? Also, any thoughts about sharing two cells with (slightly) different widths?

passes/opt/opt_share.cc
```diff
@@ -0,0 +1,459 @@
...
+
+struct OptRmdffPass : public Pass {
+	OptRmdffPass() : Pass("opt_share", "merge arithmetic operators that share an operand") {}
+	void help() YS_OVERRIDE
+	{
+		//   |---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|
+		log("\n");
+		log("    opt_share [selection]\n");
+		log("\n");
+		log("This pass identifies arithmetic operators that share an operand and whose\n");
```
eddiehung: Suggestion -- be more explicit:

This pass identifies mutually exclusive `$alu` arithmetic cells that:
(a) share an input operand
(b) drive the same `$mux`, `$_MUX_`, or `$pmux` multiplexing cell
allowing the `$alu` cell to be merged and the multiplexer to be moved 
from multiplexing its output to multiplexing the non-shared input operands.

bogdanvuk:
Thanks @eddiehung, I implemented your suggestions.

Can you please explain what is a "fixed-point iteration style"?

**Here are my thoughts regarding the opt_share complexity**:

For each $alu that is connected to a MUX, opt_share makes a set of other $alu-s connected to the rest of MUX's input ports. It then needs to check with how many other $alu-s from the set the first $alu shares an operand. The following is then performed for each operand of the first $alu: the set of all users of the operand is retrieved and then it is checked how many of the other $alu-s from the set are in the same time users of the operand.

The number of checks needed can be made to be either the number of operand users or the number of other $alu-s in the MUX set. Say we use the second approach. The number of $alu-s in the MUX set is less then or equal to the number of MUX inputs, say **N_mux_s**. Worst case scenario for this procedure is if all $alu-s in the module are connected to the different inputs of the same MUX, so If the number of $alu-s in the modules equals **N_alu**, this procedure has a complexity of:

O(**N_mux_s** * **N_alu**)

In order to build various other helper structures, and to check which $alu-s are connected to multiple users, each bit of the module netlist needs to be traversed. If there are **N_bits** bits in the netlist, I guess the final complexity is :

O(opt_share) = O(**N_mux_s** * **N_alu** + **N_bits**)

bogdanvuk:
@cliffordwolf With the latest commit I supported most of the binary operators: all bitwise, relational, reduction, logical, shift operators and most arithmetic operators. Handling of $alu is still limited to when its Y output is used exclusively and might need to be revisited later.

I tried including opt_share as a part of the opt step (I didn't commit this), and after fixing fiew issues, all tests passed. So it looks like opt_share is conservative enough.

It looks that now arbitrary combinatorial 2 input 1 output cells could be easily supported when I figure out how to determine whether a cell is combinatorial (I am thinking about user defined modules). Is there a straight forward way of doing this?

Regarding cell widths, right now the muxed operands can have different widths and are automatically extended. I'll have to think about some more if and when a similar thing can be done for the shared operand and the cell output.

passes/opt/opt_share.cc
```diff
@@ -0,0 +1,661 @@
...
+	void help() YS_OVERRIDE
+	{
+		//   |---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|
+		log("\n");
+		log("    opt_share [selection]\n");
+		log("\n");
+
+		log("This pass identifies mutually exclusive cells of the same type that:\n");
+		log("    (a) share an input signal\n");
+		log("    (b) drive the same $mux, $_MUX_, or $pmux multiplexing cell allowing\n");
```
eddiehung: Sorry I wasn't clear. I was trying to suggest that "allowing the cell to be merged..." does not belong inside the list and should be on the next line with no indent.

bogdanvuk: Hope I got it right this time :)

bogdanvuk:
@eddiehung Regarding the fixed point run, I did observe that opt_share found mergers on the second pass of the opt while loop for a larger design I have. opt_share does not remove the output $mux that is pushed back, but only rewires its inputs. This is done because sometimes only a portion of inputs of the $mux can be merged. So, opt_clean does the cleanup for the opt_share. This means that after the cleanup it might be worth it to check whether new opportunities for the merge arose.

passes/opt/opt_share.cc
eddiehung: I don't see a change yet?

eddiehung:
> @eddiehung Regarding the fixed point run, I did observe that opt_share found mergers on the second pass of the opt while loop for a larger design I have. opt_share does not remove the output $mux that is pushed back, but only rewires its inputs. This is done because sometimes only a portion of inputs of the $mux can be merged. So, opt_clean does the cleanup for the opt_share. This means that after the cleanup it might be worth it to check whether new opportunities for the merge arose.

As a point of record, can you share the results on this design (and the design if possible) for:
* Without `opt_share`
* A single `opt_share` run
* `opt_share` inside `opt` loop?
Thanks!

Let's also hear @cliffordwolf's opinion about whether we should have this inside the `opt` loop.

eddiehung:
After discussing with @cliffordwolf, we think it's reasonable to add `opt_share` as part of `opt -full`.

PR #1232, opened 2019-07-29
Add support for writing gzip-compressed files

daveshah1:
gzip compression is automatically enabled when an output file has the extension ".gz".

cc @mithro


daveshah1:
That does seem the best way forward. Implemented in 74a9333

PR #1233, opened 2019-07-29
Call "read_verilog" with -defer from "read"

cliffordwolf:
Small code change, big behavior change. This makes non-verific Yosys behave a bit more like verific-yosys.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,17 +130,14 @@ commands and ``help <command>`` to print details on the specified command:
 
 	yosys> help help
 
-reading the design using the Verilog frontend:
+reading and elaborating the design using the Verilog frontend:
 
-	yosys> read_verilog tests/simple/fiedler-cooley.v
+	yosys> read -sv tests/simple/fiedler-cooley.v
+	yosys> hierarchy -top up3down5
 
 writing the design to the console in Yosys's internal format:
 
 	yosys> write_ilang
-
-elaborate design hierarchy:
-
-	yosys> hierarchy
 
 convert processes (``always`` blocks) to netlist elements and perform
 some simple optimizations:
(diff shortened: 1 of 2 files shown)


daveshah1:
I am happy with the improved consistency this brings. But I'm wondering if we should have a warning if one attempts to run `check` on a design with only abstract modules, to explain why people are suddenly seeing empty designs and thus reduce confusion from the change (even if `read` isn't widely used compared to `read_verilog`)

README.md
```diff
@@ -130,18 +130,15 @@ commands and ``help <command>`` to print details on the specified command:
 
 	yosys> help help
 
-reading the design using the Verilog frontend:
+reading and elaborating the design using the Verilog frontend:
 
-	yosys> read_verilog tests/simple/fiedler-cooley.v
+	yosys> read -sv tests/simple/fiedler-cooley.v
```
eddiehung: Is `-sv` necessary?

eddiehung:
In general, I'm in two minds about `-defer`. On one hand, it should be default since there is very utility for Yosys to have a non-elaborated design, but on the other, we can run into issues when the default parameters of certain modules are absurd (think ASIC memories) or into simulation cells that only support certain modes -- e.g. https://github.com/YosysHQ/yosys/blob/ca94c2d3c4785c45a2fefdb659e9ff94f2f8c7b3/techlibs/xilinx/cells_sim.v#L465-L494
that I had to wrap behind an `ifdef`.
On a maybe related note, when removing this `ifdef` just now, I'm getting:
```
share/xilinx/cells_sim.v:466: ERROR: Can't resolve task name `$fatal'
```

daveshah1:
> On a maybe related note, when removing this ifdef just now, I'm getting:
  ```share/xilinx/cells_sim.v:466: ERROR: Can't resolve task name `$fatal'```

I think Yosys only supports `$fatal` etc outside of initial, in the main module body.

README.md
cliffordwolf: You can also use `-vlog95` or `-vlog2k` or `-sv2005` etc, but `read` does expect an option indicating the file type as first argument.

PR #1239, opened 2019-07-31
Fix formatting for msys2 mingw build

mmicko:
- msys2 builds anyway can not be compiled with plugins so made config so compile work by default
- %zu is not supported in msys mingw, mxe and VisualStudio older then 2015 so made define for size_t formatting
- update places where wrong formatting have been used for size_t values
- fixed cast and not defined symbol in launcher.c

TODO: check mxe build after these changes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -869,9 +869,11 @@ config-mxe: clean
 
 config-msys2: clean
 	echo 'CONFIG := msys2' > Makefile.conf
+	echo 'ENABLE_PLUGINS := 0' >> Makefile.conf
 
 config-msys2-64: clean
 	echo 'CONFIG := msys2-64' > Makefile.conf
+	echo 'ENABLE_PLUGINS := 0' >> Makefile.conf
 
 config-cygwin: clean
 	echo 'CONFIG := cygwin' > Makefile.conf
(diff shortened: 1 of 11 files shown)


mmicko:
Added fixes to enable MXE builds. Used https://github.com/mmicko/mxe-docker as environment

cliffordwolf:
I think instead of `SIZE_T_FMT` we should just use `%d` and cast the `size_t` argument to `int`. That's what we do in all the other cases where we print a `size_t` (those few instances this PR touches are relatively new code).

cliffordwolf:
Also, instead of `x.size()` one should generally use `GetSize(x)` in Yosys code anyways. That already returns an `int`. (Yes, I know we consistently use size_t for pass args, but that's just because this pattern is older than `GetSize()`.)

cliffordwolf:
I did cherry pick 73dcf31862182a5dd5f45e684621c27f7adc5d0f into 0.9-rc now.

mmicko:
@cliffordwolf Right, most is code that came after previous release. Will take care and convert code to use GetSize  so %d is then fine

mmicko:
Cleaned a bit pull request. 
Last two 3f633690ae3d0c8753866d136851b8f0129cb643 and 7a65ed19a5921ff219fd208d2712e09b5e5894dd are needed for latest MXE version build. If you could send me version you have deployed just to confirm that it works without this on old one.

PR #1241, opened 2019-08-01
Improved JSON attr/param encoding

cliffordwolf:
Quoting new format description:

> Numeric 32-bit parameter and attribute values are written as decimal values.
Bit verctors of different sizes, or ones containing 'x' or 'z' bits, are written
as string holding the binary representation of the value. Strings are written
as strings, with an appended blank in cases of strings of the form /[01xz]* */.

### DIFF
--- a/backends/json/json.cc
+++ b/backends/json/json.cc
@@ -83,20 +83,43 @@ struct JsonWriter
 		return str + " ]";
 	}
 
+	void write_parameter_value(const Const &value)
+	{
+		if ((value.flags & RTLIL::ConstFlags::CONST_FLAG_STRING) != 0) {
+			string str = value.decode_string();
+			int state = 0;
+			for (char c : str) {
+				if (state == 0) {
+					if (c == '0' || c == '1' || c == 'x' || c == 'z')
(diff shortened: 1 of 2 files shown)


daveshah1:
I agree with this change, however, some major restructuring will be needed in nextpnr which assumes anything <32 bits will be written as an integer at the moment - and this will need to be done in a way that is backwards compatible too. Let me think about this a bit more.

cliffordwolf:
> in nextpnr which assumes anything <32 bits will be written as an integer at the moment

That sounds like a bug in any case. I'd say simply adding MSB zero bits to a parameter probably shouldn't break a cell instantiation. This sound like padding a parameter until it's >32 bits will break in the current nextpnr implementation.

daveshah1:
Yes. I am working on replacing the current infrastructure in nextpnr with RTLIL::Const so we can deal with all these cases and round-trip JSON correctly.

daveshah1:
Currently working on this in YosysHQ/nextpnr#306. This will probably take a couple of days to finish & regression test, can we hold off merging this until then?

PR #1250, opened 2019-08-05
techlibs/intel: Clean up Makefile

bwidawsk:
Use gmake's foreach iterator and remove nonexistent files. Gmake is
already a requirement of the build system.

Signed-off-by: Ben Widawsky <ben@bwidawsk.net>

@dh73 
@ZirconiumX


ZirconiumX:
Build failure appears to be spurious?

PR #1255, opened 2019-08-06
Suppress some compiler warnings occurring when applying NDEBUG

eddiehung:
Specifically, during `NDEBUG`:

* aigerparse.cc's log_debug2 now calls `if (0) log_debug(__VA_ARGS__);` (where `log_debug` is a NOOP) in order to suppress unused warnings.
* `log_debug` now calls a `constexpr` function that always returns false and thus will never print any debug message
* `log_assert` now explicitly ignores the expression using `(void)(expr);`

### DIFF
--- a/kernel/log.h
+++ b/kernel/log.h
@@ -89,11 +89,10 @@ YS_NORETURN void log_cmd_error(const char *format, ...) YS_ATTRIBUTE(format(prin
 
 #ifndef NDEBUG
 static inline bool ys_debug(int n = 0) { if (log_force_debug) return true; log_debug_suppressed += n; return false; }
-#  define log_debug(...) do { if (ys_debug(1)) log(__VA_ARGS__); } while (0)
 #else
-static inline bool ys_debug(int n = 0) { (void) n; return false; }
-#  define log_debug(_fmt, ...) do { } while (0)
+static constexpr inline bool ys_debug(int /*n*/ = 0) { return false; }
 #endif
+#define log_debug(...) do { if (ys_debug(1)) log(__VA_ARGS__); } while (0)
 
 static inline void log_suppressed() {
 	if (log_debug_suppressed && !log_make_debug) {
--- a/kernel/log.h
+++ b/kernel/log.h
@@ -91,7 +91,7 @@ YS_NORETURN void log_cmd_error(const char *format, ...) YS_ATTRIBUTE(format(prin
 static inline bool ys_debug(int n = 0) { if (log_force_debug) return true; log_debug_suppressed += n; return false; }
 #  define log_debug(...) do { if (ys_debug(1)) log(__VA_ARGS__); } while (0)
 #else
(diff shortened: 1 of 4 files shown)


kernel/log.h
```diff
@@ -154,7 +153,7 @@ static inline void log_assert_worker(bool cond, const char *expr, const char *fi
 }
 #  define log_assert(_assert_expr_) YOSYS_NAMESPACE_PREFIX log_assert_worker(_assert_expr_, #_assert_expr_, __FILE__, __LINE__)
 #else
-#  define log_assert(_assert_expr_)
+#  define log_assert(_assert_expr_) (void)(_assert_expr_);
```
cliffordwolf: If `_assert_expr_` has side effects, or the compiler can't tell that it has none, it is still being executed. We explicitly do not want that, as `_assert_expr_` may do expensive things like looking up IdStrings. Better would be for example `do { if (0) ... }`.

kernel/macc.h
```diff
@@ -107,10 +107,8 @@ struct Macc
 		std::vector<RTLIL::State> config_bits = cell->getParam("\\CONFIG").bits;
 		int config_cursor = 0;
 
-#ifndef NDEBUG
 		int config_width = cell->getParam("\\CONFIG_WIDTH").as_int();
 		log_assert(GetSize(config_bits) >= config_width);
-#endif
```
cliffordwolf: This would slow down NDEBUG builds because now it's performing unnecessary operations, such as looking up the `CONFIG_WIDTH` IdString.

eddiehung: Correct. I had to make this change because, now that we reference the `log_assert()` expression, it fails when `NDEBUG` is enabled.
I am assuming that a optimising compiler (particularly one where `NDEBUG` is enabled) would be able to figure out that the function has no side effects and its return value is discarded, and optimise the whole thing away. I am aware that this isn't guaranteed though...

The only two alternatives I can think of are:
- wrap `log_assert()`s using `config_width` behind `NDEBUG` too
- to not cache `config_width` and call `getParam` inside every `log_assert()`

unless you can think of a better way?

kernel/log.h
cliffordwolf: Why was this marked "resolved"? I haven't seen this issue addressed.

eddiehung: Oops sorry. I forgot to push.

cliffordwolf: Ah.  :+1:

kernel/macc.h
cliffordwolf: `"\\CONFIG_WIDTH"` is being converted to an IdString. That alone should have enough side effect for the compiler to keep this in. Maybe instead of changing the debug macros, a better solution would be to `#ifndef NDEBUG` the variables that are in fact unused with `NDEBUG`, as happened here originally, and in ilang_backend.cc.

PR #1258, opened 2019-08-06
Cleanup a few barnacles across codebase

eddiehung:
Mainly because they were an eyesore for me...

- Use `IdString::in()` instead of `IdString::operator==() || IdString::operator==() || ...`
- Use `State::S{0,1}` instead of `Const({0,1}, 1)` or `RTLIL::S{0,1}`
- Use `IdString::begins_with()` instead of `IdString::substr(0, ...) == `
- Use `std::string::compare()` instead of `std::string::substr() == `
- `make clean` to not trigger rebuilding `abc`


frontends/verific/verific.cc
```diff
@@ -2244,7 +2244,7 @@ struct VerificPass : public Pass {
 					continue;
 				}
 				if (args[argidx] == "-L" && argidx+1 < GetSize(args)) {
-					verific_sva_fsm_limit = atoi(args[++argidx].c_str());
+					verific_sva_fsm_limit = std::stoi(args[++argidx]);
```
cliffordwolf: This can throw an exception that we do not catch (instead of returning zero). Is this really the behavior we want?

eddiehung: Ack. Good point. I didn't realise `stoi` could raise an exception (which is arguably better than silently returning zero) but let's not get into that rabbit hole right now.

PR #1261, opened 2019-08-07
Automatically prune init attributes in verific front-end

### LINKED ISSUE #1237: VHDL Initial values not ignored when also set in combinatorial process

Consider the following VHDL test case:

```vhdl
library ieee;
use ieee.std_logic_1164.all;

entity tst is
  port (
    i_clk    : In   std_logic;
    i_val    : In   std_logic;
    o_val    : out  std_logic);
end entity tst;

architecture implementation of tst is
	signal	w_val : std_logic := '0';
	signal	r_val : std_logic := '1';
begin

	w_val <= r_val;

	process(i_clk)
	begin
		if (i_clk'event and i_clk = '1') then
			r_val <= i_val;
		end if;
	end process;

	o_val <= w_val;

end architecture;
```

processed with the following SBY script,

```
[options]
mode prove
depth 5

[engines]
smtbmc

[script]
read -vhdl tst.vhd
prep -top tst

[files]
tst.vhd
```

Notice in this construct that contradictory initial values are given to w_val.

## Expected behavior

I understand this is a common VHDL coding style.  Looking through a popular vendor's code, I'm seeing many examples of it.  I am not expecting this to generate a warmup failure, but rather for the initial value to be ignored when overridden by a combinatorial assignment.

## Actual behavior

Warmup failure.

cliffordwolf:
Fixes #1237

### DIFF
--- a/frontends/verific/verific.cc
+++ b/frontends/verific/verific.cc
@@ -19,6 +19,7 @@
 
 #include "kernel/yosys.h"
 #include "kernel/sigtools.h"
+#include "kernel/celltypes.h"
 #include "kernel/log.h"
 #include <stdlib.h>
 #include <stdio.h>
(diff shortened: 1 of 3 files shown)


daveshah1:
Don't have a quick way to test this, but looking at the implementation of the fix I don't think it would handle this case?

```vhdl
library ieee;
use ieee.std_logic_1164.all;

entity tst is
  port (
    o_val    : out  std_logic);
end entity tst;

architecture implementation of tst is
	signal	r_val : std_logic := '1';
begin
	r_val <= '0';
	o_val <= r_val;
end architecture;
```

cliffordwolf:
> Don't have a quick way to test this, but looking at the implementation of the fix I don't think it would handle this case?

It handles this case just fine. For a value driven with it's init value, verific produces a constant driver and an init attribute.

Even if verific would not produce the constant driver but leave the net floating, this commit would not change the behavior, because we only remove init attributes from wires that have a non-FF driving cell.

daveshah1:
Thanks for checking!

PR #1266, opened 2019-08-07
Wrap SB_LUT+SB_CARRY into $__ICE40_CARRY_WRAPPER

### LINKED ISSUE #1205: Use $__ICE40_FULL_ADDER throughout synth_ice40
ice40 has a special carry structure that taps the input of a LUTs, rather than the output:
image
(from [Lattice datasheet](http://www.latticesemi.com/~/media/LatticeSemi/Documents/DataSheets/iCE/iCE40LPHXFamilyDataSheet.pdf])) thus requiring the LUT input ordering to be preserved.

Currently, in non `abc9` flows this done by instantiating a directly `SB_CARRY` and `SB_LUT` cell for each adder bit. Then during techmapping, these cells are treated as regular module instantiations and are transformed into PI/POs and thus not touched by the techmapper.

After techmapping, `ice40_unlut` then converts those `SB_LUT` cells back into Yosys internal cells `$lut` and `opt_lut` then attempts to merge `$lut` cells together while preserving the input order for those attached to an `SB_CARRY`.

In #1186, a new `$__ICE40_FULL_ADDER` cell (encapsulating the `SB_CARRY` and `SB_LUT`) was instantiated instead for `abc9` flows, necessary because `abc9` (a) now has knowledge of the logic and delay through these cells; and (b) the has the capability to optimise either of them away independently. Thus, there was a need to wrap them together as one all-or-nothing atomic unit for optimisation.

As suggested by @whitequark [here](https://github.com/YosysHQ/yosys/pull/1186#issuecomment-511519300) it may be effective to utilise `$__ICE40_FULL_ADDER` beyond `abc9`.

I don't believe it will make any difference to `abc` (or any other techmappers in their current form) but it could help clean up some of the code surrounding `ice40_opt`, `ice40_unlut`, `opt_lut`, etc.

Also necessary is a way to identify `SB_CARRY` and `SB_LUT` patterns that are instantiated by the user (rather than inferred from behavioural code) so that they also get the same treatment.

eddiehung:
Fixes #1205 

* Renamed `$__ICE40_FULL_ADDER` to `$__ICE40_CARRY_WRAPPER` to capture the fact that the LUT it wraps may not be (just) computing the sum (i.e. when manually instantiated), and thus we should preserve its I0 and I3 inputs, as well as the LUT mask
* `arith_map.v` to techmap into `$__ICE40_CARRY_WRAPPER` regardless of `abc9`
* Removed `ice40_unlut`
* Added pattern matcher pass `ice40_wrapcarry` to identify and encapsulate manually instantiated `SB_CARRY` and `SB_LUT4` pairs into `$__ICE40_CARRY_WRAPPER`.


eddiehung:
@whitequark Did you have other optimisations in mind, now that we're encapsulating?

whitequark:
> Did you have other optimisations in mind, now that we're encapsulating?

I think so, but I don't recall exactly and I'm not feeling well right now. Regardless, I looked over the structure of the pass and this is a very good improvement, so once I get back to this it should not be a problem to add any optimizations I had in mind.

whitequark:
Before merging, please make sure that opt_lut did not regress the optimization it was designed for: folding muxes into adders.

eddiehung:
Thanks for the tip @whitequark. Do I understand you correctly in saying that there is no test at the moment that checks that `opt_lut` doesn't touch specific LUTs?

whitequark:
> Do I understand you correctly in saying that there is no test at the moment that checks that `opt_lut` doesn't touch specific LUTs?

There is a test (`tests/opt/opt_lut_elim`) that checks that `opt_lut` can combine `$lut` cells, and a test (`tests/opt/opt_lut`) that checks that `opt_lut` does not break a design. But there is no test that positively guarantees that a mux will be folded into an adder, nor a test that tests the `-dlogic` option. 

It makes sense to write the first test, but probably not the second; instead, it makes more sense to remove the `-dlogic` option entirely and feed `opt_lut` a netlist that contains `$__ICE40_CARRY_WRAPPER` cells, as this would actually be easier and better in long term.

That said, when writing the previous comment, I misunderstood the way you updated `synth_ice40` in this PR. I suspected it will cause a regression but I now see that it will not.

eddiehung:
> `tests/opt/opt_lut`

I have lightly modified this test to change it to do an `equiv_opt` on `synth_ice40`, which now does `-relut` by default. However, as I understand, this test didn't actually check that `-dlogic` inputs were not modified, right, which would bring us to no worse than before?
I was also thinking of writing some sort of `select` incantation that makes sure we have specific LUTs preserved by `-dlogic`... but it's not very elegant. 
Alternatively, we should have something like an `ice40_check` which checks for this sort of arch-specific intricacies.

I also haven't changed the previous test `opt_lut_elim` and it continues to pass. Would this be sufficient?

> It makes sense to write the first test, but probably not the second; instead, it makes more sense to remove the -dlogic option entirely and feed opt_lut a netlist that contains $__ICE40_CARRY_WRAPPER cells, as this would actually be easier and better in long term.

Sorry I'm not following. Are you saying we should have a `$__ICE40_CARRY_WRAPPER` specific optimisation pass (which would make it even more niche than `opt_lut`) or are you saying we should generic "whitebox" optimisation?

whitequark:
> However, as I understand, this test didn't actually check that -dlogic inputs were not modified, right, which would bring us to no worse than before?

Yes.

> Alternatively, we should have something like an `ice40_check` which checks for this sort of arch-specific intricacies.

This seems hard, because there is no requirement that any given SB_CARRY in the input netlist is legal; we'd have to somehow check only Yosys-introduced carries.

> I also haven't changed the previous test `opt_lut_elim` and it continues to pass. Would this be sufficient?

Not entirely, but there is probably no urgent need to improve on this front.

> Sorry I'm not following. Are you saying we should have a $__ICE40_CARRY_WRAPPER specific optimisation pass (which would make it even more niche than opt_lut) or are you saying we should generic "whitebox" optimisation?

Not quite either: I was trying to say that opt_lut, instead of using `-dlogic`, should look at both `$lut` and `$__ICE40_FULL_ADDER` cells. The latter cells are sufficiently "LUT-like" that it could do the same sort of optimizations on them, *except* it should never reorder inputs.

However, now that you mentioned whitebox optimizations, I looked at ECP5's `CCU2C` and it looks like even the extended opt_lut I propose above would not be able to handle this. So a stronger approach is needed--probably something involving true whiteboxes...

<hr>

To summarize: please disregard my comments earlier. I think this PR should be fine as it is.

PR #1270, opened 2019-08-08
Add a few comments to document $alu and $lcu

eddiehung:


daveshah1:
It would be good to add a brief summary of the cell's function in a comment above the cell too (see the bitwise cells above, for example). This will then appear when you do `help $alu`

PR #1272, opened 2019-08-09
Propagate parameters for Travis build

mmicko:
Let's try out if this is sufficient


mmicko:
All builds now have proper version of build tools reported after start. So think this is ready for merge.

PR #1275, opened 2019-08-09
New ID() macro and now also use it

cliffordwolf:
This picks up on the ID() macro introduced a long time ago in 75423169c but was never used.

Usage examples: `ID(A)`, `ID($add)`

86c39d3 updates the `ID()` macro to provide the new API.

14c0289 applies the `ID()` macro to wherever it was easy to do in `passes/opt/*`.

### DIFF
--- a/kernel/yosys.h
+++ b/kernel/yosys.h
@@ -305,8 +305,16 @@ RTLIL::IdString new_id(std::string file, int line, std::string func);
 #define NEW_ID \
 	YOSYS_NAMESPACE_PREFIX new_id(__FILE__, __LINE__, __FUNCTION__)
 
-#define ID(_str) \
-	([]() { static YOSYS_NAMESPACE_PREFIX RTLIL::IdString _id(_str); return _id; })()
+// Create a statically allocated IdString object, using for example ID(A) or ID($add).
+//
+// Recipe for Converting old code that is using conversion of strings like "\\A" and
+// "$add" for creating IdStrings: Run below SED command on the .cc file and then use for
+// example "meld foo.cc foo.cc.orig" to manually compile errors, if necessary.
+//
+//  sed -i.orig -r 's/"\\\\([a-zA-Z0-9_]+)"/ID(\1)/g; s/"(\$[a-zA-Z0-9_]+)"/ID(\1)/g;' <filename>
+//
+#define ID(_id) ([]() { const char *p = "\\" #_id, *q = p[1] == '$' ? p+1 : p; \
+        static const YOSYS_NAMESPACE_PREFIX RTLIL::IdString id(q); return id; })()
 
 RTLIL::Design *yosys_get_design();
 std::string proc_self_dirname();
--- a/kernel/cellaigs.cc
+++ b/kernel/cellaigs.cc
@@ -268,9 +268,9 @@ Aig::Aig(Cell *cell)
 	cell->parameters.sort();
 	for (auto p : cell->parameters)
 	{
-		if (p.first == "\\A_WIDTH" && mkname_a_signed) {
+		if (p.first == ID(A_WIDTH) && mkname_a_signed) {
 			name = mkname_last + stringf(":%d%c", p.second.as_int(), mkname_is_signed ? 'S' : 'U');
-		} else if (p.first == "\\B_WIDTH" && mkname_b_signed) {
+		} else if (p.first == ID(B_WIDTH) && mkname_b_signed) {
 			name = mkname_last + stringf(":%d%c", p.second.as_int(), mkname_is_signed ? 'S' : 'U');
 		} else {
 			mkname_last = name;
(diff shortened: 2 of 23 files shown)


kernel/yosys.h
```diff
@@ -305,8 +305,8 @@ RTLIL::IdString new_id(std::string file, int line, std::string func);
 #define NEW_ID \
 	YOSYS_NAMESPACE_PREFIX new_id(__FILE__, __LINE__, __FUNCTION__)
 
-#define ID(_str) \
-	([]() { static YOSYS_NAMESPACE_PREFIX RTLIL::IdString _id(_str); return _id; })()
+#define ID(_id) ([]() { const char *p = "\\" #_id, *q = p[1] == '$' ? p+1 : p; \
```
daveshah1: Can you add a brief comment here explaining the purpose/reason for this macro (i.e. performance compared to just using a plain string and relying on conversion?)

eddiehung: It's not entirely clear what this does -- create and run a lambda that contains a const char* with the identifier, and then construct and move-return a new RTLIL::IdString? -- and what advantage this has over the implicit construction of an RTLIL::IdString which we had before?

daveshah1: AIUI, the purpose is the created IdString is `static const`, so it will only be constructed (with associated expensive hash table lookup) once at startup rather than every time it is used.

eddiehung: Ooh yes. I missed that. 

I think I initially discarded it because I was hoping this PR would be guaranteed to share the same IdString object between all users of each ID, but now I'm beginning to see that it only helps multiple constructions of the same ID for each particular usage location, and maybe a smart compiler would be able to identify that it can merge multiple `static const` IDs together. 

I am concerned that there is no _guarantee_ that the lambda will be inlined and executed just once though.

> so it will only be constructed (with associated expensive hash table lookup) once at startup rather than every time it is used.

My understanding is that it will only be executed upon the first invocation, after which it will not construct again. 

Also, wouldn't keeping all these statics around increase memory usage, since every appearance of each identifier could be duplicated?

cliffordwolf: The memory requirements for the static IdStrings should be minimal. It's one `int` for the IdString and one `bool` the compiler adds to guard the static initialization code.

Another advantage of the `ID(A)` and `ID($add)` syntax over the current coding style is that it is much easier to search-and-replace when/if we are creating something like dedicated pre-allocated `ID_A` and `ID_add` later.

My guess would be that with `ID()`, reference counting has a bigger performance impact than ID allocation. So imo the main advantage of something like `ID_A` would be that we could build it in a way so that we don't need to reference count those IDs. (I also have on my todo an ifdef hack that disables reference counting, so we can benchmark that to get a baseline figure for how much overhead that really creates.)

I'd also guess that the overhead of having to allocate IdString objects the first time a code path is executed is neglectable. The bigger issue is that it has to check every time if this is the first run.

cliffordwolf:
As an experiment I compiled and disassembled the following function:

```
IdString foobar() { return ID($foobar); }
```

Here is the x86 assembly, broken in pieces.

First, check if constructor needs to be run:

```
(gdb) disassemble foobar
Dump of assembler code for function foobar():
   0x00000000009ae814 <+0>:	push   %rbx
   0x00000000009ae815 <+1>:	mov    %rdi,%rbx
   0x00000000009ae818 <+4>:	mov    0x11fef7a(%rip),%al        # 0x1bad798 <_ZGVZZ6foobarvENK3$_0clEvE2id>
   0x00000000009ae81e <+10>:	test   %al,%al
   0x00000000009ae820 <+12>:	je     0x9ae842 <foobar()+46>
```

Next, the frequent part for when we did run this code before. Notice that this code contains no function calls.

```
   0x00000000009ae822 <+14>:	movslq 0x11fef67(%rip),%rax        # 0x1bad790 <_ZZZ6foobarvENK3$_0clEvE2id>
   0x00000000009ae829 <+21>:	test   %rax,%rax
   0x00000000009ae82c <+24>:	je     0x9ae83b <foobar()+39>
   0x00000000009ae82e <+26>:	lea    0x11fd2bb(%rip),%rcx        # 0x1babaf0 <_ZN5Yosys5RTLIL8IdString24global_refcount_storage_E>
   0x00000000009ae835 <+33>:	mov    (%rcx),%rcx
   0x00000000009ae838 <+36>:	incl   (%rcx,%rax,4)
   0x00000000009ae83b <+39>:	mov    %eax,(%rbx)
   0x00000000009ae83d <+41>:	mov    %rbx,%rax
   0x00000000009ae840 <+44>:	pop    %rbx
   0x00000000009ae841 <+45>:	retq   
```

Finally the messy part that only runs when the function is called the first time:

```
   0x00000000009ae842 <+46>:	lea    0x11fef4f(%rip),%rdi        # 0x1bad798 <_ZGVZZ6foobarvENK3$_0clEvE2id>
   0x00000000009ae849 <+53>:	callq  0x7842d0 <__cxa_guard_acquire@plt>
   0x00000000009ae84e <+58>:	test   %eax,%eax
   0x00000000009ae850 <+60>:	je     0x9ae822 <foobar()+14>
   0x00000000009ae852 <+62>:	lea    0x9b9621(%rip),%rdi        # 0x1367e7a
   0x00000000009ae859 <+69>:	callq  0x7e5780 <Yosys::RTLIL::IdString::get_reference(char const*)>
   0x00000000009ae85e <+74>:	mov    %eax,0x11fef2c(%rip)        # 0x1bad790 <_ZZZ6foobarvENK3$_0clEvE2id>
   0x00000000009ae864 <+80>:	lea    0x11fef25(%rip),%rsi        # 0x1bad790 <_ZZZ6foobarvENK3$_0clEvE2id>
   0x00000000009ae86b <+87>:	lea    0x7bf0(%rip),%rdi        # 0x9b6462 <Yosys::RTLIL::IdString::~IdString()>
   0x00000000009ae872 <+94>:	lea    0x11f81af(%rip),%rdx        # 0x1ba6a28
   0x00000000009ae879 <+101>:	callq  0x784c10 <__cxa_atexit@plt>
   0x00000000009ae87e <+106>:	lea    0x11fef13(%rip),%rdi        # 0x1bad798 <_ZGVZZ6foobarvENK3$_0clEvE2id>
   0x00000000009ae885 <+113>:	callq  0x784690 <__cxa_guard_release@plt>
   0x00000000009ae88a <+118>:	jmp    0x9ae822 <foobar()+14>
   0x00000000009ae88c <+120>:	mov    %rax,%rbx
   0x00000000009ae88f <+123>:	lea    0x11fef02(%rip),%rdi        # 0x1bad798 <_ZGVZZ6foobarvENK3$_0clEvE2id>
   0x00000000009ae896 <+130>:	callq  0x784400 <__cxa_guard_abort@plt>
   0x00000000009ae89b <+135>:	mov    %rbx,%rdi
   0x00000000009ae89e <+138>:	callq  0x784830 <_Unwind_Resume@plt>
```

cliffordwolf:
Now this also adds `ID::A` etc. This implementation of `ID::` is far from perfect and I have various ideas for better solutions. But it's easier to play with and benchmark implementation ideas when there's a codebase that is already using the new API.

Speaking of benchmarks: I see a slight performance improvement with this PR, but it's not much. Maybe 3% in my tests. This would suggest that maybe ID lookups aren't that expensive after all.

kernel/yosys.h
eddiehung: All sounds good. It wasn't clear if this whether this was the complete solution, so maybe I shouldn't be critiquing it with the complete bar.

eddiehung:
> Speaking of benchmarks: I see a slight performance improvement with this PR, but it's not much. Maybe 3% in my tests. This would suggest that maybe ID lookups aren't that expensive after all.

Is this with what you've committed (some `ID()` macros, some `ID::`) or after using more `ID::`? In my profiling in the past I've seen all ID lookups take as much as 10%, but that would be hard-coded IdStrings as well as user design ones...

cliffordwolf:
> Is this with what you've committed [..]

That's just what I committed. I.e. changes to `passes/opt/*` and `kernel/*`, but only the simple cases, i.e. still a lot of cases where we do things like operations on `std::string` objects to construct IDs.

> but that would be hard-coded IdStrings as well as user design ones

The user-defined ones are rarely constructed from strings. Usually they are just compared or looked up in a dict. I'd assume reference counting is the bigger issue with those. I'll make the experiment with disabled reference counting later today.

cliffordwolf:
>  I'll make the experiment with disabled reference counting later today.

I did that now (see `YOSYS_NO_IDS_REFCNT` added in b25cf36) and in my very quick test it looks like this alone is in the neighborhood of a 10% speedup.

PR #1277, opened 2019-08-09
opt_expr -fine to now trim LSBs of $alu cells too

### LINKED ISSUE #1262: opt_expr $add/$sub width reduction to also work on $alu
Seems that after #1213 `synth_xilinx` is still using 8xMUXCY 9xXORCY, because `alumacc` is transforming $add/$sub into $alu.

See: https://github.com/YosysHQ/yosys/pull/1213#issuecomment-519187669

eddiehung:
Extends #1213 for `$alu` cells, since `opt_expr -fine` is only run after `alumacc` in some flows.

Fixes #1262 

```
module opt_expr_add_test(input [3:0] i, input [7:0] j, output [8:0] o);
    assign o = (i << 4) + j;
endmodule
```
on `synth_xilinx` now produces:
```
   Number of cells:                 17
     LUT2                            8
     MUXCY                           4
     XORCY                           5
```


passes/opt/opt_expr.cc
```diff
@@ -666,6 +667,53 @@ void replace_const_cells(RTLIL::Design *design, RTLIL::Module *module, bool cons
...
+					else if (!sub && sig_a.at(i, State::Sx) == State::S0 && sig_b.at(i, State::Sx) != State::Sx) {
+						module->connect(sig_x[i], sig_b[i]);
+						module->connect(sig_y[i], sig_b[i]);
+						module->connect(sig_co[i], State::S0);
+					}
+					else
+						break;
+				}
+				if (i > 0) {
+					cover_list("opt.opt_expr.fine", "$alu", cell->type.str());
```
cliffordwolf: Two correction here:
- The first argument to `cover_list()` should be unique. `opt.opt_expr.fine.$alu` would be consistent with other `cover_list()`.
- `cover_list()` is used when there are multiple cases to cover. Here you only give one case (`"$alu"`), so that's a case for regular `cover()`.

Therefore, this line should probably just be: `cover("opt.opt_expr.fine.$alu");`

eddiehung:
Sorry, am no good at early morning commits. Will fix in a couple of hours

PR #1280, opened 2019-08-10
Revert "Wrap SB_LUT+SB_CARRY into $__ICE40_CARRY_WRAPPER"

daveshah1:
This seems to be causing combinational loop errors in nextpnr on picorv32 and reported QoR regressions. I suggest that we revert this until we figure out exactly what the problem is and how to fix it.


eddiehung:
Sounds good. Can you give me a test case?

smunaut:
@eddiehung You can try the picosoc in the picorv32 repo and just set CATCH_MISALIGN to 0 in picorv32 for instance.

Issue #1225, opened 2019-07-25
synth_xilinx is discarding tristate logic


daveshah1:
## Steps to reproduce the issue
```verilog
module top(inout pin, input dout, sel, output din);
assign pin = sel ? dout : 1'bz;
assign din = pin;
endmodule
```
`yosys -p "synth_xilinx; dump" tristate.v`
## Expected behavior

Yosys either instantiates a tristate buffer that downstream tools can understand; or at least prints a clearer warning/error that the tristate is being totally discarded for `synth_xilinx`, rather than the ambiguous `Yosys has only limited support for tri-state logic at the moment.` warning.

## Actual behavior

```
attribute \top 1
attribute \src "tristate.v:1"
module \top

  attribute \src "tristate.v:1"
  wire output 4 \din

  attribute \src "tristate.v:1"
  wire input 2 \dout

  attribute \src "tristate.v:1"
  wire inout 1 \pin

  attribute \src "tristate.v:1"
  wire input 3 \sel

  connect \din \dout
  connect \pin \dout
end
```

koriakin:
The bug happens in `opt_expr -full`.

PR #1283, opened 2019-08-11
Fix various NDEBUG compiler warnings

cliffordwolf:
This is an alternative to #1255. With this NDEBUG compiles without warnings for me.

### DIFF
--- a/frontends/aiger/aigerparse.cc
+++ b/frontends/aiger/aigerparse.cc
@@ -901,11 +901,6 @@ void AigerReader::post_process()
 				RTLIL::Cell* cell = module->cell(stringf("$__box%d__", variable));
 				if (cell) { // ABC could have optimised this box away
 					module->rename(cell, escaped_s);
-				#ifndef NDEBUG
-					RTLIL::Module* box_module = design->module(cell->type);
-					log_assert(box_module);
-				#endif
-
 					for (const auto &i : cell->connections()) {
 						RTLIL::IdString port_name = i.first;
 						RTLIL::SigSpec rhs = i.second;
--- a/frontends/aiger/aigerparse.cc
+++ b/frontends/aiger/aigerparse.cc
@@ -67,7 +67,7 @@ struct ConstEvalAig
 				continue;
 			for (auto &it2 : it.second->connections())
 				if (yosys_celltypes.cell_output(it.second->type, it2.first)) {
-					auto r = sig2driver.insert(std::make_pair(it2.second, it.second));
+					auto r YS_ATTRIBUTE(unused) = sig2driver.insert(std::make_pair(it2.second, it.second));
 					log_assert(r.second);
 				}
 		}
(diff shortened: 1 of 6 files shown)


eddiehung:
To be clear: the grand goal of #1255 was to make it such that use of a variable inside `log_debug()` and `log_assert()` would block the unused-variable warning, even under `#define NDEBUG`, without any additional user intervention.

Requiring `YS_ATTRIBUTE(unused)` -- whilst it does suppress the warning (which admittedly is the main title of both PRs) -- my position is that this (a) imposes a cognitive load for developers, who don't typically compile with `NDEBUG`, and (b) is a suppression that will persist even if the `log_*()` is later removed, resulting in a false negative.

cliffordwolf:
> To be clear: the grand goal of #1255 was to make it such that use of a variable inside log_debug() and log_assert() would block the unused-variable warning

I think someone building with `NDEBUG` doesn't want unnecessary code to run. So I guess we should use those warning to guide us to actually remove that unnecessary code in the NDEBUG case.

Simply to trust that the compiler will optimize away is overly optimistic in my opinion. And generally not a good guess from my experience.

So just making the compiler warning go away is imo a step in the wrong direction. We should actually fix the underlying condition the compiler is warning us about, which is that we are spending cycles on calculating things that are not used.

> Requiring YS_ATTRIBUTE(unused) -- [..] even if the log_*() is later removed, resulting in a false negative.

And that's why it should only be used in cases where using `#ifndef` blocks would be cumbersome (for example because it's a function parameter), or where we just want `log_assert` to check the return code of a function we want to call anyway, such as the toposort examples in my patch.

frontends/aiger/aigerparse.cc
```diff
@@ -440,16 +444,18 @@ void AigerReader::parse_xaiger(const dict<int,IdString> &box_lookup)
 			}
 			else if (c == 'h') {
 				f.ignore(sizeof(uint32_t));
+			#ifndef NDEBUG
```
daveshah1: `parse_xaiger_literal` has an `istream::read` in it, so disabling it in NDEBUG looks to be incorrect?

cliffordwolf: Thanks! I misread that code.. Should be fixed now.

eddiehung:
I agree with your points. We should use `YS_ATTRIBUTE(unused)` sparingly, but (assuming we get many more liberal uses of `log_{assert,debug}()`) this would assume that we have to apply it to every variable which is only used by those functions. I would still not prefer that, but not something I would die fighting for. 

Your suggested change from #1225 _should_ be obvious enough for a compiler to optimise, no?
```
- #  define log_assert(_assert_expr_)
+ #  define log_assert(_assert_expr_) do { if (0) (void)(_assert_expr_); } while (0)
```

cliffordwolf:
> more liberal uses of log_{assert,debug}()) this would assume that we have to apply it to every variable which is only used by those functions.

I'd argue that in this case the code should be structured in a way so that there is one big `#ifndef` block containing all the variables and `log_debug()` calls, instead of creating individual variables here and there just so we can log them later.

> Your suggested change from #1225 should be obvious enough for a compiler to optimise, no?

My point is that we actually want to get that warning message for unused variables, as they are indicators of us potentially performing unnecessary computations. This change would suppress the warning, without doing anything about the underlying problem the compiler is warning us about.

eddiehung:
> I'd argue that in this case the code should be structured in a way so that there is one big `#ifndef` block containing all the variables and `log_debug()` calls, instead of creating individual variables here and there just so we can log them later.

Normally, that is the case. In a few rare bits of my code, I call functions that have side-effects and that return something. In a debug scenario, I'd like to output those return values, or maybe assert on them. During `NDEBUG` those variables are no longer consumed, and so cause a warning. Since this only happens rarely, I can `YS_ATTRIBUTE(unused)` them, then.

frontends/verific/verific.cc
```diff
@@ -1789,8 +1789,10 @@ struct VerificExtNets
 				new_net = new Net(name.c_str());
 				nl->Add(new_net);
 
+			#ifndef NDEBUG
 				Net *n = route_up(new_net, port->IsOutput(), ca_nl, ca_net);
```
eddiehung: It really looks like this call should have side effects -- are you sure it is meant to be inside `NDEBUG`?

cliffordwolf: Oooops. Fixed in 27d59dc.

PR #1285, opened 2019-08-12
Add job 'docker' to Travis CI

eine:
Ref #1152

This PR adds a job named 'docker' to the Travis CI configuration file. In this job, the docker image is built. This is useful to ensure that the image is buildable. However, it is not pushed anywhere. Hence, users willing to use it need to build it themselves.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -131,3 +131,10 @@ matrix:
 #     osx_image: xcode9.4
 #     env:
 #       - MATRIX_EVAL="CONFIG=clang && CC=clang && CXX=clang++"
+
+    - services: docker
+      language: minimal
+      install: skip
+      before_install:
+      script: docker build -t yosys/yosys .
+      after_success:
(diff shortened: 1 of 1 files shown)


mmicko:
Due to limited time we can use for Travis build, we are not able to build docker image on each commit.

eine:
@mmicko, I understand that. Precisely, in org GHDL we moved from Travis to GitHub Actions due to the time limit (50min vs 6h). That's not a problem with GHDL itself, but yosys, nextpnr, etc. might take longer than 50min if tests are executed too.

Anyway, this PR was just an excuse to bring up the discussion, since it would be almost useless anyway. I'd be glad if you could please have a look at #1287, which is the relevant issue indeed. See also librecores/docker-images#33 and ghdl/docker#21.

mmicko:
@eine Thank you for feedback. We are working on new build/test environment, and will keep this in mind.

PR #1292, opened 2019-08-12
Add clock buffer insertion pass, improve iopadmap.

koriakin:
A few new attributes are defined for use in cell libraries:

- iopad_external_pin: marks PAD cell's external-facing pin.  Pad
  insertion will be skipped for ports that are already connected
  to such a pin.
- clkbuf_sink: marks an input pin as a clock pin, requesting clock
  buffer insertion.
- clkbuf_driver: marks an output pin as a clock buffer output pin.
  Clock buffer insertion will be skipped for nets that are already
  driven by such a pin.

All three are module attributes that should be set to a comma-separeted
list of pin names.

Clock buffer insertion itself works as follows:

1. All cell ports, starting from bottom up, can be marked as clock sinks
   (requesting clock buffer insertion) or as clock buffer outputs.
2. If a wire in a given module is driven by a cell port that is a clock
   buffer output, it is in turn also considered a clock buffer output.
3. If an input port in a non-top module is connected to a clock sink in a
   contained cell, it is also in turn considered a clock sink.
4. If a wire in a module is driven by a non-clock-buffer cell, and is
   also connected to a clock sink port in a contained cell, a clock
   buffer is inserted in this module.
5. For the top module, a clock buffer is also inserted on input ports
   connected to clock sinks, optionally with a special kind of input
   PAD (such as IBUFG for Xilinx).
6. Clock buffer insertion on a given wire is skipped if the clkbuf_inhibit
   attribute is set on it.


techlibs/xilinx/synth_xilinx.cc
```diff
@@ -78,6 +81,12 @@ struct SynthXilinxPass : public ScriptPass
 		log("    -nowidelut\n");
 		log("        do not use MUXF[78] resources to implement LUTs larger than LUT6s\n");
 		log("\n");
+		log("    -iopads\n");
```
eddiehung: Perhaps we should have just one option instead of two: `-iopad` (non plural to be consistent) to make it clearer.
In the unlikely scenario that the user wishes to disable under `-ise`, perhaps they can do `-iopad=false`?

techlibs/xilinx/synth_xilinx.cc
```diff
@@ -410,6 +434,17 @@ struct SynthXilinxPass : public ScriptPass
 			run("clean");
 		}
 
+		if (check_label("finalize")) {
+			bool do_iopads = iopads || (ise && !noiopads);
+			if (help_mode || do_iopads)
+				run("clkbufmap -buf BUFG O:I -inpad IBUFG O:I", "(-inpad passed if '-iopads' or '-ise' and not '-noiopads')");
```
eddiehung: Is there any harm in specifying `-inpad IBUFG O:I` all the time? For example, what if the user instantiated some IBUFGs, but did not want any extra tool promotion to happen?

passes/techmap/clkbufmap.cc
```diff
@@ -0,0 +1,299 @@
...
+						driven_wire_bits.insert(port.second[i]);
+
+			// Insert buffers.
+			std::vector<pair<Wire *, Wire *>> input_queue;
+			for (auto wire : module->selected_wires())
+			{
+				// Should not happen.
+				if (wire->port_input && wire->port_output)
+					continue;
+				if (wire->get_bool_attribute("\\clkbuf_inhibit"))
```
eddiehung: Perhaps we can make this a user defined option given to the `clkbufmap` pass? That way it can be generic to match whatever the style guide or vendor rules exist

koriakin: That'd be a useful feature, but wouldn't it be better to use the generic `attrmap` pass instead of coding it in here?

techlibs/xilinx/synth_xilinx.cc
koriakin: I don't quite follow? If pad insertion is disbled, that's presumably because the user wants some kind of out-of-context mode, and we definitely shouldn't insert the IBUFG.  On the other hand, if pad insertion is enabled, and the user wants to avoid IBUFG, they could just manually instantiate another pad type and/or manually instantiate just the BUFG (which will trigger plain IBUF insertion in the following iopadmap pass).

Or do you mean you want an option to disable BUFG insertion altogether?  I agree that would be useful, I'm going to add it.

techlibs/xilinx/synth_xilinx.cc
koriakin: Certainly could be done, but I see no other option in yosys that takes a bool like that, and there is a precedent for -<option> and -no<option> (read_verilog -[no]mem2reg, -[no]overwrite). Agreed about non plural.

techlibs/xilinx/synth_xilinx.cc
eddiehung: Sorry, I misread the use of this flag! But allow clock promotion to be disabled is a good thing to do.

techlibs/xilinx/synth_xilinx.cc
eddiehung: Just an idea to reduce the number of yes/no flags, but you're right, there's no precedent. Let's leave as-is.

techlibs/xilinx/synth_xilinx.cc
```diff
@@ -63,6 +63,9 @@ struct SynthXilinxPass : public ScriptPass
 		log("        generate an output netlist (and BLIF file) suitable for VPR\n");
 		log("        (this feature is experimental and incomplete)\n");
 		log("\n");
+		log("    -ise\n");
+		log("        generate an output netlist suitable for ISE\n");
```
eddiehung: Let's explicitly mention that this enables `-iopad` here.

CHANGELOG
```diff
@@ -19,6 +19,9 @@ Yosys 0.9 .. Yosys 0.9-dev
     - Added automatic gzip compression (based on filename extension) for backends
     - Improve attribute and parameter encoding in JSON to avoid ambiguities between
       bit vectors and strings containing [01xz]*
+    - Added "clkbufmap" pass
+    - Added "synth_xilinx -ise" for Spartan 6 (experimental)
```
eddiehung: I would just leave it as `synth_xilinx -ise` since it is not Spartan 6 specific.
You can add another entry that adds `synth_xilinx -family xc6s` for Spartan 6 (experimental)
I would also add entries for `-iopad/-noipad`

CHANGELOG
```diff
@@ -19,6 +19,9 @@ Yosys 0.9 .. Yosys 0.9-dev
...
+    - "synth_xilinx" now automatically inserts clock buffers
```
eddiehung: Add "(`-noclkbuf` to disable)"?

eddiehung:
Oh wait. Perhaps you can add a few arch-neutral tests for the new `clkbufmap` pass?

passes/techmap/clkbufmap.cc
```diff
@@ -0,0 +1,287 @@
...
+		// Process submodules before module using them.
+		std::vector<Module *> modules_sorted;
+		pool<Module *> modules_processed;
+		for (auto module : design->selected_modules())
+			module_queue(design, module, modules_sorted, modules_processed);
+
+		for (auto module : modules_sorted)
+		{
+			if (module->get_blackbox_attribute()) {
+				auto it = module->attributes.find("\\clkbuf_driver");
```
cliffordwolf: These new attributes `\clkbuf_driver`, `\clkbuf_sink`, etc. need to be added to the list of attributes yosys supports in the README file.

passes/techmap/clkbufmap.cc
```diff
@@ -0,0 +1,287 @@
...
+				if (it != module->attributes.end()) {
+					auto value = it->second.decode_string();
+					for (auto name : split_tokens(value, ",")) {
+						auto wire = module->wire(RTLIL::escape_id(name));
```
cliffordwolf: Wouldn't it be better to simply put a boolean `\clkbuf_driver` attribute on the wire, instead of listing the wire names in an attribute on the module?

passes/techmap/clkbufmap.cc
```diff
@@ -0,0 +1,287 @@
...
+						if (!wire)
+							log_error("Module %s does not have port %s.\n", log_id(module), log_id(name));
+						for (int i = 0; i < GetSize(wire); i++)
+							buf_ports.insert(make_pair(module->name, make_pair(RTLIL::escape_id(name), i)));
+					}
+				}
+				it = module->attributes.find("\\clkbuf_sink");
+				if (it != module->attributes.end()) {
+					auto value = it->second.decode_string();
+					for (auto name : split_tokens(value, ",")) {
```
cliffordwolf: Same as above.

passes/techmap/clkbufmap.cc
eddiehung: Absolutely possible through `attrmap`, but that's another walk over the design when you're doing one here anyway

koriakin:
Rewritten cells_xtra.sh to Python to support attaching per-wire attributes.  Aside from the new attributes, the only difference in the generated cells_xtra.v file is that input/output/inout lines with more than one port are now split to separate lines.

passes/techmap/clkbufmap.cc
```diff
@@ -0,0 +1,273 @@
...
+		// Process submodules before module using them.
+		std::vector<Module *> modules_sorted;
+		pool<Module *> modules_processed;
+		for (auto module : design->selected_modules())
+			module_queue(design, module, modules_sorted, modules_processed);
+
+		for (auto module : modules_sorted)
+		{
+			if (module->get_blackbox_attribute()) {
+				for (auto wire : module->wires()) {
```
cliffordwolf: As mentioned in the zoom call, jfyi, you can efficiently iterate over module ports like this:

```
for (auto port_name : module->ports) {
    auto wire = module->wire(port_name);
```

Not really a big deal in this case since most wires on blackbox modules (and even whitebox modules) are module ports anyway.

passes/techmap/clkbufmap.cc
eddiehung: Another reason for a user specified attribute: https://github.com/YosysHQ/nextpnr/pull/317 as opposed to a hardcoded one (and to reduce the number of hardcoded attributes Yosys has to support)

eddiehung:
@koriakin to review my changes.

passes/techmap/clkbufmap.cc
```diff
@@ -166,13 +166,24 @@ struct ClkbufmapPass : public Pass {
 
 			// Insert buffers.
 			std::vector<pair<Wire *, Wire *>> input_queue;
-			for (auto wire : module->selected_wires())
+			// Copy current wire list, as we will be adding new ones during iteration.
+			std::vector<Wire *> wires(module->wires());
+			for (auto wire : wires)
```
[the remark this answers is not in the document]
koriakin: Yes, got an assert failure.

PR #1294, opened 2019-08-14
Revert "Since $_ANDNOT_ is not symmetric, do not sort leaves"

eddiehung:
Reverts YosysHQ/yosys#1288

Seems to break yosys-tests on internal Jenkins (but can't reproduce locally, indicating order issue?)

### DIFF
--- a/passes/techmap/extract_fa.cc
+++ b/passes/techmap/extract_fa.cc
@@ -153,10 +153,12 @@ struct ExtractFaWorker
 		}
 	}
 
-	void check_partition(SigBit root, const pool<SigBit> &leaves)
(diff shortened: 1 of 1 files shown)


cliffordwolf:
So... thinking more about this: The reason `leaves.sort()` is here is so that keys in `func3` / `func2` are canonical. There should be no problem regarding asymetric cells such as `$_ANDNOT_`, because the cells are just evaluated using `ce.eval()`, which knows which input is which in the circuit under test. There's no need for the `A` variable to correspond to the `A` input of the cell for this to work correctly.

All that being said: Afair `extract_fa` never fully worked. So there's definitely a bug somewhere in there, but it wasn't that.

PR #1296, opened 2019-08-14
Since $_ANDNOT_ is not symmetric, do not sort leaves (retry)

eddiehung:
Retry of #1288 which caused yosys-tests to fail on Jenkins, but that I couldn't reproduce locally. I suspect that indicates that sorting was necessary to get a canonical result...

This is a much less heavy-handed approach, which only blocks sorting for `$_ANDNOT_` and `$_ORNOT_` cells. Are there any other 2-input or 3-input cells we should be blocking sort on?

### DIFF
--- a/passes/techmap/extract_fa.cc
+++ b/passes/techmap/extract_fa.cc
@@ -153,11 +153,12 @@ struct ExtractFaWorker
 		}
 	}
 
-	void check_partition(SigBit root, pool<SigBit> &leaves)
+	void check_partition(SigBit root, pool<SigBit> &leaves, IdString cell_type)
 	{
 		if (config.enable_ha && GetSize(leaves) == 2)
 		{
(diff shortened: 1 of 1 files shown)


eddiehung:
Sorry, missed @cliffordwolf's reply here: https://github.com/YosysHQ/yosys/pull/1294#issuecomment-521377975
> So... thinking more about this: The reason leaves.sort() is here is so that keys in func3 / func2 are canonical. There should be no problem regarding asymetric cells such as $_ANDNOT_, because the cells are just evaluated using ce.eval(), which knows which input is which in the circuit under test. There's no need for the A variable to correspond to the A input of the cell for this to work correctly.
>
>All that being said: Afair extract_fa never fully worked. So there's definitely a bug somewhere in there, but it wasn't that.

That makes sense. Let me keep digging.

PR #1300, opened 2019-08-15
Use ID::{A,B,Y,keep,blackbox,whitebox} instead of ID()

eddiehung:
Thought I'd try using the existing globals too.

Note I've aliased the `RTLIL::ID` namespace as the `ID` namespace:
https://github.com/YosysHQ/yosys/blob/453a9429b67551f94f8c714c24f12929e472052c/kernel/yosys.h#L228

Not sure what's possible with dollar-identifiers (e.g. `$and`) globals, though...

### DIFF
--- a/kernel/cellaigs.cc
+++ b/kernel/cellaigs.cc
@@ -292,19 +292,19 @@ Aig::Aig(Cell *cell)
 
 	if (cell->type.in(ID($not), ID($_NOT_), ID($pos), ID($_BUF_)))
 	{
-		for (int i = 0; i < GetSize(cell->getPort(ID(Y))); i++) {
-			int A = mk.inport(ID(A), i);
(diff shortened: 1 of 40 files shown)


eddiehung:
Grrr seems like this line doesn't go down well with gcc-4.8: https://github.com/YosysHQ/yosys/blob/453a9429b67551f94f8c714c24f12929e472052c/kernel/yosys.h#L228

mmicko:
Just remove "using namespace RTLIL;" from wreduce.cc and use RTLIL::Sx instead of Sx and RTLIL::S0 instead of S0. This is anyway how it is used on all other parts of code. Tested and this works on gcc-4.8 and others as well

eddiehung:
Thanks @mmicko! I never noticed that there was a stray `using namespace RTLIL;` inside `wreduce.cc` and thought it affected other files too!

PR #1301, opened 2019-08-16
Add support for extracting init values of ports to techmap, add proper support for init values on Spartan 6 FFs.

koriakin:


koriakin:
Added a testcase, added changelog entry (only for the techmap improvement, since synth_xilinx changes count towards adding Spartan 6 support).

passes/techmap/techmap.cc
```diff
@@ -1032,6 +1079,14 @@ struct TechmapPass : public Pass {
 		log("        former has a 1-bit for each constant input bit and the latter has the\n");
 		log("        value for this bit. The unused bits of the latter are set to undef (x).\n");
 		log("\n");
+		log("    _TECHMAP_WIREINIT_<port-name>_\n");
+		log("        When a parameter with this name exists, it will be set to the initial\n");
+		log("        value of the wire(s) connected to the given port, as specified by the init\n");
+		log("        attribute. If the attribute doesn't exist, x will be filled for the\n");
+		log("        missing bits. If this module is used for mapping, the init attribute bits\n");
```
eddiehung: If this parameter is present, does this makes it mandatory for the rule to consume it (because otherwise it will get blown away?) This seems like quite an onerous things to do...

daveshah1: Perhaps we could have a `_TECHMAP_REMOVEINIT_x_` wire, where a 1 in a bit of this wire removes `(* init *)` from that bit of port x. This is similar to how wires trigger actions in techmap more generally.

koriakin: @daveshah1 that sounds like a good idea, I'll implement it.

PR #1302, opened 2019-08-16
DFFLIBMAP pass regression fix

mmicko:


passes/techmap/dfflibmap.cc
```diff
@@ -156,10 +156,10 @@ static void find_cell(LibertyAst *ast, IdString cell_type, bool clkpol, bool has
 		}
 
 		std::map<IdString, char> this_cell_ports;
-		this_cell_ports[cell_clk_pin] = 'C';
+		this_cell_ports[ID(cell_clk_pin)] = 'C';
```
cliffordwolf: That can't be right. This will literally create an ID `cell_clk_pin`. But the intention of the code is obviously to convert the content of the string stored in the cell_clk_pin variable into an ID.

mmicko: That line is exactly where it fails with:
ERROR: Assert `p[0] == '$' || p[0] == '\\'' failed in ./kernel/rtlil.h:153.

mmicko: I see what you are pointing at, after macro expansion is not right for sure. looking where things got wrong actually.

cliffordwolf: 9f98241010481588d643c6d4e24d7b9af2b37c2f changed some of the maps to use `IdString` as key instead of `std::string`, without making sure those strings are actually convertible into an `IdString`. We can either switch those back to use `std::string` keys, or run the key values through `RTLIL::escape_id()` befor converting to `IdString`.

mmicko: Since LibertyAST is using strings internally believe it is proper to use it here until that one is converted to use AST. cell_name is one I need to escape for now.

PR #1304, opened 2019-08-16
Refactor abc9 to use port attributes, not module attributes

eddiehung:
And doc.

### DIFF
--- a/techlibs/xilinx/cells_sim.v
+++ b/techlibs/xilinx/cells_sim.v
@@ -183,9 +183,9 @@ endmodule
 
 (* abc_box_id = 4, lib_whitebox *)
 module CARRY4(
-  (* abc_carry_out *) output [3:0] CO,
+  (* abc_carry *) output [3:0] CO,
   output [3:0] O,
-  (* abc_carry_in *) input CI,
+  (* abc_carry *) input CI,
   input        CYINIT,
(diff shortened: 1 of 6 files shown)


eddiehung:
CC @dh73 to review

techlibs/xilinx/cells_sim.v
```diff
@@ -181,8 +181,14 @@ module XORCY(output O, input CI, LI);
   assign O = CI ^ LI;
 endmodule
 
-(* abc_box_id = 4, abc_carry="CI,CO", lib_whitebox *)
-module CARRY4(output [3:0] CO, O, input CI, CYINIT, input [3:0] DI, S);
+(* abc_box_id = 4, lib_whitebox *)
+module CARRY4(
+  (* abc_carry_out *) output [3:0] CO,
```
mithro: Why can't you use the fact this port is an output to detect it is a carry out?

eddiehung: There are two outputs, no?

mithro: I mean `(* abc_carry *)` and use the port being output to mean `abc_carry_out`.

eddiehung: I see what you mean now. Good suggestion!

PR #1308, opened 2019-08-19
Handle real values when deriving ast modules

jakobwenzel:
Currently, when real values are used as parameters, they are converted to their string value and marked as reals. When the parameterized module is derived, the string's value is used as value for the real.

This PR attempts to fix that by checking for the real values' flag.


eddiehung:
Can you provide the testcase you identified this issue with?

jakobwenzel:
Sure!

Verilog Code:
```
module child();
	parameter real p = 0;
	assert property(p==1.0);
	assert property(p!="1.000000");
endmodule

module parent();
	child #(1.0) child_real();
	child #(1) child_int();
endmodule
```

Yosys Commands:
```
read_verilog real_params.v
hierarchy -top parent
write_ilang
```

Without the PR, child_real's asserts fail.

cliffordwolf:
Added test case in 6ffb910.

PR #1310, opened 2019-08-20
Add new tests for ice40 architecture

SergeyDegtyar:


tests/ice40/add_sub.ys
```diff
@@ -0,0 +1,7 @@
+equiv_opt -map ../../techlibs/ice40/cells_sim.v synth_ice40
```
eddiehung: Consider the more elegant `+/ice40/cells_sim.v` special path

tests/ice40/add_sub.ys
```diff
@@ -0,0 +1,7 @@
...
+synth_ice40
+select -assert-count 12 t:SB_LUT4
+select -assert-count 7 t:SB_CARRY
+select -assert-count 2 t:$logic_and
```
eddiehung: I would expect any of these internal cells here. They should go away after you do a `design -load`

tests/ice40/memory.ys
```diff
@@ -0,0 +1,7 @@
+proc
+memory
+equiv_opt -map ../../techlibs/ice40/cells_sim.v synth_ice40
+synth_ice40
+select -assert-count 8 t:SB_DFF
+select -assert-count 512 t:SB_DFFE
```
eddiehung: This should map to a memory, but is creating 500+ flops instead. How come?

SergeyDegtyar: If use just synth_ice40 we have this result:
     SB_RAM40_4K                     1

but equiv_opt reports: ERROR: Gold module contains memories or procresses. Run 'memory' or 'proc' respectively.

But when use this commands:
read_verilog memory.v
proc
memory
synth_ice40

I have this result (like in current test case):

     SB_DFF                          8
     SB_DFFE                       512
     SB_LUT4                       464


When use the same commands without proc, the result will be:
     $mem                            1
     SB_DFF                         23

Maybe there is a bug but without memory command I cant use equiv_opt, and with it I don't have an SB_RAM40_4K.

tests/ice40/add_sub.ys
eddiehung: Sorry: I *wouldn't* expect...

tests/ice40/add_sub.ys
```diff
@@ -0,0 +1,7 @@
+synth_ice40
```
eddiehung: Actually, can you replace the first `synth_ice40` call with `synth -coarse` which causes it to do technology independent synthesis. That way you are checking that technology-independent synthesis is equivalent to technology-specific, rather than checking `synth_ice40` against itself.

May I suggest this template for all tests:
```
hierarchy -top top
synth -run coarse # technology-independent coarse grained synthesis
equiv_opt -assert -map +/ice40/cells_sim.v synth_ice40 # equivalency check same as technology-dependent fine-grained synthesis
select -assert-count 11 t:SB_LUT4
select -assert-count 6 t:SB_CARRY
select -assert-none t:SB_LUT4 t:SB_CARRY %% t:* %D # Check that there is no other cells except for these two. Explained, this select command means: select all SB_LUT4 cells as first element of stack. Then push all SB_CARRY cells as second element of stack. "%%" means to find union of all elements of stack. "t:*" selects all cells, and "%D" subtracts the top-1 element (i.e. all SB_LUT4 and SB_CARRY) from the top element (all cells) and ensures that no cells are leftover.
```

tests/ice40/tribuf.v
```diff
@@ -0,0 +1,23 @@
+module tristate (en, i, o);
+    input en;
+    input i;
+    output reg o;
+
+    always @(en or i)
+		o <= (en)? i : 1'bZ;
+endmodule
+
```
eddiehung: Can you also add the tristate template described in: https://www.latticesemi.com/-/media/LatticeSemi/Documents/UserManuals/EI/iCEcube201701UserGuide.ashx?document_id=52071
please?

tests/ice40/run-test.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -e
+for x in *.v; do
+  echo "Running $x.."
+  ../../yosys -q -s ${x%.v}.ys -l ./temp/${x%.v}.log $x
```
eddiehung: I think it might be failing because the `temp` dir doesn't exist.

Can you change the script to be more similar to this:
https://github.com/YosysHQ/yosys/blob/71dd412ac55860cbf51d91d26088515978f70116/tests/various/run-test.sh

(but without the `*.sh` loop?)

tests/ice40/dffs.v
```diff
@@ -0,0 +1,108 @@
+module adff
```
eddiehung: So it seems like `equiv_*` does not support asynchronous or SR flops.
Can you please:
(a) Add a boring-old dff test (without enable)
(b) Move these async/SR flops into a different test that does not run equiv_opt, but instead just does `select -assert-count` and uses simulation (see `tests/simple` for an `test_autotb` approach)

tests/ice40/div_mod.ys
```diff
@@ -0,0 +1,6 @@
+synth_ice40
+#equiv_opt -assert -map +/ice40/cells_sim.v synth_ice40
```
eddiehung: This works for me:
```
hierarchy -top top
synth -flatten -run coarse
equiv_opt -assert -map +/ice40/cells_sim.v synth_ice40
#equiv_opt -map +/ice40/cells_sim.v synth_ice40
design -load postopt
cd top
select -assert-count 88 t:SB_LUT4
select -assert-count 65 t:SB_CARRY
```

tests/ice40/latches.ys
```diff
@@ -0,0 +1,6 @@
+synth_ice40
+#equiv_opt -assert -map +/ice40/cells_sim.v synth_ice40
```
eddiehung: Looks like `equiv_*` does not support latches, either, so perhaps we should resort to simulation for these. Take a look at `tests/simple` for an approach that auto generates a testbench.

tests/ice40/mul_pow.v
```diff
@@ -0,0 +1,13 @@
...
+(
+ input [3:0] x,
+ input [3:0] y,
+
+ output [3:0] A,
+ output [3:0] B
+ );
+
+assign A =  x * y;
+assign B =  x ** y;
```
eddiehung: `$pow` is not a typically synthesisable construct. Please remove.

Change circuit to be wider:
```
module top
(
 input [5:0] x,
 input [5:0] y,

 output [11:0] A,
 );

assign A =  x * y;

endmodule
```

tests/ice40/mul_pow.ys
```diff
@@ -0,0 +1,7 @@
+synth_ice40
+#equiv_opt -assert -map +/ice40/cells_sim.v synth_ice40
```
eddiehung: ... and run `synth_ice40` with the `-dsp` option to infer multipliers.

tests/ice40/memory.ys
eddiehung: Running `proc` and `memory` before `synth_ice40` blocks it from inferring a RAM.
However, even if you fix this issue and others, it looks like there's an issue with equivalency checking memories, so maybe we should resort to just `select -assert-count` and simulation for this particular test.

tests/ice40/memory.v
```diff
@@ -0,0 +1,21 @@
+module top
```
eddiehung: Consider memory templates from: https://www.latticesemi.com/-/media/LatticeSemi/Documents/UserManuals/EI/iCEcube201701UserGuide.ashx?document_id=52071

tests/ice40/tribuf.v
SergeyDegtyar: If I use this template I also have $_TBUF_ cell and error from equiv_opt about unproven cells

tests/ice40/add_sub.ys
SergeyDegtyar: done

tests/ice40/memory.ys
SergeyDegtyar: done

tests/ice40/run-test.sh
SergeyDegtyar: done

tests/ice40/dffs.v
SergeyDegtyar: done

tests/ice40/div_mod.ys
SergeyDegtyar: done

tests/ice40/latches.ys
SergeyDegtyar: done

tests/ice40/mul_pow.v
SergeyDegtyar: done

tests/ice40/mul_pow.ys
SergeyDegtyar: for now the result is the same. Will change this case.

eddiehung: That's because the multiply operation isn't wide enough:
```
Change circuit to be wider:

module top
(
 input [5:0] x,
 input [5:0] y,

 output [11:0] A,
 );

assign A =  x * y;

endmodule
```
now works.

tests/ice40/run-test.sh
```diff
@@ -0,0 +1,33 @@
...
+echo "all::"
+for x in *.ys; do
+	echo "all:: run-$x"
+	echo "run-$x:"
+	echo "	@echo 'Running $x..'"
+	echo "	@../../yosys -ql ${x%.ys}.log $x -w 'Yosys has only limited support for tri-state logic at the moment.'"
+done
+for t in *_tb.v; do
+	echo "all:: run-$t"
+	echo "run-$t:"
```
mmicko: change this line to :
echo "run-$t: ${t%_tb.v}_synth.v"

mmicko:
This should fix it : 

```
diff --git a/tests/ice40/.gitignore b/tests/ice40/.gitignore
index 1d329c93..3742b7a7 100644
--- a/tests/ice40/.gitignore
+++ b/tests/ice40/.gitignore
@@ -1,2 +1,4 @@
 *.log
+*_synth.v
+*_testbench
 /run-test.mk
diff --git a/tests/ice40/run-test.sh b/tests/ice40/run-test.sh
index 75aa0833..11b0ce1f 100755
--- a/tests/ice40/run-test.sh
+++ b/tests/ice40/run-test.sh
@@ -13,14 +13,13 @@ for x in *.ys; do
 	echo "run-$x:"
 	echo "	@echo 'Running $x..'"
 	echo "	@../../yosys -ql ${x%.ys}.log $x -w 'Yosys has only limited support for tri-state logic at the moment.'"
+	if [ -f "${x%.ys}_tb.v" ]; then
+		echo "	@echo 'Running ${x%.ys}_tb.v..'"
+		echo "	@iverilog -o ${x%.ys}_testbench $t ${x%.ys}_synth.v common.v $TECHLIBS_PREFIX/ice40/cells_sim.v"
+		echo "	@vvp -N ${x%.ys}_testbench"
+	fi
 done
-for t in *_tb.v; do
-	echo "all:: run-$t"
-	echo "run-$t: ${t%_tb.v}_synth.v"
-	echo "	@echo 'Running $t..'"
-	echo "	@iverilog -o ${t%_tb.v}_testbench  $t ${t%_tb.v}_synth.v common.v $TECHLIBS_PREFIX/ice40/cells_sim.v"
-	echo "	@vvp -N ${t%_tb.v}_testbench"
-done
+
 for s in *.sh; do
 	if [ "$s" != "run-test.sh" ]; then
 		echo "all:: run-$s"
```

mmicko:
@eddiehung can this be merged or there are some more changes needed ?

tests/ice40/run-test.sh
SergeyDegtyar: done

tests/ice40/memory.v
SergeyDegtyar: done

SergeyDegtyar:
@eddiehung yes, all comments were resolved.

tests/ice40/memory.v
SergeyDegtyar: You are right. From this document i added dpram.v. But this case is very similar to one of examples from this guide.

tests/ice40/dpram.v
```diff
@@ -0,0 +1,20 @@
+module top (din, write_en, waddr, wclk, raddr, rclk, dout);
```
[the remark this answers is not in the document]
SergeyDegtyar: OK. I will add comments to this cases.

eddiehung:
There are two issues:
- Please resolve https://github.com/YosysHQ/yosys/pull/1310#discussion_r318756332
- Please comment out https://github.com/YosysHQ/yosys/blob/5dda8f39a6b13610084d259e5490604380350eba/tests/ice40/run-test.sh#L17-L21 as we have no more testbenches (I commented out the wrong lines before!)

and I think we'll be good to merge!

tests/ice40/dpram.v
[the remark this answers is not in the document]
SergeyDegtyar: Done

SergeyDegtyar:
"
- Please resolve #1310 (comment)
- Please comment out
"
Both done.

I have also added several tests from yosys-cover/architecture_ice40 - I think, now it is not necessary, have two test groups for ice40

tests/ice40/dpram.v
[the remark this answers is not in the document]
eddiehung: @SergeyDegtyar I'm still not seeing any changes to this file?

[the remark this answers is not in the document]
SergeyDegtyar: https://github.com/SergeyDegtyar/yosys/edit/master/tests/ice40/dpram.v?pr=%2FYosysHQ%2Fyosys%2Fpull%2F1310

/*
Example from: https://www.latticesemi.com/-/media/LatticeSemi/Documents/UserManuals/EI/iCEcube201701UserGuide.ashx?document_id=52071 [p. 72].
*/
module top (din, write_en, waddr, wclk, raddr, rclk, dout);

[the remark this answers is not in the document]
SergeyDegtyar: added by this commit:
https://github.com/YosysHQ/yosys/pull/1310/commits/d588c6898fb7cfebe52a71a48d6fb21d1623e61b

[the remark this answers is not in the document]
eddiehung: My apologies. GH didn't seem to want to show it (or I'm losing my marbles!)

PR #1312, opened 2019-08-21
Allow arrival times of sequential outputs to be specified to abc9 

eddiehung:
* Specify arrival times on cells by attaching `(* abc_arrival=<integer> *)` to port
* Removes `abc_scc_break` attribute as no longer necessary
* `RAMD*` and `SRL*` are now blackbox (with arrival times) to reflect (their more restrictive) sequential behaviour, but combinatorial behaviour is maintained by using a box
    * This process is achieved by calling `techmap -map abc_map.v` before `abc9` to separate those primitives into two internal ones that ABC can understand, and then `techmap -map abc_unmap.v` afterwards to undo this.
* Arrival times added to `FD*` and `RAM{D,B}* and SRL*` outputs for Xilinx
* Arrival times added to `SB_DFF*` and `SB_RAM40*` for ice40

Broken
- ~Shift register inference~ (fixed by #1321)

~Help needed for ecp5 timings.~
ECP5 timings deferred for a future PR.


techlibs/xilinx/abc_map.v
```diff
@@ -0,0 +1,125 @@
...
+  parameter [15:0] INIT = 16'h0000;
+  parameter [0:0] IS_CLK_INVERTED = 1'b0;
+  wire \$Q ;
+  SRL16E #(
+    .INIT(INIT), .IS_CLK_INVERTED(IS_CLK_INVERTED)
+  ) _TECHMAP_REPLACE_ (
+    .Q(\$Q ),
+    .A0(A0), .A1(A1), .A2(A2), .A3(A3), .CE(CE), .CLK(CLK), .D(D)
+  );
+  // TODO: Check if SRL uses fast inputs or slow inputs
```
eddiehung: TODO

techlibs/ecp5/abc_5g.box
```diff
@@ -15,16 +15,16 @@ CCU2C   1      1   9      3
...
-# name            ID  w/b   ins   outs
-TRELLIS_DPR16X4   2     0   14    4
-
-#DI0   DI1   DI2   DI3   RAD0   RAD1   RAD2   RAD3   WAD0    WAD1   WAD2   WAD3  WCK   WRE
--      -     -     -     141    379    275    379    -       -      -      -     -     -
--      -     -     -     141    379    275    379    -       -      -      -     -     -
--      -     -     -     141    379    275    379    -       -      -      -     -     -
--      -     -     -     141    379    275    379    -       -      -      -     -     -
+# name               ID  w/b   ins   outs
+$__ABC_RAM16X2_COMB  2     0   8    4
```
daveshah1: Shouldn't this be 16X4 (the actual implementation is indeed two 16x2 RAM SLICEs, but that's not what seems to be going on here)

techlibs/xilinx/abc_xc7.box
```diff
@@ -37,22 +39,15 @@ CARRY4 4 1 10 8
 580 526 507 398 385 508 528 378 380 114
 
 # SLICEM/A6LUT
-# Inputs: A0 A1 A2 A3 A4 D DPRA0 DPRA1 DPRA2 DPRA3 DPRA4 WCLK WE
+# Box to emulate comb/seq behaviour of RAMD{32,64} and SRL{16,32}
```
daveshah1: Do you think you could expand this comment a bit, for anyone that comes across this in the future?

daveshah1:
Although the attribute suffices for now imo, in the long term it doesn't represent configurable modules well:
 - ECP5 BRAM arrival times reduce from 5.8ns to 0.9ns (slowest -6 grade) by enabling the extra latency-2 output register (which admittedly Yosys can't infer yet) and I expect Xilinx would be similar
 - DSPs have all kinds of possibilities for registers inside them.

eddiehung:
> Though the attribute suffices for now imo, in the long term it doesn't represent configurable modules well:
> 
> * ECP5 BRAM arrival times reduce from 5.8ns to 0.9ns (slowest -6 grade) by enabling the extra latency-2 output register (which admittedly Yosys can't infer yet) and I expect Xilinx would be similar
> * DSPs have all kinds of possibilities for registers inside them.

Yes indeed. The plan is to have `abc_map.v` techmap those parameterised cells into non parameterised "internal" boxes that represents that specific combination of registers (and parameters), and then `abc_unmap.v` to undo that transformation. I'll be doing that first for the DSP branch.

In light of how ABC handles boxes, there's no getting round this fact, except to dress it up a bit nicer so that it doesn't impact the regular Yosys flow as much as possible, and perhaps to auto-generate those box descriptions from specify blocks in the future.

techlibs/ecp5/abc_5g.box
eddiehung: Good spot. That's just an internal name that ABC uses for the box, it doesn't affect the flow otherwise which is why it wasn't picked up.

PR #1313, opened 2019-08-21
Call abc/abc9 with '-F' option, so no need to echo ourselves

eddiehung:
To make debugging ABC and generating MCVEs a little easier.


eddiehung:
> Very well. Looks like not everyone shares my strong preference for `+ ` over `abc - > ` as prefix for those messages, and my equally strong preference for not including semicolons in those messages. (Which is literally the reason why I didn't want to use `-F` in the first place.)

Fair enough. This is not a must-have for me, but it does make working with Yosys-generated ABC scripts a little bit easier -- there's a little less clutter in the file, and when adding new options to ABC commands then I only need to change one line.

PR #1318, opened 2019-08-22
Intel M9K Fixes for SDP

dh73:
Hi,
This changes fixes the current templates for matching M9K block ram in Intel devices. A little explanation:

* **techlibs/intel/common/brams_m9k.txt**: Corrected rules for data and address memory dimensions supported by M9K.
* **techlibs/intel/common/brams_map_m9k.v**: Fixed technology mapping template that actually matches to a SDP.
* **techlibs/intel/common/m9k_bb.v**: Updated parameters needed for M9K in SDP mode.

Still initialization and clock enable/polarity are WIP.


eddiehung:
Can you add a few tests, preferably using vendor templates (with attribution), in a similar format to #1310, please?

ZirconiumX:
If I'm honest, I don't think `altsyncram` is the right primitive to use here; maybe `generic_m9k`.

dh73:
Hi,
The cell templates have to match with technology primitive names to be useful for the P&R tool. If Yosys provide a netlist with a primitive name that is not in the device database, it will introduce an error.

Whereas ```altsyncram``` is the universal wrapper for the memory elements of Intel devices (selectable by ```lpm_type``` parameter), I don't know or haven't hear any primitive ```generic_m9k``` that exists in the device database, therefore I'm not sure what is the advantage of renaming it.

There's one level below ```altsyncram``` that uses two 4K memories for each device, but for now whatever I'm pushing is generic (have to work for most devices) and specific modifications can be made from that easily. I'm not adding anything about M10K.

Thank you.

PS, just tried to find any primitive or parameter with suggested name in Quartus and doesn't seems to exist.

ZirconiumX:
There's a `generic_m10k`, and I was hoping there would be a `generic_m9k`, but no. I think the name you're looking for here is either `cycloneive_ram_block` or `fiftyfivenm_ram_block`.

PR #1320, opened 2019-08-22
do not require boost if pyosys is not used

mmicko:
Just a small update for macOS builds


Makefile
```diff
@@ -682,6 +689,7 @@ endif
 
 test: $(TARGETS) $(EXTRA_TARGETS)
 	+cd tests/simple && bash run-test.sh $(SEEDOPT)
+	+cd tests/simple_abc9 && bash run-test.sh $(SEEDOPT)
```
eddiehung: This looks like an unrelated change?

mmicko: uf, my bad, let me check, guess I did not update something well

PR #1325, opened 2019-08-22
In sat: 'x' in init attr should be ignored

eddiehung:
Attached testcase works if `sat -enable_undef` is specified, even though it shouldn't be necessary as there is no ambiguity.

Also removal of one colon is necessary otherwise signal is collapsed and no init.

### DIFF
--- a/passes/sat/sat.cc
+++ b/passes/sat/sat.cc
@@ -268,9 +268,7 @@ struct SatHelper
 				RTLIL::SigSpec removed_bits;
 				for (int i = 0; i < lhs.size(); i++) {
 					RTLIL::SigSpec bit = lhs.extract(i, 1);
-					if (bit.is_fully_const() && rhs[i] == State::Sx)
-						rhs[i] = bit;
-					if (!satgen.initial_state.check_all(bit)) {
+					if (rhs[i] == State::Sx || !satgen.initial_state.check_all(bit)) {
 						removed_bits.append(bit);
 						lhs.remove(i, 1);
 						rhs.remove(i, 1);
--- a/passes/sat/sat.cc
+++ b/passes/sat/sat.cc
@@ -268,6 +268,8 @@ struct SatHelper
 				RTLIL::SigSpec removed_bits;
 				for (int i = 0; i < lhs.size(); i++) {
 					RTLIL::SigSpec bit = lhs.extract(i, 1);
+					if (bit.is_fully_const() && rhs[i] == State::Sx)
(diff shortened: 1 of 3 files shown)


tests/sat/initval.v
```diff
@@ -11,5 +12,8 @@ module test(input clk, input [3:0] bar, output [3:0] foo);
   always @(posedge clk)
     last_bar <= bar;
 
+  always @*
+    asdf[2:0] <= 3'b111;
```
cliffordwolf: This should be using blocking assignment (`=`).

passes/sat/sat.cc
```diff
@@ -268,6 +268,8 @@ struct SatHelper
 				RTLIL::SigSpec removed_bits;
 				for (int i = 0; i < lhs.size(); i++) {
 					RTLIL::SigSpec bit = lhs.extract(i, 1);
+					if (bit.is_fully_const() && rhs[i] == State::Sx)
+						rhs[i] = bit;
```
cliffordwolf: I think this is not fixing the underlying problem, which is that we keep `init` attributes on things that aren't FFs. (If the init value is or isn't `x` doesn't really matter.) We fixed that for verific in part in 9260e97aa28b245ee88d81d162bb6b83cbc5eab0. (But even this is still not a satisfying solution for latches that may or may not be enabled in the first cycle. For this we'd also need changes to `async2sync` and `clk2fflogic`.) But this problem effects `sat` as well as `btor`, `smt2`, `smv`, and (I think) `aiger`. So I think just working around this in only some cases for only for `sat` is not a satisfying solution.

eddiehung: Consider my updated test case, which is a wire with part FF and part non-FF. The `(* init= *)` attribute has to be applied to the whole bus, thus a reasonable thing to do would be set to set the part of the signal that is not init-relevant to `1'bx`?

My testcase is contrived, granted, but I have seen it in the wild. The cause of this could be something that reduces to `asdf <= {foobar, 3'b111};` that causes the LSBs to be constant drivers.

cliffordwolf: Forget what I wrote in the previous comment. Actually we are using x-bits in init to denote unused (uninitialized) bits. So I think the right thing here would be to just ignore bits in the init vector that are `x`, regardless of if the signal is driven by a constant or not.

If in `-enable_undef` mode you like all uninitialized flops to start at `x` then that's done with `-set-init-undef`, independent of init attributes.

PR #1328, opened 2019-08-24
Don't filter commits by author when composing YOSYS_VER

whitequark:
I'm not sure why it tries to filter by email, but this seems wrong. This came to my attention because YOSYS_VER causes issues downstream, like in NixOS, which has to patch these constructs out as `.git` is not available.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,7 +115,7 @@ LDFLAGS += -rdynamic
 LDLIBS += -lrt
 endif
 
-YOSYS_VER := 0.8+$(shell cd $(YOSYS_SRC) && test -e .git && { git log --author=clifford@clifford.at --oneline 4d4665b.. 2> /dev/null | wc -l; })
+YOSYS_VER := 0.8+$(shell cd $(YOSYS_SRC) && test -e .git && { git log --oneline 4d4665b.. 2> /dev/null | wc -l; })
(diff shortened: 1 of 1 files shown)


cliffordwolf:
> I'm not sure why it tries to filter by email, but this seems wrong.

The reason for this is so it doesn't count revisions up when someone is forking yosys. Otherwise they report issues with say 0.8+691 when in fact it's 0.8+600 with 91 of their commits ontop of it.

whitequark:
> The reason for this is so it doesn't count revisions up when someone is forking yosys.

So if (say) Eddie makes a PR and I merge it, that doesn't count revisions up either. Which seems bad.

cliffordwolf:
> So if (say) Eddie makes a PR and I merge it, that doesn't count revisions up either. Which seems bad.

The idea is that I will commit often enough to give it a bump every now and then.

> Also, wouldn't the git hash disambiguate this anyway?

No, because when someone is forking of their branch then the git hash they have is nothing I can look up, so I will have no way of telling what version they have branched of from.

A better solution would probably be having a bot that just updates a version number or time stamp every night.

GCC has something like that: https://github.com/riscv/riscv-gcc/commit/bbafc6b5ba0ec7c105afec27272c4ac5d363b5cc

cliffordwolf:
See fdbcf78. For now I'll just run it from time to time. We can create a bot for it later.

PR #1332, opened 2019-08-27
ecp5: Add GSR and SGSR support

daveshah1:
This adds support for SGSR and GSR for ECP5, and also mirrors Diamond behaviour thus:

 - the GSR parameter is set to "DISABLED" when no GSR primitive is used and "ENABLED" otherwise. A non-standard override `(* nogsr *)` can also be used to override this
 - active low async FF resets connected to GSR are pruned

See also YosysHQ/nextpnr#319

cc @whitequark


whitequark:
Thank you, this looks excellent.

PR #1290, opened 2019-08-12
Wrap SB_LUT+SB_CARRY into $__ICE40_CARRY_WRAPPER (retry)

eddiehung:
Retry of #1266 

Culprit was: https://github.com/YosysHQ/yosys/commit/8a2480526f2ff80cbdf0f0f4279e8989b53b6e9b (sorry!)

CC: @smunaut

PR #1337, opened 2019-08-29
Fix $__ICE40_CARRY_WRAPPER, restore abc9 functionality

eddiehung:
Sorry for my incompetence! I've just realised that `abc9` ice40 has not been using boxes at all, and discovered these remnants from #1290 and #1260 which addressed #1205 originally.

~I've also had to remove optimising `$__ICE40_CARRY_WRAPPER` (or the deprecated `$__ICE40_FULL_ADDER`) from `ice40_opt` since we can't reason with it the same way -- it can't be guaranteed to implement the sum part of the full adder since the LUT mask could be different, and I0 and I3 could have be non-zero and non-CI respectively.
*EDIT:* Not sure if this is desirable for non-abc9 flows. For `-abc9` it is possible that abc9 may be able to reason about the box and optimise it away entirely (but not demote it back to a soft `$lut` and use that during techmapping). Perhaps it is better to restore its handling in `ice40_opt` and only operate on the "common" case of its use as a full-adder? Let me sit on it a little more, and RFC please.~

### DIFF
--- a/techlibs/ice40/abc_u.box
+++ b/techlibs/ice40/abc_u.box
@@ -8,6 +8,6 @@
 #   (NB: carry chain input/output must be last
 #        input/output and have been moved there
 #        overriding the alphabetical ordering)
-$__ICE40_FULL_ADDER 1 1 5 2
+$__ICE40_CARRY_WRAPPER 1 1 5 2
 1231 1205 1285 874 874
 675  609  -    -   278
--- a/techlibs/ice40/abc_u.box
+++ b/techlibs/ice40/abc_u.box
@@ -3,11 +3,11 @@
 # NB: Inputs/Outputs must be ordered alphabetically
 #     (with exceptions for carry in/out)
 
-# Inputs: A B CI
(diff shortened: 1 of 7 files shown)


eddiehung:
Okay on a rethink the `ice40_opt` is only analysing the `SB_CARRY` to remove it. Thus we can always demote the LUT part of the wrapper back to soft logic.

techlibs/ice40/abc_u.box
```diff
@@ -3,11 +3,11 @@
...
-# Inputs: A B CI
+# Inputs: A B I0 I3 CI
 # Outputs: O CO
 #   (NB: carry chain input/output must be last
 #        input/output and have been moved there
 #        overriding the alphabetical ordering)
-$__ICE40_FULL_ADDER 1 1 3 2
-1231 1205 874
-675  609  278
+$__ICE40_FULL_ADDER 1 1 5 2
```
whitequark: Aren't you removing `$__ICE40_FULL_ADDER` below? Or am I missing something?

PR #1338, opened 2019-08-29
hierarchy -auto-top to work with (* top *) modules from read/read_verilog -defer

eddiehung:
This PR propagates module attributes through to deferred modules (named `$abstract*`) -- I can't see why this would be undesirable?

This includes the `(* top *)` attribute, allowing `hierarchy -auto-top` to (with some changes) recognise that value on `$abstract*` modules and elaborate accordingly.

Fixes a regression likely caused by the introduction of #1233.


frontends/ast/ast.cc
```diff
@@ -1100,6 +1095,12 @@ static AstModule* process_module(AstNode *ast, bool defer, AstNode *original_ast
 		ignoreThisSignalsInInitial = RTLIL::SigSpec();
 	}
 
+	for (auto &attr : ast->attributes) {
+		if (attr.second->type != AST_CONSTANT)
+			log_file_error(ast->filename, ast->linenum, "Attribute `%s' with non-constant value!\n", attr.first.c_str());
+		current_module->attributes[attr.first] = attr.second->asAttrConst();
+	}
+
```
cliffordwolf: This will fail on attributes that have a compound expression as value instead of a literal.

```
(* foo = 1 + 2 *)
module test;
endmodule
```

```
$ ./yosys -Qp 'read_verilog -defer test.v'

-- Running command `read_verilog -defer test.v' --

Storing AST representation for module `$abstract\test'.
test.v:2: ERROR: Attribute `\foo' with non-constant value!
```

cliffordwolf: Hmm.. now it doesn't produce an error message if the expression is not constant. Maybe it would make sense to _only_ change the behavior for `top` attributes?

> The fact that it must be a constant expression should mean that it is evaluate-able even for deferred elaboration.

The whole idea of `-defer` is that we defer simplifying and evaluating the AST, but in order to evaluate attributes we'd need to do exactly that, since the expression for the attribute value may call functions, depend on parameters, etc.

eddiehung: > Hmm.. now it doesn't produce an error message if the expression is not constant.
Yes, attributes with those expressions are now ignored (just like before, when we were ignoring all such attributes) -- I've changed it to only propagate attributes that *are* constant.
This makes it an improvement over what we have in master right now, with nothing lost. 

> Maybe it would make sense to only change the behavior for top attributes?
Of course, the assumption is that `(* top=<constant literal> *)`. We can, of course, whitelist just the `top` attribute if that makes you more comfortable?

cliffordwolf: >  I've changed it to only propagate attributes that are constant.

Yes, I think 6510297712729f9742d00f6a4cd4fc3a0e530758 fixes the issue.

> For module attributes, can it access any parameters?

Icarus Verilog, Verilator, and Verific all seem to accept the following.

```
(* foo = bar + 1 *)
module test #( parameter integer bar = 0 );
endmodule
```

passes/hierarchy/hierarchy.cc
```diff
@@ -808,6 +808,37 @@ struct HierarchyPass : public Pass {
 				if (mod_it.second->get_bool_attribute("\\top"))
 					top_mod = mod_it.second;
 
+		if (top_mod != nullptr && auto_top_mode) {
```
cliffordwolf: I don't think this is interacting with `-auto-top` correctly.

The idea of `-auto-top` is that it finds a suitable top module when the design contains no marked top module. But this case is one where in fact the design contains a marked top-module. This would derive from a marked abstract top module when `-auto-top` is used, but not derive from it otherwise?!

It inherently doesn't make any sense imo to combine `-auto-top` with `-chparam`, but this seems to actually require to combine the two.

I think the following is the intended behavior.

```
                if (top_mod != nullptr && top_mod->name.begins_with("$abstract"))
                {
                        IdString top_name = top_mod->name.substr(strlen("$abstract"));

                        dict<RTLIL::IdString, RTLIL::Const> top_parameters;
                        for (auto &para : parameters) {
                                SigSpec sig_value;
                                if (!RTLIL::SigSpec::parse(sig_value, NULL, para.second))
                                        log_cmd_error("Can't decode value '%s'!\n", para.second.c_str());
                                top_parameters[RTLIL::escape_id(para.first)] = sig_value.as_const();
                        }

                        top_mod = design->module(top_mod->derive(design, top_parameters));

                        if (top_mod != nullptr && top_mod->name != top_name) {
                                Module *m = top_mod->clone();
                                m->name = top_name;
                                Module *old_mod = design->module(top_name);
                                if (old_mod)
                                        design->remove(old_mod);
                                design->add(m);
                                top_mod = m;
                        }
                }
```

tests/various/hierarchy_defer.ys
```diff
@@ -0,0 +1,19 @@
+read -vlog2k <<EOT
```
cliffordwolf: This should either be preceded by `read -noverific` or it should be `read_verilog` instead of `read`. Because here document don't work with Verific. (Yet. Feel free to add support for that.)

tests/various/hierarchy_defer.ys
```diff
@@ -0,0 +1,19 @@
...
+sub s0(i, o);
+endmodule
+
+(* constant_expression=1+1?2*2:3/3 *)
+module sub(input i, output o);
+assign o = ~i;
+endmodule
+EOT
+
+hierarchy -auto-top
```
cliffordwolf: `hierarchy -auto-top` is meant for cases where there is no dedicated top module (marked with the `top` attribute). That's the opposite of what you are testing here. So it should just be `hierarchy` without `-auto-top`.

eddiehung: The reason I am particularly interested with `-auto-top` is that all the `synth_*` commands call `hierarchy` with this option when there is no `-top`.
It is also my understanding that `-auto-top` is only treated as a fall-back when `(* top *)` cannot be find.

tests/various/hierarchy_defer.ys
eddiehung: The difference between `read` and `read_verilog` is that the former calls `read_verilog -defer`. Let me add `read -noverific` then.

passes/hierarchy/hierarchy.cc
eddiehung: >  This would derive from a marked abstract top module when -auto-top is used, but not derive from it otherwise?!

No, the intention is that -- with `(* top *)` present, it would behave the same with and without `-auto-top` since `synth_*` calls `hierarchy -auto-top` in the absence of a specific `-top`.

I see that my PR currently doesn't do that... let me fix it.

eddiehung: > It inherently doesn't make any sense imo to combine -auto-top with -chparam, but this seems to actually require to combine the two.

I can't see any reason why this wouldn't be desirable?

Remember, ultimately I want to see `read_verilog -defer` behave the same as `read_verilog` (equally, `read` before and after #1233) when there is a `(* top *)` involved, including with `-chparam`.

PR #1339, opened 2019-08-29
Add blackbox definitions for IBUFG, and Spartan6-only PLL_ADV, BUFIO2

eddiehung:
I've found these to be things that are instantiated by LiteX projects...


eddiehung:
Closed in favour of a more complete solution like what we have for `cells_xtra.py`

Issue #1341, opened 2019-08-29
0.9 generates .json parameter that nextpnr can't parse due to extra <space> character (worked in 0.8)


ret:
## Steps to reproduce the issue
See [gist files](https://gist.github.com/ret/3b762574d65b05736adf7d7382cc48b5) to help reproduce.

The issue is present in `Yosys 0.9+1 (git sha1 a0b666f2, clang 10.0.1 -fPIC -Os)`. Worked in `Yosys 0.8+     615 (git sha1 a02d1720, clang 10.0.1 -fPIC -Os)`

## Expected behavior
No extra spaces in .json parameter (PULLUP), i.e. `"1"`.

## Actual behavior
Extra space `"1 "` in .json parameter (PULLUP), which `nextpnr` fails on.

## Details
`Yosys 0.9+1 (git sha1 a0b666f2, clang 10.0.1 -fPIC -Os)` produces an extra space in `"1 "` for the `PULLUP` parameter below:
```json

        "uart_0__rx_0": {
          ...
          "parameters": {
            "IO_STANDARD": "SB_LVTTL",
            "PIN_TYPE": 1,
            "PULLUP": "1 "
          },
	  ...

        "uart_0__tx_0": {
          ...
          "parameters": {
            "IO_STANDARD": "SB_LVTTL",
            "PIN_TYPE": 25,
            "PULLUP": "1 "
          },
          ...
```

This trips up nextpnr to error out with
```console
ERROR: expected numeric value for parameter 'PULLUP' on cell 'uart_0__rx_0'
```

Version `Yosys 0.8+     615 (git sha1 a02d1720, clang 10.0.1 -fPIC -Os)` did not produce the extra space:
```json

        "uart_0__rx_0": {
          ...
          "parameters": {
            "IO_STANDARD": "SB_LVTTL",
            "PIN_TYPE": 1,
            "PULLUP": "1"
          },
	  ...

        "uart_0__tx_0": {
          ...
          "parameters": {
            "IO_STANDARD": "SB_LVTTL",
            "PIN_TYPE": 25,
            "PULLUP": "1"
          },
          ...
```
Reverting https://github.com/YosysHQ/yosys/commit/15fae357f6c7a5c205d2733cd0b2640eecd17a2d removes the problem (as a workaround).

I believe https://github.com/YosysHQ/yosys/blob/master/backends/json/json.cc#L102 may be the problematic line and will send a separate pull request to change it to add the space only if `state == 1`.

[...]

PR #1342, opened 2019-08-29
Fix extra space issue in parameter of json backend.

ret:
This pull request fixes (what I think is) the root cause of [1341](https://github.com/YosysHQ/yosys/issues/1341). The yosys, nextpnr-ice40, and -ecp5 tests pass locally, but I'm nervous that this change may affect parts of the toolchain I'm unaware of. Please think of it as a mere starting point for a real fix.

### DIFF
--- a/backends/json/json.cc
+++ b/backends/json/json.cc
@@ -99,7 +99,7 @@ struct JsonWriter
 				} else if (state == 1 && c != ' ')
 					state = 2;
 			}
-			if (state < 2)
+			if (state == 1)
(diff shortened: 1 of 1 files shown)


ret:
Per discussion of issue [1342](https://github.com/YosysHQ/yosys/pull/1342), this may need to a change to the nmigen-board definition for the iCEBreaker UART so nMigen won't generate `PULLUP "1"`, but `1'b1`, or `1` (anything, but a string), instead.

PR #1343, opened 2019-08-30
Add/update every Diamond FF primitive

whitequark:
This should now be complete, although latches are still missing (I'm not sure how to make a latch with Trellis).


daveshah1:
> I'm not sure how to make a latch with Trellis

The only hard latches in the ECP5 are in the IO tiles but I don't think these have been fuzzed yet. Diamond uses soft logic for all other primitives. This could be implemented similar how to Yosys creates soft latches: https://github.com/YosysHQ/yosys/blob/master/techlibs/ecp5/latches_map.v

There is actually an undocumented "async preload" mode of the flipflops that could be used to create hard latches. But I'm a bit wary relying on this because Diamond's synthesis never uses it (the only way to activate it is with the undocumented ncl route to create post pnr designs). I don't think the nextpnr support for this is complete either, but maybe if someone has an interesting use case I can work on this more.

Issue #1211, opened 2019-07-19
ice40: SB_MAC model incorrect for signed inputs


eddiehung:
## Steps to reproduce the issue

`techlibs/ice40/tests/test_dsp_model.sh` (which compares `cells_sim.v` model with vendor sim model) fails on new branch https://github.com/YosysHQ/yosys/compare/eddie/signed_ice40_dsp containing new test cases where `A_SIGNED` or `B_SIGNED` are set.

## Expected behavior

Pass.

## Actual behavior

Fail.

PR #1344, opened 2019-08-30
ice40_dsp to allow signed multipliers

eddiehung:
@cliffordwolf what was the reason behind why signed multipliers were rejected in: https://github.com/YosysHQ/yosys/commit/dca65d83a0037539464d303ea8751a3e06a92e03?

Also modified existing testcase to be wider to go over our thresholds for DSP inference.


daveshah1:
I remember fixing a signedness bug in the sim model a month or too ago (looks like it did land in master too). I don't know if that was part of the problem?

eddiehung:
> I remember fixing a signedness bug in the sim model a month or two ago (looks like it did land in master too). I don't know if that was part of the problem?

Thanks David. Perhaps #1211 was the original cause. It seems to pass `equiv_opt` now.

PR #1355, opened 2019-09-04
peepopt_dffmux -- bit optimisations for word level $dff + (enable/reset) $mux cells

eddiehung:
`wreduce` has the ability to trim the size of `$dff` cells, if the `$dff` is just capturing a sign-extended result.

However, this ability does not work for `$dff` + `$mux` combos that are used to describe an flop-with-enable (and which would be transformed into `$dffe` cells by `dff2dffe`).
EDIT: Also added support for optional reset mux.

This is a pmgen-based peepopt that performs this optimisation, and I've found it to be necessary for DSP register packing.


eddiehung:
TODO: Move peepopt tests to various and use `equiv_opt` as well as `select`.

tests/various/peepopt.ys
```diff
@@ -0,0 +1,63 @@
...
+endmodule
+EOT
+
+prep -nokeepdc
+equiv_opt peepopt
+design -load postopt
+clean
+select -assert-count 1 t:$shr
+select -assert-count 1 t:$mul
+select -assert-count 0 t:$shr t:$mul %% t:* %D
```
eddiehung: @cliffordwolf can you check that `peepopt` rightfully cannot optimise this `$shr`?
I wasn't satisfied with the existing tests being in `tests/simple` (which just checks functionality through sim) and moved them to `tests/various` which uses `equiv_opt` as well as `select -assert-count` to ensure that things really are optimised away...

passes/pmgen/peepopt_shiftmul.pmg
```diff
@@ -50,8 +50,9 @@ code
 	if (GetSize(const_factor_cnst) > 20)
 		reject;
 
-	if (GetSize(port(shift, \Y)) > const_factor)
-		reject;
+	if (shift->type.in($shift, $shiftx))
+		if (GetSize(port(shift, \Y)) > const_factor)
+			reject;
```
cliffordwolf: The following script demonstrates that disabling this check for `$shr` shift cells yields incorrect results.

```
read_verilog <<EOT
module test(input [11:0] D, input [1:0] S, output [11:0] Y);
	assign Y = D >> (S*3);
endmodule
EOT

prep

design -save gold
peepopt
design -stash gate

design -import gold -as gold test
design -import gate -as gate test

miter -equiv -make_assert -make_outputs -ignore_gold_x -flatten gold gate miter
sat -show-public -enable_undef -prove-asserts miter
```

Afaict it's only okay to skip this check if `const_factor` is a power of two. I think in that case it should be okay to skip it for all shift cell types, not just `$shr`.

eddiehung: Okay, I've reverted e2c2d78 now, and added this as a new test.

tests/various/peepopt.ys
eddiehung: @cliffordwolf This remains the sticking point. In the above test and in the new one you gave me that I added in a2eee9eb, it seems that no peep optimisation is being done -- since the design wasn't changed, of course it will remain functionally equivalent. However, we have a multiplier before, and a multiplier after `peepopt`. 
This means that:
1. Either no peepopt is possible with these testcases
2. peepopt is broken
I think it would be more efficient for you to look at at this, please.

eddiehung:
To preserve chronological ordering of the comment train: the latest status is that this PR is held up as explained above (https://github.com/YosysHQ/yosys/pull/1355#discussion_r324399652)

tests/various/peepopt.ys
cliffordwolf: I've double checked `peepopt_shiftmul_1` and it's intended that this is not optimized. There are actually two reasons for not optimizing this case: (1) the output is larger than the constant factor, so padding the 3-bit chunks to 4 bit would cause incorrect result (at least the way we do padding) and (2) the multiply output has a small bit-width so there may be overflows that would not be handled correctly by our replacement circuit.

(1) is checked by `(GetSize(port(shift, \Y)) > const_factor)` and (2) is checked by `(GetSize(shamt) < factor_bits+GetSize(mul_din))`.

eddiehung: Thank you. I shall go ahead and merge this, unless you want to write a test that is eligible to be optimised.

PR #1356, opened 2019-09-04
Use $(shell :; ...) in Makefile to force shell

emilazy:
Did you think that `$(shell command -v ...)` would actually get run by
the shell? Foolish mortal; GNU Make is obviously far more wise than
thee, as it optimizes it to a direct -- and hence broken (since
`command` is a shell builtin) -- exec. This horrifying contortion
ensures that an actual shell runs the command and fixes the behaviour.

@Shizmob found the source of this misbehaviour; turns out gmake has a
hard-coded, incomplete list of shell builtins:

https://github.com/mirror/make/blob/715c787dc69bac37827a7d6ea6d40a86c55b5583/src/job.c#L2691

This contains `command`, but the whole function is full of horrible
heuristic garbage so who knows. I'm so sorry.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,7 +88,7 @@ ifeq ($(OS), Darwin)
 PLUGIN_LDFLAGS += -undefined dynamic_lookup
 
 # homebrew search paths
-ifneq ($(shell command -v brew),)
+ifneq ($(shell :; command -v brew),)
 BREW_PREFIX := $(shell brew --prefix)/opt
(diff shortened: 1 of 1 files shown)


emilazy:
Mystery solved:

```
00:49 <Shiz> emily: https://github.com/mirror/make/blob/4.2.1/job.c#L2508
00:49 <Shiz> no command in that list in 4.2.1
```

So I guess the next GNU Make version will have slightly less buggy behaviour in this one particular case. What a mess.

PR #1359, opened 2019-09-07
DSP inference for Xilinx (improved for ice40, initial support for ecp5)

eddiehung:
`synth_xilinx` now automatically infers `DSP48E1` cells, support multiply with (any?) combination of `([ABCDMP]|AD)REG=[12]` (with and without reset and enable), pre-adder, post-adder/accumulation, with C->P connections converted to PCOUT->PCIN cascading where possible, with limited [AB]COUT -> [AB]CIN cascading too. `-nodsp` to disable.

Also containing two (even more) experimental DSP48E1 features:
- Use SIMD mode based on presence of a `(* use_dsp="simd" *)` attribute
- Automatic inference of pattern detector to detect overflow (terms and conditions apply), based on presence of a greater-than-or-equal `>=` operation on the (registered) P output

See `help xilinx_dsp` for more details.

`synth_ecp5` also infers `MULT18X18D`. `-nodsp` to disable.

In all cases, wide multipliers are now decomposed into several multiplier and shift-add cells.
In the xilinx and ice40 cases, these shift-adds are packed into the primitive where possible.


tests/xilinx/macc.ys
```diff
@@ -0,0 +1,31 @@
+read_verilog macc.v
+design -save read
+
+proc
+hierarchy -top macc
+#equiv_opt -assert -map +/xilinx/cells_sim.v synth_xilinx ### TODO
```
eddiehung: TODO

passes/pmgen/xilinx_dsp.cc
```diff
@@ -0,0 +1,597 @@
...
+		}
+		else if (st.ffP && st.sigC == st.sigP)
+			opmode[4] = State::S0;
+		else
+			opmode[4] = State::S1;
+		opmode[6] = State::S0;
+		opmode[5] = State::S1;
+
+		if (opmode[4] != State::S0) {
+			//if (st.postAddMuxAB == "\\A")
```
eddiehung: FIXME

passes/pmgen/xilinx_dsp.pmg
```diff
@@ -0,0 +1,621 @@
...
+	filter nusers(sigP.extract_end(GetSize(unextend(port(postAdd, AB))))) <= 1
+	set postAddAB AB
+	optional
+endmatch
+
+code sigC sigP
+	if (postAdd) {
+		sigC = port(postAdd, postAddAB == \A ? \B : \A);
+
+		// TODO for DSP48E1, which will have sign extended inputs/outputs
```
eddiehung: TODO

passes/pmgen/ice40_dsp.cc
```diff
@@ -135,21 +137,43 @@ void create_ice40_dsp(ice40_dsp_pm &pm)
...
 
 	// SB_MAC16 Output Interface
 
-	SigSpec O = st.ffS ? st.sigS : st.sigY;
+	SigSpec O = st.sigO;
+	int O_width = GetSize(O);
+	if (O_width == 33) {
+		log_assert(st.addAB);
+		// If we have a signed multiply-add, then perform sign extension
+		// TODO: Need to check CD[31:16] is sign extension of CD[15:0]?
```
eddiehung: TODO

passes/pmgen/ice40_dsp.pmg
```diff
@@ -1,163 +1,272 @@
...
 
 		clock = c;
 		clock_pol = cp;
-		clock_vld = true;
 	}
+
+	sigO = sigH;
 endcode
 
 match addA
```
eddiehung: TODO: Use `choice`

passes/pmgen/ice40_dsp.pmg
```diff
@@ -1,163 +1,272 @@
...
+		sigO = port(addAB, \Y);
 	}
 endcode
 
 match muxA
-	if addAB
 	select muxA->type.in($mux)
-	select nusers(port(muxA, \A)) == 2
-	index <SigSpec> port(muxA, \A) === port(addAB, \Y)
+	index <int> nusers(port(muxA, \A)) === 2
```
eddiehung: TODO: Revert back to `select`

passes/pmgen/ice40_dsp.pmg
```diff
@@ -1,163 +1,272 @@
...
 	optional
 endmatch
 
 match muxB
-	if addAB
 	if !muxA
 	select muxB->type.in($mux)
-	select nusers(port(muxB, \B)) == 2
-	index <SigSpec> port(muxB, \B) === port(addAB, \Y)
+	index <int> nusers(port(muxB, \B)) === 2
```
eddiehung: TODO: Revert back to select

tests/xilinx/macc.ys
eddiehung: Why doesn't his work?!

daveshah1: I expect this is a problem with the clock enable meaning there are combinations of inputs for which the outputs of gate and gold will never synchronise, causing induction to fail

techlibs/xilinx/abc_xc7.box
```diff
@@ -59,3 +59,1107 @@ $__ABC_LUT6 2000 0 7 1
...
+#   disconnecting A when AREG, MREG or PREG is enabled)
+#   this mux captures the existence of a replacement path
+#   between AREG/BREG/CREG/etc. and P/PCOUT.
+#   Since the Aq/ADq/Bq/etc. inputs are assumed to arrive at
+#   the mux at zero time, the combinatorial delay through
+#   these muxes thus represents the clock-to-q delay at
+#   P/PCOUT.
+$__ABC_DSP48E1_MULT_P_MUX 2100 0 103 48
+# A AD    B    C D                                                                                               I    M P                                                                                                Pq
+2952 - 2813 1687 - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - 1671 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 329
```
daveshah1: What about a Python script to generate this massive array of numbers (maybe I have missed one?)

eddiehung: Right now, this information is imported from a spreadsheet by hand. The vision is to generate box/lut files automatically, but until that day, I don't believe there's a lot of value in writing and maintaining a script to generate this information since it is expected to be constant once finalised. If someone else wants to do one and maintain it, I'd be happy to help.

tests/xilinx/macc.ys
eddiehung: Please forgive my ignorance, but can you elaborate on this a little? Any clock enable input would be fed equally to gate and gold, which would be expected to implement identical behaviour, no?

daveshah1: The initial state will be random and not necessarily the same between gate and gold. Induction checks that, given the output is the same for K cycles, it will be the same at cycle K+1. As one possible input sequence is CE low for those K cycle and high on K+1 (i.e. so the state between the two never synchronises); induction will not pass

eddiehung: I see. That makes a lot of sense, thanks! Is there any way to guide induction to throw a CE transition or two in there as part of those K cycles, and whether that would help things at all?

PR #1362, opened 2019-09-09
MSVC2 fixes

xobs:
This fixes two issues with Yosys when building for MSVC2:

1. The Makefile assumes the compiler is always called `gcc`, which isn't always true.
2. Patch 3d3779b0376b8204ed7637053176a07b7271ac1d added a static map that contains a static class with static members, which might get initialized in an incorrect order.

### DIFF
--- a/backends/smt2/Makefile.inc
+++ b/backends/smt2/Makefile.inc
@@ -16,7 +16,7 @@ yosys-smtbmc-script.py: backends/smt2/smtbmc.py
 		-e "s|#!/usr/bin/env python3|#!$(PYTHON)|" < $< > $@
 
 yosys-smtbmc.exe: misc/launcher.c yosys-smtbmc-script.py
-	$(P) gcc -DGUI=0 -O -s -o $@ $<
+	$(P) $(CXX) -DGUI=0 -O -s -o $@ $<
 # Other targets
 else
 TARGETS += yosys-smtbmc
--- a/misc/launcher.c
+++ b/misc/launcher.c
@@ -65,7 +65,7 @@ SOFTWARE. */
 
 int child_pid=0;
 
-int fail(char *format, char *data) {
+int fail(const char *format, const char *data) {
     /* Print error message to stderr and return 2 */
     fprintf(stderr, format, data);
     return 2;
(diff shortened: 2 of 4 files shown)


xobs:
This fixes yosys so it now produces a working binary.

I am not sure why all checks have failed -- I haven't changed anything that impacts those systems.

daveshah1:
The Travis failure is a small (non problematic) QoR change in the new arch tests which require an exact number of LUTs to be synthesised, probably from a previous patch. (I said this would be a PITA at the time this check was added, but no one listened to me...)

mmicko:
This looks fine, static initialization order is UB so this is proper way of doing it. Thanks

eddiehung:
> The Travis failure is a small (non problematic) QoR change in the new arch tests which require an exact number of LUTs to be synthesised, probably from a previous patch. (I said this would be a PITA at the time this check was added, but no one listened to me...)

My opinion is to disagree. Any change of QoR, for better or worse, is important to know, especially in light of a lack of any other QoR system currently. Yes, it makes it more of a pain to maintain, but the arguments for it include:
1. I want to know that all platforms and compilers and return the same exact result (not simply that it falls under the upper bound)
2. Developers should rightfully be alerted to any and all changes. Perhaps some new functionality they added had an unintended (but still beneficial) effect (which would otherwise fall under the radar) and could be helpful visibility into indirect tasks such as debugging UB. Or it could be helpful for identifying an optimisation opportunity that could be applied more broadly.
3. The test itself may not be correct/sensitive enough. Optimising everything away to zero resources may mean that the output is constant-zero and has fewer resources than the bound, and that would be a false negative.

This is so far the first case where it's come to bite us. I would like to give it a little more time before deciding whether the RoI is not worth it.

xobs:
I made the change as suggested by @eddiehung .  I'm not sure why my changes caused this test to fail, but it ought to pass now.

eddiehung:
> I made the change as suggested by @eddiehung . I'm not sure why my changes caused this test to fail, but it ought to pass now.

I imagine we were "relying" on UB before, which I hope is no longer the case now!

PR #1363, opened 2019-09-09
Add tests for Xilinx architecture

SergeyDegtyar:


tests/xilinx/latches.ys
```diff
@@ -0,0 +1,19 @@
...
+design -load postopt # load the post-opt design (otherwise equiv_opt loads the pre-opt design)
+
+design -load read
+
+synth_xilinx
+flatten
+cd top
+select -assert-count 1 t:LUT1
+select -assert-count 2 t:LUT3
+select -assert-count 3 t:$_DLATCH_P_
```
daveshah1: Given this is obviously broken (`$_DLATCH_P_` is a Yosys cell that Vivado won't recognise), is this really a useful test?

tests/xilinx/tribuf.ys
```diff
@@ -0,0 +1,11 @@
+read_verilog tribuf.v
+hierarchy -top top
+proc
+tribuf
+flatten
+synth
+equiv_opt -assert -map +/xilinx/cells_sim.v -map +/simcells.v synth_xilinx # equivalency check
+design -load postopt # load the post-opt design (otherwise equiv_opt loads the pre-opt design)
+cd top # Constrain all select calls below inside the top module
+select -assert-count 1 t:$_TBUF_
```
daveshah1: Comment about `$_DLATCH_*_` also applies here

tests/xilinx/adffs.ys
```diff
@@ -0,0 +1,14 @@
+read_verilog adffs.v
+proc
+async2sync # converts async flops to a 'sync' variant clocked by a 'super'-clock
```
daveshah1: If this is an attempt to test mapping of asynchronous flipflops, then the `async2sync` here defeats that purpose by making the flipflops synchronous.

In any case, if you did want to test asynchronous flipflops properly, the Xilinx arch doesn't support both async set and reset so these would be left unmapped.

tests/xilinx/latches.ys
SergeyDegtyar: I checked this case in Xilinx Vivado and it synthesize LDCE cells, but Yosis don't. I think, this test is useful and will be updated when LDCE will be supported. Or I'm not right?

tests/xilinx/tribuf.ys
SergeyDegtyar: The same thoughts as with latches. Vivado synthesizes OBUFT cells. Maybe, I should report new issues to support this cases?

tests/xilinx/adffs.ys
SergeyDegtyar: Xilinx Vivado synthesizes FDCE cells that have asynchronous reset for this case. And when run synth_xilinx without async2sync it also synthesizes FDCE cells, but I have an: "ERROR: Found 8 unproven $equiv cells in 'equiv_status -assert'" from equiv_opt -assert command. And this test fails.

SergeyDegtyar: Does Yosys have any optimizations for this case? Vivado synthesizes two ffs and each of them have one control.

tests/xilinx/latches.ys
daveshah1: The concern is someone seeing this test might think it is correct behaviour to have `$_DLATCH_P_` cells at the end of synthesis. Perhaps you could add a comment and/or create an issue?

tests/xilinx/adffs.ys
daveshah1: See #1370 for a solution to the equivalence checking issue by adding `-multiclock` to `equiv_opt`.

> "Xilinx arch doesn't support both async set and reset" Does Yosys have any optimizations for this case? Vivado synthesizes two ffs and each of them have one control.

The solution Vivado uses is a horrible hack involving a LUT used as a latch, iirc, to work around the lack of native support. But I don't think this is something we want to emulate in Yosys, at least by default.

tests/xilinx/latches.ys
SergeyDegtyar: OK. Will add a comment for expected behavior. And a new issue.

tests/xilinx/adffs.ys
SergeyDegtyar: OK. Thank you for your description. Will review this test. Maybe will remove both set/reset usage and will add synchronous controls.

eddiehung: Sorry `async2sync` was my suggestion. I think what I really wanted was `clk2fflogic` which is dealt with in #1370.

tests/xilinx/latches.ys
SergeyDegtyar: An issue for LDCE cell support:
[](https://github.com/YosysHQ/yosys/issues/1387)

tests/xilinx/tribuf.ys
SergeyDegtyar: An issue for BUFT cell support:
[#1388 ](https://github.com/YosysHQ/yosys/issues/1388)

tests/xilinx/run-test.sh
```diff
@@ -6,7 +6,7 @@ for x in *.ys; do
 	echo "all:: run-$x"
 	echo "run-$x:"
 	echo "	@echo 'Running $x..'"
-	echo "	@../../yosys -ql ${x%.ys}.log $x"
+	echo "	@../../yosys -ql ${x%.ys}.log $x -w 'Yosys has only limited support for tri-state logic at the moment.'"
```
mmicko: Please move $x at end of line, like in https://github.com/YosysHQ/yosys/pull/1409/commits/7f0eec8270dbef06e3e6970af20dae2d6f89f6b9

tests/xilinx/latches.ys
eddiehung: Please update test with: https://github.com/YosysHQ/yosys/pull/1396/commits/8c2dd518b908644c9181526e193b8211734e8cef

tests/xilinx/tribuf.ys
eddiehung: Please add issue number to comment

eddiehung: Please add issue link to comment

tests/xilinx/latches.ys
eddiehung: Resolved in #1420

tests/xilinx/tribuf.ys
eddiehung: Resolved in #1420

tests/xilinx/run-test.sh
eddiehung: Resolved in #1420

eddiehung:
Continued in #1420

PR #1367, opened 2019-09-10
Implement macromodule inlining

andres-erbsen:
Proposed commit message:

> The `macromodule` keyword can now be used instead of `module` in places
> where `module` was previously accepted. Instantiations of such modules
> will be synthesized by inlining. Further, macromodule inlining is
> performed before task/function inlining, so tasks defined in a
> macromodule can be called from its parent module. This last behavior is
> experimental, undocumented, and may disappear at any time.

This is my first contribution to yosys. I think it is far along enough to be worth reviewing, but not ready to be merged quite yet. There are a few specific todo-s I would like advice on:

- [ ] are the changes to `ast.cc` a good way to keep around the unsimplified AST of each module?
- [ ] should children of inlined modules be always appended to the current module, not the current block?
- [ ] should `prefixWith` do something different for `GENBLOCK`?
- [ ] is it important to support inlining for `macromodule` instantiations whose parameters are set using `defparam`?
- [ ] are the tests sufficient?
- [ ] TODO store macromodule/module distinction in a better-named field than `is_signed`, but what?

### DIFF
--- a/frontends/ast/ast.cc
+++ b/frontends/ast/ast.cc
@@ -940,6 +940,7 @@ static AstModule* process_module(AstNode *ast, bool defer, AstNode *original_ast
 	current_module->attributes["\\src"] = stringf("%s:%d", ast->filename.c_str(), ast->linenum);
 	current_module->set_bool_attribute("\\cells_not_processed");
 
+	ast = ast->clone();
 	current_ast_mod = ast;
 	AstNode *ast_before_simplify;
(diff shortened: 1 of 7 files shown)


cliffordwolf:
Hello! Thanks for this submission. It's been a while since there was last some activity on this PR and it looks to have been abandoned. Please leave a comment if there's any new information.

Issue #1364, opened 2019-09-09
Parsing single Hex digits followed by ?


jakobwenzel:
## Steps to reproduce the issue

I am trying to parse some Verilog code:

```
module hexparse();
wire o;
wire a;
wire b;
wire [3:0] i;
assign o = i == 4'hb ? a:b;
endmodule
```
## Expected behavior

Yosys successfully parses the module.

## Actual behavior

Yosys errors out:

```
xyz.v:6: ERROR: syntax error, unexpected TOK_ID
```

The ? seems to get pulled into the number that is checked against. This only seems to happen for `4'hb` and `4'hd`, not for others.

jakobwenzel:
https://github.com/YosysHQ/yosys/blob/fda94311ee6e3ec8de0e85e91251a2744673abaf/frontends/verilog/verilog_lexer.l#L242

The Radix characters allowed to appear many times. Shouldn't this be 
```
[0-9]*[ \t]*\'s?[bodhBODH]?[ \t\r\n]*[0-9a-fA-FzxZX?_]+ { 
```
?

Issue #1022, opened 2019-05-21
Error parsing systemverilog integer literal constants 


AaronKel:
## Expected behavior

I expect the following code

```
module test(
  output logic [31:0] b
);

  assign b = '0;

endmodule
```

to be equivalent to

```
module test(
  output logic [31:0] b
);

  assign b = 32'b0;

endmodule
```

Looking at the IEEE Std 1800-2017 - 5.7.1 Integer literal constants 

> An unsized single-bit value can be specified by preceding the single-bit value with an apostrophe ( ' ), but without the base specifier. All bits of the unsized value shall be set to the value of the specified bit. In a self-determined context, an unsized single-bit value shall have a width of 1 bit, and the value shall be treated as unsigned.

## Actual behavior

Parsing SystemVerilog input from `test.v' to AST representation.
test.v:6: ERROR: syntax error, unexpected $undefined

AaronKel:
This is also the case for '1, 'x and 'z.

sumit0190:
I encountered the same problem a couple of days ago and made this half-researched attempt to fix it:

```diff
diff --git a/frontends/verilog/const2ast.cc b/frontends/verilog/const2ast.cc
index 7848c626..13145212 100644
--- a/frontends/verilog/const2ast.cc
+++ b/frontends/verilog/const2ast.cc
@@ -224,7 +224,12 @@ AstNode *VERILOG_FRONTEND::const2ast(std::string code, char case_type, bool warn
 			my_strtobin(data, endptr+2, len_in_bits, 16, case_type);
 			break;
 		default:
-			return NULL;
+            char next_char = char(tolower(*(endptr+1)));
+            if (next_char == '0' || next_char == '1' || next_char == 'x' || next_char == 'X') {
+                my_strtobin(data, endptr+2, -1, 2, case_type);
+            } else {
+	          return NULL;
+            }
 		}
 		if (len_in_bits < 0) {
 			if (is_signed && data.back() == RTLIL::S1)
diff --git a/frontends/verilog/verilog_lexer.l b/frontends/verilog/verilog_lexer.l
index 142d05d4..d89e144a 100644
--- a/frontends/verilog/verilog_lexer.l
+++ b/frontends/verilog/verilog_lexer.l
@@ -232,7 +232,7 @@ YOSYS_NAMESPACE_END
 	return TOK_CONSTVAL;
 }
 
[...]

PR #1373, opened 2019-09-12
Fix lexing of integer literals

cliffordwolf:
This reverts a change introduced in 34417ce55 by @mmicko that caused #1364. It also adds support for uppercase `S` as signedness indicator.


mmicko:
@cliffordwolf Well now originally fixed issue is broken again https://github.com/YosysHQ/yosys/issues/1022
Plain unsized  works :  assign b = 'b0; 
But unsized and "untyped" still does not : assign b = '0;

cliffordwolf:
> But unsized and "untyped" still does not : assign b = '0;

Ah! I see.. It should have been `?` instead of `*`. I think now it should be fine.

PR #1374, opened 2019-09-12
Fix two non-deterministic behaviours that cause divergence between compilers

### LINKED ISSUE #1371: Variation on synthesis output based on compiler and compiler flags

This replication is intended to be run on linux, but I expect any attempt to compare output from clang-compiled yosys and gcc-compiled yosys will replicate the issue.

1) Untar [yosys_variation.tar.gz](https://github.com/YosysHQ/yosys/files/3602665/yosys_variation.tar.gz)
2) Run "./make_conda_yosys.sh" to build a gcc and clang compiled version
3) Run "./run_all.sh" to compile example circuit under both versions of yosys

Examples of difference in output:
```
6273c6273
< Extracted 12460 gates and 14102 wires to a netlist network with 1640 inputs and 1387 outputs.
---
> Extracted 12459 gates and 14101 wires to a netlist network with 1640 inputs and 1387 outputs.
```

```
6733,6734c6746,6747
<    Number of wires:               4784
<    Number of wire bits:           8096
---
>    Number of wires:               4776
>    Number of wire bits:           8088
6740c6753
<    Number of cells:               6248
---
>    Number of cells:               6240
6748c6761
<      SB_LUT4                      4607
---
>      SB_LUT4                      4599
```

This issue was original replicated between clang and gcc, but even with the same gcc with different flags results in different outputs.

Known variations:
1) `gcc -fPIC -Os`
2) `clang-7 -fPIC -Os`
3) `clang-8 -fPIC -Os`
4) `gcc -fvisibility-inlines-hidden -fmessage-length=0 -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -fPIC -Os` 

## Expected behavior

Same input files, at the same git commit, outputs the same final circuit.

## Actual behavior

Variation on circuit output of the same input, at the same git commit, is unexpected.

eddiehung:
1. Default Compare function of `SigSet<T, std::less<T>>` causes pointer addresses to be compared leading to non-deterministic behaviour.
2. Code inside `alumacc` allows compiler the free choice of function call order:
https://github.com/YosysHQ/yosys/blob/4da6e19fe1c326bb86e68f4446597c2163f3fbb2/passes/techmap/alumacc.cc#L58
The problem with this is that `Or()`, `get_lt()`, etc. have side effects (i.e. use `NEW_ID` which increments `autoidx`) thus causing signals and cells to be named differently and thus a different downstream trajectory.

Fixes #1371

### DIFF
--- a/kernel/consteval.h
+++ b/kernel/consteval.h
@@ -33,7 +33,7 @@ struct ConstEval
 	SigMap assign_map;
 	SigMap values_map;
 	SigPool stop_signals;
-	SigSet<RTLIL::Cell*> sig2driver;
+	SigSet<RTLIL::Cell*, RTLIL::sort_by_name_id<RTLIL::Cell>> sig2driver;
 	std::set<RTLIL::Cell*> busy;
 	std::vector<SigMap> stack;
 	RTLIL::State defaultval;
--- a/kernel/consteval.h
+++ b/kernel/consteval.h
@@ -33,7 +33,7 @@ struct ConstEval
 	SigMap assign_map;
 	SigMap values_map;
 	SigPool stop_signals;
-	SigSet<RTLIL::Cell*, RTLIL::sort_by_name_id<RTLIL::Cell>> sig2driver;
+	SigSet<RTLIL::Cell*> sig2driver;
 	std::set<RTLIL::Cell*> busy;
 	std::vector<SigMap> stack;
 	RTLIL::State defaultval;
--- a/kernel/sigtools.h
+++ b/kernel/sigtools.h
@@ -138,6 +138,8 @@ struct SigPool
 template <typename T, class Compare = std::less<T>>
 struct SigSet
 {
+	static_assert(!std::is_pointer<T>::value || !std::is_same<Compare, std::less<T>>::value, "Explicit `Compare' class require for SigSet with pointer-type values!");
+
 	struct bitDef_t : public std::pair<RTLIL::Wire*, int> {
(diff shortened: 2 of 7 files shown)


whitequark:
> I guess that kind of performance optimization isn't a pressing issue for `SigSet<>`, but we wouldn't want to add similar restrictions for more fundamental container types.

IMO determinism is far more important than performance, so nondeterministic sorting order should only be opt-in.

eddiehung:
> However, it can be argued that it's fine to sort by pointer values in cases where one never iterates over the set. (And then it wouldn't be necessary to dereference two pointers in order to compare them.) 
Admittedly I didn't check before, but I just over all the instances of `SigSet<Cell*>` above and value-set is iterated over in all cases.

> IMO determinism is far more important than performance, so nondeterministic sorting order should only be opt-in.
100% agree, which is why I went as far as to ban it outright.

However, I've now added some template magic so that `SigSet<Cell*>` defaults to `sort_by_id`, but still allows `SigSet<Cell*, std::less<Cell*>>` for those who are really masochistic.

eddiehung:
#1371 reporter confirms issue is resolved now. Will merge when CI goes green.

cliffordwolf:
> IMO determinism is far more important than performance, so nondeterministic sorting order should only be opt-in.

Just on a side note: This is only an issue because SigSet is still using `std::set`. The iterator order for `dict` and `pool` only depends on insertion order.

PR #1377, opened 2019-09-13
Fix handling of z_digit "?" and fix optimization of cmp with "z"

cliffordwolf:
See https://github.com/m-labs/nmigen/issues/202#issuecomment-530329710 for details.

### DIFF
--- a/passes/opt/opt_expr.cc
+++ b/passes/opt/opt_expr.cc
@@ -953,6 +953,10 @@ void replace_const_cells(RTLIL::Design *design, RTLIL::Module *module, bool cons
 			}
 
 			if (b.is_fully_const()) {
+				if (b.is_fully_undef()) {
+					RTLIL::SigSpec input = b;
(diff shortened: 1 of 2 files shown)


passes/opt/opt_expr.cc
```diff
@@ -953,6 +953,10 @@ void replace_const_cells(RTLIL::Design *design, RTLIL::Module *module, bool cons
 			}
 
 			if (b.is_fully_const()) {
+				if (b.is_fully_undef()) {
```
whitequark: Should this be `is_fully_undef` and not `is_any_undef`? My reading of the standard is that *any* x or z bits in any operand cause the result to be x.
Screenshot_20190913_114627

cliffordwolf: This code is only for 1 bit expressions, so fully_undef and any_undef is the same.. However, answering the more general point, for example `2'b00 == 2'b1x` is `1'b0`. That's what's meant by "if [...] the relation is ambiguous".

PR #1379, opened 2019-09-15
Added simulation models for Efinix and Anlogic

mmicko:


techlibs/anlogic/cells_sim.v
```diff
@@ -9,6 +9,70 @@ module AL_MAP_SEQ (
 	parameter REGSET = "RESET"; //RESET/SET
 	parameter SRMUX = "SR"; //SR/INV
 	parameter SRMODE = "SYNC"; //SYNC/ASYNC
+
+	wire clk_ce;
+	assign clk_ce = ce ? clk : 1'b0;
```
daveshah1: This doesn't look like a correct implementation (unless the hardware is also like this). If clk is high and ce is toggling then the FF will clock, normally it doesn't work like this.

techlibs/anlogic/cells_sim.v
```diff
@@ -9,6 +9,70 @@ module AL_MAP_SEQ (
...
+						q <= regset;
+					else 
+						q <= d;	
+			end
+		end
+		else
+		begin
+			if (SRMODE == "ASYNC") 
+			begin
+				always @(clk_ce, srmux)
```
daveshah1: I think this should be `, posedge srmux`

techlibs/efinix/cells_sim.v
```diff
@@ -6,6 +6,7 @@ module EFX_LUT4(
    input I3
 );
    parameter LUTMASK  = 16'h0000;
+	assign O = LUTMASK >> {I3, I2, I1, I0};
```
daveshah1: Using a cascade of `?:` like https://github.com/YosysHQ/yosys/blob/master/techlibs/ice40/cells_sim.v#L135-L141 is needed for correct handling of 1'bx when an input is "don't care" due to the init value and state of other values.

techlibs/efinix/cells_sim.v
```diff
@@ -33,6 +42,51 @@ module EFX_FF(
...
+            always @(posedge clk)
+               if (sr)
+                  Q <= SR_VALUE;
+               else if (ce)
+                  Q <= d;
+         end
+         else
+         begin
+            always @(posedge clk)
+               if (ce)
```
daveshah1: Should work fine, but a `begin` and `end` here would disambiguate the dangling else better

techlibs/efinix/cells_sim.v
```diff
@@ -41,6 +95,12 @@ module EFX_GBUFCE(
    output O
 );
    parameter CE_POLARITY = 1'b1;
+
+   wire ce;
+   assign ce = CE_POLARITY ? CE : ~CE;
+   
+   assign O = I & ce;
```
daveshah1: I expect the hardware is really a glitchless gate?

techlibs/anlogic/cells_sim.v
mmicko: Will send you vendor sim model on email, do not have them near me right now

techlibs/efinix/cells_sim.v
mmicko: vendor sim model confirms that

techlibs/anlogic/cells_sim.v
mmicko: Vendor simulation model says it should be level. Please note this is latch mode

techlibs/efinix/cells_sim.v
mmicko: update this in new commit

techlibs/efinix/cells_sim.v
mmicko: Fixed on anlogic and efinix as well

techlibs/anlogic/cells_sim.v
daveshah1: Of course that makes sense, for some reason I thought it was a DDR flipflop mode

mmicko:
Added comment for "future self" to be aware it is for latch :)

PR #1384, opened 2019-09-19
Add techmap_autopurge attribute

### LINKED ISSUE #1381: techmap-ing module with underspecified input port connections creates new undriven signal

```
read_verilog <<EOT
module sub(input i, output o, input j);
foobar f(i, o, j);
endmodule
EOT
design -stash techmap

read_verilog <<EOT
(* blackbox *)
module sub(input i, output o, input j);
endmodule

(* blackbox *)
module foobar(input i, output o, input j);
endmodule

sub s0(i, o);
endmodule
EOT

### techmap -map %techmap
hierarchy -check
check -assert
```
Remove `###` to error out on `check -assert`

## Expected behavior

Identical behaviour as with `techmap` commented out -- no error.

## Actual behavior

```
5. Executing CHECK pass (checking for obvious problems).
checking module top..
Warning: Wire top.\s0.j is used but has no driver.
found and reported 1 problems.
```
where resulting netlist is:
```
module \top
  attribute \src "<<EOT:9"
  wire input 1 \i
  attribute \src "<<EOT:9"
  wire output 2 \o
  attribute \src "<<EOT:10|<<EOT:1"
  wire \s0.i
  attribute \src "<<EOT:10|<<EOT:1"
  wire \s0.j
  attribute \src "<<EOT:10|<<EOT:1"
  wire \s0.o
  attribute \src "<<EOT:10|<<EOT:2"
  cell \foobar \s0.f
    connect \i \i
    connect \j \s0.j
    connect \o \o
  end
end
```

cliffordwolf:
Fixes #1381

### DIFF
--- a/passes/techmap/techmap.cc
+++ b/passes/techmap/techmap.cc
@@ -206,10 +206,27 @@ struct TechmapWorker
 
 		std::map<RTLIL::IdString, RTLIL::IdString> positional_ports;
 		dict<Wire*, IdString> temp_renamed_wires;
+		pool<SigBit> autopurge_tpl_bits;
 
-		for (auto &it : tpl->wires_) {
+		for (auto &it : tpl->wires_)
+		{
 			if (it.second->port_id > 0)
-				positional_ports[stringf("$%d", it.second->port_id)] = it.first;
+			{
+				IdString posportname = stringf("$%d", it.second->port_id);
+				positional_ports[posportname] = it.first;
+
+				if (!flatten_mode && it.second->get_bool_attribute(ID(techmap_autopurge)) &&
+						(!cell->hasPort(it.second->name) || !GetSize(cell->getPort(it.second->name))) &&
+						(!cell->hasPort(posportname) || !GetSize(cell->getPort(posportname))))
+				{
+					if (sigmaps.count(tpl) == 0)
+						sigmaps[tpl].set(tpl);
+
+					for (auto bit : sigmaps.at(tpl)(it.second))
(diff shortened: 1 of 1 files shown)


eddiehung:
@clairexen Would you consider adding this to the `techmap` doc? In solving #3775 I (once again) re-discovered a need for this, only to find to find out this wasn't the first time...!

nakengelhardt:
@eddiehung Which doc? The attribute is described in the [techmap help](https://yosyshq.readthedocs.io/projects/yosys/en/latest/cmd/techmap.html), is techmap also documented elsewhere that we need to update?

eddiehung:
> @eddiehung Which doc? The attribute is described in the [techmap help](https://yosyshq.readthedocs.io/projects/yosys/en/latest/cmd/techmap.html), is techmap also documented elsewhere that we need to update?

I'm so sorry! I completely missed that it was. I didn't see it in this pull request and expected my squishy brain to have spotted it after skimming those exact docs. The only other place I might expect to see if would be https://github.com/YosysHQ/yosys#verilog-attributes-and-non-standard-features but there are likely many other attributes missing from that list. Sorry again for the noise!!

nakengelhardt:
No worries! It is actually in this pull request too but somehow I also only noticed it on the third look...

I don't know if the techmap features should go in the general verilog frontend feature description, IMO that should be a separate section somewhere so that there's no confusion with what goes into user input designs. I'll put it on the list of all the things that should be changed or added when we overhaul the manual.

PR #1389, opened 2019-09-20
Switch parser to GLR and implement SV typedef

daveshah1:
As far as I can tell, any further SystemVerilog support pretty much requires a GLR parser to implement cleanly, due to the arbitrary amount of lookahead needed to disambiguate wires and cells (even more so between packed wires and arrays of cells, where the range might contain an expression of arbitrary complexity).

I can't see any major downsides to this but I may have missed something. The main issue I see is a possible performance reduction parsing cell instances where the parser will briefly "fork". But I'm happy to look at other options if there's a downside that I've missed.


daveshah1:
N.B. there is currently a test failure due to a ambiguity between typedef ports and interface ports, this needs a fix still by treating interface ports as a special case of this.

whitequark:
Very nice work! Are you at all interested in improving the grammar for `always_ff` and friends?

daveshah1:
I need to look into that, but I expect it shouldn't be too hard. With this change, structs and enums shouldn't be too hard to follow on with either.

daveshah1:
I imagine the only parser change needed for `always_ff`, etc is setting an extra attribute depending on the token used, which the appropriate `proc` pass would then take into consideration

hzeller:
Very nice!

FYI I was running a subset of tests (that are known to work with at least one other tool) from our work-in-progress test suite for SystemVerilog compatibility https://github.com/SymbiFlow/sv-tests , and with this pull request a significant amount of more tests started to pass :) 117 -> 181 passing tests.

(and no regression, i.e. things that passed before and not afterwards)
Baseline yosys is f3781f98

```
+-------+--------------+----------+
| Yosys | Yosys_pr1389 | count(*) |
+-------+--------------+----------+
| false | false        |       92 |
| false | true         |       64 |
| true  | true         |      117 |
+-------+--------------+----------+
```

[results.csv.gz](https://github.com/YosysHQ/yosys/files/3637755/results.csv.gz)

frontends/ast/simplify.cc
```diff
@@ -780,6 +782,100 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
 		delete_children();
 	}
 
+	// resolve typedefs
+	if (type == AST_TYPEDEF) {
+		log_assert(children.size() == 1);
+		log_assert(children[0]->type == AST_WIRE || children[0]->type == AST_MEMORY);
+		while(children[0]->simplify(const_fold, at_zero, in_lvalue, stage, width_hint, sign_hint, in_param)) {
+			did_something = true;
+		};
```
cliffordwolf: Style nitpick: No `{..}` for a single statement and certainly no `};`.  ;)

frontends/verilog/verilog_parser.y
```diff
@@ -112,6 +112,10 @@ struct specify_rise_fall {
 
 %define api.prefix {frontend_verilog_yy}
 
+%glr-parser
+%expect 21
+%expect-rr 2
```
cliffordwolf: Are we sure that those 21 shift-reduce conflicts are all justified? Is there anything to gain from using `%expect` if the number is that large? (I'm assuming that it's okay to have conflicts without `%expect` statement in GLR mode.)

udif: How are we sure that these 21 shift-reduce conflicts are harmless? Did you actually check each of those?
Is it possible to silence individual shift-reduce warnings just to be sure we won't be covering new conflicts that might appear in the future?
I think I've tried GLR in the past when I worked on my private branch trying to add typedefs, but gave up on all those shift-reduce. I was able to restructure the parser a bit and reduce the number of conflicts, but not eliminate them entirely. I came to the conclusion that the {...} code sections within the rules is one of the reasons for these conflicts because you cannot look ahead past them, or something like that.
I haven't put time into fully understanding why the earlier approach (doing symbol lookup at the lexer) was ruled out by packages.

frontends/ast/simplify.cc
daveshah1: Oops, will fix once next at a PC

frontends/verilog/verilog_parser.y
daveshah1: I haven't tried to work out the conflicts, I'm not sure if there is a good way to investigate them? AIUI, in the worst case a conflict will produce a run time error rather than a mis-parse so they should easily be caught (as was the case with SV interfaces before). 

The `%expect` was mostly added to suppress a warning that may confuse people building Yosys and make sure we know of any new conflicts added from now on.

imo, doing resolution in the lexer is going to make it very difficult to handle scoping correctly

PR #1393, opened 2019-09-22
write_verilog: do not print (*init*) attributes on regs

whitequark:
If an init value is emitted for a reg, an (\*init\*) attribute is never
necessary, since it is exactly equivalent. On the other hand, some
tools that consume Verilog (ISE, Vivado, Quartus) complain about
(\*init\*) attributes because their interpretation differs from Yosys.

All (\*init\*) attributes that would not become reg init values anyway
are emitted as before.

### DIFF
--- a/backends/verilog/verilog_backend.cc
+++ b/backends/verilog/verilog_backend.cc
@@ -371,13 +371,14 @@ void dump_sigspec(std::ostream &f, const RTLIL::SigSpec &sig)
 	}
 }
 
-void dump_attributes(std::ostream &f, std::string indent, dict<RTLIL::IdString, RTLIL::Const> &attributes, char term = '\n', bool modattr = false, bool as_comment = false)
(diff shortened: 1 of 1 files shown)


whitequark:
Can we backport this to a release branch? Not being able to use nMigen/Yosys with proprietary toolchains for a year is pretty annoying. There is no workaround (using `attrmap` would prevent both forms of initialization from being emitted) other than somehing like patching the Verilog output with a regexp.

whitequark:
@cliffordwolf @eddiehung Are there any objections to the change itself, regardless of whether it goes to a release branch?

eddiehung:
> @cliffordwolf @eddiehung Are there any objections to the change itself, regardless of whether it goes to a release branch?

I haven't looked at this in detail, but I recall that @cliffordwolf expressed some concerns about whether there were any obscure corner cases that could be affected, and that he's still intent on reviewing it properly ... Hope that helps!

whitequark:
Thanks, that's fine with me. I was worried it fell through the cracks.

PR #1396, opened 2019-09-23
synth_xilinx: Support latches, remove used-up FF init values.

### LINKED ISSUE #1387: Please support LDCE cell type for xilinx architecture.

module latchp
    ( input d, clk, en, output reg q );
	always @*
		if ( en )
			q <= d;
endmodule

read_verilog latchp.v
synth_xilinx

## Expected behavior

     LDCE                                1

## Actual behavior

     $_DLATCH_P_                  1

koriakin:
Fixes #1387.

### DIFF
--- a/techlibs/xilinx/xc6s_ff_map.v
+++ b/techlibs/xilinx/xc6s_ff_map.v
@@ -18,7 +18,12 @@
  */
 
 // ============================================================================
-// FF mapping
+// FF mapping for Spartan 6.  The primitives used are the same as Series 7,
+// but with one major difference: the initial value is implied by the
+// primitive type used (FFs with reset pin must have INIT set to 0 or x, FFs
+// with set pin must have INIT set to 1 or x).  For Yosys primitives without
+// set/reset, this means we have to pick the primitive type based on the INIT
+// value.
 
 `ifndef _NO_FFS
 
@@ -29,6 +34,7 @@ module  \$_DFF_N_   (input D, C, output Q);
   else
     FDRE_1 #(.INIT(_TECHMAP_WIREINIT_Q_)) _TECHMAP_REPLACE_ (.D(D), .Q(Q), .C(C), .CE(1'b1), .R(1'b0));
   endgenerate
+  wire _TECHMAP_REMOVEINIT_Q_ = 1;
 endmodule
 module  \$_DFF_P_   (input D, C, output Q);
   parameter [0:0] _TECHMAP_WIREINIT_Q_ = 1'bx;
@@ -37,6 +43,7 @@ module  \$_DFF_P_   (input D, C, output Q);
   else
     FDRE   #(.INIT(_TECHMAP_WIREINIT_Q_)) _TECHMAP_REPLACE_ (.D(D), .Q(Q), .C(C), .CE(1'b1), .R(1'b0));
   endgenerate
+  wire _TECHMAP_REMOVEINIT_Q_ = 1;
 endmodule
 
 module  \$_DFFE_NP_ (input D, C, E, output Q);
@@ -46,6 +53,7 @@ module  \$_DFFE_NP_ (input D, C, E, output Q);
   else
     FDRE_1 #(.INIT(_TECHMAP_WIREINIT_Q_)) _TECHMAP_REPLACE_ (.D(D), .Q(Q), .C(C), .CE(E),    .R(1'b0));
   endgenerate
+  wire _TECHMAP_REMOVEINIT_Q_ = 1;
 endmodule
 module  \$_DFFE_PP_ (input D, C, E, output Q);
   parameter [0:0] _TECHMAP_WIREINIT_Q_ = 1'bx;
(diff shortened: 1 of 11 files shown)


eddiehung:
Sorry for the delay for approving this. I have on my to-do list to investigate how to build tests for latches... I have some an outstanding effort based on `async2sync` which I've yet to return to, to see if it is sufficient to recommend here...

eddiehung:
Following script seems to fail for me

`./yosys -e "unprocessed" ff.ys`
```
read_verilog <<EOT
module top #(parameter W=2) (input clk, input [W:0] din, output [W:0] qout);
reg [W:0] r = 3'b101;
always @(posedge clk) r <= din;
assign qout = r;
endmodule
EOT
#chparam -set W 2 top
synth_xilinx
```
Works when uncommenting the `chparam` command though.

koriakin:
Oops. I think I see the problem — I added `_TECHMAP_REMOVEINIT_` wire handling in the same place where other special techmap wires are handled, which is only evaluated if the current cell is not covered by `techmap_do_cache`, effectively only removing the init values for the first instance of a given cell...

I'll make a separate fix PR + a regression test, sorry about that.

techlibs/xilinx/xc6s_ff_map.v
```diff
@@ -120,6 +131,26 @@ module  \$_DFF_PP1_ (input D, C, R, output Q);
   else
     FDPE   #(.INIT(_TECHMAP_WIREINIT_Q_)) _TECHMAP_REPLACE_ (.D(D), .Q(Q), .C(C), .CE(1'b1), .PRE( R));
   endgenerate
+  wire _TECHMAP_REMOVEINIT_Q_ = 1;
+endmodule
+
+module  \$_DLATCH_N_ (input E, D, output Q);
```
eddiehung: Why does this differ from the xc7_ff_map.v entry herehttps://github.com/YosysHQ/yosys/pull/1396/files#diff-a3e9bf8ae76f9e5867c42731096cf446R89-R97 ?
It appears that you are branching on the value of `_TECHMAP_WIREINIT_Q_`, which you already provide as the `INIT` parameter. Is it because ISE has issues with this parameter such that you need to instantiate the latch with the correct default value?

koriakin: Latches on xc6s need to have reset value equal to the init value, and we need to provide both (the INIT param may be reduntant, but won't hurt)

tests/xilinx/latches.ys
```diff
@@ -0,0 +1,16 @@
...
+flatten
+equiv_opt -assert -run :prove -map +/xilinx/cells_sim.v synth_xilinx # equivalency check
+async2sync
+equiv_opt -assert -run prove: -map +/xilinx/cells_sim.v synth_xilinx # equivalency check
+design -load postopt # load the post-opt design (otherwise equiv_opt loads the pre-opt design)
+
+design -load preopt
+synth_xilinx
+cd top
+select -assert-count 1 t:LUT1
```
koriakin: This part is worrying — instead of getting an inverter on the G pin, we should just get LDCE_1 used here.  I'll investigate it...

tests/xilinx/latches.ys
```diff
@@ -0,0 +1,16 @@
+read_verilog latches.v
+
+proc
+flatten
+equiv_opt -assert -run :prove -map +/xilinx/cells_sim.v synth_xilinx # equivalency check
+async2sync
+equiv_opt -assert -run prove: -map +/xilinx/cells_sim.v synth_xilinx # equivalency check
+design -load postopt # load the post-opt design (otherwise equiv_opt loads the pre-opt design)
+
+design -load preopt
```
koriakin: Why two `design -load` in a row? The first one is getting wholly overwritten with the second, no?

eddiehung: Good spot! Removed.

techlibs/xilinx/xc6s_ff_map.v
eddiehung: Perhaps leaving a breadcrumb here in the form of a comment would be useful here...

koriakin: OK, though I'll add the comment at the top of the module (since it applies to FFs as well).

koriakin: And done

koriakin:
@eddiehung Your changes LGTM, with the exception of the LDCE_1 issue — but that's a problem with the opt_expr pass (or its usage), which we may want to handle in another PR.  Should I merge this PR now?

koriakin:
One more thing: since we don't actually have a blackbox for LDCE_1, I've changed the mapping rule to use LDCE with IS_G_INVERTED.  I could revert this change and make simulation models for LD[CP]E_1 instead, but I'm not sure it's worth it to use _1 variants.  What do you think?

eddiehung:
Merging this sounds good!

PR #1352, opened 2019-09-03
[WIP] Add new tests for ECP5 architecture

SergeyDegtyar:
Problems:
- dpram.ys. SAT failed with error: Failed to import cell gate.mem.0.0.0 (type DP16KD) to SAT database.
- fsm.ys. remove -assert option for equiv_opt because of unproven cells.

PR #1401, opened 2019-09-24
ICE40 tests. adffs test update (equiv_opt -multiclock).

SergeyDegtyar:
The same changes as in https://github.com/YosysHQ/yosys/pull/1352/commits/c597c2f2ae10111b2dc28e9738ea336d8ad22b31, but for ICE40 architecture.

### DIFF
--- a/tests/ice40/adffs.v
+++ b/tests/ice40/adffs.v
@@ -22,30 +22,26 @@ module adffn
             q <= d;
 endmodule
 
-module dffsr
+module dffs
     ( input d, clk, pre, clr, output reg q );
     initial begin
       q = 0;
     end
-	always @( posedge clk, posedge pre, posedge clr )
-		if ( clr )
-			q <= 1'b0;
-		else if ( pre )
+	always @( posedge clk )
+		if ( pre )
 			q <= 1'b1;
 		else
             q <= d;
 endmodule
 
-module ndffnsnr
+module ndffnr
     ( input d, clk, pre, clr, output reg q );
     initial begin
       q = 0;
     end
-	always @( negedge clk, negedge pre, negedge clr )
+	always @( negedge clk )
 		if ( !clr )
 			q <= 1'b0;
-		else if ( !pre )
-			q <= 1'b1;
 		else
             q <= d;
 endmodule
@@ -58,7 +54,7 @@ input a,
 output b,b1,b2,b3
 );
 
-dffsr u_dffsr (
+dffs u_dffs (
         .clk (clk ),
         .clr (clr),
         .pre (pre),
@@ -66,7 +62,7 @@ dffsr u_dffsr (
         .q (b )
     );
 
-ndffnsnr u_ndffnsnr (
+ndffnr u_ndffnr (
         .clk (clk ),
         .clr (clr),
         .pre (pre),
--- a/tests/ice40/adffs.v
+++ b/tests/ice40/adffs.v
@@ -27,7 +27,7 @@ module dffs
     initial begin
       q = 0;
     end
-	always @( posedge clk )
+	always @( posedge clk, posedge pre )
 		if ( pre )
 			q <= 1'b1;
 		else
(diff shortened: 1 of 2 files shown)


tests/ice40/adffs.v
```diff
@@ -22,30 +22,26 @@ module adffn
             q <= d;
 endmodule
 
-module dffsr
+module dffs
     ( input d, clk, pre, clr, output reg q );
     initial begin
       q = 0;
     end
-	always @( posedge clk, posedge pre, posedge clr )
```
eddiehung: Removing `posedge pre` means this is no longer async?

tests/ice40/adffs.v
```diff
@@ -22,30 +22,26 @@ module adffn
...
             q <= d;
 endmodule
 
-module ndffnsnr
+module ndffnr
     ( input d, clk, pre, clr, output reg q );
     initial begin
       q = 0;
     end
-	always @( negedge clk, negedge pre, negedge clr )
```
eddiehung: Removing `negedge clr` means this is no longer async?

tests/ice40/adffs.ys
```diff
@@ -1,12 +1,11 @@
...
-equiv_opt -assert -map +/ice40/cells_sim.v synth_ice40 # equivalency check
+equiv_opt -multiclock -assert -map +/ice40/cells_sim.v synth_ice40 # equivalency check
 design -load postopt # load the post-opt design (otherwise equiv_opt loads the pre-opt design)
 cd top # Constrain all select calls below inside the top module
-select -assert-count 1 t:SB_DFF
-select -assert-count 1 t:SB_DFFN
-select -assert-count 2 t:SB_DFFSR
-select -assert-count 7 t:SB_LUT4
-select -assert-none t:SB_DFF t:SB_DFFN t:SB_DFFSR t:SB_LUT4 %% t:* %D
+select -assert-count 1 t:SB_DFFNSR
```
eddiehung: We should take a closer look at this: two flops described in the HDL results in 4 flop primitives, and a LUT?

tests/ice40/adffs.v
SergeyDegtyar: Yes. Now, this control is synchronous.

tests/ice40/adffs.ys
SergeyDegtyar: I think the result is right. In the test, we have four modules with FF descriptions and four FFs as a result. LUTs used for inverted controls. I think Lattice doesn't have FFs with inverted controls?

tests/ice40/adffs.v
SergeyDegtyar: The same.

tests/ice40/adffs.v
eddiehung: Isn't this file meant to test asynchronous flops, which seem to be supported for ice40? This is the whole reason for using the `-multiclock`...

daveshah1: iCE40 doesn't have flops with both async preload and clear, only one or the other; and there is no alternative implementation that is likely to work reliably

tests/ice40/adffs.ys
daveshah1: You're right, positive controls only, so the LUT is needed

tests/ice40/adffs.v
daveshah1: What you should probably do is change this to one asynchronous control signal rather than one synchronous control signal

tests/ice40/adffs.ys
eddiehung: Sorry. I miscounted and thought there were only 2 flops here! As you were.

tests/ice40/adffs.v
SergeyDegtyar: Ok. Will fix it.

SergeyDegtyar: Done.

PR #1406, opened 2019-09-27
rpc: new frontend

whitequark:
A new pass, `connect_rpc`, allows any HDL frontend that can read/write JSON from/to stdin/stdout or an unix socket or a named pipe to participate in elaboration as a first class citizen, such that any other HDL supported by Yosys directly or indirectly can transparently instantiate modules handled by this frontend.

Recognizing that many HDL frontends emit Verilog, it allows the RPC frontend to direct Yosys to process the result of instantiation via any built-in Yosys frontend. The resulting RTLIL is then hygienically integrated into the overall design.

<hr>

To see this in action with a real frontend, check out the `rpc` branch of [nMigen](https://github.com/m-labs/nmigen/tree/rpc), run `python3 setup.py develop --user`, then use `synthesize.sh` or `simulate.sh` scripts from https://gist.github.com/whitequark/5283570958f139fac285d047f6788fe1.

<hr>

This PR has been tested on Linux and Windows, and the mock frontend includes Windows support, but since the Yosys testsuite requires bash, the tests currently only run on Unix.

I found the amount and quality of ad-hoc JSON code in Yosys highly aggravating, so I took the liberty of importing json11 into the tree. It is very compact, reasonably fast, and uses a compatible license.


frontends/rpc/rpc_frontend.cc
```diff
@@ -0,0 +1,589 @@
...
+ *  MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ *  ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ *  WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ *  ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ *  OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *
+ */
+
+// The reason the -path mode of connect_rpc uses byte-oriented and not message-oriented sockets, even though
+// it is a message-orietned interface, is that the system can place various limits on the message size, which
```
daveshah1: nitpick: oriented

frontends/rpc/rpc_frontend.cc
```diff
@@ -0,0 +1,589 @@
...
+			}
+		} else is_valid = false;
+		if (!is_valid)
+			log_cmd_error("RPC frontend returned malformed response: %s\n", response.dump().c_str());
+		return modules;
+	}
+
+	std::pair<std::string, std::string> derive_module(const std::string &module, const dict<RTLIL::IdString, RTLIL::Const> &parameters) {
+		Json::object json_parameters;
+		for (auto &param : parameters) {
```
daveshah1: I'm wondering if we should try and unify this with how we write parameters in JSON netlists:

> Numeric 32-bit parameter and attribute values are written as decimal values.
> Bit verctors of different sizes, or ones containing 'x' or 'z' bits, are written
> as string holding the binary representation of the value. Strings are written
> as strings, with an appended blank in cases of strings of the form /[01xz]* */.

Even though this isn't the nicest approach (due to backwards compatibility), consistency might be helpful?

whitequark: Do you mean change the JSON backend protocol, or change the RPC frontend protocol? I was specifically trying to avoid the ad-hoc things the JSON backend did, even at cost of losing some readability. Also, having only one encoding for numbers is a nice simplification for the RPC frontend side parser.

passes/hierarchy/hierarchy.cc
```diff
@@ -833,7 +833,7 @@ struct HierarchyPass : public Pass {
 		}
 
 		if (top_mod == nullptr && auto_top_mode) {
-			log_header(design, "Finding top of design hierarchy..\n");
+			log_header(design, "Finding top of design hierarchy...\n");
```
cliffordwolf: There's a pretty consistent use of `..` instead of `...` at the end of log messages like this in Yosys. See `git grep -E '[^.]\.\.\.(\\n)?"'` vs `git grep -E '[^.]\.\.(\\n)?"'`.

We can have a discussion if it should be `...`, but I'm against just changing it here and nowhere else.

whitequark: I didn't realize it wasn't a typo. Removed that commit.

PR #1409, opened 2019-09-27
Change order of parameters, to work on other OS

mmicko:
Due to a difference how getopt works on GNU library and other implementation, changed order of parameters, since in strict systems non-option arguments must be at the end of argument list.


mmicko:
This makes all tests pass on macOS and also works with getopt implementation for Windows build

PR #1412, opened 2019-09-27
equiv_opt to call async2sync when not -multiclock like SymbiYosys

eddiehung:
Just like sby does here:
https://github.com/YosysHQ/SymbiYosys/blob/06cc339957a56bf413da911d6c0aa01503590802/sbysrc/sby_core.py#L346-L349
and cleans up small barnacles like this:
https://github.com/YosysHQ/yosys/blob/8c2dd518b908644c9181526e193b8211734e8cef/tests/xilinx/latches.ys#L5-L7

### DIFF
--- a/passes/equiv/equiv_opt.cc
+++ b/passes/equiv/equiv_opt.cc
@@ -156,6 +156,8 @@ struct EquivOptPass:public ScriptPass
 		if (check_label("prove")) {
 			if (multiclock || help_mode)
 				run("clk2fflogic", "(only with -multiclock)");
+			else
+				run("async2sync", "(only without -multiclock)");
 			run("equiv_make gold gate equiv");
 			if (help_mode)
 				run("equiv_induct [-undef] equiv");


whitequark:
I'm afraid I don't understand how `async2sync` works.

eddiehung:
> I'm afraid I don't understand how `async2sync` works.

I can't say I had a solid grasp until I'm asked to try explain it, so here goes...
Take this script which infers a latch:
```
read_verilog <<EOT
module top(input D, E, output reg Q);
initial Q = 1'b0;
always @*
    if (E) Q = D;
endmodule
EOT

proc
opt
show -format png -prefix before
async2sync
show -format png -prefix after
```

Since induction "operates" using a clock, it is unable to verify circuits with latches inside, such as the one above. What I believe `async2sync` does is that it transforms such asynchronous elements (async flops and latches) into "synchronous" ones that are clocked by an imaginary super clock that operates infinitely fast (global clock in SV speak?). Thus, equivalence checking can proceed.

The result of the transformation is from this: 
before to 
after

I don't believe it should have no effect on regular user clocks... which makes it plausible to use by default when `-multiclock` is not specified.

whitequark:
Thank you for the explanation, makes a lot of sense.

daveshah1:
One comment on this PR.

Previously, if a pass accidentally transformed an async reset DFF into a sync one or vice versa, then `equiv_opt` without multiclock would rightfully fail. After this PR I believe it will pass despite equivalence no longer true. This is fine so long as `-multiclock` is considered part of the contract for designs containing async devices.

whitequark:
I think we should @daveshah1's observation that to the help text for this pass.

cliffordwolf:
> The result of the transformation is from this:
> before
> to
> after
> -- which includes an `$ff` cell understood to be clocked by this global clock.

Huh. That doesn't look right. The `$ff` input should be the module `Q` port, not `D`. Weird..

cliffordwolf:
> Huh. That doesn't look right. The `$ff` input should be the module `Q` port, not `D`. Weird..

#1417 should fix this. I have not tested this change. So let's see if it breaks something or improves the overall situation.  :D

eddiehung:
For completeness, post #1417 the "after" fig looks like this now:
after

eddiehung:
@daveshah1 I've been trying to reproduce your comment:

> Previously, if a pass accidentally transformed an async reset DFF into a sync one or vice versa, then `equiv_opt` without multiclock would rightfully fail. After this PR I believe it will pass despite equivalence no longer true. This is fine so long as `-multiclock` is considered part of the contract for designs containing async devices.

```
read_verilog -formal <<EOT
module gate(input C, D, R, output reg Q);
initial Q = 1'b0;
always @(posedge C or posedge R)
    if (R) Q <= 1'b0; else Q <= D;
endmodule

module gold(input C, D, R, output reg Q);
initial Q = 1'b0;
always @(posedge C /*or posedge R*/)
    if (R) Q <= 1'b0; else Q <= D;
endmodule
EOT

proc
async2sync
#clk2fflogic
equiv_make gold gate equiv
equiv_induct equiv
equiv_status -assert equiv
```

For this particular circuit, it fails in both the `async2sync` and `clk2fflogic` case. Were you not expecting this to be the case, or can my testcase be improved?

daveshah1:
This was a misunderstanding on my part as to how `async2sync` works (I think this was might have been a fairly recent change to give more useful behaviour in common cases)?

However, consider the following which does pass with `async2sync` despite a lack of equivalence (consider R being asserted and deasserted without a clock edge) - it correctly fails with `clk2fflogic`:
```
read_verilog -formal <<EOT
module gate(input C, D, R, output reg Q);
initial Q = 1'b0;
always @(posedge C or posedge R)
    if (R) Q <= 1'b0; else Q <= D;
endmodule

module gold(input C, D, R, output reg Q);
initial Qp = 1'b0;
always @(posedge C /*or posedge R*/)
    if (R) Qp <= 1'b0; else Qp <= D;
assign Q = Qp & !R;
endmodule
EOT

proc
async2sync
#clk2fflogic
equiv_make gold gate equiv
equiv_induct equiv
equiv_status -assert equiv
```

cliffordwolf:
Just an idea: Maybe it would make sense to add a `-singleclock` option that is exclusive with `-multiclock` and only run `async2sync` when `-singleclock` is present?

eddiehung:
Thanks @daveshah1! @cliffordwolf my thinking was that we only want to run `async2sync` for latches, since they wouldn't pass induction regardless. Can anyone see a problem with that?

I'm not particularly keen on `-singleclock` as that should be default in the absence of `-multiclock`. 

P.S. @cliffordwolf Does this above findings affect SymbiYosys?

Issue #1407, opened 2019-09-27
in MXE build abc9 crashes on read <abc-temp-dir>/input.xaig 


Chopper455:
## Steps to reproduce the issue
Compiled 143f82d with 32-bit and 64-bit MXE, run in Win 7.
Used simple 4-bit counter:

[counter4.zip](https://github.com/YosysHQ/yosys/files/3661661/counter4.zip)

As a target used xc7 with commands:
>read_verilog counter4.v
synth_xilinx -family xc7 -abc9 -top counter4 -edif proj_xilinx.edf

## Expected behavior

In xubuntu 18.04 with 64-bit build block compiles successfully for the same 143f82d tag.
>2.39.1.4. Executing ABC9.
Running ABC command: <yosys-exe-dir>/yosys-abc -s -f <abc-temp-dir>/abc.script 2>&1
ABC: ABC command line: "source <abc-temp-dir>/abc.script".
ABC: 
ABC: + read_lut <yosys-exe-dir>/share/xilinx/abc_xc7.lut 
ABC: + read_box -v <yosys-exe-dir>/share/xilinx/abc_xc7.box 
ABC: + &read <abc-temp-dir>/input.xaig 
ABC: + &ps 
ABC: <abc-temp-dir>/input : i/o =      6/     10  and =       0  lev =    0 (0.00)  mem = 0.00 MB  box = 1  bb = 0
ABC: + &st 
ABC: + &scorr 
ABC: Warning: The network is combinational.
ABC: + &sweep 
ABC: + &dc2 
ABC: + &st 
ABC: + &dch -f 
ABC: + &ps 
ABC: <abc-temp-dir>/input : i/o =      6/     10  and =       0  lev =    0 (0.00)  mem = 0.00 MB  ch =    0  box = 1  bb = 0
ABC: + &if -W 300 -v 
ABC: K = 8. Memory (bytes): Truth =    0. Cut =   64. Obj =  144. Set =  672. CutMin = no
ABC: Node =       0.  Ch =     0.  Total mem =    0.01 MB. Peak cut mem =    0.00 MB.
ABC: P:  Del =  921.00.  Ar =       0.0.  Edge =        0.  Cut =        0.  T =     0.00 sec
ABC: P:  Del =  921.00.  Ar =       0.0.  Edge =        0.  Cut =        0.  T =     0.00 sec
ABC: P:  Del =  921.00.  Ar =       0.0.  Edge =        0.  Cut =        0.  T =     0.00 sec
[...]

PR #1416, opened 2019-09-29
Open aig frontend as binary file

mmicko:
Reading of generated aig files fails when they are in binary format, made change to support frontends with binary input, and use binary stream to open xaig output in ABC9 after generate.
This fixes hang on read in issue #1407

### DIFF
--- a/frontends/aiger/aigerparse.cc
+++ b/frontends/aiger/aigerparse.cc
@@ -1056,7 +1056,7 @@ struct AigerFrontend : public Frontend {
 			}
 			break;
 		}
-		extra_args(f, filename, args, argidx);
+		extra_args(f, filename, args, argidx, true);
 
 		if (module_name.empty()) {
 #ifdef _WIN32
--- a/frontends/aiger/aigerparse.cc
+++ b/frontends/aiger/aigerparse.cc
@@ -285,6 +285,8 @@ end_of_header:
 		}
 		else if (c == 'c') {
 			f.ignore(1);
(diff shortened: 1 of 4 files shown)


mmicko:
This should be it, looks like there was error on "read_aiger" for some time at least on windows, so this solves that problem as well. Tested ascii and binary aig files, and both are now readable on windows. It also works if windows generated files are read in linux and other way arround.

frontends/aiger/aigerparse.cc
```diff
@@ -285,6 +285,8 @@ void AigerReader::parse_aiger()
 		}
 		else if (c == 'c') {
 			f.ignore(1);
+			if (f.peek() == '\r')
```
eddiehung: Do you still need this if you're opening in `std::ifstream::binary`? If so, this would mean ABC is emitting '\r\n' right?

mmicko:
Well our own write_aiger -ascii file.aig would open as text file so this
enables reading such files.

eddiehung:
> Well our own write_aiger -ascii file.aig would open as text file so this enables reading such files.

Good thinking. I forgot about the ASCII mode. The standard isn't specific about whether to use Windows/Linux line endings, and indeed, the reference library does not write in binary mode either:

https://github.com/zimmski/aiger/blob/ba826db1798022fdabe2de71ea173bd2b6dcac9d/aiger.c#L1853

It does read in binary mode, and does ignore '\r' as part of ignoring whitespace, though.

PR #1420, opened 2019-09-30
Add tests for Xilinx architecture (contd)

eddiehung:
Continued from #1363, resolving all outstanding issues.


tests/xilinx/shifter.v
```diff
@@ -0,0 +1,22 @@
...
+clk,
+in
+);
+    output [7:0] out;
+    input signed clk, in;
+    reg signed [7:0] out = 0;
+
+    always @(posedge clk)
+	begin
+`ifndef BUG
```
eddiehung: BUG is not used

mmicko: fixed here and in tribuf.v as well

Issue #1358, opened 2019-09-06
Missing ECP5 simulation models


daveshah1:
Simulation models need to be created for the following cells that are otherwise supported in the flow:
 - [ ] DP16KD (block RAM)
 - [ ] MULT18X18D (DSP multiplier)
 - [ ] CLKDIVF (clock divider)
 - [ ] DCCA (clock gate)
 - [ ] OSCG (oscillator)
 - [ ] DDR I/O primitives

PR #1421, opened 2019-09-30
Add tests for ECP5 architecture (contd)

eddiehung:
Continued from #1363, resolving all but one issue regarding latches:

https://github.com/YosysHQ/yosys/pull/1352#discussion_r329809414


tests/ecp5/macc.ys
```diff
@@ -0,0 +1,15 @@
+read_verilog macc.v
+proc
+hierarchy -top top
+#Failed because of 14 unproven cells.
+equiv_opt -assert -map +/ecp5/cells_sim.v synth_ecp5 # equivalency check
+#equiv_opt -map +/ecp5/cells_sim.v synth_ecp5 # equivalency check
```
mmicko: Seams that now MULT18X18D is properly generated, but we do not have simulation models yet.
Adding something like "#Blocked by issue #1358 (Missing ECP5 simulation models)" will be fine for now I guess

PR #1423, opened 2019-10-01
RFC: techmap to recognise wires named "_TECHMAP_REPLACE_.<suffix>"

eddiehung:
... and to substitute the original cell name and alias them to the auto-generated wire.

Useful for creating wires that have a predictable name (not `$techmapNNN...`) that can be recovered by other passes.

Could be used to create `(* hierconn *)` signals, for example, too.

### DIFF
--- a/passes/techmap/techmap.cc
+++ b/passes/techmap/techmap.cc
@@ -257,6 +257,12 @@ struct TechmapWorker
 					w->add_strpool_attribute(ID(src), extra_src_attrs);
 			}
 			design->select(module, w);
+
+			if (it.second->name.begins_with("\\_TECHMAP_REPLACE_.")) {
+				IdString replace_name = stringf("%s%s", orig_cell_name.c_str(), it.second->name.c_str() + strlen("\\_TECHMAP_REPLACE_"));
+				Wire *replace_w = module->addWire(replace_name, it.second);
+				module->connect(replace_w, w);
+			}
 		}
 
 		SigMap tpl_sigmap(tpl);
@@ -1198,6 +1204,10 @@ struct TechmapPass : public Pass {
 		log("\n");
 		log("A cell with the name _TECHMAP_REPLACE_ in the map file will inherit the name\n");
 		log("and attributes of the cell that is being replaced.\n");
+		log("A wire with a name of the form `_TECHMAP_REPLACE_.<suffix>` in the map file will\n");
+		log("cause a new wire alias to be created with its name set to the original but with\n");
+		log("its `_TECHMAP_REPLACE_' prefix to be substituted with the name of the cell being\n");
+		log("replaced.\n");
 		log("\n");
 		log("See 'help extract' for a pass that does the opposite thing.\n");
 		log("\n");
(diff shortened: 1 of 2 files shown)


cliffordwolf:
hmm.. quickly summarizing the current behavior, using the following top-level (from your test case):

```
sub s0(i, o);
endmodule
```

The following will create a foobar instance called `s0`. Everything else in that module is mapped using the prefix `$techmap%d`. The reason for not using the prefix `s0` is so that when the new `s0` is techmapped to something else, and that something else would create `s0.asdf` that would conflict with `s0.asdf` frtom `sub`:

```
module sub(input i, output o, input j);
foobar _TECHMAP_REPLACE_(i, o, j);
wire asdf = i ;
endmodule
EOT
```

And the following would create `s0.xyz` and `s0.asdf`, as no `_TECHMAP_REPLACE_` cell is present, and therefore no `s0` cell exists after techmap, so no risk of name collisions down the road when a `s0` is mapped.

```
module sub(input i, output o, input j);
foobar xyz (i, o, j);
wire asdf = i ;
endmodule
EOT
```

What is the use-case for this PR where that latter version without `_TECHMAP_REPLACE_` is insufficient? I think it makes sense to avoid creating entities with names like `s0` and `s0.foo` that would potentially result in name collisions when `s0` is techmapped to something else.

eddiehung:
> What is the use-case for this PR where that latter version without _TECHMAP_REPLACE_ is insufficient? I think it makes sense to avoid creating entities with names like s0 and s0.foo that would potentially result in name collisions when s0 is techmapped to something else.

The use case is that I need to use `_TECHMAP_REPLACE_`. Specifically, I have a techmap module that wraps (using `-max_iter 1` to prevent infinite recursion) an existing cell with existing cell + new cell, while preserving everything about the existing cell:
https://github.com/YosysHQ/yosys/blob/1b96d29174d7c56a14031bc117a7da5fa5192c81/techlibs/xilinx/abc_map.v#L38-L54
In this manner, I really want to have a predictable name for (specific) wires inside, as then I can do a lookup in a downstream pass to get that generated wire, based only on information from the cell. With the `stringf($techmap%d, autoidx++)` prefix, I don't believe I can. 

An alternative would be to preserve the value of autoidx using a mechanism like:
https://github.com/YosysHQ/yosys/blob/d963e8c2c6207ad98d48dc528922ad58c030173f/passes/cmds/scc.cc#L248-L251

cliffordwolf:
Okay. SGTM.

Maybe it would make sense to add a similar functionality for cells named `_TECHMAP_REPLACE_.*` here:

https://github.com/YosysHQ/yosys/blob/369652d4b99181e2f7b875b6c458c1a5a3b0381e/passes/techmap/techmap.cc#L380-L388

eddiehung:
The alternative is that we *only* use the `$techmap%d` prefix /iff/ there is going to be a name collision (i.e. do so lazily). Are there any problems with that? I think we should at least try and behave the same whether or not there is a `_TECHMAP_REPLACE_` present...

cliffordwolf:
We have recently added functionality to techmap to deal with name collisions for wires. We would just need to do the same for cells, and then we can get rid of the `$techmap%d` prefix mechanism.

PR #1429, opened 2019-10-02
Add "check -mapped"

cliffordwolf:


daveshah1:
For iCE40 and ECP5 flows the contract with the PnR tool allows `$_TBUF_` cells in the mapped netlist. Although longer term we might want to look for a better solution, how about an extra `-allow-tbuf` option?

cliffordwolf:
> how about an extra -allow-tbuf option?

Good point. Added in 3e27b28.

PR #1430, opened 2019-10-02
Default to "synth_xilinx -nodsp" for all families but "xc7"

eddiehung:


daveshah1:
Vivado upgrades primitives automatically, confirmed working even going through EDIF, so imo there's no real harm in leaving DSP inference turned on for UltraScale (I'll look at doing it properly some time but it's a lot better than nothing)

koriakin:
You can merge it if you want, but I intend to do a better patch this week (with proper support for techmapping to DSP48A1/DSP48E2, but no pmgen yet)

PR #1432, opened 2019-10-03
Refactor peepopt_dffmux and be sensitive to \init when trimming

### LINKED ISSUE #1427: peepopt performs unsound optimization

Tested on Yosys 0.9+899 (git sha1 da347b9f).

Testcase:
<details>

```
attribute \top 1
module \top
  wire input 5 \clk_if_0__io
  attribute \init 2'00
  wire width 2 \fsm.addr
  wire width 2 \fsm.addr$next
  wire \fsm.next_state
  attribute \init 1'0
  wire \fsm.state
  wire output 1 \led_0__io
  wire output 2 \led_1__io
  wire output 3 \led_2__io
  wire output 4 \led_3__io
  cell $sshl $2
    parameter \A_SIGNED 0
    parameter \A_WIDTH 1
    parameter \B_SIGNED 0
    parameter \B_WIDTH 2
    parameter \Y_WIDTH 4
    connect \A 1'1
    connect \B \fsm.addr
    connect \Y { \led_3__io \led_2__io \led_1__io \led_0__io }
  end
  cell $dff $techmap\fsm.$procdff$39
    parameter \CLK_POLARITY 1'1
    parameter \WIDTH 2
    connect \CLK \clk_if_0__io
    connect \D \fsm.addr$next
    connect \Q \fsm.addr
  end
  cell $dff $techmap\fsm.$procdff$40
    parameter \CLK_POLARITY 1'1
    parameter \WIDTH 1
    connect \CLK \clk_if_0__io
    connect \D \fsm.next_state
    connect \Q \fsm.state
  end
  cell $mux $techmap\fsm.$procmux$25
    parameter \WIDTH 2
    connect \A 2'11
    connect \B \fsm.addr
    connect \S \fsm.state
    connect \Y \fsm.addr$next
  end
  attribute \full_case 1
  cell $mux $techmap\fsm.$procmux$37
    parameter \WIDTH 1
    connect \A 1'0
    connect \B 1'1
    connect \S \fsm.state
    connect \Y \fsm.next_state
  end
end
```
</details>

The testcase has a design that looks like this:
Screenshot_20191002_044044

Namely, it assigns `\fsm.addr` first to 0b00, and then to 0b11, then encodes it as one-hot and displays on LEDs. Running `synth_ice40` on it produces rather interesting results: it lights up `led1`, which is clearly absurd:

Screenshot_20191002_044203

If I comment out the `peepopt` call in `synth_ice40`, it produces the expected optimized design:

Screenshot_20191002_044312

eddiehung:
Fixes #1427. 
`equiv_opt synth_ice40` now passes for that testcase.

Issue was that `peepopt_dffmux` was not checking the `\init` value before trimming.

EDIT: Depends on #1431 -- please let that get merged first.

### DIFF
--- a/tests/various/peepopt.ys
+++ b/tests/various/peepopt.ys
@@ -188,6 +188,13 @@ endmodule
 EOT
 
 proc
+# NB: equiv_opt uses equiv_induct which covers
+#     only the induction half of temporal induction
+#     --- missing the base-case half
+#     This makes it akin to `sat -tempinduct-inductonly`
+#     instead of `sat -tempinduct-baseonly` or
+#     `sat -tempinduct` which is necessary for this
+#     testcase
 #equiv_opt -assert peepopt
 
 design -save gold
@@ -197,7 +204,7 @@ design -stash gate
 design -import gold -as gold
 design -import gate -as gate
 miter -equiv -flatten -make_assert -make_outputs gold gate miter
-sat -seq 1 -verify -prove-asserts -show-ports miter
+sat -tempinduct -verify -prove-asserts -show-ports miter
 
 design -load gate
 select -assert-count 1 t:$dff r:WIDTH=4 %i
--- a/tests/various/peepopt.ys
+++ b/tests/various/peepopt.ys
@@ -188,19 +188,9 @@ endmodule
 EOT
 
 proc
-#equiv_opt -assert peepopt
-
-design -save gold
-peepopt
-design -stash gate
-design -import gold -as gold
-design -import gate -as gate
-miter -equiv -flatten -make_assert -make_outputs gold gate miter
-sat -seq 1 -verify -prove-asserts -show-ports miter
-
+equiv_opt -assert peepopt
 design -load postopt
-wreduce
-select -assert-count 1 t:$dff r:WIDTH=2 %i
+select -assert-count 1 t:$dff r:WIDTH=4 %i
 select -assert-count 2 t:$mux
-select -assert-count 2 t:$mux r:WIDTH=2 %i
+select -assert-count 2 t:$mux r:WIDTH=4 %i
 select -assert-count 0 t:$logic_not t:$dff t:$mux %% t:* %D
--- a/tests/various/peepopt.ys
+++ b/tests/various/peepopt.ys
@@ -131,8 +131,8 @@ EOT
 proc
 equiv_opt -assert peepopt
 design -load postopt
-select -assert-count 1 t:$dff r:WIDTH=5 %i
-select -assert-count 1 t:$mux r:WIDTH=5 %i
+select -assert-count 1 t:$dff r:WIDTH=4 %i
+select -assert-count 1 t:$mux r:WIDTH=4 %i
 select -assert-count 0 t:$dff t:$mux %% t:* %D
 
 ####################
--- a/tests/various/peepopt.ys
+++ b/tests/various/peepopt.ys
@@ -188,8 +188,18 @@ endmodule
 EOT
 
 proc
-equiv_opt -assert peepopt
-design -load postopt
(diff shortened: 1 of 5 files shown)


tests/various/peepopt.ys
```diff
@@ -173,3 +173,34 @@ select -assert-count 1 t:$dff r:WIDTH=2 %i
...
+    initial o <= 4'b0010;
+    always @(posedge clk) begin
+        if (ce) o <= i;
+        if (!rstn) o <= 4'b1111;
+    end
+endmodule
+EOT
+
+proc
+#equiv_opt -assert peepopt
```
whitequark: Is there some deficiency in `equiv_opt` that requires escaping from it here? Or is it just for readability?

In either case, as the author of `equiv_opt`, IMO it should either default to `miter`+`sat` or at least provide it as an option. The reason it does not use those is because I did not understand how to use them at the time. I am still not very certain of this, but I think `miter`+`sat` are strictly more useful than `equiv_induct`. (Correct me if I'm wrong.)

eddiehung: Sorry, let me check again. I couldn't get it to fail in equiv_opt and chalked it up to init funniness...

whitequark: That's one more reason to change `equiv_opt` to use `sat`+`miter` then.

eddiehung: Okay, I've just force pushed some commits that are easier to work with.

The fix is d99810ad8a0d56a13cca75c4b129f932491d9c80.
Regardless of this commit, `equiv_opt` passes.

Without this commit `miter`+`sat` fails, but *only* with #1427 (i.e. it fails if c6a55d948aebc8993cc3e33372b9c403b0b90554 is reverted).

Fun fun fun!

To be honest, I don't think I fully understand all the pros and cons of each method either, except that `miter`+`sat` is bounded and temporal induction is not, however the latter suffers when (from my own empirical experience) initial conditions or clock enables are involved. From a developer point of view, it would be great to get an absolutely bomb-proof equivalence check for any (sequential or combinatorial) transformation I make...

whitequark: Yeah, agreed on the latter. @cliffordwolf might have a better understanding of this?

daveshah1: `sat` has `-tempinduct` for induction, too. I think that `equiv_induct` also adds extra assertions that registers driving the same wire name are equivalent which significantly helps induction, as induction isn't really a useful form of blackbox verification. I don't think `miter` can do this but there might be some other trickery involving the undef modelling in `sat` (although my experience is that comes at a high performance cost)

daveshah1: In fact, `sat -tempinduct` looks stronger than `equiv_induct` as it checks the base case too whereas `equiv_induct` appears just to do the induction part.

eddiehung: Conversations with @cliffordwolf seem to indicate that `equiv_induct` can/should be augmented to cover the base case (requiring about a day's work) but would that actually help here?

eddiehung: It seems like it would. `sat -tempinduct-inductonly` (mirroring `equiv_induct`) passes, as expected, but `sat -tempinduct-baseonly` or `sat -tempinduct` fails which is a good sign. The expectation is that `equiv_induct` that also covers the base case would be a desirable improvement.

daveshah1: Adding the base case to `equiv_induct` would be helpful, I think. I wonder if counterexamples would be possible too?

eddiehung:
Dependency on #1431 only relevant for failing test case. Since #1431 still needs a review, and this PR fixes a case where we're creating incorrect logic, I'm going to go ahead and merge this.

PR #1433, opened 2019-10-03
async2sync to be called by equiv_opt only when -async2sync given

eddiehung:
Make #1412 more conservative by only calling `async2sync` when new option `equiv_opt -async2sync` is given, since @daveshah1 kindly provided a counter example: https://github.com/YosysHQ/yosys/pull/1412#issuecomment-536680733

This new `-async2sync` option is mutex to `-multiclock`

Also disables equiv checking in ice40/latches.ys since it was broken.

### DIFF
--- a/passes/equiv/equiv_opt.cc
+++ b/passes/equiv/equiv_opt.cc
@@ -58,7 +58,7 @@ struct EquivOptPass:public ScriptPass
 	}
 
 	std::string command, techmap_opts;
-	bool assert, undef, multiclock;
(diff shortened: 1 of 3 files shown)


passes/equiv/equiv_opt.cc
```diff
@@ -33,7 +33,7 @@ struct EquivOptPass:public ScriptPass
 		log("    equiv_opt [options] [command]\n");
```
daveshah1: Please add `-async2sync` in the list of options in the help

PR #1443, opened 2019-10-08
sat passes when it should fail if sigmap(wire) has dupes, but differing \init

eddiehung:
Retry of #1431 since that got merged in by accident

### DIFF
--- a/passes/sat/sat.cc
+++ b/passes/sat/sat.cc
@@ -265,15 +265,18 @@ struct SatHelper
 				RTLIL::SigSpec rhs = it.second->attributes.at("\\init");
 				log_assert(lhs.size() == rhs.size());
 
+				dict<RTLIL::SigBit,SigBit> seen_init;
 				RTLIL::SigSpec removed_bits;
 				for (int i = 0; i < lhs.size(); i++) {
 					RTLIL::SigSpec bit = lhs.extract(i, 1);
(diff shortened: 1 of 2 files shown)


cliffordwolf:
Here is a test case that show that this is not really fixing something, just changing one incorrect behavior into another (this is expected to pass):

```
read_verilog -formal <<EOT
module test (input clk, x, y, z, w);
        reg [2:0] a, b, c, d;

        initial a = 2'b01;
        initial b = 2'b10;
        initial c = 2'b01;
        initial d = 2'b10;

        always @(posedge clk) a[0] <= x;
        always @(posedge clk) b[0] <= y;
        always @(posedge clk) c[1] <= z;
        always @(posedge clk) d[1] <= w;

        always @* a[1] = a[0];
        always @* b[1] = b[0];
        always @* c[0] = c[1];
        always @* d[0] = d[1];

        reg first = 1;
        always @(posedge clk) first <= 0;

        always @* begin
                if (first) begin
                        assert (a == 2'b11);
                        assert (b == 2'b00);
                        assert (c == 2'b00);
                        assert (d == 2'b11);
                end
        end
endmodule
EOT

prep -top test
sat -seq 10 -prove-asserts -verify -show-public test
```

The real fix would be to change `proc` to eliminate init attributes on things that don't hold any state.

nakengelhardt:
@eddiehung Should this suffer the same fate as #1566?

PR #1448, opened 2019-10-14
Typedef support (with wrong syntax)

cliffordwolf:
This work is based on #1389. In order to decouple the GLR discussion from the rest of the original PR, this adds typedef support with a wrong syntax, that doesn't require a GLR parser, so we can merge the rest of the PR and figure out what to do about GLR later.

The following would re-enable the correct syntax, using GLR:

```
--- frontends/verilog/verilog_parser.y
+++ frontends/verilog/verilog_parser.y
@@ -180,7 +180,9 @@ struct specify_rise_fall {
 %right UNARY_OPS
 
 %define parse.error verbose
-%define parse.lac full
+//%define parse.lac full
+%glr-parser
+%expect 21
 
 %nonassoc FAKE_THEN
 %nonassoc TOK_ELSE
@@ -292,7 +294,7 @@ hierarchical_id:
        };
 
 hierarchical_type_id:
-       '(' hierarchical_id ')' { $$ = $2; };
+       hierarchical_id { $$ = $1; };
 
 module:
        attr TOK_MODULE TOK_ID {
```

Compared to #1389 this version doesn't have the two rr-conflicts.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -708,6 +708,7 @@ test: $(TARGETS) $(EXTRA_TARGETS)
 	+cd tests/various && bash run-test.sh
 	+cd tests/sat && bash run-test.sh
 	+cd tests/svinterfaces && bash run-test.sh $(SEEDOPT)
+	+cd tests/svtypes && bash run-test.sh $(SEEDOPT)
 	+cd tests/proc && bash run-test.sh
 	+cd tests/opt && bash run-test.sh
 	+cd tests/aiger && bash run-test.sh $(ABCOPT)
--- a/README.md
+++ b/README.md
@@ -510,6 +510,8 @@ from SystemVerilog:
   into a design with ``read_verilog``, all its packages are available to
   SystemVerilog files being read into the same design afterwards.
 
+- typedefs are supported (including inside packages)
+
 - SystemVerilog interfaces (SVIs) are supported. Modports for specifying whether
   ports are inputs or outputs are supported.
 
--- a/frontends/ast/ast.cc
+++ b/frontends/ast/ast.cc
@@ -164,6 +164,8 @@ std::string AST::type2str(AstNodeType type)
 	X(AST_MODPORT)
 	X(AST_MODPORTMEMBER)
 	X(AST_PACKAGE)
+	X(AST_WIRETYPE)
+	X(AST_TYPEDEF)
 #undef X
 	default:
 		log_abort();
@@ -206,6 +208,7 @@ AstNode::AstNode(AstNodeType type, AstNode *child1, AstNode *child2, AstNode *ch
 	was_checked = false;
 	range_valid = false;
 	range_swapped = false;
+	is_custom_type = false;
 	port_id = 0;
 	range_left = -1;
 	range_right = 0;
--- a/frontends/ast/ast.h
+++ b/frontends/ast/ast.h
@@ -148,7 +148,10 @@ namespace AST
 		AST_INTERFACEPORTTYPE,
 		AST_MODPORT,
 		AST_MODPORTMEMBER,
-		AST_PACKAGE
+		AST_PACKAGE,
+
+		AST_WIRETYPE,
+		AST_TYPEDEF
 	};
 
 	// convert an node type to a string (e.g. for debug output)
@@ -174,7 +177,7 @@ namespace AST
 		// node content - most of it is unused in most node types
 		std::string str;
 		std::vector<RTLIL::State> bits;
-		bool is_input, is_output, is_reg, is_logic, is_signed, is_string, is_wand, is_wor, range_valid, range_swapped, was_checked, is_unsized;
+		bool is_input, is_output, is_reg, is_logic, is_signed, is_string, is_wand, is_wor, range_valid, range_swapped, was_checked, is_unsized, is_custom_type;
 		int port_id, range_left, range_right;
 		uint32_t integer;
 		double realvalue;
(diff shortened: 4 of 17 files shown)


cliffordwolf:
> investigating the downsides of GLR more thoroughly and testing on more cases (perhaps even enumerating all RR conflicts)

There are no RR conflicts left with this version, even when switching to the correct syntax for types.

But there are still 21 shift-reduce conflicts left.

My two primary concerns are

(1) what if in the future we encounter test cases that are broken and we are at `[large N]` shift-reduce conflicts. how will we be able to fix it.

(2) the moment we switch to GLR we will never be able to inspect the parse state inside the lexer for any reason, because we will never be sure how far the lexer is ahead of the parser.

> looking at ways of rewriting the parser so GLR isn't needed. But because of the arbitrary expressions possible in either a packed array type or an array of cells before the point of disambiguation; I think this could be quite tricky.

Another way would be to write a horrible look-ahead hack in the lexer that would cause the lexer to return TOK_TYPE_ID instead of TOK_ID when it is a type name. I don't think the flex "trailing context" feature isn't nearly sufficient to do that, so likely that would mean hand-writing a lexer.

In summary: I think there are only bad options. The question is which of them is the least bad option.

daveshah1:
FWIW, inspecting the Bison output which I had missed before, it seems like the unexpectedly large number of SR conflicts mostly come from cases where the type is optional (because the type name could also be the wire/parameter name without a type) as well as the more obvious conflict with instantiations.

I am in agreement that neither option is great. To me GLR certainly seems like a cleaner solution. My understanding of Bison isn't great, but from what I've seen and read irresolvable conflicts will be a runtime error rather than an incorrect result, so testing it shouldn't be too hard.

As for inspecting parser state inside the lexer; I would have thought that the whole point of GLR is to avoid the need to ever do that, are there any cases where think it might still be needed?

cliffordwolf:
>  it seems like the unexpectedly large number of SR conflicts mostly come from cases where the type is optional

Ah! I see. I noticed that only one SR conflict comes from the conflict with cell instances. So maybe we can actually reduce the number of SR conflicts to just one? That would be great. Then I'd be totally happy with the GLR solution.

> My understanding of Bison isn't great, but from what I've seen and read irresolvable conflicts will be a runtime error rather than an incorrect result, so testing it shouldn't be too hard.

Yep. I'm more concerned about what do we do if testing reveals a problem with the parser but we don't really understand the parser because we just ignored all the SR conflicts.

> As for inspecting parser state inside the lexer; I would have thought that the whole point of GLR is to avoid the need to ever do that, are there any cases where think it might still be needed?

You'd definitely need it if the grammar isn't context free.

Otherwise I think the opposite case would be more interesting: Inspecting the lexer state from the parser. This is still possible with GLR parsers if one simply puts the relevant parser state directly in the lexer tokens. It's just less convenient. For example: Allow proper per-net settings of `default_nettype`.

I don't know if there is a feature in bison GLR that allows you to tell the parser to force it being in a resolved state (or otherwise abort) at a certain point in the grammar, such as the end of a statement. That would simplify things a bit in this regard.

cliffordwolf:
> I can't immediately think how to do this, at least without very significant restructuring

I've had a closer look yesterday. At least some of the cases have the form `... wire_type {...} foobar ...` and there wouldn't be a conflict if it were `... wire_type foobar {...} ...` instead. But that means that right now `foobar` operates on the state set up by `{...}` and that would need to be changed so that instead `foobar` would return something that `{...}` would then process. In the cases I looked at this indeed would be significant restructuring.

swetland:
Having the lexer resolve ID vs TYPE_ID feels straightforward(-ish) to me...  Is there a reason the identifier for tokens:
```
[a-zA-Z_$][a-zA-Z0-9_$]* {
        frontend_verilog_yylval.string = new std::string(std::string("\\") + yytext);
        return TOK_ID;
}
```
Couldn't see if the token string exists in the type symbol table, doing something like:
`return is_type_id(yytext) ? TOK_TYPE_ID : TOK_ID;`

daveshah1:
The main problem with this is scoping around packages, as package content is processed at the AST level long after the tokeniser is done - this would need quite a bit of refactoring

PeterCrozier:
This isn't as hard as you think. As **swetland** notes, all you have to do is have the lexer lookup the token is the symbol table to see if it is a known type.  The problem is that there is no symbol table: there is only the AST tree.  You can't search the tree looking for types as types and params from packages are only appended to the tree after parsing in preparation for the simplify stage when the tree is patched up.

However we can fake a symbol table with a simple vector of type names shared between the lexer and the parser.  When the parser identifies a typedef statement it saves the name in the list and the lexer uses this list to separate `TOK_ID` from `TOK_TYPE_ID`.

At the end of each compilation unit we can scan the completed trees in the design looking for packages and make a list of the types therein with their package qualified names.  We share this with the lexer and it checks any package qualified symbol names, `pkg::name`, against that list to identify a type name.  It would be more elegant to build this package list as we compile packages but I am trying to keep this simple.

This is less than a hundred lines of code.  I took some code from my old implementation from 2 years ago and added it to the current yosys build.  It works in that it passes the `svtypes` tests when the grammar is reverted to standard.  I will post a PR later tonight.

Trying to resolve an ambiguity like this with GLR seems wishful thinking.

Looking to the future, if you do have problems that don't fit LALR(1) I would not be scared of going down the route of a hand written lexer.  They are easy to write and give you a lot of flexibility to get round gnarly grammars.  

You have spent 4 months talking about this.  I suggest you apply this PR and get the show on the road.

PeterCrozier:
PR #1725

PR #1449, opened 2019-10-16
Improvements for gowin support

pepijndevos:
I had these changes laying around, thinking I'd submit a PR once it was "complete", but seems @dh73 and others are actually [doing stuff with them](https://twitter.com/dhdezr73/status/1093336976399441920), I figured better get it upstream now.

My branch has:
 * bram init
 * lots of dff types
 * some fixes
 * updated for Gowin IDE 1.9


mmicko:
@pepijndevos yes, please merge in latest master into this branch and try adding new tests for gowin (in yosys/tests/arch/gowin) . For simplicity just copy giles from efinix and check for proper expected results. Ping me if there are questions

pepijndevos:
Compared to efinix, the following tests are incomplete:

- Latches are not implemented yet
- FSM has a similar bug to efinix and seems too complex to pin down in numbers
- Memory maps to dram due to small size, only the number of dram cells is tested.
- Tribuf is not implemented yet

techlibs/gowin/cells_sim.v
```diff
@@ -69,10 +271,80 @@ module GSR (input GSRI);
...
+parameter MULT = 9;
+
+parameter ALU_MODE = 0;
+
+reg S, C;
+
+assign SUM = S ^ CIN;
+assign COUT = S? CIN : C;
+
+always @(I0, I1, I3,CIN) begin
```
daveshah1: Please use `always @*`

techlibs/gowin/cells_sim.v
```diff
@@ -69,10 +271,80 @@ module GSR (input GSRI);
...
+module ALU (SUM, COUT, I0, I1, I3, CIN);
+
+input I0;
+input I1;
+input I3;
+input CIN;
+output SUM;
+output COUT;
+
+parameter ADD = 0;
```
daveshah1: These should be `localparam`, not `parameter`

techlibs/gowin/cells_sim.v
pepijndevos: Okay. As a Verilog noob: Why?

daveshah1: No risk of very weird behaviour (and sim/synth mismatches, but that's not so relevant here) if you add another signal as an input to the block; but forget to update the sensitivity list

techlibs/gowin/cells_sim.v
pepijndevos: `ALU_MODE` should be a parameter though, right? If I googled correctly...

daveshah1: Yes, sorry for the ambiguity, I was referring to the internally-used constants

pepijndevos:
So I just added some tweaks to the ALU sim model as suggested by @daveshah1, and somehow it changed the number of LUTs used in the MUX test. Just to indicate how fragile these tests are...

techlibs/gowin/brams_map.v
```diff
@@ -94,7 +112,10 @@ module \$__GW1NR_SDP (CLK2, CLK3, A1ADDR, A1DATA, A1EN, B1ADDR, B1DATA, B1EN);
 			.WREA(A1EN),   .OCE(1'b0),
 			.WREB(1'b0),   .CEB(B1EN), .CEA(1'b1),
 			.RESETA(1'b0), .RESETB(1'b0), .BLKSEL(3'b000),
-                        .DI(A1DATA), .DO(B1DATA), .ADA(A1ADDR), .ADB(B1ADDR)
+			.DI({{(32-CFG_DBITS){1'b0}}, A1DATA}),
+			.DO({open, B1DATA}),
+			.ADA({A1ADDR, {(12-CFG_ABITS){1'b0}}, 2'b11}),
```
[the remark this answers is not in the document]
pepijndevos: From the primitive manual:
> If SDP is configured as 16bit/32bit and SDPX9 is configured as
18bit/36bit, the byte enable function of BSRAM can be realized; i.e., the
data written to memory are controlled by the low four bit of AD, high-level
enable. ADA[0] controls whether to write DI[7:0]/DI[8:0] to memory; ADA[1]
controls whether to write DI[15:8]/DI[17:9] to memory; ADA[2] controls
whether to write DI[23:16]/DI[26:18] to memory; ADA[3] controls whether to
write DI[31:24]/DI[35:27] to memory.

Uhm, I'd have to think about that a bit more to see what is going on and what the current code is actually doing. A lot of this was @dh73 s work I think. At least I can't really remember what I was trying to do here.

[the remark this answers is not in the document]
daveshah1: That description sounds _very_ similar to how byte enables work on ECP5 too, albeit for ECP5 36-bit is a different primitive so appears different from a synthesis point of view (but the hardware still uses the 3 lowest address wires)

pepijndevos:
Okay, I've added the byte-enable as @daveshah1 suggested. Here are some other things I'm aware of that are missing

- [ ] Wide LUTs.
- [ ] DSP stuff.
- [ ] More BRAM modes, especially `SDPX9`. My current understanding is that single port and true dual port don't map to Yosys.
- [ ] More DRAM modes.
- [ ] MUXes, but they make synthesis worse, so not really.
- [ ] Different ALU modes, but not sure how they map to Yosys. Currently using `ADDSUB` which maps nicely to `$alu` cells.
- [ ] Tri-state buffers.
- [ ] Latches. Seems nobody uses them but there is always this "support latches" issue on every HDL thing ever.

pepijndevos:
Sorry, I fucked up formatting pretty badly. I had my editor set wrong, and while trying to fix it, vim decided it was ok to reformat inside comments and *strings*. So after a few tries fixing the problem manually, I used some git fu from https://stackoverflow.com/questions/5669358/can-i-do-a-partial-revert-in-git to just... revert all the madness.

In other news, I've added wide luts with optional abc9. But for [this specific case](https://electronics.stackexchange.com/a/387438/94884), abc9 actually seems to do worse:

```
=== 4 luts, abc ===

   Number of wires:                267
   Number of wire bits:            293
   Number of public wire bits:      16
   Number of cells:                285
     IBUF                            8
     LUT2                           39
     LUT3                           25
     LUT4                          205
     OBUF                            8

=== wide luts, abc ===

   Number of wires:                162
   Number of wire bits:            188
   Number of public wire bits:      16
   Number of cells:                180
     IBUF                            8
     LUT2                            1
     LUT4                           85
     MUX2_LUT5                      42
     MUX2_LUT6                      21
     MUX2_LUT7                      10
     MUX2_LUT8                       5
     OBUF                            8

=== 4 luts, abc9

   Number of wires:                286
   Number of wire bits:            312
   Number of public wire bits:      16
   Number of cells:                304
     IBUF                            8
     LUT2                           49
     LUT3                           26
     LUT4                          213
     OBUF                            8

=== wide luts, abc9

   Number of wires:                246
   Number of wire bits:            272
   Number of public wire bits:      16
   Number of cells:                264
     IBUF                            8
     LUT4                          128
     MUX2_LUT5                      64
     MUX2_LUT6                      32
     MUX2_LUT7                      16
     MUX2_LUT8                       8
     OBUF                            8

```

daveshah1:
The area benefit from abc9 when using wide LUTs comes from the support for LUT libraries with per-input delays. Without these it's fundamentally got the same broken model as abc to work with.

dh73:
> Okay, I've added the byte-enable as @daveshah1 suggested. Here are some other things I'm aware of that are missing
> 
> * [x]  Wide LUTs.
> * [x]  Tri-state buffers.
>   
>   * [x]  TBUF
>   * [x]  IOB
> * [ ]  DSP stuff.
> * [ ]  More BRAM modes, especially `SDPX9`. My current understanding is that single port and true dual port don't map to Yosys.
> * [ ]  More DRAM modes.
> * [ ]  Different ALU modes, but not sure how they map to Yosys. Currently using `ADDSUB` which maps nicely to `$alu` cells.
> * [ ]  Latches. Seems nobody uses them but there is always this "support latches" issue on every HDL thing ever.
> * [ ]  MUXes, but they make synthesis worse, so not really.

This is only an **advice**: I would work on all those, let's call it, "advanced features" right before having the "basic ones" already working in the end-to-end flow (Yosys->NextPNR). Because, for example, to get the best from abc9, you need to know primitive delays, and that's the info you got by fuzzing/knowing the device. I would not add support for all types of memories and DSP yet, to Yosys, unless I really need them for a design that I'm actually using for fuzzing configuration bits of that specific primitive. 
LUT models, RAM16S4/SDP, IO buffers are sufficient for a PoC. Once you reach that point, fixing the QoR must be the next goal.

pepijndevos:
This is why counting instances might not be a very good testcase: https://travis-ci.org/YosysHQ/yosys/builds/603937909

So current status of the tests are:

- Latches are not implemented yet
- FSM has a bug in equivalence
- Memory maps to dram
- Mux test is unstable and has an equivalence bug with wide muxes

PR #1455, opened 2019-10-18
Add BRAM and URAM mapping for UltraScale[+]

daveshah1:


techlibs/xilinx/synth_xilinx.cc
```diff
@@ -296,15 +304,21 @@ struct SynthXilinxPass : public ScriptPass
 				run("read_verilog -lib +/xilinx/xc6v_cells_xtra.v");
 			else if (family == "xc7")
 				run("read_verilog -lib +/xilinx/xc7_cells_xtra.v");
-			else if (family == "xcu" || family == "xcup")
+			else if (family == "xcu" || family == "xcup") {
+				// Also read xc7 blackboxes; as many old cells are upgraded
+				// automatically, and some designs rely on this
+				run("read_verilog -lib +/xilinx/xc7_cells_xtra.v");
 				run("read_verilog -lib +/xilinx/xcu_cells_xtra.v");
```
koriakin: This doesn't look good.  Which old cells do you care about here?  IMO we should either cherry-pick the cells into the xcu list (if there's just a handful of them), or just merge all the blackbox files into a single `cells_xtra.v` (if there are lots).

techlibs/xilinx/xcu_cells_xtra.v
```diff
@@ -8330,412 +8330,6 @@ module FIFO36E2 (...);
     input WREN;
 endmodule
 
-module RAMB18E2 (...);
```
koriakin: Please do not modify this code without corresponding modifications to the generator.

techlibs/xilinx/xcu_brams.txt
```diff
@@ -0,0 +1,105 @@
+
```
koriakin: This file is identical to `xc7_brams.txt`, can't we just reuse it?

eddiehung:
> Looks OK in general, modulo inline comments. The blackbox library needs a bit of discussion though (I think the simplest way would be to just merge all the blackbox libraries into single `cells_xtra.v`, and also merge `*_brams_bb.v` into it — IMO there's no reason for these to be a separate file). What do you think? cc @eddiehung on that.

Agreed. There should be no compatibility issues with overlapping blackboxes, however, if you use the wrong instantiation (a URAM) on a older arch it will still pass through the flow error free. Otherwise, macros?

koriakin:
@eddiehung by "macros" do you mean "`ifdef"? If so, I agree.  I can make the patch, but I'd rather avoid conflicting with this PR — if we agree that a merged file + ifdefs is the correct way forward, I'd suggest merging this PR as-is (modulo other comments), and then I'll make a new PR on top of the new master.

techlibs/xilinx/synth_xilinx.cc
daveshah1: Added the PLLE2 and MMCME2 cells for the purpose of this PR, as LiteX and possibly some other things need them (due to the high cross-compatibility), but it seems like this will only be temporary based on the comments below.

techlibs/xilinx/xcu_cells_xtra.v
daveshah1: Done. Also removed some latches from the generator that were no longer in cells_xtra.

techlibs/xilinx/xcu_brams.txt
daveshah1: Done

daveshah1: Done

daveshah1:
I would agree that we should have all older primitives available in cells_xtra; and merge the files to avoid duplication. `ifdef seems like as a good a solution as any.

PR #1457, opened 2019-10-19
Makefile: don't assume python is called `python3`

xobs:
On some architectures, notably on Windows, the official name for the
Python binary from python.org is `python`.  The build system assumes
that python is called `python3`, which breaks under this architecture.

There is already infrastructure in place to determine the name of the
Python binary when building PYOSYS.  Since Python is now always required
to build Yosys, enable this check universally which sets the
`PYTHON_EXECUTABLE` variable.

Then, reuse this variable in other Makefiles as necessary, rather than
hardcoding `python3` everywhere.


mmicko:
Thanks, LGTM

PR #1466, opened 2019-10-24
Added flag to write_blif

maustin8:
Hi,

For a tool that I am helping develop, we need a completely flattened BLIF file without .subckt or .gate attributes in it. The libraries used by the tool are inspired by ABC and have similar capabilities in its BLIF reader meaning that it cannot handle any hierarchy, so I added a flag to the write_blif command in /backends/blif/blif.cc, called "flatten", which is based on the internal BLIF files that are created to be used by ABC command in /passes/techmap/abc.cc.

Thank you.

-Max Austin

### DIFF
--- a/backends/blif/blif.cc
+++ b/backends/blif/blif.cc
@@ -31,6 +31,52 @@
 USING_YOSYS_NAMESPACE
 PRIVATE_NAMESPACE_BEGIN
 
+enum class gate_type_t {
+	G_NONE,
+	G_FF,
+	G_BUF,
+	G_NOT,
+	G_AND,
+	G_NAND,
+	G_OR,
+	G_NOR,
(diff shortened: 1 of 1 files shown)


cliffordwolf:
> For a tool that I am helping develop, we need a completely flattened BLIF file without .subckt or .gate attributes in it.

Why not simply run `flatten` before running `write_blif` then?

maustin8:
The flatten command still uses "templates" to replace cells and the directives .subckt or .gate are still used to reference them when the file is written which our BLIF parser does not currently handle.

cliffordwolf:
Hello! Thanks for this submission. It's been a while since there was last some activity on this PR and it looks to have been abandoned. Please leave a comment if there's any new information.

PR #1470, opened 2019-10-31
Add CodingReadme section on script passes

cliffordwolf:


ZirconiumX:
For whatever it's worth, I proof-read it and it looks good to me.

PR #1483, opened 2019-11-11
[WIP] Add WASI platform support

whitequark:
WASI stands for "Web Assembly System Interface". Web Assembly is a portable compilation target for systems languages like C++ that can be built once and then compiled again down to any modern 32- or 64-bit ISA. Web Assembly System Interface is a POSIX-like ABI for Web Assembly that enables most libc and a few common POSIX functions. A Web Assembly file built against WASI is a single artifact that can be run with reasonable performance on many major OSes and architectures as easy as `wasm-compiler application.wasm` (where `wasm-compiler` is one of the many implementations like [wasmtime](https://github.com/CraneStation/wasmtime) or [lucet](https://github.com/fastly/lucet)).

The reason I'm interested in WASI is that it would be convenient to ship a fallback Yosys binary built to Web Assembly to lower the effort required to use nMigen, which has Yosys as a hard dependency. Of course, it is possible to build Yosys for all major platforms and architectures directly, but it would be much nicer if building artifacts for a half dozen OS/architecture combinations was someone else's problem, wouldn't it?

This PR requires ABC commit whitequark/abc@5e71f1d0b, which is just the composition of berkeley-abc/abc#57 and berkeley-abc/abc#58. To build Yosys against WASI, download the latest [wasi-sdk](https://github.com/CraneStation/wasi-sdk/releases) and unpack it to `${WASI_PREFIX}`, then run:

```
git -C abc fetch https://github.com/whitequark/abc tmp
make CONFIG=wasi ENABLE_TCL=0 ENABLE_PLUGINS=0 ENABLE_READLINE=0 ENABLE_ZLIB=0 ABCREV=5e71f1d0b
```

This will produce `yosys.wasm`, which can be run with e.g. [wasmtime](https://github.com/CraneStation/wasmtime):

```
cargo install -f wasmtime
wasmtime -o -c --mapdir=/share::share --dir=. ./yosys.wasm -- -p synth_ice40 picorv32.v
```

or [lucet](https://github.com/fastly/lucet):

```
cargo install -f lucetc
RUSTFLAGS="-C link-args=-rdynamic" cargo install -f lucet-wasi
wget https://github.com/fastly/lucet/blob/4d12bc6da6adbbf62b0c2639819baefac84e9320/lucet-wasi/bindings.json
lucetc yosys.wasm --bindings bindings.json -o yosys.so
lucet-wasi --dir share:/share --dir . yosys.so -- -p synth_ice40 picorv32.v
```

Unfortunately... neither of these actually works.

 * wasmtime crashes in an opaque way (CraneStation/wasmtime#537); this could be a bug in wasmtime, a bug in wasi-sdk, a bug in clang, or (unlikely but possible) a bug in yosys. In different conditions (loading an ilang file and runing `synth_ice40`), the crash appears as `__cxa_guard_acquire detected recursive initialization`.
 * lucet crashes trying to compile the binary, also in an opaque way (fastly/lucet#361). The wasm file passes validation so this is definitely a bug in lucet.

In terms of unfinished work, this PR includes two FIXMEs:
 * abc requires `system`, but WASI does not have any way to spawn processes. Stubbing it out (like it's done right now) is probably just fine.
 * more importantly, Web Assembly/WASI currently do not support C++ exceptions. I would personally prefer to add a configuration flag that allows building Yosys without exceptions, since most (all?) of them are intended to crash the process anyway. Web Assembly will gain support for exceptions (and setjmp/longjmp, which are also not supported) at some point in the future but it is unlikely to be soon.

### DIFF
--- a/libs/minisat/System.cc
+++ b/libs/minisat/System.cc
@@ -73,7 +73,7 @@ static inline int memReadPeak(void)
 }
 
 double Minisat::memUsed() { return (double)memReadStat(0) * (double)getpagesize() / (1024*1024); }
-double Minisat::memUsedPeak(bool strictlyPeak) { 
+double Minisat::memUsedPeak(bool strictlyPeak) {
     double peak = memReadPeak() / (double)1024;
     return peak == 0 && !strictlyPeak ? memUsed() : peak; }
 
@@ -101,7 +101,7 @@ double Minisat::memUsedPeak(bool) { return 0; }
 #endif
 
 
-#if !defined(_MSC_VER) && !defined(__MINGW32__)
+#if !defined(_MSC_VER) && !defined(__MINGW32__) && !defined(__wasm)
 void Minisat::limitMemory(uint64_t max_mem_mb)
 {
 // FIXME: OpenBSD does not support RLIMIT_AS. Not sure how well RLIMIT_DATA works instead.
@@ -133,7 +133,7 @@ void Minisat::limitMemory(uint64_t /*max_mem_mb*/)
 #endif
 
 
-#if !defined(_MSC_VER) && !defined(__MINGW32__)
+#if !defined(_MSC_VER) && !defined(__MINGW32__) && !defined(__wasm)
 void Minisat::limitTime(uint32_t max_cpu_time)
 {
     if (max_cpu_time != 0){
@@ -156,9 +156,13 @@ void Minisat::limitTime(uint32_t /*max_cpu_time*/)
 
 void Minisat::sigTerm(void handler(int))
 {
+#if defined(__wasm)
+	(void)handler;
+#else
     signal(SIGINT, handler);
     signal(SIGTERM,handler);
 #ifdef SIGXCPU
     signal(SIGXCPU,handler);
+#endif
 #endif
 }
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ CONFIG := clang
 # CONFIG := gcc-4.8
 # CONFIG := afl-gcc
 # CONFIG := emcc
+# CONFIG := wasi
 # CONFIG := mxe
 # CONFIG := msys2
 # CONFIG := msys2-64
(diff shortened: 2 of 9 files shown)


libs/minisat/System.cc
```diff
@@ -73,7 +73,7 @@ static inline int memReadPeak(void)
 }
 
 double Minisat::memUsed() { return (double)memReadStat(0) * (double)getpagesize() / (1024*1024); }
-double Minisat::memUsedPeak(bool strictlyPeak) { 
+double Minisat::memUsedPeak(bool strictlyPeak) {
```
cliffordwolf: Changes to `libs/minisat/` should be reflected in the patch files in that directory. For example, this could go in `libs/minisat/00_PATCH_typofixes.patch`. The other stuff maybe in a new file such as `libs/minisat/00_PATCH_wasm.patch`.

whitequark:
I'm going to close this for the time being since it looks like getting Yosys to actually run as WASM isn't going to be a quick and easy job. Not so much because it has weird memory corruption, but more because right now it looks like none of the WASM compilers emit usable debug information. That's understandable though as it is a very new technology and I am using it in (AFAICT) completely novel ways.

whitequark:
I'm reopening this PR with much rejoicing, because not only is WASI Yosys clearly practical now, it actually looks very compelling!

eddiehung:
> This PR requires ABC commit [whitequark/abc@5e71f1d](https://github.com/whitequark/abc/commit/5e71f1d0b), which is just the composition of [berkeley-abc/abc#57](https://github.com/berkeley-abc/abc/pull/57) and [berkeley-abc/abc#58](https://github.com/berkeley-abc/abc/pull/58).

Would #1946 be helpful?

whitequark:
@eddiehung Yup, I was excited to see it. Do you think you can cherry-pick those for me?

whitequark:
(Also, why on earth can't I reopen this PR?)

eddiehung:
> (Also, can't I reopen this PR? Seems like GitHub didn't like my last push?)

Reopen is greyed out for me too.

whitequark:
> EDIT: Apparently it's because "The wasi branch was force-pushed or recreated."

Yeah, but... don't we force-push branches all the time?

bjorn3:
If you force-push the old commit as of closing, you should be able to re-open this PR.

whitequark:
@bjorn3 Ah, thanks for explanation. I guess there's a new one already so I won't add even more noise.

PR #1486, opened 2019-11-12
Bugfix in fsm_detect

cliffordwolf:
The old code contained the following code, which when reached would always set `clr_output = false` because the if-condition was always true.

```
if (port_it.first != "\\A" || port_it.first != "\\Y")
    clr_output = false;
```

The new code should do now what the old code was intended to do, I think.

### DIFF
--- a/passes/fsm/fsm_detect.cc
+++ b/passes/fsm/fsm_detect.cc
@@ -173,9 +173,10 @@ static void detect_fsm(RTLIL::Wire *wire)
 
 			if (set_output || clr_output) {
 				for (auto &port_it : cell->connections())
-					for (auto bit : assign_map(port_it.second))
-						if (bit.wire != nullptr && !sig_q_bits.count(bit))
-							goto next_cellport;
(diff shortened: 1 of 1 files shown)


passes/fsm/fsm_detect.cc
```diff
@@ -158,22 +158,24 @@ static void detect_fsm(RTLIL::Wire *wire)
...
+			if (cell->type.in("$eq", "$logic_not", "$reduce_and"))
 				clr_output = true;
 
-			if (!set_output && !clr_output) {
-				clr_output = true;
+			if (set_output || clr_output) {
 				for (auto &port_it : cell->connections())
-					if (port_it.first != "\\A" || port_it.first != "\\Y")
-						clr_output = false;
+					for (auto bit : assign_map(port_it.second))
```
mwkmwkmwk: You should only iterate over cell inputs here

cliffordwolf: duh! thanks for finding that. should be fixed now.

Issue #568, opened 2018-06-13
EDIF generation renumbers multi-bit values


spakin:
The EDIF code produced by `write_edif` is incorrectly renumbering the bits in multi-bit values.

## Steps to reproduce the issue

1. Write Verilog code such as the following with inputs/output containing multiple bits:

```Verilog
module badness (in, out);
   input [1:0] in;
   output [1:0] out;

   assign out[0] = ~in[0];
   assign out[1] = in[1];
endmodule
```
2. Generate EDIF code:
```
yosys> read_verilog badness.v 
yosys> write_edif badness.edif
```

I'm using Yosys 83631555dd5d57e2a3d3b7175f77baf0ad67ccd9 (2018-06-11):

```bash
$ yosys -V
Yosys 0.7+596 (git sha1 83631555, clang 6.0.0 -fPIC -Os)
```

## Expected behavior

I would have expected an EDIF `net` to associate `in[0]` with port 0 of `in` and with the input to an inverter and another `net` to associate `in[1]` with port 1 of `in` and with port 1 of `out`, as in the following EDIF code:

```Lisp
(net (rename id00003 "in[0]") (joined
  (portRef (member in 0))
  (portRef A (instanceRef id00002))
))
(net (rename id00004 "in[1]") (joined
  (portRef (member in 1))
  (portRef (member out 1))
))
(net (rename id00005 "$not$badness.v:5$1_Y") (joined
  (portRef (member out 0))
  (portRef Y (instanceRef id00002))
))
```
In fact, older versions of Yosys get this right.  An example of a good Yosys is the 0.7 release (61f681162754fa170494e567f1a1a9ae2d17eb69):

```bash
$ yosys -V
Yosys 0.7 (git sha1 61f6811, gcc 6.2.0-11ubuntu1 -O2 -fdebug-prefix-map=/build/yosys-OIL3SR/yosys-0.7=. -fstack-protector-strong -fPIC -Os)
```

## Actual behavior

What actually happens is that an EDIF `net` associates port **1** of `in` with the input to the inverter (and with `in[0]`).  Another EDIF `net` associates port **0** of `in` directly with port **0** of `out` (and with `in[1]`)—but no inverter—as in the following EDIF code:

```Lisp
(net (rename id00003 "in[1]") (joined
  (portRef (member in 0))
  (portRef (member out 0))
))
(net (rename id00004 "in[0]") (joined
  (portRef (member in 1))
  (portRef A (instanceRef id00002))
))
[...]

PR #1489, opened 2019-11-12
Add "write_edif -lsbidx"

cliffordwolf:


eddiehung:
For future lifeforms, which tool(s) have you found require this?

nakengelhardt:
@eddiehung: None; this is an incomplete workaround/bug-finding session for trying to export an edif with a blackbox module whose port widths depend on a parameter. Attempting this found several bugs:

- edif doesn't support variable-width signals, but the cell is only declared once, so the instance widths don't match the declaration width
- consequently, if the instance width is larger, the connections end up using negative indices (this is what -lsbidx works around, it reverses the connection direction so at least they are positive, but of course this results in the bits being connected in the wrong order if you instantiate it in a design in vivado - I was doing out-of-context synthesis to look at area and timing, so this didn't matter in my case)
- the cell declaration takes the default parameter width but the wire name takes the width from a random instance, so they don't match (this is what the second commit is about)
- yosys can't distinguish between a port "a" and a port "a[0:0]" and defaults to the former; but these require different syntax to connect to in edif, so the edif ends up broken if one of the instances parametrizes a port to a width of 1.

This draft PR doesn't fix these issues, it's just some code from Clifford that I used to continue tracking down the bugs. An actual fix will probably need to tie into hierarchy to figure out how port widths are affected by parameter settings, and export a separate blackbox cell for each parameter set.

nakengelhardt:
Despite arising out of a different background, this would actually be a partial fix for issue (enhancement) #568. It adds point 1 of the options laid out [here](https://github.com/YosysHQ/yosys/pull/830#issuecomment-468453251). Leaving it open for now so it can be completed later.

peterzieba:
I'm very curious about using the ATF150x CPLDs with yosys. What would be needed for these changes to be merged? I believe this might be the only thing necessary to generated EDIF for the Atmel fitter.

PR #1494, opened 2019-11-15
write_verilog: add -extmem option, to write split memory init files

whitequark:
Some toolchains (in particular Quartus) are pathologically slow if a large amount of assignments in `initial` blocks are used. In practice, this functionality is required for nMigen to be usable with such toolchains.

### DIFF
--- a/backends/verilog/verilog_backend.cc
+++ b/backends/verilog/verilog_backend.cc
@@ -33,11 +33,11 @@
 USING_YOSYS_NAMESPACE
 PRIVATE_NAMESPACE_BEGIN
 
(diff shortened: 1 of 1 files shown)


whitequark:
cc @ZirconiumX

backends/verilog/verilog_backend.cc
```diff
@@ -1885,6 +1943,11 @@ struct VerilogBackend : public Backend {
 				nostr = true;
 				continue;
 			}
+			if (arg == "-extmem") {
+				extmem = true;
```
cliffordwolf: I think there's an `extmem = false;` default assignment missing at the start of this function.

whitequark: Thanks, fixed

PR #1495, opened 2019-11-18
xilinx: Add simulation models for MULT18X18* and DSP48A*.

mwkmwkmwk:
This adds simulation models for the following primitives:

- MULT18X18 and MULT18X18S (Virtex 2*, Spartan 3)
- MULT18X18SIO (Spartan 3E, Spartan 3A)
- DSP48A (Spartan 3A DSP) — implemented in terms of DSP48A1
- DSP48A1 (Spartan 6)

Since the DSP models are already getting unwieldy (they're already
larger than all the other simulation models combined), and will be
getting even bigger when more devices are supported, they're split
off to a separate file.


mwkmwkmwk:
The original idea was to split off the DSP code into `cells_sim_dsp.v`, but I backed out of that, because I couldn't easily do it without breaking search paths somewhere.  If we want to do that, I'll put it off until another PR.

eddiehung:
As an aside that can be part of this PR or another one, since you are visiting these DSP models, is it necessary to apply `(* invertible_pin= "IS_*_INVERTED" *)` attributes?

mwkmwkmwk:
> As an aside that can be part of this PR or another one, since you are visiting these DSP models, is it necessary to apply `(* invertible_pin= "IS_*_INVERTED" *)` attributes?

No — this parameter isn't present in any of these cells (they were removed before Vivado happened and introduced the inversion params), so there are no netlists out there that would require it.  We may want to add one ourselves though, if `xilinx_dsp` ever learns about negedge flops and wants to utilize the hardware inversion.

Issue #285, opened 2016-12-26
Probem cross-compiling ABC (From Yosys)


udif:
I'm trying to cross compile Yosis for Windows using docker and the mxe target;

```
FROM debian:jessie
MAINTAINER Udi Finkelstein <github@udifink.om>
RUN \
  echo "deb http://pkg.mxe.cc/repos/apt/debian jessie main" > /etc/apt/sources.list.d/mxeapt.list && \
  apt-key adv --keyserver keyserver.ubuntu.com --recv-keys D43A795B73B16ABE9643FE1AFD8FFF16DB45C6AB && \
  apt-get update -y && \
  apt-get install -y \
  autoconf \
  bison \
  flex \
  g++ \
  gawk \
  gettext \
  git \
  libffi-dev \
  make \
  mercurial \
  mxe-i686-w64-mingw32.static-gcc \
  pkg-config \
  python3 \
   tcl-dev \
  xdot

RUN git clone https://github.com/cliffordwolf/yosys.git && cd yosys && make config-mxe && \
  ln -s /usr/lib/mxe /usr/local/src/mxe && \
  make mxebin

RUN git clone https://github.com/cliffordwolf/icestorm.git icestorm && cd icestorm && make -j$(nproc) mxebin

RUN git clone https://github.com/cseed/arachne-pnr.git && cd arachne-pnr && make mxebin
```

Compilation progresses fine, until I get the following error while compiling ABC:

```
src/proof/acec/acecSt.o: In function `Z17Acec_StatsCollectP10Gia_Man_t_i':
/yosys/abc/src/proof/acec/acecSt.c:409: undefined reference to `Dar_LibReturnClass(unsigned int)'
collect2: error: ld returned 1 exit status
make[1]: *** [abc-a4872e22c646.exe] Error 1
```

After some investigation it seems that the cross compile configs always use '-x c++' causing all files to be parsed as C++ (despite their '.c' suffix).
The file defining Dar_LibReturnClass() (sorry, forgot the path for this file) is compiled as C, while src/proof/acec/acecSt.c is compiled as C++, assuming Dar_LibReturnClass() is a C++ func and calls it with a mangled filename.
The non-cross-compile versions successfully compile ABC as C code.
This also happened when I tried to use the Ubuntu built-in mingw32 cross compile packages (I created soft links from /usr/bin/mingw... to the hardcoded path in the Yosys Makefile).

[...]

PR #1502, opened 2019-11-18
use CC not CXX to compile abc

rxrbln:
Compiling Yosys w/ "make CC=gcc CPP=cpp CXX=g++ PREFIX=/usr docdir=/usr/doc/yosys ABCREV=default" I otherwise get:
/usr/lib64/gcc/x86_64-t2-linux-gnu/8.1.0/../../../../x86_64-t2-linux-gnu/bin/ld: src/base/main/mainInit.o: in function `Abc_FrameInit':
mainInit.c:(.text+0xcf): undefined reference to `Glucose_Init(Abc_Frame_t_*)'
/usr/lib64/gcc/x86_64-t2-linux-gnu/8.1.0/../../../../x86_64-t2-linux-gnu/bin/ld: src/base/main/mainInit.o: in function `Abc_FrameEnd':
mainInit.c:(.text+0x1b1): undefined reference to `Glucose_End(Abc_Frame_t_*)'
/usr/lib64/gcc/x86_64-t2-linux-gnu/8.1.0/../../../../x86_64-t2-linux-gnu/bin/ld: src/base/acb/acbFunc.o: in function `Acb_NtkEcoPerform(Acb_Ntk_t_*, Acb_Ntk_t_*, char*, int)':
acbFunc.c:(.text+0xd125): undefined reference to `Abc_SopSynthesizeOne(char*, int)'

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -131,7 +131,7 @@ bumpversion:
 ABCREV = 623b5e8
 ABCPULL = 1
 ABCURL ?= https://github.com/berkeley-abc/abc
-ABCMKARGS = CC="$(CXX)" CXX="$(CXX)" ABC_USE_LIBSTDCXX=1
+ABCMKARGS = CC="$(CC)" CXX="$(CXX)" ABC_USE_LIBSTDCXX=1
 
 # set ABCEXTERNAL = <abc-command> to use an external ABC instance
 # Note: The in-tree ABC (yosys-abc) will not be installed when ABCEXTERNAL is set.


udif:
You might want to revisit #285 and see if it is still relevant.

rxrbln:
> You might want to revisit #285 and see if it is still relevant.

Well, it occurs with yosys f02623a from some weeks ago as packaged in #t2sde (https://t2sde.org/packages/yosys), coming natively. Looks like C++ symbol mangling vs. plain C. gcc vs. g++ mismatch, ..?

mmicko:
Actually it will be enough to use make CC=gcc CXX=gcc PREFIX=/usr docdir=/usr/doc/yosys ABCREV=default"  
This way you would replicate exactly what yosys is doing inside it's Makefile 
Or even better do it two-step:
make config-gcc
make PREFIX=/usr docdir=/usr/doc/yosys ABCREV=default"

rxrbln:
not really best practice, right?

mmicko:
Right, but issue is on abc compile side, and mixing C and C++ there, and you were right, name mangling is actual problem there

PR #1503, opened 2019-11-19
`verific` pass to print help message when command syntax error

eddiehung:
Making it consistent with other Yosys passes.

~The other change of note is to also print out the help message even when it wasn't compiled with Verific -- so users can see what they're missing out on!~


frontends/verific/verific.cc
```diff
@@ -2073,8 +2073,10 @@ struct VerificPass : public Pass {
 	{
 		static bool set_verific_global_flags = true;
 
-		if (check_noverific_env())
+		if (check_noverific_env()) {
+			help();
 			log_cmd_error("This version of Yosys is built without Verific support.\n");
```
cliffordwolf: That's not a command syntax error. I don't think we actually want to print the help message here.

frontends/verific/verific.cc
```diff
@@ -2490,6 +2492,7 @@ struct VerificPass : public Pass {
 	}
 #else /* YOSYS_ENABLE_VERIFIC */
 	void execute(std::vector<std::string>, RTLIL::Design *) YS_OVERRIDE {
+		help();
 		log_cmd_error("This version of Yosys is built without Verific support.\n");
```
cliffordwolf: Same as above. Not a command syntax error.

frontends/verific/verific.cc
eddiehung: Is there no interest in printing out the help message even on non Verific builds so that users can "see what you could have won!!11!!"?

cliffordwolf: Wouldn't that be `help verific`?

eddiehung: Fair enough. Reverted in d730bba6d

PR #1505, opened 2019-11-20
xaig_dff to support async flops $_DFF_[NP][NP][01]_

eddiehung:


tests/various/abc9.ys
```diff
@@ -22,3 +22,19 @@ abc9 -lut 4
 select -assert-count 1 t:$lut r:LUT=2'b01 r:WIDTH=1 %i %i
 select -assert-count 1 t:unknown
 select -assert-none t:$lut t:unknown %% t: %D
+
+design -load read
+hierarchy -top abc9_test031
+proc
+async2sync
```
daveshah1: What is the purpose of the test? Doesn't async2sync here mean that the async resets are lost?

eddiehung: Oops. I meant to use `clk2fflogic`. Fixed.

PR #1508, opened 2019-11-21
xilinx: Improve flip-flop handling.

mwkmwkmwk:
This adds support for infering more kinds of flip-flops:

- FFs with async set/reset and clock enable
- FFs with sync set/reset
- FFs with sync set/reset and clock enable

Some passes have been moved (and some added) in order for dff2dffs to
work correctly.

This gives us complete coverage of Virtex 6+ and Spartan 6 flip-flop
capabilities (though not latch capabilities).  Older FPGAs also support
having both a set and a reset input, which will be handled at a later
data.

### DIFF
--- a/techlibs/xilinx/synth_xilinx.cc
+++ b/techlibs/xilinx/synth_xilinx.cc
@@ -441,6 +441,16 @@ struct SynthXilinxPass : public ScriptPass
 		}
 
 		if (check_label("map_ffram")) {
+			// Required for dffsr2dff to work.
+			run("simplemap t:$dff t:$adff t:$mux");
+			// Needs to be done before opt -mux_bool happens.
+			run("dffsr2dff");
+			if (help_mode)
+				run("dff2dffs [-match-init]", "(-match-init for xc6s only)");
+			else if (family == "xc6s")
+				run("dff2dffs -match-init");
+			else
+				run("dff2dffs");
 			if (widemux > 0)
 				run("opt -fast -mux_bool -undriven -fine"); // Necessary to omit -mux_undef otherwise muxcover
 									    // performs less efficiently
@@ -450,14 +460,11 @@ struct SynthXilinxPass : public ScriptPass
 		}
 
 		if (check_label("fine")) {
-			run("dffsr2dff");
-			run("dff2dffe");
+			run("dff2dffe -direct-match $_DFF_* -direct-match $__DFFS_*");
 			if (help_mode) {
-				run("simplemap t:$mux", "         ('-widemux' only)");
 				run("muxcover <internal options>, ('-widemux' only)");
 			}
 			else if (widemux > 0) {
-				run("simplemap t:$mux");
 				constexpr int cost_mux2 = 100;
 				std::string muxcover_args = stringf(" -nodecode -mux2=%d", cost_mux2);
 				switch (widemux) {
--- a/techlibs/xilinx/synth_xilinx.cc
+++ b/techlibs/xilinx/synth_xilinx.cc
@@ -564,6 +564,7 @@ struct SynthXilinxPass : public ScriptPass
 			else
 				techmap_args += " -map " + ff_map_file;
 			run("techmap " + techmap_args);
+			run("xilinx_dffopt");
 			run("clean");
 		}
 
--- a/techlibs/xilinx/xc6s_ff_map.v
+++ b/techlibs/xilinx/xc6s_ff_map.v
@@ -27,6 +27,8 @@
 
 `ifndef _NO_FFS
 
+// No reset.
+
 module  \$_DFF_N_   (input D, C, output Q);
   parameter [0:0] _TECHMAP_WIREINIT_Q_ = 1'bx;
   generate if (_TECHMAP_WIREINIT_Q_ === 1'b1)
@@ -45,6 +47,8 @@ module  \$_DFF_P_   (input D, C, output Q);
   endgenerate
   wire _TECHMAP_REMOVEINIT_Q_ = 1;
 endmodule
+
+// No reset, enable.
 
 module  \$_DFFE_NP_ (input D, C, E, output Q);
   parameter [0:0] _TECHMAP_WIREINIT_Q_ = 1'bx;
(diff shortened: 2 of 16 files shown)


techlibs/xilinx/synth_xilinx.cc
```diff
@@ -450,14 +460,13 @@ struct SynthXilinxPass : public ScriptPass
 		}
 
 		if (check_label("fine")) {
-			run("dffsr2dff");
-			run("dff2dffe");
+			// Required for dff2dffe to work well.
+			run("clean");
```
eddiehung: `clean` calls can be quite expensive. Can you share what this call cleans up to help `dff2dffe` do its job more efficiently?

techlibs/xilinx/synth_xilinx.cc
```diff
@@ -441,6 +441,16 @@ struct SynthXilinxPass : public ScriptPass
 		}
 
 		if (check_label("map_ffram")) {
+			// Required for dffsr2dff to work.
+			run("simplemap t:$dff t:$adff t:$mux");
+			// Needs to be done before opt -mux_bool happens.
+			run("dffsr2dff");
```
eddiehung: What's your take on making `dffsr2dff` be `$[a]dff` aware, just like `dff2dffe`?

techlibs/xilinx/xc6s_ff_map.v
```diff
@@ -139,6 +153,260 @@ module  \$_DFF_PP1_ (input D, C, R, output Q);
...
+  endgenerate
+  wire _TECHMAP_REMOVEINIT_Q_ = 1;
+endmodule
+module  \$__DFFSE_PN1 (input D, C, E, R, output Q);
+  parameter [0:0] _TECHMAP_WIREINIT_Q_ = 1'bx;
+  generate if (_TECHMAP_WIREINIT_Q_ === 1'b0)
+    $error("Spartan 6 doesn't support FFs with set initialized to 0");
+  else
+    wire NR;
+    INV inv (.O(NR), .I(R));
```
daveshah1: Would it make more sense not to use a generic not (~) here which may well end up folded into a LUT? afaik hardware R inversion only starts from Ultrascale, which should probably be a special case

eddiehung: I think the abc1 flow does ff_map.v after ABC. The abc9 flow does ff_map.v before which makes this possible. Perhaps the abc1 flow needs to convert all neg polarity resets to positive polarity to work around this?

techlibs/xilinx/synth_xilinx.cc
mwkmwkmwk: First, `dff2dffe` is not `$adff` aware — it only knows about `$dffe`.

Second, fixing that would require yet more cell types, and also making the `$__DFF*` cells into first-class primitives (so that we have something to simplemap into).  I'm not against it, but it's a *big* combinatorial explosion, and we'd need to discuss it properly.

techlibs/xilinx/xc6s_ff_map.v
mwkmwkmwk: Not really — hardware R inversion has been supported since always, so the INV will always be folded into the slice (and is, in fact, the proper and only way to describe a negative-polarity reset in ISE).

daveshah1: I'm curious now, is it part of the slice configuration? Xray doesn't have bits for it for xc7, but nor have I ever seen Vivado use it.

techlibs/xilinx/synth_xilinx.cc
mwkmwkmwk: When `dff2dffs` folds a mux with a const input into the FF as a sync set/reset, it leaves the original mux still there, unused. When `dff2dffe` runs, it notices that the second mux that would have been folded into CE has more than one user (the now-dead set/reset mux), and will not perform the conversion.

techlibs/xilinx/xc6s_ff_map.v
mwkmwkmwk: Alright, I'd like to retract my earlier comment.  As it turns out, the inverter *was* there up to Spartan 3 / Virtex 4, but was removed on Spartan 6 / Virtex 5. So this is actually a valid concern, and I will look into how we can handle it in the flow...

eddiehung: Wait. What I'm suggesting should already be in place: https://github.com/YosysHQ/yosys/blob/0ac330bb81946081a3ab9ae45ef5aa7122dcb786/techlibs/xilinx/cells_map.v#L21-29. For architectures that do support R inversion, this could be disabled but right now it should work?

techlibs/xilinx/synth_xilinx.cc
eddiehung: Sorry. I meant `$dff` (word level cells) aware. Long term, it would be good to preserve as much of the word level operations as possible, with a view to word level synthesis, but that's beyond the scope here.

Discussions with @cliffordwolf have led to the consensus we should add 16 new `$_DFFE_*` cells, since it is a common enough pattern.

mwkmwkmwk: Which cells do you mean?  As far as I can tell, we're missing the following cells that could be plausibly advocated for here:

- word-level ff with async set/reset and enable
- word-level ff with sync set/reset
- word-level ff with sync set/reset and enable
- bit-level ff with async set/reset and enable (currently `$__DFFE_[NP][NP][01]` — 8 cells)
- bit-level ff with sync set/reset (currently `$__DFFS_[NP][NP][01]_` — 8 cells)
- bit-level ff with sync set/reset and enable (currently `$__DFFSE_[NP][NP][01]` — 8 cells)

This doesn't count further cell types that we would need to support FFs with both set and reset (another 3 coarse + 24 fine by my count)

techlibs/xilinx/xc6s_ff_map.v
mwkmwkmwk: Not yet — it needs to be extended to cover the newly-emitted cells (with enable and with sync reset) as well.  But that seems to be an easy change...

mwkmwkmwk: Added more rules there — seems to work.  It seems to me that the rules for negative-reset flops in `*_ff_maps.v` are effectively unreachable and can be completely removed, correct?

techlibs/xilinx/synth_xilinx.cc
daveshah1: The other question is whether you then also add variants with enable-over-sync-reset priority; as required for iCE40 and I think Intel FPGAs; which would be another 8 cells

techlibs/xilinx/xc6s_ff_map.v
eddiehung: Yes, I think it's reasonable to remove them for now. We can add handling back for the parts that do support SR inversion later; this would be low priority though, as it would only be helpful for cases where it would save a logic level right?

techlibs/xilinx/synth_xilinx.cc
eddiehung: Have you considered putting `clean` call after `dff2dffs` so that `-run fine:` doesn't start with a clean?

techlibs/xilinx/xc6s_ff_map.v
mwkmwkmwk: I think we wouldn't even need to do that — P&R already looks for inverters to merge itself. At most we may need to make sure that INV cell is used instead of LUT1 (which can be done by changing `$lut` mapping).

techlibs/xilinx/synth_xilinx.cc
eddiehung: Let's revisit this in a future PR

techlibs/xilinx/synth_xilinx.cc
eddiehung: Wait, wouldn't https://github.com/YosysHQ/yosys/blob/c5f2a530c4f3b661508fe1c96d928a252628fc68/techlibs/xilinx/synth_xilinx.cc#L454-L458 already run `opt_clean`?

techlibs/xilinx/xc6s_ff_map.v
eddiehung: Yup. But it might be useful for Yosys to do this when making area/timing decisions though.

techlibs/xilinx/synth_xilinx.cc
mwkmwkmwk: Good point, removed the clean call.

mwkmwkmwk:
Sorry for the push-force; the changes are:

- rebase on master
- change LUT1 to INV in tests to match current techmapping on master
- fix cell names in `cells_map.v` (accidentally had three undescores instead of two)

PR #1509, opened 2019-11-21
xilinx: Add simulation model for DSP48 (Virtex 4).

mwkmwkmwk:


mwkmwkmwk:
Added tests and updated; The tests uncovered a few places where Xilinx model does nonsensical things which don't match the documentation (and didn't match my own model either).  I've changed my model to match Xilinx model, but I'd really like to see what the whole thing does on real hw.

PR #1510, opened 2019-11-21
handle anonymous unions to fix #1080

### LINKED ISSUE #1080: python: SigBit data and offset attributes not available through python
The SigBit class is exposed through the python interface, and its members are `wire` and an anonymous union of `data` and `offset`. `py_wrap_generator.py` sees the enclosing {} around the union and treats it as its own namespace and therefore does not look for members within it, but the anonymous union members are part of the enclosing scope and can be treated like any other declared members for the purpose of that script. This works since the class members are accessed by set/get functions that reference the original C++ object.

I locally hacked py_wrap_generator.py to ignore the union enclosure and got the desired behavior with offset and data visible in SigBit objects, but it's not robustly written as I only have the one union in SigBit to test it against.

```
cd yosys/examples/python-api
python3 -i script.py
cell = design.top_module().selected_cells()[0]
sigbit = cell.connections_[ys.IdString('\\A')].to_sigbit_vector()[0]
sigbit.offset
```

## Expected behavior

offset int value printed

## Actual behavior

```
  File "<stdin>", line 1, in <module>
AttributeError: 'SigBit' object has no attribute 'offset'
```

## py_wrap_generator.py hack
		# for anonymous unions, ignore union enclosure by skipping start line and replacing end line with new line
		if 'union {' in line:
			j = i+1
			while j < len(source_text):
				union_line = source_text[j]
				if '};' in union_line:
					source_text[j] = '\n'
					break
				j += 1
			if j != len(source_text):
				i += 1
				continue

pumbor:
PR with suggested fix from #1080

### DIFF
--- a/misc/py_wrap_generator.py
+++ b/misc/py_wrap_generator.py
@@ -1935,6 +1935,19 @@ def parse_header(source):
 		line = source_text[i].replace("YOSYS_NAMESPACE_BEGIN", "                    namespace YOSYS_NAMESPACE{").replace("YOSYS_NAMESPACE_END","                    }")
 		ugly_line = unpretty_string(line)
 
+		# for anonymous unions, ignore union enclosure by skipping start line and replacing end line with new line
+		if 'union {' in line:
+			j = i+1
+			while j < len(source_text):
+				union_line = source_text[j]
+				if '};' in union_line:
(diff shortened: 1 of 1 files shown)


daveshah1:
cc @btut

Issue #1193, opened 2019-07-14
SystemVerilog frontend does not implement semantics of always_{ff,comb,latch}


whitequark:
## Steps to reproduce the issue

Testcase `x.sv`:
```sv
module test (input e, a, output reg b);
  always_comb
  	if (e)
  		b = a;
endmodule
```
Run as: `yosys x.sv -p proc -b ilang`.

## Expected behavior

Latches are not silently inferred in an `always_comb` block.

## Actual behavior

Latches are silently inferred in an `always_comb` block.

## Remarks

Per IEEE Std 1800-2009:

![Screenshot_20190714_143615](https://user-images.githubusercontent.com/54771/61185105-bc617000-a644-11e9-8dfd-1f6a7fce3348.png)

Clearly, Yosys does not perform such checks, since it does not produce a warning here. (Why does the spec even specify a warning here and not an error?)

I'll investigate writing a PR to handle SV `always_*` variants in the future.

PR #1511, opened 2019-11-21
sv: Error checking for always_comb, always_latch and always_ff

daveshah1:
This handles always_comb, always_latch and always_ff correctly. Sensitivity lists after always_(comb|latch) are now a syntax error as they should be; and an attribute is set on the process according to the type of always.

Checking of the types is done in `proc_dlatch`. Checking for correct use of `always_ff` here feels a bit iffy here; but the sequence inside `proc` means that doing it inside `proc_dff` seems tricky.

My understanding is that latches in always_comb; and vice versa; are only supposed to be a warning according to the standard. This PR implements what 99% of users would expect and throws an error in these cases, happy to change that if strict standards compliance is preferred.


whitequark:
Fixes #1193!

PR #1517, opened 2019-11-22
Add "opt_mem" pass

cliffordwolf:
With the most basic optimization: Remove a memory that has neither write port nor init data.

### DIFF
--- a/passes/opt/opt_mem.cc
+++ b/passes/opt/opt_mem.cc
@@ -1,0 +1,143 @@
+/*
+ *  yosys -- Yosys Open SYnthesis Suite
+ *
+ *  Copyright (C) 2012  Clifford Wolf <clifford@clifford.at>
+ *
+ *  Permission to use, copy, modify, and/or distribute this software for any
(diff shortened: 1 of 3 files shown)


passes/opt/opt_mem.cc
```diff
@@ -0,0 +1,143 @@
...
+		//   |---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|
+		log("\n");
+		log("    opt_mem [options] [selection]\n");
+		log("\n");
+		log("This pass performs various optimizations on memories in the design.\n");
+		log("\n");
+	}
+	void execute(std::vector<std::string> args, RTLIL::Design *design) YS_OVERRIDE
+	{
+		log_header(design, "Executing OPT_MEM pass (detect identical cells).\n");
```
daveshah1: nitpick - "detect identical cells" seems to be copied from another pass and incorrect

cliffordwolf: Oops. Fixed. Thanks.

PR #1518, opened 2019-11-22
[WIP] New 'clkpart' to {,un}partition design according to clock/enable

eddiehung:
Factors out this operation from the `abc` command, with the future goal of sharing it between `abc` and `abc9`, as well as for decomposition purposes.

Opening a draft PR to solicit feedback about this approach
TODO:
* ~use `submod` attribute to execute `submod` pass just once~
* ~on top of not doing anything if there is only one domain, do not call `submod` on the largest domain (leaving it inside the original)~
* Handle other `$_DFF*` and `$__DFF*` cells too


whitequark:
Can you elaborate on what happens with combinatorial cells after this pass? For example, let's say I have a `$and` cell sampled by two DFFs with different control sets. Where does it end up?

eddiehung:
> Can you elaborate on what happens with combinatorial cells after this pass? For example, let's say I have a `$and` cell sampled by two DFFs with different control sets. Where does it end up?

Excellent question! Currently, I believe it puts the `$and` into an arbitrary domain. My understanding is that ABC is only capable of technology mapping for one clock domain at a time, and primary inputs (as well as inputs from other domains) are assumed to arrive at time zero.
When Yosys gets a proper timing analyser in the future, it might be possible to do better and propagate these arrival times...

whitequark:
> Excellent question! Currently, I believe it would put the `$and` into an arbitrary domain.

Can you document this?

eddiehung:
Concerns about heavy-rewriting of the design were raised. Closing for now.

whitequark:
Can you summarize the concerns? I had similar ideas (actually I was wondering about using this pass as well) so it'd be useful to know why it's a no-go.

eddiehung:
@whitequark The background of this pass was for use as part of `abc9`. In order for `abc9` to perform correct sequential synthesis (involving reachability analysis of flop state in order to identify equivalent flops) then it must only operate on one clock domain at a time. For this particular use-case, @cliffordwolf raised issues that we were rewriting the design many times, forwards and backwards, for a task that shouldn't need it beyond prototyping. I still maintain that it is useful to have such a pass in the toolbox for others to use (perhaps you can share your use case for extra ammunition) but it's no longer on my critical path right now so I'm parking it.

whitequark:
The cxxrtl backend naturally works per-clock-domain in some regards. But I already decided I want it to handle processes, so this excluded using clkpart.

PR #1519, opened 2019-11-23
submod: several bugfixes

eddiehung:
Should solve a number of issues:
* Wire `c` is not detected as a primary output, and won't be made into a submod output port
```
module top(input a, output [1:0] b);
wire c;
(* submod="bar" *) sub s1(a, c);
assign b[0] = c;
endmodule
```
* Wire `b` is only partially used, but the entire bus will be made into a submodule output port, causing driver-driver conflict for `b[0]`:
```
module top(input a, output [1:0] b);
(* submod="bar" *) sub s1(a, b[1]);
assign b[0] = 1'b0;
endmodule
```
* `(* init = ... *)` is copied, not moved, into submodule causing `check -noinit` to give out warnings:
```
module top(input d, c, (* init = 1'b1 *) output reg q);
(* submod="bar" *) DFF s1(.D(d), .C(c), .Q(q));
endmodule

module DFF(input D, C, output Q);
parameter INIT = 1'b0;
endmodule
```
since top.q and top_bar.q would otherwise both have `(* init *)`

* Add `submod -hidden`:
```
    -hidden
        instead of creating submodule ports with public names, create ports with
        private names so that a subsequent 'flatten; clean' call will restore the
        original module with original public names.
```


eddiehung:
Should solve a number of issues:
* Wire `c` is not detected as a primary output, and won't be made into a submod output port
```
module top(input a, output [1:0] b);
wire c;
(* submod="bar" *) sub s1(a, c);
assign b[0] = c;
endmodule
```
* Wire `b` is only partially used, but the entire bus will be made into a submodule output port, causing driver-driver conflict for `b[0]`:
```
module top(input a, output [1:0] b);
(* submod="bar" *) sub s1(a, b[1]);
assign b[0] = 1'b0;
endmodule
```
* In designs similar to the one below, prevents `b` submodule port from being an inout, which I've seen to cause issues with `flatten` to trigger this error: https://github.com/YosysHQ/yosys/blob/c03b6a3e9cab9fc05b2d5b256676f5ddc6c2d763/passes/techmap/techmap.cc#L336-L346
but I've yet to come up with a minimal testcase yet...

eddiehung:
Note to self -- investigate behaviour surrounding unknown cell instantiations...

PR #1521, opened 2019-11-24
Adding support for Xilinx memory attribute 'block' in single port mode.

dh73:
The following PR is to open a discussion about supporting `ram_style` attributes in `memory_bram` match rules. This approach needs Dave's `memory_collect.cc` changes from PR #1501 .

Having the following module with `ram_style="block"` attribute defined:
```verilog
`default_nettype none
module memory #(parameter W=8, D=64)
   (input wire clk, wr_en,
    input wire [W-1:0]  wr_addr, rd_addr,
    input wire [W-1:0]  din,
    output reg [W-1:0]  dout);

   (*ram_style = "block"*)
   reg [W-1:0]          mem [D-1:0];

   always @(posedge clk) begin
      if(wr_en) mem[wr_addr] <= din;
      dout <= mem[rd_addr];
   end
endmodule // memory
```
Yosys is able to match to a RAMB18E1, even if the `min bits` and `min efficiency` are below the constraints:
```
=== memory ===

   Number of wires:                  9
   Number of wire bits:             72
   Number of public wires:           6
   Number of public wire bits:      34
     BUFG                            1
     RAMB18E1                        1

   Estimated number of LCs:          0
```

As @eddiehung and I discussed, the apporach is to have a new match rule `attribute` into `xc7_xcu_brams.txt` as follows:
```
match $__XILINX_RAMB18_SDP
  min bits 4096
  min efficiency 5
  shuffle_enable B
  make_transp
  attribute ram_style block
  or_next_if_better
endmatch
``` 
So, if the attribute in the RTL object match to the one defined in the match rule, we bypass the `min bits` and `min efficiency` checks. This is done just for `RAMB18E1`.


passes/memory/memory_bram.cc
```diff
@@ -403,6 +409,7 @@ bool replace_cell(Cell *cell, const rules_t &rules, const rules_t::bram_t &bram,
 	int mem_abits = cell->getParam("\\ABITS").as_int();
 	int mem_width = cell->getParam("\\WIDTH").as_int();
 	// int mem_offset = cell->getParam("\\OFFSET").as_int();
+        bool memory_attribute = cell->attributes["\\ram_style"].decode_string()=="block";
```
eddiehung: Is this, and a number of other lookups below, intended to be hard coded to `ram_style`/`block` rather than taking the value from the rules file?

techlibs/xilinx/xc7_xcu_brams.txt
```diff
@@ -85,6 +85,7 @@ match $__XILINX_RAMB18_SDP
   min efficiency 5
   shuffle_enable B
   make_transp
+  attribute ram_style block
```
eddiehung: My original suggestion would be to:
(a) add support for `attribute` and
(b) use it in a new set of match blocks (if it is indeed possible to have multiple match blocks for the same `bram` entry) as I would think that would be more flexible....

Take for example that you don't want to ignore `min_bits` or`efficiency` (or any other condition) but instead have different thresholds or conditions outright.... do you think that would be a good idea?

daveshah1:
This PR also needs to consider the case where an attribute is used to _not_ map a memory; e.g. to use LUTRAM/ROM or FFRAM instead of BRAM for timing or resource reasons

eddiehung:
> This PR also needs to consider the case where an attribute is used to _not_ map a memory; e.g. to use LUTRAM/ROM or FFRAM instead of BRAM for timing or resource reasons

Perhaps this could be achieved with an `attribute <name>` entry (no value) on the existing `match` to indicate that it should be considered if the attribute is _not_ present?

techlibs/xilinx/xc7_xcu_brams.txt
dh73: The thing is, `parse_match` function does not have a way to know if the match name is the same, but with different rules, it parses line by line starting from `match` keyword. If we put two match rules with same name, the memory will ended into a RAMB18E1 all the time. That's why I didn't used that approach.

There are more attributes than `block`, that can be applied to a memory, so I think the flexibility would be more into `memory_bram` pass instead of the match rules. For instance, attribute `block` forces Vivado to use RAMB18 iff memory words are greather than or equal to 2 (therefore, min bits and efficiency are ignored), so for this specific case we could use `min words N` (maybe more than 2 if we don't want to use Vivado thresholds). Quick way would be to copy and rename both `$__XILINX_RAMB18_SDP` match rule and Verilog techmap instantiation, (like ```$__XILINX_RAMB18_SDP_ATTR```) and define there the `min word` constraint, but we'll have redundant RAMB18E1 techmap. We could also manage that case in the `memory_bram` pass, but we need a mechanism to know every supported family attributes if we don't plan to make this Xilinx spcefic. Or, make `memory_bram` recognise different match rules for same memory.

passes/memory/memory_bram.cc
dh73: The commit I just pushed gets the attribute name from `$mem` cell and compares it to the one defined in the match rule, instead of fixing it to `block`.

passes/memory/memory_bram.cc
```diff
@@ -362,6 +368,20 @@ struct rules_t
 	}
 };
 
+bool check_memory_attribute(Cell *cell, std::string match)
+{
+	std::string memory_attribute = cell->attributes["\\ram_style"].decode_string();
```
eddiehung: Looks like this is still hardcoded to `ram_style`?

passes/memory/memory_bram.cc
```diff
@@ -997,6 +1018,7 @@ void handle_cell(Cell *cell, const rules_t &rules)
 	log("Processing %s.%s:\n", log_id(cell->module), log_id(cell));
 
 	bool cell_init = !SigSpec(cell->getParam("\\INIT")).is_fully_undef();
+	std::string cell_attribute = cell->attributes["\\ram_style"].decode_string();
```
eddiehung: And here

techlibs/xilinx/xc7_xcu_brams.txt
eddiehung: > The thing is, parse_match function does not have a way to know if the match name is the same, but with different rules, it parses line by line starting from match keyword. 

If I'm understanding correctly, then it's possible to have multiple `match` blocks with the same "name" (referring to the same `bram` entry) but with different conditions -- right? That's actually what I was hoping for, because you can have multiple entries:

```
match $__XILINX_RAMB18_SDP
  min bits 4096
  min efficiency 5
  shuffle_enable B
  make_transp
  attribute ram_style # No/empty value indicating this attribute must not exist
  or_next_if_better
endmatch

# This match block has no min bits/efficiency conditions, just ram_style="block"
match $__XILINX_RAMB18_SDP
  shuffle_enable B
  make_transp
  attribute ram_style block
  or_next_if_better
endmatch
```
Is that possible, or am I missing something?

> If we put two match rules with same name, the memory will ended into a RAMB18E1 all the time. That's why I didn't used that approach.

Not sure what you mean here -- can you elaborate please?

eddiehung: Resolved in 89be614

eddiehung:
@dh73 Looks like you're still missing this: https://github.com/YosysHQ/yosys/blob/5383e639e34a027600417152078d78131d184ece/techlibs/xilinx/xc7_xcu_brams.txt#L83-L97
in order to
(1) Use existing match rules for when `(* ram_style *)` doesn't exist
(2) Use new match rule (with no min/max conditions) when `(* ram_style = "block" *)` exists

whitequark:
IEEE 1364.1 also defines attributes `(* synthesis, ram_block [= "<optional>"] *)` and `(* synthesis, rom_block [= "<optional>"] *)`. IMO these should be also supported and aliased to `(* ram_style/rom_style = "block" *)`.

eddiehung:
> IEEE 1364.1 also defines attributes `(* synthesis, ram_block [= "<optional>"] *)` and `(* synthesis, rom_block [= "<optional>"] *)`. IMO these should be also supported and aliased to `(* ram_style/rom_style = "block" *)`, since they make designs more portable between families.

Apparently 1364.1-2002 has been withdrawn/superseded [https://standards.ieee.org/standard/1364_1-2002.html] by 1364.2005 which no longer has mention of `(ram|rom)_block`.  Also notable is the use of `logic_block` to force register usage.

Nonetheless I agree we should still support it.

Let me propose the grammar be changed to:

`attribute attr1=value1 [attrN=valueN]*` for the existence of attr=value1 OR attrN=valueN

and

`attribute !key1` to require the omission of key1

Thus:
```
attribute !ram_style
attribute !logic_block
```
would match `$mem` cells without `(* ram_style *)` AND without `(* logic_block *)`

and
```
attribute ram_block=1 ram_style=block
```
would match `$mem` cells with `(* ram_block[=1] *)` OR `(* ram_style="block" *)`

This grammar can be used to match `(* attr *)` since this comes up as `(* attr=1 *)` but cannot match `attr=<anything>`, which I don't think is necessary.

Thoughts?

whitequark:
> Apparently 1364.1-2002 has been withdrawn/superseded [[linky]](https://standards.ieee.org/standard/1364_1-2002.html) by 1364.2005 which no longer has mention of `(ram|rom)_block`.

Do you mean 62142-2005? That's the direct successor of 1364.1-2002 (it's not called 1364.1-2005 because of some weird co-publishing with IEC, but it's morally that), and essentially identical, including `(ram|rom)_block`. In general, 1364-YEAR is the base Verilog standard and 1364.1-YEAR is the synthesizable subset.

Of course, I have to mention that all of the 1364 series standards have been withdrawn, because from the point of view of IEEE, there is no more Verilog, only SystemVerilog. (Sort of like dinosaurs and birds.) There are two things here though: Yosys still tries hard to support 1364-only tools, and 1800 makes no attempt to standardize anything synthesis-related, i.e. 1364.1-2002 / 62142-2005 has no successor at all. This is why I think it makes sense to still consider 1364.1-2002 / 62142-2005.

> Let me propose the grammar be changed to:

Is the semantics of `attribute` you are proposing here essentially a product-of-sums? I.e. all top-level `attribute` statements must match, but within a single `attribute` statement any nonzero amount of terms must match. If so, that sounds fine to me.

eddiehung:
> Is the semantics of `attribute` you are proposing here essentially a product-of-sums? I.e. all top-level `attribute` statements must match, but within a single `attribute` statement any nonzero amount of terms must match. If so, that sounds fine to me.

Indeed. Each condition in a `match` block are treated as products (AND-ed together) so this just adds "sum" support for `attribute` conditions.

Also TODO: add a test or two

passes/memory/memory_bram.cc
```diff
@@ -326,7 +328,18 @@ struct rules_t
 				data.or_next_if_better = true;
 				continue;
 			}
+	
+			if (GetSize(tokens) == 3 && tokens[0] == "!attribute") {
```
eddiehung: This would mean you can't have a rule like so:
`attribute !ram_style ram_style=block`
which will allow either the omission of `ram_style`, or if it was set to the value `block`
Would this be a problem?

passes/memory/memory_bram.cc
```diff
@@ -1073,6 +1107,43 @@ void handle_cell(Cell *cell, const rules_t &rules)
 						i+1, log_id(bram.name), bram.variant);
 				goto next_match_rule;
 			}
+			
+			for (auto synattr: match.attr_syn) {
+				auto it = cell->attributes.find(synattr.first);
+				if(it != cell->attributes.end()) {
+					if(it->first == "\\logic_block" || it->first == "\\ram_style") {
```
eddiehung: Should we still hard code these? This seems a backward step from what we had before.

passes/memory/memory_bram.cc
```diff
@@ -1233,6 +1309,10 @@ struct MemoryBramPass : public Pass {
 		log("has a higher efficiency than the next match (and the one after that if\n");
 		log("the next also has 'or_next_if_better' set, and so forth).\n");
 		log("\n");
+		log("A match containing the condition '!attribute' followed by a name\n");
+                log("requires that the memory does not contains the given attribute name\n");
+                log("or that the attribute is not present.\n");
```
eddiehung: The two statements: "does not contains the given attribute name or that the attribute is not present." seem identical, no? Also, spacing is a little off here, and perhaps it would be better to put this description right after the `attribute` description above.

techlibs/xilinx/xc7_xcu_brams.txt
```diff
@@ -81,10 +81,18 @@ match $__XILINX_RAMB36_SDP
 endmatch
 
 match $__XILINX_RAMB18_SDP
-  min bits 4096
```
eddiehung: Isn't this change part of  #1533 and shouldn't belong here?

passes/memory/memory_collect.cc
```diff
@@ -218,6 +218,10 @@ Cell *handle_memory(Module *module, RTLIL::Memory *memory)
 	mem->setPort("\\RD_DATA", sig_rd_data);
 	mem->setPort("\\RD_EN", sig_rd_en);
 
+	// Copy attributes from RTLIL memory to $mem
```
eddiehung: This looks to be from #1501, which has already been merged into master. Do you know why it's appearing here again?

tests/arch/xilinx/memory_attributes.ys
```diff
@@ -0,0 +1,30 @@
+# RAMB18E1 Converted to registers
+read_verilog ../common/memory_attributes/ram_test_logic_block.v
+hierarchy -top ram_test_logic_block
+synth_xilinx -abc9 -flatten
```
eddiehung: No need to use `-abc9` here.

passes/memory/memory_bram.cc
```diff
@@ -1100,6 +1136,42 @@ void handle_cell(Cell *cell, const rules_t &rules)
...
+					auto it = cell->attributes.find(iter.first);
+
+					if (iter.second.empty()) {
+						log("    Rule for bram type %s is rejected: requirement 'attribute %s=\"%s\"' not met.\n",
+								log_id(match.name), iter.first.c_str(), iter.second.decode_string().c_str());
+						continue;
+					}
+
+					if (it != cell->attributes.end()) {
+						if (it->second == iter.second)
```
dh73: The PR #1501  copies the value of the attribute, if given, so this comparison can be made easily. In the case of,`synthesis, ram_block`, no value is assigned to the attribute, therefore, `memory_collect.cc` adds a value of 1, meaning the attribute is present. In that case, the comparison is wrong, since data types are different.

techlibs/xilinx/xc7_xcu_brams.txt
```diff
@@ -85,6 +88,9 @@ match $__XILINX_RAMB36_SDP
 endmatch
 
 match $__XILINX_RAMB18_SDP
+  attribute ram_style=block ram_block=1
+  attribute !ram_style
```
eddiehung: Can you double check this? My reading is that this condition is:
If ram_style == block OR ram_block == 1
AND ram_style attribute is not present # <<< Do the tests show this to be a conflict?
AND logic_block attribute is not present

tests/arch/xilinx/attributes_test.ys
```diff
@@ -0,0 +1,47 @@
...
+ 
+# Check that distributed memory without parameters is not modified
+design -reset
+read_verilog ../common/memory_attributes/attributes_test.v
+hierarchy -top distributed_ram
+synth_xilinx -top distributed_ram
+cd distributed_ram # Constrain all select calls below inside the top module
+select -assert-count 8 t:RAM32X1D
+ 
+# Set ram_style distributed to blockram memory; will be implemented as distributed
```
eddiehung: Am I right in reading that this doesn't test test that we 100% implement this as distributed RAMs, simply that it won't get implemented as block RAMs and thus falls through to the next level in the hierarchy which are LUTRAMs...

tests/arch/xilinx/attributes_test.ys
```diff
@@ -0,0 +1,47 @@
...
+design -reset
+read_verilog ../common/memory_attributes/attributes_test.v
+prep
```
eddiehung: Why is `prep` necessary here?

tests/arch/common/memory_attributes/attributes_test.v
```diff
@@ -0,0 +1,88 @@
+`default_nettype none
```
eddiehung: Why not re-use https://github.com/YosysHQ/yosys/blob/87e21b0122bd682db8aeffae3e1ac503c9cea2d2/tests/arch/common/blockram_params.v and use `setattr` to set attributes on wires, as you have done?

PR #1524, opened 2019-11-25
Gowin: add and test DFF init values

pepijndevos:
In the previous Gowin patch, it was found that DFF init behavior was not supported in hardware and removed. Upon closer inspection it was found that in fact the hardware initializes to the reset value. In the vendor tools, the following code
```verilog
module myDFF (output reg Q = 1, input CLK, D);
	always @(posedge CLK)
		Q <= D;
endmodule
```
Synthesizes to the following DFFS with the reset tied to ground.
```verilog
DFFS Q_0 (
	.Q(Q_c),
	.D(D_c),
	.CLK(CLK_c),
	.SET(GND)
);
```
This pull request adds the same behavior to all Gowin flip-flops that don't use a reset.
In case a reset is used, the wire init parameter is only removed if it is compatible with the reset value.

This PR also adds tests for all DFF types and their init values, and fixed a few typos in previously untested DFF types.

Finally, it fixes a small discrepancy in the help. A `-noiopads` option was mentioned, but not actually implemented. Now it is.

One thing I don't like about this PR is that the test code is basically a copy of `cell_sim.v` with different names. Directly using the cell library resulted in weird issues.

### DIFF
--- a/techlibs/gowin/cells_map.v
+++ b/techlibs/gowin/cells_map.v
@@ -1,82 +1,231 @@
+`default_nettype none
 //All DFF* have INIT, but the hardware is always initialised to the reset
 //value regardless. The parameter is ignored.
 
 // DFFN      D Flip-Flop with Negative-Edge Clock
-module  \$_DFF_N_ (input D, C, output Q); DFFN _TECHMAP_REPLACE_ (.D(D), .Q(Q), .CLK(C)); endmodule
+module  \$_DFF_N_ #(parameter _TECHMAP_WIREINIT_Q_ = 1'bx) (input D, C, output Q);
+	generate
+    if (_TECHMAP_WIREINIT_Q_ === 1'b1)
+			DFFNS _TECHMAP_REPLACE_ (.D(D), .Q(Q), .CLK(C), .SET(1'b0));
+    else
+			DFFN _TECHMAP_REPLACE_ (.D(D), .Q(Q), .CLK(C));
+  endgenerate
+	wire _TECHMAP_REMOVEINIT_Q_ = 1;
+endmodule
+
 // DFF       D Flip-Flop
-module  \$_DFF_P_ (input D, C, output Q); DFF _TECHMAP_REPLACE_ (.D(D), .Q(Q), .CLK(C)); endmodule
+module  \$_DFF_P_ #(parameter _TECHMAP_WIREINIT_Q_ = 1'bx) (input D, C, output Q);
+	generate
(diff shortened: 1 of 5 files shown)


techlibs/gowin/cells_map.v
```diff
@@ -1,133 +1,282 @@
...
+endmodule
+module	\$__DFFSE_NP0 #(parameter _TECHMAP_WIREINIT_Q_ = 1'bx) (input D, C, R, E, output Q);
+	DFFNRE _TECHMAP_REPLACE_ (.D(D), .Q(Q), .CLK(C), .RESET(R), .CE(E));
+	wire _TECHMAP_REMOVEINIT_Q_ = _TECHMAP_WIREINIT_Q_ !== 1'b1;
+endmodule
 
-// DFFS      D Flip-Flop with Synchronous Set
-module  \$__DFFS_PN1_ (input D, C, R, output Q); DFFS _TECHMAP_REPLACE_ (.D(D), .Q(Q), .CLK(C), .SET(!R)); endmodule
-module  \$__DFFS_PP1_ (input D, C, R, output Q); DFFS _TECHMAP_REPLACE_ (.D(D), .Q(Q), .CLK(C), .SET(R)); endmodule
+// DFFS			 D Flip-Flop with Synchronous Set
```
daveshah1: For the DFFs with synchronous set/reset; it would probably be better to map the set/reset back to logic (which can model it exactly, i.e. an OR or ANDNOT gate) if it doesn't match init - it's only async DFFs where the init value can't be handled.

pepijndevos: IMO it should be extremely rare that the init value is different from the reset value, but you are right. How do you model init with an OR or ANDNOT gate though?

daveshah1: You don't model init with the OR or ANDNOT gate - you remove the sync set/reset on the primitive, replace it with one that matches the init value and use an OR or ANDNOT gate on the D input with set/reset as appropriate

mwkmwkmwk: I'd just like to remind you that `dff2dffs` has a `-match-init` attribute now, which prevents creating such cells in the first place.

pepijndevos: Oh wow, that sounds much better.

pepijndevos: Okay, I added `-match-init` and adjusted the testcase. I left the `_TECHMAP_REMOVEINIT_Q_` on the sync reset flops for now, it seems like a good sanity check, maybe?

PR #1527, opened 2019-11-26
iopadmap: Refactor and fix tristate buffer mapping.

mwkmwkmwk:
The previous code for rerouting wires when inserting tristate buffers
was overcomplicated and didn't handle all cases correctly (in
particular, only cell connections were rewired — internal connections
were not).

### DIFF
--- a/passes/techmap/iopadmap.cc
+++ b/passes/techmap/iopadmap.cc
@@ -180,33 +180,24 @@ struct IopadmapPass : public Pass {
 
 		for (auto module : design->selected_modules())
 		{
-			dict<IdString, pool<int>> skip_wires;
 			pool<SigBit> skip_wire_bits;
-			SigMap sigmap(module);
+			dict<Wire *, dict<int, pair<Cell *, IdString>>> rewrite_bits;
 
 			for (auto cell : module->cells())
 			for (auto port : cell->connections())
 				if (ignore.count(make_pair(cell->type, port.first)))
-					for (auto bit : sigmap(port.second))
+					for (auto bit : port.second)
 						skip_wire_bits.insert(bit);
 
 			if (!toutpad_celltype.empty() || !tinoutpad_celltype.empty())
 			{
-				dict<SigBit, pair<IdString, pool<IdString>>> tbuf_bits;
-				pool<pair<IdString, IdString>> norewrites;
-				SigMap rewrites;
+				dict<SigBit, Cell *> tbuf_bits;
 
 				for (auto cell : module->cells())
 					if (cell->type == ID($_TBUF_)) {
-						SigBit bit = sigmap(cell->getPort(ID::Y).as_bit());
-						tbuf_bits[bit].first = cell->name;
+						SigBit bit = cell->getPort(ID::Y).as_bit();
+						tbuf_bits[bit] = cell;
 					}
-
-				for (auto cell : module->cells())
-				for (auto port : cell->connections())
-				for (auto bit : sigmap(port.second))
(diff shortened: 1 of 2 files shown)


passes/techmap/iopadmap.cc
```diff
@@ -238,37 +224,16 @@ struct IopadmapPass : public Pass {
 							log("Mapping port %s.%s[%d] using %s.\n", log_id(module), log_id(wire), i, tinoutpad_celltype.c_str());
 
 							Cell *cell = module->addCell(NEW_ID, RTLIL::escape_id(tinoutpad_celltype));
-							Wire *owire = module->addWire(NEW_ID);
 
 							cell->setPort(RTLIL::escape_id(tinoutpad_portname), en_sig);
-							cell->setPort(RTLIL::escape_id(tinoutpad_portname2), owire);
+							cell->setPort(RTLIL::escape_id(tinoutpad_portname2), wire_bit);
```
eddiehung: Can you use better variable names than `portname[0-9]?` to help readability?

passes/techmap/iopadmap.cc
```diff
@@ -415,15 +334,47 @@ struct IopadmapPass : public Pass {
...
+					wire->port_input = false;
+					wire->port_output = false;
+				}
+			}
+
+			for (auto &it : rewrite_bits) {
+				RTLIL::Wire *wire = it.first;
+				RTLIL::Wire *new_wire = module->addWire(NEW_ID, wire);
+				module->swap_names(new_wire, wire);
+				wire->attributes.clear();
```
eddiehung: Is it desirable to remove all attributes on this wire? What if there was a user attribute?

passes/techmap/iopadmap.cc
```diff
@@ -415,15 +334,47 @@ struct IopadmapPass : public Pass {
...
+				for (int i = 0; i < wire->width; i++)
+				{
+					SigBit wire_bit(wire, i);
+					if (!it.second.count(i)) {
+						if (wire->port_output)
+							module->connect(SigSpec(new_wire, i), SigSpec(wire, i));
```
eddiehung: Is there a difference between `assign a = b` and `assign b = a`?

mwkmwkmwk: I don't know the exact semantics assumed by all our passes, but it definitely affects direction of BUFs in the `show` command.

passes/techmap/iopadmap.cc
mwkmwkmwk: Then the user attributes will now be on the new wire.  This dance is a roundabout way of creating a new plain anonymous wire and reconnecting all cells/connects from the original wire to it — since it turns out to be simpler to move the name and attributes to a new wire than to move connections, we do it this way.

passes/techmap/iopadmap.cc
mwkmwkmwk: Done.

passes/techmap/iopadmap.cc
eddiehung: Pardon me if I'm missing something, but all I see is `wire->attributes.clear()` and I'm not seeing where `new_wire` adopts those attributes...?

mwkmwkmwk: Right when it is created — `module->addWire(NEW_ID, wire)` copies the attributes of the original wire.

eddiehung: Ah-ha! Thanks for pointing that out.

whitequark:
Excellent work!

PR #1533, opened 2019-11-26
Adjust Xilinx xc7/xcu BRAM min bits threshold for RAMB18E1

dh73:
Based on [UG473](https://www.xilinx.com/support/documentation/user_guides/ug473_7Series_Memory_Resources.pdf) and memory analysis/behavior we have seen in benchmark system, the `min bits 4096` is too high for Yosys, to infer a RAMB18E1.

Vivado infers RAMB18E1 (TDP/SDP) if min bits are 1024. Adjusting this value significantly reduces LUTRAM usage in 4 designs, without affecting speed.


techlibs/xilinx/xc7_xcu_brams.txt
```diff
@@ -81,7 +81,7 @@ match $__XILINX_RAMB36_SDP
 endmatch
 
 match $__XILINX_RAMB18_SDP
```
eddiehung: Please cite link to thresholds here

eddiehung:
Do the RAMB36 thresholds need changing too?

dh73:
Probably yes, but RAMB36 needs also port configuration, not only min bits. I will check that one too.

dh73:
Regarding if whether or not RAMB36 needs an adjustment, the rules to satisfy that type of memory inference are already covered by the `xc7_xcu_brams.txt` file (see table below). Experiments doesn't show any advantage in modification of them.
image

With current settings, Yosys yields a very similar result as Vivado:
image

[This test](https://github.com/dh73/BRAM-test) are trivial, they are just to compare results between Vivado and Yosys using Vivado's synthesis as reference. I don't think they are needed (benchmarks also helped to improve this), but if I'm wrong, let me know so I can update the PR.

techlibs/xilinx/xc7_xcu_brams.txt
```diff
@@ -73,33 +73,36 @@ bram $__XILINX_RAMB18_TDP
...
 
 match $__XILINX_RAMB18_TDP
-  min bits 4096
+  min bits 1024
   min efficiency 5
   shuffle_enable B
   make_transp
 endmatch
 
+# [[CITE]] 7 Series FPGAs Memory Resources User Guide (UG473),
```
eddiehung: Can you move this to before the first `match` block, and state that the `min bits` value was taken from this document (with a link) please?

tests/arch/xilinx/memory_params.ys
```diff
@@ -0,0 +1,45 @@
...
+# Anything memory bits < 1024 -> LUTRAM
+design -reset
+read_verilog ../common/memory_params.v
+chparam -set ADDRESS_WIDTH 8 -set DATA_WIDTH 2 sync_ram_sdp
+synth_xilinx -top sync_ram_sdp
+cd sync_ram_sdp
+select -assert-count 0 t:RAMB18E1
+select -assert-count 4 t:RAM128X1D
+
+# More than 18K bits and addr <= 36: -> RAMB36E1
```
eddiehung: Double check `addr <= 36`?

PR #1545, opened 2019-12-03
Preserve SB_CARRY name and attributes when using $__ICE40_CARRY_WRAPPER

### LINKED ISSUE #1543: ice40_wrapcarry does not appear to respect (* keep *) attribute on SB_CARRY cells

I am attempting to do experiments with tapped delay lines in ice40. I am using (for example)  a delay element as in the attached file delay_element.v. Note that I have applied (* keep *) attributes to all the cells which I need to remain there throughout synthesis.

In yosys I run the script

```
 read_verilog delay_line.v
synth_ice40
stat

   Number of wires:                  9
   Number of wire bits:              9
   Number of public wires:           9
   Number of public wire bits:       9
     SB_DFFER                        1

```

As can be seen, the synthesis has optimised away the SB_CARRY (and hence also the SB_LUT4) that I explicitly need to be kept.

Having investigated in a little more detail, when the synth_ice40 command executes the ice40_opt pass, it is optimising away the $__ICE40_CARRY_WRAPPER cell (see int_4.v for netlist after ice40_opt) which was generated by ice40_wrapcarry (see the file int_3.v for netlist after ice40_wrapcarry). 

I think the fault lies in ice40_wrapcarry, as can be seen from int_3.v, there is no (* keep *) attribute on the $__ICE40_CARRY_WRAPPER  cell. I'm not familiar enough with yosys internals to know if this is actually the problem, though.

[delayline.zip](https://github.com/YosysHQ/yosys/files/3914011/delayline.zip)

## Expected behavior

synthesized netlist contains an SB_CARRY cell, which was marked as (* keep *)

## Actual behavior

synthesized netlist contains only SB_DFF** cell, SB_LUT4 and SB_CARRY have been optimised away.

eddiehung:
Fixes #1543 

The only thing that isn't preserved is the SB_LUT4's name or attributes... I don't think this is critical right now though.


whitequark:
What about the case where I have a `keep` attribute on the `SB_LUT4` cell but not `SB_CARRY`? I think it makes sense to merge the `keep` and `src` attribute at least. I think it's fine to copy all others from the `SB_CARRY` cell unconditionally. What do you think?

eddiehung:
> I think it makes sense to merge the `keep` and `src` attribute at least.
That sounds reasonable without having to add new `techmap` features, thanks for the suggestion! Incoming.

passes/pmgen/ice40_wrapcarry.cc
```diff
@@ -51,6 +51,7 @@ void create_ice40_wrapcarry(ice40_wrapcarry_pm &pm)
 	cell->setParam("\\LUT", st.lut->getParam("\\LUT_INIT"));
 
 	cell->attributes = std::move(st.carry->attributes);
+	cell->attributes.insert(st.lut->attributes.begin(), st.lut->attributes.end());
```
whitequark: I think this isn't quite right: this won't merge the `\src` attribute (instead it would replace its content with LUT's `\src`), nor the `\keep` attribute (something like `(* keep = 0 *)` is valid as currently recognized by Yosys and equivalent to no `\keep` attribute at all, so `(*keep*) SB_CARRY` + `(*keep=0*) SB_LUT` would result in wrong behavior).

I think we should have something like `AttrObject::merge` that takes care of attributes known to Yosys itself.

whitequark: Another place where `AttrObject::merge_attributes` would be useful is in the `opt_lut` pass.

eddiehung: > this won't merge the \src attribute (instead it would replace its content with LUT's \src)

I'm pretty sure the `insert()` semantic is not to overwrite map entries? Indeed, my test shows that the SB_CARRY's `answer = 42` is preserved:
https://github.com/YosysHQ/yosys/blob/a7e0cca48003dbb212326c37c0b8a5580c2e6d2c/tests/arch/ice40/wrapcarry.ys#L25-L50

Regarding `(* keep *)` I will treat this as a special case and OR the two results.

whitequark: > I'm pretty sure the `insert()` semantic is not to overwrite map entries?

Right, sorry, I forgot how maps work in STL and Yosys. But that has the same fundamental problem, which is that the value of `\src` is derived from just one cell instead of both of them.

eddiehung: > > I'm pretty sure the `insert()` semantic is not to overwrite map entries?
> 
> Right, sorry, I forgot how maps work in STL and Yosys. But that has the same fundamental problem, which is that the value of `\src` is derived from just one cell instead of both of them.

Yes, we can only pick one or the other and I've given `SB_CARRY` priority over the `SB_LUT4`. I can append the `src` value (separated by `|`)  but that breaks their usual meaning of indicating the progression one particular cell made.

To summarise, the `SB_CARRY` attributes have priority over `SB_LUT4` except in the case of `(* keep = 0 *)`. On unwrapping `SB_CARRY` gets all the attributes and `SB_LUT4` gets none. It sounds like this lossiness is still not acceptable, then I'd have to preserve both sets of attributes separately and create a custom `ice40_wrapcarry -unwrap` pass instead of using a `techmap` call. Does that sound better?

whitequark: > but that breaks their usual meaning of indicating the progression one particular cell made.

I didn't understand that Yosys `\src` attributes have this invariant. In that case I retract my request, and I think the changes here are fine. (I'm not sure if this is a *good* invariant, but this discussion shouldn't block this PR, and I'll raise it elsewhere in the future if I think it's important.)

eddiehung: Okay, I'm now preserving all attributes on both `SB_CARRY` and `SB_LUT4` and added a `ice40_wrapcarry -unwrap` option to undo that

PR #1554, opened 2019-12-05
synth_intel_alm: replacement flow for ALM-based Intel FPGAs.

ZirconiumX:
`synth_intel` has long been neglected, and I think it should be divided into two back-end sub-scripts to better represent architectural differences between LUT4+FF ("Logic Element") based FPGAs, such as the Cyclone IV and Cyclone 10LP; and LUT6+4xFF ("Adapatable Logic Module") based FPGAs, such as the Stratix II, Cyclone V and Cyclone 10GX. For now, this covers the ALM family; specifically the Cyclone V. 

The vendor LUT primitives (`{stratixii,cyclonev,cyclone10gx}_lcell_comb`) for the ALM architectures have been replaced with Yosys-internal `MISTRAL_ALMk` primitives; this allows Intel passes to operate on a single primitive family instead of multiple, deferring vendor primitive mapping until the end.

The equivalency tests have been ported from Efinix, barring `memory.ys` (I don't currently have any RAM models) and `latches.ys` (I don't have any latch models).

This is missing a lot of things such as DSP mapping, RAM initialisation, but it's a start.

### DIFF
--- a/techlibs/intel_alm/common/alm_map.v
+++ b/techlibs/intel_alm/common/alm_map.v
@@ -1,0 +1,56 @@
+module \$lut (A, Y);
+
+parameter WIDTH = 1;
+parameter LUT = 0;
+
+input [WIDTH-1:0] A;
+output Y;
+
+generate
+    if (WIDTH == 1) begin
+        generate
+            if (LUT == 2'b00) begin
+                assign Y = 1'b0;
+            end
+            else if (LUT == 2'b01) begin
+                MISTRAL_NOT _TECHMAP_REPLACE_(
+                    .A(A[0]), .Q(Y)
+                );
+            end
+            else if (LUT == 2'b10) begin
+                assign Y = A;
+            end
+            else if (LUT == 2'b11) begin
+                assign Y = 1'b1;
+            end
+        endgenerate
+    end else
+    if (WIDTH == 2) begin
+        MISTRAL_ALUT2 #(.LUT(LUT)) _TECHMAP_REPLACE_(
+            .A(A[0]), .B(A[1]), .Q(Y)
+        );
+    end else
+    if (WIDTH == 3) begin
+        MISTRAL_ALUT3 #(.LUT(LUT)) _TECHMAP_REPLACE_(
+            .A(A[0]), .B(A[1]), .C(A[2]), .Q(Y)
+        );
+    end else
+    if (WIDTH == 4) begin
+        MISTRAL_ALUT4 #(.LUT(LUT)) _TECHMAP_REPLACE_(
+            .A(A[0]), .B(A[1]), .C(A[2]), .D(A[3]), .Q(Y)
+        );
+    end else
+    if (WIDTH == 5) begin
+        MISTRAL_ALUT5 #(.LUT(LUT)) _TECHMAP_REPLACE_ (
+            .A(A[0]), .B(A[1]), .C(A[2]), .D(A[3]), .E(A[4]), .Q(Y)
+        );
+    end else
+    if (WIDTH == 6) begin
+        MISTRAL_ALUT6 #(.LUT(LUT)) _TECHMAP_REPLACE_ (
+            .A(A[0]), .B(A[1]), .C(A[2]), .D(A[3]), .E(A[4]), .F(A[5]), .Q(Y)
+        );
+    end else begin
+        wire _TECHMAP_FAIL_ = 1'b1;
+    end
+endgenerate
+endmodule
--- a/techlibs/intel_alm/common/alm_sim.v
+++ b/techlibs/intel_alm/common/alm_sim.v
@@ -1,0 +1,482 @@
+`default_nettype none
+
+(* abc9_lut=2, lib_whitebox *)
+module MISTRAL_ALUT6(input A, B, C, D, E, F, output Q);
+
+parameter [63:0] LUT = 64'h0000_0000_0000_0000;
+
+`ifdef cyclonev
+specify
+    (A => Q) = 602;
+    (B => Q) = 584;
+    (C => Q) = 510;
+    (D => Q) = 510;
+    (E => Q) = 339;
+    (F => Q) = 94;
+endspecify
+`endif
+`ifdef cyclone10gx
+specify
+    (A => Q) = 275;
+    (B => Q) = 272;
+    (C => Q) = 175;
+    (D => Q) = 165;
+    (E => Q) = 162;
+    (F => Q) = 53;
+endspecify
+`endif
+
+assign Q = LUT >> {F, E, D, C, B, A};
+
+endmodule
+
+
+(* abc9_lut=1, lib_whitebox *)
+module MISTRAL_ALUT5(input A, B, C, D, E, output Q);
+
+parameter [31:0] LUT = 32'h0000_0000;
+
+`ifdef cyclonev
+specify
+    (A => Q) = 584;
+    (B => Q) = 510;
+    (C => Q) = 510;
+    (D => Q) = 339;
+    (E => Q) = 94;
+endspecify
+`endif
+`ifdef cyclone10gx
+specify
+    (A => Q) = 272;
+    (B => Q) = 175;
+    (C => Q) = 165;
+    (D => Q) = 162;
+    (E => Q) = 53;
+endspecify
+`endif
+
+assign Q = LUT >> {E, D, C, B, A};
+
+endmodule
+
+
+(* abc9_lut=1, lib_whitebox *)
+module MISTRAL_ALUT4(input A, B, C, D, output Q);
+
+parameter [15:0] LUT = 16'h0000;
+
+`ifdef cyclonev
+specify
+    (A => Q) = 510;
+    (B => Q) = 510;
+    (C => Q) = 339;
+    (D => Q) = 94;
+endspecify
+`endif
+`ifdef cyclone10gx
+specify
+    (A => Q) = 175;
+    (B => Q) = 165;
+    (C => Q) = 162;
+    (D => Q) = 53;
+endspecify
+`endif
+
+assign Q = LUT >> {D, C, B, A};
+
+endmodule
+
+
+(* abc9_lut=1, lib_whitebox *)
+module MISTRAL_ALUT3(input A, B, C, output Q);
+
+parameter [7:0] LUT = 8'h00;
+
+`ifdef cyclonev
+specify
+    (A => Q) = 510;
+    (B => Q) = 339;
+    (C => Q) = 94;
+endspecify
+`endif
+`ifdef cyclone10gx
+specify
+    (A => Q) = 165;
+    (B => Q) = 162;
+    (C => Q) = 53;
+endspecify
+`endif
+
+assign Q = LUT >> {C, B, A};
+
+endmodule
+
+
+(* abc9_lut=1, lib_whitebox *)
+module MISTRAL_ALUT2(input A, B, output Q);
+
+parameter [3:0] LUT = 4'h0;
+
+`ifdef cyclonev
+specify
+    (A => Q) = 339;
+    (B => Q) = 94;
+endspecify
+`endif
+`ifdef cyclone10gx
+specify
+    (A => Q) = 162;
+    (B => Q) = 53;
+endspecify
+`endif
+
+assign Q = LUT >> {B, A};
+
+endmodule
+
+
+(* abc9_lut=1, lib_whitebox *)
+module MISTRAL_NOT(input A, output Q);
+
+`ifdef cyclonev
+specify
+    (A => Q) = 94;
+endspecify
+`endif
+`ifdef cyclone10gx
+specify
+    (A => Q) = 53;
+endspecify
+`endif
+
+assign Q = ~A;
+
+endmodule
+
+(* abc9_box, lib_whitebox *)
+module MISTRAL_ALUT_ARITH(input A, B, C, D0, D1, (* abc9_carry *) input CI, output SO, (* abc9_carry *) output CO);
+
+parameter LUT0 = 16'h0000;
+parameter LUT1 = 16'h0000;
+
+`ifdef cyclonev
+specify
+    (A => SO) = 1283;
+    (B => SO) = 1167;
+    (C => SO) = 866;
+    (D0 => SO) = 756;
+    (D1 => SO) = 756;
+    (CI => SO) = 355;
+    (A => CO) = 950;
+    (B => CO) = 1039;
+    (C => CO) = 820;
+    (D0 => CO) = 1006;
+    (D1 => CO) = 1006;
+    (CI => CO) = 23;
+endspecify
+`endif
+`ifdef cyclone10gx
+specify
+    (A => SO) = 644;
+    (B => SO) = 477;
+    (C => SO) = 416;
+    (D0 => SO) = 380;
+    (D1 => SO) = 431;
+    (CI => SO) = 276;
+    (A => CO) = 525;
+    (B => CO) = 433;
+    (C => CO) = 712;
+    (D0 => CO) = 653;
+    (D1 => CO) = 593;
+    (CI => CO) = 16;
+endspecify
+`endif
+
+wire q0, q1;
+
+assign q0 = LUT0 >> {D0, C, B, A};
+assign q1 = LUT1 >> {D1, C, B, A};
+
+assign {CO, SO} = q0 + !q1 + CI;
+
+endmodule
+
+
+/*
+// A, B, C0, C1, E0, E1, F0, F1: data inputs
+// CARRYIN: carry input
+// SHAREIN: shared-arithmetic input
+// CLK0, CLK1, CLK2: clock inputs
+//
+// COMB0, COMB1: combinational outputs
+// FF0, FF1, FF2, FF3: DFF outputs
+// SUM0, SUM1: adder outputs
+// CARRYOUT: carry output
+// SHAREOUT: shared-arithmetic output
+module MISTRAL_ALM(
+    input A, B, C0, C1, E0, E1, F0, F1, CARRYIN, SHAREIN, // LUT path
+    input CLK0, CLK1, CLK2, AC0, AC1,                     // FF path
+    output COMB0, COMB1, SUM0, SUM1, CARRYOUT, SHAREOUT,
+    output FF0, FF1, FF2, FF3
+);
+
+parameter LUT0 = 16'b0000;
+parameter LUT1 = 16'b0000;
+parameter LUT2 = 16'b0000;
+parameter LUT3 = 16'b0000;
+
+parameter INIT0 = 1'b0;
+parameter INIT1 = 1'b0;
+parameter INIT2 = 1'b0;
+parameter INIT3 = 1'b0;
+
+parameter C0_MUX = "C0";
+parameter C1_MUX = "C1";
+
+parameter F0_MUX = "VCC";
+parameter F1_MUX = "GND";
+
+parameter FEEDBACK0 = "FF0";
+parameter FEEDBACK1 = "FF2";
+
+parameter ADD_MUX = "LUT";
+
+parameter DFF01_DATA_MUX = "COMB";
+parameter DFF23_DATA_MUX = "COMB";
+
+parameter DFF0_CLK = "CLK0";
+parameter DFF1_CLK = "CLK0";
+parameter DFF2_CLK = "CLK0";
+parameter DFF3_CLK = "CLK0";
+
+parameter DFF0_AC  = "AC0";
+parameter DFF1_AC  = "AC0";
+parameter DFF2_AC  = "AC0";
+parameter DFF3_AC  = "AC0";
+
+// Feedback muxes from the flip-flop outputs.
+wire ff_feedback_mux0, ff_feedback_mux1;
+
+// C-input muxes which can be set to also use the F-input.
+wire c0_input_mux, c1_input_mux;
+
+// F-input muxes which can be set to a constant to allow LUT5 use.
+wire f0_input_mux, f1_input_mux;
+
+// Adder input muxes to select between shared-arithmetic mode and arithmetic mode.
+wire add0_input_mux, add1_input_mux;
+
+// Combinational-output muxes for LUT #1 and LUT #3
+wire lut1_comb_mux, lut3_comb_mux;
+
+// Sum-output muxes for LUT #1 and LUT #3
+wire lut1_sum_mux, lut3_sum_mux;
+
+// DFF data-input muxes
+wire dff01_data_mux, dff23_data_mux;
+
+// DFF clock selectors
+wire dff0_clk, dff1_clk, dff2_clk, dff3_clk;
+
+// DFF asynchronous-clear selectors
+wire dff0_ac, dff1_ac, dff2_ac, dff3_ac;
+
+// LUT, DFF and adder output wires for routing.
+wire lut0_out, lut1a_out, lut1b_out, lut2_out, lut3a_out, lut3b_out;
+wire dff0_out, dff1_out, dff2_out, dff3_out;
+wire add0_sum, add1_sum, add0_carry, add1_carry;
+
+generate
+    if (FEEDBACK0 === "FF0")
+        assign ff_feedback_mux0 = dff0_out;
+    else if (FEEDBACK0 === "FF1")
+        assign ff_feedback_mux0 = dff1_out;
+    else
+        $error("Invalid FEEDBACK0 setting!");
+
+    if (FEEDBACK1 == "FF2")
+        assign ff_feedback_mux1 = dff2_out;
+    else if (FEEDBACK1 == "FF3")
+        assign ff_feedback_mux1 = dff3_out;
+    else
+        $error("Invalid FEEDBACK1 setting!");
+
+    if (C0_MUX === "C0")
+        assign c0_input_mux = C0;
+    else if (C0_MUX === "F1")
+        assign c0_input_mux = F1;
+    else if (C0_MUX === "FEEDBACK1")
+        assign c0_input_mux = ff_feedback_mux1;
+    else
+        $error("Invalid C0_MUX setting!");
+
+    if (C1_MUX === "C1")
+        assign c1_input_mux = C1;
+    else if (C1_MUX === "F0")
+        assign c1_input_mux = F0;
+    else if (C1_MUX === "FEEDBACK0")
+        assign c1_input_mux = ff_feedback_mux0;
+    else
+        $error("Invalid C1_MUX setting!");
+
+    // F0 == VCC is LUT5
+    // F0 == F0 is LUT6
+    // F0 == FEEDBACK is unknown
+    if (F0_MUX === "VCC")
+        assign f0_input_mux = 1'b1;
+    else if (F0_MUX === "F0")
+        assign f0_input_mux = F0;
+    else if (F0_MUX === "FEEDBACK0")
+        assign f0_input_mux = ff_feedback_mux0;
+    else
+        $error("Invalid F0_MUX setting!");
+
+    // F1 == GND is LUT5
+    // F1 == F1 is LUT6
+    // F1 == FEEDBACK is unknown
+    if (F1_MUX === "GND")
+        assign f1_input_mux = 1'b0;
+    else if (F1_MUX === "F1")
+        assign f1_input_mux = F1;
+    else if (F1_MUX === "FEEDBACK1")
+        assign f1_input_mux = ff_feedback_mux1;
+    else
+        $error("Invalid F1_MUX setting!");
+
+    if (ADD_MUX === "LUT") begin
+        assign add0_input_mux = ~lut1_sum_mux;
+        assign add1_input_mux = ~lut3_sum_mux;
+    end else if (ADD_MUX === "SHARE") begin
+        assign add0_input_mux = SHAREIN;
+        assign add1_input_mux = lut1_comb_mux;
+    end else
+        $error("Invalid ADD_MUX setting!");
+
+    if (DFF01_DATA_MUX === "COMB")
+        assign dff01_data_mux = COMB0;
+    else if (DFF01_DATA_MUX === "SUM")
+        assign dff01_data_mux = SUM0;
+    else
+        $error("Invalid DFF01_DATA_MUX setting!");
+
+    if (DFF23_DATA_MUX === "COMB")
+        assign dff23_data_mux = COMB0;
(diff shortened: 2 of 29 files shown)


techlibs/intel/synth_intel_alm.cc
```diff
@@ -0,0 +1,208 @@
...
+			run("flatten");
+			run("tribuf -logic");
+			run("deminout");
+		}
+
+		if (check_label("coarse")) {
+			run("synth -run coarse");
+		}
+
+		if (!nolutram && check_label("map_lutram", "(skip if -nolutram)")) {
```
daveshah1: Usually BRAM is mapped before LUTRAM - as BRAM is preferred for large memories so mapped first, then anything that doesn't meet the criteria maps to LUTRAM

techlibs/intel/common/alm_sim.v
```diff
@@ -0,0 +1,67 @@
...
+assign      Q = A ? s1[1]       : s1[0];
+
+endmodule
+
+
+module MISTRAL_NOT(input A, output Q);
+
+assign      Q = ~A;
+
+endmodule
```
daveshah1: nitpick - a few files are missing final newlines

ZirconiumX:
So, as tempted as I am to just keep adding stuff to this, I'm going to leave this branch as-is and put my changes in a new branch to be PR'd after this.

@cliffordwolf Can you review this please?

ZirconiumX:
I'll also mention that the `MISTRAL_ALM*` cells also pass an equivalence test against the vendor `cyclonev_lcell_comb` cell model.

ZirconiumX:
At present, please don't merge this due to issues with Quartus importing this from Yosys.

- VQM apparently uses a separate or restricted parser that rejects valid Verilog statements.
- Verilog import works, but gives an ICE that nobody else has ever encountered before.
- EDIF import complains that the ground net is used multiple times.

Dealing with Quartus is pain, and I am looking forward to the day I never have to use it again.

techlibs/intel/common/alm_map.v
```diff
@@ -0,0 +1,43 @@
...
+parameter LUT = 0;
+
+input [WIDTH-1:0] A;
+output Y;
+
+generate
+    if (WIDTH == 1) begin
+        MISTRAL_NOT _TECHMAP_REPLACE_(
+            .A(A[0]), .Q(Y)
+        );
```
mwkmwkmwk: What if a `$lut` with a different function gets here somehow?  This shouldn't happen in a reasonable flow, but it would be good to at least have a check here (generate-if + `$error`)

techlibs/intel/common/dff_map.v
```diff
@@ -0,0 +1,102 @@
+`default_nettype none
+
+// D flip-flops
+module \$_DFF_P_ (input D, C, output Q);
+parameter _TECHMAP_WIREINIT_Q_ = 1'b0;
+parameter _TECHMAP_REMOVEINIT_Q_ = 1'b1;
```
mwkmwkmwk: These need to be wires, not parameters.

techlibs/intel/common/dff_map.v
```diff
@@ -0,0 +1,102 @@
...
+module \$_DFF_PP0_ (input D, C, R, output Q);
+parameter _TECHMAP_WIREINIT_Q_ = 1'b0;
+parameter _TECHMAP_REMOVEINIT_Q_ = 1'b1;
+MISTRAL_FF #(.INIT(_TECHMAP_WIREINIT_Q_)) _TECHMAP_REPLACE_(.D(D), .CLK(C), .ACn(~R), .ALD(1'b0), .AD(1'b0), .EN(1'b1), .Q(Q));
+endmodule
+
+module \$_DFF_PP1_ (input D, C, R, output Q);
+parameter _TECHMAP_WIREINIT_Q_ = 1'b0;
+parameter _TECHMAP_REMOVEINIT_Q_ = 1'b1;
+MISTRAL_FF #(.INIT(_TECHMAP_WIREINIT_Q_)) _TECHMAP_REPLACE_(.D(~D), .CLK(C), .ACn(~R), .ALD(1'b0), .AD(1'b0), .EN(1'b1), .Q(~Q));
```
mwkmwkmwk: You need to invert the INIT value here as well.  Also, the .Q assignment is wrong and results in a multiple driver error — the inverter needs to be plugged the other way around.

techlibs/intel/common/dff_sim.v
```diff
@@ -0,0 +1,27 @@
...
+// Q: data output
+//
+// Note: the DFFEAS primitive has an async set, but it's a lie; Intel emulate it in hardware with async clear.
+module MISTRAL_FF(input D, CLK, ACn, ALD, AD, EN, output reg Q);
+
+parameter [0:0] INIT = 1'b0;
+
+initial Q = INIT;
+
+always @(posedge CLK or negedge ACn or posedge ALD) begin
```
mwkmwkmwk: Unfortunately, this is a syntehsis-simulation mismatch — consider what happens when `AD` is 1, both `ACn` and `ALD` are asserted, and then `ACn` gets deasserted: your model won't see the edge, and will wrongly keep `Q` at 0, while the hardware will output 1.  We had a long talk about this internally, and came to the conclusion that the only reasonable workaround is to emulate the proper semantics using several FFs, latches and muxes — you can look at my `FDCPE` model for inspiration.

techlibs/intel/cyclonev/quartus_rename.v
```diff
@@ -0,0 +1,45 @@
...
+parameter LUT = 4'h0;
+
+cyclonev_lcell_comb #(.lut_mask({16{LUT}})) _TECHMAP_REPLACE_ (.dataa(A), .datab(B), .combout(Q));
+
+endmodule
+
+
+module MISTRAL_NOT(input A, output Q);
+
+assign Q = ~A;
```
mwkmwkmwk: This will result in a yosys-specific `$not` cell being left in the output — shouldn't you be actually mapping this to something?

techlibs/intel/synth_intel_alm.cc
```diff
@@ -0,0 +1,222 @@
...
+
+        if (!nobram && check_label("map_bram", "(skip if -nobram)")) {
+            run("memory_bram -rules +/intel/common/bram_m10k.txt", "(for Cyclone V)");
+            run("techmap -map +/intel/common/bram_m10k_map.v", "(for Cyclone V)");
+        }
+
+        if (!nolutram && check_label("map_lutram", "(skip if -nolutram)")) {
+			run("memory_bram -rules +/intel/common/lutram_mlab.txt", "(for Cyclone V)");
+			run("techmap -map +/intel/common/lutram_mlab_map.v", "(for Cyclone V)");
+		}
```
mwkmwkmwk: Please use tabs (mixed tabs/spaces here).

techlibs/intel/cyclonev/quartus_rename.v
ZirconiumX: I'll try using the Quartus `NOT` primitive and hope for the best.

techlibs/intel/common/dff_sim.v
ZirconiumX: Since I evidently stumbled into one of the corner cases of the Verilog standard, is this any better?
```verilog
module MISTRAL_FF(input D, CLK, ACn, ALD, AD, EN, output Q);

parameter [0:0] INIT = 1'b0;

reg Qsync, Qasync, Qlast;

initial begin
	Qsync = INIT;
	Qasync = INIT;
	Qlast = 0;
end

always @(posedge CLK) begin
	if (EN) begin
		Qsync <= D;
		Qlast <= 0;
	end
end

always @* begin
	if (!ACn) begin
		Qasync = 0;
		Qlast = 1;
	end
	else begin
		if (ALD) begin
			Qasync = AD;
			Qlast = 1;
		end
	end
end

assign Q = (Qlast == 1'b0) ? Qsync : Qasync;

endmodule
```

ZirconiumX: Actually this has both blocking and nonblocking assignments to `Qlast`. Which is wonderful.

mwkmwkmwk: It's still not synthesizable, but I think it at least simulates properly.  You could improve it to be synthesizable with something like this:

```
wire go_comb = !AC || ALD;
always @(posedge CLK, posedge go_comb)
  if (go_comb) Qlast <= 1;
  else Qlast <= 0;
```

ZirconiumX:
The build failed because apt couldn't update. That's...concerning.

techlibs/intel/common/dff_sim.v
ZirconiumX: Couldn't that be simplified to `Qlast <= go_comb`?

ZirconiumX: Or perhaps `Qlast <= (go_comb === 1)` to be more robust against `'x`?

mwkmwkmwk: No, because then it won't match the IEEE 1364.1 async set pattern.  Also, you're not allowed to use `===` in synthesizable code.

dh73:
> At present, please don't merge this due to issues with Quartus importing this from Yosys.
> 
> * VQM apparently uses a separate or restricted parser that rejects valid Verilog statements.
> * Verilog import works, but gives an ICE that nobody else has ever encountered before.
> * EDIF import complains that the ground net is used multiple times.
> 
> Dealing with Quartus is pain, and I am looking forward to the day I never have to use it again.

From *VQM Extractor and Language Functional Description, Version 2.0*: 

* Contrary to the Verilog standard, 1’b0, 1’b1 and 1’bz are the only constants supported in VQM. Hence, expressions like 1’b0101 will not be honored in VQM.
* VQM is originally a machine generated design file. The Quartus extractor therefore expects minimal user errors. Syntax errors will often result in a general indication of an error exists on a particular line without going into details on the nature of the error. The parser does not have a strong semantic checking phase.
* All identifiers must satisfy the requirement for identifiers in Verilog.
* Any Verilog language semantic requirement that governs the VQM subset must be
satisfied.

This excerpt from that document found in the QUIP latest release, satisifes what we discussed in IRC. I think is the only VQM documentation out there.

techlibs/intel_alm/synth_intel_alm.cc
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	void execute(std::vector<std::string> args, RTLIL::Design *design) YS_OVERRIDE
+	{
+		string run_from, run_to;
+		clear_flags();
+
+		size_t argidx;
+		for (argidx = 1; argidx < args.size(); argidx++) {
+			if (args[argidx] == "-family" && argidx + 1 < args.size()) {
```
mwkmwkmwk: This option is not mentioned in the help message.

techlibs/intel_alm/synth_intel_alm.cc
```diff
@@ -0,0 +1,231 @@
...
+				continue;
+			}
+			break;
+		}
+		extra_args(args, argidx, design);
+
+		if (!design->full_selection())
+			log_cmd_error("This command only operates on fully selected designs!\n");
+
+		if (family_opt == "cyclonev") {
```
mwkmwkmwk: Suggestion: set `family_opt` and `bram_type` to `<family>` and `<bram_type>` if in help mode

techlibs/intel_alm/synth_intel_alm.cc
```diff
@@ -0,0 +1,231 @@
...
+			run(stringf("abc9 -lut +/intel_alm/%s/abc9_lut.lut -box +/intel_alm/%s/abc9_box.box -W 200", family_opt.c_str(), family_opt.c_str()));
+			run("hilomap -singleton -hicell __MISTRAL_VCC Q -locell __MISTRAL_GND Q");
+			run("techmap -map +/intel_alm/common/alm_map.v");
+			run("opt -fast");
+			run("autoname");
+			run("clean");
+		}
+
+		if (check_label("check")) {
+			//run("hierarchy -check");
```
mwkmwkmwk: Why is this commented out?

techlibs/intel_alm/synth_intel_alm.cc
```diff
@@ -0,0 +1,231 @@
...
+			run("stat");
+			run("check");
+		}
+
+		if (check_label("vqm")) {
+			if (!vout_file.empty() || help_mode) {
+				run(stringf("techmap -map +/intel_alm/%s/quartus_rename.v", family_opt.c_str()));
```
mwkmwkmwk: This will cause different synthesis results depending on whether edif/vqm output is enabled — perhaps use `design -save/-load`, or add another option?

techlibs/intel_alm/common/alm_map.v
```diff
@@ -0,0 +1,98 @@
...
+    end else
+    if (WIDTH == 6) begin
+        MISTRAL_ALUT6 #(.LUT(LUT)) _TECHMAP_REPLACE_ (
+            .A(A[0]), .B(A[1]), .C(A[2]), .D(A[3]), .E(A[4]), .F(A[5]), .Q(Y)
+        );
+    end else
+    if (WIDTH == 7) begin
+        wire q0, q1;
+
+        MISTRAL_ALUT6 #(.LUT(LUT[63:0])) lut0 (
```
mwkmwkmwk: This doesn't seem correct?  AFAIU, you should split to sub-LUTs based on `A[6]`, not `A[0]` (that or rearrange the LUT bits in a different and messy way)

techlibs/intel_alm/common/alm_map.v
```diff
@@ -0,0 +1,98 @@
...
+
+        MISTRAL_ALUT6 #(.LUT(LUT[191:128])) lut2 (
+            .A(A[2]), .B(A[3]), .C(A[4]), .D(A[5]), .E(A[6]), .F(A[7]), .Q(q2)
+        );
+
+        MISTRAL_ALUT6 #(.LUT(LUT[255:192])) lut3 (
+            .A(A[2]), .B(A[3]), .C(A[4]), .D(A[5]), .E(A[6]), .F(A[7]), .Q(q3)
+        );
+
+        MISTRAL_ALUT6 #(.LUT(64'b1111111100000000111100001111000011001100110011001010101010101010)) mux (
```
mwkmwkmwk: ditto

techlibs/intel_alm/common/alm_map.v
```diff
@@ -0,0 +1,98 @@
...
+            .A(q0), .B(q1), .C(q2), .D(q3), .E(A[1]), .F(A[0]), .Q(Y)
+        );
+    end else begin
+        wire _TECHMAP_FAIL_ = 1'b1;
+    end
+endgenerate
+endmodule
+
+module __MISTRAL_VCC(output Q); MISTRAL_ALUT2 #(.LUT(4'b1111)) _TECHMAP_REPLACE_ (.A(1), .B(1), .Q(Q)); endmodule
```
mwkmwkmwk: How is this supposed to work?  A VCC mapping that uses a const 1 itself doesn't seem like a good plan to me; perhaps just leave const 0/1 in the netlist without mapping them, and let the (hypothetical) nextpnr backend deal with it + add an appropriate mapping to `quartus_rename.v`?

techlibs/intel_alm/common/alm_sim.v
```diff
@@ -0,0 +1,383 @@
+`default_nettype none
+
+// This is just $lut but renamed.
+module lut (A, Y);
```
mwkmwkmwk: Consider renaming it to `__MISTRAL_LUT` or something like that, to avoid namespace pollution

techlibs/intel_alm/common/bram_m10k.txt
```diff
@@ -0,0 +1,33 @@
...
+    dbits 16   @D512x16
+    dbits 20   @D512x20
+    abits  8   @D256x32 @D256x40
+    dbits 32   @D256x32
+    dbits 40   @D256x40
+    groups 2
+    ports  1 1
+    wrmode 1 0
+    # read enable; write enable + byte enables (only for multiples of 8)
+    enable 1 1
```
mwkmwkmwk: This describes a blockram with only a single write enable (not byte enables) — is that correct? You can look at Xilinx `xc6s_brams.txt` for an example with byte enables.

techlibs/intel_alm/common/alm_map.v
ZirconiumX: This is a hack around the significantly *less* hypothetical Quartus VQM format. It's quite picky about where it allows integer constants: multi-bit constants outside of parameters are forbidden, but single-bit constants are allowed (which is why it's legal to set A and B here) as well as connections to a constant driver LUT.

This is the only way I could think of that gives me some level of control over write_verilog not outputting multi-bit constants.

mwkmwkmwk: Fair enough; since this is a Quartus-specific hack though, what about handling that in `quartus_rename.v` instead?

techlibs/intel_alm/common/alm_map.v
ZirconiumX: Thinking about this, I'm really not sure who is correct here, due to how ABC9 orders LUT inputs. I'd have to look at that.

techlibs/intel_alm/common/alm_sim.v
ZirconiumX: Since this primitive isn't actually synthesisable I'm in a pickle here. I'm pretty sure the "right shift it by the input" approach doesn't properly propagate X, but I'm not sure how best to implement it then.

techlibs/intel_alm/common/bram_m10k.txt
ZirconiumX: I'm trying not to run before I can walk; the `altsyncram` primitive is an utter mess and wacky definitions of what a byte is depending on how you configure the memory. For now, a single enable will do.

techlibs/intel_alm/synth_intel_alm.cc
ZirconiumX: I don't see much point in picking one primitive list over another, and being intentionally leaky like this makes it easier to write tests to additionally check that the right Quartus primitives are produced.

techlibs/intel_alm/synth_intel_alm.cc
ZirconiumX: `hierarchy -check` complains about `__MISTRAL_{VCC,GND}` not being defined despite having been mapped out by this point. I don't understand why.

mwkmwkmwk: Which test can I reproduce that with?

ZirconiumX:
Any of them if you uncomment it. It seems to fail whatever I try.

mwkmwkmwk:
@ZirconiumX the issue is that `hilomap` works on `whitebox` modules (and turns const drivers into `__MISTRAL_VCC/GND` cells eg. inside the mistral sim models, while `techmap` does not (and thus the cells are left over).  That should be partially solved by moving it to near `quartus_rename.v`

ZirconiumX:
@eddiehung While trying to migrate to `specify` blocks in 8f10480, I keep hitting the following error:
```
ERROR: Assert `box_module->attributes.count("\\abc9_box_id") || box_module->get_bool_attribute("\\abc9_flop")' failed in backends/aiger/xaiger.cc:319.
```
Where am I messing up?

mwkmwkmwk:
> @eddiehung While trying to migrate to `specify` blocks in [8f10480](https://github.com/YosysHQ/yosys/commit/8f10480a1700d2489e4ff322ad22632496e7b421), I keep hitting the following error:
> 
> ```
> ERROR: Assert `box_module->attributes.count("\\abc9_box_id") || box_module->get_bool_attribute("\\abc9_flop")' failed in backends/aiger/xaiger.cc:319.
> ```
> 
> Where am I messing up?

I can't help you here, but please do specify exact repro instructions (failing command) for things like this.

ZirconiumX:
Take something like `picorv32_large.v` from yosys-bench, and run `yosys -p "synth_intel_alm" picorv32_large.v`, at which point you get this:

```
2.22.7.2. Executing XAIGER backend.
ERROR: Assert `box_module->attributes.count("\\abc9_box_id") || box_module->get_bool_attribute("\\abc9_flop")' failed in backends/aiger/xaiger.cc:319.
```

techlibs/intel_alm/synth_intel_alm.cc
ZirconiumX: I added the `-quartus` option to produce a Quartus-compatible netlist.

backends/aiger/xaiger.cc
```diff
@@ -280,6 +280,10 @@ struct XAigerWriter
 				if (abc9_flop)
 					continue;
 			}
+			else {
```
mwkmwkmwk, line 283: This shouldn't be part of this PR

techlibs/intel_alm/common/arith_alm_map.v
```diff
@@ -0,0 +1,67 @@
...
+
+wire [Y_WIDTH-1:0] A_buf, B_buf;
+\$pos #(.A_SIGNED(A_SIGNED), .A_WIDTH(A_WIDTH), .Y_WIDTH(Y_WIDTH)) A_conv (.A(A), .Y(A_buf));
+\$pos #(.A_SIGNED(B_SIGNED), .A_WIDTH(B_WIDTH), .Y_WIDTH(Y_WIDTH)) B_conv (.A(B), .Y(B_buf));
+
+wire [Y_WIDTH-1:0] AA = A_buf;
+wire [Y_WIDTH-1:0] BB = BI ? ~B_buf : B_buf;
+wire [Y_WIDTH-1:0] BX = B_buf;
+wire [Y_WIDTH-1:0] C;
+wire [Y_WIDTH+1:0] ALM_CARRY;
```
mwkmwkmwk, line 27: `[Y_WIDTH:0]` should suffice

techlibs/intel_alm/common/arith_alm_map.v
```diff
@@ -0,0 +1,67 @@
...
+        .LUT1(16'b1100_0011_1100_0011), // Q = C ? B : ~B (LUT1's input to the adder is inverted)
+    ) alm_i (
+        .A(AA[i]), .B(BX[i]), .C(BI), .D0(1'b1), .D1(1'b1),
+        .CI(ALM_CARRY[i]),
+        .SO(Y[i]),
+        .CO(ALM_CARRY[i+1])
+    );
+
+    // ALM carry chain is not directly accessible, so calculate the carry through soft logic if really needed.
+    assign CO[i] = (AA[i] && BB[i]) || (C[i] && (AA[i] || BB[i]));
```
mwkmwkmwk, line 62: `C` doesn't appear to actually be assigned anywhere, I think it should be removed altogether and `CO` calculated from `Y` instead (except for the highest `CO` bit, which could be easier to obtain by adding another cell to the carry chain?)

techlibs/intel_alm/common/alm_sim.v
```diff
@@ -0,0 +1,482 @@
...
+MISTRAL_ALUT4 #(.LUT(LUT1)) lut1_sum  (.A(A), .B(B), .C(C0), .D(E0), .Q(lut1_sum_mux));
+
+// F1 on the Quartus diagram
+MISTRAL_ALUT4 #(.LUT(LUT2)) lut2 (.A(A), .B(B), .C(C1), .D(c0_input_mux), .Q(lut2_out));
+
+// F3 on the Quartus diagram
+MISTRAL_ALUT4 #(.LUT(LUT3)) lut3_comb (.A(A), .B(B), .C(C1), .D(c0_input_mux), .Q(lut3_comb_mux));
+MISTRAL_ALUT4 #(.LUT(LUT3)) lut3_sum  (.A(A), .B(B), .C(C1), .D(E1), .Q(lut3_sum_mux));
+
+MISTRAL_FF #(.INIT(INIT0)) dff0 (.D(dff01_data_mux), .CLK(dff0_clk), .ACn(dff0_ac), .Q(dff0_out));
```
mwkmwkmwk, line 449: The `ALD` and `EN` ports should be connected to the proper consts here

techlibs/intel_alm/cyclone10gx/quartus_rename.v
```diff
@@ -0,0 +1,68 @@
...
+module MISTRAL_ALUT_ARITH(input A, B, C, D0, D1, CI, output SO, CO);
+parameter LUT0 = 16'h0000;
+parameter LUT1 = 16'h0000;
+
+cyclone10gx_lcell_comb #(.lut_mask({16'h0, LUT1, 16'h0, LUT0}), .shared_arith("on")) _TECHMAP_REPLACE_ (.dataa(A), .datab(B), .datac(C), .datad(D0), .dataf(D1), .cin(CI), .sumout(SO), .cout(CO));
+
+endmodule
+
+
+module __MISTRAL_VCC(output Q);
```
mwkmwkmwk, line 57: this is a duplicate of the rules already in the common file

techlibs/intel_alm/common/alm_sim.v
ZirconiumX: This is part of a rough ALM simulation; it's not complete and is presently commented out, especially since nothing currently emits it. (I had it in mind as a target for `intel_almpack` but I don't know if people even care about that)

techlibs/intel_alm/common/arith_alm_map.v
ZirconiumX: Turns out I forgot a small but important part of the ECP5 copypasta.

techlibs/intel_alm/common/arith_alm_map.v
```diff
@@ -23,8 +23,8 @@ wire [Y_WIDTH-1:0] A_buf, B_buf;
 wire [Y_WIDTH-1:0] AA = A_buf;
 wire [Y_WIDTH-1:0] BB = BI ? ~B_buf : B_buf;
 wire [Y_WIDTH-1:0] BX = B_buf;
-wire [Y_WIDTH-1:0] C;
-wire [Y_WIDTH+1:0] ALM_CARRY;
+wire [Y_WIDTH-1:0] C = {CO, CI};
+wire [Y_WIDTH:0] ALM_CARRY;
 
 assign C[0] = CI;
```
mwkmwkmwk, line 29: This is now a double assignment

techlibs/intel_alm/common/arith_alm_map.v
```diff
@@ -0,0 +1,67 @@
...
+        .LUT1(16'b1100_0011_1100_0011), // Q = C ? B : ~B (LUT1's input to the adder is inverted)
+    ) alm_i (
+        .A(AA[i]), .B(BX[i]), .C(BI), .D0(1'b1), .D1(1'b1),
+        .CI(ALM_CARRY[i]),
+        .SO(Y[i]),
+        .CO(ALM_CARRY[i+1])
+    );
+
+    // ALM carry chain is not directly accessible, so calculate the carry through soft logic if really needed.
+    assign CO[i] = (AA[i] && BB[i]) || (C[i] && (AA[i] || BB[i]));
```
mwkmwkmwk, line 62: I'd suggest: `assign CO[i] = (AA[i] && BB[i]) || ((Y[i] ^ AA[i] ^ BB[i]) && (AA[i] || BB[i]));` and nuke `C` entirely

techlibs/intel_alm/common/dff_map.v
```diff
@@ -0,0 +1,124 @@
+`default_nettype none
+
+// D flip-flops
+module \$_DFF_P_ (input D, C, output Q);
+parameter _TECHMAP_WIREINIT_Q_ = 1'b0;
+if (_TECHMAP_WIREINIT_Q_ === 1'b0 || _TECHMAP_WIREINIT_Q_ === 1'bx) begin
```
mwkmwkmwk, line 6: `if (_TECHMAP_WIREINIT_Q_ !== 1'b1)` would be shorter

techlibs/intel_alm/common/dff_map.v
```diff
@@ -0,0 +1,124 @@
...
+    wire _TECHMAP_REMOVEINIT_Q_ = 1'b1;
+    MISTRAL_FF _TECHMAP_REPLACE_(.DATAIN(D), .CLK(C), .ACLR(1'b1), .ENA(1'b1), .SCLR(1'b0), .SLOAD(1'b0), .SDATA(1'b0), .Q(Q));
+end
```
mwkmwkmwk, line 9: `else $error(...)`?

ZirconiumX: My thinking here is that `check` will warn on unprocessed init values, which should be a bit more useful?

mwkmwkmwk: Yeah, but then you'll have a *warning* and a completely mis-synthed circuit (you're actually discarding the flop entirely if the init value doesn't match).  Much better to do a hard error in this case.

techlibs/intel_alm/synth_intel_alm.cc
```diff
@@ -0,0 +1,240 @@
...
+		}
+
+		if (check_label("map_ffram")) {
+			run("memory_map");
+			run("opt -full");
+		}
+
+		if (check_label("map_ffs")) {
+			run("dff2dffe -direct-match $_DFF_*");
+			run("zinit -all");
```
mwkmwkmwk, line 203: Remove the `-all` — it's actively harmful here by forcing `0` intialization for otherwise-uninitialized flops with async set, preventing you from implementing them.

techlibs/intel_alm/common/dff_map.v
```diff
@@ -0,0 +1,124 @@
...
+    wire _TECHMAP_REMOVEINIT_Q_ = 1'b1;
+    MISTRAL_FF _TECHMAP_REPLACE_(.DATAIN(D), .CLK(C), .ACLR(1'b1), .ENA(1'b1), .SCLR(1'b0), .SLOAD(1'b0), .SDATA(1'b0), .Q(Q));
+end else $error("Unsupported flop: $_DFF_P_ with INIT=1");
+endmodule
+
+module \$_DFF_N_ (input D, C, output Q);
+parameter _TECHMAP_WIREINIT_Q_ = 1'b0;
+if (_TECHMAP_WIREINIT_Q_ !== 1'b1) begin
+    wire _TECHMAP_REMOVEINIT_Q_ = 1'b1;
+    MISTRAL_FF _TECHMAP_REPLACE_(.D(D), .CLK(~C), .ACLR(1'b1), .ENA(1'b1), .SCLR(1'b0), .SLOAD(1'b0), .SDATA(1'b0), .Q(Q));
```
mwkmwkmwk, line 16: `.DATAIN(D)`

PR #1557, opened 2019-12-06
xilinx: Add xilinx_dffopt pass

mwkmwkmwk:


techlibs/xilinx/xilinx_dffopt.cc
```diff
@@ -0,0 +1,344 @@
+/*
+ *  yosys -- Yosys Open SYnthesis Suite
+ *
+ *  Copyright (C) 2012  Clifford Wolf <clifford@clifford.at>
```
mithro: FYI This copyright looks wrong?

techlibs/xilinx/xilinx_dffopt.cc
```diff
@@ -0,0 +1,344 @@
...
+			for (auto wire : module->wires())
+				if (wire->port_output || wire->port_input)
+					for (int i = 0; i < GetSize(wire); i++)
+						bit_uses[sigmap(SigBit(wire, i))]++;
+
+			// Iterate through FFs.
+			for (auto cell : module->selected_cells())
+			{
+				bool has_s = false, has_r = false;
+				if (cell->type.in(ID(FDCE), ID(FDPE), ID(FDCPE))) {
```
eddiehung: What about the `FD*_1` variants?

techlibs/xilinx/xilinx_dffopt.cc
```diff
@@ -0,0 +1,344 @@
...
+					has_r = true;
+					has_s = true;
+				} else {
+					// Not a FF.
+					continue;
+				}
+
+				// Don't bother if D has more than one use.
+				SigBit sig_D = sigmap(cell->getPort(ID(D)));
+				if (bit_uses[sig_D] > 2)
```
eddiehung: Did you consider using the pattern matcher for this pass? It would appear that it would abstract away/re-use many of the things you're doing here

techlibs/xilinx/xilinx_dffopt.cc
```diff
@@ -0,0 +1,344 @@
...
+					final_lut = lut_d_post_ce;
+					log("  Merging CE LUT for %s/%s (%d -> %d)\n", log_id(cell), log_id(sig_Q.wire), GetSize(lut_d.second), GetSize(final_lut.second));
+				} else {
+					// Nothing to do here.
+					continue;
+				}
+
+				// Okay, we're doing it.  Unmap ports.
+				if (worthy_post_r) {
+					cell->unsetParam(ID(IS_R_INVERTED));
```
eddiehung: Consider that not all `FD*` cells (such as `FD_1` ones) have `IS_*_INVERTED` params

techlibs/xilinx/xilinx_dffopt.cc
```diff
@@ -0,0 +1,344 @@
...
+					cell->setPort(ID(R), Const(0, 1));
+				}
+				if (has_s && (worthy_post_r || worthy_post_s)) {
+					cell->unsetParam(ID(IS_S_INVERTED));
+					cell->setPort(ID(S), Const(0, 1));
+				}
+				cell->setPort(ID(CE), Const(1, 1));
+				cell->unsetParam(ID(IS_D_INVERTED));
+
+				// Create the new LUT.
```
eddiehung: Is it worth moving/inheriting any useful attributes (e.g. `src`) from the original cell?

mwkmwkmwk:
> Can I suggest a test for when `CE`, `S` and `R` are emulated? And maybe also a `CE` and `R` are both emulated? Thanks!

Sure, will add tests.

techlibs/xilinx/xilinx_dffopt.cc
mwkmwkmwk: Good point, will add them as well.

techlibs/xilinx/xilinx_dffopt.cc
mwkmwkmwk: Correct, but I don't see how it matters — this pass doesn't treat lack of such parameter as an error, and it only ever removes such parameters — never adds or changes them.

techlibs/xilinx/xilinx_dffopt.cc
mwkmwkmwk: I suppose I could try propagating the `src` attribute, but it's not trivial — there are, in fact, several source LUTs being merged here, so the attributes would somehow have to be merged as well.

techlibs/xilinx/xilinx_dffopt.cc
mwkmwkmwk: Yes, and I came to the conclusion that it would not help — the pattern being matched here is quite trivial (LUT inputs to FF), and most boilerplate code here would be present in a pmgen version as well.

techlibs/xilinx/xilinx_dffopt.cc
mwkmwkmwk: Copied all attributes from the original cell; added rejecting LUTs and FFs with keep attribute.

techlibs/xilinx/xilinx_dffopt.cc
mwkmwkmwk: Done.

techlibs/xilinx/xilinx_dffopt.cc
eddiehung: Righty. I missed that you were just unsetting, and that unsetting a non-existent param does not throw any exceptions.

mwkmwkmwk:
> Oh wait. Still need some tests SVP.

Correct (I haven't forgotten, just ran out of timeslice after doing the first few review fixes).

mwkmwkmwk:
Added tests.

techlibs/xilinx/xilinx_dffopt.cc
```diff
@@ -0,0 +1,350 @@
...
+	XilinxDffOptPass() : Pass("xilinx_dffopt", "Xilinx: optimize FF control signal usage") { }
+	void help() YS_OVERRIDE
+	{
+		//   |---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|
+		log("\n");
+		log("    xilinx_dffopt [options] [selection]\n");
+		log("\n");
+		log("Converts hardware clock enable and set/reset signals on FFs to emulation\n");
+		log("using LUTs, if doing so would improve area.\n");
+		log("\n");
```
eddiehung: Perhaps worth re-emphasising here that it operates on technology-mapped Xilinx LUT* and FD* cells only (as opposed to `$lut` or `$_DFF_*` cells?)

techlibs/xilinx/xilinx_dffopt.cc
```diff
@@ -0,0 +1,350 @@
...
+	}
+	std::vector<int> idx_sel;
+	for (auto bit : select.second) {
+		int idx = -1;
+		for (int i = 0; i < GetSize(result.second); i++)
+			if (result.second[i] == bit)
+				idx = i;
+		if (idx == -1) {
+			idx = GetSize(result.second);
+			result.second.push_back(bit);
```
eddiehung: Another late thought I had: this function adds control signals to the last input of a LUT, right? The last input of a LUT is also the fastest one, so would you be slowing down the timing of all the existing `D` path in favour of these control? If so, would it make more sense to put it at the beginning since these control inputs are pretty slow to begin with?

mwkmwkmwk: I don't think it matters at this point — the P&R tool permutes LUT inputs anyhow, so if our arrangement is an obstacle to meeting timing, it'll be automatically fixed.

techlibs/xilinx/xilinx_dffopt.cc
mwkmwkmwk: Done.

techlibs/xilinx/xilinx_dffopt.cc
eddiehung: Not sure I agree with your statement entirely. My view is that post-techmapping we should take care in minimising timing impact of any physical optimisations that are made. For one, on the Yosys roadmap is for it to gain static timing analysis capabilities, and it would be unnecessary for this analysis to have to do LUT input permutations. Secondly, LUT input permutations may be done as late as at the routing stage, in which case any decisions made prior to this could be made based on less-than-optimal arrangement.

What do you think?

mwkmwkmwk: I don't consider any of these effects to be very important (any mistakes we make about timing will probably be minor compared to the interconnect effect), but I'll do it — the change doesn't seem too hard.

Issue #1556, opened 2019-12-06
tests: efinix: mux16 case in mux.ys fails due to wrong number of LUT4 on macOS


kowalewskijan:
**Environment:**
- macOS Sierra (10.12.6)
- clang (8.1.0)
- iverilog (7f95abc)
- Yosys (7dece795)

Yosys built using official instructions from Readme.

**Test run on macOS:**
```
cd tests/arch/efinix
bash run_test.sh
```
**Error log:**
```
Running mux.ys..
ERROR: Assertion failed: selection contains 11 elements instead of the asserted 12: t:EFX_LUT4
Selection contains:
mux16/$abc$653$auto$blifparse.cc:498:parse_blif$654
mux16/$abc$653$auto$blifparse.cc:498:parse_blif$655
mux16/$abc$653$auto$blifparse.cc:498:parse_blif$656
mux16/$abc$653$auto$blifparse.cc:498:parse_blif$657
mux16/$abc$653$auto$blifparse.cc:498:parse_blif$658
mux16/$abc$653$auto$blifparse.cc:498:parse_blif$659
mux16/$abc$653$auto$blifparse.cc:498:parse_blif$660
mux16/$abc$653$auto$blifparse.cc:498:parse_blif$661
mux16/$abc$653$auto$blifparse.cc:498:parse_blif$662
mux16/$abc$653$auto$blifparse.cc:498:parse_blif$663
mux16/$abc$653$auto$blifparse.cc:498:parse_blif$664
make: *** [run-mux.ys] Error 1
```

I also ran the test on Linux and the error did not occur.
In attached blif files you can see that indeed there is a difference in number of LUTs.

[blif_files.zip](https://github.com/YosysHQ/yosys/files/3931483/blif_files.zip)

mmicko:
Thank you for reporting this issue. Was able to reproduce it on macOS machine.

eddiehung:
Just in case it gets lost, there was some discussion in https://github.com/YosysHQ/yosys/pull/1559#issuecomment-565581875 on the root cause:

> After much debugging with @mmicko it turns out the destruction order of STL containers in macOS is different to gcc/clang/VS: [apple/foundationdb#1862](https://github.com/apple/foundationdb/issues/1862)
> 
> Thus containers with IdString's in are destructed in a different order on macOS, and thus get recycled in a different order. And when other containers sort on the IdString index, we then get inconsistent behaviour.
> 
[...]

PR #1559, opened 2019-12-07
Fix for non-deterministic test

mmicko:
There seams to be non-deterministic behavior. Output blif that yosys generates contains always same number of elements:

Extracted 49 gates and 69 wires to a netlist network with 20 inputs and 1 outputs.

But order of those is not deterministic, so abc produce different number of $lut  cells, when we just run mux16 then it is 11, when we run others before we have 12.
Strange is that with clang and gcc on Linux and msvc on Windows we have same problem (12 luts), but on clang 10 on macOS (11 luts) it works fine even others are executed before.

This is fix for #1556

We also have same issue appearing on xilinx pull request for same mux test.

### DIFF
--- a/tests/arch/efinix/mux.ys
+++ b/tests/arch/efinix/mux.ys
@@ -36,6 +36,6 @@ proc
 equiv_opt -assert -map +/efinix/cells_sim.v synth_efinix # equivalency check
 design -load postopt # load the post-opt design (otherwise equiv_opt loads the pre-opt design)
 cd mux16 # Constrain all select calls below inside the top module
-select -assert-min 11 t:EFX_LUT4
+select -assert-max 12 t:EFX_LUT4
 
 select -assert-none t:EFX_LUT4 %% t:* %D
--- a/tests/arch/efinix/mux.ys
+++ b/tests/arch/efinix/mux.ys
@@ -36,6 +36,6 @@ proc
 equiv_opt -assert -map +/efinix/cells_sim.v synth_efinix # equivalency check
 design -load postopt # load the post-opt design (otherwise equiv_opt loads the pre-opt design)
 cd mux16 # Constrain all select calls below inside the top module
-select -assert-count 12 t:EFX_LUT4
+select -assert-min 11 t:EFX_LUT4
 
 select -assert-none t:EFX_LUT4 %% t:* %D


tests/arch/efinix/mux.ys
```diff
@@ -36,6 +36,6 @@ proc
 equiv_opt -assert -map +/efinix/cells_sim.v synth_efinix # equivalency check
 design -load postopt # load the post-opt design (otherwise equiv_opt loads the pre-opt design)
 cd mux16 # Constrain all select calls below inside the top module
-select -assert-count 12 t:EFX_LUT4
+select -assert-min 11 t:EFX_LUT4
```
eddiehung: Wouldn't it be better to do `-assert-max`?

mmicko: Was thinking more :
select -assert-min 11 t:EFX_LUT4
select -assert-max 12 t:EFX_LUT4
That will be then just covering two possible cases

eddiehung:
Please hold on merging this. I'm investigating why there is a divergence between platforms...

eddiehung:
After much debugging with @mmicko it turns out the destruction order of STL containers in macOS is different to gcc/clang/VS: https://github.com/apple/foundationdb/issues/1862

Thus containers with IdString's in are destructed in a different order on macOS, and thus get recycled in a different order. And when other containers sort on the IdString index, we then get inconsistent behaviour.

Is there any good reason why we should be recycling 32b indices in the first place? https://github.com/YosysHQ/yosys/blob/9ab1feeaf11adb6b675ac4034e246cb137d07db9/kernel/rtlil.h#L82

Alternatives include sorting the recycle list (at some cost): https://github.com/YosysHQ/yosys/blob/9ab1feeaf11adb6b675ac4034e246cb137d07db9/kernel/rtlil.h#L80

mmicko:
Just to confirm that  after change from :
#undef YOSYS_NO_IDS_REFCNT
to:
#define YOSYS_NO_IDS_REFCNT 
So with recycling not being done, it start to behave consistently on all platforms, and also got some result if I run just mux16 or all tests from mux.ys (and that was giving different results before as well)

eddiehung:
Potential solution: keep free list in sorted order all the time (i.e. use a heap)

tests/arch/efinix/mux.ys
nakengelhardt: Use -assert-max 12 (as discussed)

mmicko: Updated

PR #1566, opened 2019-12-11
[WIP] proc_dlatch: discard \init on bits that get transformed into direct connections

eddiehung:
Fixes #1443

### DIFF
--- a/passes/proc/proc_dlatch.cc
+++ b/passes/proc/proc_dlatch.cc
@@ -393,6 +393,15 @@ void proc_dlatch(proc_dlatch_db_t &db, RTLIL::Process *proc)
 		else
 			log("No latch inferred for signal `%s.%s' from process `%s.%s'.\n",
 					db.module->name.c_str(), log_signal(lhs), db.module->name.c_str(), proc->name.c_str());
+		auto it = chunk.wire->attributes.find(ID(init));
+		if (it != chunk.wire->attributes.end())
+			for (int i = offset; i < offset+chunk.width; i++)
+				if (it->second[i] != State::Sx) {
+					log_warning("Discarding 'init' attribute on signal `%s.%s'.\n",
+						db.module->name.c_str(), log_signal(lhs[i]));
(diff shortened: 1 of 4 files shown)


eddiehung:
Actually, I'm not sure this works. This change discards `\init` signals on direct connections:
```
module top(input i, output reg o);
initial o = 1'b0;
always @* o = i;
endmodule
```

but what about situations like:
```
module top(input clk, i, output reg o, p);
initial p = 1'b0;
always @(posedge clk) o <= i;
always @* p = o;
endmodule
```
where the `\init` of `p` would otherwise be used to initialise the FF on `o`?

And what about situations like:
```
module top(input clk, i, output reg o, p);
initial o = 1'b0;
initial p = 1'b0;
always @(posedge clk) o <= i;
always @* p = i;
endmodule
```
where the `\init` values are consistent?

cliffordwolf:
> but what about situations like:
> where the \init of p would otherwise be used to initialise the FF on o?

```
module top(input clk, i, output reg o, p);
initial p = 1'b0;
always @(posedge clk) o <= i;
always @* p = o;
endmodule
```

I don't think this is "legal" synthesizable verilog code anyway.

> And what about situations like:
where the \init values are consistent (is there any legal way to get here?)

```
module top(input clk, i, output reg o, p);
initial o = 1'b0;
initial p = 1'b0;
always @(posedge clk) o <= i;
always @* p = i;
endmodule
```

Same for this afaict.

eddiehung:
> ```
> module top(input clk, i, output reg o, p);
> initial p = 1'b0;
> always @(posedge clk) o <= i;
> always @* p = o;
> endmodule
> ```
> 
> 
> I don't think this is "legal" synthesizable verilog code anyway.

I seem to recall that I minimised this from a larger testcase... but it's been so long that I can't remember now. It may not be intentionally illegal, but some other optimisations could leave this as an intermediate stage.

Since so much time as has passed, I think there's little value in pursuing this any more so I'm going to close it.

PR #1568, opened 2019-12-11
zinit: fixes for $_DFF_[NP][NP][01]_and $adff cells with init = 1'b1

eddiehung:
* Fixes `zinit` to handle coarse and fine-grained async flops
* Disables support for `$dffsr` and `$_DFFSR_[NP][NP][NP]_`
* `(* init *)` attributes only erased when consumed by `zinit`

### DIFF
--- a/passes/techmap/zinit.cc
+++ b/passes/techmap/zinit.cc
@@ -143,7 +143,7 @@ struct ZinitPass : public Pass {
 				/*if (cell->type.in(ID($_DFFSR_NNN_), ID($_DFFSR_NNP_), ID($_DFFSR_NPN_), ID($_DFFSR_NPP_),
 							ID($_DFFSR_PNN_), ID($_DFFSR_PNP_), ID($_DFFSR_PPN_), ID($_DFFSR_PPP_)))
 				{
-					// TODO: I think I need a $_DFFRS_* cell where R has priority over S...
+					// TODO: I think I need a $_DFFRS_* cell where S has priority over R...
 					std::swap(cell->connections_.at(ID(R)), cell->connections_.at(ID(S)));
 				}
 				else*/ if (cell->type.in(ID($_DFF_NN0_), ID($_DFF_NN1_), ID($_DFF_NP0_), ID($_DFF_NP1_),
--- a/passes/techmap/zinit.cc
+++ b/passes/techmap/zinit.cc
@@ -141,9 +141,10 @@ struct ZinitPass : public Pass {
 				cell->setPort(ID::Q, initwire);
 			}
 
-			for (auto &it : initbits)
-				if (donebits.count(it.first) == 0)
-					log_error("Failed to handle init bit %s = %s.\n", log_signal(it.first), log_signal(it.second));
+			if (!design->selected_whole_module(module))
+				for (auto &it : initbits)
+					if (donebits.count(it.first) == 0)
+						log_error("Failed to handle init bit %s = %s.\n", log_signal(it.first), log_signal(it.second));
 		}
 	}
 } ZinitPass;
--- a/passes/techmap/zinit.cc
+++ b/passes/techmap/zinit.cc
@@ -117,7 +117,7 @@ struct ZinitPass : public Pass {
 						const auto &d = initbits.at(sig_q[i]);
 						initval.bits.push_back(d.first);
 						const auto &b = d.second;
-						b.wire->attributes.at(ID::init)[b.offset];
+						b.wire->attributes.at(ID::init)[b.offset] = State::Sx;
 					} else
 						initval.bits.push_back(all_mode ? State::S0 : State::Sx);
 				}
@@ -126,11 +126,11 @@ struct ZinitPass : public Pass {
 				initwire->attributes[ID::init] = initval;
 
 				for (int i = 0; i < GetSize(initwire); i++)
-					if (initval.bits.at(i) == State::S1)
+					if (initval[i] == State::S1)
 					{
 						sig_d[i] = module->NotGate(NEW_ID, sig_d[i]);
 						module->addNotGate(NEW_ID, SigSpec(initwire, i), sig_q[i]);
-						initwire->attributes[ID::init].bits.at(i) = State::S0;
+						initwire->attributes[ID::init][i] = State::S0;
 					}
 					else
 					{
@@ -145,8 +145,9 @@ struct ZinitPass : public Pass {
 
 				if (cell->type == ID($adff)) {
 					auto val = cell->getParam(ID::ARST_VALUE);
-					for (auto &b : val)
-						b = (b == State::S1 ? State::S0 : State::S1);
+					for (int i = 0; i < GetSize(initwire); i++)
+						if (initval[i] == State::S1)
+							val[i] = (val[i] == State::S1 ? State::S0 : State::S1);
 					cell->setParam(ID::ARST_VALUE, std::move(val));
 				}
 				else if (cell->type.in(ID($_DFF_NN0_), ID($_DFF_NN1_), ID($_DFF_NP0_), ID($_DFF_NP1_),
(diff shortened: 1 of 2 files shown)


passes/techmap/zinit.cc
```diff
@@ -139,11 +139,26 @@ struct ZinitPass : public Pass {
 
 				cell->setPort(ID(D), sig_d);
 				cell->setPort(ID(Q), initwire);
+
+				/*if (cell->type.in(ID($_DFFSR_NNN_), ID($_DFFSR_NNP_), ID($_DFFSR_NPN_), ID($_DFFSR_NPP_),
+							ID($_DFFSR_PNN_), ID($_DFFSR_PNP_), ID($_DFFSR_PPN_), ID($_DFFSR_PPP_)))
+				{
+					// TODO: I think I need a $_DFFRS_* cell where R has priority over S...
```
mwkmwkmwk: You mean "where S has priority over R" (R over S is the standard pretty much everywhere, including our own cells)

eddiehung: Oops right you are. Let me change that. I had assumed that the `SR` moniker meant that `S` > `R`

passes/techmap/zinit.cc
```diff
@@ -139,11 +139,26 @@ struct ZinitPass : public Pass {
...
+					std::string t = cell->type.str();
+					t[8] = (t[8] == '0' ? '1' : '0');
+					cell->type = t;
+				}
 			}
 
-			for (auto &it : initbits)
-				if (donebits.count(it.first) == 0)
-					log_error("Failed to handle init bit %s = %s.\n", log_signal(it.first), log_signal(it.second));
+			if (!design->selected_whole_module(module))
```
mwkmwkmwk: Why this if?  Why are we only reporting failure to handle init for partial selections?

mwkmwkmwk:
The fix for `$_DFF_[NP][NP][01]_` cells LGTM.  However, this pass is still rather buggy:

- `$adff` and `$dffsr` handled wrong
- if an unsupported storage cell is present, the `init` attribute for it will be needlessly nuked
- latches are not supported (which coupled with the above results in always nuking their init value)

If you don't want to fix this issues now, could you at least add a TODO for them?

passes/techmap/zinit.cc
eddiehung: I'm not sure why this pass was so aggressive in firing an error if ANY `(* init *)` entry was left behind, but rather than thinking much about it I stuck it behind an `if` so that it won't throw any errors when operating on entire modules, to keep forward momentum.

eddiehung:
> * a

Some excellent points. I didn't notice that it blew away all `(* init *)`-s .. that's bad. Very bad. But explains why it errors out if it can't handle some. A re-think of this pass is necessary.... (or to chuck it away)

nakengelhardt:
@eddiehung Are you still planning on working on this? If yes, can you please assign it to yourself? Thanks!

passes/techmap/zinit.cc
```diff
@@ -139,11 +139,26 @@ struct ZinitPass : public Pass {
...
 				cell->setPort(ID(Q), initwire);
+
+				/*if (cell->type.in(ID($_DFFSR_NNN_), ID($_DFFSR_NNP_), ID($_DFFSR_NPN_), ID($_DFFSR_NPP_),
+							ID($_DFFSR_PNN_), ID($_DFFSR_PNP_), ID($_DFFSR_PPN_), ID($_DFFSR_PPP_)))
+				{
+					// TODO: I think I need a $_DFFRS_* cell where S has priority over R...
+					std::swap(cell->connections_.at(ID(R)), cell->connections_.at(ID(S)));
+				}
+				else*/ if (cell->type.in(ID($_DFF_NN0_), ID($_DFF_NN1_), ID($_DFF_NP0_), ID($_DFF_NP1_),
+							ID($_DFF_PN0_), ID($_DFF_PN1_), ID($_DFF_PP0_), ID($_DFF_PP1_)))
```
cliffordwolf: LGTM

passes/techmap/zinit.cc
```diff
@@ -139,11 +139,26 @@ struct ZinitPass : public Pass {
 
 				cell->setPort(ID(D), sig_d);
 				cell->setPort(ID(Q), initwire);
+
+				/*if (cell->type.in(ID($_DFFSR_NNN_), ID($_DFFSR_NNP_), ID($_DFFSR_NPN_), ID($_DFFSR_NPP_),
+							ID($_DFFSR_PNN_), ID($_DFFSR_PNP_), ID($_DFFSR_PPN_), ID($_DFFSR_PPP_)))
```
cliffordwolf: Maybe we just want to remove those cell types from `dff_types` for now?

passes/techmap/zinit.cc
cliffordwolf: I think this should fail if we did not handle all init bits. When `zinit` is used then it's usually part of a flow where the later stages ignore init attributes and just assume everything is zero initialized. So not failing here will result in an incorrect output netlist for the purpose of whatever flow this is part of.

eddiehung: I've noticed that there are a lot of `\init` attributes leftover, even after a full `synth_*` run. We will need to eliminate them all otherwise this pass won't be usable at all.

eddiehung: Okay, I've elected to only erase the `(* init *)` bits that were actually consumed by a supported cell.

passes/techmap/zinit.cc
```diff
@@ -113,8 +114,10 @@ struct ZinitPass : public Pass {
 
 				for (int i = 0; i < GetSize(sig_q); i++) {
 					if (initbits.count(sig_q[i])) {
-						initval.bits.push_back(initbits.at(sig_q[i]));
-						donebits.insert(sig_q[i]);
+						const auto &d = initbits.at(sig_q[i]);
+						initval.bits.push_back(d.first);
+						const auto &b = d.second;
+						b.wire->attributes.at(ID::init)[b.offset];
```
mwkmwkmwk, line 120: shouldn't this line be... doing... something?

passes/techmap/zinit.cc
```diff
@@ -139,11 +142,21 @@ struct ZinitPass : public Pass {
...
 				cell->setPort(ID::D, sig_d);
 				cell->setPort(ID::Q, initwire);
-			}
 
-			for (auto &it : initbits)
-				if (donebits.count(it.first) == 0)
-					log_error("Failed to handle init bit %s = %s.\n", log_signal(it.first), log_signal(it.second));
+				if (cell->type == ID($adff)) {
+					auto val = cell->getParam(ID::ARST_VALUE);
+					for (auto &b : val)
```
mwkmwkmwk, line 148: You should only invert bits that actually correspond to State::S1 in initval.  Also note that this doesn't consider the State::Sx possibility in the reset value.

passes/techmap/zinit.cc
```diff
@@ -139,11 +142,21 @@ struct ZinitPass : public Pass {
...
+						b = (b == State::S1 ? State::S0 : State::S1);
+					cell->setParam(ID::ARST_VALUE, std::move(val));
+				}
+				else if (cell->type.in(ID($_DFF_NN0_), ID($_DFF_NN1_), ID($_DFF_NP0_), ID($_DFF_NP1_),
+							ID($_DFF_PN0_), ID($_DFF_PN1_), ID($_DFF_PP0_), ID($_DFF_PP1_)))
+				{
+					std::string t = cell->type.str();
```
mwkmwkmwk, line 155: ditto — verify that a negation is actually necessary

tests/techmap/zinit.ys
```diff
@@ -0,0 +1,23 @@
+read_verilog -icells <<EOT
```
mwkmwkmwk, line 1: This test would be much more useful if it actually verified the init=0 case as well.

eddiehung:
Thanks for the detailed review @mwkmwkmwk! I think I've resolved all your concerns now.

PR #1572, opened 2019-12-12
add a command to read/modify scratchpad contents

nakengelhardt:


passes/cmds/scratchpad.cc
```diff
@@ -0,0 +1,86 @@
...
+USING_YOSYS_NAMESPACE
+PRIVATE_NAMESPACE_BEGIN
+
+struct ScratchpadPass : public Pass {
+	ScratchpadPass() : Pass("scratchpad", "get/set values in the scratchpad") { }
+	void help() YS_OVERRIDE
+	{
+		//   |---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|
+		log("\n");
+		log("    scratchpad [ -get id | -set id val | -unset id | -copy id1 id2 ]\n");
```
eddiehung: For consistency, please adopt the same format for help messages as other passes, e.g. https://github.com/YosysHQ/yosys/blob/2666482282421bb54213ba01054111eadc401373/passes/techmap/attrmvcp.cc#L30-L54

passes/cmds/scratchpad.cc
```diff
@@ -0,0 +1,86 @@
+/*
+ *  yosys -- Yosys Open SYnthesis Suite
+ *
+ *  Copyright (C) 2012  Clifford Wolf <clifford@clifford.at>
```
eddiehung: Add yourself here!

passes/cmds/scratchpad.cc
```diff
@@ -30,14 +31,18 @@ struct ScratchpadPass : public Pass {
...
 		//   |---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|
 		log("\n");
-		log("    scratchpad [ -get id | -set id val | -unset id | -copy id1 id2 ]\n");
+		log("    scratchpad [options]\n");
 		log("\n");
 		log("This pass allows to read and modify values from the scratchpad of the current\n");
 		log("design. Options:\n");
 		log("    -get <identifier>\n");
+		log("        print the value saved in the scratchpad under the given identifier\n");
 		log("    -set <identifier> <value>\n");
```
eddiehung: Looks like you're still missing newlines in between the options, and a full stop at the end of the description

nakengelhardt: Oh, right. Starting with a lower case letter but ending with a period is kind of a counterintuitive style choice...

eddiehung:
Another suggestion I've just thought of would be to have `scratchpad -assert <key> <value>` option which would be useful for testing.  What do you think?

nakengelhardt:
Sure. I've added `-assert-set` and `-assert-unset` for good measure, now the test can be a `.ys` file.

passes/cmds/scratchpad.cc
```diff
@@ -0,0 +1,130 @@
...
+					log_error("Assertion failed: scratchpad entry '%s' is not defined\n", identifier.c_str());
+				continue;
+			}
+			if (args[argidx] == "-assert-unset" && argidx+1 < args.size()) {
+				string identifier = args[++argidx];
+				if (design->scratchpad.count(identifier) > 0)
+					log_error("Assertion failed: scratchpad entry '%s' is defined\n", identifier.c_str());
+				continue;
+			}
+			log("Unrecognized argument: %s\n", args[argidx].c_str());
```
eddiehung: It is typical to fall through to the `extra_args()` like so:
https://github.com/YosysHQ/yosys/blob/a73f96594f4688afc85098b485ef7788e79f5c33/passes/hierarchy/hierarchy.cc#L749-L751
as it produces a nice error message point to exactly what argument was failing and where it was in the command.

PR #1574, opened 2019-12-13
xilinx: add LUTRAM rules for RAM32M, RAM64M

### LINKED ISSUE #1460: synth_xilinx infers RAM32X1D instead of RAM32M

The following code:

```
module register_file(
    input wire clk,
    input wire write_enable,
    input wire [63:0] write_data,
    input wire [4:0] write_reg,
    input wire [4:0] read1_reg,
    input wire [4:0] read2_reg,
    input wire [4:0] read3_reg,
    output reg [63:0] read1_data,
    output reg [63:0] read2_data,
    output reg [63:0] read3_data
    );

    reg [63:0] registers[0:31];

    always @(posedge clk) begin
      if (write_enable == 1'b1) begin
        registers[write_reg] <= write_data;
      end
    end

    always @(all) begin
      read1_data <= registers[read1_reg];
      read2_data <= registers[read2_reg];
      read3_data <= registers[read3_reg];
    end
endmodule
```

When built with:

```
yosys -p "read_verilog register_file.v; synth_xilinx"
```

Consumes 192 RAM32X1D (dual-port 32 x 1-bit RAM) instances:

```
=== register_file ===

   Number of wires:                 12
   Number of wire bits:            280
   Number of public wires:          10
   Number of public wire bits:     278
   Number of cells:                193
     BUFG                            1
     RAM32X1D                      192
```

It seems like this could also use 32 RAM32M (quad-port 32 x 2-bit RAM) instances. If I'm reading it right, we currently use 384 LUTs, and with RAM32M instances we'd use 128 LUTs, 3x less.

eddiehung:
Where `RAM32M` and `RAM64M` are used in a 1 write port, 3 read port form (which is still an area win over using `RAM*X1D` variants since the write port is amortised over 3 instead of 1) since Yosys can't currently support a combined read/write port.

Fixes #1460


techlibs/xilinx/cells_sim.v
```diff
@@ -1185,10 +1185,10 @@ module RAM64M (
   output DOB,
   output DOC,
   output DOD,
-  input [4:0] ADDRA,
-  input [4:0] ADDRB,
-  input [4:0] ADDRC,
-  input [4:0] ADDRD,
+  input [5:0] ADDRA,
```
eddiehung: @mwkmwkmwk can you double check this for me? (does `RAM64M8` need the same treatment?)

mwkmwkmwk: Oops.  Yes, it does.

techlibs/xilinx/lutrams.txt
```diff
@@ -1,4 +1,17 @@
 
+bram $__XILINX_RAM16X1D
```
mwkmwkmwk: Why this rule?  `RAM16X1D` is a LUT4 architecture leftover, and is not natively supported by LUT6 architectures — it will just be converted to `RAM32X1D`.

PR #1576, opened 2019-12-13
opt_merge: discard \init of '$' cells with 'Q' port when merging

eddiehung:


passes/opt/opt_merge.cc
```diff
@@ -222,7 +222,8 @@ struct OptMergeWorker
 			return true;
 		}
 
-		if (cell1->type.begins_with("$") && conn1.count(ID(Q)) != 0) {
+		if (conn1.count(ID(Q)) != 0 && (cell1->type.begins_with("$dff") || cell1->type.begins_with("$dlatch") || cell1->type.in("$adff", "$sr", "$ff") ||
+					cell1->type.begins_with("$_DFF") || cell1->type.begins_with("$_DLATCH"))) {
```
mwkmwkmwk: There's also `$_FF_` and `$_SR_*`

passes/opt/opt_merge.cc
```diff
@@ -222,7 +222,8 @@ struct OptMergeWorker
 			return true;
 		}
 
-		if (cell1->type.begins_with("$") && conn1.count(ID(Q)) != 0) {
```
mwkmwkmwk: Awww, that was a perfect "featherless biped"-style definition :(

passes/opt/opt_merge.cc
eddiehung: In a future state element refactor, I propose that we add a `bool cell_ff(IdString type)` function (amongst others) to: https://github.com/YosysHQ/yosys/blob/319cba70d37eafdb8bbd3ddf6a0f9c238d53d0c2/kernel/celltypes.h#L34

whitequark: See also: https://github.com/YosysHQ/yosys/pull/1562/files#diff-94b3e6687aa5c05fbb80cfdfba068adcR173-R198

nakengelhardt:
@eddiehung I think there was no problem with this PR but it was forgotten over the holidays. Can you fix the conflict and merge this?

eddiehung:
> @eddiehung I think there was no problem with this PR but it was forgotten over the holidays. Can you fix the conflict and merge this?

Done. Sorry for mixing in some unrelated changes which I've backed out -- they were the cause of the conflict in the first place.

PR #1581, opened 2019-12-17
Fix sim for assignments with lhs<rhs size

### LINKED ISSUE #1565: sim pass assertion failure on unflattened design with unused port

To reproduce, create the following Verilog file,

```verilog
module check(clk, rst, value, out);
	input	wire	clk, rst, value;
	output	wire	out;

	submod m1(.clk(clk), .rst(rst), .value(value), .out(out), .count());
endmodule

module submod(clk, rst, value, out, count);
	input	wire	clk, rst, value;
	output	wire	out;
	output	wire	[7:0]	count;

	always @(posedge clk)
	if (rst)
		count <= 0;
	else
		count <= count + 1;

	assign	out = value;
endmodule
```

Together with the Yosys script,

```text
read_verilog -sv check.v
prep -top check
sim -d -clock clk -reset rst -rstlen 1 -n 1 -w check
```

Then run `yosys -s script.ys`.

## Expected behavior

I expect the tool to simulate a single step without errors.

## Actual behavior

The core crashes with an internal log_assert() error in passes/sat/sim.cc:set_state()

At issue is the null reference to the `count` output of the submodule.  Inside set_state(), the simulation tries to set a zero length value--associated with the count output of the submodule that ... doesn't go anywhere.

If the script file is adjusted to include `flatten`, the assertion failure vanishes.

cliffordwolf:
Fixes #1565

### DIFF
--- a/passes/sat/sim.cc
+++ b/passes/sat/sim.cc
@@ -230,7 +230,7 @@ struct SimInstance
 		bool did_something = false;
 
 		sig = sigmap(sig);
-		log_assert(GetSize(sig) == GetSize(value));
+		log_assert(GetSize(sig) <= GetSize(value));
 
 		for (int i = 0; i < GetSize(sig); i++)
 			if (state_nets.at(sig[i]) != value[i]) {


ZipCPU:
Looks good to me!

PR #1582, opened 2019-12-17
Check scratchpad for abc options

nakengelhardt:
Modifies the `abc9` and `abc` passes to check the scratchpad for options, e.g. `abc.script` would override the `-script` argument of abc script, which would then be overridden again by any value specified in the pass.

To be more specific:
```
scratchpad -set abc.script "foo" # Override default -script value to be "foo"
abc # Run with -script = "foo"

scratchpad -set abc.script "foo"
abc -script "bar" # Run with -script = "bar"
```


passes/techmap/abc.cc
```diff
@@ -732,6 +732,8 @@ void abc_module(RTLIL::Design *design, RTLIL::Module *current_module, std::strin
 					abc_script += script_file[i];
 		} else
 			abc_script += stringf("source %s", script_file.c_str());
+	} else if (design->scratchpad.count("abc.script")) {
```
eddiehung: Would it be better to instead do:
```
if (!script_file.empty())
    script_file = design->scratchpad_get_string("abc.script");
if (!script_file.empty()) {
		if (script_file[0] == '+') {
			for (size_t i = 1; i < script_file.size(); i++)
...
```
So that `abc{,9}.script` has the same meaning as the `-script` option and goes through the same codepath (yet be unaffected by commas)? That would also eliminate the need for a separate `-scriptfile` command...

nakengelhardt: I'm not a fan of in-band signaling in general, but I can implement it that way if you want. Is there any necessity for the scratchpad options to mirror the command options? It's an internal interface, and we might want it to be more granular than is reasonable for the command...

eddiehung: I'm not following what you're saying about in-band signalling -- can you elaborate please?

The first cut of the vision I had in mind for this is for it to override existing, public facing options without having to break `abc{,9}` out from their `synth_*` scripts. These would be ideally have the same behaviour as those existing options, so that separate documentation is not needed, even when they are promoted from `scratchpad` into something less internal.
(It that's the case, neither my suggested changes nor your original ones would have this override behaviour...)

For `scratchpad` settings that have no or much less overlap, then it makes sense.

Did you have a different vision in mind? Please share it for discussion!

nakengelhardt: Ah, another word for it is [magic values](https://akrzemi1.wordpress.com/2012/06/19/beware-of-magic-values/)? Basically, mixing control and data, in this case by using a "special" value to change the meaning of the parameter. It's unintuitive (you'd never guess the behavior without reading the doc) and always introduces the risk of confusion between the special signal and the regular value that just happens to be the same. In this case it's not a big deal since it's not very likely someone would start a file name with `+`, but it is legal to do that on most file systems.

As for the abc9 parameters, I agree that anything that is passed in the command arguments has higher priority than scratchpad values. And keeping the interfaces identical would definitely make it clearer exactly which parameter overrides which scratchpad entry. I was thinking that there would be more values that might be controllable via scratchpad, e.g. the -K or -W options for &if - these probably only make sense to modify in a development setting, so maybe they shouldn't be user-facing. But there's an argument to be made that that's far enough off the beaten path that at that point you should just specify the whole script. I guess let's implement the existing abc9 options first, and then see if we want/need anything else as we run our experiments - it's all academic at this point.

eddiehung: I hadn't heard of magic values, thanks! There are few places inside Yosys where this practice is, rightly or wrongly, taken (and I'm at fault for adding more). However, since `-script` is already out in the wide, I don't think it's the right place to change them up or add ambiguity here...

nakengelhardt:
How do you feel about this solution? Since we said the command line and scratchpad arguments should match as much as possible, I tried to make it work with the same handling code for both.

passes/techmap/abc9.cc
```diff
@@ -952,6 +952,42 @@ struct Abc9Pass : public Pass {
 #endif
 #endif
 
+		// get arguments from scratchpad first, then override by command arguments
+		std::string lut_arg, luts_arg;
```
nakengelhardt: By the way, some of the `string` declarations in this file use `std::` and some don't. Is there a preferred usage? I believe it's `yosys.h` that imports it...

passes/techmap/abc9.cc
```diff
@@ -952,6 +952,42 @@ struct Abc9Pass : public Pass {
...
+			delay_target = "-D " + design->scratchpad_get_string("abc9.D");
+		}
+		if (design->scratchpad.count("abc9.lut")) {
+			lut_arg = design->scratchpad_get_string("abc9.lut");
+		}
+		if (design->scratchpad.count("abc9.luts")) {
+			luts_arg = design->scratchpad_get_string("abc9.luts");
+		}
+		if (design->scratchpad.count("abc9.fast")) {
+			fast_mode = design->scratchpad_get_bool("abc9.fast", false /* default value if not bool-like */);
```
eddiehung: How about `design->scratchpad_get_bool("abc9.fast", fast_mode)` so that it inherits the default value?

passes/techmap/abc9.cc
eddiehung: There's no official guidance as far as I know, it's a bit of a inconsistent mess right now....

PR #1590, opened 2019-12-22
iopadmap: Emit tristate buffers with const OE for some edge cases.

mwkmwkmwk:


tests/techmap/iopadmap.ys
```diff
@@ -97,3 +111,12 @@ select -assert-count 1 @oeb %co %co @iob %i
 select -assert-count 1 @iob %co %co @o2b %i
 select -assert-count 1 @iob %co %co t:$_NOT_ %i
 select -assert-count 1 @o3b %ci %ci t:$_NOT_ %i
+
+select -assert-count 2 f/t:obuft
```
mwkmwkmwk: If you have some good idea on how to assert that 0 is in fact connected to the oe pin, please tell.

passes/techmap/iopadmap.cc
```diff
@@ -192,53 +192,97 @@ struct IopadmapPass : public Pass {
...
+							continue;
+						driven_bits.insert(dstbit);
 					}
 
 				for (auto wire : module->selected_wires())
 				{
 					if (!wire->port_output)
 						continue;
 
+					// Don't handle inout ports if we have no suitable buffer type.
```
ZirconiumX: These feel like they should emit warnings to the user if they ever trigger.

tests/techmap/iopadmap.ys
eddiehung: Perhaps look at `sat`? You can either get it to prove that a signal is a specific value, or you can throw some `assert`s in?

mwkmwkmwk: The signal is a constant wired straight to a cell port though, how would I get at it?

passes/techmap/iopadmap.cc
mwkmwkmwk: Not necessarily — in theory someone could be running `iopadmap` in several passes to insert various types of buffers (though I can't think of a combination where you'd have tinoutpad but not toutpad or the other way around — mostly you'd be inserting tristate buffers separately from non-tristate ones).  @eddiehung what do you think?

eddiehung: Hmm. I think if the user only specifies `-tinoutpad` or `-toutpad` but not the other then they would likely have a good reason for doing so; failing that a `$_TBUF_` would be left behind anyway. Perhaps a debug message might be useful here just in case?

tests/techmap/iopadmap.ys
eddiehung: Ah, you're right, sorry! I don't think there's an elegant way then.

passes/techmap/iopadmap.cc
mwkmwkmwk: I decided against adding it, it would require a refactor to do it right — if we just put a warning here, it will trigger whether tristate buffers are actually present or not,

tests/techmap/iopadmap.ys
eddiehung: Was traipsing through the Yosys docs when I stumbled across this: http://www.clifford.at/yosys/cmd_hilomap.html -- sounds like it should help?

mwkmwkmwk: Ohhhh, that would work, thanks for the idea!

PR #1600, opened 2019-12-28
Nitpick cleanup for ecp5

eddiehung:


eddiehung:
> I presume abc9 now supports submodules in boxes (this is why it is this way)?

I think what I removed is a remanent from before `(* lib_whitebox *)` existed.

PR #1603, opened 2020-01-01
ice40/ecp5: add support for both 1364.1 and Synplify/LSE RAM/ROM attributes

whitequark:
This commit tries to carefully follow the documented behavior of LSE
and Synplify. It will use `syn_ramstyle` attribute if there are any
write ports, and `syn_romstyle` attribute otherwise.
  * LSE supports both `syn_ramstyle` and `syn_romstyle`.
  * Synplify only supports `syn_ramstyle`, with same values as LSE.
  * Synplify also supports `syn_rw_conflict_logic`, which is not
    documented as supported for LSE.

Limitations of the Yosys implementation:
  * ~LSE/Synplify appear to interpret attribute values insensitive
    to case. There is currently no way to do this in Yosys (attrmap
    can only change case of attribute names).~
  * LSE/Synplify support `syn_ramstyle="block_ram,no_rw_check"`
    syntax to turn off insertion of transparency logic. There is
    currently no way to support multiple valued attributes in
    memory_bram. It is also not clear if that is a good idea, since
    it can cause sim/synth mismatches.
  * LSE/Synplify/1364.1 support block ROM inference from full case
    statements. Yosys does not currently perform this transformation.
  * LSE/Synplify propagate `syn_ramstyle`/`syn_romstyle` attributes
    from the module to the inner memories. There is currently no way
    to do this in Yosys (attrmvcp only works on cells and wires).


whitequark:
> * LSE/Synplify appear to interpret attribute values insensitive
>   to case.

I've added some code to `memory_bram` to handle this case as well.

whitequark:
I did a similar update of ECP5 BRAM rules, taking into account LUTRAM.

techlibs/ecp5/bram.txt
```diff
@@ -37,16 +37,78 @@ bram $__ECP5_DP16KD
...
+# https://www.latticesemi.com/-/media/LatticeSemi/Documents/Tutorials/AK/LatticeDiamondTutorial311.ashx
+attr_icase 1
+
 match $__ECP5_PDPW16KD
+	# implicitly requested RAM or ROM
+	attribute !syn_ramstyle syn_ramstyle=auto
+	attribute !syn_romstyle syn_romstyle=auto
+	attribute !ram_block
+	attribute !rom_block
+	attribute !logic_block
```
daveshah1: nitpick: looks like tabs and spaces have been mixed here

whitequark: ah good catch, I fixed this for ice40 but forgot ecp5

whitequark:
I've also adjusted `memory_map` so that it does not map memories that are explicitly required to use *not* FFRAM (i.e. BRAM or, if available, LUTRAM). This fixes the longstanding Yosys issue where if you make a mistake describing a BRAM behaviorally, it goes off to spend ten minutes mapping a megabit of RAM to DFFs.

whitequark:
Thinking out loud about possible improvements (I have no plans to implement any of this):
  * It seems valuable to have a single pass that would case-normalize Synposys attributes and where applicable map them to Yosys ones (`syn_keep` → `keep` seems particularly important). You could go quite far with just `attrmap`, but it seems ugly to duplicate it every time a vendor toolchain uses Synopsys-compatible attributes, since most of them do.
  * Multiple valued attributes should probably be supported similarly to "strpool" attributes, except that they are split by `,`, the accessors operate on a `vector<string>` in addition to `pool<string>`, and whitespace is stripped.
  * Extending `attrmvcp` to handle modules seems valuable because there are quite a few Synopsys attributes that need it (`syn_use_carry_chain`, `syn_useenables`, ...) but also there are Lattice-specific ones (`GSR`), so just doing this in Synopsys-specific code is probably not enough.
  * Existing `memory_bram` logic should be able to handle `no_rw_check` just fine, so whoever thinks that this is a good idea (I am unconvinced) can probably implement it easily once multiple valued attributes are in.

eddiehung:
Another related thing to just mention for discussion is that Synopsys directives can also appear as comments, so if there are improvements ahead it may be worth planning for that too...

whitequark:
> Another related thing to just mention for discussion is that Synopsys directives can also appear as comments, so if there are improvements ahead it may be worth planning for that too...

Let me elaborate a bit on my strategy here. When adding attributes like this one, it is good practice to do the least surprising thing for the HDL designer. In this case, to me, this means not inventing new attributes or meanings if at all possible; and so I am reusing Synopsys attribute names and values (even though they are inconsistent and a bit clunky), as well as 1364.1 ones (which I suspect were influenced by Synopsys).

Separately from that there is the issue of hot comments. I think they are an abomination, and Yosys certainly does not disagree, as the warnings suggest. I have actually considered implementing them and spent a bit of time before deciding that some things are best left in the past. It's been almost a decade since Verilog-2001, isn't it finally time to start using it? (Yes, you can't really do `// synthesis translate off` with attributes, but here we are discussing syntax that is exactly as powerful as Verilog-2001 attributes, except worse in every other aspect.)

eddiehung:
> Separately from that there is the issue of hot comments. I think they are an abomination, and Yosys certainly does not disagree, as the warnings suggest. I have actually considered implementing them and spent a bit of time before deciding that some things are best left in the past. It's been almost a decade since Verilog-2001, isn't it finally time to start using it? (Yes, you can't really do `// synthesis translate off` with attributes, but here we are discussing syntax that is exactly as powerful as Verilog-2001 attributes, except worse in every other aspect.)

I'm not advocating their continued use, but it may be an unavoidable necessity for supporting legacy code. One solution may be to convert those hot comments into attributes.

daveshah1:
Perhaps comments of this format should produce a warning advising the correct attribute but _not_ actually be honoured, to push people away from them.

whitequark:
> One solution may be to convert those hot comments into attributes.

If Yosys were to support such hot comments, I think converting them into attributes is the only realistic option. But I really like @daveshah1's suggestion of parsing them without honoring them. If there is no strong objection against it I might implement it; it's certainly true that an unfortunate amount of documentation, even recent, suggests using hot comments.

> it may be an unavoidable necessity for supporting legacy code

That is true. In my view, the best way to handle hot comments in legacy code is to release a tool that automatically converts them to Verilog-2001 attributes. Perhaps one already exists? If not, then one of the recently developed Verilog parsers may be useful (in terms of providing precise source locations) for such a tool. I wrote a few tools like that and could certainly do another.

whitequark:
Another option would be to (re)use Yosys' lexer to make a much more hacky code transformation tool tailored specifically to this issue. Without relying on a tooling-grade parser it is harder to make one that is 100% reliable because you can't easily check if the transformation you just done is indeed correct, but it might still be good enough for most uses.

whitequark:
Rebased the PR on top of master. Could I get a review please?

cc @eddiehung @cliffordwolf since this changes `memory_map` and `memory_bram`
cc @daveshah1 since this changes iCE40/ECP5 brams.txt
cc @mwkmwkmwk since this does *not* change Xilinx brams.txt; I could probably do the change but I don't use those devices myself and lack the context right now
cc @ZirconiumX since this does *not* change Intel brams.txt and I'm not even sure if it recognizes these attributes

anyone I forgot? Achronix, Anlogic, Efinix, Gowin? I'm not sure if those targets care about BRAM attributes yet.

ZirconiumX:
Quartus [does](https://www.intel.com/content/www/us/en/programmable/quartushelp/17.0/hdl/vlog/vlog_file_dir_ram.htm) recognise `syn_ramstyle` but interprets its contents a little differently.

passes/memory/memory_map.cc
```diff
@@ -356,17 +407,71 @@ struct MemoryMapPass : public Pass {
...
+				if (eq_at != std::string::npos) {
+					name  = attr_arg.substr(0, eq_at);
+					value = attr_arg.substr(eq_at + 1);
+				} else {
+					name  = attr_arg;
+					value = RTLIL::Const(1);
+				}
+				if (attr_arg.size() > 1 && attr_arg[0] == '!') {
+					if (value != RTLIL::Const(1))
+						break; // we don't support -attr !<name>=<value>
```
cliffordwolf: At this point we have already executed `++argidx`, therefore it will look like `extra_args()` as if `!<name>=<value>` is a select expression. We need to `--argidx` before break to make sure `extra_args()` sees `-attr` and produces an `Unknown option or option in arguments.` error message. Otherwise the changes to memory_map LGTM.

nakengelhardt:
@whitequark Sorry, I didn't quite manage to follow the status of this PR. Can you give me a bit of context? Do you want Claire to look at the whole thing or just the changes to memory_map/memory_bram? Is this ready to merge otherwise?

whitequark:
@nakengelhardt I fixed the two requested minor changes. There were no other changes to the PR. It is totally ready to merge otherwise.

whitequark:
Other than that, @ZirconiumX is already aware of the changes (which is all I wanted here), and @mwkmwkmwk never responded here but there is no particular action needed from her, I just wanted to give a heads up about implementing Synplify attributes so that other targets could take it into account.

PR #1607, opened 2020-01-01
ast: avoid intermediate wires/assigns when lowering to AST_MEMINIT

whitequark:
Before this commit, every initial assignment to a memory generated
two wires and four assigns in a process. For unknown reasons (I did
not investigate), large amounts of assigns cause quadratic slowdown
later in the AST frontend, in processAst/removeSignalFromCaseTree.
As a consequence, common and reasonable Verilog code, such as:

    reg [`WIDTH:0] mem [0:`DEPTH];
    integer i; initial for (i = 0; i <= `DEPTH; i++) mem[i] = 0;

took extremely long time to be processed; around 80 s for a 8-wide,
8192-deep memory.

After this commit, initial assignments where address and/or data are
constant (after `generate`) do not incur the cost of intermediate
wires; expressions like `mem[i+1]=i^(i<<1)` are considered constant.
This results in speedups of orders of magnitude for common memory
sizes; it now takes merely 0.4 s to process a 8-wide, 8192-deep
memory, and only 5.8 s to process a 8-wide, 131072-deep one.

As a bonus, this change also results in nontrivial speedups later
in the synthesis pipeline, since pass sequencing issues meant that
all of these intermediate wires were subject to transformations such
as width reduction, even though they existed solely to be constant
folded away in `memory_collect`.

### DIFF
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -831,7 +831,7 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
 			// Insert clones children from template at beginning
 			for (int i  = 0; i < GetSize(templ->children); i++)
 				children.insert(children.begin() + i, templ->children[i]->clone());
-			
+
 			if (type == AST_MEMORY && GetSize(children) == 1) {
 				// Single-bit memories must have [0:0] range
 				AstNode *rng = new AstNode(AST_RANGE);
@@ -878,7 +878,7 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
 			did_something = true;
 		}
 		log_assert(!is_custom_type);
-	}	
+	}
 
 	// resolve constant prefixes
 	if (type == AST_PREFIX) {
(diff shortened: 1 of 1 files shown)


whitequark:
Assigned to myself as there are failing tests which I haven't had time to fix yet.

whitequark:
I fixed the failing tests, and I'm reasonably confident this change is correct now.

frontends/ast/simplify.cc
```diff
@@ -1897,38 +1922,32 @@ skip_dynamic_range_lvalue_expansion:;
...
 		wrnode->id2ast = children[0]->id2ast;
-		wrnode->children[0]->str = id_addr;
-		wrnode->children[1]->str = id_data;
-		if (current_always->type != AST_INITIAL)
-			wrnode->children[2]->str = id_en;
 		current_ast_mod->children.push_back(wrnode);
 
+		if (newNode->children.empty()) {
+			delete newNode;
+			newNode = new AstNode();
```
cliffordwolf: I have not really digested this PR but does this mean we might potentially end up with an AstNode in the AST that has type `AST_NONE`?

whitequark: IIRC no, there's a condition elsewhere. But this PR has been in review too long and I forgot how these changes work. I would have to spend a hour or two to remember how this part of the frontend works.

whitequark:
Thanks!

PR #1620, opened 2020-01-09
abc9: add some scripts/options into "scratchpad"

eddiehung:
~DEPENDENCY: #1622~

Example usage:
`scratchpad -copy abc9.script.flow3 abc9.script` before `synth_xilinx -abc9`

* Added an `RTLIL::constpad` global for storing constant things that can be read from and copied into a design's scratchpad (initialised by `Pass::on_register()`) -- readable but not writable by the `scratchpad` command
* Added the following constpad entries for now:
```
abc9.script.default # (as before)
abc9.script.default.area # (as abc9.script.default, but with the "-a" option to &if)
abc9.script.default.fast # (as before)
abc9.script.flow
abc9.script.flow2
abc9.script.flow3
```
* `abc9.if.C` to control `&if -C`
* `abc9.if.R` to control `&if -R`
* `abc9.debug` as alias to `abc9.nocleanup` and `abc9.showtmp`
* `abc9.verify` to add `verify [-s]` to end of script

### DIFF
--- a/kernel/yosys.cc
+++ b/kernel/yosys.cc
@@ -524,6 +524,15 @@ void yosys_setup()
 		PyRun_SimpleString("import sys");
 	#endif
 
+	RTLIL::constpad["abc9.script.default"] = "&scorr; &sweep; &dc2; &dch -f; &ps; &if {C} {W} {D} -v; &mfs";
+	RTLIL::constpad["abc9.script.default.area"] = "&scorr; &sweep; &dc2; &dch -f; &ps; &if {C} {W} {D} -a -v; &mfs";
+	RTLIL::constpad["abc9.script.default.fast"] = "&if {W} {D}";
+	RTLIL::constpad["abc9.script.flow3"] = "&scorr; &sweep;" \
+		"&if {C} {W} {D}; &save; &st; &syn2; &if {C} {W} {D} -v; &save; &load; "\
+		"&st; &if {C} -g -K 6; &dch -f; &if {C} {W} {D} -v; &save; &load; "\
+		"&st; &if {C} -g -K 6; &synch2; &if {C} {W} {D} -v; &save; &load; "\
+		"&mfs";
+
 	Pass::init_register();
 	yosys_design = new RTLIL::Design;
 	yosys_celltypes.setup();
--- a/kernel/yosys.cc
+++ b/kernel/yosys.cc
@@ -524,13 +524,39 @@ void yosys_setup()
 		PyRun_SimpleString("import sys");
 	#endif
 
-	RTLIL::constpad["abc9.script.default"] = "&scorr; &sweep; &dc2; &dch -f; &ps; &if {C} {W} {D} -v; &mfs";
-	RTLIL::constpad["abc9.script.default.area"] = "&scorr; &sweep; &dc2; &dch -f; &ps; &if {C} {W} {D} -a -v; &mfs";
-	RTLIL::constpad["abc9.script.default.fast"] = "&if {W} {D}";
+	RTLIL::constpad["abc9.script.default"] = "&scorr; &sweep; &dc2; &dch -f; &ps; &if {C} {W} {D} {R} -v; &mfs";
+	RTLIL::constpad["abc9.script.default.area"] = "&scorr; &sweep; &dc2; &dch -f; &ps; &if {C} {W} {D} {R} -a -v; &mfs";
+	RTLIL::constpad["abc9.script.default.fast"] = "&if {C} {W} {D} {R}";
+	// Based on ABC's &flow
+	RTLIL::constpad["abc9.script.flow"] = "&scorr; &sweep;" \
+		/* Round 1 */ \
+		"&unmap; &if {C} {W} {D} {R}; &mfs;" \
+		"&st; &dsdb;" \
+		"&unmap; &if {C} {W} {D} {R}; &mfs;" \
+		"&st; &syn2 -m -R 10; &dsdb;" \
+		"&blut -a -K 6;" \
+		"&unmap; &if {C} {W} {D} {R}; &mfs;" \
+		/* Round 2 */ \
+		"&st; &sopb;" \
+		"&unmap; &if {C} {W} {D} {R}; &mfs;" \
+		"&st; &dsdb;" \
+		"&unmap; &if {C} {W} {D} {R}; &mfs;" \
+		"&st; &syn2 -m -R 10; &dsdb;" \
+		"&blut -a -K 6;" \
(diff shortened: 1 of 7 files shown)


passes/techmap/abc9.cc
```diff
@@ -281,18 +267,24 @@ void abc9_module(RTLIL::Design *design, RTLIL::Module *module, std::string scrip
...
-		if (script_file[0] == '+') {
-			for (size_t i = 1; i < script_file.size(); i++)
-				if (script_file[i] == '\'')
-					abc9_script += "'\\''";
-				else if (script_file[i] == ',')
-					abc9_script += " ";
-				else
-					abc9_script += script_file[i];
-		} else
+		if (check_file_exists(script_file))
```
nakengelhardt: So with this change we can have file names starting with `+` (cool!) and unadorned scripts (also cool!)
but if the file doesn't exist it just puts the bare file name in the script. Then abc prints the message `ABC: Generic file reader requires a known file extension to open "nonexistent.script".` but that isn't actually an error, it continues until abc9 fails with ```ERROR: Assert `existing_cell' failed in passes/techmap/abc9.cc:520.``` Do you think that is explicit enough for users to figure out the problem or should we try harder to find out if the string is meant to be a filename or a script? (E.g., can you have an abc9 script that doesn't contain `&` or `;`?)

passes/techmap/abc9.cc
```diff
@@ -305,11 +297,23 @@ void abc9_module(RTLIL::Design *design, RTLIL::Module *module, std::string scrip
 	for (size_t pos = abc9_script.find("{W}"); pos != std::string::npos; pos = abc9_script.find("{W}", pos))
 		abc9_script = abc9_script.substr(0, pos) + wire_delay + abc9_script.substr(pos+3);
 
+	std::string C;
+	if (design->scratchpad.count("abc9.if.C"))
```
nakengelhardt: Do you want to add the -C and -R options to the docstring?

eddiehung: No, let's keep them as undocumented options for now... other scratchpad entries are not documented yet either.

kernel/yosys.cc
```diff
@@ -524,6 +524,41 @@ void yosys_setup()
 		PyRun_SimpleString("import sys");
 	#endif
 
+	RTLIL::constpad["abc9.script.default"] = "&scorr; &sweep; &dc2; &dch -f; &ps; &if {C} {W} {D} {R} -v; &mfs";
```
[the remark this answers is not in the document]
eddiehung: Resolved with 67c9c41f7e566f5604a3e38e7ad402d6b5c80fd8 with the help of #1622

passes/techmap/abc9.cc
eddiehung: As discussed, reverted change so that all non-files must start with `+` (but unlike the command line, scratchpad entries can contain spaces)

passes/techmap/abc9.cc
```diff
@@ -305,11 +292,23 @@ void abc9_module(RTLIL::Design *design, RTLIL::Module *module, std::string scrip
...
+		C = "-C " + design->scratchpad_get_string("abc9.if.C");
+	for (size_t pos = abc9_script.find("{C}"); pos != std::string::npos; pos = abc9_script.find("{C}", pos))
+		abc9_script = abc9_script.substr(0, pos) + C + abc9_script.substr(pos+3);
+
+	std::string R;
+	if (design->scratchpad.count("abc9.if.R"))
+		C = "-C " + design->scratchpad_get_string("abc9.if.R");
```
eddiehung: Note to self: `C` -> `R`

passes/techmap/abc9.cc
```diff
@@ -305,11 +292,23 @@ void abc9_module(RTLIL::Design *design, RTLIL::Module *module, std::string scrip
...
+	for (size_t pos = abc9_script.find("{R}"); pos != std::string::npos; pos = abc9_script.find("{R}", pos))
+		abc9_script = abc9_script.substr(0, pos) + C + abc9_script.substr(pos+3);
```
eddiehung: And again

passes/techmap/abc9.cc
```diff
@@ -758,18 +802,15 @@ struct Abc9Pass : public Pass {
 		log("\n");
 		log("        if no -script parameter is given, the following scripts are used:\n");
 		log("\n");
-		log("        for -lut/-luts (only one LUT size):\n");
-		log("%s\n", fold_abc9_cmd(ABC_COMMAND_LUT).c_str());
-		log("\n");
-		log("        for -lut/-luts (different LUT sizes):\n");
-		log("%s\n", fold_abc9_cmd(ABC_COMMAND_LUT).c_str());
+		log("        for -lut/-luts:\n");
+		log("%s\n", fold_abc9_cmd(RTLIL::constpad.at("abc9.script.default")).c_str());
```
eddiehung: `.c_str()+1` to rid the `+`

passes/techmap/abc9.cc
```diff
@@ -758,18 +802,15 @@ struct Abc9Pass : public Pass {
...
 		log("\n");
 		log("    -fast\n");
 		log("        use different default scripts that are slightly faster (at the cost\n");
 		log("        of output quality):\n");
 		log("\n");
 		log("        for -lut/-luts:\n");
-		log("%s\n", fold_abc9_cmd(ABC_FAST_COMMAND_LUT).c_str());
+		log("%s\n", fold_abc9_cmd(RTLIL::constpad.at("abc9.script.default.fast")).c_str());
```
eddiehung: Same as above

passes/techmap/abc9.cc
```diff
@@ -758,18 +809,15 @@ struct Abc9Pass : public Pass {
...
+		log("        for -lut/-luts:\n");
+		log("%s\n", fold_abc9_cmd(RTLIL::constpad.at("abc9.script.default")).c_str()+1);
 		log("\n");
 		log("    -fast\n");
 		log("        use different default scripts that are slightly faster (at the cost\n");
 		log("        of output quality):\n");
 		log("\n");
 		log("        for -lut/-luts:\n");
-		log("%s\n", fold_abc9_cmd(ABC_FAST_COMMAND_LUT).c_str());
+		log("%s\n", fold_abc9_cmd(RTLIL::constpad.at("abc9.script.default.fast")).c_str()+1);
```
nakengelhardt: This cuts the semicolon off the end of the script rather than the + at the beginning...

eddiehung: No, it starts the string one character in; however, `fold_abc9_cmd()` returns a string with spaces in front, and so I was just chopping off a space. Thanks for spotting!

PR #1642, opened 2020-01-17
Enum support

### LINKED ISSUE #248: SystemVerilog enum types
Hi,
apparently, enums are not supported (yet?) by yosys, even when setting the -sv flag for read_verilog. This small example fails:

```
module foo(input clk, input rst);
   typedef enum {S0, S1, S2} state_t;
   state_t state;

    if (rst) begin
      state <= S0;
    end else begin
      state <= state.next();     
    end
  end
endmodule
```

There are workarounds (rewrite code using localparams or defines), but a better support for SystemVerilog features would be very nice.

jjj11x:
I rebased PeterCrozier's enum work onto current master
    
I tried to keep only the enum-related changes, and minimize the diff. (The original commit also had a lot of work done to get typedefs working, but yosys has diverged quite a bit since the 2018-03-09 commit, with a new typedef implementation https://github.com/YosysHQ/yosys/pull/1389.) I did not include the import related changes either.

I added support for enum typedefs at the toplevel scope and in generate if blocks, and some simple tests. It works so far for my purposes, but any comments are welcome. The enums are currently not strongly typed.

Resolves #248
    
Original commit:
"Initial implementation of enum, typedef, import.  Still a WIP."
https://github.com/PeterCrozier/yosys/commit/881833aa738e7404987646ea8076284e911fce3f

### DIFF
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -324,7 +324,7 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
 		const_fold = true;
 
 	// in certain cases a function must be evaluated constant. this is what in_param controls.
-	if (type == AST_PARAMETER || type == AST_LOCALPARAM || type == AST_LOCALPARAM || type == AST_ENUM_ITEM || type == AST_DEFPARAM || type == AST_PARASET || type == AST_PREFIX)
+	if (type == AST_PARAMETER || type == AST_LOCALPARAM || type == AST_ENUM_ITEM || type == AST_DEFPARAM || type == AST_PARASET || type == AST_PREFIX)
 		in_param = true;
 
 	std::map<std::string, AstNode*> backup_scope;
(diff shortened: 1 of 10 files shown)


whitequark:
cc @RobertBaruch

whitequark:
@jjj11x Do you think you could add some code that packs the enum key-value map into an attribute, attached to the RTLIL wires that were declared with the type of that enum? This would be greatly helpful for producing symbolic enum values with formal verification tools.

jjj11x:
> @jjj11x Do you think you could add some code that packs the enum key-value map into an attribute, attached to the RTLIL wires that were declared with the type of that enum? This would be greatly helpful for producing symbolic enum values with formal verification tools.

I can take a look, although I'm really not an expert in any of this. (Without PeterCrozier's changes I would've really had no idea where to start. This is my first time looking looking at the Yosys source code...trying to read through the manual now.)

daveshah1:
Also FYI, the reason brackets are currently required around typedef names is discussed in #1448. I'm not sure what the best solution long term is here.

jjj11x:
@whitequark @RobertBaruch I added code to add the enum key-value map to attributes attached to the RTLIL wires that were declared with the type of that enum.

e.g. given: 
```

File: frontends/ast/simplify.cc [hunk 1]
```diff
@@ -822,11 +863,68 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
 			range_swapped = templ->range_swapped;
 			range_left = templ->range_left;
 			range_right = templ->range_right;
+			attributes["\\wiretype"] = mkconst_str(resolved_type->str);
```
cliffordwolf: New attributes should be added to the attributes section of the README file.

frontends/ast/simplify.cc
```diff
@@ -2513,6 +2637,10 @@ skip_dynamic_range_lvalue_expansion:;
 					wire->is_output = false;
 					wire->is_reg = true;
 					wire->attributes["\\nosync"] = AstNode::mkconst_int(1, false);
+					if (child->type == AST_ENUM_ITEM){
+						wire->attributes["\\enum_base_type"] = child->attributes["\\enum_base_type"];
+
+					}
```
cliffordwolf: Nitpick: unnecessary empty line.

frontends/ast/simplify.cc
jjj11x: added some documentation

PR #1648, opened 2020-01-21
"techmap -map +/cmp2lcu.v" for decomposing arithmetic compares to $lcu

eddiehung:
More efficient than treating `$gt` `$ge` `$lt` `$le` as ALU operations by having each stage of the carry chain compare more than one pair of bits (number of bits controlled by `` `LUT_WIDTH`` macro).

Question remains whether it is more efficient to use the carry-chain at all, since if it was all soft-logic then technology mapping into LUTs has more flexibility...

Also, for specific technologies, it may be more efficient to instantiate a primitive (e.g. `LUT6_2` for Xilinx) in order to make even more efficient use of the fracturable LUT that feeds the carry.

Currently demonstrated inside `synth_xilinx` using `-D LUT_WIDTH=6` since there is no guarantee that `-D LUT_WIDTH=5` will cause two fracture-compatible LUTs (i.e. with shared inputs) to be created. Ideas welcome here. `+/xilinx/arith_map.v` then maps `$lcu` to `CARRY4`.


techlibs/common/cmp2lut.v
```diff
@@ -75,10 +75,10 @@ generate
...
 	if (_TECHMAP_CELLTYPE_ == "$ge")
 		localparam operation = 3;
-	if (_TECHMAP_CELLTYPE_ == "$eq")
-		localparam operation = 4;
-	if (_TECHMAP_CELLTYPE_ == "$ne")
-		localparam operation = 5;
+//	if (_TECHMAP_CELLTYPE_ == "$eq")
+//		localparam operation = 4;
+//	if (_TECHMAP_CELLTYPE_ == "$ne")
+//		localparam operation = 5;
```
whitequark: Just remove these (and handling of operation = 5), I don't like having useless code around.

whitequark:
Looks good to me overall, and much more appropriate than my quick `cmp2lut` hack that I made without an in-depth understanding of arithmetic operation synthesis. (For example I did not know there is a `$lcu` cell back then.)

techlibs/common/cmp2lut.v
eddiehung: Thanks! Done in 970beee95.

whitequark:
I'm currently learning a few things about adders, so no review yet, but can I ask you to do some housekeeping? Normally when I first learn how a cell works, I add it to the manual with a detailed explanation. Right now it says:

```
\begin{fixme}
Add information about {\tt \$alu}, {\tt \$macc}, {\tt \$fa}, and {\tt \$lcu} cells.
\end{fixme}
```

I think it would be really helpful if you documented `$alu`, `$fa` and `$lcu` cells, since you're working on this anyway. Ideally with a link to some paper/slides explaining how they work in detail (e.g. why `$alu` has a `X` output), but only if you already know one.

eddiehung:
> I'm currently learning a few things about adders, so no review yet, but can I ask you to do some housekeeping? Normally when I first learn how a cell works, I add it to the manual with a detailed explanation. Right now it says:
> 
> ```
> \begin{fixme}
> Add information about {\tt \$alu}, {\tt \$macc}, {\tt \$fa}, and {\tt \$lcu} cells.
> \end{fixme}
> ```
> 
> I think it would be really helpful if you documented `$alu`, `$fa` and `$lcu` cells, since you're working on this anyway. Ideally with a link to some paper/slides explaining how they work in detail (e.g. why `$alu` has a `X` output), but only if you already know one.

When I went through the same exercise, I added my `$alu` and `$lcu` learnings into `+/simlib.v` which is exposed as so:
```
yosys -Qp "help \$alu"

-- Running command `help $alu' --

    $alu (A, B, CI, BI, X, Y, CO)

Arithmetic logic unit.
A building block supporting both binary addition/subtraction operations, and
indirectly, comparison operations.
Typically created by the `alumacc` pass, which transforms:
  $add, $sub, $lt, $le, $ge, $gt, $eq, $eqx, $ne, $nex
cells into this $alu cell.

Run 'help $alu+' to display the Verilog model for this cell type.
```

and

```
./yosys -Qp "help \$alu+"

-- Running command `help $alu+' --

module \$alu (A, B, CI, BI, X, Y, CO);

    parameter A_SIGNED = 0;
    parameter B_SIGNED = 0;
    parameter A_WIDTH = 1;
    parameter B_WIDTH = 1;
    parameter Y_WIDTH = 1;

    input [A_WIDTH-1:0] A;      // Input operand
    input [B_WIDTH-1:0] B;      // Input operand
    output [Y_WIDTH-1:0] X;     // A xor B (sign-extended, optional B inversion,
                                //          used in combination with
                                //          reduction-AND for $eq/$ne ops)
    output [Y_WIDTH-1:0] Y;     // Sum

    input CI;                   // Carry-in (set for $sub)
    input BI;                   // Invert-B (set for $sub)
    output [Y_WIDTH-1:0] CO;    // Carry-out
<<SNIP>>
```
You're right though: these should also go into the manual.

whitequark:
OK, so if I understand correctly, what this pass does is. It recognizes that to compute `a < b`, it is not necessary to know all bits of the result; it is only necessary to know whether `a - b` generates a carry. Because carry can be computed faster than the entire result, it is advantageous to compute only carry and use that. Is that correct?

eddiehung:
> OK, so if I understand correctly, what this pass does is. It recognizes that to compute `a < b`, it is not necessary to know all bits of the result; it is only necessary to know whether `a - b` generates a carry. Because carry can be computed faster than the entire result, it is advantageous to compute only carry and use that. Is that correct?

Not quite, and sorry I didn't actually explain it at all. Basically, there are two things:
1. I am "abusing" the carry chain (`$lcu`) in order to propagate the "true so far" status (and thus, the final carry out would be the final status) as opposed to using general purpose routing, with the expectation that the carry chain will be faster.
2. As opposed to the `alumacc` approach, which you can think of as "comparing" one pair of bits per carry-chain stage, this work typically compares more than one pair (the number of pairs is controlled by `` `LUT_WIDTH``) per stage. Furthermore, if both pairs of bits being compared are non-constant, then this consumes two LUT input (`COST`s 2, reducing `BUDGET` by this amount), if just one of the two pairs is constant then it costs one LUT input, and if both pairs are constant, then it consumes zero inputs. LUT input `BUDGET` is decremented until it can fit no more comparisons, in which case it moves onto the next `$lcu` stage.

Credit to @mwkmwkmwk for the original idea and helping me flesh it out. Currently, this is done via recursive techmapping, which can be quite slow. I imagine it might be possible to improve this by using a `for` inside a process for a more opaque solution (but I don't think it's possible to use a `generate`-`for` elegantly, which was my first stab, since the increment index must be constant...)

whitequark:
I understand it now, thanks for explanation. I agree that this pass seems close to a superset of `cmp2lut` and I would be very happy to see the latter removed. Let's, then, think of how to do it.

In the past I've advocated for using LUT mapping instead of AIG mapping via abc. However, that has some downsides; flowmap doesn't do logic synthesis, and it's not nearly as battle-tested. So for a short-medium term solution, I propose a pass that finds carry chains ($alu and $lcu cells) that can be fit in a LUT, and techmaps those.

I think it could be done as a two-part pass: a generic LUT mapper for any const-evaluatable cells (and this could be reused in many other places, I suspect), plus an algorithm (via pmgen perhaps?) that recognizes the constructs that are advantageous.

eddiehung:
> I understand it now, thanks for explanation. I agree that this pass seems close to a superset of `cmp2lut` and I would be very happy to see the latter removed. Let's, then, think of how to do it.

Sounds good. I'm beginning to think that this PR is not a superset of `cmp2lut` and both of these can be mutually exclusive. To summarise:
* `cmp2lut` does two things: (a) blocks `alumacc` from taking small compares and irreversibly mapping them to carry chains, and (b) maps them directly to a `$lut` cell (come to think of it, `abc9` for will blow this `$lut` back into AIG anyway, though this could be avoided)
* `cmp2lcu` breaks down a big compare into multiple small compares (each of which ideally fitting into a LUT, but as I discuss below, turns out not to be the case) -- this is only advantageous when the carry-chain is on the output side of a LUT, and not the input side like in iCE40.

If this is the case, would you be okay with merging this PR in its current form?

My current interpretation is that the biggest value `cmp2lut` provides is point (a). I think ABC still remains the best place to perform LUT mapping, since it is timing aware and able to do advanced things like don't-care/false-path optimisations and sequential synthesis.

> In the past I've advocated for using LUT mapping instead of AIG mapping via abc. However, that has some downsides; flowmap doesn't do logic synthesis, and it's not nearly as battle-tested. So for a short-medium term solution, I propose a pass that finds carry chains ($alu and $lcu cells) that can be fit in a LUT, and techmaps those.

I think such a pass could be at least prototyped using `techmap` rules as I do here. As I said originally, I see the greatest value being that we block `alumacc` from doing its stuff, because once it does so, it stays that way. To refine this a little more, `alumacc` creates `$alu` cells and then typically `+/<arch>/arith_map.v` maps that to carry-chains. Perhaps `arith_map.v` should be the arbiter of what goes into carry chains and what remains as soft-logic?

One thing which ABC cannot do is technology map multi-output cells, and this is where I think a custom pass might have an edge. In this case, I'm specifically referring to Xilinx architecture which has a fracturable LUT (`LUT6_2`) which in the simple form is two independent 5-input LUTs sharing the same 5 inputs. These two inputs are directly connected to the "propagate" and "generate" inputs of a `$lcu`, and it would be advantageous to map to this `LUT6_2` directly. Currently, ABC has the flexibility of choosing any sized LUTs for `P` and `G` with no constraint on the number of shared inputs, meaning there is no guarantee or any sort of preference that they be compatible for fracturing, which is inefficient.

This is not dissimilar to the iCE40 restriction where we have to stamp out a `$lut` in front of `SB_CARRY` (the pair of which gets blackboxed for ABC, and whiteboxed for ABC9) with `opt_lut` used to pack in as much as possible. 

> I think it could be done as a two-part pass: a generic LUT mapper for any const-evaluatable cells (and this could be reused in many other places, I suspect), plus an algorithm (via pmgen perhaps?) that recognizes the constructs that are advantageous.

Without finishing the discussion above, I haven't put a lot of thought into what the algorithm would look like. However, one thing that has crossed my mind is that (at least currently, though this might change) `abc9` calls `aigmap` in order to blow everything into AIGs. Calling something like `opt_lut` at this point would make it trivially simple to perform LUT mapping...

whitequark:
> * `cmp2lcu` breaks down a big compare into multiple small compares (each of which ideally fitting into a LUT, but as I discuss below, turns out not to be the case) -- this is only advantageous when the carry-chain is on the output side of a LUT, and not the input side like in iCE40.

Ahhh I see, my thinking is a bit skewed towards iCE40.

> If this is the case, would you be okay with merging this PR in its current form?

Certainly. I'm very happy to see any techmapping improvements.

> My current interpretation is that the biggest value `cmp2lut` provides is point (a). I think ABC still remains the best place to perform LUT mapping, since it is timing aware and able to do advanced things like don't-care/false-path optimisations and sequential synthesis.

Indeed, and this is not different from my understanding; the reason `cmp2lut` maps to LUTs specifically is that it's really easy to map combinatorial functions to LUTs. IIRC, I did not intend for those LUTs to specifically survive through the entire chain. (On the testcases that I ran, I think they did happen to survive, since the mapping happened to be optimal.)

> Calling something like `opt_lut` at this point would make it trivially simple to perform LUT mapping...

Funny that you mention it... The story goes, I was pretty frustrated with abc, and floated the idea of writing a new LUT mapper, when someone (Dave maybe?) told me it's a hard and nasty problem few people want to tackle. That's when I wrote `gate2lut` (and `cmp2lut`), which, when combined with `opt_lut`, had similar Fmax and twice the area of pre-abc9 `synth_ice40`. Given that it took me only a day to do so, and none of these passes makes even any attempt at optimization, it demonstrated that maybe the problem isn't so hard and nasty, and inspired my interest in flowmap.

In other words, it would, and it does :)

eddiehung:
>>Calling something like opt_lut at this point would make it trivially simple to perform LUT mapping...

> In other words, it would, and it does :)

Sorry, my original statement was along the lines of stamping out a `LUT6_2` (much like stamping out a `SB_LUT4`) and using `opt_lut` to cram things into it. As I understand, `opt_lut` currently merges whole LUTs opportunistically (please fact check that, I haven't studied it in detail) while allowing input constraints to be preserved. Then you would present that as a whitebox to `abc9` so that its logic and delay can be considered during techmapping of the rest of the soft-logic.

whitequark:
> Sorry, my original statement was along the lines of stamping out a `LUT6_2` (much like stamping out a `SB_LUT4`) and using `opt_lut` to cram things into it.

Right--that's exactly how `gate2lut` works, no? Or am I misunderstanding?

> As I understand, `opt_lut` currently merges whole LUTs opportunistically

Correct. Usually same sized LUTs, but it can also merge differently sized LUTs in some cases.

> while allowing input constraints to be preserved

Sooooort of. The `-dlogic` option is probably not really suitable for much besides iCE40. I tried to make it generic but I think I failed.

ZirconiumX:
Not directly related to the topic, but could this be put in `synth -lut K` like `techmap -map +/cmp2lut.v` already is? #1554 uses `synth -lut K -run :fine` for its coarse stage, and I think the coarse stage of `synth` is already unnecessarily duplicated across ECP5, iCE40 and Xilinx. (perhaps `synth` should grow a `-mul2dsp` option?)

eddiehung:
> > Sorry, my original statement was along the lines of stamping out a `LUT6_2` (much like stamping out a `SB_LUT4`) and using `opt_lut` to cram things into it.
> 
> Right--that's exactly how `gate2lut` works, no? Or am I misunderstanding?

Kind of? `gate2lut` is only necessary to create bigger-than-necessary `$lut` cells that `opt_lut` can then recognise. An alternative would be to skip `gate2lut` entirely and work on any built-in Yosys (bit-level) cell.

eddiehung:
> Not directly related to the topic, but could this be put in `synth -lut K` like `techmap -map +/cmp2lut.v` already is? #1554 uses `synth -lut K -run :fine` for its coarse stage

Yes, I think that's a reasonable suggestion.

> I think the coarse stage of `synth` is already unnecessarily duplicated across ECP5, iCE40 and Xilinx. (perhaps `synth` should grow a `-mul2dsp` option?)

As you've spotted, the initial reason we needed to break away from `synth` is when we started adding multiplier/DSP support, however, since then there has been some further deviation, namely wide multiplexer support too, which could also potentially go into this more general `synth` under some option -- this is something @mwkmwkmwk has been pushing for too. Let's discuss this in a separate issue though.

techlibs/xilinx/arith_map.v
```diff
@@ -36,13 +36,12 @@ module _80_xilinx_lcu (P, G, CI, CO);
 `ifdef _EXPLICIT_CARRY
```
mwkmwkmwk: This needs to be rebased out of existence.

techlibs/xilinx/synth_xilinx.cc
```diff
@@ -410,6 +408,7 @@ struct SynthXilinxPass : public ScriptPass
 		}
 
 		if (check_label("coarse")) {
+			run("techmap -map +/cmp2lut.v -map +/cmp2lcu.v -D LUT_WIDTH=6");
```
mwkmwkmwk: This needs to be changed to use `lut_size_s`

techlibs/common/cmp2lcu.v
```diff
@@ -0,0 +1,116 @@
+// This pass performs an optimisation that decomposes wide arithmetic
+//   comparisons into LUT-size chunks (as guided by the `LUT_WIDTH
+//   macro) connected to a single lookahead-carry-unit $lcu cell,
+//   which is typically mapped to dedicated (and fast) FPGA
+//   carry-chains.
+(* techmap_celltype = "$lt $le $gt $ge" *)
+module _90_lcu_cmp_ (A, B, Y);
```
cliffordwolf: Considering that this is competing with `_90_lut_cmp_` in the same techmap call in `synth` and `synth_xilinx`, I think it would make sense to pick a number different from 90 to make clear which one should be considered first by techmap.

PR #1651, opened 2020-01-22
read_verilog: allow width detection of unreachable ternary operation branches to fail

### LINKED ISSUE #1649: Width detection fails for genvar used with conditional operator
I am trying to synthesize https://github.com/jotego/jt12/blob/master/hdl/mixer/jt12_comb.v for the ice40. It seems there is an issue with genvar width detection inside a conditional operator when synthesizing using yosys. I am running git hash f165a748

I have created a simpler example which shows the issue:

```
module delay #(parameter width=2, taps=1)(
       input clk,
       input [width-1:0] d_in,
       output reg [width-1:0] d_out
);

generate
        genvar k;
        reg [width-1:0] mem[0:taps-1];

        for (k=0;k<taps;k=k+1) begin : delay_gen
                always @(posedge clk) begin
                        mem[k] <= k==0 ? d_in : mem[k-1]; // <<<<<<<<< ERROR IS HERE
                end
        end
endgenerate

always @(posedge clk) begin
        b <= mem[taps-1];
end

endmodule


module top (input clk, input wire [3:0] a, output wire [3:0] b);
        delay #(.width(4), .taps(1))  d(clk, a, b);
endmodule
```

running ` yosys -p "synth_ice40 -top top" genvar_test.v`, I get:

```
... license info removed for brevity ...

 Yosys 0.9+1706 (git sha1 f165a748, clang 6.0.0-1ubuntu2 -fPIC -Os)


-- Parsing `genvar_test.v' using frontend `verilog' --

1. Executing Verilog-2005 frontend: genvar_test.v
Parsing Verilog input from `genvar_test.v' to AST representation.
Generating RTLIL representation for module `\delay'.
Warning: wire '\b' is assigned in a block at genvar_test.v:19.
Warning: Replacing memory \mem with list of registers. See genvar_test.v:13
genvar_test.v:13: ERROR: Failed to resolve identifier \mem[-1] for width detection!
```

## Expected behavior

It should synthesize because `mem[k-1]` is unreachable for k==0. Replacing the conditional operator line with a full `if`/`else`/`end` allows yosys to synthesize the code correctly. It's not totally clear to me from the verilog 2005 standard if conditional operators are allowed with genvars, but it seems to be roughly supported anyway.

## Actual behavior

See error output above.

eddiehung:
Fixes #1649, and fixes even something as simple as this:
```
module top(output o1, o2);
wire i [0:0]; // Only errors out on unpacked...
assign o1 = (0 == 0) ? i[0] : i[-1];
assign o2 = (1 > 1) ? i[1] : i[0];
endmodule
```
where one branch is unreachable.

I'm not sure how other tools handle both testcases (perhaps OP @jeremyherbert would be so kind to check) but I see no reason not to implement this. I've been bitten a few times by this that I've committed using `generate-if` to muscle-memory at the expense of elegance.

### DIFF
--- a/frontends/ast/ast.h
+++ b/frontends/ast/ast.h
@@ -250,8 +250,8 @@ namespace AST
 		void dumpVlog(FILE *f, std::string indent) const;
 
 		// used by genRTLIL() for detecting expression width and sign
-		void detectSignWidthWorker(int &width_hint, bool &sign_hint, bool *found_real = NULL);
-		void detectSignWidth(int &width_hint, bool &sign_hint, bool *found_real = NULL);
+		void detectSignWidthWorker(int &width_hint, bool &sign_hint, bool *found_real = NULL, bool can_fail = false);
+		void detectSignWidth(int &width_hint, bool &sign_hint, bool *found_real = NULL, bool can_fail = false);
 
 		// create RTLIL code for this AST node
 		// for expressions the resulting signal vector is returned
--- a/frontends/ast/genrtlil.cc
+++ b/frontends/ast/genrtlil.cc
@@ -580,7 +580,7 @@ struct AST_INTERNAL::ProcessGenerator
 };
 
 // detect sign and width of an expression
-void AstNode::detectSignWidthWorker(int &width_hint, bool &sign_hint, bool *found_real)
+void AstNode::detectSignWidthWorker(int &width_hint, bool &sign_hint, bool *found_real, bool can_fail)
 {
 	std::string type_name;
 	bool sub_sign_hint = true;
@@ -610,8 +610,11 @@ void AstNode::detectSignWidthWorker(int &width_hint, bool &sign_hint, bool *foun
 		id_ast = id2ast;
 		if (id_ast == NULL && current_scope.count(str))
 			id_ast = current_scope.at(str);
-		if (!id_ast)
+		if (!id_ast) {
+			if (can_fail)
+				return;
 			log_file_error(filename, linenum, "Failed to resolve identifier %s for width detection!\n", str.c_str());
+		}
 		if (id_ast->type == AST_PARAMETER || id_ast->type == AST_LOCALPARAM) {
 			if (id_ast->children.size() > 1 && id_ast->children[1]->range_valid) {
 				this_width = id_ast->children[1]->range_left - id_ast->children[1]->range_right + 1;
@@ -769,8 +772,10 @@ void AstNode::detectSignWidthWorker(int &width_hint, bool &sign_hint, bool *foun
 		break;
 
 	case AST_TERNARY:
-		children.at(1)->detectSignWidthWorker(width_hint, sign_hint, found_real);
-		children.at(2)->detectSignWidthWorker(width_hint, sign_hint, found_real);
+		children.at(1)->detectSignWidthWorker(width_hint, sign_hint, found_real,
+				children.at(0)->type == AST_CONSTANT && !children.at(0)->asBool() /* can_fail */);
+		children.at(2)->detectSignWidthWorker(width_hint, sign_hint, found_real,
+				children.at(0)->type == AST_CONSTANT && children.at(0)->asBool() /* can_fail */);
 		break;
 
 	case AST_MEMRD:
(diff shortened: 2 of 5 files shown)


cliffordwolf:
I don't think this is a valid transformation. Consider the following code:

```
wire foo, bar;
assign foo = &(1 ? bar : something);
```

If the width of `something` is >1 then foo should be assigned constant zero and if the width of `something` is 1 then foo should be assigned bar. (I actually reported a bug just like that in quartus a long time ago: http://www.clifford.at/yosys/vloghammer_bugs/issue_001_quartus.html)

eddiehung:
> I don't think this is a valid transformation. Consider the following code:
> 
> ```
> wire foo, bar;
> assign foo = &(1 ? bar : something);
> ```
> 
> If the width of `something` is >1 then foo should be assigned constant zero and if the width of `something` is 1 then foo should be assigned bar. (I actually reported a bug just like that in quartus a long time ago: http://www.clifford.at/yosys/vloghammer_bugs/issue_001_quartus.html)

Did you try it? I checked the linked test-case on master and this branch and they both give:
```
./yosys top.v -qp "opt" -b ilang
# Generated by Yosys 0.9+1706 (git sha1 6c5490d2, clang 9.0.0-2 -fPIC -Os)
autoidx 2
attribute \cells_not_processed 1
attribute \src "top.v:1"
module \issue_001
  attribute \src "top.v:2"
  wire width 3 input 1 \a
  attribute \src "top.v:3"
  wire width 4 input 2 \b
  attribute \src "top.v:4"
  wire output 3 \y
  connect \y 1'0
end
```
which I believe is the correct solution.

Note that my patch makes no changes to width detection except allowing it to fail (i.e. not calling `log_error` and silently `return`-ing) when one branch is unreachable, and so it should not affect this test-case.

The only other thing it changes is that it does not generate the RTLIL for cases where the condition is constant. This prevents a further ``Warning: Identifier `i[-1]' is implicitly declared.`` for the unpacked identifier in the testcase above.

cliffordwolf:
> Did you try it? I checked the linked test-case on master and this branch and they both give:

The test case for the quartus bug does not involve failing width detection. Consider the following case:

```
wire [1:0] t [0:0]; 
assign o = &((0 == 0) ? i : t[-1]);
endmodule
```

This should tie `o` to constant zero.

eddiehung:
> > Did you try it? I checked the linked test-case on master and this branch and they both give:
> 
> The test case for the quartus bug does not involve failing width detection. Consider the following case:
> 
> ```
> module top(input i, output o);
> wire [1:0] t [0:0]; 
> assign o = &((0 == 0) ? i : t[-1]);
> endmodule
> ```
> 
> This should tie `o` to constant zero.

So on master (46ed0db2e) I get the following warning:
```
<<EOT:3: ERROR: Failed to resolve identifier \t[-1] for width detection!
```
I think we can agree this is not correct.

The intention of this PR is to allow this failure, and thus adopt the width from the other branch. Your comment above states that this is not correct, and the SEDA suite would agree with you. Can you explain how you get an out-of-bounds access will result in zero (and what size should that zero be?)

clairexen:
> I think we can agree this is not correct.

But your PR synthesizes this as `o = i`, which also is not correct. And I rather have an error message than an incorrect synthesis result.

> Can you explain how you get an out-of-bounds access will result in zero

Because `t[-1]` is two bits it forces the size of the `?:` operator to be two bits. Because the condition is constant true this means `(0 == 0) ? i : t[-1]` is just `{1'b0, i}`. The AND-reduce will thus return a constant zero bit because the MSB of it's input is always zero.

If you do not correctly detect the width of `t[-1]` then you can not determine the size of the `?:` operator, thus can not decide if you have to zero-pad the value to evaluate this expression correctly.

>  (and what size should that zero be?)

The result of a self-sized reduce operation is always 1 bit.

----

```
$ cat test.v
wire [1:0] t [0:0]; 
assign o = &((0 == 0) ? i : t[-1]);
endmodule
```

```
$ yosys -p 'verific -sv test.v; prep -top top; dump'
...
  connect \o 1'0
...
```

```
$ ./yosys-pr1651 -p 'read_verilog test.v; prep -top top; dump'
...
  connect \o \i
...
```

eddiehung:
> > I think we can agree this is not correct.
> 
> But your PR synthesizes this as `o = i`, which also is not correct. And I rather have an error message than an incorrect synthesis result.

I understand that my PR is not the correctly answer, which is why I've moved this into draft state. Ideally, we would want to have it behave correctly -- which is even better than an error message -- and this is what I'm striving towards by continuing this conversation. If there is no motivation then let's just close this ticket.

> > Can you explain how you get an out-of-bounds access will result in zero
> 
> Because `t[-1]` is two bits it forces the size of the `?:` operator to be two bits. Because the condition is constant true this means `(0 == 0) ? i : t[-1]` is just `{1'b0, i}`. The AND-reduce will thus return a constant zero bit because the MSB of it's input is always zero.
> 
> If you do not correctly detect the width of `t[-1]` then you can not determine the size of the `?:` operator, thus can not decide if you have to zero-pad the value to evaluate this expression correctly.

As I understand, there doesn't seem to be an elegant way to recover the `t` identifier in order to detect its width, so it looks like forward progress is being blocked by that.

arcturusannamalai:
Good concept and code; I would also recommend we can absorb this peephole optimization/check during constant folding pass which it doesn't seem to be done now.

daglem:
Yosys now correctly handles the examples above, so perhaps this PR should simply be closed?

PR #1657, opened 2020-01-27
synth_xilinx: add -dsp-multonly

daveshah1:
This is intended to get some kind of DSP usage - simple combinational multiplies only - working with open source flows, without the complexity of dealing with cascades, etc, to start with. It is a bit of a hack so happy for this not to go upstream if you prefer.


daveshah1:
That makes sense. Should the check for the scratchpad option be in synth_xilinx or in xilinx_dsp itself (that way a subset of options could eventually be added, like registers but not cascading)?

eddiehung:
Good point. Let's stick it in `xilinx_dsp` then. Perhaps just let it do the extra work pattern matching extra work and then return early from https://github.com/YosysHQ/yosys/blob/6fd9cae5ca5c7b154371f5cd328a7d96ef00f259/passes/pmgen/xilinx_dsp.cc#L259.

PR #1659, opened 2020-01-27
Add log_experimental() and experimental() API and "yosys -x"

cliffordwolf:


kernel/log.h
```diff
@@ -50,7 +50,7 @@ extern std::vector<FILE*> log_files;
 extern std::vector<std::ostream*> log_streams;
 extern std::map<std::string, std::set<std::string>> log_hdump;
 extern std::vector<std::regex> log_warn_regexes, log_nowarn_regexes, log_werror_regexes;
-extern std::set<std::string> log_warnings;
+extern std::set<std::string> log_warnings, log_experimentals;
```
mwkmwkmwk: It does the opposite of what its name suggests, maybe let's call it `log_experimental_ignores`?

cliffordwolf: hmm... the main idea is to just filter out duplicates, like log_warnings. the fact that `-x` also uses this is more a hack. maybe `-x` should use a different set? See updated PR.

PR #1661, opened 2020-01-27
abc9: add support for required times

eddiehung:
With arrival times (clock-to-q time) added in #1312, this adds support for required (setup) times using a similar interface (an `(* abc9_required=<positive-int> *)` attribute). Since `abc9` does not fully support this, emulate it using dynamically generated 1-input 1-output blackboxes.

With this, the box entries for `abc9_flop`-s are also dynamically generated too from arrival/required specifications.


ZirconiumX:
So, this probably seems like an odd place to put it, but I can't think of anywhere better to ask.

Looking through the changes here, you've added setup times to the (mostly-sanely-designed) Xilinx RAM primitives by attaching an `(* abc9_required *)` attribute to the address inputs (if I'm reading this right).

In Intel/Altera land, things are not so straightforward. We have the `altsyncram` megafunction to represent memories, which synthesis lowers to target-specific `XXX_ram_block` primitives for BRAMs and `XXX_mlab_cell` for LUTRAM. These are still highly generic functions, however: the RAM dimensions are parametric, for example.

The most naive approach would be to very manually map these individual RAM modes into separate blackboxes, one for each mode, but an MLAB has three address widths, an M10K (Cyclone V BRAM) has ten address/data combinations, and an M20K (Cyclone 10GX BRAM) has another ten address/data combinations. These are then multiplied by two for the synchronous/combinational memory modes. This would be a nightmare to maintain.

Perhaps there could be an `abc9_memory_{width,depth}=<PARAM>` and `abc9_arrival_if=(width=foo,depth=bar) 1 2 3` type thing. You'd probably have a better design for it than I would, but it's the kind of problem I have at present.

eddiehung:
@ZirconiumX Take a look at how DSP arrival/required times are handled here using a `constexpr` function approach: https://github.com/YosysHQ/yosys/blob/be8bc63f8443ea0aa2261bbbe4f30de476fa4a61/techlibs/xilinx/cells_sim.v#L2553-L2555

ZirconiumX:
Would a combinational memory be better represented with boxes or with arrival times?

eddiehung:
> Would a combinational memory be better represented with boxes or with arrival times?

Arrival/required times are only applicable to sequential cells like flops or BRAMs. 

Comb cells must be modelled as boxes, and conversely so -- boxes can only model comb cells.
In the case of hybrid comb/seq cells (LUTRAMs are the obvious case) then acrobatics need to be done, see: 
https://github.com/YosysHQ/yosys/blob/be8bc63f8443ea0aa2261bbbe4f30de476fa4a61/techlibs/xilinx/abc9_map.v#L378-L402
which wraps the original cell (all non-box cells are treated as sequential) with a comb box to capture the comb path.

Let me know if you have any other questions!

eddiehung:
@ZirconiumX Just to let you know I'm slowly rethinking seq synthesis. There are some two issues that make it deserve it:
- In `abc9` flows, flops are mapped to tech specific primitives before calling `abc9`. This is not the most efficient as we then have to extract this tech specific behaviour and describe it to ABC, and even after it comes back out of ABC it blocks downstream Yosys optimisations. 
- Extracting this tech specific behaviour is messy and requires a lot of manual labour, making supporting new architectures more difficult than it needs to be.
- Since `abc9`'s sequential synthesis requires that the initial state of the flop is zero, I currently use soft-logic (inefficiently) to enforce that, and I'm looking at doing more acrobatics to eliminate that.
In summary, I would suggest de-prioritising adding seq synthesis support, if that is what you're looking at doing, until I have a more solid proposal we can discuss.

ZirconiumX:
Noted. ABC9 has always felt rather like a black box to me, not only in
terms of code (e.g. I needed Dave to explain what -W does and how the lut
and box formats work), but also in terms of process (most ABC9 PRs - which
Intel synthesis vitally depends on for good results - appear out of the
blue to me), so I appreciate the plans and insight.

Many of the papers on these subjects feel like indecipherable gobbledegook
to me, since I never got beyond high-school math. Much of this behaviour
feels like it could be more easily implemented in flowmap, but the paper is
impenetrable.

PR #1665, opened 2020-01-29
Preserve wires with keep attribute in EDIF back-end

cliffordwolf:
Make sure we preserve "keep" nets in EDIF back-end. Requires testing @mmicko.


mmicko:
Needed to add fixes from 245b8c4ab64c5c3bd7b9f71f94316a76a2576fd1 to build fine.
EDIF now do contain all needed wire names.

PR #1666, opened 2020-01-29
Makefile improvements for packaging scripts

Xiretza:
While working on the [yosys-git](https://aur.archlinux.org/packages/yosys-git/) AUR package, I noticed a few hickups in the make system that cause some trouble for packaging. Apart from fixes for build failures, the only observable change this PR should have is that `libyosys.so` is no longer installed to `BINDIR`, and the copy installed to `PYTHON_DESTDIR` is a symlink pointing to `LIBDIR` instead. Also, a 30% reduction in package size.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -763,8 +763,7 @@ clean-unit-test:
 	@$(MAKE) -C $(UNITESTPATH) clean
 
 install: $(TARGETS) $(EXTRA_TARGETS)
-	$(INSTALL_SUDO) mkdir -p $(DESTDIR)$(BINDIR)
-	$(INSTALL_SUDO) cp $(TARGETS) $(DESTDIR)$(BINDIR)
+	$(INSTALL_SUDO) install -D -t $(DESTDIR)$(BINDIR) $(filter-out libyosys.so,$(TARGETS))
 ifneq ($(filter yosys,$(TARGETS)),)
 	$(INSTALL_SUDO) $(STRIP) -S $(DESTDIR)$(BINDIR)/yosys
 endif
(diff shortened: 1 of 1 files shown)


mmicko:
@Xiretza Have to check how this affect mingw build, since not sure if install is actually available there, also making symbolic links are not working, so assumption is that current way is there for that reason.

mmicko:
Have tested this on Windows and there were no issues with other parts except symbolic link, so will just add additional commit check if mingw configuration is used and do copy in that case and not linking

mmicko:
just a note: windows build of python library is not working anyway out of box, have made hack changes in order to build and ln -s will in msys case create a copy of file so no need to change that anyway.

mmicko:
@Xiretza Sorry, had to revert this, since this is breaking macOS install. 
This https://github.com/YosysHQ/yosys/commit/a0cc795e85541b0326b6d4396a726142f0d0f8bb was useful so have added it after to Makefile

Xiretza:
What excatly was causing the problem? Could the ABCREV stuff be added back again as well?

mmicko:
Install -D does not exist on macOS, and it was intention to fail when git errors out, but in case you do not want this git check to be done just set ABCREV=default

Xiretza:
Ah, fair enough about `install -D`. Maybe packaging could be added to the CI so something like this can be caught early?

Regarding ABC though: isn't that exactly what my patch does? If git fails to run, the assignment to `rev` fails, then the whole pipeline fails, and ABC is fetched.

Makefile
```diff
@@ -668,7 +668,8 @@ ifneq ($(ABCREV),default)
 	$(Q) if ( cd abc 2> /dev/null && ! git diff-index --quiet HEAD; ); then \
 		echo 'REEBE: NOP pbagnvaf ybpny zbqvsvpngvbaf! Frg NOPERI=qrsnhyg va Lbflf Znxrsvyr!' | tr 'A-Za-z' 'N-ZA-Mn-za-m'; false; \
 	fi
-	$(Q) if test "`cd abc 2> /dev/null && git rev-parse --short HEAD`" != "$(ABCREV)"; then \
+	# set a variable so the test fails if git fails to run - when comparing outputs directly, empty string would match empty string
+	$(Q) if ! { cd abc && rev="`git rev-parse $(ABCREV)`" && test "`git rev-parse HEAD`" == "$$rev"; }; then \
```
Xiretza: Oh, I just found one potential issue: if `make` runs its commands in a shell that doesn't support group commands, the `cd` is executed in the context of the top-level shell. Using a subshell instead solves the problem:

```
if ! (cd abc && rev="`git rev-parse $(ABCREV)`" && test "`git rev-parse HEAD`" == "$$rev"); then
```

mmicko:
Can you please create separate pull request and explain intention on ABCREV change, so @cliffordwolf can check that ?

Xiretza:
Done in #1764. The only other thing that's neither in the new PR nor reintroduced in master is linking `libyosys.so` to the python directory instead of copying it - I think that should be supported on all platforms (with windows simply copying it anyway).

mmicko:
Thanks for adding that. Will check tomorrow with @cliffordwolf and do some more testing

PR #1668, opened 2020-01-30
abc9: Fix regression breaking support for use of ABCEXTERNAL

gsomlo:


passes/techmap/Makefile.inc
```diff
@@ -13,6 +13,7 @@ OBJS += passes/techmap/abc9_ops.o
 ifneq ($(ABCEXTERNAL),)
 passes/techmap/abc.o: CXXFLAGS += -DABCEXTERNAL='"$(ABCEXTERNAL)"'
 passes/techmap/abc9.o: CXXFLAGS += -DABCEXTERNAL='"$(ABCEXTERNAL)"'
```
eddiehung: Do we still need this line?

gsomlo: Do you mean the one for `abc9.o`? There still is an `#ifdef ABCEXTERNAL` in there, but only for the `help()` message. So in absolute terms, the answer is "yes", unless you can (or want/need to) remove the remaining `#ifdef` from `techmap/abc9.cc`.

eddiehung: Good thinking. I forgot we had that line!

PR #1669, opened 2020-01-30
Make RTLIL attributes accessible via pyosys

thasti:
Currently, the pyosys wrapper generator disregards any information on inheritance. As most of the RTLIL classes (Module, Cell, Wire, ...) derive their attribute capabilites from AttrObject, their accessors are lost during wrapper generation.

This patch adds a basic notion of inheritance to the wrapper generation process. Now one can do:

```
[thasti@pc]$ cat test.v 
module test();
    (* my_attr = 1 *) wire attrwire;
endmodule

[thasti@pc]$ cat yo.py 

design = ys.Design()
ys.run_pass("read_verilog test.v", design)
mod = design.module(ys.IdString("\\test"))
print("Module Source Attribute: %s " % mod.get_src_attribute())
wire = mod.wire(ys.IdString(ys.IdString("\\attrwire")))
print("All Wire Attributes: %s" % wire.attributes)

[thasti@pc]$ python yo.py 

-- Running command `read_verilog test.v' --

Module Source Attribute: test.v:1 
All Wire Attributes: {IdString "\my_attr": Const "00000000000000000000000000000001", IdString "\src": Const "0111010001100101011100110111010000101110011101100011101000110010"}
[thasti@pc]$ 
```

### DIFF
--- a/misc/py_wrap_generator.py
+++ b/misc/py_wrap_generator.py
@@ -914,7 +914,7 @@ class WClass:
 		body = self.gen_boost_py_body()
 		base_info = ""
 		if self.base_class is not None:
-			base_info = ", bases<%s::%s>" % (self.base_class.namespace, self.base_class.name)
+			base_info = ", bases<" + (self.base_class.name) + ">"
 
 		if self.link_type == link_types.derive:
 			text = "\n\t\tclass_<" + self.name + base_info + ">(\"Cpp" + self.name + "\""
--- a/misc/py_wrap_generator.py
+++ b/misc/py_wrap_generator.py
@@ -721,6 +721,7 @@ class WClass:
 	name = None
 	namespace = None
 	link_type = None
+	base_class = None
 	id_ = None
 	string_id = None
 	hash_id = None
@@ -732,6 +733,7 @@ class WClass:
 	def __init__(self, name, link_type, id_, string_id = None, hash_id = None, needs_clone = False):
 		self.name = name
 		self.namespace = None
+		self.base_class = None
 		self.link_type = link_type
 		self.id_ = id_
 		self.string_id = string_id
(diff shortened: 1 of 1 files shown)


thasti:
@btut could you perhaps take a look at this one?

btut:
Looks good to me. Great to have attributes available through Python now as well, thanks!

I'm wondering whether it is possible to expose the class hierarchy to Python too. As far as I can tell with your solution there is no way to tell that a Wire is derived from an AttrObject, but I guess that would be a lot of work to achieve and not really worth it, right? With Pythons dynamic typing there is no real benefit anyway.

thasti:
Thanks for reviewing! I don't quite know if/how that would be possible, I'm not quite an expert in boost-python things. I agree that having things like isinstance(some_wire, ys.AttrObject) would be nice to have, but currently don't see an easy way to get there.

btut:
I played around a little and this seems easy enough to achieve. Two steps are needed:

- Classes must be polymorphic (add virtual destructor to all inheritors)
- Declare the inheritance in the boost::python section (instead of i.e. `class_<Wire>("Wire", no_init)` write `class_<Wire, bases<AttrObject>>("Wire", no_init)`

I tested this out by just hacking it up in the generated python_wrappers.cc file and it works (isinstance(wire, ys.AttrObject) returns true for your code above), it would be great if you could amend the generator script to generate this code.

Another thing that came to mind is multiple inheritance. Again, this is probably not needed as Yosys does not use multiple inheritance, but who knows whether that changes in the future...

thasti:
That actually did the trick, thanks for the pointers!

```
[thasti@pc]$ cat test.v 
module test();
    (* my_attr = 1 *) wire attrwire;
endmodule

[thasti@pc]$ cat yo.py 

design = ys.Design()
ys.run_pass("read_verilog test.v", design)
mod = design.module(ys.IdString("\\test"))
print("Module Source Attribute: %s " % mod.get_src_attribute())
wire = mod.wire(ys.IdString(ys.IdString("\\attrwire")))
print("All Wire Attributes: %s" % wire.attributes)

print("Wire is a ys.Wire?: %s" % isinstance(wire, ys.Wire))
print("Wire is a ys.Module?: %s" % isinstance(wire, ys.Module))
print("Wire is a ys.AttrObject?: %s" % isinstance(wire, ys.AttrObject))

[thasti@pc]$ python yo.py 
Module Source Attribute: test.v:1 
All Wire Attributes: {IdString "\my_attr": Const "00000000000000000000000000000001", IdString "\src": Const "0111010001100101011100110111010000101110011101100011101000110010"}
Wire is a ys.Wire?: True
Wire is a ys.Module?: False
Wire is a ys.AttrObject?: True
```

Re multiple inheritance: That would require some more involved parsing of the class declaration line, with quite a number of special cases to be taken considered. If acceptable, I'd prefer not to go there for this PR.

But while on the subject: I am under the impression pyosys is not part of Travis at the moment - would at least some basic tests (such as this one) to confirm 'it still compiles and imports correctly' be a useful addition to catch such problems as they arise?

btut:
Multiple inheritance was just an Idea, I don't think it is necessary for this PR.

Yes, adding Travis tests would make sense. I can't remember why we decided not to include them in the inital pyosys-PR.

thasti:
Thanks for approval! I may add something to Travis in a separate PR in the future when I get around to preparing some more useful testcases.

PR #1670, opened 2020-01-31
$readmem[hb] file inclusion is now relative to the Verilog file

rodrigomelo9:
Until now, the $readmem[hb] implementation looking for a memory content file in a relative path to the Yosys execution. Now, it is relative to the Verilog file, which in case of being in the path where Yosys is executed, it also works. I checked that at least ISE and Vivado works in this manner.

Here a [test.tar.gz](https://github.com/YosysHQ/yosys/files/4141641/test.tar.gz). Running `yosys ram.ys` without this change, give you:
```hdl/ram.v:13: ERROR: Can not open file `data/memory.dat` for \$readmemb.```

Let me know if you would like to have this test (and please, point me where to put).

This solution seems to me at least an improvement. I am not an experienced C++ developer, so suggestions for a better implementation are welcome.

### CODE REVIEW

**Comment by daveshah1** [comment]
Although this is in itself a reasonable idea, the problem here is that only looking relative to the source file will break existing projects that relied on the old behaviour. Perhaps more of a "search path" approach would be better.

**Comment by rodrigomelo9** [comment]
Hi @daveshah1, are there an example in Yosys or you are talking in a hypothetical way?

Anyway, what about this solution as default and if the file is not found, the old one? Or vice versa, as you wish. Probably, also the specification of where to search could be useful, but I think that I have not idea how to implement it :P

**Comment by rodrigomelo9** [comment]
I modified the code to be backward compatible:

* Old behaviour: the memory content file is searched in a relative location according to the execution directory.
* New behaviour: if the file is not founded in the first location, is searched relative to the Verilog file location.

@cliffordwolf I have doubts about how I used `yosys_input_files.insert`. Please, take a look.

File: CHANGELOG [hunk 1]
```diff
@@ -53,12 +53,13 @@ Yosys 0.9 .. Yosys 0.9-dev
     - Added support for flip-flops with synchronous reset to synth_xilinx
     - Added support for flip-flops with reset and enable to synth_xilinx
     - Added "check -mapped"
-    - Added checking of SystemVerilog always block types (always_comb,
+    - Added checking of SystemVerilog always block types (always_comb,
```
cliffordwolf: Do we really want to force a line break in the rendered output here?

frontends/ast/simplify.cc
```diff
@@ -2886,9 +2886,14 @@ AstNode *AstNode::readmem(bool is_readmemh, std::string mem_filename, AstNode *m
 	int meminit_size=0;
 
 	std::ifstream f;
+	std::string path = filename.substr(0, filename.find_last_of("\\/")+1);
```
cliffordwolf: What happens here if `filename` does not contain a `/`?

CHANGELOG
rodrigomelo9: Sorry. I didn't make the line break, but I added the space accidentally (because I also thought that the line jump is not needed and checked the line width). How do you prefer that I fix the problem? Removing the line jump in a new commit is ok?

frontends/ast/simplify.cc
rodrigomelo9: Mmm, I suppose that something will be broken XD (good catch). I will test and fix it!

eddiehung: Are Windows paths considered by this new code too?

rodrigomelo9: @eddiehung I have not Windows near to check (I am a GNU/Linux user) but I did my best:
* `find_last_of("\\/")` is to take in account the both possible slashes.
* `+1` is to use the slash provided bt the filename.
* I used `std::string` (I understand that is supported in both OS)

@cliffordwolf it works also when `filename` does not contain a slash. I added `test/memfile`. With the old implementation:

```
$ ./run-test.sh 
* Creating Memory Content Files
* Running from the parent directory
  * Memory Content File: tempfile1.dat
memfile/memory.v:15: ERROR: Can not open file `tempfile1.dat` for \$readmemb.
  * Memory Content File: temp/tempfile2.dat
memfile/memory.v:15: ERROR: Can not open file `temp/tempfile2.dat` for \$readmemb.
* Running from the same directory
  * Memory Content File: tempfile1.dat
  * Memory Content File: temp/tempfile2.dat
* Running from a child directory
  * Memory Content File: tempfile1.dat
../memory.v:15: ERROR: Can not open file `tempfile1.dat` for \$readmemb.
  * Memory Content File: temp/tempfile2.dat
../memory.v:15: ERROR: Can not open file `temp/tempfile2.dat` for \$readmemb.
  * Memory Content File: tempfile2.dat
../memory.v:15: ERROR: Can not open file `temp/tempfile2.dat` for \$readmemb.
* Done
```

And with the new feature:

```
$ ./run-test.sh 
* Creating Memory Content Files
* Running from the parent directory
  * Memory Content File: tempfile1.dat
  * Memory Content File: temp/tempfile2.dat
* Running from the same directory
  * Memory Content File: tempfile1.dat
  * Memory Content File: temp/tempfile2.dat
* Running from a child directory
  * Memory Content File: tempfile1.dat
  * Memory Content File: temp/tempfile2.dat
  * Memory Content File: tempfile2.dat
* Done
```

The test is not included to be run automatically. Let me know all the needed to be accepted.

Regards

frontends/ast/simplify.cc
```diff
@@ -2886,10 +2886,15 @@ AstNode *AstNode::readmem(bool is_readmemh, std::string mem_filename, AstNode *m
...
-	yosys_input_files.insert(mem_filename);
-
-	if (f.fail())
+	if (f.fail()) {
+		f.open(path + mem_filename.c_str());
+		yosys_input_files.insert(path + mem_filename);
+	} else {
+		yosys_input_files.insert(mem_filename);
+	}
+	if (f.fail() || strlen(mem_filename.c_str()) == 0)
```
rodrigomelo9: I added `strlen(mem_filename.c_str()` to avoid a hung when the included file is "" and `path != ""`. As I said, I am not a C++ expert, so maybe there is an option to fail when not a file? (I didn't find in a quick search).

CHANGELOG
rodrigomelo9: I removed the line jump (there are longest lines than the resulting).

frontends/ast/simplify.cc
rodrigomelo9: I added it to `make test` with an output format similar to the other tests.

tests/memfile/run-test.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+for i in {1..64}; do
+	echo "00001111000000001111111100000000" >> tempfile1.dat
```
eddiehung: Wouldn't this keep appending to `tempfile1.dat` on each run? Why not just commit this file?

frontends/ast/simplify.cc
eddiehung: > * `find_last_of("\\/")` is to take in account the both possible slashes.

Good point. I misread that.. I think it should be okay. There's no good reason why you should have a forward slash in Windows paths, or a backwards slash in Linux paths. The alternative would be an `#ifdef` to decide what the character to search for.

tests/memfile/run-test.sh
```diff
@@ -0,0 +1,33 @@
...
+done
+
+mkdir -p temp
+cp tempfile1.dat temp/tempfile2.dat
+
+cd ..
+
+echo "Running from the parent directory with tempfile1.dat"
+../yosys -qp "read_verilog -defer memfile/memory.v; chparam -set MEMFILE \"tempfile1.dat\" memory; synth -top memory"
```
eddiehung: Do we need to do `synth` at all, since any error should occur in the front-end, no?

tests/memfile/run-test.sh
rodrigomelo9: XD I deleted a previous rm of the file :S Ok, I will add tempfile1.dat but tempfile2.dat will be still a copy into temp, to simulate a file in another directory.

tests/memfile/run-test.sh
rodrigomelo9: I am new with Yosys. I thought that because of I used `-defer`, I needed something to force the use of `chparam`, but seems not. I remove the `synth` and it works.

With yosys compiled without this new feature:

```
Running from the parent directory with content1.dat
memfile/memory.v:15: ERROR: Can not open file `content1.dat` for \$readmemb.
Running from the parent directory with temp/content2.dat
memfile/memory.v:15: ERROR: Can not open file `temp/content2.dat` for \$readmemb.
Running from the parent directory with memfile/temp/content2.dat
Running from the same directory with content1.dat
Running from the same directory with temp/content2.dat
Running from a child directory with content1.dat
../memory.v:15: ERROR: Can not open file `content1.dat` for \$readmemb.
Running from a child directory with temp/content2.dat
../memory.v:15: ERROR: Can not open file `temp/content2.dat` for \$readmemb.
Running from a child directory with content2.dat
../memory.v:15: ERROR: Can not open file `temp/content2.dat` for \$readmemb.
```

With Yosys compiled with this new feature:

```
Running from the parent directory with content1.dat
Running from the parent directory with temp/content2.dat
Running from the parent directory with memfile/temp/content2.dat
Running from the same directory with content1.dat
Running from the same directory with temp/content2.dat
Running from a child directory with content1.dat
Running from a child directory with temp/content2.dat
Running from a child directory with content2.dat
```

tests/memfile/run-test.sh
rodrigomelo9: FYI. Because is not more a `tempfile` I changed the name to content1.dat and content2.dat.

frontends/ast/simplify.cc
daveshah1: Please use `GetSize(mem_filename) == 0` rather than `strlen(mem_filename.c_str()) == 0`

frontends/ast/simplify.cc
rodrigomelo9: I moved the path detection into the `if` where it is used:

```
	if (f.fail()) {
		std::string path = filename.substr(0, filename.find_last_of("\\/")+1);
		f.open(path + mem_filename.c_str());
		yosys_input_files.insert(path + mem_filename);
	} else {
```

Do you prefer the use of `#ifdef` as follows? 

```
	if (f.fail()) {
#ifdef _WIN32
		std::string slash = "\\";
#else
		std::string slash = "/";
#endif
		std::string path = filename.substr(0, filename.find_last_of(slash)+1);
		f.open(path + mem_filename.c_str());
		yosys_input_files.insert(path + mem_filename);
	} else {
```

frontends/ast/simplify.cc
rodrigomelo9: Done. I saw that it is specified in the CodingReadme, sorry.

CHANGELOG
rodrigomelo9: There was a conflict due to changes in master. I solved it reverting the change in that line.

frontends/ast/simplify.cc
eddiehung: Now that you're offering it: sure! I think that can only improve things as opposed to make things worse.
The other minor nitpick is to use `char slash = ` since you really are only searching for one character as opposed to multiple.

rodrigomelo9: Done. Tested and working, at least with Linux :-D

tests/memfile/run-test.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+
+set -e
```
rodrigomelo9: @eddiehung FYI: I added this and two more checks at the bottom of the file (for situations where the execution must fail).

rodrigomelo9: @eddiehung it says that some checks haven't completed but if you enter to the details all the jobs had finished. There was a problem getting an Ubuntu package, it was re-launched, passed, but is marked still as pending.

frontends/ast/simplify.cc
```diff
@@ -2904,7 +2904,12 @@ AstNode *AstNode::readmem(bool is_readmemh, std::string mem_filename, AstNode *m
 	std::ifstream f;
 	f.open(mem_filename.c_str());
 	if (f.fail()) {
-		std::string path = filename.substr(0, filename.find_last_of("\\/")+1);
+#ifdef _WIN32
+		char slash = '\\';
```
whitequark, line 2908: This is actually incorrect, since on Windows you can use both slashes, and this is useful in portable code.

PR #1673, opened 2020-02-01
Add opt_lut_ins pass.

mwkmwkmwk:


tests/arch/ecp5/opt_lut_ins.ys
```diff
@@ -0,0 +1,36 @@
...
+    connect \B \B
+    connect \C \C
+    connect \D \D
+    connect \Z \Z
+  end
+end
+
+EOF
+
+write_verilog crap.v
```
eddiehung: No need?

tests/arch/xilinx/opt_lut_ins.ys
```diff
@@ -0,0 +1,27 @@
...
+    connect \I1 \A [1]
+    connect \I2 \A [2]
+    connect \I3 \A [3]
+    connect \O \O
+  end
+end
+
+EOF
+
+write_verilog crap.v
```
eddiehung: No need?

tests/arch/ecp5/opt_lut_ins.ys
```diff
@@ -0,0 +1,36 @@
...
+
+read_verilog -lib +/ecp5/cells_sim.v
+
+equiv_opt -assert -map +/ecp5/cells_sim.v opt_lut_ins -tech ecp5
+
+design -load postopt
+
+dump
```
eddiehung: No need?

eddiehung:
For the history books, can you drop a few lines about when you've seen this optimisation get be required? On the surface, one would not expect ABC to return a LUT mapping with unused inputs...

ZirconiumX:
To be honest I think `-tech` is a poor design choice: as the codebase gets older, it'll likely be considered a "good idea" to add FPGA families to it, making the codebase unnecessarily ratty. The only technology-specific hack I see is the ECP5 minimum-LUT size. Since ECP5 techmap maps these into `LUT4` cells with constant drivers anyway, I see little need for it, providing this is called after `abc9` and before `techmap`ping into LUT cells.

ZirconiumX:
Okay, reading it again and reflecting on it, it seems the purpose of this is that *after* techmapping, certain LUTs don't care about some inputs, and thus those inputs can be removed (so the cells pack better?). This still feels like it's insufficiently generic, honestly, but I'm not smart enough to figure out how to better approach this.

mwkmwkmwk:
@ZirconiumX The long-term plan we came up with at the meeting is that we should move LUT techmapping to a later stage in a flow, keeping `$lut` for as long as possible, making more passes use generic cells (in particular `xilinx_dffopt` and wide LUT breakdown), thus removing the need for  the `-tech` parameter.  However, this requires some infrastructure we don't currently have (in particular, new FF types).

whitequark:
> The long-term plan we came up with at the meeting is that we should move LUT techmapping to a later stage in a flow, keeping `$lut` for as long as possible

That would be really nice to see.

eddiehung:
> To be honest I think `-tech` is a poor design choice: as the codebase gets older, it'll likely be considered a "good idea" to add FPGA families to it, making the codebase unnecessarily ratty. The only technology-specific hack I see is the ECP5 minimum-LUT size. Since ECP5 techmap maps these into `LUT4` cells with constant drivers anyway, I see little need for it, providing this is called after `abc9` and before `techmap`ping into LUT cells.

@mwkmwkmwk can likely explain better, but the need for this appears to be driven by two events, which have lead to unnecessary routing as well as DRC errors:
* Mapping wide `$lut`s into tech-specific LUTs plus muxes can sometimes leave unused inputs behind
* `xilinx_dffopt` can also leave unused inputs behind, this must be called after this

I agree the general strategy should be to stay generic as long as possible which is why this pass does work on `$lut` cells too, but we should move towards that goal in small manageable PRs while continuing to move forward with new enhancements.

mwkmwkmwk:
There's a lot of dependencies here:

- `xilinx_dffopt` was written as target specific, because we don't yet have generic FF types with matching behavior
- we need `opt_lut_ins` to run after `xilinx_dffopt` (because `xilinx_dffopt` can produce LUTs with redundant inputs)
- we also need `opt_lut_ins` to run after LUT mapping (so that wide LUTs are lowered to possibly-redundant sub-LUTs and multiplexers), which produces target-specific LUTs

All that can be solved:

- split the LUT mapping into two steps: mapping wide `$lut` into smaller `$lut` and technology-specific (or maybe even generic?) wide mux cells (which would happen before `opt_lut_ins`) and afterwards mapping small `$lut` into target-specific LUT types (after `opt_lut_ins`)
- make `xilinx_dffopt` operate on generic `$lut`s mixed with target-specific FF types, to fit in with the above flow

However, that will require possibly-invasive changes to pass ordering and splitting up the `*_map.v` files, which IMO should be done in separate PRs.  Thus, this pass supports the `-tech` option (which should be considered transitional, and can be removed when we no longer have a need for it).

mwkmwkmwk:
Another point raised by @daveshah1 on IRC: for some targets (ecp5) we need to avoid optimizing away the LUT to a contant driver (because the wide mux code in nextpnr requires a complete set of LUTs). While this can be fixed for this particular case, we should try to maintain compatibility with older nextpnr versions, and we may come across other targets requiring similar behavior.  I suppose in a generic flow we would model this case with a pass option and a 0-width `$lut`?

PR #1677, opened 2020-02-03
xilinx: Initial support for LUT4 devices.

mwkmwkmwk:


techlibs/xilinx/synth_xilinx.cc
```diff
@@ -82,10 +97,10 @@ struct SynthXilinxPass : public ScriptPass
 		log("        do not use XORCY/MUXCY/CARRY4 cells in output netlist\n");
 		log("\n");
 		log("    -nowidelut\n");
-		log("        do not use MUXF[78] resources to implement LUTs larger than LUT6s\n");
+		log("        do not use MUXF[56789] resources to implement LUTs larger than LUT4s/LUT6s\n");
```
ZirconiumX: I feel like `do not use MUXF[56789] resources to implement LUTs larger than native to hardware` is less confusing.

techlibs/xilinx/lut4_arith_map.v
```diff
@@ -0,0 +1,101 @@
+/*
```
eddiehung: How does the arithmetic mapping for LUT4 differ from LUT6? Do LUT4 architectures not accept CARRY4?

techlibs/xilinx/synth_xilinx.cc
```diff
@@ -49,10 +49,25 @@ struct SynthXilinxPass : public ScriptPass
...
 		log("\n");
-		log("    -family {xcup|xcu|xc7|xc6v|xc5v|xc6s}\n");
+		log("    -family <family>\n");
 		log("        run synthesis for the specified Xilinx architecture\n");
 		log("        generate the synthesis netlist for the specified family.\n");
-		log("        default: xc7\n");
+		log("        supported values:\n");
+		log("        - xcup: Ultrascale Plus\n");
+		log("        - xcu: Ultrascale\n");
+		log("        - xc7: Series 7\n");
```
eddiehung: Can we restore doc pointing to "xc7" as the default value?

techlibs/xilinx/synth_xilinx.cc
```diff
@@ -49,10 +49,25 @@ struct SynthXilinxPass : public ScriptPass
...
+		log("        - xc6s: Spartan 6\n");
+		log("        - xc6v: Virtex 6\n");
+		log("        - xc5v: Virtex 5 (EXPERIMENTAL)\n");
+		log("        - xc4v: Virtex 4 (EXPERIMENTAL)\n");
+		log("        - xc3sda: Spartan 3A DSP (EXPERIMENTAL)\n");
```
eddiehung: Is there enough of a difference between the Spartan3 families to warrant a separate entry?

techlibs/xilinx/synth_xilinx.cc
```diff
@@ -49,10 +49,25 @@ struct SynthXilinxPass : public ScriptPass
...
+		log("        - xc3sa: Spartan 3A (EXPERIMENTAL)\n");
+		log("        - xc3se: Spartan 3E (EXPERIMENTAL)\n");
+		log("        - xc3s: Spartan 3 (EXPERIMENTAL)\n");
+		log("        - xc2vp: Virtex 2 Pro (EXPERIMENTAL)\n");
```
eddiehung: Also for Virtex 2?

techlibs/xilinx/synth_xilinx.cc
eddiehung: Or `do not use MUXF[56789] resources to implement LUTs larger than supported natively`?

techlibs/xilinx/synth_xilinx.cc
```diff
@@ -270,9 +288,38 @@ struct SynthXilinxPass : public ScriptPass
...
+			widelut_size = 6;
+		} else
 			log_cmd_error("Invalid Xilinx -family setting: '%s'.\n", family.c_str());
 
+		if (widemux != 0 && lut_size != 6)
+			log_cmd_error("-widemux is not currently supported for LUT4-based architectures.\n");
+
+		if (lut_size != 6) {
+			// Features not yet supported for LUT4-based architectures.
+			nosrl = true;
```
eddiehung: Please warn user

techlibs/xilinx/synth_xilinx.cc
eddiehung: I see from #1678 that Spartan 3/3E, Spartan 3A, Spartan 3A DSP all have different block RAM rules!

techlibs/xilinx/lut4_arith_map.v
mwkmwkmwk: Indeed, they do not (the slice on these architectures was 2 LUTs tall, while the "4" in CARRY4 refers to 4 LUTs in the LUT6-based slices)

techlibs/xilinx/synth_xilinx.cc
mwkmwkmwk: Correct, and Spartan 3 vs Spartan 3E/3A also has different multiplier primitives (which we don't currently make use of beyond the basics that are identical, but we should do so in the future).

techlibs/xilinx/synth_xilinx.cc
mwkmwkmwk: Virtex 2 should be identical to Virtex 2 Pro wrt synthesis (they differ by supported I/O standards and by the presence of gigabit transceivers and PPC cores), but I'd rather accept both values as valid for consistency.  Also, if we ever get timing information for them, it's likely to be different.

techlibs/xilinx/lut4_arith_map.v
eddiehung: Righty. Given there is so much shared code between them, would you consider merging them back by reverting 5c589244df2ec4fc5fde0bcdc69dee727f4b8e79 to go back to providing MUXCY/XORCY as an option?

mwkmwkmwk: There is a slight but important difference in `$alu` mapping between the old removed code and my LUT4 code: the old code used `AA & BB` as the `DI` input (which is fine for LUT6 architectures, as they have fracturable LUTs and can compute it using the `O5` output), while my LUT4 code uses `AA` input (which is the only reasonable selection for LUT4 architectures that doesn't result in extra LUTs somewhere else).  However, it seems to me that this would be a reasonable option for LUT6 architectures as well — should I just merge these two files and provide an option to use `MUXCY` for LUT6 architectures as well?

eddiehung: Hmm. Let me write out my thought pattern. Consider this diagram: https://www.researchgate.net/profile/Raffaele_Giordano4/publication/291019054/figure/fig4/AS:536738776838151@1504979904869/The-CARRY4-primitive-available-in-7-Series-Xilinx-FPGAs.png -- the `S` (sum) input is the select input to the `MUXCY` -- if the sum is 1, the carry is propagated (`CO = CI`); the `DI` input is the "generate", i.e. override the carry with this value. Since `S == 0` is only true if `AA == BB`, meaning that `DI = AA && BB` is redundant. We should definitely change it then!! Thanks for revisiting this.

Issue #1550, opened 2019-12-04
xilinx: Add support for Block RAM inference on remaining devices.


mwkmwkmwk:
- [x] Virtex 2, Spartan 3, Spartan 3E
- [x] Spartan 3A
- [x] Spartan 3A DSP
- [ ] Virtex 4
- [ ] Virtex 5

PR #1678, opened 2020-02-03
xilinx: Add support for Virtex 2* and Spartan 3* block RAMs.

mwkmwkmwk:
Part of #1550


techlibs/xilinx/xc3sda_brams.txt
```diff
@@ -0,0 +1,33 @@
+# Spartan 3A DSP block RAM rules.
+
+bram $__XILINX_RAMB16BWER_TDP
```
eddiehung: Is it more accurate to say that this is an `SDP` rule since Yosys doesn't currently support `TDP`?
I know this does not deviate from the norm and that this is necessary `xc6s_brams_map`, but do we need to update all the rules to be more accurate in this respect (doesn't mean it should be done here)

mwkmwkmwk: It is a SDP rule like all our rules are, but it's called TDP because it passes "TDP" as the RAM_MODE for the underlying primitive (which does not support any modes *other* than TDP, but the 8-kbit version does, and we use that in both variants).

Issue #1676, opened 2020-02-03
Preservation of escaped identifier


towoe:
SystemVerilog allows escaped identifiers in Section 5.6.1.
The rule is that they must start with an backslash and end with a white space. All printable ASCII characters are included in the name.

Yosys can read those identifiers, but when using `write_ilang` or `write_smt2` there is a problem if the identifiers are not in the top level of the hierarchy. This is because the path to the identifier is stored without the direct backslash of the identifier.

I attached a simple example which will generate a `.ilang` and `.smt2` file of the design.
Here is the gist of it.
```verilog
module top ();
  bottom UUT ();
endmodule

module bottom ();
  reg \escaped[2] ;
endmodule
```

The signal is then referred to as `\UUT.escaped[1]`.
I am not sure if this is correct or incorrect, or the result of some Yosys design choices.
But if this identifier is used by SymbiYosys, which has the ability to create a Verilog test bench, the resulting identifier is `UUT.escaped[1]`, which will be an invalid path for a simulation.
I included the `.smt2` file as this is the input to sby/Yosys and here both versions of the identifier are stored (`\UUT.escaped[1]` and `UUT.escaped[1]`). The one without any backslash is part of a Yosys specific comment which is used for the test bench creation.
My assumption is that Yosys removes all backslashes and uses one in general as a prefix?
This would be problematic if the original identifier needs to be recreated.

[escaped_name.zip](https://github.com/YosysHQ/yosys/files/4147990/escaped_name.zip)

whitequark:
It looks to me that the problem here is that SymbiYosys doesn't have an understanding of hierarchy. AFAIU, in SV, the signal should be referred to as `UUT.\escaped[1] `, i.e. with each level individually escaped. This is a problem because Yosys currently uses "in-band signaling" for hierarchy: an escaped identifier `\UUT.escaped[1]` and a hierarchical identifier `UUT.\escaped[1]` would result in the same flattened RTLIL name.

Does this match your understanding?

towoe:
I also think that `UUT.\escaped[1]` would be the correct reference to the signal.
Yes, I think Yosys will generate a flattened RTLIL name independent of the actual escaped signal name.
So this would mean that the RTLIL name handling would need updating, in order to have the correct escaped hierarchy name at a later stage? This would sound like a major change.

whitequark:
[...]

PR #1680, opened 2020-02-04
Fix trace tb generation

towoe:
This PR provides some improvements for the generation of Verilog test benches.
Please have a look at the commit messages for a detailed description of the changes.
There are still some problems. But it would be nice to get some feedback if the changes make sense.
If I have a solution for #1676 I can update this PR to include the changes. Or this can be done independently.

### DIFF
--- a/backends/smt2/smt2.cc
+++ b/backends/smt2/smt2.cc
@@ -532,10 +532,10 @@ struct Smt2Worker
 			if (cell->type.in("$anyconst", "$anyseq", "$allconst", "$allseq"))
 			{
 				registers.insert(cell);
-				string infostr = cell->attributes.count("\\src") ? cell->attributes.at("\\src").decode_string().c_str() : get_id(cell);
+				string infostr;
 				if (cell->attributes.count("\\reg"))
-					infostr += " " + cell->attributes.at("\\reg").decode_string();
-				decls.push_back(stringf("; yosys-smt2-%s %s#%d %d %s\n", cell->type.c_str() + 1, get_id(module), idcounter, GetSize(cell->getPort("\\Y")), infostr.c_str()));
+					infostr += cell->attributes.at("\\reg").decode_string();
+				decls.push_back(stringf("; yosys-smt2-%s %s#%d %d %s %s\n", cell->type.c_str() + 1, get_id(module), idcounter, GetSize(cell->getPort("\\Y")), get_id(cell->getPort("\\Y").as_wire()), infostr.c_str()));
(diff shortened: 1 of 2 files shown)


backends/smt2/smt2.cc
```diff
@@ -532,10 +532,10 @@ struct Smt2Worker
 			if (cell->type.in("$anyconst", "$anyseq", "$allconst", "$allseq"))
 			{
 				registers.insert(cell);
-				string infostr = cell->attributes.count("\\src") ? cell->attributes.at("\\src").decode_string().c_str() : get_id(cell);
+				string infostr;
 				if (cell->attributes.count("\\reg"))
-					infostr += " " + cell->attributes.at("\\reg").decode_string();
-				decls.push_back(stringf("; yosys-smt2-%s %s#%d %d %s\n", cell->type.c_str() + 1, get_id(module), idcounter, GetSize(cell->getPort("\\Y")), infostr.c_str()));
+					infostr += cell->attributes.at("\\reg").decode_string();
+				decls.push_back(stringf("; yosys-smt2-%s %s#%d %d %s %s\n", cell->type.c_str() + 1, get_id(module), idcounter, GetSize(cell->getPort("\\Y")), get_id(cell->getPort("\\Y").as_wire()), infostr.c_str()));
```
cliffordwolf: There's value to providing a src attribute in the smt comment when present (this change completely removes that information) and there's no guarantee that `cell->getPort("\\Y").as_wire()` doesn't return a NULL pointer afaict.

towoe: Thanks for the feedback.
Would you advice to add this info, if available, as an addition? Not sure anymore how this would affect retrieving of the info, I think I tried to align this with other cases.
Or does this mean changes to this comment should not be made?

cliffordwolf: I think I'm not too picky about the details, just think this information should be present, and that `.as_wire()` is no good here for the reasons stated above.

nakengelhardt:
@towoe are you still interested in finishing this PR? What is blocking it?

towoe:
@nakengelhardt I think it would make sense to get the hierarchy information in for const signals as well so this can be reused in a similar way as with the other signals.
But I am not familiar with the code at the moment to see how it can be done without `.as_wire()` and if the format is different from the other signal types it just makes it harder to retrieve the information at a later point.

nakengelhardt:
I'm not familiar with the smt2 format. Is everything in `infostr`, including the space and what's after it, part of the comment? Is it possible to add more spaces? If so, could you change the order and put the source location at the end?

If you have a small testcase that I could play around with to see the output, and the error that it causes, that might help.

For the `.as_wire()` problem, you can guard it with `if(x.is_wire())`, but you have to figure out what the appropriate thing to do is when it's not a wire...

PR #1682, opened 2020-02-06
synth_*: call 'opt -fast' after 'techmap'

eddiehung:
Necessary post-#1650-with-[fix](4c1d3a126d3978aeacbab623532c0ec9400b8959) at least on `tests/arch/anlogic/mux.ys`, which transforms:
before
into
after
which leads to the "optimal" solution of 5xLUT6...
To be honest, I think only a `clean` was strictly necessary but I think there may be value in doing `opt -fast` as done in `synth_xilinx`.

Even though it wasn't necessary for other arches to get back to passing status, I've done the same change for them...

### DIFF
--- a/techlibs/anlogic/synth_anlogic.cc
+++ b/techlibs/anlogic/synth_anlogic.cc
@@ -175,6 +175,7 @@ struct SynthAnlogicPass : public ScriptPass
 		if (check_label("map_gates"))
 		{
 			run("techmap -map +/techmap.v -map +/anlogic/arith_map.v");
+			run("opt -fast");
 			if (retime || help_mode)
 				run("abc -dff -D 1", "(only if -retime)");
 		}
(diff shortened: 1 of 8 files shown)


eddiehung:
In the interests of fixing master, I'm going to merge this in now and do a longer post mortem tomorrow.

eddiehung:
So the post-mortem finds that the BLIF file given to `abc` before and after this PR are identical, save for the signal ordering. This perturbation leads to a different (sub-optimal) solution with three logic levels instead of two. Adding `opt -fast` in this PR creates a different perturbation which puts it back on the straight and narrow.
See https://github.com/berkeley-abc/abc/issues/67 for another case of this, along with the comment that there's no good way of determining the best signal ordering upfront.

I'm also being told that multiplexer rich circuits can also be quite hard to map optimally since combinatorial optimisation (rewriting) will destroy this mux structure making it harder to map into LUTs...

PR #1684, opened 2020-02-06
Fix/cleanup +/xilinx/arith_map.v

eddiehung:


techlibs/xilinx/arith_map.v
```diff
@@ -36,13 +36,12 @@ module _80_xilinx_lcu (P, G, CI, CO);
 `ifdef _EXPLICIT_CARRY
 
 	wire [WIDTH-1:0] C = {CO, CI};
-	wire [WIDTH-1:0] S = P & ~G;
```
mwkmwkmwk: Could you explain that part? It seems to me that `P & ~G` was necessary to match `$lcu` model behavior when both `P` and `G` are set.  Is there some rule that they can never be both set at once?  If so, shouldn't we have `$lcu` output `x` in this case?

eddiehung: Good spot. You're right: there is no rule stating that `P` and `G` must be mutually exclusive, and the `$lcu` model is indeed: `CO[i] = G[i] || (P[i] && CO[i-1]);`. I was focused solely on the case that `$lcu`-s would be mapped to a multiplexer. Let me revert.

PR #1685, opened 2020-02-06
Removing cells_sim from GoWin bram techmap

dh73:
Answering @eddiehung reviews request, I'm fixing one of the comments: 
https://github.com/YosysHQ/yosys/commit/f9272fc56d7179f04a9f776bf056eedfc33dd358#r37144384

the other one: https://github.com/YosysHQ/yosys/commit/f9272fc56d7179f04a9f776bf056eedfc33dd358#r37144371
Makes perfect sense to remove `run("techmap -map +/techmap.v");` since is irrelevant, but `fine` synthesis seems to be replaced for `map_ffram` and `map_gates` in current head, so these lines does not exists in master anymore.

### DIFF
--- a/techlibs/gowin/synth_gowin.cc
+++ b/techlibs/gowin/synth_gowin.cc
@@ -191,7 +191,7 @@ struct SynthGowinPass : public ScriptPass
 		if (!nobram && check_label("map_bram", "(skip if -nobram)"))
 		{
 			run("memory_bram -rules +/gowin/brams.txt");
-			run("techmap -map +/gowin/brams_map.v -map +/gowin/cells_sim.v");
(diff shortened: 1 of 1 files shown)


eddiehung:
Thanks Diego! This was something I spotted working on #1682 .
Re: https://github.com/YosysHQ/yosys/commit/f9272fc56d7179f04a9f776bf056eedfc33dd358#r37144371 I think I changed it here: https://github.com/YosysHQ/yosys/pull/1682/files#diff-83149add259aef87d7424c0bd00533c0L214

Also, can you re-do your branch with just the 931a68f commit and not the previous two which seem to be no-ops?

dh73:
Yeah, one of these two commits was too old. Sorry for that.

Issue #1474, opened 2019-11-05
ERROR: Assert `lut_slacks[node] >= 0' failed in passes/techmap/flowmap.cc:1295.


ZirconiumX:
## Steps to reproduce the issue

Using 
[flowmap_1295.tar.gz](https://github.com/YosysHQ/yosys/files/3806602/flowmap_1295.tar.gz), run
`yosys -p "flowmap -maxlut 6 -optarea 4"  flowmap_1295.il`.

## Expected behavior

`flowmap` completes successfully.

## Actual behavior

```
2. Executing FLOWMAP pass (pack LUTs with FlowMap-r).
Labeling cells.
Mapped to 1 LUTs with maximum depth 1.

Optimizing area.
Relaxing with depth bound 1.
  Relaxed to 1 (+0) LUTs.
  Slack exhausted.
Relaxing with depth bound 2.
  Breaking LUT $auto$bugpoint.cc:222:simplify_something$247 to extract LUT $procmux$266.B_AND_S [10] (potential 150).
  Relaxed to 2 (+1) LUTs.
  Slack exhausted.
Relaxing with depth bound 3.
  Breaking LUT $auto$bugpoint.cc:222:simplify_something$247 to extract LUT \o_o [10] (potential 700).
  Breaking LUT $auto$bugpoint.cc:222:simplify_something$247 to extract LUT $auto$simplemap.cc:127:simplemap_reduce$342 [5] (potential 500).
  Breaking LUT $auto$bugpoint.cc:222:simplify_something$247 to extract LUT $auto$simplemap.cc:127:simplemap_reduce$351 [2] (potential 300).
  Breaking LUT $auto$bugpoint.cc:222:simplify_something$247 to extract LUT $auto$simplemap.cc:127:simplemap_reduce$356 [1] (potential 300).
  Breaking LUT $auto$simplemap.cc:127:simplemap_reduce$356 [1] to extract LUT $procmux$266.B_AND_S [10] (potential 150).
  Relaxed to 3 (+1) LUTs.
  Slack exhausted.
Relaxing with depth bound 4.
  Breaking LUT $auto$simplemap.cc:127:simplemap_reduce$356 [1] to extract LUT \o_o [10] (potential 700).
  Breaking LUT $auto$simplemap.cc:127:simplemap_reduce$356 [1] to extract LUT $auto$simplemap.cc:127:simplemap_reduce$342 [5] (potential 500).
[...]

PR #1691, opened 2020-02-08
Use flowmap instead of gate2lut in noabc modes

ZirconiumX:
This is more of a discussion point than a pull request.

The PR itself is simple: use the `flowmap` pass from #775 instead of `techmap -map +/gate2lut.v` for better synthesis quality when using `synth{,_ice40} -noabc`.

firefox_2020-02-08_17-22-32

But should we just remove `-noabc`, given that only iCE40 even exposes an option for it?
Or should we add `-noabc` option that uses `flowmap` to all flows, so that Yosys does not require ABC for FPGA synthesis?
What about CMOS synthesis, which currently *does* require ABC?

### DIFF
--- a/techlibs/common/synth.cc
+++ b/techlibs/common/synth.cc
@@ -78,6 +78,9 @@ struct SynthPass : public ScriptPass
 		log("    -abc9\n");
 		log("        use new ABC9 flow (EXPERIMENTAL)\n");
 		log("\n");
+		log("    -flowmap\n");
+		log("        use FlowMap LUT techmapping instead of ABC\n");
+		log("\n");
 		log("\n");
 		log("The following commands are executed by this synthesis command:\n");
 		help_script();
@@ -85,7 +88,7 @@ struct SynthPass : public ScriptPass
 	}
 
 	string top_module, fsm_opts, memory_opts, abc;
-	bool autotop, flatten, noalumacc, nofsm, noabc, noshare;
+	bool autotop, flatten, noalumacc, nofsm, noabc, noshare, flowmap;
 	int lut;
 
 	void clear_flags() YS_OVERRIDE
@@ -101,6 +104,7 @@ struct SynthPass : public ScriptPass
 		nofsm = false;
 		noabc = false;
 		noshare = false;
+		flowmap = false;
 		abc = "abc";
 	}
 
@@ -167,6 +171,10 @@ struct SynthPass : public ScriptPass
 				abc = "abc9";
 				continue;
 			}
+			if (args[argidx] == "-flowmap") {
+				flowmap = true;
+				continue;
+			}
 			break;
 		}
 		extra_args(args, argidx, design);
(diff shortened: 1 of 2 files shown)


whitequark:
There is also `synth -noabc`.

ZirconiumX:
Yes, this PR changes that.

whitequark:
Right--I meant that even if iCE40 ends up always using `abc`, it seems worthwhile to keep it in `synth`, re this part:

> But should we just remove `-noabc`, given that only iCE40 even exposes an option for it?

ZirconiumX:
So in the process of researching this I did a *lot* of literature review. The `flowmap` pass contains a lot of useful infrastructure for cut-enumeration based mapping approaches. However, the state of the art has moved on significantly, building on the legacy of FlowMap, in particular [DAOMap](https://ieeexplore.ieee.org/document/1382677).

Fundamentally, the problem with exhaustive cut-enumeration approaches is that they are `O(N^K)` where N is the number of nodes in the tree and K is the number of LUT inputs. FlowMap itself targeted K = 5, but at K = 6 and beyond (ECP5, LUT6 Xilinx and ALM Intel) the exponential term begins to rear its ugly head. [Mishchenko et al](https://people.eecs.berkeley.edu/~alanmi/publications/2006/tcad06_map.pdf) show various ways of more efficiently enumerating cuts to keep the depth-optimality with lower average cuts per node, but it feels to me like punting the problem down the road.

ABC uses a [cutless approach also from Mishchenko](https://people.eecs.berkeley.edu/~alanmi/publications/2007/tech07_fast.pdf) which does not offer the depth-optimality guarantees that FlowMap does, but is significantly faster.

eddiehung:
Be aware there is at least one outstanding flowmap issue described in https://github.com/YosysHQ/yosys/issues/1474

whitequark:
> Be aware there is at least one outstanding flowmap issue described in #1474

That one is using `-optarea`, which is an unfinished feature that doesn't currently do anything useful (it makes the area worse). The base flowmap flow (without using `-relax` or `-optarea`) should be quite reliable.

ZirconiumX:
Something for consideration: should this be added for ECP5 (at least)? While it does have tables for up to LUT7 which is expensive, in `-nowidelut` mode the LUT4s should be easy to handle.

nakengelhardt:
Do we need extra tests for this or do the existing flowmap and ice40 tests cover enough bases?

Apart from that I'd say let's merge this and you can make a separate PR for ECP5 later if you find it makes sense to add it there too.

ZirconiumX:
I have two goals with this: first to have something better in the
admittedly rare case of building Yosys without ABC and second to expose it
publicly, so it can get some measure of testing on real-world code by
people curious what the new -flowmap option does.

>

nakengelhardt:
Right, but nevertheless it's probably not going to be so frequently used that we can rely on users to yell if it breaks. I'm just wondering how likely it would be to accidentally break this option without failing any of the existing tests, but I'm not familiar with the ice40 code at all.

PR #1694, opened 2020-02-09
json: Change compat mode to directly emit ints <= 32 bits

rqou:
This increases compatibility with certain older parsers in some cases that worked before commit 15fae357 but do not work with the current compat-int mode

The specific case that I encountered involved flip-flop cells with `INIT` attributes set to 0 (these attributes are created by the `dffinit` pass). For some reason that I didn't fully investigate (because it does not seem to cause any breakage), yosys seems to think that these attributes are 0 bits wide (e.g. `write_verilog` will output `.INIT({0{1'b0}})`). The code prior to commit 15fae357 would have output a 0 into the resulting JSON file, but the code that is currently in master will output an empty string (even with `-compat-int`).

By making `-compat-int` write attributes with <= 32 bits (rather than == 32 bits) as JSON numbers, the behavior should exactly match the pre-15fae357 behavior.

### DIFF
--- a/backends/json/json.cc
+++ b/backends/json/json.cc
@@ -104,7 +104,7 @@ struct JsonWriter
 			if (state < 2)
 				str += " ";
 			f << get_string(str);
-		} else if (compat_int_mode && GetSize(value) == 32 && value.is_fully_def()) {
+		} else if (compat_int_mode && GetSize(value) <= 32 && value.is_fully_def()) {
 			if ((value.flags & RTLIL::ConstFlags::CONST_FLAG_SIGNED) != 0)
 				f << stringf("%d", value.as_int());
 			else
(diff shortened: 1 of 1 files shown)


mwkmwkmwk:
This seems like it warrants a patch to `dffinit` as well.

Looks good to me. @cliffordwolf — since this is exactly the kind of old parsers that we were talking about that people actually have problem with, and because 15fae357f6c7a5c205d2733cd0b2640eecd17a2d hasn't landed in an actual release yet, I'd say this should be the behavior of `-compat-int`.  I'll merge in this PR if there is no opposition in a few days.

Kmanfi:
I am using old style parsers and now I need to change all my scripts with this -compat-int.
One option is also keep "old" as a default and make switch for this new behaviour.

whitequark:
IMO: the "old" JSON format was clearly and unambiguously broken. The best thing would have been to never ship it. But the next best thing would be to not keep it as default.

Kmanfi:
😄 👍 
I am happy to update my scripts with the new switch. Actually now I can modify those (int str) if/thens away.

PR #1696, opened 2020-02-09
write_verilog: dump zero width sigspecs correctly

whitequark:
See commit 4ff44d85a5 for details.

### DIFF
--- a/backends/verilog/verilog_backend.cc
+++ b/backends/verilog/verilog_backend.cc
@@ -355,7 +355,8 @@ void dump_sigchunk(std::ostream &f, const RTLIL::SigChunk &chunk, bool no_decima
 void dump_sigspec(std::ostream &f, const RTLIL::SigSpec &sig)
 {
 	if (GetSize(sig) == 0) {
-		f << "\"\"";
+		// See IEEE 1364-2005 Clause 5.1.14.
+		f << "{0{1'b0}}";
 		return;
 	}
 	if (sig.is_chunk()) {


mwkmwkmwk:
Are you sure that's correct? from the very chapter you refered to:

"A replication operation may have a replication constant with a value of zero. This is useful in parameterized code. A replication with a zero replication constant is considered to have a size of zero and is ignored. Such a replication shall appear only within a concatenation in which at least one of the operands of the concatenation has a positive size."

It seems we're not in a concatenation?

whitequark:
You're absolutely right: this isn't a legal transformation. Thanks for catching this.

The problem here is that the current code is blatantly wrong, too.

whitequark:
There are essentially two cases here:
  * Empty sigchunk in a sigspec. That can be just skipped, and AFAICT already gets skipped (or at least I can't construct a pathological input in textual IR).
  * Empty sigspec. That means the entire assignment should be gone. But... that'd require replacing like a hundred places in write_verilog.

Thoughts?

mwkmwkmwk:
I came to the same conclusion: verilog really doesn't like 0-bit assignments, so you have to skip the assignment entirely, and there's probably no way around it.

whitequark:
Right. I mostly mean the specific approach to skipping them. Ideally without modifying half of `write_verilog`.

mithro:
Could you do a prepass which eliminates the zero entries before running the `write_verilog` pass?

whitequark:
Yes, but it's not ideal because AFAIK there's no Yosys pass which does *just* that, and e.g. for nMigen it's important for the Verilog output to stay close to the RTLIL it emits. (nMigen is not affected by this issue directly because it strips out assignments to zero width sigspecs on its own, effectively implementing the pass you suggest.)

PR #1699, opened 2020-02-13
iopadmap: move \init attributes from outpad output to its input

eddiehung:
Fixes flop initialisation

### DIFF
--- a/passes/techmap/iopadmap.cc
+++ b/passes/techmap/iopadmap.cc
@@ -408,18 +408,35 @@ struct IopadmapPass : public Pass {
 				RTLIL::Wire *wire = it.first;
 				RTLIL::Wire *new_wire = module->addWire(NEW_ID, wire);
 				module->swap_names(new_wire, wire);
-				wire->attributes.clear();
 				for (int i = 0; i < wire->width; i++)
 				{
 					SigBit wire_bit(wire, i);
 					if (!it.second.count(i)) {
-						if (wire->port_output)
+						if (wire->port_output) {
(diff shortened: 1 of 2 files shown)


passes/techmap/iopadmap.cc
```diff
@@ -408,18 +408,35 @@ struct IopadmapPass : public Pass {
 				RTLIL::Wire *wire = it.first;
 				RTLIL::Wire *new_wire = module->addWire(NEW_ID, wire);
 				module->swap_names(new_wire, wire);
-				wire->attributes.clear();
 				for (int i = 0; i < wire->width; i++)
 				{
 					SigBit wire_bit(wire, i);
 					if (!it.second.count(i)) {
-						if (wire->port_output)
+						if (wire->port_output) {
```
mwkmwkmwk: I don't understand the structure of this code at all.

1. This is broken if pads are inserted only for a subset of bits in a wire (granted, this is a rare case, but...) — if the first bit does not have an inserted pad, the clear will be executed and init attribute nuked.
2. Why is the clear in both branches of the if?
3. Why do we deal with it inside this for loop at all?  The code that handles the attribute deals with the whole wire, and not individual bits.
4. I don't really see how what we want to do depends on whether a buffer is inserted or just a plain connection is made.

IMO what we should do is just keep the `clear` above, and right afterwards move the `init` attribute (if any) from `new_wire` back to `wire`.

eddiehung: Sorry about such a bad patch and thanks for your detailed review. I think I've resolved 2-4, but I'm not sure about 1 -- whether it is still relevant or how to construct a testcase that exercises it.

PR #1702, opened 2020-02-13
Fix the README for macOS Catalina

BracketMaster:
brew changed a bit in Catalina...

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ Similarily, on Mac OS X Homebrew can be used to install dependencies:
 	
 The brew bundle is not currently working on Catalina, so do:
 
-	$ brew install python3 bison flex readline gawk libffi git graphviz pkgconfig boost zlib tcl tcl-tk
+	$ brew install python3 bison flex readline gawk libffi git graphviz pkg-config boost zlib tcl tcl-tk xdot
 
 or MacPorts:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -72,6 +72,10 @@ prerequisites for building yosys:
 Similarily, on Mac OS X Homebrew can be used to install dependencies:
 
 	$ brew tap Homebrew/bundle && brew bundle
+	
+The brew bundle is not currently working on Catalina, so do:
+
+	$ brew install python3 bison flex readline gawk libffi git graphviz pkgconfig boost zlib tcl tcl-tk
```
mmicko: Can you change pkgconfig -> pkg-config (since first is just alias)
xdot is missing in list (required for show command)
Also please do update Brewfile to reflect addition of boost (needed for pyosys)

Guess tcl and zlib are actually not needed to be there, since they will anyway be installed by other requirements, can you confirm that.

BracketMaster: Oh this one is on me.
``$brew bundle`` didn't work because I wasn't inside ``/yosys``
SMH - staying up late is not good.

Perhaps it might be a good idea to add ``$cd yosys`` before  ``brew bundle`` to the README for programmers like me who stay up past their bedtime?

Things work fine on Catalina :)
The brewfile is fine.

mmicko:
Since brew file do work, there is actually no need to repeat content of it in README.md again, but have added more explanation in https://github.com/YosysHQ/yosys/commit/a7df492243dbf186037b577657a65e357e8e0fec

PR #1709, opened 2020-02-17
Improve CoolRunner-II optimization by using extract_counter pass

rqou:
The current implementation tends to implement counters in a very inefficient way using a very large number of pterms. Using the `extract_counter` pass, it is possible to detect most common ways of describing a counter in HDL and map them to a manually-optimized implementation that uses significantly fewer resources (max 2 pterms per bit)


rqou:
cc @azonenberg for the changes to extract_counter

PR #1715, opened 2020-02-22
Closes #1714. Fix make failure when NDEBUG=1.

### LINKED ISSUE #1714: Make fails when NDEBUG=1
make clean
make config-clang
# set NDEBUG=1 in Makefile with your preferred editor
make

## Expected behavior

Build success

## Actual behavior

Build fails with:

```
passes/techmap/abc9_ops.cc:778:21: error: use of undeclared identifier 'box_module'
                        auto abc9_flop = box_module->attributes.count("\\abc9_flop");
                                         ^
passes/techmap/abc9_ops.cc:789:46: error: use of undeclared identifier 'derived_type'
                        for (const auto &port_name : box_ports.at(derived_type)) {
                                                                  ^
passes/techmap/abc9_ops.cc:790:22: error: use of undeclared identifier 'box_module'
                                RTLIL::Wire *w = box_module->wire(port_name);
```

boqwxp:
Remove the #ifndef NDEBUG guard in passes/techmap/abc9_ops.cc

### DIFF
--- a/passes/techmap/abc9_ops.cc
+++ b/passes/techmap/abc9_ops.cc
@@ -752,13 +752,11 @@ void reintegrate(RTLIL::Module *module)
 				continue;
 			}
 
-#ifndef NDEBUG
 			RTLIL::Module* box_module = design->module(existing_cell->type);
 			IdString derived_type = box_module->derive(design, existing_cell->parameters);
 			RTLIL::Module* derived_module = design->module(derived_type);
 			log_assert(derived_module);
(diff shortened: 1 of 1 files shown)


eddiehung:
Thanks for the patch. My original intention is to also guard some of the code that supports the `log_assert`, but that has incorrectly included `box_module` too. On revisiting this, it can be expected that the `derived_module` lookup *should* get optimised away if it isn't used by the assert.

@mmicko should we have a `NDEBUG` build on CI too?

PR #1716, opened 2020-02-22
ecp5: Add missing parameter to \$__ECP5_PDPW16KD

zeldin:
This caused yosys to fail when running with implicit nets turned off, even if no bram was used...


daveshah1:
I think it would be better just to remove the `WRITEMODE_A` localparam altogether? Looks like some dead code I left behind from the other BRAM rules.

PR #1718, opened 2020-02-23
Closes #1717. Add more precise Verilog source location information to AST and RTLIL nodes.

### LINKED ISSUE #1717: Add more precise Verilog source file location information to AST and RTLIL nodes
See `frontends/ast/ast.h`, in definition of `struct AstNode`:
```
		// this is the original sourcecode location that resulted in this AST node
		// it is automatically set by the constructor using AST::current_filename and
		// the AST::get_line_num() callback function.
		std::string filename;
		int linenum;
```

In other words, only the source file name and *a* line number are available, but no other information.

## Desired behavior

AST and RTLIL nodes should ideally have precise location information, not just a line number.

In `ast.h`, define a new:
```
	struct AstSrcLocType
	{
		unsigned int first_line, last_line;
		unsigned int first_column, last_column;
	};
```

Add a new `AstSrcLocType location;` member variable to `AstNode`.  Optionally remove `int linenum` and replace its usages with `location.first_line`, or perhaps a more sophisticated replacement that preserves/reports the entire location information.

Update `frontends/verilog/verilog_lexer.l` to track the precise location information in `YY_USER_ACTION` and `frontends/verilog/verilog_parser.y` to assign to the generated AST nodes locations from things on the stack.  Parsers/lexers for other file formats can also be modified to track and record precise location information, but I am personally only interested in Verilog for now.

Files like `frontends/ast/ast.cc`, `frontends/ast/genrtlil.cc`, `frontends/verilog/verilog_frontend.cc`, and perhaps other files require some modification.

This would make precise location information available in the AST and RTLIL data structures, enabling better error messages and other applications that might find such location information useful, e.g. Vennsa OnPoint-like automated RTL source level error localization.

boqwxp:
I am not sure if other frontends (verific, ghdl, others ?) use the `linenum` member.  If so, then it needs to be decided whether to keep both `linenum` and `location` and synchronize their updates, to keep only `location` and require those frontends to change to using `location`, or some other strategy.

Maybe the best strategy is to keep the `use_internal_line_num()`, `set_line_num()`, and `get_line_num()` interface, add a `set_precise_location()` and `get_precise_location()` interface, and update all existing references to `linenum` to use `get_line_num()` or `get_precise_location()` as appropriate.

@eddiehung Thoughts?

### DIFF
--- a/frontends/ast/ast.cc
+++ b/frontends/ast/ast.cc
@@ -182,7 +182,7 @@ bool AstNode::get_bool_attribute(RTLIL::IdString id)
 
 	AstNode *attr = attributes.at(id);
 	if (attr->type != AST_CONSTANT)
-		log_file_error(attr->filename, attr->linenum, "Attribute `%s' with non-constant value!\n", id.c_str());
+		log_file_error(attr->filename, attr->location.first_line, "Attribute `%s' with non-constant value!\n", id.c_str());
 
 	return attr->integer != 0;
 }
@@ -197,7 +197,6 @@ AstNode::AstNode(AstNodeType type, AstNode *child1, AstNode *child2, AstNode *ch
 
 	this->type = type;
 	filename = current_filename;
-	linenum = get_line_num();
 	is_input = false;
 	is_output = false;
 	is_reg = false;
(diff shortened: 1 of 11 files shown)


cliffordwolf:
> I am not sure if other frontends (verific, ghdl, others ?) use the linenum member.

Afaik no other front-end currently goes through uses the AST framework. At least verific goes directly from verific internal representation to RTLIL, so there is definitely no issue here. If any out-of-tree front-end uses AST then they should be easy enough to fix for that change. Thus the approach in this PR looks fine to me.

boqwxp:
OK, I've marked this ready for review.  It would be nice in the future to make a proper interface as described above, so I will make a separate issue for that.  Otherwise I think it's ready to merge.

PR #1721, opened 2020-02-25
deminout: Don't demote inouts with unused bits

daveshah1:
Previously, `deminout` was converting `inout` ports with unused bits to outputs with those bits driving 'x', which doesn't seem like a sound transformation to me as those unused bits should be high-z.

example case:
```verilog
module top(input clk, inout [7:0] x);

reg [3:0] ctr;
always @(posedge clk) ctr <= ctr + 1'b1;

assign x[7:4] = ctr;
endmodule
````
```
yosys -p "synth_ecp5; dump" tri.v
```

### DIFF
--- a/tests/various/deminout_unused.ys
+++ b/tests/various/deminout_unused.ys
@@ -1,0 +1,14 @@
+read_verilog <<EOT
+module top(input clk, inout [7:0] x);
+
+reg [3:0] ctr;
+always @(posedge clk) ctr <= ctr + 1'b1;
+
+assign x[7:4] = ctr;
+endmodule
+EOT
+proc
+tribuf
+deminout
+select -assert-count 1 i:x o:x %i
+
--- a/passes/techmap/deminout.cc
+++ b/passes/techmap/deminout.cc
@@ -121,8 +121,7 @@ struct DeminoutPass : public Pass {
 									goto tribuf_bit;
 							} else {
 						tribuf_bit:
-								if (bits_used.count(bit))
-									new_input = true;
+								new_input = true;
 							}
 						}
 

### CODE REVIEW

**Comment by eddiehung** [comment]
Test, plz.

File: tests/various/deminout_unused.ys [hunk 1]
```diff
@@ -0,0 +1,12 @@
+read_verilog <<EOT
+module top(input clk, inout [7:0] x);
+
+reg [3:0] ctr;
+always @(posedge clk) ctr <= ctr + 1'b1;
+
+assign x[7:4] = ctr;
+endmodule
+EOT
+synth_ecp5
```
eddiehung: > I can change this to a generic test though
Please do!

daveshah1: Done

mwkmwkmwk:
I would argue that `deminout` is doing a sound transformation here, it's `opt -undriven` that is at fault and performs an unsound transformation — it's changing unconnected outputs (that would be high-z by default) to x-driving outputs.

mwkmwkmwk:
On second thought, according to IEEE 1364.1, assignments don't propagate `z` states, and `output` port connections count as assignments (while AFAICT `inout` connections do not), so perhaps this transform is *also* unsound...

daveshah1:
It seems to be vendor FPGA tool convention that unused top level outputs drive 1'b0 whereas unused top level inouts are high-impedance.

mwkmwkmwk:
> It seems to be vendor FPGA tool convention that unused top level outputs drive 1'b0 whereas unused top level inouts are high-impedance.

Aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa

mwkmwkmwk:
... and on ISE, unused inout is an input, while unused output is a DRC error caught *at bitgen* (the IOB is configured as perma-on output buffer with data input pin disconnected). Wonderful.

PR #1725, opened 2020-02-27
Support standard typedef grammar

PeterCrozier:


swetland:
Is there an objection to picking this up?  It looks like a pretty straight forward solution that gets typedef support using the correct syntax, which seems like a big win from over here...

eddiehung:
> Is there an objection to picking this up? It looks like a pretty straight forward solution that gets typedef support using the correct syntax, which seems like a big win from over here...

I have no objections, but I'm just not familiar enough with this code to review it properly and pull the trigger. Adding reviewers from #1448.

daveshah1:
Overall LGTM - there are a few edge cases that are legal SV that this won't work on and would need some more handling of scope in the parser, but they are unlikely to crop up in real designs (e.g. defining a typedef inside a genblock and then using a wire of the same name outside of the genblock)

nakengelhardt:
@swetland @PeterCrozier FYI the verilog frontend has been designated a sensitive code area that requires special review by Claire (@cliffordwolf) because it's so easy to subtly break, so it's always going to take a bit longer to review any PRs that make changes to it. I'll try to sneak it into her schedule but she's on business trip for the next two weeks...

frontends/verilog/verilog_frontend.h
```diff
@@ -45,6 +45,12 @@ namespace VERILOG_FRONTEND
 	// this function converts a Verilog constant to an AST_CONSTANT node
 	AST::AstNode *const2ast(std::string code, char case_type = 0, bool warn_z = false);
 
+	// names of locally typedef'ed types
+	extern std::map<std::string, AST::AstNode*> user_types;
+
+	// names of package typedef'ed types
+	extern std::map<std::string, AST::AstNode*> pkg_user_types;
```
cliffordwolf: This map should probably go into `RTLIL::Design`, like `verilog_packages`, `verilog_globals`, and `verilog_defines`, and should be reset with `deisgn -reset-vlog`.

PeterCrozier: I am not following you here.  These two maps are temporary lookup tables belonging to the parser and they are not persisted in the design.

The `user_types` map is used only inside the parser/lexer and the `pkg_user_types` is rebuilt from `design->verilog_packages` in `verilog_frontend.cc` as `yyparse()` is not passed any reference to the design class and can't directly reference the packages.  Declaring these maps in the parser seems like the right place to me.

However I do take your point about `design -reset-vlog` which clears the packages in the design and has no visibility of the `pkg_user_types` map.  I think the best way to handle this is to check if the design package list is empty before calling the parser and if so clear the package types table just in case. 

I have committed a fix for this.  This commit also has a fix to support typedefs in interfaces which weren't around in yosys when I did this the first time.

cliffordwolf: Yes, my comment related to `pkg_user_types` only. I don't think your fix is sufficient, considering that there's not only `design -reset-vlog`, but also `design -...` to push, pop, stash, save and load designs, and each of those ops would invalidate `pkg_user_types` afaict. That's why I think it should go into `RTLIL::Design`, so that `pkg_user_types` is moved around together with the rest of the design on those operations.

PeterCrozier: I'm still not convinced that `pkg_user_types` belongs in the design.  As it stands it is just a temporary lookup mechanism internal to the parser/lexer.  If moved to the design then all the design code (push/pop/...) will have to take account of it and I really don't want to go there.

So I have made a commit whereby the map gets built before the parser and not afterwards when the tree was processed.  This way it will always use the latest design and the lifetime of the map is just the parse phase.  I have no idea why I thought it had to be afterwards.  Maybe I originally built it from the tree.

There is a second commit fixing a bug with typedef enums in packages caused by simplify not being called for them.

cliffordwolf:
LGTM. Please resolve merge conflicts.  ;)

nakengelhardt:
Merged as #1803.

PR #1732, opened 2020-03-02
sta: very crude static timing analysis pass

eddiehung:
Currently computing latest arrival time through flattened circuit.


ZirconiumX:
This could perhaps use an interconnect delay parameter, much like ABC[9] has. I can imagine a situation where one path goes through a DSP while another goes through like a multiplexer tree, and this would select the DSP as the potential slowest path rather than the mux tree.

eddiehung:
> This could perhaps use an interconnect delay parameter, much like ABC[9] has. I can imagine a situation where one path goes through a DSP while another goes through like a multiplexer tree, and this would select the DSP as the potential slowest path rather than the mux tree.

`abc9`'s `-W` parameter is passed straight through to ABC, where its `&if` command simply adds this delay to each LUT input delay; it is not considered for anything else:
```
    // add wire delay to LUT library delays
    if ( pPars->WireDelay > 0 && pPars->pLutLib )
    {
        int i, k;
        for ( i = 0; i <= pPars->pLutLib->LutMax; i++ )
            for ( k = 0; k <= i; k++ )
                pPars->pLutLib->pLutDelays[i][k] += pPars->WireDelay;
    }
    // perform mapping
    pNew = Gia_ManPerformMapping( pAbc->pGia, pPars );
    // subtract wire delay from LUT library delays
    if ( pPars->WireDelay > 0 && pPars->pLutLib )
    {
        int i, k;
        for ( i = 0; i <= pPars->pLutLib->LutMax; i++ )
            for ( k = 0; k <= i; k++ )
                pPars->pLutLib->pLutDelays[i][k] -= pPars->WireDelay;
    }
```
(pasted because https://github.com/berkeley-abc/abc/blob/master/src/base/abci/abc.c is too big for GH to display)

I had considered doing exactly the same for this `sta` command too so that one could correlate against what ABC sees, but in the end decided against it in favour of allowing `sta` to examine pure logic delays. I'm also not sure what value adding a homogeneous net delay would be (there is a small value in doing so for ABC's `&if`, as it can encourage the use of bigger LUTs instead of smaller ones which on the surface would appear to be quicker:
https://github.com/YosysHQ/yosys/blob/8b074cc473676288eb9cb04b09fb37ba6d11461a/techlibs/xilinx/synth_xilinx.cc#L35-L36
)

ZirconiumX:
> I'm also not sure what value adding a homogeneous net delay would be

As I said previously, a situation where, when you include some amount of interconnect delay the longest path changes. It doesn't have to be accurate, but it could be the difference between a path that goes through a single, slow, cell, and a longer path that goes between multiple faster cells.

eddiehung:
> > I'm also not sure what value adding a homogeneous net delay would be
> 
> As I said previously, a situation where, when you include some amount of interconnect delay the longest path changes. It doesn't have to be accurate, but it could be the difference between a path that goes through a single, slow, cell, and a longer path that goes between multiple faster cells.

It would certainly be easy to add in a delay for each net, but I'm still not seeing any elegant solutions to:
* how would one choose the one delay value to put in there in order to expose the _true_ critical path, if that is indeed possible?
* how would/should `sta` know which nets to add delays to? For example, some nets (e.g. carry chains, LUT to FF connections, etc.) would travel on fast connections as opposed to the general routing fabric to incur that net delay.

The originating purpose of this pass is to provide a quick and dirty, and fair, way to compare the pure synthesis quality between different flows (and indeed even those generated outside of Yosys) before entering P&R.

passes/cmds/sta.cc
```diff
@@ -0,0 +1,295 @@
...
+						maxarrival = new_arrival;
+						maxbit = dst_bit;
+					}
+				}
+			}
+		}
+
+		log("Latest arrival time in '%s' is %d:\n", log_id(module), maxarrival);
+		auto b = maxbit;
+		const auto &e = endpoints.at(maxbit);
```
ZirconiumX: I've been playing around with adding Gowin specify blocks, and while working on it keep hitting aborts when `maxbit` is not in `endpoints`. Could you add some better error handling for that?

eddiehung: Just to check we're on the same page first: would you agree that that if `maxbit` is not in `endpoints`, that means the latest arrival time does not terminate at a primary output or sequential input, and so handling it would mean just emitting an error message? Or am I missing something?

ZirconiumX: Yeah, an error message would be much more graceful than a `std::out_of_range` exception.

Though I feel I should ask what the definition of sequential input is here; a `posedge` specify?

Additionally, while discussing this pass, I get warnings on not having timing arcs for `VCC`/`GND`, which as the name implies are just constant drivers; as far as I can tell this isn't actually something you can communicate to `sta`.

eddiehung: > Yeah, an error message would be much more graceful than a `std::out_of_range` exception.

Thanks for pointing that out. I admit it is not very friendly to any one but me right now. Resolved in 5b26d4b64c8586c29e5e8e65ead44ed0af945282.
 
> Though I feel I should ask what the definition of sequential input is here; a `posedge` specify?

Endpoints are cell input ports with "required" (`$setup`) time constraints: https://github.com/YosysHQ/yosys/blob/5b26d4b64c8586c29e5e8e65ead44ed0af945282/passes/cmds/sta.cc#L94-L97

Or are primary outputs:
https://github.com/YosysHQ/yosys/blob/5b26d4b64c8586c29e5e8e65ead44ed0af945282/passes/cmds/sta.cc#L140-L143


> 
> Additionally, while discussing this pass, I get warnings on not having timing arcs for `VCC`/`GND`, which as the name implies are just constant drivers; as far as I can tell this isn't actually something you can communicate to `sta`.

ZirconiumX: > Can you think of a system to identify these calls *without* having to tag them with an attribute to suppress the no-timing-arcs warning?

The most obvious property that comes to mind is that at the very least Gowin `VCC`/`GND` have no input ports (I can arrange this for `synth_intel_alm` too, as the `__MISTRAL_{VCC,GND}` cells are also like this). Lacking any inputs means they must surely be constant at runtime, unless there's some Verilog trickery that I'm not aware of.

ZirconiumX: Alternatively you could check if a cell has only constant inputs, but for a simple test, the "no inputs" method seems like a good rule of thumb.

eddiehung: @ZirconiumX The "no timing arcs" warning message should now be suppressed by ee52e3a. Is it what you was looking for?

ZirconiumX: Yep!

ZirconiumX:
I realise this kinda defeats the "crudeness" of this particular pass, but I've noticed that Intel, Gowin and by the looks of it iCE40 all have unateness timings. Perhaps in the distant future `sta` (and maybe nextpnr? I don't know too much about it) can take into account unateness for critical path timing calculation.

daveshah1:
There has been some consideration in the nextpnr API for this, I'm not sure how useful it will be for logic but it could definitely be useful to determine clock duty cycle skew (eg for posedge->negedge paths)

Issue #1740, opened 2020-03-05
Better document exists-forall problem support


boqwxp:
General
====
I don't see much in the documentation aside from:

> FIXME: Add information about $assert,$assume,$live,$fair,$cover,$equiv,$initstate,$anyconst,$anyseq,$allconst,$allseqcells

Some of these have somewhat decent descriptions at https://symbiyosys.readthedocs.io/en/latest/verilog.html#unconstrained-variables .

Questions to Clarify
====
Looking over commit b13e6bd375dc19fc2d6a3e67cdc6c045da732200, I'm not super clear on what exactly are `ex_state_eq` and `ex_input_eq` and their role in the problem formulation.

I suppose that's partly a consequence of my lack of understanding of the whole process for setting up the problem for solving (the yosys-smtbmc part, I guess).  A general overview of how the emitted SMTLIBv2 differs from normal, how yosys-smtbmc treats it differently, etc. would be very helpful.

For example, I see things like (`backends/smt2/smt2.cc`):
```
ex_state_eq.push_back(stringf("(= %s %s)", get_bv(cell->getPort("\\Y")).c_str(), get_bv(cell->getPort("\\Y"), "other_state").c_str()));
```

I wonder what exactly is the purpose of `other_state`.  Is the equality constraint between application with `state` and application with `other_state`, (or, rather with `s0` and `s0_` in `examples/smtbmc/demo8`), what ensures that it is the *same* constant chosen both in the outer (i.e. outside `forall`) and inner (inside `forall`) scope?

However it really works, this seems to be a funny way of defining an exists-forall problem. Doesn't it make more sense to just `declare-const` the existentially-quantified variable?

Also, it seems that `$anyseq` and `$allseq` support sequential circuits--I mean, that's the raison d'etre. Yet `backends/smt2/smtio.py` seems to imply that unrolling circuits in exists-forall problems is not supported.  Is that true?

boqwxp:
@cliffordwolf Could you take a look at this when you get a chance?  I have some other plans for exists-forall stuff and it would be great to nail this down.

ShinyKate:
@boqwxp do you have any open questions here or am I good to close?

PR #1741, opened 2020-03-05
Recognize `$anyconst` and `$anyseq` cell types as flags for an exists-forall problem

boqwxp:
Maybe I don't properly understand the difference between `$anyconst` and `$anyseq` vs. `$allconst` and `$allseq`, but according to [this](https://symbiyosys.readthedocs.io/en/latest/verilog.html#unconstrained-variables) they should all be valid ways to declare an existentially-quantified variable for an exists-forall problem.

Yet, `backends/smt2/smt2.cc` only sets `forallmode = true` and emits the (necessary for proper interpretation by `yosys-smtbmc`) `; yosys-smt2-forall` directive if it detects `$allconst` or `$allseq`.  So, an exists-forall problem using only `$anyconst` and `$anyseq` won't properly be detected and elaborated.

If I'm wrong, I'm very interested to read why--and this should probably also go in #1740.

### DIFF
--- a/backends/smt2/smt2.cc
+++ b/backends/smt2/smt2.cc
@@ -1532,7 +1532,7 @@ struct Smt2Backend : public Backend {
 
 		for (auto module : sorted_modules)
 			for (auto cell : module->cells())
-				if (cell->type.in("$allconst", "$allseq"))
+				if (cell->type.in("$anyconst", "$anyseq", "$allconst", "$allseq"))
 					goto found_forall;
 		if (0) {
 	found_forall:


boqwxp:
If any of this is incorrect, please correct me.

I think I better understand now: I did indeed misunderstand the difference between `$anyconst` and `$allconst`.  From what I understand now, it's very simple: `$anyconst` is used to create an existentially-quantified variable, and `$allconst` is used to create a universally-quantified variable.

So, as I understand, when setting up an exists-forall problem, the top module should have no inputs or outputs and the existential and universal variables in the unit under test should be assigned `$anyconst` or `$allconst` as appropriate (or `$anyseq` or `$allseq`, but I'm exclusively contemplating combinational circuits right now).

For example:
```
module demo9;
	wire[7:0] h = $anyconst;
	wire [7:0] i = $allconst;

	wire [7:0] t0 = ((i << 8'b00000010) + 8'b00000011);
	wire trigger = (t0 > h)	&& (h < 8'b00000100);

	always @* begin
		assert(trigger == 1'b1);
	end
endmodule
```
is how to describe the exists-forall problem `\exists \vec{h} \forall \vec {i} : trigger`.


WRT this pull request, I am still unsure what the correct course of action is.  Can one describe a regular SAT problem (not an exists-forall problem) with just `$anyconst`?  Can't they also then describe a universally-quantified SAT problem with just `$allconst`?  If these are true, then simply detecting `$anyconst` *or* `$allconst` is not the correct way to detect an exists-forall problem.

I am also not sure, BTW, if `forallmode = true` *only* applies for exists-forall problems, or if it also applies to universally-quantified SAT problems.

cliffordwolf:
> From what I understand now, it's very simple: $anyconst is used to create an existentially-quantified variable, and $allconst is used to create a universally-quantified variable.

Yes, this is correct. Therefore, if the module contains no $allconst or $allseq then the module does not describe an exist-forall problem but rather a regular "exist-only problem". Thus I'm closing this PR.

> is how to describe the exists-forall problem \exists \vec{h} \forall \vec{i} : trigger(\vec{h}, \vec{i}).

$allconst behaves as $anyconst for everything except the parts that feed into assumptions. Thus, you probably want to do something like the following instead:

```
module demo9;
	wire[7:0] h = $anyconst;
	wire [7:0] i = $allconst;

	wire [7:0] t0 = ((i << 8'b00000010) + 8'b00000011);
	wire trigger = (t0 > h) && (h < 8'b00000100);

	always @* begin
		assume (trigger == 1'b1);
		cover (1);
	end
endmodule
```

(Assuming you are looking for ∃ h ∀ i : trigger(h, i) == 1.)

boqwxp:
Thanks, Claire, that really helps to clarify things!  Also, TIL that UTF has quantifier symbols :).

PR #1746, opened 2020-03-08
Add support for optimizing exists-forall problems.

boqwxp:
* Modifies smt2 backend to recognize `$anyconst` etc. assigned to a wire with the `maximize` or `minimize` attribute and emit a `; yosys-smt2-maximize` or `; yosys-smt2-minimize` directive as appropriate.
* Modifies `backends/smt2/smtbmc.py` and `smtio.py` to recognize those directives and emit a `(maximize ...)` or `(minimize ...)` command at the end of `smt_forall_assert()`, as described in the paper [νZ - An Optimizing SMT Solver](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/nbjorner-nuz.pdf) by Nikolaj Bjørner et al.
* Adds an example `examples/smtbmc/demo9.v` to show how it can be used.

AFAIK this is only supported by z3, but cvc4 might be able to do it, too.

Multiple objectives appear to be supported, but I have added an assertion to limit to only one `maximize` or `minimize` directive for now.  Unless someone else has a good use case for multiple objectives, I prefer to keep the assertion, as multiple objectives might very likely be a bug or user mistake.

Further improvements might include full weighted `assert-soft` constraint support.

### DIFF
--- a/backends/smt2/smt2.cc
+++ b/backends/smt2/smt2.cc
@@ -536,6 +536,14 @@ struct Smt2Worker
 				if (cell->attributes.count("\\reg"))
 					infostr += " " + cell->attributes.at("\\reg").decode_string();
 				decls.push_back(stringf("; yosys-smt2-%s %s#%d %d %s\n", cell->type.c_str() + 1, get_id(module), idcounter, GetSize(cell->getPort("\\Y")), infostr.c_str()));
+				if (cell->getPort("\\Y").is_wire() && cell->getPort("\\Y").as_wire()->get_bool_attribute("\\maximize")){
+					decls.push_back(stringf("; yosys-smt2-maximize %s#%d\n", get_id(module), idcounter));
+					log("Wire %s is maximized\n", cell->getPort("\\Y").as_wire()->name.str().c_str());
+				}
+				else if (cell->getPort("\\Y").is_wire() && cell->getPort("\\Y").as_wire()->get_bool_attribute("\\minimize")){
+					decls.push_back(stringf("; yosys-smt2-minimize %s#%d\n", get_id(module), idcounter));
+					log("Wire %s is minimized\n", cell->getPort("\\Y").as_wire()->name.str().c_str());
+				}
 				makebits(stringf("%s#%d", get_id(module), idcounter), GetSize(cell->getPort("\\Y")), log_signal(cell->getPort("\\Y")));
 				if (cell->type == "$anyseq")
 					ex_input_eq.push_back(stringf("  (= (|%s#%d| state) (|%s#%d| other_state))", get_id(module), idcounter, get_id(module), idcounter));
--- a/backends/smt2/smtbmc.py
+++ b/backends/smt2/smtbmc.py
@@ -1158,6 +1158,8 @@ def smt_forall_assert():
     global asserts_cache_dirty
     asserts_cache_dirty = False
 
+    assert (len(smt.modinfo[topmod].maximize) + len(smt.modinfo[topmod].minimize) <= 1)
+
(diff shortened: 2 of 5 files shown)


backends/smt2/smtio.py
```diff
@@ -696,7 +704,9 @@ def check_sat(self):
                     if msg is not None:
                         print("%s waiting for solver (%s)" % (self.timestamp(), msg), flush=True)
 
-        result = self.read()
+        result = ""
```
boqwxp: These changes are necessary because when solving with a `(maximize ...)` or `(minimize ...)` command, the solver outputs at least one line like:
```
((_ extract 8 1) s0) |-> 3
```
before the line with `sat` or `unsat`.

backends/smt2/smt2.cc
```diff
@@ -536,6 +536,14 @@ struct Smt2Worker
 				if (cell->attributes.count("\\reg"))
 					infostr += " " + cell->attributes.at("\\reg").decode_string();
 				decls.push_back(stringf("; yosys-smt2-%s %s#%d %d %s\n", cell->type.c_str() + 1, get_id(module), idcounter, GetSize(cell->getPort("\\Y")), infostr.c_str()));
+				if (cell->getPort("\\Y").is_wire() && cell->getPort("\\Y").as_wire()->attributes.count("\\maximize") > 0){
```
cliffordwolf: This should use `as_wire()->get_bool_attribute("\\maximize")` instead.

backends/smt2/smtbmc.py
```diff
@@ -1158,6 +1158,8 @@ def smt_forall_assert():
     global asserts_cache_dirty
     asserts_cache_dirty = False
 
+    assert (len(smt.modinfo[topmod].maximize) <= 1 and len(smt.modinfo[topmod].minimize) <= 1 and (len(smt.modinfo[topmod].maximize) + len(smt.modinfo[topmod].minimize)) <= 1)
```
cliffordwolf: I think the last expression (len(..) + len(..) <= 1) implies the first two.

boqwxp: It does in the real numbers, but there is the (perhaps irrelevantly small) chance of overflow which the full expression here avoids.

backends/smt2/smtio.py
cliffordwolf: I have now reverted this one change in master in c9555c9 because it broke other smt-based flows where yosys-smtbmc was not able to correctly identify situations where the solver produced unexpected output. Please create a new PR that keeps reading on the specific text patterns you need yosys-smtbmc to skip, not just anything that is not "sat" or "unsat".

boqwxp: OK, no problem.  I think I will basically just have it skip lines with `|->` for now.

Can you give me an example failing `yosys-smtbmc` flow so I have something to test against?

nakengelhardt: If you have SymbiYosys installed, try the following:
```
cd docs/examples/puzzles
sby -f primegen.sby primes_fail
```
The solver times out relatively quickly and returns `unknown`.

boqwxp: OK, thanks.  Do you have any examples where the value returned is not `sat`, `unsat`, or `unknown`?  If it's always one of those three, I would prefer to keep this behavior (as it's necessary for `yosys-smtbmc` to work with optimizing exists-forall problems) and change the following line to:
```
        while result not in ["sat", "unsat", "unknown"]:
```

WDYT?

boqwxp: See #1815

PR #1749, opened 2020-03-09
Extend `add` command to allow adding $assert cells.

boqwxp:


nakengelhardt:
What does this do that you can't do with `add -mod $assert`?

nakengelhardt:
I thought there should already be a generic way to add cells, but the example I was thinking of was using the C++ interface, and actually I don't see any command that does this. What do you think of a generic `add -cell` that can be used instead?

boqwxp:
As I see it, the main difficulty with a generic `add -cell` is how to specify all ports and parameters.  It certainly can be done, but it's significantly more involved than this and I happen to have the need for this specifically.

I'm totally open to a more general way to add cells, but my selfish interest is to get this merged now, and revisit this later in a more general way.  I do, however, understand if you don't want to merge a band aid like this...

nakengelhardt:
You're right, it would be a lot of boilerplate doing a giant switch for all the `Module::addX` functions and the interface would be unwieldy to use. Maybe a compromise could be `add -formal` for just assert/assume/cover/live/fair? (There's also equiv but it has a different signature and there's already an `equiv_add` pass for it.) That way the wire-finding code would be the same for all arguments and it's just the cell type that has to change.

PR #1751, opened 2020-03-09
Extend `add` command to allow adding $assert cells.

boqwxp:
See #1749 .  I accidentally closed the PR with a premature force push.

@nakengelhardt is this more like what you had in mind?

### DIFF
--- a/passes/cmds/add.cc
+++ b/passes/cmds/add.cc
@@ -22,6 +22,42 @@
 USING_YOSYS_NAMESPACE
 PRIVATE_NAMESPACE_BEGIN
 
+static bool is_formal_celltype(const std::string &celltype)
+{
+	if(celltype == "assert" || celltype == "assume" || celltype == "live" || celltype == "fair" || celltype == "cover")
+		return true;
+	else
+		return false;
+}
+
+static void add_formal(RTLIL::Module *module, const std::string &celltype, const std::string &name, const std::string &enable_name)
+{
+	std::string escaped_name = RTLIL::escape_id(name);
+	std::string escaped_enable_name = (enable_name != "") ? RTLIL::escape_id(enable_name) : "";
+	RTLIL::Wire *wire = module->wire(escaped_name);
+	log_assert(is_formal_celltype(celltype));
+
+	if (wire == nullptr) {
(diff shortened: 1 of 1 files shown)


nakengelhardt:
Claire also suggested adding an optional `-if x` argument to connect the enable port to a wire instead of constant 1.

boqwxp:
OK, code style has been improved.

> Claire also suggested adding an optional `-if x` argument to connect the enable port to a wire instead of constant 1.

Good idea.


How's this?

passes/cmds/add.cc
```diff
@@ -106,6 +142,12 @@ struct AddPass : public Pass {
 		log("selected modules.\n");
 		log("\n");
 		log("\n");
+		log("    add {-assert|-assume|-live|-fair|-cover} [-if <name2>] <name1>\n");
```
nakengelhardt: This description doesn't match what the code does: it says you should use `add -assert -if en foo` but what actually works is `add -assert foo -if en` (which is what we want).

And I'm not sure how the argument parsing ends up there, but if you do attempt to use `add -assert -if en foo` or even just `add -assert -if foo` then no error message is generated but nothing happens either...

boqwxp: Oops, good catch.  This is actually a very interesting bug!

With `add -assert -if en foo`, the '-assert' is parsed correctly, but assigns "-if" to `arg_name`.  This never has the chance to generate an error message, because the test `if(!design->selected_whole_module(module->name))` succeeds.

This test succeeds because the extra arguments `en` and `foo` are handled by `Pass::extra_args()` in `kernel/register.cc`.  `en` does not begin with '-', so it's not detected as an erroneous argument, and `select` is its default `true` because the call to `extra_args()` does not include a final `false` parameter.  So it treats `en` and `foo` as arguments to a `select` command and calls `handle_extra_select_args()`.

This is definitely very buggy behavior but is actually outside the scope of this PR... I will create a separate PR to fix the call to `extra_args()`.

boqwxp:
@nakengelhardt I think this is ready to merge.

nakengelhardt:
Thanks!

PR #1754, opened 2020-03-10
Set AST node source location in more parser rules.

boqwxp:
See #1717 and #1718. This PR sets the `AstSrcLocType` member `location` for more AST nodes generated in the parser.

### DIFF
--- a/frontends/verilog/verilog_parser.y
+++ b/frontends/verilog/verilog_parser.y
@@ -2113,18 +2113,22 @@ simple_behavioral_stmt:
 	lvalue '=' delay expr {
 		AstNode *node = new AstNode(AST_ASSIGN_EQ, $1, $4);
 		ast_stack.back()->children.push_back(node);
+		SET_AST_NODE_LOC(node, @1, @4);
 	} |
 	lvalue TOK_INCREMENT {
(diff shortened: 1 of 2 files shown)


nakengelhardt:
This looks harmless enough, I think I can merge it without bothering Claire even though I don't really understand much about the frontend...

nakengelhardt:
Thanks @eddiehung! You've had a look at this PR then? Any opinions?

eddiehung:
I believe [Travis is failing](https://travis-ci.org/github/YosysHQ/yosys/jobs/661090766?utm_medium=notification&utm_source=github_status) on just the gcc-4.8 run because that compiler version has broken regex.

boqwxp:
Is this blocked by the CI failure on gcc-4.8?  It would be a shame to hold this PR up just for that; really quite a lot of AST nodes are currently missing location information.

Is there a workaround, or a way to disable the test on gcc-4.8?

nakengelhardt:
@boqwxp The previous change in #1718 also broke [mcy gui](https://github.com/YosysHQ/mcy) as we discovered yesterday. @mmicko is working on finding solutions for both issues, and we'll get back to this once we figure out which direction to go.

boqwxp:
Ah, I see.  Thanks for pointing that out.

PR #1755, opened 2020-03-10
Clean up `passes/cmds/add.cc` code style.

boqwxp:
Review of #1751 showed outdated style and pseudo-private member usage in `passes/cmds/add.cc`.  This PR (hopefully) cleans this file up to current standards.

### DIFF
--- a/passes/cmds/add.cc
+++ b/passes/cmds/add.cc
@@ -24,24 +24,23 @@ PRIVATE_NAMESPACE_BEGIN
 
 static void add_wire(RTLIL::Design *design, RTLIL::Module *module, std::string name, int width, bool flag_input, bool flag_output, bool flag_global)
 {
-	RTLIL::Wire *wire = NULL;
+	RTLIL::Wire *wire = nullptr;
 	name = RTLIL::escape_id(name);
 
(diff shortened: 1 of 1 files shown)


passes/cmds/add.cc
```diff
@@ -53,7 +52,6 @@ static void add_wire(RTLIL::Design *design, RTLIL::Module *module, std::string n
 		wire->port_output = flag_output;
 
 		if (flag_input || flag_output) {
-			wire->port_id = module->wires_.size();
 			module->fixup_ports();
 		}
```
boqwxp: As far as I can tell, the assignment to `wire->port_id` is superfluous, as the immediately following call to `module->fixup_ports()` will re-assign all `port_id`s for all module wires that have `port_input` or `port_output`.
So rather than update this section verbatim simply to avoid using `module->wires_` I decided just to remove that assignment.

nakengelhardt:
Great, thanks! I don't know if CI will use `add` much so I should probably test this manually just to be sure, give me a second.

Looks like this doesn't touch any lines in common with #1751 so I should be able to merge them in any order, right?

PR #1756, opened 2020-03-10
Fix bug in `add` command argument processing.

boqwxp:
This is the `add` command, not a `select` command, so extra arguments should not be treated as if they are arguments to a `select` command.

### DIFF
--- a/passes/cmds/add.cc
+++ b/passes/cmds/add.cc
@@ -150,7 +150,7 @@ struct AddPass : public Pass {
 			return;
 		}
 
-		extra_args(args, argidx, design);
+		extra_args(args, argidx, design, false);
 
 		for (auto module : design->modules())
 		{


eddiehung:
Are you aware that the `add` command supports a `[selection]` in order to specify which modules to add this new object to:

https://github.com/YosysHQ/yosys/blob/f310dc31292a576b33b57309b395c819e11db44a/passes/cmds/add.cc#L87-L90

boqwxp:
I see...  I don't know how many peoples' scripts depend on this behavior, but it seems like a violation of separation of concerns.  IMO this should be deprecated and people should use an explicit `select` command if they wish to manipulate/add to only a selection.

Anyway, it certainly cannot be correct that e.g. `add -wire foo 1 ods1opkf se88fslkf` simply does nothing, reporting no error messages.  Maybe the problem is really that there is no feedback either that modules `ods1opkf` and `se88fskf` do not exist to select, or that *no* wire was added.

eddiehung:
If you look across all of Yosys' commands, you'll see that this is quite a common pattern that allows you to select something /just/ for this one command instead of wrapping it in two other commands. Globally, I would say a lot of scripts depend on this feature, even if removing it from `add` does not break CI tests.

boqwxp:
OK.  Then this should be closed, and the issue is really the lack of warnings or errors.  I've opened #1762 for that.

Issue #1745, opened 2020-03-07
Yosys gives bogus error message on ternary operator with missing ? token


zeldin:
Parsing this trivial Verilog module
```
module inverter(input a, output y);

   assign y = (a == 1'b0? 1'b1 : 1'b0);
   
endmodule // inverter
```
yosys gives the error
```
Parsing Verilog input from `/tmp/inverter.v' to AST representation.
/tmp/inverter.v:3: ERROR: Digit larger than 1 used in in base-2 constant.
```
If a space is inserted before the `?`, then the parse is successful.

Reproducable in current git master.

mwkmwkmwk:
This is not a bug — `?` is a valid binary digit in Verilog in some contexts, and it's supposed to be parsed as part of the same lexeme.  At most, this is a misleading error message.

The `?` digit can be used eg. in cases:

```
module top(input [1:0] a, output b);

always @*
        case (a)
                2'b1?: b <= 1;
                2'b?1: b <= 1;
                default: b <= 0;
        endcase

endmodule
```

ZirconiumX:
In those circumstances I think `?` is interpreted as `z`, so `1'b0?` is `1'b0z`, which is a) longer than you specified, and b) probably an unintended bug.

So I'd agree that Yosys is correct to reject this code, but the warning could be improved.

zeldin:
@mwkmwkmwk I don't think this is one of those contexts, and if I do put a `z` or `x` there I do not get the error.

mwkmwkmwk:
This doesn't matter — lexemes are the same everywhere, so the `?` has to be parsed as part of the constant.  Also, you do get the same error for `x` and `z`:

```
module inverter(input a, output y);

   assign y = (a == 1'b0x 1'b1 : 1'b0);

endmodule // inverter
```

fails with the same error.

zeldin:
Fascinating.  So if you remove the `?` between `1'b0x` and `1'b1`, then yosys finds a "digit larger than 1" which it did not find when there is a `?` there.  Is it actually the (second) `b` that it complains about?  But then must scan _past_ whitespace, so then how does having a space before the `?` help in the original example?  Something very odd is going on here.

ZirconiumX:
I don't know the Verilog grammar, but I would imagine here it's interpreted
as `<value> <whitespace>+ ? <whitespace>+ <value> <whitespace>+ :
<whitespace>+ <value>`

Without at least one whitespace character, it would be parsed as part of
[...]

PR #1759, opened 2020-03-11
refixed parsing of constant with comment between size and value

zeldin:
The code to handle comments between size and value in constants by arbitrarily concatenating two TOK_CONSTVAL which was introduced in ee8ad72fd950e1ee204e5c97155a50b8b1445dec has three problems:

* It accepts invalid code; e.g. the expression `12'h4 /*foo*/'b0` is accepted as the hexadecimal constant `12'h4b0` despite being syntactically incorrect
* It generates misleading diagnostics; e.g. the expression `a == 1'b0? 1'b1 : 1'b0`
  yields the diagnostic "Digit larger than 1 used in in base-2 constant" (issue #1745)
* It is incomplete - It does not accept any spaces between the comment and the `'`, and no comments between the base and the digits even though spaces are allowed there.

This PR fixes all three problems by making the three parts of a based constant (size, base, digits) into three separate tokens, allowing the linear whitespace (including comments) between them to be treated as normal inter-token whitespace.

### DIFF
--- a/frontends/verilog/verilog_lexer.l
+++ b/frontends/verilog/verilog_lexer.l
@@ -113,8 +113,10 @@ extern int frontend_verilog_yylex(YYSTYPE *yylval_param, YYLTYPE *yyloc_param);
 %x SYNOPSYS_TRANSLATE_OFF
 %x SYNOPSYS_FLAGS
 %x IMPORT_DPI
+%x BASED_CONST
 
 %%
+	int comment_caller;
 
 <INITIAL,SYNOPSYS_TRANSLATE_OFF>"`file_push "[^\n]* {
 	fn_stack.push_back(current_filename);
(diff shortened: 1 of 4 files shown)


eddiehung:
Thanks for the patch! Can you add tests for all three of those bullet points please?

eddiehung:
And just to be clear, this doesn't "fix" the parsing of `a == 1'b0? 1'b1 : 1'b0` by interpreting the '?' as a ternary operator, but now gives a more relevant error message:

`top.v:2: ERROR: syntax error, unexpected OP_EQ, expecting TOK_PACKAGESEP or '=' or '.' or '['`

zeldin:
@eddiehung I'm unclear how to add a test that something should give an error (two of the three bullet points).  Is there an earlier test of that type I could use for reference?

I could add a test for `4 /*foo*/ 'baa` (with a space before the `'`) which was broken before.  I can't add a test for `4'b /*foo*/ aa` though since the test system tries to compare the result against iverilog which can't handle that expression...

zeldin:
@eddiehung BTW, what is your `top.v`?  The expected error (and the one I'm getting) for that expression is "ERROR: syntax error, unexpected TOK_CONSTVAL" (where the TOK_CONSTVAL is the `1` immediately following the `?`).

eddiehung:
> @eddiehung I'm unclear how to add a test that something should give an error (two of the three bullet points). Is there an earlier test of that type I could use for reference?

Check out the new `logger` pass added recently: https://github.com/YosysHQ/yosys/pull/1705
Be careful about a CI breakage on gcc-4.8 that could result from using this, due to a broken regex implementation: https://github.com/YosysHQ/yosys/pull/1754#issuecomment-597914630

> @eddiehung BTW, what is your top.v? The expected error (and the one I'm getting) for that expression is "ERROR: syntax error, unexpected TOK_CONSTVAL" (where the TOK_CONSTVAL is the 1 immediately following the ?).

2647fe2e4 + `zeldin:constant_with_comment_redux`

`./yosys ../master/top.v` on
```
module top(output a);
assign a == 1'b0? 1'b1 : 1'b0;
endmodule
```
gives:
```
../master/top.v:2: ERROR: syntax error, unexpected OP_EQ, expecting TOK_PACKAGESEP or '=' or '.' or '['
```

zeldin:
I added tests for all three bullet points.

mmicko:
@zeldin  Ah we have problematic case. Please just try running :+1: 
./yosys techlibs/common/gate2lut.v -DLUT_WIDTH=4
Noticed this when running yosys-tests

zeldin:
@mmicko That file has a legitimate syntax error in it.  The first digit is _not_ allowed to be an `_`.
Just replace the `_`  with an actual space and everything is fine.

mmicko:
yeah, understand that, just need to double check with standard

mmicko:
It was invalid so fixed in https://github.com/YosysHQ/yosys/commit/acb341745d436987ad764c9198ff0c6ab63c572c

PR #1761, opened 2020-03-11
opt_merge: speedup

eddiehung:
Previously, a `std::map` was used to keep the cells seen so far, using a custom comparator function which first looks at a cell's type, as well as some other factors like the existence of a keep attribute, before using a custom hash (cached so only run once per cell) to quickly check if same.

Transformed the `std::map` container into a ~`std::unordered_map`~ `dict` that maps directly from hash-string to cell, with a match triggering a more complex equality comparison (as before) to check that they really are identical before merging. Optimised many things inside the comparator and hash function along the way. 

*****

Now I've built this mostly preserving the existing (and potentially expensive) hashing function, however, we need not keep it so complex, since we are resistant to collisions anyway by always doing an full compare. For one, we are SHA1-ing a long string to get a short string, which then gets hashed again into a `size_t` when using the ~`unordered_map`~ `dict`.

### DIFF
--- a/passes/opt/opt_merge.cc
+++ b/passes/opt/opt_merge.cc
@@ -26,7 +26,6 @@
 #include <stdio.h>
 #include <set>
 
-#define USE_CELL_HASH_CACHE
 
 USING_YOSYS_NAMESPACE
 PRIVATE_NAMESPACE_BEGIN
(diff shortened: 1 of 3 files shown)


passes/opt/opt_merge.cc
```diff
@@ -312,42 +273,51 @@ struct OptMergeWorker
 			}
 
 			did_something = false;
-			std::map<RTLIL::Cell*, RTLIL::Cell*, CompareCells> sharemap(CompareCells(this));
+			std::unordered_map<std::string, RTLIL::Cell*> sharemap;
```
cliffordwolf: Why not use `dict`? Otherwise LGTM.

eddiehung: Come to think of it, is there any reason not to hash directly to a `size_t` instead of a SHA-1?

PR #1766, opened 2020-03-13
Regex support for GCC 4.8

mmicko:
Since std::regex does not work (it compiles) on gcc 4.8, there were few possible ways to go
 - boost::regex - it is 1:1 with std::regex but there is a big issue, in case boost is compiled on later gcc and linked with gcc 4.8 then linking will fail, also requires additional library to be linked
 - boost::xpressive - compatible but slightly different syntax for some parts, header only, so no additional library to link

### DIFF
--- a/kernel/log.h
+++ b/kernel/log.h
@@ -26,7 +26,7 @@
 
 // In GCC 4.8 std::regex is not working correctlty, in order to make features
 // using regular expressions to work replacement regex library is used
-#if defined(__GNUC__) && ( __GNUC__ == 4 && __GNUC_MINOR__ <= 8)
+#if defined(__GNUC__) && !defined( __clang__) && ( __GNUC__ == 4 && __GNUC_MINOR__ <= 8)
 	#include <boost/xpressive/xpressive.hpp>
 	#define YS_REGEX_TYPE boost::xpressive::sregex
 	#define YS_REGEX_NS boost::xpressive
--- a/kernel/log.h
+++ b/kernel/log.h
@@ -23,18 +23,21 @@
 #define LOG_H
 
 #include <time.h>
-#if defined(__GNUC__) && ( __GNUC__ == 4 && __GNUC_MINOR__ == 8)
+
+// In GCC 4.8 std::regex is not working correctlty, in order to make features
+// using regular expressions to work replacement regex library is used
+#if defined(__GNUC__) && ( __GNUC__ == 4 && __GNUC_MINOR__ <= 8)
 	#include <boost/xpressive/xpressive.hpp>
-	#define REGEX_TYPE boost::xpressive::sregex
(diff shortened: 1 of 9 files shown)


kernel/log.h
```diff
@@ -23,7 +23,22 @@
 #define LOG_H
 
 #include <time.h>
-#include <regex>
+#if defined(__GNUC__) && ( __GNUC__ == 4 && __GNUC_MINOR__ == 8)
+	#include <boost/xpressive/xpressive.hpp>
+	#define REGEX_TYPE boost::xpressive::sregex
+	#define REGEX_NS boost::xpressive
+	#define REGEX_COMPILE(param) boost::xpressive::sregex::compile(param, \
```
cliffordwolf: Could we prefix those defines with `YS_` and add a comment here explaining why we need this? Also, is it only 4.8 that's broken or also gcc versions that predate 4.8? (not sure it would build on older gcc but it might be cleaner to check for <4.8 istead of =4.8 anyway.)

mmicko: Will add YS_ prefix and explanation, thing is that for version before 4.8 build will be broken anyway due to other c++11 missing features, I can however put <= just in case someone actually using it on ancient compilers.

boqwxp:
This adds code cruft to continue supporting a very old compiler version.  Ubuntu 16.04 LTS has gcc 5.3, [for example](https://packages.ubuntu.com/xenial/gcc).  Even Debian 8 is on [gcc-4.9](https://packages.debian.org/jessie/gcc).  The only issue might be CentOS 7, which is [indeed](https://access.redhat.com/solutions/19458) on gcc-4.8.  But really, users can either build a non-ancient compiler from source or just use a different OS to build/test yosys.

Are there lots of users, or important users, using gcc-4.8?  Does it not make more sense simply to remove support for gcc-4.8?

eddiehung:
> Are there lots of users, or important users, using gcc-4.8? Does it not make more sense simply to remove support for gcc-4.8?

CentOS7 and RHEL7 are quite common in commercial deployments.

boqwxp:
I guess if there are paying commercial customers of Symbiotic EDA who demand to continue using gcc-4.8 then I understand, but I also don't think it's unreasonable to ask CentOS/RHEL 7 folks to install a more modern compiler. GCC 4.8 [was first released](https://gcc.gnu.org/releases.html) 7 years ago, and the last release was almost 5 years ago.

It's not even difficult.  CentOS/RHEL 7 folks can just follow the [official RedHat advice](https://developers.redhat.com/blog/2019/03/05/yum-install-gcc-8-clang-6/) to install a working compiler.

boqwxp:
I really think this merits reconsideration.  Again, it's not difficult for RHEL 7 users to install a working compiler.  I think it makes much more sense to drop gcc-4.8 support.

PR #1767, opened 2020-03-13
IdString: use more ID::*, make them easier to use, speed up IdString::in()

eddiehung:
Most of the patch is for using more `ID::*` entries, which are now easier to add by adopting the same approach as [nextpnr](http://github.com/YosysHQ/nextpnr)'s `kernel/constids.inc`.

Lastly, speed up the variadic template function `IdString::in(...)` by using ~perfect forwarding to prevent IdStrings from being copied on every recursive invocation~ a folding expression hack from [https://articles.emptycrate.com/2016/05/14/folds_in_cpp11_ish.html](https://articles.emptycrate.com/2016/05/14/folds_in_cpp11_ish.html) to eliminate recursion. Profiling shows this to make a noticeable difference as it's called heavily in `opt_expr`


kernel/rtlil.cc
```diff
@@ -40,12 +40,10 @@ int RTLIL::IdString::last_created_idx_[8];
...
 
-IdString RTLIL::ID::A;
-IdString RTLIL::ID::B;
-IdString RTLIL::ID::Y;
-IdString RTLIL::ID::keep;
-IdString RTLIL::ID::whitebox;
-IdString RTLIL::ID::blackbox;
+#define X(_id) IdString RTLIL::ID::_id;
+#include "constids.inc"
+#undef X
```
boqwxp: This is a neat trick. I'm sure I'll use this sometime in the future.

backends/btor/btor.cc
```diff
@@ -394,18 +394,18 @@ struct BtorWorker
 			if (cell->type == "$neg") btor_op = "neg";
 			log_assert(!btor_op.empty());
 
-			int width = GetSize(cell->getPort("\\Y"));
-			width = std::max(width, GetSize(cell->getPort("\\A")));
+			int width = GetSize(cell->getPort(ID::Y));
+			width = std::max(width, GetSize(cell->getPort(ID::A)));
```
boqwxp:
```suggestion
			int width = std::max(GetSize(cell->getPort(ID::A)), GetSize(cell->getPort(ID::Y)));
```

kernel/rtlil.h
```diff
@@ -358,9 +361,14 @@ namespace RTLIL
...
-		bool in(T first, Args... rest) const {
-			return in(first) || in(rest...);
+		template<typename... Args>
+		bool in(Args... args) const {
+			//return in(first) || in(rest...);
+
+            // Credit: https://articles.emptycrate.com/2016/05/14/folds_in_cpp11_ish.html
+            bool result = false;
+            (void) std::initializer_list<int>{ (result = result || in(args), 0)... };
+            return result;
```
boqwxp: Wonky formatting.

Tjoppen:
One small thing: changing IdString::in(IdString rhs) to const& reduces the number of calls to IdString::put_reference() for one example locally from 61.6M to 45.4M as judged by callgrind. The patch is quite simple:

```
diff --git a/kernel/rtlil.h b/kernel/rtlil.h
index 34128d80..5ec43b30 100644
--- a/kernel/rtlil.h
+++ b/kernel/rtlil.h
@@ -369,7 +369,7 @@ namespace RTLIL
                        return result;
                }
 
-               bool in(IdString rhs) const { return *this == rhs; }
+               bool in(const IdString& rhs) const { return *this == rhs; }
                bool in(const char *rhs) const { return *this == rhs; }
                bool in(const std::string &rhs) const { return *this == rhs; }
                bool in(const pool<IdString> &rhs) const { return rhs.count(*this) != 0; }
```

eddiehung:
@Tjoppen Yes indeed! I have avoided that specific optimisation as I want to start a broader conversation about whether `IdString`-s should be passed by value at all (which appears to be convention), since each occurrence triggers a reference increment on copy and a decrement on destruction shortly after. I tried removing the implicit copy constructor to prevent this but there were way too many things to fix. Then I tried disabling the reference counting altogether and got some really small (~1% outside-of-callgrind usertime saving) which made it not seem a good ROI right now.

What are your thoughts?

P.S. Are you not seeing `put_reference` get inlined?

Tjoppen:
> @Tjoppen Yes indeed! I have avoided that specific optimisation as I want to start a broader conversation about whether `IdString`-s should be passed by value at all (which appears to be convention), since each occurrence triggers a reference increment on copy and a decrement on destruction shortly after. I tried removing the implicit copy constructor to prevent this but there were way too many things to fix. Then I tried disabling the reference counting altogether and got some really small (~1% outside-of-callgrind usertime saving) which made it not seem a good ROI right now.
> 
> What are your thoughts?

I see a whole bunch of places where const refs would make a huge improvement in reducing the number of allocations. Not just in IdString but also SigPool, SigMap, SigSet etc.

I also think all getters should be implemented in the headers so that they can be inlined.

> P.S. Are you not seeing `put_reference` get inlined?

I'm using ENABLE_DEBUG at the moment, which may be preventing it from being inlined.

boqwxp:
I think that broader conversation is a good one, but I think this is a great one-line change for now, especially for a PR to "speed up IdString::in()" :).

eddiehung:
> I see a whole bunch of places where const refs would make a huge improvement in reducing the number of allocations. Not just in IdString but also SigPool, SigMap, SigSet etc.

I have a patch for that incoming.

Tjoppen:
Part of the reason why the refcounting is still in when not compiling as DEBUG is because of an #undef YOSYS_NO_IDS_REFCNT in rtlil.h

eddiehung:
> Part of the reason why the refcounting is still in when not compiling as DEBUG is because of an #undef YOSYS_NO_IDS_REFCNT in rtlil.h

No, refcounting is always in, `ENABLE_DEBUG` or not. I have no idea what the value of these lines are since they are always in force:
https://github.com/YosysHQ/yosys/blob/3e46faa58c86fb85ee722abcaef5588f76bde564/kernel/rtlil.h#L79-L82
For my no-ref-count-gave-~1%-improvement-experiment, I had to replace the `#undef YOSYS_NO_IDS_REFCNT` with a `#define`.

PR #1769, opened 2020-03-14
Clean up code style and pseudo-private member usage in `passes/cmds/select.cc`

boqwxp:
Also adds a `contains()` member function to `ObjRange` in `kernel/rtlil.h`.

I think a benchmark comparison with and without this PR must be done before merging.  My concern is that in the original code, `it` has the `RTLIL::IdString` local, next to the `RTLIL::Module *` in memory.  In the new code that uses the `ObjRange` accessor functions, the name that is passed to many functions no longer local.  Instead, it's an indirect jump away, at e.g. `wire->name`.

### DIFF
--- a/passes/cmds/select.cc
+++ b/passes/cmds/select.cc
@@ -228,7 +228,7 @@ static void select_op_submod(RTLIL::Design *design, RTLIL::Selection &lhs)
 		{
 			for (auto cell : mod->cells())
 			{
-				if (!design->has(cell->type))
+				if (design->module(cell->type) == nullptr)
 					continue;
 				lhs.selected_modules.insert(cell->type);
 			}
@@ -242,7 +242,7 @@ static void select_op_cells_to_modules(RTLIL::Design *design, RTLIL::Selection &
 	for (auto mod : design->modules())
 		if (lhs.selected_module(mod->name))
 			for (auto cell : mod->cells())
-				if (lhs.selected_member(mod->name, cell->name) && design->has(cell->type))
+				if (lhs.selected_member(mod->name, cell->name) && (design->module(cell->type) != nullptr))
 					new_sel.selected_modules.insert(cell->type);
 	lhs = new_sel;
 }
@@ -252,7 +252,7 @@ static void select_op_module_to_cells(RTLIL::Design *design, RTLIL::Selection &l
 	RTLIL::Selection new_sel(false);
 	for (auto mod : design->modules())
 		for (auto cell : mod->cells())
-			if (design->has(cell->type) && lhs.selected_whole_module(cell->type))
+			if ((design->module(cell->type) != nullptr) && lhs.selected_whole_module(cell->type))
 				new_sel.selected_members[mod->name].insert(cell->name);
 	lhs = new_sel;
 }
@@ -333,7 +333,7 @@ static void select_op_diff(RTLIL::Design *design, RTLIL::Selection &lhs, const R
 
 	for (auto &it : rhs.selected_members)
 	{
-		if (!design->has(it.first))
+		if (design->module(it.first) == nullptr)
 			continue;
 
 		RTLIL::Module *mod = design->module(it.first);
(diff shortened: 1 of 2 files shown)


boqwxp:
@eddiehung would you mind running the benchmark suite on this?  Or, alternatively, could you point me to where I can download it?

passes/cmds/select.cc
```diff
@@ -1581,15 +1578,15 @@ struct CdPass : public Pass {
...
-			RTLIL::Module *module = NULL;
-			if (design->modules_.count(design->selected_active_module) > 0)
-				module = design->modules_.at(design->selected_active_module);
-			if (module != NULL && module->cells_.count(modname) > 0)
-				modname = module->cells_.at(modname)->type.str();
+		if (!design->has(modname) == 0 && !design->selected_active_module.empty()) {
+			RTLIL::Module *module = nullptr;
+			if (design->has(design->selected_active_module) > 0)
+				module = design->module(design->selected_active_module);
+			if (module != nullptr && module->cells().contains(modname))
```
eddiehung: I don't think we need a `ObjRange::contains()` method; one can look at whether the `Module::cell()` returns a null pointer or not.

passes/cmds/select.cc
```diff
@@ -1581,15 +1578,15 @@ struct CdPass : public Pass {
...
+				modname = module->cell(modname)->type.str();
 		}
 
-		if (design->modules_.count(modname) > 0) {
+		if (design->has(modname)) {
```
eddiehung: In fact, I would go as far as to deprecate `Design::has()` in favour of `Design::module()` as the latter is a superset of the former.

eddiehung:
> My concern is that in the original code, it has the RTLIL::IdString local, next to the RTLIL::Module * in memory...

I'm not sure what you mean by this. Can you elaborate? As it stands, I don't see any performance concern. 

> @eddiehung would you mind running the benchmark suite on this? Or, alternatively, could you point me to where I can download it?

Unfortunately, we don't really have a decent benchmark suite for measuring runtime, we have a repository for benchmarks at github.com/YosysHQ/yosys-bench, but not any decent infrastructure to run it. I just run a few ad-hoc ones locally (if that) which is far from rigorous. @mithro to the rescue?

passes/cmds/select.cc
boqwxp: Got it.

passes/cmds/select.cc
boqwxp: OK, I will make sure my code uses this style.

boqwxp:
I mean that iterating over `dict<RTLIL::IdString, RTLIL::Cell*> cells_` might be more performant for code that mostly passes that `IdString` and `Cell*` to other helper functions.  Maybe I'm wrong, but I would guess that the `IdString` and `Cell*` are close to each other in memory, maybe even on the same cache line.  If instead that `IdString` needs to be obtained by dereferencing that `Cell*`, that indirect memory access could take a relatively long time.

> we have a repository for benchmarks at YosysHQ/yosys-bench, but not any decent infrastructure to run it. 

Got it.  I'll check it out and hack something together just to see if there's any impact.

eddiehung:
> I mean that iterating over `dict<RTLIL::IdString, RTLIL::Cell*> cells_` might be more performant for code that mostly passes that `IdString` and `Cell*` to other helper functions. Maybe I'm wrong, but I would guess that the `IdString` and `Cell*` are close to each other in memory, maybe even on the same cache line. If instead that `IdString` needs to be obtained by dereferencing that `Cell*`, that indirect memory access could take a relatively long time.

Sorry, still not following. If you want to lookup a `Cell*` from its `IdString`-name, then that's O(1). Can you give an example where `IdString` and `Cell*` are passed to a helper function?

boqwxp:
OK, I looked over the changes and I can't find any instance of what I feared.  All new `->name` accesses don't seem to incur any additional indirect memory access that wasn't already there.  So this probably won't have a big impact on performance.

eddiehung:
> OK, I looked over the changes and I can't find any instance of what I feared. All new `->name` accesses don't seem to incur any additional indirect memory access that wasn't already there. So this probably won't have a big impact on performance.

Okay, I think I'm getting what you mean now. For https://github.com/YosysHQ/yosys/pull/1769/files#diff-8f49103b623e40720504e1f73fb0f31aL155-R157 it was previously returning a `std::pair<>` containing the `IdString` name as well as a `Module*`. Doing `std::pair<>.first` is arguably cheaper than `(Module*)->name` but I doubt you could get measure anything unless you had a truly humongous `select` benchmark, and I think it is superseded by cleaner more meaningful code.

PR #1774, opened 2020-03-16
Add `exec` command to allow running shell commands from inside Yosys scripts

boqwxp:


ZirconiumX:
This gives me the creeps a little. Consider that Yosys already has a feature for executing arbitrary Yosys scripts from a wire in `techmap` and I *think* `script -wire`. Now you have the potential of executing arbitrary code from Verilog. Is that such a smart idea?

ZirconiumX:
Actually never mind, you can already achieve external shell access by prefixing your command with `!`.

daveshah1:
The risk is low given that in both cases need some degree of control over the Yosys script, whereas stuff like this is possible in Verilog directly (please don't actually run this!)

```verilog
module test;
  import "DPI-C" function fork = integer fork ();
  
  genvar i;
  generate
	 for (i=0;1;i++)
  		integer foo = fork();
  endgenerate
endmodule
```

I believe you can already run shell commands from within a Yosys script by prefixing with ! anyway.

nakengelhardt:
Yes, `!` already allows executing shell commands. But it doesn't allow checking the return value, which we were complaining about in JF just last month IIRC! So this is actually by accident implementing a wished-for feature.

boqwxp:
Aha, obviously I did not know that `!` existed as a command.  It should be listed in `help`.

WRT expected return value or stdout contents, I'm not sure what the best syntax would be for the `!` operator.  Maybe it makes sense to merge this PR, anyway, and have a bit of duplication of effort, because it has a decent way to specify those things.

passes/cmds/exec.cc
```diff
@@ -0,0 +1,190 @@
...
+		log("\n");
+		log("\n");
+		log("    Example: exec -q -expect-return 0 -- echo \"bananapie\" | grep \"nana\"\n");
+		log("\n");
+		log("\n");
+	}
+	void execute(std::vector<std::string> args, RTLIL::Design *design) YS_OVERRIDE
+	{
+		std::string cmd = "";
+		char buf[4096] = {};
```
mmicko: Guess making buffer this big is not needed. Usually output of executable will be quite limited, and later you anyway loop in case it is larger then buffer. Understand this is making it more performant but not sure if needed.

passes/cmds/exec.cc
```diff
@@ -0,0 +1,190 @@
...
+		if(WIFEXITED(status)) {
+		    retval = WEXITSTATUS(status);
+		}
+		else if(WIFSIGNALED(status)) {
+		    retval = WTERMSIG(status);
+		}
+		else if(WIFSTOPPED(status)) {
+		    retval = WSTOPSIG(status);
+		}
+
```
mmicko: Since this is not supported on all platforms, I would suggest something like https://github.com/YosysHQ/yosys/blob/master/kernel/yosys.cc#L363-L369

passes/cmds/exec.cc
```diff
@@ -0,0 +1,190 @@
...
+ *  WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ *  ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ *  OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *
+ */
+
+#include "kernel/register.h"
+#include "kernel/log.h"
+#include <cstdio>
+#include <sys/wait.h>
```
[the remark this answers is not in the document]
boqwxp: It looks like `WIFEXITED` etc. which are defined in this header on POSIX systems does not exist anywhere in MinGW.  git defines them directly: https://github.com/git-for-windows/git/blob/4861720f7d6e2ae6e2bab5945c9900e5f1f328bc/compat/mingw.h so I guess I will, too.

passes/cmds/exec.cc
boqwxp: OK, sure, I'll change to 1KiB.  I don't think it makes much difference either way.

passes/cmds/exec.cc
boqwxp: Should be taken care of by defining these macros manually for Win32 platforms.

passes/cmds/exec.cc
```diff
@@ -0,0 +1,203 @@
+/*
+ *  yosys -- Yosys Open SYnthesis Suite
+ *
+ *  Copyright (C) 2012  Clifford Wolf <clifford@clifford.at>
```
nakengelhardt: Dunno who that is but I'm pretty sure it's not you :D

ckqee:
Does the TCL option not allow the use of exec from TCL?
I'm currently building from master.

passes/cmds/exec.cc
boqwxp: Sorry, I just copied the header from another existing file.  The email address for this nym is not super stable (or frequently checked), and I am certainly not going to exercise any sort of legal rights, so I hereby assign the copyright for this PR to Claire Wolf.  In the alternative, if a copyright assignment is not legally enforceable in a given jurisdiction, I offer the ISC license as usual.

boqwxp:
@ckqee I dunno about the `tcl` command, but even if it does:
- it doesn't support expected return value or stdout contents
- it (looks like it) requires a separate file, so doesn't lend itself well to shell commands directly in yosys scripts

Issue #1728, opened 2020-03-01
Use a proper interface for Verilog source location information


boqwxp:
Follow up to #1718 .

> Maybe the best strategy is to keep the use_internal_line_num(), set_line_num(), and 
> get_line_num() interface, add a set_precise_location() and get_precise_location() interface, and
> update all existing references to linenum to use get_line_num() or get_precise_location() as
> appropriate.

PR #1775, opened 2020-03-17
Add precise locations for asserts

huaixv:
This commit add precise locations for `assume`, `assert`, etc. in verilog parser rules. And it also fills `location` member in new assert nodes created during simplification.

A better approach may be implementing the new location interfaces mentioned in #1728 and do initialization in `AstNode` constructor.

### DIFF
--- a/frontends/verilog/verilog_parser.y
+++ b/frontends/verilog/verilog_parser.y
@@ -1955,6 +1955,7 @@ assert:
 			delete $5;
 		} else {
 			AstNode *node = new AstNode(assume_asserts_mode ? AST_ASSUME : AST_ASSERT, $5);
+			SET_AST_NODE_LOC(node, @1, @6);
 			if ($1 != nullptr)
 				node->str = *$1;
 			ast_stack.back()->children.push_back(node);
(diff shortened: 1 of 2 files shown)


nakengelhardt:
@boqwxp Can you take a look at this?

frontends/verilog/verilog_parser.y
```diff
@@ -1955,6 +1955,7 @@ assert:
 			delete $5;
 		} else {
 			AstNode *node = new AstNode(assume_asserts_mode ? AST_ASSUME : AST_ASSERT, $5);
+			SET_AST_NODE_LOC(node, @1, @7);
```
boqwxp:
```suggestion
			SET_AST_NODE_LOC(node, @1, @6);
```
The trailing semicolon probably should not be counted. The same goes for the following rules.

huaixv: Corrected

boqwxp:
Looks good.  cd82ccd2581f93b17e2de017b7a2504d7733f2df is ready to merge IMO.

nakengelhardt:
Thanks!

PR #1778, opened 2020-03-17
Add support for SystemVerilog-style `define to Verilog frontend

rswarbrick:
This patch should support things like

    `define foo(a, b = 3, c)   a+b+c
    `foo(1, ,2)

which will evaluate to 1+3+2. It also spots mistakes like

    `foo(1)

(the 3rd argument doesn't have a default value, so a call site is
required to set it).

Most of the patch is a simple parser for the format in preproc.cc, but
I've also taken the opportunity to wrap up the "name -> definition"
map in a type, rather than use multiple std::map's.

Since this type needs to be visible to code that touches defines, I've
pulled it (and the frontend_verilog_preproc declaration) out into a
new file at frontends/verilog/preproc.h and included that where
necessary.

Finally, the patch adds a few tests in tests/various to check that we
are parsing everything correctly.

### DIFF
--- a/frontends/verilog/verilog_frontend.cc
+++ b/frontends/verilog/verilog_frontend.cc
@@ -27,6 +27,7 @@
  */
 
 #include "verilog_frontend.h"
+#include "preproc.h"
 #include "kernel/yosys.h"
 #include "libs/sha1/sha1.h"
 #include <stdarg.h>
@@ -253,7 +254,8 @@ struct VerilogFrontend : public Frontend {
 		bool flag_defer = false;
 		bool flag_noblackbox = false;
 		bool flag_nowb = false;
-		std::map<std::string, std::string> defines_map;
+		define_map_t defines_map;
+
 		std::list<std::string> include_dirs;
 		std::list<std::string> attributes;
 
@@ -369,7 +371,7 @@ struct VerilogFrontend : public Frontend {
 			}
 			if (arg == "-lib") {
 				lib_mode = true;
-				defines_map["BLACKBOX"] = string();
+				defines_map.add("BLACKBOX", "");
 				continue;
 			}
 			if (arg == "-nowb") {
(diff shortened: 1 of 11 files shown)


frontends/verilog/verilog_frontend.cc
```diff
@@ -353,7 +355,7 @@ struct VerilogFrontend : public Frontend {
 			}
 			if (arg == "-lib") {
 				lib_mode = true;
-				defines_map["BLACKBOX"] = string();
+                defines_map.add("BLACKBOX", "");
```
eddiehung: Spacing

rswarbrick: Wow, I don't know what happened there. Anyway, fixed.

rswarbrick:
Hi there! Any progress on reviewing this change? It's been a week since I pushed up the last version.

nakengelhardt:
It's on the list! Please patient a little more. Anything that touches the verilog frontend requires extra scrutiny since it's a very sensitive and fragile code area, and has to be reviewed by @cliffordwolf specifically, but she's still out of office for a few more days!

rswarbrick:
Oh, great, no problem. I just wanted to make sure this hadn't fallen through a crack.

frontends/verilog/verilog_frontend.cc
```diff
@@ -237,7 +238,8 @@ struct VerilogFrontend : public Frontend {
 		bool flag_defer = false;
 		bool flag_noblackbox = false;
 		bool flag_nowb = false;
-		std::map<std::string, std::string> defines_map;
+        define_map_t defines_map;
```
cliffordwolf: Indenting

rswarbrick: Gah, Emacs let me down again! Fixed in latest push.

rswarbrick:
Thanks for the review. I've just rebased onto master (and fixed the indentation bug).

PR #1783, opened 2020-03-19
Clean up pseudo-private member usage in `frontends/ast/ast.cc`.

boqwxp:


nakengelhardt:
I think for the verilog frontend the calculation for updating the code might be different. It's such a fragile code area, we try to minimize changes there, especially changes that might be difficult to revert later after several commits have been added on top of them. This kind of style update probably doesn't bring enough benefits to be worth the risk here.

boqwxp:
Roger that.  I'll figure out what broke the build--likely the changes around line 1459--and submit a less ambitious patch for a better chance of getting merged.

boqwxp:
@nakengelhardt I was right, those were the offending changes.  60405939943cd812e146b84848be8bc9307702db reverts those to a much less aggressive patch.
Note that `wires_.erase()` is now `mod->remove()`.  This also `delete`s the dummy interface port, in addition to `erase()`ing it from `wires_`.  AFAIK this wasn't `delete`d before, and there is no problem `delete`ing it now, so I think this PR also fixes a memory leak.

Other than that, the PR is now quite straightforward, so I don't think there's too much risk of anything breaking again... (famous last words :))

nakengelhardt:
Great! @eddiehung volunteered to review this one last week since I don't know anything about the frontend.

frontends/ast/ast.cc
```diff
@@ -1457,8 +1457,10 @@ RTLIL::IdString AstModule::derive(RTLIL::Design *design, dict<RTLIL::IdString, R
 
 		// Now that the interfaces have been exploded, we can delete the dummy port related to every interface.
 		for(auto &intf : interfaces) {
-			if(mod->wires_.count(intf.first)) {
-				mod->wires_.erase(intf.first);
+			if(mod->wire(intf.first) != nullptr) {
+				pool<RTLIL::Wire*> to_remove;
+				to_remove.insert(mod->wire(intf.first));
```
eddiehung: Removing wires is generally an expensive operation. Is it possible to collect all to-be-removed wires and remove them once outside the loop?

boqwxp: That is exactly what I thought, and I tried to do that at first.  But somehow that broke the build. :-\

See the Travis logs with and without 60405939943cd812e146b84848be8bc9307702db

eddiehung: Sorry, I didn't notice you had tried that already. I figured out what was going on, let me add it as a comment.

frontends/ast/ast.cc
```diff
@@ -1457,8 +1457,10 @@ RTLIL::IdString AstModule::derive(RTLIL::Design *design, dict<RTLIL::IdString, R
...
+				mod->remove(to_remove);
```
eddiehung:
```suggestion
				// Normally, removing wires would be batched together as it's an
				//   expensive operation, however, in this case doing so would mean
				//   that a cell with the same name cannot be created (below)...
				// Since we won't expect many interfaces to exist in a module,
				//   we can let this slide...				
				pool<RTLIL::Wire*> to_remove;
				to_remove.insert(mod->wire(intf.first));
				mod->remove(to_remove);
```

frontends/ast/ast.cc
```diff
@@ -1469,6 +1471,7 @@ RTLIL::IdString AstModule::derive(RTLIL::Design *design, dict<RTLIL::IdString, R
 				log_error("No port with matching name found (%s) in %s. Stopping\n", log_id(intf.first), modname.c_str());
 			}
 		}
+		mod->fixup_ports();
```
eddiehung:
```suggestion
```

boqwxp: Oup, forgot to remove that in 60405939943cd812e146b84848be8bc9307702db

PR #1786, opened 2020-03-19
Clean up pseudo-private member usage in `passes/hierarchy/hierarchy.cc`.

boqwxp:


passes/hierarchy/hierarchy.cc
```diff
@@ -493,7 +490,7 @@ void hierarchy_clean(RTLIL::Design *design, RTLIL::Module *top, bool purge_lib)
 		if (!purge_lib && mod->get_blackbox_attribute())
 			continue;
 		log("Removing unused module `%s'.\n", mod->name.c_str());
-		design->modules_.erase(mod->name);
+		design->remove(mod);
```
eddiehung:
```suggestion
```
`Design::remove()` will have called `delete` already. Perhaps this is why CI is failing?

boqwxp: D'oh!  You're probably right, although my preferred change is to use the proper `remove()` method and get rid of the manual `delete`.

eddiehung: You're doing the right thing! I had meant to suggest removing the `delete` statement. Sorry about that.

PR #1790, opened 2020-03-20
opt_expr: optimise $xor/$xnor/$_XOR_/$_XNOR_ -s with constant inputs

eddiehung:
Interestingly, it seems that `opt_expr` will only optimise word-level `$and`/`$or` cells if they are one-bit wide:
https://github.com/YosysHQ/yosys/blob/f828cb5132b5d1a2c2d4d26ffaac6d492c7cc69e/passes/opt/opt_expr.cc#L443
https://github.com/YosysHQ/yosys/blob/f828cb5132b5d1a2c2d4d26ffaac6d492c7cc69e/passes/opt/opt_expr.cc#L449

Thus something like: `assign y = a & 2'b11` would not get optimised until simplemap breaks it down to gate-level/1-bit wide `$_AND_` cells.

In the case of `$xor`/`$xnor`/`$_XOR_`/`$_XNOR_` it seems that no optimisation was done at all, so this PR fixes that (but still only operates on 1-bit word cells).

### DIFF
--- a/passes/opt/opt_expr.cc
+++ b/passes/opt/opt_expr.cc
@@ -502,7 +502,7 @@ void replace_const_cells(RTLIL::Design *design, RTLIL::Module *module, bool cons
 			SigBit sig_b = assign_map(cell->getPort(ID::B));
 			if (!sig_a.wire)
 				std::swap(sig_a, sig_b);
-			if (sig_b == State::S0 || sig_b == State::S1)
+			if (sig_b == State::S0 || sig_b == State::S1) {
 				if (cell->type.in(ID($xor), ID($_XOR_))) {
 					cover("opt.opt_expr.xor_buffer");
 					replace_cell(assign_map, module, cell, "xor_buffer", ID::Y, sig_b == State::S1 ? module->NotGate(NEW_ID, sig_a) : sig_a);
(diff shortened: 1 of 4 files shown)


passes/opt/opt_expr.cc
```diff
@@ -842,8 +878,6 @@ void replace_const_cells(RTLIL::Design *design, RTLIL::Module *module, bool cons
 			if (input.match("11")) ACTION_DO_Y(0);
 			if (input.match(" *")) ACTION_DO_Y(x);
 			if (input.match("* ")) ACTION_DO_Y(x);
-			if (input.match(" 0")) ACTION_DO(ID::Y, input.extract(1, 1));
-			if (input.match("0 ")) ACTION_DO(ID::Y, input.extract(0, 1));
```
boqwxp: I'm just perusing these changes; I haven't sat down and fully understood this part of the code.  But this strikes me as interesting--why are these removed now?  Are they guaranteed not to apply because of the changes above?  Why?  Should there be an assert there (maybe `ifndef NDEBUG`-guarded) to check these never happen?

eddiehung: These have become redundant, because: https://github.com/YosysHQ/yosys/blob/0c0dc4ffc3c907fbe6973c925e9de07c2286cd6a/passes/opt/opt_expr.cc#L499-L505
will already check if either input is a `1'b0` or `1'b1` and optimise it before execution reaches this point.

I believe there are more redundancies in the other `match()` calls above and below here, but that's out of the scope of this PR.

PR #1794, opened 2020-03-20
ice40: Map unmapped 'mince' DFFs to gate level

daveshah1:
Otherwise, they are not mapped to `SB_DFF` before ABC, which causes a 'no loops' failure in `abc9` (and would also mean that required times etc were wrong).

cc @smunaut


smunaut:
I can confirm that fixes the issue I was having.

PR #1739, opened 2020-03-05
ice40: fix specify for -device {lp,u}

eddiehung:
Fixes #1738

PR #1796, opened 2020-03-20
tests: avoid using bash4's assoc arrays as macOS no likey

eddiehung:
Resolves https://github.com/YosysHQ/yosys/pull/1739#issuecomment-599516034

### DIFF
--- a/tests/arch/run-test.sh
+++ b/tests/arch/run-test.sh
@@ -2,14 +2,16 @@
 
 set -e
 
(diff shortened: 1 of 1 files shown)


nakengelhardt:
I pulled in bash 4 via Homebrew in #1780 to fix this already, would you prefer to avoid that anyway?

eddiehung:
Fixed already by https://github.com/YosysHQ/yosys/pull/1780

PR #1797, opened 2020-03-21
Keep file information when emitting firrtl

sahandKashani:
Hi,

This PR preserves yosys' `\src` attribute when emitting firrtl. The ilang backend does something similar, but this functionality was missing in the firrtl backend. It's a shame to lose this information downstream in other tools as it would help link back netlist transformations back to their origin signals.

Best,
Sahand


backends/firrtl/firrtl.cc
```diff
@@ -194,7 +265,7 @@ struct FirrtlWorker
 			}
 		 }
 		// We need a default constructor for the dict insert.
-	   memory() : pCell(0), read_latency(0), write_latency(1), init_file(""), init_file_srcFileSpec(""){}
+		 memory() : pCell(0), read_latency(0), write_latency(1), init_file(""), init_file_srcFileSpec(""){}
```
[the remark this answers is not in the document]
sahandKashani: Thanks for pointing this out.

backends/firrtl/firrtl.cc
```diff
@@ -42,6 +42,77 @@ static const FDirection FD_OUT = 0x2;
 static const FDirection FD_INOUT = 0x3;
 static const int FIRRTL_MAX_DSH_WIDTH_ERROR = 20; // For historic reasons, this is actually one greater than the maximum allowed shift width
 
+// Shamelessly copied from ilang_backend.cc. Something better is surely possible here.
```
eddiehung: If this is character for character identical then why not just call that function `ILANG_BACKEND::dump_const()`?

backends/firrtl/firrtl.cc
```diff
@@ -42,6 +42,77 @@ static const FDirection FD_OUT = 0x2;
...
+				f << str[i];
+		}
+		f << stringf("\"");
+	}
+}
+
+std::string getFileinfo(const dict<RTLIL::IdString, RTLIL::Const> &attributes)
+{
+	std::ostringstream fileinfo;
+	for (auto &it : attributes) {
```
eddiehung: Why iterate through all entries when you can do a lookup?
Why not use `{Cell,Wire}::get_src_attribute()` (which is implemented in the base class here: https://github.com/YosysHQ/yosys/blob/3e46faa58c86fb85ee722abcaef5588f76bde564/kernel/rtlil.cc#L284-L290)

backends/firrtl/firrtl.cc
sahandKashani: The various yosys backends seemed independent to me so I didn't want to introduce a dependency between the firrtl and ilang backend (though I admit I didn't look very deeply). But I suppose the ilang backend has little chance of changing and I could just call its `dump_const()` function (it is indeed a character for character copy).

backends/firrtl/firrtl.cc
sahandKashani: Originally I intended to store all yosys' attributes in the target firrtl file, hence why I was looping over them all. But in the end I just ended up keeping the src attribute (firrtl doesn't support other attributes than `Info` inline after each statement that is outputted).

Thanks for pointing this function out. Hadn't spotted that one before.

backends/firrtl/firrtl.cc
```diff
@@ -42,6 +43,23 @@ static const FDirection FD_OUT = 0x2;
...
+	std::string src(design_entity->get_src_attribute());
+
+	std::ostringstream fileinfo;
+	if (!src.empty()) {
+		fileinfo << "@[" << src << "]";
+	}
+
+	// Remove quotes from src attribute as firrtl automatically escapes and
+	// double-quotes them.
+	std::string fileinfo_str(fileinfo.str());
```
boqwxp:
```suggestion
	std::string fileinfo_str = "@[" + src + "]";
```

backends/firrtl/firrtl.cc
```diff
@@ -329,7 +348,8 @@ struct FirrtlWorker
 			log_warning("No instance for %s.%s\n", cell_type.c_str(), cell_name.c_str());
 			return;
 		}
-		wire_exprs.push_back(stringf("%s" "inst %s%s of %s", indent.c_str(), cell_name.c_str(), cell_name_comment.c_str(), instanceOf.c_str()));
+		auto cellFileinfo = getFileinfo(cell);
```
boqwxp: FYI project maintainers don't like `auto` for simple types like `std::string`

backends/firrtl/firrtl.cc
```diff
@@ -394,12 +414,15 @@ struct FirrtlWorker
 
 	void run()
 	{
-		f << stringf("  module %s:\n", make_id(module->name));
+		auto moduleFileinfo = getFileinfo(module);
```
boqwxp: Same here.

backends/firrtl/firrtl.cc
```diff
@@ -394,12 +414,15 @@ struct FirrtlWorker
...
+		f << stringf("  module %s: %s\n", make_id(module->name), moduleFileinfo.c_str());
 		vector<string> port_decls, wire_decls, cell_exprs, wire_exprs;
 
 		for (auto wire : module->wires())
 		{
 			const auto wireName = make_id(wire->name);
+			auto wireFileinfo = getFileinfo(wire);
```
boqwxp: And here

backends/firrtl/firrtl.cc
```diff
@@ -1123,7 +1155,8 @@ struct FirrtlBackend : public Backend {
 		if (top == nullptr)
 			top = last;
 
-		*f << stringf("circuit %s:\n", make_id(top->name));
+		auto circuitFileinfo = getFileinfo(top);
```
boqwxp: And here

backends/firrtl/firrtl.cc
sahandKashani: That's indeed way simpler :)
I'll update it.

backends/firrtl/firrtl.cc
sahandKashani: Done.

sahandKashani:
> Superficially LGTM after these trivial changes, but I have not tested it :).

I tested this with some small circuits and a much larger pipelined floating-point core design. The source attributes were correctly generated and firrtl could automatically ingest them downstream :)

ucbjrl:
Passes all my tests.

backends/firrtl/firrtl.cc
eddiehung: I would say the `ILANG_BACKEND::*` are okay to call from outside because ilang is probably the most native format (and other files like `kernel/log.cc` and `passes/cmds/bugpoint.cc` also call `ILANG_BACKEND::*` functions)

backends/firrtl/firrtl.cc
```diff
@@ -42,6 +43,18 @@ static const FDirection FD_OUT = 0x2;
...
 static const int FIRRTL_MAX_DSH_WIDTH_ERROR = 20; // For historic reasons, this is actually one greater than the maximum allowed shift width
 
+std::string getFileinfo(const RTLIL::AttrObject *design_entity)
+{
+	std::string src(design_entity->get_src_attribute());
+	std::string fileinfo_str = src.empty() ? "" : "@[" + src + "]";
+
+	// Remove quotes from src attribute as firrtl automatically escapes and
+	// double-quotes them.
+	fileinfo_str.erase(std::remove(fileinfo_str.begin(), fileinfo_str.end(), '\"'), fileinfo_str.end());
```
whitequark, line 53: Where do the quotes in the attribute come from?

sahandKashani: You are entirely right to bring this up.  In the past I was calling the `ILANG_BACKEND::dump_const()` function to dump values to an output stream. When these values were strings, the function used to add quotes around them. The firrtl compiler, in turn, escaped these quotes and the result was double-quoted strings which were hard to read.

However I'm now calling `design_entity->get_src_attribute()` directly and there is no additional quote being put around it, so we can safely remove the unnecessary call to str.erase() here.

backends/firrtl/firrtl.cc
```diff
@@ -505,16 +499,16 @@ struct FirrtlWorker
...
+				register_reverse_wire_map(y_id, cell->getPort(ID::Y));
 
 				continue;
 			}
-			if (cell->type.in("$add", "$sub", "$mul", "$div", "$mod", "$xor", "$xnor", "$and", "$or", "$eq", "$eqx",
-				"$gt", "$ge", "$lt", "$le", "$ne", "$nex", "$shr", "$sshr", "$sshl", "$shl",
-				"$logic_and", "$logic_or", "$pow"))
+			if (cell->type.in(ID($add), ID($sub), ID($mul), ID($div), ID($mod), ID($xor), ID($xnor), ID($and), ID($or), ID($eq), ID($eqx),
+												ID($gt), ID($ge), ID($lt), ID($le), ID($ne), ID($nex), ID($shr), ID($sshr), ID($sshl), ID($shl),
+												ID($logic_and), ID($logic_or), ID($pow)))
```
boqwxp, line 508: Messed up indentation

backends/firrtl/firrtl.cc
```diff
@@ -599,8 +593,8 @@ struct FirrtlWorker
 					primop = "eq";
 					always_uint = true;
 					firrtl_width = 1;
-				}
-				else if ((cell->type == "$ne") | (cell->type == "$nex")) {
+			    }
```
boqwxp: Please use tabs for indentation.

backends/firrtl/firrtl.cc
sahandKashani: Sorry about that. I knew yosys used tabs for indentation and the code looked perfectly aligned in my editor, but I was having tabs displayed as 2 characters. I read through the yosys style guide and indented the long statement with 2 extra tabs as specified.
https://github.com/YosysHQ/yosys/blob/master/CodingReadme

backends/firrtl/firrtl.cc
sahandKashani: Same as above. Fixed now after reading the yosys style guide. Thanks for pointing it out.

PR #1812, opened 2020-03-23
opt_whitebox: Initial implementation

daveshah1:
This is intended to be a start at implementing abc9-style whitebox optimisations inside Yosys itself. It's not necessarily complete or optimal yet, but I am interesting in discussing about the future direction for this. 

At the moment this uses a truth-table type approach to determine whitebox outputs that, based on the connectivity and constant status of inputs, are either stuck high/low or equivalent to a input signal.

Future more powerful optimisations might also be possible either by converting boxes to some kind of AIG representation, and/or by adding whitebox support to the SAT framework

### DIFF
--- a/passes/opt/Makefile.inc
+++ b/passes/opt/Makefile.inc
@@ -18,4 +18,5 @@ OBJS += passes/opt/opt_lut.o
 OBJS += passes/opt/opt_lut_ins.o
 OBJS += passes/opt/pmux2shiftx.o
 OBJS += passes/opt/muxpack.o
+OBJS += passes/opt/opt_whitebox.o
 endif
--- a/passes/opt/opt_whitebox.cc
+++ b/passes/opt/opt_whitebox.cc
@@ -1,0 +1,308 @@
+/*
+ *  yosys -- Yosys Open SYnthesis Suite
+ *
+ *  Copyright (C) 2012  Claire Wolf <claire@symbioticeda.com>
+ *  Copyright (C) 2020  David Shah <dave@ds0.me>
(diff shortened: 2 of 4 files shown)


eddiehung:
Haven't tried it out yet, but looks like a great idea. Since you mention ABC9, then my understanding is that it also performs this optimisation, and so I'd be curious to see if this pass in its current state finds something that ABC9 can't find...

daveshah1:
While I look at supporting more complex cases, would it be best to leave the dervied `proc`'d whiteboxes around after passes so if other passes need the boxes they don't have to rerun derive? I am now looking at whitebox SAT and cleaning up derive boxes at the end of SatGen would be far from ideal as passes like freduce call into SAT many times.

eddiehung:
> While I look at supporting more complex cases, would it be best to leave the dervied `proc`'d whiteboxes around after passes so if other passes need the boxes they don't have to rerun derive? I am now looking at whitebox SAT and cleaning up derived boxes at the end of SatGen would be far from ideal as passes like freduce call into SAT many times.

I personally think it's fine to `proc` whiteboxes and leave them as such as I don't see any value in having a pre-`proc` representation around, and I do exactly that here: https://github.com/YosysHQ/yosys/blob/c9555c9adeba886a308c60615ac794ec20d9276e/passes/techmap/abc9_ops.cc#L314-L315
In any case one can re-derive from AST if they really want to undo `proc`.

gatecat:
Going to close this to help tidy things up; it was an experiment that went nowhere in the end.

PR #1815, opened 2020-03-26
Fix solver output parsing for exists-forall optimization 

boqwxp:
This assumes the only desired output from the solver will be "sat", "unsat", or "unknown".

### DIFF
--- a/backends/smt2/smtio.py
+++ b/backends/smt2/smtio.py
@@ -704,7 +704,9 @@ class SmtIo:
                     if msg is not None:
                         print("%s waiting for solver (%s)" % (self.timestamp(), msg), flush=True)
 
-        result = self.read()
+        result = ""
+        while result not in ["sat", "unsat", "unknown"]:
+            result = self.read()
 
         if self.debug_file:
             print("(set-info :status %s)" % result, file=self.debug_file)
--- a/backends/smt2/smtio.py
+++ b/backends/smt2/smtio.py
@@ -704,8 +704,12 @@ class SmtIo:
                     if msg is not None:
                         print("%s waiting for solver (%s)" % (self.timestamp(), msg), flush=True)
 
-        result = ""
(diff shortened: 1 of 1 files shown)


backends/smt2/smtio.py
```diff
@@ -704,7 +704,9 @@ def check_sat(self):
                     if msg is not None:
                         print("%s waiting for solver (%s)" % (self.timestamp(), msg), flush=True)
 
-        result = self.read()
+        result = ""
+        while result not in ["sat", "unsat", "unknown"]:
+            result = self.read()
```
cliffordwolf: please specify the patterns you want to skip, not the patterns where you want to stop skipping. (or, if that is really not an option, only skip stuff in exists-forall mode)

boqwxp: You're right, I was just being lazy.  But you spurred me to investigate again exactly what the skipped lines were.  Apparently there are no longer (with z3 master HEAD) lines with `|->`; maybe there used to be an extraneous `(get-model)` in one of my files.  In any case, the only skipped line now is:
`WARNING: optimization with quantified constraints is not supported`

I am not super convinced this is true, but I have some other experiments where the final value with optimized quantifiers in z3 does not match up with the value obtained with an iterated solving / binary search with yices, so it's probably true and just coincidence that simple examples like this can be optimized.

In any case, I have updated this PR to leave the non-exists-forall code path unchanged.  In exists-forall mode, unexpected lines are still skipped, but they are logged.  A separate PR will introduce the `qbfsat` command, where it will automatically `log_warn()` lines from `yosys-smtbmc` that contain a warning from z3.

Finally, it still makes sense to keep the existing yosys code for optimizing exists-forall problems.  Even if it's not fully supported by z3, I intend to write an additional iterative solving / binary search optimizing mode for the `qbfsat` command.

PR #1816, opened 2020-03-26
Add support for QuickLogic devices

kgugala:
This PR adds support for the following QuickLogic devices:

* EOS-S3 (https://www.quicklogic.com/products/eos-s3/)
* PolarPro 3E (https://www.quicklogic.com/products/fpga/fpgas-sram/#polarpro3e)

CC @mithro

### DIFF
--- a/techlibs/quicklogic/cells_map.v
+++ b/techlibs/quicklogic/cells_map.v
@@ -1,63 +1,141 @@
-module \$_MUX_ (
-   input A,
-   input B,
-   input S,
-   output Y
-);
-    QLOGIC_MUX _TECHMAP_REPLACE_ (.A(A), .B(B), .S(S), .Y(Y));
+module \$lut (A, Y);
+    parameter WIDTH = 0;
+    parameter LUT = 0;
+
+    input [WIDTH-1:0] A;
+    output Y;
+
+    generate
+        if (WIDTH == 1)
+        begin
+            LUT1 #(.INIT(LUT)) _TECHMAP_REPLACE_ (.O(Y), .I0(A[0]));
+        end
+        else if (WIDTH == 2)
+        begin
+            LUT2 #(.INIT(LUT)) _TECHMAP_REPLACE_ (.O(Y), .I0(A[0]), .I1(A[1]));
+        end
+        else if (WIDTH == 3)
+        begin
+            LUT3 #(.INIT(LUT)) _TECHMAP_REPLACE_ (.O(Y), .I0(A[0]), .I1(A[1]), .I2(A[2]));
+        end
+        else if (WIDTH == 4)
+        begin
+            LUT4 #(.INIT(LUT)) _TECHMAP_REPLACE_ (.O(Y), .I0(A[0]), .I1(A[1]), .I2(A[2]), .I3(A[3]));
+        end
+        else
+        begin
+            wire _TECHMAP_FAIL_ = 1;
+        end
+    endgenerate
 endmodule
 
-module \$_DFF_N_ (
-   input D,
-   input C,
-   output Q
-);
-    wire nC = ~ C;
-    QLOGIC_DFFE _TECHMAP_REPLACE_ (.S(1'b0), .D(D), .E(1'b1), .QCK(nC), .R(1'b0), .QZ(Q));
+module \$_MUX_ (A, B, S, Y);
+
+    parameter WIDTH = 0;
+
+    input [WIDTH-1:0] A, B;
+    input S;
+    output reg [WIDTH-1:0] Y;
+
+    generate
+        if (WIDTH == 1)
+        begin
+            mux2x0 _TECHMAP_REPLACE_ (.Q(Y[0]), .S(S), .A(A[0]), .B(B[0]));
+        end
+        else
+        begin
+            wire _TECHMAP_FAIL_ = 1;
+        end
+    endgenerate
 endmodule
 
-module \$_DFF_P_ (
-   input D,
-   input C,
-   output Q
-);
-    QLOGIC_DFFE _TECHMAP_REPLACE_ (.S(1'b0), .D(D), .E(1'b1), .QCK(C), .R(1'b0), .QZ(Q));
+module \$_NOT_ (A, Y);
+    input A;
+    output Y;
+    inv _TECHMAP_REPLACE_ (.Q(Y), .A(A));
 endmodule
 
-module \$_DFF_PN0_ (
-   input D,
-   input C,
-   input R,
-   output Q
-);
-    wire nR = ~ R;
-    QLOGIC_DFFE _TECHMAP_REPLACE_ (.S(1'b0), .D(D), .E(1'b1), .QCK(C), .R(nR), .QZ(Q));
+module \$_AND_ (A, B, Y);
+    input A;
+    input B;
+    output Y;
+    AND2I0 _TECHMAP_REPLACE_ (.Q(Q), .A(A), .B(B));
 endmodule
 
-module \$_DFF_PP0_ (
-   input D,
-   input C,
-   input R,
-   output Q
-);
-    QLOGIC_DFFE _TECHMAP_REPLACE_ (.S(1'b0), .D(D), .E(1'b1), .QCK(C), .R(R), .QZ(Q));
+module \$_DFF_N_ (D, Q, C);
+    input D;
+    input C;
+    output Q;
+    dff _TECHMAP_REPLACE_ (.Q(Q), .D(D), .CLK(~C));
 endmodule
 
-module \$__DFFE_PP0_ (
-   input D,
-   input C,
-   input E,
-   input R,
-   output Q
-);
-    wire nE = ~ E;
-    QLOGIC_DFFE _TECHMAP_REPLACE_ (.S(1'b0), .D(D), .E(nE), .QCK(C), .R(R), .QZ(Q));
+module \$_DFF_P_ (D, Q, C);
+    input D;
+    input C;
+    output Q;
+    dff _TECHMAP_REPLACE_ (.Q(Q), .D(D), .CLK(C));
 endmodule
 
-module \$_NOT_ (
-   input A,
-   output Y
-);
-    QLOGIC_NOT _TECHMAP_REPLACE_ (.A(A), .Y(Y));
+module \$_DFF_NN0_ (D, Q, C, R);
+    input D;
+    input C;
+    input R;
+    output Q;
+    dffc _TECHMAP_REPLACE_ (.Q(Q), .D(D), .CLK(~C), .CLR(~R));
 endmodule
 
+module \$_DFF_NN1_ (D, Q, C, R);
+    input D;
+    input C;
+    input R;
+    output Q;
+    dffp _TECHMAP_REPLACE_ (.Q(Q), .D(D), .CLK(~C), .PRE(~R));
+endmodule
+
+module \$_DFF_NP0_ (D, Q, C, R);
+    input D;
+    input C;
+    input R;
+    output Q;
+    dffc _TECHMAP_REPLACE_ (.Q(Q), .D(D), .CLK(~C), .CLR(R));
+endmodule
+
+module \$_DFF_NP1_ (D, Q, C, R);
+    input D;
+    input C;
+    input R;
+    output Q;
+    dffp _TECHMAP_REPLACE_ (.Q(Q), .D(D), .CLK(~C), .PRE(R));
+endmodule
+
+module \$_DFF_PN0_ (D, Q, C, R);
+    input D;
+    input C;
+    input R;
+    output Q;
+    dffc _TECHMAP_REPLACE_ (.Q(Q), .D(D), .CLK(C), .CLR(~R));
+endmodule
+
+module \$_DFF_PN1_ (D, Q, C, R);
+    input D;
+    input C;
+    input R;
+    output Q;
+    dffp _TECHMAP_REPLACE_ (.Q(Q), .D(D), .CLK(C), .PRE(~R));
+endmodule
+
+module \$_DFF_PP0_ (D, Q, C, R);
+    input D;
+    input C;
+    input R;
+    output Q;
+    dffc _TECHMAP_REPLACE_ (.Q(Q), .D(D), .CLK(C), .CLR(R));
+endmodule
+
+module \$_DFF_PP1_ (D, Q, C, R);
+    input D;
+    input C;
+    input R;
+    output Q;
+    dffp _TECHMAP_REPLACE_ (.Q(Q), .D(D), .CLK(C), .PRE(R));
+endmodule
--- a/techlibs/quicklogic/cells_map.v
+++ b/techlibs/quicklogic/cells_map.v
@@ -66,7 +66,9 @@ module \$_DFF_N_ (D, Q, C);
     input D;
     input C;
     output Q;
-    dff _TECHMAP_REPLACE_ (.Q(Q), .D(D), .CLK(~C));
+    wire C_INV;
+    inv clkinv (.Q(C_INV), .A(C));
+    dff tmpdff (.Q(Q), .D(D), .CLK(C_INV));
 endmodule
 
 module \$_DFF_P_ (D, Q, C);
@@ -81,7 +83,11 @@ module \$_DFF_NN0_ (D, Q, C, R);
     input C;
     input R;
     output Q;
-    dffc _TECHMAP_REPLACE_ (.Q(Q), .D(D), .CLK(~C), .CLR(~R));
+    wire C_INV;
+    inv clkinv (.Q(C_INV), .A(C));
+    wire R_INV;
+    inv clrinv (.Q(R_INV), .A(R));
+    dffc tmpdffc (.Q(Q), .D(D), .CLK(C_INV), .CLR(R_INV));
 endmodule
 
 module \$_DFF_NN1_ (D, Q, C, R);
@@ -89,7 +95,11 @@ module \$_DFF_NN1_ (D, Q, C, R);
     input C;
     input R;
     output Q;
-    dffp _TECHMAP_REPLACE_ (.Q(Q), .D(D), .CLK(~C), .PRE(~R));
+    wire C_INV;
+    inv clkinv (.Q(C_INV), .A(C));
+    wire R_INV;
+    inv preinv (.Q(R_INV), .A(R));
+    dffp tmpdffp (.Q(Q), .D(D), .CLK(C_INV), .PRE(R_INV));
 endmodule
 
 module \$_DFF_NP0_ (D, Q, C, R);
@@ -97,7 +107,9 @@ module \$_DFF_NP0_ (D, Q, C, R);
     input C;
     input R;
     output Q;
-    dffc _TECHMAP_REPLACE_ (.Q(Q), .D(D), .CLK(~C), .CLR(R));
+    wire C_INV;
+    inv clkinv (.Q(C_INV), .A(C));
+    dffc tmpdffc (.Q(Q), .D(D), .CLK(C_INV), .CLR(R));
 endmodule
 
 module \$_DFF_NP1_ (D, Q, C, R);
@@ -105,7 +117,9 @@ module \$_DFF_NP1_ (D, Q, C, R);
     input C;
     input R;
     output Q;
-    dffp _TECHMAP_REPLACE_ (.Q(Q), .D(D), .CLK(~C), .PRE(R));
+    wire C_INV;
+    inv clkinv (.Q(C_INV), .A(C));
+    dffp tmpdffp (.Q(Q), .D(D), .CLK(C_INV), .PRE(R));
 endmodule
 
 module \$_DFF_PN0_ (D, Q, C, R);
@@ -113,7 +127,9 @@ module \$_DFF_PN0_ (D, Q, C, R);
     input C;
     input R;
     output Q;
-    dffc _TECHMAP_REPLACE_ (.Q(Q), .D(D), .CLK(C), .CLR(~R));
+    wire R_INV;
+    inv preinv (.Q(R_INV), .A(R));
+    dffc tmpdffc (.Q(Q), .D(D), .CLK(C), .CLR(R_INV));
 endmodule
 
 module \$_DFF_PN1_ (D, Q, C, R);
@@ -121,7 +137,9 @@ module \$_DFF_PN1_ (D, Q, C, R);
     input C;
     input R;
     output Q;
-    dffp _TECHMAP_REPLACE_ (.Q(Q), .D(D), .CLK(C), .PRE(~R));
+    wire R_INV;
+    inv preinv (.Q(R_INV), .A(R));
+    dffp tmpdffp (.Q(Q), .D(D), .CLK(C), .PRE(R_INV));
 endmodule
 
 module \$_DFF_PP0_ (D, Q, C, R);
--- a/techlibs/quicklogic/cells_map.v
+++ b/techlibs/quicklogic/cells_map.v
@@ -194,3 +194,20 @@ module \$_DFFSR_PPP_ (D, Q, C, R, S);
     output Q;
     dffpc tmpdffpc (.Q(Q), .D(D), .CLK(C), .CLR(R), .PRE(S));
 endmodule
+
+module qlal4s3_mult_16x16_cell (
+    input [15:0] Amult,
+    input [15:0] Bmult,
+    input [1:0] Valid_mult,
+    input [31:0] Cmult);
+
+    wire [63:0] Cmult_w;
+    wire [31:0] Amult_w;
+    wire [31:0] Bmult_w;
+
+    assign Amult_w = {{16{Amult[15]}}, Amult};
+    assign Bmult_w = {{16{Bmult[15]}}, Bmult};
+    assign Cmult = Cmult_w[31:0];
+
+    qlal4s3_mult_32x32_cell I1 ( .Amult(Amult_w), .Bmult(Bmult_w), .Cmult(Cmult_w));
+endmodule
--- a/techlibs/quicklogic/cells_map.v
+++ b/techlibs/quicklogic/cells_map.v
@@ -1,0 +1,63 @@
+module \$_MUX_ (
+   input A,
+   input B,
+   input S,
+   output Y
+);
+    QLOGIC_MUX _TECHMAP_REPLACE_ (.A(A), .B(B), .S(S), .Y(Y));
+endmodule
+
+module \$_DFF_N_ (
+   input D,
+   input C,
+   output Q
+);
+    wire nC = ~ C;
+    QLOGIC_DFFE _TECHMAP_REPLACE_ (.S(1'b0), .D(D), .E(1'b1), .QCK(nC), .R(1'b0), .QZ(Q));
+endmodule
+
+module \$_DFF_P_ (
+   input D,
+   input C,
+   output Q
+);
+    QLOGIC_DFFE _TECHMAP_REPLACE_ (.S(1'b0), .D(D), .E(1'b1), .QCK(C), .R(1'b0), .QZ(Q));
+endmodule
+
+module \$_DFF_PN0_ (
+   input D,
+   input C,
+   input R,
+   output Q
+);
+    wire nR = ~ R;
+    QLOGIC_DFFE _TECHMAP_REPLACE_ (.S(1'b0), .D(D), .E(1'b1), .QCK(C), .R(nR), .QZ(Q));
+endmodule
+
+module \$_DFF_PP0_ (
+   input D,
+   input C,
+   input R,
+   output Q
+);
+    QLOGIC_DFFE _TECHMAP_REPLACE_ (.S(1'b0), .D(D), .E(1'b1), .QCK(C), .R(R), .QZ(Q));
+endmodule
+
+module \$__DFFE_PP0_ (
+   input D,
+   input C,
+   input E,
+   input R,
+   output Q
+);
+    wire nE = ~ E;
+    QLOGIC_DFFE _TECHMAP_REPLACE_ (.S(1'b0), .D(D), .E(nE), .QCK(C), .R(R), .QZ(Q));
+endmodule
+
+module \$_NOT_ (
+   input A,
+   output Y
+);
+    QLOGIC_NOT _TECHMAP_REPLACE_ (.A(A), .Y(Y));
+endmodule
+
--- a/techlibs/quicklogic/cells_map.v
+++ b/techlibs/quicklogic/cells_map.v
@@ -5,22 +5,39 @@ module \$lut (A, Y);
     input [WIDTH-1:0] A;
     output Y;
 
+    function [WIDTH ** 2 - 1:0] INIT_address_inverse;
+        input [WIDTH ** 2 - 1: 0] arr;
+        integer i;
+        integer n;
+        reg [WIDTH - 1:0] tmp;
+        reg [WIDTH - 1:0] tmp2;
+        tmp = 0;
+        tmp2 = 0;
+        for (i = 0; i < WIDTH ** 2; i++) begin
+            INIT_address_inverse[tmp2] = arr[tmp];
+            tmp = tmp + 1;
+            for (n = 0; n < WIDTH; n++) begin
+                tmp2[WIDTH - 1 - n] = tmp[n];
+            end
+        end
+    endfunction
+
     generate
         if (WIDTH == 1)
         begin
-            LUT1 #(.INIT(LUT)) _TECHMAP_REPLACE_ (.O(Y), .I0(A[0]));
+            LUT1 #(.INIT(INIT_address_inverse(LUT))) _TECHMAP_REPLACE_ (.O(Y), .I0(A[0]));
         end
         else if (WIDTH == 2)
         begin
-            LUT2 #(.INIT(LUT)) _TECHMAP_REPLACE_ (.O(Y), .I0(A[0]), .I1(A[1]));
+            LUT2 #(.INIT(INIT_address_inverse(LUT))) _TECHMAP_REPLACE_ (.O(Y), .I0(A[0]), .I1(A[1]));
         end
         else if (WIDTH == 3)
         begin
-            LUT3 #(.INIT(LUT)) _TECHMAP_REPLACE_ (.O(Y), .I0(A[0]), .I1(A[1]), .I2(A[2]));
+            LUT3 #(.INIT(INIT_address_inverse(LUT))) _TECHMAP_REPLACE_ (.O(Y), .I0(A[0]), .I1(A[1]), .I2(A[2]));
         end
         else if (WIDTH == 4)
         begin
-            LUT4 #(.INIT(LUT)) _TECHMAP_REPLACE_ (.O(Y), .I0(A[0]), .I1(A[1]), .I2(A[2]), .I3(A[3]));
+            LUT4 #(.INIT(INIT_address_inverse(LUT))) _TECHMAP_REPLACE_ (.O(Y), .I0(A[0]), .I1(A[1]), .I2(A[2]), .I3(A[3]));
         end
         else
         begin
--- a/techlibs/quicklogic/cells_map.v
+++ b/techlibs/quicklogic/cells_map.v
@@ -29,15 +29,15 @@ module \$lut (A, Y);
         end
         else if (WIDTH == 2)
         begin
-            LUT2 #(.INIT(INIT_address_inverse(LUT))) _TECHMAP_REPLACE_ (.O(Y), .I0(A[0]), .I1(A[1]));
+            LUT2 #(.INIT(INIT_address_inverse(LUT))) _TECHMAP_REPLACE_ (.O(Y), .I0(A[1]), .I1(A[0]));
         end
         else if (WIDTH == 3)
         begin
-            LUT3 #(.INIT(INIT_address_inverse(LUT))) _TECHMAP_REPLACE_ (.O(Y), .I0(A[0]), .I1(A[1]), .I2(A[2]));
+            LUT3 #(.INIT(INIT_address_inverse(LUT))) _TECHMAP_REPLACE_ (.O(Y), .I0(A[2]), .I1(A[1]), .I2(A[0]));
         end
         else if (WIDTH == 4)
         begin
-            LUT4 #(.INIT(INIT_address_inverse(LUT))) _TECHMAP_REPLACE_ (.O(Y), .I0(A[0]), .I1(A[1]), .I2(A[2]), .I3(A[3]));
+            LUT4 #(.INIT(INIT_address_inverse(LUT))) _TECHMAP_REPLACE_ (.O(Y), .I0(A[3]), .I1(A[2]), .I2(A[1]), .I3(A[0]));
         end
         else
         begin
@@ -76,7 +76,7 @@ module \$_AND_ (A, B, Y);
     input A;
     input B;
     output Y;
-    AND2I0 _TECHMAP_REPLACE_ (.Q(Q), .A(A), .B(B));
+    AND2I0 _TECHMAP_REPLACE_ (.Q(Y), .A(A), .B(B));
 endmodule
 
 module \$_DFF_N_ (D, Q, C);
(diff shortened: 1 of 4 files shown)


ZirconiumX:
By tradition, scripts are named by their manufacturer, so this would be `synth_antmicro` not `synth_quicklogic`.

techlibs/quicklogic/synth_quicklogic.cc
```diff
@@ -0,0 +1,172 @@
...
+            run("opt_clean");
+            run("proc");
+            run("techmap -map +/quicklogic/cells_map.v");
+            run("clean");
+            run("opt");
+            run("check");
+            run("peepopt");
+            run("opt_clean -purge");
+            run("clean -purge");
+            /* avoid sefault when this code is run without active design (e.g when calling help) */
```
ZirconiumX: We have the `help_mode` boolean for this.

kgugala:
> By tradition, scripts are named by their manufacturer, so this would be `synth_antmicro` not `synth_quicklogic`.

those scripts control synthesis process for QuickLogic's chips, not Antmicro's. It's QuickLogic who manufactures the chips.

ZirconiumX:
This also needs tests.

ZirconiumX:
Adding black-boxes for multiplier cells is not enough for Yosys to use them; look at `mul2dsp.v` for a helper script to map to your cells.

kgugala:
@ZirconiumX thanks for the review, we'll add the tests and take a look on the multiplier cells.

BTW can you give me a hint how to debug the problem causing CI to fail? It fails only in GCC 4.8 run. I wasn't able to reproduce it locally. I used Ubuntu 16.04 docker image with GCC 4.8 in local tests. Any help here is appreciated.

ZirconiumX:
I'm reasonably sure that whatever it is, it's not related to this PR.

kgugala:
> I'm reasonably sure that whatever it is, it's not related to this PR.

indeed, doesn't seem to be related

techlibs/quicklogic/cells_map.v
```diff
@@ -0,0 +1,315 @@
...
+        else
+        begin
+            wire _TECHMAP_FAIL_ = 1;
+        end
+    endgenerate
+endmodule
+
+module \$_MUX_ (A, B, S, Y);
+
+    parameter WIDTH = 0;
```
daveshah1: $_MUX_ cells (and all other uppercase $_*_ cells) are gate-level and therefore don't have a width parameter.

techlibs/quicklogic/synth_quicklogic.cc
```diff
@@ -0,0 +1,172 @@
...
+            if (!this->active_design)
+                this->currmodule = "top";
+            else
+                this->currmodule = this->active_design->top_module()->name.str();
+            if (this->currmodule.size() > 0)
+                run(stringf("select -module %s", this->currmodule.c_str()));
+            run("select -set clock_inputs */t:dff* %x:+[CLK,CLR,PRE] */t:dff* %d");
```
daveshah1: You probably want to be using clkbufmap here

techlibs/quicklogic/cells_map.v
```diff
@@ -0,0 +1,315 @@
...
+    endgenerate
+endmodule
+
+module \$_NOT_ (A, Y);
+    input A;
+    output Y;
+    inv _TECHMAP_REPLACE_ (.Q(Y), .A(A));
+endmodule
+
+module \$_AND_ (A, B, Y);
```
daveshah1: There will be no AND gates left after LUT mapping so this rule shouldn't do anything (and trying to map them before LUT mapping is a very bad idea as it will prevent many optimisations from working).

techlibs/quicklogic/cells_sim.v
```diff
@@ -0,0 +1,343 @@
...
+endmodule
+
+// O:  CZ
+// I0: TA1 TA2 TB1 TB2 BA1 BA2 BB1 BB2
+// I1: TSL BSL
+// I2: TAB BAB
+// I3: TBS
+module LUT4(output O, input I0, I1, I2, I3);
+    parameter [15:0] INIT = 0;
+    assign O = INIT[{I3, I2, I1, I0}];
```
daveshah1: If you care about it, beware that this LUT model is overly pessimistic in terms for X-propagation, see the cascades of ?: in https://github.com/YosysHQ/yosys/blob/master/techlibs/ecp5/cells_sim.v#L6 for a more accurate model

techlibs/quicklogic/cells_sim.v
```diff
@@ -0,0 +1,343 @@
...
+        else if (EN)
+            Q <= D;
+    end
+endmodule
+
+//                 QZ        QDI    QCK  QRT
+module dffsc(output reg Q, input D, CLK, CLR);
+    parameter [0:0] INIT = 1'b0;
+    initial Q = INIT;
+    always @(posedge CLK, posedge CLR) begin
```
daveshah1: If the "s" stands for synchronous, this is not a synchronous clear

techlibs/quicklogic/cells_sim.v
```diff
@@ -0,0 +1,343 @@
...
+
+endmodule
+
+(* blackbox *)
+module ram8k_2x1_cell_macro (
+    input [10:0] A1_0,
+    input [10:0] A1_1,
+    input [10:0] A2_0,
+    input [10:0] A2_1,
+    input CLK1_0 /* synthesis syn_isclock=1 */,
```
daveshah1: "synthesis" hot comments in place of attributes are not supported by Yosys. Verilog has only had attributes for 19 years now, please use them.

techlibs/quicklogic/synth_quicklogic.cc
```diff
@@ -0,0 +1,172 @@
...
+
+        if (check_label("prepare")) {
+            run("proc");
+            if (flatten || help_mode)
+                run("flatten", "(with '-flatten')");
+            run("tribuf -logic");
+            run("opt_expr");
+            run("proc");
+            run("opt_clean");
+            run("proc");
```
daveshah1: Why are you calling proc so many times?

techlibs/quicklogic/synth_quicklogic.cc
```diff
@@ -0,0 +1,172 @@
...
+            run("techmap");
+            run("abc -lut 1:4");
+            run("opt_clean");
+            run("proc");
+            run("techmap -map +/quicklogic/cells_map.v");
+            run("clean");
+            run("opt");
+            run("check");
+            run("peepopt");
```
daveshah1: peepopt is focused on word level transformation, its not going to do anything well after techmap and abc

techlibs/quicklogic/synth_quicklogic.cc
```diff
@@ -0,0 +1,172 @@
...
+            run("opt_clean -purge");
+            run("clean -purge");
```
daveshah1: Why do you need both opt_clean -purge and clean -purge, they should behave the same just with different verbosity.

techlibs/quicklogic/synth_quicklogic.cc
```diff
@@ -0,0 +1,172 @@
...
+    void help() YS_OVERRIDE
+    {
+        log("\n");
+        log("   synth_quicklogic [options]\n");
+        log("This command runs synthesis for QuickLogic FPGAs\n");
+        log("\n");
+        log("    -top <module>\n");
+        log("         use the specified module as top module\n");
+        log("\n");
+        log("    -exe <command>\n");
```
daveshah1: Seems like redundant copypasta from the abc command? Doesn't seem like -exe does anything

techlibs/quicklogic/cells_sim.v
```diff
@@ -194,77 +194,151 @@ module logic_1(output a);
...
+assign FZ = FS ? F2 : F1;
+assign TZ = TZI;
+assign CZ = CZI;
+assign QCKP = QCKS ? QCK : ~QCK;
+
+always @(posedge QCKP)
+	if(~QRT && ~QST)
+		if(QEN)
+		QZ = QZI;
+always @(QRT or QST)
```
daveshah1, line 263: `always @*` would probably be cleaner here?

techlibs/quicklogic/cells_sim.v
```diff
@@ -194,77 +194,151 @@ module logic_1(output a);
...
+assign QZI = QDS ? QDI : CZI ;
+assign FZ = FS ? F2 : F1;
+assign TZ = TZI;
+assign CZ = CZI;
+assign QCKP = QCKS ? QCK : ~QCK;
+
+always @(posedge QCKP)
+	if(~QRT && ~QST)
+		if(QEN)
+		QZ = QZI;
```
daveshah1, line 262: This should be `QZ <= QZI;` as it appears to be modelling a flop

techlibs/quicklogic/cells_sim.v
mkurc-ant: I've re-written the model. Now it uses only non-blocking assignments.

techlibs/quicklogic/cells_sim.v
mkurc-ant: Fixed, see the comment above.

techlibs/quicklogic/cells_sim.v
mkurc-ant: Even if its not the model correspond to a FF type accepted by Quicklogic P&R toolchain. Yosys doesn't emit it but it has to be here so it can be used explicitly in a design.

techlibs/quicklogic/cells_map.v
mkurc-ant: I removed the techmap since is shouldn't be needed as there will be no `$_MUX_` left after LUT mapping.

techlibs/quicklogic/cells_map.v
mkurc-ant: Fixed (removed techmap)

techlibs/quicklogic/cells_sim.v
mkurc-ant: I'd leave it as it is for now.

mkurc-ant:
@daveshah1 I think we've addressed all the issues. Could you re-review?

techlibs/quicklogic/synth_quicklogic.cc
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    void script() YS_OVERRIDE
+    {
+        if (check_label("begin")) {
+            run("read_verilog -lib +/quicklogic/cells_sim.v");
+            run(stringf("hierarchy -check %s", top_opt.c_str()));
+        }
+
+        if (check_label("prepare")) {
```
daveshah1, line 97: All the other Yosys synth passes are split up in to labels, rather than putting everything in "prepare"

techlibs/quicklogic/cells_sim.v
```diff
@@ -0,0 +1,474 @@
...
+
+endmodule /* ram8k_2x1_cell_macro */
+
+module gpio_cell_macro (
+    input DS, ESEL, FIXHOLD, IE, INEN, IQC, IQCS, IQE, IQR, OQE, OQI, OSEL, WPD,
+    inout IP,
+    output IZ);
+
+    assign IZ = INEN ? IP : 1'b0;
+    assign IP = IEN ? OQI : 1'bz;
```
daveshah1, line 425: This is causing a CI failure, as IEN is not declared anywhere

techlibs/quicklogic/cells_sim.v
```diff
@@ -0,0 +1,474 @@
...
+    (* clkbuf_sink *)
+    input CLK
+);
+    parameter [0:0] INIT = 1'b0;
+    initial Q = INIT;
+    always @(posedge CLK)
+        Q <= D;
+endmodule
+
+module dffc(
```
daveshah1, line 53: If you want to use the dffc/dffp models for verification with Yosys, or probably Verilator too, you need to combine the two always blocks driving Q into one.

techlibs/quicklogic/cells_sim.v
mkurc-ant: Fixed.

techlibs/quicklogic/cells_sim.v
mkurc-ant: I've merged those always blocks. For now the verification whether these cells are inferred, as it is done in the tests, is enough.

techlibs/quicklogic/synth_quicklogic.cc
mkurc-ant: Ok, I've split the flow into labels.

mkurc-ant:
@daveshah1 I've addressed your comments and pushed fixes.

techlibs/quicklogic/synth_quicklogic.cc
```diff
@@ -0,0 +1,162 @@
...
+            run("proc");
+            if (flatten || help_mode)
+                run("flatten", "(with '-flatten')");
+        }
+
+        if (check_label("corase")) {
```
nakengelhardt, line 103: That's a typo for "coarse" I assume?

mkurc-ant: Yep, fixed that.

nakengelhardt:
Sorry, I'm not the right person to review this. (Unless you wanted me to confirm that you've spelled "coarse" correctly this time?)

Do you have a SEDA support contract? If so, get in touch with your rep and we'll assign someone to work with you asap.

mgielda:
Hi @daveshah1, could you re-review this then?

techlibs/quicklogic/synth_quicklogic.cc
```diff
@@ -0,0 +1,162 @@
...
+            run("clean");
+            run("check");
+            run("opt_clean -purge");
+        }
+
+        if (check_label("iomap")) {
+            if (help_mode) {
+                run("select -module top");
+            } else {
+                this->currmodule = this->active_design->top_module()->name.str();
```
daveshah1, line 127: I don't think all this selection stuff is needed, you can just do `iopadmap ... A:top` to select the top module (which has the `top` attribute set), like here: https://github.com/YosysHQ/yosys/blob/master/techlibs/xilinx/synth_xilinx.cc#L590

techlibs/quicklogic/synth_quicklogic.cc
```diff
@@ -0,0 +1,162 @@
...
+            run("techmap");
+            run("abc -lut 1:4");
+            run("opt_clean");
+        }
+
+        if (check_label("map")) {
+            run("techmap -map +/quicklogic/cells_map.v");
+            run("clean");
+            run("check");
+            run("opt_clean -purge");
```
daveshah1, line 120: Are you sure that the `-purge` is needed here? This will throw away all alias names for wires but one, which might make clock constraints harder as you will have to guess the name Yosys chooses.

techlibs/quicklogic/synth_quicklogic.cc
```diff
@@ -0,0 +1,162 @@
...
+        }
+
+        if (check_label("prepare")) {
+            run("proc");
+            if (flatten || help_mode)
+                run("flatten", "(with '-flatten')");
+        }
+
+        if (check_label("coarse")) {
+            run("tribuf -logic");
```
daveshah1, line 104: Usually there would be a `deminout` after `tribuf`, to simplify unnecessary `inout`s.

techlibs/quicklogic/synth_quicklogic.cc
```diff
@@ -0,0 +1,162 @@
...
+                run("select -clear");
+            } else if (this->currmodule.size() > 0)
+                run("select -clear");
+        }
+
+        if (check_label("finalize")) {
+            run("splitnets -ports -format ()");
+            run("hilomap -hicell logic_1 a -locell logic_0 a -singleton");
+            run("techmap -map +/quicklogic/cells_map.v");
+            run("setundef -zero -params -undriven");
```
daveshah1, line 143: `setundef` should probably go before `hilomap`, otherwise any zero drivers it creates (for undriven signals) won't be mapped to a constant driver cell.

techlibs/quicklogic/synth_quicklogic.cc
```diff
@@ -0,0 +1,162 @@
...
+            run("opt_clean");
+            run("clean -purge");
+            run("opt");
```
daveshah1, line 146: I doubt that `opt` here will do anything more than `opt_clean` or `clean` would

techlibs/quicklogic/synth_quicklogic.cc
```diff
@@ -0,0 +1,162 @@
...
+            } else if (this->currmodule.size() > 0)
+                run("select -clear");
+        }
+
+        if (check_label("finalize")) {
+            run("splitnets -ports -format ()");
+            run("hilomap -hicell logic_1 a -locell logic_0 a -singleton");
+            run("techmap -map +/quicklogic/cells_map.v");
+            run("setundef -zero -params -undriven");
+            run("opt_clean");
```
daveshah1, line 144: `opt_clean` is redundant before `clean -purge` (which does everything it does and more), as above are you sure the `-purge` is needed?

techlibs/quicklogic/synth_quicklogic.cc
mkurc-ant: Allright, fixed that.

techlibs/quicklogic/synth_quicklogic.cc
mkurc-ant: It turned out that all `-purge` were redundant / not necessary. Removed.

techlibs/quicklogic/synth_quicklogic.cc
mkurc-ant: Fixed that, and added a test

techlibs/quicklogic/synth_quicklogic.cc
mkurc-ant: That makes sense, added the command after `tribuf`

techlibs/quicklogic/synth_quicklogic.cc
mkurc-ant: Fixed

techlibs/quicklogic/synth_quicklogic.cc
mkurc-ant: Fixed

daveshah1:
Looks like all the issues in my most recent review are now resolved.

mkurc-ant:
@daveshah1 Right, Waiting for the CI now

techlibs/quicklogic/synth_quicklogic.cc
```diff
@@ -0,0 +1,150 @@
...
+            run("tribuf -logic");
+            run("deminout");
+            run("opt");
+            run("opt_clean");
+            run("check");
+            run("peepopt");
+            run("opt_clean");
+            run("check");
+            run("techmap");
+            run("abc -lut 1:4");
```
ZirconiumX, line 113: This looks *very* wrong. From the `abc` help:
```
    -lut <w1>:<w2>
        generate netlist using luts of (max) the specified width <w2>. All
        luts with width <= <w1> have constant cost. for luts larger than <w1>
        the area cost doubles with each additional input bit. the delay cost
        is still constant for all lut widths.
```

What this means is that a LUT2 costs twice as much area as a LUT1, a LUT3 four times as much area as a LUT1 and a LUT4 eight times as much area as a LUT1. This doesn't match the hardware at all: reading the EOS S3 datasheet, a Logic Cell can implement 2 LUT3s, a LUT4, or use the hard multiplexers to mux up to larger multiplexers. That means a LUT1, a LUT2 and LUT3 all have the same area cost.

You could consider either `abc -lut 3:4` to aim to reduce area by producing LUT3s that might be packed together into an LC; `abc -lut 4` to produce the fastest logic that fits in single LCs, or `abc -lut 4:6` to consider packing together multiple LCs to reduce logic depth. But `abc -lut 1:4` is surely wrong.

techlibs/quicklogic/synth_quicklogic.cc
```diff
@@ -0,0 +1,150 @@
...
+            if (flatten || help_mode)
+                run("flatten", "(with '-flatten')");
+        }
+
+        if (check_label("coarse")) {
+            run("tribuf -logic");
+            run("deminout");
+            run("opt");
+            run("opt_clean");
+            run("check");
```
ZirconiumX, line 108: There are a lot of these `check` calls that are redundant; there's little point obsessively calling `check` after a few passes.

techlibs/quicklogic/synth_quicklogic.cc
```diff
@@ -0,0 +1,150 @@
...
+        log("\n");
+        log("    -edif <file>\n");
+        log("        write the design to the specified edif file. writing of an output file\n");
+        log("        is omitted if this parameter is not specified.\n");
+        log("\n");
+        log("    -blif <file>\n");
+        log("        write the design to the specified BLIF file. writing of an output file\n");
+        log("        is omitted if this parameter is not specified.\n");
+        log("\n");
+        log("    -flatten\n");
```
ZirconiumX, line 30: General Yosys precedent from iCE40, ECP5, Gowin and Intel (Xilinx is an exception) is that you should flatten by default.

techlibs/quicklogic/synth_quicklogic.cc
ZirconiumX: You should also use `-dress` to attempt to preserve wire names in the netlist.

techlibs/quicklogic/synth_quicklogic.cc
```diff
@@ -0,0 +1,150 @@
...
+            run("techmap");
+            run("abc -lut 1:4");
+            run("opt_clean");
+            run("check");
+        }
+
+        if (check_label("map")) {
+            run("techmap -map +/quicklogic/cells_map.v");
+            run("opt_clean");
+            run("check");
```
ZirconiumX, line 121: Consider running `autoname` to discover more wires names to preserve.

smunaut:
Looking fwd to this.
FWIW, I built this branch merged with master yesterday. Seems to work ok.

techlibs/quicklogic/synth_quicklogic.cc
kgugala: fixed - removed the obsessive checks

techlibs/quicklogic/synth_quicklogic.cc
kgugala: I have switched to `abc -luts 1,2,2,4` also added MUX4 and MUX8 inference (this gives additional optimization for bigger desgins)

The problem I see is how to tell yosys that LUT1/MUX2 can be packed together with LUT2/3 pair or a single LUT4. @ZirconiumX any hints?

techlibs/quicklogic/synth_quicklogic.cc
kgugala: done

techlibs/quicklogic/synth_quicklogic.cc
kgugala: added

techlibs/quicklogic/synth_quicklogic.cc
ZirconiumX: I don't think that's presently possible. I think it would be better to let ABC just have the MUX2s, to see if it can perform logic optimisations on it, and then just pack as best as you can.

kgugala: @ZirconiumX can we land this PR and work on further optimizations later?

ZirconiumX:
I don't have merge powers, so it's not up to me whether this PR lands or
not. I'm just giving feedback from experience.

techlibs/quicklogic/cells_sim.v
```diff
@@ -0,0 +1,470 @@
...
+    assign TZ = TZI;
+    assign CZ = CZI;
+    assign QCKP = QCKS ? QCK : ~QCK;
+
+    always @(posedge QCKP or posedge QRT or posedge QST)
+        if (QRT)
+            QZ <= 1'b0;
+        else if (QST)
+            QZ <= 1'b1;
+        else
```
mwkmwkmwk, line 294: `QEN`?

techlibs/quicklogic/cells_map.v
```diff
@@ -0,0 +1,294 @@
+module \$lut (A, Y);
+    parameter WIDTH = 0;
+    parameter LUT = 0;
+
+    input [WIDTH-1:0] A;
+    output Y;
+
+    function [WIDTH ** 2 - 1:0] INIT_address_inverse;
```
mwkmwkmwk, line 8: Why are you doing all this, instead of just giving the inputs to the `LUT*` cells in the same order as in the original `$lut`?

techlibs/quicklogic/synth_quicklogic.cc
```diff
@@ -0,0 +1,141 @@
...
+
+        log_header(design, "Executing SYNTH_QUICKLOGIC pass.\n");
+        log_push();
+
+        run_script(design, run_from, run_to);
+
+        log_pop();
+    }
+
+    void script() YS_OVERRIDE
```
mwkmwkmwk, line 81: this needs to be changed to `override` because of changes on current master

techlibs/quicklogic/synth_quicklogic.cc
```diff
@@ -0,0 +1,141 @@
...
+    {
+        if (check_label("begin")) {
+            run("read_verilog -lib +/quicklogic/cells_sim.v");
+            run(stringf("hierarchy -check %s", (top_opt.empty()) ? "-auto-top" : top_opt.c_str()));
```
mwkmwkmwk, line 85: AFAICS `top_opt` cannot be empty? just `top_opt.c_str()` should be fine

techlibs/quicklogic/cells_sim.v
```diff
@@ -0,0 +1,470 @@
...
+);
+    parameter [0:0] INIT = 1'b0;
+    initial Q = INIT;
+
+    always @(posedge CLK or posedge CLR)
+        if (CLR)
+            Q <= 1'b0;
+        else if (EN)
+            Q <= D;
+endmodule
```
mwkmwkmwk, line 142: shouldn't there also be a `dffep`?

techlibs/quicklogic/cells_sim.v
```diff
@@ -0,0 +1,470 @@
...
+// E BA1
+// F BA2
+// G BB1
+// H BB2
+// Q CZ
+module mux8x0(output Q, input S0, S1, S2, A, B, C, D, E, F, G, H);
+    assign Q = S2 ? (S1 ? (S0 ? H : G) : (S0 ? F : E)) : (S1 ? (S0 ? D : C) : (S0 ? B : A));
+endmodule
+
+module inpad(output Q, input P);
```
mwkmwkmwk, line 201: these could use the `(* iopad_external_pin *)` attribute to avoid pad insertion if already instantiated

techlibs/quicklogic/synth_quicklogic.cc
```diff
@@ -0,0 +1,141 @@
...
+        if (check_label("map")) {
+            run("techmap -map +/quicklogic/cells_map.v");
+            run("opt_clean");
+            run("check");
+            run("autoname");
+        }
+
+        if (check_label("iomap")) {
+            run("clkbufmap -buf $_BUF_ Y:A -inpad ckpad Q:P");
+            run("iopadmap -bits -outpad outpad A:P -inpad inpad Q:P -tinoutpad bipad EN:Q:A:P -ignore ckpad P A:top");
```
mwkmwkmwk, line 117: add `(* iopad_external_pin *)` on `ckpad` in the sim library instead of an `-ignore` here

techlibs/quicklogic/cells_map.v
```diff
@@ -0,0 +1,294 @@
...
+    input S;
+    input R;
+    input D;
+    output Q;
+    wire SEDQ;
+    LUT4 #(.INIT(16'b11111111110010)) sedqlut (.O(SEDQ), .I3(S), .I2(E), .I1(D), .I0(Q));
+    LUT2 #(.INIT(4'b0010)) sedr (.O(Q), .I1(R), .I0(SEDQ));
+endmodule
+
+module qlal4s3_mult_16x16_cell (
```
mwkmwkmwk, line 279: what exactly does this rule do and why is it necessary here? isn't 16x16 a hw-supported cell? why is `Valid_mult` not connected, and what even is that? (the manual is not exactly verbose)

techlibs/quicklogic/cells_sim.v
```diff
@@ -0,0 +1,470 @@
...
+module qlal4s3_mult_cell_macro(
+    input [31:0] Amult,
+    input [31:0] Bmult,
+    input [1:0] Valid_mult,
+    input sel_mul_32x32,
+    output [63:0] Cmult);
+
+endmodule
+
+// module DFFSEC(output Q, input D, CLR, EN, CLK, N_11);
```
mwkmwkmwk, line 452: that looks like something that should be removed

techlibs/quicklogic/cells_sim.v
kgugala: I fixed the model. `QEN` is handled correctly

techlibs/quicklogic/cells_map.v
kgugala: This is how the P&R tool expects it. Both opensource and proprietary versions.

techlibs/quicklogic/synth_quicklogic.cc
kgugala: done

techlibs/quicklogic/synth_quicklogic.cc
kgugala: fixed

techlibs/quicklogic/cells_sim.v
kgugala: This can be added. However, the proprietary P&R tool does not expect those.

techlibs/quicklogic/cells_sim.v
kgugala: done

techlibs/quicklogic/synth_quicklogic.cc
kgugala: done

techlibs/quicklogic/cells_map.v
kgugala: this is a macro used in some of the designs. This will map to half of the 32x32 mult. This map is set only for compatibility.
I fixed the `Valid_mult` connection

techlibs/quicklogic/cells_sim.v
kgugala: done

techlibs/quicklogic/cells_sim.v
```diff
@@ -0,0 +1,4959 @@
...
+    assign BZI = BAB ? BBI : BAI;
+    assign CZI = TBS ? BZI : TZI;
+    assign QZI = QDS ? QDI : CZI ;
+    assign FZ = FS ? F2 : F1;
+    assign TZ = TZI;
+    assign CZ = CZI;
+    assign QCKP = QCKS ? QCK : ~QCK;
+
+
+    always @(posedge QCKP)
```
mwkmwkmwk, line 315: this is non-synthesizable, and thus cannot be used eg. for equivalence checking; please use the standard Verilog single-process pattern for FF with set+reset.

techlibs/quicklogic/cells_sim.v
mwkmwkmwk: ... did you seriously "fix" this by bringing back earlier, more broken version?

techlibs/quicklogic/cells_map.v
mwkmwkmwk: That doesn't explain why you're inverting the LUT input order — you could just keep the order of inputs (`.I0(A[0]), .I1(A[1]), .I2(A[2])`)  and not have to do this ugly transform in the first place.

techlibs/quicklogic/cells_sim.v
mwkmwkmwk: Wait, so the vendor tool has `dffc`, `dffp`, `dffpc`, `dffec`, and `dffepc`, but not `dffep`?  That's great...

techlibs/quicklogic/synth_quicklogic.cc
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    void script() override
+    {
+        if (check_label("begin")) {
+            run("read_verilog -lib +/quicklogic/cells_sim.v");
+            run(stringf("hierarchy -check %s", top_opt.c_str()));
+        }
+
+        if (check_label("prepare")) {
```
mwkmwkmwk, line 88: The flow structure here seems quite weird and mismatched with other synth flows.  While yosys isn't very good at consistency here (something we should fix one day...), there are some major issues with this pass set:

- `memory` is never called, so Verilog/RTLIL memories will just end up at the output stage and break it
- `fsm` is never called
- `cmp2lcu` is not used (though that may not be a problem since you don't seem to have a `$lcu` model?)
- LUTs and FFs are mapped in the same step. While this works, it is suboptimal — for one, it can prevent merging FF mapping detritus (like the reset/set/enable inverters) with surrounding logic.  The standard approach in newer targets is to map FFs in a separate step before abc and LUT mapping.

These three steps (prepare/coarse/map) should be replaced by something resembling `synth_intel_alm`'s flatten/coarse/map_ffram/map_ffs/map_luts steps — it's probably our cleanest pass at the moment (it seems the only thing you will need to add wrt intel_alm is `muxpack` + `muxcover`, the rest is already there).

techlibs/quicklogic/cells_map.v
```diff
@@ -0,0 +1,296 @@
...
+    mux8x0 _TECHMAP_REPLACE_ (.A(A), .B(B), .C(C), .D(D), .E(E), .F(F), .G(G), .H(H), .S0(S), .S1(T), .S2(U), .Q(Y));
+endmodule
+
+module \$_MUX4_ (A, B, C, D, S, T, U, Y);
+    input A, B, C, D, S, T, U;
+    output Y;
+    mux4x0 _TECHMAP_REPLACE_ (.A(A), .B(B), .C(C), .D(D), .S0(S), .S1(T), .Q(Y));
+endmodule
+
+module \$_DFF_N_ (D, Q, C);
```
mwkmwkmwk, line 61: What about quicklogic's FF initialization capabilities?  Can `INIT` take an arbitrary value for all FF types?  If so, let's use `_TECHMAP_WIREINIT_Q_` to handle that (see `techlibs/xilinx/xc7_ff_map.v` for an example)

mwkmwkmwk:
One more general question: where can I find some proper documentation for these chips, and in particular the cell library?  The S3 TRM has a grand total of one page dedicated to the logic cell...

techlibs/quicklogic/cells_sim.v
```diff
@@ -0,0 +1,4959 @@
...
+        .FB_PKfbData_9S     (FB_PKfbData_9S),
+        .FB_PKfbData_6S     (FB_PKfbData_6S),
+        .Sys_PKfb_ClkS      (Sys_PKfb_ClkS),
+        .FB_BusyS           (FB_BusyS),
+        .WB_CLKS            (WB_CLKS)
+		);
+
+endmodule /* qlal4s3b_cell_macro */
+
+(* blackbox *)
```
mwkmwkmwk, line 2760: why is this marked a blackbox when it clearly has a usable implementation?

techlibs/quicklogic/cells_sim.v
```diff
@@ -0,0 +1,4959 @@
...
+                     .WEN1_1({ WEN1_1[1:0] }),
+                     .WIDTH_SELECT1_0({ WIDTH_SELECT1_0[1:0] }),
+                     .WIDTH_SELECT1_1({ WIDTH_SELECT1_1[1:0] }),
+                     .WIDTH_SELECT2_0({ WIDTH_SELECT2_0[1:0] }),
+                     .WIDTH_SELECT2_1({ WIDTH_SELECT2_1[1:0] }) );
+
+endmodule /* ram8k_2x1_cell_macro */
+
+(* blackbox *)
+module gpio_cell_macro (
```
mwkmwkmwk, line 4661: also shouldn't be a blackbox

techlibs/quicklogic/cells_sim.v
```diff
@@ -0,0 +1,4959 @@
...
+		FIXHOLD,
+		IZ,
+		IQZ,
+		IQE,
+		IQC,
+		IQCS,
+		IQR,
+		WPD,
+		INEN,
+		IP
```
mwkmwkmwk, line 4678: `(* iopad_external_pin *)`?

techlibs/quicklogic/synth_quicklogic.cc
```diff
@@ -0,0 +1,141 @@
...
+        if (check_label("map")) {
+            run("techmap -map +/quicklogic/cells_map.v");
+            run("opt_clean");
+            run("check");
+            run("autoname");
+        }
+
+        if (check_label("iomap")) {
+            run("clkbufmap -buf $_BUF_ Y:A -inpad ckpad Q:P");
+            run("iopadmap -bits -outpad outpad A:P -inpad inpad Q:P -tinoutpad bipad EN:Q:A:P A:top");
```
mwkmwkmwk, line 117: this is missing a `-toutpad` option — seems to me that just adding `-toutpad bipad EN:A:P` should work (leaving Q unconnected)

kgugala: added

techlibs/quicklogic/cells_sim.v
kgugala: fixed

techlibs/quicklogic/cells_sim.v
kgugala: indeed, this was a leftover. Removed it.

techlibs/quicklogic/cells_sim.v
kgugala: removed it

techlibs/quicklogic/cells_sim.v
kgugala: fixed

techlibs/quicklogic/cells_sim.v
mwkmwkmwk: ... did you seriously "fix" this *again*, by bringing back _the other_ slightly less broken version that still doesn't have `QEN` signal?

techlibs/quicklogic/synth_quicklogic.cc
```diff
@@ -0,0 +1,141 @@
...
+
+        if (check_label("map")) {
+            run("techmap -map +/quicklogic/cells_map.v");
+            run("opt_clean");
+            run("check");
+            run("autoname");
+        }
+
+        if (check_label("iomap")) {
+            run("clkbufmap -buf $_BUF_ Y:A -inpad ckpad Q:P");
```
mwkmwkmwk, line 116: With my recent `clkbufmap` improvements in #2230, the `-buf $_BUF_ Y:A` part is no longer necessary and can be dropped

techlibs/quicklogic/synth_quicklogic.cc
mwkmwkmwk: Also, an update on the current FF type refactoring effort: recently, I added the `dfflegalize` pass which converts all yosys FF cells to ones actually supported by the target, and will soon be required in all synth flows.  This will actually simplify your mapping code, since now you're only required to describe the exact "native" cell types and don't have to eg. manually add inverters to match control signal polarity, or upconvert to more capable cell types — just list the cell types you directly support on the `dfflegalize` command line and only keep techmap rules for those.

techlibs/quicklogic/synth_quicklogic.cc
kgugala: rebased and removed the `-buf` switch

Ravenslofty:
Okay, this PR has been sat here with no activity since July. Let's give it a bump, because while there's a lot of cleanup still necessary, it's better for everyone if we get this merged sooner rather than later.

tests/arch/quicklogic/logic.ys
```diff
@@ -0,0 +1,16 @@
+read_verilog v/logic.v
```
Ravenslofty, line 1: Please use `common/logic.v` instead of unnecessarily bundling a different copy of it.

Ravenslofty:
Here's what *I* think needs to be done to make this mergeable:
- The LUT permutation code needs to be removed, because you can just use a different input ordering.
- Much of the FF techmap rules can be replaced with a `dfflegalize` call to Do The Right Thing.
  - It's probably best to just have a simple `$_DLATCH_` rule, much like iCE40 has.
- The FF mapping needs to be a separate step before the LUT mapping, rather than both performed simultaneously.
- Please actually, properly fix the `logic_cell_macro` model
- Use the existing common tests where available, instead of different copies.

Willmish:
Hey, is there any intention to get this merged upstream? (I know its been a while, but it would be cool to see this in the repo!) @kgugala

PR #1818, opened 2020-03-26
Add new FF types to library.

mwkmwkmwk:
This pull request adds the following new FF types to the library, in both coarse and fine variants:

- DFF with async set or reset and clock enable
- DFF with sync set or reset
- DFF with sync set or reset and clock enable, with set/reset having priority over clock enable

Do not merge this PR — it's meant to serve as basis for discussion about the new FF types.

If accepted, further steps are:

- recognize the new cells in all places looking for FFs
- switch over dff2dffe/dff2dffs to use the new types, as well as all flows currently using them
- add support for the coarse versions in dff2dffe/dff2dffs
- add support for the coarse versions in simplemap
- for the xilinx flow, move dff2dffe/dff2dffs upwards in the flow, change xilinx_dsp to recognize the new cells


daveshah1:
If we want to use these new types for iCE40 (and Intera iirc) too, then we would need a "enable having priority over set/reset" variant too.

While you are at it, it might be worth adding "DFFSR but with S over R priority" too so we have an even more complete set of FF types.

mwkmwkmwk:
Yeah, that's on the roadmap — this is just the first batch (specifically the subset of types prioritized so that we can clean up `xilinx_dsp` ASAP)

daveshah1:
Great, just making sure it wasn't forgotten

mwkmwkmwk:
Rebased on top of the current master — the changes are the same as in the original, but the actual fine cells are now python-generated, so the important part of the diff is much smaller.

eddiehung:
Would adding these new types to https://github.com/YosysHQ/yosys/blob/master/kernel/celltypes.h be appropriate for this PR? Otherwise, good to go.

mwkmwkmwk:
> Would adding these new types to https://github.com/YosysHQ/yosys/blob/master/kernel/celltypes.h be appropriate for this PR? Otherwise, good to go.

Not just celltypes.h, I've got a list of like 15 different files I'll need to add these cells to, there's a reason this is still just a draft PR...

eddiehung:
> > Would adding these new types to https://github.com/YosysHQ/yosys/blob/master/kernel/celltypes.h be appropriate for this PR? Otherwise, good to go.
> 
> Not just celltypes.h, I've got a list of like 15 different files I'll need to add these cells to, 

My thought pattern was that celltypes.h was what is used to `check` a design to ensure built-in cells maintain validity, not to use actually use them yet. 

> there's a reason this is still just a draft PR...

Is the intention to merge this PR, and then perform each of the bullet-pointed steps in follow-up PRs (in which case this can go in) or to do all of those points in this one PR?

mwkmwkmwk:
Alright, I consider the base PR ready for review now (though we may want to hold off merging it until it is actually tested by the follow-up PRs). Also note that this depends on #1884.

mwkmwkmwk:
Also: I changed table formatting in the manual.  I'm not entirely convinced I did the right thing, but at least the layout is not completely broken right now.

mwkmwkmwk:
Added three more kinds of FFs while I'm at it:

- DFF with async set, reset, and enable (`$dffsre`)
- DFF with sync set/reset and enable, enable has priority (`$dffes`)
- D latch with set or reset, but not both (`$dlatchr`)

kernel/rtlil.cc
```diff
@@ -47,6 +47,136 @@ int RTLIL::IdString::last_created_idx_ptr_;
 
 dict<std::string, std::string> RTLIL::constpad;
 
+const pool<IdString> RTLIL::builtin_ff_cell_types = {
```
cliffordwolf, line 50: This might be a C++ Static Initialization Order Fiasco bug.

mwkmwkmwk: Yeah, I've been worried about that. Any suggestions on a better place to do that?

(also, this is actually part of #1884, maybe let's review that first?)

whitequark:
IMO, `$dffs` should be called `$sdff`, because currently the naming is ambiguous with the `sr` variant.

mwkmwkmwk:
Renamed the new FF types, see the discussion at https://freenode.irclog.whitequark.org/~h~openfpga/2020-04-11

mwkmwkmwk:
@clairexen I think this PR should be mergeable in the current state.  There's more to come, but I'd rather have that in separate PRs:

1. Adding support for the new FF types to `write_verilog` and CXXRTL
2. The new `dfflegalize` pass
3. Converting the backends to use `dfflegalize`
4. Replacing `dff2dffe` and `dff2dffs` with `opt_dff` pass, and dealing with fallout from that (eg. `xilinx_dsp` needs to learn about new FF types)

PR #1828, opened 2020-03-26
kernel: share a single CellTypes within a pass

eddiehung:
Profiling showed that `CellTypes` was being created for each module in the design when inside `share`  and `memory_share`; rewrite to avoid doing that because constructing one can be expensive...

### DIFF
--- a/passes/memory/memory_share.cc
+++ b/passes/memory/memory_share.cc
@@ -666,7 +666,6 @@ struct MemoryShareWorker
 	// -------------
 
 	MemoryShareWorker(RTLIL::Design *design) : design(design), modwalker(design) {}
-	}
 
 	void operator()(RTLIL::Module* module)
 	{
--- a/passes/memory/memory_share.cc
+++ b/passes/memory/memory_share.cc
@@ -665,9 +665,7 @@ struct MemoryShareWorker
 	// Setup and run
 	// -------------
 
-	MemoryShareWorker(RTLIL::Design *design) :
-			design(design), modwalker(design)
-	{
+	MemoryShareWorker(RTLIL::Design *design) : design(design), modwalker(design) {}
 	}
 
 	void operator()(RTLIL::Module* module)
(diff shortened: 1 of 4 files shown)


passes/memory/memory_share.cc
```diff
@@ -665,9 +665,18 @@ struct MemoryShareWorker
 	// Setup and run
 	// -------------
 
-	MemoryShareWorker(RTLIL::Design *design, RTLIL::Module *module) :
-			design(design), module(module), sigmap(module)
+	MemoryShareWorker(RTLIL::Design *design) :
+			design(design), modwalker(design)
 	{
```
boqwxp: Why not just:
```suggestion
	MemoryShareWorker(RTLIL::Design *design) : design(design), modwalker(design) {}
```

passes/memory/memory_share.cc
```diff
@@ -665,9 +665,18 @@ struct MemoryShareWorker
...
+	}
+
+	void operator()(RTLIL::Module* module)
+	{
+		this->module = module;
+		sigmap.set(module);
+		sig_to_mux.clear();
+		conditions_logic_cache.clear();
+
 		std::map<std::string, std::pair<std::vector<RTLIL::Cell*>, std::vector<RTLIL::Cell*>>> memindex;
```
boqwxp: Kind of ugly.  Why not move the `memindex` declaration to the very top of the function?

passes/memory/memory_share.cc
```diff
@@ -755,8 +764,11 @@ struct MemorySharePass : public Pass {
 	void execute(std::vector<std::string> args, RTLIL::Design *design) YS_OVERRIDE {
 		log_header(design, "Executing MEMORY_SHARE pass (consolidating $memrd/$memwr cells).\n");
 		extra_args(args, 1, design);
+
```
boqwxp:
```suggestion
```

passes/opt/share.cc
```diff
@@ -1140,8 +1138,28 @@ struct ShareWorker
...
+	void operator()(RTLIL::Module *module) {
+		this->module = module;
 
-		modwalker.setup(design, module);
+	#ifndef NDEBUG
+		bool before_scc = module_has_scc();
+	#endif
+
+		limit = config.limit;
+
```
boqwxp:
```suggestion
```

passes/opt/share.cc
```diff
@@ -1140,8 +1138,28 @@ struct ShareWorker
...
+		modwalker.setup(module);
+
+		cells_to_remove.clear();
+		recursion_state.clear();;
```
boqwxp:
```suggestion
		recursion_state.clear();
```

PR #1829, opened 2020-03-26
simcells.v: Generate the fine FF cell types by a python script.

### LINKED ISSUE #1824: simcells.v: Generate the `$_DFF*` cells from python script
This will enable easy expansion to future large number of FF cell types.

mwkmwkmwk:
This makes adding more FF types in the future much more manageable.

Fixes #1824.


mithro:
@mwkmwkmwk -- I have a [spreadsheet documenting flip flops](https://docs.google.com/spreadsheets/d/16yvScRkedOkPCRBSInFuHE_cUMD8Ls-b9KFmHocpC-o/edit#gid=0).

I also had scripts which generate models at [`flipflop-download-csv.py`](https://github.com/mithro/symbiflow-arch-defs/blob/9b5df77cb29b65dbb4468d956c660b129537295a/utils/flipflop-download-csv.py), [`flipflop-generate.py`](https://github.com/mithro/symbiflow-arch-defs/blob/9b5df77cb29b65dbb4468d956c660b129537295a/utils/flipflop-generate.py) and [`flipflop.py`](https://github.com/mithro/symbiflow-arch-defs/blob/9b5df77cb29b65dbb4468d956c660b129537295a/utils/flipflop.py). The goal was to generate Yosys techmaps, good looking documentation (truth tables, schematic diagrams, etc).

mwkmwkmwk:
Perhaps; the question is whether the "in-place" generator is even a good idea in this case — for a proper makefile flow, I suppose we should have the non-generated cells in a separate file (`simcells_src.v`)?

eddiehung:
> Perhaps; the question is whether the "in-place" generator is even a good idea in this case — for a proper makefile flow, I suppose we should have the non-generated cells in a separate file (`simcells_src.v`)?

I'd be fine either way, especially as the way you've built it means that nothing changes if the script is run when it doesn't need to be.

PR #1832, opened 2020-03-28
Clean up pseudo-private member usage in `passes/cmds/design.cc`.

boqwxp:
Mostly straightforward.

The interesting parts are things like (orig. line 233):
```
			if (copy_to_design->modules_.count(prefix))
				delete copy_to_design->modules_.at(prefix);
```
changing to:
```
			if (copy_to_design->module(prefix) != nullptr)
				copy_to_design->remove(copy_to_design->module(prefix));
```

The `if` condition is a standard change.  But the mechanics change from `delete`ing it to `remove()`ing it.  `remove()` does `delete` it, too, but it also erases the entry in `modules_` (which should have been done before, no?) and notifies any monitors.

Similarly, the lines changed to e.g:
```
						RTLIL::Module *t = fmod->clone();
						t->name = trg_name;
						t->design = copy_to_design;
						t->attributes.erase("\\top");
						copy_to_design->add(t);
```
now use the proper `add()` method rather than brute forcing it into `modules_` directly.

### DIFF
--- a/passes/cmds/design.cc
+++ b/passes/cmds/design.cc
@@ -200,7 +200,7 @@ struct DesignPass : public Pass {
 					continue;
 				}
 				if (sel.selected_module(mod->name))
-					log_cmd_error("Module %s is only partly selected.\n", RTLIL::id2cstr(mod->name));
+					log_cmd_error("Module %s is only partly selected.\n", log_id(mod->name));
 			}
 
 			if (import_mode) {
--- a/passes/cmds/design.cc
+++ b/passes/cmds/design.cc
@@ -194,13 +194,13 @@ struct DesignPass : public Pass {
 				argidx = args.size();
 			}
 
-			for (auto &it : copy_from_design->modules_) {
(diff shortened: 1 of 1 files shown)


passes/cmds/design.cc
```diff
@@ -194,13 +194,13 @@ struct DesignPass : public Pass {
...
-					copy_src_modules.push_back(it.second);
+			for (auto mod : copy_from_design->modules()) {
+				if (sel.selected_whole_module(mod->name)) {
+					copy_src_modules.push_back(mod);
 					continue;
 				}
-				if (sel.selected_module(it.first))
-					log_cmd_error("Module %s is only partly selected.\n", RTLIL::id2cstr(it.first));
+				if (sel.selected_module(mod->name))
+					log_cmd_error("Module %s is only partly selected.\n", RTLIL::id2cstr(mod->name));
```
eddiehung: Since we're touching this code, I'd like to move towards getting `RTLIL::id2cstr()` and use `log_id()` directly please

PR #1833, opened 2020-03-28
Clean up pseudo-private member usage in `passes/sat/freduce.cc`.

boqwxp:


passes/sat/freduce.cc
```diff
@@ -828,8 +828,7 @@ struct FreducePass : public Pass {
 		extra_args(args, argidx, design);
 
 		int bitcount = 0;
-		for (auto &mod_it : design->modules_) {
-			RTLIL::Module *module = mod_it.second;
+		for (auto module : design->modules()) {
 			if (design->selected(module))
```
eddiehung:
```suggestion
		for (auto module : design->selected_modules()) {
```

boqwxp: Good catch.

PR #1840, opened 2020-03-30
greenpak4: Add final clean -purge

rqou:
gp4par does not like netnames that have bits set to constant values ("0" or "1"). get rid of them with a final `clean -purge`

### DIFF
--- a/techlibs/greenpak4/synth_greenpak4.cc
+++ b/techlibs/greenpak4/synth_greenpak4.cc
@@ -188,7 +188,7 @@ struct SynthGreenPAK4Pass : public ScriptPass
 			run("attrmvcp -attr src -attr LOC -driven t:GP_IBUF n:*");
 			run("techmap -map +/greenpak4/cells_map.v");
 			run("greenpak4_dffinv");
-			run("clean");
+			run("clean -purge");
 		}
 
 		if (check_label("check"))


eddiehung:
Not sure I'm following, can you show an example of a netname that has bits set to constants? Best would be a real simple test-case in `tests/arch/greenpak4` so that this gets fixed and stays fixed.

My concern is that a `clean -purge` is quite a heavy hammer that can blow away non-offending net names that would make for difficult debug.

rqou:
Sure, https://github.com/azonenberg/openfpga/blob/master/tests/greenpak4/slg46620v/Blinky.v demonstrates this problem

cliffordwolf:
One problem I see is that `clean -purge` doesn't actually fix the issue afaict. Only if the whole wire is assigned a constant value. But if only a few bits of the wire are assigned constant values that those assignment would still survive `clean -purge`. Thus I think the proper solution would be to fix gp4par.

whitequark:
I agree with the others here: Yosys should never call `clean -purge` on its own as this breaks end-to-end public net mapping, and gp4par should cope with constant bits.

whitequark:
There's also the `hilomap` pass that could perhaps make the gp4par changes simpler.

PR #1844, opened 2020-03-30
verilog: Add location info for generate constructs

### LINKED ISSUE #1839: Error gives the wrong (0) line number

* I ran `yosys -p "synth_ecp5 -json buglet.json"  buglet.v` where `buglet.v` had

```
module sad();
   reg [10:0] i;
   generate begin
      genvar i;
      for (i = 0; i < N; i = i + 1) begin : gen1
      end
   end
   endgenerate
endmodule
```

## Expected behavior
```
...
Generating RTLIL representation for module `\sad'.
buglet.v:5: ERROR: Left hand side of 1st expression of generate for-loop is not a gen var!
```

## Actual behavior
I got
```
...
Generating RTLIL representation for module `\sad'.
buglet.v:0: ERROR: Left hand side of 1st expression of generate for-loop is not a gen var!
```

In this boiled-down example it's obvious, but when there are many for-loop it would be nice to get the actual line.

daveshah1:
Fixes #1839

### DIFF
--- a/frontends/verilog/verilog_parser.y
+++ b/frontends/verilog/verilog_parser.y
@@ -2522,6 +2522,7 @@ gen_stmt:
 	} simple_behavioral_stmt ';' expr {
 		ast_stack.back()->children.push_back($6);
 	} ';' simple_behavioral_stmt ')' gen_stmt_block {
+		SET_AST_NODE_LOC(ast_stack.back(), @1, @11);
 		ast_stack.pop_back();
 	} |
 	TOK_IF '(' expr ')' {
@@ -2530,6 +2531,7 @@ gen_stmt:
 		ast_stack.push_back(node);
 		ast_stack.back()->children.push_back($3);
 	} gen_stmt_block opt_gen_else {
+		SET_AST_NODE_LOC(ast_stack.back(), @1, @7);
 		ast_stack.pop_back();
 	} |
 	case_type '(' expr ')' {
@@ -2538,6 +2540,7 @@ gen_stmt:
 		ast_stack.push_back(node);
 	} gen_case_body TOK_ENDCASE {
 		case_type_stack.pop_back();
+		SET_AST_NODE_LOC(ast_stack.back(), @1, @7);
 		ast_stack.pop_back();
 	} |
 	TOK_BEGIN {
@@ -2551,6 +2554,7 @@ gen_stmt:
 		exitTypeScope();
 		delete $3;
 		delete $7;
+		SET_AST_NODE_LOC(ast_stack.back(), @1, @7);
 		ast_stack.pop_back();
 	} |
 	TOK_MSG_TASKS {
@@ -2560,6 +2564,7 @@ gen_stmt:
 		ast_stack.back()->children.push_back(node);
 		ast_stack.push_back(node);
 	} opt_arg_list ';'{
+		SET_AST_NODE_LOC(ast_stack.back(), @1, @3);
 		ast_stack.pop_back();
 	};
 
@@ -2569,6 +2574,7 @@ gen_stmt_block:
 		ast_stack.back()->children.push_back(node);
 		ast_stack.push_back(node);
 	} gen_stmt_or_module_body_stmt {
+		SET_AST_NODE_LOC(ast_stack.back(), @2, @2);
 		ast_stack.pop_back();
 	};
 


frontends/verilog/verilog_parser.y
```diff
@@ -2522,6 +2522,7 @@ gen_stmt:
 	} simple_behavioral_stmt ';' expr {
 		ast_stack.back()->children.push_back($6);
 	} ';' simple_behavioral_stmt ')' gen_stmt_block {
+		SET_AST_NODE_LOC(ast_stack.back(), @1, @9);
```
boqwxp: If I'm counting correctly, the location of the `AST_GENFOR` node should begin with the location of the `TOK_FOR` (OK, I agree) and end with the location of the final `simple_behavioral_stmt`, i.e. the `i++` in `for(i = 0; i < 10; i++)`?

Should it not include the contents of the `gen_stmt_block` at `@11`?  I am honestly not sure.

frontends/verilog/verilog_parser.y
```diff
@@ -2530,6 +2531,7 @@ gen_stmt:
 		ast_stack.push_back(node);
 		ast_stack.back()->children.push_back($3);
 	} gen_stmt_block opt_gen_else {
+		SET_AST_NODE_LOC(ast_stack.back(), @1, @6);
```
boqwxp: Again, if I'm counting correctly, `@6` should be `gen_stmt_block`, not `opt_gen_else`.  Is this intentional?  Why is `opt_gen_else` not counted as part of the location for the `AST_GENIF`?  Does that `opt_gen_else` have its own cell type that gets its location properly set?

I'm not saying this is wrong; I honestly don't know.

frontends/verilog/verilog_parser.y
```diff
@@ -2538,6 +2540,7 @@ gen_stmt:
 		ast_stack.push_back(node);
 	} gen_case_body TOK_ENDCASE {
 		case_type_stack.pop_back();
+		SET_AST_NODE_LOC(ast_stack.back(), @1, @6);
```
boqwxp: Similarly, should the location include the `TOK_ENDCASE`?

frontends/verilog/verilog_parser.y
```diff
@@ -2551,6 +2554,7 @@ gen_stmt:
 		exitTypeScope();
 		delete $3;
 		delete $7;
+		SET_AST_NODE_LOC(ast_stack.back(), @1, @5);
```
boqwxp: And should this include `TOK_END`?

frontends/verilog/verilog_parser.y
```diff
@@ -2569,6 +2574,7 @@ gen_stmt_block:
 		ast_stack.back()->children.push_back(node);
 		ast_stack.push_back(node);
 	} gen_stmt_or_module_body_stmt {
+		SET_AST_NODE_LOC(ast_stack.back(), @1, @1);
```
boqwxp: Should this not be:
```suggestion
		SET_AST_NODE_LOC(ast_stack.back(), @2, @2);
```
?  I.e., the location of the `gen_stmt_or_module_body_stmt`?

frontends/verilog/verilog_parser.y
daveshah1: Ahh, does the counting include actions as well as tokens?

boqwxp: Yep.  See: https://www.gnu.org/software/bison/manual/html_node/Using-Midrule-Actions.html

```
The midrule action itself counts as one of the components of the rule. This makes a difference when there is another action later in the same rule (and usually there is another at the end): you have to count the actions along with the symbols when working out which number n to use in $n. 
```

daveshah1: Thanks for clearing up my misunderstanding! I've recounted them all and I think all the review comments should be cleared up, please could you take another look?

frontends/verilog/verilog_parser.y
```diff
@@ -2530,6 +2531,7 @@ gen_stmt:
 		ast_stack.push_back(node);
 		ast_stack.back()->children.push_back($3);
 	} gen_stmt_block opt_gen_else {
+		SET_AST_NODE_LOC(ast_stack.back(), @1, @7);
```
boqwxp, line 2534: I still do not know if this should be `(@1, @6)` or `(@1, @7)`.  Should the `opt_gen_else` be included in the `if` location?  Does the `opt_gen_else` have its own location assigned?  It looks like it doesn't.  I am not sure exactly how this generate stuff is parsed and represented as an AST, but if I have:
```
<<EOT
if (condition1)
...
else
...
EOT
```

Doesn't it make more sense for the `AST_GENIF` to be at `1.1-2.X` and the `opt_gen_else` node (if it exists?) be at `3.1-4.Y` rather than the `AST_GENIF` to be at `1.1-4.Y`?

frontends/verilog/verilog_parser.y
```diff
@@ -2551,6 +2554,7 @@ gen_stmt:
 		exitTypeScope();
 		delete $3;
 		delete $7;
+		SET_AST_NODE_LOC(ast_stack.back(), @1, @7);
```
boqwxp, line 2557: So the location ends at the label rather than `TOK_END`?  I am not sure.  I don't have a strong preference either way, though.

frontends/verilog/verilog_parser.y
```diff
@@ -2560,6 +2564,7 @@ gen_stmt:
 		ast_stack.back()->children.push_back(node);
 		ast_stack.push_back(node);
 	} opt_arg_list ';'{
+		SET_AST_NODE_LOC(ast_stack.back(), @1, @4);
```
boqwxp, line 2567: Don't include the trailing ';'.

frontends/verilog/verilog_parser.y
daveshah1: Looking at the `opt_gen_else` clause there isn't really an obvious else node, so I think it makes sense to associate it with the `if`.

frontends/verilog/verilog_parser.y
daveshah1: The label to me feels as much part of the construct as the end token

frontends/verilog/verilog_parser.y
daveshah1: Fixed

frontends/verilog/verilog_parser.y
boqwxp: OK, that's fine, then.  BTW, do you have an example `-dump-ast1` output from a relevant input file?

frontends/verilog/verilog_parser.y
boqwxp: Is it?  The label is just the label; the structure ends with the `TOK_END`...  Again, I don't have a strong preference here, but my opinion is that it should end with `TOK_END`'s location.

frontends/verilog/verilog_parser.y
daveshah1:
```verilog
module top(input a, output q);
	parameter MODE = 1'b0;
	generate
	if (MODE)
		assign q = !a;
	else
		assign q = a;
	endgenerate
endmodule
```

gives
```
    AST_MODULE <genif.v:1.1-9.10> [0x5628d920e900] str='\top'
      AST_WIRE <genif.v:1.18-1.19> [0x5628d920dde0] str='\a' input port=1
      AST_WIRE <genif.v:1.28-1.29> [0x5628d920df20] str='\q' output port=2
      AST_PARAMETER <genif.v:0.0-0.0> [0x5628d920e680] str='\MODE'
        AST_CONSTANT <genif.v:0.0-0.0> [0x5628d920e2e0] bits='0'(1) range=[0:0]
      AST_GENIF <genif.v:4.2-7.16> [0x5628d920e1a0]
        AST_IDENTIFIER <genif.v:4.6-4.10> [0x5628d920e060] str='\MODE'
        AST_GENBLOCK <genif.v:5.3-5.17> [0x5628d920e7c0]
          AST_ASSIGN <genif.v:5.10-5.16> [0x5628d920eb80]
            AST_IDENTIFIER <genif.v:5.10-5.11> [0x5628d9209d40] str='\q'
            AST_LOGIC_NOT <genif.v:5.14-5.16> [0x5628d920ede0]
              AST_IDENTIFIER <genif.v:5.15-5.16> [0x5628d920ef20] str='\a'
        AST_GENBLOCK <genif.v:7.3-7.16> [0x5628d92091d0]
          AST_ASSIGN <genif.v:7.10-7.15> [0x5628d9218ab0]
            AST_IDENTIFIER <genif.v:7.10-7.11> [0x5628d9207de0] str='\q'
            AST_IDENTIFIER <genif.v:7.14-7.15> [0x5628d9209860] str='\a'
```

boqwxp: OK, cool, thanks.

daveshah1: It doesn't seem related to this PR in particular though, as the same behaviour occurs with regular, non-generate if statements too:
```verilog
module top(input a, output reg q);
	always @*
		if (1'b1)
			q = !a;
endmodule
```
gives
```
    AST_MODULE <if.v:1.1-5.10> [0x55aa0194cf20] str='\top'
      AST_WIRE <if.v:1.18-1.19> [0x55aa0194c2e0] str='\a' input port=1
      AST_WIRE <if.v:1.32-1.33> [0x55aa0194c680] str='\q' output reg port=2
      AST_ALWAYS <if.v:2.2-4.11> [0x55aa0194c7c0]
        AST_BLOCK <if.v:2.11-4.11> [0x55aa0194c900]
          AST_CASE <if.v:3.3-4.11> [0x55aa0194cb80]
            AST_REDUCE_BOOL <if.v:0.0-0.0> [0x55aa01945de0]
              AST_CONSTANT <if.v:0.0-0.0> [0x55aa0194ca40] bits='1'(1) range=[0:0] int=1
            AST_COND <if.v:3.7-3.11> [0x55aa0194c1a0]
              AST_CONSTANT <if.v:0.0-0.0> [0x55aa019471d0] bits='1'(1) range=[0:0] int=1
              AST_BLOCK <if.v:4.4-4.11> [0x55aa0194cde0]
                AST_ASSIGN_EQ <if.v:4.4-4.10> [0x55aa0194bf20]
                  AST_IDENTIFIER <if.v:4.4-4.5> [0x55aa01947860] str='\q'
                  AST_LOGIC_NOT <if.v:4.8-4.10> [0x55aa0194bde0]
                    AST_IDENTIFIER <if.v:4.9-4.10> [0x55aa01947d40] str='\a'
```

boqwxp: Hmm, indeed, I see.  Maybe this is that annoying longstanding bug with `TOK_ID` that `verilog_parser.y` has a FIXME about...

PR #1845, opened 2020-03-30
kernel: speedup by using more pass-by-const-ref

eddiehung:
Especially `SigSpec` and `SigBit`.

Also speedup wire deletion by directly modifying `Module`'s private members.

Retry of #1827...

### DIFF
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -1586,6 +1586,7 @@ void RTLIL::Module::remove(const pool<RTLIL::Wire*> &wires)
 		const pool<RTLIL::Wire*> *wires_p;
 
 		void operator()(RTLIL::SigSpec &sig) {
+			sig.pack();
 			for (auto &c : sig.chunks_)
 				if (c.wire != NULL && wires_p->count(c.wire)) {
 					c.wire = module->addWire(NEW_ID, c.width);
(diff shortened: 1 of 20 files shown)


kernel/rtlil.cc
```diff
@@ -168,25 +170,24 @@ RTLIL::Const RTLIL::Const::from_string(std::string str)
 std::string RTLIL::Const::decode_string() const
 {
 	std::string string;
-	std::vector<char> string_chars;
-	for (int i = 0; i < int (bits.size()); i += 8) {
+	string.reserve(GetSize(bits)/8);
```
boqwxp: What about non-multiple-of-8 sizes?

kernel/rtlil.cc
```diff
@@ -168,25 +170,24 @@ RTLIL::Const RTLIL::Const::from_string(std::string str)
...
+	for (int i = 0; i < GetSize(bits); i += 8) {
 		char ch = 0;
 		for (int j = 0; j < 8 && i + j < int (bits.size()); j++)
 			if (bits[i + j] == RTLIL::State::S1)
 				ch |= 1 << j;
 		if (ch != 0)
-			string_chars.push_back(ch);
+			string.append({ch});
```
boqwxp: Is this `{ch}` in order to use the `std::string::append(const char *)` method?  If so, this is really dangerous and should be `{ch, 0}`.  But really you should probably just use:
```suggestion
			string.append(1, ch);
```

eddiehung: No, that's using C++11 an initializer_list.

kernel/rtlil.cc
eddiehung: The original code does not handle non-multiple-of-8, and it's not the intention of this PR to do so either.

kernel/rtlil.cc
boqwxp: Ah, phew, OK.  Still, isn't it clearer (potentially also to the compiler?) to do `append(1, ch)`, and also more performant?  I don't know enough about C++11 to have confidence, but doesn't the `append({ch})` way require instantiating an `initializer_list<char>` object?

kernel/rtlil.cc
boqwxp: OK, gotcha.

kernel/rtlil.cc
boqwxp, line 180: So, doesn't this cause an unnecessary instantiation of an `initializer_list<char>`?  Is it not more efficient to do `string.append(1, ch)`?  Maybe I misunderstand something, because you're even careful enough to use `emplace_back()` instead of `push_back()` above.

whitequark: An `initializer_list<char>` is a pair of pointers to stack; in principle, `append(initializer_list<char>)` should cost the same as `append(int, char)`.

PR #1846, opened 2020-03-30
Adding error message for when size (width) of number literal is zero

### LINKED ISSUE #1808: Weird JSON output for 0-bit constants
Take this Verilog code:
```verilog
module test2(input [1:0] i, output o);
  assign o = i;
endmodule
module test(output o);
  test2 x(0'b1, o);
endmodule
```
The following JSON is generated from it:
```json
        "x": {
          "hide_name": 0,
          "type": "test2",
          "parameters": {},
          "attributes": {
            "module_not_derived": 1,
            "src": "test.sv:5"
          },
          "port_directions": {
            "i": "input",
            "o": "output"
          },
          "connections": {
            "i": [],
            "o": [
              2
            ]
          }
        }
      },
```
See that there is nothing connected to the "i" input.

In other cases, Yosys makes sure that the input connection has the required number of bits. For example, if the `0'b1` is changed to `1'b1`, we now get:
```json
          "connections": {
            "i": [
              "1",
              "0"
            ],
            "o": [
              2
            ]
          }
```
So Yosys correctly zero-extended the one-bit constant to two bits.

I'm reporting this issue, because in my project, yosys2digitaljs, I'm relying on the connections being correctly extended. It works fine for every sensible example I've seen, but the zero-bit constant breaks things. I'm using yosys2digitaljs for teaching, and the test example comes from one of my students, who mistakenly thought `0'b1` is a 1-bit constant with value 0.

Is this a Yosys issue, or is it normal and I should accept these kinds of inputs?

dh73:
The following PR aims to close #1808 . Given that this is a front-end specific change, and even if the change may look simple, please feel free to comment if doesn't looks good enough or should be applied at a different level.

### DIFF
--- a/frontends/verilog/const2ast.cc
+++ b/frontends/verilog/const2ast.cc
@@ -139,6 +139,10 @@ static void my_strtobin(std::vector<RTLIL::State> &data, const char *str, int le
 		data.resize(len_in_bits, msb);
 	}
 
+	if (len_in_bits == 0)
+		log_error("Illegal integer constant size of zero in %s:%d (IEEE 1800-2012, 5.7).\n",
+				current_filename.c_str(), get_line_num());
+
 	if (len > len_in_bits)
 		log_warning("Literal has a width of %d bit, but value requires %d bit. (%s:%d)\n",
 			len_in_bits, len, current_filename.c_str(), get_line_num());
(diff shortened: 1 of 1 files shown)


frontends/verilog/const2ast.cc
```diff
@@ -139,6 +139,10 @@ static void my_strtobin(std::vector<RTLIL::State> &data, const char *str, int le
 		data.resize(len_in_bits, msb);
 	}
 
+	if (len_in_bits == 0)
+		log_error("Illegal integer constant size of zero in %s:%d (IEEE 1800-2012, 5.7).\n",
```
daveshah1: Shouldn't this be using log_file_error for consistency?

dh73: Good catch, thanks!

PR #1848, opened 2020-03-31
 ast: simplify to fully populate dynamic slicing case transformation

eddiehung:


eddiehung:
> This seems to have introduced a few warnings in `techlibs/xilinx/cells_sim.v`, specifically around L1706 and L1837 ([build log](https://travis-ci.org/github/YosysHQ/yosys/jobs/669380184#L3987)). Are they harmless?

Thanks for bringing this up. These warnings _shouldn't_ break anything, but of course it would be better if they were not there.

PR #1851, opened 2020-04-01
Improved rewrite code for writing to bit slice

cliffordwolf:
This adds the new rewrite rule. But it's still missing a check that makes
sure the new rewrite rule is actually a valid substitute in the always
block being processed. Therefore the new rewrite rule is just disabled
for now.

### DIFF
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -1721,8 +1721,10 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
 	}
 
 	// replace dynamic ranges in left-hand side expressions (e.g. "foo[bar] <= 1'b1;") with
-	// a big case block that selects the correct single-bit assignment.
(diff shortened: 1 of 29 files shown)


frontends/ast/simplify.cc
```diff
@@ -1725,18 +1730,65 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
 				log_file_error(filename, location.first_line, "Unsupported expression on dynamic range select on signal `%s'!\n", str.c_str());
 			result_width = abs(int(left_at_zero_ast->integer - right_at_zero_ast->integer)) + 1;
 		}
-		did_something = true;
-		newNode = new AstNode(AST_CASE, shift_expr);
-		for (int i = 0; i < source_width; i++) {
-			int start_bit = children[0]->id2ast->range_right + i;
-			AstNode *cond = new AstNode(AST_COND, mkconst_int(start_bit, true));
+
+		if (1)
```
cliffordwolf, line 1734: This needs to be changed to `if (0)` to test the new rewrite rule.

PR #1858, opened 2020-04-03
kernel: include "kernel/constids.inc"

### LINKED ISSUE #1856: ghdlsynth build breaks
ghdlsynth uses the following to compile its interface module:
```
yosys-config --exec --cxx -c --cxxflags -o $@ $<
```

A recent change to yosys breaks the inclusion of rtlil.h... at line 380 it tries to include "constids.inc" which appears only in the kernel directory, and is not in the include path.  A quick hack to put the yosys/kernel directory in the path manually results in a working ghdl module.

I had a quick look at the changesets and it's not immediately obvious which change causes this, or if the build break is caused by an incomplete build or install issue, but it is in the last 2 days.  I have been tracking both ghdlsynth and yosys development, and this is a new bug.

Cheers, and thanks to all the Yosys developers.  In conjunction with the GHDL developer's work, very much progress has been made in just the last few weeks.

eddiehung:
Fixes #1856

### DIFF
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -42,7 +42,7 @@ int RTLIL::IdString::last_created_idx_ptr_;
 #endif
 
 #define X(_id) IdString RTLIL::ID::_id;
-#include "constids.inc"
+#include "kernel/constids.inc"
 #undef X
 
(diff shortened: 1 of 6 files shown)


donnie-j:
Now seems to be fixed.  Thanks all.

PR #1859, opened 2020-04-03
Add `-push-copy` option to the `design` command.

boqwxp:
There is currently no way to push the current design onto the design stack without also clearing the current design.  This PR adds the `-duplicate` option to the `design` command, which allows exactly that functionality.


boqwxp:
How about `-push-copy`?

whitequark:
Also sounds reasonable to me.

PR #1861, opened 2020-04-05
Clean up `passes/hierarchy/submod.cc`.

boqwxp:


passes/hierarchy/submod.cc
```diff
@@ -274,12 +273,11 @@ struct SubmodWorker
 
 		if (opt_name.empty())
 		{
-			for (auto &it : module->wires_)
-				it.second->attributes.erase(ID::submod);
+			for (auto w : module->wires())
+				w->attributes.erase(ID::submod);
```
whitequark, line 277: nit: I prefer spelling out `wire` rather than single-letter variables (and elsewhere we have `cell` and `wire` and so on).

PR #1862, opened 2020-04-05
Clean up `passes/techmap/techmap.cc`

boqwxp:
This one has a few non-trivial changes around lines 496, 751, 769, and 1428.


passes/techmap/techmap.cc
```diff
@@ -1425,15 +1423,11 @@ struct FlattenPass : public Pass {
...
-			dict<RTLIL::IdString, RTLIL::Module*> new_modules;
-			for (auto mod : vector<Module*>(design->modules()))
-				if (used_modules[mod->name] || mod->get_blackbox_attribute(worker.ignore_wb)) {
-					new_modules[mod->name] = mod;
-				} else {
+			for (auto mod : design->modules())
+				if (!used_modules[mod->name] && !mod->get_blackbox_attribute(worker.ignore_wb)) {
 					log("Deleting now unused module %s.\n", log_id(mod));
-					delete mod;
+					design->remove(mod);
```
eddiehung, line 1429: Is calling `design->remove()` legal when iterating over `design->modules()`?
I think that's what the old code was avoiding.
It looks `Design::remove(Module *)` is missing an assert similar to this one:
https://github.com/YosysHQ/yosys/blob/516857f3ba662f467d56df562eeca9ed214ac702/kernel/rtlil.cc#L1661

boqwxp: Oops, good call.  Updated in b47215df6ce119819c7c1f8839a374fa3c978c13 .

passes/techmap/techmap.cc
```diff
@@ -1425,15 +1423,14 @@ struct FlattenPass : public Pass {
 							new_used_modules.insert(cell->type);
 			}
 
-			dict<RTLIL::IdString, RTLIL::Module*> new_modules;
-			for (auto mod : vector<Module*>(design->modules()))
-				if (used_modules[mod->name] || mod->get_blackbox_attribute(worker.ignore_wb)) {
-					new_modules[mod->name] = mod;
-				} else {
+			std::set<RTLIL::Module *> to_remove;
```
eddiehung, line 1426:
```suggestion
			pool<RTLIL::Module *> to_remove;
```

eddiehung:
To be honest, there's a lot of `std::map` and `std::set` use inside `techmap` that I've long wanted to get rid of. Would you do the honours please?

Also, can I suggest turning these datastructures:
https://github.com/YosysHQ/yosys/blob/b47215df6ce119819c7c1f8839a374fa3c978c13/passes/techmap/techmap.cc#L82
https://github.com/YosysHQ/yosys/blob/b47215df6ce119819c7c1f8839a374fa3c978c13/passes/techmap/techmap.cc#L808
to `IdString` instead of `std::string`?

boqwxp:
> Would you do the honours please?

With pleasure.

> Also, can I suggest turning these datastructures to IdString instead of std::string?

Roger that.

boqwxp:
@eddiehung This is ready for review.

kernel/rtlil.h
```diff
@@ -150,7 +150,7 @@ namespace RTLIL
 			if (!p[0])
 				return 0;
 
-			log_assert(p[0] == '$' || p[0] == '\\');
+			log_assert(p[0] == '$' || p[0] == '\\' || strncmp(p, "_TECHMAP_", strlen("_TECHMAP_")) == 0);
```
eddiehung, line 153: Why is this necessary? All IdString-s really should have `$` (private) or `\\` (public).

passes/techmap/techmap.cc
```diff
@@ -51,21 +51,21 @@ void apply_prefix(IdString prefix, RTLIL::SigSpec &sig, RTLIL::Module *module)
...
 struct TechmapWorker
 {
-	std::map<RTLIL::IdString, void(*)(RTLIL::Module*, RTLIL::Cell*)> simplemap_mappers;
-	std::map<std::pair<RTLIL::IdString, std::map<RTLIL::IdString, RTLIL::Const>>, RTLIL::Module*> techmap_cache;
-	std::map<RTLIL::Module*, bool> techmap_do_cache;
-	std::set<RTLIL::Module*, RTLIL::IdString::compare_ptr_by_name<RTLIL::Module>> module_queue;
+	dict<IdString, void(*)(RTLIL::Module*, RTLIL::Cell*)> simplemap_mappers;
+	dict<std::pair<IdString, dict<IdString, RTLIL::Const>>, RTLIL::Module*> techmap_cache;
+	dict<RTLIL::Module*, bool> techmap_do_cache;
+	std::set<RTLIL::Module*, IdString::compare_ptr_by_name<RTLIL::Module>> module_queue;
```
eddiehung, line 68: Can this not be transformed into a `pool`?

passes/techmap/techmap.cc
```diff
@@ -377,43 +377,43 @@ struct TechmapWorker
...
+			RTLIL::Cell *c = module->addCell(c_name, tpl_cell);
 			design->select(module, c);
 
 			if (!flatten_mode && c->type.begins_with("\\$"))
 				c->type = c->type.substr(1);
 
 			vector<IdString> autopurge_ports;
 
-			for (auto &it2 : c->connections_)
+			for (auto &conn : c->connections_)
```
eddiehung, line 400:
```suggestion
			for (auto &conn : c->connections())
```

passes/techmap/techmap.cc
```diff
@@ -566,9 +566,9 @@ struct TechmapWorker
 
 			for (auto &tpl_name : celltypeMap.at(cell_type))
 			{
-				RTLIL::IdString derived_name = tpl_name;
-				RTLIL::Module *tpl = map->modules_[tpl_name];
-				std::map<RTLIL::IdString, RTLIL::Const> parameters(cell->parameters.begin(), cell->parameters.end());
+				IdString derived_name = tpl_name;
+				RTLIL::Module *tpl = map->module(tpl_name);
+				dict<IdString, RTLIL::Const> parameters(cell->parameters.begin(), cell->parameters.end());
```
eddiehung, line 571:
```suggestion
				dict<IdString, RTLIL::Const> parameters(cell->parameters);
```

passes/techmap/techmap.cc
```diff
@@ -763,28 +763,31 @@ struct TechmapWorker
...
 				} else {
-					std::pair<RTLIL::IdString, std::map<RTLIL::IdString, RTLIL::Const>> key(tpl_name, parameters);
+					std::pair<IdString, dict<IdString, RTLIL::Const>> key(tpl_name, parameters);
 					if (techmap_cache.count(key) > 0) {
 						tpl = techmap_cache[key];
 					} else {
 						if (parameters.size() != 0) {
 							mkdebug.on();
-							derived_name = tpl->derive(map, dict<RTLIL::IdString, RTLIL::Const>(parameters.begin(), parameters.end()));
+							derived_name = tpl->derive(map, dict<IdString, RTLIL::Const>(parameters.begin(), parameters.end()));
```
eddiehung, line 790:
```suggestion
							derived_name = tpl->derive(map, parameters);
```

passes/techmap/techmap.cc
```diff
@@ -763,28 +763,31 @@ struct TechmapWorker
...
 				}
 
 				if (0) {
 			use_wrapper_tpl:;
 					// do not register techmap_wrap modules with techmap_cache
 				} else {
-					std::pair<RTLIL::IdString, std::map<RTLIL::IdString, RTLIL::Const>> key(tpl_name, parameters);
+					std::pair<IdString, dict<IdString, RTLIL::Const>> key(tpl_name, parameters);
 					if (techmap_cache.count(key) > 0) {
 						tpl = techmap_cache[key];
```
eddiehung, line 786:
```suggestion
					auto it = techmap_cache.find(std::make_pair(tpl_name, parameters));
					if (it != techmap_cache.end) {
						tpl = it->second;
```

kernel/rtlil.h
boqwxp: It's necessary in order to use `dict`s instead of `std::map`sIdStrings for one of the changes.  I'm sorry, I forgot which right now; I can find out which modification otherwise causes the crash but you'll have to give me a while.

passes/techmap/techmap.cc
boqwxp: This and the other remaining `std::set`s use a custom comparator, like `IdString::compare_ptr_by_name<>`.  It could be transformed into a `pool`, I just have not tried to make that work yet.

passes/techmap/techmap.cc
boqwxp: This can't be done trivially; the `apply_prefix()` at line 415 and the subsequent line require a non-`const` `conn`.  This can be refactored, I just haven't gotten there yet.

passes/techmap/techmap.cc
eddiehung: I think it might be okay to get rid of the custom comparator. It exists to ensure that we get a deterministic order -- without which you would be sorting pointer addresses.

passes/techmap/techmap.cc
```diff
@@ -1300,18 +1304,17 @@ struct TechmapPass : public Pass {
 
 		log_header(design, "Continuing TECHMAP pass.\n");
 
-		std::map<RTLIL::IdString, std::set<RTLIL::IdString, RTLIL::sort_by_id_str>> celltypeMap;
-		for (auto &it : map->modules_) {
-			if (it.second->attributes.count(ID::techmap_celltype) && !it.second->attributes.at(ID::techmap_celltype).bits.empty()) {
-				char *p = strdup(it.second->attributes.at(ID::techmap_celltype).decode_string().c_str());
-				for (char *q = strtok(p, " \t\r\n"); q; q = strtok(NULL, " \t\r\n"))
-					celltypeMap[RTLIL::escape_id(q)].insert(it.first);
+		dict<IdString, std::set<IdString, RTLIL::sort_by_id_str>> celltypeMap;
```
eddiehung, line 1307: For tracking purposes, this is the other `std::set` occurrence. This should also be safe to get rid of...

kernel/rtlil.h
eddiehung: See f3250f43

boqwxp: Wait, sorry, I'm confused.  Were the reverted changes around line 786 intentional or not?

passes/techmap/techmap.cc
boqwxp: OK, I see.  Deterministic ordering is already guaranteed by `pool`, right?  I'll try moving to `pool` without a special comparator and see what happens.

kernel/rtlil.h
eddiehung: My original commit (and the several that followed it) reverted your change here: https://github.com/YosysHQ/yosys/commit/f3250f4372cadc53fb7eecafe0596c71cc98881d#diff-f7c403795140175b79df71d870dd8971R572
But the latest force-push doesn't.

eddiehung: Oh dear. I see a few more. I've no idea why it would do this since it shouldn't have let me push if I wasn't up to date. Standby for another force push.

boqwxp: Gotcha.  OK, no problem.

eddiehung: e173d68b

boqwxp:
@eddiehung Nope, that appears to have caused a test failure:
```
cd tests/arch/ice40 && bash run-test.sh ""
make[1]: Entering directory '/home/user/github/yosys/tests/arch/ice40'
Running add_sub.ys..
ERROR: Assertion failed: selection contains 10 elements instead of the asserted 11: t:SB_LUT4
Selection contains:
top/A_SB_LUT4_O_I0_SB_LUT4_O
top/A_SB_LUT4_O_1
top/A_SB_LUT4_O
top/A_SB_LUT4_O_I3_SB_LUT4_O
top/B_SB_LUT4_O_2
top/B_SB_LUT4_O_I0_SB_LUT4_O
top/B_SB_LUT4_O_1
top/B_SB_LUT4_O
top/A_SB_LUT4_O_3
top/A_SB_LUT4_O_2
```

eddiehung:
> @eddiehung Nope, [dffd059](https://github.com/YosysHQ/yosys/commit/dffd059f176a9348d54703066cd19f602a0615ef) appears to have caused a test failure:
> 
> ```
> cd tests/arch/ice40 && bash run-test.sh ""
> make[1]: Entering directory '/home/user/github/yosys/tests/arch/ice40'
> Running add_sub.ys..
> ERROR: Assertion failed: selection contains 10 elements instead of the asserted 11: t:SB_LUT4
> Selection contains:
> top/A_SB_LUT4_O_I0_SB_LUT4_O
> top/A_SB_LUT4_O_1
> top/A_SB_LUT4_O
> top/A_SB_LUT4_O_I3_SB_LUT4_O
> top/B_SB_LUT4_O_2
> top/B_SB_LUT4_O_I0_SB_LUT4_O
> top/B_SB_LUT4_O_1
> top/B_SB_LUT4_O
> top/A_SB_LUT4_O_3
> top/A_SB_LUT4_O_2
> ```
> 
> Looks like I will have to implement the special comparator, or just leave them as `std::set`s.

Let me take a look

kernel/rtlil.h
eddiehung: In general, static IdString-s should be referenced with `ID::<name>` or `ID(name)` -- the latter being necessary for names beginning with `$`.

boqwxp:
I guess that works too.  The only thing left is to deal with the remaining pseudo-private `connections_` usage on line 401, right?

On an unrelated note, can you explain exactly when one should use `log_push()` and `log_pop()`?

passes/techmap/techmap.cc
```diff
@@ -377,43 +378,43 @@ struct TechmapWorker
...
 				}
 
 				if (autopurge) {
-					autopurge_ports.push_back(it2.first);
+					autopurge_ports.push_back(conn.first);
 				} else {
-					apply_prefix(cell->name, it2.second, module);
-					port_signal_map.apply(it2.second);
+					apply_prefix(cell->name, const_cast<RTLIL::SigSpec&>(conn.second), module);
+					port_signal_map.apply(const_cast<RTLIL::SigSpec&>(conn.second));
```
boqwxp, line 417: Are you cool with just `const_cast<>()`ing these?  This could also be:
```
                                       RTLIL::SigSpec new_conn = conn.second;
                                       apply_prefix(cell->name, new_conn, module);
                                       port_signal_map.apply(new_conn);
                                       c->setPort(conn.first, new_conn);
```

However, this is really an inner loop, and I'm hesitant to introduce a new copy when it's not really necessary, after all.

eddiehung: No `const_cast`, please! To be honest I don't think performance is that big an issue there -- and in any case, without any hard evidence we shouldn't overthink it. Let me try.

boqwxp: OK, no problem.  That's why I asked ;).

passes/techmap/techmap.cc
```diff
@@ -377,43 +378,45 @@ struct TechmapWorker
...
 						}
 				}
 
 				if (autopurge) {
-					autopurge_ports.push_back(it2.first);
+					autopurge_ports.push_back(conn.first);
 				} else {
-					apply_prefix(cell->name, it2.second, module);
-					port_signal_map.apply(it2.second);
+					RTLIL::SigSpec new_conn(std::move(conn.second));
```
eddiehung, line 416: I don't think this is triggering a move, since `conn.second` is const. If you can't write to it, then you can't move it to anything else.

boqwxp: I think you're right.

kernel/hashlib.h
```diff
@@ -100,6 +106,20 @@ template<typename P, typename Q> struct hash_ops<std::pair<P, Q>> {
 	}
 };
 
+template<typename P, typename Q> struct hash_ops<dict<P, Q>> {
```
eddiehung, line 109: Template specialisation is only necssary for data structures that we have no control over. Things that we do have control over should provide a `T::hash()` function so that the default template takes over:

https://github.com/YosysHQ/yosys/blob/da16a5b83a71db4f1d15b5ee3d4f6ab6f54fc0d2/kernel/hashlib.h#L59-L66

boqwxp: Oh, I see.

kernel/hashlib.h
```diff
@@ -630,6 +615,15 @@ class dict
...
 	}
 
+	unsigned int hash() const {
+		unsigned int h = mkhash_init;
+		for (auto &it : entries) {
+			h = mkhash(h, hash_ops<K>::hash(it.udata.first));
+			h = mkhash(h, hash_ops<T>::hash(it.udata.second));
+		}
+		return h;
+	}
```
cliffordwolf, line 625: This says that two dict are different if they only differ in iterator order. ("different" meaning they might map to different hash values.) I don't think this is a reasonable definition. Especially considering that `==` is written in a way that says that two dicts are equivalent if they contain the same key-value pairs, regardless of iterator order.

boqwxp: Good point.  c062047b02aa1d9fd001cb9a36b556358ccad450 should fix that.

boqwxp:
@eddiehung This one is also ready for review.

kernel/hashlib.h
eddiehung: @cliffordwolf What do you think of https://github.com/YosysHQ/yosys/commit/c062047b02aa1d9fd001cb9a36b556358ccad450? 
Making a copy of keys and values can be expensive; however, not copying the values will require a lookup (perhaps a pointer to the value could help).
Alternatively, I'd be fine with breaking const-ness and sorting all the entries, since `dict` doesn't guarantee a specific ordering (like insertion), but instead guarantees:
https://github.com/YosysHQ/yosys/blob/d1c8837572fb343baccff4d766a8aa0bca26aab7/CodingReadme#L79-L80

eddiehung:
For top-level visibility: https://github.com/YosysHQ/yosys/pull/1862#discussion_r419540197

kernel/hashlib.h
boqwxp: If breaking `const`ness is OK, then IMO would be even better to keep a `bool sorted` in `dict` that is set when sorted and unset when any write happens.  This way we can avoid a call to `std::sort` on an already-sorted dict every time we call `hash()`.

eddiehung: > If breaking `const`ness is OK, then IMO would be even better to keep a `bool sorted` in `dict` that is set when sorted and unset when any insert/emplace happens. This way we can avoid a call to `std::sort` on an already-sorted dict every time we call `hash()`.

Not sure I agree with that. Keeping the `dict` sorted is expensive, and we would only need it in case we hash. As evident by the fact we only needed a hash function now, I think we can infer that this necessity is quite rare.

daveshah1: I wonder, would it be possible to design a way of calculating the hash that is order-indepenedent (ie a linear combination of some function on keys/values) but still a good enough hash not to affect performance?

boqwxp: @eddiehung Nonono, I am not proposing to keep it sorted.  Just to keep a flag that is true when it *is* sorted, because we might insert some stuff, at some point sort it when call `hash()`, then never insert to it again.  In that case, subsequent calls to `hash()` will needlessly call `std::sort` on an already sorted list of entries.  Of course, if we *do* insert/emplace to it after we set the sorted flag, we should unset that flag and not worry about sorting again until we have to do it anyway inside `hash()`.

eddiehung: Sorry @boqwxp, I misread your answer. Let's revisit this after we've decided on the best approach to hashing.
@daveshah1 Thanks for the "order-independent hash" term; a quick smoogle led me to this:
https://stackoverflow.com/a/30735385, specifically
> The JDK itself proposes the following solution to this problem. The contract of the java.util.Set interface states:
>
>>Returns the hash code value for this set. The hash code of a set is defined to be the sum of the hash codes of the elements in the set, where the hash code of a null element is defined to be zero. This ensures that s1.equals(s2) implies that s1.hashCode()==s2.hashCode() for any two sets s1 and s2, as required by the general contract of Object.hashCode().
>
>An alternative to using the sum of the entries' hash codes would be to use, for example, the ^ (XOR) operator.
>
>The Scala language uses an ordering-invariant version of the Murmurhash algorithm (cf. the private scala.util.hashing.MurmurHash3 class) to implement the hashCode (or ##) method of its immutable sets and similar collections.

Summing sounds like a decent idea actually...

boqwxp: Should I just do it?

boqwxp: @eddiehung It's not immediately clear to me how to implement this for `dict` with the current hash function.  `mkhash()` takes an initial hash value along with the item to hash; in other words the hash value of each item is itself order-dependent.  Something like:
```
for(auto &it : entries) {
	h += mkhash(mkhash_init, hash_ops<K>::hash(it.udata.first));
	h += mkhash(mkhash_init, hash_ops<T>::hash(it.udata.second));
}
```
might work, but I assume that kind of usage is not intended, and I don't know if it is somehow problematic.

eddiehung: `mkhash` is a function that combines two hashes together, and right now you're combining it with the constant `mkhash_init` which would work but likely won't give an extra value than just:
```
for(auto &it : entries) {
	h += hash_ops<K>::hash(it.udata.first);
	h += hash_ops<T>::hash(it.udata.second);
}
```

boqwxp: Ah, OK, silly me.  I thought there was a need to use `mkhash` for some reason.

boqwxp: @eddiehung Nope, didn't work.  5b0cb2b causes a test failure.  Should I just revert it, or is there something else you'd like to try?

eddiehung: Actually, `mux.ys` is a very brittle test that keeps breaking (I can take significant blame for putting it in, that whole system didn't work out quite how I thought) so I wouldn't worry too much that it gets broken; these changes means that techmap iterates through the design in a slightly different order resulting in a slightly different circuit.

Out of interest, what OS do you have?

boqwxp: I see.  I am compiling and testing on Debian 10-based [Whonix 15 workstation](https://www.whonix.org/wiki/Main_Page#overview).

Anyway I changed it to use XOR and now it passes the test just fine. ¯\_(ツ)_/¯

boqwxp:
This is ready for final review by @cliffordwolf

passes/techmap/techmap.cc
boqwxp: It hasn't been lost exactly; I updated the for loop to:
```
for (auto mod : design->modules().to_vector())
```
It should be legal to delete modules while iterating over a vector copy of `modules()`, no?

In fact the original code did make a vector copy, but also did some buffer-swapping.  I think that was over-conservative and that simply adding `to_vector()` is cleaner than a few extra lines with an extra variable and for loop.

PR #1874, opened 2020-04-06
Clean up `passes/cmds/show.cc`.

boqwxp:


passes/cmds/show.cc
```diff
@@ -366,22 +364,20 @@ struct ShowWorker
 		std::set<std::string> all_sources, all_sinks;
 
 		std::map<std::string, std::string> wires_on_demand;
-		for (auto &it : module->wires_) {
-			if (!design->selected_member(module->name, it.first))
-				continue;
+		for (auto w : module->selected_wires()) {
```
whitequark, line 367: Same concern re: `w`/`wire` as in the other PR (but matters more in this PR).

Issue #1855, opened 2020-04-01
Source locations for some elements are missing or off by 1


boqwxp:
## Steps to reproduce the issue
```
module top(input a, output reg q);
	always @*
		if (1'b1)
			q = !a;
endmodule
```

## Actual behavior

`read_verilog -dump_ast1` gives:

```
    AST_MODULE <if.v:1.1-5.10> [0x55aa0194cf20] str='\top'
      AST_WIRE <if.v:1.18-1.19> [0x55aa0194c2e0] str='\a' input port=1
      AST_WIRE <if.v:1.32-1.33> [0x55aa0194c680] str='\q' output reg port=2
      AST_ALWAYS <if.v:2.2-4.11> [0x55aa0194c7c0]
        AST_BLOCK <if.v:2.11-4.11> [0x55aa0194c900]
          AST_CASE <if.v:3.3-4.11> [0x55aa0194cb80]
            AST_REDUCE_BOOL <if.v:0.0-0.0> [0x55aa01945de0]
              AST_CONSTANT <if.v:0.0-0.0> [0x55aa0194ca40] bits='1'(1) range=[0:0] int=1
            AST_COND <if.v:3.7-3.11> [0x55aa0194c1a0]
              AST_CONSTANT <if.v:0.0-0.0> [0x55aa019471d0] bits='1'(1) range=[0:0] int=1
              AST_BLOCK <if.v:4.4-4.11> [0x55aa0194cde0]
                AST_ASSIGN_EQ <if.v:4.4-4.10> [0x55aa0194bf20]
                  AST_IDENTIFIER <if.v:4.4-4.5> [0x55aa01947860] str='\q'
                  AST_LOGIC_NOT <if.v:4.8-4.10> [0x55aa0194bde0]
                    AST_IDENTIFIER <if.v:4.9-4.10> [0x55aa01947d40] str='\a'
```
## Expected behavior

`AST_IDENTIFIER <if.v:4.9-4.10> [0x55aa01947d40] str='\a'` should be `AST_IDENTIFIER <if.v:4.9-4.9> [0x55aa01947d40] str='\a'`

`AST_BLOCK <if.v:4.4-4.11> [0x55aa0194cde0]` should be `AST_BLOCK <if.v:4.4-4.9> [0x55aa0194cde0]`.  4.11 is an invalid location.  This is actually an off-by-two error, but probably just two compounded off-by-one errors.

## Credit

Thanks to @daveshah1 for the example.

[...]

PR #1877, opened 2020-04-06
Verilog frontend: Do not include semicolon in `behavioral_stmt` location.

boqwxp:
See issue #1855.

### DIFF
--- a/frontends/verilog/verilog_parser.y
+++ b/frontends/verilog/verilog_parser.y
@@ -2210,7 +2210,10 @@ simple_behavioral_stmt:
 behavioral_stmt:
 	defattr | assert | wire_decl | param_decl | localparam_decl | typedef_decl |
 	non_opt_delay behavioral_stmt |
-	simple_behavioral_stmt ';' | ';' |
+	simple_behavioral_stmt ';' {
+		SET_RULE_LOC(@$, @1, @1);
+	} |
+	';' |
 	hierarchical_id attr {
 		AstNode *node = new AstNode(AST_TCALL);
 		node->str = *$1;


eddiehung:
I think this is a PR for this comment https://github.com/YosysHQ/yosys/issues/1855#issuecomment-609846010 right?

In which case, I think the better fix would be to include the semi-colon for the block and the assignment. Otherwise, what does the semicolon belong to? 
Hypothetically, if an application down the line was to use these locations to rewrite the design, then wouldn't there be a stray semicolon left behind?

I may have got the wrong end of the stick, in which case a minimal test would be helpful to understand what you're fixing here please, as well as ensure it stays fixed.

boqwxp:
> I think this is a PR for this comment #1855 (comment) right?
Yep.

I don't think a trailing semicolon should be included in anything, in fact.  Semicolons are not meaningful in the design except as a way to separate items for the parser.

> Hypothetically, if an application down the line was to use these locations to rewrite the design, then wouldn't there be a stray semicolon left behind?

I actually already have [an application](https://www.epfl.ch/labs/lap/wp-content/uploads/2018/05/BeckerNov15_FudgeFactorSyntaxGuidedSynthesisForAccurateRtlErrorLocalizationAndCorrection_HVC15.pdf) using these locations to rewrite the design (based off a very old and quite diverged fork, not upstreamed except for these source locations).

What that application does is use the location information (and other stuff) to permute the AST after parsing but before translation to RTLIL.  If you want a Verilog circuit to look at, you can just use `write_verilog` as normal.  So there is no semicolon left behind.

I can think of other potential applications that do not use a full-blown parser to rewrite a file, but:
1) If Yosys is helping to rewrite the design, then ideally Yosys itself will do the rewriting, and Yosys doesn't care about source locations internally.  Yosys has a lot of commands to manipulate the design, so quite a lot of modifying applications should be fine to implement as a Yosys script (unless they need control flow).
2) If the application cannot be implemented in Yosys, one can still use the Yosys Verilog frontend to make a robust application.  And this is a good idea not just because of semicolon BS :).
3) Just how blindly is this application going to be copying/rewriting things?  If I know I am about to emit an `if()`, I know I don't need to also emit a semicolon.  Or, if I know I'm about to emit an `assign`, I will just emit a semicolon afterwards.  Maybe there's some marginal additional complexity for such dumb application, but it's not much.

eddiehung:
The application I mean when I say rewrite is to rewrite the original source file on disk, not any in memory representation. For example, the preservation of comments, #ifdefs, etc. minimising diff delta, would all be require surgical text modifications. Let's try and be open minded instead of discarding everything it just because you don't see a need for it. 

Let's take a step back and think about what source locations are used for. They are used for human debugging (in which case a semicolon doesn't make a difference), IDEs (where it could make a difference), and text manipulation applications (where this does matter).

boqwxp:
> Let's try and be open minded instead of discarding everything it just because you don't see a need for it.

I don't think that's a fair characterization of my position.  In my post above I argued for not including semicolons, yes, but I laid out a three-point breakdown of why I believed that any sort of source-to-source transformation would not be seriously affected by a decision not to include semicolons.

To be clear, my arguing for a position does not mean I do not keep an open mind.  It just means I would like a refutation of my arguments or an example of something I did not consider.

> the preservation of comments, #ifdefs, etc. minimising diff delta, would all be require surgical text modifications

Preservation of comments and #ifdefs is unrelated to including semicolons in source locations, no?  If I see there's a `behavioral_stmt` at the half-open range `1.3-1.15`, the only confusion is whether or not that includes the trailing semicolon.  Unless there are comments and #ifdefs between the `behavioral_stmt` and the trailing semicolon, there should be no relation.

Minimizing diff deltas is a good example I did not foresee.  In that case you may really be concerned about potential whitespace or comments between a `behavioral_stmt` and the trailing semicolon.  I still think this is not a big problem, though.  If you want to remove a `behavioral_stmt`, you know that there is a trailing semicolon and, assuming there are not comments between the `behavioral_stmt` and the trailing semicolon, you can find it simply by reading characters from the file until you get a semicolon.  I agree this is not ideal, but it's not that bad.

> They are used for human debugging (in which case a semicolon doesn't make a difference), IDEs (where it could make a difference), and text manipulation applications (where this does matter).

One of those text manipulation applications might require mapping back from source location to the actual string from the Verilog file.  It just seems more logical to me that mapping from the location of a `behavioral_stmt` would give e.g. `a = 4'b1010` vs. potentially `a = 4'b1010  /*magic number?*/;`.

Also, it's relatively easy to go from the former to the latter: Just keep reading until you get a semicolon, skipping over comments.  But going from the latter to the former is a bit more complicated: You start at the semicolon, then read backwards, looking ahead for comments, until you find the first character that's not whitespace or part of a comment.  The location of that character is one before the end location of the `behavioral_stmt` if the semicolon were not included.


Ultimately I think this is a matter of preference.  For some applications it doesn't matter, while for others it matters a bit, but some applications are more naturally implemented with semicolons included in source locations, and some are more naturally implemented with them excluded.

PR #1878, opened 2020-04-06
Gowin ABC9 support

Ravenslofty:
With the recent addition of timing data for Apicula, that can now be added to `specify` blocks and used for ABC9 synthesis and Eddie's static timing analysis.

There are a few things to note:
- Memory arrival times and I/O buffer delays haven't been reverse-engineered yet, so they're set to zero.
- The timings here are for GW1N (Littlebee) speed grade 4, making them very conservative. I could add GW2A (Arora) timings if desired, but I only know of a single GW2A devboard.
- The formatting looks a little weird in areas because I'm using Yosys-standard tabs while the surrounding code is two-space indented.
- ~~I've reduced the maximum LUT size from 8 to 6, because inter-CLU timings are presently unknown, and I'd rather not feed ABC9 fudged data.~~

Here are some numbers from the VexRiscV core in yosys-bench:
- ABC1 max LUT4 (`synth_gowin -nowidelut`): 25465ps critical path; 221 + 792 + 2848 + 2188 = 6049 LUTs
- ~~ABC1 max LUT6 (`synth_gowin`): 23981ps critical path; 537 + 714 + 2443 + 2735 = 6429 LUTs~~
- ABC1 max LUT8 (`synth_gowin`): 22473ps critical path; 729 + 852 + 1938 + 3757 = 7276 LUTs
- ABC9 max LUT4 (`synth_gowin -abc9 -nowidelut`): 19671ps critical path;   6 + 849 + 2825 + 2185 = 5865 LUTs
- ~~ABC9 max LUT6 (`synth_gowin -abc9`): 15652ps critical path; 114 + 819 + 1894 + 3223 = 6050 LUTs~~
- ABC9 max LUT8 (`synth_gowin -abc9`): 15576ps critical path; 187 + 1111 + 1913 + 3275 = 6486 LUTs

And that's a pretty solid win for ABC9.

### DIFF
--- a/techlibs/gowin/cells_sim.v
+++ b/techlibs/gowin/cells_sim.v
@@ -1,33 +1,112 @@
+(* abc9_lut=1 *)
 module LUT1(output F, input I0);
 	parameter [1:0] INIT = 0;
+	specify
+		(I0 => F) = (555, 902);
+	endspecify
 	assign F = I0 ? INIT[1] : INIT[0];
 endmodule
 
+(* abc9_lut=1 *)
 module LUT2(output F, input I0, I1);
 	parameter [3:0] INIT = 0;
+	specify
+		(I0 => F) = (867, 1184);
+		(I1 => F) = (555, 902);
+	endspecify
 	wire [ 1: 0] s1 = I1 ? INIT[ 3: 2] : INIT[ 1: 0];
 	assign F = I0 ? s1[1] : s1[0];
 endmodule
 
+(* abc9_lut=1 *)
 module LUT3(output F, input I0, I1, I2);
 	parameter [7:0] INIT = 0;
+	specify
+		(I0 => F) = (1054, 1486);
+		(I1 => F) = (867, 1184);
+		(I2 => F) = (555, 902);
+	endspecify	
 	wire [ 3: 0] s2 = I2 ? INIT[ 7: 4] : INIT[ 3: 0];
 	wire [ 1: 0] s1 = I1 ?   s2[ 3: 2] :   s2[ 1: 0];
 	assign F = I0 ? s1[1] : s1[0];
 endmodule
 
+(* abc9_lut=1 *)
 module LUT4(output F, input I0, I1, I2, I3);
 	parameter [15:0] INIT = 0;
+	specify
+		(I0 => F) = (1054, 1486);
+		(I1 => F) = (1053, 1583);
+		(I2 => F) = (867, 1184);
+		(I3 => F) = (555, 902);
+	endspecify	
 	wire [ 7: 0] s3 = I3 ? INIT[15: 8] : INIT[ 7: 0];
 	wire [ 3: 0] s2 = I2 ?   s3[ 7: 4] :   s3[ 3: 0];
 	wire [ 1: 0] s1 = I1 ?   s2[ 3: 2] :   s2[ 1: 0];
 	assign F = I0 ? s1[1] : s1[0];
 endmodule
 
+(* abc9_lut=2 *)
+module __APICULA_LUT5(output F, input I0, I1, I2, I3, M0);
+	specify
+		(I0 => F) = (1187, 1638);
+		(I1 => F) = (1184, 1638);
+		(I2 => F) = (995, 1371);
+		(I3 => F) = (808, 1116);
+		(M0 => F) = (486, 680);
+	endspecify	
+endmodule
+
+(* abc9_lut=4 *)
+module __APICULA_LUT6(output F, input I0, I1, I2, I3, M0, M1);
+	specify
+		(I0 => F) = (1187 + 136, 1638 + 255);
+		(I1 => F) = (1184 + 136, 1638 + 255);
+		(I2 => F) = (995 + 136, 1371 + 255);
+		(I3 => F) = (808 + 136, 1116 + 255);
+		(M0 => F) = (486 + 136, 680 + 255);
+		(M1 => F) = (478, 723);
+	endspecify	
+endmodule
+
+(* abc9_lut=8 *)
+module __APICULA_LUT7(output F, input I0, I1, I2, I3, M0, M1, M2);
+	specify
+		(I0 => F) = (1187 + 136 + 136, 1638 + 255 + 255);
+		(I1 => F) = (1184 + 136 + 136, 1638 + 255 + 255);
+		(I2 => F) = (995 + 136 + 136, 1371 + 255 + 255);
+		(I3 => F) = (808 + 136 + 136, 1116 + 255 + 255);
+		(M0 => F) = (486 + 136 + 136, 680 + 255 + 255);
+		(M1 => F) = (478 + 136, 723 + 255);
+		(M2 => F) = (478, 723);
+	endspecify	
+endmodule
+
+(* abc9_lut=16 *)
+module __APICULA_LUT8(output F, input I0, I1, I2, I3, M0, M1, M2, M3);
+		specify
+		(I0 => F) = (1187 + 136 + 136 + 136, 1638 + 255 + 255 + 255);
+		(I1 => F) = (1184 + 136 + 136 + 136, 1638 + 255 + 255 + 255);
+		(I2 => F) = (995 + 136 + 136 + 136, 1371 + 255 + 255 + 255);
+		(I3 => F) = (808 + 136 + 136 + 136, 1116 + 255 + 255 + 255);
+		(M0 => F) = (486 + 136 + 136 + 136, 680 + 255 + 255 + 255);
+		(M1 => F) = (478 + 136 + 136, 723 + 255 + 255);
+		(M2 => F) = (478 + 136, 723 + 255);
+		(M3 => F) = (478, 723);
+		endspecify	
+	endmodule
+
 module MUX2 (O, I0, I1, S0);
   input I0,I1;
   input S0;
   output O;
+
+	specify
+		(I0 => O) = (141, 160);
+		(I1 => O) = (141, 160);
+		(S0 => O) = (486, 680);
+	endspecify
+
   assign O = S0 ? I1 : I0;
 endmodule
 
@@ -35,6 +114,13 @@ module MUX2_LUT5 (O, I0, I1, S0);
   input I0,I1;
   input S0;
   output O;
+
+	specify
+		(I0 => O) = (141, 160);
+		(I1 => O) = (141, 160);
+		(S0 => O) = (486, 680);
+	endspecify
+
   MUX2 mux2_lut5 (O, I0, I1, S0);
 endmodule
 
@@ -42,6 +128,13 @@ module MUX2_LUT6 (O, I0, I1, S0);
   input I0,I1;
   input S0;
   output O;
+
+	specify
+		(I0 => O) = (136, 255);
+		(I1 => O) = (136, 255);
+		(S0 => O) = (478, 723);
+	endspecify
+
   MUX2 mux2_lut6 (O, I0, I1, S0);
 endmodule
 
@@ -49,6 +142,13 @@ module MUX2_LUT7 (O, I0, I1, S0);
   input I0,I1;
   input S0;
   output O;
+
+	specify
+		(I0 => O) = (136, 255);
+		(I1 => O) = (136, 255);
+		(S0 => O) = (478, 723);
+	endspecify
+
   MUX2 mux2_lut7 (O, I0, I1, S0);
 endmodule
 
@@ -56,29 +156,58 @@ module MUX2_LUT8 (O, I0, I1, S0);
   input I0,I1;
   input S0;
   output O;
+
+	specify
+		(I0 => O) = (136, 255);
+		(I1 => O) = (136, 255);
+		(S0 => O) = (478, 723);
+	endspecify
+
   MUX2 mux2_lut8 (O, I0, I1, S0);
 endmodule
 
+(* abc9_flop, lib_whitebox *)
 module DFF (output reg Q, input CLK, D);
 	parameter [0:0] INIT = 1'b0;
 	initial Q = INIT;
+
+	specify
+		(posedge CLK => (Q : D)) = (480, 660);
+		$setup(D, posedge CLK, 576);
+	endspecify
+
 	always @(posedge CLK)
 		Q <= D;
 endmodule
 
+(* abc9_flop, lib_whitebox *)
 module DFFE (output reg Q, input D, CLK, CE);
   parameter [0:0] INIT = 1'b0;
   initial Q = INIT;
+
+	specify
+		if (CE) (posedge CLK => (Q : D)) = (480, 660);
+		$setup(D, posedge CLK &&& CE, 576);
+		$setup(CE, posedge CLK, 63);
+	endspecify
+
   always @(posedge CLK) begin
     if (CE)
       Q <= D;
   end
 endmodule // DFFE (positive clock edge; clock enable)
 
-
+(* abc9_box, lib_whitebox *)
 module DFFS (output reg Q, input D, CLK, SET);
   parameter [0:0] INIT = 1'b1;
   initial Q = INIT;
+
+	specify
+		(posedge CLK => (Q : D)) = (480, 660);
+		$setup(D, posedge CLK, 576);
+		$setup(SET, posedge CLK, 63);
+	endspecify
+
   always @(posedge CLK) begin
     if (SET)
       Q <= 1'b1;
@@ -87,10 +216,18 @@ module DFFS (output reg Q, input D, CLK, SET);
   end
 endmodule // DFFS (positive clock edge; synchronous set)
 
-
+(* abc9_box, lib_whitebox *)
 module DFFSE (output reg Q, input D, CLK, CE, SET);
   parameter [0:0] INIT = 1'b1;
   initial Q = INIT;
+
+	specify
+		if (CE) (posedge CLK => (Q : D)) = (480, 660);
+		$setup(D, posedge CLK &&& CE, 576);
+		$setup(CE, posedge CLK, 63);
+		$setup(SET, posedge CLK, 63);
+	endspecify
+
   always @(posedge CLK) begin
     if (SET)
       Q <= 1'b1;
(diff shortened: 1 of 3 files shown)


techlibs/gowin/cells_sim.v
```diff
@@ -394,7 +680,7 @@ end
 
 endmodule
 
-
+(* abc9_box *)
```
eddiehung, line 683: LUTRAMs are not combinatorial-only boxes (they have comb/async read behavioru but sync write behaviour) and should not be marked as a box. See: https://github.com/YosysHQ/yosys/blob/516857f3ba662f467d56df562eeca9ed214ac702/techlibs/xilinx/cells_sim.v#L1385-L1439

In order to capture the comb read path, there needs to be another "bypass" box. In the above Xilinx LUTRAM, this is handled by this techmap rule:
https://github.com/YosysHQ/yosys/blob/516857f3ba662f467d56df562eeca9ed214ac702/techlibs/xilinx/abc9_map.v#L405-L426
On my TODO list is to do this transformation automatically.

techlibs/gowin/cells_sim.v
```diff
@@ -434,7 +726,7 @@ module RAM16S4 (DO, DI, AD, WRE, CLK);
 endmodule // RAM16S4
 
 
-(* blackbox *)
+(* abc9_box, blackbox *)
```
eddiehung, line 729: Even more so for BRAMs. It has no combinatorial paths through it whatsoever.

ZirconiumX: ...I kinda got into the mental habit of "uses `specify` blocks == needs `(* abc9_box *)`". Evidently this isn't the case.

pepijndevos:
I've looked a bit into the LUT8 issue. This is what it looks like when placed.
mux
In terms of routing, it's all hardwired. If I unpack the bitstream, I only see one single wire connected to OF7, the MUX2_LUT8 output.

In the timing data we see the following names, as used by the vendor tools.
```
parse_lut
{
'a_f': [0.7260000109672546, 0.7319999933242798, 1.0260000228881836, 1.031999945640564],
'b_f': [0.7239999771118164, 0.7310000061988831, 1.0609999895095825, 1.0989999771118164],
'c_f': [0.5559999942779541, 0.6019999980926514, 0.8019999861717224, 0.8220000267028809],
'd_f': [0.38499999046325684, 0.3720000088214874, 0.625, 0.6259999871253967],
'a_ofx': [0.8240000009536743, 0.8209999799728394, 1.1759999990463257, 1.180999994277954],
'b_ofx': [0.8219999670982361, 0.8199999928474426, 1.2109999656677246, 1.2480000257492065],
'c_ofx': [0.6539999842643738, 0.6909999847412109, 0.9520000219345093, 0.9710000157356262],
'd_ofx': [0.4829999804496765, 0.4610000252723694, 0.7749999761581421, 0.7749999761581421],
'm0_ofx0': [0.3370000123977661, 0.33399999141693115, 0.47200000286102295, 0.43700000643730164],
'm1_ofx1': [0.2750000059604645, 0.3319999873638153, 0.3880000114440918, 0.5019999742507935],
'fx_ofx1': [0.06599999964237213, 0.09399999678134918, 0.1770000010728836, 0.16300000250339508]}
```

Based on this, we are pretty certain that `[a-d]_f` is from lut input to output, `[a-d]_ofx` is from lut input to mux output. We think `m[01]_ofx[01]` are the delay from the select lines to the mux output of respectively the slice-local `MUX2_LUT5` and the inter-slice `MUX2_LUT[6-8]`. And I'm guessing `fx_ofx1` might refer to the extra delay introduced by a `MUX2_LUT[6-8]`.

At first sight it is a bit puzzling that there is no separate delay data for the three different `MUX2_LUT[6-8]`. But I think this might be reasonable, since the delay of the trace length is most likely negligible compared to the MOSFET parasitics. I would appreciate the thoughts of someone with more experience with other architectures on this.

ZirconiumX:
I've re-enabled LUT7/LUT8 in light of this, although it seems to drag area down for what appears to be minimal gain...

pepijndevos:
Would it be worth adding `-widelut n` syntax so the designer can choose the best trade-off?

ZirconiumX:
I'd say `-maxlut N` might be clearer, and has precedent in ABC9, but I'm a little concerned about option overload. Consider for example, that to use ABC9 to target LUT4 and then produce a (hypothetical) nextpnr netlist you'd need `synth_gowin -abc9 -maxlut 4 -json output.json`.

pepijndevos:
Here is the latest timing output with SRAM and BRAM.
[GW1NR-9_timing.txt](https://github.com/YosysHQ/yosys/files/4466136/GW1NR-9_timing.txt)
[GW1N-1_timing.txt](https://github.com/YosysHQ/yosys/files/4466138/GW1N-1_timing.txt)

And this is a dump of the exported symbols of the vendor library. I'm not sure what numbers you need for IO buffers and maybe other stuff, but if it's in here, it can be documented further.
[TmModel.txt](https://github.com/YosysHQ/yosys/files/4466137/TmModel.txt)

techlibs/gowin/cells_sim.v
```diff
@@ -1,84 +1,213 @@
...
+	specify
+		(I0 => O) = (136, 255);
+		(I1 => O) = (136, 255);
+		(S0 => O) = (478, 723);
+	endspecify
+
   MUX2 mux2_lut8 (O, I0, I1, S0);
 endmodule
 
+(* abc9_flop *)
```
eddiehung, line 169: Just to be clear: before #1926, this attribute will have no effect (but there's no harm).
After #1926, when there /should/ be an effect, then you will be needing `(* lib_whitebox *)` in there as well.

Same for all other flops below.

ZirconiumX:
Since `-W` is as much a fudge factor as anything, trying to optimise it for the correct value is quite tricky. If you have any ideas on *how* to optimise it, I'd like to hear.

techlibs/gowin/cells_sim.v
ZirconiumX: Fixed here and also in #1554.

techlibs/gowin/cells_sim.v
```diff
@@ -123,10 +275,17 @@ module DFFRE (output reg Q, input D, CLK, CE, RESET);
   end
 endmodule // DFFRE (positive clock edge; synchronous reset takes precedence over clock enable)
 
-
+(* abc9_flop, lib_whitebox *)
```
eddiehung, line 278: Sorry, only just thought to check this. Flops with async control cannot be modelled in ABC9 and so you shouldn't attach this here.

eddiehung:
> Since `-W` is as much a fudge factor as anything, trying to optimise it for the correct value is quite tricky. If you have any ideas on _how_ to optimise it, I'd like to hear.

It absolutely is a fudge. The only reason I use this fudge is a small value to encourage ABC9 to use a whole 6-input LUT for a 6-input gate, rather than use multiple LUTs and thus incurring _some_ wire delay. For example, setting `-W 0` here means ABC9 uses a LUT2+LUT5. At `-W 500` it also uses LUT8 for 8-input gate -- is this better area/delay than two LUTs?

techlibs/gowin/cells_sim.v
ZirconiumX: So anything that has an async set/reset shouldn't be an `abc9_flop`?

ZirconiumX:
> For example, setting `-W 0` here means ABC9 uses a LUT2+LUT5 [instead of a LUT6]. At `-W 500` it also uses LUT8 for 8-input gate -- is this better area/delay than two LUTs?

How do you figure? `1638 + 255 [LUT6 delay]` < `1638 [LUT5 delay] + 1184 [LUT2 delay]`, and a LUT6 is 4 LUT4s, while a LUT5 + LUT2 is 5 LUT4s, so I don't see why it would choose not to use a LUT6. However, at 16 LUT4s per LUT8, the main reason *not* to use a LUT8 would be that it could be formulated to use less area for less-critical paths (e.g. a LUT7 and a LUT2 is 9 LUT4s, not 16).

The built-in muxes for GW1N are surprisingly fast, honestly.

techlibs/gowin/cells_sim.v
ZirconiumX: Hopefully I got it right?

eddiehung:
> > For example, setting `-W 0` here means ABC9 uses a LUT2+LUT5 [instead of a LUT6]. At `-W 500` it also uses LUT8 for 8-input gate -- is this better area/delay than two LUTs?
> 
> How do you figure? `1638 + 255 [LUT6 delay]` < `1638 [LUT5 delay] + 1184 [LUT2 delay]`, and a LUT6 is 4 LUT4s, while a LUT5 + LUT2 is 5 LUT4s, so I don't see why it would choose not to use a LUT6. 

With `-W 0`, the delay is LUT2: 1184 + _680_ = 1864 -- since the LUT2 will use the fastest LUT5 input.
With `-W >0` the delay is LUT6: 1638 + 255 = 1893
Not much in it, but still something.

> However, at 16 LUT4s per LUT8, the main reason _not_ to use a LUT8 would be that it could be formulated to use less area for less-critical paths (e.g. a LUT7 and a LUT2 is 9 LUT4s, not 16).

In https://github.com/YosysHQ/yosys/pull/1878#issuecomment-612393741 you expressed that the current setup was expending area for minimal delay gain. In my 8-input gate example, your can use a LUT8 for 2403ps and 16xLUT4s, or at `-W 0` it is 2xLUT2 + 1xLUT6 (for 6xLUT4) for 2119ps. In this other extreme, is saving two trips through the interconnect worth almost 3 times the number of LUT4s?

pepijndevos:
@eddiehung I've been looking into the failed build, and it appears that `lib_whitebox` is causing extra init parameters to remain on the flop outputs. What would be the proper solution for this?

The `init.ys` test fails, because it detects 20 init parameters instead of the expected 8. Looking at the logs, there are 8 flops that gave a warning as expected. The async ones with the wrong init. The sync ones which synthed to a LUT seem to be fine, but the 12 that became a correct flop with reset seem to have a leftover init on their Q.

eddiehung:
It turns out that you were counting the init attributes inside whiteboxes too; as a workaround before #1949 comes online is to delete those whiteboxes for now. If the test was constructed with one top module, then you would be able to just select inside that top module.

Things I've noticed that can be improved in a different PR:
- Test should have one top module, then you can really test `synth_gowin` end-to-end without having to pull out `read_verilog` just to avoid `hierarchy`
- And to avoid `flatten`, you should rename the label and have `-noflatten` just control the `flatten` pass, not the entire block.

pepijndevos:
My thought was to only remove the init value if it corresponds with the reset value of the flop, so that it gives a warning in the weird edgecase where you have an async flop with a reset and init value that are different. It seems a bit harsh to fail completely, and bad to mismap silently. I'm open to alternatives of course.

eddiehung:
> My thought was to only remove the init value if it corresponds with the reset value of the flop, so that it gives a warning in the weird edgecase where you have an async flop with a reset and init value that are different. 

Unless I'm missing something, I don't think that's a weird edgecase. Initial states could come from RTL (perhaps even written for another architecture) or could be a result of some internal Yosys transformation/optimisation which the designer would never have imagined. That's why `ecp5_ffinit` exists for example (or for a more techmap-py replacement of that: #1925).

> It seems a bit harsh to fail completely, and bad to mismap silently. I'm open to alternatives of course.

Would it not be even harsher for a designer to miss the warning and then see it not behave as expected on hardware?
Restrictions on initial value is not uncommon and solutions do exist, take a look at #1925 and https://github.com/YosysHQ/yosys/blob/master/techlibs/xilinx/xc6s_ff_map.v

techlibs/gowin/cells_sim.v
```diff
@@ -681,10 +681,13 @@ module RAM16S4 (DO, DI, AD, WRE, CLK);
    input 	      CLK;
    input 	      WRE;
 
-  // TODO: check these numbers
   specify
-    if (WRE) (posedge CLK => (DO : DI)) = 0;
-    (AD => DO) = 0;
+    (AD => DO) = (270, 405);
+	// TODO: ask Eddie about positive/negative edge setup times.
```
eddiehung, line 686: Negative edge setup times are not supported by ABC9. They could be worked around inside Yosys by moving the zero reference point, but since I expect the magnitude of the negative value to be small, I haven't bothered. Instead they should be clamped to zero, e.g.:
https://github.com/YosysHQ/yosys/blob/ee3322af3340fd04fb0dba6c5159400b8c864800/techlibs/xilinx/cells_sim.v#L518

ZirconiumX:
You're parsing "negative edge setup times" wrong here, Eddie. I'm referring
to a positive setup time delay for a negative clock edge. As in,
$setup(..., negedge clk, ...).

eddiehung:
> You're parsing "negative edge setup times" wrong here, Eddie. I'm referring to a positive setup time delay for a negative clock edge. As in, $setup(..., negedge clk, ...).

Sorry. In which case, it doesn't appear that `RAM16S4` supports clock inversion nor is it DDR? Maybe I shouldn't jump the gun. 

In case it is still relevant, I've supported clock inversion like so:

https://github.com/YosysHQ/yosys/blob/ee3322af3340fd04fb0dba6c5159400b8c864800/techlibs/xilinx/cells_sim.v#L518-L519

The alternative is to wrap it inside a `generate`.

ZirconiumX:
Yes, the model is positive edge, but the delay given here is actually the
*negative* edge setup time (62.4ps), as that was higher than the *positive*
edge setup time (54ps).

eddiehung:
> Yes, the model is positive edge, but the delay given here is actually the *negative* edge setup time (62.4ps), as that was higher than the *positive* edge setup time (54ps).

Do you mean the setup time for rising data is different to falling data? From Yosys' point of view, we only care about the worst-case. If you really wanted to put this information in, it's not clear how you can do that. For simple/edge-sensitive paths, you can use `(rise,fall)`, however for timing constraint checks it doesn't look like you can? https://insights.sigasi.com/tech/systemverilog.ebnf/#setup_timing_check

ZirconiumX:
@eddiehung could you re-review this?

pepijndevos:
It's a semi-dual port block ram, described in the [Gowin FPGA Primitive User Guide](https://drive.google.com/file/d/1srsVjNKopNJeVP50kSqlagxAM6lKwwO2/view?usp=sharing) and has a full fledged simulation model at `IDE/simlib/gw1n/prim_sim.v`

clairexen:
@eddiehung @ZirconiumX is this ready to be merged or are there still open questions/issues?

ZirconiumX:
Ready to merge as far as I'm concerned.

pepijndevos:
I think we might need to double check Eddies concerns. My suspicion is the
extra DO constraint should be dropped, because it's a semi-dual-port so
there should be nothing between the write port and data out if I understand
correctly.
This is likely a mistake on our part since the same timing data is used for
the true dual port RAM.

ZirconiumX:
Which "extra DO constraint"? I don't have the simulation model here, and while I assume it's the `CLKA` constraint that has to go, could you double check?

pepijndevos:
I checked with the timing info, and indeed CLKA does not drive DO, so this line should be removed. As far as READ_MODE is concerned, I'd have to study it a bit more. Do you think there should be some conditional statements in the specify block that depend on the parameters?

techlibs/gowin/cells_sim.v
```diff
@@ -294,11 +542,25 @@ module GND(output G);
 	assign G = 0;
 endmodule
 
+(* abc9_box *)
 module IBUF(output O, input I);
+
+	// TODO: check these numbers.
+	specify
+		(I => O) = 0;
```
mwkmwkmwk, line 550: given that IBUF/OBUF implies leaving the chip, I doubt this does much good for synthesis

techlibs/gowin/cells_sim.v
```diff
@@ -408,6 +681,15 @@ module RAM16S4 (DO, DI, AD, WRE, CLK);
    input 	      CLK;
    input 	      WRE;
 
+  specify
+    (AD => DO) = (270, 405);
+	// TODO: ask Eddie about positive/negative edge setup times.
```
mwkmwkmwk, line 686: Is this still a TODO?

techlibs/gowin/cells_sim.v
Ravenslofty: ABC9 etc needs to know there's a path through this, otherwise it will error.

techlibs/gowin/cells_sim.v
pepijndevos: No, I think we resolved this. IIRC you just take the worst case because Yosys doesn't really differentiate. So the comment can go I think.

Ravenslofty:
Added `(* abc9_box *)` to flops that power up to one, removed the TODOs.

pepijndevos:
So at this point I think the basic functionality is there, and there is a big gain in QoR. There is a small open issue about non-standard BRAM modes, with the result that ABC might be on the conservative side. I don't currently have a lot of bandwidth to look into this due to hopefully graduating in August. I propose to merge this now and address more advanced features in a future PR.

techlibs/gowin/cells_sim.v
mwkmwkmwk: The TODO above should be removed, then?

techlibs/gowin/cells_sim.v
```diff
@@ -171,27 +353,48 @@ module DFFCE (output reg Q, input D, CLK, CE, CLEAR);
...
+	endspecify
+
   always @(negedge CLK) begin
     if (CE)
       Q <= D;
   end
 endmodule // DFFNE (negative clock edge; clock enable)
 
-
+(* abc9_flop, lib_whitebox *)
```
mwkmwkmwk, line 387: abc9_box

techlibs/gowin/cells_sim.v
```diff
@@ -200,10 +403,18 @@ module DFFNS (output reg Q, input D, CLK, SET);
   end
 endmodule // DFFNS (negative clock edge; synchronous set)
 
-
+(* abc9_flop, lib_whitebox *)
```
mwkmwkmwk, line 406: abc9_box

PR #1882, opened 2020-04-08
Clean up pseudo-private member usage in `passes/cmds/rename.cc`.

boqwxp:
Unfortunately this might have performance implications.  Instead of iterating through `wires_` or `cells_` once directly, creating one renamed copy, and swapping members, we now iterate through once to create `selected_wires()` or `selected_cells()`, then potentially iterate through everything again, where we then rename each element in place.

### DIFF
--- a/passes/cmds/rename.cc
+++ b/passes/cmds/rename.cc
@@ -258,22 +258,18 @@ struct RenamePass : public Pass {
 
 				for (auto wire : module->selected_wires())
 					if (wire->name[0] == '$') {
-						std::ostringstream buf;
-						do {
-							buf.str("");
-							buf << "\\" << pattern_prefix << counter++ << pattern_suffix;
-						} while (module->count_id(buf.str()) > 0);
-						new_wire_names[wire] = buf.str();
+						RTLIL::IdString buf;
+						do buf = stringf("\\%s%d%s", pattern_prefix.c_str(), counter++, pattern_suffix.c_str());
+						while (module->wire(buf) != nullptr);
+						new_wire_names[wire] = buf;
 					}
 
 				for (auto cell : module->selected_cells())
 					if (cell->name[0] == '$') {
-						std::ostringstream buf;
-						do {
-							buf.str("");
-							buf << "\\" << pattern_prefix << counter++ << pattern_suffix;
-						} while (module->count_id(buf.str()) > 0);
-						new_cell_names[cell] = buf.str();
+						RTLIL::IdString buf;
+						do buf = stringf("\\%s%d%s", pattern_prefix.c_str(), counter++, pattern_suffix.c_str());
+						while (module->cell(buf) != nullptr);
+						new_cell_names[cell] = buf;
 					}
 
 				for (auto &it : new_wire_names)
--- a/passes/cmds/rename.cc
+++ b/passes/cmds/rename.cc
@@ -213,16 +213,16 @@ struct RenamePass : public Pass {
 			for (auto module : design->selected_modules())
 			{
 				int counter = 0;
-				std::map<RTLIL::Wire *, IdString> new_wire_names;
-				std::map<RTLIL::Cell *, IdString> new_cell_names;
+				dict<RTLIL::Wire *, IdString> new_wire_names;
+				dict<RTLIL::Cell *, IdString> new_cell_names;
 
 				for (auto wire : module->selected_wires())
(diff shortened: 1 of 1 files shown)


passes/cmds/rename.cc
```diff
@@ -210,114 +210,102 @@ struct RenamePass : public Pass {
...
-			for (auto &mod : design->modules_)
+			for (auto module : design->selected_modules())
 			{
 				int counter = 0;
+				std::map<RTLIL::Wire *, IdString> new_wire_names;
+				std::map<RTLIL::Cell *, IdString> new_cell_names;
+
+				for (auto wire : module->selected_wires())
+					if (wire->name[0] == '$') {
+						std::ostringstream buf;
```
whitequark, line 261: I don't think using `ostringstream` is an improvement in readability over `stringf`. STL iostreams have a notoriously baroque interface, which is particularly bad when you need to use any formatting options, whereas printf-style functions don't have that issue. It is also more verbose.

whitequark:
I believe the performance implications don't matter in this case. `rename` is not a part of any synthesis flow (`autoname` is), so this is a pass that people will only call manually, and it is O(n) both before and after the PR.

passes/cmds/rename.cc
boqwxp: TBH I don't mind either way, I was just updating to use C++ stdlib instead of libc functions.

boqwxp:
> and it is O(n) both before and after the PR.

Well, sure, 2n is still O(n), but it's definitely worse than n.  But I get your point that it's not really on any critical path and so doesn't matter.

passes/cmds/rename.cc
nakengelhardt: Even though it looks like one, `stringf` isn't actually a libc function: it's defined in `yosys.h`. We prefer it over stringstreams so please don't convert use of `stringf`!

boqwxp: Oh, I see.  My bad :)

passes/cmds/rename.cc
```diff
@@ -210,114 +210,98 @@ struct RenamePass : public Pass {
 		{
 			extra_args(args, argidx, design);
 
-			for (auto &mod : design->modules_)
+			for (auto module : design->selected_modules())
 			{
 				int counter = 0;
+				std::map<RTLIL::Wire *, IdString> new_wire_names;
+				std::map<RTLIL::Cell *, IdString> new_cell_names;
```
eddiehung, line 217: Prefer dict to map

passes/cmds/rename.cc
```diff
@@ -210,114 +210,98 @@ struct RenamePass : public Pass {
...
-
-				dict<RTLIL::IdString, RTLIL::Cell*> new_cells;
-				for (auto &it : module->cells_) {
-					if (it.first[0] == '$' && design->selected(module, it.second))
-						it.second->name = derive_name_from_wire(*it.second);
-					new_cells[it.second->name] = it.second;
-				}
-				module->cells_.swap(new_cells);
+			for (auto module : design->selected_modules()) {
+				std::map<RTLIL::Cell *, IdString> new_cell_names;
```
eddiehung, line 240: Prefer dict to map

passes/cmds/rename.cc
eddiehung: Besides:
https://github.com/YosysHQ/yosys/blob/840bb170896c1acbbd321c4214df1f3b15108466/CodingReadme#L53-L80
using a `std::map<T*,U>` and then iterating over will lead to nondeterminism.

passes/cmds/rename.cc
```diff
@@ -210,114 +210,98 @@ struct RenamePass : public Pass {
...
 		if (flag_enumerate)
 		{
 			extra_args(args, argidx, design);
 
-			for (auto &mod : design->modules_)
+			for (auto module : design->selected_modules())
 			{
 				int counter = 0;
+				std::map<RTLIL::Wire *, IdString> new_wire_names;
+				std::map<RTLIL::Cell *, IdString> new_cell_names;
```
eddiehung, line 257: Prefer dict to map

whitequark:
@boqwxp Should the `kernel/hashlib.h` changes be in this PR?

boqwxp:
Oops, I forgot to update those.  Thanks.  I just rebased on master.

PR #1884, opened 2020-04-08
[NFCI] Deduplicate builtin FF cell types list

mwkmwkmwk:
A few passes included the same list of FF cell types.  Make it a global
const instead.

The zinit pass also seems to include a list like that, but given that
it seems to be completely broken at the time (see #1568 discussion),
I'm going to pretend I didn't see that.

### DIFF
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -47,6 +47,53 @@ int RTLIL::IdString::last_created_idx_ptr_;
 
 dict<std::string, std::string> RTLIL::constpad;
 
+const pool<IdString> &RTLIL::builtin_ff_cell_types() {
+	static const pool<IdString> res = {
+		ID($sr),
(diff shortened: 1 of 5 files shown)


kernel/rtlil.cc
```diff
@@ -47,6 +47,50 @@ int RTLIL::IdString::last_created_idx_ptr_;
 
 dict<std::string, std::string> RTLIL::constpad;
 
+const pool<IdString> RTLIL::builtin_ff_cell_types = {
```
cliffordwolf, line 50: Is this an instance of the C++ Static Initialization Order Fiasco? (I think the textbook solution to this would be to put this as static member into a function that returns a const reference to it.)

mwkmwkmwk: OK, does this look better?

PR #1885, opened 2020-04-08
Fix modulo/remainder semantics

### LINKED ISSUE #1523: Yosys modulus operator has remainder semantics
As per Wikipedia, usually modulus has the sign of the divisor, while remainder has the sign of the dividend. In languages that only have one `%` operator, you can get either behavior. https://en.wikipedia.org/wiki/Modulo_operation#In_programming_languages

However, VHDL is a language that has both a `mod` and a `rem` operator, where `rem` has the behavior of Yosys/Verilog `$mod`/`%`, and there is no way to represent actual VHDL `mod`.

Below is a program that when synthesized with GHDL currently throws an error. When using Verific, both statements result in a `$mod` operator, which is clearly incorrect.

```vhdl
library ieee;
use ieee.std_logic_1164.all;
use ieee.numeric_std.all;

entity lut2 is
    port (a: in integer;
          b: in integer;
          y: out integer;
          z: out integer);
end lut2;

architecture synth of lut2 is
begin
  y <= a mod b;
  z <= a rem b;
end synth;
```
yosys_show
Here is the offending Verific code:
https://github.com/YosysHQ/yosys/blob/db2268703f657fb977e1aa5506748683fbeb6cb1/frontends/verific/verific.cc#L483-L487

Xiretza:
Many languages have both modulo and remainder operations. They perform the same basic task (remainder of an integer division), but the sign of the output changes:
- The result of a remainder operation has the sign of the numerator (dividend)
- The result of a modulo operation has the sign of the denominator (divisor)

The `$mod` cell, being targeted at the Verilog `%` operator, used remainder semantics, making it hard to add a cell for actual modulo operations. `$mod` was renamed to `$rem`, and a new `$mod2` cell that performs denominator-signed modulo was added.

Fixes #1523.

### DIFF
--- a/backends/smv/smv.cc
+++ b/backends/smv/smv.cc
@@ -358,7 +358,8 @@ struct SmvWorker
 				continue;
 			}
 
-			if (cell->type.in(ID($div), ID($mod)))
+			// SMV has a "mod" operator, but its semantics don't seem to be well-defined - to be safe, don't generate it at all
+			if (cell->type.in(ID($div)/*, ID($mod), ID($modfloor)*/))
 			{
 				int width_y = GetSize(cell->getPort(ID::Y));
 				int width = max(width_y, GetSize(cell->getPort(ID::A)));
@@ -366,7 +367,7 @@ struct SmvWorker
 				string expr_a, expr_b, op;
 
 				if (cell->type == ID($div))  op = "/";
-				if (cell->type == ID($mod))  op = "mod";
+				//if (cell->type == ID($mod))  op = "mod";
 
 				if (cell->getParam(ID::A_SIGNED).as_bool())
 				{
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -948,7 +948,7 @@ namespace {
 				return;
 			}
 
-			if (cell->type.in(ID($add), ID($sub), ID($mul), ID($div), ID($mod), ID($pow))) {
+			if (cell->type.in(ID($add), ID($sub), ID($mul), ID($div), ID($mod), ID($modfloor), ID($pow))) {
 				param_bool(ID::A_SIGNED);
 				param_bool(ID::B_SIGNED);
 				port(ID::A, param(ID::A_WIDTH));
@@ -1949,6 +1949,7 @@ DEF_METHOD(Sub,      max(sig_a.size(), sig_b.size()), ID($sub))
 DEF_METHOD(Mul,      max(sig_a.size(), sig_b.size()), ID($mul))
 DEF_METHOD(Div,      max(sig_a.size(), sig_b.size()), ID($div))
 DEF_METHOD(Mod,      max(sig_a.size(), sig_b.size()), ID($mod))
+DEF_METHOD(ModFloor, max(sig_a.size(), sig_b.size()), ID($modfloor))
 DEF_METHOD(LogicAnd, 1, ID($logic_and))
 DEF_METHOD(LogicOr,  1, ID($logic_or))
 #undef DEF_METHOD
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -948,7 +948,7 @@ namespace {
 				return;
 			}
 
-			if (cell->type.in(ID($add), ID($sub), ID($mul), ID($div), ID($mod), ID($modfloor), ID($pow))) {
+			if (cell->type.in(ID($add), ID($sub), ID($mul), ID($div), ID($mod), ID($divfloor), ID($modfloor), ID($pow))) {
 				param_bool(ID::A_SIGNED);
 				param_bool(ID::B_SIGNED);
 				port(ID::A, param(ID::A_WIDTH));
@@ -1949,6 +1949,7 @@ DEF_METHOD(Sub,      max(sig_a.size(), sig_b.size()), ID($sub))
 DEF_METHOD(Mul,      max(sig_a.size(), sig_b.size()), ID($mul))
 DEF_METHOD(Div,      max(sig_a.size(), sig_b.size()), ID($div))
 DEF_METHOD(Mod,      max(sig_a.size(), sig_b.size()), ID($mod))
+DEF_METHOD(DivFloor, max(sig_a.size(), sig_b.size()), ID($divfloor))
 DEF_METHOD(ModFloor, max(sig_a.size(), sig_b.size()), ID($modfloor))
 DEF_METHOD(LogicAnd, 1, ID($logic_and))
 DEF_METHOD(LogicOr,  1, ID($logic_or))
--- a/techlibs/common/techmap.v
+++ b/techlibs/common/techmap.v
@@ -364,7 +364,8 @@ module \$__div_mod_u (A, B, Y, R);
 	end endgenerate
 endmodule
 
-module \$__div_mod (A, B, Y, R);
+// truncating signed division/modulo
+module \$__div_mod_trunc (A, B, Y, R);
 	parameter A_SIGNED = 0;
 	parameter B_SIGNED = 0;
 	parameter A_WIDTH = 1;
@@ -420,7 +421,7 @@ module _90_div (A, B, Y);
 	(* force_downto *)
 	output [Y_WIDTH-1:0] Y;
 
-	\$__div_mod #(
+	\$__div_mod_trunc #(
 		.A_SIGNED(A_SIGNED),
 		.B_SIGNED(B_SIGNED),
 		.A_WIDTH(A_WIDTH),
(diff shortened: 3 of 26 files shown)


Xiretza:
I've used `$mod2` as the name for the new modulo gate for now, suggestions welcome.

ZirconiumX:
cc @whitequark because I think she wanted this for nMigen.

whitequark:
As proposed, this PR is a breaking change in Yosys. That is, once this pull request is merged, nMigen will be no longer compatible with Yosys.

Xiretza:
Right, I think the only breaking change is the removal of `RTLIL::Module::addMod()` and `RTLIL::Module::Mod()`, if API compatibility should be kept there, I can add those back.

ZirconiumX:
@whitequark There is one quirk that might be relevant: post-patch Yosys will emit $mod2/$rem which won't be compatible with pre-patch Yosys.

Xiretza:
Oops, looks like I misunderstood the definition of "mod" myself. It's not simply the negation of the remainder, I'll try to find a way to elegantly model it in verilog for simlib and techmap.

Xiretza:
Alright, I cleaned the commits up a bit - `$mod2` behaviour should now be correct everywhere, and the remainder operator is now still called `$mod` instead of `$rem`, since that was apparently decided to be the best option. As for the name for the actual modulo operator, @mwkmwkmwk suggested `$truemod`, which sounds good to me - I'll wait on an official decision before substituting it though.

whitequark:
What happens if you pass `$truemod` to the Verilog backend?

Xiretza:
It produces an instance of the `$truemod` cell. Should it generate equivalent verilog code instead or is it enough if the cell is in simlib?

whitequark:
It should generate equivalent Verilog code; a significant part of the utility of the Verilog backend is being able to convert RTLIL to something the vendor tools can understand.

Xiretza:
Makes sense, it's now handled using a ternary operator - is that okay or should it be a simpler netlist?

whitequark:
I'm not the best person to ask here, unfortunately--I'm not very familiar with efficient implementations of arithmetics in hardware.

Xiretza:
Oh it's not really about efficiency (the remainder operator will far overshadow any glue logic anyway), but rather if using a ternary operator in `write_verilog` output is okay or if it should only use more primitive language features.

whitequark:
That should be perfectly fine, I think it will emit ternaries for muxes already.

whitequark:
> The div and mod definitions used by Verilog are commonly referred to as "T-definition" (truncation):

Should we have `$divt`, `$modt`, `$divf`, `$modf` cells then, rather than the current ambiguous names? I think this would satisfy your preference (having the cell for Verilog modulo be called `$mod`, or at least something close to that) and my preference (having the cell names be completely unambiguous). Also, if we do this, then the upgrade path becomes easy to understand too, with the old cells simply being aliases for the new T-semantics cells.

Xiretza:
Those figures are very helpful, thanks. As discussed previously, the `$mod` cell won't be renamed. From [this handy chart on Wikipedia](https://en.wikipedia.org/wiki/Modulo_operation#In_programming_languages) it's obvious that almost all possible semantics can be achieved with "mod" in some language, so `$mod` can stay the same, it's ambiguous either way. While "rem"/"remainder" seems to be used exclusively for F-definition, the best naming would probably avoid both of these options though, and rather make the rounding mode explicit. I think this gives three possible options:

1. Keep only `$mod` and `$div` and parameterize them for different rounding modes (not sure if this is possible with yosys' cell architecture)
2. Keep `$mod` and `$div` for truncating operations, add new `$modf`/`$divf` cells for flooring operations (and possibley more in the future, should other rounding modes ever be supported)
3. As @whitequark suggests, drop `$mod` and `$div` in a backwards-compatible way and add separate mod/div cells for each rounding mode

Which of these, if any, would be preferred?

cliffordwolf:
> Should we have $divt, $modt, $divf, $modf cells then, rather than the current ambiguous names?

I'm a bit reluctant to change the current cell names.

I was even considering to suggest `$divf` and `$modf` for the new cells, but I don't quite like those names because of possible confusion with the `modf()` std C function.

My personal preference would be to stick with `$mod` and `$div` for truncating behavior, and use `$modfloor` and `$divfloor` for the new cells.

kernel/rtlil.cc
```diff
@@ -1892,7 +1892,9 @@ DEF_METHOD(Add,      max(sig_a.size(), sig_b.size()), ID($add))
 DEF_METHOD(Sub,      max(sig_a.size(), sig_b.size()), ID($sub))
 DEF_METHOD(Mul,      max(sig_a.size(), sig_b.size()), ID($mul))
 DEF_METHOD(Div,      max(sig_a.size(), sig_b.size()), ID($div))
+DEF_METHOD(Mod2,     max(sig_a.size(), sig_b.size()), ID($mod2))
 DEF_METHOD(Mod,      max(sig_a.size(), sig_b.size()), ID($mod))
+DEF_METHOD(Rem,      max(sig_a.size(), sig_b.size()), ID($mod))
```
cliffordwolf, line 1897: I don't see the need for adding a "Rem" method for creating `$mod` cells in addition to the "Mod" method.

techlibs/common/techmap.v
```diff
@@ -310,10 +310,10 @@ endmodule
 
 
 // --------------------------------------------------------
-// Divide and Modulo
+// Division, Modulo, Remainder
```
cliffordwolf, line 313: ~Remainder~

techlibs/common/techmap.v
```diff
@@ -310,10 +310,10 @@ endmodule
...
 // --------------------------------------------------------
 
-module \$__div_mod_u (A, B, Y, R);
+module \$__div_rem_u (A, B, Y, R);
```
cliffordwolf, line 316: I don't think renaming this makes much sense

techlibs/common/techmap.v
```diff
@@ -339,7 +339,7 @@ module \$__div_mod_u (A, B, Y, R);
 	end endgenerate
 endmodule
 
-module \$__div_mod (A, B, Y, R);
+module \$__div_mod_rem (A, B, Y, M, R);
```
cliffordwolf, line 342: Same here. This is the assumption that "mod" and "rem" are clearly defined terms, which they are not.

techlibs/common/techmap.v
```diff
@@ -352,17 +352,17 @@ module \$__div_mod (A, B, Y, R);
 
 	input [A_WIDTH-1:0] A;
 	input [B_WIDTH-1:0] B;
-	output [Y_WIDTH-1:0] Y, R;
+	output [Y_WIDTH-1:0] Y, M, R;
```
cliffordwolf, line 355: Maybe it would be better to have a separate `\$__div_mod_floor` module for flooring mod and div.

techlibs/common/techmap.v
```diff
@@ -400,6 +405,31 @@ module _90_div (A, B, Y);
 	);
 endmodule
 
+(* techmap_celltype = "$mod2" *)
+module _90_mod2 (A, B, Y);
```
cliffordwolf, line 409: I think a better name instead of `$mod2` would be `$modfloor`, and if we add that we should also add a matching `$divfloor`.

whitequark:
> My personal preference would be to stick with `$mod` and `$div` for truncating behavior, and use `$modfloor` and `$divfloor` for the new cells.

Sounds good to me.

backends/smv/smv.cc
```diff
@@ -358,15 +358,16 @@ struct SmvWorker
 				continue;
 			}
 
-			if (cell->type.in(ID($div), ID($mod)))
+			// SMV has a "mod" operator, but its semantics don't seem to be well-defined - to be safe, don't generate it at all
+			if (cell->type.in(ID($div)/*, ID($mod), ID($modfloor)*/))
```
Xiretza, line 362: @cliffordwolf: you implemented this back in 8a86162ae9a4e2939af162edfc6fa281931f786e, do you have any further information on the exact semantics of SMV's "mod"?

Xiretza:
What's the opinion on `RTLIL::Module::[add](Div/Mod)Trunc()` methods that do exactly the same as the existing `[add](Div/Mod)()`? It would help external API consistency somewhat by making explicit names available for every operation, but because it would be the only place "truncating" is used in the context of division, it would also be somewhat internally inconsistent.

backends/smv/smv.cc
cliffordwolf: Good question. I have no idea.

cliffordwolf:
> What's the opinion on RTLIL::Module::[add](Div/Mod)Trunc() methods that do exactly the same as the existing [add](Div/Mod)()?

I'm not a fan, similar to https://github.com/YosysHQ/yosys/pull/1885#discussion_r411357362.

Xiretza:
Alright, I removed the methods and rebased on master. Ready to merge from my side.

whitequark:
@Xiretza Given that we are stuck with some of the more confusing possible naming schemes, please describe the precise behavior of all of the cells in the manual.

Xiretza:
Added a section to the manual, added some short cell help, and incorporated aee43936. Let me know if you want the documentation expanded in any further way.

whitequark:
All looks good to me now, thank you!

PR #1913, opened 2020-04-13
opt_expr: Add more $alu optimizations.

### LINKED ISSUE #1912: optimise $alu cells with known propagate/generate states
Consider the following example based on some code I wrote:

```verilog
module example(input [63:0] i, output [63:0] o);
wire tmp;
assign tmp = i & 64'hFEFEFEFEFEFEFEFE;
assign o = i & -i;
endmodule
```

Here, Yosys should notice that carries can neither propagate nor generate from one byte to the next, and bisect the 64-bit `$alu` cell into 8 8-bit `$alu` cells, drastically shortening the carry propagation delay.

It's also possible that in the middle of a `$alu` there's a generate term (e.g. 1+1). The `$alu` cell could be bisected at that point too.

mwkmwkmwk:
Detect the places in the $alu where the carry bit is constant (due to
const A[i] == B[i] ^ BI) and split it into smaller $alu at these points.

Also, make the existing const-carry detection for low bits more generic
(now handles cases where both BI and CI are constant, but not equal to
one another).

Fixes #1912.

### DIFF
--- a/passes/opt/opt_expr.cc
+++ b/passes/opt/opt_expr.cc
@@ -717,31 +717,27 @@ void replace_const_cells(RTLIL::Design *design, RTLIL::Module *module, bool cons
 				RTLIL::SigSpec sig_y = cell->getPort(ID::Y);
 				RTLIL::SigSpec sig_co = cell->getPort(ID::CO);
 
-				bool sub = (sig_ci == State::S1 && sig_bi == State::S1);
+				if (sig_bi != State::S0 && sig_bi != State::S1)
+					goto skip_fine_alu;
+				if (sig_ci != State::S0 && sig_ci != State::S1)
+					goto skip_fine_alu;
 
-				// If not a subtraction, yet there is a carry or B is inverted
-				//   then no optimisation is possible as carry will not be constant
-				if (!sub && (sig_ci != State::S0 || sig_bi != State::S0))
-					goto next_cell;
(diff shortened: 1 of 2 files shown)


mwkmwkmwk:
(draft because tests pending)

passes/opt/opt_expr.cc
```diff
@@ -1530,6 +1528,101 @@ void replace_const_cells(RTLIL::Design *design, RTLIL::Module *module, bool cons
...
+				sig_ci = p.second;
+			}
+
+			cover("opt.opt_expr.alu_split");
+			module->remove(cell);
+
+			did_something = true;
+			goto next_cell;
+
+
```
eddiehung, line 1622: Stray newlines

tests/opt/opt_expr_alu.ys
```diff
@@ -61,3 +61,51 @@ equiv_opt opt_expr -fine
...
+
+design -reset
+read_verilog <<EOT
+module test(input [1:0] a, output [3:0] y);
+assign y = -{a[1], 2'b10, a[0]};
+endmodule
+EOT
+
+alumacc
+equiv_opt opt -fine
```
eddiehung, line 74: Just noticed that this is missing here and in everything (that I did) above:
```suggestion
equiv_opt -assert opt -fine
```

mwkmwkmwk:
(fixed up review comments)

PR #1915, opened 2020-04-14
kernel: Add `dict` support for rvalue references and C++11 move semantics.

boqwxp:
This avoids an unnecessary copy of the `std::pair<K, T>` in the `entry_t` constructor when the inserted key/value pair is an rvalue and also always avoids an unnecessary `std::move()` into the `entries` member variable by instead constructing the `entry_t` in place with `emplace_back()`.

Inserting an rvalue into a `dict` is actually not super uncommon throughout the code base:
```
user@host:~/github/yosys$ find . -type f | xargs grep -n "\.insert(.*std::make_pair" | grep -v "^Binary file"
./techlibs/ice40/ice40_opt.cc:128:							new_attr.insert(std::make_pair(a.first, a.second));
./passes/hierarchy/submod.cc:187:						auto jt = new_wire->attributes.insert(std::make_pair(ID::init, Const(State::Sx, GetSize(sig)))).first;
./passes/pmgen/ice40_wrapcarry.cc:147:						carry->attributes.insert(std::make_pair(ID::src, src));
./passes/pmgen/ice40_wrapcarry.cc:148:						lut->attributes.insert(std::make_pair(ID::src, src));
./passes/techmap/shregmap.cc:143:					auto r = sigbit_chain_next.insert(std::make_pair(d_bit, cell));
./passes/techmap/shregmap.cc:157:						sigbit_chain_next.insert(std::make_pair(wire, cell));
./passes/techmap/abc9_ops.cc:50:			auto r = box_lookup.insert(std::make_pair(stringf("$__boxid%d", id), m->name));
./passes/techmap/abc9_ops.cc:143:		auto r = clk_to_mergeability.insert(std::make_pair(abc9_clock, clk_to_mergeability.size() + 1));
./passes/techmap/abc9_ops.cc:180:			replace.insert(std::make_pair(Q,D));
./passes/opt/opt_merge.cc:281:				auto r = sharemap.insert(std::make_pair(hash, cell));
./backends/aiger/xaiger.cc:225:					auto r YS_ATTRIBUTE(unused) = ff_bits.insert(std::make_pair(D, cell));
./frontends/aiger/aigerparse.cc:72:					auto r YS_ATTRIBUTE(unused) = sig2driver.insert(std::make_pair(it2.second, it.second));
```

### DIFF
--- a/kernel/hashlib.h
+++ b/kernel/hashlib.h
@@ -314,11 +314,11 @@ class dict
 	int do_insert(const K &key, int &hash)
 	{
 		if (hashtable.empty()) {
-			entries.push_back(entry_t(std::pair<K, T>(key, T()), -1));
+			entries.emplace_back(std::pair<K, T>(key, T()), -1);
 			do_rehash();
 			hash = do_hash(key);
 		} else {
-			entries.push_back(entry_t(std::pair<K, T>(key, T()), hashtable[hash]));
+			entries.emplace_back(std::pair<K, T>(key, T()), hashtable[hash]);
 			hashtable[hash] = entries.size() - 1;
 		}
 		return entries.size() - 1;
@@ -327,11 +327,25 @@ class dict
 	int do_insert(const std::pair<K, T> &value, int &hash)
 	{
 		if (hashtable.empty()) {
-			entries.push_back(entry_t(value, -1));
+			entries.emplace_back(value, -1);
 			do_rehash();
 			hash = do_hash(value.first);
 		} else {
-			entries.push_back(entry_t(value, hashtable[hash]));
+			entries.emplace_back(value, hashtable[hash]);
+			hashtable[hash] = entries.size() - 1;
+		}
+		return entries.size() - 1;
+	}
+
+	int do_insert(std::pair<K, T> &&rvalue, int &hash)
+	{
+		if (hashtable.empty()) {
+			auto key = rvalue.first;
+			entries.emplace_back(std::forward<std::pair<K, T>>(rvalue), -1);
+			do_rehash();
+			hash = do_hash(key);
+		} else {
+			entries.emplace_back(std::forward<std::pair<K, T>>(rvalue), hashtable[hash]);
 			hashtable[hash] = entries.size() - 1;
 		}
 		return entries.size() - 1;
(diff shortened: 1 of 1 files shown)


kernel/hashlib.h
```diff
@@ -441,6 +455,26 @@ class dict
...
+	{
+		int hash = do_hash(rvalue.first);
+		int i = do_lookup(rvalue.first, hash);
+		if (i >= 0)
+			return std::pair<iterator, bool>(iterator(this, i), false);
+		i = do_insert(std::forward<std::pair<K, T>>(rvalue), hash);
+		return std::pair<iterator, bool>(iterator(this, i), true);
+	}
+
+	std::pair<iterator, bool> insert(K const &key, T &&rvalue)
```
eddiehung, line 468: In keeping with STL semantics, shouldn't this be an `emplace()`?

boqwxp: Maybe... I actually had played with the `emplace()` name originally.  Technically I think a proper `emplace()` also uses some fancy template varargs stuff that I am not familiar with.  Its usage does *look* the same, though.

TBH my main reason for including this specific function wasn't to make a proper `emplace()` but to avoid an ugly explicit `std::make_pair()` in `passes/cmds/scatter.cc`.

If you think this non-template-varargs code is fine to call `emplace()`, then sure, why not?

eddiehung: My reason for wanting `emplace()` is that these datastructures look and feel like STL containers already, it would make sense to continue adopting those semantics. If I did not examine the headers carefully, I would first try and use `emplace()` and see that it doesn't compile, and then switch back to `insert(std::make_pair())` -- I would not expect `insert()` overload to accept two arguments.

boqwxp: So you think it's better just to rename this to `emplace()`, even if I don't implement it "properly" with the varargs template stuff?  OK, no problem.

eddiehung: Hmm. I've never attempted to implement `emplace()` before, but I'm guessing the templated varargs approach would be able to capture lvalue and rvalue without separate overloads?

boqwxp: That's my understanding.

PR #1916, opened 2020-04-14
kernel: Module::makeblackbox() to clear connections too

eddiehung:
Not removing connections can lead to the equivalent of:
`assign 1'bx = 1'bx;`
(since wires have been deleted) which fires an assertion in `Module::check()`

### DIFF
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -767,6 +767,8 @@ void RTLIL::Module::makeblackbox()
 		delete it->second;
 	processes.clear();
 
+	connections_.clear();
+
 	remove(delwires);
 	set_bool_attribute(ID::blackbox);
 }


kernel/rtlil.cc
```diff
@@ -767,7 +761,14 @@ void RTLIL::Module::makeblackbox()
 		delete it->second;
 	processes.clear();
 
+	connections_.clear();
+
+	pool<RTLIL::Wire*> delwires;
+	for (auto it = wires_.begin(); it != wires_.end(); ++it)
+		if (!it->second->port_input && !it->second->port_output)
+			delwires.insert(it->second);
 	remove(delwires);
```
mwkmwkmwk, line 770: Uh, it seems that wires were already removed last, it's just that the list of wires to delete was *collected* before (which shouldn't matter performance-wise).

eddiehung: Ah, that's right. I misread it. Clearing connections fix still stands.

PR #1918, opened 2020-04-14
ast/simplify: improve enum handling

whitequark:
Before this commit, enum values were serialized as attributes of form

    \enum_<width>_<value>

where \<value> was a decimal signed integer.

This has multiple drawbacks:
  * Enums with large values would be hard to process for downstream tooling that cannot easily parse arbitrary precision decimals. (In fact Yosys also did not correctly process enums with large values, and would overflow `int` in that case, which is UB.)
  * Enum value attributes were not confined to their own namespace, making it harder for downstream tooling to enumerate all such attributes, as opposed to looking up any specific value.
  * Enum values could not include `x` or `z`, which are explicitly permitted in the SystemVerilog standard.

After this commit, enum values are serialized as attributes of form

    \enum_value_<value>

where \<value> is a bit sequence of the appropriate width.

cc @RobertBaruch @jjj11x

### DIFF
--- a/README.md
+++ b/README.md
@@ -443,8 +443,8 @@ Verilog Attributes and non-standard features
 - The ``wiretype`` attribute is added by the verilog parser for wires of a
   typedef'd type to indicate the type identifier.
 
-- Various ``enum_{width}_{value}`` attributes are added to wires of an
-  enumerated type to give a map of possible enum items to their values.
+- Various ``enum_value_{value}`` attributes are added to wires of an enumerated type
+  to give a map of possible enum items to their values.
 
 - The ``enum_base_type`` attribute is added to enum items to indicate which
   enum they belong to (enums -- anonymous and otherwise -- are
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -443,8 +443,8 @@ Verilog Attributes and non-standard features
 - The ``wiretype`` attribute is added by the verilog parser for wires of a
   typedef'd type to indicate the type identifier.
 
-- Various ``enum_{width}_{value}`` attributes are added to wires of an
-  enumerated type to give a map of possible enum items to their values.
+- Various ``enum_{value}`` attributes are added to wires of an enumerated type
+  to give a map of possible enum items to their values.
```
cliffordwolf, line 447: This says the attribute name is `enum_{value}` but looking at the code it seems it actually is `enum_value_{value}`.

whitequark: Good catch, thanks!

PR #1923, opened 2020-04-14
abc9_ops: Add an assert ensuring that connected port actually exists.

mwkmwkmwk:


mwkmwkmwk:
(this otherwise blows up with a segmentation fault, as noticed in #1554 development)

eddiehung:
This would turn the segfault into an assertion, but not give any more meaningful information. Is it worth putting something a bit more ~user~ developer friendly here?

mwkmwkmwk:
Changed to a `log_error`.

PR #1925, opened 2020-04-14
[WIP] ecp5: replace ecp5_ffinit with techmap rules + dff2dffs -match-init

eddiehung:
~Driver for this was that I need to do the init-matching before I do `zinit` so that ABC9 can operate on zero-initialised flops.~
A weaker reason would be to get rid of custom C++ passes for things that can be done using `techmap`?

There are a few TODO items that I'm not sure about:
* ~Why is `LSRMUX="INV"` not used?~
* Is is true that ECP5 cannot support FFs with asynchronous set initialised to 0/asynchronous reset initialised to 1?
* When `LSRMODE="PRLD"` (used for latches and SR flops) can it support any initial value? The sim model: https://github.com/YosysHQ/yosys/blob/7025881a5e8c4feb077d296f996683094e2a313a/techlibs/ecp5/cells_sim.v#L319-L326
indicates that the initial value would be the input `M`... Rather conservatively, I only allow `init=1'bx` here.


techlibs/ecp5/cells_map.v
```diff
@@ -1,65 +1,223 @@
...
+module  \$_DFF_xxx_ (input D, C, R, output Q);
+    parameter [0:0] _TECHMAP_WIREINIT_Q_ = 1'bx;
+	parameter _TECHMAP_CELLTYPE_ = "";
+    wire _TECHMAP_REMOVEINIT_Q_ = 1'b1;
+    generate
+        if (_TECHMAP_CELLTYPE_[3*8+:8] == "N")
+            localparam CLKMUX = "INV";
+        else
+            localparam CLKMUX = "CLK";
+        // TODO: Why not use LSRMUX param?
```
daveshah1, line 51: LSRMUX seems to sometimes be broken

eddiehung:
Hmm. Perhaps `dff2dffs -match-init` is unnecessary if I'm doing that (and other init matching) in the techmap file anyway.

ZirconiumX:
> Is it true that ECP5 cannot support FFs with asynchronous set initialised to 0/asynchronous reset initialised to 1?

It turns out that Intel flops have exactly this problem; they don't have async set, and initialise to zero.

eddiehung:
On thinking and testing some more, this always help me. The releavnt, abridged, parts of the ECP5 flow in #1926 is:
```
dff2dffs -match-init
zinit
techmap +/ecp5/cells_map.v
```
Since `dff2dffs -match-init` creates `$__DFFS_*` cells, but `zinit` can't yet handle them. Does that mean this PR isn't actually necessary for what I need?

mwkmwkmwk:
Superseded by #2237.

PR #1927, opened 2020-04-14
kernel: Design::remove(RTLIL::Module *) to check refcount_modules_

eddiehung:
To prevent insertion during iteration; in line with `Design::add()`, `Module::{add,remove}()`, etc.

https://github.com/YosysHQ/yosys/pull/1862#discussion_r406340722

### DIFF
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -597,6 +597,7 @@ void RTLIL::Design::remove(RTLIL::Module *module)
 	}
 
 	log_assert(modules_.at(module->name) == module);
+	log_assert(refcount_modules_ == 0);
 	modules_.erase(module->name);
 	delete module;
 }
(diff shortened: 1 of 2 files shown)


kernel/rtlil.cc
```diff
@@ -597,6 +597,7 @@ void RTLIL::Design::remove(RTLIL::Module *module)
 	}
 
 	log_assert(modules_.at(module->name) == module);
+	log_assert(refcount_modules_ == 0);
```
cliffordwolf, line 600: The whole reason `refcount_modules_` exists is for this assert. Why do we want to allow removing modules while there is an active ObjRange?

ZirconiumX:
This broke `bugpoint`, though I'm pretty sure it has no tests to catch this, since it passed CI.

ZirconiumX:
Perhaps it's better to say that `bugpoint` was already broken, but this exposed that brokenness which didn't get caught by a test.

mwkmwkmwk:
#1978 should fix it.

PR #1936, opened 2020-04-14
opt_expr: Fix X and CO outputs for $alu identity-mapping rules.

mwkmwkmwk:


tests/opt/opt_expr_alu.ys
```diff
@@ -60,7 +60,8 @@ alumacc
 equiv_opt -assert opt_expr -fine
 design -load postopt
 select -assert-count 1 t:$pos
-select -assert-count none t:$pos t:* %D
+select -assert-count 1 t:$not
+select -assert-count none t:$pos t:$not %% t:* %D
```
ZirconiumX, line 64: Wouldn't it be slightly terser to use `-assert-none` here?

passes/opt/opt_expr.cc
```diff
@@ -1135,9 +1135,24 @@ void replace_const_cells(RTLIL::Design *design, RTLIL::Module *module, bool cons
...
 
 				if (cell->type == ID($alu)) {
+					bool a_signed = cell->parameters[ID::A_SIGNED].as_bool();
+					bool b_signed = cell->parameters[ID::B_SIGNED].as_bool();
+					bool is_signed = a_signed && b_signed;
+					RTLIL::SigBit sig_ci = assign_map(cell->getPort(ID::CI));
 					int y_width = GetSize(cell->getPort(ID::Y));
-					module->connect(cell->getPort(ID::X), RTLIL::Const(State::S0, y_width));
-					module->connect(cell->getPort(ID::CO), RTLIL::Const(State::S0, y_width));
+					if (sig_ci.data == State::S1) {
```
eddiehung, line 1143:
```suggestion
					if (sig_ci == State::S1) {
```

tests/opt/opt_expr_alu.ys
eddiehung: Actually, `-assert-count none` relies on `none` being interpreted as `0`:
https://github.com/YosysHQ/yosys/blob/3c4758c60e33386e19049d81fd6f72c8e7f316e0/passes/cmds/select.cc#L1265-L1268
Sorry, another issue from copy-pasting from my tests.
So either `-assert-count 0` for spacing, or `-assert-none` as @ZirconiumX suggests.

PR #1938, opened 2020-04-14
kernel: Trap in `log_error()` when a debugger is attached.

boqwxp:
This PR adds a multi-platform debug break facility to enable `log_error()` to instead issue a debug trap instead of simply exiting.
However, when no debugger is attached, the behavior remains unchanged.

See: https://stackoverflow.com/questions/3596781/how-to-detect-if-the-current-process-is-being-run-by-gdb (baseip's answer)
and: https://github.com/scottt/debugbreak

Note that the debugbreak code above upon which this is partially based is licensed under BSD 2-clause.  I think that should be compatible with the ISC license in the rest of Yosys, but I may be wrong.

This has only been tested on Linux so far.  Testing on Windows with MSVC and MingW should be done before this is merged.

### DIFF
--- a/kernel/log.h
+++ b/kernel/log.h
@@ -53,6 +53,8 @@
 #ifndef _WIN32
 #  include <sys/time.h>
 #  include <sys/resource.h>
+#  include <sys/ptrace.h>
+#  include <signal.h>
 #endif
 
 #if defined(_MSC_VER)
@@ -68,6 +70,22 @@ YOSYS_NAMESPACE_BEGIN
 #define S__LINE__sub2(x) #x
 #define S__LINE__sub1(x) S__LINE__sub2(x)
 #define S__LINE__ S__LINE__sub1(__LINE__)
+
+#if defined(_MSC_VER)
+# define YS_DEBUGTRAP __debugbreak()
+#elif defined(__unix__)
+# if defined(__clang__)
+#  if defined(__builtin_debugtrap)
+#   define YS_DEBUGTRAP __builtin_debugtrap()
+#  else
+#   define YS_DEBUGTRAP do{if(ptrace(PTRACE_TRACEME, 0, nullptr, 0) == -1){ raise(SIGTRAP); }} while(0);
+#  endif
+# else
+#  define YS_DEBUGTRAP do{if(ptrace(PTRACE_TRACEME, 0, nullptr, 0) == -1){ raise(SIGTRAP); }} while(0);
+# endif
+#else
+# define YS_DEBUGTRAP do{} while(0);
+#endif
 
 struct log_cmd_error_exception { };
 
--- a/kernel/log.cc
+++ b/kernel/log.cc
@@ -354,6 +354,9 @@ static void logv_error_with_prefix(const char *prefix,
 
 	if (check_expected_logs)
 		log_check_expected();
+
+	YS_DEBUGTRAP; //trap into debugger if one is attached
+
 #ifdef EMSCRIPTEN
 	log_files = backup_log_files;
 	throw 0;


whitequark:
I am strongly opposed to this change. A HDL toolchain is not a place to add platform-specific inline assembly on rarely used code paths, and whichever improvement in convenience this brings will be compensated by the added portability and reliability issues.

whitequark:
Also, it is not even clear to me why this level of effort is necessary. What is wrong with:

```c
void debug_die()
{
#if defined(EMSCRIPTEN)
  throw 0;
#elif defined(_WIN32)
  if (IsDebuggerPresent())
    DebugBreak();
#elif defined(__unix__)
  if (ptrace(PTRACE_TRACEME, 0, NULL, 0) == -1)
    raise(SIGTRAP);
#endif
  _exit(1);
}
```

boqwxp:
Well, there is platform-specific inline assembly, but for quite a lot of platforms ;).  And there is a generic POSIX-compliant fallback.  I haven't done extensive testing on multiple platforms yet, but considering the popularity of the base code I think it will probably work just fine.

Besides, this inline assembly is *only* called when we want to either die (original functionality) or trap into a debugger (ideally).  What's the worst that can happen--a totally invalid instruction is emitted?  Then we die, just as we are currently supposed to, albeit in a different and slightly less elegant way.

Besides, the convenience factor is huge, at least for me.  Without this existing in the code base, I frequently finding myself modifying `kernel/log.h` to put a `__builtin_trap()`, which necessitates a recompilation of *everything* that on my computer takes ~20 min.

Manually setting breakpoints is another option, but when debugging during development, one often doesn't know all the points that assertions might fail or `log_error()` might be triggered, making it a tedious whack-a-mole to get a debugger attached with the state one wants to inspect--this all the more so if the errors are triggered after a long elaboration process.

whitequark:
> Well, there is platform-specific inline assembly, but for quite a lot of platforms ;)

There is no reason for a HDL toolchain--a tool that, fundamentally, transforms byte sequences to byte sequences--to ever use anything but portable, platform-agnostic C/C++ code.

> Manually setting breakpoints is another option, but when debugging during development, one often doesn't know all the points that assertions might fail or `log_error()` might be triggered, making it a tedious whack-a-mole to get a debugger attached with the state one wants to inspect--this all the more so if the errors are triggered after a long elaboration process.

If you care primarily about gdb, you could use [the `.gdb_debug_scripts` section](https://sourceware.org/gdb/onlinedocs/gdb/dotdebug_005fgdb_005fscripts-section.htmll) to get the same results without any of the invasive changes.

boqwxp:
>      There is no reason for a HDL toolchain--a tool that, fundamentally, transforms byte sequences to byte sequences--to ever use anything but portable, platform-agnostic C/C++ code.

That's an assertion, not really an argument.  Also keep in mind there is a portable, platform-agnostic fallback.  These are just specializations.

I may be wrong, but I believe the purpose of the inline assembly is to act as a code motion barrier.  Maybe @scottt can give some insight here?  This is currently pointless for our purposes, but in the future someone else may want to make `debug_break()` non-static and break in the middle of some code path rather than just before dying.

To be clear I am not necessarily opposed to doing it your way.  It's a lot of boilerplate and preprocessor grossness, and sure, it might be overkill for now.  But I don't share your extreme aversion to a few platform-specific code paths with one inline assembly instruction each.

whitequark:
> It's a lot of boilerplate and preprocessor grossness, and sure, it might be overkill for now.

It is a lot less boilerplate than your 250 line patch.

boqwxp:
> It is a lot less boilerplate than your 250 line patch.

? That's what I was talking about...

whitequark:
> Also keep in mind there is a portable, platform-agnostic fallback.

If you remove everything but the fallback code then this PR would look a whole lot better than it does right now.

whitequark:
> ? That's what I was talking about...

Ah sorry, I misunderstood what you meant.

boqwxp:
I'll leave it to someone else to decide.  I don't really have a strong preference either way, so if the purpose of the inline assembly is just to prevent code motion optimization (would be great to hear from @scottt on this) and other people prefer your way or have no preference, I will eliminate the inline assembly and simplify the PR.

whitequark:
We discussed this on IRC and came to agreement on the way forward on this PR.

boqwxp:
The current state of this code works fine, but when a debugger is not attached on Linux/BSD, it aborts rather than exiting gracefully.

Unfortunately, @whitequark refuses to cooperate with me on this after I apparently upset them by saying, "hey dude", and then apparently patronized them by quoting Wikipedia to point out that it's been a unisex term for decades.

This experience has left a bad taste in my mouth and I'm no longer interested in working on this PR.  Feel free to close this or assign it to anyone willing to finish it.

nakengelhardt:
@boqwxp that was staggeringly awful of you. I find I am suddenly not very interested in cooperating with you either.

boqwxp:
@nakengelhardt Please help me to understand how it was staggeringly awful.

I am a native speaker.  "Dude" is a gender-neutral term in common use by males and females to refer both to other males and females.  Addressing an ambiguously-gendered ("whitequark" gives me no information about their gender) person with "dude", while informal, does not ascribe a gender.

I'm genuinely confused why y'all have such extreme reactions to a totally banal use of language.  I mean, even if it did imply the speaker assumes the subject is a male--again, it does not--that seems a minor offense and nowhere approaching "staggeringly awful".

nakengelhardt:
It's the bit where you double down on addressing someone in a manner they have told you they don't want to be addressed, and trying to justify it as somehow the objectively right choice, instead of just respecting that preference. If you'd just said "oh sorry" and used "hey there" next time I wouldn't have a problem with it. (And "is dude gender neutral" has been hotly debated for many years now, it's by no means as clear cut as you say.)

Also, I was just about to edit the previous comment because I wanted to disambiguate that the close is unrelated to the comment; IIRC in last week's call everyone else was in favor of closing the PR anyway but we were leaving it to whitequark as she seemed to have an interest in it.

boqwxp:
You're right, I should have just said sorry and moved on.  I made the mistake of assuming she was not a native speaker and trying to correct a potential misunderstanding about the word.

>  "is dude gender neutral" has been hotly debated for many years now
Maybe academically, like "is singular 'they' grammatical?" but not in reality.  I hear it all the time in the community used by female speakers to refer to people of either gender.  In any case, if someone has a particular aversion to being called "dude"--as is now clear to me, but was not at all when I "patronized" her--I will not address them so.

I still don't understand how it's staggeringly awful.  I'm mystified about the incredible amount of drama here.  I guess I'll just leave it at this: If you feel uncomfortable working with me for any reason, then by all means, please don't!

nakengelhardt:
It's not academical: as you are currently discovering, many people do in fact carry bad memories of past fights on the topic which lead to heightened reactions. Which "community" you are a part of will affect your experiences a lot. I'll downgrade my reaction to "that was pretty insensitive" now that I've calmed down again.

whitequark:
> You're right, I should have just said sorry and moved on.

@boqwxp As a matter of fact, if you did that, I would have put that unfortunate exchange behind us and moved on to what I'd rather do, which is improve Yosys. Frankly, I did not even expect you to apologize, just to not double down immediately. I hope you see now that, just like how my actions left a bad taste in your mouth, yours did not exactly brighten my day either.

> IIRC in last week's call everyone else was in favor of closing the PR anyway but we were leaving it to whitequark as she seemed to have an interest in it

@nakengelhardt Speaking on purely technical grounds, I think the approach of having a debug trap in `log_error()` would be a noticeable improvement in ergonomics that I would certainly like to have, and I believe the remaining minor technical issue with the implementation in this PR could be easily resolved.

nakengelhardt:
There was also the issue of making sure that the license of the included code is compatible with the yosys license, which I'm not sure if we figured out conclusively.

nakengelhardt:
I guess I can leave it open until Thursday and bring it up again to see if anyone else is interested in taking over.

whitequark:
> There was also the issue of making sure that the license of the included code is compatible with the yosys license, which I'm not sure if we figured out conclusively.

This PR was completely rewritten; it doesn't include any third-party code now, and is also much simpler.

> I guess I can leave it open until Thursday and bring it up again to see if anyone else is interested in taking over.

If necessary I will do that.

nakengelhardt:
Oh sorry, I missed that. In that case opinions on last week's code are irrelevant. I'll add it to the agenda.

kernel/log.h
```diff
@@ -69,6 +71,22 @@ YOSYS_NAMESPACE_BEGIN
...
 #define S__LINE__ S__LINE__sub1(__LINE__)
 
+#if defined(_MSC_VER)
+# define YS_DEBUGTRAP __debugbreak()
+#elif defined(__unix__)
+# if defined(__clang__)
+#  if defined(__builtin_debugtrap)
+#   define YS_DEBUGTRAP __builtin_debugtrap()
+#  else
+#   define YS_DEBUGTRAP do{if(ptrace(PTRACE_TRACEME, 0, nullptr, 0) == -1){ raise(SIGTRAP); }} while(0);
```
whitequark, line 81: Style nit: if the definitions that expand into functions don't end with a `;`, then the defines that expand into do-while loops shouldn't either.

whitequark:
I've completely reimplemented this PR in #2000. I believe #2000 addresses concerns raised by everyone so far. And it turned out it's possible to do it in a way that's both more portable *and* more reliable than the one I suggested in [this comment](https://github.com/YosysHQ/yosys/pull/1938#issuecomment-613839683).

PR #1939, opened 2020-04-15
Use const auto& in range loops to avoid copies.

hzeller:
This is some low hanging performance fruit, avoiding
copies of non-trivial objects when not needed.
Consistently applying it to the code-base also provides
some good practice template for new code.

This was automatically found using clang-tidy; it only
suggests this for non-trivial types when only const methods
are accessed.

https://clang.llvm.org/extra/clang-tidy/checks/performance-for-range-copy.html

Tests pass, and the whole 'make test' runs about 3-4% faster on my machine.
(might well be in the noise, but there is no degradation).

Signed-off-by: Henner Zeller <h.zeller@acm.org>


backends/aiger/aiger.cc
```diff
@@ -161,10 +161,10 @@ struct AigerWriter
 			}
 		}
 
-		for (auto bit : input_bits)
+		for (const auto& bit : input_bits)
```
eddiehung, line 164: Yosys style is (mostly) `const auto &bit` -- can you adopt that instead please? 
I'm also surprised that the compiler can't do this optimisation automatically.

hzeller: done.

cliffordwolf:
Most of those seem to be about `SigBit` iterators. Do we have any data that suggests that using const references here is actually faster than value copies?

Kmanfi:
FYI. #1942

hzeller:
Claire, here is a godbolt example comparing a reduced example of a SigBit with the two loop constructs, compiled with the latest gcc with `-O3`. Here, it saves two instructions. So even with this small struct, it can save.

https://godbolt.org/z/zhe4tH   `for (SigBit sb : container)`

https://godbolt.org/z/ywFsbZ  `for (const SigBit &sb : container)`

cliffordwolf:
> Here, it saves two instructions. So even with this small struct, it can save.

My main concern is aliasing. Can the compiler see that this `const SigBit &sb` reference isn't aliasing any other sigbit that's also in scope at the same time, and are we thus forcing the compiler to re-fetch `sb` whenever another SigBit is written in the same code?

The main motivation behind the design of SigBit was to have something small enough so that we can copy it by value and don't need to think about pointer aliasing when writing code..

Your code completely avoids this question because in your code the `<` is just an external reference, so we never see anything being done with those objects. We never see if the compiler would create redundant loads when inlining that function. (In that case it wouldn't because this object is the only SigBit around. And that's exactly my point. You need to look at an example where pointer aliasing may become an issue.)

So consider this example instead: https://godbolt.org/z/jGsmub

Here `maxval = a` is being compiled into the following asm code:

```
        mov     rdx, QWORD PTR [rax]
        mov     QWORD PTR [rdi], rdx
        mov     edx, DWORD PTR [rax+8]
        mov     DWORD PTR [rdi+8], edx
        mov     rdx, QWORD PTR [rax]
```

However, if we change `auto &a : A` to `auto a : A` then it's only two instructions:

```
        mov     QWORD PTR [rdi], rax
        mov     DWORD PTR [rdi+8], ecx
```

We end up with the additional instructions in the first case because we have to re-load the data from `a` just in case `a` happens to alias `maxval`.

Finally, instruction count is a poor measure of performance, especially on x86. And `-O3` does not exactly optimize for a small number of instructions anyway. So when I said "Do we have any data that suggests that using const references here is actually faster than value copies?" I meant benchmark results on the actual code base, not instruction count on a made-up code snippet.

hzeller:
The benchmark results were that `make test` was running about 3-4% faster with this change, tested over 5 runs.
But this is overall, not specifically targeting a particular the SigBit struct. Is there a particular benchmark that you'd like me to run ?

hzeller:
I took the example above and made a little benchmark out of it; attached. For some reason github does not like *.cc files, so it is gzipped.
[sigbit-loop-bench.cc.gz](https://github.com/YosysHQ/yosys/files/4520702/sigbit-loop-bench.cc.gz)

The relevant functions are
```c++
// Updating an out-param from a copy.
void CopySetOutparam(std::vector<SigBit> &A, SigBit &maxval, SigBit &minval){
    for (const auto a : A)
    {
        if (maxval < a)
            maxval = a;
        if (a < minval)
            minval = a;
    }
}

// This function has to worry about aliasing.
void RefSetOutparam(std::vector<SigBit> &A, SigBit &maxval, SigBit &minval){
    for (const auto &a : A)
    {
        if (maxval < a)
            maxval = a;
        if (a < minval)
            minval = a;
    }
}

// Updating an out-param after local accumulation. Using copy-loop.
void CopyLocalUpdate(std::vector<SigBit> &A, SigBit &maxval, SigBit &minval) {
    std::pair<SigBit, SigBit> result = { minval, maxval };
    for (const auto a : A)
    {
        if (result.second < a)
            result.second = a;
        if (a < result.first)
            result.first = a;
    }
    minval = result.first;
    maxval = result.second;
}

// Updating an out-param after local accumulation. Using ref-loop.
void RefLocalUpdate(std::vector<SigBit> &A, SigBit &maxval, SigBit &minval) {
    std::pair<SigBit, SigBit> result = { minval, maxval };
    for (const auto &a : A)
    {
        if (result.second < a)
            result.second = a;
        if (a < result.first)
            result.first = a;
    }
    minval = result.first;
    maxval = result.second;
}
```

Results
```
$ for CXX in g++ clang++ ; do echo $CXX ; $CXX -O3 sigbit-loop-bench.cc -o sigbit-loop-bench && for i in {1..10} ; do ./sigbit-loop-bench  ; done ; done
g++
CopySetOutparam:16753.4   RefSetOutparam:15507      CopyLocalUpdate:30266.7     RefLocalUpdate:15139.6
CopySetOutparam:16662.2   RefSetOutparam:15460.1    CopyLocalUpdate:29739.9     RefLocalUpdate:14896.9
CopySetOutparam:17365.8   RefSetOutparam:15806.7    CopyLocalUpdate:30831.8     RefLocalUpdate:15249.2
CopySetOutparam:17328.3   RefSetOutparam:15838.7    CopyLocalUpdate:30689.7     RefLocalUpdate:15269.9
CopySetOutparam:17323.1   RefSetOutparam:15789.7    CopyLocalUpdate:30740.5     RefLocalUpdate:15238.4
CopySetOutparam:16730.3   RefSetOutparam:15710.2    CopyLocalUpdate:29864.9     RefLocalUpdate:14958.2
CopySetOutparam:16871.9   RefSetOutparam:15661.2    CopyLocalUpdate:29995.8     RefLocalUpdate:14935.2
CopySetOutparam:17422     RefSetOutparam:15822.3    CopyLocalUpdate:30737.9     RefLocalUpdate:15274.6
CopySetOutparam:17203.7   RefSetOutparam:15676.5    CopyLocalUpdate:30803.9     RefLocalUpdate:15168.7
CopySetOutparam:16986.6   RefSetOutparam:15473.1    CopyLocalUpdate:29682.6     RefLocalUpdate:14887.5
clang++
CopySetOutparam:14881.6   RefSetOutparam:15034.4    CopyLocalUpdate:17475.1     RefLocalUpdate:16294.3
CopySetOutparam:15381.9   RefSetOutparam:15437.3    CopyLocalUpdate:18017.4     RefLocalUpdate:17118.5
CopySetOutparam:15275     RefSetOutparam:15330.3    CopyLocalUpdate:17818.7     RefLocalUpdate:16629
CopySetOutparam:14930.7   RefSetOutparam:15036.1    CopyLocalUpdate:17445.2     RefLocalUpdate:16346.7
CopySetOutparam:15309.6   RefSetOutparam:15364.5    CopyLocalUpdate:17857.5     RefLocalUpdate:16658.1
CopySetOutparam:15317.8   RefSetOutparam:15531.4    CopyLocalUpdate:18071.7     RefLocalUpdate:16721.5
CopySetOutparam:14897.3   RefSetOutparam:14974.8    CopyLocalUpdate:17404.5     RefLocalUpdate:16292.7
CopySetOutparam:15379.9   RefSetOutparam:15422.1    CopyLocalUpdate:17903.5     RefLocalUpdate:16730.2
CopySetOutparam:15514.4   RefSetOutparam:15432.6    CopyLocalUpdate:17938.8     RefLocalUpdate:16717.3
CopySetOutparam:15312.9   RefSetOutparam:15394.4    CopyLocalUpdate:17923.8     RefLocalUpdate:16740
```

Findings (on mobile Intel(R) Core(TM) i7-7500U CPU @ 2.70GHz )

  * g++ (version 9.3. 0) -O3
    * SetOutparam: Copy _slower_ than Ref 
    * Local Update: Copy _much slower_ than Ref
 * clang++ (version 9.0.1-12)  -O3
   * SetOutparam: Copy slightly _faster_ than Ref (by a very small margin)
   * LocalUpdate: Copy _slower_ than Ref.

Copy typically is slower, with gcc in some case generating code that uses twice the processing time than using a reference.
Copy is very slightly faster in the aliasing case for clang++ but by a very small margin.

Given the numbers above, I'd say using const reference instead of a copy in the range loop is a net positive.

backends/aiger/xaiger.cc
```diff
@@ -355,7 +355,7 @@ struct XAigerWriter
 				}
 			}
 
-			for (auto port_name : r.first->second) {
+			for (const auto &port_name : r.first->second) {
```
cliffordwolf, line 358: This is literally creating a pointer to an int instead of copying the int by value. That can't possibly be an improvement. There are a few more places in this patch where this is done.

frontends/rpc/rpc_frontend.cc
```diff
@@ -209,7 +209,7 @@ struct RpcModule : RTLIL::Module {
 					if (name_mangling.count(cell->type.str()))
 						cell->type = name_mangling[cell->type.str()];
 
-			for (auto module : derived_design->modules_) {
+			for (const auto &module : derived_design->modules_) {
```
cliffordwolf, line 212: This creates a reference to a pair containing an IdString (just and int) and a pointer. The code in this block is only using module.first (the int) and module.second (the pointer) directly, never the whole pair. In what way is this code improved by creating a reference to that pair instead of loading this two members right here directly?

kernel/rtlil.cc
```diff
@@ -382,27 +382,27 @@ void RTLIL::Selection::optimize(RTLIL::Design *design)
 	std::vector<RTLIL::IdString> del_list, add_list;
 
 	del_list.clear();
-	for (auto mod_name : selected_modules) {
+	for (const auto &mod_name : selected_modules) {
```
cliffordwolf, line 385: These and the ones below are all references to integers. How is it an improvement to keep a 64-bit pointer around that has to be dereferenced every time it's used instead of just copying the singe 32-bit integer that is being referenced?

passes/cmds/splitnets.cc
```diff
@@ -226,7 +226,7 @@ struct SplitnetsPass : public Pass {
 
 					vector<IdString> &new_ports = module_ports_db[module->name][wire->name];
 
-					for (SigSpec c : sig.chunks())
+					for (const SigSpec &c : sig.chunks())
```
cliffordwolf, line 229: So `sig.chunks()` returns a reference to a `std::vector<RTLIL::SigChunk>` we then iterate over that list and convert each element into a `SigSpec`. In what way is it an improvement to then create a reference to that freshly created SigSpec? This is not avoiding any copy, it's just adding an additional indirection.

passes/equiv/equiv_simple.cc
```diff
@@ -374,7 +374,7 @@ struct EquivSimplePass : public Pass {
 				it.second.sort();
 
 				vector<Cell*> cells;
-				for (auto it2 : it.second)
+				for (const auto &it2 : it.second)
```
cliffordwolf, line 377: This creates a reference to a pointer. How can that possibly be better than just copying the pointer?

passes/opt/rmports.cc
```diff
@@ -74,7 +74,7 @@ struct RmportsPassPass : public Pass {
 			}
 
 			auto ports_to_remove = removed_ports[cell->type];
-			for(auto p : ports_to_remove)
+			for(const auto &p : ports_to_remove)
```
cliffordwolf, line 77: That's a funny one! `auto ports_to_remove` obviously should be `const auto ports_to_remove&`, but isn't. But instead of fixing that we still copy the whole container of ints, and then patch the code so it creates a reference to each of the individual ints. Why?

cliffordwolf:
> The benchmark results were that make test was running about 3-4% faster with this change, tested over 5 runs.

I can not reproduce this result.

I have taken your patch and stripped all changes that relate to SigBit or IdString, and then ran `make test` 10x for the unmodified Yosys master, the stripped version of your patch, and the full version of your patch. Then I calculated mean and std deviation for those three sets of 10 data points each. If I plot the corresponding gauss curves (normalized so that they all max at 1):

image

image

According to the (meaningless) data I collected my stripped version of the patch is about 1% faster than the master branch, both in terms of wall clock time and CPU time.

The unstripped version of the patch performs 0.05% faster than master in terms of CPU time and 1.5% worse than master in terms of wall clock time.

I think the main takeaway is that `make test` is a very poor benchmark. The variance is just too large to get any meaningful results from just 10 (or even 5) data points.

Here are my files from this experiment: 
[hzeller-performance-eval.zip](https://github.com/YosysHQ/yosys/files/4531754/hzeller-performance-eval.zip)

I ran my tests on a Scaleway GP1-XS quad core instance running Ubuntu 18.04.2 LTS with Clang 6.0.0-1ubuntu2.

backends/aiger/xaiger.cc
hzeller: If it is a `const integer &`,  the compiler automatically choose to do the right thing choosing a copy or reference (when it doesn't have to worry about aliasing).

I see how this might not _look_ optimal, but there is no difference in what the compiler generates here ( Play with the reference `&` here: https://godbolt.org/z/v-Huvy ).

As someone reading the code, it is a readability improvement because they don't have to worry about if this is an potentially expensive copy or not, as the compiler will choose the fastest way with the const reference, but might not have a chance to do the fastest when it is a copy.

frontends/rpc/rpc_frontend.cc
hzeller: Similar here, the code output in this case is the same https://godbolt.org/z/FQeBge

kernel/rtlil.cc
hzeller: it is not doing that, it actually will store it locally and make not any difference in the code output (similar to https://godbolt.org/z/v-Huvy ).
Here the `const auto &` merely serves as readability improvement for the programmer when scanning the code, because it makes sure there is no _expensive_ copy.

passes/cmds/splitnets.cc
hzeller: Again, in this case there will not be any change in the code output (it is like assigning a local variable to a const reference). Here as well, this will only be for readability improvement.

passes/equiv/equiv_simple.cc
hzeller: Same thing, it will collapse to just using the pointer directly.

passes/opt/rmports.cc
hzeller: Oh, the reason why the `auto ports_to_remove` is not constified is because this automated change ( https://clang.llvm.org/extra/clang-tidy/checks/performance-for-range-copy.html ) is only concerned about range copies, not where the inputs are coming from. Agreed, that this would probably be a good idea :)

The `const auto &p` vs. `const auto p` will make a difference here if p is a string.

hzeller:
Thanks for your very thorough review, Claire!

I think a lot of the places you came across (reference to int or pointer instead of the thing itself) are actually dealt with by the compiler in the way we'd expect: it would not create an extra dereference so the `for (const int a : A)` vs. `for (const int &a : A)` are identical. So as a developer, we don't have to worry if it will take a reference or a copy, it will do the fastest thing in the context. But if we don't use a reference, it will _always_ create a copy, and that might not be optimal.

I see that the automated patch does not properly recognize opportunities with avoiding aliasing (the earlier discussion), the other very simple integer loops are not making a difference, as the compiler has to 'flatten' the reference to a value anyway, and any optimization it does will end up with the same code.

In general, i'd see this as an improvement of readability and improvement for someone just reading a snippet of code, because it conveys that a particular loop does not do potentially expensive copies (in particular if it is not clear what `auto` might be in that context), and that any use of _not_ `const auto &` would be a deliberate choice. Any 'overzealous' use of `const auto &` does not harm either _except_ potentially in the case of aliasing.

Should we 
  1. go ahead with this patch but do the stripping you propose (I do think it improves readability even if performance equal) or
  2. don't merge this now and wait for the clang-tidy thing to be a bit smarter detecting potential aliasing situations ? Which I agree is potentially a downside of this (though overall I personally still think the overall benefits outweigh this possibly rare occurence).

(We should find some large beefy Verilog as benchmark that exercises a lot of Yosys internals to get a better feel for performance suggestions; though in this case, I also strongly favor the readability improvement of this patch).

backends/aiger/xaiger.cc
cliffordwolf: The compilerexplorer code you have posted here is in no way representative for the code in question. The code in question contains external function calls interleaved with uses of `port_name`, thus the compiler has to re-fetch the value of `port_name` multiple times.

Please stop posting these misleading code snippets that seem to almost intentionally avoid the specific issue I've been pointing out numerous times now in the code in question.

frontends/rpc/rpc_frontend.cc
cliffordwolf: Same as before. `module.first` is being used, then external functions are called that the compiler has no visibility into, then `module.first` is used again, forcing the compiler to re-load the values from memory.

Your compiler explorer example is carefully avoiding doing exactly that. Thus it has no relevance whatsoever for explaining what will happen with this code.

kernel/rtlil.cc
cliffordwolf: Only if the compiler has perfect visibility into the `count()`, `push_back()`, and `erase()` methods of the containers in question _and_ can use it to prove that there is no aliasing happening here, which may or may not be the case, but your compiler explorer code example does nothing to answer that question.

cliffordwolf:
> I think a lot of the places you came across (reference to int or pointer instead of the thing itself) are actually dealt with by the compiler in the way we'd expect: it would not create an extra dereference so the for (const int a : A) vs. for (const int &a : A) are identical.

This is incorrect. Whenever any int in memory is being overwritten, `a` has be loaded from memory again in the reference case if the compiler can't prove that `a` can be a reference to that memory location. Whenever an external function is called, `a` has to be loaded from memory again because the compiler doesn't know if that external function modified the memory location `a` is a reference to.

See this simple example: https://godbolt.org/z/wHPNTG

The two additional `mov edi, DWORD PTR [rbx-4]` instructions are the compiler fetching `a` from memory again and again. Without the reference, `a` is only fetched once and kept in a register for the entire loop body.

Moreover, code like this prevents reordering and other optimizations on `a` that might otherwise be possible.

> Should we [..]

This PR claims to be a performance improvement and proposes to change the Yosys coding style of consistently use value copies for SigBit and IdString.

If this is a performance improvement then there must be a benchmark than shows the improved performance.

I have some ideas for how to set up such a benchmark, but I'm pretty busy right now so I'm not sure when I will get to it.

backends/aiger/xaiger.cc
hzeller: To be fair, your immediate raised concern in this example was that this was using a const reference to the int-wrapping `IdString`; so I responded directly to that and wanted to show in a reduced example that this is not typically a concern: the compiler will not necessarily dereference a const ref every time.

In generalization, when we can use a const reference, the compiler will have a chance to do this optimization, but if we force it to do a copy, the compiler might not have a chance to do that in case as it will have to call an expensive copy constructor. But you are also right that this requires enough visibility for the compiler to know that it is safe.

This very loop is actually a good example: in its current implementation, the `IdString` is not just a simple integer, there is also some global reference counting going on in the constructor/destructor (unless `YOSYS_NO_IDS_REFCNT` is defined (which it is not in the default compilation)). So in this loop, with the forced copy, there is actually one more instance of incrementing/decrementing the reference going on, making this loop slower with the copy vs. the const reference.

Since you're skeptical of my potentially biased godbolt-reductions, we can observe this directly in the actual piece of compiled code; do the following once with the stock xaiger.cc, then edit line 358 to use a const reference: `for (const auto &port_name : r.first->second)`

This command-line then would cut out the approximate area of code we're interested in:
```
make && objdump -l -C -d -S ./backends/aiger/xaiger.o | awk '/xaiger.cc:358/ { do_print=1; }  { if (do_print) print $0; } /xaiger.cc:388/ { do_print=0; }' > xaiger-loop.1
```
With a stock compilation with clang 8.0 this shows that there is at least one round more of reference counting going on. In that case the reference version clearly has an advantage.

This is an example where an unassumingly expensive constructor makes the difference that this pull request intends to address: avoiding these copies so that the compile has a chance to generate more optimal code.

If `YOSYS_NO_IDS_REFCNT` is defined and `IdString` is actually boiling down to an integer, the code is similar but requires a few more indirections (but I am not sure if in that case the clang-tidy would've suggested it).

frontends/rpc/rpc_frontend.cc
hzeller: Looking at the actual object file we also see more activity with the code that copies vs. the const reference due to the reference counting business:

(compare this vs. changing line 212 with `for (const auto &module : ...`)
```
make && objdump -l -C -d -S ./frontends/rpc/rpc_frontend.o | awk '/rpc_frontend.cc:212/ { do_print=1; }  { if (do_print) print $0; } /rpc_frontend.cc:226/ { do_print=0; }' > rpcfrontend.1
```

kernel/rtlil.cc
hzeller: That is true, in case of aliasing, the compiler might have to generate more expensive defensive code.

Here, we see a significant shorter code generated with the const reference; with the copy, there is our friend, the reference counter in and out of `IdString()/~IdString()`
```
make && objdump -l -C -d -S ./kernel/rtlil.o | awk '/rtlil.cc:385$/ { do_print=1; }  { if (do_print) print $0; } /rtlil.cc:390$/ { do_print=0; }' > rtlil.1
```

hzeller:
If you have some ideas on such a benchmark I could work on it as well, but I understand if just explaining it could eat more time for something not very important.
Since this is generally a minor change, we can as well let it rest and revisit later ? I can generate a pull request at that time.

backends/aiger/xaiger.cc
cliffordwolf: >  there is also some global reference counting going on

which, as you also point out, can be turned off using `YOSYS_NO_IDS_REFCNT`, which is what should be done if one is concerned about ref count performance. thus a change that improves performance for enabled reference counting, but worsens it with disabled reference counting, is counterproductive, because it worsens performance for those who care the most.

hzeller:
Closing this now to be re-visited later when clang-tidy can take potential aliasing impact into account and we have a benchmark to provide solid data for a decision.

PR #1945, opened 2020-04-16
Add parameter positional and default value information.

mwkmwkmwk:


kernel/rtlil.cc
```diff
@@ -2612,7 +2613,16 @@ void RTLIL::Cell::setParam(RTLIL::IdString paramname, RTLIL::Const value)
 
 const RTLIL::Const &RTLIL::Cell::getParam(RTLIL::IdString paramname) const
 {
-	return parameters.at(paramname);
+	static const RTLIL::Const empty;
```
eddiehung, line 2616: Is it a concern that `getParam()` would previously fail with an exception, and now silently returns an empty constant?

mwkmwkmwk: I don't know, do you think it is?

AFAICT no place in yosys actually relies on an exception being raised here (there's barly any try..catch in yosys code, as there should be).  For internal cells, the checks still validate the parameter presence, and for vendor cells, this function will actually return the default value from `cells_sim.v` etc files, which should be reasonable.

eddiehung: I think it is of value for developers, and their fat fingers. This realigns it with `getPort` too behaviour too.

mwkmwkmwk: So, what would you propose? Raising an exception if the parameter doesn't exist and it has no default value?

eddiehung: Yes, that's what I would revert it to in a new PR unless you object?

mwkmwkmwk: OK, sounds good

eddiehung: #1984

PR #1946, opened 2020-04-16
abc: use YosysHQ/abc instead of upstream berkeley-abc/abc

eddiehung:
Enabling modifications.

Currently, the only modification is YosysHQ/abc#1 which is a proposed fix for Yosys' `abc9 -dff` operating on multi clock designs.


daveshah1:
Need to remove the "unmodified" here too: https://github.com/YosysHQ/yosys/blob/master/manual/CHAPTER_Auxprogs.tex#L22

PR #1949, opened 2020-04-16
select: do not select inside black-/white- boxes unless '=' prefix used

eddiehung:
Without which `select c:*` will select specify cells inside blackboxes, or anything inside a whitebox.

### DIFF
--- a/passes/cmds/select.cc
+++ b/passes/cmds/select.cc
@@ -1113,8 +1113,8 @@ struct SelectPass : public Pass {
 		log("    <obj_pattern>\n");
 		log("        select the specified object(s) from the current module\n");
 		log("\n");
-		log("Prefix the following patterns with '=' if the pattern should match black-/\n");
-		log("white-box modules and their contents.\n");
+		log("By default, patterns will not match black/white-box modules or their");
+		log("contents. To include such objects, prefix the pattern with '='.\n");
 		log("\n");
 		log("A <mod_pattern> can be a module name, wildcard expression (*, ?, [..])\n");
 		log("matching module names, or one of the following:\n");
--- a/passes/cmds/select.cc
+++ b/passes/cmds/select.cc
@@ -761,10 +761,10 @@ static void select_stmt(RTLIL::Design *design, std::string arg, bool disable_emp
 		return;
 	}
 
-	bool select_blackboxes = true;
+	bool select_blackboxes = false;
 	if (arg.substr(0, 1) == "=") {
 		arg = arg.substr(1);
-		select_blackboxes = false;
+		select_blackboxes = true;
 	}
 
 	if (!design->selected_active_module.empty()) {
@@ -1113,6 +1113,9 @@ struct SelectPass : public Pass {
 		log("    <obj_pattern>\n");
 		log("        select the specified object(s) from the current module\n");
 		log("\n");
+		log("Prefix the following patterns with '=' if the pattern should match black-/\n");
+		log("white-box modules and their contents.\n");
+		log("\n");
 		log("A <mod_pattern> can be a module name, wildcard expression (*, ?, [..])\n");
 		log("matching module names, or one of the following:\n");
 		log("\n");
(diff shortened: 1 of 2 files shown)


eddiehung:
My argument would be that `select` should not be reaching inside blackbox or whiteboxes, which is the behaviour Yosys passes should be doing across the board -- for all intents and purposes they should invisible. Besides pass specific options like `techmap -wb`, there is also `wbflip` to enable such (rare) operations to happen.

Now I've already forgotten why I needed it for #1926, but another example would be the `select -assert-count` that we often use for tests -- without constraining the search, these should not start counting things inside black/white boxes because those are things we almost-always don't really care about. Thoughts?

cliffordwolf:
> My argument would be that select should not be reaching inside blackbox or whiteboxes, which is the behaviour Yosys passes should be doing across the board

At least so far the behavior always was that the select expressions reach into blackbox modules, but the passes ignore the blackbox modules in the selection. There should be no need to explicitly exclude backbox modules at the level of selection expressions.

However, if we do filter black boxes in select expressions, then we should at least do it consistently. With your code, `SB*/*` would not match the contents of backbox modules, but `SB*` or `*/*` would.

> Besides pass specific options like techmap -wb

I think `techmap -wb` is actually an excellent example for why select expressions should match blackboxes as well. Otherwise a command like `techmap -wb <pattern>` is pointless.

To keep this conversation going I have now pushed d834cc7 to this branch. It adds a `=` prefix for patterns so that we can create patterns that exclude backboxes. The default behavior is unchanged.

eddiehung:
As discussed, inverted your commit so that black/white-boxes are only selected if `=` prefix is present, and updated help text and test.

passes/cmds/select.cc
```diff
@@ -1113,6 +1113,9 @@ struct SelectPass : public Pass {
 		log("    <obj_pattern>\n");
 		log("        select the specified object(s) from the current module\n");
 		log("\n");
+		log("Prefix the following patterns with '=' if the pattern should match black-/\n");
```
cliffordwolf, line 1116: It says "the _following_ patterns" but actually that's in the middle of the section describing patterns.

passes/cmds/select.cc
```diff
@@ -1113,6 +1113,9 @@ struct SelectPass : public Pass {
...
+		log("white-box modules and their contents.\n");
```
cliffordwolf, line 1117: That inversion of the previous statement is at least misleading. The this is written it suggests that `=` selects only black-/white-box module. But it selects them in addition to other modules.

passes/cmds/select.cc
eddiehung: I moved it from being in between describing mod_pattern and obj_pattern to before both, or it could after both.
Would this be clearer?
```suggestion
		log("Prefix <[mod|obj]_pattern> with '=' if the pattern should match black-/\n");
```

passes/cmds/select.cc
eddiehung: Good point. See new suggestion.

passes/cmds/select.cc
```diff
@@ -1099,6 +1113,9 @@ struct SelectPass : public Pass {
...
```
eddiehung, line 1117:
```suggestion
		log("By default, <[mod|obj]_pattern> will not match black/white-box modules or");
		log("their contents. To include such modules, prefix these patterns with '='.\n");
```

passes/cmds/select.cc
cliffordwolf: Why not just "prefix the pattern"? I just don't like the use of "following" at a position that does not precede the section on patterns but is in the middle of it.  (moving it up is fine and definitely an improvement but it's still between the paragraphs that outline the general format and the paragraphs the explain the details.)

The issue I have with "Prefix <[mod|obj]_pattern>" is that in `<mod_pattern>/<obj_pattern>` you can put the prefix before  `<mod_pattern>`, but not `<obj_pattern>`, because it's a prefix to the entire pattern, not one of the two components.

passes/cmds/select.cc
cliffordwolf: `<[mod|obj]_pattern>` are components of a pattern. `=` is a prefix to the entire pattern, not to either component.

eddiehung:
```suggestion
		log("By default, patterns will not match black/white-box modules or their");
		log("contents. To include such objects, prefix the pattern with '='.\n");
```

eddiehung: If that doesn't do it let's stop playing whack a mole. Just edit my suggestion and commit it.

PR #1950, opened 2020-04-16
design: -import to not count black/white-boxes as candidates for top

eddiehung:


passes/cmds/design.cc
```diff
@@ -228,14 +228,20 @@ struct DesignPass : public Pass {
...
+						candidates.clear();
+						candidates.push_back(module);
 						break;
 					}
+					if (!module->get_blackbox_attribute())
+						candidates.push_back(module);
 				}
+
+				if (GetSize(candidates) == 1)
+					copy_src_modules = std::move(candidates);
```
cliffordwolf, line 244: It looks to me like we want to do this always. Otherwise `copy_src_modules` will contain everything (including black boxes) for a design with two non-blackbox modules, but only the one module for a design with only one non-blackbox module. To me that sounds like an odd inconsistency. (Before we only had that when the one module is marked as top, which arguably already was odd, but now we also have that for a case with one non-top non-blackbox module.)

eddiehung: Sorry, I'm not following. `design -import` will only succeed if there is one top module (either marked explicitly with `(* top *)` or if there is only one non-black/white-box module):
https://github.com/YosysHQ/yosys/blob/9eace8f36032665394e582a60fc58be0a7c9303b/passes/cmds/design.cc#L266-L267
Once it passes that check, it then imports only those modules that are instantiated beyond that.

PR #1957, opened 2020-04-17
intel_alm: direct LUTRAM cell instantiation

ZirconiumX:
By instantiating the LUTRAM cell directly, we avoid a trip through altsyncram, which speeds up Quartus synthesis time. This also gives a little more flexibility, as Yosys can build RAMs out of individual 32x1 LUTRAM cells.

While working on this, I discovered that the mem_init0 parameter of `<family>_mlab_cell` gets ignored by Quartus. That means that until `memory_bram` can pass information about a cell RAM init file, LUTRAMs can only initialise to zero (which is the same as uninitialisable to `memory_bram`).

I'm aware of the duplication here between Cyclone V and Cyclone 10GX, but since there's already a lot of duplication, next on the TODO list is a big deduplication commit.


ZirconiumX:
For the sake of transparency to the - zero? - people watching this: it seems like trying to get synthesis working using Quartus as a back-end is really tricky. I can make a series of non-functional changes and end up with broken synthesis. Some *major* test additions are in order before this gets anywhere.

ZirconiumX:
ping @mwkmwkmwk

techlibs/intel_alm/synth_intel_alm.cc
```diff
@@ -164,6 +164,7 @@ struct SynthIntelALMPass : public ScriptPass {
 			run(stringf("read_verilog -sv -lib +/intel/%s/cells_sim.v", family_opt.c_str()));
 			run(stringf("read_verilog -specify -lib -D %s +/intel_alm/common/alm_sim.v", family_opt.c_str()));
 			run(stringf("read_verilog -specify -lib -D %s +/intel_alm/common/dff_sim.v", family_opt.c_str()));
+            run(stringf("read_verilog -specify -lib -D %s +/intel_alm/common/mem_sim.v", family_opt.c_str()));
```
mwkmwkmwk, line 167: mismatched indent style

techlibs/intel_alm/common/mem_sim.v
```diff
@@ -0,0 +1,59 @@
...
+// is high, writes it to the address in A1ADDR. A1EN can therefore be used to
+// conditionally write data to the MLAB.
+//
+// Simultaneously, the MLAB reads data from B1ADDR, and outputs it to B1DATA,
+// asynchronous to CLK1 and ignoring A1EN. If a synchronous read is needed
+// then the output can be fed to embedded flops. Presently, Yosys assumes
+// Quartus will pack external flops into the MLAB, but this is an assumption
+// that needs testing.
+//
+// If data is simultaneously read from and written to an address, the read
```
mwkmwkmwk, line 46: outdated

techlibs/intel_alm/synth_intel_alm.cc
mwkmwkmwk: uhh, still

PR #1961, opened 2020-04-18
ast, rpc: record original name of $paramod\* as \hdlname attribute

whitequark:
The `$paramod` name mangling is not invertible (the `\` character, which separates the module name from the parameters, is valid in the module name itself), which does not stop people from trying to invert it.

This commit makes it easy to invert the name mangling by storing the original name explicitly in the `\hdlname` attribute, and fixes the firrtl backend to use the newly introduced attribute.

**Unresolved question:** is `\hdlname` the appropriate attribute here? Currently it's only used in the `unique` pass, and I think the intent is the same, but I'd rather double check.

### DIFF
--- a/backends/firrtl/firrtl.cc
+++ b/backends/firrtl/firrtl.cc
@@ -306,17 +306,8 @@ struct FirrtlWorker
 		// If this is a parameterized module, its parent module is encoded in the cell type
 		if (cell->type.begins_with("$paramod"))
 		{
-			std::string::iterator it;
-			for (it = cell_type.begin(); it < cell_type.end(); it++)
-			{
-				switch (*it) {
-					case '\\': /* FALL_THROUGH */
-					case '=': /* FALL_THROUGH */
-					case '\'': /* FALL_THROUGH */
-					case '$': instanceOf.append("_"); break;
-					default: instanceOf.append(1, *it); break;
-				}
-			}
+			log_assert(cell->has_attribute(ID::hdlname));
+			instanceOf = cell->get_string_attribute(ID::hdlname);
 		}
 		else
 		{
--- a/frontends/ast/ast.cc
+++ b/frontends/ast/ast.cc
@@ -1565,6 +1565,9 @@ std::string AstModule::derive_common(RTLIL::Design *design, const dict<RTLIL::Id
 	rewritten.reserve(GetSize(parameters));
 
 	AstNode *new_ast = ast->clone();
+	if (!new_ast->attributes.count(ID::hdlname))
+		new_ast->attributes[ID::hdlname] = AstNode::mkconst_str(stripped_name);
+
 	para_counter = 0;
 	for (auto child : new_ast->children) {
 		if (child->type != AST_PARAMETER)
--- a/frontends/rpc/rpc_frontend.cc
+++ b/frontends/rpc/rpc_frontend.cc
@@ -217,6 +217,8 @@ struct RpcModule : RTLIL::Module {
 				module.second->name = mangled_name;
 				module.second->design = design;
 				module.second->attributes.erase(ID::top);
+				if (!module.second->has_attribute(ID::hdlname))
+					module.second->set_string_attribute(ID::hdlname, module.first.str());
 				design->modules_[mangled_name] = module.second;
 				derived_design->modules_.erase(module.first);
 			}


eddiehung:
For context, why do you need just the hdlname? In what I've been working on recently, I find I need this name and the cell parameters in order to go back to what I had originally..

whitequark:
> For context, why do you need just the hdlname?

In the FIRRTL backend the changes speak for themselves. In the CXXRTL backend, the cell parameters are taken from the cell (unless I'm mistaken and they aren't guaranteed to be preserved there?).

eddiehung:
> > For context, why do you need just the hdlname?
> 
> In the FIRRTL backend the changes speak for themselves. In the CXXRTL backend, the cell parameters are taken from the cell (unless I'm mistaken and they aren't guaranteed to be preserved there?).

My understanding is `hierarchy` will "uniquify" the instantiated modules by (a) deriving the module according to the cell's parameters, (b) rewrite the cell to instantiate that module directly, dropping any parameters that exist.

Exemplified by:
```
read_verilog <<EOT
module sub(input [WIDTH-1:0] i, output [WIDTH-1:0] o);
parameter WIDTH = 0;
assign o = i;
endmodule

module top(input i, o);
wire w1;
wire [1:0] w2;
sub #(.WIDTH(1)) s1 (.i({1{i}}), .o(w1));
sub #(.WIDTH(2)) s2 (.i({2{i}}), .o(w2));
assign o = ^{w1,w2};
endmodule
EOT
dump c:s?
hierarchy -top top
dump c:s?
```
giving:
```
1. Executing Verilog-2005 frontend: <<EOT
Parsing Verilog input from `<<EOT' to AST representation.
Generating RTLIL representation for module `\sub'.


  attribute \src "<<EOT:10.18-10.41"
  cell \sub \s2
    parameter signed \WIDTH 2
    connect \o \w2
    connect \i { \i \i }
  end

  attribute \src "<<EOT:9.18-9.41"
  cell \sub \s1
    parameter signed \WIDTH 1
    connect \o \w1
    connect \i \i
  end

2. Executing HIERARCHY pass (managing design hierarchy).

2.1. Analyzing design hierarchy..
Top module:  \top
Used module:     \sub
Parameter \WIDTH = 2

2.2. Executing AST frontend in derive mode using pre-parsed AST for module `\sub'.
Parameter \WIDTH = 2
Generating RTLIL representation for module `$paramod\sub\WIDTH=2'.
Parameter \WIDTH = 1

2.3. Executing AST frontend in derive mode using pre-parsed AST for module `\sub'.
Parameter \WIDTH = 1
Generating RTLIL representation for module `$paramod\sub\WIDTH=1'.

2.4. Analyzing design hierarchy..
Top module:  \top
Used module:     $paramod\sub\WIDTH=2
Used module:     $paramod\sub\WIDTH=1

2.5. Analyzing design hierarchy..
Top module:  \top
Used module:     $paramod\sub\WIDTH=2
Used module:     $paramod\sub\WIDTH=1
Removing unused module `\sub'.
Removed 1 unused modules.


  attribute \src "<<EOT:10.18-10.41"
  cell $paramod\sub\WIDTH=2 \s2
    connect \o \w2
    connect \i { \i \i }
  end

  attribute \src "<<EOT:9.18-9.41"
  cell $paramod\sub\WIDTH=1 \s1
    connect \o \w1
    connect \i \i
  end
```

If short enough, the parameters are placed in the module name. For longer parameters, a SHA1 hash is used.

whitequark:
> My understanding is `hierarchy` will "uniquify" the instantiated modules by (a) deriving the module according to the cell's parameters, (b) rewrite the cell to instantiate that module directly, dropping any parameters that exist.

Thanks for explanation. I didn't realize that (b) will happen, but it makes sense. It means that I am, indeed, mistaken, and the functionality introduced here would not be useful for cxxrtl. (This also explains why black boxes are never derived!)

I've edited the PR description to not mention cxxrtl. I think the other considerations still apply. For example, a tool examining post-`hierarchy` netlist might want to show an error message that mentions the name of the module in the source code, or perhaps use that to highlight this module in an IDE.

In principle, I have no objection to extending this PR such that it includes the original parameters somehow, although I don't need it myself.

eddiehung:
> Thanks for explanation. I didn't realize that (b) will happen, but it makes sense. It means that I am, indeed, mistaken, and the functionality introduced here would not be useful for cxxrtl. (This also explains why black boxes are never derived!)

The extreme case is that you don't want to derive a new module for every unique LUT mask, otherwise you could end up with a lot of them clogging up the design for no additional value. Incidentally, `techmap` does this right now when mapping from `$lut` and is one big reason why it is so slow. We have a few ideas about how to improve this, but it's of low priority right now.

> I've edited the PR description to not mention cxxrtl. I think the other considerations still apply. For example, a tool examining post-`hierarchy` netlist might want to show an error message that mentions the name of the module in the source code, or perhaps use that to highlight this module in an IDE.

I think these are enough reasons to merge it.

> I'm fine with extending this PR such that it includes the original parameters somehow, although I don't need it myself. Let me know if you have a specific mapping you'd like me to implement.

I don't think there's an elegant way to preserve the original parameters. In what I've been working on, I dynamically create a new named design containing techmap rules that can go from derived cell back to the original cell with parameters via `techmap -map %<name>`.

PR #1968, opened 2020-04-19
Fixed issue #1837 by clearing leftover AST_INTERNAL state.

### LINKED ISSUE #1837: File order dependent segmentation fault when parsing parameterized SV interfaces
I've provided a minimal working example in [minpass.zip](https://github.com/YosysHQ/yosys/files/4397697/minpass.zip) and a similar failing example in [minfail.zip](https://github.com/YosysHQ/yosys/files/4397696/minfail.zip).

Tried on newest commit 1dbc7017
Simply running `yosys if2.sv if1.sv main.sv` causes a segmentation fault with a warning: `Warning: Deep recursion in AST simplifier.` Note that running 
`yosys if2.sv main.sv if1.sv` intererstingly does not.

I also noticed that the tests in tests/svinterfaces fail with segmentation faults for me.

I originally started writing this error report with my current install (bfeba9ad) where I got the following error message instead: `if1.sv:0: ERROR: Failed to detect width for identifier \ADDR_WIDTH!`
Not sure if that is relevant.

Version info: `Yosys 0.9+1706 (git sha1 1dbc7017, gcc 7.4.0-1ubuntu1~18.04.1 -fPIC -Os)`

## Expected behavior
File ordering should not cause a segmentation fault.

## Actual behavior
File ordering can cause a segmentation fault.

Afromaniac:
Pull request that Closes #1837 
Once the issue was found the fix seems simple. Th tests succeeded and my original error has disappeared. The `AST` preprocessing code is a bit too dense for me to tell if this just cures a symptom or the cause itself.

### DIFF
--- a/frontends/ast/ast.cc
+++ b/frontends/ast/ast.cc
@@ -1152,8 +1152,8 @@ static AstModule* process_module(AstNode *ast, bool defer, AstNode *original_ast
 void AST::process(RTLIL::Design *design, AstNode *ast, bool dump_ast1, bool dump_ast2, bool no_dump_ptr, bool dump_vlog1, bool dump_vlog2, bool dump_rtlil,
 		bool nolatches, bool nomeminit, bool nomem2reg, bool mem2reg, bool noblackbox, bool lib, bool nowb, bool noopt, bool icells, bool pwires, bool nooverwrite, bool overwrite, bool defer, bool autowire)
 {
-    current_scope.clear();
-    
+	current_scope.clear();
+	
 	current_ast = ast;
 	current_ast_mod = nullptr;
 	flag_dump_ast1 = dump_ast1;
@@ -1429,10 +1429,9 @@ RTLIL::IdString AstModule::derive(RTLIL::Design *design, const dict<RTLIL::IdStr
 	}
 
 	std::string new_modname = modname;
-	if (has_interfaces)
-    {
+	if (has_interfaces) {
 		new_modname += "$interfaces$" + interf_info;
-    }
+	}
 
 	if (!design->has(new_modname)) {
 		if (!new_ast) {
--- a/frontends/ast/ast.cc
+++ b/frontends/ast/ast.cc
@@ -993,7 +993,7 @@ static AstModule* process_module(AstNode *ast, bool defer, AstNode *original_ast
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


frontends/ast/ast.cc
```diff
@@ -1152,6 +1152,8 @@ static AstModule* process_module(AstNode *ast, bool defer, AstNode *original_ast
 void AST::process(RTLIL::Design *design, AstNode *ast, bool dump_ast1, bool dump_ast2, bool no_dump_ptr, bool dump_vlog1, bool dump_vlog2, bool dump_rtlil,
 		bool nolatches, bool nomeminit, bool nomem2reg, bool mem2reg, bool noblackbox, bool lib, bool nowb, bool noopt, bool icells, bool pwires, bool nooverwrite, bool overwrite, bool defer, bool autowire)
 {
+    current_scope.clear();
```
boqwxp, line 1155: Use tabs to indent rather than four spaces.

frontends/ast/ast.cc
```diff
@@ -1428,8 +1430,9 @@ RTLIL::IdString AstModule::derive(RTLIL::Design *design, const dict<RTLIL::IdStr
 
 	std::string new_modname = modname;
 	if (has_interfaces)
+    {
 		new_modname += "$interfaces$" + interf_info;
-
+    }
```
boqwxp, line 1435: Same here.

eddiehung: Actually it looks like these braces add no additional value?

frontends/ast/ast.cc
Afromaniac: Oh, you're completely right, my bad!

frontends/ast/ast.cc
Afromaniac: The braces are a leftover from the debugging process. You're completely right in saying that they're not necessary but I left them in to match the local code style.

eddiehung: Generally, it is good practice to minimise the size of the PR so that it's easier to review.

eddiehung:
Is it easy to add a testcase to make sure that this gets fixed and stays fixed?

cliffordwolf:
Thanks for tracking this down. However, I think #1981 is the correct fix for this.

PR #1969, opened 2020-04-20
kernel: Add `pool` support for rvalue references and C++11 move semantics.

boqwxp:
This is similar to the recent changes to `dict`.

This always avoids a `std::move` by `emplace_back()`ing `entry_t`s in-place, and avoids making a copy when `insert()` is given an rvalue reference by `std::forward<>()`ing and `std::move()`ing it into the `udata` member of `entry_t`.


eddiehung:
@boqwxp To preserve semantics I added an emplace method too.

Issue #1733, opened 2020-03-04
EDIF backend produces a file with multi-driven net.


tmichalak:
## Steps to reproduce the issue

I have a SoC design with a VexRiscv core that when synthesized with the latest Yosys (``Yosys 0.9+1706 (git sha1 0ec97144, clang 8.0.0 -fPIC -Os``) fails in Vivado during EDIF parsing:
```
Parsing EDIF File [./design.edif]
WARNING: [EDIF 20-100] Multiple nets with the same name 'GND_NET' specified. Folding connectivity of these nets into a single net. [/work/test2/zephyr_broken_edif/build/design.edif:27427]
ERROR: [EDIF 20-80] Cannot connect net 'IBusCachedPlugin_predictionJumpInterface_payload[0]' to pin '$auto$simplemap.cc:420:simplemap_dff$7519/Q' in cell 'VexRiscv'. This pin is already connected to net '_zz_115_[0]'. The new connection will be ignored. [/work/test2/zephyr_broken_edif/build/design.edif:68684]
Finished Parsing EDIF File [./design.edif]
```

I attached a test case:
[zephyr_broken_edif.zip](https://github.com/YosysHQ/yosys/files/4291600/zephyr_broken_edif.zip)
Simply run with ``make yosys``

I ran a `git bisect` on the current `master` and it pointed me to the following commit:
```
  commit 50d70288d0f1b516dc04f38494eea94612104974
Author: Claire Wolf <clifford@clifford.at>
Date:   Wed Jan 29 14:07:11 2020 +0100

    Preserve wires with keep attribute in EDIF back-end
    
    Signed-off-by: Claire Wolf <clifford@clifford.at>
```

## Expected behavior

Vivado should successfully parse the EDIF and continue with P&R.

## Actual behavior

When inspecting the `design.edif` I can clearly see that we have a multi-driven net:
```
...
          (net (rename id09476 "IBusCachedPlugin_fetchPc_pcReg[0]") (joined
            (portRef (member io_cpu_fetch_pc 31) (instanceRef IBusCachedPlugin_cache))
            (portRef D (instanceRef id04838))
            (portRef Q (instanceRef id05980))
            )
          )
...
          (net (rename id09507 "_zz_115_[0]") (joined
            (portRef (member io_cpu_decode_pc 31) (instanceRef IBusCachedPlugin_cache))
            (portRef D (instanceRef id05515))
            (portRef Q (instanceRef id04838))
            )
          )
...
[...]

PR #1971, opened 2020-04-20
Ignore conflicting keep attributes, unless asked not to

cliffordwolf:
This replicated the XST behavior of simply ignoring nets with conflicting KEEP attributes, as described in https://github.com/YosysHQ/yosys/issues/1733#issuecomment-595923804.

I've investigated options for writing BUF cells for those nets directly in the EDIF file, but because of the way EDIF works and the way the current back-end is written this turns out to be not feasible without a major rewrite of the EDIF back-end.

An alternative would be a different pass that adds the buffers before even calling the EDIF back-end.

### DIFF
--- a/backends/edif/edif.cc
+++ b/backends/edif/edif.cc
@@ -345,6 +345,70 @@ struct EdifBackend : public Backend {
 			*f << stringf("        (viewType NETLIST)\n");
 			*f << stringf("        (interface\n");
 
+			for (auto cell : module->cells()) {
+				for (auto &conn : cell->connections())
+					if (cell->output(conn.first))
+						sigmap.add(conn.second);
+			}
+
+			for (auto wire : module->wires())
+				for (auto b1 : SigSpec(wire))
+				{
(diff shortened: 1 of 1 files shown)


mwkmwkmwk:
This code fails in situations where it should work:
```
module top(input wire i, output wire o);

(* keep *)
wire a;

assign a = ~i;
assign o = ~a;

endmodule
```

`yosys t.v -o t.edif`:
`Warning: Ignoring conflicting 'keep' property on net a. Use -keep to generate the extra net nevertheless.`

IMO we should do a first pass through all wires and, when a keep wire is noticed, make the pass use this wire's bits as the "representative" (overriding whatever sigmap suggests), unless doing so would conflict with another keep wire (which should cause a warning)

cliffordwolf:
> IMO we should do a first pass through all wires and, when a keep wire is noticed, make the pass use this wire's bits as the "representative" (overriding whatever sigmap suggests), unless doing so would conflict with another keep wire (which should cause a warning)

Done.

PR #1973, opened 2020-04-20
tests: fix various/plugin.sh when PREFIX != /usr/local/share

### LINKED ISSUE #1966: tests: various/plugin.sh fails when PREFIX != /usr/local/share
This has already been brought up in https://github.com/YosysHQ/yosys/commit/f11dd6e20838bce1b1fa8e085570d2126484d10f#r38433102, but I'll add it here for better visibility and tracking.

```
make -j8 PREFIX=/usr/share test
```

## Expected behavior

The tests complete without error.

## Actual behavior

`tests/various/plugin.cc` fails to compile with:

```
plugin.cc:1:10: fatal error: 'kernel/rtlil.h' file not found
#include "kernel/rtlil.h"
         ^~~~~~~~~~~~~~~~
```

This is due to the replacement of include path flags that happens in `plugin.sh`, which assumes `PREFIX=/usr/local/share`.

eddiehung:
Fixes #1966

### DIFF
--- a/misc/yosys-config.in
+++ b/misc/yosys-config.in
@@ -6,7 +6,7 @@ help() {
 		echo "Usage: $0 [--exec] [--prefix pf] args.."
 		echo "       $0 --build modname.so cppsources.."
 		echo ""
-		echo "Replecement args:"
+		echo "Replacement args:"
 		echo "    --cxx         @CXX@"
 		echo "    --cxxflags    $( echo '@CXXFLAGS@' | fmt -w60 | sed ':a;N;$!ba;s/\n/ \\\n                      /g' )"
 		echo "    --ldflags     @LDFLAGS@"
--- a/tests/various/plugin.sh
+++ b/tests/various/plugin.sh
@@ -1,6 +1,8 @@
 set -e
 rm -f plugin.so
 CXXFLAGS=$(../../yosys-config --cxxflags)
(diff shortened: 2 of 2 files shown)


cliffordwolf:
Just `DATDIR` is a bit unspecific imo for a Yosys-specific environment variable. I'd say let's use something like `YOSYS_DATDIR` instead and extend `proc_share_dirname()` (in kernel/yosys.cc) to also check that environment variable.

eddiehung:
> proc_share_dirname

`DATDIR` comes from https://github.com/YosysHQ/yosys/blob/c32b4bded582e121056a497131bf617684448cc4/Makefile#L61

Actually, since `--datdir` exists I could use that as a search/replace...

eddiehung:
Didn't realise you could actually get `DATDIR` from `yosys-config`; using that instead now.

cliffordwolf:
> DATDIR comes from `yosys/Makefile`

My point was that with the original PR the installed `yosys-config` program would be sensitive to a DATDIR environment variable, and I think the executables we install should only be sensitive to standard environment variables (such as PATH), and environment variables that clearly are Yosys-specific (such as ones prefixed with `YOSYS_`).

> The only reason I want to replace DATDIR is for a test.

But the executable would always be sensitive to the DATDIR environment variable, not just in the context of that test. People might for example use `DATDIR` in their own scripts or Makefiles for something entirely different and then wonder why their Yosys plugins do not build.

(Just explaining the reasoning behind my original comment. The new PR is looking fine.)

PR #1978, opened 2020-04-21
bugpoint: Don't remove modules while iterating over them.

mwkmwkmwk:
Reported by @ZirconiumX.


eddiehung:
Is it possible to add a small `bugpoint` test?

PR #1987, opened 2020-04-23
Preserve 'signed'-ness of an AST_WIRE in it's attribute

jersey99:
Prior to this signedness of an AST_WIRE was lost prior to RTLIL transformation (which is OK for synthesis).

However, preserving this information in attributes (suggested by @mwkmwkmwk (?) ) can help automation tools that rely on Yosys.

constids.inc: Adds a constant SIGNED
ast.cc: signedness of an AST_WIRE is added as an attribute with ID:SIGNED


eddiehung:
I would suggest the attribute be lower case instead of upper case, as that fits other attributes. Also, does it make sense for backends (in particular, Verilog) to emit this natively?

jersey99:
> I would suggest the attribute be lower case instead of upper case, as that fits other attributes.

I tried that and finally went with upper case, as "signed" is a C++11 keyword @eddiehung, maybe one could do "Signed"?

> Also, does it make sense for backends (in particular, Verilog) to emit this natively?

Re-parsing a verilog output for automation, is an option I like less as I have had a lot of success with `write_json` recently.

cliffordwolf:
Since `upto` and `start_offset` are members of RTLIL::Wire, I would suggest making this a native member of that struct as well.

> However, preserving this information in attributes can help automation tools that rely on Yosys.

I'd like to see one of those use cases. Who is doing what with this information?

eddiehung:
What about re-using the existing `wiretype` attribute?
https://github.com/YosysHQ/yosys/blame/master/README.md#L443-L444

jersey99:
> I'd like to see one of those use cases. Who is doing what with this information?

@cliffordwolf : @ldoolitt and I have been using verilog parsers for register bus automation (decoder generation and register information export). We started by parsing comments and then graduated to using attributes (using iverilog parser). I personally find the yosys parser cleaner and also think that we can benefit a lot from RTLIL into the future (especially now that a GHDL plugin is in the works). Hope that answers your question.

> What about re-using the existing wiretype attribute?

@eddiehung Could you elaborate a little more. I see that currently wiretype is an attribute that is used once in simplify.cc, which is linked to is_custom_type set inside the parser.

I am going to look into making signedness a native field of RTLIL::Wire struct for now.

eddiehung:
> > What about re-using the existing wiretype attribute?
> 
> @eddiehung Could you elaborate a little more. I see that currently wiretype is an attribute that is used once in simplify.cc, which is linked to is_custom_type set inside the parser.

Scratch that idea. I spoke with the BDFL and they didn't like it.

jersey99:
Closing this, see #2006

PR #1992, opened 2020-04-23
bugpoint: improve help text

eddiehung:
It wasn't clear how to use `bugpoint` from the help text, for example, so I rewrote it.

If others also think it's clearer too, then feel free to merge. Otherwise it's just me not really thinking it through!


mwkmwkmwk:
I'm not sure using `keep` for this purpose is a good idea, as the presence/absence of `keep` is something that could be involved in the actual bug being tracked down.  How about using a separate attribute? `(* bugpoint_keep *)`?

whitequark:
> How about using a separate attribute? `(* bugpoint_keep *)`?

Agreed re: separate attribute. Should it be `bugpoint_keep` or `bugpoint.keep`? I've been using the latter with cxxrtl (I asked and it looks like no one opined on it.)

whitequark:
Oh wait, Claire did tell me the convention is to use an underscore, I just missed it. Now I need to fix cxxrtl...

passes/cmds/bugpoint.cc
```diff
@@ -304,6 +319,9 @@ struct BugpointPass : public Pass {
 		bool fast = false, clean = false;
 		bool modules = false, ports = false, cells = false, connections = false, assigns = false, updates = false, has_part = false;
 
+		log_header(design, "Executing BUGPOINT pass (minimize testcases).\n");
```
whitequark, line 322: I vaguely recall intentionally not adding a header for some reason. But if you think a header looks OK, let's keep it.

PR #1995, opened 2020-04-23
verific: do not assert if wire not found; warn instead

eddiehung:


ZipCPU:
This seems wrong to me.  Why should a "valid" design result in a warning?  If it's a problem within the tool then shouldn't we fix the tool instead so that this warning no longer takes place?

cliffordwolf:
> This seems wrong to me. Why should a "valid" design result in a warning? If it's a problem within the tool then shouldn't we fix the tool instead so that this warning no longer takes place?

I'd say let's un-break this first, then investigate further what is going wrong.

Do we have a minimal example that triggers the warning?

eddiehung:
See #1999

PR #1996, opened 2020-04-24
frontend: Include complete source location instead of just `location.first_line` in `frontends/ast/genrtlil.cc`.

boqwxp:
This PR adds the complete source location, including `first_line`, `first_column`, `last_line`, and `last_column`, in all the places in `genrtlil.cc` where source location attributes are currently set.


boqwxp:
@eddiehung Can you please take a look at this?

PR #2000, opened 2020-04-24
kernel: Trap in `log_error()` when a debugger is attached

whitequark:
The workflow of debugging fatal pass errors in Yosys is flawed in three ways:
 1. Running Yosys under a debugger is sufficient for the debugger to catch some fatal errors (segfaults, aborts, STL exceptions) but not others (`log_error()`, `log_cmd_error()`). This is neither obvious nor easy to remember.
 2. To catch Yosys-specific fatal errors, it is necessary to set a breakpoint at `logv_error_with_prefix()`, or at least, `logv_error()`. This is neither obvious nor easy to remember, and GDB's autocomplete takes many seconds to suggest function names due to the large amount of symbols in Yosys.
 3. If a breakpoint is not set and Yosys encounters with such a fatal error, the process terminates. When debugging a crash that takes a long time to reproduce (or a nondeterministic crash) this can waste a significant amount of time.

To solve this problem, add a macro `YS_DEBUGTRAP` that acts as a hard breakpoint (if available), and a macro `YS_DEBUGTRAP_IF_DEBUGGING` that acts as a hard breakpoint only if debugger is present.

Then, use `YS_DEBUGTRAP_IF_DEBUGGING` in `logv_error_with_prefix()` to obviate the need for a breakpoint on nearly every platform.

<hr>

This PR is a complete rewrite of #1938 that makes it much simpler and also fully portable. I believe it addresses all of the use cases requested by @boqwxp. Currently, `YS_DEBUGTRAP` is not used anywhere in Yosys; it could be used for debugging, but we could also use it e.g. to request the compiler to trip after a certain pass, [like Swift does](https://github.com/apple/swift/blob/master/lib/SILOptimizer/PassManager/PassManager.cpp#L339).

This PR has been tested on Linux glibc and Windows MinGW. It should build and work on any Win32 platform and on any POSIX platform, regardless of the compiler and architecture, since it only relies on the most basic primitives provided by either. I did not try it on MSVC since the code seems obviously correct to me, but I will do so (and any other reasonable platform) on request.

I admit that the manipulations with `signal()` I do on *nix platforms look weird. However, not only are all the alternatives much worse...
 1. `ptrace()` interface differs between Linux and BSDs and `ptrace(PTRACE_TRACEME)` can fail for a lot of reasons, such as "having a security policy preventing ptrace from running" or "being run under strace" or "process called PTRACE_TRACEME already" with no way to distinguish those;
 2. reading from `/proc/self/status` to get tracer PID is just gross, but also Linux-specific, wouldn't necessarily work in a container, and would misidentify strace as a debugger.

... but also the solution I implemented is nearly perfect because it is guaranteed by POSIX to do nothing when not debugged, does not disturb strace (which records a trap and moves on), and is virtually certain to trap on a real debugger (both gdb and lldb do trap here).

The only downside of it I can think of is that it's not MT-safe. (`raise` is guaranteed to deliver `SIGTRAP` to the current thread, unlike the usual, but there is no way to ignore a signal just on one thread.) So in theory, something else could race with this code for changing the SIGTRAP handler. It seems profoundly unlikely though.

### DIFF
--- a/kernel/log.h
+++ b/kernel/log.h
@@ -50,9 +50,12 @@
 					std::regex_constants::egrep)
 #endif
 
-#ifndef _WIN32
+#if defined(_WIN32)
+#  include <intrin.h>
+#else
 #  include <sys/time.h>
 #  include <sys/resource.h>
+#  include <signal.h>
 #endif
 
 #if defined(_MSC_VER)
@@ -68,6 +71,41 @@ YOSYS_NAMESPACE_BEGIN
 #define S__LINE__sub2(x) #x
 #define S__LINE__sub1(x) S__LINE__sub2(x)
 #define S__LINE__ S__LINE__sub1(__LINE__)
+
+// YS_DEBUGTRAP is a macro that is functionally equivalent to a breakpoint
+// if the platform provides such functionality, and does nothing otherwise.
+// If no debugger is attached, it starts a just-in-time debugger if available,
+// and crashes the process otherwise.
+#if defined(_WIN32)
+# define YS_DEBUGTRAP __debugbreak()
+#else
+# ifndef __has_builtin
+// __has_builtin is a GCC/Clang extension; on a different compiler (or old enough GCC/Clang)
+// that does not have it, using __has_builtin(...) is a syntax error.
+#  define __has_builtin(x) 0
+# endif
+# if __has_builtin(__builtin_debugtrap)
+#  define YS_DEBUGTRAP __builtin_debugtrap()
+# elif defined(__unix__)
+#  define YS_DEBUGTRAP raise(SIGTRAP)
+# else
+#  define YS_DEBUGTRAP do {} while(0)
+# endif
+#endif
+
+// YS_DEBUGTRAP_IF_DEBUGGING is a macro that is functionally equivalent to a breakpoint
+// if a debugger is attached, and does nothing otherwise.
+#if defined(_WIN32)
+# define YS_DEBUGTRAP_IF_DEBUGGING do { if (IsDebuggerPresent()) DebugBreak(); } while(0)
+#elif defined(__unix__)
+// There is no reliable (or portable) *nix equivalent of IsDebuggerPresent(). However,
+// debuggers will stop when SIGTRAP is raised, even if the action is set to ignore.
+# define YS_DEBUGTRAP_IF_DEBUGGING do { \
+		sighandler_t old = signal(SIGTRAP, SIG_IGN); raise(SIGTRAP); signal(SIGTRAP, old); \
+	} while(0)
+#else
+# define YS_DEBUGTRAP_IF_DEBUGGING do {} while(0)
+#endif
 
 struct log_cmd_error_exception { };
 
--- a/kernel/log.cc
+++ b/kernel/log.cc
@@ -354,6 +354,9 @@ static void logv_error_with_prefix(const char *prefix,
 
 	if (check_expected_logs)
 		log_check_expected();
+
+	YS_DEBUGTRAP_IF_DEBUGGING;
+
 #ifdef EMSCRIPTEN
 	log_files = backup_log_files;
 	throw 0;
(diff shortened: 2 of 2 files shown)


kernel/log.h
```diff
@@ -69,6 +70,41 @@ YOSYS_NAMESPACE_BEGIN
...
+#  define YS_DEBUGTRAP raise(SIGTRAP)
+# else
+#  define YS_DEBUGTRAP do {} while(0)
+# endif
+#endif
+
+// YS_DEBUGTRAP_IF_DEBUGGING is a macro that is functionally equivalent to a breakpoint
+// if a debugger is attached, and does nothing otherwise.
+#if defined(_WIN32)
+# define YS_DEBUGTRAP_IF_DEBUGGING do { if (IsDebuggerPresent()) DebugBreak(); } while(0)
```
mwkmwkmwk, line 97: Why `DebugBreak` here and `__debugbreak` above?  What's the difference between those?

mwkmwkmwk:
Very good, this is exactly the approach I wanted to suggest :)

I guess the obvious follow-up would be to add a "debugbreak" pass?

kernel/log.h
whitequark: `__debugbreak` is an intrinsic, `DebugBreak` is a kernel32 function. `YS_DEBUGTRAP` tries to expand to a platform-independent intrinsic (so that it works on platforms like WASM or Graal well--I checked and both of them should prefer that).

Conversely, `YS_DEBUGTRAP_IF_DEBUGGING` is apriori platform dependent so it doesn't try that.

There's no actual difference in practice, I just wrote both of them idiomatically.

mwkmwkmwk: Ah, alright.  Thanks for the explanation.

whitequark:
Ah, I realized I forgot to address one objection raised in #1938, which is that this code might not inhibit code motion. I checked and Clang treats `__builtin_debugtrap` as a barrier. `raise` of course is a barrier because it might never return. `__debugbreak` must be a barrier for the same reason as `raise`: it raises a SEH exception and might never return. So I'm quite certain there will be no code motion anywhere around `YS_DEBUGBREAK`.

whitequark:
> I guess the obvious follow-up would be to add a "debugbreak" pass?

Ah, that didn't occur to me, actually. But it is very logical.

kernel/log.h
```diff
@@ -69,6 +70,41 @@ YOSYS_NAMESPACE_BEGIN
 #define S__LINE__sub1(x) S__LINE__sub2(x)
 #define S__LINE__ S__LINE__sub1(__LINE__)
 
+// YS_DEBUGTRAP is a macro that is functionally equivalent to a breakpoint
+// if the platform provides such functionality, and does nothing otherwise.
+// If no debugger is attached, it starts a just-in-time debugger if available,
+// and crashes the process otherwise.
+#if defined(_WIN32)
+# define YS_DEBUGTRAP __debugbreak()
```
boqwxp, line 78: Is this intrinsic guaranteed to exist for all compilers when `_WIN32` is defined?  Am I wrong, or does this need to be guarded by some checks for its existence like below?

whitequark: > I just wonder about the case of compiling for Windows with a non-MSVC compiler

The [MSDN docs](https://docs.microsoft.com/en-us/cpp/intrinsics/compiler-intrinsics?view=vs-2019) say:

> The use of intrinsics affects the portability of code, because intrinsics that are available in Visual C++ might not be available if the code is compiled with other compilers

In practice, MinGW does support this intrinsic, and Yosys currently doesn't support any other Win32 compilers. Since there is no way to check for an intrinsic other than by adding a guard checking for specific compiler vendors, I would say the guard is redundant and it should be added when (or if) Yosys is ported to another Win32 toolchain, which is in all likelihood will be never.

I would also be fine with using `DebugBreak()` instead.

> Also it looks like you need to `#include <intrin.h>`, too

I missed that, good catch.

boqwxp: Aha, I see.  OK, works for me.

Issue #1990, opened 2020-04-23
Dynamic bit write with signed expression fails


daveshah1:
Minimised from an example by @smunaut 

```verilog
module top(input [7:0] wr_mask, output reg [7:0] mem_dm_w);

	localparam DATA_WIDTH = 32;
	localparam MASK_WIDTH = 8;
	localparam NW = 2;

	always @(*)
	begin : map
		integer n, x, o;

		// Map actual bits
		for (n=0; n<MASK_WIDTH; n=n+1)
		begin
			// Determine position
			x = n % NW;	// Which SPRAM
			o = n / NW;	// Which nibble inside that SPRAM

			// Map IO
			mem_dm_w[( 4*x)+   o    ] = ~wr_mask[n];
		end
	end
endmodule
```
gives
```
ERROR: Found error in internal cell \top.$shl$signed_bitsel.v:0$134 ($shl) at kernel/rtlil.cc:850:
  attribute \src "signed_bitsel.v:0"
  cell $shl $shl$signed_bitsel.v:0$134
    parameter \Y_WIDTH 8
    parameter \B_WIDTH 32
    parameter \A_WIDTH 8
    parameter \B_SIGNED 1
    parameter \A_SIGNED 0
    connect \Y $shl$signed_bitsel.v:0$134_Y
    connect \B $add$signed_bitsel.v:19$133_Y
    connect \A $and$signed_bitsel.v:0$131_Y
  end
```

This appears to be a regression caused by #1851, reverting the merge commit causes it to build fine.

daveshah1:
Possible patch, although I don't know if this handles all cases correctly (e.g. if one of the vector bounds is negative, is this even legal?)

```patch
diff --git a/frontends/ast/simplify.cc b/frontends/ast/simplify.cc
index 837c14ad..5e12e3ac 100644
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -1774,7 +1774,7 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
                AstNode *range = children[0]->children[0];
 
                if (range->children.size() == 1) {
-                       shift_expr = range->children[0]->clone();
+                       shift_expr = new AstNode(AST_TO_UNSIGNED, range->children[0]->clone());
                } else {
                        shift_expr = range->children[1]->clone();
                        AstNode *left_at_zero_ast = range->children[0]->clone();
```

daveshah1:
[...]

PR #2005, opened 2020-04-27
Add "nowrshmsk" attribute, fix shift-and-mask bit slice write for signed offset

cliffordwolf:
Fixes some of #1990.

Remaining issues:
- Disable shift-and-mask when the code describes latches
- `[x+:N]` doesn't work properly for signed `x` with vectors with negative indices


cliffordwolf:
Signed indices and vectors that have negative start offsets should work fine now. I've also added some test cases for that.

PR #2006, opened 2020-04-27
Preserve 'signed'-ness of a verilog wire through RTLIL

jersey99:
As per suggestion made in https://github.com/YosysHQ/yosys/pull/1987, now:

`RTLIL::Wire` now holds an is_signed field.
`signed` is exported via JSON backend, and read in via JSON frontend
Also exported via dump_rtlil command, and read back via ilang_parser

I am sure I missed a few other places, and would love to add a test somewhere. Please feel free to advise!

### DIFF
--- a/backends/ilang/ilang_backend.cc
+++ b/backends/ilang/ilang_backend.cc
@@ -131,6 +131,8 @@ void ILANG_BACKEND::dump_wire(std::ostream &f, std::string indent, const RTLIL::
 		f << stringf("output %d ", wire->port_id);
 	if (wire->port_input && wire->port_output)
 		f << stringf("inout %d ", wire->port_id);
+	if (wire->is_signed)
+		f << stringf("signed ");
 	f << stringf("%s\n", wire->name.c_str());
 }
 
--- a/backends/json/json.cc
+++ b/backends/json/json.cc
@@ -160,6 +160,8 @@ struct JsonWriter
 				f << stringf("          \"offset\": %d,\n", w->start_offset);
 			if (w->upto)
 				f << stringf("          \"upto\": 1,\n");
+			if (w->is_signed)
+				f << stringf("          \"signed\": %d,\n", w->is_signed);
 			f << stringf("          \"bits\": %s\n", get_bits(w).c_str());
 			f << stringf("        }");
 			first = false;
@@ -227,6 +229,8 @@ struct JsonWriter
 				f << stringf("          \"offset\": %d,\n", w->start_offset);
 			if (w->upto)
 				f << stringf("          \"upto\": 1,\n");
+			if (w->is_signed)
+				f << stringf("          \"signed\": %d,\n", w->is_signed);
 			f << stringf("          \"attributes\": {");
 			write_parameters(w->attributes);
 			f << stringf("\n          }\n");
--- a/frontends/ast/genrtlil.cc
+++ b/frontends/ast/genrtlil.cc
@@ -1058,6 +1058,7 @@ RTLIL::SigSpec AstNode::genRTLIL(int width_hint, bool sign_hint)
 			wire->port_input = is_input;
 			wire->port_output = is_output;
 			wire->upto = range_swapped;
+			wire->is_signed = is_signed;
 
 			for (auto &attr : attributes) {
 				if (attr.second->type != AST_CONSTANT)
--- a/frontends/ilang/ilang_parser.y
+++ b/frontends/ilang/ilang_parser.y
@@ -192,6 +192,9 @@ wire_options:
 	wire_options TOK_UPTO {
 		current_wire->upto = true;
 	} |
+	wire_options TOK_SIGNED {
+		current_wire->is_signed = true;
+	} |
 	wire_options TOK_OFFSET TOK_INT {
 		current_wire->start_offset = $3;
 	} |
--- a/frontends/json/jsonparse.cc
+++ b/frontends/json/jsonparse.cc
@@ -309,6 +309,12 @@ void json_import(Design *design, string &modname, JsonNode *node)
 					port_wire->upto = val->data_number != 0;
 			}
 
+			if (port_node->data_dict.count("signed") != 0) {
+				JsonNode *val = port_node->data_dict.at("signed");
+				if (val->type == 'N')
+					port_wire->is_signed = val->data_number != 0;
+			}
+
 			if (port_node->data_dict.count("offset") != 0) {
 				JsonNode *val = port_node->data_dict.at("offset");
 				if (val->type == 'N')
@@ -573,4 +579,3 @@ struct JsonFrontend : public Frontend {
 } JsonFrontend;
 
 YOSYS_NAMESPACE_END
-
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -1862,6 +1862,7 @@ RTLIL::Wire *RTLIL::Module::addWire(RTLIL::IdString name, const RTLIL::Wire *oth
 	wire->port_input = other->port_input;
 	wire->port_output = other->port_output;
 	wire->upto = other->upto;
+	wire->is_signed = other->is_signed;
 	wire->attributes = other->attributes;
 	return wire;
 }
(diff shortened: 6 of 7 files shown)


whitequark:
It would be nice to have this used in the Verilog backend, but without extensive testing it is likely that this change would backfire.

jersey99:
@whitequark Have a look. I tested it with some verilog files at hand. However, I tried to look for something like dump_module or *Backend::execute in /tests, but failed to see anything that I can append a test to.

whitequark:
@jersey99 I'm not comfortable signing off on this without some VLogHammer testing. (@cliffordwolf could greenlight this, but I suspect her view would be the same as mine here...)

cliffordwolf:
This implementation LGTM. I do share @whitequark's concerns regarding the verilog backend bit. A simple way of "playing it save", if we really want this in the verilog back-end, would be to always use signed variables with `$unsigned()` in the generated verilog code.

However, as previously pointed out in https://github.com/YosysHQ/yosys/pull/1987#issuecomment-618293763 I'd like to see a use-case for this.

jersey99:
@cliffordwolf My use case for this is basically reflected in my comment in #1987. We have been using attributes heavily for bus independent code-generation and register map automation. I find that using Yosys parser and rtlil will provide more benefits to us into the future. However, if we want to get hold of the output of write_json or dump_rtlil, the signedness is lost and this PR will help preserve it.

@whitequark I spent some time with VlogHammer after your suggestion. syn_yosys and check_yosys run fine. However, there is no exercise of verilog_backend inside VlogHammer. From what I can see read_verilog, write_ilang and read_ilang are the ones used. The report.html generated between this PR and master is practically identical. Any more guidance is appreciated!

whitequark:
@cliffordwolf I'm in favor of having this functionality. It seems entirely in line with having `upto` and `start_offset`: all three are not directly useful to Yosys but could be used by downstream tooling; by PNR and codegen tools in case of `upto` and `start_offset`, and codegen tools alone in case of `signed`. If we have `upto` and `start_offset`, I think there is a good case to have `signed`.

It is true that Yosys seems to be biased towards synthesis tools, but I see no reason it should remain biased. nMigen is pushing Yosys towards HDL conversion and simulation, I think it's good that @jersey99 is pushing Yosys to be more useful for code generators.

jersey99:
@cliffordwolf  @whitequark To give you both a little more context (maybe a bit too much :) ), although this isn't quite at the level of (n)-migen and/or Litex's AutoCSR (where python-DSL within python makes life easier), this is at the moment more verilog-centric and could use some clean-up.

Code here is taken from https://github.com/BerkeleyLab/Bedrock/tree/newad_yosys_take1/rtsim

Take a verilog module (prng) defined in file (prng.v):

```

`timescale 1ns / 1ns

`include "prng_auto.vh"
// Almost useless shell around two tt880v modules
module prng(
	input clk,
	output [31:0] rnda,
	output [31:0] rndb,
	(* external *)
	input [0:0] random_run,  // external
	(* external, signal_type="plus-we" *)
	input [31:0] iva,  // external plus-we
	(* external, signal_type="plus-we" *)
	input [31:0] ivb,  // external plus-we
	// Everything else above is robust and pretty good.
	// Minor loss in efficiency from having two 32-bit registers when zero would do.
	// The following get hacked in.
	input iva_we,  // special for above, keep a trailing _we
	input ivb_we  // special for above, keep a trailing _we

);

reg iva_strobe=0; always @(posedge clk) iva_strobe <= iva_we; wire tt800_ena = random_run | iva_strobe;
reg ivb_strobe=0; always @(posedge clk) ivb_strobe <= ivb_we; wire tt800_enb = random_run | ivb_strobe;

tt800 tt800a(.clk(clk), .en(tt800_ena), .init(iva_strobe), .initv(iva), .y(rnda));
tt800 tt800b(.clk(clk), .en(tt800_enb), .init(ivb_strobe), .initv(ivb), .y(rndb));

endmodule

```

Being instantiated in another file (cav_mech.v):

```
`timescale 1ns / 1ns
`define AUTOMATIC_self
`define AUTOMATIC_decode
`define AUTOMATIC_noise_couple
`define AUTOMATIC_resonator
`define AUTOMATIC_prng
`include "cav_mech_auto.vh"

module cav_mech(
  input clk,
  input start_eig,
  input start_outer,
  output signed [17:0] mech_x,
  output signed [17:0] noise_eig_drive,
  input signed [17:0] eig_drive,
  // Output status
  output res_clip,
  `AUTOMATIC_self
);

`AUTOMATIC_decode

parameter n_mech_modes = 7;
parameter n_cycles = n_mech_modes * 2;

// Couple randomness to mechanical drive
wire signed [17:0] environment;  // filled in later
(*lb_automatic*)
outer_prod noise_couple  // auto
        (.clk(clk), .start(start_outer), .x(environment), .result(noise_eig_drive),
         `AUTOMATIC_noise_couple
);

// Instantiate the mechanical resonance computer
(*lb_automatic*)
resonator resonator // auto
  (.clk(clk), .start(start_eig),
         .drive(eig_drive),
         .position(mech_x), .clip(res_clip),
         `AUTOMATIC_resonator
);
// Pseudorandom number subsystem
wire [31:0] rnda, rndb;

(*lb_automatic*)
prng prng  // auto
  (.clk(clk), .rnda(rnda), .rndb(rndb),
   `AUTOMATIC_prng);

// Create a white noise term for environment
// This is a strangely-scaled CIC filter
// Each iteration adds a variance of 12, consuming 6 bits of PRNG
// Result has variance of n_cycles*12, mean of 0, possible peak n_cycles*7
// e.g. if n_cycles=14, std.dev.=12.96, peak=98, peak/rms=7.56
reg signed [11:0] noise_accum=0, noise_1=0, noise_out=0;
always @(posedge clk) begin
         noise_accum <= noise_accum + rndb[2:0] - rndb[5:3];
         if (start_eig) begin
                  noise_1 <= noise_accum;
                  noise_out <= noise_1-noise_accum;
         end
end
assign environment = noise_out;

endmodule // cav_mech
```

1. User settable registers are marked (* external *)
2. Those signals marked (* external *) are automatically decoded (bus independent) 
   externally (higher up the hierarchy) and routed through the module instantiation sites.
   Handling generation of write strobes, and arrays etc.
3. Instantiation sites that need routing these signals through are marked with (* lb_automatic *)
4. So far, we haven't ventured into modifying verilog files directly, and all the code generation
   is done within `*_auto.vh` files. This generated code is for macros that are sprinkled
   throughout the verilog. We are thinking of directly modifying verilog and avoid using macros and 
   header files with RTLIL/JSON.
5. With this, writing error-prone boiler plate code can be avoided, by automatically generating:
   a) Register decoder at \`AUTOMATIC\_decode in cav\_mech.v
   b) Instantiation site wiring with macros like \`AUTOMATIC\_prng and \`AUTOMATIC_resonator
   c) As the decoder is generated, addresses for these registers is automated as well.
6. Things not seen here:
   a) automated clock domain crossing for some of these registers
   b) generate loop handling
   c) comments exported as notes inside the register map etc.
   d) Register readback

There are several things to improve/iron out, to list a few:
1. The ugly \`define 's at the top of the file (there to ensure the code is valid verilog
   before code generation step). These are in place as we decided to not modify the source
   verilog directly.
2. Code with sprayed attributes and missing signals may not be readable for users unfamiliar
   with the automation. But readable code can be re-emitted.
3. Compatibility with VHDL.

Yosys makes this easier to deal with than iverilog or verilog-perl. Also promising to integrate
with ghdl with ghdl-yosys-plugin.

If you are interested in talking more about this, perhaps we can take the discussion off-line. Also, notifying @ldoolitt in case he wants to add his thoughts here.

whitequark:
Just to be clear--I was always in favor of the functionality, I was only concerned about testing the Verilog backend.

clairexen:
I, too, am concerned about testing of the Verilog back-end changes. Could we maybe factor the Verilog back-end changes into a separate pull request and merge the rest now?

jersey99:
> Could we maybe factor the Verilog back-end changes into a separate pull request and merge the rest now?
@clairexen Let me do that. Also do note that I ran VlogHammer against changes in the verilog_backend. We can pick this up at the other PR.

jersey99:
@clairexen @whitequark verilog_backend changes were in a single commit. So I undid that commit and force pushed. Let me know if there is anything else you need from me. Once we clear this I will open the other PR.

PR #2007, opened 2020-04-28
[Draft] MIF file generator from Yosys INIT parameter for Intel memory cells

dh73:
Since Quartus requires either an hex file or a mif file, to be created as part of the synhtesized memories, for all families, this PR takes the Yosys `INIT` parameter and converts it to a Quartus MIF file.
A `PicoSoC` example has been used to test this pass as shown [here](https://twitter.com/dhdezr73/status/1175991636955938816).  Also, as a quick example, consider this module:
```verilog
`default_nettype none
`define _MEM_INIT_
module ram #(parameter DW=8, AW=10)
   (input  wire [DW-1:0]  i_din,
    input  wire [AW-1:0]  i_wain, i_rain,
    output logic [DW-1:0] o_dout,
    input  wire           i_clk, i_en);

`ifdef _ZERO_INIT_
   var logic [DW-1:0]     array [(2**AW)-1:0] /* synthesis syn_ramstyle = "MLAB"*/ = '{default:0};
`endif

`ifdef _MEM_INIT_
   logic [DW-1:0]     array [(2**AW)-1:0] /* synthesis syn_ramstyle = "MLAB"*/;
   initial $readmemh("f0.hex", array);
`endif

   always_ff @(posedge i_clk) begin
      if (i_en) array[i_wain] <= i_din;
   end

   always_ff @(posedge i_clk) o_dout <= array[i_rain];
endmodule // ram
```
with `f0.hex` as:
```
1FF
1DE
1AD
100
1FF
1DE
1AD
100
1FF
1DE
1AD
100
1FF
1DE
1AD
100
1FF
1DE
1AD
100
1FF
1DE
1AD
100
```
Both Quartus and MIFGen creates a MIF file with the same contents (left Yosys, right Quartus).
image

I'm creating this PR as a draft, because there is still some work needed in the testing side, and some other improvements that might be needed to cover all Intel memory blocks. This might be of interest to @ZirconiumX .

### DIFF
--- a/techlibs/intel/Makefile.inc
+++ b/techlibs/intel/Makefile.inc
@@ -1,5 +1,6 @@
 
 OBJS += techlibs/intel/synth_intel.o
+OBJS += techlibs/intel/intel_genmif.o
 
 $(eval $(call add_share_file,share/intel/common,techlibs/intel/common/m9k_bb.v))
 $(eval $(call add_share_file,share/intel/common,techlibs/intel/common/altpll_bb.v))
@@ -11,4 +12,3 @@ families := max10 arria10gx cyclonev cyclone10lp cycloneiv cycloneive
 $(foreach family,$(families), $(eval $(call add_share_file,share/intel/$(family),techlibs/intel/$(family)/cells_sim.v)))
 $(foreach family,$(families), $(eval $(call add_share_file,share/intel/$(family),techlibs/intel/$(family)/cells_map.v)))
 #$(eval $(call add_share_file,share/intel/cycloneive,techlibs/intel/cycloneive/arith_map.v))
-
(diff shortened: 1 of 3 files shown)


ZirconiumX:
Since Quartus accepts .hex files, I think this should use that format instead of .mif because it makes the pass notably more generic and potentially useful to other flows. Additionally, this pass looks for specifically `altsyncram` cells, which I'm moving away from in the name of compile time; rather than checking cell name, just use the user-selected modules.

dh73:
I'm not sure if hex file is the way to go. Native Quartus format is MIF, and anyway it will convert the hex file to MIF all the time. What I know and have been experimenting is, if Quartus cannot convert the hex file to the native format, it will just either eat the entire system RAM, or crash without any apparent reason. 
But the pass could handle both formats.

ZirconiumX:
Right, then I'll swallow my distaste for MIF. It still needs to be more general than `altsyncram` cells for it to be any use to me at all though.

mmicko:
Sorry I have to close this one since source location for files do not exist any more and that makes issues with CI.

Issue #2011, opened 2020-05-01
Yosys compiler configuration is fundamentally incorrect and very misleading


whitequark:
Yosys currently uses four variables in Makefiles whose names do not match their intended contents:

 * `$(CXX)` contains a path to universal C/C++ compiler frontend that runs in C mode
 * `$(CXXFLAGS)` contains flags for both the C and C++ code
 * `$(LD)` contains a path to the C++ compiler frontend
 * `$(LDFLAGS)` contains linker flags escaped with `-Wl,`

Not only the names are misleading and differ from established use (e.g. any autoconf-based project expects *linker* flags in `LDFLAGS`, not *compiler frontend* flags), but they also cause practical problems.

## Steps to reproduce the issue

Configure Yosys build with `CXX := clang++` and `ENABLE_ABC := 1`.

## Expected behavior

abc build succeeds.

## Actual behavior

abc build breaks because Yosys implicitly assumes that `$(CXX)` expands to a C compiler frontend.

whitequark:
Fixed by @RCoeurjoly! :tada:

(It only took around four years since I filed the issue.)

PR #2012, opened 2020-05-01
Fix WASI builds with abc enabled

whitequark:
This PR works around #2011 and updates ABC to a more recent commit with WASI support.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -136,7 +136,7 @@ bumpversion:
 # is just a symlink to your actual ABC working directory, as 'make mrproper'
 # will remove the 'abc' directory and you do not want to accidentally
 # delete your work on ABC..
-ABCREV = d14acd8
+ABCREV = fd2c9b1
 ABCPULL = 1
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -281,13 +281,13 @@ yosys.html: misc/yosys.html
 
 else ifeq ($(CONFIG),wasi)
 ifeq ($(WASI_SDK),)
-CXX = clang++
+CXX = clang
```
eddiehung, line 284: Why this change?

whitequark: Because the Yosys build system is [fundamentally broken](https://github.com/YosysHQ/yosys/issues/2011) at the moment. All the other configurations alo have `CXX = <c-compiler-frontend>` in them.

eddiehung: Ah, I see this is the default:
https://github.com/YosysHQ/yosys/blob/ff7a8d0e1e1e6fab3e508bb8c237502ef570132b/Makefile#L185

PR #2016, opened 2020-05-03
qbfsat: Add `-solver` option and allow choice of Z3 or Yices, making Yices the default.

boqwxp:
Overview:
----
This PR adds support for using Yices in exists-forall mode to `yosys-smtbmc` and `qbfsat`.  This requires some tweaks like ensuring the logic specification is only `BV` and appears before any uses of operators or types from that logic.

This PR also makes Yices the default solver, replacing Z3.  This is because I have [found](https://github.com/boqwxp/yosys-examples/tree/master/example1) that Yices is *much* faster than Z3.  In this example, Yices found a solution to a bisection iteration problem in ~6m on my machine (~3m when linking Yices with `libtcmalloc`), using no more than 400MB of memory.  Z3, on the other hand, took approximately the same amount of time to gobble up 8GB of memory and fail.  While this is only one data point, it agrees with my experience that Yices is (often by far) the most performant exists-forall `BV` solver.

Depends:
----
- [x] [Yices](https://github.com/SRI-CSL/yices2) >= git commit 186434ba7212ecc436729152e212a30c08c65b06
- [x] PR #1996 
- [x] PR #2015

### DIFF
--- a/backends/smt2/smtio.py
+++ b/backends/smt2/smtio.py
@@ -172,7 +172,7 @@ class SmtIo:
             self.unroll = False
 
         if self.solver == "yices":
-            if self.noincr:
+            if self.noincr or self.forall:
                 self.popen_vargs = ['yices-smt2'] + self.solver_opts
             else:
                 self.popen_vargs = ['yices-smt2', '--incremental'] + self.solver_opts
@@ -232,16 +232,20 @@ class SmtIo:
             if self.logic_uf: self.logic += "UF"
             if self.logic_bv: self.logic += "BV"
             if self.logic_dt: self.logic = "ALL"
+            if self.solver == "yices" and self.forall: self.logic = "BV"
 
         self.setup_done = True
 
-        for stmt in self.info_stmts:
-            self.write(stmt)
+        if self.forall and self.solver == "yices":
+            self.write("(set-option :yices-ef-max-iters 1000000000)")
 
         if self.produce_models:
             self.write("(set-option :produce-models true)")
 
         self.write("(set-logic %s)" % self.logic)
+
+        for stmt in self.info_stmts:
+            self.write(stmt)
 
     def timestamp(self):
         secs = int(time() - self.start_time)
--- a/backends/smt2/smtio.py
+++ b/backends/smt2/smtio.py
@@ -236,6 +236,9 @@ class SmtIo:
 
         self.setup_done = True
 
+        for stmt in self.info_stmts:
+            self.write(stmt)
+
         if self.forall and self.solver == "yices":
             self.write("(set-option :yices-ef-max-iters 1000000000)")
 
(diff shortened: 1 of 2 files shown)


backends/smt2/smtio.py
```diff
@@ -232,17 +232,21 @@ def setup(self):
...
             if self.logic_bv: self.logic += "BV"
             if self.logic_dt: self.logic = "ALL"
+            if self.solver == "yices" and self.forall: self.logic = "BV"
 
         self.setup_done = True
 
-        for stmt in self.info_stmts:
-            self.write(stmt)
+        if self.forall and self.solver == "yices":
+            self.write("(set-option :yices-ef-max-iters 1000000000)")
```
boqwxp, line 240: Yices has a default exists-forall mode iteration limit of 100.  This is not nearly enough to solve pretty much any problem I've ever generated.  I set the iteration limit to a billion here because we don't care how many iterations the solver can do.  At most we care about limiting the time the solver can take, but that is a separate issue and is addressed in another forthcoming PR.

passes/sat/qbfsat.cc
```diff
@@ -280,84 +333,149 @@ void assume_miter_outputs(RTLIL::Module *module, const QbfSolveOptions &opt) {
...
+				success = value.as_const().as_int();
+				log("Problem is satisfiable with %s = %d.\n", wire_to_optimize_name.c_str(), success);
+				Pass::call(design, "design -pop");
+				module = design->module(module_name);
+
+				//sometimes this happens if we get an 'unknown' or timeout
+				if (!maximize && success < failure)
+					break;
+				else if (maximize && success > failure)
+					break;
```
boqwxp, line 453: This is basically to accommodate a forthcoming PR adding a timeout functionality.

Generally speaking, as the problem gets "harder" (i.e. the threshold approaches the true optimum), the solver takes longer.  So, if you keep a static per-iteration timeout, you might expect this could never happen.  However, the solver time depends on many factors, including random choices, and it's totally possible that the solver is able to find a more optimal value--better than `cur_thresh`--after failing (perhaps timing out) on a more relaxed constraint.

backends/smt2/smtio.py
```diff
@@ -232,17 +232,21 @@ def setup(self):
...
 
         if self.produce_models:
             self.write("(set-option :produce-models true)")
 
         self.write("(set-logic %s)" % self.logic)
 
+        for stmt in self.info_stmts:
+            self.write(stmt)
```
clairexen, line 248: Why move the info statements to after `set-logic`?

boqwxp: I was having trouble with yices requiring `(set-logic)` before any types were used (e.g. `BitVec` in `(define-sort |demo9_s| () (_ BitVec 17))`) and so tried to put `(set-logic)` as high up as possible.  But you're right that this is unnecessary and undesirable, and I will move it back.

Issue #210, opened 2016-08-21
Bug in Verilog frontend


adamwalker:
I believe that Yosys (git SHA a93fcec) fails to handle the following example correctly:
```
module Main_fir_sfir(result);
  output signed [9:0] result;
  
  wire signed [9:0] intermediate [0:2];
  
  function integer depth2Index;
    input integer depth;
  
    depth2Index = depth;
  endfunction
  
  assign intermediate[depth2Index(2)] = 0;  
  assign result = intermediate[2];
endmodule
```
Specifically, when I show the schematic using the show command, intermediate[2] is not driven, but it should be driven by the second to last line of the module. As a result, when I optimise it with opt_clean -purge, the whole circuit disappears.

This is a simplification of the Verilog generated by the [Clash](http://www.clash-lang.org/) compiler for the FIR filter example on that page.

It works correctly if I replace the depthToIndex(2) call with the literal 2. i.e. I remove the depthToIndex function call on the second to last line.

cliffordwolf:
Thanks for the bug report and test case. That problem is fixed now in commit 82a4a02.

The mechanism in Yosys that handles Verilog arrays that are not describing memories (mem2reg) is complicated and there are many strange corner cases. At the same time it's not very well tested because most (hand written) Verilog code avoids constructs like that. I would not be surprised if you'd find more bugs there..

adamwalker:
Thanks for the quick fix!
Looking at the schematic, it appears to be synthesizing the full FIR example correctly now.

PR #2018, opened 2020-05-03
smtbmc and qbfsat: Add timeout option to set solver timeouts for Z3, Yices, and CVC4.

boqwxp:
Overview:
----
This PR adds support to `yosys-smtbmc` and `qbfsat` for setting solver timeouts.

This is mostly straightforward, ~but `yices-smt2` currently has [a bug](https://github.com/SRI-CSL/yices2/issues/210) that renders the proper command line argument mechanism for specifying a timeout ineffective.~

~The draft PR includes an additional commit to what is intended to merge that works around this issue by using the `timeout` program.  Obviously this is not portable and this workaround commit will be removed once the Yices issue is fixed.  Until then, the commit exists to help test Yosys and `yosys-smtbmc` behavior and the PR will remain as a draft.~

**Update**: Yices has been fixed and the workaround commit has been removed.  This PR is now ready for review.

Depends:
----
- [x] ~Fix for [Yices issue #210](https://github.com/SRI-CSL/yices2/issues/210)~ Yices >= git commit dc13a01fb366741b1b9221126b3ad8934e704aad
- [x] PR #1996 
- [x] PR #2015 
- [x] PR #2016
- [x] PR #2017

### DIFF
--- a/backends/smt2/smtio.py
+++ b/backends/smt2/smtio.py
@@ -121,6 +121,7 @@ class SmtIo:
         self.logic_bv = True
         self.logic_dt = False
         self.forall = False
+        self.timeout = 0
         self.produce_models = True
         self.smt2cache = [list()]
         self.p = None
@@ -135,6 +136,7 @@ class SmtIo:
             self.debug_file = opts.debug_file
             self.dummy_file = opts.dummy_file
             self.timeinfo = opts.timeinfo
+            self.timeout = opts.timeout
             self.unroll = opts.unroll
             self.noincr = opts.noincr
             self.info_stmts = opts.info_stmts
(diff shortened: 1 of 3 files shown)


backends/smt2/smtio.py
```diff
@@ -445,6 +468,10 @@ def info(self, stmt):
 
         fields = stmt.split()
 
+        if fields[1] == "yosys-smt2-timeout":
+            self.timeout = int(fields[2])
+            assert self.timeout > 0
```
clairexen, line 473: I'm not convinced that passing a timeout through the SMT file is a sensible thing to do. We don't do that for other similar parameters, such as which SMT solver to use.

boqwxp: Fair enough.  It already works fine without it, so it's superfluous, anyway.

PR #2022, opened 2020-05-04
Avoid switch fall-through warnings

Xiretza:
Add a few comments and an unreachable `break` in cxxrtl to silence implicit fallthrough warnings. For some reason, the g++9 configuration on travis doesn't allow comments to [silence the warnings](https://travis-ci.org/github/YosysHQ/yosys/jobs/682632651#L661), even though this [should be enabled](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html) by `-Wextra`. Not that this is too big of a deal, but specifying the warning option explicitly might just fix it.

### DIFF
--- a/backends/cxxrtl/cxxrtl.cc
+++ b/backends/cxxrtl/cxxrtl.cc
@@ -2346,16 +2346,22 @@ struct CxxrtlBackend : public Backend {
 			case 6:
 				worker.max_opt_level = true;
 				worker.run_proc_flatten = true;
+				YS_FALLTHROUGH
 			case 5:
 				worker.run_opt_clean_purge = true;
+				YS_FALLTHROUGH
 			case 4:
 				worker.localize_public = true;
+				YS_FALLTHROUGH
 			case 3:
 				worker.elide_public = true;
+				YS_FALLTHROUGH
 			case 2:
 				worker.localize_internal = true;
+				YS_FALLTHROUGH
 			case 1:
 				worker.elide_internal = true;
+				YS_FALLTHROUGH
 			case 0:
 				break;
 			default:
--- a/backends/cxxrtl/cxxrtl.cc
+++ b/backends/cxxrtl/cxxrtl.cc
@@ -1943,13 +1943,13 @@ struct CxxrtlWorker {
 						case RTLIL::STa:
 							break;
 
+						case RTLIL::STg:
+							log_cmd_error("Global clock is not supported.\n");
+
 						// Handling of init-type sync rules is delegated to the `proc_init` pass, so we can use the wire
 						// attribute regardless of input.
 						case RTLIL::STi:
 							log_assert(false);
-
-						case RTLIL::STg:
-							log_cmd_error("Global clock is not supported.\n");
 					}
 			}
 
--- a/kernel/yosys.h
+++ b/kernel/yosys.h
@@ -155,6 +155,16 @@ extern Tcl_Obj *Tcl_ObjSetVar2(Tcl_Interp *interp, Tcl_Obj *part1Ptr, Tcl_Obj *p
 #  define YS_NORETURN
 #endif
 
+#if __cplusplus >= 201703L
+#  define YS_FALLTHROUGH [[fallthrough]];
+#elif defined(__GNUC__)
+#  define YS_FALLTHROUGH [[gnu::fallthrough]];
+#elif defined(__clang__)
+#  define YS_FALLTHROUGH [[clang::fallthrough]];
+#else
+#  define YS_FALLTHROUGH
+#endif
+
 YOSYS_NAMESPACE_BEGIN
 
 // Note: All headers included in hashlib.h must be included
--- a/backends/firrtl/firrtl.cc
+++ b/backends/firrtl/firrtl.cc
@@ -343,7 +343,7 @@ struct FirrtlWorker
 				switch (dir) {
 					case FD_INOUT:
 						log_warning("Instance port connection %s.%s is INOUT; treating as OUT\n", cell_type.c_str(), log_signal(it->second));
-						/* FALLTHRU */
+						YS_FALLTHROUGH
 					case FD_OUT:
 						sourceExpr = firstName;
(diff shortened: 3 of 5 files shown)


Makefile
```diff
@@ -81,7 +81,7 @@ all: top-all
 YOSYS_SRC := $(dir $(firstword $(MAKEFILE_LIST)))
 VPATH := $(YOSYS_SRC)
 
-CXXFLAGS := $(CXXFLAGS) -Wall -Wextra -ggdb -I. -I"$(YOSYS_SRC)" -MD -D_YOSYS_ -fPIC -I$(PREFIX)/include
+CXXFLAGS := $(CXXFLAGS) -Wall -Wextra -Wimplicit-fallthrough=3 -ggdb -I. -I"$(YOSYS_SRC)" -MD -D_YOSYS_ -fPIC -I$(PREFIX)/include
```
Xiretza, line 84: Not even this seems to work: https://travis-ci.org/github/YosysHQ/yosys/jobs/683030101#L850, even though it's clearly marked as fallthrough: https://github.com/YosysHQ/yosys/blob/584780d776c92bc91731dbc2710dd8d9a624dc70/backends/firrtl/firrtl.cc#L344-L347

Any ideas?

kernel/yosys.h
```diff
@@ -155,6 +155,14 @@ extern Tcl_Obj *Tcl_ObjSetVar2(Tcl_Interp *interp, Tcl_Obj *part1Ptr, Tcl_Obj *p
 #  define YS_NORETURN
 #endif
 
+#if __cplusplus >= 201703L
+#  define YS_FALLTHROUGH [[fallthrough]];
+#elif defined(__GNUC__)
+#  define YS_FALLTHROUGH [[gnu::fallthrough]];
+#elif defined(__clang__)
+#  define YS_FALLTHROUGH [[clang::fallthrough]];
```
whitequark, line 163: This breaks on pre-C++17 MSVC.

Xiretza: Doh, forgot to declare a fallback. Should it default to empty string or to a `/* fall through */` comment?

Xiretza: As far as I can see the only compiler that will be affected by the fallback is MSVC, which doesn't have an implicit fallthrough warning anyway: https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warnings-by-compiler-version. Definining it as empty by default should be fine then.

backends/cxxrtl/cxxrtl.cc
```diff
@@ -1947,6 +1947,8 @@ struct CxxrtlWorker {
 						// attribute regardless of input.
 						case RTLIL::STi:
 							log_assert(false);
+							// unreachable, but compiler still warns about implicit fall-through
```
whitequark, line 1950: This seems like a bug we should fix one way or another. Either figure out why the warning persists, or add `log_unreachable()`.

Xiretza: With `NDEBUG=1` this warning is completely legit, since `log_assert` becomes a no-op. Maybe using `log_abort` would be a better choice here?

whitequark: Oh, I guess you could order this case the last to avoid the fallthrough.

Xiretza: That works too, done.

cliffordwolf:
jfyi, I have pushed a bugfix for this PR in 8ec3b6d to avoid the following compiler warning with clang.

```
claire@hopper:~/Work/yosys$ make PRETTY=0 frontends/ast/genrtlil.o
[Makefile.conf] INSTALL_SUDO := sudo
[Makefile.conf] ENABLE_VERIFIC:= 1
mkdir -p frontends/ast/
clang -o frontends/ast/genrtlil.o -c   -Wall -Wextra -ggdb -I. -I"./" -MD -D_YOSYS_ -fPIC -I/usr/local/include -std=c++11 -Os -DYOSYS_ENABLE_READLINE  -DYOSYS_ENABLE_PLUGINS -DYOSYS_ENABLE_GLOB -DYOSYS_ENABLE_ZLIB -I/usr/include/tcl8.6 -DYOSYS_ENABLE_TCL -DYOSYS_ENABLE_ABC -I/usr/local/src/verific_lib/verilog -I/usr/local/src/verific_lib/vhdl -I/usr/local/src/verific_lib/database -I/usr/local/src/verific_lib/util -I/usr/local/src/verific_lib/containers -I/usr/local/src/verific_lib/hier_tree -DYOSYS_ENABLE_VERIFIC -DYOSYS_ENABLE_COVER frontends/ast/genrtlil.cc
frontends/ast/genrtlil.cc:933:3: warning: unknown attribute 'fallthrough' ignored [-Wunknown-attributes]
                YS_FALLTHROUGH
                ^
./kernel/yosys.h:161:28: note: expanded from macro 'YS_FALLTHROUGH'
#  define YS_FALLTHROUGH [[gnu::fallthrough]];
                           ^
frontends/ast/genrtlil.cc:1029:3: warning: unknown attribute 'fallthrough' ignored [-Wunknown-attributes]
                YS_FALLTHROUGH
                ^
./kernel/yosys.h:161:28: note: expanded from macro 'YS_FALLTHROUGH'
#  define YS_FALLTHROUGH [[gnu::fallthrough]];
                           ^
frontends/ast/genrtlil.cc:1826:3: warning: unknown attribute 'fallthrough' ignored [-Wunknown-attributes]
                YS_FALLTHROUGH
                ^
./kernel/yosys.h:161:28: note: expanded from macro 'YS_FALLTHROUGH'
#  define YS_FALLTHROUGH [[gnu::fallthrough]];
                           ^
3 warnings generated.



claire@hopper:~/Work/yosys$ clang --version
clang version 6.0.0-1ubuntu2 (tags/RELEASE_600/final)
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

whitequark:
Ah oops, I missed this. Thanks for the fix!

Issue #2021, opened 2020-05-04
Condition without statement syntax error


towoe:
## Steps to reproduce the issue

```verilog
module empty ();
	parameter on = 0;
	if (on);
endmodule
```

## Expected behavior

I am not completely sure if this is actually a valid Verilog syntax, but from the spec I read the following: "if (expression) statement_or_null" so I am assuming that with the "null" part is should be legal to have a condition without a statement.

## Actual behavior

1. Executing Verilog-2005 frontend: empty.v
Parsing Verilog input from `empty.v' to AST representation.
empty.v:3: ERROR: syntax error, unexpected ';'

Sorry if this is just a question and not a bug report and thanks for the support!

dh73:
Hello, 
In the IEEE 1800-2015,  `A.6.6 Conditional statements`, shows the same syntax as you pointed out here. In the `if (on);` statement, the `;` is the null construct. However, the example is not correct, since conditional expressions can be used either in procedural assignments or generated instantiations (generate if). In that example, conditional statement is not under any of those.

This is correct, Yosys generates an empty module as expected:
```Verilog
module test ();
        localparam y = 1;
        always @(*) 
                if (y) ;
endmodule
```
Another example with a case statement:
```Verilog
module test (input a, output y);
        always @(*)
                case (a)
                        1'b0: y = 1'b1;
                        default: ;
                endcase
endmodule
```
Your example can be modeled under a `generate-if` construct. Although it will be an empty module instantiating no cell.

nakengelhardt:
Tangentially, `statement_or_null` is defined in A.6.4 as `statement | { attribute_instance } ;` but if I try to add the optional attribute it doesn't work:
```
module test ();
	localparam y = 1;
	always @(*)
		if (y) (* foo *) ;
endmodule
```
```
1. Executing Verilog-2005 frontend: stmt.v
Parsing Verilog input from `stmt.v' to AST representation.
stmt.v:4: ERROR: syntax error, unexpected ';'
```
Am I doing it wrong, or is there some restriction on attributes I didn't notice?

towoe:
Thanks @dh73 for the detailed reply!

The first example does work for me.

The example I posted was actually just an excerpt of something else. So my minimal example was not correct, but thanks for clarifying what is possible and correct.

But the following with the `generate` does not work as well. @dh73 should it?

```
module empty ();
	localparam on = 1;
	generate
		if (on);
	endgenerate
endmodule
```

The following does work
```
[...]

PR #2028, opened 2020-05-05
verilog: allow null gen-if then block

zachjs:
This addresses #2021. This wraps the then block of gen-ifs in an implicit block, similar to the strategy employed for procedural if statements. I am not sure if my strategy for "asserting" which branch is taken via the port bindings is reasonable. I am eager to receive any feedback you may have!

### DIFF
--- a/tests/various/gen_if_null.ys
+++ b/tests/various/gen_if_null.ys
@@ -1,0 +1,4 @@
+read_verilog gen_if_null.v
+select -assert-count 1 test/x
+select -assert-count 1 test/y
+select -assert-count 1 test/z
--- a/frontends/verilog/verilog_parser.y
+++ b/frontends/verilog/verilog_parser.y
@@ -2533,7 +2533,12 @@ gen_stmt:
 		ast_stack.back()->children.push_back(node);
 		ast_stack.push_back(node);
 		ast_stack.back()->children.push_back($3);
-	} gen_stmt_block opt_gen_else {
+		AstNode *block = new AstNode(AST_GENBLOCK);
+		ast_stack.back()->children.push_back(block);
+		ast_stack.push_back(block);
+	} gen_stmt_or_null {
+		ast_stack.pop_back();
+	} opt_gen_else {
 		SET_AST_NODE_LOC(ast_stack.back(), @1, @7);
 		ast_stack.pop_back();
 	} |
--- a/tests/various/gen_if_null.v
+++ b/tests/various/gen_if_null.v
@@ -1,0 +1,13 @@
+module test(x, y, z);
+	localparam OFF = 0;
+	generate
+		if (OFF) ;
+		else input x;
+		if (!OFF) input y;
+		else ;
+		if (OFF) ;
+		else ;
+		if (OFF) ;
+		input z;
+	endgenerate
+endmodule


tests/various/gen_if_null.ys
```diff
@@ -0,0 +1,3 @@
+read_verilog gen_if_null.v
+design -save read
+prep
```
eddiehung, line 3: These two lines aren't necessary for the test.
`design -save read` is copied from another test -- its purpose is to save a snapshot of the design under the `read` label so that we don't have to read it from disk again when we re-use it for a different test.
`prep` is part of synthesis, and isn't necessary for a test that just touches the Verilog frontend.

Amusingly, both iverilog and verilator give a warning:
```
$ iverilog -V
Icarus Verilog version 10.3 (stable) ()
<<SNIP>>
$ iverilog -s top gen_if_null.v 
gen_if_null.v:7: syntax error
gen_if_null.v:7: error: invalid module item.
gen_if_null.v:9: syntax error
gen_if_null.v:9: error: invalid module item.
```
```
$ verilator -V
Verilator 4.016 2019-06-16 rev UNKNOWN_REV
<<SNIP>>
$ verilator --lint-only --top-module top gen_if_null.v 
%Error: gen_if_null.v:5: syntax error, unexpected input
%Error: gen_if_null.v:6: syntax error, unexpected input
%Error: gen_if_null.v:11: syntax error, unexpected input
%Error: Exiting due to 3 error(s)
```

zachjs: I removed the `design -save read`. Thanks for explaining!

I think `prep`, or something like it is needed. In a prior implementation, I had made the frontend parse successfully, but the branches would be mis-associated. This test ensures the correct branch is taken by declaring the ports in the desired branches. It seems that yosys does not check the ports until I run `prep`. Perhaps there is a better choice for triggering this behavior?

eddiehung: Can you recall what you had in the prior implementation?

`read_verilog` is responsible for generating the AST and for elaborating the design into a netlist (RTLIL), which requires port knowledge. Without this PR, it fails on `read_verilog` with:
```
 ./yosys -q gen_if_null.v 
gen_if_null.v:4: ERROR: syntax error, unexpected ';'
```

zachjs: I believe the overall patch for the naive implementation is simply:
```patch
diff --git a/frontends/verilog/verilog_parser.y b/frontends/verilog/verilog_parser.y
index 4a5aba79..8faede80 100644
--- a/frontends/verilog/verilog_parser.y
+++ b/frontends/verilog/verilog_parser.y
@@ -2533,7 +2533,7 @@ gen_stmt:
                ast_stack.back()->children.push_back(node);
                ast_stack.push_back(node);
                ast_stack.back()->children.push_back($3);
-       } gen_stmt_block opt_gen_else {
+       } gen_stmt_or_null opt_gen_else {
                SET_AST_NODE_LOC(ast_stack.back(), @1, @7);
                ast_stack.pop_back();
        } |
```

When built with this patch, `read_verilog` on the test file runs without issue, but running `prep` yields:
```
2.1. Executing HIERARCHY pass (managing design hierarchy).
ERROR: Module `test' referenced in module `top' in cell `test' has only 2 ports, requested port 3.
```

eddiehung: Ooh, fascinating! Thanks for digging into it @zachjs -- it was more for my curiosity than it was doubting what you were doing.

So with your naive patch, it seems that `test` is elaborated into:
```
module test(y, z);
  input y;
  input z;
endmodule
```
which is a mismatch when it comes to `module test(x, y, z);` from the original input. I would have expected this to give an error, meaning you wouldn't need to resolve cell instantiations (done by the `hierarchy` pass) in order to see an error.

eddiehung: Rather than relying on an instantiation, you can also consider checking for the existence of ports explicitly:
```
select -assert-count 1 i:x
select -assert-count 1 i:y
select -assert-count 1 i:z
```

zachjs: Thanks @eddiehung! That test procedure makes a lot more sense.

PR #2031, opened 2020-05-05
Add extmodule support to firrtl backend

sahandKashani:
The current firrtl backend emits blackboxes as standard modules
with an empty body, but this causes the firrtl compiler to
optimize out entire circuits due to the absence of any drivers.

Yosys already tags blackboxes with a (*blackbox*) attribute, so this
commit just propagates this change to firrtl's syntax for blackboxes.

### DIFF
--- a/backends/firrtl/firrtl.cc
+++ b/backends/firrtl/firrtl.cc
@@ -392,7 +392,37 @@ struct FirrtlWorker
 		return result;
 	}
 
-	void run()
+	void emit_extmodule()
+	{
+		std::string moduleFileinfo = getFileinfo(module);
+		f << stringf("  extmodule %s: %s\n", make_id(module->name), moduleFileinfo.c_str());
+		vector<string> port_decls;
+
+		for (auto wire : module->wires())
+		{
+			const auto wireName = make_id(wire->name);
+			std::string wireFileinfo = getFileinfo(wire);
+
+			// Maybe not needed?
+			if (wire->port_id)
+			{
+				if (wire->port_input && wire->port_output)
+				{
+					log_error("Module port %s.%s is inout!\n", log_id(module), log_id(wire));
+				}
+				port_decls.push_back(stringf("    %s %s: UInt<%d> %s\n", wire->port_input ? "input" : "output",
+						wireName, wire->width, wireFileinfo.c_str()));
+			}
+		}
+
+		for (auto str : port_decls) {
+			f << str;
+		}
+
+		f << stringf("\n");
+	}
+
+	void emit_module()
 	{
 		std::string moduleFileinfo = getFileinfo(module);
 		f << stringf("  module %s: %s\n", make_id(module->name), moduleFileinfo.c_str());
@@ -1076,6 +1106,22 @@ struct FirrtlWorker
 
 		for (auto str : wire_exprs)
 			f << str;
+
+		f << stringf("\n");
+	}
+
+	void run()
+	{
+		// Blackboxes should be emitted as `extmodule`s in firrtl. Only ports are
+		// emitted in such a case.
+		if (module->get_blackbox_attribute())
+		{
+			emit_extmodule();
+		}
+		else
+		{
+			emit_module();
+		}
 	}
 };
 
--- a/backends/firrtl/firrtl.cc
+++ b/backends/firrtl/firrtl.cc
@@ -396,26 +396,23 @@ struct FirrtlWorker
 	{
 		std::string moduleFileinfo = getFileinfo(module);
 		f << stringf("  extmodule %s: %s\n", make_id(module->name), moduleFileinfo.c_str());
-		vector<string> port_decls;
+		vector<std::string> port_decls;
 
 		for (auto wire : module->wires())
 		{
 			const auto wireName = make_id(wire->name);
 			std::string wireFileinfo = getFileinfo(wire);
 
-			// Maybe not needed?
-			if (wire->port_id)
+			if (wire->port_input && wire->port_output)
 			{
-				if (wire->port_input && wire->port_output)
-				{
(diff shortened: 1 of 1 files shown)


backends/firrtl/firrtl.cc
```diff
@@ -392,7 +392,37 @@ struct FirrtlWorker
...
+
+		for (auto wire : module->wires())
+		{
+			const auto wireName = make_id(wire->name);
+			std::string wireFileinfo = getFileinfo(wire);
+
+			// Maybe not needed?
+			if (wire->port_id)
+			{
+				if (wire->port_input && wire->port_output)
```
whitequark, line 409: AFAIK if `port_input` or `port_output` are set you don't need to check `port_id`.

backends/firrtl/firrtl.cc
```diff
@@ -392,7 +392,37 @@ struct FirrtlWorker
 		return result;
 	}
 
-	void run()
+	void emit_extmodule()
+	{
+		std::string moduleFileinfo = getFileinfo(module);
+		f << stringf("  extmodule %s: %s\n", make_id(module->name), moduleFileinfo.c_str());
+		vector<string> port_decls;
```
boqwxp, line 399:
```suggestion
		vector<std::string> port_decls;
```

backends/firrtl/firrtl.cc
```diff
@@ -392,7 +392,37 @@ struct FirrtlWorker
...
+
+		for (auto wire : module->wires())
+		{
```
boqwxp, line 402: There is an opening brace on a new line here but not for the `for` on line 418.  Personally, I prefer to keep the opening brace on the same line, but either way, it should be consistent.

backends/firrtl/firrtl.cc
```diff
@@ -392,7 +392,37 @@ struct FirrtlWorker
...
+				if (wire->port_input && wire->port_output)
+				{
+					log_error("Module port %s.%s is inout!\n", log_id(module), log_id(wire));
+				}
+				port_decls.push_back(stringf("    %s %s: UInt<%d> %s\n", wire->port_input ? "input" : "output",
+						wireName, wire->width, wireFileinfo.c_str()));
+			}
+		}
+
+		for (auto str : port_decls) {
```
boqwxp, line 418: This should probably be:
```suggestion
		for (auto &str : port_decls) {
```
But I don't think it really makes a performance difference in this case.

backends/firrtl/firrtl.cc
```diff
@@ -392,7 +392,37 @@ struct FirrtlWorker
...
+			f << str;
+		}
+
+		f << stringf("\n");
```
boqwxp, line 422: You don't need `stringf()` if you're not using the format specifier.

```suggestion
		f << "\n";
```
(and on line 1110)

backends/firrtl/firrtl.cc
```diff
@@ -1076,6 +1106,22 @@ struct FirrtlWorker
...
+		// Blackboxes should be emitted as `extmodule`s in firrtl. Only ports are
+		// emitted in such a case.
+		if (module->get_blackbox_attribute())
+		{
+			emit_extmodule();
+		}
+		else
+		{
+			emit_module();
+		}
```
boqwxp, line 1124: OK, already covered by the review of line 402, but doesn't:
```suggestion
		if (module->get_blackbox_attribute())
			emit_extmodule();
		else
			emit_module();
```
just look a lot cleaner?

backends/firrtl/firrtl.cc
sahandKashani: Good to know. Thanks!

backends/firrtl/firrtl.cc
sahandKashani: Nice catch. My muscle memory put the brace on the same line as I also prefer it, but (most of) the rest of this source file had them on the following line, so I stuck with that convention.

backends/firrtl/firrtl.cc
sahandKashani: I agree that there is no format specifier here and calling this function doesn't help, but the rest of the code used `stringf` everywhere as a convention and I stuck to it. I can go ahead and change it everywhere if you think it would be necessary?

backends/firrtl/firrtl.cc
sahandKashani: Indeed, done.

backends/firrtl/firrtl.cc
boqwxp: I don't think it's necessary, no.  Frankly I don't think it makes much of a difference; I was just pointing it out.

sahandKashani:
@whitequark Do you think we can merge the PR? I believe it's pending your final approval.

whitequark:
> I believe it's pending your final approval.

I don't think I have the final say over the FIRRTL backend given that I've never even used FIRRTL, but if no one else steps up then I could probably take over maintaining this part of Yosys.

sahandKashani:
> I don't think I have the final say over the FIRRTL backend given that I've never even used FIRRTL, but if no one else steps up then I could probably take over maintaining this part of Yosys.

Sorry for the confusion. I meant that you were listed as one of the reviewers of the PR, hence why I was asking.

Do you know if there is a list of yosys maintainers anywhere? I understand that the firrtl backend doesn't see too many updates, but I think having a maintainer nonetheless would be great. Thanks for being open to taking this up!

whitequark:
Looks like Claire did the original work, then a few people updated it as needed.

I'm not comfortable taking over it at the moment because there are some issues I see (like the read-under-write behavior not matching RTLIL semantics) that I can't properly fix as I don't use FIRRTL yet. But I think I can merge this particular PR just fine.

PR #2034, opened 2020-05-06
Makefile: git fetch $(ABCURL) explicitly for local ABC checkout

eddiehung:
Should resolve issue described at https://github.com/YosysHQ/yosys/commit/ae409d4d81f3a15aba58d5144e68b565cf70b79a#commitcomment-38972956

CC: @Xiretza @boqwxp

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -722,7 +722,7 @@ ifneq ($(ABCREV),default)
 		test $(ABCPULL) -ne 0 || { echo 'REEBE: NOP abg hc gb qngr naq NOPCHYY frg gb 0 va Znxrsvyr!' | tr 'A-Za-z' 'N-ZA-Mn-za-m'; exit 1; }; \
 		echo "Pulling ABC from $(ABCURL):"; set -x; \
 		test -d abc || git clone $(ABCURL) abc; \
-		cd abc && $(MAKE) DEP= clean && git fetch origin master && git checkout $(ABCREV); \
+		cd abc && $(MAKE) DEP= clean && git fetch $(ABCURL) && git checkout $(ABCREV); \
 	fi
 endif
 	$(Q) rm -f abc/abc-[0-9a-f]*


Makefile
```diff
@@ -721,8 +721,14 @@ ifneq ($(ABCREV),default)
...
-		test -d abc || git clone $(ABCURL) abc; \
-		cd abc && $(MAKE) DEP= clean && git fetch origin master && git checkout $(ABCREV); \
+		if test -d abc; then \
+			cd abc; \
+			git remote set-url origin $(ABCURL); \
+		else \
+			git clone $(ABCURL) abc; \
+			cd abc; \
+		fi; \
+		$(MAKE) DEP= clean && git fetch origin master && git checkout $(ABCREV); \
```
Xiretza, line 731: This might call `make clean` on yosys if one the `cd` above fails for whatever reason.

boqwxp: Good point.  I guess just adding `cd ..; \` after the `git remote` command, removing `cd abc; \` from the `else`, and reverting the `$(MAKE)` line should do the trick.

Xiretza: Alternatively, `git -C abc remote ...`.

eddiehung: Oooh I never knew about the `-C` option! Thanks for the suggestion; adopted in 3c0d2af.

boqwxp: Me either :).  TIL.

Makefile
```diff
@@ -721,7 +721,11 @@ ifneq ($(ABCREV),default)
 	$(Q) if ! (cd abc && rev="`git rev-parse $(ABCREV)`" && test "`git rev-parse HEAD`" == "$$rev"); then \
 		test $(ABCPULL) -ne 0 || { echo 'REEBE: NOP abg hc gb qngr naq NOPCHYY frg gb 0 va Znxrsvyr!' | tr 'A-Za-z' 'N-ZA-Mn-za-m'; exit 1; }; \
 		echo "Pulling ABC from $(ABCURL):"; set -x; \
-		test -d abc || git clone $(ABCURL) abc; \
+		if test -d abc; then \
+			git -C abc remote set-url origin $(ABCURL); \
+		else \
+			git clone $(ABCURL) abc; \
+		fi; \
 		cd abc && $(MAKE) DEP= clean && git fetch origin master && git checkout $(ABCREV); \
```
whitequark, line 729: What about using `git fetch $(ABCURL) master && git checkout $(ABCREV)` instead? That's a lot cleaner than clobbering user's choice of remote.

eddiehung: I would expect that the `abc/` subdir would be checked out and maintained by Yosys. For users that want to set their own remote, I would also expect that they would set `ABCREV=default` to block this auto-updating functionality.

eddiehung: Nevertheless, your suggestion is cleaner, and adopted in 
8eb98b1. The only change was that, since the YosysHQ is not on the master branch, I've resorted to getting all commits as opposed to those just on master: https://github.com/YosysHQ/yosys/blob/8eb98b12c726149ee64c3901005eec4d527f12ad/Makefile#L725

whitequark: > I would expect that the `abc/` subdir would be checked out and maintained by Yosys. For users that want to set their own remote, I would also expect that they would set `ABCREV=default` to block this auto-updating functionality.

I agree in principle, but there are some sentences in the docs or the README (I don't recall where exactly) that suggest being careful with the `abc` directory while working on Yosys lest the Yosys build system wipes your work. It seems like we should rather avoid making things worse.

whitequark: And it's very easy to accidentally run `make` on the Yosys source tree in somewhere other than the place where your `Makefile.conf` lives.

PR #2044, opened 2020-05-11
verilog: allow attributes on behavioural statements (including null statement)

### LINKED ISSUE #2037: Syntax error for attribute in front of statement
While looking at #2021 I noticed that the verilog standard allows attributes in more places than yosys currently accepts, such as e.g. in front of any statement (including the null statement).
```
module test ();
	localparam y = 1;
	always @(*)
		if (y) (* foo *) ;
endmodule
```
```
1. Executing Verilog-2005 frontend: stmt.v
Parsing Verilog input from `stmt.v' to AST representation.
stmt.v:4: ERROR: syntax error, unexpected ';'
```
As attributes have no defined meaning, we should accept and ignore any unknown attribute in any of the allowed locations.

eddiehung:
Fixes #2037. 

To support the null statement I had to remove the `gen_stmt_or_null` rule as modifying it to allow attributes created a shift/reduce conflict.

### DIFF
--- a/frontends/verilog/verilog_parser.y
+++ b/frontends/verilog/verilog_parser.y
@@ -2230,7 +2230,6 @@ behavioral_stmt:
 	non_opt_delay behavioral_stmt |
 	attr simple_behavioral_stmt ';' |
 	attr ';' {
-		log_file_warning(current_filename, get_line_num(), "Attribute(s) attached to null statement. Ignoring.\n");
 		free_attr($1);
 	} |
 	attr hierarchical_id {
@@ -2534,7 +2533,6 @@ module_gen_body:
 gen_stmt_or_module_body_stmt:
 	gen_stmt | module_body_stmt |
 	attr ';' {
-		log_file_warning(current_filename, get_line_num(), "Attribute(s) attached to null statement. Ignoring.\n");
 		free_attr($1);
 	};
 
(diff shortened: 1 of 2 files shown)


nakengelhardt:
Sorry, the issue I opened wasn't super clear on this, I meant to go through the standard and find all the places that it's allowed to have an attribute, but the short version is "practically anywhere":

>  An `attribute_instance` can appear in the Verilog description as a prefix attached to a declaration, a module item, a statement, or a port connection.

The null statement isn't special at all here, for instance the following is afaict perfectly standard-compliant:
```
module test ();
	localparam y = 1;
	reg x = 1'b0;
	always @(*) begin
		if (y)
			(* foo *) x <= 1'b1;
	end
endmodule
```

I also don't think we should handle attributes on null statements any differently than elsewhere. If it wouldn't cause a warning in front of `x <= 1'b1;`, it shouldn't cause a warning in front of `;` either - we don't know what `(* foo *)` means, so it was probably intended for a different tool, and who knows if it makes perfect sense to have it in front of an empty statement there. (There's a discussion to be had about general warning philosophy here, maybe.)

eddiehung:
The null statement is special because we discard it -- there is nothing to attach its atteibutes to in the AST (and subsequently in RTLIL). In your example, the attribute should be preserved correctly on the assignment. The worry I have is accidentally misplaced attributes that may not be obvious, for example, through macro expansion or the like.

eddiehung:
You're right again @nakengelhardt! I've added your testcase which I can't fix without introducing shift/reduce conflicts, so I'm switching this over to a draft PR while we work through it.

mmicko:
@eddiehung please do rebase to latest, since there were more changes to logger.cc

PR #2045, opened 2020-05-11
verilog: error if no direction given for task arguments, default to input in SV mode

### LINKED ISSUE #2042: Verilog Parser Failure
Yosys Version: 06104249406972de01d0360df63a32cafcdf2ec5

Run the following command: yosys -f "verilog -sv " -p "synth_ice40 -top Task_Test_Top -json Task_Test_Top.json" Task_Test_Top.v

with the below code:

**Task_Test_Top.v**
```
module Task_Test_Top
(
);

    task SomeTaskName(a)
    endtask

endmodule
```
## Expected behavior

Some sort of syntax error message.

## Actual behavior

Yosys parser fails with the below error message:

**ERROR**: Internal error - should not happen - no AST_WIRE node.

eddiehung:
Fixes #2042.

~Supporting this is left as a future enhancement.~

### DIFF
--- a/tests/verilog/bug2042.ys
+++ b/tests/verilog/bug2042.ys
@@ -1,4 +1,4 @@
-logger -expect error "Non-ANSI style task/function arguments not currently supported" 1
+logger -expect error "task/function argument direction missing" 1
 read_verilog <<EOT
 module Task_Test_Top
 (
@@ -9,4 +9,3 @@ module Task_Test_Top
 
 endmodule
 EOT
-
--- a/tests/verilog/bug2042.ys
+++ b/tests/verilog/bug2042.ys
@@ -1,0 +1,12 @@
+logger -expect error "Non-ANSI style task/function arguments not currently supported" 1
+read_verilog <<EOT
+module Task_Test_Top
+(
+);
+
+    task SomeTaskName(a)
+    endtask
+
+endmodule
+EOT
+
--- a/Makefile
+++ b/Makefile
@@ -780,6 +780,7 @@ test: $(TARGETS) $(EXTRA_TARGETS)
 	+cd tests/arch/intel_alm && bash run-test.sh $(SEEDOPT)
 	+cd tests/rpc && bash run-test.sh
 	+cd tests/memfile && bash run-test.sh
+	+cd tests/verilog && bash run-test.sh
 	@echo ""
 	@echo "  Passed \"make test\"."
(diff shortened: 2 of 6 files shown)


whitequark:
What does "non-ANSI" actually mean?

nakengelhardt:
Good question. Google leads me to believe that "ANSI-style" means putting the input/output ports of the task on the same line, as with modules, is that what you meant @eddiehung? Because that is supported, the thing that's not supported is SystemVerilog's default direction that allows you to omit the "input" keyword.

On the same example iverilog reports `task.v:1: error: missing task/function port direction.`, which IMO is a clearer error message.

tests/verilog/bug2042.ys
```diff
@@ -0,0 +1,12 @@
+logger -expect error "Non-ANSI style task/function arguments not currently supported" 1
+read_verilog <<EOT
+module Task_Test_Top
+(
+);
+
+    task SomeTaskName(a)
```
nakengelhardt, line 7: I would suggest fixing the other syntax error (missing `;` at the end of the line) in this test, in case we change the frontend in a way that finds the other error first at some point...

eddiehung:
> Good question. Google leads me to believe that "ANSI-style" means putting the input/output ports of the task on the same line, as with modules, is that what you meant @eddiehung? Because that is supported, the thing that's not supported is SystemVerilog's default direction that allows you to omit the "input" keyword.

Yes that is what I meant. However, I wasn't aware that SV treats this case as an input implicitly. I'd have to look again to see if that can be supported.

> On the same example iverilog reports `task.v:1: error: missing task/function port direction.`, which IMO is a clearer error message.

I would argue that the user may not be omitting the port direction entirely, as it can still appear later in the task/func body as for modules.

nakengelhardt:
> I would argue that the user may not be omitting the port direction entirely, as it can still appear later in the task/func body as for modules.

I don't think that is legal; the standard specifies two options, "ansi-style":
```
task SomeTaskName(input a);
endtask
```
or "non-ansi-style":
```
task SomeTaskName;
  input a;
endtask
```
but I don't think it allows (what I read what you are saying as):
```
task SomeTaskName(a);
  input a;
endtask
```

eddiehung:
@nakengelhardt You're absolutely right; I'm so sorry! I've updated the error message, and default it to `input` when in SV mode now.

PR #2046, opened 2020-05-13
Extend YS_DEBUGTRAP to MacOS.

PeterCrozier:
PR #2000 was only implemented for Windows and Linux.  This enables the Unix code under MacOS where the signal type is different.

### DIFF
--- a/kernel/log.h
+++ b/kernel/log.h
@@ -97,15 +97,12 @@ YOSYS_NAMESPACE_BEGIN
 // if a debugger is attached, and does nothing otherwise.
 #if defined(_WIN32)
 # define YS_DEBUGTRAP_IF_DEBUGGING do { if (IsDebuggerPresent()) DebugBreak(); } while(0)
-#elif defined(__unix__)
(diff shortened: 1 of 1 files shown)


kernel/log.h
```diff
@@ -103,6 +103,9 @@ YOSYS_NAMESPACE_BEGIN
 # define YS_DEBUGTRAP_IF_DEBUGGING do { \
 		sighandler_t old = signal(SIGTRAP, SIG_IGN); raise(SIGTRAP); signal(SIGTRAP, old); \
 	} while(0)
+#elif defined(__APPLE__) && defined(__MACH__)
+// MacOS
+#define YS_DEBUGTRAP_IF_DEBUGGING { sig_t old = signal(SIGTRAP, SIG_IGN); raise(SIGTRAP); signal(SIGTRAP, old); }
```
whitequark, line 108: Just change the common *nix code to use `auto old = ...`

PeterCrozier: Good idea.

Issue #1772, opened 2020-03-14
Deprecate and then remove support for gcc-4.8


boqwxp:
GCC 4.8 is a [very old](https://gcc.gnu.org/releases.html) compiler not used by [Ubuntu 16.04 LTS](https://packages.ubuntu.com/xenial/gcc) or [Debian 8](https://packages.debian.org/jessie/gcc).  [CentOS/RHEL 7](https://access.redhat.com/solutions/19458) seems to be the only currently-in-use distro that has gcc-4.8 as its default compiler.

However, there is an officially-recommended and supported [path](https://developers.redhat.com/blog/2019/03/05/yum-install-gcc-8-clang-6/) to installing a much more modern compiler that will work just fine for Yosys.

Additionally, nothing prevents users from using a more modern compiler built from source.  It's not that difficult, and users--sorry, users *who wish to compile Yosys*--who are on a severely outdated platform can IMO be expected to pay that small cost.

Support for gcc-4.8 does not come for free.  For example, see the code cruft added in #1766.  Or consider that 1/4 of the Travis CI resources are used making gcc-4.8 builds, presumably because it frequently breaks.  I think it's a good idea at least to deprecate gcc-4.8 support, warn on builds with gcc-4.8 that support will soon end, and then remove it entirely.

KrystalDelusion:
Done with #3620

PR #2055, opened 2020-05-14
logger: fix for multiple calls with same pattern

eddiehung:


passes/cmds/logger.cc
```diff
@@ -58,7 +58,8 @@ struct LoggerPass : public Pass {
 		log("        do not print warnings for the specified experimental feature\n");
 		log("\n");
 		log("    -expect <type> <regex> <expected_count>\n");
-		log("        expect log,warning or error to appear. In case of error return code is 0.\n");
+		log("        expect log, warning or error to appear. matched errors will terminate\n");
```
boqwxp, line 61:
```suggestion
		log("        expect log, warning or error to appear. Matched errors will terminate\n");
```

Actually, the general lack of capitalization and the strange grammar in the help message bother me.  I will submit a separate PR...

passes/cmds/logger.cc
```diff
@@ -158,12 +159,40 @@ struct LoggerPass : public Pass {
...
+						auto it = log_expect_log.begin();
+						auto ie = log_expect_log.end();
+						for (; it != ie; it++)
+							if (it->second.pattern == pattern) {
+								it->second.expected_count = count;
+								break;
+							}
+						if (it == ie)
+							log_expect_log.emplace_back(YS_REGEX_COMPILE(pattern), LogExpectedItem(pattern, count));
+					}
```
boqwxp, line 194: Does something like this not work because of difficulties comparing the elements of the `std::pair`?

Make `log_expect_error`, etc. `std::set` instead of `std::vector` and:

```suggestion

					if (type == "error")
						log_expect_error.insert(std::make_pair(YS_REGEX_COMPILE(pattern), LogExpectedItem(pattern, count)));
					else if (type == "warning")
						log_expect_warning.insert(std::make_pair(YS_REGEX_COMPILE(pattern), LogExpectedItem(pattern, count)));
					else if (type == "log")
						log_expect_log.insert(std::make_pair(YS_REGEX_COMPILE(pattern), LogExpectedItem(pattern, count)));
```

passes/cmds/logger.cc
eddiehung: I think if you look across most of Yosys' docs it does not capitalise the beginning of a paragraph, nor after full stops. I tried to be rebellious and fix that in the past, but now I don't bother and just comply (and even revert) to the majority.

passes/cmds/logger.cc
eddiehung: It's unclear whether all instances of `YS_REGEX_COMPILE` resolve to objects that are less-than-comparable. Note that I am walking over `log_expect_{error,warning,log}` and equality-checking `<pair>.second.pattern`.

boqwxp: I see, because there is some uncertainty about what exactly `YS_REGEX_COMPILE` is (at least until #1772 is closed), it's better to take the safe route.  But it would be good to check, because this is significantly more verbose.  And even if there is a problem with comparison, the `log_expect_*` should still be `std::set`, because that more closely expresses the design intent: Each element should be unique.

passes/cmds/logger.cc
boqwxp: Maybe we can start a revolution :).

PR #2056, opened 2020-05-14
log: Use `std::set` instead of `std::vector` for `log_expect_{error, warning, log}` to better express the intent that each element is unique.

boqwxp:
See https://github.com/YosysHQ/yosys/pull/2055#discussion_r425470742 .


boqwxp:
I see now that the standard prohibits non-const dereferencing of `std::set` iterators. ...

PR #2057, opened 2020-05-14
verilog: support attributes before (not after) task identifier (but 13 s/r conflicts)

eddiehung:
Attributes go after the function identifier for function calls, but before the task identifier for task enable statements.

This fixes it, but introduces 13 shift-reduce conflicts. Can someone smarter than I help me figure out where the conflict is please?
EDIT: Actually, it goes further and breaks a task related test...

Potentially blocking #2044 which also has 13 shift-reduce conflicts when I try and add support for attributes for `simple_behavioural_stmt`.

### DIFF
--- a/frontends/verilog/verilog_parser.y
+++ b/frontends/verilog/verilog_parser.y
@@ -2217,23 +2217,23 @@ behavioral_stmt:
 	defattr | assert | wire_decl | param_decl | localparam_decl | typedef_decl |
 	non_opt_delay behavioral_stmt |
 	simple_behavioral_stmt ';' | ';' |
-	hierarchical_id attr {
+	attr hierarchical_id {
 		AstNode *node = new AstNode(AST_TCALL);
-		node->str = *$1;
-		delete $1;
+		node->str = *$2;
+		delete $2;
 		ast_stack.back()->children.push_back(node);
 		ast_stack.push_back(node);
-		append_attr(node, $2);
+		append_attr(node, $1);
 	} opt_arg_list ';'{
 		ast_stack.pop_back();
 	} |
-	TOK_MSG_TASKS attr {
+	attr TOK_MSG_TASKS {
 		AstNode *node = new AstNode(AST_TCALL);
-		node->str = *$1;
-		delete $1;
+		node->str = *$2;
+		delete $2;
 		ast_stack.back()->children.push_back(node);
 		ast_stack.push_back(node);
-		append_attr(node, $2);
+		append_attr(node, $1);
 	} opt_arg_list ';'{
 		ast_stack.pop_back();
 	} |
@@ -2329,8 +2329,6 @@ behavioral_stmt:
 		case_type_stack.pop_back();
 		ast_stack.pop_back();
 	};
-
-	;
 
 unique_case_attr:
 	/* empty */ {
--- a/frontends/verilog/verilog_parser.y
+++ b/frontends/verilog/verilog_parser.y
@@ -2216,7 +2216,7 @@ simple_behavioral_stmt:
 behavioral_stmt:
 	defattr | assert | wire_decl | param_decl | localparam_decl | typedef_decl |
 	non_opt_delay behavioral_stmt |
-	simple_behavioral_stmt ';' | ';' |
+	attr simple_behavioral_stmt ';' | ';' |
 	attr hierarchical_id {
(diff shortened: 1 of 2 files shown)


boqwxp:
The first shift/reduce conflict happens in State 401.  The conflict is between a shift on `TOK_ID` and go to state 40:
```
State 40

   25 hierarchical_id: TOK_ID .

    $default  reduce using rule 25 (hierarchical_id)
```

and a reduce using rule 12:
```
    12 $@2: . %empty  [TOK_ID, TOK_MSG_TASKS, TOK_USER_TYPE, TOK_PKG_USER_TYPE, ATTR_BEGIN, TOK_PARAMETER, TOK_LOCALPARAM, TOK_VAR, TOK_INPUT, TOK_OUTPUT, TOK_INOUT, TOK_WIRE, TOK_WAND, TOK_WOR, TOK_REG, TOK_LOGIC, TOK_INTEGER, TOK_SIGNED, TOK_BEGIN, TOK_IF, TOK_FOR, TOK_WHILE, TOK_REPEAT, TOK_CASE, TOK_CASEX, TOK_CASEZ, TOK_GENVAR, TOK_SUPPLY0, TOK_SUPPLY1, TOK_RAND, TOK_CONST, TOK_UNIQUE, TOK_PRIORITY, '(']
```

TBH I am not sure where this rule 12 comes from.  I guess it's the first action for `attr` before `attr_opt` matches?

It looks like the default shift action is correct.  So I guess there just needs to be some precedence set to avoid warnings about the conflict: https://www.gnu.org/software/bison/manual/html_node/Non-Operators.html .  I'm not sure what, exactly, though...  I tried setting precedence for `TOK_ID` and inventing a lower-precedence `EMPTY` terminal to manually set the precedence for the empty rule in `attr_opt`, but nothing changed; still 13 s/r conflicts.

mwkmwkmwk:
Unfortunately, your grammar after the change is not LR(1), thus yacc cannot deal with it.  Consider the following simplified grammar which exhibits the same problem:

```
%token ATTR ID
%%
top: top stmt | ;
attr: ATTR | ;
stmt: attr ID ';'  /* rule 1 */
 | ID '=' ID ';'  /* rule 2 */
```

Consider the situation when you have to parse `abc = xyz;` or `abc;`. When yacc wants to parse a stmt and sees `abc` (as lookahead), there are two rules it could match: for rule 1, it would have to immediately perform the `attr: /*empty*/` reduction (before shifting the ID), and for rule 2 it would have to shift the ID immediately.  However, it cannot do the decision without looking one token further (LR(1) implies only one token of lookahead and it's `abc` right now).  If you resolve the conflict in favor of shift, rule 1 can never match, and you won't be able to parse `abc;`, and if you resolve in favor of reduce, you won't be able to parse `abc = xyz;`.

A simple way out of this would be to add an optional attribute in front of a simple_behavioral_stmt as well — this way both rules that begin with an ID will require the same action (reducing `attr: /*empty*/`) and there will be no conflict.

boqwxp:
That makes perfect sense and is easy to see with the simplified grammar.

Can you please walk me through how you deduced this conflict from the Bison output?  And what exactly is rule 12, and why are there so many potential token matches for it?

mwkmwkmwk:
I have no idea, I didn't look at Bison output at all, I just looked at the `.y` file and looked for rules with similar prefixes that could cause trouble.

mwkmwkmwk:
That said, rule 12 looks like it is simply the "empty attr" rule, and the tokens listed for it are the lookahead tokens for which bison wants to invoke this rule, ie. all possible tokens that could happen after `attr` in any behavioral_stmt rule (see first set / follow set in any good compiler textbook)

boqwxp:
Thanks :).

frontends/verilog/verilog_parser.y
```diff
@@ -2217,23 +2217,23 @@ behavioral_stmt:
 	defattr | assert | wire_decl | param_decl | localparam_decl | typedef_decl |
 	non_opt_delay behavioral_stmt |
 	simple_behavioral_stmt ';' | ';' |
```
boqwxp, line 2219:
```suggestion
	attr simple_behavioral_stmt ';' | ';' |
```

Credit to @mwkmwkmwk (just formalizing the diff for the GitHub interface)

PR #2064, opened 2020-05-17
Use conditional assignment for feature flags in Makefile

anishathalye:
This allows enabling/disabling features without having to edit the Makefile, e.g. with `make ENABLE_TCL=0 ENABLE_PYOSYS=1`.


whitequark:
You don't need to edit the `Makefile`; you can create a `Makefile.conf` with the changed values. But I like your proposal regardless.

anishathalye:
Thank you, I hadn't noticed that! That solves my immediate problem of building a minimal version of Yosys in CI. But yeah, I think having the command-line options basically for free is nice.

nakengelhardt:
Variable definitions on the command line [override definitions in the file anyway](https://www.gnu.org/software/make/manual/html_node/Overriding.html), even without `?=`. But I'm not sure if this behavior is gnu make specific?

anishathalye:
Ah yes, you're right. The `?=` is necessary for being able to override with environment variables, though, so it's useful if we want `ENABLE_TCL=0 make` or `export ENABLE_TCL=0; make` to work.

whitequark:
Does yosys even build with non-GNU Make?

nakengelhardt:
I don't even know what non-gnu make implementations still exist.

But speaking of environment variables, are we sure that other tools don't set the same variables? Something like `ENABLE_PLUGINS` is very generic...

whitequark:
There's the BSD make, used on macOS. (You'd use `gmake` for most software there.)

nakengelhardt:
Which system version? Under 10.14, I have
```
> which make
/usr/bin/make
> make -v
GNU Make 3.81
```
which I think is the one that came with Xcode...

anishathalye:
I don't think macOS ships with BSD make. on macOS 10.14.6:

```console
$ /Applications/Xcode.app/Contents/Developer/usr/bin/make --version                                                                                                            anish at mbp (--)(cow)
GNU Make 3.81
Copyright (C) 2006  Free Software Foundation, Inc.
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
PARTICULAR PURPOSE.

This program built for i386-apple-darwin11.3.0
```

whitequark:
Ok, no idea then--I remember I had to deal with non-GNU make on macOS at some point but maybe I did something wrong.

nakengelhardt:
Discussed this internally, and we actually want to discourage using environment variables for make settings, to avoid conflicts with other software, and issues around needing to clean every time the value changes. Using `Makefile.conf` or the `config-xxx` targets for this purpose is preferred.

PR #2066, opened 2020-05-18
ilang_parser: reject negative literals in constant position

whitequark:
Consider this IL code:

    module \foo
      attribute \init -2147483647
      wire width 64 \-2**31+1
    end

Before this commit, it would not roundtrip through the IL backend:

    module \foo
      attribute \init 32'10000000000000000000000000000001
      wire width 64 \-2**31+1
    end

Neither would it result in Verilog output with a negative number:

    module foo();
      (* init = 32'd2147483649 *)
      wire [63:0] \-2**31+1 ;
    endmodule

To avoid ambiguity, prohibit these entirely. This PR also includes a fix for an out-of-range check that previously would have never fired.


whitequark:
Converting to draft as issues were raised during internal review.

Issue #2036, opened 2020-05-07
pyosys autogeneration: Parser skips a few functions due to line breaks in header


hackfin:
## Steps to reproduce the issue

* Build yosys with pyosys Python extension
* Run code below:

```
from pyosys import libyosys as ys

def addSignal(m, name, n, public = False):
    w = m.addWire(name, n)
    return ys.SigSpec(w)

def ID(x):
    return ys.IdString("$" + x)

d = ys.Design()
m = d.addModule(ID("test1"))
clk, r, d, q = [ addSignal(m, ID("sig_%d" % i), 1) for i in range(4)]
c = ys.Const(0, 1)
m.addAdff(ID("adff"), clk, r, d, q, c, 0, 0)
```
## Expected behavior

`$adff` Element being added to the module

## Actual behavior

Method not found, because not generated in the python module from the parser auxiliary `misc/py_wrap_generator.py`, due to undetected line breaks in `rtlil.h`

Quick workaround: Remove line breaks. See also patch.

[patch-pyosys.txt](https://github.com/YosysHQ/yosys/files/4592229/patch-pyosys.txt)

mmicko:
These methods are now available. Code breaks later due to unrelated issue.

Issue #2043, opened 2020-05-11
pyosys API: Silent aborts (segfault/assert())


hackfin:
## Steps to reproduce the issue

A number of Python API calls can 'crash' the Python interpreter due to segfaults or assert() exits from the yosys kernel side.

I've collected them in a binder with a fairly recent pyosys build:

[![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/hackfin/myhdl.git/jupyosys?filepath=src%2Fmyhdl%2Fmyhdl-yosys%2Fexample%2Fipynb%2Fpyosys_issues.ipynb)

## Expected behavior

Return a Python exception.

## Actual behavior

Python interpreter or IPython kernel terminate without notice. When called from the shell, the error can be guessed. Inside IPython, the kernel dies non-verbosely and it seems impossible to debug the problem through stderr capture.

## Comment

Maybe it would be time to start a test suite for the pyosys API. Any ideas, opinions on `py.test`? Testing might get tricky due the module crashing (i.e. aborting the test)

hackfin:
Making another move on this one: Added docker hub py.test setup on this branch that is running a couple of basic tests.

https://github.com/hackfin/yosys/tree/pytest

Quickstart from local docker shell:
`docker run -it --rm  hackfin/yosys /home/pyosys/src/tests/pyosys/run-test.sh`

The logging test is currently disabled, to provoke the segfault, run

```
docker run -it --rm  hackfin/yosys bash -c "cd /home/pyosys/src/tests/pyosys && mv _test_logging.py test_logging.py && py.test-3"
```

However, it can be run as a single test without crash:

`docker run -it --rm  hackfin/yosys py.test-3  /home/pyosys/src/tests/pyosys/_test_logging.py`

As I don't really have the insights on the entire testing situation and status of the current Dockerfile, it's probably too early to do formal PRs. Also, I'm not sure about the future of dockerhub autobuilds.
On the travis side, it might likely just require an additional 'py.test' package installer line in the .yml to spool within `make test`

mmicko:
pyosys wrapper was redone using pybind11 but segfaults and asserts in kernel code are fatal errors so they could not be handled on python side

PR #2070, opened 2020-05-19
Pyosys API: idict type handling

hackfin:
Q&D patch for:

- Support of access of idict type members
- Hotfix (no line breaks) in order to have all methods parsed (see also #2036)

Hereby requesting review.

Question: are there any test suites for the Python API planned?

### DIFF
--- a/kernel/rtlil.h
+++ b/kernel/rtlil.h
@@ -1222,13 +1222,10 @@ public:
 	RTLIL::Cell* addFf    (RTLIL::IdString name, const RTLIL::SigSpec &sig_d, const RTLIL::SigSpec &sig_q, const std::string &src = "");
 	RTLIL::Cell* addDff   (RTLIL::IdString name, const RTLIL::SigSpec &sig_clk, const RTLIL::SigSpec &sig_d,   const RTLIL::SigSpec &sig_q, bool clk_polarity = true, const std::string &src = "");
 	RTLIL::Cell* addDffe  (RTLIL::IdString name, const RTLIL::SigSpec &sig_clk, const RTLIL::SigSpec &sig_en,  const RTLIL::SigSpec &sig_d, const RTLIL::SigSpec &sig_q, bool clk_polarity = true, bool en_polarity = true, const std::string &src = "");
-	RTLIL::Cell* addDffsr (RTLIL::IdString name, const RTLIL::SigSpec &sig_clk, const RTLIL::SigSpec &sig_set, const RTLIL::SigSpec &sig_clr,
-			RTLIL::SigSpec sig_d, const RTLIL::SigSpec &sig_q, bool clk_polarity = true, bool set_polarity = true, bool clr_polarity = true, const std::string &src = "");
-	RTLIL::Cell* addAdff (RTLIL::IdString name, const RTLIL::SigSpec &sig_clk, const RTLIL::SigSpec &sig_arst, const RTLIL::SigSpec &sig_d, const RTLIL::SigSpec &sig_q,
-			RTLIL::Const arst_value, bool clk_polarity = true, bool arst_polarity = true, const std::string &src = "");
+	RTLIL::Cell* addDffsr (RTLIL::IdString name, const RTLIL::SigSpec &sig_clk, const RTLIL::SigSpec &sig_set, const RTLIL::SigSpec &sig_clr, RTLIL::SigSpec sig_d, const RTLIL::SigSpec &sig_q, bool clk_polarity = true, bool set_polarity = true, bool clr_polarity = true, const std::string &src = "");
+	RTLIL::Cell* addAdff (RTLIL::IdString name, const RTLIL::SigSpec &sig_clk, const RTLIL::SigSpec &sig_arst, const RTLIL::SigSpec &sig_d, const RTLIL::SigSpec &sig_q, RTLIL::Const arst_value, bool clk_polarity = true, bool arst_polarity = true, const std::string &src = "");
 	RTLIL::Cell* addDlatch (RTLIL::IdString name, const RTLIL::SigSpec &sig_en, const RTLIL::SigSpec &sig_d, const RTLIL::SigSpec &sig_q, bool en_polarity = true, const std::string &src = "");
-	RTLIL::Cell* addDlatchsr (RTLIL::IdString name, const RTLIL::SigSpec &sig_en, const RTLIL::SigSpec &sig_set, const RTLIL::SigSpec &sig_clr,
-			RTLIL::SigSpec sig_d, const RTLIL::SigSpec &sig_q, bool en_polarity = true, bool set_polarity = true, bool clr_polarity = true, const std::string &src = "");
+	RTLIL::Cell* addDlatchsr (RTLIL::IdString name, const RTLIL::SigSpec &sig_en, const RTLIL::SigSpec &sig_set, const RTLIL::SigSpec &sig_clr, RTLIL::SigSpec sig_d, const RTLIL::SigSpec &sig_q, bool en_polarity = true, bool set_polarity = true, bool clr_polarity = true, const std::string &src = "");
 
 	RTLIL::Cell* addBufGate    (RTLIL::IdString name, const RTLIL::SigBit &sig_a, const RTLIL::SigBit &sig_y, const std::string &src = "");
 	RTLIL::Cell* addNotGate    (RTLIL::IdString name, const RTLIL::SigBit &sig_a, const RTLIL::SigBit &sig_y, const std::string &src = "");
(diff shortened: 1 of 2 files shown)


kernel/rtlil.h
```diff
@@ -1222,13 +1222,10 @@ struct RTLIL::Module : public RTLIL::AttrObject
 	RTLIL::Cell* addFf    (RTLIL::IdString name, const RTLIL::SigSpec &sig_d, const RTLIL::SigSpec &sig_q, const std::string &src = "");
 	RTLIL::Cell* addDff   (RTLIL::IdString name, const RTLIL::SigSpec &sig_clk, const RTLIL::SigSpec &sig_d,   const RTLIL::SigSpec &sig_q, bool clk_polarity = true, const std::string &src = "");
 	RTLIL::Cell* addDffe  (RTLIL::IdString name, const RTLIL::SigSpec &sig_clk, const RTLIL::SigSpec &sig_en,  const RTLIL::SigSpec &sig_d, const RTLIL::SigSpec &sig_q, bool clk_polarity = true, bool en_polarity = true, const std::string &src = "");
-	RTLIL::Cell* addDffsr (RTLIL::IdString name, const RTLIL::SigSpec &sig_clk, const RTLIL::SigSpec &sig_set, const RTLIL::SigSpec &sig_clr,
-			RTLIL::SigSpec sig_d, const RTLIL::SigSpec &sig_q, bool clk_polarity = true, bool set_polarity = true, bool clr_polarity = true, const std::string &src = "");
-	RTLIL::Cell* addAdff (RTLIL::IdString name, const RTLIL::SigSpec &sig_clk, const RTLIL::SigSpec &sig_arst, const RTLIL::SigSpec &sig_d, const RTLIL::SigSpec &sig_q,
-			RTLIL::Const arst_value, bool clk_polarity = true, bool arst_polarity = true, const std::string &src = "");
```
whitequark, line 1228: IMO the parser in pyosys should be fixed rather than the code in core Yosys files be modified to work around its deficiencies.

nakengelhardt:
@btut Can you review this please?

kernel/rtlil.h
hackfin: Agreed. There were more issues in the parser though that I didn't want to open that box.
It might be worth evaluating if the header parsing business could be circumvented.

whitequark: I'm wondering if SWIG could be used instead of boost::python; as a side effect the API wouldn't be just exported to Python but to many other languages as well.

hackfin: I've used SWIG in the past, it used to be problematic with nested user-defined object structures and ownerships, might have gotten better. Maybe a wrapper class would solve some problems, but also introduce an (internal) API change, possibly. The other option I've found quite effective is Cython (python-wrapping only, obviously). This allows some tight interaction with the C++ code while taking care of the reference counting and object ownerships.
I have found little references on the pyosys API and its testing so far, maybe someone from the maintainer's side can share more insight?

whitequark: > it used to be problematic with nested user-defined object structures and ownerships, might have gotten better.

AFAICT the ownership management in pyosys is a rather invasive hack that relies on `get_all_*()` functions. SWIG can't possibly be worse because you can just keep using that.

btut: I agree that it would be better to fix the parser. I am not sure but I thought multi-line definitions were supported.

> AFAICT the ownership management in pyosys is a rather invasive hack that relies on get_all_*() functions. SWIG can't possibly be worse because you can just keep using that.

True. I never worked with SWIG and have no idea how much work it would be to get it up and running. I expect the result to be much better though as the current parser only supports (most of) the c++ style that was used when I wrote it.

misc/py_wrap_generator.py
```diff
@@ -349,19 +349,24 @@ def translate_cpp(c, varname, types, prefix, ref):
 		text += prefix + "}"
 		return text
 
+class IDictTranslator(PythonListTranslator):
+	typename = "boost::python::list"
+	orig_name = "idict"
+	insert_name = ""
+
```
btut, line 356: LGTM

whitequark:
I have some interest in SWIG so if there's rough consensus that it's the way to go, I might be able to try and port pyosys to SWIG.

kernel/rtlil.h
nakengelhardt: > I have found little references on the pyosys API and its testing so far, maybe someone from the maintainer's side can share more insight?

The pyosys API was developed as a student project by @btut. It doesn't have as many users since it isn't enabled by default, so it's relatively rarely touched, and I believe none of the maintainers has worked with it enough to develop a thorough understanding of the code yet. As far as I know, there are no tests, at least in `make test` (which is what runs in our CI).

btut: I do know of some people that are using it, but there are probably not many out there.

I don't think the userbase is important here though. Ideally, the interfaces would stay the same no matter which tool (parser+boostpython vs SWIG) does the translation.

If the parser stays, adding some tests would definitely be a good idea. What do you have in mind? test the build, check if a set of functions are exposed to python, and then do some basic interaction with Yosys?

nakengelhardt: Oh no, the userbase comment wasn't meant about usefulness, just about testing. Since our test suite still has a lot of deficiencies, we rely to a significant extent on someone yelling if we break something 😅 Because pyosys isn't used by as many people, it takes longer for someone to notice problems. (And we also look at the code less often than other parts of yosys since there are fewer issues/PRs about it.)

A few unit tests would definitely be an improvement. The tests included in ``make test`` should be short (the whole thing has to fit in travis' 1 hour limit), so just exercising the basic functions on a small design sounds good. We could also add a larger integration test, perhaps trying to synthesize a more complex design, to yosys-tests (which we are in the process of setting up as an additional, longer-running CI task).

btut:
> I have some interest in SWIG so if there's rough consensus that it's the way to go, I might be able to try and port pyosys to SWIG.

I think this highly depends on the amount of tweaking that's needed for SWIG. If it comes with a ton of code that needs to be maintained it might not be worth the effort. I definitely looked into using SWIG and cython back when I wrote the wrappers, but unfortunately, I cannot exactly remember why I went with boost::python.
I think the main reason was ownership. Very early in the development, I had a meeting with Claire where we agreed on the get_all_* type access to objects to avoid nasty issues with ownership. I think none of the libraries offered a way to implement that way of access.

If you want to look into SWIG, I'll tell you what my biggest issues (other than ownership) were during development so you can watch out for those early on:

- The pass and monitor interface: You need to be able to have python classes that are children of C++ classes in order to register passes and monitors.
- Initialization for different kind of usages
  - python passes used by C++ yosys: In this case, no initialization must be done since that was already done in C++
  - python passes used in python yosys: same as above
  - loading the pyosys library in python: initializations must be done
- System wide installation: this still is not working properly on some systems, as the initialization is done twice. If you just use a libyosys.so in the same folder as your script, it should be fine. If you load that same .so from the python library path, an exception occurs.
- support for custom translators for containers

kernel/rtlil.h
hackfin: > I don't think the userbase is important here though. Ideally, the interfaces would stay the same no matter which tool (parser+boostpython vs SWIG) does the translation.

For me, API changes are no problem, as I'm wrapping things up to a certain point.
The project (direct synthesis from Python/MyHDL) is here: https://github.com/hackfin/myhdl/tree/jupyosys.
The minor issues I had are mostly the ones in #2043, however there are some API specifics that appear 'unpythonic', for instance yosys dictionary types access (where only the full assignment of a dict is supported and insertion by key would silently be ineffective).

> If the parser stays, adding some tests would definitely be a good idea. What do you have in mind?

I keep relying on py.test mostly, you'll find a few scenarios in the above MyHDL fork. There's also a nice integration with jupyter notebooks so you can make sure that documented examples also are up to date (you'll find these in the secondary-pulled 'hacker space' from the jupyosys binder environment)

hackfin:
@btut, you're spot on, I guess, comes close to what I remember and forced us to redesign  object handling kernels WRT ownership in the past. 
I am still occasionally having issues with the readability of SWIG (and maintenance of several code layers). I'm all up for elegant prototypes, but if the parser can be fixed easily and tests are in the CI, I'd rather not touch the running system at this point and revisit some issues once there are relevant use cases.

Just my few cents opinion on strategies that generally worked well:
* C/C++ wrapper as thin as possible, but robust (error codes/exceptions covered WRT #2043)
* all fancy API stuff on top as a Python module, so that the API is as pythonic as possible and the wrapper layer can remain somewhat dirty/volatile (as long as it does the refcounting right).

Do you think some of those initialization issues could be solved by splitting functionality (like cell/module/signals/wire vs. passes) into different submodules?

nakengelhardt:
Let's merge this for now. There's still the open issue #2036 to remind us to fix the parsing, and removing a few line breaks is better than leaving things broken on a matter of principle IMO...

btut:
> Do you think some of those initialization issues could be solved by splitting functionality (like cell/module/signals/wire vs. passes) into different submodules?

I don't see how that could help. The crucial thing is to find out in what scenarios and when to call yosys_setup().

PR #2073, opened 2020-05-23
intel_alm: ABC9 sequential optimisations

Ravenslofty:
This plugs in #1926 for `synth_intel_alm`. The benefits are [modest](https://gist.github.com/ZirconiumX/a9b2c4bc7f5be27962415d698e5495cf) but might be worth it to somebody.

Some small mapping files are used to distinguish between flops with asynchronous clear (which should be `(* abc9_box *)`) and those without (which can be passed to ABC9 as `(* abc9_flop *)`).

This lead to some cleanup; flops and LUTRAM are a bit better specified, and I also reorganised the help to put more useful options near the top.

cc @mwkmwkmwk @eddiehung

### DIFF
--- a/techlibs/intel_alm/common/dff_sim.v
+++ b/techlibs/intel_alm/common/dff_sim.v
@@ -90,7 +90,7 @@ specify
     $setup(SLOAD, posedge CLK, `SYNCSETUP);
     $setup(SDATA, posedge CLK, `SYNCSETUP);
 
-    (ACLR => Q) = `COMBPATH;
+    if (!ACLR) (ACLR => Q) = `COMBPATH;
 endspecify
 
 initial begin
--- a/techlibs/intel_alm/common/dff_sim.v
+++ b/techlibs/intel_alm/common/dff_sim.v
@@ -79,10 +79,9 @@ module MISTRAL_FF(
 );
 
 specify
-    if (ENA) (posedge CLK => (Q : DATAIN)) = `SYNCPATH;
-    if (ENA) (posedge CLK => (Q : SCLR)) = `SYNCPATH;
-    if (ENA) (posedge CLK => (Q : SLOAD)) = `SYNCPATH;
-    if (ENA) (posedge CLK => (Q : SDATA)) = `SYNCPATH;
+    if (ENA && ACLR !== 1'b0 && !SCLR && !SLOAD) (posedge CLK => (Q : DATAIN)) = `SYNCPATH;
+    if (ENA && SCLR) (posedge CLK => (Q : 1'b0)) = `SYNCPATH;
+    if (ENA && !SCLR && SLOAD) (posedge CLK => (Q : SDATA)) = `SYNCPATH;
 
     $setup(DATAIN, posedge CLK, `SYNCSETUP);
     $setup(ENA, posedge CLK, `SYNCSETUP);
@@ -90,7 +89,7 @@ specify
     $setup(SLOAD, posedge CLK, `SYNCSETUP);
     $setup(SDATA, posedge CLK, `SYNCSETUP);
 
-    if (!ACLR) (ACLR => Q) = `COMBPATH;
+    if (ACLR === 1'b0) (ACLR => Q) = `COMBPATH;
 endspecify
 
 initial begin
--- a/techlibs/intel_alm/common/dff_sim.v
+++ b/techlibs/intel_alm/common/dff_sim.v
@@ -53,23 +53,45 @@
 // Q: data output
 //
 // Note: the DFFEAS primitive is mostly emulated; it does not reflect what the hardware implements.
+
+`ifdef cyclonev
+`define SYNCPATH 262
+`define SYNCSETUP 522
+`define COMBPATH 0
+`endif
+`ifdef cyclone10gx
+`define SYNCPATH 219
+`define SYNCSETUP 268
+`define COMBPATH 0
+`endif
+
+// fallback for when a family isn't detected (e.g. when techmapping for equivalence)
+`ifndef SYNCPATH
+`define SYNCPATH 0
+`define SYNCSETUP 0
+`define COMBPATH 0
+`endif
+
+(* abc9_box, lib_whitebox *)
 module MISTRAL_FF(
     input DATAIN, CLK, ACLR, ENA, SCLR, SLOAD, SDATA,
     output reg Q
 );
 
-`ifdef cyclonev
 specify
-    (posedge CLK => (Q : DATAIN)) = 262;
-    $setup(DATAIN, posedge CLK, 522);
+    if (ENA) (posedge CLK => (Q : DATAIN)) = `SYNCPATH;
+    if (ENA) (posedge CLK => (Q : SCLR)) = `SYNCPATH;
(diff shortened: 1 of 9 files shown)


ZirconiumX:
I could add a parameter for it, sure, but Intel flops let you use async/sync clears anyway, so it's not directly comparable.

ZirconiumX:
I partly want to keep `MISTRAL_FF` close to what the hardware does, though, so hmm...

eddiehung:
> I could add a parameter for it, sure, but Intel flops let you use async/sync clears anyway, so it's not directly comparable.

Ah that's right. I was thinking of using: 
https://github.com/YosysHQ/yosys/blob/7067f9d84736bae31b9ca31479db19e3698579c4/techlibs/intel_alm/common/abc9_map.v#L10
but I now realise that `cells_sim.v` is not a techmap file...

Let's sit on this a bit more.

ZirconiumX:
So, I've been prototyping this, and have the following:

```verilog
(* abc9_box=USES_ASYNC_CLEAR, abc9_flop=!USES_ASYNC_CLEAR, lib_whitebox *)
module MISTRAL_FF(
    input DATAIN, CLK, ACLR, ENA, SCLR, SLOAD, SDATA,
    output reg Q
);

parameter USES_ASYNC_CLEAR = 0;

specify
    if (ENA) (posedge CLK => (Q : DATAIN)) = `SYNCPATH;
    if (ENA) (posedge CLK => (Q : SCLR)) = `SYNCPATH;
    if (ENA) (posedge CLK => (Q : SLOAD)) = `SYNCPATH;
    if (ENA) (posedge CLK => (Q : SDATA)) = `SYNCPATH;

    $setup(DATAIN, posedge CLK, `SYNCSETUP);
    $setup(ENA, posedge CLK, `SYNCSETUP);
    $setup(SCLR, posedge CLK, `SYNCSETUP);
    $setup(SLOAD, posedge CLK, `SYNCSETUP);
    $setup(SDATA, posedge CLK, `SYNCSETUP);

    (ACLR => Q) = `COMBPATH;
endspecify

initial begin
    // Altera flops initialise to zero.
	Q = 0;
end

always @(posedge CLK, negedge ACLR) begin
    // Asynchronous clear
    if (!ACLR) Q <= 0;
    // Clock-enable
	else if (ENA) begin
        // Synchronous clear
        if (SCLR) Q <= 0;
        // Synchronous load
        else if (SLOAD) Q <= SDATA;
        else Q <= DATAIN;
    end
end

endmodule
```

which brings up errors like this:
```
ERROR: Module '$paramod\MISTRAL_FF\USES_ASYNC_CLEAR=0' with (* abc9_flop *) contains a $adff cell, which is not supported for sequential synthesis.
```

However, since `USES_ASYNC_CLEAR=0` and thus `ACLR=1` (negative true), `opt_rmdff` *should* have turned it into a `$dff`...

eddiehung:
It doesn't look like the `USES_ASYNC_CLEAR` parameter actually affects the whitebox contents so Yosys is still synthesising an async dff, when looking at just the `MISTRAL_FF` contents.

To be honest, the `abc9_{map,unmap}.v` isn't too bad. It should never be seen by any user anyway; to make this explicit I would suggest you use the `$__` prefix for internal cells that should never see the light of day.

Personally, I would also move that cell from `cells_sim.v` into `abc9_model.v`, and the model could instantiate `MISTRAL_FF` which /should/ allow `opt_rmdff` to change `$adff` back into `$dff` to save a redundant description of its contents.

ZirconiumX:
> to make this explicit I would suggest you use the `$__` prefix for internal cells that should never see the light of day.

Here's a fun bug for you: if it's `$__` prefixed, the ABC9 pass appears to treat the `(* abc9_flop *)` as an `(* abc9_box *)` - it doesn't generate `$__MISTRAL_FF_SYNCONLY_$abc9_flop` cells, but instead keeps them as `$__MISTRAL_FF_SYNCONLY`, and so ABC doesn't print anything about `boxffs`.

eddiehung:
> > to make this explicit I would suggest you use the `$__` prefix for internal cells that should never see the light of day.
> 
> Here's a fun bug for you: if it's `$__` prefixed, the ABC9 pass appears to treat the `(* abc9_flop *)` as an `(* abc9_box *)` - it doesn't generate `$__MISTRAL_FF_SYNCONLY_$abc9_flop` cells, but instead keeps them as `$__MISTRAL_FF_SYNCONLY`, and so ABC doesn't print anything about `boxffs`.

Looks like you were missing `-icells`: https://github.com/YosysHQ/yosys/pull/2073/commits/c7cb721d4d80f8df37cb50d2acf7932821f84376#diff-8b924da3546f92926c0ca2b539e3e05cL176-R176
Without this the `$__` prefix is treated as a public name resulting in `\\$__`.

eddiehung:
Actually, all that was needed was this: https://github.com/YosysHQ/yosys/blob/0461cc89edb1b75431d5805646e483f019c41acd/techlibs/intel_alm/common/dff_sim.v#L93

ZirconiumX:
You broke CI with your changes here, @eddiehung.

techlibs/intel_alm/common/dff_sim.v
```diff
@@ -53,23 +53,45 @@
...
-endspecify
-`endif
-`ifdef cyclone10gx
 specify
-    (posedge CLK => (Q : DATAIN)) = 219;
-    $setup(DATAIN, posedge CLK, 268);
+    if (ENA) (posedge CLK => (Q : DATAIN)) = `SYNCPATH;
+    if (ENA) (posedge CLK => (Q : SCLR)) = `SYNCPATH;
+    if (ENA) (posedge CLK => (Q : SLOAD)) = `SYNCPATH;
+    if (ENA) (posedge CLK => (Q : SDATA)) = `SYNCPATH;
```
eddiehung, line 85: Not that it makes a difference for timing analysis:
```suggestion
    if (ENA && ACLR !== 1'b0 && !SCLR && !SLOAD) (posedge CLK => (Q : DATAIN)) = `SYNCPATH;
    if (ENA && SCLR) (posedge CLK => (Q : 1'b0)) = `SYNCPATH;
    if (ENA && SLOAD) (posedge CLK => (Q : SDATA)) = `SYNCPATH;
```

techlibs/intel_alm/common/dff_sim.v
```diff
@@ -53,23 +53,45 @@
...
+
+    $setup(DATAIN, posedge CLK, `SYNCSETUP);
+    $setup(ENA, posedge CLK, `SYNCSETUP);
+    $setup(SCLR, posedge CLK, `SYNCSETUP);
+    $setup(SLOAD, posedge CLK, `SYNCSETUP);
+    $setup(SDATA, posedge CLK, `SYNCSETUP);
+
+    if (!ACLR) (ACLR => Q) = `COMBPATH;
```
eddiehung, line 93: To be more strict...?
```suggestion
    if (ACLR === 1'b0) (ACLR => Q) = `COMBPATH;
```

eddiehung:
Is there anything outstanding here?

Ravenslofty:
No, nothing outstanding; this is ready to merge.

PR #2076, opened 2020-05-25
clean: remove unused wires even with (* init *) on

eddiehung:
`(* init *)` is not guaranteed to be attached to a DFF cell's Q port. Thus when that DFF gets blown away (e.g. by `abc9 -dff`) then it would desirable to also remove unused wires -- even with `(* init *)` on -- that get left behind.

### DIFF
--- a/passes/opt/opt_clean.cc
+++ b/passes/opt/opt_clean.cc
@@ -323,13 +323,17 @@ bool rmunused_module_signals(RTLIL::Module *module, bool purge_mode, bool verbos
 			initval.bits.resize(GetSize(wire), State::Sx);
 		if (initval.is_fully_undef())
 			wire->attributes.erase(ID::init);
+		Const used_initval = initval;
+		for (auto i = 0; i < GetSize(wire); i++)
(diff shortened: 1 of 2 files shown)


eddiehung:
> By convention, the init attribute should be removed _when the DFF is removed_, see for example `opt_rmdff`. This allows us to check for unhandled init attributes in the `check` pass. By just removing any dangling init attribute, we are effectively getting rid of that check in some scenarios.

My tentative argument was that if a wire is completely unused and would otherwise (if it did not have `(* init *)` be optimised away, does that still count as an unhandled dangling init attribute?
I was hoping to rely on `clean` to remove such signals, since it seems a little wasteful for `abc9` to build a `SigMap` just for the purpose of handling `(* init *)`.

Another solution I think would work is to introduce an invariant that `(* init *)` must be attached to the `$_DFF*.Q` wire; the following benchmark demonstrates when this doesn't happen:
```
read_verilog -icells <<EOT
module top(input clk, output public);
(* init = 1'b1 *) wire private;
$_DFF_P_ dff (.C(clk), .D(1'b1), .Q(private));
assign public = private;
endmodule
EOT
rename -hide w:private
clean
dump
```
as it prefers the public name over the private name even though the private one has an `(* init *)` attribute on.

PR #2077, opened 2020-05-25
abc9: -dff improvements

eddiehung:
I noticed that `abc9 -dff` could leave behind some `(* init *)` attributes, which should be fixed by:
* Re-attaching the `$_DFF_[NP]_.Q` signal to the replaced flop, so that the original `(* init *)` attribute is linked and can be removed (also has the nice side effect of preserving public signals on flop outputs)
* Left-over `(* init *)` attributes for unused flops removed by ABC9 are queued to be fixed by #2076 (dependency)

### DIFF
--- a/backends/aiger/xaiger.cc
+++ b/backends/aiger/xaiger.cc
@@ -85,7 +85,7 @@ struct XAigerWriter
 	dict<SigBit, SigBit> not_map, alias_map;
 	dict<SigBit, pair<SigBit, SigBit>> and_map;
 	vector<SigBit> ci_bits, co_bits;
-	dict<SigBit, Cell*> ff_bits;
+	vector<Cell*> ff_list;
 	dict<SigBit, float> arrival_times;
 
 	vector<pair<int, int>> aig_gates;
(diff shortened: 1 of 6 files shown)


whitequark:
> * also has the nice side effect of preserving public signals on flop outputs

According to @cliffordwolf this isn't just a nice side effect but actually an (undocumented?) invariant that Yosys guarantees.

eddiehung:
> > * also has the nice side effect of preserving public signals on flop outputs
> 
> According to @cliffordwolf this isn't just a nice side effect but actually an (undocumented?) invariant that Yosys guarantees.

Sorry, I wasn't making any grand claims to inventing this. It was only through fixing this unrelated issue that I realised it was an oversight of mine that also needed fixing, fortunately it was a two-birds-with-one-stone jobby.

whitequark:
> Sorry, I wasn't making any grand claims to inventing this.

Right, I didn't understand you that way--rather I meant that according to @clairexen it is actually *necessary* to preserve this invariant.

> I'm not sure that invariant can hold when it comes to going through abc-mapping, especially since there is a possibility that (a) sequential optimisations can identify redundant flops which get removed entirely

I discussed this with @clairexen and she explained that in this case, the pass should retain all of the public names but reconnect them to the same flops.

> and (b) after retiming (which `abc` supports but not `abc9`)

This is indeed a tough problem to solve. I'm not sure how other toolchains fare here.

Vinayakamk:
hey i am intrested in doing retiming,
after 
reading verilog,
techmap, 
dfflibmap -liberty ..../...lib, here if i read liberty most of the things are skipping
abc-liberty ......./.lib ,abc -dff 
is this process correct.?

and if i want to compare results after retiming what i can do
i expect timings

PR #2079, opened 2020-05-25
Fixing implicit declaration of nets in dynamic_part_select test

dh73:
@eddiehung noticed implicit reset declaration in `reset_test.v`, fixing that and setting the default net type to none for the rest of files.


tests/various/dynamic_part_select/multiple_blocking_gate.v
```diff
@@ -1,8 +1,9 @@
+`default_nettype wire
```
nakengelhardt, line 1: Why is it `wire` for this file?

dh73: Good catch, thanks!

dh73:
Redundant due #2133 .

PR #2081, opened 2020-05-25
blackbox: use Module::makeblackbox() method

eddiehung:
~Currently, `blackbox` only sets the RTLIL blackbox attribute. Re-jig it so that the AST attribute is also set, such that derived modules created by `Module::derive()` also inherit this blackbox attribute.~

### DIFF
--- a/passes/cmds/blackbox.cc
+++ b/passes/cmds/blackbox.cc
@@ -48,31 +48,7 @@ struct BlackboxPass : public Pass {
 
 		for (auto module : design->selected_whole_modules_warn())
 		{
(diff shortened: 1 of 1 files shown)


whitequark:
I'm not 100% sure what kind of input would trigger this code path, so a testcase would be helpful here.

eddiehung:
Actually, I think it's no longer necessary to set the AST attribute. 
I think there is a little value in removing the redundant code inside `blackbox` though, so I've forced-pushed without 46b56aa and updated the title.

PR #2083, opened 2020-05-26
Pass IdString by reference in rtlil.h

rswarbrick:
Although an IdString contains just a single int (the index_ field), it
acts like a ref-counted smart pointer. This means that if I have a
function call like this:

    void foo (IdString arg);

    void bar (IdString arg) {
         foo(arg);
    }

then the implementation of bar needs to construct a new IdString and
then destruct it afterwards. Unless YOSYS_NO_IDS_REFCNT is
defined (it's not defined by default), this means calls to
get_reference and then put_reference.

Compiling with GCC 9.3 with default settings, this patch decreases
.text size from 8562669 to 8346296 (~2.5% smaller).

### DIFF
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -273,12 +273,12 @@ bool RTLIL::Const::is_fully_undef() const
 	return true;
 }
 
-bool RTLIL::AttrObject::has_attribute(RTLIL::IdString id) const
+bool RTLIL::AttrObject::has_attribute(const RTLIL::IdString &id) const
 {
 	return attributes.count(id);
 }
 
-void RTLIL::AttrObject::set_bool_attribute(RTLIL::IdString id, bool value)
+void RTLIL::AttrObject::set_bool_attribute(const RTLIL::IdString &id, bool value)
 {
 	if (value)
 		attributes[id] = RTLIL::Const(1);
@@ -286,7 +286,7 @@ void RTLIL::AttrObject::set_bool_attribute(RTLIL::IdString id, bool value)
 		attributes.erase(id);
 }
 
-bool RTLIL::AttrObject::get_bool_attribute(RTLIL::IdString id) const
+bool RTLIL::AttrObject::get_bool_attribute(const RTLIL::IdString &id) const
 {
 	const auto it = attributes.find(id);
 	if (it == attributes.end())
@@ -294,7 +294,7 @@ bool RTLIL::AttrObject::get_bool_attribute(RTLIL::IdString id) const
 	return it->second.as_bool();
 }
 
-void RTLIL::AttrObject::set_string_attribute(RTLIL::IdString id, string value)
+void RTLIL::AttrObject::set_string_attribute(const RTLIL::IdString &id, const string &value)
 {
 	if (value.empty())
 		attributes.erase(id);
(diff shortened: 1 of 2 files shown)


ZirconiumX:
This was [discussed](https://github.com/YosysHQ/yosys/pull/1939#discussion_r418535806) and rejected by Claire; that performance improvement is why `YOSYS_NO_IDS_REFCNT` exists.

rswarbrick:
Oh, that's sad. Is the intention that `YOSYS_NO_IDS_REFCNT` becomes the default behaviour at some point in the future then? At the moment, the default behaviour is to generate unnecessarily slow code. This patch makes it less bad, pessimizing performance in a mode that isn't enabled at the moment.

If we don't want to do that, and aren't going to enable `YOSYS_NO_IDS_REFCNT` any time soon, maybe we should switch references on and off depending on `YOSYS_NO_IDS_REFCNT`? I think I can see how to do that cleanly with a macro. This will give the best behaviour in either case, at the cost of macro invocations in the code.

Please can someone weigh in on whether that's a good idea? If so, I can do it easily.

ZirconiumX:
> At the moment, the default behaviour is to generate unnecessarily slow code

At the moment, the default behaviour is to *catch bugs*. Yosys has a wide variety of internal bug-catching mechanisms - such as ref-counting strings - and sure, they're slow, but they have [caught](https://github.com/yosyshq/yosys/issues?q=is%3Aissue+assert+failed) a lot of bugs. I'm only a developer, but I am immensely grateful for bug-finding mode being the norm, and not the exception.

rswarbrick:
Sorry, I don't think I was very clear. I'm a great fan of the default behaviour being to catch bugs! My question is whether we can make it more performant while keeping the safety.

ZirconiumX:
So, I conducted an experiment, synthesising [the ULX3S SNES core](https://github.com/daveshah1/SNES_MiSTer_ulx3s/) with and without `-DYOSYS_NO_IDS_REFCNT`:

with refcounting (user time): 2m51.141s, 2m52.219s, 2m45.219s.
without refcounting (user time): 3m4.031s, 2m53.313s, 2m51.938s.

The noise levels of that are high, I admit, but refcounting simply isn't a notable hotspot, and optimising based on code size is *entirely* the wrong measurement to use here.

rswarbrick:
Is this after my patch? If so, that's exactly the behaviour that I would expect (and is consistent with what Claire wrote in the PR you linked to). Or have I misunderstood?

rswarbrick:
Following up after checking on IRC: The timing tests are without this patch, and demonstrate that refcounting doesn't have a significant impact on that benchmark. (I don't understand why the numbers actually get worse with the flag - maybe the change is just completely swamped by the noise)

I think the remaining question is whether we just do the increment/decrement refcount dance on every function entry (boo!) or do something clever to avoid it when refcounting is enabled, without passing stuff by reference when it isn't (to avoid the pessimisation that Claire mentioned in the linked discussion).

I'm happy either way, and am leaving this PR open to get some feedback.

clairexen:
> Is this after my patch? If so, that's exactly the behaviour that I would expect (and is consistent with what Claire wrote in the PR you linked to). Or have I misunderstood?

With `-DYOSYS_NO_IDS_REFCNT` there's no refcounting at all, thus it is an upper bound for any performance gain that can be achieved by decreasing the frequency of refcount increment/decrement.

I'm closing this PR since it claims to be a performance improvement but benchmarking seems to show that this is not a performance bottleneck.

PR #2085, opened 2020-05-26
Silence warning in select.cc and pass some more args by ref

rswarbrick:
With GCC 9.3, at least, compiling select.cc spits out a warning about
an implausible bound being passed to strncmp.

While tidying up the warning (and looking at the generated code), I
realised that we're creating several std::string objects and generally
repeating computations. This version should have the same behaviour,
but be rather more efficient.

Since I was visiting the file anyway, I also changed functions that
pass a std::string or standard library container by value to pass by
reference. The resulting code is ~6.4kb smaller when compiled with GCC
9.3 and standard settings.

### DIFF
--- a/kernel/register.h
+++ b/kernel/register.h
@@ -125,7 +125,7 @@ struct Backend : Pass
 };
 
 // implemented in passes/cmds/select.cc
-extern void handle_extra_select_args(Pass *pass, std::vector<std::string> args, size_t argidx, size_t args_size, RTLIL::Design *design);
+extern void handle_extra_select_args(Pass *pass, const std::vector<std::string> &args, size_t argidx, size_t args_size, RTLIL::Design *design);
 extern RTLIL::Selection eval_select_args(const vector<string> &args, RTLIL::Design *design);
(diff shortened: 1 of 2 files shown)


rswarbrick:
Ok. Would you like this split into 2 PRs or 2 commits in 1 PR?

whitequark:
Two commits are good.

rswarbrick:
Actually, I split into 3 ("fix warning"; "improve function"; "don't pass by reference"). The last touches the most code, but should be easy to review since it's essentially a find/replace to add "const ... &" to function declarations.

whitequark:
First and third commits look good to me, and the size reduction is welcome.

I didn't review the second too closely because both the original logic and the new logic are way too hard to understand to me; I'm not a fan of microoptimizations in obviously very cold code (`select` is a debugging command) so I'll let someone else review that.

rswarbrick:
@whitequark: Thanks for the review. If you don't have an opinion on the 2nd commit, could you clear the change request, so that the PR can get merged if someone else approves it?

whitequark:
Done

PR #2088, opened 2020-05-26
Minor optimisation in Module::wire() and Module::cell()

rswarbrick:
The existing code does a search to figure out whether id is in the
dict (with the call to count()), and then looks it up again to get the
result (with the call to at()). This version calls find() instead,
avoiding the double lookup.

Code size increases slightly (6kb). I think this is because the
contents of find() are getting inlined, and then inlined into lots of
the callsites for cell() and wire().

Looking at the compiled code before this patch, you just get
a (non-inlined) call to count() followed by a call to at(). After the
patch, the contents of find() have been inlined (so you see do_hash,
then do_lookup). The result for each function is about 30 bytes / 40%
bigger, which presumably also enlarges call-sites that inline it.

### DIFF
--- a/kernel/rtlil.h
+++ b/kernel/rtlil.h
@@ -1134,8 +1134,14 @@ public:
 		return design->selected_member(name, member->name);
 	}
 
-	RTLIL::Wire* wire(RTLIL::IdString id) { return wires_.count(id) ? wires_.at(id) : nullptr; }
-	RTLIL::Cell* cell(RTLIL::IdString id) { return cells_.count(id) ? cells_.at(id) : nullptr; }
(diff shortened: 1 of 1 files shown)


whitequark:
You say it's bigger, but is it actually faster?

rswarbrick:
I'd expect so on a micro-benchmark, since it does less work. The code size change is from GCC's inlining heuristics: I only reported it in case anyone noticed and wondered what was going on.

whitequark:
I guess there's no reason to not do this.

rswarbrick:
Yeah, I realise that it's not a big deal, but `count(...); at(...)` is a reasonably standard anti-pattern in C++ code, so I thought I may as well fix it as I went past.

PR #2091, opened 2020-05-27
Add `printattrs` command to print attributes of currently selected objects.

boqwxp:
AFAIK in general there is no way to know if an attribute is a string attribute, a bool attribute, or bit string attribute.  So the attribute value is printed first as `decode_string()` and then as `as_string()`, so all kinds of attribute values will be visible.

### DIFF
--- a/passes/cmds/printattrs.cc
+++ b/passes/cmds/printattrs.cc
@@ -34,6 +34,16 @@ struct PrintAttrsPass : public Pass {
 		log("\n");
 		log("\n");
 	}
+
+	static void log_const(const RTLIL::IdString &s, const RTLIL::Const &x, const unsigned int indent) {
+		if (x.flags == RTLIL::CONST_FLAG_STRING)
+			log("%s(* %s=\"%s\" *)\n", stringf(stringf("%%%ds", indent).c_str(), " ").c_str(), log_id(s), x.decode_string().c_str());
+		else if (x.flags == RTLIL::CONST_FLAG_NONE)
+			log("%s(* %s=%s *)\n", stringf(stringf("%%%ds", indent).c_str(), " ").c_str(), log_id(s), x.as_string().c_str());
+		else
+			log_assert(x.flags == RTLIL::CONST_FLAG_STRING || x.flags == RTLIL::CONST_FLAG_NONE); //intended to fail
+	}
+
 	void execute(std::vector<std::string> args, RTLIL::Design *design) YS_OVERRIDE
 	{
 		size_t argidx = 1;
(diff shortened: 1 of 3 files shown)


whitequark:
> AFAIK in general there is no way to know if an attribute is a string attribute, a bool attribute, or bit string attribute.

There is: the `flags` member of `RTLIL::Const`.

boqwxp:
Oh, I see, thanks!  I'll update this PR appropriately.

whitequark:
Take a look at `connect_rpc` or `write_cxxrtl` to see how to interpret the flags field correctly.

passes/cmds/printattrs.cc
```diff
@@ -0,0 +1,86 @@
...
+		log("    printattrs [selection]\n");
+		log("\n");
+		log("Print all attributes of the selected objects.\n");
+		log("\n");
+		log("\n");
+	}
+
+	static void log_const(const RTLIL::IdString &s, const RTLIL::Const &x, const unsigned int indent) {
+		if (x.flags == RTLIL::CONST_FLAG_STRING)
+			log("%s(* %s=\"%s\" *)\n", stringf(stringf("%%%ds", indent).c_str(), " ").c_str(), log_id(s), x.decode_string().c_str());
```
whitequark, line 40: Can you extract `stringf(stringf("%%%ds", indent).c_str(), " ")` into a function? It's repetitive and it wasn't immediately obvious to me what it does.

boqwxp: Done.

boqwxp:
The pass is useful because it is the only way to view attributes in scripts or from the REPL (or am I wrong? If so, please tell me how!).  It exists because I needed that ability while I was developing/testing/debugging b3040c31e2d3d9dd9b12173a75cebccda9f74e42.

whitequark:
> Don't you think it's a bit odd to have a `setattr` command but not a `printattrs` command?

Oh, that makes sense! I normally print the RTLIL and use my text editor.

boqwxp:
> Oh, that makes sense! I normally print the RTLIL and use my text editor.

Good call, I wish I'd have thought of that.  Anyway, it's done now, and I still think it makes sense to have this command, and maybe a `printparams` command, too.

passes/cmds/printattrs.cc
```diff
@@ -0,0 +1,92 @@
...
+		log("\n");
+		log("Print all attributes of the selected objects.\n");
+		log("\n");
+		log("\n");
+	}
+
+	static std::string get_indent_str(const unsigned int indent) {
+		//Build the format string (e.g. "%4s")
+		std::string format_str = stringf("%%%ds", indent);
+		return stringf(format_str.c_str(), " "); //Use the format string with " " as %s
```
Xiretza, line 41:
```suggestion
		return stringf("%*s", indent, "");
```

Would seem a bit cleaner to me.

passes/cmds/printattrs.cc
```diff
@@ -0,0 +1,90 @@
...
+	PrintAttrsPass() : Pass("printattrs", "print attributes of selected objects") { }
+	void help() YS_OVERRIDE
+	{
+		//   |---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|
+		log("\n");
+		log("    printattrs [selection]\n");
+		log("\n");
+		log("Print all attributes of the selected objects.\n");
+		log("\n");
+		log("\n");
```
whitequark, line 35: Stray newline? Passes without options only have one at the end, see `blackbox`.

whitequark: Actually it looks like passes are totally inconsistent on this aspect.

PR #2092, opened 2020-05-27
Restrict RTLIL::IdString to not contain whitespace or control chars

whitequark:
This is an existing invariant (most backends can't cope with these) but one that was not checked or documented.

### DIFF
--- a/kernel/rtlil.h
+++ b/kernel/rtlil.h
@@ -150,9 +150,6 @@ namespace RTLIL
 			if (!p[0])
 				return 0;
 
-			log_assert(p[0] == '$' || p[0] == '\\');
-			log_assert(p[1] != 0);
-
 			auto it = global_id_index_.find((char*)p);
 			if (it != global_id_index_.end()) {
 		#ifndef YOSYS_NO_IDS_REFCNT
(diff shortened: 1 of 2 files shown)


kernel/rtlil.h
```diff
@@ -152,6 +152,8 @@ namespace RTLIL
 
 			log_assert(p[0] == '$' || p[0] == '\\');
 			log_assert(p[1] != 0);
+			for (const char *c = p; *c; c++)
+				log_assert(*c > ' ');
```
mwkmwkmwk, line 156: Two things from JF:

- what about characters `>= 0x7f`? Should the check just be `*c > ' ' && *c < 0x7f`?
- this (and probably the existing two asserts) would be better positioned in the insert path (after the following `if`), so that we don't run it for every lookup

whitequark: > * what about characters `>= 0x7f`? Should the check just be `*c > ' ' && *c < 0x7f`?

I see no particular reason to forbid extended charset identifiers at the level of Yosys core. Yosys doesn't care, some backends probably care but backends have to make sure they properly escape various special characters in identifiers anyway.

That said, I wouldn't oppose prohibiting them if there's consensus that it's the right thing.

> this (and probably the existing two asserts) would be better positioned in the insert path (after the following if), so that we don't run it for every lookup

Will do.

mwkmwkmwk: There's a little problem here though: since `char` may or may not be signed, you may or may not be already rejecting extended charset identifiers (because they could be negative) — if we want to allow them, the condition needs changing to `*c > ' ' || *c < 0`

whitequark: Argh, I forgot about that. Yeah, I actually meant `(unsigned)*c > (unsigned)' '`.

whitequark: Done

Issue #1942, opened 2020-04-16
Speed opt. Compared std 11, 14, 17, 2a.


Kmanfi:
## Steps to reproduce the issue
Compiled Yosys with different std versions.

Yosys 0.9+2406 (git sha1 bb42b641, clang 11.0.0 -O3 -DNDEBUG -fPIC)

`time make test`
std=c++11
real	11m13.547s
user	10m25.694s
sys	0m35.091s

std=c++14
real	10m58.735s
user	10m14.206s
sys	0m32.700s

std=c++17
real	10m33.387s
user	9m49.292s
sys	0m31.639s

std=c++2a
real	9m56.299s
user	9m12.764s
sys	0m31.032s

Would be nice to have a check of highest std available and use it.

PR #2093, opened 2020-05-27
qbfsat: Multiple bugfixes

boqwxp:
This PR fixes multiple bugs in the `qbfsat` optimization procedure and hole specialization.

The first commit, b3040c31e2d3d9dd9b12173a75cebccda9f74e42 , updates the specialization code to use a bit-precise mapping for hole/$anyconst values and use a much more robust solution file format.  This fixes a bug preventing solutions being used to specialize other modules, e.g. one part of a miter vs. the entire miter.  Solution files now have to ways to identify each bit of the solution: the source location of the corresponding $anyconst cell, and the name of the wire for the appropriate SigBit.  Wire name matches are preferred, but we fall back to matching source locations (in a robust way).

The second commit, fe48e6a0bc8f658055625b873fc35979d6a1d758 , fixes three bugs in the optimization procedure.
1. Infinite loop in the optimization procedure when the first solution found while maximizing is at zero.
2. A signed-ness issue when maximizing.
3. Erroneously entering bisection mode with no wire to optimize.

In my experience this PR is necessary in order to use ABC to optimize some problems, because the $anyconst port Y `SigSpec` can have multiple chunks pointing to different wires.

Depends:
----
- [x] PR #2016
- [x] PR #2017 
- [x] PR #2018 

(no strict dependence on these, but it would be a PITA to backport)

### DIFF
--- a/kernel/rtlil.h
+++ b/kernel/rtlil.h
@@ -754,6 +754,7 @@ struct RTLIL::SigBit
 	SigBit(const RTLIL::SigBit &sigbit) = default;
 	RTLIL::SigBit &operator =(const RTLIL::SigBit &other) = default;
 
+	std::string str() const;
 	bool operator <(const RTLIL::SigBit &other) const;
 	bool operator ==(const RTLIL::SigBit &other) const;
 	bool operator !=(const RTLIL::SigBit &other) const;
@@ -1546,6 +1547,13 @@ inline RTLIL::SigBit::SigBit(RTLIL::Wire *wire) : wire(wire), offset(0) { log_as
 inline RTLIL::SigBit::SigBit(RTLIL::Wire *wire, int offset) : wire(wire), offset(offset) { log_assert(wire != nullptr); }
 inline RTLIL::SigBit::SigBit(const RTLIL::SigChunk &chunk) : wire(chunk.wire) { log_assert(chunk.width == 1); if (wire) offset = chunk.offset; else data = chunk.data[0]; }
 inline RTLIL::SigBit::SigBit(const RTLIL::SigChunk &chunk, int index) : wire(chunk.wire) { if (wire) offset = chunk.offset + index; else data = chunk.data[index]; }
+
+inline std::string RTLIL::SigBit::str() const {
+	if (wire != nullptr)
+		return stringf("%s[%d]", wire->name.c_str(), offset);
+	else
+		return stringf("%u", data);
+}
 
 inline bool RTLIL::SigBit::operator<(const RTLIL::SigBit &other) const {
 	if (wire == other.wire)
--- a/kernel/rtlil.h
+++ b/kernel/rtlil.h
@@ -754,7 +754,6 @@ struct RTLIL::SigBit
 	SigBit(const RTLIL::SigBit &sigbit) = default;
 	RTLIL::SigBit &operator =(const RTLIL::SigBit &other) = default;
 
-	std::string str() const;
 	bool operator <(const RTLIL::SigBit &other) const;
 	bool operator ==(const RTLIL::SigBit &other) const;
 	bool operator !=(const RTLIL::SigBit &other) const;
@@ -1547,13 +1546,6 @@ inline RTLIL::SigBit::SigBit(RTLIL::Wire *wire) : wire(wire), offset(0) { log_as
 inline RTLIL::SigBit::SigBit(RTLIL::Wire *wire, int offset) : wire(wire), offset(offset) { log_assert(wire != nullptr); }
 inline RTLIL::SigBit::SigBit(const RTLIL::SigChunk &chunk) : wire(chunk.wire) { log_assert(chunk.width == 1); if (wire) offset = chunk.offset; else data = chunk.data[0]; }
 inline RTLIL::SigBit::SigBit(const RTLIL::SigChunk &chunk, int index) : wire(chunk.wire) { if (wire) offset = chunk.offset + index; else data = chunk.data[index]; }
-
-inline std::string RTLIL::SigBit::str() const {
-	if (wire != nullptr)
-		return stringf("%s[%d]", wire->name.c_str(), offset);
-	else
-		return stringf("%u", data);
-}
 
 inline bool RTLIL::SigBit::operator<(const RTLIL::SigBit &other) const {
 	if (wire == other.wire)
--- a/passes/sat/qbfsat.cc
+++ b/passes/sat/qbfsat.cc
@@ -231,8 +231,8 @@ void write_solution(RTLIL::Module *module, const QbfSolutionType &sol, const std
 }
 
 void specialize_from_file(RTLIL::Module *module, const std::string &file) {
-	YS_REGEX_TYPE hole_bit_assn_regex = YS_REGEX_COMPILE_WITH_SUBS("^(.*) ([0-9]+) ([^ ]*) \\[([0-9]+)] = ([01])$");
-	YS_REGEX_TYPE hole_assn_regex = YS_REGEX_COMPILE_WITH_SUBS("^(.*) ([0-9]+) ([^ ]*) = ([01])$"); //if no index specified
+	YS_REGEX_TYPE hole_bit_assn_regex = YS_REGEX_COMPILE_WITH_SUBS("^(.+) ([0-9]+) ([^ ]+) \\[([0-9]+)] = ([01])$");
+	YS_REGEX_TYPE hole_assn_regex = YS_REGEX_COMPILE_WITH_SUBS("^(.+) ([0-9]+) ([^ ]+) = ([01])$"); //if no index specified
 	YS_REGEX_MATCH_TYPE bit_m, m;
 	//(hole_loc, hole_bit, hole_name, hole_offset) -> (value, found)
 	dict<pool<std::string>, RTLIL::Cell*> anyconst_loc_to_cell;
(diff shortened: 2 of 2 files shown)


kernel/rtlil.h
```diff
@@ -728,6 +728,7 @@ struct RTLIL::SigBit
 	SigBit(const RTLIL::SigBit &sigbit) = default;
 	RTLIL::SigBit &operator =(const RTLIL::SigBit &other) = default;
 
+	std::string str() const;
```
clairexen, line 731: The canonical form of creating a string representation of a signal is `log_signal()`. (Yes, `log_signal()` would add a blank between the wire name and the index. If that's a problem then the pass should have an application specific function for turning a SigBit into a string. No global `str()` function with a format different from the canonical `log_signal()` format should be needed.)

passes/sat/qbfsat.cc
```diff
@@ -28,9 +28,16 @@
 USING_YOSYS_NAMESPACE
 PRIVATE_NAMESPACE_BEGIN
 
+static inline unsigned int difference(unsigned int a, unsigned int b) {
+	if (a < b)
+		return b - a;
+	else
+		return a - b;
+}
+
```
clairexen, line 37: Why not just `abs(a-b)`?

passes/sat/qbfsat.cc
```diff
@@ -91,7 +98,10 @@ void recover_solution(QbfSolutionType &sol) {
 			log_assert(YS_REGEX_NS::regex_search(loc, hole_loc_regex));
 			log_assert(YS_REGEX_NS::regex_search(val, hole_val_regex));
 #endif
-			sol.hole_to_value[loc] = val;
+			RTLIL::AttrObject tmp;
+			tmp.set_src_attribute(loc);
+			pool<std::string> loc_pool = tmp.get_strpool_attribute(ID::src);
+			sol.hole_to_value[loc_pool] = val;
```
clairexen, line 104: Shouldn't that be identical to `auto locs =  split_tokens(loc, "|"); pool<std::string> loc_pool(locs.first, locs.second); ...`? No need to create a temp AttrObject.

passes/sat/qbfsat.cc
boqwxp: Because `a` and `b` are `unsigned int`s.  If `b` > `a`, then `a - b` underflows (or maybe is undefined?  I am not familiar with the spec in this regard).  One option is to make `a` and/or `b` (in this case `success` and `failure` on line 470) signed `int`s, but that limits the maximum value that  `a` and `b` can have.  Another option is to use a bigger signed type, like `long long`, but that's not really the point.  I think a simple difference function is elegant, functionally obvious, has no signedness issues, and allows the use of the full `unsigned` value range.

FWIW using `abs(success - failure)` with `unsigned int`s on line 470 gives this compiler error:
```
passes/sat/qbfsat.cc: In function ‘{anonymous}::QbfSolutionType {anonymous}::qbf_solve(Yosys::RTLIL::Module*, const {anonymous}::QbfSolveOptions&)’:
passes/sat/qbfsat.cc:479:47: error: call of overloaded ‘abs(unsigned int)’ is ambiguous
   while (failure == 0 || abs(success - failure) > 1) {
                                               ^
In file included from /usr/include/c++/8/cstdlib:75,
                 from /usr/include/c++/8/ext/string_conversions.h:41,
                 from /usr/include/c++/8/bits/basic_string.h:6400,
                 from /usr/include/c++/8/string:52,
                 from /usr/include/c++/8/stdexcept:39,
                 from /usr/include/c++/8/array:39,
                 from /usr/include/c++/8/tuple:39,
                 from /usr/include/c++/8/bits/stl_map.h:63,
                 from /usr/include/c++/8/map:61,
                 from ./kernel/yosys.h:42,
                 from passes/sat/qbfsat.cc:20:
/usr/include/stdlib.h:837:12: note: candidate: ‘int abs(int)’
 extern int abs (int __x) __THROW __attribute__ ((__const__)) __wur;
            ^~~
In file included from /usr/include/c++/8/cstdlib:77,
                 from /usr/include/c++/8/ext/string_conversions.h:41,
                 from /usr/include/c++/8/bits/basic_string.h:6400,
                 from /usr/include/c++/8/string:52,
                 from /usr/include/c++/8/stdexcept:39,
                 from /usr/include/c++/8/array:39,
                 from /usr/include/c++/8/tuple:39,
                 from /usr/include/c++/8/bits/stl_map.h:63,
                 from /usr/include/c++/8/map:61,
                 from ./kernel/yosys.h:42,
                 from passes/sat/qbfsat.cc:20:
/usr/include/c++/8/bits/std_abs.h:78:3: note: candidate: ‘constexpr long double std::abs(long double)’
   abs(long double __x)
   ^~~
/usr/include/c++/8/bits/std_abs.h:74:3: note: candidate: ‘constexpr float std::abs(float)’
   abs(float __x)
   ^~~
/usr/include/c++/8/bits/std_abs.h:70:3: note: candidate: ‘constexpr double std::abs(double)’
   abs(double __x)
   ^~~
/usr/include/c++/8/bits/std_abs.h:61:3: note: candidate: ‘long long int std::abs(long long int)’
   abs(long long __x) { return __builtin_llabs (__x); }
   ^~~
/usr/include/c++/8/bits/std_abs.h:56:3: note: candidate: ‘long int std::abs(long int)’
   abs(long __i) { return __builtin_labs(__i); }
   ^~~
make: *** [Makefile:673: passes/sat/qbfsat.o] Error 1
```

passes/sat/qbfsat.cc
boqwxp: Yes, that should work fine, too.  This way, though, if the storage format of a strpool attribute changes (maybe there is a need to store `|` in some strpool attribute), this code will not need updating.  I don't really mind either way, so I'll change it if you prefer the `split_tokens()` approach.

kernel/rtlil.h
boqwxp: That's fair.  Actually, I think it's preferable to go that route anyway, as a space character is the only character I'm 100% positive will always be disallowed in wire names, and is visually much nicer as a separator character than `[`, `]`, and `#`.

whitequark: > as a space character is the only character I'm 100% positive will always be disallowed in wire names, and is visually much nicer as a separator character than `[`, `]`, and `#`.

To be clear, all characters except control and space are explicitly *allowed* in wire names--I spent some time clarifying this just recently.

boqwxp: I see, thanks for the clarification.  I updated the PR to ensure the wire name is always surrounded by spaces and is described in the recovery regex as `[^ ]*` instead of `.*`.

passes/sat/qbfsat.cc
whitequark: > This way, though, if the storage format of a strpool attribute changes (maybe there is a need to store `|` in some strpool attribute)

We'll have to escape it for a rigorous solution. On Linux the only thing that isn't valid in a path is `\0`, but using that as a separator is just asking for trouble.

boqwxp: OK, I changed it.

passes/sat/qbfsat.cc
```diff
@@ -187,113 +199,146 @@ void write_solution(RTLIL::Module *module, const QbfSolutionType &sol, const std
...
+	//location bit name = value
+	//location bit name [offset] = value
+	//
+	//where '[', ']', and '=' are literal symbols, "location" is the $anyconst cell source location attribute,
+	//"bit" is the index of the $anyconst cell, "name" is the `wire->name` field of the SigBit corresponding
+	//to the current bit of the $anyconst cell->getPort(ID::Y), "offset" is the `offset` field of that same
+	//SigBit, and "value", which is either '0' or '1', represents the assignment for that bit.
+	dict<std::pair<pool<std::string>, int>, RTLIL::SigBit> hole_loc_idx_to_sigbit = get_hole_loc_idx_sigbit_map(module, sol);
+	for (auto &x : sol.hole_to_value) {
+		std::string src_as_str = std::accumulate(x.first.begin(), x.first.end(), std::string(), [](const std::string &a, const std::string &b){return a + "|" + b;});
```
boqwxp, line 227: I would have liked to have used auto types for the lambda parameters, but that would require compiling with `--std=c++14`.  See also #1942 .

passes/sat/qbfsat.cc
```diff
@@ -187,113 +199,146 @@ void write_solution(RTLIL::Module *module, const QbfSolutionType &sol, const std
...
+	//1. By the source location of the $anyconst cells
+	//2. By the name(s) of the wire(s) connected to each SigBit of the $anyconst cell->getPort(ID::Y) SigSpec.
+	//
+	//Option 1 has the benefit of being very precise.  There is very limited potential for confusion, as long
+	//as the source attribute has been set.  However, if the source attribute is not set, this won't work.
+	//More importantly, we want to have the ability to port hole assignments to other modules with compatible
+	//hole names and widths.  Obviously in those cases source locations of the $anyconst cells will not match.
+	//
+	//Option 2 has the benefits previously described, but wire names can be changed automatically by 
+	//optimization or techmapping passes, especially when (ex/im)porting from BLIF for optimization with ABC.
```
whitequark, line 213: Would the `hdlname` attribute help here?

boqwxp: I'm not sure.

Maybe it could be used in place of source locations?  But source locations are still quite a good way to track an object; I'm not sure matching by `hdlname` gives any advantage.

I don't think it can replace regular names, with which we already first try to match: The point of name matching is to allow the use of the same hole assignments found by a solver to specialize a different module with compatible holes.  I understand the `hdlname` attribute is rightfully "sticky" and e.g. does not update when wires are renamed (right? please correct me if wrong), so matching with it might impede the kind of flexibility that name matching is trying to allow.  And if it doesn't impede flexibility, then it's not clear to me how it differs from regular name matching.

boqwxp: I can't think of any specific benefit it brings.  I'll keep it in mind and who knows, maybe a benefit will become clear at some point.  But for now at least, I think I'll leave it as it is.

PR #2098, opened 2020-05-29
Add codeowners file

nakengelhardt:
No ownership implied; on a voluntary basis, add yourself to this file to have github request a review from you when a file of interest is changed. You can add commits to this PR, or leave a comment here or on slack with lines you'd like me to add.

We also write down who was responsible for major external contributions, but anyone without write access will not automatically get notified by github if I understand correctly. It's still a convenient place to keep this information.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -23,6 +23,9 @@ passes/opt/opt_lut.cc          @whitequark
 # requests automatically, but we add information for other
 # contributors here too, so we know who to ask to take a look.
 
+# synth_intel_alm
+techlibs/intel_alm/*           @ZirconiumX
+
 # pyosys
 misc/*.py                      @btut
 
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,25 @@
+## CODE NOTIFICATIONS
+# Register yourself here to be notified about modifications
+# for any files you have an interest in/know your way around.
+
+# Each line is a file pattern followed by one or more users.
+# Both github usernames and email addresses are supported.
+# Order is important; the last matching pattern takes the most
+# precedence. Previous matches will not be applied.
+
+
+# PATH (can use glob)          USERNAME(S)
+
+passes/cmds/scratchpad.cc      @nakengelhardt
+
+
+
+## External Contributors
+# Only users with write permission to the repository get review
+# requests automatically, but we add information for other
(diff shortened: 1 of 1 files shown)


ZirconiumX:
> anyone without write access will not automatically get notified by github if I understand correctly.

It's a little annoying that I won't get pinged, but could you add `techlibs/intel_alm/* @ZirconiumX` (presumably as external contributor)?

whitequark:
It's possible to set things up so that external contributors *do* get pinged with GitHub actions.

nakengelhardt:
Oh nice. How does that work? Will that also pull the info from the codeowners file (in which case we'd want to be more careful about asking first before adding people) or do you have to set it up separately?

whitequark:
It's done by arbitrary code we can write, so basically whichever functionality you want is doable. Take a look at what [wasmtime does](https://github.com/bytecodealliance/wasmtime/tree/master/.github): they have auto-addition of labels, and then auto-subscription to labels, chained together.

nakengelhardt:
OK, so after testing a little, the precise behavior if a line lists a code owner that does not have write permissions is that the review request does not get sent automatically, but the line *does* override (and hence potentially disable) previous lines.

nakengelhardt:
Let's squash merge this PR at the end to not have so many redundant commits.

nakengelhardt:
This is currently not of much practical relevance, but putting `*` at the end of the path [explicitly excludes any subdirectories](https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#example-of-a-codeowners-file), did you mean to do that?

@whitequark  For `rpc` and `cxxrtl` there aren't any subdirectories yet, but would you want to be notified if someone adds one?

@ZirconiumX For `intel_alm` it won't be automatic anyway but I took the liberty of removing the `*` since you wrote the contents of the subdirectory `common/` as well...

ZirconiumX:
That's fine. In practice I'm the only person working on and using intel_alm
presently anyway, so I expect changes to the codebase there to be rare in
practice.

whitequark:
> @whitequark For `rpc` and `cxxrtl` there aren't any subdirectories yet, but would you want to be notified if someone adds one?

Yep, I didn't read the syntax of the codeowners file very carefully. Let's ditch `*`.

nakengelhardt:
OK, done.

PR #2101, opened 2020-06-01
Support asymmetric memories for verific frontend

mmicko:


whitequark:
How does this work? I thought Yosys core does not support asymmetric memories.

clairexen:
> How does this work? I thought Yosys core does not support asymmetric memories.

The `for (int i = 0; i < numchunks; i++)` loop creates multiple ports. (support asymmetric memories was already there, but only for powers-of-two sub-dimensions.)

PR #2105, opened 2020-06-03
Split `flatten` from `techmap` and simplify it

whitequark:
Although the two passes started out very similar, they diverged over time and now have little in common. Moreover, `techmap` is extremely complex while `flatten` does not have to be, and this complexity interferes with improving `flatten`.

This is a fairly large PR, and it is best reviewed commit by commit: it starts by copying the techmap code wholesale and then simplifying it to make it easier to convince yourself that the changes are indeed correct. It can also be helpful to append `?w=1` to the URL to omit whitespace changes from the output.

The only actual improvement to `flatten` in this PR is using topological order, but it sets the stage for adding support for `hdlname` and processes (without `proc`).

### DIFF
--- a/passes/techmap/flatten.cc
+++ b/passes/techmap/flatten.cc
@@ -249,7 +249,7 @@ struct FlattenWorker
 		}
 	}
 
-	bool flatten_module(RTLIL::Design *design, RTLIL::Module *module, pool<RTLIL::Cell*> &handled_cells)
+	bool flatten_module(RTLIL::Design *design, RTLIL::Module *module)
 	{
 		if (!design->selected(module) || module->get_blackbox_attribute(ignore_wb))
 			return false;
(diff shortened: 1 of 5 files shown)


whitequark:
The "topologically sort modules" commit started out as a simplification (because I didn't fully understand the logic) and ended up as more of an optimization (but now the logic is documented!). I hope that's ok.

passes/techmap/flatten.cc
```diff
@@ -0,0 +1,357 @@
...
+				if (tpl != nullptr) {
+					if (topo_modules.database.count(tpl) == 0)
+						worklist.insert(tpl);
+					topo_modules.edge(tpl, module);
+				}
+			}
+		}
+
+		if (!topo_modules.sort())
+			log_error("Cannot flatten a design that contains loops.\n");
```
daveshah1, line 341: This should probably be phrased better, `loops` to me means combinational loops rather than loops in the hierarchy

whitequark: Makes sense. Any suggestion for phrasing? `Cannot flatten a recursive design`?

daveshah1: `design containing recursive instantiations`?

whitequark:
Thanks a lot!

PR #2107, opened 2020-06-04
flatten: preserve original object names

whitequark:
This PR depends on #2120.

This PR adjusts `flatten` to track the exact original name in the `hdlname` attribute, and describes the semantics of that attribute in the manual.

This addition lays the groundwork for robust end-to-end tracking of public names, from frontend to bitstream. I would like to see a future where tooling can find out where every bit of every register ended up without ever having to guess, even though registers could have a name that contains dots, have been removed, duplicated, renamed to avoid conflicts, or split.

### DIFF
--- a/manual/CHAPTER_Overview.tex
+++ b/manual/CHAPTER_Overview.tex
@@ -193,6 +193,13 @@ Violating these rules results in a runtime error.
 
 All RTLIL identifiers are case sensitive.
 
+Some transformations, such as flattening, may have to change identifiers provided by the user
+to avoid name collisions. When that happens, attribute ``{\tt hdlname}`` is attached to the object
+with the changed identifier. This attribute contains one name (if emitted directly by the frontend,
+or is a result of disambiguation) or multiple names separated by spaces (if a result of flattening).
+All names specified in the ``{\tt hdlname}`` attribute are public and do not include the leading
+``\textbackslash``.
+
 \subsection{RTLIL::Design and RTLIL::Module}
 
 The RTLIL::Design object is basically just a container for RTLIL::Module objects. In addition to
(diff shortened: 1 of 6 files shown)


manual/CHAPTER_Overview.tex
```diff
@@ -193,6 +193,12 @@ \subsection{RTLIL Identifiers}
 
 All RTLIL identifiers are case sensitive.
 
+Some transformations, such as flattening, may have to change identifiers provided by the user
+to avoid name collisions. When that happens, attribute ``{\tt hdlname}`` is attached to the object
+with the changed identifier. The contents of this attribute follows the same rules as RTLIL
+identifiers, and in addition, multiple levels of hierarchy are represented by separating multiple
```
daveshah1, line 199: nitpick - can you add the 'single space character' convention to the `hdlname` bullet point in the README too?

whitequark: Done, thanks for pointing out!

whitequark:
@clairexen There is one unresolved question here: are internal names ever allowed, or going to be allowed, in the `hdlname` attribute? I'm not completely sure what the semantics of internal names is, but assuming they are *only* for unnamed wires produced by the frontend for temporaries, or by passes during optimization, it makes sense to make internal names irrepresentable in `hdlname` by stripping the leading `\` and dropping `hdlname` if any internal name is anywhere in a path. Currently the leading `\` is present but internal names are still not allowed.

clairexen:
@whitequark yes, I think it would make sense to only use hdlname for non-internal names and thus it's okay, and probably preferred by users I'd assume, to strip the leading `\\`.

whitequark:
Great, that's what I'm also leaning towards. I'll fix the PR to do this instead.

whitequark:
Fixed and ready for review.

whitequark:
Thanks for the quick review :)

PR #2112, opened 2020-06-04
Add latch detection for use_case_method in part-select write

### LINKED ISSUE #2040: Part-select write in always @* causes logic loop warnings

The following is a simplified version of the test case from #1990.

```
module test #(
        parameter BUG = 1
) (
        (* nowrshmsk = !BUG *)
        output reg [1:0] x
);
        wire z = 0;
        integer i;
        always @*
                for (i = 0; i < 2; i=i+1)
                        x[z^i] = z^i;
endmodule
```

Run with `yosys -p 'prep; show' test.v`.

## Expected behavior

This should simply reduce to `x = 2'b10`. (It does with `BUG = 0`.)

## Actual behavior

Current Yosys master (8ec3b6d) produces a circuit with a topological loop.

image

(Running `opt -full` will optimize the loop away, but in the current flows there will be a warning about the loop in the intermediate circuit before it is being optimized away.)

clairexen:
Fixes #2040

### DIFF
--- a/frontends/ast/ast.h
+++ b/frontends/ast/ast.h
@@ -257,6 +257,7 @@ namespace AST
 		bool mem2reg_check(pool<AstNode*> &mem2reg_set);
 		void mem2reg_remove(pool<AstNode*> &mem2reg_set, vector<AstNode*> &delnodes);
 		void meminfo(int &mem_width, int &mem_size, int &addr_bits);
+		bool detect_latch(const std::string &var);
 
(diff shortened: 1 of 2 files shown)


clairexen:
@dh73 you have done some tests on a previous version of this. Can you look at this version and verify that this is now correct for all those test cases? Also, can we add those test cases to `make test` or yosys-tests?

dh73:
Sure, I'll work on it.

PR #2113, opened 2020-06-05
cxxrtl: fix implementation of $sshr cell

### LINKED ISSUE #2111: cxxrtl backend uses unsigned shift for signed shift cell
As far as I can tell, this should use `value::sshr` or set the `Signed` template parameter to true.

https://github.com/YosysHQ/yosys/blob/352731df4e0bf344cd20533a5561bc8ba3e07853/backends/cxxrtl/cxxrtl.h#L922-L925

whitequark:
Fixes #2111.

### DIFF
--- a/backends/cxxrtl/cxxrtl.h
+++ b/backends/cxxrtl/cxxrtl.h
@@ -921,7 +921,7 @@ value<BitsY> sshr_uu(const value<BitsA> &a, const value<BitsB> &b) {
 
 template<size_t BitsY, size_t BitsA, size_t BitsB>
 value<BitsY> sshr_su(const value<BitsA> &a, const value<BitsB> &b) {
-	return a.template shr(b).template scast<BitsY>();
+	return a.template sshr(b).template scast<BitsY>();
 }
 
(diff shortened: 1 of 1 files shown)


programmerjake:
according to https://github.com/YosysHQ/yosys/blob/94c10353897c6b2b3f960bdd6647a5da9c1d9f2c/kernel/celltypes.h#L261-L293
`sshr_su` is the only one that actually translates to an arithmetic right shift, confusingly enough.

whitequark:
@programmerjake Great catch, thank you! Should be fixed

programmerjake:
LGTM

adding unit tests comparing with some other verilog implementation or with yosys's const evaluator would be good, though that would probably be a different PR

whitequark:
> adding unit tests comparing with some other verilog implementation or with yosys's const evaluator would be good, though that would probably be a different PR

The idea was to use randomized testing comparing nMigen's pysim, nMigen's cxxsim (once it arrives), Yosys' `eval`, and some Verilog simulator.

PR #2115, opened 2020-06-06
cxxrtl: add debug information to the C++ API, and add introspection via a new C API

whitequark:
See individual commits for details.

This PR has two main opportunities for improvement, to be addressed in future PRs:
  1. Memories cannot be updated in a race-free way either through the C++ debug information or the C API.
  2. Localized and elided public signals do not appear in the debug information and cannot be introspected.

This PR significantly benefits from (but does not depend on) #2107.


whitequark:
cc @programmerjake @bracketmaster @jfng

PR #2122, opened 2020-06-08
Support 2D bit arrays in structures. Optimise array indexing.

PeterCrozier:
Support bit slicing of array elements in structs or unions e.g.

    s.a[i][p:q]

Array indexing when the array stride is a power of two is now done with a shift rather than a multiply.

### DIFF
--- a/tests/svtypes/struct_array.sv
+++ b/tests/svtypes/struct_array.sv
@@ -1,9 +1,9 @@
 // test for array indexing in structures
 
 module top;
-	
+
 	struct packed {
-		bit [7:0] [7:0] a;	// 8 element packed array of bytes
+		bit [5:0] [7:0] a;	// 6 element packed array of bytes
 		bit [15:0] b;		// filler for non-zero offset
 	} s;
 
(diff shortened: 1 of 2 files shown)


whitequark:
This PR has two changes. Are they related? If not, can they be split into two separate commits?

PeterCrozier:
The changes were connected.  The main commit was to support 2D array indexing in structures which I had left as a TODO from my previous PR.  I just optimised a special case of this specific  indexing when the array stride was a power of 2 and a shift seemed to be better than a multiply.

I have no benchmarks to justify this guess, and it is far from performance critical code, so I have removed the shift substitution.

whitequark:
If Yosys optimization passes would not take care of a multiply-by-power-of-2 I would consider it a bug.

tests/svtypes/struct_array.sv
```diff
@@ -3,7 +3,7 @@
 module top;
 	
 	struct packed {
-		bit [5:0] [7:0] a;	// 6 element packed array of bytes
+		bit [7:0] [7:0] a;	// 8 element packed array of bytes
```
clairexen, line 6: Maybe there's value to testing both cases, not just replacing the test for 6 elements with a test for 8 elements?

PR #2128, opened 2020-06-09
flatten: accept processes

whitequark:
There is no reason not to do this (the code that accepts processes is actually as long as the code that would print the error message rejecting them), it makes the pass nicer to use, and it allows some interesting experiments with other passes like CXXRTL (turns out `-noproc -flatten` has 15% lower compile time, which I was only able to measure after doing this change).


whitequark:
Thanks!

PR #2133, opened 2020-06-09
Adding latch tests for shift&mask AST dynamic part-select enhancements

dh73:
Enhancing tests as per  #2112 request. This PR is mainly #2079 with test for latch detection mechanism.

This test is already enabled and passing
```
make -j$(nproc) test | tee log && grep -e "dynamic" log && grep -e "make test" log 
```
`Running dynamic_part_select.ys..
  Passed "make test".
`

### DIFF
--- a/tests/various/dynamic_part_select.ys
+++ b/tests/various/dynamic_part_select.ys
@@ -3,105 +3,105 @@ read_verilog ./dynamic_part_select/original.v
 proc
 rename -top gold
 design -stash gold
- 
+
 read_verilog ./dynamic_part_select/original_gate.v
 proc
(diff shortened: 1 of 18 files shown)


tests/various/dynamic_part_select.ys
```diff
@@ -3,18 +3,18 @@ read_verilog ./dynamic_part_select/original.v
 proc
 rename -top gold
 design -stash gold
-
+
```
clairexen, line 6: oops. :D what happened here?

whitequark:
Can we merge this? This PR only adds tests and they pass.

PR #2136, opened 2020-06-10
Allow constant function calls in for loops and generate if and case

### LINKED ISSUE #2123: Allow constant functions within generate if condition

Run Yosys on the following file (`read_verilog` is sufficient):
```Verilog
module Example(out);
    parameter OUTPUT = "FOO";
    output wire [11:0] out;
    function automatic [31:0] flip;
        input [11:0] inp;
        flip = ~inp;
    endfunction
    generate
        if (flip(OUTPUT) == flip("BAR")) /* trouble here */
            assign out = OUTPUT;
        else
            assign out = 0;
    endgenerate
endmodule

module top(out);
    wire [11:0] o1, o2, o3, o4;
    Example          e1(o1);
    Example #("FOO") e2(o2);
    Example #("BAR") e3(o3);
    Example #("BAZ") e4(o4);
    output wire [12 * 4 - 1 :0] out;
    assign out = {o1, o2, o3, o4};
endmodule
```

`top` is not necessary to reproduce the bug, but is useful to ensure the module is being instantiated correctly.

## Expected behavior

Given that `flip` is a constant function, its use should be permitted within the condition of an `if_generate_construct`. Yosys should successfully read the modules and produce a result which yields the appropriate values on the output port of `top`.

## Actual behavior

On ` Yosys 0.9+2406 (git sha1 83f84afc, clang 11.0.0 -fPIC -Os)` I get:
```
Generating RTLIL representation for module `\Example'.
test.v:9: ERROR: Condition for generate if is not constant!
```

It works as expected if the troublesome line is replaced with:
```Verilog
        if (OUTPUT == "BAR")
```
More interestingly, it also works if the line is replaced with:
```Verilog
        localparam X = flip(OUTPUT) == flip("BAR");
        if (X)
```

While the `localparam` workaround is serviceable for the time being, I wanted to make this issue known.

zachjs:
This resolves #2123. I'm not completely sure this is the right fix, but it seems to work without breaking anything.

### DIFF
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -1126,6 +1126,10 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
 			bool in_param_here = in_param;
 			if (i == 0 && (type == AST_REPLICATE || type == AST_WIRE))
 				const_fold_here = true, in_param_here = true;
+			if (i == 0 && (type == AST_GENIF || type == AST_GENCASE))
+				in_param_here = true;
+			if (i == 1 && (type == AST_FOR || type == AST_GENFOR))
+				in_param_here = true;
 			if (type == AST_PARAMETER || type == AST_LOCALPARAM)
 				const_fold_here = true;
 			if (i == 0 && (type == AST_ASSIGN || type == AST_ASSIGN_EQ || type == AST_ASSIGN_LE))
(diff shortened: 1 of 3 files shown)


zachjs:
Any update on this PR? I'd be happy to make changes if there are issues with this solution.

nakengelhardt:
Claire's very busy so she usually reviews things all at once, and unfortunately your PR came just a day after the last batch! Unofficial policy is that we try not to leave things lying around for more than two weeks, so I'll try to get her to take a look by next week.

zachjs:
I have updated this change to also allow constant function calls in the condition of a generate for loop.

zachjs:
I've further extended it to allow constant function calls in the conditions of procedural for loops.

PR #2147, opened 2020-06-11
intel_alm: direct M10K instantiation

Ravenslofty:
Much like #1957, this directly instantiates M10K cells for block RAM, avoiding the pesky altsyncram cell, and permitting ABC9 timings and sequential optimisations like in #2073.

The M10K simulation cell is a bit awkward, but the test complains if I try to use a memory directly. I'd appreciate any advice on how to tidy it up.


ZirconiumX:
cc @mwkmwkmwk

ZirconiumX:
The SAT verification appears to run out of memory on the Travis instances. I'll disable it for now.

dh73:
With all respect, I want to do just two comments.

1. The `altsyncram` is mainly used as a flexible way to select the block ram resources and even let Quartus to change the selection if logic does not fit due resource issues. If you don't make it available at least as blackbox, the front-end will not be compatible with other project that directly instantiates that quartus LPM (unless the user generates the netlist instead of the IP blackbox). A good approach would be something similar as is done here. converting the `altsyncram` into N blocks of `<family>_ram_cell` after `memory_bram` pass. The problem is that most of the designs out there references `altsyncram` and not the conversion itself. CycloneV is the exception, I think. At least I haven't seen any `altsyncram` reference on it.

2. If you're not using techmap pass after `M10K` inference, then you will have problems with `CFG_ABITS` and `CFG_DBITS` since they will be dumped with the fixed value defined in the renaming rules. I could be wrong, didn't looked it with detail.

ZirconiumX:
>     1. The `altsyncram` is mainly used as a flexible way to select the block ram resources and even let Quartus to change the selection if logic does not fit due resource issues.

At present `altsyncram` is passed through to Quartus, which achieves this. Long term, I'll make `altsyncram` a whitebox that Yosys turns into `$mem`.

> If you don't make it available at least as blackbox, the front-end will not be compatible with other project that directly instantiates that quartus LPM (unless the user generates the netlist instead of the IP blackbox).

This has been in Yosys master since `synth_intel_alm` was merged.

https://github.com/YosysHQ/yosys/blob/a5f0cb4eba5d78f31c31a7c21b0dd0309363cd38/techlibs/intel_alm/common/megafunction_bb.v#L54-L108

>     2. If you're not using techmap pass after `M10K` inference, then you will have problems with `CFG_ABITS` and `CFG_DBITS` since they will be dumped with the fixed value defined in the renaming rules. I could be wrong, didn't looked it with detail.

This is the purpose of `quartus_rename.v`, which maps the `MISTRAL_*` primitives to the Quartus equivalents.

dh73:
I apologize, I did not explained myself correctly. I hope this is more clear: 

top.sv
```verilog
`default_nettype none
module top #(parameter A=8, D=16)
   (input wire clk, we, input wire [A-1:0] ra, wa,
    input [D-1:0] di, output logic [D-1:0] d0);
   logic [D-1:0]  m [0:(2**A-1)];
      if (we) m[wa] <= di;
      d0 <= m[ra];
   end
endmodule
```

Command:
```
./yosys -p "read_verilog -sv top.sv; synth_intel_alm -top top -family cyclonev -vqm top.vqm"`
```

See `port_x_last_address`, should be `255`, also, address with is incremented by 1 bit.
```
  defparam \m.0.0.0 .logical_ram_name = "MISTRAL_M10K";
  defparam \m.0.0.0 .operation_mode = "dual_port";
  defparam \m.0.0.0 .port_a_address_width = 9;
  defparam \m.0.0.0 .port_a_data_width = 16;
  defparam \m.0.0.0 .port_a_first_address = 0;
  defparam \m.0.0.0 .port_a_first_bit_number = 0;
  defparam \m.0.0.0 .port_a_last_address = 65535;
```

ZirconiumX:
> See `port_x_last_address`, should be `255`, also, address with is incremented by 1 bit.

Except the M10K does not have an 8-bit address by 16-bit data mode; only a 9-bit address by 16-bit data mode. Yosys knows how to map the 8x16 to a 9x16 (add a zero at the top of the address), so you get a 9x16. 

While it may not look pretty, this is a legal and correct mapping, so there's no real problem.

dh73:
I am not sure of the veracity of that statement. The Intel memory table is not clear, but it doesn't states that a size, for example, `256 × 32` is only for a memory of 8x32, but that `256` words infer 1 hard memory block if read port is in range from 1 to 32. Therefore, a memory of ADDR=8 and DAT=16 fits perfectly in the `256x40` range. 

By selecting `9x16`, you're doubling the number of words to select the `512x16`, unused ports will need to be processed by Quartus, that might not be a big deal if the `port_N_logical_ram_depth` is updated accordingly, but, at least for me, that's uncertain. 

There's no need to synthesize anything, if you go and play with the megawizard, it will tell you the scenarios where Quartus will use 1 M10K depending on the memory size. Using the sizes from my example, see `Resource Usage`:

* For 1-port:
image

* For two port is the same:
image

PR #2148, opened 2020-06-11
intel_alm: DSP inference

Ravenslofty:
This adds basic support for the multiply-accumulate slices within the Cyclone V, using `mul2dsp.v` to map them. They're rather unusual in that a slice can be three 9x9 multipliers, two 18x18 multipliers or a 27x27 multiplier.  Ideally this would have a Yosys pass to attempt to pack adders and flops into the cell, but this works fine as a minimum viable product.

cc @mwkmwkmwk

### DIFF
--- a/techlibs/intel_alm/common/megafunction_bb.v
+++ b/techlibs/intel_alm/common/megafunction_bb.v
@@ -129,3 +129,31 @@ output [data_width-1:0] portbdataout;
 input ena0, clk0, clk1;
 
 endmodule
+
+(* blackbox *)
+module cyclonev_mac(ax, ay, resulta);
+
+parameter ax_width = 9;
+parameter ay_scan_in_width = 9;
+parameter result_a_width = 18;
+parameter operation_mode = "M9x9";
+
+input [ax_width-1:0] ax;
+input [ay_scan_in_width-1:0] ay;
+output [result_a_width-1:0] resulta;
+
+endmodule
+
+(* blackbox *)
+module cyclone10gx_mac(ax, ay, resulta);
+
+parameter ax_width = 18;
+parameter ay_scan_in_width = 18;
+parameter result_a_width = 36;
+parameter operation_mode = "M18X18_FULL";
+
+input [ax_width-1:0] ax;
+input [ay_scan_in_width-1:0] ay;
+output [result_a_width-1:0] resulta;
+
+endmodule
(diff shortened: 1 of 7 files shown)


techlibs/intel_alm/common/quartus_rename.v
```diff
@@ -119,3 +121,54 @@ module MISTRAL_MLAB(input [4:0] A1ADDR, input A1DATA, A1EN, CLK1, input [4:0] B1
 );
 
 endmodule
+
+
+module MISTRAL_MUL27X27(A, B, Y);
+
+parameter A_SIGNED = 0;
```
mwkmwkmwk, line 128: why do these cells have parameters here but not in `dsp_sim.v`?  IMO these should be dropped

techlibs/intel_alm/common/quartus_rename.v
```diff
@@ -119,3 +121,54 @@ module MISTRAL_MLAB(input [4:0] A1ADDR, input A1DATA, A1EN, CLK1, input [4:0] B1
...
+
+endmodule
+
+
+module MISTRAL_MUL18X18(A, B, Y);
+
+parameter A_SIGNED = 0;
+parameter B_SIGNED = 1;
+parameter A_WIDTH = 18;
+parameter B_WIDTH = 19;
```
mwkmwkmwk, line 148: this doesn't match `dsp_sim.v`

techlibs/intel_alm/common/dff_sim.v
```diff
@@ -61,12 +61,14 @@ module MISTRAL_FF(
 `ifdef cyclonev
 specify
     (posedge CLK => (Q : DATAIN)) = 262;
+    (posedge CLK => (Q : SDATA)) = 262;
```
mwkmwkmwk, line 64: please do not sneak unrelated changes like that

techlibs/intel_alm/common/quartus_rename.v
Ravenslofty: I dropped them, yeah. Not entirely sure what I was thinking.

techlibs/intel_alm/common/megafunction_bb.v
```diff
@@ -129,3 +129,17 @@ output [data_width-1:0] portbdataout;
 input ena0, clk0, clk1;
 
 endmodule
+
```
mwkmwkmwk, line 132: shouldn't there be a cyclone10gx_mac as well?

PR #2149, opened 2020-06-11
cxxrtl: unbuffer output wires of toplevel module

whitequark:
Without unbuffering output wires of, at least, toplevel modules, it is not possible to have most designs that rely on IO via toplevel ports (as opposed to using exclusively blackboxes) converge within one delta cycle. That seriously impairs the performance of CXXRTL.


tomverbeure:
module outputs (led__red, led__green, ...) aren't present in the struct with this PR.

[ExampleTop.sim.ilang.gz](https://github.com/YosysHQ/yosys/files/4767949/ExampleTop.sim.ilang.gz)

PR #2154, opened 2020-06-13
cutpoint: Clarify control flow and improve efficiency

boqwxp:
This PR makes a number of changes to clarify control flow; multiple instances like:
```
for(...) {
    if (!cond)
        continue;
    //...
}
```

into the clearer:
```
for(...) {
    if (cond) {
        //...
    }
}
```

It also changes a loop like:
```
for (auto wire : module->wires()) {
    if (!wire->port_input)
        continue;
    //...
}
```

into the more efficient:
```
for (auto id : module->ports) {
    RTLIL::Wire *wire = module->wire(id);
    if (wire->port_input) {
        //...
    }
}
```

The PR is split into three commits for easier review, but I am happy to rebase this into one commit if preferred.

### DIFF
--- a/passes/sat/cutpoint.cc
+++ b/passes/sat/cutpoint.cc
@@ -94,11 +94,11 @@ struct CutpointPass : public Pass {
 					wire->port_id = 0;
 					wire->port_input = false;
 					wire->port_output = false;
-					continue;
+				} else {
+					log("Making wire %s.%s a cutpoint.\n", log_id(module), log_id(wire));
+					for (auto bit : sigmap(wire))
+						cutpoint_bits.insert(bit);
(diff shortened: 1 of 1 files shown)


nakengelhardt:
The downside of course is that it adds indentation levels, so that now an even greater portion of the default github viewport is occupied by whitespace (and everyone loves scrolling sideways...) I guess I'll put it to a vote at the thursday meeting.

clairexen:
We actually think that early exit is the cleaner way to go. We are not alone. For example, LLVM: https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code

PR #2156, opened 2020-06-14
Fix Verilator sim warnings: 1 BLKSEQ and 3 WIDTH

XarkLabs:
Sylvain "tnt" Munaut suggested I create this PR to solve a few warnings from Verilator in an ice40 simulation file when I asked about them. -Thanks!


nakengelhardt:
The original is also legal verilog but I don't see any reason not to merge this. @daveshah1 you can probably tell in one look, can you think of any?

daveshah1:
Can't see any problems here myself.

PR #2158, opened 2020-06-14
cxxrtl: fix sshr sign-extension.

### LINKED ISSUE #2157: cxxrtl: signed shr doesn't sign-extend correctly for values smaller than chunk size
When (signed) shr handles the final chunk
https://github.com/YosysHQ/yosys/blob/74e93e083ff23f3381fe2409e5847f9843840b17/backends/cxxrtl/cxxrtl.h#L382

if, for example, chunk::bits is 32, Bits is 16, and the shift is 1 then this will only fill bit 31 of the chunk.
The RHS should be equivalent to `chunk::mask << (Bits - shift_bits)` when the value is smaller than the chunk size.

Also I think there's an off-by-one, it should assign to `result.data[chunks - shift_chunks - 1]`.

miek:
This fixes #2157

I think this covers all the cases now. It works by finding the most significant chunk after the shift, calculating the number of bits in that chunk, then masking the rest. In the edge case where a small shift lands right on a chunk boundary, it will mask off the whole chunk above - I think that's reasonable so it didn't seem worth adding a special case.

This won't guarantee masking all chunks above the new "valid" region (ie: if a small shift moves the value out of chunk[1] and completely into chunk[0], this won't mask chunk[1]). I wasn't sure whether that was required as the result will get sliced down immediately anyway?

### DIFF
--- a/backends/cxxrtl/cxxrtl.h
+++ b/backends/cxxrtl/cxxrtl.h
@@ -376,10 +376,12 @@ struct value : public expr_base<value<Bits>> {
 				: data[chunks - 1 - n] << (chunk::bits - shift_bits);
 		}
 		if (Signed && is_neg()) {
-			for (size_t n = chunks - shift_chunks; n < chunks; n++)
+			size_t top_chunk_idx  = (Bits - shift_bits) / chunk::bits;
+			size_t top_chunk_bits = (Bits - shift_bits) % chunk::bits;
+			for (size_t n = top_chunk_idx + 1; n < chunks; n++)
 				result.data[n] = chunk::mask;
 			if (shift_bits != 0)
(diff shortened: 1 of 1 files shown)


whitequark:
>  I wasn't sure whether that was required as the result will get sliced down immediately anyway?

Could you elaborate here? I'm not sure what you mean by "sliced down immediately", the `sshr` operation is self-contained.

miek:
> Could you elaborate here? I'm not sure what you mean by "sliced down immediately", the `sshr` operation is self-contained.
Sorry, I was misreading what was going on in my generated code - I thought it had a call to slice() on every sshr_su result that was trimming it to size.

miek:
That should be good now - I altered the loop to mask all chunks above the most significant one.

PR #2159, opened 2020-06-14
cxxrtl: Implement chunk-wise multiplication

MerryMage:
Simple chunk-wise multiplication. Given typical bitsizes, using a more advanced algorithm seemed like overkill. Reduces to a single multiply for Bits <= 32.

---

Regarding overflow:

The maximal value in a `wide_result` element is `0x1'ffff'fffe`.
The maximal result of multiplication is `0xffff'fffe'0000'0001`.
Therefore their sum will never overflow.

### DIFF
--- a/backends/cxxrtl/cxxrtl.h
+++ b/backends/cxxrtl/cxxrtl.h
@@ -66,6 +66,7 @@ namespace cxxrtl {
 // Therefore, using relatively wide chunks and clearing the high bits explicitly and only when we know they may be
 // clobbered results in simpler generated code.
 typedef uint32_t chunk_t;
+typedef uint64_t wide_chunk_t;
 
 template<typename T>
 struct chunk_traits {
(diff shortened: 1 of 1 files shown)


backends/cxxrtl/cxxrtl.h
```diff
@@ -439,6 +440,27 @@ struct value : public expr_base<value<Bits>> {
...
+			}
+		}
+
+		for (size_t n = 0; n < result.chunks; n++) {
+			result.data[n] = wide_result[n];
+		}
+		result.data[result.chunks - 1] &= result.msb_mask;
+		return result;
+	}
+
```
whitequark, line 463: I'd probably put `mul` below the `*cmp` to group similar operations together, but this isn't important.

backends/cxxrtl/cxxrtl.h
```diff
@@ -1304,28 +1326,14 @@ value<BitsY> sub_ss(const value<BitsA> &a, const value<BitsB> &b) {
...
-		if (multiplier.bit(step)) {
-			multiplicand = multiplicand.shl(value<32> { multiplicand_shift });
-			product = product.add(multiplicand);
-			multiplicand_shift = 0;
-		}
-		multiplicand_shift++;
-	}
-	return product;
+	constexpr size_t BitsM = BitsA >= BitsB ? BitsA : BitsB;
+	return a.template zcast<BitsM>().mul<BitsY>(b.template zcast<BitsM>());
```
kbeckmann, line 1330: I had to make a small change here to make it build though, had to add `.template ` before the call to mul():

```
return a.template zcast<BitsM>().template mul<BitsY>(b.template zcast<BitsM>());
```

This also goes for mul_ss.

PR #2164, opened 2020-06-17
Get yosys building on Visual Studio

madebr:
These patches will make yosys build on MSVC 15 2017:

- The TRANSPARENT patch is similar to the one in the [mingw_fix branch](https://github.com/YosysHQ/yosys/tree/mingw_fix)
- The `__builtin_unreachable` patch can probably be generalized to a header to a `#define YOSYS_UNREACHABLE`, but since this is the only usage of `__builtin_unreachable`, I just did the change there.

### DIFF
--- a/frontends/verilog/preproc.cc
+++ b/frontends/verilog/preproc.cc
@@ -591,7 +591,7 @@ read_define_args()
 
 		default:
 			// The only FSM states are 0-2 and we dealt with 2 at the start of the loop.
-			__builtin_unreachable();
+			log_assert(false);
 		}
 	}
 
--- a/kernel/yosys.h
+++ b/kernel/yosys.h
@@ -117,11 +117,11 @@ extern Tcl_Obj *Tcl_ObjSetVar2(Tcl_Interp *interp, Tcl_Obj *part1Ptr, Tcl_Obj *p
 #    define PATH_MAX MAX_PATH
 #    define isatty _isatty
 #    define fileno _fileno
-#  else
-//   mingw includes `wingdi.h` which defines a TRANSPARENT macro
-//   that conflicts with X(TRANSPARENT) entry in kernel/constids.inc
-#    undef TRANSPARENT
 #  endif
+
+// mingw and msvc include `wingdi.h` which defines a TRANSPARENT macro
+// that conflicts with X(TRANSPARENT) entry in kernel/constids.inc
+#  undef TRANSPARENT
 #endif
 
 #ifndef PATH_MAX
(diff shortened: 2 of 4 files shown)


mmicko:
Please take a look at https://github.com/YosysHQ/yosys/blob/master/kernel/yosys.h#L123  can you please just make that undef to be for all WIN32 and not mingw, that should remove need to change all files using TRANSPARENT.

frontends/verilog/preproc.cc
```diff
@@ -591,7 +591,11 @@ read_define_args()
 
 		default:
 			// The only FSM states are 0-2 and we dealt with 2 at the start of the loop.
-			__builtin_unreachable();
```
whitequark, line 594: Just make this `log_assert(false);` instead of this mess.

madebr: I have applied your suggestion.

kernel/yosys.h
```diff
@@ -117,11 +117,11 @@ extern Tcl_Obj *Tcl_ObjSetVar2(Tcl_Interp *interp, Tcl_Obj *part1Ptr, Tcl_Obj *p
...
 #    define fileno _fileno
-#  else
-//   mingw includes `wingdi.h` which defines a TRANSPARENT macro
-//   that conflicts with X(TRANSPARENT) entry in kernel/constids.inc
-#    undef TRANSPARENT
 #  endif
+
+// mingw and msvc include `wingdi.h` which defines a TRANSPARENT macro
+// that conflicts with X(TRANSPARENT) entry in kernel/constids.inc
+#  undef TRANSPARENT
```
madebr, line 124: @mmicko 
What do you think about this change?

passes/techmap/extract_counter.cc
```diff
@@ -795,11 +795,11 @@ struct ExtractCounterPass : public Pass {
 		pool<RTLIL::IdString> _parallel_cells;
 		CounterExtractionSettings settings
 		{
-			.parallel_cells = _parallel_cells,
-			.maxwidth = 64,
-			.minwidth = 2,
-			.allow_arst = true,
-			.allowed_dirs = 0,
+			_parallel_cells,
+			64,
```
mwkmwkmwk, line 799: could you retain the field names as comments here, for clarity?

kernel/yosys.h
mmicko: yes, that should do it

passes/techmap/extract_counter.cc
madebr: Is there *prior art* about how to format these?
Since the `.editorconfig` does not specify `tab_width`, tabs cannot be used reliantly to align these.

mwkmwkmwk: *shrug* just `64, // maxwidth` would be enough

madebr: Thanks. I've found the `CodingReadme`.

PR #2168, opened 2020-06-17
Use (and ignore) the expression provided to log_assert in NDEBUG builds

whitequark:
This avoids warnings in NDEBUG builds emitted when a variable is only used in log_assert, but is always defined.

I've done this to avoid a warning introduced by commit 3c4e974, and fixed up the other places where this problem came up before.


eddiehung:
Note I came across this in the past: #1255; the preferred resolution was https://github.com/YosysHQ/yosys/commit/0c5db07cd6cc3c19b926da21a46599f97592b20f

whitequark:
I disagree that that "solution" is better.

mwkmwkmwk:
> I disagree that that solution is better. All it does is kick the problem down the road to the first person who uses a non-gcc-compatible compiler. With #2164 being merged this should be finally fixed properly.

Agreed.  Also, this has the major advantage of fixing it once and for all, instead of requiring lots of YS_ATTRIBUTE everywhere

PR #2172, opened 2020-06-18
Add `partition` command to do balanced hypergraph partitioning on a gate-level module

boqwxp:
This PR adds the `partition` command to do [balanced hypergraph partitioning](https://kahypar.org/#what-is-a-hypergraph-what-is-hypergraph-partitioning) on a flattened gate-level module.  Finding an optimal partitioning is generally an [NP-hard problem](https://en.wikipedia.org/wiki/Graph_partition#Problem_complexity), but fortunately there are good heuristics for it.

Miscellaneous changes:
----
This PR makes a few various enabling changes around the code base:
- ~Add a `Design::select()` interface for selecting whole modules~ (See PR #2178 )
- ~Add `operator+=()` to `dict<>::iterator` and `dict<>::const_iterator` and add `operator+()` and `operator+=()` to `ObjIterator`.  This enables indexing into `dict`s with an iterator, like `auto value = *(dict.begin() + offset)`, which helps reduce memory by avoiding the need for a node number -> cell `dict<int, Cell *>`~ (See PR #2177 )
- ~Break out `statdata_t` to the Yosys namespace.  This is to support the `-s` option, where we need to measure the size of our module~
- Add `unit_gate_cost()`, where all gates except NOT and BUF have a unit cost. This is the default cell type cost used for node weights in the hypergraph ~and for determining module size in the `-s` option~

Solvers, interface, integration:
----
[hMETIS](http://glaros.dtc.umn.edu/gkhome/metis/hmetis/overview) seems to be the most performant solver, but has a very restrictive license.  [KaHIP](https://github.com/KaHIP/KaHIP) has an MIT license but seems to have a nasty dependency on an old version of OpenMPI.  [minipart](https://github.com/Coloquinte/minipart2) is also MIT licensed, but brings in boost dependencies.

@daveshah1 has indicated that he is developing hypergraph partitioning in-tree in nextpnr, so when that is done, I think we all agree to bring that in-tree, and I will update this PR and mark it as ready for review.

Until then, this command makes a temporary directory, writes a `.hgr` file (spec in [this document](http://glaros.dtc.umn.edu/gkhome/fetch/sw/hmetis/manual.pdf)), invokes `minipart` (which must be on the `PATH`), and reads a `.sol` file with the partition assignments.

Depends:
----
- [x] PR #2135 
- [x] PR #2177 
- [x] PR #2178

### DIFF
--- a/kernel/cost.h
+++ b/kernel/cost.h
@@ -70,6 +70,28 @@ struct CellCosts
 		return db;
 	}
 
+	static const dict<RTLIL::IdString, int>& unit_gate_cost() {
+		static const dict<RTLIL::IdString, int> db = {
+			{ ID($_BUF_),     0 },
+			{ ID($_NOT_),     0 },
+			{ ID($_AND_),     1 },
+			{ ID($_NAND_),    1 },
+			{ ID($_OR_),      1 },
+			{ ID($_NOR_),     1 },
+			{ ID($_ANDNOT_),  1 },
+			{ ID($_ORNOT_),   1 },
+			{ ID($_XOR_),     1 }, //arguably should be higher
+			{ ID($_XNOR_),    1 }, //arguably should be higher
+			{ ID($_AOI3_),    2 },
+			{ ID($_OAI3_),    2 },
+			{ ID($_AOI4_),    3 },
+			{ ID($_OAI4_),    3 },
+			{ ID($_MUX_),     3 },
+			{ ID($_NMUX_),    3 }
+		};
+		return db;
+	}
+
 	dict<RTLIL::IdString, int> mod_cost_cache;
 	const dict<RTLIL::IdString, int> *gate_cost = nullptr;
 	Design *design = nullptr;
--- a/passes/cmds/partition.cc
+++ b/passes/cmds/partition.cc
@@ -363,7 +363,7 @@ struct PartitionWorker {
 struct PartitionPass : public Pass {
 	PartitionPass() : Pass("partition", "split techmapped module into equal-sized parts") { }
 
-	void help() YS_OVERRIDE
+	void help() override
 	{
 		//   |---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|
 		log("\n");
@@ -399,7 +399,7 @@ struct PartitionPass : public Pass {
 		log("\n");
 	}
 
-	void execute(std::vector<std::string> args, RTLIL::Design *design) YS_OVERRIDE
+	void execute(std::vector<std::string> args, RTLIL::Design *design) override
 	{
 		int opt_k = -1, opt_imbalance = 5;
 		bool opt_verbose = false, opt_nocleanup = false;
--- a/passes/cmds/partition.cc
+++ b/passes/cmds/partition.cc
@@ -1,0 +1,482 @@
+/*
+ *  yosys -- Yosys Open SYnthesis Suite
+ *
+ *  Copyright (C) 2020  Alberto Gonzalez <boqwxp@airmail.cc>
+ *
+ *  Permission to use, copy, modify, and/or distribute this software for any
+ *  purpose with or without fee is hereby granted, provided that the above
+ *  copyright notice and this permission notice appear in all copies.
+ *
+ *  THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ *  WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ *  MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ *  ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ *  WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ *  ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ *  OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *
+ */
+
+#include "kernel/yosys.h"
+#include "kernel/cost.h"
+
+USING_YOSYS_NAMESPACE
+PRIVATE_NAMESPACE_BEGIN
+
+struct PartitionPass : public Pass {
+	//Constructs a hypergraph from a flat techmapped module, partitions the hypergraph,
+	//recovers the cell partition assignments, creates partition modules with the appropriate
+	//cells, ports, and intra-partition connections, then replaces the original module
+	//with one that instantiates and connects the partition modules.
+	int opt_k, opt_imbalance;
+	bool opt_verbose, opt_nocleanup;
+	RTLIL::Module *module, *new_module;
+	RTLIL::Design *design;
+	dict<RTLIL::IdString, int> cell_type_costs;
+	dict<RTLIL::SigBit, int> sigbit_to_edgenum;
+	dict<int, pool<int>> edge_nodes;
+	dict<int, pool<int>> partition_nodes;
+	dict<int, bool> edge_cut;
+	pool<RTLIL::IdString> cut_wires;
+	int node_ctr, edge_ctr, cut_edge_ctr, partitions;
+	std::vector<RTLIL::Module *> partition_modules;
+
+	PartitionPass() : Pass("partition", "split techmapped module into equal-sized parts"), opt_k(-1), opt_s(-1),
+	                  opt_imbalance(5), opt_verbose(false), opt_nocleanup(false), module(nullptr), new_module(nullptr),
+	                  design(nullptr), cell_type_costs(CellCosts::unit_gate_cost()), node_ctr(0), edge_ctr(0),
+	                  cut_edge_ctr(0), partitions(2) { }
+
+	void reset() {
+		opt_k = -1, opt_imbalance = 5;
+		opt_verbose = false, opt_nocleanup = false;
+		module = nullptr, new_module = nullptr;
+		design = nullptr;
+		cell_type_costs = CellCosts::unit_gate_cost();
+		sigbit_to_edgenum.clear();
+		edge_nodes.clear();
+		partition_nodes.clear();
+		edge_cut.clear();
+		cut_wires.clear();
+		node_ctr = 0, edge_ctr = 0, cut_edge_ctr = 0, partitions = 2;
+		partition_modules.clear();
+	}
+
+	void create_hypergraph() {
+		//Initial set up on sigbit_to_edgenum for the cell output ports and on edge_nodes with just that output node for now:
+		for (auto cell : module->cells()) {
+			if (cell_type_costs.find(cell->type) == cell_type_costs.end())
+				log_cmd_error("Module must be techmapped; encountered unknown cell type %s\n", cell->type.c_str());
+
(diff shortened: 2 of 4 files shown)


passes/cmds/partition.cc
```diff
@@ -0,0 +1,501 @@
...
+	dict<RTLIL::IdString, int> cell_type_costs;
+	dict<RTLIL::SigBit, int> sigbit_to_edgenum;
+	dict<int, pool<int>> edge_nodes;
+	dict<int, pool<int>> partition_nodes;
+	dict<int, bool> edge_cut;
+	pool<RTLIL::IdString> cut_wires;
+	int node_ctr, edge_ctr, cut_edge_ctr, partitions;
+	std::vector<RTLIL::Module *> partition_modules;
+
+	PartitionPass() : Pass("partition", "split techmapped module into equal-sized parts"), opt_k(-1), opt_s(-1),
```
whitequark, line 44: Can you use C++11 member initializers (I think that's what they're called) instead of these gigantic lists that (a) produce warnings if options are out of order and (b) make it easy to omit a member leading to UB?

```
RTLIL::Design *design = nullptr;
```

passes/cmds/partition.cc
```diff
@@ -0,0 +1,501 @@
...
+	                  opt_imbalance(5), opt_verbose(false), opt_nocleanup(false), module(nullptr), new_module(nullptr),
+	                  design(nullptr), cell_type_costs(CellCosts::unit_gate_cost()), node_ctr(0), edge_ctr(0),
+	                  cut_edge_ctr(0), partitions(2) { }
+
+	void reset() {
```
whitequark, line 49: And this ought to be something like `*this = {};`, or otherwise use the `<Pass>Worker` pattern that's present elsewhere.

passes/cmds/partition.cc
```diff
@@ -0,0 +1,501 @@
...
+ */
+
+#include "kernel/yosys.h"
+#include "kernel/cost.h"
+
+USING_YOSYS_NAMESPACE
+PRIVATE_NAMESPACE_BEGIN
+
+struct PartitionPass : public Pass {
+	//Constructs a hypergraph from a flat techmapped module, partitions the hypergraph,
```
whitequark, line 27: I'm not sure I understand the "techmapped" restriction. Do you actually mean that the module is restricted to working on netlists with Yosys fine-grained cells like `$_AND_`?

passes/cmds/partition.cc
```diff
@@ -0,0 +1,501 @@
...
+
+	void create_hypergraph() {
+		//Initial set up on sigbit_to_edgenum for the cell output ports and on edge_nodes with just that output node for now:
+		for (auto cell : module->cells()) {
+			if (cell_type_costs.find(cell->type) == cell_type_costs.end())
+				log_cmd_error("Module must be techmapped; encountered unknown cell type %s\n", cell->type.c_str());
+
+			if (opt_verbose) log("Adding cell %s as node #%d\n", cell->name.c_str(), node_ctr);
+			for (auto port : cell->connections()) {
+				if (cell->output(port.first)) {
```
whitequark, line 72: The amount of rightward drift and nested conditionals in this code (here specifically and everywhere below) makes it excessively hard to review. After looking at this PR I find myself in strong agreement with [Claire's view](https://github.com/YosysHQ/yosys/pull/2154#issuecomment-646136574).

passes/cmds/partition.cc
```diff
@@ -0,0 +1,501 @@
...
+			dict<IdString, double> cell_area;
+			statdata_t module_stats(design, module, false, cell_area, "gates");
+			unsigned long num_gates = 0;
+			for (auto &it : module_stats.num_cells_by_type)
+				num_gates += cell_type_costs[it.first] * it.second;
+			partitions = num_gates / opt_s;
+			if (partitions <= 1)
+				return; //nothing to do
+		}
+		const std::string partitions_str = opt_k != 1? stringf(" -k %d", partitions) : "";
```
whitequark, line 164: Is there any benefit in having `const` here, or does it just add visual noise?

kernel/cost.h
```diff
@@ -111,6 +111,223 @@ struct CellCosts
...
+		}
+
+		for (auto &it : mod->processes) {
+			if (!design->selected(mod, it.second))
+				continue;
+			num_processes++;
+		}
+	}
+
+	void log_data(RTLIL::IdString mod_name, bool top_mod)
```
whitequark, line 250: This is an implementation detail of the `stat` pass and shouldn't be a part of global Yosys namespace. This change would make it harder to introduce the `stat -json` option proposed on IRC, for example.

I'm also not really convinced that using `statdata_t` wholesale is the right way to go here.

passes/cmds/partition.cc
boqwxp: You mean [initializer_list](https://en.cppreference.com/w/cpp/utility/initializer_list)?  Not sure I really see the benefit here.  Can you ELI5?

passes/cmds/partition.cc
boqwxp: Oh, good call. `*this = {};` or `*this = PartitionPass();` is much cleaner.

passes/cmds/partition.cc
whitequark: I mean [in-class member initializers](https://www.stroustrup.com/C++11FAQ.html#member-init).

passes/cmds/partition.cc
boqwxp: There are two restrictions:
1. Hierarchy is not supported, though it shouldn't be too hard to add.
2. The module must only contain cells that are contained in `cell_type_costs`.  See `kernel/costs.h` in this PR for the definition of `CellCosts::unit_gate_cost()` which has a list of cell types.  It is my understanding that the module generally must have been run through `techmap` for this to be the case.  TL;DR yes, I mean it's restricted to `$_AND_` etc.

passes/cmds/partition.cc
boqwxp: I think the problem is really GitHub's crappy UI.  This PR looks just fine to me even on a 1368x768 laptop screen with tabstop=4.

But if you're really insistent, I can obfuscate the control flow with `continue;` instead of `if()` guards etc.

passes/cmds/partition.cc
boqwxp: I suppose it's just visual noise.  Arguably anything beyond `auto` is visual noise...

passes/cmds/partition.cc
whitequark: Right. I would say that "techmapped" is confusing here because most Yosys techmapping does *not* produce those gates--it is done as a part of synthesis flows that produce vendor cells. Something like "fine grained" or "gate level" would be easier to understand.

Or just not make that restriction a part of the design short description, since it's not fundamental--I would very much like `partition` to work for CXXRTL some day. So, mentioning that currently it only works on gate level designs in the long description works too IMO.

kernel/cost.h
boqwxp: Well, there does need to be some way to measure the size of the module for the `-s` option.  But maybe this is controversial and problematic enough (and frankly unimportant to me) to just remove the `-s` option entirely.

passes/cmds/partition.cc
whitequark: Tabstops solve a part of the problem, but I consider nested conditionals obfuscation, not early exits.

passes/cmds/partition.cc
boqwxp: Oh shit, how have I missed that C++11 feature?! Yeah, that's way better.

passes/cmds/partition.cc
whitequark: Sometimes you're writing out the type because you use it in list-initialization; and sometimes you're casting (or avoiding casting) the result of a function call, where the undefinedness of out-of-range numeric conversions makes the explicitness important. But yes.

kernel/cost.h
whitequark: To be more clear, I think it's reasonable to have *some* cost function in Yosys core; I just don't it is reasonable to pul the implementation details of the `stat` pass as-is. If it's not important for you I would say ditch it, but I would also be happy to brainstorm a nicer way to do it than the current approach.

passes/cmds/partition.cc
whitequark: Also if you're feeling like doing it, I'd be on board with doing this conversion for the rest of Yosys.

passes/cmds/partition.cc
boqwxp: Nope, either one of those gives me:
```
ERROR: Assert `first_queued_pass == NULL' failed in kernel/register.cc:135.
```

kernel/cost.h
boqwxp: I think the target size idea is nice--hence my implementing it--but anyway I haven't had a clear idea how to work with a module so partitioned.  A lack of control flow in yosys scripts means there is no scriptable way to detect the number of created partitions and iterate over them (or whatever one needs to do with the created partitions).  And, in a REPL session, the user can run `stat` themselves and easily calculate the desired number of partitions.  So maybe it's not really so useful anyway.

passes/cmds/partition.cc
whitequark: I think that the `reset` function as you wrote it is error-prone in the long run (it requires any updates to class member initializers to be manually reflected in the reset function), and also not very nice conceptually because it effectively makes all of those members global variables, since that's how pass registration works.

I haven't looked into this issue before (I more or less cargo-culted the worker pattern in my passes), but given the above I think it is the reasonable solution here, barring a major redesign of the pass system.

kernel/cost.h
whitequark: Let's ditch it for now; we can always bring it back later.

passes/cmds/partition.cc
boqwxp: Agreed it is a maintenance burden.  Also a good point about initialization costs; I didn't consider that.  Anyway I gave up after 3 min trying to figure out the pass registration system to avoid it, so I'll just add a worker struct.

passes/cmds/partition.cc
boqwxp: Maybe some day.  But not today :P.

passes/cmds/partition.cc
boqwxp: I understand there's legitimate room for disagreement, and you seem insistent, so I'll change it.

whitequark: Thank you for accomodating!

passes/cmds/partition.cc
boqwxp: Fair enough. And I agree, it would be great to extend it to be more general.

passes/sat/qbfsat.cc
```diff
@@ -187,113 +200,146 @@ void write_solution(RTLIL::Module *module, const QbfSolutionType &sol, const std
...
+	//1. By the source location of the $anyconst cells
+	//2. By the name(s) of the wire(s) connected to each SigBit of the $anyconst cell->getPort(ID::Y) SigSpec.
+	//
+	//Option 1 has the benefit of being very precise.  There is very limited potential for confusion, as long
+	//as the source attribute has been set.  However, if the source attribute is not set, this won't work.
+	//More importantly, we want to have the ability to port hole assignments to other modules with compatible
+	//hole names and widths.  Obviously in those cases source locations of the $anyconst cells will not match.
+	//
+	//Option 2 has the benefits previously described, but wire names can be changed automatically by 
+	//optimization or techmapping passes, especially when (ex/im)porting from BLIF for optimization with ABC.
```
whitequark, line 214: Would the `hdlname` attribute help here?

boqwxp: This is outside the scope of this PR.  Can you please recreate the question on PR #2093?

PR #2173, opened 2020-06-18
Use C++11 final/override/[[noreturn]]

whitequark:
It's been almost 9 years since C++11 came out, the people who still don't have a working C++11 compiler should upgrade. Even RHEL comes with instructions on how to get a compiler that can build modern C++ software.


daveshah1:
As a data point, nextpnr uses C++11 and I don't think I've had a single complaint about that.

PR #2177, opened 2020-06-19
hashlib, rtlil: Add `operator+()` and `operator+=()` to `dict` iterators

boqwxp:
Currently, with `dict` iterators like you'd get from e.g. `module->cells().begin()`, you can increment them: `module->cells().begin()++` but you can't offset from them: `module->cells().begin() + 5`.  This PR changes that.

Specifically, this PR adds `operator+=()` to `dict<>::iterator` and `dict<>::const_iterator` and adds `operator+()` and `operator+=()` to `ObjIterator`. 

This PR is used in PR #2172 in which it helps to reduce memory usage by obviating the need for a `dict<int, Cell *>` mapping node numbers to cells with: `RTLIL::Cell *cell = *(module->cells().begin() += i);`  Note that one can use `operator+=()` on any temporary object to improve efficiency vs. `operator+()`, which makes a copy of the iterator object.

### DIFF
--- a/kernel/hashlib.h
+++ b/kernel/hashlib.h
@@ -363,7 +363,7 @@ public:
 	public:
 		const_iterator() { }
 		const_iterator operator++() { index--; return *this; }
-		const_iterator operator+=(int amt) { index -= amt; if(index < 0) index = -1; return *this; }
+		const_iterator operator+=(int amt) { index -= amt; return *this; }
 		bool operator<(const const_iterator &other) const { return index > other.index; }
 		bool operator==(const const_iterator &other) const { return index == other.index; }
 		bool operator!=(const const_iterator &other) const { return index != other.index; }
@@ -381,7 +381,7 @@ public:
 	public:
 		iterator() { }
 		iterator operator++() { index--; return *this; }
-		iterator operator+=(int amt) { index -= amt; if(index < 0) index = -1; return *this; }
+		iterator operator+=(int amt) { index -= amt; return *this; }
 		bool operator<(const iterator &other) const { return index > other.index; }
 		bool operator==(const iterator &other) const { return index == other.index; }
 		bool operator!=(const iterator &other) const { return index != other.index; }
--- a/kernel/hashlib.h
+++ b/kernel/hashlib.h
@@ -363,6 +363,7 @@ public:
 	public:
 		const_iterator() { }
 		const_iterator operator++() { index--; return *this; }
+		const_iterator operator+=(int amt) { index -= amt; if(index < 0) index = -1; return *this; }
 		bool operator<(const const_iterator &other) const { return index > other.index; }
 		bool operator==(const const_iterator &other) const { return index == other.index; }
 		bool operator!=(const const_iterator &other) const { return index != other.index; }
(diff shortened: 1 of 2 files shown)


kernel/hashlib.h
```diff
@@ -363,6 +363,7 @@ class dict
 	public:
 		const_iterator() { }
 		const_iterator operator++() { index--; return *this; }
+		const_iterator operator+=(int amt) { index -= amt; if(index < 0) index = -1; return *this; }
```
whitequark, line 366: What does the condition do?

boqwxp: `dict::end()` is defined as:
```
iterator end() { return iterator(nullptr, -1); }
```
i.e. an `index` of -1.

`dict::iterator::operator==()` is defined as:
```
bool operator==(const iterator &other) const { return index == other.index; }
```

So, in order to make things like `if(it != mydict.end())` work properly, any seek beyond the end of the dict must return an iterator with `index` exactly equal to -1.

whitequark: Isn't it generally a bug to seek past the end of an iterator?

boqwxp: Assuming we're basically copying STL container interfaces and semantics here, [here](https://en.cppreference.com/w/cpp/named_req/RandomAccessIterator) is the semantic behavior for `+=`:

```
difference_type m = n;

if (m >= 0) while (m--) ++r;
else while (m++) --r;
return r; 
```

So, seeking beyond the end should return the same thing as `mydict.end()++`.  I'm not actually sure what that is :).

whitequark: That's UB; it will be (nearly) always caught by tools like sanitizers or valgrind. The reason I asked is that I suggest removing the condition in its current form, and replacing it with either an assert (in which case we probably should add an assert to `operator++` too) or just nothing. The current behavior doesn't match STL containers and it can actually *impede* bug-finding by turning a crash into a subtle logic bug.

whitequark: (Yes, even constructing a pointer beyond one-past-the-end is UB, not just using it. This is a bit subtle.)

boqwxp: I see what you mean.  I'm OK with adding asserts.

boqwxp: Oh, crap, `log_assert()` isn't defined in `hashlib.h`... I guess you're right and it's best just to remove the check completely so that at least valgrind etc. can spot the bad behavior.

kernel/rtlil.h
```diff
@@ -554,6 +554,29 @@ namespace RTLIL
...
 		}
 
+		inline ObjIterator<T>& operator+=(int amt) {
+			log_assert(list_p != nullptr);
+			it += amt;
+			if (it == list_p->end()) {
+				(*refcount_p)--;
+				list_p = nullptr;
+				refcount_p = nullptr;
+			}
```
boqwxp, line 564: @whitequark I'm not sure how this is affected by d71a9badda20d48bcd4031f030d2d0901f837261 .  I guess it doesn't really matter as this is bad behavior anyway, but it would be nice to understand all the implications.

kernel/hashlib.h
whitequark: Note that Valgrind isn't a static analyzer.

kernel/rtlil.h
whitequark: Seems fine to me.

PR #2184, opened 2020-06-23
verilog_backend: Add handling for all FF types.

mwkmwkmwk:


mwkmwkmwk:
Note: I'm not quite convinced of the `$ff` / `$_FF_` handling, but it matches `simlib.v`.

mwkmwkmwk:
> Note: I'm not quite convinced of the `$ff` / `$_FF_` handling, but it matches `simlib.v`.

... scratch that, it breaks tests.  I'll just fall back to cell instantiation.

mwkmwkmwk:
@clairexen refactored to use FfData, changed to use blocking assignments for latches.

PR #2188, opened 2020-06-24
Add logic-assignments operators

kgugala:
This PR adds support for the following operators:

* `+=`
* `|=`
* `^=`


daveshah1:
Do you think you would be able to add the other combined assignment operators? It seems a bit inconsistent supporting += but not -=, for example.

kgugala:
> Do you think you would be able to add the other combined assignment operators? It seems a bit inconsistent supporting += but not -=, for example.

@daveshah1 makes sense, we'll add `&=` and `-=` to this PR. Later we can open more PRs with the rest of the combined operators.

PR #2195, opened 2020-06-26
Add a few more gate types to the manual.

mwkmwkmwk:
If anyone has some idea on how to insert linebreaks into `\lstinline` and thus fit the MUX4/8/16 cells into the main table, please let me know.  Also, if anyone has some idea how to insert a `$` character inside `\lstlisting`, please let me know as well.

I hate tex.

### DIFF
--- a/manual/CHAPTER_CellLib.tex
+++ b/manual/CHAPTER_CellLib.tex
@@ -490,6 +490,7 @@ source tree.
 \begin{tabular}[t]{ll}
 Verilog & Cell Type \\
 \hline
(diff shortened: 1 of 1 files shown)


boqwxp:
> Also, if anyone has some idea how to insert a $ character inside \lstlisting, please let me know as well.

Generally you just write `$`.  The problem has to do with `mathescape`.

whitequark:
@mwkmwkmwk The docs all look good to me. The TeX I hate no less than you and can't really help with.

PR #2200, opened 2020-06-29
expose pass fix

mmicko:
After change https://github.com/YosysHQ/yosys/commit/696660351f657a42ec5d4785d8ab547e0b568c94 expose pass was broken due to assertion caused by reference counter (code was adding new wires while looping trough them, which is not allowed). 

This change, adds new all new wires after loop have been finished.

### DIFF
--- a/passes/sat/expose.cc
+++ b/passes/sat/expose.cc
@@ -445,6 +445,8 @@ struct ExposePass : public Pass {
 
 			SigMap out_to_in_map;
 
+			std::map<RTLIL::Wire*, RTLIL::IdString> wire_map;
+
 			for (auto w : module->wires())
 			{
 				if (flag_shared) {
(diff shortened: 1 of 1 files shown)


passes/sat/expose.cc
```diff
@@ -474,12 +475,22 @@ struct ExposePass : public Pass {
...
 					}
 				}
 			}
+			for (auto &wm : wire_map)
+			{
+				if (flag_input) {
+						RTLIL::Wire *in_wire = module->addWire(wm.second, GetSize(wm.first));
+						out_to_in_map.add(wm.first, in_wire);
+				}
+				if (flag_cut) {
```
whitequark, line 488: Are you sure the logic here is the same? Above there is a `if (flag_input) ... else ... if (flag_cut)` construct, this is just two independent conditions. I would probably use two wire maps to make this change obviously correct.

mmicko: Actually it can even be placed in blocks bellow, since we have flag_input or flag_cut they are mutually exclusive.

whitequark: Ah ok if they are mutually exclusive that's fine

whitequark: Do you think you can also make it an error to pass both `-cut` and `-input`? Right now one of the flags is silently ignored which isn't very nice.

mmicko: Sure

PR #2201, opened 2020-06-29
Use ID macro to fix assertion

mmicko:
Fixes test case for test_cell -f with ilang file.
Without this change there is assertion in get_reference  of IdString ( log_assert(p[0] == '$' || p[0] == '\\'); )


whitequark:
How did this code ever work? I thought there is no way in Yosys to create an IdString not starting with `$` or `\`?

whitequark:
> How did this code ever work?

Oh right, it just used that identifier internally.

Issue #2198, opened 2020-06-29
test_pmgen -generate xilinx_srl.fixed failing


mmicko:
Due to change in https://github.com/YosysHQ/yosys/commit/86ab7d3a6ed994c06654fbae9770caa030f115d6 executing `test_pmgen -generate xilinx_srl.fixed` will fail with error trying to get non-existing parameter values for IS_C_INVERTED, IS_R_INVERTED and IS_D_INVERTED.

It is possible to fix crash and make test pass by change instance to something like `param(next, \IS_C_INVERTED, false)` but is it a proper fix ?

mwkmwkmwk:
Yes, it is — some FF types (particularly the `_1` types) just don't have the inversion parameters...

mwkmwkmwk:
So this doesn't seem to be a `xilinx_srl` bug to me — the problem here is that `xilinx_srl` expects the `+/xilinx/cells_sim.v` file to be loaded, so that the default parameter values are resolved automatically.  Running `read_verilog -lib +/xilinx/cells_sim.v` before running `test_pmgen` fixes the problem.

Perhaps we need a documentation fix instead?

mmicko:
Thanks. Added this into proper yosys-tests test case already

mwkmwkmwk:
This can be closed, right?

mmicko:
All fine and confirmed by yosys-test

PR #2202, opened 2020-06-29
xilinx_srl - fix default value for non-existing parameters

mmicko:
This fixes issue #2198


mwkmwkmwk:
Hold on, now that I look at it, it seems the correct logic is already there (only looking at the parameters for `FDRE` cells and not `FDRE_1`), perhaps something is amiss with the loaded sim models

PR #2215, opened 2020-07-02
qbfsat, smt2, smtio: Add `-solver-option` to allow specifying SMT-LIBv2 `(set-option ...)` commands

boqwxp:
`qbfsat` and `write_smt2` take `-solver-option` arguments that get emitted as `; yosys-smt2-solver-option` info statements.  In `smtio.py`, those info statements are translated into `(set-option ...)` commands.

Depends:
----
- [x] ~PR #2208~

### DIFF
--- a/backends/smt2/smtio.py
+++ b/backends/smt2/smtio.py
@@ -124,6 +124,7 @@ class SmtIo:
         self.timeout = 0
         self.produce_models = True
         self.smt2cache = [list()]
+        self.smt2_options = dict()
         self.p = None
         self.p_index = solvers_index
         solvers_index += 1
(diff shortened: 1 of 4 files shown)


backends/smt2/smtio.py
```diff
@@ -258,14 +259,17 @@ def setup(self):
...
             self.write("(set-option :produce-models true)")
 
         self.write("(set-logic %s)" % self.logic)
 
+        if self.forall and self.solver == "yices":
+            self.write("(set-option :yices-ef-max-iters 1000000000)")
+
+        for key, val in self.smt2_options.items():
+            self.write("(set-option {} {})".format(key, val))
+
```
boqwxp, line 272: Yices at least has some very... particular requirements for the order of `(set-option ..)` and `(set-logic ...)` commands.

The `(set-option :produce-models true)` *must* come before the `(set-logic BV)`, however, in Yices's `set-logic` SMT-LIBv2 handler, all other (except `:produce-assignments` and the `:yices-ef*`) options are set to their default values.  So if you want your `(set-option ...)` commands to have any effect, they must come *after* the `(set-logic BV)` (except for `:produce-models` as described above).

Again, this does *not* apply to the exists-forall options prefixed with `:yices-ef`--those do not get set to default values in the `set-logic` handler.  I moved the `:yices-ef-max-iters` block down just to keep it logically with the rest of the options.

nakengelhardt: Interesting, does that apply to `(set-option :random-seed ...)` as well? In which case I have a sby PR from yesterday I need to fix...

clairexen: Section 4.1.7 of SMT-LIB std 2.6 states that different options are only supported before or after `set-logic`. For example, `produce-models`, `produce-proofs`, `random-seed`, and many more are only allowed in "start mode", i.e. before `set-logic`.

boqwxp: Ah, great, thanks for pointing that out.  I will update the PR to filter out those `mode: start` options from Section 4.1.7 and emit them before `(set-logic ...)`.

PR #2219, opened 2020-07-03
Add newlines to help text for dfflegalize

rswarbrick:
I think these were probably missed by accident. Spotted because GCC
spits out lots of messages like this:

```
passes/techmap/dfflegalize.cc:114:7: warning: zero-length gnu_printf format string [-Wformat-zero-length]
  114 |   log("");
      |       ^~
```

(because we tell GCC that the first argument to log() looks like a
printf control string in log.h, and a zero length such string triggers
a warning).


rswarbrick:
If the the help text shouldn't include the newlines implied by the shape of the source code, I think we should just delete the `log("")` calls (but I think that makes the help text less readable, which is why I pushed up this version).

PR #2220, opened 2020-07-03
Respect \A_SIGNED for $shift

### LINKED ISSUE #2196: $shift and $shiftx ignore \A_SIGNED

```
$ cat /tmp/test.il
module \shifttest
  wire width 8 output 1 \Y_shift
  wire width 8 output 2 \Y_shr

  cell $shift \UUT_shift
    parameter \A_SIGNED 1
    parameter \A_WIDTH 4
    parameter \B_SIGNED 0
    parameter \B_WIDTH 4
    parameter \Y_WIDTH 8
    connect \A 4'1010
    connect \B 4'0001
    connect \Y \Y_shift
  end

  cell $shr \UUT_shr
    parameter \A_SIGNED 1
    parameter \A_WIDTH 4
    parameter \B_SIGNED 0
    parameter \B_WIDTH 4
    parameter \Y_WIDTH 8
    connect \A 4'1010
    connect \B 4'0001
    connect \Y \Y_shr
  end
end
$ ./yosys -p 'check; eval' /tmp/test.il

 |                                                                            |
 |                                                                            |
 |  Copyright (C) 2012 - 2020  Claire Wolf <claire@symbioticeda.com>          |
 |                                                                            |
 |  Permission to use, copy, modify, and/or distribute this software for any  |
 |  purpose with or without fee is hereby granted, provided that the above    |
 |  copyright notice and this permission notice appear in all copies.         |
 |                                                                            |
 |  THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES  |
 |  WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF          |
 |  MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR   |
 |  ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES    |
 |  WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN     |
 |  ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF   |
 |  OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.            |
 |                                                                            |

 Yosys 0.9+2406 (git sha1 b822beb1, gcc 10.1.0 -fPIC -Os)


-- Parsing `/tmp/test.il' using frontend `ilang' --

1. Executing ILANG frontend.
Input filename: /tmp/test.il

-- Running command `check; eval' --

2. Executing CHECK pass (checking for obvious problems).
checking module shifttest..
found and reported 0 problems.

3. Executing EVAL pass (evaluate the circuit given an input).
Eval result: \Y_shr = 8'01111101.
Eval result: \Y_shift = 8'00000101.

End of script. Logfile hash: 47cd0e26f9, CPU: user 0.01s system 0.00s, MEM: 16.33 MB peak
Yosys 0.9+2406 (git sha1 b822beb1, gcc 10.1.0 -fPIC -Os)
Time spent: 49% 1x eval (0 sec), 27% 1x read_ilang (0 sec), ...
```

## Expected behavior

`$shift` behaves the same as `$shr` for positive/unsigned `B` operands.

## Actual behavior

`$shr` sign-extends its `A` operand according to `\A_SIGNED` before shifting, `$shift` always zero-extends.

If this is the desired behaviour, `check` should error on an asserted `\A_SIGNED` for `$shift`/`$shiftx`, since it cannot possibly have the desired effect.

Xiretza:
This reflects the behaviour of `$shr`/`$shl`, which sign-extend their A operands to the size of their output, then do a logical shift (shift in 0-bits).

Fixes #2196, replaces #2217.

### DIFF
--- a/backends/btor/test_cells.sh
+++ b/backends/btor/test_cells.sh
@@ -6,7 +6,7 @@ rm -rf test_cells.tmp
 mkdir -p test_cells.tmp
 cd test_cells.tmp
 
-../../../yosys -p 'test_cell -n 5 -w test all /$alu /$fa /$lcu /$lut /$sop /$macc /$mul /$div /$mod /$divfloor /$modfloor'
+../../../yosys -p 'test_cell -n 5 -w test all /$alu /$fa /$lcu /$lut /$sop /$macc /$mul /$div /$mod /$divfloor /$modfloor /$shiftx'
 
 for fn in test_*.il; do
 	../../../yosys -p "
@@ -19,7 +19,7 @@ for fn in test_*.il; do
 		hierarchy -top main
 		write_btor ${fn%.il}.btor
 	"
-	boolectormc -kmax 1 --trace-gen --stop-first -v ${fn%.il}.btor > ${fn%.il}.out
+	btormc -kmax 1 --trace-gen --stop-first -v ${fn%.il}.btor > ${fn%.il}.out
 	if grep " SATISFIABLE" ${fn%.il}.out; then
 		echo "Check failed for ${fn%.il}."
 		exit 1
--- a/backends/verilog/verilog_backend.cc
+++ b/backends/verilog/verilog_backend.cc
@@ -750,21 +750,19 @@ bool dump_cell_expr(std::ostream &f, std::string indent, RTLIL::Cell *cell)
 		f << stringf(" = ");
 		if (cell->getParam(ID::B_SIGNED).as_bool())
 		{
-			f << stringf("$signed(");
-			dump_sigspec(f, cell->getPort(ID::B));
-			f << stringf(")");
+			dump_cell_expr_port(f, cell, "B", true);
 			f << stringf(" < 0 ? ");
-			dump_sigspec(f, cell->getPort(ID::A));
+			dump_cell_expr_port(f, cell, "A", true);
 			f << stringf(" << - ");
 			dump_sigspec(f, cell->getPort(ID::B));
 			f << stringf(" : ");
-			dump_sigspec(f, cell->getPort(ID::A));
+			dump_cell_expr_port(f, cell, "A", true);
 			f << stringf(" >> ");
 			dump_sigspec(f, cell->getPort(ID::B));
 		}
 		else
 		{
-			dump_sigspec(f, cell->getPort(ID::A));
+			dump_cell_expr_port(f, cell, "A", true);
 			f << stringf(" >> ");
 			dump_sigspec(f, cell->getPort(ID::B));
 		}
--- a/techlibs/common/techmap.v
+++ b/techlibs/common/techmap.v
@@ -141,6 +141,7 @@ module _90_shift_shiftx (A, B, Y);
 	parameter [B_WIDTH-1:0] _TECHMAP_CONSTVAL_B_ = 0;
 
 	localparam extbit = _TECHMAP_CELLTYPE_ == "$shift" ? 1'b0 : 1'bx;
+	wire a_padding = _TECHMAP_CELLTYPE_ == "$shiftx" ? extbit : (A_SIGNED ? A[A_WIDTH-1] : 1'b0);
 
 	generate
 `ifndef NO_LSB_FIRST_SHIFT_SHIFTX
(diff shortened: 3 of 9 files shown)


Xiretza:
Apart from the testsuite, I tested the changes with the following:

```bash
#!/bin/bash
set -xeu

cells=('$shl' '$shr' '$sshl' '$sshr' '$shift' '$shiftx')

./yosys -p "test_cell ${cells[*]}"
./yosys -p "test_cell -simlib ${cells[*]}"

cat > verilog_test.ys <<EOF
cd ..
design -stash orig
design -copy-from orig gate
write_verilog design.v
design -reset
read_verilog design.v
design -copy-to orig gate
design -load orig
cd gate
EOF

./yosys -p "test_cell -script verilog_test.ys ${cells[*]}"

cd backends/btor
./test_cells.sh
```

Are there any other areas to check?

backends/btor/test_cells.sh
```diff
@@ -6,7 +6,7 @@ rm -rf test_cells.tmp
 mkdir -p test_cells.tmp
 cd test_cells.tmp
 
-../../../yosys -p 'test_cell -n 5 -w test all /$alu /$fa /$lcu /$lut /$sop /$macc /$mul /$div /$mod /$divfloor /$modfloor'
+../../../yosys -p 'test_cell -n 5 -w test all /$alu /$fa /$lcu /$lut /$sop /$macc /$mul /$div /$mod /$divfloor /$modfloor /$shiftx'
```
mwkmwkmwk, line 9: Could you explain the rationale for changes to this file?

Xiretza: As far as I can tell, there is no way to test `$shiftx` with the method used in this file. The `gate` module is passed through `synth`, which collapses all the possible `x` outputs to either `0` or `1`. Meanwhile, the BTOR model for `gold` still produces `x` outputs, which are modeled as additional BTOR `input`s. Therefore, it is always possible to find some combination of BTOR inputs that collapses the undefs differently to how `synth` did it, resulting in a test failure.

`miter -equiv -ignore_gold_x` doesn't help either, because `== x` is just modeled as a comparison with another new input that can take on whatever value.

Xiretza:
Should be fixed now. `$shiftx` behaviour is now unchanged, with only an additonal `check` to make sure `\A_SIGNED` isn't set (it has no effect). `$shift` now behaves the same as `$shl`/`$shr` by first extending its `A` input to the width of `Y`, then shifting.

techlibs/common/techmap.v
```diff
@@ -160,7 +161,7 @@ module _90_shift_shiftx (A, B, Y);
 			localparam entries = (A_WIDTH+Y_WIDTH-1)/Y_WIDTH2;
 			localparam len = Y_WIDTH2 * ((entries+1)/2);
 			wire [len-1:0] AA;
-			wire [(A_WIDTH+Y_WIDTH2+Y_WIDTH-1)-1:0] Apad = {{(Y_WIDTH2+Y_WIDTH-1){extbit}}, A};
+			wire [(A_WIDTH+Y_WIDTH2+Y_WIDTH-1)-1:0] Apad = {{(Y_WIDTH2+Y_WIDTH-1){a_padding}}, A};
```
mwkmwkmwk, line 164: I don't think this is correct, but since this path is already broken in master for unrelated reasons (see #2345), I'll just ignore it and fix it myself later.

backends/verilog/verilog_backend.cc
```diff
@@ -860,19 +858,12 @@ bool dump_cell_expr(std::ostream &f, std::string indent, RTLIL::Cell *cell)
...
-
 		f << stringf("%s" "assign ", indent.c_str());
 		dump_sigspec(f, cell->getPort(ID::Y));
-		f << stringf(" = %s[", temp_id.c_str());
-		if (cell->getParam(ID::B_SIGNED).as_bool())
-			f << stringf("$signed(");
-		dump_sigspec(f, cell->getPort(ID::B));
-		if (cell->getParam(ID::B_SIGNED).as_bool())
-			f << stringf(")");
+		f << stringf(" = ");
```
mwkmwkmwk, line 863: This change needs to be rolled back — the intermediate wire is, in fact, necessary when A is not a whole wire (ie. a concatenation, or a bit select, or a constant — part selection in Verilog only works on registers / wires / memory elements, not on arbitrary expressions).

techlibs/common/techmap.v
Xiretza: I hardly understand what's going on here myself, so thanks for looking into that.

backends/verilog/verilog_backend.cc
Xiretza: fixed

mwkmwkmwk:
Merged, thanks!

PR #2224, opened 2020-07-04
sf2: Use dfflegalize.

mwkmwkmwk:
Note: I added support for FFs with clock enable while I'm at it, as well as latches — it seems to be correct based on the sim models, but please let me know if there's some issue that means we shouldn't actually be using it.

Also: is there support for initalized FFs on sf2? The current flow doesn't support them, and there is no matching parameter on the FF cell, but the code contains a commented out call to `sf2_ffinit` — or is that just a copypaste leftover that should be nuked?


mwkmwkmwk:
Have thrown in sync reset FFs as well.

PR #2226, opened 2020-07-04
efinix: Nuke efinix_gbuf in favor of clkbufmap.

mwkmwkmwk:


mwkmwkmwk:
@mmicko there's one thing that I need to ask though: why was `efinix_gbuf` only inserting buffers for external clocks (ie. directly wired to input ports)? The documentation says it can be used for all clocks, including core-generated ones, so did you have any particular reason for disallowing those?

mmicko:
@mwkmwkmwk was long time ago, but think this was like their own tool generated it. Tomorrow I will not be available but can test this for you and confirm with hardware

PR #2229, opened 2020-07-04
sf2: replace sf2_iobs with {clkbuf,iopad}map

Ravenslofty:
I had to add a lot of blackboxes to annotate everything, but this PR drops `sf2_iobs` for a combination of `clkbufmap` and `iopadmap`, handily adding tristate buffer support for free.


mwkmwkmwk:
(note: please do not merge this yet, this PR is pending a better version that depends on #2230 )

techlibs/sf2/synth_sf2.cc
```diff
@@ -209,10 +209,14 @@ struct SynthSf2Pass : public ScriptPass
 
 		if (check_label("map_iobs"))
 		{
-			if (help_mode)
-				run("sf2_iobs [-clkbuf]", "(unless -noiobs)");
-			else if (iobs)
-				run(clkbuf ? "sf2_iobs -clkbuf" : "sf2_iobs");
+			if (help_mode || iobs) {
+				if (help_mode || clkbuf) {
+					run("clkbufmap -buf $_BUF_ Y:A -inpad CLKBUF Y:PAD", "(if -clkbuf, unless -noiobs)");
```
mwkmwkmwk, line 214: now that #2230 has landed, these three lines can be replaced by `clkbufmap -buf CLKINT Y:A -inpad CLKBUF Y:PAD`

mmicko:
@Ravenslofty  Before we got CLKINT added for each clock signal, so simple design with dff for example had one. Since default is changed there is none, which is fine, but with -clkbuf option we just get CLKBUF and not CLKINT as before, not sure what vendor tools generate, but behavior of yosys have been changed.

mwkmwkmwk:
@mmicko Could you give us a simple example design where output changed?

mmicko:
module dff ( input d, clk, output reg q );
	always @( posedge clk )
            q <= d;
endmodule

this one generated CLKINT but not CLKBUF

PR #2232, opened 2020-07-05
gowin: Fix INIT values in sim library.

mwkmwkmwk:


pepijndevos:
Related note: Eddie suggested that techmap should fail if init and reset are incompatible, rather than leave the init value, which generates a warning. Not required to fix that in this PR, but something that needs to happen at some point maybe.

mwkmwkmwk:
Fun fact: this is exactly what `dfflegalize` does in #2239 (among many other useful things) :p

PR #2238, opened 2020-07-05
anlogic: Use dfflegalize.

mwkmwkmwk:
Notes:

1. This fixes the sim library to something sane (with proper clock gating and latch model). Hopefully this actually matches the hardware.  If not, the corresponding functionality needs to be disabled, because there's just no hope for it.
2. This adds support for mapping DFFs with both async reset and enable, with sync reset (and maybe enable), and proper latch support.  The new DFF types are not actually emitted yet — this will be done in the future `opt_dff` pass.

### DIFF
--- a/techlibs/anlogic/cells_sim.v
+++ b/techlibs/anlogic/cells_sim.v
@@ -10,9 +10,6 @@ module AL_MAP_SEQ (
 	parameter SRMUX = "SR"; //SR/INV
 	parameter SRMODE = "SYNC"; //SYNC/ASYNC
 
-	wire clk_ce;
-	assign clk_ce = ce ? clk : 1'b0;
-
 	wire srmux;
 	generate
 		case (SRMUX)
@@ -20,7 +17,7 @@ module AL_MAP_SEQ (
 			"INV": assign srmux = ~sr;
 			default: assign srmux = sr;
 		endcase
-	endgenerate	
+	endgenerate
 
 	wire regset;
 	generate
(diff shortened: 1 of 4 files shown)


techlibs/anlogic/cells_sim.v
```diff
@@ -38,40 +35,29 @@ module AL_MAP_SEQ (
...
 					if (srmux)
 						q <= regset;
-					else 
+					else if (ce)
 						q <= d;	
 			end 
 			else
 			begin
-				always @(posedge clk_ce)
+				always @(posedge clk)
```
daveshah1, line 46: Looks like as well as fixing the broken CE implementation, this also changes semantics from CE-over-SR to SR-over-CE. I think the latter is actually correct for Anlogic (generally it follows ECP5 semantics), but might be worth checking.

mmicko: Just double checked with their simulation model and David is right, there it is also CE over SR

mmicko: But that is just simulation, if you can see any reliable way of checking this with hardware I can check it.

mwkmwkmwk: Just instantiate AL_MAP_SEQ with the right parameters, wire the inputs straight to buttons/switches, outputs to LEDs, and get the full truth table out of that thing manually?

Also another thing I'd like to verify in hardware: is the enable of the latch indeed `clk & ce`? (for xilinx it would be `~clk & ce`, and there's a good hardware reason for that, so I'd like to double-check it)

mwkmwkmwk: I've spent some time reading through the `AL_PHY_*SLICE` vendor models, and they confirm what I thought about sync set/reset and clock enable, but the jury is still out on `clk` polarity on latches — on one hand, the hardware matches xilinx; on the other, it's quite plausible that the toolchain auto-inverts the polarity when `AL_MAP_SEQ` is used.  I guess I'll just have to test the damn thing on hardware, after all...

mwkmwkmwk: As it turns out, the LATCH does have active-low clk input after all.  I have tested the exact behavior of this cell on hardware, and updated `cells_sim.v` accordingly (in particular, note the lovely D-latch with sync reset).

PR #2239, opened 2020-07-05
gowin: Use dfflegalize.

mwkmwkmwk:


pepijndevos:
This looks like it makes stuff a lot simpler, but I have no idea how `dfflegalise` works, so need to do some study before I can meaningfully review this.

tests/arch/gowin/init.ys
```diff
@@ -38,8 +38,8 @@ chparam -set INIT 1 myDFF myDFFN myDFFE myDFFNE
 
 # async should give a warning
```
pepijndevos, line 39: This comment is now incorrect.

tests/arch/gowin/init.ys
```diff
@@ -38,8 +38,8 @@ chparam -set INIT 1 myDFF myDFFN myDFFE myDFFNE
...
 # sync should synth to a mux
-chparam -set INIT 0 myDFF*S* myDFF*P*
-chparam -set INIT 1 myDFF*R* myDFF*C*
+chparam -set INIT 0 myDFF*S*
+chparam -set INIT 1 myDFF*R*
```
pepijndevos, line 42: Can you explain the effect of removing the Preset/Clear async flops here? If I'm understanding the docs correctly, this should be handled by the `- emulating adff when the (reset, init) value combination is unsupported by
  dff + adff + dlatch + mux` case. In this case, I'd expect it to be reflected in changes to the following `-assert-count` lines. Or do these cases produce an error now? Is there a way to test that they do in fact result in an error?

mwkmwkmwk: That would be true if `synth_gowin` actually supported latches — without them, it's an error.

pepijndevos: Ah true. I guess I'll add those later. Should be fairly easy. I guess all I need to do once I add those is add a `-cell $_DLATCH_[NP]_ ???` directive to tell it these are supported.

mwkmwkmwk: Depends, what capabilities do these latches have? If they're anything like FFs (ie. single s/r pin of const polarity, varying polarity enable, init tied to reset value), then `-cell $_DLATCH_?P?_ r` would be correct

tests/arch/gowin/init.ys
mwkmwkmwk: Removed

tests/arch/gowin/init.ys
mwkmwkmwk: I've added an error test for now.

PR #2243, opened 2020-07-06
dfflibmap: Refactor to use dfflegalize internally.

mwkmwkmwk:


mwkmwkmwk:
Note: code should be ready, but I still need to add proper tests...

mwkmwkmwk:
Also, a few notes:

- this pass runs dfflegalize only on actual FFs (`$_DFF*` and `$_SDFF*`), and ignores latches — this is for compatibility with how it behaved before (in particular, coolrunner2 relies on this behavior, because it maps latches via techmap instead)
- the options passed to dfflegalize imply that all cells involved have full initialization capabilities — this is again for compatibility (coolrunner2 later runs `dffinit` to fix that up)

I'm not particularly happy with this mess, and perhaps this pass should be deprecated and/or grow more options to better control its behavior.  Thoughts?

PR #2260, opened 2020-07-12
proc_dlatch: Remove init values for combinatorial processes.

### LINKED ISSUE #2258: DFFLEGALIZE pass: ERROR: Conflicting init values for signal 1'0

I'm building ECP5 SoC's with Litex/migen. Yosys (b822beb1) from 2 weeks is working, but when using Yosys `eef0ec6a` it fails.

I've created a minimal example which produces the same failure: https://github.com/gregdavill/OrangeCrab-examples/blob/minimal/litex/SoC-minimal.py

I have attached the generated files from Litex which are enough to reproduce the issue with just yosys.
[gateware.zip](https://github.com/YosysHQ/yosys/files/4906551/gateware.zip)

This was filed as an issue to me: https://github.com/gregdavill/OrangeCrab-examples/issues/3
The reporter suggests that this recent commit might be to blame: https://github.com/YosysHQ/yosys/commit/372521ca56c2ca27bb562098aa65b4e457959e77

## Expected behavior

When running `yosys orangecrab.ys` I would expect it to complete and produce a `.json` file. 
This is the behaviour of an older version which I was previously using (b822beb1, ~15 days old)

## Actual behavior

Running `yosys orangecrab.ys` results in the following error and terminates.

```
4.39. Executing DFFLEGALIZE pass (convert FFs to types supported by the target).
ERROR: Conflicting init values for signal 1'0 (\interface2_bank_bus_dat_r [2] = 1'x != 1'0).
```

mwkmwkmwk:
Fixes #2258.

### DIFF
--- a/passes/proc/proc_dlatch.cc
+++ b/passes/proc/proc_dlatch.cc
@@ -37,6 +37,7 @@ struct proc_dlatch_db_t
 	dict<Cell*, vector<SigBit>> mux_srcbits;
 	dict<SigBit, pair<Cell*, int>> mux_drivers;
 	dict<SigBit, int> sigusers;
+	dict<SigBit, std::pair<State,SigBit>> initbits;
 
 	proc_dlatch_db_t(Module *module) : module(module), sigmap(module)
 	{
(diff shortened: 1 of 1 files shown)


mwkmwkmwk:
Merging this without review due to serious master breakage.

PR #2282, opened 2020-07-20
Only allow "sat" and "unsat" smt solver responses in yosys-smtbmc

clairexen:
See https://github.com/YosysHQ/SymbiYosys/pull/102. cc @boqwxp

### DIFF
--- a/backends/smt2/smtbmc.py
+++ b/backends/smt2/smtbmc.py
@@ -1275,10 +1275,10 @@ def smt_pop():
     asserts_consequent_cache.pop()
     smt.write("(pop 1)")
 
-def smt_check_sat():
+def smt_check_sat(expected=["sat", "unsat"]):
     if asserts_cache_dirty:
         smt_forall_assert()
-    return smt.check_sat()
+    return smt.check_sat(expected=expected)
 
 if tempind:
     retstatus = "FAILED"
--- a/backends/smt2/smtio.py
+++ b/backends/smt2/smtio.py
@@ -653,7 +653,7 @@ class SmtIo:
 
         return stmt
 
-    def check_sat(self):
+    def check_sat(self, expected=["sat", "unsat", "unknown", "timeout", "interrupted"]):
         if self.debug_print:
             print("> (check-sat)")
         if self.debug_file and not self.nocomments:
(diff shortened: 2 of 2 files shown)


boqwxp:
I'm not sure this is the best way to solve the issue.  `check_sat()` is only ever called by `smt_check_sat()`, so why copy the whole list of `["sat", "unsat", "unknown", "timeout", "interrupted"]` to the default argument when it will only ever be called with that argument set to `["sat", "unsat"]`?  Why not just eliminate the default argument, or change it to just `["sat", "unsat"]`?

FWIW, in the context of the `qbfsat` pass, everything still works (i.e., there are no crashes and `retstatus` is still `FAILED`, so the exit code doesn't change).  Here is how the output changes:

Original:
```
smtbmc output: ##   0:00:00  Solver: yices
smtbmc output: ##   0:00:03  Solving for step 0..
smtbmc output: ##   0:00:13  No solution found! (interrupted)
smtbmc output: ##   0:00:13  Status: FAILED
Solver finished in 13.742 seconds.
Warning: solver timed out
```

With this PR:
```
smtbmc output: ##   0:00:00  Solver: yices
smtbmc output: ##   0:00:03  Solving for step 0..
smtbmc output: ##   0:00:13  Unexpected response from solver: interrupted
Solver finished in 13.890 seconds.
```

The `Warning: solver timed out` is just detected by regex matching, so I could recover that information with a slightly changed regex.  But it seems to me that the original output is correct (a timeout is not unexpected, it's a totally valid response, and it gets handled gracefully).  The new output makes a timeout seem like an error.

clairexen:
> in the context of the qbfsat pass, everything still works

:+1:

> is only ever called by smt_check_sat()

the tool is split into a library part and a tool part so that other programs can also use the smtio infrastructure.

boqwxp:
Aha, I see.  Just for my edification, are there other programs using it that you know of?  I would naïvely expect other programs should probably prefer using Yosys or SymbiYosys directly (at least for hardware stuff) or something like [PySMT](https://github.com/pysmt/pysmt).

clairexen:
> are there other programs using it that you know of

we built some scripts for custom proofs a while back. I'm not sure if any of it is in the public tho and likely it doesn't work anymore with current smtio.py.

> or something like PySMT

I'm not sure how one would use PySMT with the Yosys SMT output.

boqwxp:
> we built some scripts for custom proofs a while back.

Gotcha.

> I'm not sure how one would use PySMT with the Yosys SMT output.

I meant that if doing hardware stuff, I would expect that one would use (Symbi)Yosys and not deal directly with smtio.py at all.  It feels like this file is kind of in a gray area between "stable interface for external use, do not break" and "internal use only, change at will".  As you say, out-of-tree code that interfaces directly with SmtIo will probably break.

On the other hand, if not doing hardware stuff, it probably makes more sense to use something like PySMT than to involve (Symbi)Yosys or smtio.py at all.

PR #2287, opened 2020-07-21
satgen: Add support for dffe, sdff, sdffe, sdffce cells.

mwkmwkmwk:
Depends on #2277, #2286.


Fakerr:
Curious here. Since this PR depends on another, would you be interested by a [Github app](https://www.dpulls.com/) that prevents a PR from being merged until its dependency is itself merged ?

PR #2297, opened 2020-07-24
opt_expr: Remove -clkinv option, make it the default.

mwkmwkmwk:
Adds -noclkinv option just in case the old behavior was actually useful
to someone.


mwkmwkmwk:
*sigh* of course this uncovered another latent problem.  To be fixed in an upcoming async2sync PR.

mwkmwkmwk:
Depends on #2184, #2298.

PR #2299, opened 2020-07-25
Avoid generating wires for function args which are constant

### LINKED ISSUE #2296: Constant arg loop elaboration in function with some non-constant args

```Verilog
module top;
	function automatic [31:0] operation;
		input reg [3:0] rounds;
		input integer x;
		integer i;
		begin
			for (i = 0; i < rounds[3:0]; i = i + 1)
				x = x * 2;
			operation = x;
		end
	endfunction

	wire a;
	assign a = 2;

	parameter A = 3;

	wire [31:0] x;
	assign x = operation(A, a);

// `define VERIFY
`ifdef VERIFY
    assert property (a == 2);
    assert property (A == 3);
    assert property (x == 16);
`endif
endmodule
```

## Expected behavior

The loop in `operation` should be successfully elaborated given that is invoked with `rounds` as a constant.

## Actual behavior

```
test.v:7: ERROR: 2nd expression of generate for-loop is not constant!
```

zachjs:
Resolves #2296. Any feedback is greatly appreciated!

### DIFF
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -483,6 +483,27 @@ static AstNode *make_packed_struct(AstNode *template_node, std::string &name)
 	return wnode;
 }
 
+// check if a node or its children contains an assignment to the given variable
+static bool node_contains_assignment_to(const AstNode* node, const AstNode* var)
+{
+	if (node->type == AST_ASSIGN_EQ || node->type == AST_ASSIGN_LE) {
+		// current node is iteslf an assignment
+		log_assert(node->children.size() >= 2);
+		const AstNode* lhs = node->children[0];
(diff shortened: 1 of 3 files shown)


mmicko:
Hi @zachjs we noticed one thing that is not working any more (worked before this change), please try next case 

```
module test (
  input [7:0] a,
  output [7:0] b
);
  function [7:0] f;
    input [7:0] v;
    input off;
    integer off;
    begin
      f = v + off;
    end
  endfunction
  assign b = f(a, 42);
endmodule
```
Just use `yosys test.v ` and it will produce error. Verific and other tools (iverilog and verilator) works just fine.
For reference error text is :
`test.v:13: ERROR: Incompatible re-declaration of wire \off`

PR #2291, opened 2020-07-22
Add opt_dff pass.

mwkmwkmwk:
Depends on #2277, #2286, #2287, #2288.

PR #2309, opened 2020-07-28
synth_ice40: Use opt_dff.

mwkmwkmwk:
The main part is converting ice40_dsp to recognize the new FF types
created in opt_dff instead of trying to recognize the mux patterns on
its own.
    
The fsm call has been moved upwards because the passes cannot deal with
$dffe/$sdff*, and other optimizations don't help it much anyway.

Depends on #2308, #2291 .


mwkmwkmwk:
Note: DO NOT MERGE YET, it breaks test_pmgen pattern generation.

PR #2312, opened 2020-07-29
Add support for user types in IOs

kgugala:
This PR adds support for using user defined types in module's IO list

### DIFF
--- a/frontends/verilog/verilog_parser.y
+++ b/frontends/verilog/verilog_parser.y
@@ -282,7 +282,7 @@ static void rewriteAsMemoryNode(AstNode *node, AstNode *rangeNode)
 %token TOK_OR_ASSIGN TOK_XOR_ASSIGN TOK_AND_ASSIGN TOK_SUB_ASSIGN
 
 %type <ast> range range_or_multirange  non_opt_range non_opt_multirange range_or_signed_int
-%type <ast> wire_type expr basic_expr concat_list rvalue lvalue lvalue_concat_list
+%type <ast> wire_type expr basic_expr concat_list rvalue lvalue lvalue_concat_list non_io_wire_type io_wire_type
 %type <string> opt_label opt_sva_label tok_prim_wrapper hierarchical_id hierarchical_type_id integral_number
 %type <string> type_name
 %type <ast> opt_enum_init enum_type struct_type non_wire_data_type
@@ -619,26 +619,19 @@ non_opt_delay:
 delay:
 	non_opt_delay | %empty;
 
-wire_type:
-	{
-		astbuf3 = new AstNode(AST_WIRE);
-		current_wire_rand = false;
-		current_wire_const = false;
-	} wire_type_token_list {
-		$$ = astbuf3;
-		SET_RULE_LOC(@$, @2, @$);
-	};
+io_wire_type:
+	{ astbuf3 = new AstNode(AST_WIRE); current_wire_rand = false; current_wire_const = false; }
+	wire_type_token_io wire_type_const_rand opt_wire_type_token wire_type_signedness
+	{ $$ = astbuf3; SET_RULE_LOC(@$, @2, @$); };
 
-wire_type_token_list:
-	wire_type_token |
-	wire_type_token_list wire_type_token |
-	wire_type_token_io |
-	hierarchical_type_id {
-		astbuf3->is_custom_type = true;
-		astbuf3->children.push_back(new AstNode(AST_WIRETYPE));
-		astbuf3->children.back()->str = *$1;
-		delete $1;
-	};
+non_io_wire_type:
+	{ astbuf3 = new AstNode(AST_WIRE); current_wire_rand = false; current_wire_const = false; }
+	wire_type_const_rand wire_type_token wire_type_signedness
+	{ $$ = astbuf3; SET_RULE_LOC(@$, @2, @$); };
+
+wire_type:
+	io_wire_type  |
+	non_io_wire_type;
 
 wire_type_token_io:
 	TOK_INPUT {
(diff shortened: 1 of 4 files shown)


dsvf:
Tested this PR with my workflow. Works great for my needs (typedefs and typedef structs) and found no bugs so far. I'm not an experienced developer though, so take this with a grain of salt.

whitequark:

frontends/verilog/verilog_parser.y
```diff
@@ -610,26 +610,26 @@ non_opt_delay:
...
+
+not_io_wire_type:
+	wire_type_const_rand wire_type_token wire_type_signed;
+
+no_implicite_wire_type:
+	wire_type_const_rand wire_type_token wire_type_signed;
+
+wire_type_without_implicit:
+	{ astbuf3 = new AstNode(AST_WIRE); current_wire_rand = false; current_wire_const = false; }
+	no_implicite_wire_type { $$ = astbuf3; SET_RULE_LOC(@$, @2, @$); };
```
zachjs, line 624: I think I understand how these grammar changes resolve the `TOK_USERTYPE` conflict in the `typedef` rule, but I'm wondering if we can simplify it a little. Could we have both `io_wire_type` and `non_io_wire_type`, both of which set and use `astbuf3`, have `wire_type` be either of those, and then have the `typedef` rule use `non_io_wire_type` directly? This would eliminate `wire_type_without_implicit`, `no_implicite_wire_type`, and `wire_type_wrapper`.

frontends/verilog/verilog_parser.y
```diff
@@ -643,7 +643,25 @@ wire_type_token_io:
 		astbuf3->is_output = true;
 	};
 
+wire_type_signed:
```
zachjs, line 646: Perhaps `wire_type_signedness` to match up with other code?

frontends/verilog/verilog_parser.y
```diff
@@ -643,7 +643,25 @@ wire_type_token_io:
...
+	TOK_SIGNED   { astbuf3->is_signed = true;  } |
+	TOK_UNSIGNED { astbuf3->is_signed = false; } |
+	%empty;
+
+wire_type_const_rand:
+	TOK_CONST { current_wire_const = true; } |
+	TOK_RAND { current_wire_rand = true; } |
```
zachjs, line 653: Can we use the same extra alignment spacing as in the above rule?

tests/svtypes/typedef_struct_port.sv
```diff
@@ -0,0 +1,26 @@
...
+
+typedef struct packed {
+	byte a;
+	byte b;
+} p_t;
+
+endpackage
+
+module foo1(
+	input p::p_t p
```
zachjs, line 11: Given the non-trivial changes to the parser, I'd appreciate more substantial test coverage for these changes. We should have coverage for typedefs aliasing other typenames (both in another package and local), declaring such ports in the module body rather than the module header, and the effects of including the other qualifiers in these declarations.

tests/svtypes/typedef_struct_port.ys
```diff
@@ -0,0 +1 @@
+read_verilog -sv typedef_struct_port.sv
```
zachjs, line 1: This test might pass even without the changes in `simplify.cc`. Can we extend it?

zachjs (changes requested):

frontends/verilog/verilog_parser.y
kamilrakoczy: Fixed.

frontends/verilog/verilog_parser.y
kamilrakoczy: Fixed.

frontends/verilog/verilog_parser.y
kamilrakoczy: Fixed.

tests/svtypes/typedef_struct_port.sv
kamilrakoczy: Added.

tests/svtypes/typedef_struct_port.ys
kamilrakoczy: I extended it to check asserts. Is this enough?

frontends/verilog/verilog_parser.y
```diff
@@ -619,26 +619,19 @@ non_opt_delay:
...
-		delete $1;
-	};
+non_io_wire_type:
+	{ astbuf3 = new AstNode(AST_WIRE); current_wire_rand = false; current_wire_const = false; }
+	wire_type_const_rand wire_type_token wire_type_signedness
+	{ $$ = astbuf3; SET_RULE_LOC(@$, @2, @$); };
+
+wire_type:
+	io_wire_type  |
+	non_io_wire_type; // This types can't be merged, as it would give possibility to produce only empty type.
```
zachjs, line 634: I'm not convinced that this comment is still accurate. Perhaps it can be updated or dropped?

tests/svtypes/typedef_struct_port.ys
zachjs: I think this test coverage is substantially improved. Thank you!

For things in `test_parser` which the comments say should be signed or unsigned, can we verify this in programmatically in the test? I could imagine checking how the ports sign extend or asserting the equivalence of certain declarations would work. Perhaps it would be possible to inspect the presence of the `signed` attribute on the generated RTLIL wires?

frontends/verilog/verilog_parser.y
kamilrakoczy: It is true, I dropped this comment, as we do not want to merge this types.

tests/svtypes/typedef_struct_port.ys
kamilrakoczy: I updated tests to assert with expected signed/unsigned value.

zachjs (commented):
Can we squash the commit history and get this rebased to run it against the latest tests? Otherwise, LGTM.

kamilrakoczy:
I squashed the commit history and rebased on top of current master.

zachjs (approved):

mmicko:
@zachjs can you please check ?

This Verilog file now fail loading https://github.com/YosysHQ/yosys-tests/blob/master/frontends/read_verilog/top_dpi.v

"wire logic" is valid construct but it now fails parsing.

zachjs:
@mmicko I've opened a PR with a fix for that issue.

While I've begun running `make ystests` locally for my larger changes, it seems this feedback is missing in the public CI pipeline. Would we consider adding this to avoid missing such issues in the future?

mmicko:
Thanks @zachjs, it takes a while to run all those tests so for now it is not possible to run them for each commit.

PR #2313, opened 2020-07-29
opt_expr: Fix handling of $_XNOR_ cells with A = B.

### LINKED ISSUE #2311: Error while executing OPT_EXPR pass (perform const folding).
While trying to run synthesis for the RISC V based Aardonyx SoC using Yosys, an error is encountered.

Steps to reproduce the issue:
1) Run the synthesis for the verilog files in 
[soc.tar.gz](https://github.com/YosysHQ/yosys/files/4992943/soc.tar.gz) using the OpenLane flow porting Skywater PDK.

Expected Behavior:
The synthesis should be completed successfully as there are no bugs in the verilog source codes.

Actual Behavior:
The error message reads as below:
67.1. Executing OPT_EXPR pass (perform const folding).
what(): Cell::getParam()
[ERROR]: during executing: "yosys -c /openLANE_flow/scripts//synth.tcl -l /openLANE_flow/designs/aardonyx/runs/28-07_19-08//logs/synthesis/yosys.log |& tee >&@stdout"
[ERROR]: Last 10 lines:
child killed: SIGABRT

[yosys.log](https://github.com/YosysHQ/yosys/files/4992859/yosys.log)
[error.log](https://github.com/YosysHQ/yosys/files/4992862/error.log)

mwkmwkmwk:
Fixes #2311.

### DIFF
--- a/passes/opt/opt_expr.cc
+++ b/passes/opt/opt_expr.cc
@@ -604,7 +604,7 @@ void replace_const_cells(RTLIL::Design *design, RTLIL::Module *module, bool cons
 				if (cell->type.in(ID($xnor), ID($_XNOR_))) {
 					cover("opt.opt_expr.const_xnor");
 					// For consistency since simplemap does $xnor -> $_XOR_ + $_NOT_
-					int width = cell->getParam(ID::Y_WIDTH).as_int();
+					int width = GetSize(cell->getPort(ID::Y));
 					replace_cell(assign_map, module, cell, "const_xnor", ID::Y, SigSpec(RTLIL::State::S1, width));
 					goto next_cell;
 				}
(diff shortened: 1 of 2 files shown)


passes/opt/opt_expr.cc
```diff
@@ -604,7 +604,7 @@ void replace_const_cells(RTLIL::Design *design, RTLIL::Module *module, bool cons
 				if (cell->type.in(ID($xnor), ID($_XNOR_))) {
 					cover("opt.opt_expr.const_xnor");
 					// For consistency since simplemap does $xnor -> $_XOR_ + $_NOT_
-					int width = cell->getParam(ID::Y_WIDTH).as_int();
+					int width = GetSize(cell->getPort(ID::Y));
```
whitequark, line 607: Wait, why does this make a difference? Shouldn't a module where these two don't match fail `check()`?

mwkmwkmwk: They always match for `$xnor`, but `Y_WIDTH` is not present at all for `$_XNOR_`, causing the crash reported in #2311.

PR #2323, opened 2020-08-03
Link with libatomic always

mohamed:
libatomic must be linked explicitly under Android. Hence, we always link to it.
This change has been tested with:
 - Ubuntu 18.04 LTS + GCC 8.4.0
 - Android 9.0 + clang 10.0.1

Signed-off-by: Mohamed A. Bamakhrama <mohamed@alumni.tum.de>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,7 +83,7 @@ YOSYS_SRC := $(dir $(firstword $(MAKEFILE_LIST)))
 VPATH := $(YOSYS_SRC)
 
 CXXFLAGS := $(CXXFLAGS) -Wall -Wextra -ggdb -I. -I"$(YOSYS_SRC)" -MD -D_YOSYS_ -fPIC -I$(PREFIX)/include
-LDLIBS := $(LDLIBS) -lstdc++ -lm
+LDLIBS := $(LDLIBS) -lstdc++ -lm -latomic
 PLUGIN_LDFLAGS :=
 
 PKG_CONFIG ?= pkg-config


nakengelhardt:
@whitequark will this affect yowasp or any of your targets?

@mmicko is this fine on windows and mac (works for me on 10.14)?

whitequark:
> will this affect yowasp or any of your targets?

Yep, it will break WASI and (I think) Emscripten.

mohamed:
@nakengelhardt @whitequark I will update the change to link to libatomic if `uname -o` equals `Android`. Is this fine with you?

whitequark:
> I will update the change to link to libatomic if `uname -o` equals `Android`. Is this fine with you?

No, of course not, this breaks cross-compilation: you should not determine target features by checking what the host system is.

mohamed:
> No, of course not, this breaks cross-compilation: you should not determine target features by checking what the host system is.

OK. What about adding a new `LINK_ATOMIC` variable and using it to control the linking? The user can set it explicitly to `1` when building on Android.

whitequark:
Why not just set `LDLIBS=-latomic` directly in that case? That requires no changes in Yosys. (An alternative would be to add a `config-android` `Makefile.conf` option that does the same thing, for better discoverability.)

mmicko:
I can only confirm that it is also needed by other ARM builds, guess you were compiling for Android but for ARM cpu,

mohamed:
> I can only confirm that it is also needed by other ARM builds, guess you were compiling for Android but for ARM cpu,

Yes. I am developing this on an Arm-based Android device

whitequark:
Ah right. This is actually a [compiler bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=81358).

mohamed:
> Why not just set `LDLIBS=-latomic` directly in that case? 

I tried this at the very beginning. `make LDLIBS=-latomic` doesn't work because all assignments in the makefile will be ignored (see: https://www.gnu.org/software/make/manual/html_node/Override-Directive.html#Override-Directive).
I also tried `make LDLIBS+=-latomic` and it also fails.
If I change line 87 into `LDLIBS := $(LDLIBS_USER) -lstdc++ -lm` and then override `LDLIBS_USER` on the command line, then the linker succeeds.

whitequark:
Try `LDLIBS=-latomic make`

mohamed:
> Try `LDLIBS=-latomic make`

This works indeed. There is still my PR in https://github.com/YosysHQ/abc/pull/7 to enable yosys-abc on Android

PR #2331, opened 2020-08-05
Add support for unpacked arrays assignment

kgugala:
This PR adds support for copying unpacked arrays via assignment statement

### DIFF
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -899,6 +899,73 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
 				if (!is_rand_reg)
 					log_warning("reg '%s' is assigned in a continuous assignment at %s:%d.%d-%d.%d.\n", children[0]->str.c_str(), filename.c_str(), location.first_line, location.first_column, location.last_line, location.last_column);
 			}
+
+			// Copy unpacked array
+			if (children[0]->id2ast->type == AST_MEMORY &&
+					children[1]->type == AST_IDENTIFIER && children[1]->id2ast && children[1]->id2ast->type == AST_MEMORY &&
+					children[0]->children.size() == 0 && children[1]->children.size() == 0) {
+				const auto& lhs_unpacked_range = children[0]->id2ast->children[1];
+				log_assert(lhs_unpacked_range->basic_prep);
+				const auto& rhs_unpacked_range = children[1]->id2ast->children[1];
+				log_assert(rhs_unpacked_range->basic_prep);
+
+				const auto lhs_width = lhs_unpacked_range->range_left - lhs_unpacked_range->range_right + 1;
+				const auto rhs_width = rhs_unpacked_range->range_left - rhs_unpacked_range->range_right + 1;
+
+				if (lhs_width != rhs_width)
+					log_error("Array '%s' and '%s' differs in size: %s:%d.%d-%d.%d.\n",
+						children[0]->str.c_str(), children[1]->str.c_str(), filename.c_str(),
+						location.first_line, location.first_column, location.last_line, location.last_column);
+
+				// Create template from assignment
+				const auto* tmpl = this->clone();
+
+				// Replace assignment with block
+				auto* block = this;
+				switch (type)
+				{
+				case AST_ASSIGN_EQ:
+				case AST_ASSIGN_LE:
+					block->type = AST_BLOCK;
+					break;
+
+				case AST_ASSIGN:
+					block->type = AST_GENBLOCK;
+					break;
+
+				default: // should not happen
+					log_assert(false);
+					break;
+				}
+				block->children.clear();
+
+				// Generate series of assignments
+				for (int i = 0 ; i < lhs_width ; ++i) {
+					auto* assign = tmpl->clone();
+					assign->children[0]->children.push_back(new AstNode(AST_RANGE));
+					assign->children[0]->children[0]->children.push_back(AstNode::mkconst_int(i + lhs_unpacked_range->range_right, true));
+
+					assign->children[1]->children.push_back(new AstNode(AST_RANGE));
+					assign->children[1]->children[0]->children.push_back(AstNode::mkconst_int(i + rhs_unpacked_range->range_right, true));
+
+					// simplify
+					assign->children[0]->basic_prep = false;
+					while (!assign->children[0]->basic_prep && assign->children[0]->simplify(false, false, true, stage, -1, false, in_param) == true)
+						did_something = true;
+					assign->children[1]->basic_prep = false;
+					while (!assign->children[1]->basic_prep && assign->children[1]->simplify(false, false, true, stage, -1, false, in_param) == true)
+						did_something = true;
+
+					assign->children[0]->was_checked = true;
+					assign->children[1]->was_checked = true;
+
+					block->children.push_back(assign);
+				}
+
+				log_assert((size_t)block->children.size() == (size_t)lhs_width);
+				did_something = true;
+			}
+
 			children[0]->was_checked = true;
 		}
 		break;
--- a/tests/svtypes/array_copy.sv
+++ b/tests/svtypes/array_copy.sv
@@ -1,0 +1,22 @@
+// test multirange array assignement (copy)
+
+module top;
+
+logic a [1:0][1:0];
+logic b [1:0][1:0];
+
+assign a[0][0] = 1'b0;
+assign a[0][1] = 1'b1;
+assign a[1][0] = 1'b1;
+assign a[1][1] = 1'b1;
+
+assign b = a;
+
+always_comb begin
+	assert(b[0][0] == 1'b0);
+	assert(b[0][1] == 1'b1);
+	assert(b[1][0] == 1'b1);
+	assert(b[1][1] == 1'b1);
+end
+
+endmodule


frontends/ast/simplify.cc
```diff
@@ -899,6 +899,73 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
...
+					while (!assign->children[1]->basic_prep && assign->children[1]->simplify(false, false, true, stage, -1, false, in_param) == true)
+						did_something = true;
+
+					assign->children[0]->was_checked = true;
+					assign->children[1]->was_checked = true;
+
+					block->children.push_back(assign);
+				}
+
+				log_assert((size_t)block->children.size() == (size_t)lhs_width);
```
daveshah1, line 965: Yosys convention is to use GetSize in these sort of cases, which casts `.size()` to int

frontends/ast/simplify.cc
```diff
@@ -899,6 +899,73 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
...
+					auto* assign = tmpl->clone();
+					assign->children[0]->children.push_back(new AstNode(AST_RANGE));
+					assign->children[0]->children[0]->children.push_back(AstNode::mkconst_int(i + lhs_unpacked_range->range_right, true));
+
+					assign->children[1]->children.push_back(new AstNode(AST_RANGE));
+					assign->children[1]->children[0]->children.push_back(AstNode::mkconst_int(i + rhs_unpacked_range->range_right, true));
+
+					// simplify
+					assign->children[0]->basic_prep = false;
+					while (!assign->children[0]->basic_prep && assign->children[0]->simplify(false, false, true, stage, -1, false, in_param) == true)
```
daveshah1, line 953: Is `== true` really needed here?

kgugala: It's not really needed, just followed the convention used in this file e.g:
```
while (!children[0]->basic_prep && children[0]->simplify(false, false, true, stage, -1, false, in_param) == true)
```
in line 873.

We can change this if you want, but then it'd good to update all the other `== true` lines so the code style is consistent in the file.

frontends/ast/simplify.cc
kgugala: fixed

whitequark:

frontends/ast/simplify.cc
```diff
@@ -899,6 +899,73 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
 				if (!is_rand_reg)
 					log_warning("reg '%s' is assigned in a continuous assignment at %s:%d.%d-%d.%d.\n", children[0]->str.c_str(), filename.c_str(), location.first_line, location.first_column, location.last_line, location.last_column);
 			}
+
+			// Copy unpacked array
```
zachjs, line 903: Could we clarify that this is for copying an _entire_ unpacked array?

frontends/ast/simplify.cc
```diff
@@ -899,6 +899,73 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
...
+			if (children[0]->id2ast->type == AST_MEMORY &&
+					children[1]->type == AST_IDENTIFIER && children[1]->id2ast && children[1]->id2ast->type == AST_MEMORY &&
+					children[0]->children.size() == 0 && children[1]->children.size() == 0) {
```
zachjs, line 906: Could we use `.empty()` over `.size() == 0`?

frontends/ast/simplify.cc
```diff
@@ -899,6 +899,73 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
...
+				const auto& lhs_unpacked_range = children[0]->id2ast->children[1];
+				log_assert(lhs_unpacked_range->basic_prep);
+				const auto& rhs_unpacked_range = children[1]->id2ast->children[1];
+				log_assert(rhs_unpacked_range->basic_prep);
+
+				const auto lhs_width = lhs_unpacked_range->range_left - lhs_unpacked_range->range_right + 1;
+				const auto rhs_width = rhs_unpacked_range->range_left - rhs_unpacked_range->range_right + 1;
```
zachjs, line 913: Surrounding code in this file conventionally uses just `int` for `range_left` and `range_right` rather than `const auto`.

frontends/ast/simplify.cc
```diff
@@ -899,6 +899,73 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
...
+				if (lhs_width != rhs_width)
+					log_error("Array '%s' and '%s' differs in size: %s:%d.%d-%d.%d.\n",
+						children[0]->str.c_str(), children[1]->str.c_str(), filename.c_str(),
+						location.first_line, location.first_column, location.last_line, location.last_column);
+
+				// Create template from assignment
+				const auto* tmpl = this->clone();
+
+				// Replace assignment with block
+				auto* block = this;
```
zachjs, line 924: Given `block` is sparingly used, it might be preferable to drop this alias.

frontends/ast/simplify.cc
```diff
@@ -899,6 +899,73 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
...
+					break;
+
+				case AST_ASSIGN:
+					block->type = AST_GENBLOCK;
+					break;
+
+				default: // should not happen
+					log_assert(false);
+					break;
+				}
```
zachjs, line 939: `assert(false);` can be unhelpful when debugging. Would you consider something like the following?
```suggestion
				if (type == AST_ASSIGN) {
					type = AST_GENBLOCK;
				} else {
					log_assert(type == AST_ASSIGN_EQ || type == AST_ASSIGN_LE);
					type = AST_BLOCK;
				}
```

frontends/ast/simplify.cc
```diff
@@ -899,6 +899,73 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
...
+				block->children.clear();
+
+				// Generate series of assignments
+				for (int i = 0 ; i < lhs_width ; ++i) {
```
zachjs, line 943: Can we drop the extraneous spaces here?

frontends/ast/simplify.cc
zachjs: There are only a handful of instances of `== true` in this file. My preference would be to avoid replicating that style in new code given many find it problematic. We may consider removing the other usages, but some may prefer avoiding the noise in the commit history.

zachjs (changes requested):
This change does not seem to work correctly if the source and destination dimensions are not in the same direction. For example, the following slightly modified test case does not pass:
```SystemVerilog
module top;
logic a [1:0][1:0];
logic b [1:0][0:1];

assign a[0][0] = 1'b0;
assign a[0][1] = 1'b1;
assign a[1][0] = 1'b1;
assign a[1][1] = 1'b1;

assign b = a;

always_comb begin
  assert(b[0][1] == 1'b0);
  assert(b[0][0] == 1'b1);
  assert(b[1][1] == 1'b1);
  assert(b[1][0] == 1'b1);
end
endmodule
```

frontends/ast/simplify.cc
ldalek-antmicro: Sure. I'll fix that.

frontends/ast/simplify.cc
ldalek-antmicro: Okay.

frontends/ast/simplify.cc
ldalek-antmicro: Okay

frontends/ast/simplify.cc
ldalek-antmicro: Fixed.

frontends/ast/simplify.cc
ldalek-antmicro: Done

frontends/ast/simplify.cc
ldalek-antmicro: How about:
```			
log_assert(type == AST_ASSIGN_EQ || type == AST_ASSIGN_LE || type == AST_ASSIGN);
if (type == AST_ASSIGN) {
  this->type = AST_GENBLOCK;
} else {
  this->type = AST_BLOCK;
}
```

ldalek-antmicro:
Hey @zachjs,
Thanks for comments. I applied your review comments and fixed problem with reversed dimensions.

zachjs:
I think this is missing some test coverage and error checking.

* There is no coverage for dimensions besides `[1:0]` and `[0:1]`
* There is no coverage for memories with more than 2 unpacked dimensions
* The following is accepted, despite other tools rejecting the mismatched dimensions:
    ```SystemVerilog
    logic a [1:0][1:0];
    logic b [0:0][0:3];
    assign a = b;
    ```
* The `differs in size` error is given for two distinct failures


We should also consider supporting some of the following variations, or ensuring the current strategy can be adapted for them if support is desired down the road.

```SystemVerilog
logic toggle;
logic a1 [2:0][3:0];
logic a2 [2:0][3:0];
logic b [2:0][3:0];
assign b = toggle ? a1 : a2;
```

```SystemVerilog
logic toggle;
logic a [1:0][2:0][3:0];
logic b [2:0][3:0];
assign b = a[toggle];
```

```SystemVerilog
logic toggle;
logic a [1:0][2:0][3:0];
logic b [2:0][3:0];
assign b = toggle ? a[0] : a[1];
```

Handling the elaboration stepwise like that may also eliminate the need for the internal recursive lambda.

widlarizer:
Reworked in #5630

Issue #1592, opened 2019-12-22
Clock in interface port mis-synthesized away (but accepted in modport)


tux3:
## Steps to reproduce the issue

Attachment [interface_clk_issue.zip](https://github.com/YosysHQ/yosys/files/3992740/interface_clk_issue.zip) contains `good.sv`, `bad.sv`, and a Makefile.  
Run `make good` to observe the expected output, and `make bad` to reproduce the mis-synthesis.  
This is reproduced on the current master (aa1adb0f), using `synth_ice40`.

The file `bad.sv` connects a clock to the interface instance:

```systemverilog
interface dummybus(input clk, input rst);
    modport host(input clk, input rst);
    modport device(input clk, input rst);
endinterface

module top(input clk, input rst);
    dummybus bus(.clk, .rst); // bus driven here
    device dev(.bus);
    host host(.clk, .rst, .bus);
endmodule

module host(input clk, input rst, dummybus.host bus);
    // This page intentionally left blank
endmodule

module device(dummybus.device bus);
    (* keep *)
    logic [63:0] keep;
    always @(posedge bus.clk)
        if (bus.rst) begin
            keep <= 'b0;
        end else begin
            keep <= keep + 'b1;
        end
endmodule
```

And `good.sv` connects the clock to a modport, which works as expected:

```systemverilog
interface dummybus();
    logic clk;
    logic rst;
    modport host(output clk, output rst);
    modport device(input clk, input rst);
endinterface

module top(input clk, input rst);
    dummybus bus();
    device dev(.bus);
    host host(.clk, .rst, .bus);
endmodule

module host(input clk, input rst, dummybus.host bus);
    // This works, the clock is recognized
    assign bus.clk = clk;
    assign bus.rst = rst;
endmodule

module device (dummybus.device bus);
    (* keep *)
    logic [63:0] keep;
    always @(posedge bus.clk)
        if (bus.rst) begin
            keep <= 'b0;
        end else begin
            keep <= keep + 'b1;
        end
endmodule
```

The result of synthetizing `bad.sv` is zero cells and a warning that the bus clock has no drivers.  
Simulation in ALDEC shows the same expected behavior for both files (`keep` is incremented).

## Expected behavior

Yosys should see that the clock in `bad.sv` is driven from the top.
Synthesis results for `bad.sv` should be functionally equivalent to `good.sv`.

## Actual behavior

Yosys misses the clock in `bad.sv` and removes live code.

[...]

PR #2337, opened 2020-08-09
opt_clean: Fix module keep rules.

mwkmwkmwk:
- wires with keep attribute now force a module to be kept
- presence of $memwr and $meminit cells no longer forces a module to be
  kept


mwkmwkmwk:
This fixes a problem in #1592 (but is unrelated to the main issue at hand).

PR #2340, opened 2020-08-12
cxxrtl.h: Fix incorrect CarryOut in alu when Bits % 32 != 0 && Invert == False

andy-knowles:
This PR fixes the alu function in cxxrtl.h to return correct CarryOut when Bits % 32 != 0 && Invert == False, by updating carry after the last chunk is masked.
(CarryOut is currently only used by ucmp(), which always sets Invert = True, so this bug has not been triggered)

### DIFF
--- a/backends/cxxrtl/cxxrtl.h
+++ b/backends/cxxrtl/cxxrtl.h
@@ -450,12 +450,18 @@ struct value : public expr_base<value<Bits>> {
 	std::pair<value<Bits>, bool /*CarryOut*/> alu(const value<Bits> &other) const {
 		value<Bits> result;
 		bool carry = CarryIn;
-		for (size_t n = 0; n < result.chunks; n++) {
+		// Handle full chunks first
+		for (size_t n = 0; n < result.chunks - 1; n++) {
 			result.data[n] = data[n] + (Invert ? ~other.data[n] : other.data[n]) + carry;
 			carry = (result.data[n] <  data[n]) ||
 			        (result.data[n] == data[n] && carry);
 		}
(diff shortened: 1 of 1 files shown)


backends/cxxrtl/cxxrtl.h
```diff
@@ -450,12 +450,18 @@ struct value : public expr_base<value<Bits>> {
...
+		for (size_t n = 0; n < result.chunks - 1; n++) {
 			result.data[n] = data[n] + (Invert ? ~other.data[n] : other.data[n]) + carry;
 			carry = (result.data[n] <  data[n]) ||
 			        (result.data[n] == data[n] && carry);
 		}
-		result.data[result.chunks - 1] &= result.msb_mask;
+		// Handle last chunk (mask before updating carry)
+		constexpr size_t last = result.chunks - 1;
+		result.data[last] = data[last] + (Invert ? ~other.data[last] : other.data[last]) + carry;
+		result.data[last] &= result.msb_mask;
```
whitequark, line 462: I'm pretty sure you can just use a conditional instead of duplicating all this code, this loop gets unrolled anyways.

andy-knowles: Ok, i wasn't sure if I could rely on it always being unrolled and optimised away.

whitequark: There isn't any pragmas to force it to unroll the loop, but in general, the performance goals of CXXRTL can only be achieved with a C++ compiler that aggressively unrolls loops with a small amount of conditionals in them. (There's already ternaries in all the loops!) So it's fine to do this here, too.

whitequark:
Thanks!

PR #2349, opened 2020-08-20
Ensure smt2 comments are associated with accessors

nmoroze:
Hi,

I noticed an issue recently where accessor comment annotations may end up no longer associated with the correct accessor in Yosys's SMT2 output. For example:

The following module:
```verilog
module top (
    input in,
    output out
);
wire in_n;
assign in_n = in + 1'b1;
assign out = ~in_n;
endmodule
```

When compiled with the following command:
`yosys -p 'read_verilog example.v' -p 'prep -flatten -top top -nordff' -p 'write_smt2 out.smt2'`

Gives output like so:
```
; SMT-LIBv2 description generated by Yosys 0.9+3491 (git sha1 87b9ee33, clang 11.0.3 -fPIC -Os)
; yosys-smt2-module top
(declare-sort |top_s| 0)
(declare-fun |top_is| (|top_s|) Bool)
; yosys-smt2-input in 1
(declare-fun |top#0| (|top_s|) Bool) ; \in
(define-fun |top_n in| ((state |top_s|)) Bool (|top#0| state))
; yosys-smt2-output out 1
(define-fun |top#1| ((state |top_s|)) (_ BitVec 1) (bvadd (ite (|top#0| state) #b1 #b0) #b1)) ; \in_n
(define-fun |top#2| ((state |top_s|)) (_ BitVec 1) (bvnot (|top#1| state))) ; \out
(define-fun |top_n out| ((state |top_s|)) Bool (= ((_ extract 0 0) (|top#2| state)) #b1))
(define-fun |top_a| ((state |top_s|)) Bool true)
(define-fun |top_u| ((state |top_s|)) Bool true)
(define-fun |top_i| ((state |top_s|)) Bool true)
(define-fun |top_h| ((state |top_s|)) Bool true)
(define-fun |top_t| ((state |top_s|) (next_state |top_s|)) Bool true) ; end of module top
; yosys-smt2-topmod top
; end of yosys output
```

This seems wrong to me - the comment annotation for `out` precedes a completely different signal, and even the comment annotation for `in` is above the function with a numbered name, not the accessor with the nice name. 

I discovered the problem is that the functions `get_bool` and `get_bv` that return a reference to a signal (which are used to generate the body of these accessor functions) may also export cells themselves. This pushes additional strings onto the `decls` vector in between the comment and the actual accessor. 

This patch solves the problem by not pushing the comment write away, and precomputing the `sig_bool` or `sig_bv` return value before pushing the accessor function. This way, the comment and the accessor function are guaranteed to be pushed together, with the right relative ordering. With this patch, the output from above turns into:
```
; SMT-LIBv2 description generated by Yosys 0.9+3491 (git sha1 87b9ee33, clang 11.0.3 -fPIC -Os)
; yosys-smt2-module top
(declare-sort |top_s| 0)
(declare-fun |top_is| (|top_s|) Bool)
(declare-fun |top#0| (|top_s|) Bool) ; \in
; yosys-smt2-input in 1
(define-fun |top_n in| ((state |top_s|)) Bool (|top#0| state))
(define-fun |top#1| ((state |top_s|)) (_ BitVec 1) (bvadd (ite (|top#0| state) #b1 #b0) #b1)) ; \in_n
(define-fun |top#2| ((state |top_s|)) (_ BitVec 1) (bvnot (|top#1| state))) ; \out
; yosys-smt2-output out 1
(define-fun |top_n out| ((state |top_s|)) Bool (= ((_ extract 0 0) (|top#2| state)) #b1))
(define-fun |top_a| ((state |top_s|)) Bool true)
(define-fun |top_u| ((state |top_s|)) Bool true)
(define-fun |top_i| ((state |top_s|)) Bool true)
(define-fun |top_h| ((state |top_s|)) Bool true)
(define-fun |top_t| ((state |top_s|) (next_state |top_s|)) Bool true) ; end of module top
; yosys-smt2-topmod top
; end of yosys output
```

One note: I wasn't able to test this patch on any example design using write_smt2 with the `-nobv` flag. Every time I call the backend with this flag, I get an error message stating that basic cell types ($add, $not, $dff, etc.) are not defined. 

Thanks in advance for taking a look! If you have any feedback on a better way to tackle this issue, I'd be happy to adjust the PR accordingly.


backends/smt2/smt2.cc
```diff
@@ -824,38 +824,49 @@ struct Smt2Worker
...
+						decls.push_back(comment);
 					decls.push_back(stringf("(define-fun |%s_n %s| ((state |%s_s|)) (_ BitVec %d) %s)\n",
-							get_id(module), get_id(wire), get_id(module), GetSize(sig), get_bv(sig).c_str()));
+							get_id(module), get_id(wire), get_id(module), GetSize(sig), sig_bv.c_str()));
 					if (wire->port_input)
 						ex_input_eq.push_back(stringf("  (= (|%s_n %s| state) (|%s_n %s| other_state))",
 								get_id(module), get_id(wire), get_id(module), get_id(wire)));
 				} else {
-					for (int i = 0; i < GetSize(sig); i++)
+					std::string sig_bool[GetSize(sig)];
```
nakengelhardt, line 849: This apparently isn't standard C++ and isn't supported in older versions of clang (cf the CI failure), but you should be able to use a `std::vector<std::string>` to the same effect.

nmoroze:
Thanks for taking a look - I just fixed the commit to use a vector instead of a C-style array!

I also noticed that my original commit broke the case where there are more than one comment annotation on an accessor, but I fixed that up as well (by populating a vector of all relevant comments and inserting those, rather than a single string).

PR #2351, opened 2020-08-20
Add -nomux switch to proc

pbsds:
Running `proc -nomux` will with this change omit the `proc_mux` pass.

This PR was prompted by nmigen/nmigen#479


whitequark:
Thanks! I've confirmed this works with nMigen.

PR #2355, opened 2020-08-26
Add formal apps and template generators

mmicko:
Yosys part of recent improvements added to SEDA specific Verific library.


clairexen (approved):
LGTM

PR #2356, opened 2020-08-26
flatten, techmap: don't canonicalize tpl driven bits via sigmap

### LINKED ISSUE #2183: ERROR: Mismatch in directionality for cell port...

Possibly the same issue as https://github.com/YosysHQ/yosys/issues/2053

My code uses arrays where some lines are inputs and some are outputs.  With recent Yosys builds, some small part of this code is that used to work for a long time is no-longer compiling.  The code still functions in Vivado, fyi.

Boiled down, the following snippet illustrates this.

There is a 9-bit array, which is marked as `inout` because of the mix of signals.  The subordinate module, `worker` assigns the parts of the bundle.

``` verilog

module worker(
        inout  [8:0] data_in,
        inout  [8:0] data_out
    );
    assign data_out[7:0] = data_in[7:0];
    assign data_in[ 8 ] = data_out[ 8 ];
endmodule

module tests_tbx (
        inout [8:0] data
    );
    wire [8:0] data_shadow;
    assign data_shadow[7:0] = 8'H55;
    worker w( data_shadow, data );
endmodule

```
## Expected behavior

This should compile.

## Actual behavior

```
$ make
yosys -q -p 'synth_ice40 -top tests_tbx -json tests_tbx.json' ../projects/tests_tbx/tests_tbx.v
ERROR: Mismatch in directionality for cell port tests_tbx.w.data_in: \data_shadow <= \w.data_in
Makefile:30: recipe for target 'tests_tbx.json' failed
make: *** [tests_tbx.json] Error 1
```

whitequark:
For connection `assign a = b;`, `sigmap(a)` returns `b`. This is
exactly the opposite of the desired canonicalization for driven bits.
Consider the following code:

    module foo(inout a, b);
      assign a = b;
    endmodule
    module bar(output c);
      foo f(c, 1'b0);
    endmodule

Before this commit, the inout ports would be swapped after flattening
(and cause a crash while attempting to drive a constant value).

This issue was introduced in 9f772eb9.

Fixes #2183.

### DIFF
--- a/passes/techmap/flatten.cc
+++ b/passes/techmap/flatten.cc
@@ -152,15 +152,14 @@ struct FlattenWorker
 
 		// Attach port connections of the flattened cell
 
-		SigMap tpl_sigmap(tpl);
 		pool<SigBit> tpl_driven;
 		for (auto tpl_cell : tpl->cells())
 			for (auto &tpl_conn : tpl_cell->connections())
 				if (tpl_cell->output(tpl_conn.first))
-					for (auto bit : tpl_sigmap(tpl_conn.second))
+					for (auto bit : tpl_conn.second)
 						tpl_driven.insert(bit);
 		for (auto &tpl_conn : tpl->connections())
-			for (auto bit : tpl_sigmap(tpl_conn.first))
+			for (auto bit : tpl_conn.first)
 				tpl_driven.insert(bit);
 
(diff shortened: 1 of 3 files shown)


clairexen (approved):
Thanks for looking into this :+1:

PR #2358, opened 2020-08-26
Replace "ILANG" with "RTLIL" everywhere

whitequark:
The only difference between "RTLIL" and "ILANG" is that the latter is the text representation of the former, as opposed to the in-memory graph representation. This distinction serves no purpose but confuses people: it is not obvious that the ILANG backend writes RTLIL graphs.

Passes `write_ilang` and `read_ilang` are provided as aliases to `write_rtlil` and `read_rtlil` for compatibility.


clairexen (approved):
LGTM

PR #2359, opened 2020-08-26
intel_alm: Add multiply signedness to cells

Ravenslofty:
Quartus assumes unsigned multiplication by default, breaking signed multiplies, so add an input signedness parameter to the MISTRAL_MUL* cells to propagate to Quartus' <family>_mac cells.

Also added some multiplier tests to avoid #2342 again.

### DIFF
--- a/techlibs/intel_alm/common/dsp_sim.v
+++ b/techlibs/intel_alm/common/dsp_sim.v
@@ -1,18 +1,36 @@
 (* abc9_box *)
 module MISTRAL_MUL27X27(input [26:0] A, input [26:0] B, output [53:0] Y);
 
+parameter A_SIGNED = 1;
+parameter B_SIGNED = 1;
+
 // TODO: Cyclone 10 GX timings; the below are for Cyclone V
 specify
     (A *> Y) = 3732;
     (B *> Y) = 3928;
 endspecify
 
-assign Y = $signed(A) * $signed(B);
+wire [53:0] A_, B_;
+
+if (A_SIGNED)
+    assign A_ = $signed(A);
+else
+    assign A_ = $unsigned(A);
+
+if (B_SIGNED)
+    assign B_ = $signed(B);
+else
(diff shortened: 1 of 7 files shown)


techlibs/intel_alm/common/quartus_rename.v
```diff
@@ -174,20 +174,62 @@ endmodule
 
 module MISTRAL_MUL27X27(input [26:0] A, B, output [53:0] Y);
 
-`MAC #(.ax_width(27), .ay_scan_in_width(27), .result_a_width(54), .operation_mode("M27x27")) _TECHMAP_REPLACE_ (.ax(A), .ay(B), .resulta(Y));
+parameter A_SIGNED = 1;
+parameter B_SIGNED = 1;
+
+`MAC #(
+    .ax_width(27),
+    .signed_max(A_SIGNED ? "true" : "false"),
```
mwkmwkmwk, line 182: why is it "true" and "false" here, but 1 in the default parameter values in the blackbox?

Ravenslofty: Mistake on my part, sorry.

techlibs/intel_alm/common/dsp_sim.v
```diff
@@ -1,38 +1,56 @@
...
+parameter B_SIGNED = 1;
+
 // TODO: Cyclone 10 GX timings; the below are for Cyclone V
 specify
     (A *> Y) = 3732;
     (B *> Y) = 3928;
 endspecify
 
-assign Y = $signed(A) * $signed(B);
+if (A_SIGNED && B_SIGNED)
```
mwkmwkmwk, line 13: There is a problem here: mul2dsp can emit cells with `A_SIGNED != B_SIGNED`, which need special handling here (separate sign extension for A and B in separate wires).

mwkmwkmwk (approved):

PR #2362, opened 2020-08-27
Add the CMake build system script.

RonxBulld:


nakengelhardt:
Can you add some context for this PR about where it comes from, like what problems does it solve, does it work on all platforms, how does it change user flows? (I note that it doesn't update any documentation...)

PR #2365, opened 2020-08-29
Fix constant args used with function ports split across declarations

zachjs:
This fixes the issue I inadvertently introduced in #2299. Thanks to @mmicko for spotting and reporting this!


mmicko:
Thanks for the fix @zachjs , can confirm that all failing tests are fixed now and that there are no new issues noticed with this patch applied.

PR #2368, opened 2020-08-30
Fix import of VHDL enums

mmicko:
Current code fails to import VHDL design when one of ports is of type boolean.

### DIFF
--- a/frontends/verific/verific.cc
+++ b/frontends/verific/verific.cc
@@ -203,12 +203,6 @@ void VerificImporter::import_attributes(dict<RTLIL::IdString, RTLIL::Const> &att
 				auto p = v;
 				if (p) {
 					if (*p != '"') {
-						auto *q = p;
-						for (; *q != '\0'; q++)
-							if (*q != '0' && *q != '1') {
-								p = nullptr;
-								break;
-							}
 						if (p != nullptr)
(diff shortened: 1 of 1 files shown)


mmicko:
Added test in https://github.com/YosysHQ/yosys-tests/blob/master/verific/typerange/ to support this case as well.

frontends/verific/verific.cc
```diff
@@ -199,12 +199,17 @@ void VerificImporter::import_attributes(dict<RTLIL::IdString, RTLIL::Const> &att
...
+						auto l = strlen(p);
+						auto q = (char*)malloc(l+1);
+						strncpy(q, p, l);
+						q[l] = '\0';
+						for(char *ptr = q; *ptr; ++ptr )*ptr = tolower(*ptr);
+						attributes.emplace(stringf("\\enum_value_%s", q), RTLIL::escape_id(k));
+					} else {
 						auto *q = p+1;
 						for (; *q != '"'; q++)
 							if (*q != '0' && *q != '1') {
```
clairexen, line 215: This looks to me like we allow `x` but not `"x"`. I think we should not have an inconsistency like that.

clairexen (approved):
I'm merging this now so we can push a new release to the customer. Maybe we can address the comment in another PR.

PR #2378, opened 2020-09-15
Added $high(), $low(), $left(), $right()

udif:
This PR adds the $high(), $low(), $left(), $right() functions.
A future commit will add $increment(), $dimensions, $unpacked_dimensions()

### DIFF
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -1523,6 +1523,7 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
 	{
 		AstNode *index_expr = nullptr;
 
+		integer = children[0]->children.size(); // save original number of dimensions for $size() etc.
 		for (int i = 0; 2*i < GetSize(id2ast->multirange_dimensions); i++)
 		{
 			if (GetSize(children[0]->children) < i)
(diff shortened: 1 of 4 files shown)


frontends/ast/simplify.cc
```diff
@@ -2863,6 +2867,33 @@ skip_dynamic_range_lvalue_expansion:;
 						id_ast = current_scope.at(buf->str);
 					if (!id_ast)
 						log_file_error(filename, location.first_line, "Failed to resolve identifier %s for width detection!\n", buf->str.c_str());
+					// a slice of our identifier means we advance to the next dimension, e.g. $size(a[3])
+					if (buf->children.size() > 0) {
+						// we give up here because when we try to support thing such as $size(a[1][1]) the AST at this point doesn't contain
+						// the information how many indexes were given. The array is already flattened and a composite index is given in the AST instead.
+						log_file_error(filename, location.first_line, "%s() only supported for pure identifiers (no further indexing)!\n", str.c_str());
+						// something is hanging below this identifier
+						if (buf->children[0]->type == AST_RANGE)
```
daveshah1, line 2876: Isn't this code after log_file_error unreachable?

udif: Yes.
That error message wasn't meant to be the final word on this, just a workaround to get something out.
It was added there when I realized that the $size() argument was flattened, as a reminder to what the logic should do if I had the dimensions intact.
I've just added a fix that reuses the 'integer' field to keep the original level of array referencing (e.g. for a[9][8] it is 2), so now this error was removed as we can now safely handle stuff like this.
I also added the respective tests.

PR #2380, opened 2020-09-17
Clean up and parallelize testsuite

Xiretza:
- Factor out creating a Makefile from the `*.ys`/`*.v`/`*.sh` in the working directory to reduce redundancy in `run-test.sh` files
- Convert several tests to use Makefiles supporting `-j` for concurrency

Here are some crude timing benchmarks. First, on my i7-3520M (quad core) laptop:

.               | `make -j4 test` (before) | `make -j4 test` (after) | `make -j1`
--------------- | ------------------------ | ----------------------- | ----------
Wall time       | 899 s                    | 544 s                   | 1070 s
usr+sys time    | 1253 s                   | 1515 s                  | 1069 s
= concurrency   | 1.39                     | 2.78                    | 1.00

Resulting in a concurrency boost of **+100%** and a runtime change of **-39%**. I'm pretty sure the increase in total processor time is due to (increased) thermal throttling, so the measured effect on runtime is probably lower than it could be.

On my i7-4790 (8-core) desktop:

.               | `make -j8 test` (before) | `make -j8 test` (after) | `make -j1`
--------------- | ------------------------ | ----------------------- | ----------
Wall time       | 648 s                    | 328 s                   | 796 s
usr+sys time    | 975 s                    | 1282 s                  | 797 s
= concurrency   | 1.50                     | 3.91                    | 1.00

Which is a concurrency boost of **+161%** and a runtime change of **-49%**.

### DIFF
--- a/tests/arch/anlogic/run-test.sh
+++ b/tests/arch/anlogic/run-test.sh
@@ -1,20 +1,4 @@
 #!/usr/bin/env bash
-set -e
-{
-echo "all::"
-for x in *.ys; do
-	echo "all:: run-$x"
-	echo "run-$x:"
-	echo "	@echo 'Running $x..'"
-	echo "	@../../../yosys -ql ${x%.ys}.log -w 'Yosys has only limited support for tri-state logic at the moment.' $x"
-done
-for s in *.sh; do
-	if [ "$s" != "run-test.sh" ]; then
-		echo "all:: run-$s"
-		echo "run-$s:"
-		echo "	@echo 'Running $s..'"
-		echo "	@bash $s"
-	fi
-done
-} > run-test.mk
-exec ${MAKE:-make} -f run-test.mk
+set -eu
+source ../../gen-tests-makefile.sh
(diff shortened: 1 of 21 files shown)


clairexen:
I'm not a big fan of the use of `nullglob` and `failglob` and `pipefail` and so forth, as they make it harder for people to understand the scripts bc they are changing the default behavior of bash, so looking at a random line of the script might mislead people bc they lack the information that the default behavior of bash has been switched to something else in a different place in the script, or even another script included via `source`.

Xiretza:
Fair reasoning in principle, but `failglob` in particular is basically just free sanity checking, since it's exceedingly rare that one actually wants to silently expand a glob pattern to itself if there's no match. The `pipefail` is pure habit, I'll remove that. As for the `nullglob` in `tests/memories/run-test.sh`, I copied that line from `tests/simple/run-test.sh`, if it's not desired I can remove it from both.

Xiretza:
Removed all occurences of `nullglob`/`failglob`/`pipefail`.

PR #2381, opened 2020-09-18
Better error for unsupported SVA sequence

mmicko:


frontends/verific/verificsva.cc
```diff
@@ -1073,6 +1073,12 @@ struct VerificSvaImporter
 				break;
 			}
 
+			if (inst->Type() == PRIM_SVA_MATCH_ITEM_TRIGGER || inst->Type() == PRIM_SVA_MATCH_ITEM_ASSIGN)
+			{
+				parser_error("SVA sequences with local variable assignments are currently not supported.\n");
```
clairexen, line 1078: I think this should be part of `void parser_error(Instance *inst)` above, and we should still print instance type, name, and code location.

clairexen (changes requested):

frontends/verific/verificsva.cc
mmicko: updated now is it is in proper place, but displays in this case two-line error message, containing all info.

clairexen (approved):

PR #2386, opened 2020-09-25
Fixed python installation path

btut:
The path where python expects it's libraries seems to change from
operating system to operating system, but can be querried from the site
package.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -171,8 +171,7 @@ else
 PYTHON_CONFIG := $(PYTHON_EXECUTABLE)-config
 endif
 
-PYTHON_PREFIX := $(shell $(PYTHON_CONFIG) --prefix)
-PYTHON_DESTDIR := $(PYTHON_PREFIX)/lib/python$(PYTHON_VERSION)/site-packages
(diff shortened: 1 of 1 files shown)


mmicko (approved):
This seams to be fine with python2 and python3 variations I have found .

kbeckmann:
This unfortunately seems to have broken building with a different prefix, e.g. when creating a package such as the yosys-git AUR package for archlinux. I am not that familiar with the makefiles here so I don't have a suggestion on how to fix it. But when building the package (as an unprivileged user), I get the following error print:

`mkdir: cannot create directory ‘/usr/lib/python3.8/site-packages/pyosys’: Permission denied`

mmicko:
@kbeckmann Sorry for this, was not aware they were overriding this variable. Anyway should work just fine with https://github.com/YosysHQ/yosys/commit/9e00f3f141cd951b60b83df88c4a1d0e454ccc4d so please do try. DISTDIR is one that should be take into consideration anyway so it will work with current AUR scripts as well

kbeckmann:
Thank you for fixing it so fast @mmicko. Yes I have confirmed it, it works now again.

PR #2390, opened 2020-09-27
ice40: Provide default values for unconnected ports of SB_LUT4 and SB_RAM40_4K*

tpambor:
Instantiating `SB_LUT4` or `SB_RAM40_4K*` modules with unconnected inputs does not produce the expected results in simulation.

Fix by adding appropriate pull-up/downs to the signals.

### DIFF
--- a/techlibs/ice40/cells_sim.v
+++ b/techlibs/ice40/cells_sim.v
@@ -166,10 +166,16 @@ endmodule
 (* abc9_lut=1, lib_whitebox *)
 module SB_LUT4 (output O, input I0, I1, I2, I3);
 	parameter [15:0] LUT_INIT = 0;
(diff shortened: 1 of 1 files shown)


whitequark (changes requested):
I have no insight into whether the behavior that this PR adds is desirable (I'm not sure what the vendor tools do), but if it is, this PR should be using the Verilog input default values (see the `defaultvalue` attribute documentation) instead.

whitequark:
@daveshah1 Do you know if this is the right approach?

nakengelhardt:
Discussed this in JF today. The right way to fix this would be to use default assignments in the module port, as in
```
module test (input a = 0, output x); 
  assign x = a; 
endmodule
```
however icarus verilog does not support this syntax yet. I'll file a feature request there and we will wait to add defaults until it will work with iverilog too.

PR #2391, opened 2020-09-27
gowin: Fix cells_sim.v to make abc9 work

tpambor:
Fix synthesis for `gowin` with `abc9` enabled similar to a323881 for ice40.

### DIFF
--- a/techlibs/gowin/cells_sim.v
+++ b/techlibs/gowin/cells_sim.v
@@ -197,7 +197,7 @@ module DFFE (output reg Q, input D, CLK, CE);
   end
 endmodule // DFFE (positive clock edge; clock enable)
 
-(* abc9_box, lib_whitebox *)
+(* lib_whitebox *)
 module DFFS (output reg Q, input D, CLK, SET);
   parameter [0:0] INIT = 1'b1;
   initial Q = INIT;
@@ -216,7 +216,7 @@ module DFFS (output reg Q, input D, CLK, SET);
   end
 endmodule // DFFS (positive clock edge; synchronous set)
 
-(* abc9_box, lib_whitebox *)
+(* lib_whitebox *)
 module DFFSE (output reg Q, input D, CLK, CE, SET);
   parameter [0:0] INIT = 1'b1;
   initial Q = INIT;
@@ -281,8 +281,14 @@ module DFFP (output reg Q, input D, CLK, PRESET);
   initial Q = INIT;
 
 	specify
-		(posedge CLK => (Q : D)) = (480, 660);
+		if (!PRESET) (posedge CLK => (Q : D)) = (480, 660);
+`ifndef YOSYS
 		(posedge PRESET => (Q : 1'b1)) = (1800, 2679);
+`else
+		if (PRESET) (PRESET => Q) = (1800, 2679);	// Technically, this should be an edge sensitive path
+								// but for facilitating a bypass box, let's pretend it's
+								// a simple path
+`endif
 		$setup(D, posedge CLK, 576);
 	endspecify
 
@@ -300,8 +306,14 @@ module DFFPE (output reg Q, input D, CLK, CE, PRESET);
   initial Q = INIT;
 
 	specify
-		if (CE) (posedge CLK => (Q : D)) = (480, 660);
+		if (CE && !PRESET) (posedge CLK => (Q : D)) = (480, 660);
+`ifndef YOSYS
 		(posedge PRESET => (Q : 1'b1)) = (1800, 2679);
+`else
+		if (PRESET) (PRESET => Q) = (1800, 2679);	// Technically, this should be an edge sensitive path
+								// but for facilitating a bypass box, let's pretend it's
+								// a simple path
+`endif
 		$setup(D, posedge CLK &&& CE, 576);
 		$setup(CE, posedge CLK, 63);
 	endspecify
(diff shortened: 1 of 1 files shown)


Ravenslofty:
There appear to be a few misunderstandings in this patch.

First off: this patch does absolutely nothing at the moment. `(* abc9_flop *)` attributes only matter when `abc9 -dff` is used, which [it presently isn't](https://github.com/YosysHQ/yosys/blob/master/techlibs/gowin/synth_gowin.cc#L234-L241) for Gowin.

Second: even if `abc9 -dff` were used in a Gowin flow, this patch would still do absolutely nothing.

ABC9 has a fairly simplistic view of a flop, just a data input, clock input, and data output, and crucially, that the flop powers up to zero.

Gowin's flops power up to the reset value, so synchronous resets initialise to zero, but synchronous sets initialise to **one**. The flops that this patch changes have synchronous sets and thus do not match the ABC9 model; that could lead to ABC9 removing flops and changing the startup behaviour of the input. You will note that the synchronous-reset flops *do* have `(* abc9_flop *)` attributes.

Eddie's patch works because:
- iCE40 flops always power up to zero, and thus always match the ABC9 model.
- ECP5 and Xilinx flops can power up to zero or one, but ABC9 calls `zinit` to ensure they are zero-initialised.

To use ABC9 sequential synthesis, the Gowin flow would need to call `zinit` to ensure all flops are zero-initialised, and in the process, all places where you would find a `DFFS` cell that initialises to one would be replaced with a `DFFR` cell that initialises to zero, wrapped in inverters.

In other words: in the situation where the `(* abc9_flop *)` attribute is relevant for `DFFS`, the `DFFS` cell never gets instantiated in the first place. Therefore it is correct for all `DFFS`-type cells to be `(* abc9_box *)`, not `(* abc9_flop *)`.

tpambor:
Thanks for the detailed explanation. I am quite new to yosys so this was very helpful.

The error I got is:
```
2.25.10. Executing ABC9_OPS pass (helper functions for ABC9).
DFF.D[0] abc9_required = 576
DFFE.CE[0] abc9_required = 63
DFFE.D[0] abc9_required = 576
DFFRE.CE[0] abc9_required = 63
DFFRE.RESET[0] abc9_required = 63
DFFRE.D[0] abc9_required = 576

2.25.11. Executing ABC9_OPS pass (helper functions for ABC9).

2.25.12. Executing ABC9_OPS pass (helper functions for ABC9).
$__DFF_P__$abc9_flop.D abc9_required = 0
$__DFF_N__$abc9_flop.D abc9_required = 0
ERROR: Module 'DFFNS' with (* abc9_box *) has no timing (and thus no connectivity) information.
```

ECP5 uses `(* abc9_flop *)` also for `DFFS`. Whether `(* abc9_flop *)` or `(* abc9_box *)` is used only depends if set/reset is async or not, see https://github.com/YosysHQ/yosys/blob/08eb0821c9c62ef21e054b6a53a980019bbfa6ca/techlibs/ecp5/cells_sim.v#L309.

Nevertheless, it seems right that if support for `abc9 -dff` is implemented, it is necessary to run `zinit` first.

Ravenslofty:
That error is odd, because, well, [`DFFNS` has timing information](https://github.com/YosysHQ/yosys/blob/master/techlibs/gowin/cells_sim.v#L392-L396).

The ECP5 flop model is very different to Gowin's, because Lattice have [very strange primitive names](https://github.com/YosysHQ/yosys/blob/master/techlibs/ecp5/cells_ff.vh) for ECP5 flops; instead, Yosys abstracts over them with the `TRELLIS_FF` model. It's not really comparable.

Ravenslofty:
Use of `DFFNS` implies you have negative-edge synchronous logic. Could you try making all logic positive-edge synchronous? I'm wondering if ABC9 is failing to recognise the `negedge` specifier.

tpambor:
It has timing information but I think `abc9_box` (combinatorial) and `abc9_flop` (relative to clk) need different timing information. I think the error is thrown as of c52bb11 which checks more precisely if for a `abc9_box` combinatorial timing information is present.

There is no negative-edge synchronous logic in the design, I tested with a standard VexRiscv.

Ravenslofty:
Yes, you're right; boxes are combinational. In that case, I believe the correct fix would be to just remove `(* abc9_box *)` from those flops entirely.

tpambor:
Xilinx has it also for FFs with synchronous reset: https://github.com/YosysHQ/yosys/blob/08eb0821c9c62ef21e054b6a53a980019bbfa6ca/techlibs/xilinx/cells_sim.v#L535-L536

Therefore, I think the gowin techlib should align to the existing ECP5 and Xilinx techlibs.

Maybe @eddiehung, who implemented it for these, can give advice.

eddiehung:
`(* abc9_box *)` can be applied to a whitebox that has a combinatorial path through it. In the case of flops, this is applied to flops with asynchronous control because ABC9 cannot handle those, and the box feature is used to preserve the combinatorial async path.
It is also used for LUTRAMs which are async read.
`(* abc9_box *)` is supposed to be smart enough to detect combinatorial paths and put them into a box, but leave sequential paths as PI/POs so that ABC knows nothing about them. I suspect the error you're seeing is that it doesn't detect any combinatorial paths (from an input to output, without going through the FF). For this Gowin case, it is probably best to not annotate the "set" variants with anything, and just apply `(* abc9_flop *)` on the synchronous "reset" variants. Then it would be up to `zinit` to convert the former into the latter.

eddiehung:
> ABC9 has a fairly simplistic view of a flop, just a data input, clock input, and data output, and crucially, that the flop powers up to zero.

To be pedantic and for the record, it's even simpler: ABC9 doesn't know about clock inputs (they are left on the box to preserve them for un-mapping back to the original cell, after `abc9`, but are non functional). Clocking information is entered as an XAIG field, and only flops with the same clocking information are considered during sequential synthesis.

PR #2393, opened 2020-09-28
write_verilog: emit intermediate wire for constant values in sensitivity list

nakengelhardt:
While `posedge 1'b0` in a sensitivity list is technically standard-compliant verilog, it is not supported by many verilog frontends (including our own: `read_verilog` errors and `verific -sv` delivers wrong results...) To avoid these issues,  this makes `write_verilog` assign the constant value to a wire and use the wire in the sensitivity list.


whitequark:
Great work, I remember hitting this bug quite a few times.

PR #2397, opened 2020-10-05
synth_nexus: Initial implementation

daveshah1:
This adds synthesis support for the Lattice Nexus FPGA families (currently CrossLink-NX and Certus-NX), with Radiant being the current place and route option.

Although on the face of it ECP5 and Nexus are quite similar architectures, this adds a new command and set of libraries to take advantage of a 'fresh start' and change some design decisions. In particular, there is an aim for 100% compatibility with vendor tools, similar to synth_xilinx, which makes debugging easier. I plan for a similar 'fresh start' on the nextpnr arch side as part of general ongoing research into nextpnr quality improvements - for example the prjoxide/nextpnr-nexus database build will take 10 seconds rather than 10 minutes.

Hopefully in both cases this might trickle down via backporting ECP5 to the new approach so it will benefit from this work too.

Notable current omissions:
 - DSP inference
 - Accurate timing per LUT input (this hasn't been documented yet)
 - Sim models for DSP primitives, block RAM, IO registers, etc

This can currently be tested in two ways:
 - Feeding the 'vm' file into Synplify, essentially running synthesis twice
 - Skipping Lattice's synthesis, by feeding the 'vm' file into postsyn - starting from command 2 of the 'Using Command Line Shell Scripts' section of the Radiant user guide


techlibs/nexus/cells_map.v
```diff
@@ -0,0 +1,200 @@
...
+
+`ifndef NO_LUT
+module \$lut (A, Y);
+	parameter WIDTH = 0;
+	parameter LUT = 0;
+
+	input [WIDTH-1:0] A;
+	output Y;
+
+	// Need to swap input ordering, and fix init accordingly,
```
Ravenslofty, line 143: I'm pretty sure this is outdated; ABC9 sorts inputs to achieve this anyway.

techlibs/nexus/cells_map.v
```diff
@@ -0,0 +1,200 @@
+// Flipflop intermediate map level
+module \$__FF_NOLSR (input D, C, E, output Q);
```
Ravenslofty, line 2: Why not rely on `dfflegalize` to insert inverters where necessary? Do they have sufficiently specific placement requirements that turning LUT1s with inverter LUTs generated by e.g. ABC into `INV`s won't map properly?

techlibs/nexus/cells_sim.v
```diff
@@ -0,0 +1,373 @@
...
+		(SEL => Z) = 171;
+		(D => Z) = 303;
+		(C => Z) = 311;
+		(B => Z) = 309;
+		(A => Z) = 306;
+	endspecify
+endmodule
+
+// Two LUT4s and MUX2
+module WIDEFN9(input A0, B0, C0, D0, A1, B1, C1, D1, SEL, output Z);
```
Ravenslofty, line 36: No LUT6?

techlibs/nexus/synth_nexus.cc
```diff
@@ -0,0 +1,376 @@
...
+		log("    -nolutram\n");
+		log("        do not use LUT RAM cells in output netlist\n");
+		log("\n");
+		log("    -nowidelut\n");
+		log("        do not use PFU muxes to implement LUTs larger than LUT4s\n");
+		log("\n");
+		log("    -noiopad\n");
+		log("        do not insert IO buffers\n");
+		log("\n");
+		log("    -abc9\n");
```
Ravenslofty, line 92: Eddie complained, but `synth_intel_alm` uses ABC9 exclusively, and I think it's a bit difficult to argue for a non-ABC9 flow where you have even a bit of timing information.

techlibs/nexus/synth_nexus.cc
```diff
@@ -0,0 +1,376 @@
...
+
+		run_script(design, run_from, run_to);
+
+		log_pop();
+	}
+
+	void script() override
+	{
+
+		if (family != "lifcl" && family != "lfd2nx")
```
Ravenslofty, line 217: Do you *need* to use the vendor product codes in Yosys? Surely "crosslink" and "certus" would be more user friendly here.

techlibs/nexus/synth_nexus.cc
```diff
@@ -0,0 +1,376 @@
...
+		if (check_label("map_ffs"))
+		{
+			run("opt_clean");
+			std::string dfflegalize_args = " -cell $_DFF_?_ 01 -cell $_DFF_?P?_ r -cell $_SDFF_?P?_ r";
+			if (help_mode) {
+				dfflegalize_args += " [-cell $_DFFE_??_ 01 -cell $_DFFE_?P??_ r -cell $_SDFFE_?P??_ r]";
+			} else if (!nodffe) {
+				dfflegalize_args += " -cell $_DFFE_??_ 01 -cell $_DFFE_?P??_ r -cell $_SDFFE_?P??_ r";
+			}
+			dfflegalize_args += " -cell $_DLATCH_?_ x";
```
Ravenslofty, line 298: My understanding from @mwkmwkmwk is that wire loop latches are not useful from `dfflegalize`'s point of view.

techlibs/nexus/cells_map.v
daveshah1: Thanks, haven't looked at that side for a while

techlibs/nexus/cells_map.v
daveshah1: Although I do map inverter LUTs to `INV` anyway, might as well use `INV` to explicitly mark hardware inversion - place and route folds them into the flop.

techlibs/nexus/cells_sim.v
daveshah1: No not anymore

techlibs/nexus/synth_nexus.cc
daveshah1: synth_xilinx uses xc7, xc6s, etc instead of 7series, spartan6, etc so this way seems more consistent?

techlibs/nexus/synth_nexus.cc
daveshah1: Ack, I was copying this from synth_ecp5 but it seems like that could be tidied up, too

techlibs/nexus/cells_map.v
Ravenslofty: Sure, but ABC could fold it into a LUT, making it effectively free. I'm inclined to say that using dfflegalize is going to mean less code to maintain in the future. It's what it's there for, after all.

techlibs/nexus/cells_map.v
Ravenslofty: ABC1 doesn't care, and ABC9 handles it internally, so yeah I'd say this code is obsolete (possibly in ECP5 too?)

daveshah1: Yes, again this is something where ECP5 should be tidied up too

techlibs/nexus/cells_map.v
daveshah1: Yes, makes sense

techlibs/nexus/synth_nexus.cc
Ravenslofty: And synth_intel_alm uses "cyclonev" and "cyclone10gx" instead of "5c" and "10cgx".

techlibs/nexus/synth_nexus.cc
daveshah1: The main problem is the timing information is pretty janky and I'm not sure is really helping much - I think I saw a higher area with ABC9 than ABC in one case. Perhaps I can fudge the timing to fix that though.

daveshah1: Once timing is properly fuzzed, then I would be happier with abc9-as-default

Ravenslofty: You could put this in the `eddie/sta` branch and measure critical path delays. But yeah, with present timing information it's probably equivalent to unit delay at the moment.

techlibs/nexus/cells_map.v
daveshah1: Nevermind, looks like this has now been tidied up in ECP5 and I had copied an old file

techlibs/nexus/synth_nexus.cc
daveshah1: Just tried removing it for synth_ecp5 but that doesn't seem to work:

```
ERROR: FF latchp.$auto$simplemap.cc:581:simplemap_dlatch$354 (type $_DLATCH_P_) cannot be legalized: dlatch are not supported
```

Maybe I am missing something though?

techlibs/nexus/cells_map.v
daveshah1: Thanks, pushed fix

techlibs/nexus/cells_map.v
daveshah1: Thanks, pushed fix

techlibs/nexus/synth_nexus.cc
daveshah1: This is probably something that needs more general standardisation. iCE40 uses 'hx', 'lp' and 'u' which are closer to part numbers but 'u' in particular is a bit vague.

tests/arch/nexus/mul.ys
```diff
@@ -0,0 +1,28 @@
...
+
+equiv_opt -assert -map +/nexus/cells_sim.v synth_nexus -abc9
+design -load postopt # load the post-opt design (otherwise equiv_opt loads the pre-opt design)
+cd top # Constrain all select calls below inside the top module
+
+stat
+
+select -assert-count 7 t:CCU2
+select -assert-max 12 t:WIDEFN9
+select -assert-max 56 t:LUT4
```
mmicko, line 26: This should be 58 for test to pass

mmicko:
@daveshah1 is there anything pending except test fix ?

daveshah1:
Test fixed pushed; thanks; I think everything else is OK for a first implementation now

techlibs/nexus/cells_sim.v
```diff
@@ -0,0 +1,379 @@
...
+endmodule
+
+module VHI(output Z);
+	assign Z = 1'b1;
+endmodule
+
+// Vendor flipflops
+// (all have active high clock, enable and set/reset - use INV to invert)
+
+// Async preload
```
mwkmwkmwk, line 92: isn't that commonly called preset?

techlibs/nexus/synth_nexus.cc
mwkmwkmwk: tbh I'd vote for accepting more variants

techlibs/nexus/synth_nexus.cc
mwkmwkmwk: there are some plans to improve it, but until we have a better solution, `$_DLATCH_*` needs to be present here

techlibs/nexus/synth_nexus.cc
```diff
@@ -0,0 +1,376 @@
...
+		{
+			run("opt_clean");
+			std::string dfflegalize_args = " -cell $_DFF_P_ 01 -cell $_DFF_PP?_ r -cell $_SDFF_PP?_ r";
+			if (help_mode) {
+				dfflegalize_args += " [-cell $_DFFE_PP_ 01 -cell $_DFFE_PP?P_ r -cell $_SDFFE_PP?P_ r]";
+			} else if (!nodffe) {
+				dfflegalize_args += " -cell $_DFFE_PP_ 01 -cell $_DFFE_PP?P_ r -cell $_SDFFE_PP?P_ r";
+			}
+			dfflegalize_args += " -cell $_DLATCH_?_ x";
+			if (help_mode) {
```
mwkmwkmwk, line 299: huh, what is this for?

techlibs/nexus/cells_sim.v
daveshah1: Yes, that was a typo. fixed

techlibs/nexus/synth_nexus.cc
daveshah1: bad copy-pasta from synth_ecp5, thanks for noticing. fixed

tests/arch/nexus/run-test.sh
```diff
@@ -0,0 +1,20 @@
...
+for s in *.sh; do
+	if [ "$s" != "run-test.sh" ]; then
+		echo "all:: run-$s"
+		echo "run-$s:"
+		echo "	@echo 'Running $s..'"
+		echo "	@bash $s"
+	fi
+done
+} > run-test.mk
+exec ${MAKE:-make} -f run-test.mk
```
Xiretza, line 20: Could this be redone in the same way as #2380 for consistency?

PR #2398, opened 2020-10-06
smtbmc: escape identifiers in verilog testbench

jakobwenzel:
yosys-smtbmc's Verilog testbenches do not escape any identifier names. When identifiers contain special characters, the emitted code is invalid. This PR tries to fix that.

### DIFF
--- a/backends/smt2/smtbmc.py
+++ b/backends/smt2/smtbmc.py
@@ -817,6 +817,24 @@ def write_vcd_trace(steps_start, steps_stop, index):
 
         vcd.set_time(steps_stop)
 
+def char_ok_in_verilog(c,i):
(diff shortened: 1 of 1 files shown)


jakobwenzel:
Additionally, Backslashes in identifiers are converted to forward slashes in the SMT2 backend (introduced here: https://github.com/YosysHQ/yosys/commit/17233b11e185d2c863ccd06830e8cd0d2be38e83#diff-1b784d2db285d994736b273ec3a3390cR76 ). The identifiers in generated testbenches then do not match the identifiers in the actual design.

When commenting out these lines for testing, everything seemed to work. I am probably missing some corner case. Why do the backslashes need to be replaced?

clairexen (approved):
LGTM

PR #2399, opened 2020-10-08
opt_clean: Better memory handling.

mwkmwkmwk:
Previously, `$memwr` and `$meminit` cells were always preserved (along
with the memory itself).  With this change, they are instead part of the
main cell mark-and-sweep pass: a memory (and its `$meminit` and `$memwr`
cells) is only preserved iff any associated `$memrd` cell needs to be
preserved.


clairexen (approved):
LGTM

PR #2414, opened 2020-10-27
Prevent CXXFLAGS from leaking to abc Makefile

zeldin:
Under certain circumstances¹, the CXXFLAGS assigned by the main `Makefile` can leak down to `abc/Makefile`.  This causes them to be passed along to `abc/depends.sh` when depending C++ code (such as `abc/src/sat/glucose/AbcGlucose.cpp`).  With `gcc` this seems work cause no problems, but when the compiler is `clang` this causes an issue because `clang` is unhappy with getting both `-MM` (which is from `abc/depends.sh`) and `-MD` (which is from `Makefile`) at the same time, and outputs preprocessed C code instead of dependency rules.

By unexporting CXXFLAGS, the leak is plugged and the build completes normally with `clang`.

¹ This happens when the environment variable CXXFLAGS is set to something in the environment where the top level `make` command is run.  The fact that CXXFLAGS was set in the environment causes `make` to re-export the value of `$(CXXFLAGS)` that it sets to the environment where it can be picked up by the sub make process.  Note that the flags which break the build do not come from the environment variable, but the environment variable must be set for them to be exported.  This is a quirk of how variable export works in `make`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -661,6 +661,10 @@ ifeq ($(LINK_ABC),1)
 OBJS += $(PROGRAM_PREFIX)yosys-libabc.a
 endif
 
+# prevent the CXXFLAGS set by this Makefile from reaching abc/Makefile,
+# especially the -MD flag which will break the build when CXX is clang
+unexport CXXFLAGS
+
 top-all: $(TARGETS) $(EXTRA_TARGETS)
 	@echo ""
 	@echo "  Build successful."


whitequark:
After this patch, how does one pass flags to abc e.g. for cross-compilation?

zeldin:
@whitequark Well, I guess by setting `CC=` and `CXX=`, since those are explicitly forwarded to abc?
If you want CXXFLAGS from env to reach abc, then an alternative solution would be to use a different variable name from "CXXFLAGS" in the main Makefile, eg `YOSYS_CXXFLAGS := $(CXXFLAGS)` and then append to that variable instead...

zeldin:
Eh, sorry, setting `CXX=` is enough since it is forwarded as "CC" as well.  :smile:

whitequark:
> Well, I guess by setting `CC=` and `CXX=`, since those are explicitly forwarded to abc?

As in, adding compiler flags to `CC` and `CXX`?

zeldin:
> As in, adding compiler flags to `CC` and `CXX`?

Yes.  When cross compiling you'll need to set the name of your cross compiler of course (like `riscv64-unknown-linux-gnu-gcc`), but you can add compiler flags as well since the variables are expanded without quoting.  So something like `CXX="gcc -mbig-endian"` works as intended.

whitequark (approved):

whitequark:
> So something like `CXX="gcc -mbig-endian"` works as intended.

Right. That's perhaps not ideal but I think we can live with that.

Makefile
```diff
@@ -661,6 +661,8 @@ ifeq ($(LINK_ABC),1)
 OBJS += $(PROGRAM_PREFIX)yosys-libabc.a
 endif
 
+unexport CXXFLAGS
```
whitequark, line 664: Now that I give this PR another look, it would be good to have a comment here explaining the reason for `unexport`.

PR #2416, opened 2020-10-29
Adds support for defining abc location at runtime

QuantamHD:
This patch adds support for defining the ABC location at runtime instead of at compile time. This is helpful in build systems like bazel which do not have stable locations for binaries or directories during the compilation phase.

This change should be backwards compatible with the existing behavior.

### DIFF
--- a/passes/techmap/abc.cc
+++ b/passes/techmap/abc.cc
@@ -1471,7 +1471,12 @@ struct AbcPass : public Pass {
 		po_map.clear();
 
 #ifdef ABCEXTERNAL
-		std::string exe_file = ABCEXTERNAL;
+		std::string exe_file;
+		if (std::getenv("ABC")) {
+			exe_file = std::getenv("ABC");
+		} else {
+			exe_file = ABCEXTERNAL;
+		}
 #else
 		std::string exe_file = proc_self_dirname() + proc_program_prefix() + "yosys-abc";
 #endif


passes/techmap/abc.cc
```diff
@@ -1471,7 +1471,12 @@ struct AbcPass : public Pass {
 		po_map.clear();
 
 #ifdef ABCEXTERNAL
-		std::string exe_file = ABCEXTERNAL;
+		std::string exe_file;
+		if (std::getenv("ABCEXTERNAL")) {
```
whitequark, line 1475: This should definitely not be called `ABCEXTERNAL`. The convention would be to use `ABC`, like you'd have `CC`, `CXX`, and so on.

kernel/yosys.cc
```diff
@@ -848,6 +848,8 @@ std::string proc_share_dirname()
 	if (check_file_exists(proc_share_path, true))
 		return proc_share_path;
 #    ifdef YOSYS_DATDIR
+	if (std::getenv("YOSYS_DATDIR"))
+		return std::getenv("YOSYS_DATDIR");
```
whitequark, line 852: I'm not sure how I feel about this change. Does Bazel require patching virtually every other utility following standard Unix conventions, too?

passes/techmap/abc.cc
QuantamHD: Done.

kernel/yosys.cc
QuantamHD: I can't answer for certain but typically in bazel you would specify some binary in this case yosys as 

'''
"yosys": attr.label(
            default = Label("//utils/yosys:yosys"),
            executable = True,
            cfg = "exec",
        ),
'''

which will be available to the the compilation setup as `ctx.executable.yosys.path` these path components are not stable to encourage hermetic compilations. The path might be something like bazel-out{uniqueid}/utils/yosys/yosys. In this case compiling yosys is part of the build system which is very typical in bazel.

whitequark: Well, what happens with gcc or clang? Do they also use environment variables for sysroot, etc?

QuantamHD: Looks like there is some dynamic nature to the sysroot https://docs.bazel.build/versions/master/skylark/lib/CcToolchainInfo.html#sysroot

QuantamHD: From https://gcc.gnu.org/onlinedocs/gcc/Directory-Options.html it looks like sysroot is configurable. Is your concern with the configuration ability or how it's being configured? If this is better set as a flag I can make that change.

```
--sysroot=dir
Use dir as the logical root directory for headers and libraries. For example, if the compiler normally searches for headers in /usr/include and libraries in /usr/lib, it instead searches dir/usr/include and dir/usr/lib.

If you use both this option and the -isysroot option, then the --sysroot option applies to libraries, but the -isysroot option applies to header files.

The GNU linker (beginning with version 2.16) has the necessary support for this option. If your linker does not support this option, the header file aspect of --sysroot still works, but the library aspect does not.
```

whitequark: Speaking in personal capacity (ie not a Yosys maintainer): I would probably make the toolchain relocatable. But I would leave this particular decision to another maintainer.

(I'll be happy to merge the abc change though.)

QuantamHD: I removed the YOSYS_DATADIR change.

whitequark (approved):

PR #2417, opened 2020-10-29
Adds support for defining the YOSYS_DATDIR location at runtime

QuantamHD:
This is helpful in build systems like Bazel which do not have stable locations for binaries or directories during the compilation phase.

This change should be backwards compatible with the existing behavior.

This change was split from https://github.com/YosysHQ/yosys/pull/2416 where @whitequark was concerned about the dynamic toolchain location YOSYS_DATDIR, and wanted to leave it to a maintainer to decide whether or not to merge. 

Currently this change sets the value via an env variable, but it could also be set as a command line flag if the maintainers view that as better functionality for the codebase.

### Previous Conversation 

https://github.com/YosysHQ/yosys/pull/2416#discussion_r513857756

### DIFF
--- a/kernel/yosys.cc
+++ b/kernel/yosys.cc
@@ -848,6 +848,8 @@ std::string proc_share_dirname()
 	if (check_file_exists(proc_share_path, true))
 		return proc_share_path;
 #    ifdef YOSYS_DATDIR
+	if (std::getenv("YOSYS_DATDIR"))
+		return std::getenv("YOSYS_DATDIR");
 	proc_share_path = YOSYS_DATDIR "/";
 	if (check_file_exists(proc_share_path, true))
 		return proc_share_path;


stefanpie:
I know it's been a while, but I wanted to say that this is still a desired feature. This would allow for statically linked / "portable" builds of yosys where both the binaries and /share directories can be copied between machines or into different dirs, and it would still work (I think this is true, but please correct me if I am mistaken).

whitequark:
I think Yosys now checks an executable-relative share folder, does it not?

stefanpie:
After looking over the code again, I believe it does (at least targets I am looking at) according to this: https://github.com/YosysHQ/yosys/blob/99f7d79abb6a4514c141366237f510b1c1ec6679/kernel/yosys.cc#L539-L589.

I was working on a similar patch to make this work, but not sure if env var defined share paths are still a desired feature (at least for targets where that makes sense for).

whitequark:
I think an executable-relative path is strictly better for portable installs as it requires no configuration.

stefanpie:
Cool, no worries then!

PR #2420, opened 2020-10-29
backends/blif: Remove unused vector of strings

woodruffw:
For reasons that are unclear to me, this was being used to store every
result of `cstr` before returning them. The vector was never accessed otherwise,
resulting in a huge unnecessary memory sink when emitting to BLIF.

On local tests, this saves an extraordinary amount of memory (and subsequently time) during BLIF emission:
an 87MB BLIF output previously required 1GB of resident memory, and now requires virtually none.

### DIFF
--- a/backends/blif/blif.cc
+++ b/backends/blif/blif.cc
@@ -118,8 +118,6 @@ struct BlifDumper
 		return str;
 	}
 
-	#define CSTR(x) (str(x).c_str())
-
 	const std::string str_init(RTLIL::SigBit sig)
 	{
 		sigmap.apply(sig);
@@ -166,7 +164,7 @@ struct BlifDumper
 	void dump()
 	{
 		f << stringf("\n");
-		f << stringf(".model %s\n", CSTR(module->name));
+		f << stringf(".model %s\n", module->name.c_str());
 
 		std::map<int, RTLIL::Wire*> inputs, outputs;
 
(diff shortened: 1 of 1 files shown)


backends/blif/blif.cc
```diff
@@ -117,11 +115,12 @@ struct BlifDumper
 		if (sig.wire->width != 1)
 			str += stringf("[%d]", sig.wire->upto ? sig.wire->start_offset+sig.wire->width-sig.offset-1 : sig.wire->start_offset+sig.offset);
 
-		cstr_buf.push_back(str);
-		return cstr_buf.back().c_str();
+		return str;
 	}
 
-	const char *cstr_init(RTLIL::SigBit sig)
+	#define CSTR(x) (str(x).c_str())
```
woodruffw, line 121: N.B.: I stuffed this macro in here to avoid rewriting each of the `cstr` calls that were originally below. I'm happy to rewrite them, though, if you'd prefer to not have a macro.

woodruffw:
(Oh, and I forgot to say since GitHub just used my commit message for the PR body: thanks to the Yosys maintainers for their hard work and for reviewing this!)

backends/blif/blif.cc
whitequark: If you get rid of the macro I'll be happy to merge this right away.

woodruffw:
Sounds good. I’ll remove it tomorrow morning (ET).

backends/blif/blif.cc
woodruffw: @whitequark all done!

whitequark (approved):

woodruffw:
Thanks!

PR #2425, opened 2020-11-01
cxxrtl: don't assert on non-constant $meminit inputs

### LINKED ISSUE #2129: write_cxxrtl crashes with assertion failure on $meminit cells with non-constant ADDR/DATA
As requested:

`yosys -p 'read_ilang ics32_ilang; write_cxxrtl -O0 sim.cpp'`

Using the following as input:
[ics32_ilang.zip](https://github.com/YosysHQ/yosys/files/4751382/ics32_ilang.zip)

## Expected behavior

Outputs sim.cpp without error

## Actual behavior

```
ERROR: Assert `is_fully_const() && GetSize(chunks_) <= 1' failed in kernel/rtlil.cc:3801.`
```

## Version

Yosys built from c172b30c24285faa2b57f84a2a5162c5f3dcfb4c

whitequark:
Fixes #2129.

### DIFF
--- a/backends/cxxrtl/cxxrtl_backend.cc
+++ b/backends/cxxrtl/cxxrtl_backend.cc
@@ -22,6 +22,7 @@
 #include "kernel/sigtools.h"
 #include "kernel/utils.h"
 #include "kernel/celltypes.h"
+#include "kernel/mem.h"
 #include "kernel/log.h"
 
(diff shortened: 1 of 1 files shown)


whitequark:
@mwkmwkmwk LGTY? I left the `memory_collect` pass call intact so that it's more obvious to the user what's going on.

PR #2428, opened 2020-11-03
check: add support for processes

whitequark:
This is helpful for e.g. nMigen and CXXRTL, both of which support processes natively without requiring them to be expanded immediately after the frontend runs.


whitequark:
This PR currently does not enable tracking logic loops because it's not trivial to do with processes, though arguably this should be *the* place to track them.

clairexen (approved):
LGTM

PR #2430, opened 2020-11-06
YOSYS_HOME to override process path

mmicko:
To be able to use and check pyosys builds without installing yosys itself (for CI) we need to override main process location, since in this case it points out to python executable location itself.

### DIFF
--- a/kernel/yosys.cc
+++ b/kernel/yosys.cc
@@ -749,7 +749,7 @@ struct TclPass : public Pass {
 #endif
 
 #if defined(__linux__) || defined(__CYGWIN__)
-std::string proc_self_dirname()
+std::string proc_self_dirname_exe()
 {
 	char path[PATH_MAX];
 	ssize_t buflen = readlink("/proc/self/exe", path, sizeof(path));
@@ -761,7 +761,7 @@ std::string proc_self_dirname()
 	return std::string(path, buflen);
 }
 #elif defined(__FreeBSD__)
-std::string proc_self_dirname()
+std::string proc_self_dirname_exe()
 {
 	int mib[4] = {CTL_KERN, KERN_PROC, KERN_PROC_PATHNAME, -1};
 	size_t buflen;
@@ -781,7 +781,7 @@ std::string proc_self_dirname()
 	return path;
 }
 #elif defined(__APPLE__)
-std::string proc_self_dirname()
+std::string proc_self_dirname_exe()
 {
 	char *path = NULL;
 	uint32_t buflen = 0;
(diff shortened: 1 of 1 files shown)


whitequark:
Related to #2417, and I think these two should be fixed in the same way (but I'm not yet sure which). I'll try to come up with something.

mmicko:
This should also resolve same issue as #2417 but also those covered by #2416 but with providing one environment. Think #2416 should stay as well, since it gives more fine grain setup.

whitequark:
Agreed, I mainly merged #2416 because it may be useful to override abc location for local testing or such, and it's using an utterly standard way of doing so.

> in this case it points out to python executable location itself.

Thinking more about this case: shouldn't configuration in this case happen through Python? If pyosys is a native extension I would expect to be able to make it relocatable without messing with the environment of the process, i.e. through some kind of `pyosys.yosys_data_dir` hook or something. Basically, I would probably want to be able to build pyosys as a redistributable binary wheel.

mmicko:
Yeah, good point. In that case exposing abc location would be needed as well (since we need to be able to execute that one as well).

whitequark:
> In that case exposing abc location would be needed as well (since we need to be able to execute that one as well).

Agreed. Seems like they could go in two globals, initialized with the current method by default, but overridable?

mmicko:
well ABC can be overwritten only if ABCEXTERNAL is set on compile that means that we would need to have global keeping "calculated" location that can be over written by python code. And also another global for "proc_share_dirname".

Other way would be just exposing "proc_self_dirname" able to be overwritten since ABC location is deduced from it.

whitequark:
> that means that we would need to have global keeping "calculated" location that can be over written by python code. And also another global for "proc_share_dirname".

Yup, I think that would be a reasonable way to do it.

mmicko:
Created new branch for this https://github.com/YosysHQ/yosys/tree/paths_as_globals, works fine but makes python tests bit more complicated (setup fixtures), will see what I can do to improve that, using environment made test code not aware of paths.

whitequark:
> Created new branch for this https://github.com/YosysHQ/yosys/tree/paths_as_globals,

Looks reasonable to me, there are a few minor cleanups I'll ask you to do, but we can leave that until PR review.

> will see what I can do to improve that, using environment made test code not aware of paths.

I think you can still use environment in the testsuite, I am only concerned about two things:
  * To a greater degree, the fact that you currently cannot ship a pyosys wheel;
  * To a lesser degree, Yosys changing its behavior based on obscure environment variables.

Issue #2434, opened 2020-11-10
Add command or option to split assignments


pepijndevos:
Given the following example, `synth_gowin` produces code that cannot be parsed by the vendor toolchain.

```verilog
module top (input clk, output [1:0] leds);

reg [1:0] ctr;
always @(posedge clk)
	ctr <= ctr + 1'b1;

assign leds = ctr;

endmodule
```
Compile as follows. `-noiopads` is used for a minimal example, but the same behaviour can be seen _with_ iopads in more complex designs.
```
read_verilog blinky.v
synth_gowin -noiopads -vout vendorblinky.vg
dump
```
`synth_gowin` uses `splitnets` to split up nets in single wires as needed for the vendor netlist. However, it does not split up the connection. So the last line of the generated verilog will read as follows, which the vendor tool cannot parse.
```verilog
  assign leds = { \ctr[1] , \ctr[0]  };
```
The vendor tools gives
```
ERROR  (PA2000) : "/home/pepijn/code/apicula/vendorblinky.vg":47 | syntax error near token '{'
```
So a way is needed to split up the assignment itself. Note that the `-ports` option to `splitnets` does not help here, and only further complicates the LHS of the assignment.

I guess I'm kinda the de facto maintainer of `synth_gowin`, so I'm happy to take a stab at implementing this behaviour. I'm just hoping for some input from people more familiar with yosys internals on how to go about this.

Ravenslofty:
Isn't this just #1286 all over again?

pepijndevos:
You are completely right. Derp.

Issue #1286, opened 2019-08-12
Yosys emits assigns with a concatenated ecpression on the LHS


byzantic:
Yosys can emit verilog that has a concatenated signal as the recipient of an assign statement.

I.e something like:

`assign {a,b} = {c,d};`

The problem is that this causes problems for other tools; in this case OpenSTA [*], which means I can't properly verify my design.

I did not see this issue until recently, and it occurred after I introduced a `hilomap` stage, possibly because that reduces the fanin of constant buses. It's difficult to provide an example because it only seems to happen under certain circumstances.

I did try adding some `opt_clean` and `clean` stages after the `hilomap`, but that didn't seem to help.

I think I saw a previous issue/communication about this somewhere (but can't find it anymore ..), where it was stated that this was legal verilog, and therefore should be OK. Well, I can see that point of view, but it just seems it would cause less trouble if such assigns were split into multiple assigns with only one signal as their recipient.

[*] OpenSTA is really great, by the way.

pepijndevos:
I'll just add that the Gowin tools also don't like post-synthesis Verilog with this style of assignment.

whitequark:
It occurred to me that Yosys already has a pass that can avoid having such assignments be emitted: simply run `splitnets -driver` before `write_verilog`. Can you confirm this works (or does not) for you?

pepijndevos:
It seems that this would *reduce* the number of splits that are done. Post-synthesis verilog in Gowin is very limited and I'm already using plain `splitnets` to make it happy. I can try this when I look at the other Gowin changes Eddie suggested. (might be a while) What I think needs to be done is slightly different, which *does* split all the nets, but assigns individual bits to the output port. The `-port` option might actually be a workaround for Gowin, as top-level output ports are single pins anyway, but this might require adjusting pin mapping to some Yosys-generated name, which is not ideal.

byzantic:
Actually, I think this was suggested when I first raised the issue. Unfortunately, it doesn't work.

Using `splitnets -driver` does not make any difference to my output netlist - I get the same errors.
Using `splitnets` by itself makes the issue worse - I get more occurrences of the problem.

I think the point is that it's _not_ a driver issue - `splitnets` is probably doing its job quite happily.

whitequark:
Oh, you're right, and I misunderstood this issue.

pepijndevos:
Awesome!

PR #2435, opened 2020-11-10
synth_gowin: move splitnets to after iopadmap

pepijndevos:
`splitnets` will not split ports by default, but `iopadmap` creates a new internal net which could be a wider bus. By moving the command down, any generated nets are also split.

This does not completely solve #2434 but it should resolve some cases where it happens.

### DIFF
--- a/techlibs/gowin/synth_gowin.cc
+++ b/techlibs/gowin/synth_gowin.cc
@@ -278,7 +278,7 @@ struct SynthGowinPass : public ScriptPass
 			run("setundef -undriven -params -zero");
 			run("hilomap -singleton -hicell VCC V -locell GND G");
 			if (!vout_file.empty() || help_mode) // vendor output requires 1-bit wires
-				run("splitnets -ports");
+				run("splitnets -ports", "(only if -vout used)");
 			run("clean");
 			run("autoname");
 		}
--- a/techlibs/gowin/synth_gowin.cc
+++ b/techlibs/gowin/synth_gowin.cc
@@ -277,7 +277,8 @@ struct SynthGowinPass : public ScriptPass
 			run("opt_lut_ins -tech gowin");
 			run("setundef -undriven -params -zero");
 			run("hilomap -singleton -hicell VCC V -locell GND G");
-			run("splitnets -ports");
+			if (!vout_file.empty() || help_mode) // vendor output requires 1-bit wires
+				run("splitnets -ports");
 			run("clean");
 			run("autoname");
 		}
--- a/techlibs/gowin/synth_gowin.cc
+++ b/techlibs/gowin/synth_gowin.cc
@@ -238,7 +238,6 @@ struct SynthGowinPass : public ScriptPass
 			run("opt -fast");
 			if (retime || help_mode)
 				run("abc -dff -D 1", "(only if -retime)");
-			run("splitnets");
 			if (!noiopads || help_mode)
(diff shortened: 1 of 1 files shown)


pepijndevos:
I have confirmed with @kbeckmann that the initial change fixes the test case, but still brakes on his larger design. This design uses a wide top-level port, causing yosys to make a wide assign statement.

By splitting up ports this behaviour is avoided. This is not ideal, but due to the splitting format, constraints remain unchanged so there should be no disruption for the user.

I believe this now handles all cases that I'm aware of of bug #2434 but yosys could potentially still generate wide assign construct in unforeseen edge cases. I'm not sure how realistic this is, and how much work a proper fix is, so not sure what is a pragmatic solution.

pepijndevos:
#2434 is actually a duplicate of #1286 that... I forgot about.

pepijndevos:
All the tests were broken because `splitnets -ports` breaks the equivalence check. Splitnets is only needed when outputting verilog for the vendor tools, so wrapping it an if statement fixed all the tests and still makes sure the vendor PnR doesn't barf on multi-bit assignments as much.

techlibs/gowin/synth_gowin.cc
```diff
@@ -280,6 +280,8 @@ struct SynthGowinPass : public ScriptPass
 			run("opt_lut_ins -tech gowin");
 			run("setundef -undriven -params -zero");
 			run("hilomap -singleton -hicell VCC V -locell GND G");
+			if (!vout_file.empty() || help_mode) // vendor output requires 1-bit wires
```
mwkmwkmwk, line 283: note that this conition will unconditionally show this line in help mode without any annotation; change the following line to something like this:

`run("splitnets -ports", "(only if -vout used)");`

mwkmwkmwk:
Also: please rebase this branch on top of current master

pepijndevos:
What the hell did I do? I rebased on top of master and now it's completely fucked up. Great idea that. Also... how did I request a code review?? AAAAAAAH WHAT IS HAPPENING???

pepijndevos:
Oh I think I know what happened?? My local master was very old so I rebased on top of that which created a huge mess when updating the PR and because it included a bazillion old commits in my PR it requested a review of the code owner of files I "changed". Welll. sooooorrry for the mess I think it's back to normal now, saying I want to merge 5 commits and now over 9000.

mwkmwkmwk (approved):

mwkmwkmwk:
For future reference, please also squash your PRs into a single commit unless you have good reasons to keep them split.

(I'll squash this one on merge instead, since that lets us avoid another roundtrip and CI run)

pepijndevos:
What's the point in rebasing when you squash it anyway?

mwkmwkmwk:
I use rebase as the default for well-shaped PRs, since it directly keeps the author's internal commit structure while keeping history linear.  It's the preferred tool.

The "squash" button on github is somewhat inferior, since it by necessity destroys any *useful* commit structure there could be, plus requires the merging person (not the author) to summarize the PR in a new overriding commit message and description that doesn't go through review.

pepijndevos:
Sure I get that both serve their purposes, what I do not understand is first asking me to rebase, and then squashing it anyway. Unless I'm missing something that has the same effect as just squashing without me blundering around with git rebase for a while.

mwkmwkmwk:
Oh, I just usually assume "rebase" to include the "redact commit structure" part.

pepijndevos:
Ok

PR #2436, opened 2020-11-11
Fix begin/end in generate

dalance:
In my understanding, unconditional_block in generate is not allowed.
Instead of it, `if (1) begin` can be used to introduce a new name scope.

### DIFF
--- a/tests/simple/const_branch_finish.v
+++ b/tests/simple/const_branch_finish.v
@@ -21,9 +21,6 @@ module top;
 		end
 	end
 	generate
-		begin : unconditional_block
-			initial `CONSTANT_CHECK
-		end
 		if (WIDTH == 32) begin : conditional_block
 			initial `CONSTANT_CHECK
 		end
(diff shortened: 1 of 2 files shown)


zachjs:
I was surprised to learn that `module_or_generate_item` does not include `generate_block` in the specification. I'm not aware of any tools that do not support this non-standard behavior. However, with `-Wall` or `-Wanachronisms`, iverilog produces:
```
warning: Anachronistic use of named begin/end to surround generate schemes
```

dalance:
For example, Xilinx Vivado Simulator generates the following warning:

```
WARNING: [VRFC 10-2821] generate block is allowed only inside loop and conditional generate in SystemVerilog mode
```

Synopsys Design Compiler generates the following error:

```
Error: Syntax error on an obsolete Verilog 2001 construct standalone generate block.
```

whitequark:

zachjs:
The changes don’t necessarily reduce the coverage I intended for these test cases. I’m fine with merging this, but we should explore adding a similar warning or error to Yosys to bring it in line with other tools. If we opt for a warning, it would be prudent to add a new test case to ensure the frontend retains support for unconditional generate blocks.

PR #2438, opened 2020-11-11
synth_gowin: Add rPLL blackbox

kbeckmann:


pepijndevos (approved):
Seems to match the vendor blackbox.

PR #2442, opened 2020-11-19
Makefile: Add disabled-by-default ENABLE_SCCACHE config option.

cr1901:
[`sccache`](https://github.com/mozilla/sccache) is a `ccache`-like tool that supports distributed compilation and more languages. While it can be used as a `ccache`-replacement, I've found a few edge cases where symlinking `sccache` to `ccache` can choke. This includes on Windows where the symlink just plain doesn't work, and incompatibility with mixed C/C++ projects on all systems (`yosys` is unaffected by the latter).

Considering edge cases, along with using a completely separate cache and its extra distributed compilation capabilities, I think it's reasonable to add an `ENABLE_SCCACHE` option disabled by default. Like `ENABLE_EDITLINE` and `ENABLE_READLINE`, this option is mutually-exclusive with `ENABLE_CCACHE`; the former takes priority.

Two caveats:
* I can't speak for `ccache`, but `abc` doesn't currently play all that nice with `sccache`, even though the build succeeds. I filed a PR to fix this [upstream](https://github.com/berkeley-abc/abc/pull/108), and will get it into the YosysHQ fork once it's accepted.
* Compiler configs that requires special `CC` Makefile var handling in `ABCMKARGS` do not have `ccache`/`sccache` support yet. My feeling is that this should be fixed as part of #2011; this mostly affects Windows anyway and not the "typical" case of *nix.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,6 +31,8 @@ ENABLE_GPROF := 0
 ENABLE_DEBUG := 0
 ENABLE_NDEBUG := 0
 ENABLE_CCACHE := 0
+# sccache is not always a drop-in replacement for ccache in practice
+ENABLE_SCCACHE := 0
 LINK_CURSES := 0
 LINK_TERMCAP := 0
 LINK_ABC := 0
@@ -530,6 +532,10 @@ endif
 
 ifeq ($(ENABLE_CCACHE),1)
 CXX := ccache $(CXX)
+else
+ifeq ($(ENABLE_SCCACHE),1)
+CXX := sccache $(CXX)
+endif
 endif
 
 define add_share_file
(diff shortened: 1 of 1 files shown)


Ravenslofty:
> I can't speak for ccache, but abc doesn't currently play all that nice with sccache, even though the build succeeds. I filed a PR to fix this upstream, and will get it into the YosysHQ fork once it's accepted.

ABC needs to rebuilt rarely enough that it wasn't a problem for me in practice.

cr1901:
The immediate ABC issue is fixed [upstream](https://github.com/berkeley-abc/abc/commit/c3699a2043e94107dd963d5c49789807c135927c). I will cherry-pick it and submit to the YosysHQ fork once this is accepted. Or should I do it as part of this PR?

whitequark:
> Or should I do it as part of this PR?

Please do this as a part of this PR.

whitequark (approved):

Issue #2445, opened 2020-11-22
Would it be at all useful to document the RTLIL text format?


RobertBaruch:
I found myself idly doing this, and then wondered if I was wasting my time? Would this be a useful addition to the manual?

![rtlil](https://user-images.githubusercontent.com/1783950/99909399-25c3ab00-2c9d-11eb-968c-05b5cba64a90.JPG)

whitequark:
Sure, seems reasonable. The text fromat is already stable to an extent due to nMigen relying on it.

PR #2446, opened 2020-11-22
Adds appendix on RTLIL text format

RobertBaruch:
Issue #2445. Here's the excerpt: (updated) [rtlil.pdf](https://github.com/YosysHQ/yosys/files/5599027/rtlil.pdf)

### DIFF
--- a/manual/CHAPTER_TextRtlil.tex
+++ b/manual/CHAPTER_TextRtlil.tex
@@ -19,6 +19,8 @@ The grammar is not meant to represent semantic limitations. For example, process
 
 The grammar is also not meant to represent the exact grammar used in the RTLIL frontend, since that grammar is specific to processing by lex and yacc, and is somewhat less understandable than simple EBNF notation.
 
+Finally, note that all statements (rules ending in \texttt{-stmt}) terminate in an end-of-line. Because of this, statements cannot contain end-of-lines.
+
 \section{Lexical elements}
 
 \subsection{Identifiers}
@@ -87,34 +89,53 @@ A comment starts with a \texttt{\textbf{\#}} character and proceeds to the end o
 
 \section{File}
 
-A file consists of zero or more modules, attribute statements, and auto-index statements. All statements terminate in an end-of-line. Because of this, statements cannot contain end-of-lines.
-
-Attributes at the file level are applied to the following module.
+A file consists of an optional autoindex statement followed by zero or more modules.
 
 \begin{indentgrammar}{<design>}
-<file> ::= (<module> | <attr-stmt> | <autoidx-stmt>)$*$
+<file> ::= <autoidx-stmt>$?$ <module>*
+\end{indentgrammar}
+
+\subsection{Autoindex statements}
+
+The autoindex statement sets the global autoindex value used by Yosys when it needs to generate a unique name, e.g. \texttt{\textdollar{}flatten\textdollar{}N}. The N part is filled with the value of the global autoindex value, which is subsequently incremented. This global has to be dumped into RTLIL, otherwise e.g. dumping and running a pass would have different properties than just running a pass on a warm design.
+
+\begin{indentgrammar}{<autoidx-stmt>}
+<autoidx-stmt> ::= "autoidx" <integer> <eol>
 \end{indentgrammar}
 
 \subsection{Modules}
 
-Declares a module consisting of zero or more attributes, wires, memories, cells, processes, and connections.
+Declares a module, with zero or more attributes, consisting of zero or more wires, memories, cells, processes, and connections.
 
 \begin{indentgrammar}{<module-body-stmt>}
-<module> ::= <module-stmt> <module-body> <module-end-stmt>
+<module> ::= <attr-stmt>$*$ <module-stmt> <module-body> <module-end-stmt>
 
 <module-stmt> ::= "module" <id> <eol>
 
 <module-body> ::= 
 (<param-stmt>
-  \alt <attr-stmt>
-  \alt <wire-stmt>
-  \alt <memory-stmt>
+  \alt <wire>
+  \alt <memory>
   \alt <cell>
   \alt <process>
-  \alt <conn-stmt>)$*$
-\end{indentgrammar}
+  \alt <connection>)$*$
+
+<param-stmt> ::= "parameter" <id> <constant>$?$ <eol>
+
+<constant> ::= <value> | <integer> | <string>
 
 <module-end-stmt> ::= "end" <eol>
+\end{indentgrammar}
+
+\subsection{Attribute statements}
+
+Declares an attribute with the given identifier and value.
+
+\textbf{Warning:} There is currently a bug where integer constants are silently truncated to 32 bits and treated as unsigned.
+
+\begin{indentgrammar}{<attr-stmt>}
+<attr-stmt> ::= "attribute" <id> <constant> <eol>
+\end{indentgrammar}
 
 \subsection{Signal specifications}
 
@@ -130,47 +151,25 @@ See Sec.~\ref{sec:rtlil_sigspec} for an overview of signal specifications.
     \alt "\{" <sigspec>$*$ "\}"
 \end{indentgrammar}
 
-\subsection{Connection statements}
+\subsection{Connections}
 
-Declares a connection between the given signals.
+Declares a connection, with zero or more attributes, between the given signals.
 
-\begin{indentgrammar}{<conn-stmt>}
+\begin{indentgrammar}{<connection>}
+<connection> ::= <attr-stmt>$*$ <conn-stmt>
+
 <conn-stmt> ::= "connect" <sigspec> <sigspec> <eol>
 \end{indentgrammar}
 
-\subsection{Attribute statements}
+\subsection{Wires}
 
-Declares an attribute with the given identifier and value. Attributes at the file level apply to the following module. Attributes within a module apply to the following non-attribute statement.
-
-\begin{indentgrammar}{<attr-stmt>}
-<attr-stmt> ::= "attribute" <id> <constant> <eol>
-
-<constant> ::= <value> | <integer> | <string>
-\end{indentgrammar}
-
-\subsection{Autoindex statements}
-
-The function of this statement is currently undocumented.
-
-\begin{indentgrammar}{<autoidx-stmt>}
-<autoidx-stmt> ::= "autoidx" <integer> <eol>
-\end{indentgrammar}
-
-\subsection{Parameter statements}
-
-Declares a parameter with the given identifier for the enclosing module, optionally with the given default value.
-
-\begin{indentgrammar}{<param-stmt>}
-<param-stmt> ::= "parameter" <id> <constant>$?$ <eol>
-\end{indentgrammar}
-
-\subsection{Wire statements}
-
-Declares a wire with the given identifier in the enclosing module, with options.
+Declares a wire, with zero or more attributes, with the given identifier and options in the enclosing module.
 
 See Sec.~\ref{sec:rtlil_cell_wire} for an overview of wires.
 
 \begin{indentgrammar}{<wire-option>}
+<wire> ::= <attr-stmt>$*$ <wire-stmt>
+
 <wire-stmt> ::= "wire" <wire-option>$*$ <wire-id> <eol>
 
 <wire-id> ::= <id>
@@ -185,13 +184,15 @@ See Sec.~\ref{sec:rtlil_cell_wire} for an overview of wires.
   \alt "signed"
 \end{indentgrammar}
 
-\subsection{Memory statements}
+\subsection{Memories}
 
-Declares a memory with the given identifier in the enclosing module, with options.
+Declares a memory, with zero or more attributes, with the given identifier and options in the enclosing module.
 
 See Sec.~\ref{sec:rtlil_memory} for an overview of memory cells, and Sec.~\ref{sec:memcells} for details about memory cell types.
 
 \begin{indentgrammar}{<memory-option>}
+<memory> ::= <attr-stmt>$*$ <memory-stmt>
+
 <memory-stmt> ::= "memory" <memory-option>$*$ <id> <eol>
 
 <memory-option> ::= 
@@ -202,12 +203,12 @@ See Sec.~\ref{sec:rtlil_memory} for an overview of memory cells, and Sec.~\ref{s
 
 \subsection{Cells}
 
-Declares a cell with the given identifier in the enclosing module.
+Declares a cell, with zero or more attributes, with the given identifier and type in the enclosing module.
 
 See Chap.~\ref{chapter:celllib} for a detailed list of cell types.
 
 \begin{indentgrammar}{<cell-body-stmt>}
-<cell> ::= <cell-stmt> <cell-body-stmt> <cell-end-stmt>
+<cell> ::= <attr-stmt>$*$ <cell-stmt> <cell-body-stmt>$*$ <cell-end-stmt>
 
 <cell-stmt> ::= "cell" <cell-id> <cell-type> <eol>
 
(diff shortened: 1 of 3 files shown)


manual/CHAPTER_TextRtlil.tex
```diff
@@ -0,0 +1,252 @@
...
+
+\subsection{Values}
+
+A \textit{value} consists of a width in bits and a bit representation, most significant bit first. Bits may be any of:
+\begin{itemize}
+    \item \texttt{0}: A logic zero value
+    \item \texttt{1}: A logic one value
+    \item \texttt{x}: An unknown logic value
+    \item \texttt{z}: A high-impedance value
+    \item \texttt{m}: ?
```
whitequark, line 52: A marked bit. Should never appear in interchanged RTLIL files, but may be used internally by some passes.

manual/CHAPTER_TextRtlil.tex
```diff
@@ -0,0 +1,252 @@
...
+\end{indentgrammar}
+
+\subsection{Values}
+
+A \textit{value} consists of a width in bits and a bit representation, most significant bit first. Bits may be any of:
+\begin{itemize}
+    \item \texttt{0}: A logic zero value
+    \item \texttt{1}: A logic one value
+    \item \texttt{x}: An unknown logic value
+    \item \texttt{z}: A high-impedance value
```
whitequark, line 51: `x` and `z` may also appear in case patterns, in which case they behave like don't cares.

manual/CHAPTER_TextRtlil.tex
RobertBaruch: Updated, thanks.

manual/CHAPTER_TextRtlil.tex
RobertBaruch: Updated!

manual/CHAPTER_TextRtlil.tex
```diff
@@ -0,0 +1,252 @@
...
+
+<design> ::= <module> | <attr-stmt> | <autoidx-stmt>
+\end{indentgrammar}
+
+\subsection{Modules}
+
+A module consists of zero or more module statements.
+
+\begin{indentgrammar}{<module-stmt>}
+<module> ::= "module" <id> <eol> <module-stmt>$*$
```
tomverbeure, line 105: Missing  **end** *<eol>* ?

RobertBaruch: Updated, thanks!

tomverbeure: Just two hours ago, I finished a very rudimentary Python RTLIL parser. :-)

manual/CHAPTER_TextRtlil.tex
```diff
@@ -0,0 +1,252 @@
...
+\end{itemize}
+
+\begin{indentgrammar}{<autogen-id>}
+<id> ::= <public-id> | <autogen-id> | <dotted-id>
+
+<public-id> ::= "\textbackslash" <nonws>$+$
+
+<autogen-id> ::= "\textdollar" <nonws>$+$
+
+<dotted-id> ::= "." <decimal-digit>$+$
```
whitequark, line 41: Wait. What? Why are these a thing?? They are definitely not legal and cannot even be used:
```
ERROR: Assert `p[0] == '$' || p[0] == '\\'' failed in ./kernel/rtlil.h:165.
```

@clairexen These date back to the initial commit. What's up with them?

manual/CHAPTER_TextRtlil.tex
```diff
@@ -0,0 +1,252 @@
...
+    \item \texttt{\textbackslash r}: An 'r' character
+\end{itemize}
+
+\subsection{Comments}
+
+A comment starts with a \texttt{\textbf{\#}} character and proceeds to the end of the line. All comments are ignored.
+
+\section{File}
+
+A file consists of zero or more designs. A design may be a module, an attribute statement, or an autoindex statement.
```
whitequark, line 90: This... is not really true. The attributes are attached to the following module, so they are not standalone.

manual/CHAPTER_TextRtlil.tex
```diff
@@ -0,0 +1,252 @@
...
+
+Note that in general, statements are terminated by an end of line.
```
whitequark, line 92: I would say that all statments are terminated by EOL (so not the weasely "in general"), but some statements have bodies (and `end`) which continue on the following lines.

manual/CHAPTER_TextRtlil.tex
whitequark: Also "design statement", like "module statement", would be a lot better.

manual/CHAPTER_TextRtlil.tex
```diff
@@ -0,0 +1,252 @@
...
+
+<design> ::= <module> | <attr-stmt> | <autoidx-stmt>
+\end{indentgrammar}
+
+\subsection{Modules}
+
+A module consists of zero or more module statements.
+
+\begin{indentgrammar}{<module-stmt>}
+<module> ::= "module" <id> <eol> <module-stmt>$*$ "end" <eol>
```
whitequark, line 105: I said it before, but attributes really ought to be attached to the module here.

manual/CHAPTER_TextRtlil.tex
RobertBaruch: Done. I cleaned things up a bit.

manual/CHAPTER_TextRtlil.tex
RobertBaruch: Done. I've fiddled with the grammar so that rules ending in "-stmt" all terminate consistently in eol.

manual/CHAPTER_TextRtlil.tex
RobertBaruch: I can make that more explicit using the grammar, although those auto-index statements, whose function I don't know, appear to be able to be interspersed with module attributes and modules. In any case, I've also elaborated a bit on what attributes apply to in the attributes section.

RobertBaruch: For example, a possible refactor that might make this clearer is:

`<file> ::= ((<autoidx-stmt> | <attr-stmt>)* <module>)*`

whitequark: > those auto-index statements, whose function I don't know

Practically, an `autoidx` statement should always appear only at the very start of an RTLIL file (if at all), and only once. This isn't enforced but perhaps should be.

The function of them is simple. Sometimes Yosys needs to generate an unique name. It picks e.g. `$flatten$N`. The `N` part is the value of the `autoidx` global, which is subsequently incremented. This global has to be dumped into RTLIL, otherwise e.g. dumping and running a pass would have different properties than just running a pass on a warm design.

RobertBaruch: Ah, ok, I can document that. I could also again refactor the grammar so that autoidx only appears at the beginning of the file?

```
<file> ::= <autoidx-stmt>? <module>*
<module> ::= <module-attr-stmt>* <module-stmt> <module-body> <module-end-stmt>
<module-attr-stmt> := <attr-stmt>
```

whitequark: Now you got it! Although I'd probably use something like...
```
<file> ::= <autoidx-stmt>? <module>*
<module> ::= <attr-stmt>* <module-stmt> <module-body> <module-end-stmt>
```
since attributes are all the same in RTLIL no matter what they are attached to.

manual/CHAPTER_TextRtlil.tex
```diff
@@ -0,0 +1,270 @@
...
+\end{indentgrammar}
+
+\subsection{Attribute statements}
+
+Declares an attribute with the given identifier and value. Attributes at the file level apply to the following module. Attributes within a module apply to the following non-attribute statement.
+
+\begin{indentgrammar}{<attr-stmt>}
+<attr-stmt> ::= "attribute" <id> <constant> <eol>
+
+<constant> ::= <value> | <integer> | <string>
```
whitequark, line 148: It is worth noting that plain integer constants have a really weird, unintuitive behavior: whatever integer you have in the text is silently truncated to 32 bits and converted to unsigned. I had a PR to improve this (https://github.com/YosysHQ/yosys/pull/2066) but it got rejected in favor of a different approach... one we have yet to determine.

whitequark (commented):
I finally managed to review this PR top to bottom! Seems good overall; apart from what we've already discussed, I think the only thing I'd change is the way `<attr-stmt>` is represented for non-modules. (Same issue there.)

manual/CHAPTER_TextRtlil.tex
RobertBaruch: Done!

manual/CHAPTER_TextRtlil.tex
RobertBaruch: Noted in document.

RobertBaruch:
> I finally managed to review this PR top to bottom! Seems good overall; apart from what we've already discussed, I think the only thing I'd change is the way `<attr-stmt>` is represented for non-modules. (Same issue there.)

Refactored.

manual/CHAPTER_TextRtlil.tex
```diff
@@ -0,0 +1,270 @@
...
+
+<module-end-stmt> ::= "end" <eol>
+\end{indentgrammar}
+
+\subsection{Attribute statements}
+
+Declares an attribute with the given identifier and value.
+
+\textbf{Warning:} There is currently a bug where integer constants are silently truncated to 32 bits and treated as unsigned.
```
whitequark, line 134: This isn't a bug, it's intended behavior (I asked). It's just obnoxious and IMO must be changed.

manual/CHAPTER_TextRtlil.tex
```diff
@@ -0,0 +1,270 @@
...
+<sigspec> ::=
+<constant>
+    \alt <wire-id>
+    \alt <sigspec> "[" <integer> (":" <integer>)$?$ "]"
+    \alt "\{" <sigspec>$*$ "\}"
+\end{indentgrammar}
+
+\subsection{Connections}
+
+Declares a connection, with zero or more attributes, between the given signals.
```
whitequark, line 156: Connections don't have attributes (it is an error to try and attach them, try it).

manual/CHAPTER_TextRtlil.tex
```diff
@@ -0,0 +1,270 @@
...
+<cell-body-stmt> ::= 
+"parameter" ("signed" | "real")$?$ <id> <constant> <eol>
+  \alt "connect" <id> <sigspec> <eol>
+
+<cell-end-stmt> ::= "end" <eol>
+\end{indentgrammar}
+
+\subsection{Processes}
+
+Declares a process, with zero or more attributes, with the given identifier in the enclosing module.
```
whitequark, line 228: I think I would describe processes and cases a bit separately. Cases only exist in processes, yes, but it's a complex recursive structure that has attributes (which currently aren't mentioned). On the other hand, when you work with RTLIL, you tend to treat cases and processes/syncs a bit separately. Basically, the case in the process works like an `always_comb` on its own right, and then syncs are their own thing. The only time when `sync always` makes a difference vs direct assignment is for latch inference.

manual/CHAPTER_TextRtlil.tex
clairexen: ... they even date back to the initial commit in the original suberversion repo. I have no memory whatsoever of what that is.  :D

whitequark: Let's get rid of them.

manual/CHAPTER_TextRtlil.tex
RobertBaruch: Oof. Just to be clear: you can still have negative constants, and they will remain negative, unless the constant is greater than 32 bits?

whitequark: No, it's much weirder than that. For bit vectors (like `64'00000000...0000`) there is no problem, they get treated as bit vectors. (Incidentally you can't specify sign.) The problems only arise for bare integers (like `12345`) that are a part of a sigspec.

Basically the way the RTLIL parser works is that it reads a signed number into a C `long` in an ordinary way. After that, if this number is used in a sigspec, it is ANDed with 0xFFFFFFFF and turned into an RTLIL constant.

Besides just being inconsistent (check the examples I have in #2066), if you have an unsigned constant that's 0x80000000 or higher and you serialize it as a bare integer, then on platforms with 32-bit `long` (so basically on 32-bit platforms) this RTLIL file will fail to parse*. But it will parse just fine on 64-bit platforms. So any emitter of RTLIL files has to be careful to not produce files that will be readable on some platforms but not others--which is frankly preposterous.

\* It actually used to invoke UB and silently produce incorrect netlist until I made it not do that in commit ca70a1049ff18c8702fdc2ac4b3901685551cb28...

manual/CHAPTER_TextRtlil.tex
RobertBaruch: Do syncs have attributes?

manual/CHAPTER_TextRtlil.tex
whitequark: Actually, it seems like I misremembered. In fact, in commit ca70a1049ff18c8702fdc2ac4b3901685551cb28 I restricted the range of bare integer constants to the domain of 32-bit signed numbers, regardless of the platform.

So I think you can just write it in the manual that bare integers in sigspecs have a range of [-2147483648, 2147483648) and anything outside of that range must be emitted as a bit vector. Also that the sign of the bare integer (when it is used in a sigspec) is completely ignored, and the resulting constant is always 32-bit.

manual/CHAPTER_TextRtlil.tex
whitequark: https://github.com/YosysHQ/yosys/pull/2452

manual/CHAPTER_TextRtlil.tex
RobertBaruch: Done.

manual/CHAPTER_TextRtlil.tex
RobertBaruch: Done!

manual/CHAPTER_TextRtlil.tex
RobertBaruch: I've removed this from the documentation, and filed $2454 to remove officially.

manual/CHAPTER_TextRtlil.tex
RobertBaruch: Thanks for the clarification. Updated with explanations in both the `<integer>` section and the `<sigspec>` section.

RobertBaruch:
This is now ready for final review.

manual/CHAPTER_TextRtlil.tex
```diff
@@ -0,0 +1,299 @@
...
+
+The grammar is also not meant to represent the exact grammar used in the RTLIL frontend, since that grammar is specific to processing by lex and yacc, is even more permissive, and is somewhat less understandable than simple EBNF notation.
+
+Finally, note that all statements (rules ending in \texttt{-stmt}) terminate in an end-of-line. Because of this, a statement cannot be broken into multiple lines.
+
+\section{Lexical elements}
+
+\subsection{Characters}
+
+The characters accepted in an RTLIL file are those encodable in 8 bits. Unicode is not supported. For maximum safety, limit characters to the 7-bit ASCII range $[0,127]$.
```
whitequark, line 28: Wait, is there anything in Yosys that actually cares about 8-bit characters? The low block (0x20 and below) are forbidden in identifiers, but I don't think the high half is.

daveshah1: Anecdotally, I haven't had a problem with UTF-8 (provided you use an escaped identifier in the case of Verilog, anyway)

RobertBaruch: Hmm, I should probably change this. I originally meant to say that you can't use Unicode, and if you do use a multi-char codepoint, it will just look like a bunch of 8-bit characters. Let me fix this.

RobertBaruch: Ah, I found the assert in rtlil.h that prohibits anything 0x20 and below from being in identifiers.

RobertBaruch: Done. I've clarified the section on characters.

RobertBaruch:
Again, ready for final review.

whitequark (approved):
LGTM, great work!

PR #2448, opened 2020-11-24
Tcl script documentation fixes

nitz:
Found myself chasing my tail a bit trying to figure out how to run scripts. Figured I'd update the `-h` output and the manual so people after me chase their tails less! 🙂

### DIFF
--- a/kernel/driver.cc
+++ b/kernel/driver.cc
@@ -267,9 +267,11 @@ int main(int argc, char **argv)
 		printf("\n");
 		printf("    -s scriptfile\n");
 		printf("        execute the commands in the script file\n");
+#ifdef YOSYS_ENABLE_TCL
 		printf("\n");
 		printf("    -c tcl_scriptfile\n");
 		printf("        execute the commands in the tcl script file (see 'help tcl' for details)\n");
+#endif
 		printf("\n");
 		printf("    -p command\n");
 		printf("        execute the commands\n");


manual/command-reference-manual.tex
```diff
@@ -6141,21 +6141,27 @@ \section{synth\_xilinx -- synthesis for Xilinx FPGAs}
         write_blif 
 \end{lstlisting}
 
-\section{tcl -- execute a TCL script file}
```
whitequark, line 6144: This file is autogenerated from the help text in the C++ file, so your changes need to be done there.

kernel/driver.cc
```diff
@@ -268,8 +268,8 @@ int main(int argc, char **argv)
 		printf("    -s scriptfile\n");
 		printf("        execute the commands in the script file\n");
 		printf("\n");
-		printf("    -c tcl_scriptfile\n");
-		printf("        execute the commands in the tcl script file (see 'help tcl' for details)\n");
+		printf("    -c script\n");
+		printf("        execute the commands in the tcl script file (see 'help script' for details)\n");
```
whitequark, line 272: Actually, this is just completely incorrect. The Yosys `script` command does not use Tcl for scripting.

nitz: At least now I feel like my confusion was legitimate, haha!

When I ran 'yosys -h', the parameter help was the one there that came out of 'driver.cc'. But when I ran 'yosys -h tcl_scriptfile' like it suggested, that wasn't even a command. The only thing I saw when running 'help' directly, was 'script', which definitely was what I wanted.

Have I just massively got my wires crossed here?

whitequark: Where does it suggest to run `yosys -h tcl_scriptfile`? I can't find it.

nitz: Okay, let me list my numerous mistakes:

I was under the impression that the tcl scripting *was* yosys scripting. I see now that they are two separate things. That clears up a lot of my misconceptions.

Second: my apologies, it was actually suggesting to run 'yosys -h tcl'. I got "no command or cell type tcl" when I ran that.

I've now got to the bottom of *that*: I'm using a nightly build from open tool forge, that is apparently built *without* tcl support, which explains why I never could find the command -- I simply didn't have it 😬

In short: I'm sorry to have wasted your time with my misunderstanding(s)! Maybe all I really should have submitted as a PR is perhaps a conditional compile out for the -c help line if tcl support isn't enabled.

whitequark: > Maybe all I really should have submitted as a PR is perhaps a conditional compile out for the -c help line if tcl support isn't enabled.

Sounds good to me!

whitequark (approved):

nitz:
Thanks for going back and forth with me — apologies again for my massive misunderstanding, very much appreciate you nudging me in the right direction!

whitequark:
No problem.

PR #2449, opened 2020-11-24
Fix some typos in simlib.v comments that confused me for a second

tomverbeure:


mwkmwkmwk:
The changes are reasonable, but please squash them into a single commit — it's currently a strange mess.

tomverbeure:
Ok. I'll create a new pull request for that. It has something to do with the master of my clone not being a direct pull of upstream.

PR #2453, opened 2020-11-25
Generate only simple assignments in verilog backend

mmicko:
Yosys currently emits assigns with a concatenated expression on the LHS in form like:
assign {a,b} = {c,d,e};
this change generate instead (note that size is taken into consideration):
assign a = {c, d};
assign b = e;

This fixes issue #1286

### DIFF
--- a/backends/verilog/verilog_backend.cc
+++ b/backends/verilog/verilog_backend.cc
@@ -35,7 +35,7 @@
 USING_YOSYS_NAMESPACE
 PRIVATE_NAMESPACE_BEGIN
 
-bool verbose, norename, noattr, attr2comment, noexpr, nodec, nohex, nostr, extmem, defparam, decimal, siminit, systemverilog;
+bool verbose, norename, noattr, attr2comment, noexpr, nodec, nohex, nostr, extmem, defparam, decimal, siminit, systemverilog, simple_lhs;
 int auto_name_counter, auto_name_offset, auto_name_digits, extmem_counter;
 std::map<RTLIL::IdString, int> auto_name_map;
 std::set<RTLIL::IdString> reg_wires;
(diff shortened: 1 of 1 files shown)


daveshah1:
Seems reasonable, would be good to double check that it works correctly when dealing with 'upto' wires or non-zero start offsets

whitequark:
Changing this by default would negatively impact nMigen because it relies on Yosys-produced Verilog being relatively human-readable. Adding an option to enable this behavior seems fine though.

whitequark (changes requested):
Please keep the old behavior and make the new behavior opt-in.

mmicko:
> Please keep the old behavior and make the new behavior opt-in.

@whitequark sure, any idea for good option name ? 

> Seems reasonable, would be good to double check that it works correctly when dealing with 'upto' wires or non-zero start offsets
Good idea, will make some more test cases.

mmicko:
@daveshah1 Seams good to me with next example :

module a();
    wire [1:2] b;
    wire [3:1] c;

    assign {b, f} = c;
endmodule


/* Generated by Yosys 0.9+3696 (git sha1 cf67e6a3, clang 11.0.0 -fPIC -Os) /

( cells_not_processed =  1  )
( src = "a.v:1.1-6.10" )
module a();
  ( src = "a.v:2.13-2.14" )
  wire [1:2] b;
  ( src = "a.v:3.13-3.14" )
  wire [3:1] c;
  ( src = "a.v:5.13-5.14" *)
  wire f;
  assign f = c[1];
  assign b = c[3:2];
endmodule

whitequark:
> any idea for good option name ?

`-simple-lhs`?

whitequark (approved):
Thanks!

daveshah1:
One last comment, could you add a simple test for this to CI?

mmicko:
@daveshah1 for consistency sake I will add test in yosys-tests since we need to check generated verilog at the end

mmicko:
@daveshah1 added https://github.com/YosysHQ/yosys-tests/commit/d323bec22d076f673e97800b5267dd0d5403f715 (note I also did additional commit to fix eol :) )

Issue #2035, opened 2020-05-06
Illegal Verilog output after 'proc_prune'; assigns 'reg' from a constant


strobo5:
## Steps to reproduce the issue

The following procedure generates illegal Verilog (at least XST fails and quick research tells me it cannot be blamed).
`assign a_pc[1:0] = 2'h0;` (outside an always block)
When the ilang code is changed so that the default case is named explicitly (`case` -> `case 1'1`) then the Verilog code is as expected.
```
$ yosys -p "read_ilang bug.il; proc_prune; write_verilog bug.v"
[..]
2. Executing PROC_PRUNE pass (remove redundant assignments in processes).
Removed 0 redundant assignments.
Promoted 1 assignment to connection.
[..]
```
```
module\top
  wire width 1 input 0 \sel
  wire width 8 output 1 \a_pc
  process $group_1
    assign \a_pc 8'00000000
    switch \sel
      case 1'0
        assign \a_pc [7:2] 6'101010
      case
        assign \a_pc [7:2] 6'111111
    end
    sync init
  end
end
```

## Expected behavior

```
/* Generated by Yosys 0.9+2406 (git sha1 584780d7, gcc 9.2.0-r2 -O2 -march=bdver2 -mmmx -msse -msse2 -msse3 -mssse3 -msse4a -mcx16 -msahf -maes -mpclmul -mpopcnt -mabm -mlwp -mfma -mfma4 -mxop -mbmi -mtbm -mavx -msse4.2 -msse4.1 -mlzcnt -mf16c -mprfchw -mfxsr -mxsave -mtune=bdver2 -fPIC -Os) */

module top(a_pc, sel);
  output [7:0] a_pc;
  reg [7:0] a_pc;
  input sel;
  always @* begin
    a_pc = 8'h00;
    casez (sel)
      1'h0:
          a_pc[7:2] = 6'h2a;
      1'h1:
          a_pc[7:2] = 6'h3f;
    endcase
  end
  initial begin
  end
endmodule
```

## Actual behavior

```
[...]

PR #2458, opened 2020-11-28
proc_prune: do not promote partially redundant assignments.

jfng:
According to https://github.com/YosysHQ/yosys/issues/2035#issuecomment-634440650, the bug in #2035 requires intrusive changes to write_verilog to be properly fixed.

In the meantime, this PR proposes a workaround to proc_prune to avoid emitting the incriminated corner-case.

The source of the bug would still need to be fixed, but e.g. nMigen users will be able to use proc -nomux without hitting it.

### DIFF
--- a/passes/proc/proc_prune.cc
+++ b/passes/proc/proc_prune.cc
@@ -65,6 +65,7 @@ struct PruneWorker
 			pool<RTLIL::SigBit> sw_assigned = do_switch((*it), assigned, affected);
 			assigned.insert(sw_assigned.begin(), sw_assigned.end());
 		}
+		std::vector<RTLIL::SigSig> new_actions;
 		for (auto it = cs->actions.rbegin(); it != cs->actions.rend(); ) {
 			RTLIL::SigSpec lhs = sigmap(it->first);
 			bool redundant = true;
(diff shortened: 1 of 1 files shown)


whitequark:
Thanks for spearheading this! I don't like the idea of this change very much (and I assume neither do you); the coupling of `proc` passes with `write_verilog` really shouldn't be a thing. Furthermore, historically `proc_prune` has been a source of pretty obscure bugs, so I will be cautious when merging this.

My plan is to review this PR carefully and at the same time decide on the feasibility of a more tractable workaround in `write_verilog` (I have some ideas). Either way this issue will be solved.

whitequark:
@jfng By any chance, did https://github.com/YosysHQ/yosys/pull/2964 improve this?

mwkmwkmwk:
> @jfng By any chance, did #2964 improve this?

Nope, no change in behavior here.  This can only be fixed properly on the Verilog backend side, I'm afraid.  I have some vague hack ideas for `proc_prune`, but they all seem harder than just fixing the backend by adding an intermediate wire when different bits of an RTLIL wire have differing wire/reg requirements.

widlarizer:
Correct me if I'm wrong but it seems to be superseded by #4314

whitequark:
Yes, I think that's right.

PR #2459, opened 2020-11-29
[WIP] New $print cell for representing print statements

whitequark:
This PR introduces a new cell, `$print`, which can losslessly represent the behavior of Verilog formatting tasks like `$display` in `always` blocks, both clocked and continuous. It adds support for parsing Verilog tasks `$display` (as well as `b`, `o`, `h` suffix variants) and `$write` (as well as `b`, `o`, `h` suffix variants) such that they are translated to `$print` cells, and emitting `$print` cells as appropriate `always` blocks containing a `$write` task.

As a side effect, this PR greatly improves conformance with the Verilog LRM for the `$display` and `$write` family tasks used in `initial` blocks, as well as the `$sformatf` function. Unfortunately, the latter improvement is likely a breaking change.

Internally, this PR adds an abstract representation of format strings with substitutions that supports RTLIL and Veriog (de)serialization. It specifically does *not* use Verilog format strings anywhere but on the boundary. This is necessary because IEEE 1800-2017 does not precisely define the semantics of almost every aspect of its formatting facility, and real-world implementations wildly disagree with each other on even the grammar they accept.

The format string representation added by this PR is a new one I designed specifically for this task. It heavily borrows from Python f-strings (in particular, the method of escaping the substitutions is exactly identical) and is designed for simplicity of programmatic use and human readability, in that order.

<hr>

Remaining work:
 - [x] Add tests (diff against iverilog) for the `Fmt::render` facility used by Yosys to print messages when it encounters `initial $display` statements and for `$sformatf`.
 - [x] Add tests (roundtrip identity check) for the Verilog/RTLIL translation facilities used to support the `$print` cell.
 - [ ] Document the `$print` cell and its format string mini-language.
 - [ ] Implement `$print` cell handling in CXXRTL.
 - [ ] Add back `%m` support that existed in the old `$display` implementation (as an alias of `%l`) but was removed because it is unclear whether this implementation is correct.
 - [ ] Add `$time` and `%t` support to format strings.


whitequark:
Likely superseded by https://github.com/YosysHQ/yosys/pull/3721. (I'll keep this open for now.)

whitequark:
Obsoleted by https://github.com/YosysHQ/yosys/pull/3721.

PR #2460, opened 2020-11-30
add -noalu and -json option for apicula

pepijndevos:
At the moment I'm working on a Nextpnr target for Gowin, but Apicula does not yet support some of the primitives supported by the yosys backend.

This PR adds a `-noalu` option that prevents ALUs from being used, and also adds a `-json` option that will set all the right parameters for use with Apicula, similar to the `-vout` option for the vendor toolchain.

Semi-random question: any idea why there is a `-nodffe` option? Enable is basically free and even Apicula already supports it, but maybe it's more timing related?

### DIFF
--- a/techlibs/gowin/synth_gowin.cc
+++ b/techlibs/gowin/synth_gowin.cc
@@ -44,6 +44,11 @@ struct SynthGowinPass : public ScriptPass
 		log("        write the design to the specified Verilog netlist file. writing of an\n");
 		log("        output file is omitted if this parameter is not specified.\n");
 		log("\n");
+		log("    -json <file>\n");
+		log("        write the design to the specified JSON netlist file. writing of an\n");
+		log("        output file is omitted if this parameter is not specified.\n");
+		log("        This disables features not yet supported by nexpnr-gowin.\n");
+		log("\n");
 		log("    -run <from_label>:<to_label>\n");
 		log("        only run the commands between the labels (see below). an empty\n");
 		log("        from label is synonymous to 'begin', and empty to label is\n");
@@ -70,6 +75,9 @@ struct SynthGowinPass : public ScriptPass
 		log("    -noiopads\n");
 		log("        do not emit IOB at top level ports\n");
 		log("\n");
+		log("    -noalu\n");
+		log("        do not use ALU cells\n");
+		log("\n");
 		log("    -abc9\n");
 		log("        use new ABC9 flow (EXPERIMENTAL)\n");
 		log("\n");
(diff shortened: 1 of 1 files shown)


daveshah1:
The main reason for disabling DFFEs is that differing control sets can sometimes make placement harder. Its a bit of a blunt instrument though, something like ice40's dffe_min_ce_se is finer grained.

techlibs/gowin/synth_gowin.cc
```diff
@@ -112,6 +122,14 @@ struct SynthGowinPass : public ScriptPass
 				vout_file = args[++argidx];
 				continue;
 			}
+			if (args[argidx] == "-json" && argidx+1 < args.size()) {
+				json_file = args[++argidx];
+				nobram = true;
+				nolutram = true;
+				nowidelut = true;
+				noalu = true;
```
mmicko, line 130: Better do not add these 4 lines. That way yosys part should not be changed while you are improving nextpnr support. Yes command line will be longer, but you can explicitly disable/enable certain flags then.

pepijndevos: Hmm, I would like `synth_gowin` to do the right thing as much as possible without people having to know all the internal details of supported Gowin primitives. But on the other hand... yea it's be a new PR every time I support more primitives.

daveshah1: With ECP5 I added support for all primitives in Yosys by default before they were supported in nextpnr. In retrospect, this had the downside that people were including the Yosys args to disable those primitives in their own code and not removing them when those primitives were supported in nextpnr so getting worse results than needed. So I don't know which the best way is, tbh.

pepijndevos: Yea, I would be somewhat in favour of a plain `synth_gowin` to follow "current best practice" as much as possible. It might be a long time between `nextpnr-gowin` becoming usable and it supporting all the primitives. Removing those lines later should be a no-brainer.

mmicko: I agree with Daves point, but with flags as they are now, you can not enable alu for example when outputting json. From user perspective current code is better, from developer perspective you would need to recompile yosys when you add new thing in nextpnr, and basically your nextpnr version will work only with latest yosys in that case. So yeah hard decision.

pepijndevos: If you just want a JSON with all the options you can just not pass the `-json` option and issue a simple `write_json` command. Maybe the `-json` option can get a name that indicates it's specifically for Nextpnr/Apicula.

pepijndevos: It seems rare that you would have a newer Yosys than Nexpnr. The other way around would just have slightly worse results.

mmicko: Sure, there is always workaround. Anyway agree that current code is more user prone to errors.

pepijndevos: So what do you think we should do?
* merge as is
* remove those lines
* remove entire `-json` option (it's kinda useless without these lines)
* rename `-json` to... `-apicula` or something to clarify it does something special

mmicko: Think most reasonable option for now is to merge code as is. If someone have older yosys and newer nextpnr it will not be a problems but will just get not that good result.

mmicko (approved):

PR #2461, opened 2020-11-30
intel_alm: preliminary Arria V support

Ravenslofty:
The non-GZ Arria V parts are essentially just larger Cyclone Vs.

This PR simply adds Arria V timings and extends the `-family` option to accept "arriav".

cc @mwkmwkmwk

### DIFF
--- a/techlibs/intel_alm/common/quartus_rename.v
+++ b/techlibs/intel_alm/common/quartus_rename.v
@@ -2,14 +2,25 @@
 `define LCELL cyclonev_lcell_comb
 `define MAC cyclonev_mac
 `define MLAB cyclonev_mlab_cell
+`define RAM_BLOCK cyclonev_ram_block
 `define IBUF cyclonev_io_ibuf
 `define OBUF cyclonev_io_obuf
 `define CLKENA cyclonev_clkena
+`endif
+`ifdef arriav
+`define LCELL arriav_lcell_comb
+`define MAC arriav_mac
+`define MLAB arriav_mlab_cell
+`define RAM_BLOCK arriav_ram_block
+`define IBUF arriav_io_ibuf
+`define OBUF arriav_io_obuf
+`define CLKENA arriav_clkena
 `endif
 `ifdef cyclone10gx
 `define LCELL cyclone10gx_lcell_comb
 `define MAC cyclone10gx_mac
 `define MLAB cyclone10gx_mlab_cell
+`define RAM_BLOCK cyclone10gx_ram_block
 `define IBUF cyclone10gx_io_ibuf
 `define OBUF cyclone10gx_io_obuf
 `define CLKENA cyclone10gx_clkena
@@ -146,7 +157,7 @@ output [CFG_DBITS-1:0] B1DATA;
 // Much like the MLAB, the M10K has mem_init[01234] parameters which would let
 // you initialise the RAM cell via hex literals. If they were implemented.
 
-cyclonev_ram_block #(
+`RAM_BLOCK #(
     .operation_mode("dual_port"),
     .logical_ram_name(_TECHMAP_CELLNAME_),
     .port_a_address_width(CFG_ABITS),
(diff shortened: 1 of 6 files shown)


techlibs/intel_alm/common/quartus_rename.v
```diff
@@ -2,11 +2,19 @@
 `define LCELL cyclonev_lcell_comb
 `define MAC cyclonev_mac
 `define MLAB cyclonev_mlab_cell
+`define RAM_BLOCK cyclonev_ram_block
```
mwkmwkmwk, line 5: call it M10K_BLOCK perhaps?

(there are Intel FPGAs with several different sizes of memory blocks, could be good to establish the convention here to specify size in case we end up adding support for them)

techlibs/intel_alm/synth_intel_alm.cc
```diff
@@ -43,6 +43,7 @@ struct SynthIntelALMPass : public ScriptPass {
 		log("    -family <family>\n");
 		log("        target one of:\n");
 		log("        \"cyclonev\"    - Cyclone V (default)\n");
+		log("        \"arriav\"      - Arria V (non-GZ)");
```
mwkmwkmwk, line 46: I wonder if it would be a good idea to add `arriavgx` as an alias (to avoid confusion why there eventually is `arriav` and `arriavgz`)

Ravenslofty: I think that could be too specific. Remember all the other models of non-GZ Arria V; `arriavgx` could give the impression that only that specific family is supported. On the other hand, if someone with a GZ selects `arriav` instead of either `arriavgz` or (more likely) `stratixv`, the design will still be implementable on a GZ.

techlibs/intel_alm/common/quartus_rename.v
Ravenslofty: I fully intend to add native instantiation for Cyclone 10 GX, which uses M20K blocks instead. The problem then is what to define M10K_BLOCK on a chip that doesn't have it, and vice versa.

I think there *might* have been a family that mixed M144Ks with M20Ks, but I can't remember it. If not, then I think it's reasonably safe to use RAM_BLOCK to refer to BRAM.

techlibs/intel_alm/synth_intel_alm.cc
mwkmwkmwk: ... right, I confused it with some other family

what about adding aliases `arriavgx`, `arriavgt`, `arriavsx`, `arriavst`? there aren't that many of them

also, arria V GX design will not be implementable on arria V GZ due to GX using M10K and GZ using M20K...

techlibs/intel_alm/common/quartus_rename.v
mwkmwkmwk: Just don't define it and disable the substitution or have it `$error`

Converting a wrong type of primitive is a silent failure of the kind that we would really like to avoid

mwkmwkmwk: According to my notes, the only mixed-size families that involve ALMs are stratix iii and stratix iv (aka arria ii), both of them involving M9K+M144K; still, this does sound like something we could support one day

Ravenslofty: > Just don't define it and disable the substitution

I'm pretty sure an unexpanded substitution is a syntax error which would fail whenever techmap was called.

`$error` is probably okay though.

mwkmwkmwk: It's actually a "too hard" error — it will fail reading the whole techmap file whether the invalid substitution is in fact used or not, so it needs another `ifdef

techlibs/intel_alm/synth_intel_alm.cc
Ravenslofty: This would be too unwieldy to explain in the documentation, I think. If it can't be explained well enough, the feature might as well not exist.

PR #2463, opened 2020-12-01
Fix SYNTHESIS always being defined in Verilog frontend

georgerennie:
This patch fixes a bug where the verilog macro SYNTHESIS is always defined when using read_verilog even when the -formal option is used.

The bug occured as the define_map_t constructor in frontends/verilog/preproc.cc set one of the two macros, dependent on the formal_mode global variable. In VerilogFrontend::execute() in frontends/verilog/verilog_frontend.cc the object defines_map is created before the options to read_verilog have been parsed. This means that formal_mode has not been set correctly yet, so SYNTHESIS is always defined.

This is fixed by defining the verilog macros in execute(), after the read_verilog options have been parsed, instead of in the define_map_t constructor.

### DIFF
--- a/frontends/verilog/preproc.cc
+++ b/frontends/verilog/preproc.cc
@@ -321,7 +321,6 @@ struct define_body_t
 define_map_t::define_map_t()
 {
 	add("YOSYS", "1");
-	add(formal_mode ? "FORMAL" : "SYNTHESIS", "1");
 }
 
 // We must define this destructor here (rather than relying on the default), because we need to
--- a/frontends/verilog/verilog_frontend.cc
+++ b/frontends/verilog/verilog_frontend.cc
@@ -446,6 +446,9 @@ struct VerilogFrontend : public Frontend {
 			}
 			break;
 		}
+
+		defines_map.add(formal_mode ? "FORMAL" : "SYNTHESIS", "1");
+
 		extra_args(f, filename, args, argidx);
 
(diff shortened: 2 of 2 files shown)


udif:
Why would you not want SYNTHESIS to be defined for formal runs?
Typically SYNTHESIS is used for example to skip verification-only code for synthesis. Since your formal runs wants to look at the same code you synthesize, I would expect SYNTHESIS to be defined in formal runs as well.

georgerennie:
> Why would you not want SYNTHESIS to be defined for formal runs?
> Typically SYNTHESIS is used for example to skip verification-only code for synthesis. Since your formal runs wants to look at the same code you synthesize, I would expect SYNTHESIS to be defined in formal runs as well.

The documentation states that FORMAL should be defined for formal runs instead of SYNTHESIS http://www.clifford.at/yosys/cmd_read_verilog.html
```
-formal
        enable support for SystemVerilog assertions and some Yosys extensions
        replace the implicit -D SYNTHESIS with -D FORMAL
```
This was the case before https://github.com/YosysHQ/yosys/commit/044ca9dde409e3c91542fe95513d6641110f8462

The case I came up against where I wanted SYNTHESIS not to be defined in formal runs was where I am using a device specific primitive in synthesis which cannot be used in formal mode. As I had been wrapping that primitive in `ifdef SYNTHESIS, it was still being included in formal runs when it shouldnt have been. If you want to check if you are in either synthesis or formal mode within yosys at present you can use `ifdef YOSYS which is always defined.

PR #2471, opened 2020-12-03
ice40: simulate SPRAM poweron delay in cells_sim

bobtwinkles:
Makes the SB_SPRAM simulation model slightly more accurate, including a coarse simulation of how the block behaves when modifying the various power/sleep state control lines.

I'd also like to write up some tests for this, but the actual modifications to the cell simulation should be ready for review.


smunaut:
Mmm, I don't think that logic is correct. If you release power poweroff and then release sleep slightly after, you wouldn't see the 41 ns delay, because state would go POWEROFF -> SLEEP -> NORMAL.

bobtwinkles:
I've reworked this to use a handful of counters instead of a state machine, which was getting extremely unwieldy when trying to account for the various ways the power control lines can interact.

widlarizer:
While the functionality may be correct and there may be cases where users would prefer those, the simulation models within Yosys should be readable by Yosys, delays aren't synthesizable Verilog, which is what Yosys aims to support

PR #2476, opened 2020-12-06
Fix constants bound to single bit arguments (fixes #2383)

### LINKED ISSUE #2383: Task argument inherits width of value that was passed to it

Run this code through yosys:
```
module test(input wire [7:0] a, output reg [8:0] b);
	task t(input wire [7:0] c, input wire d);
		b = { c, d };
	endtask
	always @* t(a, 0);
endmodule
```

## Expected behavior

I would expect the output port b[0] to be 0; and b[8:1] to equal a[7:0].

## Actual behavior

Since commit 59c4ad8 all nine bits of output port b are zero.
I suspect that the constant zero that gets passed to the second argument of the task is 32 bits wide and gets treated as 32 bit inside the task also.

## Workaround

Change `input wire d` to `input wire [0:0] d`; or change the zero constant to `1'b0`.

zachjs:


frontends/ast/simplify.cc
```diff
@@ -3337,6 +3337,15 @@ skip_dynamic_range_lvalue_expansion:;
 						wire->type = AST_LOCALPARAM;
 						wire->attributes.erase(ID::nosync);
 						wire->children.insert(wire->children.begin(), arg->clone());
+						// args without a range implicitly have width 1
+						if (wire->children.back()->type != AST_RANGE) {
+							AstNode* range = new AstNode();
+							range->type = AST_RANGE;
+							wire->children.push_back(range);
+							std::vector<RTLIL::State> data{{}};
```
whitequark, line 3345: What does this vector initialize to, exactly? A single `S0` element?

zachjs: I'm using this to create constants of value `1'b0` for use in creating an implicit range such as `[0:0]`. Is there a cleaner way to accomplish this?

whitequark: `RTLIL::Const(0)`?

whitequark (approved):

mmicko:
@zachjs Sorry for late update, but this seams to break two test cases:
https://github.com/YosysHQ/yosys-tests/tree/master/regression/issue_00102
https://github.com/YosysHQ/yosys-tests/tree/master/regression/issue_00781

Both now give back error : ERROR: Incompatible re-declaration of wire 

If I recall good, we had these issues reappearing before as well by touching this part of code.

PR #2479, opened 2020-12-07
Allow constant function calls in constant function arguments

zachjs:


whitequark (commented):
Can you explain how this change works? It is not clear to me from the context what actually happnens here.

zachjs:
`has_const_only_constructs` is used later on in `simplify` to detect whether an `AST_FCALL` needs to be evaluated during elaboration. However, the arguments of any `AST_FCALL` are `simplify`-ed earlier on, at the site where I made my changes. This earlier simplification previously turned the function arguments into wires. With this change, if the `AST_FCALL` needs to be evaluated during elaboration, then the arguments themselves are evaluated as constants, rather than being turned into wires.

whitequark (approved):
LGTM

whitequark:
Thanks!

PR #2480, opened 2020-12-08
nexus: Add LRAM inference

daveshah1:
Because Radiant doesn't do LRAM inference at all, and LRAM blocks are limited in quantity (2 in the 40k device, 5 in the 17k device) and relatively slow, this is opt-in only via  a lram attribute.

This perhaps could be revisited one day if we get utilisation-aware RAM mapping.


whitequark:
@daveshah1 Is there anyone in particular you'd like this PR to be reviewed by?

daveshah1:
No, just a quick check by anyone that it's reasonable is fine I think

whitequark (approved):

PR #2483, opened 2020-12-09
Return nice error in pmgen generated code, fixes #2482

### LINKED ISSUE #2482: Erronously instantiated SB_CARRY leads to crash
### Description
An error in the instantiation of `SB_CARRY` (input `I1` replicated) does not exit particularly gracefully. The last line in the log is:
    
    2.30. Executing ICE40_WRAPCARRY pass (wrap carries).

   
### Version
    Yosys 0.9+3652 (git sha1 623526d, clang 3.8.0-2ubuntu4 -fPIC -Os)

### Code to replicate this
      
    module Not_a_graceful_failure
      (input  a,b,
       output cy
       );   
       SB_CARRY l_cy0( .CO(cy),.CI(1'b0),.I1(b),.I1(a));   
    endmodule

### Result
    yosys -o obj_dir/top.json -p 'synth_ice40 -json obj_dir/R2008.json'  carryerr.v -g > obj_dir/synth.log
    what():  dict::at()
    Aborted (core dumped)

mmicko:


daveshah1 (changes requested):
I think exceptions will cause problems for the wasm build. cc @whitequark

whitequark:
@daveshah1 Yes, but the wasm build already crashes each time yosys calls `log_cmd_error` anyway. Ignore any exceptions on error paths; until wasm implements exceptions (which... hopefully will happen soon-ish) those will crash but it's not any worse than the already deployed situation. Exceptions on success path would be different.

daveshah1:
Ack, that sounds like it isn't as big a problem as I thought then. LGTM

daveshah1 (approved):

whitequark:
> Ack, that sounds like it isn't as big a problem as I thought then. LGTM

Well, it is a serious usability issue in that yowasp-yosys is currently not really usable for anything that might cause Yosys to error out. Effectively you can only use it with something like nMigen, which produces valid input almost all of the time.

However, there's just no reasonable way to fix this I'm aware of, so...

PR #2487, opened 2020-12-13
CXXRTL: implement zero-cost full coverage debug information through the magic✨ of outlining🪄🎀🧹

whitequark:
This PR significantly improves almost every aspect of CXXRTL's performance. Most importantly, it eliminates the tradeoff between debug information coverage (which directly translates to available VCD traces and so on), runtime throughput, and compilation latency.

That's right: CXXRTL no longer has the `-Og` (optimize for debug coverage) option because it is now able to provide **100% debug coverage for every net in the source HDL at all times regardless of optimization settings**.

That's not all though. This PR also improves compilation latency, runtime throughput at 100% coverage, VCD writing performance, and VCD file size. On the benchmark I'm using (1M cycles of Minerva with UART blackbox), this changeset makes CXXRTL set up for full coverage **compile ~30% faster, run ~40% faster without being probed, run ~55% faster while writing every trace (except memories) to VCD, and the resulting VCD file is ~65% smaller!** Conversely, it makes CXXRTL set up for rapid iteration at partial coverage **compile ~20% faster, run at the same speed, and fit 7% more traces into a VCD file of the same size!** Sorcery, isn't it?

Here's the benchmark results (arranged in the ugliest table Markdown has ever rendered), before this changeset:

| scenario         | options   | compile time | no-probe run time | full VCD run time | VCD size | net coverage | 
|------------------|-----------|--------------|-------------------|-------------------|----------|--------------|
| full coverage    | `-O4 -g1` | ~12.1s | ~1.74s | ~24.0s | ~1784M | **100%** (1816/1816) |
| max performance  | `-O6 -g1` | ~8.2s  | **~1.05s** | ~5.5s | ~275M | 44% (800/1816) |

And here's the benchmark results for the same design after the changeset:

| scenario         | options   | compile time | no-probe run time | full VCD run time | VCD size | net coverage | 
|------------------|-----------|--------------|-------------------|-------------------|----------|--------------|
| max performance *and* full coverage | `-O6 -g3` | ~8.2s  | **~1.05s** | ~10.3s | ~597M | **100%** (1816/1816) |
| fast compilation | `-O6 -g2` | ~6.5s  | **~1.05s** | ~6.0s | ~275M | 47% (861/1816) |

<sub>Specific numbers included for entertainment purposes only. Your mileage may vary. I was more concerned about refining the instrumentation and avoiding pathological C++ compiler edge cases than improving performance, anyway.</sub>

I've experimented with DWARF-like "zero cost" debug information for RTL simulators for several months at this point, and during that time I've been calling it "zero cost" with the air quotes: while some of my solutions technically did achieve that goal, the cost was simply shifted elsewhere and increased enough to prohibit any practical use. However, this one seems like it can be called "zero cost" without reservations, since it is a strict improvement in every important aspect.

<hr>

So how does this all work? First, some context.

Until now, CXXRTL could not effectively provide full coverage because its compilation model requires close cooperation with a C++ compiler that is limited in its ability to reason about CXXRTL's algorithms. The final machine code has the best performance when the generated C++ code sticks to manipulating temporaries and locals. Storing many intermediate computation results into memory presents a significant obstacle to the optimizer.

Aside from interfering with the machine code compiler, many nets that appear in a typical HDL project serve no useful function yet require resources to keep track of. At the same time, these nets can be important for debugging. Consider, for example, unused signals on a high fanout bus. Even if the arbiters and decoders are designed in a way that lets the simulator discard those signals, they are still useful in VCD traces by confirming that the bus isn't stuck in some obscure state.

The solution is simple but surprisingly effective: generate two separate code paths that share the same state. The "performance" code path simulates only the nets that are required for the design to function, and stores in memory only the values of the registers; any intermediates remain just that--intermediates--even if they correspond to a net in the HDL source. The "coverage" code path is the opposite: it uses the stored design state to reconstruct every intermediate that corresponds to a user-defined net, and does nothing else.

For the most part, the goal of an RTL simulator is to be as fast as possible. No one will look at the vast majority of the cycles, so there's no real need to calculate and record the values of redundant nets, either. On the other hand, handling the small amount of cycles that *do* require attention is more important than anything else. These two observations determined the approach I took in this PR.

My earlier attempts at "zero cost" debug information relied on fine-grained representations of netlists to do the minimum amount of necessary work to evaluate any requested signal, under the hypothesis that 'only paying for what you use' is valuable in the context of an RTL simulator. Unfortunately, these fine-grained representations resulted in intractable compile times, and the coordination logic had enough overhead to make full trace dumps several times slower than just evaluating everything eagerly. After discussing the prototype with a few fellow FPGA designers, I understood that any solution, no matter how elegant conceptually, must justify its existence by being faster than just rebuilding the design with a different set of flags when something breaks. Eventually, I concluded that providing true "zero cost" debug information is not viable and focused on improving the experience with only partial coverage.

However, CXXRTL still had a major reason to provide full coverage in at least some capacity: nMigen. My goal of integrating CXXRTL into nMigen transparently required the former to provide access to any net, since breaking existing nMigen testbenches would not be very transparent. That's how `-Og` came into existence: it trades off however much performance is necessary to get full coverage because anything less than full coverage would be useless.

At this point, CXXRTL could be configured in one of the two modes: either more performance than coverage, or more coverage than performance. Since they are functionally identical in every way except for a few scheduler tweaks, eventually I realized I could just put both in the same build artifact, in the same namespace, with the same entry point, sharing the same class hierarchy and design state, with the option to switch between them at any time.

Knowing the highly asymmetric use patterns of RTL simulators, I decided to adapt the high coverage mode to be especially suited to the one workload every designer I talked to expressed a strong preference for: full VCD dumps. Rather than being configurable or fine-grained, it would always come with storage for every user-defined net (that is otherwise optimized to an inaccessible intermediate), and provide a single function that reconstructs the values of all of those nets. Conversely, it would feature no caching, or any other runtime logic, or even any way to affect the state of the design. Only nets driven by fully combinatorial logic can be optimized out in a way that reduces coverage, so it makes sense that the mechanism for reverting that would be fully combinatorial, too.

As expected, this plan worked well for making full VCD dumps. However, what I wasn't expecting is discovering that it is actually an approach that is faster in a wide range of circumstances. Consider that (with this changeset applied) saving a full VCD dump of a design built with `-O4` (which causes the value of every user-defined net to be recorded when it is computed) takes more than *twice the time* as it takes the same design built with `-O6` (which causes the intermediate results to be discarded after being used to advance the simulation and then computed anew once the VCD writer needs them). And this is a design that converges to fixed point in one delta cycle! For designs that converge in multiple delta cycles (e.g. non-flattened), evaluating all of this logic once after they converge multiplies the performance improvement.

Building the infrastructure that classifies and schedules netlists according to this model has resulted in another nice side effect. Contemporary HDL often comes in a very large number of interconnected modules. (Minerva has more than 40, and that's a fairly simple RISC-V CPU.) Consequently, a simulator that preserves the design hierarchy in its waveform dumps has to take into account that most nets in the design exist to preserve the correspondence between the (usually flattened) netlist and the hierarchical HDL source. (Out of 1816 user-defined nets in Minerva, 1285 are aliases.) CXXRTL had a way to represent aliases in debug information for a while, but the corresponding analysis was very conservative to make sure that e.g. the aliasee it picks will not be unexpectedly optimized out. Now that it has the ability to reconstruct every net that was optimized out, this is no longer an issue, which allows for precise alias detection, which in turn reduces the size of VCD files even as more traces are added.

What I didn't anticipate at all is that this all will be possible without increasing compilation time, and the reason for that is very silly. Being fed C++ code mechanically transformed from RTL netlists that relies on bignum arithmetics implemented with templates stacked dozens of levels deep, with so much redundancy that function merging is a hard requirement, and structured in a way that requires aggressive loop unrolling to *optimize for size*? Modern C++ compilers hardly find that a challenge, crunching through thousands of lines and far more template instantiations in just a few seconds while producing nice compact machine code. Then they spend about as long struggling with a futile attempt to optimize a function that's just 2000 coldest method calls in the entire unit, arranged in a straight line. I thought I'd have to rearrange half of the support template library to get a 2× speedup; nope, just had to place a single `__attribute__((optnone))`. Not that I complain.

### DIFF
--- a/backends/cxxrtl/cxxrtl.h
+++ b/backends/cxxrtl/cxxrtl.h
@@ -36,6 +36,7 @@
 #include <map>
 #include <algorithm>
 #include <memory>
+#include <functional>
 #include <sstream>
 
 #include <backends/cxxrtl/cxxrtl_capi.h>
@@ -843,6 +844,9 @@ typedef std::map<std::string, metadata> metadata_map;
 // Tag class to disambiguate values/wires and their aliases.
 struct debug_alias {};
 
+// Tag declaration to disambiguate values and debug outlines.
+using debug_outline = ::_cxxrtl_outline;
+
 // This structure is intended for consumption via foreign function interfaces, like Python's ctypes.
 // Because of this it uses a C-style layout that is easy to parse rather than more idiomatic C++.
 //
@@ -851,10 +855,11 @@ struct debug_alias {};
 struct debug_item : ::cxxrtl_object {
 	// Object types.
 	enum : uint32_t {
-		VALUE  = CXXRTL_VALUE,
-		WIRE   = CXXRTL_WIRE,
-		MEMORY = CXXRTL_MEMORY,
-		ALIAS  = CXXRTL_ALIAS,
+		VALUE   = CXXRTL_VALUE,
+		WIRE    = CXXRTL_WIRE,
+		MEMORY  = CXXRTL_MEMORY,
+		ALIAS   = CXXRTL_ALIAS,
+		OUTLINE = CXXRTL_OUTLINE,
 	};
 
 	// Object flags.
@@ -881,6 +886,7 @@ struct debug_item : ::cxxrtl_object {
 		zero_at = 0;
 		curr    = item.data;
 		next    = item.data;
+		outline = nullptr;
 	}
 
 	template<size_t Bits>
@@ -895,6 +901,7 @@ struct debug_item : ::cxxrtl_object {
 		zero_at = 0;
 		curr    = const_cast<chunk_t*>(item.data);
 		next    = nullptr;
+		outline = nullptr;
 	}
 
 	template<size_t Bits>
@@ -910,6 +917,7 @@ struct debug_item : ::cxxrtl_object {
 		zero_at = 0;
 		curr    = item.curr.data;
 		next    = item.next.data;
+		outline = nullptr;
 	}
 
 	template<size_t Width>
@@ -924,6 +932,7 @@ struct debug_item : ::cxxrtl_object {
 		zero_at = zero_offset;
 		curr    = item.data.empty() ? nullptr : item.data[0].data;
 		next    = nullptr;
+		outline = nullptr;
 	}
 
 	template<size_t Bits>
@@ -938,6 +947,7 @@ struct debug_item : ::cxxrtl_object {
 		zero_at = 0;
 		curr    = const_cast<chunk_t*>(item.data);
 		next    = nullptr;
+		outline = nullptr;
 	}
 
 	template<size_t Bits>
@@ -953,6 +963,22 @@ struct debug_item : ::cxxrtl_object {
 		zero_at = 0;
 		curr    = const_cast<chunk_t*>(item.curr.data);
 		next    = nullptr;
+		outline = nullptr;
+	}
+
+	template<size_t Bits>
+	debug_item(debug_outline &group, const value<Bits> &item, size_t lsb_offset = 0) {
+		static_assert(sizeof(item) == value<Bits>::chunks * sizeof(chunk_t),
+		              "value<Bits> is not compatible with C layout");
+		type    = OUTLINE;
+		flags   = DRIVEN_COMB;
+		width   = Bits;
+		lsb_at  = lsb_offset;
+		depth   = 1;
+		zero_at = 0;
+		curr    = const_cast<chunk_t*>(item.data);
+		next    = nullptr;
+		outline = &group;
 	}
 };
 static_assert(std::is_standard_layout<debug_item>::value, "debug_item is not compatible with C layout");
@@ -1029,10 +1055,15 @@ struct module {
 
 } // namespace cxxrtl
 
-// Internal structure used to communicate with the implementation of the C interface.
+// Internal structures used to communicate with the implementation of the C interface.
+
 typedef struct _cxxrtl_toplevel {
 	std::unique_ptr<cxxrtl::module> module;
 } *cxxrtl_toplevel;
+
+typedef struct _cxxrtl_outline {
+	std::function<void()> eval;
+} *cxxrtl_outline;
 
 // Definitions of internal Yosys cells. Other than the functions in this namespace, CXXRTL is fully generic
 // and indepenent of Yosys implementation details.
--- a/backends/cxxrtl/cxxrtl.h
+++ b/backends/cxxrtl/cxxrtl.h
@@ -41,17 +41,28 @@
 
 #include <backends/cxxrtl/cxxrtl_capi.h>
 
+#ifndef __has_attribute
+#	define __has_attribute(x) 0
+#endif
+
 // CXXRTL essentially uses the C++ compiler as a hygienic macro engine that feeds an instruction selector.
 // It generates a lot of specialized template functions with relatively large bodies that, when inlined
 // into the caller and (for those with loops) unrolled, often expose many new optimization opportunities.
 // Because of this, most of the CXXRTL runtime must be always inlined for best performance.
-#ifndef __has_attribute
-#	define __has_attribute(x) 0
-#endif
 #if __has_attribute(always_inline)
 #define CXXRTL_ALWAYS_INLINE inline __attribute__((__always_inline__))
 #else
 #define CXXRTL_ALWAYS_INLINE inline
+#endif
+// Conversely, some functions in the generated code are extremely large yet very cold, with both of these
+// properties being extreme enough to confuse C++ compilers into spending pathological amounts of time
+// on a futile (the code becomes worse) attempt to optimize the least important parts of code.
+#if __has_attribute(optnone)
+#define CXXRTL_EXTREMELY_COLD __attribute__((__optnone__))
+#elif __has_attribute(optimize)
+#define CXXRTL_EXTREMELY_COLD __attribute__((__optimize__(0)))
+#else
+#define CXXRTL_EXTREMELY_COLD
 #endif
 
 // CXXRTL uses assert() to check for C++ contract violations (which may result in e.g. undefined behavior
--- a/backends/cxxrtl/cxxrtl_backend.cc
+++ b/backends/cxxrtl/cxxrtl_backend.cc
@@ -539,6 +539,7 @@ struct CxxrtlWorker {
 	bool inline_public = false;
 
 	bool debug_info = false;
+	bool debug_alias = false;
 	bool debug_eval = false;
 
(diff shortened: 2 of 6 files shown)


tomverbeure:
The PR as it is right now fails on my general test case. Unfortunately, I don't have a smaller test case.

Repro:
```
git clone https://github.com/tomverbeure/cxxrtl_eval.git
cd cxxrtl_eval/cxxrtl
~/tools/yosys-whitequark/yosys -p "read_verilog ../spinal/ExampleTop.sim.v; hierarchy -check -top ExampleTop; write_cxxrtl -O6 -g3 ExampleTop.sim.default.cpp"
clang++ -g -O3 -std=c++14 -DEXAMPLE_TOP=\"ExampleTop.sim.default.cpp\" -I ~/tools/yosys-whitequark/share/include main.cpp -I../lib ../lib/cxxrtl_lib.cpp -o example_default_clang
```

Error:
```
In file included from main.cpp:10:
./ExampleTop.sim.default.cpp:18165:79: error: use of undeclared identifier 'i_flatten_5c_cpu__u__cpu_2e__5c_u__timer_2e__24_ne_24__2e__2e__2f_spinal_2f_ExampleTop_2e_sim_2e_v_3a_6075_24_1103__Y'
        items.add(path + "cpu_u_cpu u_timer_io_interrupt", debug_item(debug_alias(), i_flatten_5c_cpu__u__cpu_2e__5c_u__timer_2e__24_ne_24__2e__2e__2f_spinal_2f_ExampleTop_2e_sim_2e_v_3a_6075_24_11...
                                                                                     ^
./ExampleTop.sim.default.cpp:18179:89: error: use of undeclared identifier 'i_procmux_24_3283__Y'
        items.add(path + "cpu_u_cpu io_apb_decoder_io_input_PREADY", debug_item(debug_alias(), i_procmux_24_3283__Y, 0));
                                                                                               ^
./ExampleTop.sim.default.cpp:18180:89: error: use of undeclared identifier 'i_procmux_24_3290__Y'
        items.add(path + "cpu_u_cpu io_apb_decoder_io_input_PRDATA", debug_item(debug_alias(), i_procmux_24_3290__Y, 0));
                                                                                               ^
./ExampleTop.sim.default.cpp:18181:92: error: use of undeclared identifier 'i_procmux_24_3280__Y'
        items.add(path + "cpu_u_cpu io_apb_decoder_io_input_PSLVERROR", debug_item(debug_alias(), i_procmux_24_3280__Y, 0));
                                                                                                  ^
```

Isolation: replace `write_cxxrtl -O6 -g3` by  `write_cxxrtl -O6 -g2` and it passes.

whitequark:
@tomverbeure Fixed.

tomverbeure:
After the latest changes, `-O6 -g3` works fine, including probing the `PSEL` signals. However, surprisingly, `-O6 -g2` now fails to execute due to `PSEL` not being a `debug_item.`

Repro:
* Pull the latest version of my repo
* Execute:
 
```
cd cxxrtl
~/tools/yosys-whitequark/yosys -p "read_verilog ../spinal/ExampleTop.sim.v; hierarchy -check -top ExampleTop; write_cxxrtl -O6 -g2 ExampleTop.sim.Og.cpp"
clang++ -g -O3 -std=c++14 -DEXAMPLE_TOP=\"ExampleTop.sim.Og.cpp\" -I ~/tools/yosys-whitequark/share/include main.cpp -I../lib ../lib/cxxrtl_lib.cpp -o example_Og_clang
./example_Og_clang
```

Result:

```
tom@thinkcenter:~/projects/cxxrtl_eval/cxxrtl$ ./example_Og_clang
  what():  map::at
Aborted (core dumped)
``` 

Generated code with -g2 (failing case):

```
	items.add(path + "cpu_u_cpu u_uart io_apb_PWDATA", debug_item(debug_alias(), p_cpu__u__cpu_2e_u__cpu_2e_apbBridge_2e_io__pipelinedMemoryBus__cmd__halfPipe__regs__payload__data, 0));
	items.add(path + "cpu_u_cpu u_uart io_apb_PWRITE", debug_item(debug_alias(), p_cpu__u__cpu_2e_u__cpu_2e_apbBridge_2e_io__pipelinedMemoryBus__cmd__halfPipe__regs__payload__write, 0));
	static const value<1> const_p_cpu__u__cpu_2e_u__uart_2e_io__apb__PREADY = value<1>{0x1u};
	items.add(path + "cpu_u_cpu u_uart io_apb_PREADY", debug_item(const_p_cpu__u__cpu_2e_u__uart_2e_io__apb__PREADY, 0));
	items.add(path + "cpu_u_cpu u_uart io_apb_PENABLE", debug_item(debug_alias(), p_cpu__u__cpu_2e_u__cpu_2e_apbBridge_2e_state, 0));
```

Generated code with -g3 (passing case):

```
	items.add(path + "cpu_u_cpu u_uart io_apb_PRDATA", debug_item(debug_eval_outline, p_cpu__u__cpu_2e_u__uart_2e_io__apb__PRDATA, 0));
	items.add(path + "cpu_u_cpu u_uart io_apb_PWDATA", debug_item(debug_alias(), p_cpu__u__cpu_2e_u__cpu_2e_apbBridge_2e_io__pipelinedMemoryBus__cmd__halfPipe__regs__payload__data, 0));
	items.add(path + "cpu_u_cpu u_uart io_apb_PWRITE", debug_item(debug_alias(), p_cpu__u__cpu_2e_u__cpu_2e_apbBridge_2e_io__pipelinedMemoryBus__cmd__halfPipe__regs__payload__write, 0));
	static const value<1> const_p_cpu__u__cpu_2e_u__uart_2e_io__apb__PREADY = value<1>{0x1u};
	items.add(path + "cpu_u_cpu u_uart io_apb_PREADY", debug_item(const_p_cpu__u__cpu_2e_u__uart_2e_io__apb__PREADY, 0));
	items.add(path + "cpu_u_cpu u_uart io_apb_PENABLE", debug_item(debug_alias(), p_cpu__u__cpu_2e_u__cpu_2e_apbBridge_2e_state, 0));
	items.add(path + "cpu_u_cpu u_uart io_apb_PSEL", debug_item(debug_alias(), p_cpu__u__cpu_2e_apb3Router__1___2e_io__outputs__2__PSEL, 0));
	items.add(path + "cpu_u_cpu u_uart io_apb_PADDR", debug_item(debug_alias(), p_cpu__u__cpu_2e___zz__CpuTop__5__, 0));
```

whitequark:
> However `-O6 -g2` fails to execute due to `PSEL` not being a `debug_item.`

Yes, this is expected. `-O6 -g2` after this change is equivalent to `-O6 -g1` before this change, which did not let you examine `PSEL` either.

PR #2490, opened 2020-12-16
sta: very crude static timing analysis pass

Ravenslofty:
This is @eddiehung's static timing analysis command grafted onto a new branch. While that branch has a lot of other code changes, I'm not too familiar with the areas involved, so chose to keep the changes as minimal as feasible.

This modifies some of the ABC9 code due to a change in `TimingInfo`, but the ABC9 test passes locally, so I'm reasonably confident I didn't break anything there.

Supersedes #1732.


daveshah1:
It looks like the sta command itself is missing from this PR and it just has the surrounding changes?

Ravenslofty:
Serves me right for filing a PR at 3am...

kernel/rtlil.cc
```diff
@@ -445,6 +445,27 @@ vector<string> RTLIL::AttrObject::get_hdlname_attribute() const
...
+	attributes[id] = RTLIL::Const(attrval.str());
+}
+
+vector<int> RTLIL::AttrObject::get_intvec_attribute(RTLIL::IdString id) const
+{
+	vector<int> data;
+	auto it = attributes.find(id);
+	if (it != attributes.end())
+		for (const auto &s : split_tokens(attributes.at(id).decode_string()))
+			data.push_back(atoi(s.c_str()));
```
whitequark, line 465: Can we please not make Yosys have even more poorly defined type conversions? This one actually executes UB if the input is out of range, and also has different behavior on 32-bit and 64-bit platforms. Take a look at the RTLIL lexer to see how this can be done safely.

Ravenslofty: I more or less copied straight from that; hopefully that's good enough.

kernel/rtlil.cc
```diff
@@ -445,6 +445,34 @@ vector<string> RTLIL::AttrObject::get_hdlname_attribute() const
...
+vector<int> RTLIL::AttrObject::get_intvec_attribute(RTLIL::IdString id) const
+{
+	vector<int> data;
+	auto it = attributes.find(id);
+	if (it != attributes.end())
+		for (const auto &s : split_tokens(attributes.at(id).decode_string())) {
+			char *end = nullptr;
+			errno = 0;
+			long value = strtol(s.c_str(), &end, 10);
+			log_assert(end == s.c_str() + s.size());
```
whitequark, line 468: I believe the Yosys policy is to never assert on user-specified attributes, so this should be a more proper error (something like "has invalid format").

kernel/rtlil.cc
```diff
@@ -445,6 +445,34 @@ vector<string> RTLIL::AttrObject::get_hdlname_attribute() const
...
+			if (errno == ERANGE || value < INT_MIN || value > INT_MAX)
+				log_error("Literal for intvec attribute is out of range");
+			data.push_back(value);
```
whitequark, line 471: Because you used `log_error` rather `log_cmd_error`, this line would go ahead and invoke UB anyway for out-of-range literals. The errors should be `log_cmd_error`, which throw an exception.

whitequark (changes requested):

kernel/rtlil.cc
Ravenslofty: I added a conditional `log_cmd_error` for it.

kernel/rtlil.cc
Ravenslofty: Woops, fixed.

whitequark: Looks good now.

Ravenslofty:
Rebased to 113c9438.

passes/cmds/sta.cc
```diff
@@ -0,0 +1,311 @@
...
+
+	struct t_data {
+		Cell* driver;
+		IdString dst_port, src_port;
+		vector<tuple<SigBit,int,IdString>> fanouts;
+		SigBit backtrack;
+		t_data() : driver(nullptr) {}
+	};
+	dict<SigBit, t_data> data;
+	std::deque<SigBit> queue;
```
mwkmwkmwk, line 42: this file does not actually `#include <deque>`, breaking MSVC build (see CI)

Ravenslofty: Fixed.

PR #2491, opened 2020-12-16
Sign extend port connections where necessary

### LINKED ISSUE #1418: negative numbers are not sign expanded across module ports

yosys

[neg_demo.zip](https://github.com/YosysHQ/yosys/files/3671537/neg_demo.zip)

## Expected behavior

FPGA should match simulation

## Actual behavior

If a signed module port is mapped to a larger register, the upper bits are all set to 0, so the negative number becomes positive 2's complement.

### LINKED ISSUE #2265: Port connections do not sign extend

```Verilog
module Example(a, b);
    input wire [3:0] a;
    output wire [3:0] b;
    assign b = a;
endmodule

module top(o3, o4);
    output wire [3:0] o3, o4;
    Example e3(1'sb1, o3);
    wire [3:0] tmp;
    assign tmp = 1'sb1;
    Example e4(tmp, o4);
endmodule
```
## Expected behavior

Both `e3` and `e4` should have all 1s bound to their input ports, as per Section 12.3.11 of IEEE 1364-2005:
> Any expressions on a port shall be treated as any other expression in an assignment. It shall be typed, sized, and evaluated, and the resulting value assigned to the object on the other side of the port using the same rules as an assignment.

## Actual behavior

Only `e4`, the instantiation with the explicit indirection, receives all 1s on its input. `e3` incorrectly receives `4'b0001`.

zachjs:
- Signed cell outputs are sign extended when bound to larger wires
- Signed connections are sign extended when bound to larger cell inputs
- Sign extension is performed in hierarchy and flatten phases
- genrtlil indirects signed constants through signed wires
- Other phases producing RTLIL may need to be updated to preserve signedness information
- Resolves #1418
- Resolves #2265

This change may need further revisions, but I am eager for feedback!

### DIFF
--- a/frontends/ast/genrtlil.cc
+++ b/frontends/ast/genrtlil.cc
@@ -106,6 +106,7 @@ static RTLIL::SigSpec binop2rtlil(AstNode *that, IdString type, int result_width
 
 	RTLIL::Wire *wire = current_module->addWire(cell->name.str() + "_Y", result_width);
 	wire->attributes[ID::src] = stringf("%s:%d.%d-%d.%d", that->filename.c_str(), that->location.first_line, that->location.first_column, that->location.last_line, that->location.last_column);
+	wire->is_signed = that->is_signed;
 
 	for (auto &attr : that->attributes) {
 		if (attr.second->type != AST_CONSTANT)
@@ -1721,8 +1722,29 @@ RTLIL::SigSpec AstNode::genRTLIL(int width_hint, bool sign_hint)
 				}
 				if (child->type == AST_ARGUMENT) {
 					RTLIL::SigSpec sig;
-					if (child->children.size() > 0)
-						sig = child->children[0]->genRTLIL();
+					if (child->children.size() > 0) {
+						AstNode *arg = child->children[0];
+						int local_width_hint = -1;
+						bool local_sign_hint = false;
+						// don't inadvertently attempt to detect the width of interfaces
+						if (arg->type != AST_IDENTIFIER || !arg->id2ast || arg->id2ast->type != AST_CELL)
+							arg->detectSignWidth(local_width_hint, local_sign_hint);
+						sig = arg->genRTLIL(local_width_hint, local_sign_hint);
+						log_assert(local_sign_hint == arg->is_signed);
+						if (sig.is_wire()) {
+							// if the resulting SigSpec is a wire, its
+							// signedness should match that of the AstNode
+							log_assert(arg->is_signed == sig.as_wire()->is_signed);
+						} else if (arg->is_signed) {
+							// non-trivial signed nodes are indirected through
+							// signed wires to enable sign extension
+							RTLIL::IdString wire_name = NEW_ID;
+							RTLIL::Wire *wire = current_module->addWire(wire_name, arg->bits.size());
+							wire->is_signed = true;
+							current_module->connect(wire, sig);
+							sig = wire;
+						}
+					}
 					if (child->str.size() == 0) {
 						char buf[100];
 						snprintf(buf, 100, "$%d", ++port_counter);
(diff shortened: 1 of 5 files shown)


whitequark (changes requested):
I don't think the logic related to signed constants in RTLIL is necessary. Instead, the constant could be assigned to an intermediate signed wire, which would then be connected to a port. That would minimize the changes to the RTLIL core.

passes/hierarchy/hierarchy.cc
```diff
@@ -1233,14 +1233,26 @@ struct HierarchyPass : public Pass {
...
 							if (w->port_input && !w->port_output)
-								sig.append(Const(0, n));
+							{
+								bool is_signed = false;
+								if (sig.chunks().size() == 1)
+								{
+									const SigChunk& chunk = sig.chunks()[0];
+									is_signed = chunk.wire && chunk.wire->is_signed;
+								}
+								sig.extend_u0(GetSize(w), is_signed);
```
whitequark, line 1254: Something like:
```suggestion
								sig.extend_u0(GetSize(w), sig.is_wire() && sig.as_wire()->is_signed);
```

frontends/ast/genrtlil.cc
```diff
@@ -1720,16 +1720,31 @@ RTLIL::SigSpec AstNode::genRTLIL(int width_hint, bool sign_hint)
...
-						cell->setPort(buf, sig);
+						port_id = buf;
 					} else {
-						cell->setPort(child->str, sig);
+						port_id = child->str;
+					}
+					RTLIL::SigSpec sig;
+					if (child->children.size() > 0) {
+						AstNode *arg = child->children[0];
+						if (arg->type == AST_CONSTANT && arg->is_signed) {
```
whitequark, line 1734: I'm not convinced that this covers every possible case. I think it would be more robust to add an indirect wire each time the node connected to the port is signed and the output of `genRTLIL` is not a wire. And maybe add an assert for the case where it is a wire but the wire is not signed, since that seems like it should never happen.

whitequark (changes requested):

frontends/ast/genrtlil.cc
```diff
@@ -1720,16 +1721,39 @@ RTLIL::SigSpec AstNode::genRTLIL(int width_hint, bool sign_hint)
...
+						AstNode *arg = child->children[0];
+						int local_width_hint = -1;
+						bool local_sign_hint = false;
+						// don't inadvertently attempt to detect the width of interfaces
+						if (arg->type != AST_IDENTIFIER || !arg->id2ast || arg->id2ast->type != AST_CELL)
+							arg->detectSignWidth(local_width_hint, local_sign_hint);
+						sig = arg->genRTLIL(local_width_hint, local_sign_hint);
+						log_assert(local_sign_hint == arg->is_signed);
+						if (sig.is_wire()) {
+							// if the resulting SigSpec is a wire, its signing
```
whitequark, line 1743: `signedness`, not `signing`

frontends/ast/genrtlil.cc
```diff
@@ -1720,16 +1721,39 @@ RTLIL::SigSpec AstNode::genRTLIL(int width_hint, bool sign_hint)
...
+							// should match the signing of the AstNode
+							log_assert(arg->is_signed == sig.as_wire()->is_signed);
+						} else if (arg->is_signed) {
+							// non-trivial signed nodes are indirected through
+							// signed wires to enable sign extension
+							std::string wire_name = cell->name.str() + "_indirect_" + port_id.str() + "_tmp";
```
whitequark, line 1749: This wire name should probably use autoidx, just `NEW_ID` seems fine.

whitequark (changes requested):
LGTM after requested minor changes.

whitequark (approved):
LGTM

whitequark:
Thanks!

whitequark:
@zachjs This PR introduces a bug. To reproduce, clone https://github.com/jotego/jt51 and run `yosys hdl/*.v`. Please take a look!

zachjs:
@whitequark Please check out https://github.com/YosysHQ/yosys/pull/2501!

PR #2495, opened 2020-12-20
cxxrtl: preserve interior memory pointers across reset

whitequark:
This PR works fine in isolation. However, it changes the initialization strategy: with this PR applied, it is partly done with default-initialization (for wires) and partly explicitly (for memories). It would perhaps be nice to do all initialization explicitly, and then get rid of the weird "adopt" constructor.


whitequark:
Superseded by https://github.com/YosysHQ/yosys/pull/3105.

PR #2498, opened 2020-12-22
Fix use-after-free in LUT opt pass

StefanBruens:
RTLIL::Module::remove(Cell* cell) calls `delete cell`.

Any subsequent accesses of `cell` then causes undefined behavior.


whitequark (approved):
LGTM

PR #2501, opened 2020-12-23
genrtlil: fix mux2rtlil generated wire signedness

zachjs:


whitequark:
Do you think you could add a regression test? Genrtlil is a particularly fragile component.

whitequark (approved):
LGTM, thank you for the very prompt fix!

whitequark:
Would you be interested in taking a more active role in working on the Yosys Verilog frontend? Currently we have an unfortunate situation where none of the core Yosys maintainers can give it enough attention, but at the same time many people contribute improvements. Your PRs stood out to me as being particularly thought out, in the most complex part of the codebase no less, so if that's something you are interested in, we (speaking for the Yosys maintainers) would be happy to offer you co-maintainership of the frontend.

PR #2502, opened 2020-12-23
passes/pmgen/pmgen.py: trivial change to remove C++ compiler warnings

ldoolitt:
Verified that the result still builds and passes self-tests


ldoolitt:
Obligatory stackoverflow reference:
https://stackoverflow.com/questions/62030341/what-is-this-catching-polymorphic-type-x-by-value-wcatch-value

whitequark (approved):
LGTM

PR #2503, opened 2020-12-24
Changes to Liberty parser to make it more permissive of paranthetical expressions and keeper cells

jrrk2:
Dear Clifford,
  Keep up the good work with yosys. here is my humble contribution to the liberty parser improvement. I could understand it being done in a better way, however. The improvements are:

1. Explicitly skipping keeper cells
2. accepting state_function as a keyword as well as function (in flip-flops)
3. handling internal pins in a different way
4. handling paranthetical expressions after a keyword colon sequence


clairexen (changes requested):
The indenting of this is all messed up (spaces instead of blanks) and doesn't follow the CodyingStyle document (it looks like GNU coding style). Please update to use the proper indenting and coding style to match the surrounding code.

widlarizer:
The Liberty parser has moved on by a lot and has been exposed to a spectrum of PDK files. We've fixed a good couple of bugs and improved handling of expressions etc. Would be cool to salvage tests from this to find if we actually are ahead of the code changes here, but no tests are provided

PR #2506, opened 2020-12-26
Fix constants bound to redeclared function args

zachjs:
The changes in #2476 ensured that function inputs like `input x;` retained their single-bit size when instantiated with a constant argument and turned into a localparam. That change did not handle the possibility for an input to be redeclared later on with an explicit width, such as `integer x;`.


mmicko:
Thanks a lot

Issue #2505, opened 2020-12-26
Build on MSYS2 is broken


umarcor:
## Steps to reproduce the issue

Building Yosys on MSYS2 through a PKGBUILD recipe. See last commit in https://github.com/umarcor/yosys/commits/ci/msys2.

## Expected behavior

Yosys is successfully built on either MINGW64 or MINGW32.

## Actual behavior

On MINGW32, the failure is the following:

https://github.com/umarcor/yosys/runs/1609838330?check_suite_focus=true#step:5:88
```
/bin/sh: x86_64-w64-mingw32-g++: command not found
make: *** [Makefile:700: kernel/driver.o] Error 127
```

On MINGW64, the build was succesful until https://github.com/YosysHQ/yosys/commit/e825cf9d73a2126d518b2dc8f78114850c84ad32. See https://github.com/umarcor/yosys/actions/runs/444767078. It seems that https://github.com/YosysHQ/yosys/commit/b2221c1077d43a193c1a66d8017a7ed9ed1a2373 broke it: https://github.com/umarcor/yosys/runs/1609948936?check_suite_focus=true.

whitequark:
That doesn't seem right--the commit you mention does not change the build system.

umarcor:
It seems to be a syntax error, not related to the build system. In fact, the previous commit does work. Please, see https://github.com/umarcor/yosys/runs/1609948936?check_suite_focus=true#step:5:1236

whitequark:
Ah, yes. On Windows, it will be necessary to `#undef CONST` so that it would stop interfering with our code.

umarcor:
Thanks! That allowed to build 'master' on MINGW64 again!

What about MINGW32? Do you know whether Yosys is expected to work on Windows 32 bits? Was it ever tested/supported?

whitequark:
> What about MINGW32? Do you know whether Yosys is expected to work on Windows 32 bits? Was it ever tested/supported?

I don't know, but I think that if you fix the build system issue, it should compile and run fine.

umarcor:
My point is that `make config-msys2-64` exists, but `make config-msys2-32` does not. The build system tries using `x86_64-w64-mingw32-g++`, regardless of being MINGW64 (correct) or MINGW32 (incorrect). See https://github.com/umarcor/yosys/runs/1612345840?check_suite_focus=true#step:5:81.

With "fix the build system issue" you mean implementing `config-msys2-32` myself?

whitequark:
[...]

PR #2507, opened 2020-12-26
kernel/yosys.h: undef CONST on WIN32

umarcor:
As suggested by @whitequark in #2505, this is a minor fix for allowing Yosys to be built on MINGW64.

### DIFF
--- a/kernel/yosys.h
+++ b/kernel/yosys.h
@@ -121,8 +121,9 @@ extern Tcl_Obj *Tcl_ObjSetVar2(Tcl_Interp *interp, Tcl_Obj *part1Ptr, Tcl_Obj *p
 #    define fileno _fileno
 #  endif
 
-// mingw and msvc include `wingdi.h` which defines a TRANSPARENT macro
-// that conflicts with X(TRANSPARENT) entry in kernel/constids.inc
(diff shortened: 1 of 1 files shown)


kernel/yosys.h
```diff
@@ -124,6 +124,8 @@ extern Tcl_Obj *Tcl_ObjSetVar2(Tcl_Interp *interp, Tcl_Obj *part1Ptr, Tcl_Obj *p
 // mingw and msvc include `wingdi.h` which defines a TRANSPARENT macro
 // that conflicts with X(TRANSPARENT) entry in kernel/constids.inc
 #  undef TRANSPARENT
+
```
whitequark, line 127: A comment would be useful, perhaps grouping all undefs here under an overall "these Win32 defines conflict with our identifiers" banner.

umarcor: I put a comment before `CONST` and `TRANSPARENT`. Shall I move `NOMINMAX` and `YY_NO_UNISTD_H` down to this group too?

whitequark (approved):

kernel/yosys.h
whitequark: Those are different, I think, what you did is sufficient IMO.

PR #2512, opened 2020-12-28
plugin: enhance no-plugin error

### LINKED ISSUE #2408: Provide list of built-in plugins
When Yosys is built without dynamic plugin loading support, using option `-m` produces the following error:

> ERROR: This version of yosys is built without plugin support.

However, some plugins/frontends might be built-in. For example, ghdl-yosys-plugin can be built either with Yosys or as a shared library. Using `-m ghdl` in the former case produces the mentioned error. That's partially misleading because `ghdl` commands are available. Instead, I would like to propose listing the built-in plugins. For example:

> ERROR: This version of yosys includes the following plugins already: ghdl. Using plugins dynamically is not supported in this build. See reference_to_the_docs.

Ref open-tool-forge/fpga-toolchain#58 #1640

- Download and extract [open-tool-forge/fpga-toolchain](https://github.com/open-tool-forge/fpga-toolchain).
- `PATH=/path/to/toolchain/bin:$PATH yosys -m ghdl`

/cc @edbordin

umarcor:
Close #2408.

### DIFF
--- a/passes/cmds/plugin.cc
+++ b/passes/cmds/plugin.cc
@@ -93,7 +93,11 @@ void load_plugin(std::string filename, std::vector<std::string> aliases)
 #else
 void load_plugin(std::string, std::vector<std::string>)
 {
-	log_error("This version of yosys is built without plugin support.\n");
+	log_error(
+		"\n  This version of Yosys cannot load plugins at runtime.\n"
+		"  Some plugins may have been included at build time.\n"
+		"  Use option `-H' to see the available built-in and plugin commands.\n"
+	);
 }
 #endif
 


whitequark:
I think the new message is just as confusing in different ways. I would prefer something like:
```
log_error("This version of Yosys cannot load plugins at runtime.\n");
log_error("Some plugins may have been included at build time.\n");
log_error("Use `%s -H' to see the available built-in and plugin commands.\n", argv[0]);
```

umarcor:
I merged the three sentences in a single `log_error` call. I found that only the first one is executed. Aparto from that `argv` is not defined in the scope, so I reworded the last sentence to avoid it. Hope that's ok.

whitequark (approved):

PR #2514, opened 2020-12-28
makefile: add support for built-in ghdl-yosys-plugin

umarcor:
[ghdl-yosys-plugin](https://github.com/ghdl/ghdl-yosys-plugin) provides an open source VHDL frontend based on [GHDL](https://github.com/ghdl/ghdl), which allows Yosys to handle VHDL 1987, 1993, 2000 or 2008. There are two options for building ghdl-yosys-plugin:

- [Build as a module (shared library)](https://github.com/ghdl/ghdl-yosys-plugin#build-as-a-module-shared-library)
- [Build as part of yosys (not recommended)](https://github.com/ghdl/ghdl-yosys-plugin#build-as-part-of-yosys-not-recommended)

On Windows, Yosys does not support loading plugins as external shared libraries, they need to be built-in. Therefore, using the second (not recommended) approach is required. It involves patching Yosys' Makefile: [yosys.diff](https://github.com/ghdl/ghdl-yosys-plugin/blob/master/yosys.diff). That patch is harmless and having it upstreamed here can make building easier for less experienced users. After merging this PR, the build procedure would be:

```sh
  git clone https://github.com/YosysHQ/yosys
  git clone https://github.com/ghdl/ghdl-yosys-plugin

  cd yosys
  mv ../ghdl-yosys-plugin/src frontends/ghdl
  
  make config-*
  
  echo "ENABLE_GHDL=1" >> Makefile.conf
  # Optionally, set the prefix:
  # echo "GHDL_DIR=${MINGW_PREFIX}" >> Makefile.conf

  make
```

/cc @tgingold


Makefile
```diff
@@ -511,6 +512,14 @@ endif
 endif
 endif
 
+ifeq ($(ENABLE_GHDL),1)
+GHDL_DIR ?= /usr/local
```
whitequark, line 516: Should this not be `$(PREFIX)`? E.g.:

```suggestion
GHDL_PREFIX ?= $(PREFIX)
```

Makefile
```diff
@@ -511,6 +512,14 @@ endif
...
+GHDL_INCLUDE_DIR ?= $(GHDL_DIR)/include
+GHDL_LIB_DIR ?= $(GHDL_DIR)/lib
+CXXFLAGS += -I$(GHDL_INCLUDE_DIR) -DYOSYS_ENABLE_GHDL
+LDLIBS += $(GHDL_LIB_DIR)/libghdl.a $(shell cat $(GHDL_LIB_DIR)/libghdl.link)
```
whitequark, line 520: This would be more portable:
```suggestion
LDLIBS += $(GHDL_LIB_DIR)/libghdl.a $(file <$(GHDL_LIB_DIR)/libghdl.link)
```
(Yes, I know MinGW is needed anyway, but the existing Makefile is so horrifically slow due to forking lots of crap that you need something like kati to get a reasonable warm build time, so no need to make stuff even worse.)

whitequark (changes requested):

Makefile
umarcor: Indeed...

Makefile
umarcor: I was not aware of this possible improvement. Thanks!

whitequark (approved):

PR #2523, opened 2021-01-04
Add -nosynthesis flag for read_verilog command

tomverbeure:
Right now, `read_verilog` either defines SYNTHESIS or FORMAL, but there is no way to not define both SYNTHESIS and FORMAL.

With the existence of CXXRTL, Yosys can also be used for pure simulation, in which case both SYNTHESIS and FORMAL defines are sometimes undesirable.

### DIFF
--- a/frontends/verilog/verilog_frontend.cc
+++ b/frontends/verilog/verilog_frontend.cc
@@ -84,6 +84,9 @@ struct VerilogFrontend : public Frontend {
 		log("        enable support for SystemVerilog assertions and some Yosys extensions\n");
 		log("        replace the implicit -D SYNTHESIS with -D FORMAL\n");
 		log("\n");
+		log("    -nosynthesis\n");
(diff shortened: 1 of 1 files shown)


whitequark:
I am not entirely sure if this is the right approach. Usually you would use `ifndef SYNTHESIS` to guard non-synthesizable constructs. But CXXRTL only accepts synthesizable constructs because it is a simulator with synthesis semantics.

tomverbeure:
Different people have different rules about when to use `Ifndef SYNTHESIS`. For example, many companies code rules require it around $display and other constructs that are for debugging only, even if they are synthesizable. 

I think it’s right that Yosys declares SYNTHESIS by default (as do some other synthesis tools such as VIVADO, but not QUARTUS), but I also think that there should be ways to disable it. Adding a wrapper file with an undef is way too clunky for something that can be solved with just a command line flag? 

I could move the flag down the list in the help message if it’s too prominent up high in the list with options.

PR #2529, opened 2021-01-07
verilog: significant block scoping improvements

### LINKED ISSUE #656: Unnamed generate blocks
I am running into errors elaborating the following module:

```verilog
module yosys_genblk_scoping #(
                                parameter WIDTH = 6
                                )
    (
     input [WIDTH-1:0]  a_i,
     input [WIDTH-1:0]  b_i,

     output [WIDTH-1:0] z_o
     );

    genvar g;

    generate
        for (g = 0; g < WIDTH; g = g + 1) begin
            if(g > 2) begin//: blk_true
                wire tmp;

                assign tmp    = a_i[g] || b_i[g];
                assign z_o[g] = tmp;
            end
            else begin
                assign z_o[g] = a_i[g] && b_i[g];
            end
        end
    endgenerate

endmodule
```

## Expected behavior
It works fine if the label `blk_true` is uncommented, but it should also work without it.

## Actual behavior
Without the label, I get this error:
```
Yosys 0.7+682 (git sha1 9850de40, gcc 6.4.0-8ubuntu1 -fPIC -Os)
...
 Generating RTLIL representation for module `\yosys_genblk_scoping'.
 yosys_genblk_scoping.v:18: ERROR: Failed to detect width of signal access `$genblock$yosys_genblk_scoping.v:14$4[3].\tmp'!
```

### LINKED ISSUE #2423: yosys does not support local variables as loop iterator

```systemverilog
`timescale 1us/1us

module test(aclk, in, out);
    input bit aclk;
    input bit in;
    output bit out;
    parameter DEPTH = 3;

    bit flipflop[DEPTH:0];

    assign flipflop[0] = in;
    assign out = flipflop[DEPTH];

    always_ff @(posedge aclk) begin
        integer i;
        for (i = 1; i < DEPTH; i++) begin
            flipflop[i] <= flipflop[i -1];
        end
    end
endmodule
```

```ini
[options]
mode prove

[engines]
smtbmc

[script]
read -formal -sv test.sv
prep -top test

[files]
test.sv
```

## Expected behavior

The loop works as it does in verilator and other tools:

```
$ verilator -Wall test.sv  --cc
$
```

For that matter they support:

```
for (integer i = 1; i < DEPTH; i++)
```

## Actual behavior

```
SBY 22:01:42 [test] base: Warning: wire '\i' is assigned in a block at test.sv:16.14-16.19.
SBY 22:01:42 [test] base: Warning: wire '\i' is assigned in a block at test.sv:16.32-16.35.
SBY 22:01:42 [test] base: test.sv:16: ERROR: Left hand side of 1st expression of procedural for-loop is not a register!
```

### LINKED ISSUE #2493: Cornercase. Verilog frontend exits with assert instead of an error message
Using `Yosys 0.9+3760 (git sha1 871fc34, clang 3.8.0-2ubuntu4 -fPIC -Os)`

    module top  # ( parameter AUTOPLACE = 0 )
       (
        input  rxck270,
        input  ck27start,
        output ubufclk135
        );
       
       generate
          if ( AUTOPLACE ) begin
             //             
          end else begin
             
             wire cmb_ck135 = ck27start ? 1'b1 : ~ubufclk135;           
             (* BEL="X1/Y11/lc3" *) SB_DFF reg_clk135(.Q(ubufclk135),.C(rxck270),.D(cmb_ck135));
    
             assign ubufclk135 = reg_clk135;   // This illegal statement trigger assert error.
             
          end
       endgenerate
    endmodule

## Expected behavior
An error message of some kind, on the illegal statement, for instance:

    top.v:20: ERROR: Failed to detect width for identifier \reg_clk135!

## Actual behavior

    ERROR: Assert `count_id(wire->name) == 0' failed in kernel/rtlil.cc:1801

zachjs:
This change set contains a number of bug fixes and improvements related to scoping and resolution in generate and procedural blocks. While many of the frontend changes are interdependent, it may be possible bring the techmap changes in under a separate PR.

Declarations within unnamed generate blocks previously encountered issues because the data declarations were left un-prefixed, breaking proper scoping. The LRM outlines behavior for generating names for unnamed generate blocks. The original goal was to add this implicit labelling, but doing so exposed a number of issues downstream. Additional testing highlighted other closely related scope resolution issues, which have been fixed. This change also adds support for block item declarations within unnamed blocks in SystemVerilog mode.

1. Unlabled generate blocks are now implicitly named according to the LRM in `label_genblks`, which is invoked at the beginning of module elaboration
2. The Verilog parser no longer wraps explicitly named generate blocks in a synthetic unnamed generate block to avoid creating extra hierarchy levels where they should not exist
3. The techmap phase now allows special control identifiers to be used outside of the topmost scope, which is necessary because such wires and cells often appear in unlabeled generate blocks, which now prefix the declarations within
4. Some techlibs required modifications because they relied on the previous invalid scope resolution behavior
5. `expand_genblock` has been simplified, now only expanding the outermost scope, completely deferring the inspection and elaboration of nested scopes; names are now resolved by looking in the innermost scope and stepping outward
6. Loop variables now always become localparams during unrolling, allowing them to be resolved and shadowed like any other identifier
7. Identifiers in synthetic function call scopes are now prefixed and resolved in largely the same manner as other blocks
     before: `$func$\func_01$tests/simple/scopes.blk.v:60$5$\blk\x`
      after: `$func$\func_01$tests/simple/scopes.v:60$5.blk.x`
8. Support identifiers referencing a local generate scope nested more than 1 level deep, i.e. `B.C.x` while within generate scope `A`, or using a prefix of a current or parent scope, i.e. `B.C.D.x` while in `A.B`, `A.B.C`, or `A.B.C.D`
9. Variables can now be declared within unnamed blocks in SystemVerilog mode

Fixes #656
Fixes #2423
Fixes #2493

### DIFF
--- a/frontends/verilog/verilog_parser.y
+++ b/frontends/verilog/verilog_parser.y
@@ -770,6 +770,7 @@ module_body:
 	module_body module_body_stmt |
 	/* the following line makes the generate..endgenrate keywords optional */
 	module_body gen_stmt |
+	module_body gen_block |
 	module_body ';' |
 	%empty;
 
@@ -2459,6 +2460,16 @@ behavioral_stmt:
 		exitTypeScope();
 		if ($4 != NULL && $8 != NULL && *$4 != *$8)
 			frontend_verilog_yyerror("Begin label (%s) and end label (%s) don't match.", $4->c_str()+1, $8->c_str()+1);
+		AstNode *node = ast_stack.back();
+		// In SystemVerilog, unnamed blocks with block item declarations
+		// create an implicit hierarchy scope
+		if (sv_mode && node->str.empty())
+		    for (const AstNode* child : node->children)
+			if (child->type == AST_WIRE || child->type == AST_MEMORY || child->type == AST_PARAMETER
+				|| child->type == AST_LOCALPARAM || child->type == AST_TYPEDEF) {
+			    node->str = "$unnamed_block$" + std::to_string(autoidx++);
+			    break;
+			}
 		SET_AST_NODE_LOC(ast_stack.back(), @2, @8);
 		delete $4;
 		delete $8;
@@ -2473,6 +2484,7 @@ behavioral_stmt:
 		ast_stack.back()->children.push_back($7);
 	} ';' simple_behavioral_stmt ')' {
 		AstNode *block = new AstNode(AST_BLOCK);
+		block->str = "$for_loop$" + std::to_string(autoidx++);
 		ast_stack.back()->children.push_back(block);
 		ast_stack.push_back(block);
 	} behavioral_stmt {
@@ -2722,6 +2734,7 @@ single_arg:
 
 module_gen_body:
 	module_gen_body gen_stmt_or_module_body_stmt |
+	module_gen_body gen_block |
 	%empty;
 
 gen_stmt_or_module_body_stmt:
@@ -2747,12 +2760,7 @@ gen_stmt:
 		ast_stack.back()->children.push_back(node);
 		ast_stack.push_back(node);
 		ast_stack.back()->children.push_back($3);
-		AstNode *block = new AstNode(AST_GENBLOCK);
-		ast_stack.back()->children.push_back(block);
-		ast_stack.push_back(block);
-	} gen_stmt_block {
-		ast_stack.pop_back();
-	} opt_gen_else {
+	} gen_stmt_block opt_gen_else {
 		SET_AST_NODE_LOC(ast_stack.back(), @1, @7);
 		ast_stack.pop_back();
 	} |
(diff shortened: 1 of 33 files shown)


frontends/verilog/verilog_parser.y
```diff
@@ -2459,6 +2460,16 @@ behavioral_stmt:
...
 		if ($4 != NULL && $8 != NULL && *$4 != *$8)
 			frontend_verilog_yyerror("Begin label (%s) and end label (%s) don't match.", $4->c_str()+1, $8->c_str()+1);
+		AstNode *node = ast_stack.back();
+		// In SystemVerilog, unnamed blocks with block item declarations
+		// create an implicit hierarchy scope
+		if (sv_mode && node->str.empty())
+		    for (const AstNode* child : node->children)
+			if (child->type == AST_WIRE || child->type == AST_MEMORY || child->type == AST_PARAMETER
+				|| child->type == AST_LOCALPARAM || child->type == AST_TYPEDEF) {
+			    node->str = "$unamed_block$" + std::to_string(autoidx++);
```
TimRudy, line 2470:
```suggestion
			    node->str = "$unnamed_block$" + std::to_string(autoidx++);
```

whitequark (approved):
To the extent that I understand these changes I think they seem fine; let's merge this and fix issues that arise, if any, rather than having it stuck waiting for a perfect review.

Issue #2530, opened 2021-01-07
ABC9 flop bypass specify rules can cause combinational loops


Ravenslofty:
This testcase has been minimised from a QuickLogic benchmark:
```verilog
module top(input WBs_CLK_i, _1_, _0_, output reg Rx_overflow_detected_o);
  wire rx_fifo_cntr_rst;
  assign rx_fifo_cntr_rst = _1_ | Rx_overflow_detected_o;
  always @(posedge WBs_CLK_i, posedge rx_fifo_cntr_rst)
    if (rx_fifo_cntr_rst) Rx_overflow_detected_o <= 1'h0;
    else Rx_overflow_detected_o <= _0_;
endmodule
```
which is a flop that asynchronously resets itself depending on its output.

## Expected behavior

If you take an ABC flow, such as `synth_ice40`, you get:
```
   Number of wires:                  5
   Number of wire bits:              5
   Number of public wires:           3
   Number of public wire bits:       3
   Number of memories:               0
   Number of memory bits:            0
   Number of processes:              0
   Number of cells:                  2
     SB_DFFR                         1
     SB_LUT4                         1
```

## Actual behavior

If you take an ABC9 flow, such as `synth_ice40`, you get:
```
2.41.9. Executing ABC9_OPS pass (helper functions for ABC9).
<suppressed ~2 debug messages>
ERROR: Assert `no_loops' failed in passes/techmap/abc9_ops.cc:758.
```

The cause of this is that near-universally flop cells have specify rules to produce an ABC9 flop bypass box that models the combinational path from a set/reset pin to the output. In this particular case, however, this models a combinational loop through the flop, which causes ABC9 to assert.

The reason I'm posting this as an issue is to come to a consensus on how to resolve this given the tradeoffs.

- We could consider this a combinational loop, which is forbidden in RTLIL, keep the specify rules, and possibly extract better results, since a longer critical path means more area reduction opportunities.
- We could remove the specify rules, which I have confirmed fixes the issue, but potentially regress result quality in the rare case that the critical path involves an asynchronous set/reset.
- I could attempt to fix this in ABC9, but with Eddie gone, it's going to be a bug that stays around for a long while, since I am very unfamiliar with the code there.

cc @daveshah1 (for iCE40/ECP5/Nexus), @pepijndevos (for Gowin), @mwkmwkmwk (for Xilinx)

eddiehung:
[...]

PR #2535, opened 2021-01-11
scc: Add -specify option to find loops in boxes

Ravenslofty:
Part of the fix for #2530: enhance SCC to detect logic loops through combinational specify rules.

### DIFF
--- a/passes/cmds/scc.cc
+++ b/passes/cmds/scc.cc
@@ -37,7 +37,7 @@ struct SccWorker
 	RTLIL::Design *design;
 	RTLIL::Module *module;
 	SigMap sigmap;
-	CellTypes ct;
+	CellTypes ct, specifyCells;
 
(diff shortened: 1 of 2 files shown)


Ravenslofty:
Fixed a bug where `scc` segfaulted, since `bit.wire` may be `nullptr` if `bit` is a constant.

clairexen (approved):
LGTM. Any reason why this code only cares about `$specify2` and not `$specify3`?

Ravenslofty:
I'm happy to correct this assumption if it's wrong, but my understanding is that `$specify2` describes combinational paths (e.g. `(X => Y) = N`), while `$specify3` describes synchronous paths (e.g. `(posedge CLK => (Y : X)) = N`), and it's combinational paths that we care about in this context.

clairexen:
Well, there's `-all_cell_types` but I guess it's not relevant for the use-case.  ;)

PR #2536, opened 2021-01-12
Fixed missing goto statement in passes/techmap/abc.cc

TobiasFaller:
The goto statement seems to be missing.
Specifying "-g all" for abc produces an error: `ERROR: Command syntax error: Unsupported gate type: all`

**If this is not a bug then just close this PR**.


mmicko:
Thanks

PR #2537, opened 2021-01-13
Add buffer option to spice backend

pepijndevos:
I'm experimenting with doing mixed signal simulations with NgSpice and Yosys, similar to https://www.isotel.eu/mixedsim/intro/concept.html

Instead of a custom synth script, I'm basically trying to use the output of `synth` directly. Either way you end up having to define all the subcircuit models.

Rather than CMOS transistors, my models use the XSPICE event-driven digital models, which should be much faster than a full analog CMOS simulation. However, Yosys currently inserts DC sources to connect nets together, which is incompatible with digital signals.

So this PR introduces a `-buf` option that allows you to specify a subcircuit implementation instead of the DC source. From there you can just use the digital `d_buffer` cell to implement it. The default still uses the DC source, which works fine for analog simulations.

If there is any interest I could prepare a PR to upstream the XSPICE cell library, and maybe a `synth_xspice` command. My library basically implements most of the internal fine cells produced by `synth`, but needs an extra `dfflegalize` to handle all the flops that are not supported. Library: https://gist.github.com/pepijndevos/2f41507b821b9a3db65beb68d3330b3b Synth commands:
```
read_verilog -lib -icells +/simcells.v
synth -flatten
dfflegalize -cell \$_DFFSR_PPP_ 0
stat
write_spice -buf __BUF_ blinky.cir
```

image

### DIFF
--- a/backends/spice/spice.cc
+++ b/backends/spice/spice.cc
@@ -64,7 +64,7 @@ static void print_spice_net(std::ostream &f, RTLIL::SigBit s, std::string &neg, 
 	}
 }
 
-static void print_spice_module(std::ostream &f, RTLIL::Module *module, RTLIL::Design *design, std::string &neg, std::string &pos, std::string &ncpf, bool big_endian, bool use_inames)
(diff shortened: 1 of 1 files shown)


clairexen (approved):
LGTM

pepijndevos:
Yay thanks. Any opinion on some form of upstream XSPICE support? Or just keep that separately...

PR #2541, opened 2021-01-14
cxxrtl: Fix liveness checking for debug wires

### LINKED ISSUE #2540: assert in write_cxxrtl
m68k.v is the result of sv2v on fx68k

Do:
read -vlog95 m68k.v
hierarchy -top fx68k
write_cxxrtl m68k.cc

The write_cxxrtl step asserts in dump_sigchunk with a WireType::UNUSED that is not lhs.
ERROR: Assert `is_lhs' failed in backends/cxxrtl/cxxrtl_backend.cc:897.

[m68k.zip](https://github.com/YosysHQ/yosys/files/5815511/m68k.zip)

modwizcode:
The current liveness checking for cxxrtl breaks in certain esoteric cases. Notably when signals that were completely optimized out get reintroduced into the design (so called "outline wires") those wires are wholly calculated inside the debug pass, this is because we want to keep the optimized code for the main eval pass and only do the full computation if the debug state is requested explicitly.

Due to some annoyances of yosys itself, it's actually somewhat difficult to have public wires in a design that are unused, as at some point earlier (likely the flatten pass?) yosys has essentially renamed the public signal into a internal unnamed signal, since cxxrtl only generates this additional debug information for used public signals (i.e. ones that are optimized out but not completely unused), this feature currently mostly goes to waste.

Issue #2540 managed to find a test case that actually triggered the generation of these "compute on demand" outline wires, revealing that it wasn't tagging all the proper wires as "live" and used during debug that it should have, causing asserts to fail while generating the debug pass. The details of the problem and proposed fix are explained further in a comment below.

### DIFF
--- a/backends/cxxrtl/cxxrtl_backend.cc
+++ b/backends/cxxrtl/cxxrtl_backend.cc
@@ -2647,7 +2647,10 @@ struct CxxrtlWorker {
 					auto &debug_wire_type = debug_wire_types[wire];
 					if (wire->name.isPublic()) continue;
 
-					if (live_wires[wire].empty() && debug_live_wires[wire].empty()) {
+					// In constructing liveness for debug nodes, we skip member nodes updated during regular eval()
+					// this prevents debug_eval() from messing with eval()'s state, that process skips tagging members
+					// as live. By considering member wires live, debug wires that alias members work properly
+					if (debug_live_wires[wire].empty() && !wire_type.is_member()) {
 						continue; // wire never used
 					} else if (flow.is_inlinable(wire, debug_live_wires[wire])) {
 						log_assert(flow.wire_comb_defs[wire].size() == 1);
--- a/backends/cxxrtl/cxxrtl_backend.cc
+++ b/backends/cxxrtl/cxxrtl_backend.cc
@@ -2647,7 +2647,7 @@ struct CxxrtlWorker {
 					auto &debug_wire_type = debug_wire_types[wire];
 					if (wire->name.isPublic()) continue;
 
-					if (live_wires[wire].empty() || debug_live_wires[wire].empty()) {
+					if (live_wires[wire].empty() && debug_live_wires[wire].empty()) {
 						continue; // wire never used
 					} else if (flow.is_inlinable(wire, debug_live_wires[wire])) {
 						log_assert(flow.wire_comb_defs[wire].size() == 1);


modwizcode:
Currently, it seems that this fix can generate unused variables in the output source when on -g4 with a design that fails without the commit. I'm not positive what causes this and will have the follow the liveness checking logic in more detail to see where truely unused signals are getting reintroduced into the design.

modwizcode:
So I think I've figured out the issue here, the problem is I've not so far come up with a good way to fix it. 

What's happening is that checking both sets for liveness means that the debug method will also consider wires live if they're part of the standard `eval()` path. 

The current version (checking both sets for liveness) is I think just outright wrong, since it has to be live in both passes to pass the check, however sometimes a wire is only live for debug because it's normally inlined or otherwise optimized out in the optimized pass, that's the whole point of debug_eval, to reconstruct that information.

Doing this (checking both sets) prevents failing the asserts since signals that are inlined normally for eval are still considered "live", meaning they're marked with a type for debug (otherwise wires will default to `UNUSED` for the debug pass because of how Yosys's internal dict type works). 

The issue is that this also means that wires that are actually unused by the debug eval will be emitted. These are wires that the debug pass thinks are live but are actually not, because of the additional check allowing live wires from the optimize pass to be considered live for debug. Unfortunately some of these locals are not useful in debug because they're not reachable from nodes that are live for debug.

I think the actual problem here is that `debug_live_wires` should contain all the wires that are live for debug, but it's somehow forgetting some, at least in this case the union of `debug_live_wires` and `live_wires` contains a superset of the wires used in the design, so all that happens is a few additional wires are emitted that are unused.

To solve this problem I'm trying to figure out why the debug pass is not marking all the wires that form the input cone for the reachable debug nodes emitted to the scheduler. I think the fix that's here already is worth having, but correcting it to capture only the used variables is my current goal.

modwizcode:
# Background

After much effort I think I've fully determined the scope of the issue. 

During the first pass of debug info generation, public wires are tagged with initial types to use during the debug path, however, when a wire is tagged as `WireType::ALIAS` this redirects to the wire it's connected to. Tagging with `ALIAS` occurs under two circumstances, the first is that the wire is a member wire (exposed in the generated c++ class, either because it's buffered or because it's an actual member or outline (debug only member)), this makes sense as these wires will be precomputed from the standard pass of `eval()`.
https://github.com/YosysHQ/yosys/blob/339848b954491e6bb63a3062c60f840730123436/backends/cxxrtl/cxxrtl_backend.cc#L2600-L2610

The second condition is only done when debug eval is enabled (which is where this issue occurs i.e. `-g4`, currently the default), in this case the wire can be tagged as an alias as long as the name of the wire being aliased to is "public". Since all public wires will be exposed as outlines when `debug_eval` is enabled, we can also alias to these outline members.

The problem comes when we try to later use these aliases. Due to the way the code is structured, these aliased wires will be generated, but if they alias a member wire, it won't be tagged as "live". This is because we don't want to change non-debug state members in `debug_eval`, we just want to update the state that's *added* by allowing debug info to regenerate public signals that have been optimized away. Unfortunately skipping generating the nodes that update that state, also skips adding their connected wires to the `debug_live_wires` set as well. 

Not being tagged as live means that they'll skip the second phase of assigning `debug_wire_types` unless some other reference makes that member "live". When this is skipped, the aliased member wires won't be marked as used for when the `debug_eval` method is being generated.

# The Fix

My original fix was to consider all the wires that were live for the optimized `eval()` to be considered live for the debug pass, in addition to wires that were only live for debug. This seems to be the intent of the existing code, which seemingly had an accidental reversed condition (checking for liveness in both sets instead of either). This works, but some temporary (`WireType::LOCAL`) signals in the optimized pass get named in the debug pass, leading to unused variables in the final generated C++ file. 

We shouldn't have unused variables though, since we know all the cells being generated and what their inputs and outputs are, this should be enough to only emit the used portions. So this solution isn't ideal.

The revised solution accomplishes this by also checking if the wire is a member variable from the optimized design and considering those variables to be live also. This will result in this logic being run:
https://github.com/YosysHQ/yosys/blob/339848b954491e6bb63a3062c60f840730123436/backends/cxxrtl/cxxrtl_backend.cc#L2668-L2671

Which will accomplish the task of assigning member variables in the design as "used" by debug, but no more. This gets rid of the extra locals and seems to completely solve the problem.

mwkmwkmwk:
I'm sorry I don't have the expertise to review your PR (that will have to wait for @whitequark), but please write a better header line than "fix #bugnumber" and rename the PR appropriately — something like "cxxrtl: fix liveness checking for debug wires" would be perfect.

modwizcode:
Sorry about that, at the time I was trying to focus the discussion down into the issue itself since at first I had written an explanation there and thought it was going to be a quick simple patch. It ends up that it's not quite so straightforward and I never updated it later.

I see you updated the title for me already, so I went ahead and gave a more meaningful summary in the description too.

whitequark:
Superseded by #2871.

PR #2544, opened 2021-01-14
CXXRTL: Fix sliced bits as clock inputs

### LINKED ISSUE #2542: cxxrtl seems to lose some signals somehow
With the same verilog as Issue #2540 the generated C++ seems incorrect, no matter the -O level in write_cxxrtl.

Steps to reproduce
  read -vlog95 m68k.v
  hierarchy -top fx68k
  write_cxxrtl m68k.cc

The design (and the 68000 itself) includes a free-running lfsr-based divided-by-10 clock that is output on pin E.  The lfsr is called eCntr.  Looking at the C++ shows that nothing ever writes to p_E or p_eCntr which is not good at all.
[m68k.zip](https://github.com/YosysHQ/yosys/files/5816582/m68k.zip)

modwizcode:
Currently in cxxrtl due to an earlier change that fixed issues with constant signals driving clock inputs of clocked cells, when cells are clocked by a `SigSpec` that doesn't pass `is_wire()` most of them will be ignored. 

This causes an issue because slices of another signal (i.e. a single `SigBit`) don't return `true` for `is_wire()`. While it's arguable that there should be some `is_wire()` like function defined that passes for slices (at least for single bits), that's another issue.

In order to fix #2542, I implemented a subset of `is_wire()` internally to the cxxrtl backend, so that it can handle cells with clocks that are not directly backed by a wire. This doesn't fully convert all uses of `is_wire()` for clocks. Some of the code can map them onto their unique driving signal at the point of testing, which I think should be equivalent, but someone might want to verify that.

### DIFF
--- a/backends/cxxrtl/cxxrtl_backend.cc
+++ b/backends/cxxrtl/cxxrtl_backend.cc
@@ -601,6 +601,13 @@ struct WireType {
 	bool is_exact() const { return type == ALIAS || type == CONST; }
 };
 
+// Tests for a SigSpec that is backed by a specific slice of a wire, this is used
+// instead of .is_wire() on clocks because they can be only a portion of an underlying
+// wire
+bool is_wire_slice(const RTLIL::SigSpec& sig) {
+	return sig.is_chunk() && sig.chunks()[0].wire;
+}
+
 struct CxxrtlWorker {
 	bool split_intf = false;
 	std::string intf_filename;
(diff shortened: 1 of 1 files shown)


modwizcode:
I just realized that I included the commit from #2541 so I've forced pushed to remove that

modwizcode:
I've noticed that there's still a case this seems to fail to generate, so I've switched it to a draft as I work out that issue. 

The issue is as I suspected that if a module's input is already a vector of clocks, the way the code works I fall back on the fact that after passing through `SigMap` the unique resolved input wire will be the one . But if you were to blackbox the rest of the code and only feed in a vector and extract the clock from that, there's no way to map that bit back to it's unique driver. 

All I think this means is I need to patch the checks to be looser and not rely on `is_wire()` passing after passing it through `SigMap`

whitequark (approved):

whitequark:
Thanks!

modwizcode:
Awesome!

PR #2548, opened 2021-01-18
verilog: support for hierarchcial connections

### LINKED ISSUE #647: Internal module wires are not accessible using dot notation
The following code...

```verilog
module foo();
  wire test;
endmodule

module bar();
  foo myfoo();
  wire x = myfoo.test;
endmodule
```
...when read with yosys...
```
yosys -q -h "read_verilog" test.v
```
...generates with the following warning:
```
test.v:8: Warning: Identifier `\myfoo.test' is implicitly declared.
```

## Expected behavior

This code should read cleanly.

## Background

This issue was prompted by the method described @ZipCPU 's article: https://zipcpu.com/blog/2018/07/06/afifo.html .

The article explains how to use formal methods to verify an asychronous FIFO described in this artcle: http://www.sunburst-design.com/papers/CummingsSNUG2002SJ_FIFO1.pdf

The FIFO design uses a sub-modules to wrap up address generation RTL that is repeated twice: once for the read-side, and once for the write-side.

Unfortunately, because the yosys verilog front-end can't access internal wires, @ZipCPU had to unroll the sub-modules into the parent module so that he could access these hidden signals.

An alternative would be to expose these wires as ports on the sub-module. However, for more complex deeply nested modules this may become rather impractical.

zachjs:
References from [System]Verilog sources to internal module wires are now supported. Special autowires are automatically created and later resolved by the hierarchy pass, which turns them into appropriately dimensions wires with the `hierconn` attribute already supported downstream by the flatten pass. Because the hierarchy pass resolves modules from the top down, modules with automatic hierarchical connections are re-processed using a similar strategy to the one used to support SystemVerilog interfaces.

Fixes #647


whitequark (commented):
I'm a concerned about adding more virtual members to `RTLIL::Module` which is strictly specific to Verilog, since this creates more complexity for non-Verilog frontends. Is it really necessary for the caller to specify `hierconns` here, i.e. can't the reprocess function figure this out on its own? Should there really be separate `reprocess_module` and `reprocess_with_hierconns` entry points?

zachjs:
> I'm a concerned about adding more virtual members to `RTLIL::Module` which is strictly specific to Verilog, since this creates more complexity for non-Verilog frontends.

I went this route because it parallels the existing implementation of interfaces. We could use `dynamic_cast` instead, but note this would require including `ast.h` in `hierarchy.cc`. I agree that existing strategy is problematic given that these methods could pollute other frontends, although I'd hope the base this-is-not-supported implementation on `RTLIL::Module` would suffice for most (all?) other frontends.

> * Is it really necessary for the caller to specify `hierconns` here, i.e. can't the reprocess function figure this out on its own?

It is not necessary to pass that list. We could expose `lookup_hierconn` (likely to be renamed `resolve_hierconn_auto`) so it could be used in `ast.cc`

> * Should there really be separate `reprocess_module` and `reprocess_with_hierconns` entry points?

I'm not very familiar with the existing implementation of interfaces, although based on the what I have encountered working on sv2v and the limited existing test coverage, it may be in need of some attention. Mostly I was concerned with convoluting (and breaking) the implementations of both features. I would be happy to combine these functions. Either way, I wouldn't necessarily expect these automatic hierarchical connections to work if they refer to data within interface instances.

zachjs:
@whitequark Do you have any reservations regarding my suggestions above? If not, I'd be happy to take another pass at this PR.

whitequark:
@zachjs Your suggestions all sound reasonable to me. I believe the implementation of interfaces was not done with long term evolution of internal APIs in mind; I think it would be prudent to take more care of this now.

GenericNerdyUsername:
#2716 has been merged, is the newer version around but not linked?

PR #2549, opened 2021-01-20
adding support for passing multiple liberty files to abc

pgadfort:
This change adds support for passing multiple liberty files to abc. This is very useful when working on larger designs with multiple sources of intellectual property, such as memories and pad libraries.
The syntax for abc does not change, except for supporting additional flags: `abc -D 10000 -liberty lib1.lib -liberty lib2.lib` which translates into multiple `read_lib` commands in the resultant abc script.


whitequark (approved):
LGTM

PR #2550, opened 2021-01-20
verilog: allow spaces in macro arguments

zachjs:
@rswarbrick @clairexen

### DIFF
--- a/frontends/verilog/preproc.cc
+++ b/frontends/verilog/preproc.cc
@@ -392,7 +392,6 @@ static bool read_argument(std::string &dest)
 {
 	std::vector<char> openers;
 	for (;;) {
-		skip_spaces();
 		std::string tok = next_token(true);
 		if (tok == ")") {
 			if (openers.empty())
--- a/tests/simple/macro_arg_spaces.sv
+++ b/tests/simple/macro_arg_spaces.sv
@@ -1,0 +1,28 @@
+module top(
+	input wire [31:0] i,
+	output wire [31:0] x, y, z
+);
+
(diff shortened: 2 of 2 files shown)


whitequark (approved):

Xiretza:
I don't have a verilog LRM, so I don't know if this behavious is correct, but this introduced a breaking change:

```verilog
module test;
  `define MACRO(dummy, x) IDENT_``x``

  wire IDENT_X;
  wire `MACRO(_,Y);
  wire `MACRO(_, Z);
endmodule
```

The first two wires work as expected, the second fails with `ERROR: syntax error, unexpected TOK_ID, expecting ',' or '=' or ';' or '['` because it expands to `wire IDENT_ Z;` (notice the space).

This breaks the [techmap in symbiflow-arch-defs](https://github.com/SymbiFlow/symbiflow-arch-defs/blob/master/xc/xc7/techmap/cells_map.v#L1610), I'll report it over there if yosys' new behaviour is deemed to be correct.

zachjs:
@Xiretza If you're interested, you may be able to get a PDF copy of the specifications at:
https://ieeexplore.ieee.org/servlet/opac?punumber=8299593
https://ieeexplore.ieee.org/servlet/opac?punumber=10779

Xiretza:
> you may be able to get a PDF copy of the specifications

Thanks, but IEEE documents are a bit expensive for my taste ;)

whitequark:
@Xiretza [here](https://github.com/YosysHQ/yosys/files/5889986/verilog.zip)

PR #2551, opened 2021-01-20
sv: fix support wire and var data type modifiers

zachjs:
Fixes issue introduced in #2312


mmicko:
@zachjs seams there are still some broken cases, please take a look at https://github.com/YosysHQ/SymbiYosys/blob/master/docs/examples/puzzles/djb2hash.sv#L3 it is mentioned in issue https://github.com/YosysHQ/SymbiYosys/issues/119

PR #2554, opened 2021-01-21
Fix digit-formatting calculation for small numbers.

hzeller:
Calling log10() on zero causes a non-sensical value to be calculated. On some
compile options, I've observed yosys crashing with an illegal
instruction (SIGILL).

To make it safe, fix the calculation to do a range check; wrap it a
decimal_digits() function, and use it where the previous ceil(log10(n)) call
was used. As a side, it also improves readability.

Signed-off-by: Henner Zeller <h.zeller@acm.org>

### DIFF
--- a/frontends/aiger/aigerparse.cc
+++ b/frontends/aiger/aigerparse.cc
@@ -55,6 +55,10 @@ inline int32_t from_big_endian(int32_t i32) {
 #define log_debug2(...) ;
 //#define log_debug2(...) log_debug(__VA_ARGS__)
 
+static int decimal_digits(unsigned n) {
+	return n > 1 ? ceil(log10(n)) : 1;
+}
+
 struct ConstEvalAig
 {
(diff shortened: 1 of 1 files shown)


hzeller:
This should be an easy merge.

whitequark:
It's not clear to me that floating point was appropriate here in first place, so I won't sign off on your change either.

hzeller:
If there a simpler or more readable way to determine the number of decimal digits than with `log10()`, then I am happy to change the implementation `decimal_digits()`.

This change is at least fixing the broken use of log10 and leaves the place where things can be implemented to a single spot.

hzeller:
I replaced the floating point implementation with an integer one @whitequark 
(It's a bit longer in the implementation but in a microbenchmark it turns out to be about 10x faster)

whitequark (approved):
LGTM, sorry for the delay (I didn't get a notification).

PR #2560, opened 2021-01-25
xilinx_dffopt: Don't crash on missing IS_*_INVERTED.

### LINKED ISSUE #2559: XILINX_DFFOPT throws out_or_range exception
When I synthesize my design, yosys throws:
```
  what():  dict::at()
/opt/symbiflow/xc7/install/bin/symbiflow_synth: line 107: 168617 Aborted                 (core dumped) yosys -p "tcl ${SYNTH_TCL_PATH}" -l $LOG ${VERILOG_FILES[*]}
make: *** [Makefile:22: build/top.eblif] Error 134

```

I was able to track down the issue to file techlibs/xilinx/xilinx_dffopt.cc (line 212). Moreover, the issue is provoked when the cell is of type `FDRE_1` which makes sense because the cells_sim.v file does not have the IS_x_INVERTED parameters for this module. I was able to fix this by adding the parameters from `FDRE` to `FDRE_1`.

Unfortunately, I don't know the purpose for the _1 modules. Should they also contain the inverted parameters?
I also found https://github.com/YosysHQ/yosys/pull/2202 which might fix the issue?

I use the Symbiflow scripts but built yosys from scratch and replaced the binary and cells_sim.v

```
cd build && symbiflow_synth -t Hydrogen1 -v ../Hydrogen1.v -d artix7 -p xc7a35tftg256-1
```

[design.zip](https://github.com/YosysHQ/yosys/files/5863062/design.zip)

## Expected behavior

-

## Actual behavior

-

mwkmwkmwk:
The presence of `IS_*_INVERTED` on `FD*` cells follows Vivado, which apparently has been decided by a dice roll.  Just assume false if the
parameter doesn't exist.

Fixes #2559.

### DIFF
--- a/techlibs/xilinx/cells_sim.v
+++ b/techlibs/xilinx/cells_sim.v
@@ -633,6 +633,41 @@ module FDRSE (
       Q <= d;
 endmodule
 
+module FDRSE_1 (
+  output reg Q,
+  (* clkbuf_sink *)
+  (* invertible_pin = "IS_C_INVERTED" *)
+  input C,
+  (* invertible_pin = "IS_CE_INVERTED" *)
+  input CE,
+  (* invertible_pin = "IS_D_INVERTED" *)
+  input D,
+  (* invertible_pin = "IS_R_INVERTED" *)
+  input R,
+  (* invertible_pin = "IS_S_INVERTED" *)
+  input S
+);
+  parameter [0:0] INIT = 1'b0;
+  parameter [0:0] IS_C_INVERTED = 1'b0;
+  parameter [0:0] IS_CE_INVERTED = 1'b0;
+  parameter [0:0] IS_D_INVERTED = 1'b0;
+  parameter [0:0] IS_R_INVERTED = 1'b0;
+  parameter [0:0] IS_S_INVERTED = 1'b0;
+  initial Q <= INIT;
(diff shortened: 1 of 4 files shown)


techlibs/xilinx/cells_sim.v
```diff
@@ -633,6 +633,41 @@ module FDRSE (
...
+  input CE,
+  (* invertible_pin = "IS_D_INVERTED" *)
+  input D,
+  (* invertible_pin = "IS_R_INVERTED" *)
+  input R,
+  (* invertible_pin = "IS_S_INVERTED" *)
+  input S
+);
+  parameter [0:0] INIT = 1'b0;
+  parameter [0:0] IS_C_INVERTED = 1'b0;
```
dnltz, line 651: Should we also add these parameters for example to `FDSE_1`?

mwkmwkmwk: Not now.  The definition for these primitives follows Vivado, and while Vivado is wildly inconsistent here, using different parameter sets than vendor toolchain is asking for trouble.  We may reconsider that approach, but that requires some more thinking.

dnltz: okay. Thanks for action that fast :)

PR #2564, opened 2021-01-26
flatten: clarify confusing error message

whitequark:
This error does not always indicate any mismatch in directionality or any problem with the cell at all. For example, in this case, which was extracted from a large codebase where a developer had mistakenly left the second `assign` after some modifications:
```verilog
module m1(...);
	output wire x;
	assign x = 1'b0;
endmodule

module m2(...);
	wire y;
	assign y = 1'b0; // this was left here by mistake
	m1 c(y);
endmodule
```

The new message is still not a perfect solution (which would be to detect multiple drivers properly, something I wanted to see in Yosys for years now) but it is at least not flat out wrong.


clairexen (approved):
LGTM

Issue #2565, opened 2021-01-27
Yosys doesn't allow some delay specifications


TimRudy:
I have Verilog that I want to be common & the same for simulation and synthesis, but there's an incompatibility when Yosys gets to this #DELAY in procedural continuous assignment.

  ```
  assign #(DELAY_RISE, DELAY_FALL) Y = computed;
  ```
  ### Actual behavior:
  
  ERROR: syntax error, unexpected ','
  
  ### Expected behavior:
  
  I guess Yosys is supposed to throw away the DELAY specification (it's not for synthesis). It accepts a single DELAY value no problem. Can we enhance it to accept a pair or a triplet; that is, with commas as below in **A.2.2.3**?
  
  Incidentally Yosys also accepts:
  
  ```
  assign #(DELAY_RISE : DELAY_RISE : DELAY_FALL) Y = computed;
  ```
  ...this is min-typical-max form, which is different and isn't for my use case.
  
  ### Workaround:
  
  I've not had a need to use `ifdef SYNTHESIS so far. But without fixing this front-end parsing, I might have to use it.
  
  ### Background:
  
  I looked in the LRM. (Yes, I must be enthused about this.) Some technical definitions:
  
  ```
  delay3 ::= (From Annex A - A.2.2.3)
  # delay_value | # ( delay_value [ , delay_value [ , delay_value ] ] )
  delay2 ::=
  # delay_value | # ( delay_value [ , delay_value ] )
  delay_value ::=
    unsigned_number
  | parameter_identifier
  | specparam_identifier
  | mintypmax_expression
  ```
  ```
  mintypmax_expression ::=
    expression
  | expression : expression : expression
  ```
  
  Also we can see the reason for existing constraint in a front-end parser file:
  
  ```
  non_opt_delay:
	  '#' TOK_ID { delete $2; } |
	  '#' TOK_CONSTVAL { delete $2; } |
	  '#' TOK_REALVAL { delete $2; } |
	  '#' '(' expr ')' { delete $3; } |
	  '#' '(' expr ':' expr ':' expr ')' { delete $3; delete $5; delete $7; };
  ```

  There's no good reason for rejecting the delay2 or delay3 format, is there?
  
  _If you're good with it, do you want me to do the PR?_ Thanks

whitequark:
A PR for this is welcome!

PR #2566, opened 2021-01-27
Extend "delay" expressions to handle pair and triplet, i.e. rise, fall and turn-off

TimRudy:
- Per LRM A.2.2.3
- Not for synthesis, just for Verilog acceptance by front-end

Resolves YosysHQ/yosys#2565

### DIFF
--- a/frontends/verilog/verilog_parser.y
+++ b/frontends/verilog/verilog_parser.y
@@ -609,12 +609,17 @@ interface_body_stmt:
 	param_decl | localparam_decl | typedef_decl | defparam_decl | wire_decl | always_stmt | assign_stmt |
 	modport_stmt;
 
+mintypmax_expr:
+	expr { delete $1; } |
+	expr ':' expr ':' expr { delete $1; delete $3; delete $5; };
+
 non_opt_delay:
 	'#' TOK_ID { delete $2; } |
 	'#' TOK_CONSTVAL { delete $2; } |
 	'#' TOK_REALVAL { delete $2; } |
-	'#' '(' expr ')' { delete $3; } |
-	'#' '(' expr ':' expr ':' expr ')' { delete $3; delete $5; delete $7; };
+	'#' '(' mintypmax_expr ')' |
+	'#' '(' mintypmax_expr ',' mintypmax_expr ')' |
+	'#' '(' mintypmax_expr ',' mintypmax_expr ',' mintypmax_expr ')';
 
 delay:
 	non_opt_delay | %empty;
--- a/tests/verilog/delay_mintypmax.ys
+++ b/tests/verilog/delay_mintypmax.ys
@@ -1,0 +1,213 @@
+logger -expect-no-warnings
+read_verilog <<EOT
+module test (a, b, c, y);
+    input a;
+    input signed [1:0] b;
+    input signed [2:0] c;
+    output y;
+    assign #(12.5 : 14.5 : 20) y = ^(a ? b : c);
+endmodule
+EOT
+
+design -reset
+logger -expect-no-warnings
+read_verilog << EOT
+module test (input [7:0] a, b, c, d, output [7:0] x, y, z);
+    assign #(20:20:25) x = a + b, y = b + c, z = c + d;
+endmodule
+EOT
+
+design -reset
+logger -expect-no-warnings
+read_verilog << EOT
+module test (input [7:0] a, b, c, d, output [7:0] x, y, z);
+    assign #(20:20:25, 40:45:50, 60:65:75) x = a + b, y = b + c, z = c + d;
+endmodule
+EOT
+
+design -reset
+logger -expect-no-warnings
+read_verilog <<EOT
+module test (a, b, c, y);
+    localparam TIME_STEP = 0.011;
+    input signed [3:0] a;
+    input signed [1:0] b;
+    input signed [1:0] c;
+    output [5:0] y;
+    assign #(TIME_STEP:TIME_STEP:TIME_STEP) y = (a >> b) >>> c;
+endmodule
+EOT
+
+design -reset
+logger -expect-no-warnings
+read_verilog <<EOT
+module test;
+    wire o, a, b;
+    and #(1:2:3, 4:5:6) and_gate (o, a, b);
+    wire #(1:2:3, 4:5:6, 7:8:9) x;
+    assign o = x;
+endmodule
+EOT
+
+design -reset
+logger -expect-no-warnings
+read_verilog <<EOT
+module test;
+    localparam TIME_TYP = 0.7;
+    wire o, a, b;
+    and #(0:TIME_TYP:2) and_gate (o, a, b);
+    wire #(2:TIME_TYP:4) x;
+    assign o = x;
+endmodule
+EOT
+
+design -reset
+logger -expect warning "Yosys has only limited support for tri-state logic at the moment." 1
+read_verilog <<EOT
+module test (input en, input a, input b, output c);
+    wire [15:0] add0_res = a + b;
+    assign #(15:20:30) c = (en) ? a : 1'bz;
+endmodule
+EOT
+
+design -reset
+logger -expect-no-warnings
+read_verilog <<EOT
+module test (input en, d, t_min, t, t_max);
+    reg o;
+    always @*
+        if (en)
+            o = #(t_min : t : t_max, t_min : t : t_max) ~d;
+endmodule
+EOT
+
+design -reset
+logger -expect-no-warnings
+read_verilog <<EOT
+module test #(parameter DELAY_RISE = 0, DELAY_FALL = 0, DELAY_Z = 0)
+            (input clock, input reset, input req_0, input req_1, output gnt_0, output gnt_1);
+    parameter SIZE = 3;
+    parameter IDLE = 3'b001, GNT0 = 3'b010, GNT1 = 3'b100;
+    reg [SIZE-1:0] state;
+    reg [SIZE-1:0] next_state;
+    reg gnt_0, gnt_1;
+
+    always @ (state or req_0 or req_1)
+    begin : FSM_COMBO
+        next_state = 3'b000;
+        case (state)
+        IDLE : if (req_0 == 1'b1) begin
+                    next_state = #(DELAY_RISE-1 : DELAY_RISE : DELAY_RISE+1) GNT0;
+                end else if (req_1 == 1'b1) begin
+                    next_state = #(DELAY_FALL/1.2 : DELAY_FALL : DELAY_FALL*2.5) GNT1;
+                end else begin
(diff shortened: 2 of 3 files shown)


TimRudy:
Thanks Prof whitequark! Well that was pretty easy. Do you know if propagating new `non_opt_delay` formats -> through `delay` -> all through the language, such as `simple_behavioral_stmt`, `behavioral_stmt`, `wire_decl`, `assign_stmt`, `cell_stmt` will be transparent and not affect the system?

whitequark:
@zachjs Could you please take a look at this PR? I don't understand the relevant parts of the LRM well enough.

eddiehung:
Welcome @TimRudy! Would you be able to add a quick test to (a) demonstrate its usage, and (b) make sure it doesn't get broken in the future?

zachjs:
Relevant portions of the SV grammar: (`repeat`, `#`, `(`, `)`, `:`, and `,` are terminals)
```
delay3 ::= #delay_value | #(mintypmax_expression[,mintypmax_expression[,mintypmax_expression]])
delay2 ::= #delay_value | #(mintypmax_expression[,mintypmax_expression])
delay_value
  ::= unsigned_number
  | real_number
  | ps_identifier
  | time_literal
  | 1step
delay_control
  ::= # delay_value 
  | # ( mintypmax_expression )
mintypmax_expression
  ::= expression 
  | expression : expression : expression
delay_or_event_control
  ::= delay_control 
  | event_control
   | repeat ( expression ) event_control 
```

So you can have something like
```SystemVerilog
module top;
  wire o, a, b;
  and #(1:2:3, 4:5:6) and_gate (o, a, b);
  wire #(1:2:3, 4:5:6, 7:8:9) x;
  assign x = o;
endmodule
```

* `delay_control` and `delay3` can both be used in net declarations and continuous assignments
* `delay_control` and `event_control` are both allowed to guard procedural statements, as is the more obscure (to me) `cycle_delay`
* `delay_or_event_control` is used for procedural assignments
* `delay2` is used in primitives (`and`, `nor`, etc.) gate declarations

Both with and without these additions, there are false-positive and false-negative parses with the current grammar. That said, these delays are naturally dropped for synthesis, so it may not be worth worrying about the false-positives. I have no objections to merging this PR for now and revising our grammar for delays down the line. I can also take a stab at this in the near term if desired.

TimRudy:
My enthusiasm for open source dropped off a cliff as cygwin and building was not smooth, I was on StackOverflow which is not a good karma for a weekend passtime. (However after a little segmentation violation scare in an ABC file - solution, try again - C/C++ compilation is like riding a bike? - there is truly a lot of building done now.) To get yosys to finish I need a fresh build of iverilog from Git - but the iverilog build is giving me the bigger problems. ... OK, I did have version 10.1 installed and I seem to have found an 11 of it for Windows at http://bleyer.org/icarus, that's gotta work, it's recent.

Back to you about those tests soon

udif:
@TimRudy Have you considered using WSL2?

TimRudy:
Hi guys,
I've done `make test` and overcame a problem which was: my Windows path happened to have a space in it. Yes! This provides a chance to improve Yosys (platform builds). Should I open an issue?

It comes down to using `cygpath` to make those Windows short paths, e.g. looks like:
> Parsing Verilog input from `/cygdrive/c/LEVEL1~1/Software/yosys/share/techmap.v'

In autotest.sh I put:
autotest

Needs an issue, I guess: What is best way to do these kind of things in platform-independent .sh files?
With ifndef __CYGWIN__ or ifdef __WIN32__ in supporting .c files?

Any comments on that welcome.

TimRudy:
Then another issue I'll need to bring up: For me to use `gcc` (in cygwin context), apparently I needed to make targeted change in Makefile:
Makefile

What do you think of that? Trying to help :-) - and willing to do some legwork to improve things in the corners so convenient for strange, odd users

zachjs:
@TimRudy I don't think I'm qualified to comment on the build/script issues mentioned above. However, I'm interested in moving forward with this change. If you're currently blocked, I'd also be happy to take a stab at supporting these and other delay variants and add some test coverage.

TimRudy:
Thanks - My next step is to clone a file like bug2037.ys in verilog test
```
logger -expect-no-warnings
read_verilog <<EOT

```and write a number of examples of delays in the grammer - starting with only what we would call legitimate verilog-compliant

There are non-compliant constructs that you referred to above, I understand, and if tests were written for those they would pass, but would indicate Yosys is allowing weird delays, right? I will do the above basics, some tonight, OK?

**Comment by zachjs** [comment]
I think it's fine to not worry about coverage for the "negative" cases for the time being. I can add coverage in a future PR once the grammar includes more of the positive cases I described above.

**Comment by TimRudy** [comment]
Yeah I can tell you're interested in tweaking the grammar, so it breaks out the delay specs a bit... That would be cool, I thought about it but will leave that stuff to you

**Comment by TimRudy** [comment]
Sorry Zach, I couldn't run tests in a timely way on my machine, so will check in again on the weekend

**Comment by TimRudy** [comment]
I'll push some tests tomorrow, I got the tests to run & pass

**Comment by TimRudy** [comment]
See if you can uncover more places in the grammar, I didn't find tons of places
- The `delay2` that you suggested is not being accepted `wire #(1:2:3, 4:5:6, 7:8:9) x;`
- I also tried `time_literal` like 10us and I tried `1step` as a token and those are not accepted
- Does `repeat ( expression ) event_control` come into our delay considerations? I tried `repeat (3) @(posedge clk);` that I thought is fine and it syntax errored at `@`
- Note `always #updatePeriod` at delay_risefall.ys line 147 - I didn't explore that very much (forgot to see if it extends as a delay expression)
- Maybe I missed cases of this (RHS delay, not a continuous assign): `y = #(12 : 15 : 20) ^(a ? b : c);`
- I heard a delay could work in an initializer, example at line 49 of either file: `wire [15:0] add0_res = #(15 : 20 : 25) a+b;` (but it was not accepted)

Finally, I didn't do SV test - Cheers

**Comment by zachjs** [comment]
Thanks for adding test coverage and highlighting some of the other cases we should look into in the future. Would you mind squashing so the tests are added in the same commit as the grammar changes? Otherwise, LGTM.

**Comment by TimRudy** [comment]
More missing expressions added to grammar, may help even more people avoid syntax errors. If that's OK, let me know and I'll send the squash commit

File: frontends/verilog/verilog_parser.y [hunk 1]
```diff
@@ -616,7 +616,9 @@ non_opt_delay:
 	'#' '(' expr ')' { delete $3; } |
 	'#' '(' expr ',' expr ')' { delete $3; delete $5; } |
 	'#' '(' expr ',' expr ',' expr ')' { delete $3; delete $5; delete $7; } |
-	'#' '(' expr ':' expr ':' expr ')' { delete $3; delete $5; delete $7; };
+	'#' '(' expr ':' expr ':' expr ')' { delete $3; delete $5; delete $7; } |
+	'#' '(' expr ':' expr ':' expr ',' expr ':' expr ':' expr ')' { delete $3; delete $5; delete $7; delete $9; delete $11; delete $13; } |
+	'#' '(' expr ':' expr ':' expr ',' expr ':' expr ':' expr ',' expr ':' expr ':' expr ')' { delete $3; delete $5; delete $7; delete $9; delete $11; delete $13; delete $15; delete $17; delete $19; };
```
zachjs, line 621: My reading of the grammar is that a delay like `#(A, A:B:C, D)` should be allowed. To accomplish this and clean things up a bit, I think it would be worthwhile to split out `mintypmax_expr` into its own rule with productions for `expr` and `expr ':' expr ':' expr` (per the official grammar), and have those productions delete the expressions. Then the non-`delay_value` productions in `non_opt_delay` can be something like:
```
'#' '(' mintypmax_expr ')'
'#' '(' mintypmax_expr ',' mintypmax_expr ')'
'#' '(' mintypmax_expr ',' mintypmax_expr ',' mintypmax_expr ')'
```

TimRudy: Thanks for giving me the chance but I'm spread pretty thin, I'll revisit this if you haven't done it...!

TimRudy: Is delay_risefall.ys line 168 & 170 legitimate? Interestingly it reveals a case when a min-typ-max expression is accepted but general delay expression is not. I'll send a squash when needed

zachjs:
LGTM, let's get this squashed and I'll merge it!

frontends/verilog/verilog_parser.y
zachjs: I believe that test case is legal. I believe the LRM would allow for the parentheses to be removed, but Yosys does not currently support this. Adding support in Yosys isn't simple (or necessary for the time being) as it would introduce conflicts in the grammar.

TimRudy: OK, I don't have much of a comment (haven't used much Verilog). Seems like we went above & beyond. See you on next project.

zachjs:
Can you squash these commits before we merge? This can help make log/blame more useful on popular files like these.

eddiehung:
I see that "squash and merge" is the default setting now. Is there a need to do it on the source branch too?

TimRudy:
Does my head commit look like a squash or like a merge? Tell me if it's good to go, thanks Eddie

whitequark:
> I see that "squash and merge" is the default setting now.

It's only the default for you (it's per repo/user combination).

> Is there a need to do it on the source branch too?

You can only squash *all* commits with the GitHub merge button, but sometimes it makes sense to squash *some*. (Not sure which category this PR is in.)

eddiehung:
> > I see that "squash and merge" is the default setting now.
> 
> It's only the default for you (it's per repo/user combination).

Ooh I didn't know that, it must have remembered my last setting. I would assume that it could be enforced at the repo-wide level if one wanted. I think it's the right thing to do now (despite being careless and not doing so in the past and getting my commit count up!)

> > Is there a need to do it on the source branch too?
> 
> You can only squash _all_ commits with the GitHub merge button, but sometimes it makes sense to squash _some_. (Not sure which category this PR is in.)

I would say that when you ask a contributor to squash, they will typically squash all commits into one. Clicking the button could save that round trip.

zachjs:
It is possible to individually enable or disable each of the three merge options (merge, squash, rebase) at the repo level. While squashing into a single commit is often appropriate, I don't think we should require squashing for all changes. I personally prefer rebase-only, but this would be a significant departure for this project. I also wouldn't mind using GitHub's squash for this and other changes, so long as we are comfortable with results.

whitequark:
I agree with @zachjs here.

zachjs:
Barring any objections, I will generally opt for squashing change sets like this one. Note that the author of the squashed commit will likely be the creator of the PR, so if a PR is opened on behalf of another author, we may want to manually squash instead. I also think using rebases rather than merge commits is sometimes preferred, so we may consider trying that out in the near-term and disabling the "standard merge" option if we like it long-term.

zachjs (approved):

Issue #2058, opened 2020-05-15
Fix 0-WIDTH wire handling in techmap rules.


mwkmwkmwk:
#2027 fixed broken handling of `[-1:0]` wires that a lot of our techmap rules relied on for 0-width input handling — several cases of this were already fixed in #2027 itself, but more problems remain (eg. https://github.com/YosysHQ/yosys/commit/ee6e51e20e39dbaaed032d43697532eabd4fe8ff — reported on the IRC channel). They need to be fixed, perhaps with help of some verilog extension to better manage 0-width wires.

PR #2567, opened 2021-01-27
frontend: json: parse negative values

kgugala:
When scc breaks logic loops with `$__ABC9_SCC_BREAKER` module it adds the following entry to resulting json:

```"$__ABC9_SCC_BREAKER": {
  "attributes": {
    "blackbox": "00000000000000000000000000000001",
    "cells_not_processed": "00000000000000000000000000000001",
    "dynports": "00000000000000000000000000000001",
    "src": "/(...)/share/yosys/abc9_model.v:9.1-11.10"
  },  
  "cells": {}, 
  "netnames": {
    "I": {
      "attributes": {
        "src": "/(...)/share/yosys/abc9_model.v:9.47-9.48"
      },  
      "bits": [
        2,  
        3   
      ],  
      "hide_name": 0,
      "offset": -1, 
      "upto": 1
    }, 
```
reading the json back to yosys results in:
```
ERROR: Unexpected character in JSON file: '-'
```
in the `"offset": -1,` line

This PR fixes that

### DIFF
--- a/frontends/json/jsonparse.cc
+++ b/frontends/json/jsonparse.cc
@@ -72,10 +72,17 @@ struct JsonNode
 				break;
 			}
 
-			if ('0' <= ch && ch <= '9')
+			if (('0' <= ch && ch <= '9') || ch == '-')
 			{
+				bool negative = false;
 				type = 'N';
-				data_number = ch - '0';
+				if (ch == '-') {
+					data_number = 0;
+				       	negative = true;
+				} else {
+					data_number = ch - '0';
+				}
+
 				data_string += ch;
 
 				while (1)
@@ -97,6 +104,7 @@ struct JsonNode
 					data_string += ch;
 				}
 
+				data_number = negative ? -data_number : data_number;
 				data_string = "";
 				break;
 


whitequark (commented):
So what happens if the input is something like `123-456`?

kgugala:
> So what happens if the input is something like `123-456`?

I'll break here https://github.com/YosysHQ/yosys/blob/master/frontends/json/jsonparse.cc#L91

whitequark:
Actually, is this even the right approach? Is there a `(* force_downto *)` missing somewhere, like in #2058? cc @mwkmwkmwk

Ravenslofty:
I think we should ping @eddiehung in here too, since this is from the eddie/abc9_scc_fixes2 branch, and it would be a good idea to roll it up into his next round of ABC9 fixes.

mwkmwkmwk:
> Actually, is this even the right approach? Is there a `(* force_downto *)` missing somewhere, like in #2058? cc @mwkmwkmwk

Right or not, negative offsets are valid in Verilog either way, so this patch should be merged regardless.

mwkmwkmwk:
As in, currently converting the following to JSON and trying to read it back just fails:

```
module top(input [3:-3] I);
endmodule
```

frontends/json/jsonparse.cc
```diff
@@ -72,10 +72,17 @@ struct JsonNode
...
 
-			if ('0' <= ch && ch <= '9')
+			if (('0' <= ch && ch <= '9') || ch == '-')
 			{
+				bool negative = false;
 				type = 'N';
-				data_number = ch - '0';
+				if (ch == '-') {
+					data_number = 0;
+				       	negative = true;
```
mwkmwkmwk, line 81: whitespace error.

mwkmwkmwk (approved):

eddiehung:
I'm not sure the `$__ABC9_SCC_BREAKER` module nor `abc9` is relevant here. For one, I think other modules can trigger negative indices, plus, that cell should never leak into the outside world as it's an internal cell used between sub-passes.

PR #2572, opened 2021-02-04
verilog_parser: add label check to gen_block

kamilrakoczy:
This PR adds a check for begin/end labels mismatches in genblocks - this is similar to what is currently done for behavrioral_stmt blocks

### DIFF
--- a/tests/verilog/block_labels.ys
+++ b/tests/verilog/block_labels.ys
@@ -1,0 +1,26 @@
+read_verilog <<EOT
+module foo;
+
+	genvar a = 0;
+	for (a = 0; a < 10; a++) begin : a
+	end : a
+endmodule
+EOT
+read_verilog <<EOT
+module foo2;
+
+	genvar a = 0;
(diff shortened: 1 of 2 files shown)


tests/verilog/block_labels.ys
```diff
@@ -0,0 +1,26 @@
...
+read_verilog <<EOT
+module foo2;
+
+	genvar a = 0;
+	for (a = 0; a < 10; a++) begin : a
+	end
+endmodule
+EOT
+
+logger -expect error "Begin label (.*) and end label (.*) don't match." 1
```
zachjs, line 18: Minor nit: Can the pattern match `a` and `b` specifically, rather than `.*`? Also note that the `.` at the end of this pattern is being treated as a wildcard, so perhaps either escape it or remove it.

kamilrakoczy: Done. I also escaped ``.`` at the end.

zachjs (approved):

PR #2573, opened 2021-02-04
verilog: refactored constant function evaluation

### LINKED ISSUE #2562: Apparent bug in const checking logic
To reproduce this issue, check out [Ibex](https://github.com/lowRISC/ibex), e.g. version b295f07b (although this shouldn't be version-sensitive) and then run the synthesis flow. The instructions in `syn/README.md` should work.

The synthesis flow works by running sv2v and then running Yosys on the output. With the current head of the Yosys master branch, you get the following error message:
```
out/generated/ibex_cs_registers.v:544: ERROR: Non-constant function call in constant expression.
```
The relevant code looks like this:
```systemverilog
mstatus_d = {
   ... // some other elements in the concatenation
   sv2v_cast_2(sv2v_cast_2(csr_wdata_int[ibex_pkg_CSR_MSTATUS_MPP_BIT_HIGH:ibex_pkg_CSR_MSTATUS_MPP_BIT_LOW])),
   ... // some other elements in the concatenation
};
```
Here, `sv2v_cast_2` is defined as follows:
```systemverilog
	function automatic [1:0] sv2v_cast_2;
		input reg [1:0] inp;
		sv2v_cast_2 = inp;
	endfunction
```

Bisecting Yosys builds points to commit 2085d9a.

If you manually hack the generated code to replace the argument to the inner `sv2v_cast_2` with a constant like `8'b0` OR (more confusingly) remove one layer of `sv2v_cast_2` calls, the error message goes away.

@zachjs (since this was your change, and you probably also understand the output from sv2v better than me!)

zachjs:
Elaboration now attempts constant evaluation of any function call with only constant arguments, regardless of the context or contents of the function. This removes the concept of "recommended constant evaluation" which previously applied to functions with `for` loops or which were (sometimes erroneously) identified as recursive. Any function call in a constant context (e.g., `localparam`) or which contains a constant-only procedural construct (`while` or `repeat`) in its body will fail as before if constant evaluation does not succeed.

Fixes #2562

### DIFF
--- a/frontends/ast/ast.h
+++ b/frontends/ast/ast.h
@@ -264,10 +264,9 @@ namespace AST
 
 		// additional functionality for evaluating constant functions
 		struct varinfo_t { RTLIL::Const val; int offset; bool is_signed; };
-		bool has_const_only_constructs(bool &recommend_const_eval);
-		bool has_const_only_constructs(std::set<std::string>& visited, bool &recommend_const_eval);
-		void replace_variables(std::map<std::string, varinfo_t> &variables, AstNode *fcall);
-		AstNode *eval_const_function(AstNode *fcall);
+		bool has_const_only_constructs();
+		bool replace_variables(std::map<std::string, varinfo_t> &variables, AstNode *fcall, bool must_succeed);
+		AstNode *eval_const_function(AstNode *fcall, bool must_succeed);
 		bool is_simple_const_expr();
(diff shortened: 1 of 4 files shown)


whitequark (approved):
Looks excellent--I'll merge this unless there are objections raised.

zachjs:
@whitequark Do we feel comfortable getting this merged? Although it's been LGTM'd, I wanted to double check.

PR #2580, opened 2021-02-10
Set aside extraneous tests in simple_abc9 test suite

zachjs:
New test cases on one branch may be automatically copied from simple/ to simple_abc9/, causing failures when switching to another branch. This updates the simple_abc9 script to set aside extraneous tests in a non-destructive way.


clairexen (approved):
LGTM

PR #2582, opened 2021-02-11
`synth_machxo2` and MachXO2 techmap/sim models

cr1901:
This PR introduces the `synth_machxo2` pass and MachXO2 techmap (`cells_map.v`) and simulation (`cells_sim.v`) models. My changes here complement the `nextpnr-machxo2` backend which as of last week can generate basic bitstreams such as blinky or UART. Passing tests for functionality that's ready has been provided as well.

This PR is essentially the minimum required functionality to map from `yosys`' built-in cells to something `nextpnr-machxo2` can handle. In particular:

* Only one type of DFF is supported via `dfflegalize` and there are no provisions yet for mapping from Diamond FFs to `FACADE_FF` equivalents.
* At some, `clkbufmap` pass will be used to add global buffers, but they must be manually instantiated for now (and may not 
  even work in `nextpnr-machxo2` as of this writing).
* The `OSCH` `blackbox` primitive exists and was one of the only "extra" things I wrote to make it easier to synthesize basic designs :).


techlibs/machxo2/cells_sim.v
```diff
@@ -0,0 +1,218 @@
...
+	wire [3:0] I_pd;
+
+	genvar ii;
+	generate
+		for (ii = 0; ii < 4; ii = ii + 1'b1)
+			assign I_pd[ii] = (I[ii] === 1'bz) ? 1'b0 : I[ii];
+	endgenerate
+
+	assign I = {D, C, B, A};
+	assign Z = INIT[I_pd];
```
gatecat, line 17: This form of LUT model is unnecessarily X-pessimistic compared to doing something like https://github.com/YosysHQ/yosys/blob/master/techlibs/ecp5/cells_sim.v#L6-L9

and I'm not sure if the LUT inputs do really float low when disconnected? I know the ECP5 hardware defaults to high but maybe we discussed this and xo2 was different.

cr1901: Confirmed via LED that LUT inputs float low when disconnected:

```
module top (
  (* LOC = "13" *)
  output Z
);

  LUT4 #(.INIT(16'b1)) my_lut(.Z(Z));
endmodule
```

When programmed to Tinyfpga AX, the LED turns on.

cr1901:
I have switched the LUT4 models to the one used in ECP5. This caused one of the tests (`mux.ys`) to be able to optimize away a LUT4, and the logs show a decent reduction in the variables/clauses considered during `equiv_induct`.

According to the logs, the `abc` pass generates 11 `\$lut`s instead of 12 now for `mux16`, which causes a test failure. I'm honestly not sure how the change in models caused `abc` to optimize out a `\$lut`, considering the logs are mostly identical up to that point in the `mux16` test. However, seeing as `equiv_induct` passes, I'm guessing it's a valid optimization by `abc` that I don't understand yet. So I changed the test so it passes :).

mwkmwkmwk:
> I have switched the LUT4 models to the one used in ECP5. This caused one of the tests (`mux.ys`) to be able to optimize away a LUT4, and the logs show a decent reduction in the variables/clauses considered during `equiv_induct`.
> 
> According to the logs, the `abc` pass generates 11 `\$lut`s instead of 12 now for `mux16`, which causes a test failure. I'm honestly not sure how the change in models caused `abc` to optimize out a `\$lut`, considering the logs are mostly identical up to that point in the `mux16` test. However, seeing as `equiv_induct` passes, I'm guessing it's a valid optimization by `abc` that I don't understand yet. So I changed the test so it passes :).

In general, completely random unrelated things can perturb the IR in such a way that abc output changes — things as dumb as adding a comment line (because this changes source line attributes, and thus sorting in some places...).  This is also a good reason to *not* hardcode the exact output in tests, but instead use some reasonable min/max bounds on the amount of LUTs generated.

techlibs/machxo2/synth_machxo2.cc
```diff
@@ -0,0 +1,248 @@
...
+		log("        from label is synonymous to 'begin', and empty to label is\n");
+		log("        synonymous to the end of the command list.\n");
+		log("\n");
+		log("    -noflatten\n");
+		log("        do not flatten design before synthesis\n");
+		log("\n");
+		log("    -noiopad\n");
+		log("        do not insert IO buffers\n");
+		log("\n");
+		log("    -vpr\n");
```
mwkmwkmwk, line 66: Do you actually support VPR? If not, this should be removed.

cr1901: @mwkmwkmwk VPR support is forthcoming (as in, I agreed to implement it). Should it still be removed and put it back in once VPR support is implemented?

mwkmwkmwk: eh, fine in this case

PR #2586, opened 2021-02-12
verilog: support recursive functions using ternary expressions

### LINKED ISSUE #247: Segfault when using recursive functions
Here is a golfed example from a larger description I was writing:

``` verilog
// segfault.v
module segfault(output y);
   function get_xxx;
      input i;
      begin
           get_xxx = i == 0 ? 1 : get_xxx(0);
      end
   endfunction
   assign y = get_xxx(0);
endmodule
```

As I see it, there is enough information during compilation to unwrap this function. In fact, in this case recursion does not happen at all (it should return `1` immediately without reentering `get_xxx` function).

I run it like this:

```
yosys -p 'prep' segfault.v
```

Here is the output:

```
 Yosys 0.6 (git sha1 5869d26, gcc 6.2.0-4 -O2 -fdebug-prefix-map=/build/yosys-aIBmDV/yosys-0.6=. -fstack-protector-strong -fPIC -Os)


-- Parsing `segfault.v' using frontend `verilog' --

1. Executing Verilog-2005 frontend.
Parsing Verilog input from `segfault.v' to AST representation.
Generating RTLIL representation for module `\segfault'.
Segmentation fault
```

Am I doing something incorrectly? If so, maybe it is possible to implement a better failure mode for this case?

zachjs:
This adds a mechanism for marking certain portions of elaboration as occurring within unevaluated ternary branches. To enable elaboration of the overall ternary, this also adds width detection for these unelaborated function calls.

While this implementation is meant to be low-impact for prior inputs, I do think the introduction of `unevaluated_tern_branch` is not necessarily ideal. I'm eager for any feedback or alternative suggestions.

This fixes #247 (the original example uses a ternary in this way).

### DIFF
--- a/frontends/ast/ast.h
+++ b/frontends/ast/ast.h
@@ -270,6 +270,9 @@ namespace AST
 		bool is_simple_const_expr();
 		std::string process_format_str(const std::string &sformat, int next_arg, int stage, int width_hint, bool sign_hint);
 
+		bool is_recursive_function() const;
+		std::pair<AstNode*, AstNode*> get_tern_choice();
+
 		// create a human-readable text representation of the AST (for debugging)
 		void dumpAst(FILE *f, std::string indent) const;
 		void dumpVlog(FILE *f, std::string indent) const;
--- a/frontends/ast/genrtlil.cc
+++ b/frontends/ast/genrtlil.cc
@@ -944,6 +944,41 @@ void AstNode::detectSignWidthWorker(int &width_hint, bool &sign_hint, bool *foun
 			}
 			break;
 		}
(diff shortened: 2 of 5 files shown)


whitequark:
Related to #1651?

zachjs:
While this isn't closely related to #1651, which appears to only occur when mem2reg is used, it may be possible to use some of this machinery to accomplish the goal of #1651. This change does not attempt to "circumvent" the width detection, but rather avoid the infinite recursion in this specific edge case.

whitequark (approved):

whitequark:
Thanks!

PR #2591, opened 2021-02-18
verilog: error on macro invocations with missing argument lists

zachjs:
This would previously complain about an undefined internal macro if the unapplied macro had not already been used. If it had, it would incorrectly use the arguments from the previous invocation.

@rswarbrick

### DIFF
--- a/frontends/verilog/preproc.cc
+++ b/frontends/verilog/preproc.cc
@@ -477,7 +477,16 @@ static bool try_expand_macro(define_map_t &defines, std::string &tok)
 	std::string name = tok.substr(1);
 	std::string skipped_spaces = skip_spaces();
 	tok = next_token(false);
-	if (tok == "(" && body->has_args) {
+	if (body->has_args) {
+		if (tok != "(") {
+			if (tok.size() == 1 && iscntrl(tok[0])) {
+				char buf[5];
+				snprintf(buf, sizeof(buf), "\\x%02x", tok[0]);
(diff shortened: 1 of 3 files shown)


rswarbrick:
Oops, that definitely isn't what I intended it to do! What's the error message like for something like this?
```
`define foo(a=1) (a)

`foo
```
Looking at the patch quickly, it looks like you'll spit out something like `Expected to find '(' to begin macro arguments for 'foo', but instead found ''`. Maybe special-case the error message for end-of-line?

zachjs:
I've updated the error to special-case tokens which are control characters.

frontends/verilog/preproc.cc
```diff
@@ -477,7 +477,16 @@ static bool try_expand_macro(define_map_t &defines, std::string &tok)
 	std::string name = tok.substr(1);
 	std::string skipped_spaces = skip_spaces();
 	tok = next_token(false);
-	if (tok == "(" && body->has_args) {
+	if (body->has_args) {
+		if (tok != "(") {
+			if (tok.size() == 1 && iscntrl(tok[0])) {
+				char buf[5];
+				snprintf(buf, sizeof(buf), "\\%03o", tok[0]);
```
whitequark, line 484: Can we not do octal? No one remembers `\n` in octal.

whitequark (approved):

whitequark:
Thanks!

PR #2593, opened 2021-02-20
Adds is_wire to SigBit and SigChunk

RobertBaruch:
Useful for PYOSYS because Python can't easily check wire against NULL.

### DIFF
--- a/kernel/rtlil.h
+++ b/kernel/rtlil.h
@@ -735,7 +735,7 @@ struct RTLIL::SigChunk
 
 	RTLIL::SigChunk extract(int offset, int length) const;
 	inline int size() const { return width; }
-	inline int is_wire() const { return wire != NULL; }
+	inline bool is_wire() const { return wire != NULL; }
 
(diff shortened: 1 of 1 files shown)


RobertBaruch:
It kind of looks like the build failed temporarily due to unrelated issues, and then succeeded? I don't know when the check here is updated.

kernel/rtlil.h
```diff
@@ -760,6 +761,8 @@ struct RTLIL::SigBit
 	SigBit(const RTLIL::SigBit &sigbit) = default;
 	RTLIL::SigBit &operator =(const RTLIL::SigBit &other) = default;
 
+	inline int is_wire() const { return wire != NULL; }
```
zachjs, line 764: Arguably these should return `bool` rather than `int`. This may also give you an excuse to update the branch, which will hopefully not encounter the same failure when the tests are re-run.

RobertBaruch:
I don't know why I did that :)

mwkmwkmwk (approved):
looks good

PR #2599, opened 2021-02-23
Add $countbits, $countones, $isunknown, $onehot and $onehot0 functions

msinger:
I have implemented those functions so I can write
`assert ($onehot0(foo[2:0]));`
instead of
`assert (foo[0] + foo[1] + foo[2] <= 1);`

`$countbits(foo, 'z)` doesn't really work and `$countbits(foo, 'x)` counts both `'x` and `'z`. I guess this is because `'1&'z` results in `'x` and not `'z`. I hope this is okay.

I think that some test cases would be good, but before I take a look at that, I wanted to check if you are interested to merge this and if my code is acceptable or if I messed something up. I'm not sure about `children.size()`. It was written as `GetSize(children)` at other places I think. Is there a preference?

### DIFF
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -3148,6 +3148,34 @@ skip_dynamic_range_lvalue_expansion:;
 				goto apply_newNode;
 			}
 
+			if (str == "\\$countones" || str == "\\$isunknown" || str == "\\$onehot" || str == "\\$onehot0") {
+				if (children.size() != 1)
+					log_file_error(filename, location.first_line, "System function %s got %d arguments, expected 1.\n",
+							RTLIL::unescape_id(str).c_str(), int(children.size()));
+
+				AstNode *countbits = clone();
+				countbits->str = "\\$countbits";
+
+				if (str == "\\$countones") {
+					countbits->children.push_back(mkconst_bits({RTLIL::State::S1}, false));
+					newNode = countbits;
+				} else if (str == "\\$isunknown") {
+					countbits->children.push_back(mkconst_bits({RTLIL::Sx}, false));
+					countbits->children.push_back(mkconst_bits({RTLIL::Sz}, false));
+					newNode = new AstNode(AST_GT, countbits, mkconst_int(0, false));
+				} else if (str == "\\$onehot") {
+					countbits->children.push_back(mkconst_bits({RTLIL::State::S1}, false));
+					newNode = new AstNode(AST_EQ, countbits, mkconst_int(1, false));
+				} else if (str == "\\$onehot0") {
+					countbits->children.push_back(mkconst_bits({RTLIL::State::S1}, false));
+					newNode = new AstNode(AST_LE, countbits, mkconst_int(1, false));
+				} else {
+					log_abort();
+				}
+
+				goto apply_newNode;
+			}
+
 			if (current_scope.count(str) != 0 && current_scope[str]->type == AST_DPI_FUNCTION)
 			{
 				AstNode *dpi_decl = current_scope[str];
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -3089,6 +3089,65 @@ skip_dynamic_range_lvalue_expansion:;
 				goto apply_newNode;
 			}
 
+			if (str == "\\$countbits") {
+				if (children.size() < 2)
+					log_file_error(filename, location.first_line, "System function %s got %d arguments, expected at least 2.\n",
+							RTLIL::unescape_id(str).c_str(), int(children.size()));
+
+				std::vector<RTLIL::State> control_bits;
+
+				// Determine which bits to count
+				for (size_t i = 1; i < children.size(); i++) {
+					AstNode *node = children[i];
+					while (node->simplify(true, false, false, stage, -1, false, false)) { }
+					if (node->type != AST_CONSTANT)
+						log_file_error(filename, location.first_line, "Failed to evaluate system function `%s' with non-constant control bit argument.\n", str.c_str());
+					if (node->bits.size() != 1)
+						log_file_error(filename, location.first_line, "Failed to evaluate system function `%s' with control bit width != 1.\n", str.c_str());
+					control_bits.push_back(node->bits[0]);
+				}
+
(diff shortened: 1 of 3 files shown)


frontends/ast/simplify.cc
```diff
@@ -3089,6 +3089,138 @@ skip_dynamic_range_lvalue_expansion:;
...
+					AstNode *eqz_node;
+
+					if (count_0) {
+						/* Generate node for:  (exp & ('1 << i)) === ('0 << i) */
+						/*                                       ^^^           */
+						mask[i] = RTLIL::State::S0;
+						eq0_node = new AstNode(AST_EQX, and_node->clone(), mkconst_bits(mask, false));
+					} else {
+						eq0_node = mkconst_int(0, false);
+					}
```
zachjs, line 3145: I think these cases may benefit from abstraction into a lambda.

zachjs (changes requested):
I think it should be possible to correctly handle X and Z using a pair of shifts, rather than a mask. For example, I believe `3'b1zx << 1 >> 2 === 3'b00z`.

mwkmwkmwk:
I've submitted #2601, which will at least give a hint on how to solve the problem.

frontends/ast/simplify.cc
msinger: I collected the bit states in a vector so I can iterate over it instead of the four count_x vars.

The shifting works for 'z. Thanks.

frontends/ast/simplify.cc
```diff
@@ -3089,6 +3089,101 @@ skip_dynamic_range_lvalue_expansion:;
...
+					AstNode *lsh_node = new AstNode(AST_SHIFT_LEFT, exp->clone(), mkconst_int(i, false));
+					AstNode *rsh_node = new AstNode(AST_SHIFT_RIGHT, lsh_node, mkconst_int(exp_width - 1, false));
+
+					AstNode *or_node = nullptr;
+
+					for (auto it = do_count.begin(); it != do_count.end(); ++it) {
+						// Generate node for:  (exp << i >> ($size(exp) - 1)) === control_bit
+						//                                                    ^^^
+						cmp[0] = *it;
+						AstNode *eq_node = new AstNode(AST_EQX, rsh_node->clone(), mkconst_bits(cmp, false));
```
zachjs, line 3132: Given the the extension rules, I believe simply using `{state}` (a single element vector) would suffice here.

frontends/ast/simplify.cc
```diff
@@ -3089,6 +3089,101 @@ skip_dynamic_range_lvalue_expansion:;
...
+
+						if (it == do_count.begin()) {
+							or_node = eq_node;
+							continue;
+						}
+
+						// Or the result for each checked bit value
+						or_node = new AstNode(AST_LOGIC_OR, or_node, eq_node);
```
zachjs, line 3140:
```suggestion
						// Or the result for each checked bit value
						if (or_node)
							or_node = new AstNode(AST_LOGIC_OR, or_node, eq_node);
						else
							or_node = eq_node;
```
This also avoids duplicating the codegen for the first comparison of each bit.

frontends/ast/simplify.cc
```diff
@@ -3089,6 +3089,101 @@ skip_dynamic_range_lvalue_expansion:;
...
+					std::vector<RTLIL::State> cmp(exp_width, RTLIL::State::S0);
+
+					// Generate nodes for:  exp << i >> ($size(exp) - 1)
+					//                          ^^   ^^
+					AstNode *lsh_node = new AstNode(AST_SHIFT_LEFT, exp->clone(), mkconst_int(i, false));
+					AstNode *rsh_node = new AstNode(AST_SHIFT_RIGHT, lsh_node, mkconst_int(exp_width - 1, false));
+
+					AstNode *or_node = nullptr;
+
+					for (auto it = do_count.begin(); it != do_count.end(); ++it) {
```
zachjs, line 3128: Perhaps something like:
```C++
for (RTLIL::State control_bit : control_bits) {
```

frontends/ast/simplify.cc
```diff
@@ -3089,6 +3089,101 @@ skip_dynamic_range_lvalue_expansion:;
...
+							continue;
+						}
+
+						// Or the result for each checked bit value
+						or_node = new AstNode(AST_LOGIC_OR, or_node, eq_node);
+					}
+
+					// We should have at least one element in do_count, because we checked for the number of arguments above
+					if (!or_node)
+						log_abort();
```
zachjs, line 3145: Using `log_assert` can make for more clear error messages if this ever breaks.

frontends/ast/simplify.cc
```diff
@@ -3089,6 +3089,101 @@ skip_dynamic_range_lvalue_expansion:;
...
+
+				AstNode *countbits = clone();
+				countbits->str = "\\$countbits";
+				std::vector<RTLIL::State> v(1, RTLIL::State::S1);
+
+				if (str == "\\$countones") {
+					countbits->children.push_back(mkconst_bits(v, false));
+					newNode = countbits;
+				} else if (str == "\\$isunknown") {
+					v[0] = RTLIL::Sx;
```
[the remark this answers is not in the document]
msinger: yep

frontends/ast/simplify.cc
```diff
@@ -3089,6 +3089,101 @@ skip_dynamic_range_lvalue_expansion:;
...
+				// Detect width of exp (first argument of $countbits)
+				int  exp_width = -1;
+				bool exp_sign  = false;
+				AstNode *exp = children[0];
+				exp->detectSignWidth(exp_width, exp_sign, NULL);
+
+				newNode = mkconst_int(0, false);
+
+				for (int i = 0; i < exp_width; i++) {
+					std::vector<RTLIL::State> cmp(exp_width, RTLIL::State::S0);
```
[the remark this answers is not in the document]
msinger: I just removed this vector and used an initializer list here too. `===` still gives the same result with only one bit on that side.

frontends/ast/simplify.cc
msinger: Yes, you're right control_bits is a much better name

zachjs (approved):
Thank you for your contribution! This looks solid to me.

PR #2602, opened 2021-02-23
Fix syntax error in adff2dff.v

### LINKED ISSUE #2600: No longer able to use write_smt2 with circuits that have async-reset FFs (without clk2fflogic)
I have a project that calls the `write_smt2` backend on a circuit that has asynchronously reset flip-flops. I was originally using techmap to replace them with an equivalent circuit, but that no longer works after commit 09071afe. Here's an example input file:
```verilog
module test(
    input clk,
    input rst,
    input d,
    output reg q
);
always @(posedge clk or negedge rst) begin
    if (!rst) begin
        q <= 1'b0;
    end else begin
        q <= d;
    end
end
endmodule
```

Running the following:
```
read_verilog test.v
prep
techmap -map +/adff2dff.v
```
Gives:
```
3.1. Executing Verilog-2005 frontend: /usr/local/bin/../share/yosys/adff2dff.v
Parsing Verilog input from `/usr/local/bin/../share/yosys/adff2dff.v' to AST representation.
/usr/local/bin/../share/yosys/adff2dff.v:14: ERROR: syntax error, unexpected TOK_REG
```

I figured that this approach may now be deprecated in favor of `async2sync`, however this results in a different issue since as of 4a05cad7 async2sync replaces `adff`s with `sdff`s, which are also not supported by the write_smt2 backend. For example, running this on the same input file:
```
read_verilog test.v
prep
async2sync
write_smt2
```

Gives:
```
4. Executing SMT2 backend.
; SMT-LIBv2 description generated by Yosys 0.9+3911 (git sha1 cc7d18d2, clang 12.0.0 -fPIC -Os)
Creating SMT-LIBv2 representation of module test.
ERROR: Unsupported cell type $sdff for cell test.$procdff$3.
```

One way I can get this circuit into a state accepted by write_smt2 is by using clk2fflogic, but this feels a bit heavy-handed to me. 

Is this considered an issue? I'd be happy to help by contributing a PR, but I'm not sure what the best approach to resolving this would be. I could imagine either fixing write_smt2 to transform `sdff` cells, or modifying `async2sync` to no longer emit them (like it was prior to 4a05cad7).

mwkmwkmwk:
Fixes #2600.

### DIFF
--- a/techlibs/common/adff2dff.v
+++ b/techlibs/common/adff2dff.v
@@ -11,7 +11,7 @@ module adff2dff (CLK, ARST, D, Q);
 	(* force_downto *)
 	output reg [WIDTH-1:0] Q;
 	(* force_downto *)
-	wire reg [WIDTH-1:0] NEXT_Q;
+	reg [WIDTH-1:0] NEXT_Q;
 
 	wire [1023:0] _TECHMAP_DO_ = "proc;;";
 
--- a/tests/techmap/adff2dff.ys
+++ b/tests/techmap/adff2dff.ys
@@ -1,0 +1,19 @@
+read_verilog -icells << EOT
+module top(...);
+
+input [1:0] D;
+input C, R;
+output [1:0] Q;
+
+always @(posedge C, posedge R)
+  if (R)
+    Q <= 0;
+  else
+    Q <= D;
+
+endmodule
+EOT
+
+proc
+
+equiv_opt -async2sync techmap -map +/adff2dff.v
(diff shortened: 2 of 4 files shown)


zachjs (approved):
Are we sure we want merging this to close #2600, given that there are potentially other action items captured there?

mwkmwkmwk:
Fine, I added tests, though I'd argue these files would be better treated with a nuke.

Since all other files in `techlibs` are actually useful, they are covered by existing tests.

zachjs (approved):

PR #2604, opened 2021-02-23
frontend: Make helper functions for printing locations.

mwkmwkmwk:


frontends/ast/genrtlil.cc
```diff
@@ -45,10 +45,10 @@ static RTLIL::SigSpec uniop2rtlil(AstNode *that, IdString type, int result_width
 {
 	IdString name = stringf("%s$%s:%d$%d", type.c_str(), that->filename.c_str(), that->location.first_line, autoidx++);
 	RTLIL::Cell *cell = current_module->addCell(name, type);
-	cell->attributes[ID::src] = stringf("%s:%d.%d-%d.%d", that->filename.c_str(), that->location.first_line, that->location.first_column, that->location.last_line, that->location.last_column);
+	that->set_src_attr(cell);
```
zachjs, line 48: Overall this is a great change, but I think this warrants a quick discussion.

This (`that->set_src_attr(cell);`) is a bit counterintuitive for me, as it seems to invert the "direction" of this action. I'm guessing this was done to avoid having `rtlil.h` depend on `ast.h`. What if we made `set_src_attr` local to `genrtlil.cc`, and instead gave it a signature like:
```C++
void set_src_attr(RTLIL::AttrObject *obj, const AstNode *node);
```
so this would instead be
```C++
set_src_attr(cell, that);
```

mwkmwkmwk: It is also used in `ast.cc` in one place, but yes, making it a non-method is probably a good idea.

mwkmwkmwk:
Updated with a global function.

frontends/ast/ast.h
```diff
@@ -355,6 +358,9 @@ namespace AST
 	std::pair<std::string,std::string> split_modport_from_type(std::string name_type);
 	AstNode * find_modport(AstNode *intf, std::string name);
 	void explode_interface_port(AstNode *module_ast, RTLIL::Module * intfmodule, std::string intfname, AstNode *modport);
+
+	// Helper for setting the src attribute.
+	void set_src_attr(RTLIL::AttrObject *obj, AstNode *ast);
```
zachjs, line 363: Minor nit: Can we mark `ast` `const` to clear any potential ambiguity?

mwkmwkmwk: OK, done

zachjs (approved):

Issue #2473, opened 2020-12-04
Change in behavior results in failed BRAM inference


sylefeb:
## Summary

I observe a change in behavior in Yosys that results in failed BRAM inference, between an older and the most recent version (version details below). I include at the end a repro code, which I tried to make small.

The issue seems to come from how constant values applied to a port enable are treated. What this
example does is to either use a constant wire (assigned to 0) or a flip-flop that cannot be removed by synthesis (due to lack of initialization). 

In older versions using a constant wire has the desirable effect of simplifying the BRAM. In the latest version the constant wire leads to either a port in async or a non-recognized BRAM as below. 
Note that if the flip-flop is given an initial value of 0, yosys will remove it which also leads to a failed inference. 

## Steps to reproduce the issue

Using the attached example, commenting out USE_FLIPFLOP_ON_WENABLE at the top triggers the issue in the latest version (0.9+3743 git sha1 13a27055). 

## Expected behavior

Ideally as with version Yosys 0.9+3537 (git sha1 c75d8c74), where both cases would lead to an inferred BRAM, and using a constant wire on a port would lead to the BRAM being simplified (dual => simple dual => single => rom).

## Actual behavior

When assigning a constant wire to the wenable port, the BRAM inference fails.

## Details

* Command line `yosys -l yosys.log -p 'synth_ecp5 -abc9 -json build.json' 1_build.v`

* Using Yosys 0.9+3537 (git sha1 c75d8c74, gcc 10.1.0 -fPIC -Os)

Disabling `USE_FLIPFLOP_ON_WENABLE` the BRAM is accepted:
```
Processing top.__main.__mem__mem.buffer:
  Properties: ports=1 bits=2048 rports=1 wports=0 dbits=8 abits=8 words=256
...
 Checking rule #6 for bram type $__ECP5_DP16KD (variant 5):
    Bram geometry: abits=14 dbits=1 wports=0 rports=0
    Estimated number of duplicates for more read ports: dups=1
    Metrics for $__ECP5_DP16KD: awaste=16128 dwaste=0 bwaste=16128 waste=16128 efficiency=1
    Rule for bram type $__ECP5_DP16KD (variant 5) rejected: requirement 'attribute syn_romstyle="ebr" ...' not met.
  Selecting best of 4 rules:
    Efficiency for rule 4.3: efficiency=6, cells=2, acells=1
    Efficiency for rule 4.2: efficiency=11, cells=1, acells=1
[...]

PR #2605, opened 2021-02-23
Add memory writes to processes.

mwkmwkmwk:
As discussed on the 2021-02-15 meeting, this is the proper fix for #2473 and related issues — no more `memory_dff` converting async ports to sync ports, instead the memory writes are first emitted by the frontend as part of processes, and then converted to `$memwr` cells by `proc_memwr`, which has enough information to know the proper clock (or lack thereof).

This patch also adds more granular priority tracking in the frontend — the priority masks store information on which write ports need to have priority over which other ones, and which ones are unordered.  Currently this information is lost in `proc_memwr` (and converted to an arbitrary total order), but it will be properly hooked up and used in the upcoming memory inference changes.

TODO:

- the `tests/bram` test is now *broken* — it relies on memory write ports always having priority according to their source location, which is no longer true (writes from distinct processes can be executed in any order).
- `proc_arst` handling may not be the best (the enable signal should be ANDed with `!rst`, basically)

### DIFF
--- a/backends/rtlil/rtlil_backend.cc
+++ b/backends/rtlil/rtlil_backend.cc
@@ -242,11 +242,28 @@ void RTLIL_BACKEND::dump_proc_sync(std::ostream &f, std::string indent, const RT
 	case RTLIL::STi: f << stringf("init\n"); break;
 	}
 
-	for (auto it = sy->actions.begin(); it != sy->actions.end(); ++it) {
+	for (auto &it: sy->actions) {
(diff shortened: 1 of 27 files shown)


mwkmwkmwk:
Also: this PR removes the old transform from `memory_dff`, thus breaking compatibility with old `.il` files that contain not-quite-detected sync write ports.  If this is a problem, we can just remove the last commit from this PR.

PR #2610, opened 2021-02-25
verilog: fix handling of nested ifdef directives

zachjs:
- track depth so we know whether to consider higher-level elsifs
- error on unmatched endif/elsif/else

@rswarbrick

### DIFF
--- a/frontends/verilog/preproc.cc
+++ b/frontends/verilog/preproc.cc
@@ -727,7 +727,8 @@ frontend_verilog_preproc(std::istream                 &f,
 
 	std::vector<std::string> filename_stack;
 	int ifdef_fail_level = 0;
-	bool in_elseif = false;
+	int ifdef_pass_level = 0;
+	bool ifdef_already_satisfied = false;
 
 	output_code.clear();
 	input_buffer.clear();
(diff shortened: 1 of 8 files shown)


rswarbrick (approved):
Looks good!

I think that (before and after this change), we probably do something a little odd if someone combines ifdef and include strangely. Something like this:
```
// a.v
`ifdef FOO
AAA
`include b.v
```
and
```
// b.v
`endif
````
looks like it will expand to `AAA` if `FOO` is defined (with the endif coming from the "wrong" file). This sort of silliness is explicitly prohibited in the C spec, so I was going to suggest that we need a stack for `ifdef_pass_level` and `ifdef_fail_level`. But I went and checked IEEE 1800-2017, and can't find the equivalent wording. Maybe we don't really care (with Yosys's usual approach of "if your code is nonsense, we make no guarantees").

**Comment by zachjs** [comment]
From the beginning of Section 22.4:

> The file inclusion (`` `include ``) compiler directive is used to insert the entire contents of a source file in another file during compilation. The result is as though the contents of the included source file appear in place of the `` `include `` compiler directive.

This is one of those definitions that appears reasonable at first, but reveals problems upon further inspection. I would definitely prefer C's restrictions, but our current simplistic behavior seems in line with this (unfortunately simplistic) specification.

**Review by clairexen — APPROVED** [review]
LGTM

File: frontends/verilog/preproc.cc [hunk 1]
```diff
@@ -743,42 +744,68 @@ frontend_verilog_preproc(std::istream                 &f,
...
 			std::string name = next_token(true);
-			if (ifdef_fail_level > 0 || !defines.find(name))
+			if (ifdef_fail_level > 0 || !defines.find(name)) {
 				ifdef_fail_level++;
+			} else {
+				ifdef_pass_level++;
+				ifdef_already_satisfied = true;
+			}
+				if (ifdef_fail_level == 1)
+					ifdef_already_satisfied = false;
```
clairexen, line 794: indent messed up?

frontends/verilog/preproc.cc
```diff
@@ -743,42 +744,68 @@ frontend_verilog_preproc(std::istream                 &f,
...
 			std::string name = next_token(true);
-			if (ifdef_fail_level > 0 || defines.find(name))
+			if (ifdef_fail_level > 0 || defines.find(name)) {
 				ifdef_fail_level++;
+			} else {
+				ifdef_pass_level++;
+				ifdef_already_satisfied = true;
+			}
+				if (ifdef_fail_level == 1)
+					ifdef_already_satisfied = false;
```
clairexen, line 808: indenting?

zachjs:
With the indentation issues fixed, I'm going to merge this given the prior approvals.

PR #2618, opened 2021-02-28
sv: extended support for integer types

zachjs:
- Standard data declarations can now use any integer type
- Parameters and localparams can now use any integer type
- Function returns types can now use any integer type
- Fix `parameter logic`, `localparam reg`, etc. to be 1 bit (previously 32 bits)
- Added longint type (64 bits)
- Unified parser source for integer type widths


whitequark (approved):
LGTM, very nice!

PR #2620, opened 2021-03-01
verilog: fix sizing of ports with int types in module headers

zachjs:
Declaring the ports as standard module items already worked as expected. This adds a missing usage of `checkRange()` so that headers such as `module m(output integer x);` now work correctly.


whitequark (approved):
LGTM

PR #2625, opened 2021-03-02
fixes issue where unextend incorrectly erases 0-bit of signed signal

sylefeb:
This PR addresses an issue for ice40_dsp where `unextend` incorrectly erases the 0-sign bit of a const signed signal. This only addresses the ice40 case but the issue may be affecting `xiling_dsp*` which feature a similar `unextend` lambda.

### DIFF
--- a/passes/pmgen/ice40_dsp.pmg
+++ b/passes/pmgen/ice40_dsp.pmg
@@ -23,7 +23,7 @@ match mul
 endmatch
 
 code sigA sigB sigH
-	auto unextend = [](const SigSpec &sig) {
+	auto unextend = [](const SigSpec &sig,bool is_signed) {
 		int i;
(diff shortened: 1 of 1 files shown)


Ravenslofty:
Signedness bugs like this are common enough that I'm wondering if the generic Yosys multiplication test file should include some checks for it.

sylefeb:
I might have identified another one. I will have a closer look soon. If there is an existing framework for testing multipliers I'd be happy to help expand them? (I could not easily locate one in tests/?) Running the simulation with `ice40/cells_sim.v` worked great to reveal the issue.

sylefeb:
I have found a different but related issue that likely makes this PR obsolete, so I will close it. I'll work on a new PR, but I'd welcome guidance as I am quite new to this. More info: https://twitter.com/sylefeb/status/1368274609607413760

PR #2626, opened 2021-03-02
sv: support for parameters without default values

zachjs:
- Modules with a parameter without a default value will be automatically deferred until the hierarchy pass
- Allows for parameters without defaults as module items, rather than just int the `parameter_port_list`, despite being forbidden in the LRM
- Check for parameters without defaults that haven't been overriden
- Add location info to parameter/localparam declarations


whitequark (approved):
LGTM, very nice addition.

Issue #2622, opened 2021-03-01
Expansion of unsized single bit value not working in port map


joshtyler:
Unsized single bit values preceded by an apostrophe expand to fill the entire width of the target vector in system verilog

e.g. `logic [1:0] foo = '1; // Equivalent to = 2'b11`

Yosys appears to incorrectly only assign the lowest bit in the vector, when this style of assignment is used in a port map.

N.B. I assume this is related to #2570 , but I don't really know anything about how the yosys verilog frontend works, so can't provide an opinion!

## Steps to reproduce the issue

Save the following source as broken.sv:
```
module hier
(
    input [1:0] in,
    output [1:0] out,
);
    assign out = in;
endmodule

module broken
(
    output [1:0] out
);

    hier inst
    (
        .in('1),
        .out(out)
    );
endmodule
```
Run `yosys -p "read_verilog -sv broken.sv; synth_ice40; write_verilog"`

Observe the value assigned to `out`

I am running the following build of yosys:
`Yosys 0.9+3981 (git sha1 0e0f8429, clang 11.0.0 -fPIC -Os)`

## Expected behavior

`out` is assigned 2'h3

## Actual behavior

`out` is assigned 2'h1

zachjs:
This particular example actually does run into some of classic issues surrounding module elaboration ordering, as we'd essentially need to know the width of the port while elaborating the instantiation.

joshtyler:
Thanks very much for the explanation (and potential workaround - sv2v looks interesting!)

From my point of view as a user, it wouldn't really be a problem if yosys didn't support this (working around it was a two line change in my project). As you identify, the real issue for me was that I silently got the wrong result - effectively giving me simulation/synthesis mismatch that was a little tricky to debug.

zachjs:
The flag I described already exists, but wasn't shown in the AST dump. Something like `logic [1:0] foo = '1;` already works as expected. I have some changes locally which fix the size cast edge case (i.e., `3'('1)`) and fix the AST dump. I also have a hack to address the classic port binding usage, but there are shortcomings.

1. With my changes, your example correctly sign extends, but still yields a warning `Resizing cell port broken.inst.in from 1 bits to 2 bits.`.
2. A binding like `.inp(r ? '1 : x)` still behaves incorrectly if `inp` is wider than `x`, as the `'1` will only extend to the width of `x`. A usage like this is probably less common than something like `.inp(r ? '1 : '0)`, which also still doesn't extend correctly. 

[...]

PR #2627, opened 2021-03-03
sv: fix some edge cases for unbased unsized literals

zachjs:
- Fix explicit size cast of unbased unsized literals
- Fix unbased unsized literal bound directly to port
- Output `is_unsized` flag in `dumpAst`

Related to #2622


mmicko (approved):
LGTM. To be honest not sure if this could be done cleaner without huge code changes

PR #2630, opened 2021-03-04
kernel: When a wire is removed, change its 0-length SigChunks to const.

### LINKED ISSUE #2623: Yosys hangs when building microwatt

```
% git clone git@github.com:paulusmack/microwatt.git -b gpio
% cd microwatt/
% FPGA_TARGET="ECP5-EVN" SYNTH_ECP5_FLAGS="-noflatten" PODMAN=1 make microwatt.json
```

This runs for about 10 seconds but then produces no more output. Yosys is using 100% CPU.

Issue was first discovered by the microwatt CI in this microwatt PR:
https://github.com/antonblanchard/microwatt/pull/277
CI that shows the yosys hang is here (eventually killed by the CI)
https://github.com/antonblanchard/microwatt/pull/277/checks?check_run_id=1966479444#step:3:1499

## Expected behaviour

Yosys should run to completion.

## Actual behaviour

Yosys uses 100% of the CPU while making no forward progress.

mwkmwkmwk:
This prevents re-creating 0-length wires in a loop.

Fixes #2623.

### DIFF
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -1826,8 +1826,13 @@ void RTLIL::Module::remove(const pool<RTLIL::Wire*> &wires)
 			sig.pack();
 			for (auto &c : sig.chunks_)
 				if (c.wire != NULL && wires_p->count(c.wire)) {
-					c.wire = module->addWire(stringf("$delete_wire$%d", autoidx++), c.width);
-					c.offset = 0;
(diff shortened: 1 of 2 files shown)


mwkmwkmwk:
PR superseded by #2646 , which is considered a better solution to the underlying problem, as discussed on today's meeting.

PR #2631, opened 2021-03-04
verilog: support module scope identifiers in parametric modules

zachjs:
Extension of #2353

### DIFF
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -1691,11 +1691,15 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
 	if (type == AST_IDENTIFIER) {
 		if (current_scope.count(str) == 0) {
 			AstNode *current_scope_ast = (current_ast_mod == nullptr) ? current_ast : current_ast_mod;
-			const std::string& mod_scope = current_scope_ast->str;
-			if (str[0] == '\\' && str.substr(0, mod_scope.size()) == mod_scope) {
-				std::string new_str = "\\" + str.substr(mod_scope.size() + 1);
+			size_t pos = str.find('.', 1);
+			if (str[0] == '\\' && pos != std::string::npos) {
+				std::string new_str = "\\" + str.substr(pos + 1);
 				if (current_scope.count(new_str)) {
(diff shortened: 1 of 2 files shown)


frontends/ast/simplify.cc
```diff
@@ -1652,11 +1652,15 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
...
-				std::string new_str = "\\" + str.substr(mod_scope.size() + 1);
+			size_t pos = str.find('.', 1);
+			if (str[0] == '\\' && pos != std::string::npos) {
+				std::string new_str = "\\" + str.substr(pos + 1);
 				if (current_scope.count(new_str)) {
-					str = new_str;
+					std::string prefix = str.substr(0, pos);
+					const std::string &mod_scope = current_scope_ast->str;
+					if (mod_scope == prefix || mod_scope.find("$paramod" + prefix + '\\') == 0
+						|| (mod_scope.find("$paramod$") == 0 && mod_scope.substr(49) == prefix))
```
whitequark, line 1662: Please no. It is bad enough that we have ad-hoc manipulation of `$paramod` strings everywhere, but hardcoding the length of the hash here is too much.

whitequark (changes requested):

frontends/ast/simplify.cc
zachjs: I agree. Both this extension and my original implementation feel pretty hacky. Would changing `mod_scope.substr(49) == prefix` to something akin to `mod_scope.ends_with(prefix)` suffice? I'm very open to scrapping this whole implementation for something different if we can come up with a more principled alternative.

zachjs: @whitequark _bump_

whitequark: I think there should be a pair of functions somewhere that construct/deconstruct the `$paramod` strings and everything else should use those.

zachjs: I just discovered the `hdlname` attribute which already serves the purpose I was looking for. I've dropped the dependence on the structure of derived module names.

whitequark (approved):

PR #2632, opened 2021-03-04
verilog: impose limit on maximum expression width

### LINKED ISSUE #2290: Crash when attempting to read verilog
The attached file has some procedurally generated Verilog (from SystemVerilog using SV2V). When attempting to read the verilog into Yosys, the tool appears to crash without an error message. I gave the RTL a quick lookover and it appears to be wellformed so therefore unsure about how to debug.

[crash.tar.gz](https://github.com/YosysHQ/yosys/files/4956777/crash.tar.gz)

zachjs:
Designs with unreasonably wide expressions would previously get stuck allocating memory forever.

Fixes #2290


whitequark (approved):
LGTM, thanks! This is a common source of issues.

PR #2641, opened 2021-03-06
proc_clean: Fix empty case removal conditions.

### LINKED ISSUE #2639: `proc_clean` overzealously deletes empty cases
```
module top(...);

input a;
input clk;
input d;
output reg q;

always @(posedge clk) begin
        case (a)
                1'b0: begin
                end
                1'b0: begin
                        q <= d;
                end
        endcase
end

endmodule
```

Running this through `proc_clean` causes removal of the first case and thus changes semantics.

There is also another codepath in `proc_clean`, which removes empty last case even if the case is not considered "fully defined"; this is likewise incorrect for `(* full_case *)` cases

mwkmwkmwk:
Fixes #2639.

### DIFF
--- a/passes/proc/proc_clean.cc
+++ b/passes/proc/proc_clean.cc
@@ -76,22 +76,33 @@ void proc_clean_switch(RTLIL::SwitchRule *sw, RTLIL::CaseRule *parent, bool &did
 	}
 	else
 	{
-		bool all_fully_def = true;
 		for (auto cs : sw->cases)
-		{
 			if (max_depth != 0)
 				proc_clean_case(cs, did_something, count, max_depth-1);
-			int size = 0;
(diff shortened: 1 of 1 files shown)


whitequark:
Oops, this is probably my fault (introduced in 18291c20d2?).

mwkmwkmwk:
> Oops, this is probably my fault (introduced in [18291c2](https://github.com/YosysHQ/yosys/commit/18291c20d2b17689729d9c36b08967e928e4e32a)?).

Seems so; Anyway, it seems I have broken the intended use of it, but also it could be reintroduced by having a pass that detects `parallel_case` that is separate from `proc_mux`? I'd rather avoid duplicating that logic.

whitequark (approved):
I have no objections to merging this.

PR #2647, opened 2021-03-09
json: Add support for memories.

### LINKED ISSUE #1908: Yosys aborts on $memrd/wr cells when saving json, reloading json, and continuing the same process

Start Yosys and do this:
```
read_verilog CpuComplex.v
read_verilog -lib -D_ABC +/ice40/cells_sim.v
hierarchy -check -top CpuComplex   
proc
write_json CpuComplex.proc.json
exit
```

Restart Yosys and do this:
```
read_json CpuComplex.proc.json
flatten
```

Full repro environment can be found here: https://github.com/tomverbeure/yosys_error/tree/save_reload_crash.

Just clone and execute  `./run.sh`.

## Expected behavior

I expected to be able to use JSON to save and restore state and pick up where I left off.

[According](https://old.reddit.com/r/yosys/comments/fzrdje/yosys_aborts_when_saving_during_an_intermediate/fn5umzf/) to @daveshah1, the JSON netlist format shouldn't be used for that.

But at least, Yosys should bail with some error instead of just aborting.

## Actual behavior

Yosys aborts as follows:

```
...
Importing module CCPipelinedMemoryBusRam from JSON tree.
Importing module CCMasterArbiter from JSON tree.

2. Executing FLATTEN pass (flatten design).
Using template VexRiscv for cells of type VexRiscv.
<suppressed ~1 debug messages>
ERROR: Assert `memory_renames.count(memid) != 0' failed in passes/techmap/techmap.cc:425.
```

The abort is caused by the presence of `$memrd` and `$memwr` cells in the netlist. When I remove those, Yosys proceeds as expected.

## Work around

Use `write_ilang`/`read_ilang` instead of  `write_json`/`read_json`. That works fine.

The reason I wanted to use json is because it doesn't require a custom parser for most languages.

mwkmwkmwk:
Previously, memories were silently discarded by the JSON backend, making
round-tripping modules with them crash.

Since there are already some users using JSON to implement custom
external passes that use memories (and infer width/size from memory
ports), let's fix this by just making JSON backend and frontend support
memories as first-class objects.

Processes are still not supported, and will now cause a hard error.

Fixes #1908.

### DIFF
--- a/backends/json/json.cc
+++ b/backends/json/json.cc
@@ -135,6 +135,10 @@ struct JsonWriter
 		// reserve 0 and 1 to avoid confusion with "0" and "1"
 		sigidcounter = 2;
 
+		if (module->has_processes()) {
+			log_error("Module %s contains processes, which are not supported by JSON backend.\n", log_id(module));
+		}
+
 		f << stringf("    %s: {\n", get_name(module->name).c_str());
 
 		f << stringf("      \"attributes\": {");
@@ -215,6 +219,27 @@ struct JsonWriter
 			first = false;
 		}
 		f << stringf("\n      },\n");
+
+		if (!module->memories.empty()) {
+			f << stringf("      \"memories\": {");
+			first = true;
+			for (auto &it : module->memories) {
+				if (use_selection && !module->selected(it.second))
+					continue;
+				f << stringf("%s\n", first ? "" : ",");
+				f << stringf("        %s: {\n", get_name(it.second->name).c_str());
+				f << stringf("          \"hide_name\": %s,\n", it.second->name[0] == '$' ? "1" : "0");
+				f << stringf("          \"attributes\": {");
+				write_parameters(it.second->attributes);
+				f << stringf("\n          },\n");
+				f << stringf("          \"width\": %d,\n", it.second->width);
+				f << stringf("          \"start_offset\": %d,\n", it.second->start_offset);
+				f << stringf("          \"size\": %d\n", it.second->size);
+				f << stringf("        }");
+				first = false;
+			}
+			f << stringf("\n      },\n");
+		}
 
 		f << stringf("      \"netnames\": {");
 		first = true;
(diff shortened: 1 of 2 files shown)


clairexen (approved):

hzeller:
Hard errors for the processes are somewhat disruptive in cases where the processes are not needed.
It now breaks workflows that exported JSON but didn't necessarily worry about processes being exported.

I suggest making the [hard error](https://github.com/YosysHQ/yosys/blob/c8b45a4a826efe88e17928b3d6f526dc2866312a/backends/json/json.cc#L138-L140) a warning instead, so that existing functionality still works, but the user is made aware of the detected limitations.

mwkmwkmwk:
Which cases do you have in mind?

If it's about synthesis flow output, see #2679 for an upcoming solution; if you have any other use cases, I'd like to hear about them.  In general, we'd rather have yosys fail hard than emit broken output (and emitting a module with processes by just skipping the processes very much counts as this).

hzeller:
Just any simple module that makes use of an initial block will now fail to export as JSON.

Here is a simplified example: Let's say this simple `foo.v` in which we initialize a value:
```verilog
module foo();
  reg x = 0;
endmodule
```

So processing it with yosys will fail:

```bash
$ yosys -q -p "read_verilog foo.v; write_json foo.json"
ERROR: Module foo contains processes, which are not supported by JSON backend.

$ yosys -V
Yosys 0.9+4008 (git sha1 4f4e70876, clang 11.0.1-2 -fPIC -Os)
```

I often use JSON to structurally look at the internal representations or feed into further tooling, but I might not particularly be interested in the initializers.

But with this change, almost no file can even be processed (there is always an initializer somewhere...). I would be totally happy if I get the warning that a process could not be exported (ideally with a file/line number), but that otherwise the rest would be processed.

whitequark:
Why not run `proc` first?

hzeller:
Yes this is of course a simplified example that would work with `proc`. Just tried to find some simple example how we can show the problem. Inspecting intermediate stages of the unprocessed internal state can be useful.

My point here is, that having this as warning rather than an error would be useful.

whitequark:
If you specifically want to export everything but processes, why not run `delete %p:*` before `write_json`?

PR #2653, opened 2021-03-11
verilog: disallow overriding global parameters

zachjs:
It was previously possible to override global parameters on a per-instance basis. This could be dangerous when using positional parameter bindings, hiding oversupplied parameters.

I'm not certain that this is forbidden by the LRM. Some tools allow it, others don't. I believe the current behavior is accidental. Let's use this PR to discuss which behavior we prefer.

### DIFF
--- a/frontends/ast/ast.cc
+++ b/frontends/ast/ast.cc
@@ -1286,6 +1286,8 @@ void AST::process(RTLIL::Design *design, AstNode *ast, bool dump_ast1, bool dump
 		}
 		else {
 			// must be global definition
+			if ((*it)->type == AST_PARAMETER)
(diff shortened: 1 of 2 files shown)


whitequark:
I uh... not only I never realized this is something any Verilog tooling accepts, but I also struggle to imagine how would anyone consider this a reasonable feature when it breaks lexical scoping in such a non-discoverable way? The only other environment I know with a similar feature is Coq, but it has very different constraint.

whitequark (approved):

whitequark:
I'm going to merge this with the understanding that, should anyone complain, we can back off this change in a way that prevents unintended use of the feature, with a warning or perhaps even a frontend option.

PR #2655, opened 2021-03-12
genrtlil: fix signedness of fully sliced wire

### LINKED ISSUE #2654: Yosys asserts on signed signal assigned to unsigned instance port

Reduced example:

`test.v`:
```verilog
module test;
	wire signed [31:0] a_ff;

	my_dff #(32) u_my_dff(
		.dout(a_ff[31:0])
	);
endmodule
```

Yosys:
```
read_verilog -sv test.v
```

## Expected behavior

File gets read correctly.

## Actual behavior

```
yosys> read_verilog -sv test.v
Parsing SystemVerilog input from `test.v' to AST representation.
ERROR: Assert `arg->is_signed == sig.as_wire()->is_signed' failed in frontends/ast/genrtlil.cc:1828.
```

I don't think SystemVerilog requires equal signedness between argument and instance port, but even if it does, it should only check that during elaboration.

### Adding submodule definition -> also fail

The following case fails too, with the same error:

```verilog
module my_dff #( parameter WIDTH=1) (output logic signed [WIDTH-1:0] dout);

    assign dout = 0;

endmodule

module test;
        wire signed [31:0] a_ff;

        my_dff #(32) u_my_dff(
                .dout(a_ff[31:0])
        );
endmodule
```

```
yosys> read_verilog -sv test.v
Parsing SystemVerilog input from `test.v' to AST representation.
Generating RTLIL representation for module `\my_dff'.
ERROR: Assert `arg->is_signed == sig.as_wire()->is_signed' failed in frontends/ast/genrtlil.cc:1828.
```

### Remove `signed` qualifier -> pass

When I remove the `signed` qualifiers, Yosys processes both cases fine.

```verilog
module test;
        wire [31:0]  a_ff;

        my_dff #(32) u_my_dff(
                .dout(a_ff[31:0])
        );
endmodule
```

### Change argument from vector to 1-bit wire -> pass

Also, when I keep the argument signed, but change argument from vector to a one-bit wire, it passes too:

```verilog
module test;
        wire signed  a_ff;

        my_dff #(1) u_my_dff(
                .dout(a_ff)
        );
endmodule
```

### Change argument from multi-bit vector to 1-bit vector -> fail

```verilog
module test;
        wire signed [0:0]  a_ff;

        my_dff #(1) u_my_dff(
                .dout(a_ff[0:0])
        );
endmodule
```

zachjs:
Fixes #2654

### DIFF
--- a/frontends/ast/genrtlil.cc
+++ b/frontends/ast/genrtlil.cc
@@ -1383,6 +1383,15 @@ RTLIL::SigSpec AstNode::genRTLIL(int width_hint, bool sign_hint)
 
 			RTLIL::SigSpec sig = { RTLIL::SigSpec(RTLIL::State::Sx, add_undef_bits_msb), chunk, RTLIL::SigSpec(RTLIL::State::Sx, add_undef_bits_lsb) };
 
+			// if wire is signed, and is sliced with a range that exactly covers
+			// the size of the wire, the resulting expression should be unsigned
+			if (wire->is_signed && children.size() && !add_undef_bits_msb && !add_undef_bits_lsb){
+				RTLIL::IdString indirect_name = NEW_ID;
+				RTLIL::Wire *indirect = current_module->addWire(indirect_name, GetSize(chunk));
+				current_module->connect(indirect, chunk);
+				sig = indirect;
+			}
+
 			if (genRTLIL_subst_ptr)
 				sig.replace(*genRTLIL_subst_ptr);
 
--- a/tests/simple/signed_full_slice.v
+++ b/tests/simple/signed_full_slice.v
@@ -1,0 +1,13 @@
+module pass_through(
+    input wire [31:0] inp,
+    output wire [31:0] out
+);
+    assign out = inp;
+endmodule
+
+module top(
+    input wire signed [31:0] inp,
+    output wire [31:0] out
+);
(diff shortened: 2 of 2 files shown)


whitequark (approved):

zachjs:
The obvious issue here is that slices/bits on the left-hand side of an assignment are broken, as the inserted connection goes the wrong way. There is also an incidental issue where this is creating the indirections in some places where it is okay, but not necessary because I am not checking the width of the final chunk.

I believe the existing indirection for port connections below works because _signed_ non-wire nodes cannot be left-hand sides.
https://github.com/YosysHQ/yosys/blob/396ad17e06a7b5fa912180540206a6560b54101f/frontends/ast/genrtlil.cc#L1825-L1837

I don't think the failing assertion should be removed, as it is catching places where a slice would sign extend in a port connection when it should not. I don't think we can prevent the SigSpec from sign-extending without some form of indirection. I'm wondering if we could make LHS vs. RHS available my marking nodes, etc. Thoughts @whitequark?

zachjs:
While determining LHS vs. RHS in general is straightforward, it does run into the classic elaboration ordering issue for port bindings.

whitequark:
> Thoughts @whitequark?

Let's discuss this on the next meeting.

PR #2666, opened 2021-03-15
bugpoint: add runner option

### LINKED ISSUE #2629: bugpoint: Add a timeout option
When debugging an issue that involves an infinite loop, it would be helpful to have a timeout feature in `bugpoint`, so that testcase reduction can be done automatically.

zachjs:
This has been updated to the more general `-runner` option proposed below. *nix systems appear to have the `timeout` command available in general. I confirmed that with the recent zero-width chunk fix reverted, the following script
```
read_rtlil <<EOF
module \top
  wire output 1 \a
  wire width 0 $dummy
  cell \abc \abc
    connect \a \a
    connect \b $dummy
  end
  wire width 1 \foo
  wire width 2 \bar
end
EOF

bugpoint -yosys "./yosys" -command opt_clean -runner "timeout 5"
```
correctly runs and produces
```
autoidx 5
module \top
  wire width 0 $delete_wire$2
  wire $delete_wire$4
  cell \abc \abc
    connect \a $delete_wire$4
    connect \b $delete_wire$2
  end
end
```

Fixes #2629

### DIFF
--- a/passes/cmds/bugpoint.cc
+++ b/passes/cmds/bugpoint.cc
@@ -87,9 +87,12 @@ struct BugpointPass : public Pass {
 		log("    -updates\n");
 		log("        try to remove process updates from syncs.\n");
 		log("\n");
+		log("    -runner \"<prefix>\"\n");
+		log("        child process wrapping command, e.g., \"timeout 30\", or valgrind.\n");
+		log("\n");
 	}
 
-	bool run_yosys(RTLIL::Design *design, string yosys_cmd, string yosys_arg)
+	bool run_yosys(RTLIL::Design *design, string runner, string yosys_cmd, string yosys_arg)
 	{
 		design->sort();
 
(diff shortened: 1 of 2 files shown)


mwkmwkmwk:
Looks good, but something should be done about non-UNIX platforms (in particular, windows) — possibly just a message of "timeout is currently not supported on this platform" if no easy solution is found

passes/cmds/bugpoint.cc
```diff
@@ -87,9 +87,12 @@ struct BugpointPass : public Pass {
 		log("    -updates\n");
 		log("        try to remove process updates from syncs.\n");
 		log("\n");
+		log("    -timeout\n");
```
mwkmwkmwk, line 90: write `-timeout <num>` here, so it's clear that the option takes an argument

whitequark:
If prefixing the command with something is the only option on the table, does `-yosys "timeout n yosys"` work?

zachjs:
@whitequark The quotes around the command aren't stripped, so that will yield something like `"timeout 5 ./yosys" is not a command`.

whitequark:
Should we strip those, then?

whitequark:
Although that will probably break under `Program Files (x86)` or something. `-runner`?

zachjs:
I can get behind the more general variant. I suppose we'd just drop the quotes if present for that arg. What other runners did you imagine?

whitequark:
> What other runners did you imagine?

`valgrind` for example

whitequark (approved):

zachjs:
I rebased and force-pushed to re-run CI, which appears to have experienced a spurious failure installing some dependencies.

Issue #2667, opened 2021-03-16
Processes in whitebox causes synth_ecp5 JSON backend to fail


gatecat:
Run:

```
yosys -p "add -mod top; synth_ecp5 -json /dev/null"
```

as the ECP5 distributed RAM whiteboxes contain processes but nothing elaborates those processes; the JSON backend is now failing:

```
ERROR: Module DPR16X4C contains processes, which are not supported by JSON backend.
```

A workaround would be to unwhitebox those cells as I don't think there is much useful optimisation that can be done anyway, but I suspect this issue may well affect other situations involving boxes in processes as well.

cc @edbordin

mwkmwkmwk:
Would running `blackbox` command on all whitebox modules at the end of the synthesis flow solve this problem?

gatecat:
Yes, that seems like a better way to deal with this

PR #2668, opened 2021-03-16
ecp5: Remove lib_whitebox from distributed RAM

gatecat:
This is a workaround for #2667 to at least unbreak ECP5 builds while we think about possible better ways of dealing with this case.


eddiehung:
Just to be clear, removing these attributes will mean that ABC9 will not look inside this module to extract a bypass box that captures its async read behaviour.

gatecat:
Closing in favour of the solution of blackboxing whiteboxes before hitting the JSON backend

PR #2671, opened 2021-03-16
sv: carry over global typedefs from previous files

### LINKED ISSUE #2440: enums recognized in another file, but not typedef
I have two files: defs.vh and top.sv.

In defs.vh: 
`typedef enum logic [7:0] {TERMA=1, TERMB=2, TERMC=4, TERMD=8} term_t;`

In top.sv:
```
`default_nettype none
`include "defs.vh"
module fsm (input clk, reset, output logic [7:0] out);
  term_t foo;
  assign out = foo;
  always_ff @(posedge clk, posedge reset)
    if (reset)
      foo <= TERMA;
    else
      case(foo)
        TERMA: foo <= TERMB;
        TERMB: foo <= TERMC;
        TERMC: foo <= TERMD;
        TERMD: foo <= TERMA;
      endcase
endmodule
```

Using the current Yosys git master, I ran the following command:
`yosys -p "read_verilog -sv defs.vh top.sv; synth_ice40 -top fsm; write_verilog struct_top.sv"`

## Expected behavior

The resulting structural Verilog should have implemented a circular FSM with the help of the term_t type.

## Actual behavior

A syntax error is thrown at the line "term_t foo" in top.sv: "top.sv:15: ERROR: syntax error, unexpected ';', expecting '(' or '['".  The error doesn't make much sense, but I assume it's because it doesn't recognize term_t, despite Yosys having parsed the defs header file first.

It works if I copy the typedef from defs.vh back into top.sv, but if we're trying to keep our definitions separate, then it doesn't work.  Is this a bug or am I doing something wrong here?

zachjs:
This breaks the ability to use a global typename as a standard identifier in a subsequent input file. This is otherwise backwards compatible, including for sources which previously included conflicting typedefs in each input file.

I'm not sure why this logic is in `add_package_types`, but I figure I'd just fix it in place rather than making a larger change.

Fixes #2440

### DIFF
--- a/frontends/verilog/verilog_frontend.cc
+++ b/frontends/verilog/verilog_frontend.cc
@@ -61,8 +61,11 @@ static void add_package_types(dict<std::string, AST::AstNode *> &user_types, std
 			}
 		}
 	}
-	user_type_stack.clear();
-	user_type_stack.push_back(new UserTypeMap());
+
+	// carry over typedefs from previous files, but allow them to be overridden
+	// note that these type maps are currently never reclaimed
+	if (user_type_stack.empty() || !user_type_stack.back()->empty())
+		user_type_stack.push_back(new UserTypeMap());
 }
 
 struct VerilogFrontend : public Frontend {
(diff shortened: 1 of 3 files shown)


whitequark (approved):

zachjs:
I rebased and force-pushed to re-run CI, which appears to have experienced a spurious failure installing some dependencies.

PR #2672, opened 2021-03-16
verilog: fix buf/not primitives with multiple outputs

Xiretza:
From IEEE1364-2005 section 7.3, buf and not gates:

> These two logic gates shall have one input and one or more outputs. The last terminal in the terminal list shall connect to the input of the logic gate, and the other terminals shall connect to the outputs of the logic gate.

yosys does not follow this and instead interprets the first argument as the output, the second as the input and ignores the rest.

There exists a test in `tests/asicworld/code_verilog_tutorial_n_out_primitive.v` that should probably catch this; not sure what's up with that.

### DIFF
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -2223,6 +2223,21 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
 			children.push_back(node);
 			did_something = true;
 		}
+		else if (str == "buf" || str == "not")
+		{
+			AstNode *input = children_list.back();
+			if (str == "not")
+				input = new AstNode(AST_BIT_NOT, input);
+
+			newNode = new AstNode(AST_GENBLOCK);
+			for (auto it = children_list.begin(); it != std::prev(children_list.end()); it++) {
+				newNode->children.push_back(new AstNode(AST_ASSIGN, *it, input->clone()));
+				newNode->children.back()->was_checked = true;
+			}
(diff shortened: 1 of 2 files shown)


frontends/ast/simplify.cc
```diff
@@ -2218,6 +2218,27 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
 			children.push_back(node);
 			did_something = true;
 		}
+		else if (str == "buf" || str == "not")
+		{
+			AstNode *input = children_list.back();
+			if (str == "not")
+				input = new AstNode(AST_BIT_NOT, input);
+
+			std::vector<AstNode *> assigns;
```
zachjs, line 2227: Arguably the `newNode` should be an `AST_GENBLOCK` containing all of the new assignments, rather than adding one assignment in place and the rest at the end of the module.

zachjs (changes requested):
Can we add a test which fails before this change and is fixed by it? As you mention, the existing coverage must not be sufficient to avoid regressions.

frontends/ast/simplify.cc
Xiretza: Ah, thats's what I was looking for! I tried `AST_BLOCK`, but that failed.

Xiretza:
Added a test, the asicworld one wasn't doing anything because it doesn't have any inputs or outputs.

zachjs (approved):
Thanks!

PR #2646, opened 2021-03-09
rtlil: Disallow 0-width chunks in SigSpec.

mwkmwkmwk:
Among other problems, this also fixes equality comparisons between
SigSpec by enforcing a canonical form.

Also fix another minor issue with possible non-canonical SigSpec.

Fixes #2623.

PR #2674, opened 2021-03-17
Replace assert in get_reference with more useful error message

Ravenslofty:
This assert caught the regression caused by #2646, so it deserves a promotion into a human-readable error.

### DIFF
--- a/kernel/rtlil.h
+++ b/kernel/rtlil.h
@@ -166,7 +166,8 @@ namespace RTLIL
 			log_assert(p[0] == '$' || p[0] == '\\');
 			log_assert(p[1] != 0);
 			for (const char *c = p; *c; c++)
-				log_assert((unsigned)*c > (unsigned)' ');
+				if ((unsigned)*c <= (unsigned)' ')
+					log_error("Found control character or space (0x%02hhx) in string '%s' which is not allowed in RTLIL identifiers\n", *c, p);
 
 		#ifndef YOSYS_NO_IDS_REFCNT
 			if (global_free_idx_list_.empty()) {


mwkmwkmwk:
Note: due to a Verilog lexer bug, this error is also trivially triggerable by putting a non-tab non-space character in an escaped identifier in source.

mmicko:
@Ravenslofty Seams that 0x%02hhx is not supported on mingw, so get lot of warnings like next one (probably will not work as well), seams that 0x%02x would only work in case of mingw compiler

In file included from ./kernel/yosys.h:296,
                 from kernel/driver.cc:20:
./kernel/rtlil.h: In static member function 'static int Yosys::RTLIL::IdString::get_reference(const char*)':
./kernel/rtlil.h:170:57: warning: unknown conversion type character 'h' in format [-Wformat=]
  170 |      log_error("Found control character or space (0x%02hhx) in string '%s' which is not allowed in RTLIL identifiers\n", *c, p);
      |                                                         ^
./kernel/rtlil.h:170:73: warning: format '%s' expects argument of type 'char*', but argument 2 has type 'int' [-Wformat=]
  170 |      log_error("Found control character or space (0x%02hhx) in string '%s' which is not allowed in RTLIL identifiers\n", *c, p);
      |                                                                        ~^                                                ~~
      |                                                                         |                                                |
      |                                                                         char*                                            int
      |                                                                        %d
./kernel/rtlil.h:170:16: warning: too many arguments for format [-Wformat-extra-args]
  170 |      log_error("Found control character or space (0x%02hhx) in string '%s' which is not allowed in RTLIL identifiers\n", *c, p);
      |                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

PR #2680, opened 2021-03-17
ast: Use better parameter serialization for paramod names.

mwkmwkmwk:
Calling log_signal is problematic for several reasons:

- with recent changes, empty string is serialized as { }, which violates
  the "no spaces in IdString" rule
- the type (plain / real / signed / string) is dropped, wrongly conflating
  functionally different values and potentially introducing a subtle
  elaboration bug

Instead, use a custom simple serialization scheme.


mwkmwkmwk:
by the way, the whole serialization scheme is also unsound because the parameter name is not escaped in any way, and could theoretically clash when pathologically prepared, but at least that cannot really be triggered by accident

PR #2683, opened 2021-03-18
Plug a few memory leaks

Xiretza:
I managed to find a few leaks in verilog/ast when running the testsuite under ASan, there's at least one more related to SV struct/enum typedefs that I wasn't able to track down.

### DIFF
--- a/frontends/verilog/verilog_parser.y
+++ b/frontends/verilog/verilog_parser.y
@@ -503,18 +503,19 @@ optional_comma:
 module_arg_opt_assignment:
 	'=' expr {
 		if (ast_stack.back()->children.size() > 0 && ast_stack.back()->children.back()->type == AST_WIRE) {
-			AstNode *wire = new AstNode(AST_IDENTIFIER);
-			wire->str = ast_stack.back()->children.back()->str;
 			if (ast_stack.back()->children.back()->is_input) {
 				AstNode *n = ast_stack.back()->children.back();
 				if (n->attributes.count(ID::defaultvalue))
 					delete n->attributes.at(ID::defaultvalue);
 				n->attributes[ID::defaultvalue] = $2;
-			} else
-			if (ast_stack.back()->children.back()->is_reg || ast_stack.back()->children.back()->is_logic)
-				ast_stack.back()->children.push_back(new AstNode(AST_INITIAL, new AstNode(AST_BLOCK, new AstNode(AST_ASSIGN_LE, wire, $2))));
-			else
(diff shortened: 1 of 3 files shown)


frontends/verilog/verilog_frontend.cc
```diff
@@ -492,6 +492,11 @@ struct VerilogFrontend : public Frontend {
 		frontend_verilog_yyparse();
 		frontend_verilog_yylex_destroy();
 
+		for (auto it : user_type_stack) {
+		        delete it;
+		}
+		user_type_stack.clear();
```
Xiretza, line 498: I'm not sure about this one, should these types be kept between different frontend invocations?

Xiretza: CI answered that for me, I moved it to the pass destructor.

frontends/verilog/verilog_parser.y
```diff
@@ -151,7 +151,9 @@ static void enterTypeScope()
 
 static void exitTypeScope()
 {
+	auto types = user_type_stack.back();
 	user_type_stack.pop_back();
+	delete types;
```
zachjs, line 156: There is a comment I recently added in `verilog_frontend.cc:add_package_types` discussing this leak that should probably be updated.

zachjs (commented):
The frontend changes make sense to me, but I'd like to have someone else take a look at the Makefile and test runner changes before merging.

frontends/verilog/verilog_parser.y
Xiretza: Something I just thought of: should `design -reset[-vlog]` drop the type stack as well?

zachjs: Yes, I agree that it should.

Xiretza: The whole user type stack handling actually needed a bit more work, I redid it in #2688

Xiretza:
I dropped the  `user_type_stack` freeing in favour of  #2688 and patched two more leaks in the parser.

zachjs (commented):
Frontend changes LGTM again. @whitequark Can you review the Makefile and test runner changes?

Xiretza:
@whitequark ping

zachjs:
@Xiretza Could you drop the makefile and test runner changes from this PR to get it back in scope? Let's also get this rebased to run against the latest tests. I'm going to make a push to clean up leaks in the frontend, and naturally I'd like to get this in.

Xiretza:
Should be trimmed down to just frontend code now.

PR #2685, opened 2021-03-18
Add simple CI using github actions.

mwkmwkmwk:


.github/workflows/test.yml
```diff
@@ -0,0 +1,87 @@
+name: Build and run tests
+
+on: [push, pull_request]
```
zachjs, line 3: Given a number of PRs come from branches within this repo, it may be worth changing this to something like the following. This will prevent such PRs from running through tests twice.
```
...
on:
  push:
    branches:
      - master
  pull_request:
jobs:
...
```

zachjs (commented):
Besides this one efficiency fix, we probably want to merge this in as-is to get it running. We can refine it (e.g., matrix builds, ccache) over time.

zachjs (approved):
Per Slack thread, the `push` vs. `pull_request` have meaningfully different behavior, so we'll be keeping both.

Issue #2684, opened 2021-03-18
SV function


Kmanfi:
## Steps to reproduce the issue

bug2.sv
```
package in_types_pkg;
typedef struct packed
{
    logic unsigned  [1:0]  m_v;
} in_struct_t;
endpackage

package out_types_pkg;
typedef struct packed
{
    logic unsigned  [1:0]  m_v;
} out_struct_t;
endpackage


function automatic out_types_pkg::out_struct_t e_s (
    input in_types_pkg::in_struct_t din
);
    out_types_pkg::out_struct_t dout;
    dout.m_v = din.m_v;
    return dout;
endfunction
```
`read_verilog -sv bug2.sv`
## Expected behavior

Read wo error.
## Actual behavior

bug2.sv:16: ERROR: syntax error, unexpected TOK_PKG_USER_TYPE

zachjs:
I have a PR incoming to add support for typenames as function return types. This was pretty easy to add.

Supporting `return` as used here could be straightforward, but the general case is far more complex. Supporting the jump keywords would be a significant undertaking. https://github.com/zachjs/sv2v does handle this already, and so may be a viable workaround.

Kmanfi:
So, "return" as it's in example bug2.sv is not supported yet?

zachjs:
> So, "return" as it's in example bug2.sv is not supported yet?

That's correct.

PR #2686, opened 2021-03-18
sv: allow typenames as function return types

zachjs:
The easy half of #2684

### DIFF
--- a/frontends/verilog/verilog_parser.y
+++ b/frontends/verilog/verilog_parser.y
@@ -861,6 +861,7 @@ task_func_decl:
 			outreg->children.push_back($4);
 			outreg->is_signed = $4->is_signed;
 			$4->is_signed = false;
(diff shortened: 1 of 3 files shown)


whitequark (approved):

frontends/verilog/verilog_parser.y
```diff
@@ -871,6 +872,10 @@ task_func_decl:
 	};
 
 func_return_type:
+	hierarchical_type_id {
+		$$ = new AstNode(AST_WIRETYPE);
+		$$->str = *$1;
```
Xiretza, line 877: I think this is missing a `delete $1`?

zachjs: Sure, I've added this. This `delete` is also missing from the `wire_type` production, so you may be interested in adding that to your other PR.

PR #2688, opened 2021-03-18
verilog: rebuild user_type_stack from globals before parsing file

Xiretza:
This was actually a ticking UB bomb: after running the parser, the type maps contain pointers to children of the current AST, which is recursively deleted after the pass has executed. This leaves the pointers in `user_type_stack` dangling, which just happened to never be a problem due to another bug that causes typedefs from higher-level type maps to never be considered.

Rebuilding the type stack from the design's globals ensures the `AstNode` pointers are valid.


zachjs (approved):
Thanks for cleaning this up! This makes quite a bit more sense to me.

PR #2689, opened 2021-03-18
verilog: check entire user type stack for type definition

Xiretza:
yosys only searches the current type scope for definitions of struct and union member type, and hits an assertion failure if the typedef is in a parent scope instead. While I'm not 100% certain, I'm pretty sure this behaviour is wrong.

Tests may be failing because of newly-uncovered UB fixed in #2688, so that should be merged first.

### DIFF
--- a/tests/svtypes/typedef_scopes.sv
+++ b/tests/svtypes/typedef_scopes.sv
@@ -4,6 +4,7 @@ typedef enum logic {s0, s1} outer_enum_t;
 
 module top;
 
+	// globals are inherited
 	outer_uint4_t u4_i = 8'hA5;
 	outer_enum_t enum4_i = s0;
 	always @(*) assert(u4_i == 4'h5);
@@ -17,13 +18,22 @@ module top;
 	always @(*) assert(inner_enum1 == 3'h3);
 
 	if (1) begin: genblock
+		// type declarations in child scopes shadow their parents
 		typedef logic [7:0] inner_type;
 		parameter inner_type inner_const = 8'hA5;
  		typedef enum logic [2:0] {s5=5, s6, s7} inner_enum_t;
+
 		inner_type inner_gb_i = inner_const; //8'hA5;
  		inner_enum_t inner_gb_enum1 = s7;
 		always @(*) assert(inner_gb_i == 8'hA5);
  		always @(*) assert(inner_gb_enum1 == 3'h7);
+
+		// check that copying of struct member types works over multiple type scopes
+		typedef struct packed {
+			outer_uint4_t x;
+		} mystruct_t;
+		mystruct_t mystruct;
+		always @(*) assert($bits(mystruct) == 4);
 	end
 
 	inner_type inner_i2 = 8'h42;
--- a/frontends/verilog/verilog_parser.y
+++ b/frontends/verilog/verilog_parser.y
@@ -163,12 +163,18 @@ static bool isInLocalScope(const std::string *name)
 
 static AstNode *getTypeDefinitionNode(std::string type_name)
 {
-	// return the definition nodes from the typedef statement
-	auto user_types = user_type_stack.back();
-	log_assert(user_types->count(type_name) > 0);
-	auto typedef_node = (*user_types)[type_name];
-	log_assert(typedef_node->type == AST_TYPEDEF);
-	return typedef_node->children[0];
+	// check current scope then outer scopes for a name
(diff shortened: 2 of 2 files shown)


tests/svtypes/global_member_type.ys
```diff
@@ -0,0 +1,21 @@
...
+endmodule
+EOF
+
+read_verilog -sv <<EOF
+module second;
+	typedef logic T;
+	typedef union {
+		S s;
+		T t;
+	} myunion;
```
zachjs, line 19: I'd be interested in seeing coverage for typedefs in more deeply nested scopes, i.e., within generate blocks. The changes seem reasonable, but some extra coverage will help us avoid regressions.

tests/verilog/typedef_across_files.ys
```diff
@@ -7,12 +7,15 @@ typedef T [3:0] S;
...
+typedef logic [7:0] T;
+
 module top;
     T t;
     S s;
     always @* begin
-        assert ($bits(t) == 1);
-        assert ($bits(s) == 4);
+        assert ($bits(t) == 8);
+        assert ($bits(s) == 8);
```
Xiretza, line 18: I'm no verilog person, is this expected behaviour? I'd expect it to be 32.

tests/svtypes/global_member_type.ys
Xiretza: I added a struct member type test to `typedef_scopes.sv`.

tests/verilog/typedef_across_files.ys
zachjs: Yosys doesn't handle multiple packed dimensions, so that's why you're seeing `8` in both cases. However, I believe identifiers within a typedef should be resolved at the site of their usage, rather than from the scope in which that typedef is instantiated, so the later re-definition of `T` shouldn't affect `S`. Further, this arguably loses the existing/previous test coverage, which isn't ideal.

The way globals are currently handled is fundamentally flawed. Modules should indeed be able to shadow typedefs, parameters, etc., without affecting the definition of other globals which depend on the shadowed parameters. In this sense, global vs. module scope is no different then outer vs. inner generate block scoping.

Xiretza: I dropped the globals redefinition issue for now and only kept the original commit (struct member types not searching up the type scope stack), I'll open an issue about the former.

tests/svtypes/typedef_scopes.sv
```diff
@@ -17,20 +19,33 @@ module top;
 	always @(*) assert(inner_enum1 == 3'h3);
 
 	if (1) begin: genblock
-		typedef logic [7:0] inner_type;
+		// type must be (forward) declared before use in order to parse, but
+		// its actual signature is always that of the typedef in the local scope
```
zachjs, line 23: I don't think this behavior is in line with the LRM, given that typdefs must be declared before they are used. This comment seems to imply that the outer definition forward declares the inner definition, which does follow from the implementation, but doesn't seem like the desired behavior. Because structs use `copyTypeDefinition`, this would actually resolve to the outer definition of the type name was used within a struct definition.

There are many limitations and issues with the current struct support. While I believe the code changes here are an improvement, I'd prefer this test didn't enforce this specific, unrelated, and unusual typdef behavior.

Perhaps I'm missing something significant here. Thoughts?

Xiretza: Ah, so this would be illegal according to the LRM?

```sv
typedef ... T;

module x;
	T t;
	typedef ... T;
endmodule
```

I just added the test to document current behaviour so it's obvious if possibly breaking changes are introduced. If it's not allowed by the LRM, I'll remove it, not much sense in documenting what's basically undefined behaviour.

zachjs: That is not illegal, but it doesn't behave the way the earlier iterations of this PR expected.

```SystemVerilog
typedef logic T;
module x;
	T a;
  	typedef logic [3:0] T;
  	T b;
  	initial $display("%0d %0d", $bits(a), $bits(b));
endmodule
```
Every simulator I have tried outputs `1 4` for the above example.


From Section 6.18:
> The declaration of a user-defined data type shall precede any reference to its type_identifier. User-defined data type identifiers have the same scoping rules as data identifiers, except that hierarchical references to type_identifier shall not be allowed.

From Section 6.5:
> Data shall be declared before they are used, apart from implicit nets (see 6.10).

zachjs (approved):

PR #2694, opened 2021-03-22
abc9: fix SCC issues

### LINKED ISSUE #2530: ABC9 flop bypass specify rules can cause combinational loops

eddiehung:
Fix #2530 -- specifically `(* abc9_box, lib_whitebox *)` cells that can be bypassed using a combinatorial-only box are no longer bypassed if they form part of a combinatorial loop (based on analysing connectivity only, not if it such a condition can be reached or will cause undefined behaviour -- such are the testcases from #2530 which occurs when the output of an async flop (which causes a bypass cell to capture async behaviour) fans back into to its async set/reset input).

Improve SCC handling overall. This is achieved by no longer marking combinatorial loops with `(* abc9_keep *)` and letting `write_xaiger` turn them into `inout`s -- this doesn't work if a box output drives a PO which cannot be turned into an `inout` -- so now use an explicit `$__ABC9_SCC_BREAKER` cell. 

Also tightens up some `abc9` testing.

CC: @Ravenslofty 

Please squash merge.


CODEOWNERS
```diff
@@ -16,6 +16,7 @@ backends/cxxrtl/               @whitequark
 passes/cmds/bugpoint.cc        @whitequark
 passes/techmap/flowmap.cc      @whitequark
 passes/opt/opt_lut.cc          @whitequark
+passes/techmap/abc9*.cc        @eddiehung
```
Ravenslofty, line 19: Might as well add yourself as code owner for `xaiger.cc` too.

passes/techmap/abc9_ops.cc
```diff
@@ -1587,11 +1606,11 @@ struct Abc9OpsPass : public Pass {
...
 		log("        certain required times.\n");
 		log("\n");
-		log("    -mark_scc\n");
+		log("    -break_scc\n");
 		log("        for an arbitrarily chosen cell in each unique SCC of each selected module\n");
-		log("        (tagged with an (* abc9_scc_id = <int> *) attribute), temporarily mark all\n");
-		log("        wires driven by this cell's outputs with a (* keep *) attribute in order\n");
-		log("        to break the SCC. this temporary attribute will be removed on -reintegrate.\n");
+		log("        (tagged with an (* abc9_scc_id = <int> *) attribute), temporarily interrupt\n");
+		log("        all wires driven by this cell's outputs with a temporary $__ABC9_SCC_BREAKER\n");
```
Ravenslofty, line 1612: Maybe a bit too nitpicky, but one "temporary" is probably enough.

Ravenslofty (approved):

Ravenslofty:
Yep, looks good.

As for squash merging, you're still part of the YosysHQ org; might as well merge it yourself.

sthornington:
Can confirm this fixes a 2530 combinational loop error I have been seeing since last fall some time.

PR #2696, opened 2021-03-22
split CodingReadme into multiple files

nakengelhardt:
`CodingReadme` is a collection of random snippets of information about the codebase targeting entirely different audiences. Split it up into separate files that can be referred to more easily.


guidelines/Checklists
```diff
@@ -0,0 +1,120 @@
...
+
+	cd ~yosys
+	make update-manual
+	make manual
+	- sanity check the figures in the appnotes and presentation
+	    - if there are any odd things -> investigate
+	    - make cosmetic changes to the .tex files if necessary
+
+	cd ~yosys
+	vi README guidelines/*
```
nakengelhardt, line 39: @clairexen adjust this to your personal preferences

guidelines/UnitTests
```diff
@@ -0,0 +1,69 @@
+How to add unit test
```
Xiretza, line 1: Is this even still relevant? `tests/unit/` has 3 commits, all from december 2016, and contains no actual tests.

nakengelhardt: The associated code still exists so it's there for now - it's by no means the only thing horribly out of date in these files.  (The unit test framework was a drive-by PR from someone who promised to go on to add many unit tests for yosys but then never did. It should probably be removed altogether.)

clairexen (approved):

PR #2703, opened 2021-03-25
quicklogic: add PolarPro3 BRAM inference

Ravenslofty:
Here's the next part of the quicklogic flow: block RAM.

### DIFF
--- a/techlibs/quicklogic/pp3_brams_map.v
+++ b/techlibs/quicklogic/pp3_brams_map.v
@@ -1,0 +1,1254 @@
+`default_nettype none
+
+module \$__QUICKLOGIC_RAMB16K (
+  CLK2,
+  CLK3,
+  A1ADDR,
+  A1DATA,
+  A1EN,
+  B1ADDR,
+  B1DATA,
+  B1EN
+);
+  parameter CFG_ABITS = 9;
+  parameter CFG_DBITS = 36;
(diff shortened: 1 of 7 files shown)


Ravenslofty:
Marked as draft because I'm waiting for QuickLogic to provide code to correctly use all 18 kilobits.

Ravenslofty:
Since the code to use it as an 18-kilobit RAM never materialised, and it's better to have this upstream than not, I have rebased it and am marking this as ready for review.

clairexen (changes requested):
Can we get a source code comment explaining why two of the BRAM data bits are unused?

techlibs/quicklogic/pp3_brams_map.v
```diff
@@ -0,0 +1,1254 @@
...
+
+  input [CFG_ABITS-1:0] A1ADDR;
+  output [CFG_DBITS-1:0] A1DATA;
+  input A1EN;
+
+  input [CFG_ABITS-1:0] B1ADDR;
+  input [CFG_DBITS-1:0] B1DATA;
+  input [CFG_ENABLE_B-1:0] B1EN;
+
+  wire VCC, GND;
```
mwkmwkmwk, line 32: these are completely unnecessary, just write the constants inline

PR #2704, opened 2021-03-25
verilog: fix case expression sign and width handling

### LINKED ISSUE #2528: Case labels with different width than case expression don't match

test.v:
```
module top(output reg a, b, c);
	always @* begin
		a = 0;
		b = 0;
		case (2'b0)
			1'b0:    a = 1;
			default: b = 1;
		endcase
		c = 1'b0 == 2'b0;
	end
endmodule

```
Run:
`yosys -p "synth_ice40 -abc2 -blif test.blif" test.v`

## Expected behavior

I expect that a==1, b==0 and c==1.

## Actual behavior

In the case statement equality seems to be determined differently. 2'b0 is not equal to 1'b0 in the case statement. But they are equal for the equality operator (==).
So the result I get is a==0, b==1 and c==1.

It seems to behave correctly if the case expression is not a literal or constant, like this:
```
wire [1:0] test_wire = 0;
...
case (test_wire)
	1'b0:  ...
	default:  ...
endcase

```
In this case the label 1'b0 will be chosen, even though its width doesn't match the two bit width of the wire.

zachjs:
- The case expression and case item expressions are extended to the maximum width among them, and are only interpreted as signed if all of them are signed
- Add overall width and sign detection for AST_CASE
- Add sign argument to genWidthRTLIL helper
- Coverage for both const and non-const case statements

From Section 12.5 of IEEE 1800-2017:

> In a _case_expression_ comparison, the comparison only succeeds when each bit matches exactly with respect to the values `0`, `1`, `x`, and `z`. As a consequence, care is needed in specifying the expressions in the case statement. The bit length of all the expressions needs to be equal, so that exact bitwise matching can be performed. Therefore, the length of all the _case_item_expressions_, as well as the _case_expression_, shall be made equal to the length of the longest _case_expression_ and _case_item_expressions_. If any of these expressions is unsigned, then all of them shall be treated as unsigned. If all of these expressions are signed, then they shall be treated as signed.

Fixes #2528

### DIFF
--- a/frontends/ast/ast.h
+++ b/frontends/ast/ast.h
@@ -291,7 +291,7 @@ namespace AST
 		// for expressions the resulting signal vector is returned
 		// all generated cell instances, etc. are written to the RTLIL::Module pointed to by AST_INTERNAL::current_module
 		RTLIL::SigSpec genRTLIL(int width_hint = -1, bool sign_hint = false);
-		RTLIL::SigSpec genWidthRTLIL(int width, const dict<RTLIL::SigBit, RTLIL::SigBit> *new_subst_ptr = NULL);
+		RTLIL::SigSpec genWidthRTLIL(int width, bool sgn, const dict<RTLIL::SigBit, RTLIL::SigBit> *new_subst_ptr = NULL);
 
 		// compare AST nodes
 		bool operator==(const AstNode &other) const;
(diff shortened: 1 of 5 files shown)


zachjs:
@clairexen Can I get a quick review of: A) the "hack" of adding "overall width detection" for `AST_CASE`; and B) the added parameter for `genWidthRTLIL`?

clairexen (approved):
LGTM

PR #2722, opened 2021-04-07
Add setundef -autoports

gatecat:
Some flows require all ports to be set to defined values, even those missing from instantiations. `setundef` on its own isn't enough to deal with this as the ports won't exist at all if not included in the instantiation, this adds a new mode to `setundef` to deal with this.

### DIFF
--- a/passes/cmds/setundef.cc
+++ b/passes/cmds/setundef.cc
@@ -146,6 +146,9 @@ struct SetundefPass : public Pass {
 		log("    -params\n");
 		log("        replace undef in cell parameters\n");
 		log("\n");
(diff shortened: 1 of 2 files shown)


mwkmwkmwk:
This seems to be another iteration of the same problem as #2390, where we decided to solve it by using default values on the ports in cell library instead; what do you think about that solution?  I'd rather avoid doing it in inconcistent ways between targets.

gatecat:
Yes, that approach seems reasonable. I'll close this PR as in the short term we ended up working around this in nextpnr anyway, in order to deal with cases of cells created during PnR too.

PR #2726, opened 2021-04-09
sf2: fix name of AND modules

Riesi:
When searching through the techlibs to reuse implementations I stumbled upon this file and got a bit confused by modules called `ADD` when they are in fact AND gates. Looking at the referenced pdf confirms that the modules should be called AND2, etc.

I grep'ed through the repo and luckily nothing else is referencing or using the wrong names, so the PR should be pretty much safe. (famous last words)

Kind regards,
Riesi


mwkmwkmwk (approved):
Good catch, thanks!

PR #2727, opened 2021-04-12
quicklogic: ABC9 LUT mapping/sequential synthesis

Ravenslofty:


techlibs/quicklogic/abc9_map.v
```diff
@@ -0,0 +1,24 @@
...
+  input CLR,
+  input PRE
+);
+
+parameter INIT = 1'b0;
+
+parameter _TECHMAP_CONSTMSK_CLR_ = 1'b0;
+parameter _TECHMAP_CONSTMSK_PRE_ = 1'b0;
+
+if (_TECHMAP_CONSTMSK_CLR_ != 1'b0 && _TECHMAP_CONSTMSK_PRE_ != 1'b0)
```
mwkmwkmwk, line 19: this should also look at the `CONSTVAL` to ensure that the values are const-0 and not const-1 (the latter shouldn't ever happen in normal flow given `opt_dff`, but let's keep our transforms sound)

techlibs/quicklogic/synth_quicklogic.cc
```diff
@@ -91,6 +97,10 @@ struct SynthQuickLogicPass : public ScriptPass {
 				verilog_file = args[++argidx];
 				continue;
 			}
+			if (args[argidx] == "-abc" && argidx+1 < args.size()) {
```
mwkmwkmwk, line 100: remove the `argidx+1...` part, this option doesn't take an argument

mwkmwkmwk (approved):

PR #2735, opened 2021-04-20
Separating source build instruction from ref to binary packages

smoe:
Gave the README a small push, pointing to more distributions that ship packages for yosys.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,20 @@ front-end for Yosys, SymbiYosys:
 Setup
 ======
 
+Installation
+------------
+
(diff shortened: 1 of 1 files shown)


nakengelhardt:
Distribution packages are often out of date or unmaintained (partially our fault for the long lull in releases between 0.9 and 0.10). We don't have the bandwidth to keep up with which ones are currently up to date, so we don't want to recommend them. OSS CAD Suite is the recommended way to get yosys binaries.

smoe:
You moved your repository without telling the maintainers of those packages, so they just worried.

nakengelhardt:
Ah, that must've been before my time, I'm guessing around 2017? But again, we don't have capacity to chase up other people, we can't even keep up with our own issue tracker...

PR #2738, opened 2021-04-27
Fix use of blif name in synth_xilinx command

mdko:
## Problem
Yosys ignores the name of the BLIF file passed in with the `synth_xilinx` command.

## Before
For example, 

```bash
yosys -p "synth_xilinx -top toplevel -blif counter.blif" counter.v
```

Yosys ignores the given BLIF name (`counter.blif`), and instead prints the BLIF output to the command line.

Screen Shot 2021-04-27 at 2 36 52 AM

## After
This fix now causes `counter.blif` to be written to.

Screen Shot 2021-04-27 at 2 36 10 AM


mmicko (approved):
Thanks

PR #2752, opened 2021-05-13
Add support for the SystemVerilog bind keyword

rswarbrick:
EDIT: This used to be a large patch set, but it's now mostly merged. The remaining commit has the following commit message:

**Implement bind elaboration in the hierarchy pass**

This is surprisingly finicky, because of the difference between `bind`
directives that target modules and those that target specific
instances of those modules (i.e. specific cells).

In the first case, we can just operate on the module in place. In the
second, we have to make fresh copies of the target module and the
"path of cells" from the bind directive down to it. For instance, in
something like this:

    bind $root.foo.bar_i.baz_i qux bound_i (.*);

(where `bar_i` and `baz_i` are instances of modules called `bar` and `baz`,
respectively), we need to make a fresh clone of `baz` (let's call it
`baz1`) into which we insert an instance of `qux`. But we can't just
modify bar to re-type its `baz_i` cell from `baz` to `baz1`: what if there
are other instances of `bar` in the design? So we then need to make a
fresh clone of `bar` (let's call it `bar1`). Since `foo` here refers to the
module, rather than a particular instance, we *can* now modify `foo` to
re-type it's `bar_i` cell from `bar` to `bar1` (phew!).

To avoid some extra un-needed renames, we actually count the number of
times each cell type appears in the design (see `get_cell_type_counts`).
If the design in the example above didn't actually use `bar` except as
`bar_i` in module `foo`, we actually operate in place rather than
renaming.

### DIFF
--- a/kernel/binding.cc
+++ b/kernel/binding.cc
@@ -24,6 +24,163 @@ YOSYS_NAMESPACE_BEGIN
 RTLIL::Binding::Binding(RTLIL::IdString target_type,
                         RTLIL::IdString target_name)
 	: target_type(target_type), target_name(target_name)
-{}
+{
+	// A counter of the number of bindings that have been created so far.
+	static int binding_counter;
+	attr_name = stringf("\\bound%d", binding_counter++);
+}
+
+// Return true if mod is either called name or is derived from an abstract
+// module that was called name.
+static bool
+module_original_type_matches(const RTLIL::Module &mod, RTLIL::IdString name)
+{
+	if (mod.name == name) return true;
+
+	// A module derived from 'foo' will have a name starting with $paramod
+	// and an 'hdlname' attribute equal to '\foo'.
+	if (!mod.name.begins_with("$paramod"))
+		return false;
+
+	RTLIL::IdString hdlname = mod.get_string_attribute(ID::hdlname);
+	log_assert(!hdlname.empty());
+	return (hdlname == name);
+}
+
+// Search through the design for modules called name, or that are derived from
+// one called name.
+static std::vector<RTLIL::Module *>
+find_concrete_modules(RTLIL::Design &design, RTLIL::IdString name)
+{
+	std::vector<RTLIL::Module *> mods;
+
+	for (const auto &pr : design.modules_) {
+		if (module_original_type_matches(*pr.second, name))
+			mods.push_back(pr.second);
+	}
+
+	return mods;
+}
+
+// Try to find a cell below mod whose (dot-separated) name is given by path.
+// Modify so_far in place and return true on success.
+static bool
+find_cell_in(RTLIL::Binding::Path *so_far,
+             RTLIL::Design &design, RTLIL::Module &mod, const char *path)
+{
+	const char *dot = strchr(path, '.');
+	std::string cell_name =
+		dot ? std::string(path, dot - path) : std::string(path);
+
+	RTLIL::Cell *cell = mod.cell('\\' + cell_name);
+	if (!cell)
+		return false;
+
+	so_far->push_back(RTLIL::Binding::Item(&mod, cell));
+	if (!dot)
+		return true;
+
+	// We found a child cell, but there's more searching to do. Recurse to
+	// search the cell's module.
+	return find_cell_in(so_far, design, *cell->module, dot + 1);
+}
+
+RTLIL::Binding::Path
+RTLIL::Binding::find_rel_cell(RTLIL::Design &design, RTLIL::Module &start) const
+{
+	Path ret;
+	bool found = find_cell_in(&ret, design, start, target_name.c_str() + 1);
+	if (!found)
+		return Path();
+
+	log_assert(ret.size());
+	check_cell_type(design, *ret.back().second);
+	return ret;
+}
+
+std::vector<RTLIL::Binding::Path>
+RTLIL::Binding::find_tl_cells(RTLIL::Design &design) const
+{
+	std::vector<Path> paths;
+
+	std::string tgt_name = target_name.str();
+	size_t tgt_off = 0;
+
+	// The target name should always start with a backslash
+	log_assert(tgt_name[0] == '\\');
+
+	// Is the name absolute (starts with $root)? If so, just skip past that:
+	// we're at the top already.
+	if (tgt_name.compare(0, 7, "\\$root.") == 0)
+		tgt_off = 7;
+	else
+		tgt_off = 1;
+
+	// This function can only find cells (as opposed to modules at the top
+	// level). If there's no '.' after tgt_off, give up.
+	size_t dot_off = tgt_name.find('.', tgt_off);
+	if (dot_off == std::string::npos)
+		return paths;
+
+	// Otherwise, characters with indices {tgt_off .. dot_off - 1} should
+	// name a module and then {dot_off + 1 .. } should name the cell
+	// relative to the module.
+	//
+	// Note that we need to prefix the module name with a backslash (which
+	// is separated from the actual module name if the name started with
+	// $root) before interning the string.
+	std::string mod_name = '\\' + tgt_name.substr(tgt_off, dot_off - tgt_off);
+	const char *cell_path = tgt_name.c_str() + dot_off + 1;
+
+	std::vector<RTLIL::Module *> mods = find_concrete_modules(design, mod_name);
+	for (RTLIL::Module *mod : mods) {
+		Path path;
+		bool found = find_cell_in(&path, design, *mod, cell_path);
+		if (found) {
+			log_assert(path.size());
+			check_cell_type(design, *path.back().second);
+			paths.push_back(path);
+		}
+	}
+
+	// If no concrete modules for mod_name have a cell at cell_path, then
+	// paths will be an empty vector. If *some* have the cell (possible with
+	// a generate if/else or similar), we should probably complain.
+	if (paths.size() > 0 && paths.size() != mods.size()) {
+		log_error("Some, but not all, specialized modules "
+		          "with original name %s have a cell at path %s.",
+		          mod_name.c_str(), cell_path);
+	}
+
+	return paths;
+}
+
+std::vector<RTLIL::Module *>
+RTLIL::Binding::find_concrete_module_targets(RTLIL::Design &design) const
+{
+	// If target_type is nonempty, the bind statement had an instantiation
+	// list. In which case, we definitely don't want to interpret this as a
+	// module name.
+	if(!target_type.empty())
+		return std::vector<RTLIL::Module *>();
+
+	return find_concrete_modules(design, target_name);
+}
+
+void
+RTLIL::Binding::check_cell_type(RTLIL::Design &design,
+                                const RTLIL::Cell &cell) const
+{
+	// If the user specified the module type, check it matches
+	if (!target_type.empty()) {
+		const RTLIL::Module *cell_mod = design.module(cell.type);
+		log_assert(cell_mod);
+		if (!module_original_type_matches(*cell_mod, target_type)) {
+			log_error("The %s matches a cell, "
+				  "but that cell has the wrong type (%s).",
+				  describe().c_str(), cell_mod->name.c_str());
+		}
+	}
+}
 
 YOSYS_NAMESPACE_END
--- a/passes/hierarchy/hierarchy.cc
+++ b/passes/hierarchy/hierarchy.cc
@@ -19,6 +19,7 @@
  */
 
 #include "kernel/yosys.h"
+#include "kernel/binding.h"
 #include "frontends/verific/verific.h"
 #include <stdlib.h>
 #include <stdio.h>
@@ -439,7 +440,7 @@ void check_cell_connections(const RTLIL::Module &module, RTLIL::Cell &cell, RTLI
 	}
 }
 
-bool expand_module(RTLIL::Design *design, RTLIL::Module *module, bool flag_check, bool flag_simcheck, std::vector<std::string> &libdirs)
+bool expand_module(RTLIL::Design *design, RTLIL::Module *module, bool flag_check, bool flag_simcheck, const std::vector<std::string> &libdirs)
 {
 	bool did_something = false;
 	std::map<RTLIL::Cell*, std::pair<int, int>> array_cells;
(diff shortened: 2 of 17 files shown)


rswarbrick:
@zachjs: Thanks for the reviews you've already done for the component parts. I'll rebase and force-push to this PR as things get merged (or abandoned!) to make it easier to track what's still pending.

rswarbrick:
Force-push rebases onto master and drops #2753 (abandoned).

rswarbrick:
Ping! It's been a month: is there anything I can do to help get this reviewed and merged? Thanks, @zachjs, for looking at the frontend changes.

mmicko:
@rswarbrick Can you please update copyright messages ? ( please take a look at #2817 to see what is changed) Thanks.

rswarbrick:
Yep, happily. I think this is just the copyright headers in the new modules that got added, right? If so, everything should be updated now.

rswarbrick:
Incidentally, for proof that this actually works, I just managed to convert the Ibex icache FPV stuff that I'd done to bind the testbench in. It was surprisingly painless and (astonishingly) did something sensible. See https://github.com/lowRISC/ibex/pull/1382 for details.

rswarbrick:
Woo, just one commit to go! :-) Marking this PR as ready for review accordingly.

rswarbrick:
Hey all. It's been a couple of weeks. Is there anything I can do to help get this over the line?

tests/bind/basic.sv
```diff
@@ -8,13 +16,11 @@ module bar (input a, input b, output c);
...
+module act (input u, input v, output w);
   foo foo_i (.a (u), .b (v), .c (w));
-
   bind foo bar bound_i (.*);
+endmodule
 
-  always_comb begin
-    assert(w == u ^ v);
-  end
+module ref (input u, input v, output w);
```
zachjs, line 24: `ref` is a SystemVerilog keyword. While Yosys doesn't support it, we should avoid using it as an identifier in .sv test cases.

kernel/rtlil.cc
```diff
@@ -922,9 +922,9 @@ void RTLIL::Module::makeblackbox()
 	set_bool_attribute(ID::blackbox);
 }
 
-void RTLIL::Module::reprocess_module(RTLIL::Design *, const dict<RTLIL::IdString, RTLIL::Module *> &)
+void RTLIL::Module::expand_interfaces(RTLIL::Design *, const dict<RTLIL::IdString, RTLIL::Module *> &)
```
zachjs, line 925: Given we're adjusting all of the existing `reprocess_module` stuff, I think we should just use `dynamic_cast` at the call site and avoid this virtual-but-not-really shenanigans.

zachjs (commented):
These are just two things which jumped out at me on first glance. I'll do a full review tomorrow.

kernel/rtlil.cc
rswarbrick: I agree, but I did this to match the code for `derive`. Are you happy if we tidy this up as a follow-on?

tests/bind/basic.sv
rswarbrick: Good point! Fixed (now "gold")

kernel/binding.cc
```diff
@@ -24,6 +24,163 @@ YOSYS_NAMESPACE_BEGIN
...
+
+	// This function can only find cells (as opposed to modules at the top
+	// level). If there's no '.' after tgt_off, give up.
+	size_t dot_off = tgt_name.find('.', tgt_off);
+	if (dot_off == std::string::npos)
+		return paths;
+
+	// Otherwise, characters with indices {tgt_off .. dot_off - 1} should
+	// name a module and then {dot_off + 1 .. } should name the cell
+	// relative to the module.
```
zachjs, line 127: For target names which contain a dot, they must be a `bind_target_instance`, as `bind_target_scope` cannot contain a dot. So in this case, the target name must "specify a single instance into which the `bind_instantiation` should be inserted."

For bind directives within modules, when the target name has a dot, other tools do not allow the leading portion of the identifier to refer to a module rather than an instance. For bind directives in the compilation unit scope, when the target name has a dot, they disallow binding into a module which is instantiated twice. I believe these interpretations of a hierarchical identifier and the `bind` spec are accurate.

So, given the following modules
```SystemVerilog
module foo;
  initial $display("hello from foo");
endmodule
module bar;
  initial $display("hello from bar");
endmodule
module act;
  initial $display("hello from act");
  foo f();
endmodule
```

I believe each of the following cases should be **illegal**:
```SytemVerilog
module beep;
  act act1();
  //act act2();
endmodule
module boop;
  beep b1();
  beep b2();
endmodule
bind act.f bar b();
```
```SytemVerilog
module beep;
  act act1();
  act act2();
endmodule
module boop;
  beep b1();
endmodule
bind act.f bar b();
```
```SytemVerilog
module beep;
  act act1();
  act act2();
  bind act.f bar b();
endmodule
```

Whereas each of the following cases are **legal**:
```SystemVerilog
module beep;
  act act1();
endmodule
module boop;
  beep b1();
endmodule
bind act.f bar b();
```
```SystemVerilog
module beep;
  act act1();
endmodule
module boop;
  beep b1();
endmodule
bind act bar b();
```
```SystemVerilog
module beep;
  act act1();
  act act2();
endmodule
module boop;
  beep b1();
endmodule
bind beep.act1.f bar b();
```
```SytemVerilog
module beep;
  act act1();
  act act2();
  bind act1.f bar b();
endmodule
```

passes/hierarchy/hierarchy.cc
```diff
@@ -710,6 +711,411 @@ RTLIL::Wire *find_implicit_port_wire(Module *module, Cell *cell, const std::stri
...
+
+	// Now walk the design's modules, looking for binding statements. We
+	// walk all the modules in the design, regardless of whether we have a
+	// top module.
+	//
+	// Note that binding statements aren't allowed to nest (you can't bind
+	// something into a module that you bound in earlier), so we don't have
+	// to do any clever iterative thing.
+	std::vector<RTLIL::Module *> to_visit;
+	std::copy(design.modules().begin(), design.modules().end(),
```
zachjs, line 1087: Can't we just `to_visit = design.modules();` given the `vector` cast operator of `ObjRange`?

passes/hierarchy/hierarchy.cc
```diff
@@ -710,6 +711,411 @@ RTLIL::Wire *find_implicit_port_wire(Module *module, Cell *cell, const std::stri
...
+get_cell_type_counts_worker(RTLIL::Design &design,
+                            dict<RTLIL::IdString, int> &counts,
+                            RTLIL::Module &mod)
+{
+	for (auto cell : mod.cells()) {
+		// Set counts[cell->type] to 1 if it isn't defined
+		// yet, or increment if it is.
+		auto pr = counts.insert(std::make_pair(cell->type, 1));
+		if (!pr.second)
+			++pr.first->second;
```
zachjs, line 1032: It seems like this is equivalent to simply `counts[cell->type]++`, given the expected behavior of `operator[]` when not present in the `dict`.

passes/hierarchy/hierarchy.cc
```diff
@@ -710,6 +711,411 @@ RTLIL::Wire *find_implicit_port_wire(Module *module, Cell *cell, const std::stri
...
+		// We can add the binding directly to the module, operating in place.
+		bool did_something = false;
+		log("Applying %s to %d module%s.\n",
+		    desc.c_str(), GetSize(mods), mods.size() > 1 ? "s" : "");
+		for (auto mod : mods)
+			did_something |= binding.bind_into(design, *mod);
+		return did_something;
+	}
+
+	log_error("Couldn't find a target for %s.\n", desc.c_str());
```
zachjs, line 1015: Does the bind directive have a source location we could log with this error? If so, please update this and the other `log_error`s added elsewhere.

zachjs: It might be nice to add some `logger -expect error` test cases for these errors.

zachjs (changes requested):

kernel/binding.cc
```diff
@@ -24,6 +24,163 @@ YOSYS_NAMESPACE_BEGIN
...
+	return mods;
+}
+
+// Try to find a cell below mod whose (dot-separated) name is given by path.
+// Modify so_far in place and return true on success.
+static bool
+find_cell_in(RTLIL::Binding::Path *so_far,
+             RTLIL::Design &design, RTLIL::Module &mod, const char *path)
+{
+	const char *dot = strchr(path, '.');
```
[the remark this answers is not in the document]
rswarbrick: Oh, that's a good point. I hadn't thought about generate blocks. How are these represented in the RTLIL? Do we just have cells with names like `a.b`? (I *think* that's what I can see in `simplify.cc`, but wanted to check before I started hacking).

kernel/binding.cc
```diff
@@ -24,6 +24,163 @@ YOSYS_NAMESPACE_BEGIN
...
+
+	std::string tgt_name = target_name.str();
+	size_t tgt_off = 0;
+
+	// The target name should always start with a backslash
+	log_assert(tgt_name[0] == '\\');
+
+	// Is the name absolute (starts with $root)? If so, just skip past that:
+	// we're at the top already.
+	if (tgt_name.compare(0, 7, "\\$root.") == 0)
```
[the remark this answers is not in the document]
rswarbrick: Ah, I think it *is* imposed, because of which function we're in. The `find_tl_cells` method is supposed to split something like `A.B.C` into "the cell at path B.C within module A". That's what the stuff to compute `mod_name` and `cell_path` does just below. To interpret `A.B.C` as "the cell at path A.B.C relative to the current point" needs the `find_rel_cell` method. There are some comments explaining what's supposed to happen in `binding.h`.

Or have I misunderstood your message?

kernel/binding.cc
rswarbrick: > For bind directives within modules, when the target name has a dot, other tools do not allow the leading portion of the identifier to refer to a module rather than an instance.

Hmm. If I understand the spec correctly, both are explicitly allowed. Page 709 of IEEE 1800-2017 (section 23.3.1) has (emphasis mine):
>> `$root` allows explicit access to the top of the instantiation tree. This is useful to disambiguate a local path
(which takes precedence) from the rooted path. If `$root` is not specified, a hierarchical path is ambiguous.
**For example, `A.B.C` can mean the local `A.B.C` or the top-level `A.B.C` (assuming there is an instance A that
contains an instance B at both the top level and in the current module).** The ambiguity is resolved by giving
priority to the local scope and thereby preventing access to the top-level path. `$root` allows explicit access
to the top level in those cases in which the name of the top-level module is insufficient to uniquely identify
the path.

In the `find_tl_cells` method (where this comment points!), we only handle the top-level case. But we only call it from `insert_binding` in `hierarchy.cc` if `find_rel_cell` failed (handling the local case). I *think* that this matches the spec?

Looking one-by-one at the cases you listed:

For (1) - (3), I think that `bind act.f bar b();` parses the same in each case. `act.f` matches the hierarchical identifier in the BNF's `bind_target_instance`. The interpretation in (1) and (2) is forced to be "treat `act` as a module and point at the `f` instance in that top-level module". In (1), we get one copy and in (2), we get 2 copies. In (3), we perform a local search for `act.f` which fails and then fall back to the top-level interpretation. (A *horrible* design, but I think it's legal...)

For (4)-(7), I agree that they are legal (and I think we probably agree about what they mean).

passes/hierarchy/hierarchy.cc
rswarbrick: Oh, neat! Will change.

passes/hierarchy/hierarchy.cc
rswarbrick: If the key isn't present in the dictionary, I think the value gets default-constructed. But I don't think there's any guarantee that an `int` will be constructed as zero is there?

passes/hierarchy/hierarchy.cc
rswarbrick: Oh, good idea. I'll do that.

rswarbrick:
@zachjs: Thank you very much for the careful review. That must have taken quite some time: thanks! I've just force-pushed with the suggested simplification for how we construct `to_visit` (which can actually be simplified further, just iterating over `design.modules()` directly: I think I was making a worklist here originally, but it seems that's not needed).

I'll follow up with a nicer message for the "couldn't find a target" error and a test for it. I've responded with some queries for the other comments: when you have a sec, could you have a look?

kernel/binding.cc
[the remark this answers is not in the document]
zachjs: Yes, that's how they're represented.

kernel/binding.cc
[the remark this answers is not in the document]
zachjs: For example, the following is currently accepted, but shouldn't be, as `act` does not lie at the root of the instantiation tree.
```SystemVerilog
module foo;
  initial $display("hello from foo");
endmodule
module bar;
  initial $display("hello from bar");
endmodule
module act;
  initial $display("hello from act");
  foo f();
endmodule
module beep;
  act act1();
endmodule
module boop;
  beep b1();
endmodule
bind $root.act.f bar b();
```

passes/hierarchy/hierarchy.cc
zachjs: Given the general requirement that map values should be default constructible, I would think it would be initialized to zero. I think I have seen this behavior relied upon in the past, but I could be wrong. This is what `dict` uses in the not-present case:
```C++
i = do_insert(std::pair<K, T>(key, T()), hash);
```

While I'm not opposed to doing a clearer, unobjectionable thing, I am very interested in learning what actually happens here.

clairexen: > If the key isn't present in the dictionary, I think the value gets default-constructed. But I don't think there's any guarantee that an `int` will be constructed as zero is there?

Yes, the in default constructor sets it to zero.

However, notice that you have to run in. If you have a class member that is an int, then it's only default constructed when the constructor is called, otherwise it is left uninitialized:
https://godbolt.org/z/zPPYP5dTb

In the case of a `dict<T, int>` (or `unordered_map<T, int>` and other std containers) the values are always default constructed, i.e. ints are set to zero.

kernel/binding.cc
zachjs: You have convinced me that properly hierarchical `bind`s within a module may fall back to looking up a module rather than an instance. Hence, each of the following should and do work in this PR and in some commercial suites as well.
```SystemVerilog
module beep;
  act act1();
  bind act.f bar b();
endmodule
```
```SystemVerilog
module beep;
  act act1();
endmodule
module boop;
  beep b1();
  bind act.f bar b();
endmodule
```

> In (1), we get one copy...

There are two instances of `beep`, so in that test case, there are two instances of `act`. Indeed, there are two instances of `act` in all of (1) thru (3). I think other tools are relying on the "specify a **single instance** into which the bind_instantiation should be inserted" language to avoid having to modify multiple instances in these cases. I do think it would be a bit strange for a single `hierarchical_identifier` to referring to multiple cells.

kernel/binding.cc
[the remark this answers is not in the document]
rswarbrick: You're completely right! I hadn't been distinguishing between "top-level, so a module name" and "the top of the instantiation tree". Will fix.

passes/hierarchy/hierarchy.cc
rswarbrick: Oh, neat! I've just learned something, thanks :-)

rswarbrick: Incidentally, the code just below this note was looking at `pr.second` to see whether we'd seen a match, which isn't possible any more. The obvious translation looks like it might do a second lookup and (after compiling and looking at the generated binary) it seems that it does :-(

The result is simpler and more clearly right, so I'll switch anyway but it's a pity that GCC (10.3.8) can't fold this away.

kernel/binding.cc
rswarbrick: You're right that I can't count :-) But, ignoring that, can you explain in what way you think this code's behaviour should change? At the moment, it matches how I've understood the spec: presumably we disagree on some point there, but I'm not quite sure which one!

rswarbrick:
I've just pushed up a version that I think should:
- handle generate blocks properly (https://github.com/YosysHQ/yosys/pull/2752#discussion_r701257761)
- be more strict about `$root` (https://github.com/YosysHQ/yosys/pull/2752#discussion_r703419075)
- have a simpler way to update the counts map (https://github.com/YosysHQ/yosys/pull/2752#discussion_r700518874)

I'm going to mark those discussions resolved above, to try to avoid the PR growing without bound(!). If you don't think this resolves things properly, please feel free to un-resolve them!

Still to do:
- Reach agreement on the exact semantics of hierarchical references when they fall back to the module level (https://github.com/YosysHQ/yosys/pull/2752#discussion_r703581466)
- Improve error message, and add a test, for failing binds (https://github.com/YosysHQ/yosys/pull/2752#discussion_r700521057)

I didn't rebase, so clicking on "force-pushed" above should give a useful [diff](https://github.com/YosysHQ/yosys/compare/7462113864542342e97cf8bfaf1fe8c04a6ab0de..c117f9a9a71cdc4aa4c3d0e0609088446208c040).

passes/hierarchy/hierarchy.cc
rswarbrick: Ahah! Now I know why I didn't do that! The problem is that we modify design when iterating over the modules. Thinking harder, I'm not sure my original code was quite right: if we bind something into a module that itself specifies a binding, I think we need to make sure we run this on the modified version of the module. The latest force-push should get this right.

kernel/binding.cc
zachjs: The spec appears to state that a `bind_target_instance` is used to "specify a **single** instance." But in (1) thru (3) above, it is being used to specify **multiple** distinct instances. I have not been able to find any tool which accepts (1) thru (3), and I suspect the reason is because of this restriction.

rswarbrick: Ah, I think I understand. To implement the restriction, I guess we could propagate some "instantiation count" down from the top of the tree and only allow a hierarchical bind via module if that count happened to be 1. But it's pretty complicated and only enforces a restriction that's kind of arguable :-)

My reading of the spec's "*The second form of bind syntax can be used to specify a single instance into which the bind_instantiation should be inserted.*" is, I think, slightly different from yours. Having used `bind` in Verilator (which doesn't support this form), the annoying problem is when you have multiple instances of a module and you only want to bind into some of them. I think this form is to support that use-case. For example, suppose you wanted to do something like "bind this protocol checker into the driver for each of my AXI blocks, but not into the driver when used in in my other interfaces". There, you want a target that's something like `axi_block_module.driver_instance`.

The "just one instance in the design" interpretation would mean that's not allowed if you instantiate `axi_block_module` more than once. Instead, you'd have to write something like `$root.path.to.axi_block_instance0.driver_instance` and `$root.path.to.axi_block_instance1.driver_instance`, which is a bit horrible: it encodes the number of AXI blocks (easy to forget if you add one!) and the exact design hierarchy (causing churn when `path.to` changes).

My take is:
- The spec is slightly ambiguous here. It might really mean "just one instance in the design" or it might not.
- If it doesn't mean that, the proposed code is correct.
- If it does, the proposed code will work for any correct design and will do something reasonable if the design is invalid. Another tool might choose to choke on the invalid design, but there's no requirement for us to do so.

So I *think* I'd stick with the current code, which is in keeping with Yosys's error handling approach in general ("don't try too hard; this is a synthesis tool, not a simulator"). Of course, I wrote it, so I'm biased :-D If you disagree, do you have a suggestion for a simple way to enforce the restriction that you think should hold?

rswarbrick: Oh, I meant to say: you wrote: "I have not been able to find any tool which accepts (1) thru (3), and I suspect the reason is because of this restriction.". When you tried it out, what were the error messages you saw? Were they of the form "No, you fool! There are multiple instances of X" or were they "Aargh. Internal error". In the former case, that's a strong argument. In the latter, not so much... :-)

rswarbrick:
Hi all. I've just added a source location to the error message and added a test showing what's going on, for the cell type check and for the "couldn't find a target for..." error.

I'm not sure that it makes so much sense for the "Cannot bind into module because it didn't originally come from RTL" error, so I've left that. There's also the "Some, but not all..." error. I think this is only possible to hit if you allow hierarchical references starting with module names, which is what we're discussing in https://github.com/YosysHQ/yosys/pull/2752#discussion_r700503736. I thought I'd wait until we reach a resolution there before working on that.

I'll leave https://github.com/YosysHQ/yosys/pull/2752#discussion_r700521057 unresolved for now so that I don't forget to come back to the "Some, but not all..." error (assuming we don't completely change that code).

As before, I haven't rebased, so the [force-pushed](https://github.com/YosysHQ/yosys/compare/c117f9a9a71cdc4aa4c3d0e0609088446208c040..1f96b69272994f57948736d9ca1b5eb4e371ddb5) link above should give a helpful diff.

kernel/binding.cc
zachjs: I'm away from my computer at the moment, so I won't be able to re-run to grab those error messages. If you're interested, I often use edaplayground.com, which provides limited access to some commercial simulators for students, researchers, etc.

Regarding how to specify a limited set of modules (especially locally), rather than the following:
```SystemVerilog
module beep;
  act act1();
  act act2();
  bind act.f bar b();
endmodule
```
The spec seems to require something like:
```SystemVerilog
module beep;
  act act1();
  act act2();
  bind act1.f bar b();
  bind act2.f bar b();
endmodule
```

I think you may be right that this behavior wouldn't seem to affect definitely legal source, but I'd like to poke around a little more later today when I get the chance.

You described a rather complex strategy for implementing this restriction. I agree something that complex would be undesirable. Would it be possible to do it by erroring whenever we match multiple modules via a properly hierarchical identifier, i.e., one with dot(s)? That is the sort of restriction I had in my head, which seems like it might make certain pieces even simpler.

zachjs: Here are some relevant test results run against three different commercial simulators.

```SystemVerilog
module beep;
	act act1();
	bind act.f bar b();
endmodule
```
A) `Hierarchical name component lookup failed for 'act' at 'beep'`
B) `Failed to find 'act' in hierarchical name act.f`
C) (no complaint!)





```SystemVerilog
module beep;
	act act1();
	act act2();
	bind act.f bar b();
endmodule
```
A) `Hierarchical name component lookup failed for 'act' at 'beep'`
B) `Failed to find 'act' in hierarchical name act.f`
C) `Can only bind to modules or instances`


```SystemVerilog
module beep;
	act act1();
endmodule
bind act.f bar b();
```
A) `Hierarchical name component lookup failed for 'act' at '$unit_0x4ccdf83b'`
B) `Failed to find 'act' in hierarchical name act.f`
C) no error



```SystemVerilog
module beep;
	act act1();
	act act2();
endmodule
bind act.f bar b();
```
A) `Hierarchical name component lookup failed for 'act' at '$unit_0x4ccdf83b'`
B) `Failed to find 'act' in hierarchical name act.f`
C) `Can only bind to modules or instances`

rswarbrick:
Hi there! Sorry, I've not had any time to spend on it recently. I think that the remaining technical issue is how we enforce a requirement that binds that go through hierarchical references can only hit one thing. I had a bit of a look, and I think we'd need to do another count in the elaboration pass to figure out how many targets there are. It ends up being "how many instances of this module are there in the elaborated design". Then, to work out whether bind to `my_module.foo.bar.baz` is allowed, you need to iterate over all specializations of `my_module` and add up the counts. If the result is exactly one, you're good.

This seems quite hard work(!), and I'm not convinced it provides any actual benefit. Can you think of a cleaner way to do things?

passes/hierarchy/hierarchy.cc
```diff
@@ -710,6 +711,422 @@ RTLIL::Wire *find_implicit_port_wire(Module *module, Cell *cell, const std::stri
...
+// Returns a dictionary keyed by module name. Every module will appear in the
+// dictionary, plus any other cell that's used.
+static dict<RTLIL::IdString, int>
+get_cell_type_counts(RTLIL::Design &design)
+{
+	dict<RTLIL::IdString, int> counts;
+	for (auto mod : design.modules()) {
+		counts.insert(std::make_pair(mod->name, 0));
+		get_cell_type_counts_worker(design, counts, *mod);
+	}
```
zachjs, line 1065: After poking around some more, it seems this procedure can double count. Running `hierarchy -top c` with the below modules loaded, the count of `a` internally becomes 2 because `b` is itself visited twice.

```SystemVerilog
module m;
endmodule
module a;
endmodule
module b;
    a x();
    bind x m z();
endmodule
module c;
    b y();
endmodule
```

passes/hierarchy/hierarchy.cc
```diff
@@ -710,6 +711,422 @@ RTLIL::Wire *find_implicit_port_wire(Module *module, Cell *cell, const std::stri
...
+
+		if (count > 1) {
+			split_point = i;
+			break;
+		}
+	}
+
+	// If split_point == -1, the target module and all its parents appear
+	// just once, so we can operate in place.
+	return binding.bind_into(design, *orig_module);
```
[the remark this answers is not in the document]
zachjs: Both with and without this condition manually patched locally, the following script fails on each of the following sources, despite working as expected in other tools.

```
read_verilog -defer -sv foo.sv
hierarchy -top top
proc
flatten
opt
select -module top
sat -verify -seq 1 -prove-asserts -show-all
```


```SystemVerilog
module mod(input wire clk);
    wire x;
    assign x = 1;
endmodule

module val(input wire i);
    always @* assert (i == 0); // should fail
endmodule

module foo(input wire clk);
    mod y(clk);
    bind y val v(x);
endmodule

module bar(input wire clk);
    mod z(clk);
    bind z val v(x);
endmodule

module top;
    wire clk;
    foo f(clk);
    bar b(clk);
endmodule
```


```SystemVerilog
module mod(input wire clk);
    wire x;
    assign x = 1;
endmodule

module val(input wire i);
    always @* assert (i == 0); // should fail
endmodule

module foo(input wire clk);
    mod y(clk);
endmodule

module bar(input wire clk);
    mod z(clk);
endmodule

module top;
    wire clk;
    foo f(clk);
    bar b(clk);
    bind f.y val v(x);
    bind b.z val v(x);
endmodule

```

rswarbrick:
Hi all! Sorry for the ridiculous delay in getting back to this. I've just force-pushed, but there are no interesting changes: it's just a rebase onto latest master. I'm going to go through the rest of the discussion now and try to move things forward, but I thought it might be worth rebasing first (if nothing else, it means that the force-push diffs that follow will be a bit more informative).

kernel/binding.cc
rswarbrick: Ahah! I think I've finally understood what's going on. It turns out that the important bit is the previous paragraph of section 23.3.1 (at the bottom of p708). The point is that `my_module.a.b.c` is only allowed if `my_module` is a "top-level module".

A top-level module is a module that isn't instantiated by any other module. In particular, this will guarantee that there's at most one instance of the thing that the hierarchical reference points to.

I'll fix up the code accordingly, and include your examples from above in the tests.

rswarbrick:
Ok, after a day's staring at the code (and adding test cases), I've realised there's another problem with it (other than the one that @zachjs found above). When we bind into a module, we implement this by taking a copy of the AST it came from, shoving in the bound-in cell and then calling `process_and_replace_module`. This isn't quite right! In particular, it's not going to handle parameterised targets properly, because we're going to forget our parameterisation.

I think I've got a plan for how to fix it, but it's going to take a bit more time. (Hopefully, measured in days rather than months this time!)

PR #2753, opened 2021-05-13
Factor out AST_CELL handling in genrtlil.cc into a helper function

rswarbrick:
This is a staging commit for a following commit that adds bind
support. Since the bind construct also instantiates cells, we end up
needing exactly the same logic (but we'll be filling in a Cell object
living somewhere slightly different).

(Note: This patch is part of the larger queue at https://github.com/YosysHQ/yosys/pull/2752. The idea is to split things out to avoid having a giant PR to review all at once)

### DIFF
--- a/frontends/ast/genrtlil.cc
+++ b/frontends/ast/genrtlil.cc
@@ -731,6 +731,108 @@ struct AST_INTERNAL::ProcessGenerator
 	}
 };
 
+// helper function for filling out an RTLIL::Cell with the contents of a CELL AST node
+static void fill_cell(RTLIL::Cell *cell, const AstNode &ast_node)
+{
+	int port_counter = 0, para_counter = 0;
+
+	set_src_attr(cell, &ast_node);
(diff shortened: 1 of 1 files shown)


zachjs:
The main PR doesn't invoke `fill_cell()` in any new places, so it's not clear to me why this needs to be factored out. Am I missing something here?

rswarbrick:
Hah! Good point! When I wrote this commit, I hadn't understood `AstModule::reprocess_module` / `process_module` so I thought I had to run `fill_cell` by hand.

I think that pulling out the guts like this probably makes the code easier to read (although the commit message probably needs to change!). What do you think? If not, happy to abandon this one and rebase the main PR without it.

zachjs:
We don't have a clear precedent for factoring out an entire case of genRTLIL into a helper. I also have a secondary motive: this change would create merge conflicts with an existing PR of mine :) I will probably look into splitting out the genRTLIL changes over there into a separate PR.

rswarbrick:
Ok, that's cool :-) I'll drop this PR and remove the patch from the main queue.

PR #2754, opened 2021-05-13
Change the type of current_module to Module

rswarbrick:
The current_module global is needed so that genRTLIL has somewhere to
put cells and wires that it generates as it makes sense of expressions
that it sees. However, that doesn't actually need to be an AstModule:
the Module base class is enough.

This patch should cause no functional change, but the point is that
it's now possible to call genRTLIL with a module that isn't an
AstModule as "current_module". This will be needed for 'bind' support.


zachjs (approved):
Obviously the continued use of the  `current_module` global is not ideal, but the direction you're headed here doesn't make the situation worse at all. I think this is reasonable.

rswarbrick:
Yep, I agree. I guess any further tidy-up will still have some notion of the current module (even if it's no longer a global), so I think this change would be needed in some form even with such a tidy-up.

PR #2755, opened 2021-05-13
Simplify some RTLIL destructors

rswarbrick:
No change in behaviour, but use range-based for loops instead of
iterators.

### DIFF
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -551,8 +551,8 @@ RTLIL::Design::Design()
 
 RTLIL::Design::~Design()
 {
-	for (auto it = modules_.begin(); it != modules_.end(); ++it)
-		delete it->second;
+	for (auto &pr : modules_)
+		delete pr.second;
 	for (auto n : verilog_packages)
 		delete n;
 	for (auto n : verilog_globals)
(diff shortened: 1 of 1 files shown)


kernel/rtlil.cc
```diff
@@ -551,12 +551,9 @@ RTLIL::Design::Design()
...
 {
-	for (auto it = modules_.begin(); it != modules_.end(); ++it)
-		delete it->second;
-	for (auto n : verilog_packages)
-		delete n;
-	for (auto n : verilog_globals)
-		delete n;
+	for (auto &pr : modules_) delete pr.second;
+	for (auto n : verilog_packages) delete n;
+	for (auto n : verilog_globals) delete n;
```
zachjs, line 556: This is outside my main area, but I'm guessing the general preference would be to preserve the existing style of putting the body on its own line. I agree the range syntax (where possible) is superior, but the spacing is a matter of opinion, and the least chaotic move there would be to stick to the existing precedent.

rswarbrick: Hmm, ok. I've put the newlines back in. As you say, this simplifies the change (although the renaming means we still get diffs for all the lines, sadly).

zachjs (approved):
I think this is an improvement and relatively low noise. I'm going to go ahead and merge this.

PR #2757, opened 2021-05-13
Move interface expansion in hierarchy.cc into a helper class

rswarbrick:
There should be no functional change, but this splits up the control
flow across functions, using class fields to hold the state that's
being tracked. The result should be a bit easier to read.

This is part of work to add bind support, but I'm doing some
refactoring in the hierarchy pass to make the code a bit easier to
work with. The idea is that (eventually) the IFExpander object will
hold all the logic for expanding interfaces, and then other code can
do bind insertion.

### DIFF
--- a/passes/hierarchy/hierarchy.cc
+++ b/passes/hierarchy/hierarchy.cc
@@ -156,6 +156,168 @@ std::string basic_cell_type(const std::string celltype, int pos[3] = nullptr) {
 	return basicType;
 }
 
+// A helper struct for expanding a module's interface connections in expand_module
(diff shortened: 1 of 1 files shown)


rswarbrick:
Ping! Any chance someone could take a look at this code?

zachjs (approved):
I haven't compared line by line, but this indeed appears equivalent to the existing implementation. While it still leaves the handling of interfaces in rough state, I would argue it's an improvement. I will merge this tomorrow unless there are objections from other maintainers.

PR #2758, opened 2021-05-15
intel_alm: IO and clock buffer insertion; carry fixes

gatecat:


Ravenslofty (approved):
I'll ping @mwkmwkmwk but I'm happy with this.

techlibs/intel_alm/common/megafunction_bb.v
```diff
@@ -627,3 +627,109 @@ output [port_b_data_width-1:0] portbdataout;
 input clk0, portawe, portbre;
 
 endmodule
+
+(* blackbox *)
```
mwkmwkmwk, line 631: this (and subsequent cells) needs `(* iopad_external_pin *)` annotations

techlibs/intel_alm/common/megafunction_bb.v
```diff
@@ -627,3 +627,109 @@ output [port_b_data_width-1:0] portbdataout;
...
+parameter differential_mode = "false";
+parameter bus_hold = "false";
+parameter simulate_z_as = "Z";
+parameter lpm_type = "cyclonev_io_ibuf";
+
+input i;
+input ibar;
+input dynamicterminationcontrol;
+output o;
+
```
mwkmwkmwk, line 643: why is this a blackbox? it seems the sim model here should be just `assign o = i`?

techlibs/intel_alm/common/misc_sim.v
```diff
@@ -0,0 +1,21 @@
+module MISTRAL_IB(input PAD, output O);
```
mwkmwkmwk, line 1: likewise, `iopad_external_pin`

techlibs/intel_alm/common/misc_sim.v
```diff
@@ -0,0 +1,21 @@
...
+endmodule
+
+module MISTRAL_IO(inout PAD, input I, input OE, output O);
+	assign PAD = OE ? I : 1'bz;
+	assign O = PAD;
+endmodule
+
+// Eventually, we should support clock enables and model them here too.
+// For now, CLKENA is used as a basic entry point to global routing.
+module MISTRAL_CLKENA (
```
mwkmwkmwk, line 16: I'd say if we're planning to add clock enable pin later, this cell should have a name that does not involve "ENA"

techlibs/intel_alm/common/quartus_rename.v
```diff
@@ -233,3 +239,57 @@ parameter B_SIGNED = 1;
...
+    .bus_hold("false"),
+    .differential_mode("false")
+) _TECHMAP_REPLACE_ (
+    .i(I),
+    .o(PAD),
+    .oe(OE)
+);
+endmodule
+
+module MISTRAL_OB(output PAD, input I, OE, output O);
```
mwkmwkmwk, line 264: should be `MISTRAL_IO`

techlibs/intel_alm/synth_intel_alm.cc
```diff
@@ -231,6 +247,8 @@ struct SynthIntelALMPass : public ScriptPass {
 				}
 			}
 			run("alumacc");
+			if (!noiopad)
```
mwkmwkmwk, line 250: wrong invocation, `-tinoutpad` needs a cell name parameter

mwkmwkmwk (changes requested):

techlibs/intel_alm/common/megafunction_bb.v
gatecat: thanks, fixed

techlibs/intel_alm/common/megafunction_bb.v
gatecat: I realised there were existing sim models already for cyclonev, I've updated with them with the necessary parameters. For cyclone10gx there are no sim models at all yet, so I've left that as is for now.

techlibs/intel_alm/common/misc_sim.v
gatecat: Hmm, is adding a pin to a cell in the future a problem? cc @Ravenslofty who suggested this name

Ravenslofty: Yeah, the Quartus primitive is `family_clkena` and I didn't really see the point in deviating from that in this context.

techlibs/intel_alm/common/megafunction_bb.v
```diff
@@ -627,3 +627,73 @@ output [port_b_data_width-1:0] portbdataout;
...
+module cyclone10gx_io_obuf(i, oe, dynamicterminationcontrol, seriesterminationcontrol, parallelterminationcontrol, devoe, o, obar);
+
+parameter open_drain_output = "false";
+parameter bus_hold = "false";
+parameter shift_series_termination_control = "false";
+parameter sim_dynamic_termination_control_is_connected = "false";
+parameter lpm_type = "cyclone10gx_io_obuf";
+
+(* iopad_external_pin *) input i;
+(* iopad_external_pin *) input oe;
```
mwkmwkmwk, line 656: `oe` is not an iopad

techlibs/intel_alm/common/misc_sim.v
mwkmwkmwk: Adding an extra pin will have the usual issue of unconnected input (note that `clkbufmap` has no way of tying an extra input, and it bypasses `hierarchy` so the default value won't be applied either), making a different cell with no `ENA` would solve this problem cleanly

mwkmwkmwk (changes requested):

techlibs/intel_alm/common/misc_sim.v
gatecat: I mean dealing with the pin in nextpnr (which in practice means nothing, as the hardware defaults to enabled and is essentially what happens already) doesn't seem to present too big a problem in this case. But if another primitive is strongly preferred then I can add `MISTRAL_CLKBUF` support to nextpnr, too.

mwkmwkmwk: The preferred approach would be to have a full-featured `MISTRAL_CLKENA` cell with clock enable input, and a transient `__MISTRAL_CLKBUF` cell for `clkbufmap` that is immediately techmapped into `MISTRAL_CLKENA` — see efinix as an example of that approach

gatecat: would it be OK to do that; but not include the enable input on the `MISTRAL_CLKENA` cell yet? I don't think we fully understand, or want to implement/model, the full enable semantics at this point as they appear to be a bit icky

mwkmwkmwk: *shrug* feels icky, but if you really want to do this, fine; however it seems fairly obvious that there *is* an ENA input and it could be just as well added into the cell immediately, with a blackbox model?

gatecat: the downside of that way is that then no out-of-the-box post-synthesis simulation of a design with a clock buffer inserted is possible, even if it isn't using `ENA`

mwkmwkmwk: yes, and that's unfortunately a general problem for buffers with clock enable even when we know the exact behavior (they cannot really be modelled via synthesizable logic in a way that simulates properly), this is also why we don't (and can't) have a model for Xilinx `BUFGCE`

if you want to be safe from this problem in the future when we actually add an enable input, a separate `MISTRAL_CLKBUF` cell is the way to go

gatecat: anyway, I've had a go at implementing it this way and force-pushed.

gatecat: oops, our messages crossed, if a simulation model is going to be a pain in a future then I'll go back to the `MISTRAL_CLKBUF` idea instead

gatecat: done - also on the nextpnr side: https://github.com/YosysHQ/nextpnr/pull/707/commits/9d7f90dd89d98457e00579615ef9251d0f69f3a4

mwkmwkmwk (approved):
alright let's ship it

PR #2763, opened 2021-05-20
memory_share: Use Mem helpers.

mwkmwkmwk:


mwkmwkmwk:
*sigh* do not merge; while this doesn't introduce any *new* problems, I'd rather hold this until #2766 is resolved.

mwkmwkmwk:
No longer on hold — the bug still exists, but the offending code is now in `opt_mem_feedback` pass, and will be converted in a separate PR.

PR #2768, opened 2021-05-22
kernel/rtlil: Extract some helpers for checking memory cell types.

mwkmwkmwk:
There will soon be more (versioned) memory cells, so handle passes that
only care if a cell is memory-related by a simple helper call instead of
a hardcoded list.


eddiehung (approved):
LGTM

PR #2801, opened 2021-05-27
sv: support tasks and functions within packages

### LINKED ISSUE #2748: Parse error with function defined inside package
When reading a package file I get a parse error, the file contains:

---
package img_gfx_std_func_pack_;

   localparam IMG_CLK_GATE_FANOUT = 32;

   function automatic integer img_log2(integer d);
      integer tmp, COUNT;
      tmp  = 1;
      for (integer COUNT = 0; COUNT < 31; COUNT++) begin // N.B. Synopsys requires static loops.
         if (tmp >= d)
           return COUNT;
         tmp = tmp*2;
      end
      // Should never get here, but add return statement to avoid spyglass warning
      return d;
   endfunction : img_log2
endpackage
--

This produces a parse error but the code is valid:

3. Executing Verilog-2005 frontend: img_gfx_std_func_pack_.sv
Parsing SystemVerilog input from img_gfx_std_func_pack_.sv' to AST representation.
img_gfx_std_func_pack_.sv:43: ERROR: syntax error, unexpected TOK_FUNCTION

zachjs:
Fixes #2748

### DIFF
--- a/frontends/ast/ast.cc
+++ b/frontends/ast/ast.cc
@@ -1196,6 +1196,25 @@ static void process_module(RTLIL::Design *design, AstNode *ast, bool defer, AstN
 	design->add(current_module);
 }
 
+// renames identifiers in tasks and functions within a package
+static void rename_in_package_stmts(AstNode *pkg)
+{
+	std::unordered_set<std::string> idents;
+	for (AstNode *item : pkg->children)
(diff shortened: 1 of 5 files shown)


zachjs:
Given this only has any effect on inputs which were previously un-parseable and its relative simplicity, I am merging.

PR #2807, opened 2021-05-31
Add new memory cells.

mwkmwkmwk:
DO NOT MERGE until the following issues are fixed:

- [x] cxxrtl doesn't yet support the new features
- [ ] needs tests for init/reset value handling in various passes, and *really* needs tests for wide port handling
- [x] needs CHANGELOG entries
- [x] the manual needs updating

Also, we still don't have a new transparency mask model.  To avoid creating _v3 versions of the cells, it would be good to deal with this as well before this PR is merged.

There is also (probably) going to be a `$meminit_v2`, but that's independent from this PR.


mwkmwkmwk:
New version, now includes transparency mask, and thus depends on #2904

mwkmwkmwk:
CHANGELOG and manual now done; tests will land together with new transparency recognition in a future PR

mwkmwkmwk:
Shortening this PR to just the new cells + `proc_memwr` changes; the other two commits will be reordered and land in different PRs so they can be tested properly.

mwkmwkmwk:
(the tests became part of #2930, #2928, and #2925 in the end)

PR #2814, opened 2021-06-05
verilog: check for module scope identifiers during width detection

zachjs:
The recent fix for case expression width detection causes the width of
the expressions to be queried before they are simplified. Because the
logic supporting module scope identifiers only existed in simplify,
looking them up would fail during width detection. This moves the logic
to a common helper used in both simplify() and detectSignWidthWorker().


clairexen (approved):
LGTM

PR #2829, opened 2021-06-14
Makefile cleanups

Xiretza:
Another round of fiddling with the Makefile and trying to improve whatever I come across. This is inspired by #2011 and fixes some of the issues listed there, however `$(LD)` still points to the main compiler frontend and `$(LDFLAGS)` contains escaped flags because that seems to be The Done Thing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -354,53 +354,29 @@ TARGETS += libyosys.so
 endif
 
 ifeq ($(ENABLE_PYOSYS),1)
-
-#Detect name of boost_python library. Some distros usbe boost_python-py<version>, other boost_python<version>, some only use the major version number, some a concatenation of major and minor version numbers
-ifeq ($(OS), Darwin)
+# Detect name of boost_python library. Some distros use boost_python-py<version>, other boost_python<version>, some only use the major version number, some a concatenation of major and minor version numbers
+CHECK_BOOST_PYTHON = (echo "int main(int argc, char ** argv) {return 0;}" | $(CXX) -xc -o /dev/null $(shell $(PYTHON_CONFIG) --ldflags) -l$(1) - > /dev/null 2>&1 && echo "-l$(1)")
 BOOST_PYTHON_LIB ?= $(shell \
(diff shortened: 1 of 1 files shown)


nakengelhardt:
Hi, we discussed this at the dev JF, and here are some initial comments:
- The indentation changes are a question of coding style. We don't want to overhaul the coding style of our project in an aside in the middle of another PR, so please remove that commit.
- The `-std=blah` flags have to be per-compiler because they sometimes need to be different. Currently they happen to be mostly identical but they have diverged in the past and surely will again.
- For ABC, we need to compile with `CC="$(CXX)"` because some files in that project don’t have "extern C" guards, so if we don't explicitly say to use the C++ compiler, they are at risk of getting compiled as C and then the linker symbols will not match.

There were also concerns regarding testing this. Over time we've accumulated small compatibility fixes for a variety of platforms/versions, some more rarely used but still important, but we don't ourselves have test environments set up for all of them so we are somewhat cautious about changes. Can you motivate what problems you are trying to solve so that we can better understand the tradeoffs?

Xiretza:
This was mostly motivated by yet another person tripping over the `$(CC)`/`$(CXX)` mess. Right now I don't really feel motivated to  delve into this again, but I rebased and removed the formatting changes, so it should now be possible to easily pick pick out any commits that you find useful.

> * The indentation changes are a question of coding style. We don't want to overhaul the coding style of our project in an aside in the middle of another PR, so please remove that commit.
> * The `-std=blah` flags have to be per-compiler because they sometimes need to be different. Currently they happen to be mostly identical but they have diverged in the past and surely will again.

Removed.

> * For ABC, we need to compile with `CC="$(CXX)"` because some files in that project don’t have "extern C" guards, so if we don't explicitly say to use the C++ compiler, they are at risk of getting compiled as C and then the linker symbols will not match.

As far as I know, yosys already maintains its own ABC fork, couldn't that be fixed there?

> Can you motivate what problems you are trying to solve so that we can better understand the tradeoffs?

I started out with separating `CC`/`CXX`, then fixed whatever else jumped out at me.

1. `PYOSYS` configuration used a lot of duplicate code, which just didn't look very maintenance friendly.
2. The `DT_SONAME` problem causes reproducible build problems because it embeds build-time paths into the binary (no idea if yosys is otherwise reproducible, but this is a low-hanging fruit).
3. ABC didn't give any useful log output with `VERBOSE=0`, so I passed that down.
4. The `$(SANITIZER)` patch was separated out from #2683, where I wanted to use ASAN and UBSAN at the same time.

mmicko:
We have cherry-picked commits that are not addressing CC/CXX/LD and changes for nested conditionals since that can affect some of older make versions.

PR #2836, opened 2021-06-16
pyosys: Clear SIGINT handler after Python loads

gatecat:
This is equivalent to https://github.com/YosysHQ/nextpnr/commit/b1cbae1293fecf3ba14a7ce073d26dcfb07177f0 from way-back-when.

Python sets up its own SIGINT handler which stops Ctrl+C from working in pyosys builds, this clears that after Python has been initialised.

### DIFF
--- a/kernel/yosys.cc
+++ b/kernel/yosys.cc
@@ -71,6 +71,7 @@
 
 #include <limits.h>
 #include <errno.h>
+#include <signal.h>
 
 YOSYS_NAMESPACE_BEGIN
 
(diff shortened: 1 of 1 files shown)


smunaut:
FWIW, I confirmed here it makes CTRL-C work again.

Xiretza:
Awesome, I had already accepted this behaviour as a fact of life quite a while ago ;)

nakengelhardt:
cc @btut

btut:
LGTM, thanks!


> Awesome, I had already accepted this behaviour as a fact of life quite a while ago ;)

Me too :)

PR #2837, opened 2021-06-16
Include blif reader header in public facing extension header files.

yashton:
The LSOracle plugin uses the function "parse_blif" in frontends/blif/blifparse.h to parse the temporary output file created by the external tool. This function is not available in the share/includes directory, which would be the only headers available to compile against in an installed copy. This PR adds the required head to the share/includes directory.


mwkmwkmwk (changes requested):
This is not the way to do it — share/ is an autogenerated directory and should not exist in the source tree.

Add this include file to the list in the Makefile instead (it's around line 600).

yashton:
@mwkmwkmwk I missed that in the Makefile, I've fixed it.

mwkmwkmwk (approved):

PR #2838, opened 2021-06-17
Extract missing module support in hierarchy.cc to a helper function

rswarbrick:
*(Text edited after a force-push)*

This PR has two commits. The first is a small clean-up in `hierarchy.cc`. The second adds a test to check that this functionality interacts properly with interface expansion. The message for the first commit is:

> I think the code is now a bit easier to follow (and has lost some
> levels of indentation!).
> 
> The only non-trivial change is that I removed the check for
> `cell->type[0] != '$'` when deciding whether to complain if we couldn't
> find a module. This will always be true because of the early exit
> earlier in the function.

This is part of the commit queue in #2752, but has been split out to make reviewing easier.

### DIFF
--- a/passes/hierarchy/hierarchy.cc
+++ b/passes/hierarchy/hierarchy.cc
@@ -318,6 +318,64 @@ struct IFExpander
 	}
 };
 
+// Get a module needed by a cell, either by deriving an abstract module or by
+// loading one from a directory in libdirs.
+//
+// If the module can't be found and check is true then exit with an error
+// message. Otherwise, return a pointer to the module if we derived or loaded
+// something. or null otherwise (the module should be blackbox or we couldn't
+// find it and check is not set).
+RTLIL::Module *get_module(RTLIL::Design                  &design,
+                          RTLIL::Cell                    &cell,
+                          RTLIL::Module                  &parent,
+                          bool                            check,
+                          const std::vector<std::string> &libdirs)
+{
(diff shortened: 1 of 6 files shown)


zachjs:
> There's a small functional change: if we load or derive a module
> successfully, we'll now keep going in the cell loop, rather than
> setting the did_something flag and skipping. I think this might
> actually fix a bug where the interface code wouldn't run on a module
> that got loaded on demand.

The coverage for interface support is incredibly sparse. If it's not too much work, it would be really nice to have a test case to cover the relevant interactions here.

rswarbrick:
I agree, although I don't fully understand how this works either :-/ I can spend some time on it, but won't be able to until probably the middle of July.

rswarbrick:
Ahah! It turns out that the original code worked for this case too. Because it set the `did_something` flag, we came back and went around the loop again later. I've just reverted that part of the change (since it's neutral, maybe it's better to keep the original). However (now I've written it...) I've pushed a second patch to the PR that adds a test in the `svinterfaces` suite to make sure nothing comes unstuck in the future.

Incidentally, I built things with coverage earlier and the (small number of) svinterfaces tests do actually achieve pretty high line coverage in `hierarchy.cc`. Obviously, this isn't necessarily the best metric, but the testing maybe isn't quite as sparse as it seemed at first.

zachjs (approved):
LGTM, thanks for the changes!

PR #2844, opened 2021-06-22
Docker file & github docker hub

mrazekv:
This script adds an action to GitHub that automatically creates a pre-build docker file for Yosys

Then, it is easy to start the tool without a very long build.

```bash
docker pull ghcr.io/mrazekv/yosys:master # just for update of existing 
docker run -v $(pwd):/app -it --rm ghcr.io/mrazekv/yosys:master
```

The action takes approximately one hour. It would be also possible to upload this image to the Docker hub.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -33,6 +33,13 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
+        
+      - name: Cache sources
+        id: cache-sources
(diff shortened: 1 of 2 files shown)


widlarizer:
We're not interested in maintaining more distributions of Yosys beyond OSS CAD suite and TabbyCAD. There's open third-party Docker-based bundles as well

mrazekv:
This pull request is more than 5 years old. It is great that now it is much easier to deploy the package.

PR #2848, opened 2021-06-24
ECP5 techlibs changes to support bitfile simulation

nneonneo:
For a Pwn2Win 2021 CTF challenge, we developed some code which converts ECP5 bitfiles into Verilog. Most of that will be submitted to https://github.com/YosysHQ/prjtrellis; this PR is for the techlib bits which that code depends on for simulation.

This PR adds an apparently missing pair of connections in the CCU2 slice mode, which were necessary to get the simulated design to behave correctly.

### DIFF
--- a/techlibs/ecp5/cells_sim.v
+++ b/techlibs/ecp5/cells_sim.v
@@ -240,7 +240,7 @@ module DPR16X4C (
 		input [3:0] WAD,
 		output [3:0] DO
 );
-	// For legacy Lattice compatibility, INITIVAL is a hex
+	// For legacy Lattice compatibility, INITVAL is a hex
 	// string rather than a numeric parameter
 	parameter INITVAL = "0x0000000000000000";
 
@@ -530,6 +530,9 @@ module TRELLIS_SLICE(
 				.S0(F0), .S1(F1),
 				.COUT(FCO)
 			);
+			// TODO: confirm these connections
+			assign OFX0 = F0;
+			assign OFX1 = F1;
 		end else if (MODE == "RAMW") begin
 			assign WDO0 = C1m;
 			assign WDO1 = A1m;


mwkmwkmwk:
This doesn't sound like a good idea at all (except the missing connections fix, perhaps).

There is no such thing as a "default" value in hardware — every configuration bit is either a 0 or 1 and that determines the parameter values.  Sure, an unconfigured device may reset to one of the states, but there's no real reason to consider it a default when a bitfile is always meant to load every configuration bit.  And there is definitely no reason to reflect this in the cell library default parameters — if anything, they should be picked for user's least surprise (though relying on that is usually foolish anyway).  You have the bitfile, please output Verilog that has all parameters set according to the configuration bits, and don't rely on the defaults in the cell library — this will also help any human actually reading it.

And as for the PDPW16KD change: no.  This is a sloppily written CTF-grade model that doesn't even bother with considering the large amount of parameters supported by the actual hw cell, hardcodes a very specific configuration, and contains debug-prints.  It also doesn't in any way integrate with the actual synthesis flow.  It's not anywhere near mergeable shape.

nneonneo:
Good feedback. I had made the RAM/EBR a separate commit since I wasn't sure what the best approach to that would be. I can just drop that commit. (I may drop the corresponding EBR support in the verilog generator, because for similar reasons it only works for a specific configuration, and would not be sufficiently general to be useful).

With respect to the defaults, I was just going with Project Trellis' notion of "default", which is whatever the all-zeros bit pattern produces. This is as close to a "default" as you can get: unused components (including enums unused by a given mode, or slices unused by a given design) by default do get the all-zeros enum pattern, and the bitstream compression works best with all-zeros input. I am not sure it's a _downside_ if the defaults of `cells_sim` match the "default" zero-bit configuration.

gatecat:
The defaults in the cell library are the defaults for when the primitive is instantiated, which can often be different from the default in the bitstream when the primitive isn't instantiated at all.

nneonneo:
Thanks @mwkmwkmwk and @gatecat for correcting my understanding of the techlibs defaults. I will push the "defaults" to the verilog translator.

I dropped the RAM block; hopefully something like that can be implemented in the future with all of the modes, but I anticipate more reverse engineering would have to be done to make sure it behaves correctly.

gatecat:
> I dropped the RAM block; hopefully something like that can be implemented in the future with all of the modes, but I anticipate more reverse engineering would have to be done to make sure it behaves correctly.

Yea, unfortunately that's what it hasn't been done already

techlibs/ecp5/cells_sim.v
```diff
@@ -530,6 +530,9 @@ module TRELLIS_SLICE(
 				.S0(F0), .S1(F1),
 				.COUT(FCO)
 			);
+			// TODO: confirm these connections
+			assign OFX0 = F0;
```
gatecat, line 534: I know why you have added these, but I think the routing that appears to be from OFX0/1 in carry mode is simply because those PIPs share bits with the carry inject config bits. I think a better solution is going to be to drop the OFX to F pip (and use the F output always) when CCU2 mode is set for a slice.

nneonneo:
Thanks, @gatecat. Based on your feedback, I'll be dropping this change and instead dealing with the OFX->F0 connection in the verilog generator. That PR is now open: https://github.com/YosysHQ/prjtrellis/pull/168.

PR #2855, opened 2021-07-09
Add support for the Bitwuzla solver and the timeout option

GCHQDeveloper560:
The boolector team appears to be developing a new solver, [bitwuzla](https://bitwuzla.github.io/). I have a case where it is 3X faster than boolector, though another where it's 2X slower, so it is perhaps not always quicker. It seems to be where most of their development effort is going and was the tool they entered in the 2021 SMT competition, so I would expect it might continue to improve.

The interface unsurprisingly seems very similar to boolector, so I've added it to the same code path.

Both solvers support a timeout with the ``--time`` or ``-t`` options, so I've added support for that as well.

### DIFF
--- a/backends/smt2/smtio.py
+++ b/backends/smt2/smtio.py
@@ -203,14 +203,14 @@ class SmtIo:
                 print('timeout option is not supported for mathsat.')
                 sys.exit(1)
 
-        if self.solver == "boolector":
+        if self.solver in ["boolector", "bitwuzla"]:
             if self.noincr:
-                self.popen_vargs = ['boolector', '--smt2'] + self.solver_opts
+                self.popen_vargs = [self.solver, '--smt2'] + self.solver_opts
             else:
-                self.popen_vargs = ['boolector', '--smt2', '-i'] + self.solver_opts
+                self.popen_vargs = [self.solver, '--smt2', '-i'] + self.solver_opts
             self.unroll = True
             if self.timeout != 0:
-                print('timeout option is not supported for boolector.')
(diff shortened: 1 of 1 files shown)


backends/smt2/smtio.py
```diff
@@ -203,15 +203,14 @@ def setup(self):
...
-                self.popen_vargs = ['boolector', '--smt2'] + self.solver_opts
+                self.popen_vargs = [self.solver, '--smt2'] + self.solver_opts
             else:
-                self.popen_vargs = ['boolector', '--smt2', '-i'] + self.solver_opts
+                self.popen_vargs = [self.solver, '--smt2', '-i'] + self.solver_opts
             self.unroll = True
             if self.timeout != 0:
-                print('timeout option is not supported for boolector.')
-                sys.exit(1)
+                self.popen_vargs.append('--time=%d' % self.timeout);
```
mwkmwkmwk, line 213: this doesn't quite work properly — while the timeout will cause boolector/bitwuzla to terminate, it'll terminate with the following output:

```
[btor>main] ALARM TRIGGERED: time limit 1 seconds reached                                                                                                                                              
unknown                                                                                                                                                                                                
```

this, unfortunately, won't be properly picked up as "timeout" status by the wrapper code below

GCHQDeveloper560:
Thank you for the catch! I clearly hadn't tested the timeout portion. It looks like the wrapper code is expecting "timeout" rather than the more elaborate message being output by these solvers. I've redone the commit to just add Bitwuzla and remove the timeout change.

mwkmwkmwk (approved):

PR #2856, opened 2021-07-10
cxxrtl: Convert to Mem helpers.

mwkmwkmwk:
This *only* does conversion, but doesn't add any new functionality —
support for memory read port init/reset is still upcoming.

### DIFF
--- a/backends/cxxrtl/cxxrtl_backend.cc
+++ b/backends/cxxrtl/cxxrtl_backend.cc
@@ -1600,11 +1600,51 @@ struct CxxrtlWorker {
 			f << " = value<" << mem->width << "> {};\n";
 		dec_indent();
 		f << indent << "}\n";
-		if (has_enable) {
+		if (has_enable && !port.ce_over_srst) {
+			dec_indent();
+			f << indent << "}\n";
+		}
+		if (port.srst != State::S0) {
+			// Synchronous reset
+			std::vector<const RTLIL::Cell*> inlined_cells_srst;
+			collect_sigspec_rhs(port.srst, for_debug, inlined_cells_srst);
+			if (!inlined_cells_srst.empty())
+				dump_inlined_cells(inlined_cells_srst);
+			f << indent << "if (";
+			dump_sigspec_rhs(port.srst);
+			f << " == value<1> {1u}) {\n";
+			inc_indent();
+				f << indent;
+				dump_sigspec_lhs(port.data);
+				f << " = ";
+				dump_const(port.srst_value);
+				f << ";\n";
+			dec_indent();
+			f << indent << "}\n";
+		}
+		if (has_enable && port.ce_over_srst) {
 			dec_indent();
 			f << indent << "}\n";
 		}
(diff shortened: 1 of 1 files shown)


mwkmwkmwk:
TODOs (may or may not end up as separate PRs):

- add initial value support for read ports
- add reset support for read ports
- handle memory writes in processes

Other upcoming changes:

- support for meminit_v2 (memory initialization with mask)
- support for per-port transparency masks (blocked on model changes)

whitequark:
Really like this approach!

mwkmwkmwk:
I think it's ready for review now; some notes:

- every read port is a distinct node (to avoid creating false dependencies between one port's address and another's data), but all write ports of a single memory are a single node (I wrote it this way to make it easier to handle priority, without noticing that `update` is already fully capable of dealing with this; we could change it if necessary)
- port/memory attributes aren't printed right now, which is a minor regression; it may be a good idea to convert `kernel/mem.h` classes to inherit from `AttrObject` so that we can just `dump_attrs` them

mwkmwkmwk:
Rebased with minor fix: processes with a memory write are now always inserted into `live_nodes`

whitequark:
> * but all write ports of a single memory are a single node (I wrote it this way to make it easier to handle priority, without noticing that `update` is already fully capable of dealing with this; we could change it if necessary)

This is potentially _a lot better_ than the approach with `update` because sorting the queue vector takes a non-trivial amount of time in benchmarking. Since I think the concept of priority only really makes sense within a netlist (and not between co-simulating entities), we could ditch that entirely and have `Memory::update` do last-write-wins.

whitequark:
> * port/memory attributes aren't printed right now, which is a minor regression; it may be a good idea to convert `kernel/mem.h` classes to inherit from `AttrObject` so that we can just `dump_attrs` them



> * port/memory attributes aren't printed right now, which is a minor regression; it may be a good idea to convert `kernel/mem.h` classes to inherit from `AttrObject` so that we can just `dump_attrs` them

Would that be hard to do? It seems like one of those small things that otherwise easily fall through the cracks. (Not something I'd hold up the PR on, of course.)

mwkmwkmwk:
> > * port/memory attributes aren't printed right now, which is a minor regression; it may be a good idea to convert `kernel/mem.h` classes to inherit from `AttrObject` so that we can just `dump_attrs` them
> 
> Would that be hard to do? It seems like one of those small things that otherwise easily fall through the cracks. (Not something I'd hold up the PR on, of course.)

Don't think so; I'll have a shot at it tomorrow, probably.

mwkmwkmwk:
Done, turns out it was pretty trivial.

whitequark (approved):
LGTM

mwkmwkmwk:
Fixing one more bug noticed in own review: we need to take into account memory writes from processes as well when deciding whether to emit `commit` (ie. `writable_memories` is back)

mwkmwkmwk:
The macOS builder seems stuck, trying a dummy push-force to kick it...

PR #2858, opened 2021-07-11
rtlil: Make Process handling more uniform with Cell and Wire.

mwkmwkmwk:
- add a backlink to module from Process
- make constructor and destructor protected, expose Module functions
  to add and remove processes


whitequark (approved):
Yes please this is sooo much better.

PR #2863, opened 2021-07-12
kernel/mem: Add a coalesce_inits helper.

mwkmwkmwk:
While this helper is already useful to squash sequential initializations
into one in cxxrtl, its main purpose is to squash overlapping masked memory
initializations (when they land) and avoid having to deal with them in
cxxrtl runtime.

### DIFF
--- a/backends/cxxrtl/cxxrtl_backend.cc
+++ b/backends/cxxrtl/cxxrtl_backend.cc
@@ -1835,6 +1835,8 @@ struct CxxrtlWorker {
 			f << ",\n";
 			inc_indent();
 				for (auto &init : mem->inits) {
+					if (init.removed)
+						continue;
 					dump_attrs(&init);
 					int words = GetSize(init.data) / mem->width;
 					f << indent << "memory<" << mem->width << ">::init<" << words << "> { "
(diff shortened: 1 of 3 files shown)


backends/cxxrtl/cxxrtl_backend.cc
```diff
@@ -1835,6 +1835,8 @@ struct CxxrtlWorker {
 			f << ",\n";
 			inc_indent();
 				for (auto &init : mem->inits) {
+					if (init.removed)
+						continue;
```
whitequark, line 1839: If the inits are actually overlapping here, this is flat out unsound, right? Can you add an assert to make sure this doesn't happen?

mwkmwkmwk: Uh, why?

The way this works is:

- after creating the helper, there are no `removed` inits
- `coalesce_inits` is called immediately afterwards, non-overlapping inits are left alone, overlapping inits are marked as removed and new, coalesced, inits are created in their place
- the `dump-memory` function skips over `removed` inits to avoid printing the inits that have now been effectively replaced by the coalesced ones

whitequark: Ah, makes sense.

whitequark (approved):

PR #2868, opened 2021-07-15
verilog: save and restore overwritten macro arguments

### LINKED ISSUE #2867: Unexpected macro behavior

`Pad.sv` contains two almost identical modules, the difference is that the vector indices use a `MIN` macro in one case, and the macro is manually expanded in the other case. Loading the Verilog produces warnings in the `MIN` case.
[testcase.zip](https://github.com/YosysHQ/yosys/files/6819463/testcase.zip)

The problem is reproducible with:

`$ yosys -q -p "read_verilog -sv Pad.sv"`

## Expected behavior

No warnings.

## Actual behavior

Displays:

```
Pad.sv:6: Warning: Range [63:0] select out of bounds on signal `\y': Setting 63 MSB bits to undef.
Pad.sv:6: Warning: Ignoring assignment to constant bits:
    old assignment: { 63'xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx \y } = { 63'000000000000000000000000000000000000000000000000000000000000000 \x }
    new assignment: \y = \x.
```

zachjs:
Fixes #2867

### DIFF
--- a/frontends/verilog/preproc.cc
+++ b/frontends/verilog/preproc.cc
@@ -36,6 +36,7 @@
 #include "verilog_frontend.h"
 #include "kernel/log.h"
 #include <assert.h>
+#include <stack>
 #include <stdarg.h>
 #include <stdio.h>
 #include <string.h>
@@ -334,6 +335,11 @@ define_map_t::add(const std::string &name, const std::string &txt, const arg_map
 	defines[name] = std::unique_ptr<define_body_t>(new define_body_t(txt, args));
 }
 
+void define_map_t::add(const std::string &name, const define_body_t &body)
+{
+	defines[name] = std::unique_ptr<define_body_t>(new define_body_t(body));
+}
+
 void define_map_t::merge(const define_map_t &map)
 {
 	for (const auto &pr : map.defines) {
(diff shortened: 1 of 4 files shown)


frontends/verilog/preproc.cc
```diff
@@ -334,6 +335,11 @@ define_map_t::add(const std::string &name, const std::string &txt, const arg_map
 	defines[name] = std::unique_ptr<define_body_t>(new define_body_t(txt, args));
 }
 
+void define_map_t::add(const std::string &name, const define_body_t &body)
+{
+	defines[name] = std::unique_ptr<define_body_t>(new define_body_t(body));
```
mwkmwkmwk, line 340: this looks like a job for `make_unique`

zachjs: I was thinking the same thing, but unfortunately, `make_unique` was introduced in C++14.

frontends/verilog/preproc.cc
```diff
@@ -495,6 +511,10 @@ static bool try_expand_macro(define_map_t &defines, std::string &tok)
 			args.push_back(arg);
 		}
 		for (const auto &pr : body->args.get_vals(name, args)) {
+			if (const define_body_t *existing = defines.find(pr.first)) {
+				macro_arg_stack.push({pr.first, *existing});
+				insert_input("`__restore_macro_arg");
```
mwkmwkmwk, line 516: this needs armoring with spaces, otherwise:

```
module top(...);

`define add(x) x +

input [3:0] A;
output [3:0] B;
assign B = `add(`add(3)A)A;

endmodule
```

-->

```
-- Verilog code after preprocessor --
`file_push "t.v"
module top(...);



input [3:0] A;
output [3:0] B;
assign B = 3 +`__restore_macro_argA +A;

endmodule

`file_pop
-- END OF DUMP --
```

This, of course, blows up.

mwkmwkmwk (changes requested):
(see inline comment)

frontends/verilog/preproc.cc
mwkmwkmwk: ... okay, nevermind then

frontends/verilog/preproc.cc
zachjs: Nice catch, thanks!

mwkmwkmwk (approved):

PR #2869, opened 2021-07-15
Extract connection checking logic from expand_module in hierarchy.cc

rswarbrick:
No functional change, but pulls more logic out of the expand_module
function.

This is part of the commit queue in #2752, but has been split out to make reviewing easier.

### DIFF
--- a/passes/hierarchy/hierarchy.cc
+++ b/passes/hierarchy/hierarchy.cc
@@ -1106,8 +1106,8 @@ struct HierarchyPass : public Pass {
 
 				pool<std::pair<IdString, IdString>> params_rename;
 				for (const auto &p : cell->parameters) {
-					if (p.first[0] == '$' && '0' <= p.first[1] && p.first[1] <= '9') {
-						int id = atoi(p.first.c_str()+1);
+					int id;
+					if (read_id_num(p.first, &id)) {
 						if (id <= 0 || id > GetSize(cell_mod->avail_parameters)) {
 							log("  Failed to map positional parameter %d of cell %s.%s (%s).\n",
 									id, RTLIL::id2cstr(mod->name), RTLIL::id2cstr(cell->name), RTLIL::id2cstr(cell->type));
@@ -1134,9 +1134,9 @@ struct HierarchyPass : public Pass {
 				log("Mapping positional arguments of cell %s.%s (%s).\n",
 						RTLIL::id2cstr(module->name), RTLIL::id2cstr(cell->name), RTLIL::id2cstr(cell->type));
 				dict<RTLIL::IdString, RTLIL::SigSpec> new_connections;
-				for (auto &conn : cell->connections())
-					if (conn.first[0] == '$' && '0' <= conn.first[1] && conn.first[1] <= '9') {
-						int id = atoi(conn.first.c_str()+1);
+				for (auto &conn : cell->connections()) {
+					int id;
+					if (read_id_num(conn.first, &id)) {
 						std::pair<RTLIL::Module*,int> key(design->module(cell->type), id);
 						if (pos_map.count(key) == 0) {
 							log("  Failed to map positional argument %d of cell %s.%s (%s).\n",
@@ -1146,6 +1146,7 @@ struct HierarchyPass : public Pass {
 							new_connections[pos_map.at(key)] = conn.second;
 					} else
 						new_connections[conn.first] = conn.second;
+				}
 				cell->connections_ = new_connections;
 			}
 		}
--- a/passes/hierarchy/hierarchy.cc
+++ b/passes/hierarchy/hierarchy.cc
@@ -376,6 +376,69 @@ RTLIL::Module *get_module(RTLIL::Design                  &design,
 	return nullptr;
 }
 
+// Try to read an IdString as a numbered connection name ("$123" or similar),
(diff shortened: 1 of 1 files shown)


passes/hierarchy/hierarchy.cc
```diff
@@ -376,6 +376,65 @@ RTLIL::Module *get_module(RTLIL::Design                  &design,
 	return nullptr;
 }
 
+// Try to read an IdString as a numbered connection name ("$123" or similar),
+// writing the result to dst. If the string isn't of the right format, ignore
+// dst and return false.
+bool read_id_num(RTLIL::IdString str, int *dst)
+{
+	log_assert(dst);
```
zachjs, line 384: Would you consider taking `dst` by reference here instead?

passes/hierarchy/hierarchy.cc
```diff
@@ -376,6 +376,65 @@ RTLIL::Module *get_module(RTLIL::Design                  &design,
...
+	const char *c_str = str.c_str();
+	if (c_str[0] != '$' || !('0' <= c_str[1] && c_str[1] <= '9'))
+		return false;
+
+	*dst = atoi(c_str + 1);
+	return true;
+}
+
+// Check that all the connections and parameters on the cell match those defined
+// in its type.
```
zachjs, line 395: I think I'd like this comment to clarify that it's checking the connection/parameter _names and positions_ to make sure they exist, as "match" is a little ambiguous at first glance.

passes/hierarchy/hierarchy.cc
```diff
@@ -433,29 +492,7 @@ bool expand_module(RTLIL::Design *design, RTLIL::Module *module, bool flag_check
...
-					int id = atoi(param.first.c_str()+1);
-					if (id <= 0 || id > GetSize(mod->avail_parameters))
-						log_error("Module `%s' referenced in module `%s' in cell `%s' has only %d parameters, requested parameter %d.\n",
-								log_id(cell->type), log_id(module), log_id(cell), GetSize(mod->avail_parameters), id);
-				} else if (mod->avail_parameters.count(param.first) == 0 && param.first[0] != '$' && strchr(param.first.c_str(), '.') == NULL)
-					log_error("Module `%s' referenced in module `%s' in cell `%s' does not have a parameter named '%s'.\n",
-							log_id(cell->type), log_id(module), log_id(cell), log_id(param.first));
-			}
-
-		}
```
zachjs, line 458: If you're willing, I'd argue it would be "in scope" to also update the very similar positional re-mapping logic further down to at least use your new `read_id_num` helper.

passes/hierarchy/hierarchy.cc
rswarbrick: There's a Google (I think...) style that passes "out parameters" by pointer rather than reference, to ensure that they are more obvious at the call site. The idea is that `ret = read_id_num(name, val)` doesn't obviously modify `val`, but `ret = read_id_num(name, &val)` looks more like it might.

I don't really have strong feelings about it, but I *think* this is slightly clearer. Happy to switch it around if you think that would be better.

passes/hierarchy/hierarchy.cc
rswarbrick: Good idea. I've added a patch to use `read_id_num`. I think this could probably be more drastically refactored to avoid the "validate, then use the result" pattern, but maybe that's for a future PR.

passes/hierarchy/hierarchy.cc
```diff
@@ -1093,18 +1134,20 @@ struct HierarchyPass : public Pass {
...
 						std::pair<RTLIL::Module*,int> key(design->module(cell->type), id);
 						if (pos_map.count(key) == 0) {
 							log("  Failed to map positional argument %d of cell %s.%s (%s).\n",
-									id, RTLIL::id2cstr(module->name), RTLIL::id2cstr(cell->name), RTLIL::id2cstr(cell->type));
+							    id, RTLIL::id2cstr(module->name), RTLIL::id2cstr(cell->name), RTLIL::id2cstr(cell->type));
 							new_connections[conn.first] = conn.second;
 						} else
 							new_connections[pos_map.at(key)] = conn.second;
-					} else
+					} else {
```
zachjs, line 1147: It seems like some of these changes are not related to the actual `atoi` => `read_id_num` improvement. I'd prefer we minimize the indentation/bracing changes around here to just those necessary. Otherwise, LGTM.

rswarbrick: I can drop them if you'd like, but had to add some braces above to avoid things getting mismatched (there was `for (A) if (B) { C } else D`, which I stupidly changed to `for (A) B0; if (B1) { C } else D` which... doesn't work!). Once I'd added the necessary braces, it looked weirdly wonky not to add them for the `else` side. What do you think?

zachjs: My initial thought was to just do the minimal change, despite the current un-braced `else` branch looking a little funny. You could also consider moving up the declaration of `id` to avoid changing any of the irrelevant branches.

rswarbrick: Ok, I'll drop the later braces. Let's keep `id` with the minimal scope: no need to pollute a larger namespace than necessary.

zachjs (approved):

PR #2875, opened 2021-07-16
Generate an RTLIL representation of bind constructs

rswarbrick:
This code now takes the AST nodes of type AST_BIND and generates a
representation in the RTLIL for them.

This is a little tricky, because a binding of the form:

    bind baz foo_t foo_i (.arg (1 + bar));

means "make an instance of `foo_t` called `foo_i`, instantiate it inside
`baz` and connect the port `arg` to the result of the expression `1+bar`".
Of course, `1+bar` needs a cell for the addition. Where should that cell
live?

With this patch, the `Binding` structure that represents the construct
is itself an `AST::AstModule` module. This lets us put the adder cell
inside it. We'll pull the contents out and plonk them into 'baz' when
we actually do the binding operation as part of the hierarchy pass.

Of course, we don't want `RTLIL::Binding` to contain an
`AST::AstModule` (since kernel code shouldn't depend on a frontend), so
we define `RTLIL::Binding` as an abstract base class and put the
AST-specific code into an `AST::Binding` subclass. This is analogous to
the `AST::AstModule` class.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -565,6 +565,7 @@ $(eval $(call add_include_file,kernel/yosys.h))
 $(eval $(call add_include_file,kernel/hashlib.h))
 $(eval $(call add_include_file,kernel/log.h))
 $(eval $(call add_include_file,kernel/rtlil.h))
+$(eval $(call add_include_file,kernel/binding.h))
 $(eval $(call add_include_file,kernel/register.h))
 $(eval $(call add_include_file,kernel/celltypes.h))
 $(eval $(call add_include_file,kernel/celledges.h))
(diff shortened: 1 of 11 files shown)


rswarbrick:
Hi guys! It's been a fortnight. Anything I can do to make this easier to merge?

zachjs:
At a high level this seems reasonable to me, but I think changes to RTLIL should probably be reviewed by @clairexen.

clairexen (approved):
LGTM

clairexen:
> (Note: This patch is part of the larger queue at #2752. The idea is to split things out to avoid having a giant PR to review all at once)

I can't thank you enough for merging in your work this way.

Having the giant PR available in a draft PR while reviewing a reasonable change set at a time is gold.  <3

rswarbrick:
A pleasure! Also, I spend a lot of time reviewing patches too, so am *very* aware of how painful a 1k-line change is to look through :-)

PR #2878, opened 2021-07-17
Restructure passes/techmap/abc.cc to better support ASIC synthesis.

macd:
This PR does several things to better support ASIC synthesis.  First,
the file naming and temporary directories for file transfer to ABC
are enhanced.  The string `yosys-abc-XXXXX` is used for a top level
directory.  Each module in the design then gets its own directory based
on the module name inside of this top level directory.  These subdirectories
contain the usual abc.script, input.blif, output.blif etc files.  To
specify this, a new flag has been added to the abc command and documented
in the help as:

```
    -abctop_dir <directory name>
        set the root level of the abc work directory to be <directory name>.
        A sub-directory with the name 'yosys-abc-XXXXX' (where XXXXX will be replaced
        by a random string) will be created here. Inside of this directory,
        for each module a directory will be created for file transfer
        to and from ABC. All will be deleted on exit if cleanup=true. The default is /tmp
```

Second, the function abc_module has been broken up and out of it a new function
named abc_reintegrate has been formed.  Its only job is to reintegrate a output.blif
back into the design in place of a specified module. The function abc_module now calls
abc_reitegrate after running ABC on input.blif file.

Third, using the new abc_reintegrate function, a new pass named AbcReintegratePass has
been defined.  The corresponding command name is `abc-reint`. From the help text:

```
yosys> help abc-reint

    abc-reint [options] [selection]

This pass reintegrates ABC mapped modules back into the design

    -abcdir_name <directory name>
        set the root level of the abc work directory to be <directory name>.
        sub-directories for each module are expected here, each with an output.blif
        file produced by ABC.
```

Note that -abctop_dir and -abcdir_name do _not_ point to the same directory.  The former
is the location in which to create the `yosys-abc-XXXXX` directory.  The latter is the actual
directory named `yosys-abc-XXXXX` (after it has been created).

How does this help ASIC synthesis? As many know, synthesis is a very noisy process.  There
is no one script, even if iterated multiple times, that always gives superior results. The
variability in area and timing can be quite large.  This change allows us to run multiple
strategies to create a set of viable alternatives with which we can then use to construct
a better design.

Note that the tests/techmap all pass with the exception of mem_simple_4x1 which appears to
be a pre-existing failure.


QuantamHD:
@eddiehung Could you review this change? It would be really cool to have the functionality in yosys.

eddiehung:
Sorry, not sure I'm the right person to look at this since I'm not familiar with how ABC is used for ASIC mapping. Maybe @clairexen ?

PR #2892, opened 2021-07-27
opt_expr: Propagate constants to port connections.

mwkmwkmwk:
This adds one simple piece of functionality to opt_expr: when a cell
port is connected to a fully-constant signal (as determined by sigmap),
the port is reconnected directly to the constant value.  This is just
enough optimization to fix the "non-constant $meminit input" problem
without requiring a full opt_clean or a separate pass.


sjkelly:
We have a few cases of these in our source. Testing this branch I see simplifications which dropped our LUT count 1-2% depending on config.

mwkmwkmwk:
> We have a few cases of these in our source. Testing this branch I see simplifications which dropped our LUT count 1-2% depending on config.

That... should not happen, actually — the simplification here is a small subset of transforms done by `opt_clean`, which should be run together with `opt_expr` in optimization loop in a full synthesis flow

PR #2897, opened 2021-07-29
opt_lut: Allow more than one -dlogic per cell type.

### LINKED ISSUE #2061: ice40: opt_lut can break carry packing by moving SB_CARRY out to another input than I3

Here's a fairly minimal example that show the issue :

```
module bug #(
    parameter integer WIDTH = 12
)(  
    output reg  [WIDTH:0] cnt,
    input  wire clk,
    input  wire rst
);
    wire last_n;

    assign last_n = cnt[WIDTH];

    always @(posedge clk or posedge rst)
        if (rst)
            cnt <= 0;
        else
            cnt <= last_n ? ( cnt + { (WIDTH+1){last_n} } ) : 13'h1aaa;

endmodule
```

Build with : 

```
yosys -p 'synth_ice40  -json bug.json; show' bug.v
```

And if you look at the LUT that generates bit 12, you'll see the carry out of the previous bit is connected to I2 instead of I3.

This prevents nextpnr to properly pack the carry chain.

AFAICT this "bug" is added during the `opt_lut` pass. bit 12 doesn't have a SB_CARRY associated with it since we don't care about its carry output.

## Expected behavior

opt_lut should definitely optimize the two luts to a single one, but it should keep the carry out on I3.

## Actual behavior

This is the invalid conversion that's done :

lut_opt

mwkmwkmwk:
Fixes #2061.

### DIFF
--- a/passes/opt/opt_lut.cc
+++ b/passes/opt/opt_lut.cc
@@ -24,16 +24,22 @@
 USING_YOSYS_NAMESPACE
 PRIVATE_NAMESPACE_BEGIN
 
+struct dlogic_t {
+	IdString cell_type;
+	// LUT input idx -> hard cell's port name
+	dict<int, IdString> lut_input_port;
+};
+
 struct OptLutWorker
 {
-	dict<IdString, dict<int, IdString>> &dlogic;
+	const std::vector<dlogic_t> &dlogic;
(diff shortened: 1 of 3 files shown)


whitequark (approved):
LGTM

PR #2904, opened 2021-08-01
kernel/mem: Introduce transparency masks.

mwkmwkmwk:
This should be in mergeable state, but I'd like to delay the merge until it has been properly validated by the following follow-up PRs:

- #2807, rebased on top of this and with transparency mask support in new cells
- future PR that will add transparency logic recognition to `memory_dff`


mwkmwkmwk:
Having designed the transparency logic recognition algorithm, I am now reasonably convinced that this model change is a good direction and consider this ready for review.

PR #2911, opened 2021-08-03
remove apparently erroneously added `yosys.wasm` target

justchen1369:


justchen1369:
Elaborating: upon trying to compile for wasm, `make` errored with "could not find target yosys.wasm". upon perusing the Makefile, I found that the `yosys.wasm` both does not exist and is not necessary for compiling for wasm, hence this PR to remove it from the Makefile.

Makefile
```diff
@@ -272,7 +272,7 @@ viz.js:
 	wget -O viz.js.part https://github.com/mdaines/viz.js/releases/download/0.0.3/viz.js
 	mv viz.js.part viz.js
 
-yosysjs-$(YOSYS_VER).zip: yosys.js yosys.wasm viz.js misc/yosysjs/*
+yosysjs-$(YOSYS_VER).zip: yosys.js viz.js misc/yosysjs/*
 	rm -rf yosysjs-$(YOSYS_VER) yosysjs-$(YOSYS_VER).zip
 	mkdir -p yosysjs-$(YOSYS_VER)
 	cp viz.js misc/yosysjs/* yosys.js yosys.wasm yosysjs-$(YOSYS_VER)/
```
clairexen, line 278: This `cp` command can not possibly succeed without a yosys.wasm file

justchen1369: The `yosys.wasm` file is generated with `yosys.js`, confusingly enough.

PR #2917, opened 2021-08-05
Allow optional comma after last entry in enum

msinger:
An extra comma is already allowed in module port lists and task/function argument lists, so I thought, why not in enum.

### DIFF
--- a/frontends/verilog/verilog_parser.y
+++ b/frontends/verilog/verilog_parser.y
@@ -1624,17 +1624,18 @@ enum_type: TOK_ENUM {
 		// create the template for the names
 		astbuf1 = new AstNode(AST_ENUM_ITEM);
 		astbuf1->children.push_back(AstNode::mkconst_int(0, true));
-	 } enum_base_type '{' enum_name_list '}' {	// create template for the enum vars
-							auto tnode = astbuf1->clone();
-							delete astbuf1;
-							astbuf1 = tnode;
-							tnode->type = AST_WIRE;
-							tnode->attributes[ID::enum_type] = AstNode::mkconst_str(astbuf2->str);
-							// drop constant but keep any range
-							delete tnode->children[0];
-							tnode->children.erase(tnode->children.begin());
-							$$ = astbuf1; }
-	 ;
+	} enum_base_type '{' enum_name_list optional_comma '}' {
+		// create template for the enum vars
+		auto tnode = astbuf1->clone();
+		delete astbuf1;
+		astbuf1 = tnode;
+		tnode->type = AST_WIRE;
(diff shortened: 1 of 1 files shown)


udif:
Adding my 2c's

I don't know exactly what Yosys supports or not, but the extra comment is not simply a "why not" item. It is an explicitly supported as part of a more general (mis)feature in the Verilog standard called "null ports" (although I personally wish it would have never been added), and applies only for non-ANSI module declarations. It is not supported by ANSI-style module declarations. I have never checked whether this applies to tasks/functions as well or not.

Feel free to follow the following to read more about it:
https://github.com/verilator/verilator/issues/2874
https://github.com/verilator/verilator/issues/2875

If it was up to me I wouldn't add it for a "why not" reason, definitely without checking first if this agrees with the standard as well as other tools.

msinger:
@udif thank you very much for that information. I wonder how they came up with that idea. :/
But as long as there are no "null names" in an `enum`, I think this optional comma would be nice to have. ^^

udif:
The question that needs to be answer is if the Yosys Verilog/SystemVerilog should adhere to the standard, or add arbitrary extensions. If the latter, I would suggest enabling these via a special read_verilog flag to enable this.

zachjs (changes requested):
Optional commas are nice, and we are generally comfortable adding support in places where the LRM has not already ascribed a meaning to the stray comma, given the intent is likely clear. However, I'd like to avoid the noise of reindenting this whole block. Could we consider only changing the first line, or instead modifying `enum_name_list`?

msinger:
@zachjs, yes, I understand, I don't like it either to modify more than necessary, because as you said, the diff gets noisy and maybe the original author is sad if I disagree with their style.
I thought this was the best option here though. My intend was to add the extra comma in the same way it was added for the module ports and the function arguments. I can't see how I could easily add the comma to `enum_name_list` without unnecessarily adding an extra layer around it, because it references itself, which would allow unlimited extra commas in between.
I tried to match the indentation style with the surrounding code. By having the indentation not being dependent on the length of the first line, it makes it easier to make changes in the future, without the need of constantly reindenting everything. But if you prefer, then I change only the first line, but this messes up the alignment with the other lines.
What do you think?

zachjs (approved):
This is the only instance of this formatting in the parser. I'm usually not a fan of unnecessary whitespace changes, but this is well localized and restores some uniformity. For these reasons, I'm going to merge this as-is.

As a side note, there is a way to formulate such productions without a wrapping rule.
```
foos
  : foo optional_comma
  | foo ',' foos
```
I have heard that this can have a performance impact under certain setups, as all the `foo`s have to be shifted before they are reduced, but I'm not sure the potential impact is relevant here.

PR #2925, opened 2021-08-10
memory_dff: Recognize soft transparency logic.

mwkmwkmwk:
Depends on #2904 and #2807 .  Still needs tests.


mwkmwkmwk:
Tests added.

PR #2930, opened 2021-08-11
memory_share: Add -nosat and -nowiden options.

mwkmwkmwk:
This unlocks wide port recognition by default.

Submitting as draft, as it's still missing tests.


mwkmwkmwk:
Tests added; this PR now depends on #2933, #2932, #2931, #2925.

PR #2932, opened 2021-08-12
logger: Add -check-expected subcommand.

mwkmwkmwk:
This allows us to have multiple "expect this warning" calls in a single
long script, covering only as many passes as necessary.


mmicko (approved):
LGTM

Issue #2916, opened 2021-08-05
active low enable for tristate buffer


pepijndevos:
@yrabbit found that Gowin tristate buffers "OEN" inputs are active low, which I confirmed, but `synth_gowin` does not treat them as such.

https://github.com/YosysHQ/yosys/blob/2e421feb0ea526468493ab008f3c72beb12c2bc6/techlibs/gowin/synth_gowin.cc#L282

To reproduce, take the following code and synthesize it with Yosys and the Gowin IDE. You will see the vendor produces an IBUF wired straight to a TBUF, while `synth_gowin` generate a `LUT1` inverter between the two.

```
`default_nettype none
`define LEDS_NR 3
// TBUF
module bidir(
    inout wire io_pin,
    input wire i_output,
    input wire data_to_pin);

    assign io_pin = (i_output == 1) ? data_to_pin : 1'bz;
endmodule

//
module top (
    input wire clk,
    input wire key_n,
    inout wire [`LEDS_NR-1:0] led
);

wire [`LEDS_NR-1:0] w_dir;

bidir bd[`LEDS_NR-1:0](led, w_dir, 1'b1);

assign w_dir[`LEDS_NR-1:1] = {`LEDS_NR-1{1'b0}};
// switch the direction with the key
assign w_dir[0] = !key_n;
endmodule
```

I don't see any support in `iopadmap` for this. But it could be it needs to happen at `$_TBUF_` inference or somewhere else.

What's the correct way forward?

gatecat:
You can use the same "temporary cell and techmap" trick as Nexus and Xilinx do:

https://github.com/YosysHQ/yosys/blob/2e421feb0ea526468493ab008f3c72beb12c2bc6/techlibs/nexus/synth_nexus.cc#L336

https://github.com/YosysHQ/yosys/blob/2e421feb0ea526468493ab008f3c72beb12c2bc6/techlibs/nexus/cells_map.v#L56

pepijndevos:
My concern with that is, at least in the current flow iopadmap happens quite late so there isn't much opportunity to optimize away the inverter. But I guess that's what the `run("opt -fast");` line is for. Thanks, I'll give it a go.

mwkmwkmwk:
This is exactly why it's called so early in the Xilinx flow — move it likewise in the Gowin flow.

pepijndevos:
[...]

PR #2971, opened 2021-08-19
Gowin: deal with active-low tristate

pepijndevos:
Now that we're adding support in Apicula for inout and tristate buffers we're finding that they are actually active-low, causing inverted behavior.

This PR matches the behavior of Nexus explained in #2916 by using an intermediate cell.
It also moves the iopadmap earlier into the process, so that there is more room to optimize away the inverter.

### DIFF
--- a/techlibs/gowin/cells_map.v
+++ b/techlibs/gowin/cells_map.v
@@ -127,7 +127,7 @@ module \$__GW_IOBUF (input I, OE, output O, inout IO);
 endmodule
 
 module \$__GW_TBUF (input I, OE, output O);
-	TBUF _TECHMAP_REPLACE_ (.I(I), .O(), .OEN(~OE), .O(O));
+	TBUF _TECHMAP_REPLACE_ (.I(I), .OEN(~OE), .O(O));
 endmodule
 
 module \$lut (A, Y);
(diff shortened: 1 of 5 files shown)


techlibs/gowin/cells_map.v
```diff
@@ -122,6 +122,13 @@ module	\$_DFFE_NP0P_ (input D, C, R, E, output Q);
 	wire _TECHMAP_REMOVEINIT_Q_ = 1;
 endmodule
 
+module \$__GW_IOBUF (input I, OE, output O, inout IO);
+	IOBUF _TECHMAP_REPLACE_ (.I(I), .O(O), .OEN(~OE), .IO(IO));
+endmodule
+
+module \$__GW_TBUF (input I, OE, output O);
+	TBUF _TECHMAP_REPLACE_ (.I(I), .O(), .OEN(~OE), .O(O));
```
mwkmwkmwk, line 130: `.O` is connected twice here, please remove.

mwkmwkmwk (changes requested):
See inline comment; also, the sim models in cells_sim need changing to account for the active-low enable.

techlibs/gowin/cells_map.v
pepijndevos: Also removed in nexus where I copied this from

mwkmwkmwk (changes requested):
Please fix the tests.

pepijndevos:
Hm so on CI it errors on the LUT1 which is easy enough to fix, but locally it fails on the opt_equiv assertion, which I'm not sure how to deal with.

mwkmwkmwk (approved):

PR #2979, opened 2021-08-22
Add DLLDELD

ECP5-PCIe:


mwkmwkmwk (approved):

ECP5-PCIe:
Somehow I get a "ERROR: cell type 'DLLDELD' is unsupported (instantiated as 'U$$0.U$$0')" in NextPNR

gatecat:
Yeah, it's not supported. If you have a use case I'm happy to work on adding support (hopefully tomorrow), it shouldn't take long.

ECP5-PCIe:
Yes, that would be nice, since I want to use it to delay signals over a larger range than a DelayF can

gatecat:
Would you be able to point me to some code that's using it so I have a reference point to work with?

PR #2982, opened 2021-08-25
preprocessor: do not destroy double slash escaped identifiers

tsailer:
The preprocessor currently destroys double slash containing escaped
identifiers (for example \a//b ). This is due to next_token trying to
convert single line comments (//) into /* */ comments. This then leads
to an unintuitive error message like this:
ERROR: syntax error, unexpected '*'

This patch fixes the error by recognizing escaped identifiers and
returning them as single token. It also adds a testcase.

### DIFF
--- a/frontends/verilog/preproc.cc
+++ b/frontends/verilog/preproc.cc
@@ -142,6 +142,16 @@ static std::string next_token(bool pass_newline = false)
 				return_char(ch);
 		}
 	}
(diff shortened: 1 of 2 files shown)


zachjs (approved):
Sorry this dropped off my radar! This change seems to be exactly in line with Section 5.6.1 of IEEE 1800-2017. Thank you!

PR #2986, opened 2021-08-30
sv: support declaration in procedural for initialization

zachjs:
In line with other tools, this adds an extra wrapping block around such for loops to appropriately scope the variable.

### DIFF
--- a/frontends/verilog/verilog_parser.y
+++ b/frontends/verilog/verilog_parser.y
@@ -2607,6 +2607,53 @@ asgn_binop:
 	TOK_SSHL_ASSIGN { $$ = AST_SHIFT_SLEFT; } |
 	TOK_SSHR_ASSIGN { $$ = AST_SHIFT_SRIGHT; } ;
 
+for_initialization:
+	TOK_ID '=' expr {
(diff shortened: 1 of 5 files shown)


frontends/verilog/verilog_parser.y
```diff
@@ -2607,6 +2607,53 @@ asgn_binop:
...
+		AstNode *ident = new AstNode(AST_IDENTIFIER);
+		ident->str = *$1;
+		AstNode *node = new AstNode(AST_ASSIGN_EQ, ident, $3);
+		ast_stack.back()->children.push_back(node);
+		SET_AST_NODE_LOC(node, @1, @3);
+	} |
+	wire_type range TOK_ID {
+		frontend_verilog_yyerror("For loop variable declaration is missing initialization!");
+	} |
+	wire_type range TOK_ID '=' expr {
```
mwkmwkmwk, line 2621: `wire_type` is *very* permissive — this will allow things like `for (inout wire [3:0] i = ...`, which are patent nonsense; cleaning that up fully seems to be too much work of course, but I think at least changing to `non_io_wire_type` would disallow some things that belong here at zero cost

mwkmwkmwk (approved):

PR #2987, opened 2021-08-30
sv: support declaration in generate for initialization

zachjs:
This is accomplished by generating a unique name for the genvar, renaming references to the genvar only in the loop's initialization, guard, and incrementation, and finally adding a localparam inside the loop body with the original name so that the genvar can be shadowed as expected.

### DIFF
--- a/tests/verilog/genvar_loop_decl_1.ys
+++ b/tests/verilog/genvar_loop_decl_1.ys
@@ -1,0 +1,14 @@
+read_verilog -sv genvar_loop_decl_1.sv
+
+select -assert-count 1 gate/genblk1[0].x
+select -assert-count 1 gate/genblk1[1].x
+select -assert-count 0 gate/genblk1[2].x
+
+select -assert-count 1 gold/genblk1[0].x
+select -assert-count 1 gold/genblk1[1].x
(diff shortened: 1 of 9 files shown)


tests/verilog/genvar_loop_decl_1.ys
```diff
@@ -0,0 +1,14 @@
...
+select -assert-count 1 gate/genblk1[0].x
+select -assert-count 1 gate/genblk1[1].x
+select -assert-count 0 gate/genblk1[2].x
+
+select -assert-count 1 gold/genblk1[0].x
+select -assert-count 1 gold/genblk1[1].x
+select -assert-count 0 gold/genblk1[2].x
+
+proc
+equiv_make gold gate equiv
```
mwkmwkmwk, line 12: This equivalence check doesn't actually do anything here, since the x wires are undriven — the equiv still passes if eg. I have wrong widths for `x` in some iteration

mwkmwkmwk (changes requested):
OK to merge once something is done about the no-op equivalence check in that test — either add some drivers to the `x` wires, or just remove the equiv check entirely

PR #3000, opened 2021-09-10
yosys-smtbmc: Fix reused loop variable.

### LINKED ISSUE #2999: Writing a register array does not work
## How to reproduce

The setup: I have a memory with a read and write port, `reg [31:0] mem [0:3]`. A simple counter will serve us as address. The memory is initialized to all `0xAAAA` and we want to write it all to `0xBBBB`. Here's the Verilog code (The contents of `mem_reset.bin` is all `0xAAAA`, you can imagine):

```verilog
module MulAccVerifier (
  input               reset,
  input               clk
);
  reg [2:0] counter;
  reg [31:0] mem [0:3];

  /* Init handling */
  always @(posedge clk) begin
    if($initstate()) begin
      assume(reset);
    end
  end
  initial begin
    $readmemb("mem_reset.bin",mem);
  end

  always @(posedge clk) begin
    if(!reset) begin
      if(counter == 3'b111) begin
        /* Fail on purpose for debugging */
        assert(1'b0);
        assert((mem[counter[1:0]] == 32'hBBBB));
      end
    end
  end

  always @(posedge clk) begin
    if(reset) begin
      counter <= 3'b000;
    end else begin
      counter <= (counter + 3'd001);
      mem[counter[1:0]] <= 32'hBBBB;
    end
  end
endmodule
```

## Actual behavior

image

The writes are very sporadic. They may be cycles late, and more often than not only one of the values is written at all. This is really non-deterministic based on what the solver produces as counter-example (well each run gives the same result, but the result may vary if the net list changes), and I couldn't really figure out a pattern.

Note that I am running this through SymbiYosys because this initially came up during formal tests and the issue didn't appear in my simulation (I think it's using Verilator, but not sure). The `.sby` file is very simple, it sets `mode bmc`, `depth 15`, `[engines] smtbmc` and reads the verilog and the bin file.

## Expected behavior

image

I got this behavior by replacing the `mem[counter[1:0]] <= 32'hBBBB;` line in the above snippet with

```verilog
      int i;
      for (i = 0; i < 4; i = i + 1) begin
        if (counter[1:0] == i) begin
          mem[i] <= 32'hBBBB;
        end
      end
```

With that change, Verilog prints `Warning: Replacing memory \mem with list of registers.` which I think may be relevant to locate the root source of the bug.

I did not check in-depth for issue duplicates, but this may be related to #2570. My `yosys` version is `Yosys 0.9+4052 (git sha1 687f381b69, gcc 10.3.0 -fPIC -Os)`, I will try again with master tomorrow. `symbiyosys` is from master.

mwkmwkmwk:
Fixes #2999.

### DIFF
--- a/backends/smt2/smtbmc.py
+++ b/backends/smt2/smtbmc.py
@@ -771,12 +771,12 @@ def write_vcd_trace(steps_start, steps_stop, index):
 
                         if gotread:
                             buf = data[:]
-                            for i in reversed(range(len(tdata))):
+                            for ii in reversed(range(len(tdata))):
                                 for k in range(width):
-                                    if tdata[i][k] == "x":
-                                        tdata[i][k] = buf[k]
+                                    if tdata[ii][k] == "x":
+                                        tdata[ii][k] = buf[k]
                                     else:
-                                        buf[k] = tdata[i][k]
+                                        buf[k] = tdata[ii][k]
 
                     if not asyncwr:
                         tdata.append(data[:])


piegamesde:
Thanks for the quick overnight fix! I can confirm that it resolves the issue.

PR #3007, opened 2021-09-13
synth_gatemate: Initial implementation

pu-cc:
This PR adds synthesis support for the Cologne Chip GateMate FPGA. This includes:

- mapping of LUTs (see notes below) and dedicated MUXs
- mapping of arbitrarily-sized multipliers
- mapping of full adder cells
- block RAM rules and initialization
- blackbox cells, such as PLLs
- flip-flops, latches, iopads, clkbufmap, ...
- test cases

Primitives are decribed here:

https://www.colognechip.com/docs/ug1001-gatemate1-primitives-library-2021-08.pdf

We are currently using `abc` to map logic to LUT1-LUT4. The CPE architecture, however, is no typical LUT but a LUT2-tree with up to either 4 or 8 inputs. The place and route currently splits these primitives into (in certain circumstances several) CPEs. So there is still some open work to be done here. A mapping to LUT2-tree primitives with up to either 4 or 8 inputs would be ideal.

Since a CPE can also be used as MUX4 or MUX8, these primitives are mapped before the `abc` LUT pass.

Moreover, the FPGA architecture does not have dedicated DSP resources. Instead, each CPE has a 2x2 multiplier that can be expanded to any size using multiple CPE. We have therefore decided to provide the multiplier primitives with parameters to set their widths. The place and route takes care of the mapping it into the CPE array.


techlibs/gatemate/cells_bb.v
```diff
@@ -0,0 +1,124 @@
...
+	parameter [95:0] PLL_CFG_B = 96'bx
+)(
+	input  CLK_REF, CLK_FEEDBACK, USR_CLK_REF,
+	input  USR_LOCKED_STDY_RST, USR_SEL_A_B,
+	output USR_PLL_LOCKED_STDY, USR_PLL_LOCKED,
+	output CLK270, CLK180, CLK90, CLK0, CLK_REF_OUT
+);
+endmodule
+
+(* blackbox *)
```
mwkmwkmwk, line 48: This could use a `(* keep *)` as well (otherwise a tx-only serdes with all output pins unconnected could just disappear)

techlibs/gatemate/cells_sim.v
```diff
@@ -0,0 +1,1574 @@
...
+	endspecify
+endmodule
+
+
+module CC_MX2 (
+	input  D0, D1,
+	input  S0,
+	output reg Y
+);
+	always @(*) begin
```
mwkmwkmwk, line 414: This (and other `CC_MX`) doesn't handle X bits in the select — consider doing this via a `?:` tree instead

techlibs/gatemate/synth_gatemate.cc
```diff
@@ -0,0 +1,396 @@
...
+		}
+
+		if (check_label("map_ffram"))
+		{
+			run("opt -fast -mux_undef -undriven -fine");
+			run("memory_map");
+			run("opt -undriven -fine");
+		}
+
+		if (check_label("map_addf", "(skip if '-noaddf')"))
```
mwkmwkmwk, line 278: This step is badly named (and misdocumented), as it runs the whole bit-blasting `techmap` ruleset, not just CC_ADDF

techlibs/gatemate/synth_gatemate.cc
```diff
@@ -0,0 +1,396 @@
...
+			if (retime) {
+				run("abc -dff -D 1", "(only if -retime)");
+			}
+		}
+
+		if (check_label("map_io", "(skip if '-noiopad')"))
+		{
+			if (!noiopad) {
+				run("iopadmap -bits "
+					"-inpad $__inpad Y:I "
```
mwkmwkmwk, line 295: This could be simplified by removing `$__inpad` and `$__outpad`, since `CC_IBUF` and `CC_OBUF` can be mapped by `iopadmap` directly

techlibs/gatemate/synth_gatemate.cc
```diff
@@ -0,0 +1,396 @@
...
+				);
+				run("techmap -map +/gatemate/iob_map.v");
+				run("clean");
+			}
+		}
+
+		if (check_label("map_regs"))
+		{
+			run("opt_clean");
+			run("dfflegalize -cell $_DFF_?_ 0 -cell $_DFF_???_ 0 "
```
mwkmwkmwk, line 308: The rules for everything but `$_DFFE_????_` and `$_DLATCH_???_` are redundant and can be safely removed — `dfflegalize` already performs the job of upgrading FF types by adding always-0 reset / always-1 enable signals, and the flow doesn't benefit from special-casing them in techmap

techlibs/gatemate/synth_gatemate.cc
```diff
@@ -0,0 +1,396 @@
...
+				"-cell $_DFFE_??_ 0 -cell $_DFFE_????_ 0 "
+				"-cell $_DLATCH_?_ x -cell $_DLATCH_???_ x"
```
mwkmwkmwk, line 310: The sim models seem to imply that latches are always initialized to 0, while here the rule implies they have no initialization capability; which one is correct?

techlibs/gatemate/cells_bb.v
```diff
@@ -0,0 +1,124 @@
...
+ *  WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ *  MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ *  ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ *  WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ *  ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ *  OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *
+ */
+
+(* blackbox *)
```
mwkmwkmwk, line 20: What is the purpose of spliting this file off from `cells_sim.v`? It seems the other file also has a blackbox (the FIFO cell)

techlibs/gatemate/cells_sim.v
```diff
@@ -0,0 +1,1574 @@
...
+);
+	assign O = I;
+
+	specify
+		(I => O) = (0:0:0, 0:0:0);
+	endspecify
+endmodule
+
+
+(* blackbox *)
```
mwkmwkmwk, line 534: This (and the other BRAM) seems to have an implementation, why is it marked as `blackbox`?

techlibs/gatemate/mul_map.v
```diff
@@ -0,0 +1,79 @@
...
+ *  WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ *  MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ *  ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ *  WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ *  ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ *  OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *
+ */
+
+`define MAX(a,b) (a > b ? a : b)
```
mwkmwkmwk, line 20: `((a) > (b) ? (a) : (b))`

techlibs/gatemate/synth_gatemate.cc
pu-cc: Thanks for your review. Let us call this step "map_gates". Would you recommend including the steps from "map_mux" here?

techlibs/gatemate/synth_gatemate.cc
pu-cc: Ok - I removed everything but `$_DFFE_????_` and `$_DLATCH_???_`. Do I have to remove the techmap models in `map_regs.v` as well?

techlibs/gatemate/synth_gatemate.cc
pu-cc: Good catch. Latches and flipflops cannot be initialized during configuration. I fixed the simulation model and the `dfflegalize` call.

techlibs/gatemate/cells_bb.v
pu-cc: All blackbox cells are now in `cells_bb.v` and all other cells that have an implementation are in `cells_sim.v`.

techlibs/gatemate/synth_gatemate.cc
mwkmwkmwk: Yes, remove them — they became unused with this change.

techlibs/gatemate/mul_map.v
```diff
@@ -0,0 +1,79 @@
...
+		end
+		else begin: blkB
+			wire [MAXWIDTH-1:0] Bext = B;
+		end
+
+		if (A_WIDTH >= B_WIDTH) begin
+			CC_MULT #(
+				.A_WIDTH(MAXWIDTH),
+				.B_WIDTH(MAXWIDTH),
+				.P_WIDTH(`MIN(Y_WIDTH,MAXWIDTH+MAXWIDTH)),
```
mwkmwkmwk, line 59: this part is worrying — given big enough `Y_WIDTH`, this results in creating (in RTLIL) a cell with the `P` output of mismatched with to `P_WIDTH`

I don't think it can actually happen in the final flow (since some optimizations run first that will shorten the multiplier), but it'd be safer to provide a narrower temporary signal here, then reconstruct the actual `Y` output from it + possibly sign extension

techlibs/gatemate/reg_map.v
```diff
@@ -0,0 +1,108 @@
...
+ *  WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ *  MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ *  ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ *  WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ *  ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ *  OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *
+ */
+
+(* techmap_celltype = "$_DFF_[NP]_" *)
```
mwkmwkmwk, line 20: these extra rules still need to be removed

techlibs/gatemate/cells_sim.v
```diff
@@ -0,0 +1,1485 @@
...
+			$finish();
+		end
+		if ((WIDTH_MODE_B != 40) && (WIDTH_MODE_B != 20) && (WIDTH_MODE_B != 10) &&
+			(WIDTH_MODE_B != 5)  && (WIDTH_MODE_B != 2)  && (WIDTH_MODE_B != 1) && (WIDTH_MODE_B != 0)) begin
+			$display("ERROR: Illegal %s Port B width configuration %d.", RAM_MODE, WIDTH_MODE_B);
+			$finish();
+		end
+		// RAM initialization
+		for (i=0; i < 320; i=i+1) begin
+			memory[320*0+i]  = INIT_00[i];
```
mwkmwkmwk, line 679: interpreting this loop takes a good chunk of time when parsing this file; replace it with 320-bits-at-a-time writes like this (without loop):

```
                        memory[320*0+319:320*0]   = INIT_00;                                                             
                        memory[320*1+319:320*1]   = INIT_01;                                                 
                        memory[320*2+319:320*2]   = INIT_02; 
                        memory[320*3+319:320*3]   = INIT_03; 
                        memory[320*4+319:320*4]   = INIT_04;                                                       
                        memory[320*5+319:320*5]   = INIT_05;                                                             
                        memory[320*6+319:320*6]   = INIT_06;                                                 
[...]
```

techlibs/gatemate/cells_sim.v
```diff
@@ -0,0 +1,1485 @@
...
+			$display("ERROR: Cascade feature only supported in 1 bit data width mode.");
+			$finish();
+		end
+		if ((CAS != "NONE") && (RAM_MODE != "TDP")) begin
+			$display("ERROR: Cascade feature only supported in TDP mode.");
+			$finish();
+		end
+		// RAM initialization
+		for (i=0; i < 320; i=i+1) begin
+			memory[320*0+i]   = INIT_00[i];
```
mwkmwkmwk, line 1167: same here

techlibs/gatemate/synth_gatemate.cc
mwkmwkmwk: no, map_muxs can stay where it is

tests/arch/gatemate/run-test.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
```
mwkmwkmwk, line 1: these tests aren't doing much good, as this script isn't actually *run* from the top-level Makefile, please add it here (and fix the test failures)

techlibs/gatemate/gatemate_bramopt.cc
```diff
@@ -0,0 +1,150 @@
...
+		{
+			if (args[argidx] == "-noglwren") {
+				noglwren = true;
+				continue;
+			}
+			break;
+		}
+		extra_args(args, argidx, design);
+
+		Module *module = design->top_module();
```
mwkmwkmwk, line 137: don't look for the top module — instead, iterate over all modules in the design like most opt passes do

the current code will completely skip all non-top modules in noflatten mode

techlibs/gatemate/mul_map.v
pu-cc: Not sure I got it right. The `MAXWIDTH` parameter is now divided into A_* and B_*. Since the `CC_MULT` is a signed multiplier, we just have to make sure that the vectors are preceded by a `1'b0` in case of an unsigned multiplication, such as:

```
  CC_MULT #(
    .A_WIDTH(32'd17),
    .B_WIDTH(32'd11),
    .P_WIDTH(32'd26)
  ) _115_ (
    .A({ 1'h0, a_i }), // a_i is 16 bits
    .B({ 1'h0, b_i }), // b_i is 10 bits
    .P(_006_) // result _006_ is 26 bits
  );
```

tests/arch/gatemate/run-test.sh
pu-cc: Tests failed due to the unsupported initilization values for FF cells. I've set the init values in `dfflegalize` and in the simulation models to zero again.

mwkmwkmwk: hold on here; whatever you do in the flow has to match what the hardware actually supports

does the hardware always initialize registers to 0 after all?

if yes, the techmap rule for dlatch needs to be updated to use the removeinit thing

if no, please go back to using `x` in `dfflegalize` and sim models, then fix the tests to not use features that the hardware cannot support

techlibs/gatemate/mul_map.v
mwkmwkmwk: looks correct to me (though I'd rename `MAXWIDTH` to something else like `ADJWIDTH`, since there's no longer a MAX operation involved here)

techlibs/gatemate/mul_map.v
```diff
@@ -0,0 +1,80 @@
...
+ *  WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ *  MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ *  ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ *  WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ *  ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ *  OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *
+ */
+
+`define MAX(a,b) ((a) > (b) ? (a) : (b))
```
mwkmwkmwk, line 20: these macros are now unused and can be removed

techlibs/gatemate/mul_map.v
```diff
@@ -0,0 +1,80 @@
...
+				.P_WIDTH(Y_WIDTH),
+			) _TECHMAP_REPLACE_ (
+				.A(blkA.Aext),
+				.B(blkB.Bext),
+				.P(Y)
+			);
+		end
+		else begin // swap A,B
+			CC_MULT #(
+				.A_WIDTH(A_MAXWIDTH),
```
mwkmwkmwk, line 69: the widths also need to be swapped in this case

tests/arch/gatemate/run-test.sh
pu-cc: There is no hardware feature to pre-initialize registers during configuration. After the reset registers are zero. It can however happen that the value changes during the configuration of the CPEs. So I'd also rather use `x` in `dfflegalize` and sim models and adjust the tests.

Since it's only the `initial` blocks in `common/dffs.v` and `common/adffs.v` that lead to errors, I would add the `dffs` and `adffs` tests as verilog without `initial` to the gatemate test directory. That ok?

mwkmwkmwk: if you cannot guarantee 0 initialization, then yes, this is the correct course of action

I'd slightly prefer reusing the same test `.v` files, just with the `initial` statements wrapped in `ifndef`

pu-cc: Done as suggested in 19335d2

techlibs/gatemate/mul_map.v
pu-cc: Right, sounds better. Done in 5100daa.

techlibs/gatemate/cells_sim.v
```diff
@@ -0,0 +1,1481 @@
...
+			end
+		end
+		else if (RAM_MODE == "TDP") begin
+			// TDP port A
+			always @(posedge clka)
+			begin
+				for (i=0; i < WIDTH_MODE_A; i=i+1) begin
+					if (ena && wea && A_BM[i]) memory[addra+i] <= A_DI[i];
+
+					if (A_WR_MODE == "NO_CHANGE") begin
```
mwkmwkmwk, line 873: the read logic heredoesn't seem to correspond to the primitive documentation at all, and is a bit nonsensical

what I'd expect is roughly:

- NO_CHANGE: perform the read from memory iff `ena && !wea`
- WRITE_THROUGH: read from A_DI if `ena && wea && A_BM[i]`, otherwise read from memory

techlibs/gatemate/cells_sim.v
```diff
@@ -0,0 +1,1481 @@
...
+						if (enb) B_DO_out[k-40] <= memory[addrb+k];
+					end
+				end
+			end
+		end
+		else if (RAM_MODE == "TDP") begin
+			// TDP port A
+			always @(posedge clka)
+			begin
+				for (i=0; i < WIDTH_MODE_A; i=i+1) begin
```
mwkmwkmwk, line 1432: (same)

techlibs/gatemate/cells_sim.v
```diff
@@ -0,0 +1,1481 @@
...
+			$finish();
+		end
+		if ((A_ECC_EN != 1'b0) || (B_ECC_EN != 1'b0)) begin
+			$display("WARNING: ECC feature not supported in simulation.");
+		end
+		if ((A_ECC_EN == 1'b1) && (RAM_MODE != "SDP") && (WIDTH_MODE_A != 40)) begin
+			$display("ERROR: Illegal ECC Port A configuration. Must be SDP 40 bit, but is %s %d.", RAM_MODE, WIDTH_MODE_A);
+			$finish();
+		end
+		if ((WIDTH_MODE_A == 80) && (RAM_MODE == "TDP")) begin
```
mwkmwkmwk, line 1137: could also use validation of `WIDTH_MODE_*` for SDP case (ie. disallow width <80 with SDP)

techlibs/gatemate/cells_sim.v
```diff
@@ -0,0 +1,1481 @@
...
+	parameter INIT_78 = 320'h00000000000000000000000000000000000000000000000000000000000000000000000000000000;
+	parameter INIT_79 = 320'h00000000000000000000000000000000000000000000000000000000000000000000000000000000;
+	parameter INIT_7A = 320'h00000000000000000000000000000000000000000000000000000000000000000000000000000000;
+	parameter INIT_7B = 320'h00000000000000000000000000000000000000000000000000000000000000000000000000000000;
+	parameter INIT_7C = 320'h00000000000000000000000000000000000000000000000000000000000000000000000000000000;
+	parameter INIT_7D = 320'h00000000000000000000000000000000000000000000000000000000000000000000000000000000;
+	parameter INIT_7E = 320'h00000000000000000000000000000000000000000000000000000000000000000000000000000000;
+	parameter INIT_7F = 320'h00000000000000000000000000000000000000000000000000000000000000000000000000000000;
+
+	localparam WIDTH_MODE_A = (A_RD_WIDTH > A_WR_WIDTH) ? A_RD_WIDTH : A_WR_WIDTH;
```
mwkmwkmwk, line 1108: are mixed read-width / write-width ports supported in hardware? if so, why are the widths flattened here into one parameter without any validation?

note that there's a big new feature in progress that'd allow yosys to actually make use of such functionality via memory inference, so it'd be good to have the sim model support for that

techlibs/gatemate/brams_map.v
```diff
@@ -0,0 +1,539 @@
...
+
+	localparam INIT_CHUNK_SIZE = (CFG_DBITS <= 2) ? 256 : 320;
+
+	function [319:0] permute_init;
+		input [INIT_CHUNK_SIZE-1:0] chunk;
+		integer i;
+		begin
+			if (CFG_DBITS <= 2) begin
+				for (i = 0; i < 64; i = i + 1) begin
+					if (^chunk[i * 4 +: 4] === 1'bx) begin
```
mwkmwkmwk, line 61: I don't know what the purpose of this branch is, but it's sure to destroy some init data — note that yosys fills out init bits for unused memory cells / bitlanes with 'x, so it's likely that we end up here with such padding destroying some actual user's init data that happens to get packed into the same 4-bit-wide chunk

techlibs/gatemate/cells_sim.v
```diff
@@ -0,0 +1,1481 @@
...
+	input B_EN,
+	input A_WE,
+	input B_WE,
+	input [15:0] A_ADDR,
+	input [15:0] B_ADDR,
+	input [39:0] A_DI,
+	input [39:0] B_DI,
+	input [39:0] A_BM,
+	input [39:0] B_BM,
+	input A_CI,
```
mwkmwkmwk, line 945: the cascade logic isn't actually implemented here, making `$__CC_BRAM_CASCADE` untestable

techlibs/gatemate/brams_map.v
```diff
@@ -0,0 +1,539 @@
...
+
+	localparam INIT_CHUNK_SIZE = (CFG_DBITS <= 2) ? 256 : 320;
+
+	function [319:0] permute_init;
+		input [INIT_CHUNK_SIZE-1:0] chunk;
+		integer i;
+		begin
+			if (CFG_DBITS <= 2) begin
+				for (i = 0; i < 64; i = i + 1) begin
+					permute_init[i * 5 +: 5] = {1'b0, chunk[i * 4 +: 4]} & 5'b11111;
```
mwkmwkmwk, line 349: this `&` doesn't seem to be doing anything here

techlibs/gatemate/brams_map.v
```diff
@@ -0,0 +1,539 @@
...
+			.A_ADDR(A1ADDR),
+			.B_ADDR(B1ADDR),
+			.A_DI(A1DATA),
+			.B_DI(40'b0),
+			.A_BM(A1EN),
+			.B_BM(40'b0)
+		);
+
+		CC_BRAM_40K #(
+			`include "brams_init_40.vh"
```
mwkmwkmwk, line 507: this doesn't look correct (the two BRAMs should get two separate halves of the init, not the same half replicated)

techlibs/gatemate/gatemate_bramopt.cc
```diff
@@ -0,0 +1,146 @@
...
+	{
+		if (cell->type.in(ID(CC_DPSRAM_20K), ID(CC_DPSRAM_40K))) {
+			ram_cells.push_back(cell);
+		}
+	}
+
+	// Convert bitmask signals
+	for (auto cell : ram_cells)
+	{
+		std::vector<SigBit> mska = cell->getPort(ID(A_BM)).to_sigbit_vector();
```
mwkmwkmwk, line 46: this transformation is unsound — consider the following case:

```
reg [15:0] mem[0:1023];

initial $readmemh(...);

always @(posedge clk)
    if (we)
        mem[15:8] <= wdata;
```

In this case, the transformation as written would apply, overwriting all write enables with const-1, and making the memory write port overwrite the low 8 bits of the word on every write, which should never be changed

What *should* be done here, to get a sound transformation:

- check if there is exactly one SigBit other than S0 in the write enable, remember which indices are this SigBit and which are S0
- replace the main write enable with that SigBit, replace the corresponding bits within write masks with S1, but *don't* replace bits that were S0 originally
- if S0 is harder to materialize in P&R than S1 for some reason, an additional optimization could replace them with S1 for bit lanes which are never read (ie. their read port data bits are connected to nothing)

techlibs/gatemate/reg_map.v
```diff
@@ -0,0 +1,47 @@
...
+	parameter _TECHMAP_CELLTYPE_ = "";
+
+	CC_DFF #(
+		.CLK_INV(_TECHMAP_CELLTYPE_[39:32] == "N"),
+		.EN_INV(_TECHMAP_CELLTYPE_[15:8] == "N"),
+		.SR_INV(_TECHMAP_CELLTYPE_[31:24] == "N"),
+		.SR_VAL(_TECHMAP_CELLTYPE_[23:16] == "1")
+	) _TECHMAP_REPLACE_ (.D(D), .EN(E), .CLK(C), .SR(R), .Q(Q));
+
+	wire _TECHMAP_REMOVEINIT_Q_ = 1'b1;
```
mwkmwkmwk, line 32: this line is redundant, as initialized FFs cannot exist at this point

techlibs/gatemate/cells_sim.v
pu-cc: Actually, widths <80 for SDP are supported (makes little sense, but hardware supports it). `memory_bram` is configured to find 80-bit-wide SDP RAMs in 40K mode or 40-bit-wide SDP RAMS in 20K mode only. If we only map those two widths for SDP, we could reduce the assignments for `addra` and `addrb` to those two cases. What do you think?

techlibs/gatemate/brams_map.v
pu-cc: Absolutely right. Accidentally didn't delete that.

techlibs/gatemate/cells_sim.v
pu-cc: Just added a note that it is not yet supported. I will add this feature in my next commit, as we still have to test it in p&r...

techlibs/gatemate/brams_map.v
pu-cc: Agree, `brams_init_40.vh` now distinguishes between lower and upper cascade via define

techlibs/gatemate/gatemate_bramopt.cc
pu-cc: With the latest changes in `brams_map.v`in 09dd91e, the `gatemate_bramopt` pass has become obsolete. The current result is optimal for our p&r. We would therefore like to tend to remove this pass, if you don't mind.

tests/arch/gatemate/fsm.ys
```diff
@@ -0,0 +1,20 @@
...
+miter -equiv -make_assert -flatten gold gate miter
+stat
+sat -verify -prove-asserts -show-public -set-at 1 in_reset 1 -seq 20 -prove-skip 1 miter
+
+design -load postopt # load the post-opt design (otherwise equiv_opt loads the pre-opt design)
+cd fsm # Constrain all select calls below inside the top module
+
+select -assert-count 1 t:CC_BUFG
+select -assert-count 6 t:CC_DFF
+select -assert-count 4 t:CC_LUT2
```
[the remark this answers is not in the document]
pu-cc: Just applied the new practice to all `CC_LUT1..4` in 769955d

techlibs/gatemate/synth_gatemate.cc
```diff
@@ -0,0 +1,393 @@
...
+
+		if (check_label("map_cells"))
+		{
+			run("techmap -map +/gatemate/lut_map.v");
+			run("clean");
+		}
+
+		if (check_label("map_bufg", "(skip if '-noclkbuf')"))
+		{
+			if (!noclkbuf) {
```
Ravenslofty, line 348: Style nit:
```suggestion
		if (check_label("map_bufg", "(skip if '-noclkbuf')") && !noclkbuf)
```

techlibs/gatemate/synth_gatemate.cc
```diff
@@ -0,0 +1,393 @@
...
+		}
+
+		if (check_label("json"))
+		{
+			if (!json_file.empty() || help_mode) {
+				run(stringf("write_json %s", help_mode ? "<file-name>" : json_file.c_str()));
+			}
+		}
+
+		if (check_label("blif"))
```
Ravenslofty, line 384: Same as for EDIF: do you need this? Sure, VPR uses BLIF input, but do you have a VPR toolchain?

techlibs/gatemate/synth_gatemate.cc
```diff
@@ -0,0 +1,393 @@
...
+		log("        do not use CC_ADDF full adder cells in output netlist.\n");
+		log("\n");
+		log("    -nomult\n");
+		log("        do not use CC_MULT multiplier cells in output netlist.\n");
+		log("\n");
+		log("    -nomx8, -nomx4\n");
+		log("        do not use CC_MX{8,4} multiplexer cells in output netlist.\n");
+		log("\n");;
+		log("    -dff\n");
+		log("        run 'abc'/'abc9' with -dff option\n");
```
Ravenslofty, line 80: Since you don't run `abc9` ever, might as well delete it from this help message to avoid confusion.

techlibs/gatemate/gatemate_bramopt.cc
clairexen: SGTM

techlibs/gatemate/brams_map.v
```diff
@@ -0,0 +1,532 @@
...
+	) _TECHMAP_REPLACE_ (
+		.A_DO(B1DATA[39:0]),
+		.B_DO(B1DATA[79:40]),
+		.A_ECC_1B_ERR(A_ECC_1B_ERR),
+		.B_ECC_1B_ERR(B_ECC_1B_ERR),
+		.A_ECC_2B_ERR(A_ECC_2B_ERR),
+		.B_ECC_2B_ERR(B_ECC_2B_ERR),
+		.A_CLK(CLK2),
+		.B_CLK(CLK3),
+		.A_EN(|A1EN),
```
mwkmwkmwk, line 176: the 20K version above has `1'b1` here instead, any reason for the mismatch?

techlibs/gatemate/cells_sim.v
```diff
@@ -0,0 +1,1423 @@
...
+	wire [15:0] addra;
+	wire [15:0] addrb;
+	reg  [19:0] A_DO_out = 0, A_DO_reg = 0;
+	reg  [19:0] B_DO_out = 0, B_DO_reg = 0;
+
+	generate
+		if (RAM_MODE == "SDP") begin
+			// Port A (write)
+			if (A_WR_WIDTH <= 1) begin
+				assign addra = A_ADDR[15:7] + (A_ADDR[15:7]/4);
```
mwkmwkmwk, line 670: this looks wrong for all widths other than 40 (the correct version should be pretty similar to TDP case, I think)

same for port B and 40K version

mwkmwkmwk:
I think this is ready to be merged once the last two comments are addressed; I'm a little uneasy about lack of cascade simulation support while it's used in the mapping rules, but if there are plans to add it later, so be it.

techlibs/gatemate/iob_map.v
```diff
@@ -0,0 +1,58 @@
...
+ *  WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ *  MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ *  ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ *  WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ *  ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ *  OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *
+ */
+
+module \$__toutpad (input A, input OE, output O);
```
mwkmwkmwk, line 20: yosys has just gained (in #3073) native support for negative-polarity output enable, making this whole file no longer necessary — just change the `iopadmap` invocation to directly pass `CC_TOBUF` and `CC_IOBUF` as the target cells, with `~T` given as the output enable

pu-cc: With `~T` given as output enable in `iopadmap`, automatic IO buffer insertion seems to have an issue.

I guess this shouldn't be discussed in this PR, but consider the following example:
```
module iobuf_mwe(input do, input oe_n, output di, inout io);

	assign io = (~oe_n) ? do : 1'bz;
	assign di = io;

endmodule
```
Pre-#3073 `iopadmap` maps a bidrectional buffer as expected, e.g. with `synth_nexus` :
```
$ yosys -p 'read_verilog iobuf_mwe.v; synth_nexus -top iobuf_mwe'
```
Post-#3073 `iopadmap`, however, has an issue finding `~T`:
```
ERROR: Module `BB' referenced in module `iobuf_mwe' in cell `io_BB_B' does not have a port named '~T'.
```

mwkmwkmwk: ... thanks for the report, it seems I made a dumb typo in the PR that our tests didn't catch

mwkmwkmwk: Fixed in #3076, sorry for the mess.

techlibs/gatemate/brams_map.v
pu-cc: my mistake, fixed in 1f214092584b3f110e7a3a90bc39e51b8d5ccf74

techlibs/gatemate/cells_sim.v
pu-cc: the only _useful_ widths for SDP_20K and SDP_40K are 40 or 80 bits -- I decided to drop the other widths

techlibs/gatemate/iob_map.v
pu-cc: thank you -- just changed the `iopadmap` invocation and merged master into this branch to let the tests pass

techlibs/gatemate/cells_sim.v
```diff
@@ -1065,6 +1035,14 @@ module CC_BRAM_40K (
 			$display("ERROR: Port B width of 80 bits is only supported in SDP mode.");
 			$finish();
 		end
+		if (((CAS == "UPPER") || (CAS == "LOWER")) && (WIDTH_MODE_A > 1)) begin
```
mwkmwkmwk, line 1038: this is redundant with an already-existing check a few lines below

techlibs/gatemate/iob_map.v
pu-cc: totally forgot to remove `techmap iob_map.v` , sorry. test should pass now.

techlibs/gatemate/brams_map.v
```diff
@@ -462,7 +450,7 @@ module \$__CC_BRAM_CASCADE (CLK2, CLK3, A1ADDR, A1DATA, A1EN, B1ADDR, B1DATA, B1
 			`undef INIT_UPPER
 			.LOC("UNPLACED"),
 			.CAS("UPPER"),
-			.A_RD_WIDTH(0), .B_RD_WIDTH(CFG_DBITS),
+			.A_RD_WIDTH(CFG_DBITS), .B_RD_WIDTH(0),
```
mwkmwkmwk, line 453: This seems very wrong — the `B` port is used for reading, after all (both here and in the lower cell).

techlibs/gatemate/brams_map.v
```diff
@@ -474,8 +462,8 @@ module \$__CC_BRAM_CASCADE (CLK2, CLK3, A1ADDR, A1DATA, A1EN, B1ADDR, B1DATA, B1
 		) upper_cell (
 			.A_CI(A_CAS),
 			.B_CI(B_CAS),
-			.A_DO(A_UP_DO),
-			.B_DO(B1DATA),
+			.A_DO(B1DATA),
```
mwkmwkmwk, line 465: Also looks wrong

mwkmwkmwk (changes requested):

techlibs/gatemate/brams_map.v
pu-cc: It is actually not _wrong_ since the cells are configured as TDP. I did this to test the effects of NO_CHANGE and WRITE_THROUGH -- I'll change it back to A=write/B=read to keep it consistent with the other {20K,40K}_TDP cells

mwkmwkmwk (approved):

PR #3010, opened 2021-09-17
Fix protobuf backend build dependencies - intermittent build issue due to missing rule

the6p4c:
`backends/protobuf/protobuf.cc` depends on the source and header files generated by `protoc`, but this dependency wasn't explicitly declared. This PR adds a rule to the Makefile to fix intermittent build failures when the protobuf header/source file isn't (due to sheer luck) built before `protobuf.cc`.


lf-:
Repro of the bug being fixed:

on a fresh clone of yosys,

```
sed -i 's/^ENABLE_PROTOBUF := 0$/ENABLE_PROTOBUF := 1/' Makefile
make backends/protobuf/protobuf.o
```

PR #3017, opened 2021-09-27
Add optimization to rtlil back-end for all-x parameter values

clairexen:


mmicko:
Confirming much better results.

PR #3039, opened 2021-10-08
Add support for $aldff flip-flops to verific importer

clairexen:
Add support for `$aldff` flip-flops in verific importer. The new code will not be called unless the following change is applied to `verific.cc`:

```diff
- RuntimeFlags::SetVar("db_infer_set_reset_registers", 1);
+ RuntimeFlags::SetVar("db_infer_set_reset_registers", 0);
```

### DIFF
--- a/frontends/verific/verific.cc
+++ b/frontends/verific/verific.cc
@@ -410,6 +410,20 @@ bool VerificImporter::import_netlist_instance_gates(Instance *inst, RTLIL::IdStr
 		return true;
 	}
 
+	if (inst->Type() == PRIM_DFF)
+	{
+		VerificClocking clocking(this, inst->GetClock());
+		log_assert(clocking.disable_sig == State::S0);
+		log_assert(clocking.body_net == nullptr);
+
+		if (inst->GetAsyncVal()->IsGnd())
+			clocking.addDff(inst_name, net_map_at(inst->GetInput()), net_map_at(inst->GetOutput()));
+		else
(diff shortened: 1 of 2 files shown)


frontends/verific/verific.cc
```diff
@@ -410,6 +410,20 @@ bool VerificImporter::import_netlist_instance_gates(Instance *inst, RTLIL::IdStr
 		return true;
 	}
 
+	if (inst->Type() == PRIM_DFF)
+	{
+		VerificClocking clocking(this, inst->GetClock());
+		log_assert(clocking.disable_sig == State::S0);
+		log_assert(clocking.body_net == nullptr);
+
+		if (inst->GetAsyncVal()->IsGnd())
```
mmicko, line 419: This should be GetAsyncCond since in case of DFF that is 0 and AsyncVal is X

frontends/verific/verific.cc
```diff
@@ -410,6 +410,20 @@ bool VerificImporter::import_netlist_instance_gates(Instance *inst, RTLIL::IdStr
 		return true;
 	}
 
+	if (inst->Type() == PRIM_DFF)
```
mmicko, line 413: This block should be repeated at line 525.
Use next to test:

`yosys -p 'verific -sv dff.sv; verific -import -d dff.vo dff; show'`


```
module dff(input CLK, input D, output reg Q);
	always @(posedge CLK)
		Q <= D;
endmodule

```

PR #3043, opened 2021-10-12
Remove unused wire in procedural for initialization

kamilrakoczy:
This PR removes procedural ``for`` wire after conversion to localparams.

According to standard (``12.7.1 The for-loop``): 
```
The variables used to control a for-loop can also be declared within the loop, as part of the for_initialization
assignments. This creates an implicit begin-end block around the loop, containing declarations of the loop
variables with automatic lifetime. This block creates a new hierarchical scope, making the variables local to
the loop scope.
```
When we are leaving ``for`` scope the wire that was auto-generated for ``for_initialization`` is no longer needed.
This wire causes problems when ``for_initialization`` is inside ``always_comb`` and ``if`` as it  creates latch.

Example error:
```
ERROR: Latch inferred for signal `\top.$fordecl_block$1.i' from always_comb process `\top.$proc$<<EOT:0$3'.
```

### DIFF
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -1981,7 +1981,8 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
 			varbuf->children[0] = buf;
 		}
 
-		if (type == AST_FOR) {
+		if (type == AST_FOR && (init_ast->children[0]->id2ast->str.find("$fordecl_block") == std::string::npos &&
+					init_ast->children[0]->id2ast->str.find("$unnamed_block") == std::string::npos )) {
 			AstNode *buf = next_ast->clone();
 			delete buf->children[1];
 			buf->children[1] = varbuf->children[0]->clone();


frontends/ast/simplify.cc
```diff
@@ -1981,11 +1981,20 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
 			varbuf->children[0] = buf;
 		}
 
-		if (type == AST_FOR) {
+		if (type == AST_FOR && init_ast->children[0]->id2ast->str.find("$fordecl_block") == std::string::npos) {
 			AstNode *buf = next_ast->clone();
 			delete buf->children[1];
 			buf->children[1] = varbuf->children[0]->clone();
 			current_block->children.insert(current_block->children.begin() + current_block_idx++, buf);
+		} else {
```
zachjs, line 1989: I don't understand why we need to remove the variable from `current_scope`. Simply not generating the assignment seems like it should be enough. I double checked, and it seems the given test case does not require this portion of the changes to pass.

kamilrakoczy: It was done, to make sure we won't use this wire later in the code, but as it contains a unique prefix that prevents us from accessing it and opt_clean removes unused wires, it is true that we don't need this change.

I removed it.

PR #3048, opened 2021-10-17
dfflegalize: remove redundant check for initialized dlatch

pda:
This `if (!(supported_dlatch & ~INIT_X))` condition is repeated verbatim on lines 716 and 721, and I can't imagine a legitimate way the inputs could change in between. I imagine it's a copy/paste mistake.

It's not doing any harm, but it's presence is a bit confusing (I was encountering this error and trying to understand why).

Just to make sure I wasn't missing some subtle difference between the two, I asked `diff`, but it exits 0 with no diff:

```sh
diff -u \
    <(sed -n '716,+3p' passes/techmap/dfflegalize.cc) \
    <(sed -n '721,+3p' passes/techmap/dfflegalize.cc)
```

### DIFF
--- a/passes/techmap/dfflegalize.cc
+++ b/passes/techmap/dfflegalize.cc
@@ -718,10 +718,6 @@ flip_dqisr:;
 					goto error;
 				}
 
-				if (!(supported_dlatch & ~INIT_X)) {
-					reason = "initialized dlatch are not supported";
-					goto error;
-				}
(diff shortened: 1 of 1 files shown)


mwkmwkmwk:
Ah, thank you for noticing

This pass is slated for a full rewrite soon anyway, but I'm going to merge this one for now, as it's an obvious improvement.

mwkmwkmwk (approved):

PR #3050, opened 2021-10-19
Fix parameters using struct as a wiretype

kamilrakoczy:
This PR fixes the case when parameter is used with a wiretype of a struct.

Currently, all of the wiretype children are copied to the parameter children (this includes e.g. ``AST_STRUCT_ITEM``). This makes ``param_has_no_default`` (https://github.com/YosysHQ/yosys/blob/master/frontends/ast/ast.cc#L981) check fail as it assumes that simplified parameter can only have at most 2 children.

``AST_STRUCT_ITEM`` is unused as children to AST_PARAMETER, so I changed it to only copy AST_RANGE from wiretype node.

### DIFF
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -1613,6 +1613,7 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
 			if (template_node->type == AST_STRUCT || template_node->type == AST_UNION) {
 				// replace with wire representing the packed structure
 				newNode = make_packed_struct(template_node, str);
+				newNode->attributes[ID::wiretype] = mkconst_str(resolved_type_node->str);
 				// add original input/output attribute to resolved wire
 				newNode->is_input = this->is_input;
 				newNode->is_output = this->is_output;
@@ -1677,6 +1678,7 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
 			if (template_node->type == AST_STRUCT || template_node->type == AST_UNION) {
 				// replace with wire representing the packed structure
 				newNode = make_packed_struct(template_node, str);
+				newNode->attributes[ID::wiretype] = mkconst_str(resolved_type_node->str);
 				newNode->type = type;
 				current_scope[str] = this;
 				// copy param value, it needs to be 1st value
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -1661,18 +1661,32 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
 		if (is_custom_type) {
 			log_assert(children.size() == 2);
 			log_assert(children[1]->type == AST_WIRETYPE);
-			if (!current_scope.count(children[1]->str))
-				log_file_error(filename, location.first_line, "Unknown identifier `%s' used as type name\n", children[1]->str.c_str());
-			AstNode *resolved_type_node = current_scope.at(children[1]->str);
+			auto type_name = children[1]->str;
+			if (!current_scope.count(type_name)) {
+				log_file_error(filename, location.first_line, "Unknown identifier `%s' used as type name\n", type_name.c_str());
+			}
+			AstNode *resolved_type_node = current_scope.at(type_name);
 			if (resolved_type_node->type != AST_TYPEDEF)
-				log_file_error(filename, location.first_line, "`%s' does not name a type\n", children[1]->str.c_str());
+				log_file_error(filename, location.first_line, "`%s' does not name a type\n", type_name.c_str());
 			log_assert(resolved_type_node->children.size() == 1);
 			AstNode *template_node = resolved_type_node->children[0];
-			delete children[1];
-			children.pop_back();
 
 			// Ensure typedef itself is fully simplified
-			while(template_node->simplify(const_fold, at_zero, in_lvalue, stage, width_hint, sign_hint, in_param)) {};
+			while (template_node->simplify(const_fold, at_zero, in_lvalue, stage, width_hint, sign_hint, in_param)) {};
+
+			if (template_node->type == AST_STRUCT || template_node->type == AST_UNION) {
+				// replace with wire representing the packed structure
+				newNode = make_packed_struct(template_node, str);
+				newNode->type = type;
+				current_scope[str] = this;
+				// copy param value, it needs to be 1st value
+				delete children[1];
+				children.pop_back();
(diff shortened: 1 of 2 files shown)


zachjs:
Sorry for the delay in taking another look at this. It seems like the implementation now results in `AST_WIRE`s being created, whereas we would need localparams to enable there use in constants. We could also extend the coverage to include this by adding asserts where the checked values are indirected through localparams.

kamilrakoczy:
@zachjs Thank you for review, but I'm not sure what you mean by ``the implementation now results in AST_WIREs being created``.

Looking at the test from this PR, after simplification I don't see any new ``AST_WIRE``s. Custom type parameter is still parameter:
```
AST_PARAMETER  str='\d' logic basic_prep range=[1:0]                                                                                                                     
        ATTR \wiretype:                                                                                                                                                                                            
          AST_CONSTANT [0x563d19c71da0] str='\p::struct_t' bits='010111000111000000111010001110100111001101110100011100100111010101100011011101000101111101110100'(96) basic_prep range=[95:0] int=
1668570996                                                                                               
        AST_CONSTANT  [0x563d19c71830] bits='10'(2) basic_prep range=[1:0] int=2                                                                                                                
        AST_RANGE basic_prep range=[1:0]                                                                                                                                          
          AST_CONSTANT  [0x563d19c71af0] bits='00000000000000000000000000000001'(32) signed basic_prep range=[31:0] int=1
          AST_CONSTANT  [0x563d19c71c30] bits='00000000000000000000000000000000'(32) signed basic_prep range=[31:0]
```
and access to the fields of this parameter results in ``AST_CONSTANT`` with correct value, e.g ``d.a``:
```
          AST_ASSIGN_LE  [0x563d19c68690] basic_prep                               
            AST_IDENTIFIER  [0x563d19c682b0 -> 0x563d19c67a20] str='\e' basic_prep                                                                                                                  
            AST_CONSTANT  [0x563d19c56020] bits='1'(1) basic_prep range=[0:0] int=1
 ```
 
 Could you point me where this ``AST_WIRE``s are created?

File: tests/various/param_struct.ys [hunk 1]
```diff
@@ -0,0 +1,25 @@
+read_verilog -sv -dump_ast1 -dump_ast2 << EOF
```
zachjs, line 1: Let's drop the -dump flags.

frontends/ast/simplify.cc
```diff
@@ -1461,18 +1462,33 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
...
 			log_assert(resolved_type_node->children.size() == 1);
 			AstNode *template_node = resolved_type_node->children[0];
-			delete children[1];
-			children.pop_back();
 
 			// Ensure typedef itself is fully simplified
-			while(template_node->simplify(const_fold, at_zero, in_lvalue, stage, width_hint, sign_hint, in_param)) {};
+			while (template_node->simplify(const_fold, at_zero, in_lvalue, stage, width_hint, sign_hint, in_param)) {};
+
+			if (template_node->type == AST_STRUCT || template_node->type == AST_UNION) {
```
zachjs, line 1478: If it's not too much work, it might be nice to get some basic coverage in the below test case for A) parameters of nested structs (which seem pretty common); and B) `parameter` of `union packed`, given this code implies support.

tests/various/param_struct.ys
```diff
@@ -0,0 +1,25 @@
...
+} struct_t;
+
+parameter struct_t c = {1'b1, 1'b0};
+endpackage
+
+module dut ();
+parameter p::struct_t d = p::c;
+
+logic e = d.a;
+logic f = d.b;
```
zachjs, line 15: To have this test verify that `d.a`, etc., can be used in constant expressions, let's change `logic` to `localparam`.

zachjs (commented):
I misread the surrounding code in my initial review. This seems to work as expected. I just have some minor requests before we merge.

tests/various/param_struct.ys
kamilrakoczy: Done.

frontends/ast/simplify.cc
kamilrakoczy: I've added tests for nested structs as parameters and union packed.

tests/various/param_struct.ys
kamilrakoczy: Done.

kamilrakoczy:
@zachjs Thank you for the review. 
I fixed your comments and also added support for accessing whole nested structures.

frontends/ast/genrtlil.cc
```diff
@@ -849,6 +849,8 @@ void AstNode::detectSignWidthWorker(int &width_hint, bool &sign_hint, bool *foun
 				while (id_ast->simplify(true, false, false, 1, -1, false, true)) { }
 			if (id_ast->children[0]->type == AST_CONSTANT)
 				this_width = id_ast->children[0]->bits.size();
+			else if (id_ast->children.size() == 1 && id_ast->children[0]->type == AST_IDENTIFIER && id_ast->children[0]->id2ast)
+				this_width = id_ast->children[0]->id2ast->range_left - id_ast->children[0]->id2ast->range_right + 1;
```
zachjs, line 853: It seems like is a partial step in the direction of supporting indirecting structs through localparams, but it only resolves width detection and not the rest of simplification.

If we'd like to finish supporting this case in a separate PR, let's drop this (partial) change for now.

Once we get around to implementing it, I do have a few comments. First, it seems like `id_ast->type` should be restricted to parameters or localparams in this branch. Second, it would arguably be necessary to check if the target is a struct, etc. Finally, it might be necessary to look up the target iteratively, to support, e.g.,:
```SystemVerilog
localparam foo_t P = ...;
localparam A = P.w;
localparam B = A.x.y;
localparam C = B.z;
```

tests/verilog/struct_access.sv
```diff
@@ -77,9 +77,8 @@ module top;
         `CHECK(s.y.a, 1, 0)
         `CHECK(s.y.b, 1, 1)
 
-        // TODO(zachjs): support access to whole sub-structs and unions
-        // `CHECK(s.x, 2, 0)
-        // `CHECK(s.y, 2, 1)
+        `CHECK(s.x, 2, 0)
```
zachjs, line 80: Thanks for taking care of this!

frontends/ast/genrtlil.cc
kamilrakoczy: Ok, lets finish supporting whole sub-structs in another PR.
I've dropped support for this in this PR and I will open another one for this.

zachjs (approved):
Thank you!

PR #3052, opened 2021-10-20
genrtlil: Fix displaying debug info in packages

kamilrakoczy:
Discovered while working on https://github.com/YosysHQ/yosys/pull/3050

When packages are processed, they don't have current_ast_mod set and when some error occurs while processing package, it results in segmentation fault.

This PR adds check if current_ast_mod is set and if so, it displays whole module AST, if not it only displays AST of current node.

### DIFF
--- a/frontends/ast/genrtlil.cc
+++ b/frontends/ast/genrtlil.cc
@@ -1126,8 +1126,9 @@ void AstNode::detectSignWidthWorker(int &width_hint, bool &sign_hint, bool *foun
 
 	// everything should have been handled above -> print error if not.
 	default:
+		AstNode *current_scope_ast = current_ast_mod == nullptr ? current_ast : current_ast_mod;
 		for (auto f : log_files)
-			current_ast_mod->dumpAst(f, "verilog-ast> ");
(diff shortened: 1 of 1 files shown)


frontends/ast/genrtlil.cc
```diff
@@ -1126,8 +1126,14 @@ void AstNode::detectSignWidthWorker(int &width_hint, bool &sign_hint, bool *foun
 
 	// everything should have been handled above -> print error if not.
 	default:
-		for (auto f : log_files)
-			current_ast_mod->dumpAst(f, "verilog-ast> ");
+		// current_ast_mod is not set in e.g. packages
+		if (current_ast_mod) {
```
zachjs, line 1130: This is a great idea. Thanks! In other places, we have used the following to grab the AST when processing a package.
```C++
AstNode *current_scope_ast = current_ast_mod == nullptr ? current_ast : current_ast_mod
```
This could simplify the code a bit, and hopefully provide additional useful information.

kamilrakoczy: Done.

zachjs (approved):
Thanks!

PR #3057, opened 2021-10-21
Initial Verific impoter support for {PRIM,WIDE_OPER}_DLATCH{,RS}

clairexen:


frontends/verific/verific.cc
```diff
@@ -441,7 +441,11 @@ bool VerificImporter::import_netlist_instance_gates(Instance *inst, RTLIL::IdStr
 		return true;
 	}
 
-	// FIXME: PRIM_DLATCH
+	if (inst->Type() == PRIM_DLATCH)
```
mmicko, line 444: This block completely ignores async cond and val, so in end result S inputs do not affect latch which is wrong. When same code as in lines 575-586 is used it gives expected result ( also it matches results as when I put "db_infer_set_reset_registers" back to 1)

frontends/verific/verific.cc
```diff
@@ -872,7 +899,31 @@ bool VerificImporter::import_netlist_instance_cells(Instance *inst, RTLIL::IdStr
...
+	if (inst->Type() == OPER_WIDE_DLATCH)
+	{
+		RTLIL::SigSpec sig_d = IN;
+		RTLIL::SigSpec sig_q = OUT;
+		RTLIL::SigSpec sig_adata = IN1;
+		RTLIL::SigSpec sig_acond = IN2;
+
+		if (sig_acond.is_fully_const() && !sig_acond.as_bool()) {
+			cell = module->addDlatch(inst_name, net_map_at(inst->GetControl()), sig_d, sig_q);
+			import_attributes(cell->attributes, inst);
```
mmicko, line 911: nitpick, both import_attributes can go below if condition, but it is not a problem, I see it is same style as in OPER_WIDE_DFF

frontends/verific/verific.cc
clairexen: Thanks! TFW You realize you spent 30 minutes staring at the wrong code last night..  :D  
Fixed in 90b440f.

mmicko (approved):
LGTM

PR #3059, opened 2021-10-23
sv: remove certain unnecessary loop var ending value assignments

zachjs:


tgorochowik:
@zachjs is there something we can do to help get this is? This resolves the issue we tried to fix in #3043 so it'd be nice to get this merged

zachjs:
I discussed this with Claire earlier, and she suggested an alternative strategy which would be more robust. I'm looking to finalize the implementation by the end of the week.

zachjs:
This has been supplanted by #3150.

PR #3067, opened 2021-11-01
Update the Linux and macOS CI jobs

lethalbit:
This PR updates the Linux and macOS CI jobs to include more compiler, os, and C++ versions. 

In order to facilitate building for other C++ versions, a new Makefile variable `CXXSTD` was introduced which controls the C++ version.

For Linux the following compilers have been added:
 * clang-11
 * clang-12
 * gcc-11
 * gcc-10

For macOS the following compilers have been added:
 * gcc-10
 * gcc-latest

For both platforms all compilers are tested with C++11, 14, and 17 support, with C++20 also added on the Linux CI side.

There have been some minor issues found with the build system during this, mainly the fact the build system relies on using the C frontend for C++ erroneously, causing issues with building ABC when trying to specify the C++ compiler frontend with the `CXX` Makefile override.

There are also issues that have been found by the new CI workflows when trying to build in C++17 mode on macOS.

A windows-mingw CI set is to follow this PR soon:tm:


mmicko (approved):
This all looks good to me. It even exposes real issue in yosys.h YS_MAYBE_UNUSED macro definition. Was waiting for mingw, but that will anyway be in different file when finished, so we can merge this and fix issue in yosys.h

PR #3069, opened 2021-11-02
flatten: Keep sigmap around between flatten_cell invocations.

### LINKED ISSUE #3064: Flatten much slower compared to older yosys

I'm working with big designs with millions of cells and the flatten command is very slow with current Yosys.
But I have stored older yosys binary which I use only for flattening. The difference can be hours vs seconds.
I wanted to try if I can make a simple case to demonstrate this. I don't know when it has gotten slower.

The attached python script will generate a simple design with big number of instantiated buffer modules (just wire inside) chained together.

bufCount variable in python script defines how many buffers are created.

Flattening time behaves roughly like this when number of buffers is increased :
20k 90sec
30k 272sec
40k 496sec
50k 743sec
60k 1040sec

For comparison, even the 60k buffers are flattened in about one second with old yosys.

[gen.py.zip](https://github.com/YosysHQ/yosys/files/7427877/gen.py.zip)

Run the attached gen.py python script, it will generate issue.il file
Then do : yosys -p 'read_ilang issue.il; flatten'

## Expected behavior

Yosys 0.6+     243 (git sha1 fa1c7cc, clang 7.3.0 -fPIC -Os)
Time spent: 81% 1x flatten (1 sec), 18% 2x read_ilang (0 sec)

## Actual behavior

Yosys 0.10+51 (git sha1 19c2d6e15, clang 13.0.0 -fPIC -Os)
Time spent: 99% 1x flatten (86 sec), 0% 2x read_ilang (0 sec)

mwkmwkmwk:
Fixes #3064.

### DIFF
--- a/passes/techmap/flatten.cc
+++ b/passes/techmap/flatten.cc
@@ -77,7 +77,7 @@ struct FlattenWorker
 {
 	bool ignore_wb = false;
 
-	void flatten_cell(RTLIL::Design *design, RTLIL::Module *module, RTLIL::Cell *cell, RTLIL::Module *tpl, std::vector<RTLIL::Cell*> &new_cells)
+	void flatten_cell(RTLIL::Design *design, RTLIL::Module *module, RTLIL::Cell *cell, RTLIL::Module *tpl, SigMap &sigmap, std::vector<RTLIL::Cell*> &new_cells)
 	{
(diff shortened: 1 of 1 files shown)


whitequark:
Thank you!

PR #3070, opened 2021-11-05
WIP: machxo2: Make sure tristates are mapped to active-low enable before

cr1901:
techmapping to machxo2 cells.

It appears I missed the cutoff for `0.11` :D! Oh well...

MachXO2 tristates are active low, which `iopadmap` does not support directly. My techmap to create active-low tristates runs after ABC has created `$luts`. The end result is that designs using tristates would have a `$not` cell that wasn't properly mapped to a `LUT4`. Since `nextpnr` can't map `$not` cells on `machxo2`, this ends up being a problem creating bitstreams with tristates. These changes fix this problem so that bitstreams with tristates can be generated.

This is a WIP because the `arch/machxo2/tribuf.ys` test passes without any changes, and still produces a `$not` gate as output. I wouldn't expect this, but according to [tribuf.log](https://github.com/YosysHQ/yosys/files/7487985/tribuf.log):

```
5.2.17.1. Extracting gate netlist of module `\tristate' to `<abc-temp-dir>/input.blif'..
Extracted 0 gates and 0 wires to a netlist network with 0 inputs and 0 outputs.
Don't call ABC as there is nothing to map.
Removing temp directory.
Removed 0 unused cells and 4 unused wires.
```

`yosys` decides not to call ABC at all, and I don't understand why `yosys` thinks there is nothing to map. This is what the circuit looks like _just_ before the `map_luts` label of the `synth_machxo2` pass:

show

And this is what it looks like after ABC has run `map_luts:map_cells`:
show

Should `yosys` be adding a LUT here, or did I do something specifically wrong with mapping to active-low tristates using an intermediate `techmap`?

### DIFF
--- a/techlibs/machxo2/cells_map.v
+++ b/techlibs/machxo2/cells_map.v
@@ -30,5 +30,13 @@ module  \$_DFF_P_ (input D, C, output Q); FACADE_FF #(.CEMUX("1"), .CLKMUX("CLK"
 // IO- "$__" cells for the iopadmap pass.
 module  \$__FACADE_OUTPAD (input I, output O); FACADE_IO #(.DIR("OUTPUT")) _TECHMAP_REPLACE_ (.PAD(O), .I(I), .T(1'b0)); endmodule
 module  \$__FACADE_INPAD (input I, output O); FACADE_IO #(.DIR("INPUT")) _TECHMAP_REPLACE_ (.PAD(I), .O(O)); endmodule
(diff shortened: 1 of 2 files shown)


mwkmwkmwk:
Please do not merge this one — we're discussing a better solution

cr1901:
Closed in favor of #3073. If there's any issues I have left, it'll be on the `nextpnr-machxo2` (i.e. my) side.

PR #3077, opened 2021-11-10
Add genlib support to ABC command

clairexen:


passes/techmap/abc.cc
```diff
@@ -1343,8 +1347,13 @@ struct AbcPass : public Pass {
 		log("        generate netlists for the specified cell library (using the liberty\n");
 		log("        file format).\n");
 		log("\n");
+		log("    -genlib <file>\n");
+		log("        generate netlists for the specified cell library (using the SIS Genib\n");
```
gatecat, line 1351: nitpick: spelling

passes/techmap/abc.cc
```diff
@@ -655,8 +655,9 @@ struct abc_output_filter
 };
 
 void abc_module(RTLIL::Design *design, RTLIL::Module *current_module, std::string script_file, std::string exe_file,
```
gatecat, line 657: it's not specifically for this PR but I can't help but think a struct of configuration options would be neater than this monstrous set of function arguments at this point

clairexen: yes, that's a mood. each time I add something to it I feel that  :D

gatecat (approved):

PR #3082, opened 2021-11-18
Simplify case before trying to detect sign/width

kamilrakoczy:
This PR changes order of detecting sign/width and simplify in case.

This fixes the ``case`` where case uses items that requires simplify before trying to detect sign/width, e.g. ``$bits`` call.

### DIFF
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -1389,10 +1389,8 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
 
 	if (const_fold && type == AST_CASE)
 	{
-		int width_hint;
-		bool sign_hint;
-		detectSignWidth(width_hint, sign_hint);
 		while (children[0]->simplify(const_fold, at_zero, in_lvalue, stage, width_hint, sign_hint, in_param)) { }
+		detectSignWidth(width_hint, sign_hint);
 		if (children[0]->type == AST_CONSTANT && children[0]->bits_only_01()) {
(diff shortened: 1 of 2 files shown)


frontends/ast/simplify.cc
```diff
@@ -1389,10 +1389,8 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
 
 	if (const_fold && type == AST_CASE)
 	{
-		int width_hint;
-		bool sign_hint;
-		detectSignWidth(width_hint, sign_hint);
 		while (children[0]->simplify(const_fold, at_zero, in_lvalue, stage, width_hint, sign_hint, in_param)) { }
+		detectSignWidth(width_hint, sign_hint);
```
zachjs, line 1393: Thank you very much for spotting this issue and submitting a PR. Sorry for the delay!

Indeed `width_hint` and `sign_hint` should not have been shadowed here. My mistake!

The ordering of the `detectSignWidth` and the first `simplify` is intentional and swapping it here actually breaks things! Unfortunately, the tests I added didn't appropriately cover this scenario. For example, in the below test case, the spec requires that the case expression is evaluated with the maximum of its own width and that of the case item expressions, meaning that `1'b1 << 1` actually gets simplified to `2'b10` rather than `2'b00`.
```SystemVerilog
module top(
    output logic [5:0] out
);
always_comb
    case (1'b1 << 1)
        2'b10: out = '1;
        default: out = '0;
    endcase
endmodule
```

I have an alternative fix that I'll submit in a PR within the next day or so.

zachjs:
This has been superseded by #3116.

PR #3083, opened 2021-11-18
adding new pass `metadata` for generating design metadata

lethalbit:
As part of some work on SBY this pass is to extract design metadata for ingestion into other tools.

It is not a full netlist output, and as we did not want to consume the final design netlist JSON this pass was added.

**NOTE:** This pass is still a big WIP

The rough outline of the JSON object the pass will generate is as follows:

```json
{
  "version": "0.0.0",

  // e.g. if “formal” is here, then the dump contains assertions, assumptions, etc
  "features": ["formal", "connections"],   // optional if complete dump

  "modules": [
    {
      "name": <module_name>,
      "cell_sorts": [
        [
          {
            "type": <cell_type_or_module_name>,
            "ports": [
              {
                "name": <port_name>,
                "direction": <port_direction>,
                "range": [<msb>, <lsb>]  // (optional, default is [0,0])
              },
            ],
            "cells": [
              {
                "name": <cell_name>, // (optional)
                "attributes": {
                  <attr_name>: <val>,
                },
                "parameters": { // (optional, must not overwrite entries from sort parameters)
                  <param_name>: <val>,
                },
              },
              // cell indices that are out of bounds refer to cells without name or attributes
            ],
            "parameters": { // (optional)
            	<param_name>: <val>,
            },
          },
        ],
      ],
      "connections": [
        {
          "ports": [
            {
              "cell": [<cell_sort_idx>, <cell_idx>],
              "port": <port_idx>,
              "range": [<msb>, <lsb>]   // (optional)
            },
          ],
          "labels": [
            {
              "name": <netname>,
              "range": [<msb>, <lsb>]   // (optional)
            },
          ],
        },
      ],
    },
  ],
}
```

### DIFF
--- a/backends/metadata/metadata.cc
+++ b/backends/metadata/metadata.cc
@@ -172,6 +172,21 @@ struct MetadataWriter
         f << "          ]\n";
         f << "        }";
     }
+
+    void write_param_val(const Const& v)
+    {
+        if ((v.flags & RTLIL::ConstFlags::CONST_FLAG_STRING) == RTLIL::ConstFlags::CONST_FLAG_STRING) {
+            const auto str = v.decode_string();
+
+
+            f << get_string(str);
+        } else if ((v.flags & RTLIL::ConstFlags::CONST_FLAG_SIGNED) == RTLIL::ConstFlags::CONST_FLAG_SIGNED) {
+            f << stringf("\"%dsd %d\"", v.size(), v.as_int());
+        } else if ((v.flags & RTLIL::ConstFlags::CONST_FLAG_REAL) == RTLIL::ConstFlags::CONST_FLAG_REAL) {
+
+        } else {
+            f << get_string(v.as_string());
+        }
     }
 
     void write_cell(Cell* cell)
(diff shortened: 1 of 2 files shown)


gatecat:
> It is not a full netlist output, and as we did not want to consume the final design netlist JSON this pass was added.

I think this does give you enough information to reconstruct a netlist in its complete form, though, unless I'm missing something (and the 'cell sort' approach may well end up more compact than the existing JSON output for many post-synthesis designs)

lethalbit:
> > It is not a full netlist output, and as we did not want to consume the final design netlist JSON this pass was added.
> 
> I think this does give you enough information to reconstruct a netlist in its complete form, though, unless I'm missing something (and the 'cell sort' approach may well end up more compact than the existing JSON output for many post-synthesis designs)

I'm not sure there will be enough information in the end to reconstruct the whole netlist for every design with this output or not to be honest, but seeing as it's still very much in flux that could change, seeing as currently only the information needed for the upcoming sby changes are being exported.

Im not super clear if this would end up replacing the current JSON output or not either, as having two JSON netlist export passes seems silly, but seeing as this is still a WIP draft that can be worked out in due time i think.

backends/metadata/metadata.cc
```diff
@@ -0,0 +1,306 @@
...
+USING_YOSYS_NAMESPACE
+PRIVATE_NAMESPACE_BEGIN
+
+
+struct MetadataWriter
+{
+    private:
+        std::ostream &f;
+        bool _use_selection;
+        std::unordered_map<std::string, std::vector<Cell*>> _cells{};
```
clairexen, line 41: `dict<pair<IdString,dict<IdString,pair<int,direction>, std::vector<Cell*>>`

whitequark:
What specific issues with `write_json` does this actually solve?

lethalbit:
**NOTE:** This has been moved to #3158 due to branch rename

PR #3087, opened 2021-11-25
complete support for microsemi smartfusion2 and igloo2bar

tgingold-cern:
This PR adds:
* BRAM cells description and mapping
* $alu support
* SYSRESET and XTLOSC cells
* fix in synth_sf2 so that yosys could be used in the LiberoSoc flow.

Only files in techlibs/sf2 are modified.


Ravenslofty (commented):
I'm not all that comfortable with the `-ffinit` option here as it seems like a way of silently breaking designs; it seems strange though that the hardware cannot initialise its flops.

tgingold-cern:
smartfusion2 is flash based which could explain there is no initial ff state.

The flag is not set by default, so it's safe.
And it is not very silent: yosys reports all the unhanded init attribute.

Do you have a better suggestion ?

mwkmwkmwk:
Having an option to discard FF initialization is fine of itself, as long as it's not the default.  I'd recommend using a more descriptive name though, like `-discard-ffinit` or `-ignore-ffinit`.

Also, the current way is not quite correct: this indeed prevents `dfflegalize` from rejecting the unimplementable FFs, but leaves stray `init` attributes in the netlist (since the `cells_map.v` rules don't remove them).  I'd recommend reverting the `dfflegalize` call to its previous shape and instead running `attrmap -remove init` somewhere early in the flow (right after `proc` should be fine) if the option is set.

techlibs/sf2/cells_sim.v
```diff
@@ -435,10 +459,119 @@ module TRIBUFF_DIFF (
...
+	output POWER_ON_RESET_N);
+endmodule
+
+
+(* blackbox *)
+module XTLOSC (
+	(* iopad_external_pin *)
+	input  XTL,
+	output CLKOUT);
+	parameter [2:0] MODE = 2'h3;
```
mwkmwkmwk, line 482: should be `[1:0]`

tgingold-cern: Sure.  Good catch!

tgingold-cern:
Thank you for your feedback, that's appreciated.
I have renamed the option to -discard-ffinit and implemented using attrmap, as advised.

nakengelhardt:
Would you mind pulling out the bram inference parts into a separate PR? We'll want to implement those with the new `memory_libmap` infrastructure instead, but the rest looks fine to merge.

tgingold-cern:
Ok, I have just updated this PR without the bram inference.  Thanks!

Makefile
```diff
@@ -777,7 +777,7 @@ ifneq ($(ABCREV),default)
 	$(Q) if test -d abc/.hg; then \
 		echo 'REEBE: NOP qverpgbel vf n ut jbexvat pbcl! Erzbir nop/ naq er-eha "znxr".' | tr 'A-Za-z' 'N-ZA-Mn-za-m'; false; \
 	fi
-	$(Q) if test -d abc && test -d abc/.git && ! git -C abc diff-index --quiet HEAD; then \
```
nakengelhardt, line 780: Why this change? The original (with `git -C`) would be preferred for us since it handles errors.

techlibs/sf2/synth_sf2.cc
```diff
@@ -187,7 +197,7 @@ struct SynthSf2Pass : public ScriptPass
 
 		if (check_label("map_ffs"))
 		{
-			run("dfflegalize -cell $_DFFE_PN?P_ x -cell $_SDFFCE_PN?P_ x -cell $_DLATCH_PN?_ x");
+                        run("dfflegalize -cell $_DFFE_PN?P_ x -cell $_SDFFCE_PN?P_ x -cell $_DLATCH_PN?_ x");
```
nakengelhardt, line 200: Can you change this back to tabs?

nakengelhardt (changes requested):

tgingold-cern:
Oops, those two changes were not expected to be on this branch.  Now fixed.
Thanks!

mmicko:
Thanks @tgingold-cern will merge as soon as CI checks are done

PR #3089, opened 2021-11-25
Add read_liberty -wb

gatecat:
This enables the modules defined in a liberty file to be read as whiteboxes; similar to how we read in FPGA cell libraries.

PR #3091, opened 2021-11-27
[RFC] Add recover_names pass to recover names post-mapping 

gatecat:
This still needs some more tidying up and testing (I'm not convinced it's recovering as many names as it should in a more complex design although I still need to investigate); but it's an experiment with using a combination of random-value combinational simulation to find equivalence classes; followed by SAT; to find name pairs between a pre- and post- mapping design; in order to recover names lost during synthesis/mapping passes even if they perform significant rewriting.

It's based on abc's `dress` but integrated in Yosys to be more reliable and useful for abc9 (in theory, I haven't tried) or non-abc flows as well. Currently it's only been tested with ASIC flows, and consequently includes #3089 for this purpose.

I'd appreciate some feedback on what's implemented so far; in particular whether `flatten -wb` is a reasonable approach to be able perform evaluation and SAT on a netlist with boxes (which does add some complexity in tracking bits between the original and flattened versions of gate and gold); or whether it would be preferable to keep boxes intact (which then necessitates creating derived versions of parameterised boxes, etc).

I also need to figure out a reasonable way to add some tests for this...


eddiehung:
> or whether it would be preferable to keep boxes intact (which then necessitates creating derived versions of parameterised boxes, etc).

`abc9` necessarily does this, by the way.

maliberty:
This would be super useful for OpenROAD so I hope it can be merged.

gatecat:
Sorry, I've had a bit of a backlog recently - but I do definitely plan to get back to this!

gatecat:
Would appreciate a review of this if someone has time?

clairexen (approved):
I had a quick look and LGTM.

PR #3103, opened 2021-12-11
write_verilog: dump zero width sigspecs correctly

whitequark:
Before this commit, zero width sigspecs were dumped as "" (empty
string). Unfortunately, 1364-2005 5.2.3.3 indicates that an empty
string is equivalent to "\0", and is 8 bits wide, so that's wrong.

After this commit, a replication operation with a count of zero is
used instead, which is explicitly permitted per 1364-2005 5.1.14,
and is defined to have size zero. (Its operand has to have a non-zero
size for it to be legal, though.)

PR #1203 has addressed this issue before, but in an incomplete way.

### DIFF
--- a/backends/verilog/verilog_backend.cc
+++ b/backends/verilog/verilog_backend.cc
@@ -358,7 +358,8 @@ void dump_sigchunk(std::ostream &f, const RTLIL::SigChunk &chunk, bool no_decima
 void dump_sigspec(std::ostream &f, const RTLIL::SigSpec &sig)
 {
 	if (GetSize(sig) == 0) {
-		f << "\"\"";
+		// See IEEE 1364-2005 Clause 5.1.14.
+		f << "{0{1'b0}}";
 		return;
 	}
 	if (sig.is_chunk()) {


mwkmwkmwk:
I don't think this is actually permitted.  The full passage from the standard:

> A replication operation may have a replication constant with a value of zero. This is useful in parameterized
code. A replication with a zero replication constant is considered to have a size of zero and is ignored. Such
a replication shall appear only within a concatenation in which at least one of the operands of the
concatenation has a positive size

I think the only correct approach would be to have a pre-processing pass that removes 0-width connections (or, for builtin operators, substitutes a zero constant for source operands); however, if this patch fixes Quartus and doesn't break any other important target, it could still be worth it as a workaround?

mwkmwkmwk:
To be clear about my reading of the above: IMO `A <= {0{B}};` is not allowed, but `A <= {1'b0, {0{B}}}` is fine, since the full concatenation is 1 bit long in the end.  IEEE 1800-2017 11.4.12.1 also contains an example that agrees with this reading.

whitequark:
> however, if this patch fixes Quartus and doesn't break any other important target, it could still be worth it as a workaround?

In practice this fixes issues with Quartus and Yosys (!); before PR #1203, Yosys couldn't parse its own Verilog correctly, and I suspect it still cannot in the cases where the `dump_sigspec` code path gets hit.

mwkmwkmwk:
> > however, if this patch fixes Quartus and doesn't break any other important target, it could still be worth it as a workaround?
> 
> In practice this fixes issues with Quartus and Yosys (!); before PR #1203, Yosys couldn't parse its own Verilog correctly, and I suspect it still cannot in the cases where the `dump_sigspec` code path gets hit.

... okay, let's just merge it, then

PR #3104, opened 2021-12-11
Add clean_zerowidth pass, use it for Verilog output.

mwkmwkmwk:
This should remove instances of zero-width sigspecs in the netlist,
avoiding problems in the Verilog backend with emitting them.

See #3103.


whitequark (approved):
LGTM

PR #3116, opened 2021-12-16
fix width detection of array querying function in case and case item expressions

zachjs:
I also removed the unnecessary shadowing of `width_hint` and `sign_hint` in the corresponding case in `simplify()`.

This fixes the issue raised in #3082, while also handling the (now covered) edge case that requires the existing ordering of `detectSignWidth` and `simplify`.

### DIFF
--- a/tests/simple/case_expr_query.sv
+++ b/tests/simple/case_expr_query.sv
@@ -1,0 +1,32 @@
+module top(
+    output logic [5:0] out
+);
+always_comb begin
+    out = '0;
+    case ($bits (out)) 6:
+    case ($size (out)) 6:
+    case ($high (out)) 5:
+    case ($low  (out)) 0:
+    case ($left (out)) 5:
+    case ($right(out)) 0:
+    case (6) $bits (out):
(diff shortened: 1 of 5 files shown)


tests/simple/case_expr_query.sv
```diff
@@ -0,0 +1,30 @@
...
+    case ($low  (out)) 0:
+    case ($left (out)) 5:
+    case ($right(out)) 0:
+    case (6) $bits (out):
+    case (6) $size (out):
+    case (5) $high (out):
+    case (0) $low  (out):
+    case (5) $left (out):
+    case (0) $right(out):
+        out = '1;
```
clairexen, line 17: should `out` be initialized to `'0` at the start of this `always_comb` block?

clairexen (approved):
lgtm

tests/simple/case_expr_query.sv
zachjs: I don't think it's strictly necessary, but I'll make that change!

PR #3124, opened 2021-12-21
intel_alm: disable 256x40 M10K mode

Ravenslofty:
This BRAM mode uses both address ports of the M10K memory block, making it logically single-port.
Since memory_bram can't presently map to single-port memories, remove it.


mwkmwkmwk (changes requested):
256×32 looks like it needs the same treatment

Ravenslofty:
I took the opportunity to clean up there, leaving only the memory sizes directly implemented in hardware (the x16 is a truncated x20, and the x8 is a truncated x10).

mwkmwkmwk (approved):

Issue #3090, opened 2021-11-26
Wrong sign extension during multiplication


punzik:
Code and testbench: https://gist.github.com/punzik/0d4d5cc3e8ae26bdbeb485467744edad

In this code tests several ways to multiply a signed and unsigned number.
```verilog
    reg signed [11:0] a;
    reg [11:0] b;
    reg signed [31:0] c1;
    reg signed [31:0] c2;
    reg signed [31:0] c3;
    reg signed [31:0] c4;
 
    initial begin
        a = -12'sd20;
        b = 12'hfff;
    end

    logic signed [12:0] bs;
    assign bs = 13'(b);

    always @(posedge clock) begin
        c1 <= a * b;
        c2 <= a * $signed(13'(b));
        c3 <= 32'(a) * b;
        c4 <= a * bs;
    end
```

## Expected behavior

Tested on iverilog 11.0, QuestaSim, XCellium and VCS.

```
c1 = 16691220
c2 = -81900
c3 = -81900
c3 = -81900
```

## Actual behavior

Yosys 0.11:

```
c1 = 16691220
c2 = 20
c3 = 16691220
c3 = -81900
```

PR #3135, opened 2021-12-30
sv: fix size cast clipping expression width

zachjs:
This fixes the `c2` case in #3090. Fixes for `c3` will come separately after further investigation.


clairexen (approved):
LGTM

PR #3136, opened 2021-12-30
fixup verilog doubleslash test

zachjs:
This test was recently introduced in #2982. My original motivation was to prevent running the `verilog` suite from making the working directory dirty. Rather than avoid creating the file, I decided to extend the test to make use of it.

- add generated doubleslash.v to .gitignore
- ensure backend verilog can be read again


clairexen (approved):
LGTM

PR #3138, opened 2021-12-30
Make GIT_REV logic work in release tarballs

DanielG:
Currently GIT_REV doesn't get set properly when building a release
tarball. To fix this we arrange for .gitcommit to contain the (short)
commit hash in tarballs generated with git-archive(1) using export-subst in
gitattributes. This way the correct commit hash is (reproducibly) included
in the release tarballs while not burdening the maintainers with updating
it in the git repo.

Please note this even works on Github and similar forges as they use
git-archive for generating tarballs so this works out quite nicely.


mmicko:
@DanielG tried fixing it for make on ubuntu 18.04 but it just does not work

DanielG:
Hmm, well if the $(file ...) thing doesn't work, there's always $(shell ...) which is just a bit more inefficient :)

I'll have a play around with this on 18.04 and see if I can manage to fix it.

DanielG:
Ah, yes `$(file <...)` got introduced in gnu make 4.2 but 18.04 has only 4.1. I'll just replace it with a shell call then.

mmicko:
Thanks

DanielG:
Can you give this a try?

PR #3142, opened 2022-01-04
logger: fix unmatched expected warnings and errors

zachjs:
- Prevent unmatched expected error patterns from self-matching
- Prevent infinite recursion on unmatched expected warnings
- Always print the error message for unmatched error patterns
- Add test coverage for all unmatched message types
- Add test coverage for excess matched logs and warnings


mmicko (approved):
LGTM, have tried running yosys-tests since those were using it as well and related items pass fine.

PR #3150, opened 2022-01-07
sv: auto add nosync to certain always_comb local vars

zachjs:
If a local variable is always assigned before it is used, then adding nosync prevents latches from being needlessly generated.

This supplants #3059.

@tgorochowik @kamilrakoczy


clairexen (approved):
LGTM

zachjs:
The above force-push rebases and adds a relevant entry to the change log.

PR #3153, opened 2022-01-10
Fix always_comb latch when for is inside function

kamilrakoczy:
This PR fixes latch in ``always_comb`` block when ``for`` is inside function (and this function is used inside ``always_comb`` block.

I'm not sure if this is the correct way to fix this issue, but none of already existing tests are failing with this change.

Related to: https://github.com/YosysHQ/yosys/pull/3150

### DIFF
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -2341,15 +2341,12 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
 	{
 		expand_genblock(str + ".");
 
-		// if this is an autonamed block is in an always_comb
-		if (current_always && current_always->attributes.count(ID::always_comb)
-				&& str.at(0) == '$')
-			// track local variables in this block so we can consider adding
-			// nosync once the block has been fully elaborated
-			for (AstNode *child : children)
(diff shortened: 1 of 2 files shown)


tgorochowik:
Hi @zachjs, would you mind taking a look at this too? This change is pretty small. We'll be happy to rework this if you'd like this to be done differently

frontends/ast/simplify.cc
```diff
@@ -2341,15 +2341,12 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
 	{
 		expand_genblock(str + ".");
 
-		// if this is an autonamed block is in an always_comb
-		if (current_always && current_always->attributes.count(ID::always_comb)
-				&& str.at(0) == '$')
```
zachjs, line 2346: After further investigation, the root cause is that `str.at(0) == '$'` isn't accurate for identifying autonamed blocks. This issue even affects blocks outside of functions. Rather than remove these checks entirely, I'll open a PR to fix them.

kamilrakoczy:
This PR can be closed when https://github.com/YosysHQ/yosys/pull/3215 gets merged.

PR #3158, opened 2022-01-14
Added new jny pass for generating design metadata

lethalbit:
**NOTE**: This was previously #3083, re-creating due to rename from `metadata` to JNY

As part of some work on SBY this pass is to extract design metadata for ingestion into other tools.

It is not a full netlist output, and as we did not want to consume the final design netlist JSON this pass was added.

**NOTE:** This pass is still a big WIP

The rough outline of the JSON object the pass will generate is as follows:

```json
{
  "version": "0.0.0",

  // e.g. if “formal” is here, then the dump contains assertions, assumptions, etc
  "features": ["formal", "connections"],   // optional if complete dump

  "modules": [
    {
      "name": <module_name>,
      "cell_sorts": [
        [
          {
            "type": <cell_type_or_module_name>,
            "ports": [
              {
                "name": <port_name>,
                "direction": <port_direction>,
                "range": [<msb>, <lsb>]  // (optional, default is [0,0])
              },
            ],
            "cells": [
              {
                "name": <cell_name>, // (optional)
                "attributes": {
                  <attr_name>: <val>,
                },
                "parameters": { // (optional, must not overwrite entries from sort parameters)
                  <param_name>: <val>,
                },
              },
              // cell indices that are out of bounds refer to cells without name or attributes
            ],
            "parameters": { // (optional)
            	<param_name>: <val>,
            },
          },
        ],
      ],
      "connections": [
        {
          "ports": [
            {
              "cell": [<cell_sort_idx>, <cell_idx>],
              "port": <port_idx>,
              "range": [<msb>, <lsb>]   // (optional)
            },
          ],
          "labels": [
            {
              "name": <netname>,
              "range": [<msb>, <lsb>]   // (optional)
            },
          ],
        },
      ],
    },
  ],
}
```

### DIFF
--- a/backends/jny/jny.cc
+++ b/backends/jny/jny.cc
@@ -24,6 +24,7 @@
 #include "kernel/cellaigs.h"
 #include "kernel/log.h"
 #include <string>
+#include <algorithm>
 #include <unordered_map>
 #include <vector>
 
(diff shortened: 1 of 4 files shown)


nakengelhardt (changes requested):

backends/jny/jny.cc
```diff
@@ -0,0 +1,554 @@
...
+struct JnyWriter
+{
+    private:
+        std::ostream &f;
+        bool _use_selection;
+
+        // XXX(aki): TODO: this needs to be updated to us
+        // dict<T, V> and then coalesce_cells needs to be updated
+        // but for now for the PoC this looks to be sufficient
+        std::unordered_map<std::string, std::vector<Cell*>> _cells{};
```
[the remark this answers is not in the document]
lethalbit: For now I made `coalesce_cells` ensure the map is cleared before it does its thing.

and [claire did mention](https://github.com/YosysHQ/yosys/pull/3083#discussion_r752416071) a potential fix for that, so i'll look into getting it moved from the `unordered_map` to `dict`. Ideally it should be fairly straight forward

nakengelhardt (approved):

backends/jny/jny.cc
```diff
@@ -0,0 +1,554 @@
...
+        //   |---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|
+        log("\n");
+        log("    jny [options] [selection]\n");
+        log("\n");
+        log("Write a JSON netlist metadata for the current design\n");
+        log("\n");
+        log("    -o <filename>\n");
+        log("        write to the specified file.\n");
+        log("\n");
+        log("    -connections\n");
```
clairexen, line 480: swap the defaults and by default write all information

lethalbit: Switched all the defaults, so now you need to explicitly disable the output of the connections, attributes, and properties

nakengelhardt (approved):

PR #3161, opened 2022-01-17
[RFC] ecp5: map reduce_{and/or} to carry chains

Ravenslofty:
So, this is a neat trick I found while looking through the Altera Synthesis Cookbook: `&a` can be rephrased as the carry out of `a + 1`. We can use the LUT to compress the number of stages by taking advantage of the commutativity of AND.

Reduce-and itself isn't a particularly common operation. However, `alumacc` implements `A != B` as `|(A - B)`, which means reduce-or is a common operation, and through De Morgan's laws, reduce-or can be implemented in terms of reduce-and.

I'm tagging this as RFC, though, because I'm not happy with this yet: perhaps somebody else could shed some light into how to pack the input inverters into the LUT without getting problems caused by the LUT2. (or perhaps the answer will reveal itself while I'm asleep)


Ravenslofty:
This patch works, but I think I can do it more optimally.

Ravenslofty:
(answer: ABC9 attempts to outsmart me anyway, so there's not much point)

Ravenslofty:
Despite passing equivalence in the test this looks to be breaking SERV, and I really do not understand how; until I do I'm going to mark this as draft.

PR #3163, opened 2022-01-18
verilog: fix dynamic dynamic range asgn elab

### LINKED ISSUE #3051: Signed value used for bit select gives unexpected behavior.

`ex.v` source:
```
module top (
    input [1:0] i,
    output reg [3:0] result
);

   wire signed [3:0] idx = {2'b00,i};

   always @(*) begin
      result = 4'b0000;
      result[idx] = 1'b1;
   end

endmodule
```
At the Yosys prompt run the following commands:
```
read_verilog ex.v
synth
show
```
You will see this circuit, in which the high 3 bits of "result" are tied to zero:
bad

## Expected behavior

I expected the 2-bit "i" input to be decoded into one-hot representation in the 4-bit "result" output.

**If I remove the "signed" modifier from `idx` and use the same flow, then I get the decoder that I was expecting**:

good

## Actual behavior

The high 3 bits of "result" are always zero (incorrectly).   The low bit of "result" is asserted only when "i" is 2'b00 (correctly).   Refer to the first diagram.

**Note:** With the original circuit where I think I'm seeing incorrect behavior, the output of `synth_xilinx` is even stranger than the `synth` output shown above.

Origin: https://github.com/SymbiFlow/symbiflow-arch-defs/issues/2307

### LINKED ISSUE #3085: Unexcepted behavior of range in left-hand side expression

RTL code :
```
module top(data, sel, out);
input data;
input [1:0] sel;
output reg [3:1] out;

always @ (*) begin
  out = 0;
  out[sel] = data;
end
endmodule
```
yosys script synth.ys :
```
# read design
read -sv test.v
hierarchy -top top

# the high-level stuff
proc; fsm; opt; memory; opt

# mapping to internal cell library
techmap; opt

# mapping flip-flops to mycells.lib
dfflibmap -liberty ./examples/cmos/cmos_cells.lib

# mapping logic to mycells.lib
abc -liberty ./examples/cmos/cmos_cells.lib

# cleanup
clean

write_verilog -noexpr synth.v
```
The problem is reproducible with:
`$ yosys synth.ys`

## Actual behavior
synth.v :
```
/* Generated by Yosys 0.9+4309 (git sha1 9432400e, gcc 9.3.0-17ubuntu1~20.04 -fPIC -Os) */

(* hdlname = "\\top" *)
(* top =  1  *)
(* src = "test.v:1.1-10.10" *)
module top(data, sel, out);
  wire _0_;
  (* src = "test.v:2.7-2.11" *)
  input data;
  (* src = "test.v:4.18-4.21" *)
  output [3:1] out;
  (* src = "test.v:3.13-3.16" *)
  input [1:0] sel;
  NAND _1_ (
    .A(sel[0]),
    .B(data),
    .Y(_0_)
  );
  NOR _2_ (
    .A(sel[1]),
    .B(_0_),
    .Y(out[1])
  );
  assign out[3:2] = 2'h0;
endmodule
```
As shown in the verilog, output out[3:2] is assigned to zero, which is not as expected. 
And I believe the error is caused by the code that "replace dynamic ranges in left-hand side expressions" in simplify.cc .

zachjs:
Fixes #3085 @ZachGu
Fixes #3051 @tcal-x

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -8,6 +8,8 @@ Yosys 0.14 .. Yosys 0.14-dev
  * Verilog
     - Fixed evaluation of constant functions with variables or arguments with
       reversed dimensions
+    - Fixed elaboration of dynamic range assignments where the vector is
+      reversed or is not zero-indexed
 
 Yosys 0.13 .. Yosys 0.14
 --------------------------
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -2704,6 +2704,18 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
 			while (wire_data->simplify(true, false, false, 1, -1, false, false)) { }
 			current_ast_mod->children.push_back(wire_data);
 
+			int shamt_width_hint = -1;
+			bool shamt_sign_hint = true;
+			shift_expr->detectSignWidth(shamt_width_hint, shamt_sign_hint);
+
+			AstNode *wire_sel = new AstNode(AST_WIRE, new AstNode(AST_RANGE, mkconst_int(shamt_width_hint-1, true), mkconst_int(0, true)));
+			wire_sel->str = stringf("$bitselwrite$sel$%s:%d$%d", filename.c_str(), location.first_line, autoidx++);
+			wire_sel->attributes[ID::nosync] = AstNode::mkconst_int(1, false);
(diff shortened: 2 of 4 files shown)


ZachGu (approved):

clairexen (approved):
LGTM

zachjs:
The recent force-push simply rebased and resolved a merge conflict within the changelog.

PR #3164, opened 2022-01-18
fix dumpAst() compilation warning

zachjs:
Without this change, Apple clang 13.0.0 produces the warning below.
```
frontends/ast/ast.cc:348:23: warning: format specifies type 'int' but the argument has type 'std::__bit_const_reference<std::vector<bool>>' [-Wformat]
                         fprintf(f, " %d", v);
                                      ~~   ^
```


mmicko (approved):
LGTM

PR #3176, opened 2022-01-30
Fix the help message of synth_quicklogic command.

higuoxing:
This PR helps add some missing fields in the reference manual.

Edit:
This PR helps fix the help message of `synth_quicklogic` command.

Before:

```
yosys> help synth_quicklogic
   synth_quicklogic [options]
This command runs synthesis for QuickLogic FPGAs
...
    verilog:

```

After:

```
yosys> help synth_quicklogic
   synth_quicklogic [options]
This command runs synthesis for QuickLogic FPGAs
...
    verilog:
        write_verilog -noattr -nohex <file-name>
```

### DIFF
--- a/techlibs/quicklogic/synth_quicklogic.cc
+++ b/techlibs/quicklogic/synth_quicklogic.cc
@@ -225,8 +225,8 @@ struct SynthQuickLogicPass : public ScriptPass {
 		}
 
 		if (check_label("verilog")) {
-			if (!verilog_file.empty()) {
-				run("write_verilog -noattr -nohex " + verilog_file);
+			if (!verilog_file.empty() || help_mode) {
+				run(stringf("write_verilog -noattr -nohex %s", help_mode ? "<file-name>" : verilog_file.c_str()));
 			}
 		}
 	}


mmicko:
This file is generated out of pass help message. If some text need change or is missing, original helo message needs to be changed

higuoxing:
> This file is generated out of pass help message. If some text need change or is missing, original helo message needs to be changed

Thanks for your quick reply. So ... I have to modify the help message and copy the output of the help message to this manual? Can I reuse this PR instead of creating a new one?

mmicko:
You can, but  make sure not to affect tex file but cc files

higuoxing:
> You can, but make sure not to affect tex file but cc files

Thanks, I've updated my branch according to your comment. But it looks that I don't have permission to re-open this PR.

Xiretza:
The commits should really be squashed down into one, and you need to run `make update-manual` to actually update the TeX files.

higuoxing:
> The commits should really be squashed down into one, and you need to run `make update-manual` to actually update the TeX files.

Thanks for your tips. I just squashed my commits and update the manual. However, I see some irrelevant changes in the manual, is that ok?

Xiretza:
Yeah, someone just forgot to update the manual. I'm working on a CI check that will remind people to do that.

mmicko:
@higuoxing Please do not update tex file, this is what we do on release day

Xiretza:
> @higuoxing Please do not update tex file, this is what we do on release day

Oh, sorry, wasn't aware that that's how it's handled. I assumed that since it's checked into the repo, it should probably match up with the rest of the code.

higuoxing:
> @higuoxing Please do not update tex file, this is what we do on release day

Got it, thank you!

> Oh, sorry, wasn't aware that that's how it's handled. I assumed that since it's checked into the repo, it should probably match up with the rest of the code.

Never mind~ Now, I know how to deal with it in future :-)

PR #3185, opened 2022-02-04
Add co-simulation in sim pass

mmicko:
- support FST as output for sim pass
- "n" parameter now explicitly use clock cycles
- new parameter "r" to use FST as input for co-simulation
- "sim", "sim-cmp", "sim-gate" and "sim-gold" to select co-simulation mode
- "start","stop" or "at" parameter to specify time (default unit is ns, can use anything from zs to s)
- "scope" to specify top scope in input sim file


yuyichao:
This seems to be using zlib even with `ENABLE_ZLIB=0` and broke emcc build.

mmicko:
@yuyichao Thanks for reporting this, FST files library is using zlib directly, so in making this respect ENABLE_ZLIB will make this option not available in emcc build, which is probably fine for most users.

yuyichao:
It is actually possible to build with the `-s USE_ZLIB=1` emcc flag. However, `ENABLE_ZLIB` (and the emcc flags) aren't included by default which is the problem.

If I read the code correctly, isn't zlib just one of the output options? Shouldn't other formats still work even if zlib is disabled?

mmicko:
Yes, zlib is one of the output/input options but problem is that external API used  is not made with ZLIB as optional, so I  have to see about effort of making it optional.

yuyichao:
I see. Yeah, enabling/disabling the whole feature depending on enable_zlib should be fine for now. The user can still manually enable it and supply the right flag to pull zlib to build.

Issue #3195, opened 2022-02-10
Rename flip-flops with custom suffix


bfg86:
I'd like to have meaningful names for flip-flops (makes it easier for debugging). So I tried 
`rename -wire t:$*DFF*`
before dfflibmap. 
This gave names with the flop type as a suffix, for example
`\ready$_DFFE_PP1P_`
I made a patch so its possible to have a custom suffix when running rename -wire. Wondering if anyone else are interested in this patch.
This would make it easier to debug a netlist, and also do name-matching vs RTL in third-party tools (like LEC).

bfg86:
After applying the patch you can do
`rename -wire -suffix _reg t:$*DFF*`
and the same register would be called 
`ready_reg`
in the netlist.

dineshannayya:
Is this option give user have cell/FF in netlist with same RTL ?
My issue Yosys is:  Yosys netlist generation does not follow the industry standard naming conversion in
preserve cell instance name
Example: Currently in netlist cells instance name with number like _06637_
sky130_fd_sc_hd__nand2_2 _06637_ ( .A(_01599_), .B(_01620_), .Y(_01621_));

Preferred format <rtl_cell_name>_reg : This helps in quick gate-level, ECO and LEC and
STA debug.

Is there is example scripts to show which place this  command need to applied .

dineshannayya:
Is is custom script integrated into yosys main branch ? 

 I see tool gives with below error while integrating this patch
ERROR: TCL interpreter returned an error: wrong # args: should be "rename oldName newName"

nakengelhardt:
We don't want to provide anything that looks like a standard because we don't have any way to guarantee consistent naming of these cells in situations more complex than a simple assignment to a FF (e.g., if there are any bit selections, slices, concatenations). There will always be cases where the names would not match what another tool will do, or even what a different combination of yosys passes will create. Therefore no one should build any script relying on those names, and if they are only read by a human then it should not make a difference if it says <wirename>_reg or <wirename>$dff.

dineshannayya:
Current yosys cell naming scheme (instance name with number) is a critical drawback compare to commercial graded tools.
Netlist with cell instance with matching with source RTL name is critical function to debug the LEC failure, STA analysis and Gate level simulation and debug.

bfg86:
[...]

PR #3196, opened 2022-02-11
Add -suffix option to rename -wire

bfg86:
See #3195

### DIFF
--- a/passes/cmds/rename.cc
+++ b/passes/cmds/rename.cc
@@ -66,7 +66,7 @@ static std::string derive_name_from_src(const std::string &src, int counter)
 		return stringf("\\%s$%d", src_base.c_str(), counter);
 }
 
-static IdString derive_name_from_cell_output_wire(const RTLIL::Cell *cell)
+static IdString derive_name_from_cell_output_wire(const RTLIL::Cell *cell, string suffix)
 {
 	// Find output
 	const SigSpec *output = nullptr;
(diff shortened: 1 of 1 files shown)


passes/cmds/rename.cc
```diff
@@ -127,10 +130,12 @@ struct RenamePass : public Pass {
 		log("cells with private names.\n");
 		log("\n");
 		log("\n");
-		log("    rename -wire [selection]\n");
+		log("    rename -wire [selection] [-suffix <suffix>]\n");
 		log("\n");
 		log("Assign auto-generated names based on the wires they drive to all selected\n");
 		log("cells with private names. Ignores cells driving privatly named wires.\n");
+		log("The -suffix option can be used to set a suffix to the renamed cells.\n");
+		log("The default suffix is the cell type.\n");
```
nakengelhardt, line 138: This message is now a little confusing. There is no mention of suffixes in the first two lines so the last line reads like the default is about what happens if you provide the -suffix option with no argument. How about something like the following message instead:

Assign auto-generated names based on the wires they drive to all selected cells with private names. Ignores cells driving privatly named wires. By default, the cell is named after the wire with the cell type as suffix.
The -suffix option can be used to set the suffix to the given string instead.

nakengelhardt (changes requested):

passes/cmds/rename.cc
bfg86: I liked your text better so updated it.

nakengelhardt: Thanks!

nakengelhardt (approved):

bfg86:
@nakengelhardt This is my first PR to a open source project that actually got some traction and went through, so this is really cool for me. Thanks!

PR #3198, opened 2022-02-11
verilog: support for time scale delay values

### LINKED ISSUE #3190: yosys frontend cannot handle delays with a specified unit

After unzipping the attached zip archive:
`yosys -p 'read_verilog case.v'` -> succeeds.
`yosys -p 'read_verilog case_bad.v'` -> fails.

My yosys version:
```
$ yosys -version
Yosys 0.9+4052 (git sha1 0ccc7229, clang 11.0.1-2 -fPIC -Os)
```

## Expected behavior

I would expect that at least in the read_verilog stage, `q <= #1 d` would be treated the same way as `q <= #1ns d`. `read_verilog` should at least pass if I specify a unit on the delay in a nonblocking assignment.

## Actual behavior

`read_verilog` fails because it expects another token. See error message below.

```
Parsing Verilog input from `case_bad.v' to AST representation.
case_bad.v:8: ERROR: syntax error, unexpected TOK_ID
```

zachjs:
Fixes #3190. @ted-xie

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -10,6 +10,7 @@ Yosys 0.14 .. Yosys 0.14-dev
       reversed dimensions
     - Fixed elaboration of dynamic range assignments where the vector is
       reversed or is not zero-indexed
+    - Added frontend support for time scale delay values (e.g., `#1ns`)
 
 Yosys 0.13 .. Yosys 0.14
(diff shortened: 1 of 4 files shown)


clairexen (approved):
LGTM

PR #3203, opened 2022-02-15
Simulation for various FF types

mmicko:
Added simulation for all FF types.
Improved co-simulation performance.
Added tests per FF type.

### DIFF
--- a/passes/sat/sim.cc
+++ b/passes/sat/sim.cc
@@ -22,6 +22,7 @@
 #include "kernel/celltypes.h"
 #include "kernel/mem.h"
 #include "kernel/fstdata.h"
+#include "kernel/ff.h"
 
 #include <ctime>
 
@@ -76,6 +77,7 @@ struct SimShared
 	double stop_time = -1;
 	SimulationMode sim_mode = SimulationMode::sim;
 	bool cycles_set = false;
+	const pool<IdString> ff_types = RTLIL::builtin_ff_cell_types();
 };
 
 void zinit(State &v)
@@ -113,8 +115,13 @@ struct SimInstance
 
 	struct ff_state_t
 	{
-		State past_clock;
 		Const past_d;
+		Const past_ad;
+		SigSpec past_clk;
+		SigSpec past_ce;
+		SigSpec past_srst;
+		
+		FfData data;
 	};
 
 	struct mem_state_t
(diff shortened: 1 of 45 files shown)


passes/sat/sim.cc
```diff
@@ -76,6 +77,7 @@ struct SimShared
 	double stop_time = -1;
 	SimulationMode sim_mode = SimulationMode::sim;
 	bool cycles_set = false;
+	const pool<IdString> ff_types = RTLIL::builtin_ff_cell_types();
```
mwkmwkmwk, line 80: `builtin_ff_cell_types` just returns an address of a static, there's little point cacheing it

passes/sat/sim.cc
```diff
@@ -113,8 +115,13 @@ struct SimInstance
 
 	struct ff_state_t
 	{
-		State past_clock;
 		Const past_d;
+		Const past_ad;
+		SigSpec past_clk;
```
mwkmwkmwk, line 120: why SigSpec and not State?

mmicko: just a left over from past

passes/sat/sim.cc
```diff
@@ -466,20 +477,138 @@ struct SimInstance
 
 		for (auto &it : ff_database)
 		{
-			Cell *cell = it.first;
 			ff_state_t &ff = it.second;
+			FfData ff_data = ff.data;
```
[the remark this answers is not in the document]
mmicko: fixed

passes/sat/sim.cc
mmicko: fixed

passes/sat/sim.cc
```diff
@@ -466,21 +476,63 @@ struct SimInstance
...
+							(ff.past_clk == State::S1 && current_clk != State::S1)) {
+					bool ce = ff.past_ce == (ff_data.pol_ce ? State::S1 : State::S0);
+					// chip enable priority over reset
+					if (ff_data.ce_over_srst && ff_data.has_ce && !ce) continue;
+					// set if no ce, or ce is enabled
+					if (!ff_data.has_ce || (ff_data.has_ce && ce)) {
+						current_q = ff.past_d;
+					}
+					// override if sync reset
+					if ((ff_data.has_srst) && (ff.past_srst == (ff_data.pol_srst ? State::S1 : State::S0))) {
```
mwkmwkmwk, line 497: remove the above ce_over_srst check and use it to gate srst here instead

mwkmwkmwk (approved):

PR #3207, opened 2022-02-18
fix handling of escaped chars in json backend and frontend (mostly)

nakengelhardt:
https://www.json.org/json-en.html

It is mandatory to escape the characters `\`, `"`, and any control character (<0x20) in JSON. This uses the two-character escapes `\\ \" \r \t \n \f \b` preferentially, and `\uXXXX` escapes for others in `write_json`.

It's optionally allowed to escape any character. This accepts any of the above in `read_json`, but *does not* implement parsing of code points >0x7F (despite this being mandatory according to the standard).


clairexen (approved):
LGTM

PR #3212, opened 2022-02-23
gowin: Add support for true differential output

yrabbit:
Signed-off-by: YRabbit <rabbit@yrabbit.cyou>

### DIFF
--- a/techlibs/gowin/cells_sim.v
+++ b/techlibs/gowin/cells_sim.v
@@ -570,18 +570,29 @@ module OBUF(output O, input I);
 	assign O = I;
 endmodule
 
+(* abc9_box *)
 module TBUF (O, I, OEN);
   input I, OEN;
   output O;
   assign O = OEN ? 1'bz : I;
 endmodule
 
+(* abc9_box *)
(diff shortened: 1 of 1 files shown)


techlibs/gowin/cells_sim.v
```diff
@@ -570,12 +570,14 @@ module OBUF(output O, input I);
 	assign O = I;
 endmodule
 
+(* abc9_box *)
```
mwkmwkmwk, line 573: why? there's absolutely no purpose for `abc9_box` on I/O buffer cells

yrabbit: You mean for all IBUF, OBUF, IOBUF, etc.? I wasn't sure if it was necessary, because IBUF and OBUF already had such a line and I didn't do it. 
Should I remove them from all buffers?

mwkmwkmwk: yeah, they shouldn't be there

yrabbit: Done

PR #3215, opened 2022-02-23
sv: fix always_comb auto nosync for nested and function blocks

zachjs:
This supersedes #3153. @kamilrakoczy @tgorochowik


kamilrakoczy (approved):
Thanks for working on this. I can confirm that this PR solves my original problem.

PR #3222, opened 2022-03-01
Prune Linux CI builds

zachjs:
This PR aims to remove excess build variants from the CI. Please feel free to share any comments or concerns!

**Update:** see https://github.com/YosysHQ/yosys/pull/3222#issuecomment-1065015254

Below are the 18 current Linux matrix runs, with the 10 runs this PR removes struck.

* ubuntu-20.04, focal, clang-12, c++11
* ~~ubuntu-20.04, focal, clang-12, c++14~~
* ~~ubuntu-20.04, focal, clang-12, c++17~~
* ubuntu-20.04, focal, clang-12, c++20
* ubuntu-20.04, focal, clang-11, c++11
* ~~ubuntu-20.04, focal, clang-11, c++14~~
* ~~ubuntu-20.04, focal, clang-11, c++17~~
* ~~ubuntu-20.04, focal, clang-11, c++20~~
* ubuntu-20.04, focal, gcc-11, c++11
* ~~ubuntu-20.04, focal, gcc-11, c++14~~
* ~~ubuntu-20.04, focal, gcc-11, c++17~~
* ubuntu-20.04, focal, gcc-11, c++20
* ubuntu-20.04, focal, gcc-10, c++11
* ~~ubuntu-20.04, focal, gcc-10, c++14~~
* ~~ubuntu-20.04, focal, gcc-10, c++17~~
* ~~ubuntu-20.04, focal, gcc-10, c++20~~
* ubuntu-18.04, bionic, clang-3.9, c++11
* ubuntu-18.04, bionic, gcc-4.8, c++11

C++11 is the default standard for Yosys. To my understanding, we build using the newer standards primarily to ensure forward-compatibility.

Although there are likely exceptions, I would expect that for a particular environment, coverage for C++11 and C++20 should obviate the need to test with C++14 and C++17, especially given we are not currently planning to specifically target those standards by default.

I also dropped C++20 for the older compilers given that they likely don't provide significant additional value over the matching C++11 build combined with the newer C++20 build.

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -11,9 +11,7 @@ jobs:
           - { id: ubuntu-20.04, name: focal }
         compiler:
           - 'clang-12'
-          - 'clang-11'
           - 'gcc-11'
-          - 'gcc-10'
         cpp_std:
           - 'c++11'
           - 'c++14'
@@ -21,6 +19,12 @@ jobs:
           - 'c++20'
         include:
           # Limit the older compilers to C++11 mode
+          - os: { id: ubuntu-20.04, name: focal }
+            compiler: 'clang-11'
+            cpp_std: 'c++11'
+          - os: { id: ubuntu-20.04, name: focal }
+            compiler: 'gcc-10'
+            cpp_std: 'c++11'
           - os: { id: ubuntu-18.04, name: bionic }
             compiler: 'clang-3.9'
             cpp_std: 'c++11'


mmicko (approved):
LGTM

lethalbit (approved):

lethalbit:
We should also think about bumping GCC 4.8 to 4.9.2, as 4.8 /technically/ doesn't support the full C++11 feature set,

udif:
> We should also think about bumping GCC 4.8 to 4.9.2, as 4.8 /technically/ doesn't support the full C++11 feature set,

RHEL7/Centos7, which is common in the EDA world due to being supported by the big3, comes with gcc 4.8.5, so unless you must do otherwise, keeping 4.8.x would be useful.

lethalbit:
> > We should also think about bumping GCC 4.8 to 4.9.2, as 4.8 /technically/ doesn't support the full C++11 feature set,
> 
> RHEL7/Centos7, which is common in the EDA world due to being supported by the big3, comes with gcc 4.8.5, so unless you must do otherwise, keeping 4.8.x would be useful.

While I agree that we need to keep support, being unable to use at least 4.9.2 means we can't make use of all the C++11 features and expect them to be there, such as braced initializers in constructors.

there are scl's such as `devtoolset-5-gcc`/`devtoolset-5-gcc-c++` that provide GCC/G++ 5 and up on Centos/RHEL7, so I don't think it's too much of an ask that the minimum GCC version be raised a touch just so we get full C++11 compliant support.

That's just my take on it anyway, i'd love to hear what the others have  to say about that.

DX-MON:
Having been tracking this issue ourself, we'd note that removing testing for C++14 and 17 will remove diagnostic potential for issues that might crop up as the compiler applies different sets of diagnostics depending on standard version, which can sometimes diagnose an issue but only at a specific standard version - C++17 in particular is famous for this in GCC.

zachjs:
> Having been tracking this issue ourself, we'd note that removing testing for C++14 and 17 will remove diagnostic potential for issues that might crop up as the compiler applies different sets of diagnostics depending on standard version, which can sometimes diagnose an issue but only at a specific standard version - C++17 in particular is famous for this in GCC.

Are there diagnostics which are off in C++11, on in 14 or 17, but off again in 20? Things like "This flag is enabled by default for -std=c++17" in the GCC manual suggest this _might_ be the case, but the source generally suggests otherwise.

DX-MON:
> > Having been tracking this issue ourself, we'd note that removing testing for C++14 and 17 will remove diagnostic potential for issues that might crop up as the compiler applies different sets of diagnostics depending on standard version, which can sometimes diagnose an issue but only at a specific standard version - C++17 in particular is famous for this in GCC.
> 
> Are there diagnostics which are off in C++11, on in 14 or 17, but off again in 20? Things like "This flag is enabled by default for -std=c++17" in the GCC manual suggest this _might_ be the case, but the source generally suggests otherwise.

yes, functionally because of how several of the diagnostics work and that some of them are standard-specific but still useful.. they are per-standard - C++17 saw a pile of C++11 era diagnostic bite the dust in newer GCCs when in the newer standards mode while new diagnostics were introduced and which aren't guaranteed to be on for C++20

zachjs:
With #3228 incoming and the concerns raised above, I've restored testing of C++14 and C++17, as the overhead should now be lower. This PR now simply removes 6 of the 18 Linux runs by "demoting" GCC 10 and Clang 11 so they are only used under C++11.

PR #3227, opened 2022-03-09
intel_alm: M10K write-enable is negative-true

Ravenslofty:
This requires a techmap shim to invert the polarity of the write-enable. This feels like something that `memory_libmap` or whatever should really be able to do, but anyway.

(Finding this out has caused me so much grief over the past few months >.>)


techlibs/intel_alm/common/quartus_rename.v
```diff
@@ -157,6 +157,10 @@ output [CFG_DBITS-1:0] B1DATA;
 // Much like the MLAB, the M10K has mem_init[01234] parameters which would let
 // you initialise the RAM cell via hex literals. If they were implemented.
 
+wire A1EN_N;
+
+NOT wren_inv (.IN(A1EN), .OUT(A1EN_N));
```
gatecat, line 162: this confused me for a sec - it might be worth adding a note that this is actually uninverting the enable back to its original polarity - aiui

gatecat (approved):
LGTM

PR #3228, opened 2022-03-10
Disable tests on most of platforms

mmicko:


mmicko:
@zachjs tests for macOS are there, but just for macOS-11 with c++11 only, so one test run for Linux and one for macOS. macOS-10.5 is maybe not even needed but it is good to check compilation on older OS, and for it tests are fully disabled

zachjs:
@mmicko Ah, I misread. That makes sense to me!

PR #3238, opened 2022-03-18
fix handling of cell IDs with backslashes in sby junit interaction

nakengelhardt:
Backslashes are not allowed in smt2 format, so the smt2 backend translates them to forward slashes. This prevents sby from finding the corresponding cell in the json export.

Since sby interprets a hot comment in the smt2 file, where backslashes aren't actually an issue, print the unmodified cell name in the comment.


nakengelhardt:
Handling this in sby instead so that the smt2 file stays internally consistent. https://github.com/YosysHQ/sby/pull/142

PR #3249, opened 2022-03-25
Add -no-startoffset option to write_aiger

mmicko:


jix (approved):
LGTM, and when changing sby to use it, this fixes all the related issues I can reproduce locally

PR #3257, opened 2022-03-29
tribuf: `-formal` option: convert all to logic and detect conflicts

jix:
Using this simplifies formal verification of designs containing tristate busses. https://gist.github.com/jix/6ccf5b58f10ce77cf40d9f13ed15db86 is an example of using it via sby.

### DIFF
--- a/passes/techmap/tribuf.cc
+++ b/passes/techmap/tribuf.cc
@@ -26,10 +26,12 @@ PRIVATE_NAMESPACE_BEGIN
 struct TribufConfig {
 	bool merge_mode;
 	bool logic_mode;
+	bool formal_mode;
 
 	TribufConfig() {
 		merge_mode = false;
 		logic_mode = false;
+		formal_mode = false;
 	}
 };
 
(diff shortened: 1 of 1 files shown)


mattvenn:
Looks good. One improvement would be to improve the assertion failed message. It could be confusing as it's coming from an internal assertion rather than one I write myself.

mattvenn (approved):

jix:
I had missed a special case when there is only one tristate buffer driving a wire. I think that's the error you're seeing, this should fix it.

jix:
I'm now using a slightly different way to encode the constraints, using a separate assertion for each driver instead of just a single assertion for all drivers on the same wire. This means I can use the source location of the driver as source location for the assert which in turn means that a conflict generates errors that point to all source locations that are driving the wire during the conflict.

```
Checking assertions in step 11..
BMC failed!
Assert failed in test: test.v:37 ($tribuf_conflict$$flatten/driver_a.$verific$i6$test.v:37$46)
Assert failed in test: test.v:37 ($tribuf_conflict$$flatten/driver_b.$verific$i6$test.v:37$46)
Writing trace to VCD file: engine_0/trace.vcd
```

This might have different solving performance from the previous approach, but it shouldn't be much worse in most cases and might even be faster for common cases.

I've kept the previous version at https://github.com/jix/yosys/tree/tribuf-formal-v1

@mattvenn do you think this way the assertion failure message is clear enough? It still looks the same as an explicit assertion added by the user, but avoiding that would require separate, more involved changes to support that in general.

mattvenn:
I think it's definitely better to show all outputs involved in the violation.
Yes having an extra line that says explicitly what has failed would be good but the word tribuf is in there.

PR #3263, opened 2022-04-01
Set init values for wrapped  async control signals

mmicko:
With current code next example fails since counter starts at wrong value.
Bug is noticed due to fact that all FFs are aldffs when imported trough Verific.

```
[options]
mode bmc
multiclock on
depth 2

[engines]
smtbmc boolector


[script]
read -formal aldff.v
prep -top aldff

[file aldff.v]
module aldff(
input clk
);
parameter WIDTH=10;

reg [3:0] f_count;

initial f_count = 1;
always @(posedge clk)
if (f_count == WIDTH-1)
	f_count <= 0;
else
	f_count <= f_count + 1'b1;

always @(*)
	assert(f_count < WIDTH);

endmodule
```

### DIFF
--- a/passes/sat/clk2fflogic.cc
+++ b/passes/sat/clk2fflogic.cc
@@ -44,6 +44,7 @@ struct Clk2fflogicPass : public Pass {
 	}
 	SigSpec wrap_async_control(Module *module, SigSpec sig, bool polarity, IdString past_sig_id) {
 		Wire *past_sig = module->addWire(past_sig_id, GetSize(sig));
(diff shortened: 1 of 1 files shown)


mwkmwkmwk (approved):

mmicko:
@mwkmwkmwk wrap_async_control_gate is not used any more, all cases using it are removed in https://github.com/YosysHQ/yosys/commit/abe4e9e607c2d3d19e6342b78792fce1abedd78b. Shoud we just remove it ?

PR #3269, opened 2022-04-05
Reorder steps in -auto-top to fix synth command, fixes #3261

### LINKED ISSUE #3261: synth_ice40 not working with nextpnr since Yosys 0.12
```
module test
    (
      input [2:0] NUM
    , output wire  OUTPUT
    );

  assign OUTPUT = NUM == 3'd3;
endmodule
```

`yosys test.v -o test.json -p synth_ice40`
`nextpnr-ice40 --hx1k --package tq144 --json test.json --pcf-allow-unconstrained`

## Expected behavior

nextpnr-ice40 would finish with no errors. This is what happens with yosys 0.12

## Actual behavior

nextpnr-ice40 fails with an error
`ERROR: cell type '$eq' is unsupported`

It appears that yosys is generating different cell types post-0.12 ($eq vs SB_LUT4 here) besides just $eq, but I don't know enough about yosys to know if that is expected or not

mmicko:


mmicko:
All synth_* calls use "hierarchy -auto-top" if top is not specified. In bug it was noticed that if we call "hierarchy -auto-top" before synth call it will work, so that was hint that order is not proper.

modwizcode (approved):
LGTM.

whitequark (approved):

PR #3271, opened 2022-04-07
abc: Add support for FFs with reset in -dff

mwkmwkmwk:


Ravenslofty (commented):
Looks okay to me, for whatever that's worth,

PR #3273, opened 2022-04-08
Do not enable features that require the support of compression compiled in

modwizcode:
Currently several files are compiled always despite a dependency on inclusion of libraries that are conditionally compiled in. This fixes the issue by expressing the dependency in the build system properly.


whitequark (approved):
LGTM. This should fix the YoWASP Yosys build, which has been broken for some time.

mmicko:
Thanks @modwizcode . Main issue here is that fst library heavily depends of ZLIB (including header definition) so it is not easy to make it independent without touching external library. YoWASP builds use case is not requiring sim command so think we can go with this

PR #3285, opened 2022-04-20
Makefile: use DESTDIR in yosys-config

alanvgreen:
Modifies the yosys-config rule, adding the $(DESTDIR) prefix to
references to $(DATDIR) and $(BINDIR). This ensures that the generated
yosys-config file refers to the same data and bin directories that the
install and uninstall rules refer to.

There seem to be other DESTDIR related problems in Makefile, but this
at least addresses some of them until a more comprehensive solution is
available.

Signed-off-by: Alan Green <avg@google.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -749,9 +749,9 @@ LDLIBS_NOVERIFIC = $(LDLIBS)
 endif
 
 $(PROGRAM_PREFIX)yosys-config: misc/yosys-config.in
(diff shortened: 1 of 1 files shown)


Xiretza:
This is incorrect - yosys-config should refer to the actual paths after the package is installed, `$(DESTDIR)` is merely an artifact of packaging and must not be referenced by installed programs.

alanvgreen:
I see. That makes sense.

alanvgreen:
For future reference, the standard purpose of `DESTDIR` is to stage installs to an intermediate directory before they land in their final destination. As noted at https://www.gnu.org/prep/standards/html_node/DESTDIR.html `DESTDIR` should only be used inside `install` and `uninstall` targets.

PR #3290, opened 2022-04-24
Fix build on FreeBSD, which has no alloca.h

mpasternacki:
After this fix, `abc` compilation failed, which was fixed upstream in https://github.com/berkeley-abc/abc/pull/147. I have applied that change locally and used `ABCREV=default` to buid). Do you prefer to merge upstream changes into https://github.com/YosysHQ/abc, or should I make a PR with that change for your fork?


mmicko:
Thanks for this, planning to sync our ABC to one upstream, so it should be all in before next release

PR #3299, opened 2022-05-05
sim pass: support for memories

mmicko:
Included changes are here to support two different ways of representing memory locations in VCD/FST file, one is memory_cell<hex_address> and other memory_cell[decimal_addres], in both cases we create special memory structure with mappings of specific addresses to data inputs in FST (so mem cells can use that directly) and also add mapping to memory_cell[decimal_addres] as wire, since memory pass can convert mem cells to registers.

Also including change in btor writer to put aliases for main clock signal in info file, that then generates proper VCD output created by btorsim.


mmicko:
@nakengelhardt Can you please confirm if btor info file should contain aliased clock lines, or that should be fixed on btorsim side.

PR #3310, opened 2022-05-10
Now calls Tcl_Init after creating the interp, fixes clock format.

robinsonb5:
Without this patch, Yosys doesn't call Tcl_Init(), which means that a handful of Tcl commands don't work - for example, attempting to use "clock format" results in the following error:
ERROR: TCL interpreter returned an error: invalid command name "::tcl::clock::format"
Tcl_Init() calls the init.tcl script, part of the Tcl standard library code, which fills in some implemented-in-tcl commands in the ::tcl::clock namespace, namely clock format, clock add and clock scan.

Adding this call fixes the problem.

### DIFF
--- a/kernel/yosys.cc
+++ b/kernel/yosys.cc
@@ -734,7 +734,7 @@ extern Tcl_Interp *yosys_get_tcl_interp()
 	if (yosys_tcl_interp == NULL) {
 		yosys_tcl_interp = Tcl_CreateInterp();
 		if (Tcl_Init(yosys_tcl_interp)!=TCL_OK)
-			log("Tcl_Init() call failed\n");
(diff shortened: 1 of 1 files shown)


mmicko:
@robinsonb5 Could you please change from `log` to `log_warning`, and it would be good that message is more descriptive since it can fail if there is no init.tcl or if there is error loading it.
Are there additional name conflicts when init.tcl is loaded ?

robinsonb5:
No problem - I've updated as requested.
(Note, however, that Tcl_ErrnoMsg(Tcl_GetErrno()), which I think is the best source of error information, returns "no such file or directory" even if init.tcl does exist but contains nonsense.)

There are no extra name collisions as a result of calling Tcl_Init() - calling "yosys -import" reports collisions for "cd", "eval", "exec", "read" and "trace" with or without this patch.

PR #3321, opened 2022-05-18
techlibs/gatemate: adding option to allows dff/dlatch with an init value

trabucayre:
As mentioned in **ug1001-gatemate1-primitives 3.1 CC_DF **, *initial value during configuration is not supported*.
But with some already available code or code generator an initial value (in addition to reset signal) may be used mainly to be simulator compliant.

When it's a custom code or dedicated for this device, it's easy to fix this behavior, but in some situations it's difficult/impossible and code can't be synthetized.

This PR add an option to change the `dfflegalize` init_value from **x** to 0, to avoid **yosys** to stop synthesis.
Default behavior is kept unchanged and it's up to the user to activate this at it risk.

cc: @pu-cc

### DIFF
--- a/techlibs/gatemate/synth_gatemate.cc
+++ b/techlibs/gatemate/synth_gatemate.cc
@@ -74,6 +74,9 @@ struct SynthGateMatePass : public ScriptPass
 		log("    -dff\n");
 		log("        run 'abc' with -dff option\n");
 		log("\n");
+		log("    -ignore-initial-value\n");
+		log("        do not stop for dff or dlatch with an init value\n");
+		log("\n");
 		log("    -retime\n");
 		log("        run 'abc' with '-dff -D 1' options\n");
 		log("\n");
@@ -90,7 +93,7 @@ struct SynthGateMatePass : public ScriptPass
 	}
 
 	string top_opt, vlog_file, json_file;
-	bool noflatten, nobram, noaddf, nomult, nomx4, nomx8, luttree, dff, retime, noiopad, noclkbuf;
+	bool noflatten, nobram, noaddf, nomult, nomx4, nomx8, luttree, dff, retime, noiopad, noclkbuf, ignoreinitialvalue;
 
 	void clear_flags() override
 	{
@@ -108,6 +111,7 @@ struct SynthGateMatePass : public ScriptPass
 		retime = false;
 		noiopad = false;
 		noclkbuf = false;
+		ignoreinitialvalue = false;
 	}
 
 	void execute(std::vector<std::string> args, RTLIL::Design *design) override
@@ -136,6 +140,10 @@ struct SynthGateMatePass : public ScriptPass
 					break;
 				run_from = args[++argidx].substr(0, pos);
 				run_to = args[argidx].substr(pos+1);
+				continue;
+			}
+			if (args[argidx] == "-ignore-initial-value") {
+				ignoreinitialvalue = true;
 				continue;
 			}
 			if (args[argidx] == "-noflatten") {
@@ -212,6 +220,12 @@ struct SynthGateMatePass : public ScriptPass
 			if (!noflatten) {
 				run("flatten");
 			}
+
+			if (ignoreinitialvalue)
+			{
+				run("attrmap -remove init");
+			}
+
 			run("tribuf -logic");
 			run("deminout");
 			run("opt_expr");


techlibs/gatemate/synth_gatemate.cc
```diff
@@ -71,6 +71,9 @@ struct SynthGateMatePass : public ScriptPass
 		log("    -dff\n");
 		log("        run 'abc' with -dff option\n");
 		log("\n");
+		log("    -initial\n");
```
gatecat, line 74: I think there could be a better name for this like `-ignore-initial` or similar?

trabucayre: Yep. I'm usually really bad to find a good option name :-/

Maybe `-ignore-initial-value`: to be more explicit?
If it's good to you I update my PR.
Thanks

trabucayre: fixed

trabucayre:
build failure is now fixed.

nakengelhardt:
Instead of asking `dfflegalize` to legalize to a specific init value that doesn't actually match the hardware, wouldn't a better approach be to remove the init values (`attrmap -remove init`) at the beginning of the flow when the `-ignore-initial-value` option is set?

trabucayre:
You mean adding the test and `attrmap -remove init` just before `if (check_label("map_regs"))`
Tested like this and works as expected. Could you confirm and I update the PR accordingly.

Thanks for you feedback.

nakengelhardt:
I think I would put it at the very beginning in `prepare`, before any call to `opt_*` passes, just to make sure it is consistently ignored and not e.g. taken into account when the register can be optimized away to a constant.

trabucayre:
I have tried to put this call where you mentionned -> it's fails again.
In fact I tried to move this call and tried to synthesis a verilog known to fails: the first position where synthesis end successfully is just [before](https://github.com/YosysHQ/yosys/blob/master/techlibs/gatemate/synth_gatemate.cc#L251)  `if (check_label("map_ffram"))`

trabucayre:
With a long delay: I have updated the PR with requested changes.
This operation can"t be done earlier in the sequence: if a `CC_BRAM_20K` is filled by using `readmemh` synthesis fails otherwise.

nakengelhardt:
Thanks! LGTM, but it seems there's a merge conflict, can you rebase on latest master?

trabucayre:
Done. Thanks

trabucayre:
any news?

nakengelhardt:
Sorry, I had an idea for where the initialized FFs were coming from and wanted to have a look to confirm (and have a think about whether there's a more consistent solution for this option), but I couldn't reproduce it and then forgot. Do you have a design I could test with?

trabucayre:
I have uploaded a [design](https://kmf2.trabucayre.com/blink.tgz)
It based on @olofk blinky
You have to define `CC_TOOL`

```
export CC_TOOL="/somewhere/cc-toolchain-linux"
```

If you try using make: synthesis fails with
```
2.36. Executing DFFLEGALIZE pass (convert FFs to types supported by the target).
ERROR: FF blinky.$auto$ff.cc:266:slice$170 (type $_SDFF_PN0_) cannot be legalized: initialized D flip-flops are not supported
make: *** [Makefile:7 : blinky_synth.v] Erreur 1
```

By uncommenting Makefile's line
```
#YOSYS_OPTS = -ignore-initial-value
```

Everything is build and the **blinky_00.cfg.bit** is produces

trabucayre:
Any news?

nakengelhardt:
Thanks for the reminder! That wasn't the example I was looking for, but I was able to reproduce the issue with memory inference creating init values. I'll investigate how to best deal with it.

trabucayre:
I have updated PR by moving `attrmap -remove init` before `tribuf -logic` and after `flatten`.
I don't know why but in previous tests I have seen, when using `readmemh`, the RAM was filled with zero if `attrmap -remove init` was called before ` if (check_label("map_ffram"))`. I have tried again with
[this example](https://github.com/YosysHQ/yosys/files/10719660/yosys_patch_ram_init.tgz) and it seems RAM is correctly configured. I don't know what I did earlier. Now RAM is correctly filled with hex file content and no more crash.

trabucayre:
I applied manually PR #3688. This one fix all issues mentioned earlier and lot more user-friendly.
So I close this PR now useless.
Thanks!

PR #3324, opened 2022-05-19
select: Fix -assert-none and -assert-any error output and docs

jix:
Both of these options consider a selection containing only empty modules
as non-empty. This wasn't mentioned in the documentation nor did the
error message when using `select -assert-none` list those empty modules,
which produced a very confusing error message complaining about a
non-empty selection followed by an empty listing of the selection.

This fixes the documentation and changes the `-assert-none` and
`-assert-any` assertion error messages to also output fully selected
modules (this includes selected empty modules).

It doesn't change the messages for `-assert-count` etc. as they don't
count modules.

### DIFF
--- a/passes/cmds/select.cc
+++ b/passes/cmds/select.cc
@@ -944,12 +944,14 @@ static void select_stmt(RTLIL::Design *design, std::string arg, bool disable_emp
 	}
 }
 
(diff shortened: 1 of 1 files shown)


nakengelhardt:
The documentation change to `object or module` makes me wonder: modules are objects, right? Is the surrounding documentation consistent on the question whether modules count as objects?

PR #3328, opened 2022-05-20
verilog: Fix signedness of array querying function results

jix:
genrtlil.cc and simplify.cc had inconsistent and slightly broken
handling of signedness for array querying functions. These functions are
defined to return a signed result. Simplify always produced an unsigned
and genrtlil always a signed result. As both detectSignWidthWorker and
simplify handle context dependent conversion to an unsigned value they
should produce a signed result exactly when sign_hint is true.

Includes tests for the the relvant edge cases for context dependent
conversion to an unsigned value.


tests/verilog/sign_array_query.ys
```diff
@@ -0,0 +1,54 @@
...
+
+design -reset
+read_verilog -formal <<EOT
+module top(input clk);
+
+    reg [-1:-1] x;
+    reg y = 0;
+
+    always @(posedge clk) begin
+        case ($left(x)) 36'sh0ffffffff: y = 1; 32'h0: ; endcase
```
zachjs, line 28: It seems like this should pass regardless of whether or not `$left(x)` is signed.

frontends/ast/genrtlil.cc
```diff
@@ -1090,7 +1090,6 @@ void AstNode::detectSignWidthWorker(int &width_hint, bool &sign_hint, bool *foun
 		}
 		if (str == "\\$size" || str == "\\$bits" || str == "\\$high" || str == "\\$low" || str == "\\$left" || str == "\\$right") {
 			width_hint = 32;
```
zachjs, line 1092: Thank you for fixing this! I think this should also use `max` to avoid trimming the width of larger expressions. I wouldn't mind if that is fixed separately, if you'd prefer.

zachjs (commented):
Thank you! To make the release process a bit easier, I've been trying to get frontend changes into the changelog as they are merged. I think this is probably worthy of a small note there.

tests/verilog/sign_array_query.ys
jix: That's true, assuming the signedness rules are properly implemented. This test fails when you apply the fix to genrtlil.cc but just change `newNode = mkconst_int(result, false);` to `newNode = mkconst_int(result, true);` instead of `newNode = mkconst_int(result, backup_sign_hint);` in simplify.cc. I wanted to have a test that can distinguish these cases to make sure my understanding of the `sign_hint` logic is correct.

jix: Turns out my understanding wasn't correct, `simplify`, in general, doesn't have to handle width and signedness conversion like this. I just happened to run into a different bug in const folding for case expressions while testing this (working on a fix for that right now) and handling the conversion in simplify worked around that. This should in fact be `newNode = mkconst_int(result, true);` and I'll update this PR when the other bug is fixed.

zachjs (approved):

PR #3331, opened 2022-05-23
Change way to get commit sha

mmicko:


Xiretza:
Any clues as to what this improves over the old solution?

zachjs:
@Xiretza This is intended to work around the new(ish) git `safe.directory` restrictions, which prevent running most commands if the current user differs from the repo owner in the filesystem.

PR #3332, opened 2022-05-23
Update Verific command file documentation

mmicko:


nakengelhardt:
Can you do the hard wrap to 80 char for the help messages? I think this was already not respected before, but it's worse now...

nakengelhardt (approved):

Issue #3330, opened 2022-05-23
$divfloor not supported by write_smt2


programmerjake:
`$divfloor` isn't supported by write_smt2, apparently it was forgotten though `$modfloor` does work.

## Steps to reproduce the issue

Compile commit 0b1a1a576 (git master when issue was created)
Run the following yosys script:
```
read_verilog -icells -formal <<EOF
module uut;
    wire [7:0] a = $anyconst, b = $anyconst, y;

    \$divfloor #(
        .A_WIDTH(8),
        .B_WIDTH(8),
        .Y_WIDTH(8),
        .A_SIGNED(0),
        .B_SIGNED(0),
    ) divfloor (
        .A(a),
        .B(b),
        .Y(y)
    );

    always @* begin
        assume(b == 4);
        assert(a >> 2 == y);
    end
endmodule
EOF
prep
write_smt2 -wires
```

## Expected behavior

generate smtlib2 code

## Actual behavior

```
...
; SMT-LIBv2 description generated by Yosys 0.17+33 (git sha1 0b1a1a576, gcc 9.4.0-1ubuntu1~20.04.1 -fPIC -Os)
Creating SMT-LIBv2 representation of module uut.
ERROR: Unsupported cell type $divfloor for cell uut.divfloor.
```

Xiretza:
I don't think SMT-LIB has an operator for that (`bvsdiv` is truncating division: https://smtlib.cs.uiowa.edu/logics-all.shtml), so it'd have to be implemented by hand - that's why I left it out in #1885.

PR #3335, opened 2022-05-24
add $divfloor support to write_smt2

programmerjake:
Fixes: #3330

Also adds a simple test that `$divfloor` and `$modfloor` actually work correctly, though it does require yices to run...

### DIFF
--- a/backends/smt2/smt2.cc
+++ b/backends/smt2/smt2.cc
@@ -649,6 +649,27 @@ struct Smt2Worker
 					return export_bvop(cell, "(bvurem A B)", 'd');
 				}
 			}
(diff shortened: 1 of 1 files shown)


jix:
The additions to write_smt2 look good to me. Making yices a hard dependency for `make test` is not an option though. If you push only the first commit I can merge this and add the test to [sby](https://github.com/YosysHQ/sby/) instead. There the complete oss-cad-suite including several SMT solvers is available to the tests.

programmerjake:
I moved the test to https://github.com/YosysHQ/sby/pull/161, does it look good now?

PR #3336, opened 2022-05-24
verilog: fix signedness when removing unreachable cases

jix:
This fixes an issue where simplifying case statements by removing unreachable cases can result in the wrong signedness being used for comparison with the remaining cases.


zachjs (approved):
Nice catch! This is a great, simple fix.

PR #3337, opened 2022-05-24
verilog: fix $past's signedness

jix:
This fixes https://github.com/YosysHQ/sby/issues/132 and fixes https://github.com/YosysHQ/sby/issues/117

### DIFF
--- a/tests/verilog/past_signedness.ys
+++ b/tests/verilog/past_signedness.ys
@@ -1,0 +1,35 @@
+logger -expect-no-warnings
+
+read_verilog -formal <<EOT
+module top(input clk);
+    reg signed [3:0] value = -1;
+    reg ready = 0;
+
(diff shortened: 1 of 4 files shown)


tests/verilog/past_signedness.ys
```diff
@@ -0,0 +1,17 @@
+logger -expect-no-warnings
+
+read_verilog -formal <<EOT
+module top(input clk);
+    reg signed [3:0] value = -1;
+    reg ready = 0;
+
+    always @(posedge clk) begin
+        if (ready)
+            assert ($past(value) == -1);
```
zachjs, line 10: It would be nice if we also ensured that $past of an unsigned expression remains unsigned.

jix: Good point, added a second test for that.

PR #3342, opened 2022-05-27
smt2: smt2_model attribute for blackboxes

jix:
This is a quick prototype of an alternative to #3319 for allowing custom smtlib2 expression for formal verification. This uses blackboxes with an attribute containing the smtlib2 instead of adding a new cell type. I think this has several advantages over using a new builtin cell type:

* All code changes remain inside smt2.cc and it requires less code overall
* It allows the use of descriptive port names and multiple inputs and outputs, instead of requiring the user to concatenate and split everything on both the verilog and the smtlib2 side
* It can be extended in a backwards compatible way by adding additional attributes. e.g. for keeping state / describing possible transitions using smtlib2

This is an example usage:
```
read_verilog -formal <<EOT

(* smt2_model =
    "(= ((_ to_fp 8 24) s) (fp.add RNE ((_ to_fp 8 24) a) ((_ to_fp 8 24) b)))"
*)
module fadd(input [31:0] a, b, output [31:0] s);
endmodule

(* smt2_model = {
    "(let ((a_f ((_ to_fp 8 24) a))) ",
    "(= finite (not (or (fp.isNaN a_f) (fp.isInfinite a_f)))))"
} *)
module is_finite(input [31:0] a, output finite);
endmodule

(* smt2_model =
    "(= eq (fp.eq ((_ to_fp 8 24) a) ((_ to_fp 8 24) b)))"
*)
module feq(input [31:0] a, b, output eq);
endmodule


module top(input clk, input [31:0] a, input [31:0] b, input [31:0] c);

    wire a_finite; is_finite a_isnan(a, a_finite); always @* assume (a_finite);
    wire b_finite; is_finite b_isnan(b, b_finite); always @* assume (b_finite);
    wire c_finite; is_finite c_isnan(c, c_finite); always @* assume (c_finite);

    wire [31:0] ab; fadd ab_fadd(a, b, ab);
    wire [31:0] bc; fadd bc_fadd(b, c, bc);
    wire [31:0] ca; fadd cd_fadd(c, a, ca);


    wire [31:0] ab_c; fadd ab_c_fadd(ab, c, ab_c);
    wire [31:0] bc_a; fadd bc_a_fadd(bc, a, bc_a);
    wire [31:0] ca_b; fadd ca_b_fadd(ca, b, ca_b);

    wire ab_c_eq_bc_a; feq ab_c_eq_bc_a_feq(ab_c, bc_a, ab_c_eq_bc_a);
    wire ab_c_eq_ca_b; feq ab_c_eq_ca_b_feq(ab_c, ca_b, ab_c_eq_ca_b);

    always @* assume (ab_c_eq_bc_a);
    always @* assert (ab_c_eq_ca_b);

endmodule
EOT

hierarchy -top top
prep -flatten
write_smt2 smt2_model.smt2
exec -- yosys-smtbmc -s bitwuzla --logic QF_ABVFP --dump-vcd smt2_model.vcd smt2_model.smt2
```


programmerjake:
Quoting https://github.com/YosysHQ/yosys/pull/3319#issuecomment-1139732265:
> > As I mentioned on IRC before, while I do think this feature is quite useful, I'm not convinced that this is the right approach to implement it in yosys. Having had a closer look now, I still think that adding a new builtin cell type is not required here. I've opened a draft PR (#3342) that shows the approach I would use instead.
> 
> hmm, that could work, though I'd have to refactor the 8kloc of code I already wrote: https://gitlab.com/nmigen/nmigen/-/merge_requests/7
> 
> also, I'm not sure having the model be in the form of an implicit assumption is the best idea...what happens if we write something that's accidentally never true?
> 
> having it instead just be a cell that computes a bool which we explicitly assert/assume using yosys's existing techniques allows smtbmc to give us the exact conditions under which the assumption fails. if we went with the implicit assertion idea, then smtbmc would need a bunch of extension to add the new assumption type, which imho would be waay more complex than taking my approach.

jix:
As discussed in #3319, this should be changed to use attributes on the outputs to specify the expression that defines an output value.

This also still needs documentation and some tests.

programmerjake:
probably superseded by the new rewritten version of #3319

PR #3349, opened 2022-06-01
Make 'stat' and 'select -count' save counts to scratchpad

nakengelhardt:
E.g. so the value can be accessed for decision making in script passes.

### DIFF
--- a/passes/cmds/select.cc
+++ b/passes/cmds/select.cc
@@ -1456,7 +1456,10 @@ struct SelectPass : public Pass {
 				}
 			}
 			if (count_mode)
+			{
+				design->scratchpad_set_int("select.count", total_count);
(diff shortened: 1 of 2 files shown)


tcal-x:
Hi @nakengelhardt , I'm checking that I'm not missing something -- 

I still need to do
```
set mycount [yosys tee -q -s result.string scratchpad -get select.count]
```
to get the count into a Tcl variable that I can use, right?

tcal-x:
Ah, I found that I can do
```
set mycount [scratchpad -copy select.count result.string]
```

nakengelhardt:
Yes, those are the options right now. It would probably make sense to update `select -count` to also save the value in `result.string`.

PR #3351, opened 2022-06-02
Regression testing for read-over-write emulation

KrystalDelusion:
New memory inference interprets patterns as read-over-write and adds emulation to support this, producing higher resource counts than prior versions.

Pull request adds additional `select` tests for affected architectures (ice40, gatemate, and ecp5).

Tested and working on `Yosys 0.16 (git sha1 b63e0a0ca, clang 10.0.0-4ubuntu1 -fPIC -Os)`
Failing on current version.


KrystalDelusion:
Title is somewhat of a misnomer, read-over-write emulation isn't a regression per se, but rather a regression in tests (and code more generally) that is not _expecting_ the additional logic.  This can be avoided by using the newly added `no_rw_check` attribute.

PR #3352, opened 2022-06-02
Avoid static initialization order fiasco for global logging objects.

hzeller:
Various non-trivial objects (vector, set, dict,...) are provided as
global variables from the logging subsystem, but since global
initialization order is undefined, this is not safe and code might
access these before they are initialized.

Replace with a safe function-level static initialization.

To minimize the changes needed throughout the codebase and code
expecting the variables to exist, macros are provided that make
the change source-code compatible.

### DIFF
--- a/kernel/log.cc
+++ b/kernel/log.cc
@@ -38,12 +38,38 @@
 
 YOSYS_NAMESPACE_BEGIN
 
-std::vector<FILE*> log_files;
-std::vector<std::ostream*> log_streams;
-std::map<std::string, std::set<std::string>> log_hdump;
-std::vector<YS_REGEX_TYPE> log_warn_regexes, log_nowarn_regexes, log_werror_regexes;
-dict<std::string, LogExpectedItem> log_expect_log, log_expect_warning, log_expect_error;
-std::set<std::string> log_warnings, log_experimentals, log_experimentals_ignored;
+// Initialize the global state with a safe function initialization to avoid
+// the static initialization order fiasco for them.
+struct GlobalState {
+	std::vector<FILE*> log_files_;
+	std::vector<std::ostream*> log_streams_;
+	std::map<std::string, std::set<std::string>> log_hdump_;
+	std::vector<YS_REGEX_TYPE> log_warn_regexes_, log_nowarn_regexes_, log_werror_regexes_;
+	dict<std::string, LogExpectedItem> log_expect_log_, log_expect_warning_, log_expect_error_;
+	std::set<std::string> log_warnings_, log_experimentals_, log_experimentals_ignored_;
+	std::string log_last_error_;
+};
+
+static GlobalState &gs()
+{
+	static GlobalState gs;
+	return gs;
+}
+
+std::vector<FILE*>& gs_log_files() { return gs().log_files_; }
+std::vector<std::ostream*>& gs_log_streams() { return gs().log_streams_; }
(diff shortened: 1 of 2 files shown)


clairexen:
Do you have an example of where this is a potential issue? All code that would potentially log should not run before yosys_setup(), and before main() has set up the logging channels, anyway.

hzeller:
This was motivated by an experimental implementation of a Pass that then also accessed `yosys_get_tcl_interp()` in its constructor, which then (for some unrelated, local Tcl-installation related reason) failed to initialize, resulting in a [log_warning()](https://github.com/YosysHQ/yosys/blob/master/kernel/yosys.cc#L753) which then crashed (when compiled with a bunch of debug flags in clang) because of the log-subsystem not ready yet, triggering the UB (and I linked that Pass statically to Yosys, so its initialization was run before main()).

Independent of if this particular situation was a good idea, it highlighted that it is problematic if logging is triggered early.
Given the overall structure in which we have globally and statically registered passes before `main()`, it is conceivable that some of these might trigger some logging, which is why I think it is a good idea to have it fail-safe -- which then motivated this change. Ideally, calling logging at any time should not be fatal.

As follow-up changes it would be good if some of the logging data would present itself not as set of global variables (with undefined initialization time) but as interactions with functions to call.
I wanted to keep the initial PR small and keep the headers compatible with existing code, but I can of course work towards that goal if you like.

arcturusannamalai:
Prefer if you changed define macros to something like
Auto log_xyz = gs_log_xyz;

I think some kind of Singleton is a necessary evil. Perhaps why not place it on a global context object of the design ?

Agree generally that all the distributed static init vars have to be corralled

widlarizer:
While I don't have any strong points against this solution it does risk breaking some external code while making it safe to cause a very specific problem that I don't think people are super likely to run into. I'm open to reopening the topic whenever it comes up again, but fixing init/deinit with singletons is something I'd prefer to do primarily to less exposed functionality than logging

hzeller:
The reason why I added this was because I in fact _did_ run into that situation; it has been a while but I think it was during initialization of a linked-in pass.
In general, relying on accidental link-order initialization is always a hazard.

I agree that it would be good to have this a little less exposed, but we have to work with what the start condition is. To be compatible with existing code I had added the macros that make it source compatible.

Anyway, I'll keep adding this patch (adapted to latest code) locally, but it would of course be good if this was done upstream.

widlarizer:
What I'm wondering about is if the "don't run log in a pass constructor" footgun is something others hit or if it's something you happened to run into because of peculiar circumstance specific to one pass and worked around / learned to avoid. Because if this is a repeating problem for plugin devs, I'm willing to push potential annoying "unnecessary" code changes on plugin devs to avoid it across the board

QuantamHD:
@widlarizer This does affect Google's stuff and Henner patched Yosys internally to make sure we don't run into this. Is there a way to resolve this sufficiently for both Yosys and Google's use cases?

hzeller:
Of course I don't have stats how often that happens to folks, but having public global variables with linker-random accidental initialization sequence is probably a good thing to avoid in a robust public interface.

widlarizer:
I see, fair enough then

mmicko:
Issue addressed in https://github.com/YosysHQ/yosys/pull/6115 by using singleton

PR #3367, opened 2022-06-07
smt2: emit smtlib2_comb_expr outputs after all inputs

jix:
This is a small followup fix to #3319. Before merging, I had only tested it in unroll mode, in which the unrolling code automatically orders the unrolled definitions. When unroll mode isn't used, the `define-fun` statements need to go after any declaration/definition referred to in the new definition. This always moves smtlib2_comb_expr outputs to the end, which is sufficient for this feature.


programmerjake:
you ran into the same issue I did earlier, macos's sed is kinda lame, so doesn't support all the same options as gnu sed.

PR #3375, opened 2022-06-14
Add check for BLIF with no model name

### LINKED ISSUE #3374: Empty BLIF Model Name SegFault

The issue seems to happen when reading a BLIF file with a model with no name.
For example, the following incomplete BLIF file:
```
.model
```

To recreate, please run: 
```
echo .model > example.blif
yosys -p read_blif example.blif 
```

## Expected behavior

A warning or error message should be issued about model with no name, segmentation fault should be avoided.

## Actual behavior

When run with the current master version of Yosys, this produces a segmentation fault.
```
Segmentation fault: 11
```

Instrumenting Yosys with the Address Sanitizer and Undefined Behavior Sanitizer provides some more insight:
```
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_M_construct null not valid
Aborted (core dumped)
```

ALGCDG:
Add a check when parsing BLIF that model name is not nullptr.
Fixes #3374

### DIFF
--- a/frontends/blif/blifparse.cc
+++ b/frontends/blif/blifparse.cc
@@ -166,7 +166,10 @@ void parse_blif(RTLIL::Design *design, std::istream &f, IdString dff_name, bool 
 					goto error;
 				module = new RTLIL::Module;
 				lastcell = nullptr;
-				module->name = RTLIL::escape_id(strtok(NULL, " \t\r\n"));
+				char *name = strtok(NULL, " \t\r\n");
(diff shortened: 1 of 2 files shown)


Ravenslofty:
Hmm, I think it would also be useful to add this as a testcase.

Ravenslofty:
In case I forget: if the CI goes green, ping me and I'll merge this.

PR #3387, opened 2022-06-20
btor: add support for $pos cell

ekiwi:
This PR adds support for the `$pos` cell type to the btor backend.
This is my first PR to yosys and I would appreciate any comments on style and any possible misunderstandings wrt. how the `$pos` cell is supposed to be interpreted. Thanks!

### DIFF
--- a/backends/btor/btor.cc
+++ b/backends/btor/btor.cc
@@ -446,24 +446,27 @@ struct BtorWorker
 			goto okay;
 		}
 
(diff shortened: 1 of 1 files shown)


jix:
Thank you for the PR! This looks good to me. I don't have anything to comment on regarding the coding style. To check the intended interpretation of cells, I usually use the Yosys command `help $pos+` which prints a Verilog model for the cell. Here this confirms that `$pos` does indeed do nothing apart from the width extension common to the other cells implementing various operators.

PR #3389, opened 2022-06-22
Support parameter default values in JSON frontend and Verilog backend

gussmith23:
This enables the reading in of the `parameter_default_values` value in JSON, and the writing out of the value in Verilog.

PR #3390, opened 2022-06-22
Support ids as consts

gussmith23:
Please deal with #3389 first; leaving this as a draft until that PR is dealt with.

This enables using identifiers as constant values. Specifically, this is useful for when we want to use a module parameter to initialize a cell's parameter, e.g.:

```
module test(...);
  parameter p = ...;
  ...
  MyCell #(.my_cell_param(p) mycell0 (...);
endmodule
```

This is a bit of a hack; please advise on how to do this in a more Yosys-friendly way!


gussmith23:
I doubt this is relevant anymore, closing

PR #3391, opened 2022-06-23
add hierarchy -smtcheck

programmerjake:
like -simcheck, but allow smtlib2_module modules.

Requested in https://github.com/YosysHQ/sby/pull/170#issuecomment-1160477086


programmerjake:
Tests in https://github.com/YosysHQ/sby/pull/170

programmerjake:
thanks!

Issue #2980, opened 2021-08-23
Yosys tcl command substitution does not return the results of the command


macd:
The result of a tcl command such as `set xxx [cmd]` should set the variable xxx to the result of command cmd.  This works for normal tcl commands but not for Yosys commands.  Here is a simple example. Say we have test.tcl with the following:
```
yosys -import
set xxx [expr 1 + 2]
puts "variable xxx: $xxx"

set yyy [help zinit]
puts "variable yyy: $yyy"
```
This results in
```
yosys> tcl test.tcl
[TCL: yosys -import] Command name collision: found pre-existing command `cd' -> skip.
[TCL: yosys -import] Command name collision: found pre-existing command `eval' -> skip.
[TCL: yosys -import] Command name collision: found pre-existing command `exec' -> skip.
[TCL: yosys -import] Command name collision: found pre-existing command `read' -> skip.
[TCL: yosys -import] Command name collision: found pre-existing command `trace' -> skip.
variable xxx: 3

    zinit [options] [selection]

Add inverters as needed to make all FFs zero-initialized.

    -all
        also add zero initialization to uninitialized FFs

variable yyy: 

yosys> 
```

rockybulwinkle:
Did you find how to capture the output of yosys commands? I am struggling with this same issue.

rockybulwinkle:
Well after much trial and tribulation, I made an incredibly ugly hack to get this functionality.

Basically, It redirects stdout to a temporary file, reads that file, and returns the value. It also restores stdout before returning.

I'm at a loss why I had to make this hack, it seems like basic functionality that should be builtin to yosys. What's the point of TCL scripting if the script can't react to what yosys returns?

Anway, here's the ugly code. I do depend on tclx for access to the "dup" command (for stdout preservation). There is a way to do it without dup (read here: https://stackoverflow.com/a/8532484/600853), but ugly as it is, if we don't preserve stdout, it will mess up logging.

```
#for some reason yosys doesn't initialize basic functionality like the ability to load packages. Do it manually here.
set tcl_library "/usr/share/tcl8.6"
source /usr/share/tcl8.6/init.tcl

#require tclx for dup command
package require Tclx

#import the yosys commands
#yosys -import

proc capture_stdout {args} {
    #open a tmpfile.
    set temp [exec mktemp]
    set newstdout [open $temp w]

    #Save the old stdout to restore later
    set old_stdout [dup stdout]
    #close old stdout
    close stdout

[...]

PR #3392, opened 2022-06-23
Update tcl doc, yosys does not return data to tcl

rockybulwinkle:
This pull request is to address YosysHQ/yosys#2980.

The documentation, as originally written, does not make it clear that yosys commands, when used within a tcl script, do not return any value to the tcl script.

This pull request notes this and offers a workaround via tee as noted in the issue.

### DIFF
--- a/kernel/yosys.cc
+++ b/kernel/yosys.cc
@@ -774,6 +774,12 @@ struct TclPass : public Pass {
 		log("If any arguments are specified, these arguments are provided to the script via\n");
 		log("the standard $argc and $argv variables.\n");
 		log("\n");
+		log("Note, tcl will not recieve the output of any yosys command. If the output\n");
+		log("of the tcl commands are needed, use the yosys command 'tee' to redirect yosys's\n");
+		log("output to a temporary file.\n");
+
+
+		log("\n");
 	}
 	void execute(std::vector<std::string> args, RTLIL::Design *) override {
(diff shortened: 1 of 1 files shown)


kernel/yosys.cc
```diff
@@ -773,6 +773,12 @@ struct TclPass : public Pass {
 		log("\n");
 		log("If any arguments are specified, these arguments are provided to the script via\n");
 		log("the standard $argc and $argv variables.\n");
+		log("\n");
+		log("Note, tcl will not recieve the output of any yosys command. If the output\n");
+		log("of the tcl commands are needed, use the yosys command 'tee' to redirect yosys's\n");
+		log("output to a temporary file.\n");
+
```
mwkmwkmwk, line 780: needless two empty lines

kernel/yosys.cc
```diff
@@ -773,6 +773,12 @@ struct TclPass : public Pass {
...
+
```
mwkmwkmwk, line 781:
```suggestion
```

kernel/yosys.cc
```diff
@@ -773,6 +773,12 @@ struct TclPass : public Pass {
...
 		log("\n");
```
mwkmwkmwk, line 782:
```suggestion
		log("output to a temporary file.\n");
		log("\n");
```

PR #3394, opened 2022-06-27
memory_map: -keepdc option for formal

jix:
Use it when invoking memory_map -rom-only from write_{smt2,btor}.


mwkmwkmwk (approved):
For the record: I very much dislike this

PR #3397, opened 2022-07-03
Apicula now supports lutram

pepijndevos:


pepijndevos:
See https://github.com/YosysHQ/apicula/pull/111 and https://github.com/YosysHQ/nextpnr/pull/995

PR #3403, opened 2022-07-06
Extra tests for memory_libmap pass

KrystalDelusion:


tests/arch/common/blockram.v
```diff
@@ -45,3 +45,34 @@ module sync_ram_sdp #(parameter DATA_WIDTH=8, ADDRESS_WIDTH=10)
 
 endmodule // sync_ram_sdp
 
+
+`default_nettype none
```
mwkmwkmwk, line 49: this should only be present once, at the top of the file

tests/arch/ecp5/bug1836.mem
```diff
@@ -0,0 +1,32 @@
+0x8000,0x8324,0x8647,0x896a,0x8c8b,0x8fab,0x92c7,0x95e1,
```
mwkmwkmwk, line 1: this is a large chunk of data of unclear copyright and origin, please replace with a placeholder

KrystalDelusion: This is taken from a random hex generator, is that placeholder enough?

tests/arch/common/blockram.v
KrystalDelusion: The file already has it twice, I was just following what was already there.

tests/arch/ecp5/bug1836.mem
KrystalDelusion: Correction, not random, it's a hex sine generator.  https://www.daycounter.com/Calculators/Sine-Generator-Calculator2.phtml

PR #3406, opened 2022-07-08
Better compatibility for OpenBSD

josuah:
These three commits are all it takes to compile the yosys part!

```
$ gmake ENABLE_TCL=0 ENABLE_READLINE=0
[Makefile.conf] CONFIG := clang
[  0%] Building kernel/version_52c275a8c.cc
[  0%] Building kernel/version_52c275a8c.o
[100%] Building yosys
/usr/local/bin/ld: libs/fst/fstapi.o: in function `fstWriterEmitHdrBytes(fstWriterContext*)':
/home/josuah/Code/yosys/libs/fst/fstapi.cc:893: warning: strcpy() is almost always misused, please use strlcpy()
/usr/local/bin/ld: libs/fst/fstapi.o: in function `fstWriterClose':
/home/josuah/Code/yosys/libs/fst/fstapi.cc:1994: warning: sprintf() is often misused, please use snprintf()
[100%] Building yosys-config
[100%] Building abc/abc-5f40c47
Pulling ABC from https://github.com/YosysHQ/abc:
+ test -d abc
+ git clone https://github.com/YosysHQ/abc abc
Cloning into 'abc'...
[...]
[100%] Building share/xilinx/mux_map.v
[100%] Building share/xilinx/xc3s_mult_map.v
[100%] Building share/xilinx/xc3sda_dsp_map.v
[100%] Building share/xilinx/xc6s_dsp_map.v
[100%] Building share/xilinx/xc4v_dsp_map.v
[100%] Building share/xilinx/xc5v_dsp_map.v
[100%] Building share/xilinx/xc7_dsp_map.v
[100%] Building share/xilinx/xcu_dsp_map.v
[100%] Building share/xilinx/abc9_model.v

  Build successful.

```

And then the compilation of https://github.com/YosysHQ/abc pursue, another topic.

I am not sure how many yosys OpenBSD users there are, but yosys is on [OpenBSD ports tree with a few patches](https://github.com/openbsd/ports/tree/master/cad/yosys/patches).

[EDIT: completed the log to show the end of the build happening on my setup]

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -450,9 +450,12 @@ endif
 ifeq ($(ENABLE_TCL),1)
 TCL_VERSION ?= tcl$(shell bash -c "tclsh <(echo 'puts [info tclversion]')")
 ifeq ($(OS), $(filter $(OS),FreeBSD OpenBSD NetBSD))
+# BSDs usually use tcl8.6, but the lib is named "libtcl86"
 TCL_INCLUDE ?= /usr/local/include/$(TCL_VERSION)
+TCL_LIBS ?= -l$(subst .,,$(TCL_VERSION))
 else
 TCL_INCLUDE ?= /usr/include/$(TCL_VERSION)
+TCL_LIBS ?= -l$(TCL_VERSION)
 endif
 
 ifeq ($(CONFIG),mxe)
@@ -460,12 +463,7 @@ CXXFLAGS += -DYOSYS_ENABLE_TCL
 LDLIBS += -ltcl86 -lwsock32 -lws2_32 -lnetapi32 -lz -luserenv
 else
 CXXFLAGS += $(shell PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) $(PKG_CONFIG) --silence-errors --cflags tcl || echo -I$(TCL_INCLUDE)) -DYOSYS_ENABLE_TCL
-ifeq ($(OS), $(filter $(OS),FreeBSD OpenBSD NetBSD))
-# BSDs usually use tcl8.6, but the lib is named "libtcl86"
-LDLIBS += $(shell PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) $(PKG_CONFIG) --silence-errors --libs tcl || echo -l$(TCL_VERSION) | tr -d '.')
-else
-LDLIBS += $(shell PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) $(PKG_CONFIG) --silence-errors --libs tcl || echo -l$(TCL_VERSION))
-endif
+LDLIBS += $(shell PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) $(PKG_CONFIG) --silence-errors --libs tcl || echo $(TCL_LIBS))
 endif
 endif
 
--- a/Makefile
+++ b/Makefile
@@ -126,7 +126,9 @@ endif
 
 else
 LDFLAGS += -rdynamic
+ifneq ($(OS), OpenBSD)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -126,8 +126,10 @@ endif
 
 else
 LDFLAGS += -rdynamic
+ifneq ($(OS),OpenBSD)
```
josuah, line 129: The patches on the ports tree filters -lrt afterward (after all the `ifeq $($(CONFIG), ...)`), while this avoids to add it.

I tried to stay close to your way of doing, feel free to ask me to change as per your liking here.

kernel/yosys.cc
```diff
@@ -868,7 +868,10 @@ std::string proc_self_dirname()
 	return "/";
 }
 #else
-	#error "Don't know how to determine process executable base path!"
+std::string proc_self_dirname()
+{
+	return PREFIX "/bin";
+}
```
josuah, line 874: Again, some difference here: the ports is specific to OpenBSD (`#ifdef __OpenBSD__`) and use `${PREFIX}` inside the string, I am not sure if it works.

Here is something that could cover more OSes than just OpenBSD.

josuah:
Small problem that I will fix shortly: `ERROR: ABC: execution of command "/usr/local/binyosys-abc -s -f /tmp/yosys-abc-yblmxf/abc.script 2>&1" failed: return code 127.`

I forgot a `/`.

kernel/yosys.cc
```diff
@@ -868,7 +868,10 @@ std::string proc_self_dirname()
 	return "/";
 }
 #else
-	#error "Don't know how to determine process executable base path!"
+std::string proc_self_dirname()
+{
+	return PREFIX "/bin/";
```
Ravenslofty, line 873: The whole point of this function is to *avoid* hardcoding the directory, so that Yosys works anywhere. If you try to run the Yosys test suite (which implicitly assumes being run in the source directory), I am fairly sure it will fail because it cannot locate itself, and thus cannot locate any of the share directories that the synthesis passes rely on.

I googled around for a solution for openbsd, but I think you might need a hack to the effect of saving `argv[0]`, `getcwd()`, and `getenv("PATH")` and then attempt to locate the binary that way.

josuah: This was what was done in the OpenBSD package, which is better than not having Yosys package at all, but for something upstream including the test suites...

So let's try with the argv trick, which is portable.

kernel/driver.cc
```diff
@@ -150,6 +150,9 @@ const char *prompt()
 
 #else /* EMSCRIPTEN */
 
+char Yosys::yosys_path[PATH_MAX];
+char *Yosys::yosys_argv0;
+
```
josuah, line 155: I set that here to look alike the rest of the source, but maybe you will have a better place for them.

josuah:
With the latest git push, OpenBSD (and `__NetBSD__`, anticipating a similar behavior) is targetted by an `#ifdef`.

https://curl.se/mail/lib-2022-01/0048.html shows  `__Dragonfly__` (DragonflyBSD), `__MidnightBSD__`, `__minix`, `__AMIGA__`, ` __SYMBIAN32__`, but not sure yosys aims to support the entire operating system zoo on a per-case basis. :)

The `#error` is still there as suggested by email (not sure why the message does not appear on this thread).

Feel free to be pedantic about any detail, so that I can adjust to fit in the yosys source.

josuah:
To make the merge easier, I removed the binary path changes from this pull request, and am preparing another which would affect all unix-like platforms.

PR #3414, opened 2022-07-18
Fallback for finding the executable directory

josuah:
In reference to #3406, this change is independent and touches different files.

The path is reconstructed out of argv[0], directly or from $PATH
depending on how the binary was called.

For instance, multiple operating systems do not support "/proc/self/exe":
https://marc.info/?l=openbsd-misc&m=144987773230417&w=2

But most will support access(2) and realpath(3) as used in this
approach.

### DIFF
--- a/kernel/driver.cc
+++ b/kernel/driver.cc
@@ -498,6 +498,12 @@ int main(int argc, char **argv)
 	if (print_stats)
 		log_hasher = new SHA1;
 
+#if defined(__OpenBSD__)
+	// save the executable origin for proc_self_dirname()
+	yosys_argv0 = argv[0];
+	realpath(yosys_argv0, yosys_path);
+#endif
+
 #if defined(__linux__)
 	// set stack size to >= 128 MB
 	{
--- a/kernel/yosys.cc
+++ b/kernel/yosys.cc
@@ -871,6 +871,35 @@ std::string proc_self_dirname()
 {
 	return "/";
 }
+#elif defined(__OpenBSD__)
+char Yosys::yosys_path[PATH_MAX];
+char *Yosys::yosys_argv0;
+
+std::string proc_self_dirname(void)
+{
+	char buf[PATH_MAX + 1] = "", *path, *p;
+	// if case argv[0] contains a valid path, return it
+	if (strlen(yosys_path) > 0) {
+		p = strrchr(yosys_path, '/');
+		snprintf(buf, sizeof buf, "%*s/", (int)(yosys_path - p), yosys_path);
+		return buf;
+	}
+	// if argv[0] does not, reconstruct the path out of $PATH
+	path = strdup(getenv("PATH"));
+	if (!path)
+		log_error("getenv(\"PATH\") failed: %s\n",  strerror(errno));
+	for (p = strtok(path, ":"); p; p = strtok(NULL, ":")) {
+		snprintf(buf, sizeof buf, "%s/%s", p, yosys_argv0);
+		if (access(buf, X_OK) == 0) {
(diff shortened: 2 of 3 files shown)


josuah:
The following build action would permit to check that various test environments behave appropriately:

https://github.com/YosysHQ/yosys/actions?query=branch%3Abinary-path-fallback

Thankfully, both Linux and MacOS, affected by these changes, are running the tests.

mmicko:
Was visible in previous PR, but note that realpath is not something that exists on all supported platforms (like Visual Studio windows build). 
Would suggest to keep all other OS code unaffected and just make OpenBSD path different.
Also there is getprogpath, see how it is used at :  https://github.com/springmeyer/shared-library-path/blob/master/src/implementation.cc#L70 or https://reviews.llvm.org/D32932

josuah:
Thank you for the feedback.

Absolutely! I forgot to `#ifdef`-guard the realpath() call in `main()`.
I kept the OpenBSD code separate on the previous attempt, but I got suggested to replace the code for multiple UNIX this one in #3406.

I am force-pushing soon...

josuah:
It looks like `getprogpath()` is part of LLVM library, and use the same approach as here:

https://github.com/freebsd/freebsd-src/blob/14c99b43ef69beafa01a1574641c4cf19ecf2b9d/contrib/llvm-project/llvm/lib/Support/Unix/Path.inc#L152-L192

It is doing the same but without the use of `realpath()`.

josuah:
The new code is now within `#elif defined(__OpenBSD__)`, including the part in main().

By applying both this and #3406 I could test it:

```
# cp yosys "$(which yosys)"

$ yosys
yosys>

$ "$PWD/yosys"
yosys>

$ ./yosys
yosys>
```

kernel/driver.cc
```diff
@@ -498,6 +498,12 @@ int main(int argc, char **argv)
 	if (print_stats)
 		log_hasher = new SHA1;
 
+#if defined(__OpenBSD__)
+	// save the executable origin for proc_self_dirname()
```
mmicko, line 502: Best is to use extern defs here then in yosys.h.  just use same block here:
```
extern char *yosys_argv0;
extern char yosys_path[PATH_MAX];
```

josuah: Do you mean to remind the reader that these variables are declared elsewhere?

I love the idea, but fail to add the declaration within `main()`:

```
[  0%] Building kernel/driver.o
kernel/driver.cc:502:22: error: definition or redeclaration of 'yosys_argv0' not allowed inside a function
        extern char *Yosys::yosys_argv0;
                     ~~~~~~~^
kernel/driver.cc:503:21: error: definition or redeclaration of 'yosys_path' not allowed inside a function
        extern char Yosys::yosys_path[PATH_MAX];
                    ~~~~~~~^
2 errors generated.
gmake: *** [Makefile:732: kernel/driver.o] Error 1
```

I can add it above main though:

```diff
--- a/kernel/driver.cc
+++ b/kernel/driver.cc
@@ -192,6 +192,11 @@ void yosys_atexit()
 #endif
 }

+#if defined(__OpenBSD__)
+extern char *Yosys::yosys_argv0;
+extern char Yosys::yosys_path[PATH_MAX];
+#endif
+
 int main(int argc, char **argv)
 {
        std::string frontend_command = "auto";
```

Sorry for the extra review for this!

mmicko: Yes, please do this change and also remove changes to yosys.h file

mmicko:
Merged as part of https://github.com/YosysHQ/yosys/pull/3423

Issue #3416, opened 2022-07-19
nexus: memory_libmap pass produces unusable bitstream


danc86:
For the design I am working on, targetting Crosslink-NX17 (nexus family), the latest master branch of yosys (commit 08c319fc352fb2670b7416b5fb16ddcb9a400049) produces a bitstream that seems to be rejected by the device. I bisected the problem to commit 0a8eaca322a622610efe5a0d33d1cda83e4afa8c, which introduced the new memory_libmap pass for nexus.

The design uses explicit DPSC512K instances (LRAMs) and inferred PDPSC16K instances (EBRAMs). It seems like the memory_libmap pass is doing something invalid with those.

## Steps to reproduce the issue

1. Unpack [yosys-memory_libmap-bug-repro.zip](https://github.com/YosysHQ/yosys/files/9137249/yosys-memory_libmap-bug-repro.zip) into a new directory
2. `yosys hps_platform.ys`

Sorry I haven't minimized this reproducer yet, but just comparing the output between the two versions of yosys, bad (master branch) and good (master branch + revert of commit 0a8eaca322a622610efe5a0d33d1cda83e4afa8c), might be enough to find the problem.

## Expected behavior

Resulting bitstream should work :-)

## Actual behavior

nextpnr-nexus accepts the design and produces a bitstream successfully but it is rejected by the device.

danc86:
@tmichalak would Antmicro be able to help solve this issue?

Comparing the Verilog output from yosys in the good and bad cases, it seems the DPSC512K instances are the same in both cases, so it seems like the problem is more likely with the inferred PDPSC16K instances. I see that they are arranged and wired up quite differently between the good and bad cases. I wasn't able to spot anything immediately wrong but it might be worth a closer look.

gatecat:
This could well be an oxide or nextpnr issue too, I will try and have a look as well.

tmichalak:
@danc86 sure, we can take a look.

gatecat:
See my WIP in #3417

danc86:
I wanted to try the fixes from #3420 but it seems to cause an assertion failure in nextpnr. I'm testing with latest master branch: yosys c26b2bf543a226e65a3fb07040bb278d668accf2, nextpnr b653e39991429dc3350447f8b69030c454dd2ba7, prjoxide b6bd4bd33cacef4b7b0ff8a4137ee7e441b6bc15.

```
[...]

PR #3417, opened 2022-07-19
nexus: Fix BRAM init and CSDECODE

gatecat:
This fixes at least some of the issues in #3416. It also needs YosysHQ/nextpnr#1005 to not crash nextpnr.

I still haven't checked everything post-libmap, and I'm also not 100% sure whether my approach to INIT handling is correct for mixed port widths...

### DIFF
--- a/techlibs/nexus/brams_map.v
+++ b/techlibs/nexus/brams_map.v
@@ -32,10 +32,16 @@ output [PORT_B_WIDTH-1:0] PORT_B_RD_DATA;
 
 function [319:0] init_slice;
 	input integer idx;
-	integer i, j;
+	integer i;
 	init_slice = 0;
-	for (i = 0; i < 16; i = i + 1) begin
-		init_slice[i*20+:18] = INIT[(idx * 16 + i) * 18+:18];
+	begin
+		if (PORT_A_WIDTH <= 4) begin
+			for (i = 0; i < 32; i = i + 1'b1)
+				init_slice[i * 10 +: 10] = {2'b00, INIT[(idx * 32 + i) * 8 +: 8]};
+		end else begin
+			for (i = 0; i < 32; i = i + 1'b1)
+				init_slice[i * 10 +: 10] = {1'b0, INIT[(idx * 32 + i) * 9 +: 9]};
+		end
 	end
 endfunction
 
@@ -122,8 +128,8 @@ DP16K #(
 	.RESETMODE_B(PORT_B_OPTION_RESETMODE),
 	.ASYNC_RST_RELEASE_A(PORT_A_OPTION_RESETMODE),
 	.ASYNC_RST_RELEASE_B(PORT_B_OPTION_RESETMODE),
-	.CSDECODE_A("000"),
-	.CSDECODE_B("000"),
+	.CSDECODE_A("111"),
+	.CSDECODE_B("111"),
 	.GSR("DISABLED"),
 ) _TECHMAP_REPLACE_ (
 	.CLKA(PORT_A_CLK),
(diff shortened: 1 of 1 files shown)


gatecat:
There seems to be one remaining issue to get a LiteX type SoC working reliably, the address LSBs in nexus need to be padded with 1s not 0s. The following hack makes things work, this really needs to be a flag or something but I don't know the neatest option here:
```patch
diff --git a/passes/memory/memory_libmap.cc b/passes/memory/memory_libmap.cc
index 9e147b0bf..43a0cb8a2 100644
--- a/passes/memory/memory_libmap.cc
+++ b/passes/memory/memory_libmap.cc
@@ -1652,7 +1652,7 @@ void MemMapping::emit_port(const MemConfig &cfg, std::vector<Cell*> &cells, cons
                        hw_addr.append(addr[x]);
        }
        for (int i = 0; i < hw_wr_wide_log2 && i < hw_rd_wide_log2; i++)
-               hw_addr[i] = State::S0;
+               hw_addr[i] = State::S1;
        for (auto cell: cells)
                cell->setPort(stringf("\\PORT_%s_ADDR", name), hw_addr);
```

mwkmwkmwk:
The INIT stuff seems just plain wrong — `memory_libmap` already does what you're doing here, always providing INIT data in terms of largest available port width (ie. the INIT data you get for these primitives always involves 9-bit bytes)

mwkmwkmwk:
As for the address padding thing — that's the sort of things that IMO should be fixed up by a simple case over `PORT_*_WIDTH` in the mapping, like we already do for merging write enables in ECP5

gatecat:
> The INIT stuff seems just plain wrong — memory_libmap already does what you're doing here, always providing INIT data in terms of largest available port width (ie. the INIT data you get for these primitives always involves 9-bit bytes)

I think at least part of the PR is still needed, as INIT wasn't working properly before, but it sounds based on that the conditional on port widths <=4 is no longer needed with libmap.

> As for the address padding thing — that's the sort of things that IMO should be fixed up by a simple case over PORT_*_WIDTH in the mapping, like we already do for merging write enables in ECP5

Makes sense. What do you think about introducing a `ABITS` type parameter optionally set by libmap (like the old memory mapping stuff) to make doing the slicing and replacing a bit simpler?

mwkmwkmwk:
> > The INIT stuff seems just plain wrong — memory_libmap already does what you're doing here, always providing INIT data in terms of largest available port width (ie. the INIT data you get for these primitives always involves 9-bit bytes)
> 
> I think at least part of the PR is still needed, as INIT wasn't working properly before, but it sounds based on that the conditional on port widths <=4 is no longer needed with libmap.

oh right, the change from 18->20 to 9->10 expansion seems very much correct and necessary

> 
> > As for the address padding thing — that's the sort of things that IMO should be fixed up by a simple case over PORT_*_WIDTH in the mapping, like we already do for merging write enables in ECP5
> 
> Makes sense. What do you think about introducing a `ABITS` type parameter optionally set by libmap (like the old memory mapping stuff) to make doing the slicing and replacing a bit simpler?

I don't particularly like the extra redundancy, but I also don't like the cases that much...

We certainly need to have it be conditional on some option in libmap port description though (techmap doesn't react well to extra parameters set on the cell), do you have some proposed syntax?

gatecat:
> We certainly need to have it be conditional on some option in libmap port description though (techmap doesn't react well to extra parameters set on the cell), do you have some proposed syntax?

I don't really like it but one idea is something like a `set_params` in the port description with a set of allowed values for extra params to set (so `set_params ABITS` would add a extra `PORT_A_ABITS` parameter for example).

If it seems like solving it pure techmap might be better, then I think a solution involving `$clog2` will be workable for the one-padding, just need one conditional to deal with the >4 case of 9-bit bytes.

PR #3420, opened 2022-07-22
nexus: Fix BRAM mapping.

mwkmwkmwk:


mwkmwkmwk:
@gatecat I think this should fix #3416 with just techmap. Could you test and review it?

Issue #3410, opened 2022-07-12
verilog `resetall directive clears defined macros


imhcyx:
## Steps to reproduce the issue

Use `read_verilog -DMYMACRO resetall.v` to read the following file:

resetall.v:
```verilog
module foo;
`ifdef SYNTHESIS
initial $display("%m: SYNTHESIS defined");
`else
initial $display("%m: SYNTHESIS not defined");
`endif
`ifdef MYMACRO
initial $display("%m: MYMACRO defined");
`else
initial $display("%m: MYMACRO not defined");
`endif
endmodule

`resetall

module bar;
`ifdef SYNTHESIS
initial $display("%m: SYNTHESIS defined");
`else
initial $display("%m: SYNTHESIS not defined");
`endif
`ifdef MYMACRO
initial $display("%m: MYMACRO defined");
`else
initial $display("%m: MYMACRO not defined");
`endif
endmodule
```

## Actual behavior

```
1. Executing Verilog-2005 frontend: resetall.v
Parsing Verilog input from `resetall.v' to AST representation.
Generating RTLIL representation for module `\foo'.
foo: SYNTHESIS defined
foo: MYMACRO defined
Generating RTLIL representation for module `\bar'.
bar: SYNTHESIS not defined
bar: MYMACRO not defined
Successfully finished Verilog frontend.
```

## Expected behavior

In this example, a yosys-defined macro `SYNTHESIS` and a user-defined macro `MYMACRO` are used. After a `` `resetall`` directive, both macros are expected to be unaffected.

In Verilog-2005 spec §19.3, it says,
```
The text macro facility is not affected by the compiler directive `resetall.
```
So all macros should be preserved after a `` `resetall``.

PR #3443, opened 2022-08-10
resetall does not affect text defines, but undefineall does

mmicko:
According to specs text macros are not affected by `resetall but they are with `undefineall (which is SystemVerilog only)

This fixes https://github.com/YosysHQ/yosys/issues/3410


zachjs (approved):
Although `resetall` still isn't, e.g., resetting the `default_nettype`, this is a good step in the right direction.

mmicko:
@zachjs thanks, you are right, forgot we handled that one, other directives affected by resetall we do not support yet

PR #3445, opened 2022-08-16
dump runtime information for passes to json

nakengelhardt:
This adds the `-B <filename.json>` command line option to dump the execution time information to json file (same information as provided by -d option, but machine-readable for use in benchmarking).

### DIFF
--- a/kernel/driver.cc
+++ b/kernel/driver.cc
@@ -210,6 +210,7 @@ int main(int argc, char **argv)
 	std::string scriptfile = "";
 	std::string depsfile = "";
 	std::string topmodule = "";
+	std::string perffile = "";
 	bool scriptfile_tcl = false;
 	bool print_banner = true;
 	bool print_stats = true;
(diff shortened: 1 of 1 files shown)


kernel/driver.cc
```diff
@@ -674,6 +678,29 @@ int main(int argc, char **argv)
...
 		}
+		if(!perffile.empty())
+		{
+			FILE *f = fopen(perffile.c_str(), "wt");
+			if (f == nullptr)
+				log_error("Can't open performance log file for writing: %s\n", strerror(errno));
+
+			fprintf(f, "{\n");
+			fprintf(f, "  \"generator\": \"%s\",\n", yosys_version_str);
+			fprintf(f, "  \"total_ns\": %lu,\n", total_ns);
```
Ravenslofty, line 689: `#include <inttypes.h>` or whatever C++ wants, and then I think the portable solution is:
```suggestion
			fprintf(f, "  \"total_ns\": %" PRId64 ",\n", total_ns);
```

PR #3448, opened 2022-08-17
Instructions for people building on cygwin

davidsiaw:
I have recently built this on Cygwin and it works on clang but not on gcc.

In order for it to get working on clang, I had to remove the `-fPIC` option. I am not sure how critical it is, but it seems to work for me.

Also made a few small cleanups and added the clang requirements for anyone wanting to build with clang on Cygwin.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,7 @@ ENABLE_COVER := 1
 ENABLE_LIBYOSYS := 0
 ENABLE_PROTOBUF := 0
 ENABLE_ZLIB := 1
+ENABLE_FPIC := 1
 
 # python wrappers
 ENABLE_PYOSYS := 0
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -98,6 +99,10 @@ BISON ?= bison
 STRIP ?= strip
 AWK ?= awk
 
+ifeq ($(ENABLE_FPIC),1)
```
mmicko, line 102: Instead of adding new option I would suggest adding :
```
ifeq ($(findstring CYGWIN,$(OS)),)
CXXFLAGS += -fPIC
endif
```
You can also make sure it only do not add it in case of cygwin and clang, since not sure if fPIC is at all affecting cygwin gcc build or is being ignored

mmicko:
Cygwin support removed time ago, so closing this.

PR #3458, opened 2022-08-24
Improves ABC command runtime by 10-100x

QuantamHD:
After speaking with the author of ABC he let me know that ifraig is a very old command, and that &get; &fraig -x; &put is over 100x faster than ifraig with improved PPA results.

After making the change I confirmed that this is in fact a major speed up. On our internal designs in O(millions) of standard cells we saw multi hour reductions in runtime.

Also included is an improvement to the dress command. Using AIG based transformations removes the spec it SATs against. Providing the input blif will make sure that no matter what commands are run, the dress command can still do its job. I noticed a regression against some LUT mapping jobs that prompted me to fix this.

### DIFF
--- a/passes/techmap/abc.cc
+++ b/passes/techmap/abc.cc
@@ -29,11 +29,11 @@
 // Kahn, Arthur B. (1962), "Topological sorting of large networks", Communications of the ACM 5 (11): 558-562, doi:10.1145/368996.369025
 // http://en.wikipedia.org/wiki/Topological_sorting
 
-#define ABC_COMMAND_LIB "strash; ifraig; scorr; dc2; dretime; strash; &get -n; &dch -f; &nf {D}; &put"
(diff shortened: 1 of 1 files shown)


Ravenslofty:
Well, `&get -n` and `&put` transform the netlist to and from GIA manager, and that happens twice here (once for `&fraig -x`, and another for `&nf`). I'm wondering if there are GIA equivalents for the others, since I would imagine converting to and from twice is not the cheapest thing ever for logic on your scale.

For LUT mapping we generally recommend the `abc9` pass over `abc`, and this works exclusively in GIA manager.

QuantamHD:
To put some data to words:

This design is roughly 1 million standard cells. It takes 1600s to complete with this PR and 4800s without my changes. Across the board our designs take 50% less time to complete than they did without these changes.

## Total Profile

image

## ABC Portion Total

image

## ABC Zoomed In

image


## And & Get

In this case both &get and &put are 0.065% of the entire profile.

image

nakengelhardt:
That's pretty convincing!

PR #3459, opened 2022-08-24
Add support for subtract in preadder

gs-jgj:
Inspired in part by how it is done for xilinx_dsp48a


gussmith23:
Taking up this PR. Rebased and added a test. 

In testing this PR, I found that the pre-subtractor pattern (and thus, I think, the pre-adder pattern) is very restrictive in what it will match---specifically, bitwidths need to match exactly, with no extensions. Someone with a better understanding of pmgen can probably tell me whether this is actually true.

gs-jgj:
> In testing this PR, I found that the pre-subtractor pattern (and thus, I think, the pre-adder pattern) is very restrictive in what it will match

Yes, it was intentionally made very specific as to not produce wrong code. The goal was just to be able to somehow infer it.

Someone with a better understanding of how the system works could likely improve a lot in what it should be able to match.

Issue #3317, opened 2022-05-15
synth_ice40 -flowmap terminate called after throwing an instance of 'std::length_error'   what():  vector::reserve


corecode:
## Steps to reproduce the issue

```
read_verilog e.v
synth_ice40 -flowmap
```

e.v:
```verilog
module test
  (
   input wire [11:0] x,
   input wire [1:0] y,
   output wire [11:0] z,
   );

assign z = x >> y;

endmodule
```

## Expected behavior

No crash.

## Actual behavior

```
[...]
2.41. Executing FLOWMAP pass (pack LUTs with FlowMap).
Labeling cells.
Mapped to 24 LUTs with maximum depth 2.

Packing cells.
Packing test.$techmap$shr$/home/corecode/tmp/e.v:8$1.$auto$simplemap.cc:278:simplemap_mux$584.\Y (\z [11]).
  Packing test.$techmap$shr$/home/corecode/tmp/e.v:8$1.$auto$simplemap.cc:278:simplemap_mux$620.\Y ($techmap$shr$/home/corecode/tmp/e.v:8$1.$1\buffer[11:0] [11]).
terminate called after throwing an instance of 'std::length_error'
  what():  vector::reserve
zsh: IOT instruction (core dumped)  ./yosys
```

ColonelPhantom:
I have experienced the same issue. It seems to trace back to flowmap.cc:1409, where `minlut - input_nodes.size()` can return a negative number.

A workaround is to change the minlut parameter to 4 to equal the maxlut parameter, but quite frankly that's not ideal. I'm thinking maybe the expression should be something like `maxlut - input_node.size`, but I'm not familiar enough with the code of yosys to actually know if this will fix it, and whether it would actually be correct.

Issue #1206, opened 2019-07-17
Segmentation fault during proc_init with large wires


jfng:
## Steps to reproduce the issue

Yosys version:
`Yosys 0.8+609 (git sha1 56c00e87, clang 8.0.0 -fPIC -Os)`

Example RTLIL:
```
autoidx 1
attribute \top 1
module \top
  wire width 1073741824 $1
  wire width 1073741824 $2
  process $group_0
    sync init
  end
  connect $1 $2
end
```

Yosys script:
```
read_ilang example.il
proc_init
```

## Expected behavior

Either an error because wire width is too big, or a successful proc_init.

## Actual behavior

Yosys crashes with a segmentation fault during proc_init.

PR #3460, opened 2022-08-25
Enforce RTLIL Wire/Const limit size of 2^24

lethalbit:
This commit adds asserts to the RTLIL parser to ensure that a wire definition is no more than 2^24 bits wide, as well as that RTLIL `Const`s can not have a negative width nor be wider than 2^24 bits.

This addresses #1206 and #3317


Ravenslofty:
I think the correct thing to use here is `log_error` rather than throwing an exception. I doubt any code is going to catch it anyway.

lethalbit:
Possibly, I'll see how it feels with `log_error`, I also need to test to see how things like memories interact with the changes, although all the tests pass we don't have any that would directly exercise this code that I know of, so more work absolutely needs to be done on it.

nakengelhardt:
Looking at the surrounding code, +1 on `log_error`.

nakengelhardt:
We need to test where large consts may be created:
- memory initialization
- FSM transition tables
- others?

frontends/rtlil/rtlil_parser.y
```diff
@@ -187,7 +187,11 @@ wire_stmt:
 
 wire_options:
 	wire_options TOK_WIDTH TOK_INT {
-		current_wire->width = $3;
+		if ($3 > 0x1000000) {
```
arcturusannamalai, line 190: it maybe better to use a constexpr defined in rtlil.h instead of repeating the magic number everywhere; my 2-cents

kernel/rtlil.cc
```diff
@@ -214,6 +217,12 @@ RTLIL::Const::Const(const std::string &str)
 
 RTLIL::Const::Const(int val, int width)
 {
+	if (width < 0)
```
whitequark, line 220: This should probably be a `log_assert` rather than `log_error` since this case isn't supposed to ever happen.

widlarizer:
Since the RTLIL parser rewrite there's a 30 bit limit in parsing, so the action to take here would be to get inspired by this PR and add corresponding 30 bit asserts to all `Const`-constructing functions that log_assert the size to fit within that limit. The declaration for the size of the limit will be moved from the RTLIL parser to rtlil.h

PR #3494, opened 2022-09-28
Handle attributes imported from verific

mmicko:


mmicko:
@jix This is actually handled on Verific end. Attribute values are actually altered during parsing so in case it is value with specified width it will be properly padded and it will be in binary representation. So it is properly escaped string, binary value, 32 bit integer or unspecified value (set for strength of wire for example by Verific itself).

jix:
Ah, that's good! Ignore my comment about parsing then, I was under the wrong impression that Verific passes on the attribute unprocessed.

PR #3496, opened 2022-09-29
Rst docs conversion

KrystalDelusion:
Converting the Yosys Manual to use RST instead of pure LaTeX


KrystalDelusion:
Rebased onto master, removed unnecessary commits.

PR #3504, opened 2022-10-09
Temporal induction counterexample loop detection

ekliptik:
I have added an optional flag to smtbmc that causes failed temporal induction counterexample traces to be checked for duplicate states and reported to the user, since loops in the counterexample mean that increasing the induction depth won't help prove a design's safety properties. Designs containing memory are skipped for now. I'm open to suggestions of any kind

### DIFF
--- a/backends/smt2/smtbmc.py
+++ b/backends/smt2/smtbmc.py
@@ -993,7 +993,7 @@ def detect_state_loop(steps_start, steps_stop):
             [(tuple(path), addr, data)
              for path, addr, data in mem_trace_data.get(i, [])])
         state = tuple(value_list), tuple(mem_state)
-        if states.get(state):
+        if state in states:
             return (i, states[state])
         else:
             states[state] = i
--- a/backends/smt2/smtbmc.py
+++ b/backends/smt2/smtbmc.py
@@ -55,6 +55,7 @@ noinit = False
 binarymode = False
 keep_going = False
 check_witness = False
+find_loops = False
 so = SmtOpts()
 
 
@@ -175,6 +176,9 @@ def usage():
         check that the used witness file contains sufficient
         constraints to force an assertion failure.
 
+    --find-loops
+        check if states are unique in temporal induction counterexamples
+
 """ + so.helpmsg())
     sys.exit(1)
 
@@ -183,7 +187,7 @@ try:
     opts, args = getopt.getopt(sys.argv[1:], so.shortopts + "t:igcm:", so.longopts +
             ["final-only", "assume-skipped=", "smtc=", "cex=", "aig=", "aig-noheader", "yw=", "btorwit=", "presat",
              "dump-vcd=", "dump-yw=", "dump-vlogtb=", "vlogtb-top=", "dump-smtc=", "dump-all", "noinfo", "append=",
-             "smtc-init", "smtc-top=", "noinit", "binary", "keep-going", "check-witness"])
+             "smtc-init", "smtc-top=", "noinit", "binary", "keep-going", "check-witness", "find-loops"])
 except:
     usage()
 
@@ -264,6 +268,8 @@ for o, a in opts:
         keep_going = True
     elif o == "--check-witness":
         check_witness = True
+    elif o == "--find-loops":
+        find_loops = True
     elif so.handle(o, a):
         pass
     else:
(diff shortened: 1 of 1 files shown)


backends/smt2/smtbmc.py
```diff
@@ -969,6 +975,38 @@ def write_vcd_trace(steps_start, steps_stop, index):
...
+    path_tupled_list = [tuple(p) for p in path_list]
+    for i in range(steps_start, steps_stop):
+        value_list = smt.get_net_bin_list(topmod, path_list, "s%d" % i)
+        state = tuple(zip(path_tupled_list, value_list))
+        if states.get(state):
+            return (i, states[state])
+        else:
+            states[state] = i
+
+        if i in mem_trace_data:
```
jix, line 1003: This check for bailing out has to go above the loop-detection part, otherwise you will get false positives when a design with memory repeats the same non-memory state with changed memory contents.

Also I think using
```python
mem_state = sorted(
    (tuple(path), addr, data) for path, addr, data in words)
    for words in mem_trace_data.get(i, [])
)
state = tuple(value_list), tuple(mem_state)
```
above might be all that's needed to also support memory. I haven't tested that at all yet. Even if this works it's ok to add memory support later.

backends/smt2/smtbmc.py
```diff
@@ -1596,6 +1634,10 @@ def smt_check_sat(expected=["sat", "unsat"]):
                 print_anyconsts(num_steps)
                 print_failed_asserts(num_steps)
                 write_trace(step, num_steps+1, '%', allregs=True)
+                if find_loops:
+                    loop = find_state_loop(step, num_steps+1)
+                    if loop:
+                        print_msg(f"Loop detected, increasing induction depth will not help. Cycle {loop[0]} = cycle {loop[1]}")
```
jix, line 1640: smtbmc uses "step" instead of "cycle" for this. As a step doesn't have to correspond to a clock cycle (when using `clk2fflogic`) that's more precise.

jix (changes requested):
Thank you for contributing this! A few minor changes are necessary, see the inline comments. Beyond that, I also left some suggestions for optional improvements if you want to continue work on this.

backends/smt2/smtbmc.py
```diff
@@ -175,6 +176,9 @@ def usage():
         check that the used witness file contains sufficient
         constraints to force an assertion failure.
 
+    --find-loops
```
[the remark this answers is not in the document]
ekliptik: Implemented fully

backends/smt2/smtbmc.py
```diff
@@ -969,6 +975,38 @@ def write_vcd_trace(steps_start, steps_stop, index):
 
         vcd.set_time(steps_stop)
 
+def find_state_loop(steps_start, steps_stop):
+    print_msg(f"Looking for loops")
```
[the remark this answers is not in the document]
ekliptik: Implemented fully

backends/smt2/smtbmc.py
```diff
@@ -969,6 +975,38 @@ def write_vcd_trace(steps_start, steps_stop, index):
...
+
+    path_list = list()
+
+    for netpath in sorted(smt.hiernets(topmod)):
+        hidden_net = False
+        for n in netpath:
+            if n.startswith("$"):
+                hidden_net = True
+        if not hidden_net:
```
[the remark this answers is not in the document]
ekliptik: I picked the easy way, because it seems like handling a strange case that would take some time to write tests for, requiring more than my way of "mess around with examples/smtbmc and see if it does what I think it should". Should I open an issue?

backends/smt2/smtbmc.py
```diff
@@ -969,6 +975,38 @@ def write_vcd_trace(steps_start, steps_stop, index):
...
+        value_list = smt.get_net_bin_list(topmod, path_list, "s%d" % i)
+        state = tuple(zip(path_tupled_list, value_list))
+        if states.get(state):
+            return (i, states[state])
+        else:
+            states[state] = i
+
+        if i in mem_trace_data:
+            if mem_trace_data[i]:
+                print_msg(f"This design contains memory. Searching for counterexample loops will be skipped")
```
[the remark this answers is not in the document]
ekliptik: Not applicable anymore since I implemented memory uniqueness

backends/smt2/smtbmc.py
```diff
@@ -969,6 +975,38 @@ def write_vcd_trace(steps_start, steps_stop, index):
...
+        if not hidden_net:
+            path_list.append(netpath)
+
+    mem_trace_data = collect_mem_trace_data(steps_start, steps_stop)
+
+    states = dict()
+    path_tupled_list = [tuple(p) for p in path_list]
+    for i in range(steps_start, steps_stop):
+        value_list = smt.get_net_bin_list(topmod, path_list, "s%d" % i)
+        state = tuple(zip(path_tupled_list, value_list))
```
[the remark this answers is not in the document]
ekliptik: Implemented fully

backends/smt2/smtbmc.py
ekliptik: Implemented fully, tested on modified examples to a limited extent

backends/smt2/smtbmc.py
ekliptik: Implemented fully

backends/smt2/smtbmc.py
```diff
@@ -969,6 +976,30 @@ def write_vcd_trace(steps_start, steps_stop, index):
...
+    # Map state to index of step when it occurred
+    states = dict()
+
+    for i in range(steps_start, steps_stop):
+        value_list = smt.get_net_bin_list(topmod, path_list, "s%d" % i)
+        mem_state = sorted(
+            [(tuple(path), addr, data)
+             for path, addr, data in mem_trace_data.get(i, [])])
+        state = tuple(value_list), tuple(mem_state)
+        if states.get(state):
```
jix, line 996: Missed this the first time: When `steps_start` is 0 we can end up with a state at time step 0 which is false in a boolean context and thus a loop with that state would be ignored here. Use `if state in states:` instead.

jix (changes requested):
I found one tiny issue I missed in the first round. Apart from that it looks good to me now and I can merge this after that's addressed.

backends/smt2/smtbmc.py
ekliptik: certified 🐍 moment

PR #3507, opened 2022-10-09
Fix handling of verific -L options, add implicit "-L work"

clairexen:
As discussed on Friday: Add an implicit `-L work` at the beginning of the Verific -L option list, unless `-L work` is already present at any (other) position in the list. (That exception enables the user to specify a library with module implementations that should replace the implementations that come with the design, by specifying that library first, followed by `-L work`.)


mmicko (approved):
LGTM

PR #3508, opened 2022-10-10
backends: protobuf: removed protobuf backend

lethalbit:
This PR removes the protobuf backend as discussed in the Yosys Dev JF on 2022-10-10, as it seems to have been unused.


mmicko (approved):
LGTM

PR #3513, opened 2022-10-17
Set BISON when building with brew on mac

robtaylor:
When building on Mac M2 with brew, I hit an issue with build picking up the wrong bison. This fixes that issue.


mmicko:
Hi @robtaylor, LDFLAGS change is not needed since we do not link bison library (and your change anyway have typo, so it is not properly set). Not sure what is set in your environment, but please check if you have manually set BISON inside your Makefile.conf, since value defaults to "bison" and in line 116 PATH is already set fine, so script works without any change for others.

Issue #3520, opened 2022-10-21
Can't resolve function name


flaviens:
### Version

Yosys 0.22+29 (git sha1 12267a098, gcc 11.2.0 -fPIC -Os)

### On which OS did this happen?

Linux

### Reproduction Steps

Requires Yosys and sv2v (tested with `sv2v v0.0.10-1-gc00f508`).

```
git clone https://github.com/bsg-external/HardFloat.git
cd HardFloat/source
sv2v -IRISCV mulAddRecFN.v -w sv2v_out.v
yosys -p 'read_verilog -defer -sv sv2v_out.v; hierarchy -top mulAddRecFNToRaw -check'
```

### Expected Behavior

No error related to `mulAddRecFNToRaw.clog2`.

### Actual Behavior

```
 /----------------------------------------------------------------------------\
 |                                                                            |
 |  yosys -- Yosys Open SYnthesis Suite                                       |
 |                                                                            |
 |  Copyright (C) 2012 - 2020  Claire Xenia Wolf <claire@yosyshq.com>         |
 |                                                                            |
 |  Permission to use, copy, modify, and/or distribute this software for any  |
 |  purpose with or without fee is hereby granted, provided that the above    |
 |  copyright notice and this permission notice appear in all copies.         |
 |                                                                            |
 |  THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES  |
 |  WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF          |
 |  MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR   |
 |  ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES    |
 |  WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN     |
 |  ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF   |
 |  OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.            |
 |                                                                            |
 \----------------------------------------------------------------------------/

 Yosys 0.22+29 (git sha1 12267a098, gcc 11.2.0 -fPIC -Os)


[...]

PR #3532, opened 2022-10-29
verilog: Support module-scoped task/function calls

zachjs:
I noticed this issue while investigating https://github.com/zachjs/sv2v/issues/223 / https://github.com/YosysHQ/yosys/issues/3520. This change is primarily intended to enable the standard-permitted use of module-scoped identifiers to refer to tasks and non-constant functions. As a side-effect, this also adds support for the non-standard use of module-scoped identifiers referring to constant functions, a feature that is supported in some other tools, including Iverilog.


jix (approved):
lgtm

PR #3536, opened 2022-11-01
Add extra time at the end of a sat VCD trace

clairexen:
Otherwise the final values will not show up in gtkwave waveforms when looking at the generated traces.

Signed-off-by: Claire Xenia Wolf <claire@clairexen.net>


mmicko (approved):
LGTM, would be good to check how it affects loading VCD and generating a new one with `sim` from it

PR #3537, opened 2022-11-02
New xprop pass

jix:
Current WIP of the new xprop pass and preparatory changes.

The xprop pass itself needs more celltypes and more tests, but the current version can handle nerv after pmuxtree and dffunmap.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -881,6 +881,7 @@ test: $(TARGETS) $(EXTRA_TARGETS)
 	+cd tests/rpc && bash run-test.sh
 	+cd tests/memfile && bash run-test.sh
 	+cd tests/verilog && bash run-test.sh
+	+cd tests/xprop && bash run-test.sh $(SEEDOPT)
 	@echo ""
 	@echo "  Passed \"make test\"."
 	@echo ""
--- a/backends/btor/btor.cc
+++ b/backends/btor/btor.cc
@@ -1417,6 +1417,7 @@ struct BtorBackend : public Backend {
 		log_push();
 		Pass::call(design, "bmuxmap");
 		Pass::call(design, "demuxmap");
+		Pass::call(design, "bwmuxmap");
 		log_pop();
 
 		size_t argidx;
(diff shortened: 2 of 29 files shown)


georgerennie:
You may already be aware of this but I thought I would mention just in case not. Unless I have misunderstood how this pass works, this approach of emulating 3-state logic with 2-state logic is well suited to simulation, but does not have the most accurate X semantics that you can achieve with formal.
If you take the following example

```verilog
read_verilog -sv <<EOT
module top(
	output logic b
);

logic a;

assign a = 1'bx;
assign b = a || !a;

always assert(b !== 'x);

endmodule

EOT

proc -noopt
xprop -formal -required -split-outputs
sat -prove-asserts
```

This results in a counterexample, even though b can only ever be 1. Whether this is correct or not depends on how you interpret X in this scenario (and there a lot of ways of interpreting x propagation once you get into optimism/pessimism lmao).

I only bring this up, because Jasper's XPROP mode more accurately checks if Xs can propagate, by creating a miter with two clones of the circuit, and defines Xs as anywhere the two circuits can diverge (at least this is what I assume, I have not seen inside). This same technique can be applied to other things like SPV (security path verification), where you constrain all x sources to be the same between the two circuits, and create a cutpoint at an information source, then see if the difference in the two circuits due to this can propagate to a checker. (Equivalently, the existing glift pass does similar to this xprop pass but for information flow/path properties I believe).

I double checked the above example in Jasper XPROP mode and the property passes (modified slightly to work with Jasper).

I briefly started implementing the Jasper style XPROP earlier this year in https://github.com/georgerennie/yosys/tree/sensitizepath, hoping to add SPV style support too, but havent had much time so that code is a half finished mess.

Feel free to ignore all of this as there are pros and cons to both the 2-valued emulation and miter based approaches but I thought it worth mentioning just in case it hadn't come up as users of Jasper may expect this xprop to act the same as Jasper's XPROP when it doesn't.

clairexen (approved):

PR #3538, opened 2022-11-03
opt_muxtree: set iterations limit based on number of muxes

CAFxX:
I sometime hit the `Giving up (too many iterations)` error in some large designs.

Untested, consider this a RFC.

### DIFF
--- a/passes/opt/opt_muxtree.cc
+++ b/passes/opt/opt_muxtree.cc
@@ -36,7 +36,7 @@ struct OptMuxtreeWorker
 	RTLIL::Module *module;
 	SigMap assign_map;
 	int removed_count;
-	int glob_abort_cnt = 100000;
+	int glob_abort_cnt;
 
 	struct bitinfo_t {
 		bool seen_non_mux;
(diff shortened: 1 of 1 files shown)


nakengelhardt:
Thanks! We discussed it a little and came up with a variant that shouldn't affect smaller designs but be a bit more strict at catching grossly nonlinear behavior on large designs. I pushed that to your PR, hope you don't mind!

CAFxX:
NP, just one small note though: I ran some tests on a local build of the PR and with a multiplier of 100 I still run into the problem; with a multiplier of 1000 the limit is not hit.

passes/opt/opt_muxtree.cc
```diff
@@ -146,6 +146,10 @@ struct OptMuxtreeWorker
 			log("  No muxes found in this module.\n");
 			return;
 		}
+		
+		// Limit the number of iterations based on the number of muxes.
+		// The multiplier (1000) is arbitrary.
+		glob_abort_cnt = max(100000, GetSize(mux2info) * 100);
```
CAFxX, line 152:
```suggestion
		// The multiplier (1000) is somewhat arbitrary, but smaller values cause the
		// iteration limit to be hit on large designs.
		glob_abort_cnt = max(100000, GetSize(mux2info) * 1000);
```

nakengelhardt:
That might be that your design is one of the cases of grossly nonlinear behavior we want to catch though - how much execution time difference do you have for `* 100` vs `* 1000` (both in minutes and in % of total time)?

widlarizer:
In #5329 the cap was raised, we can't verify if there's a point to scaling with the mux count at this point, so I'm closing it but willing to reopen of we get a reproducer that justifies this scaling or something

PR #3542, opened 2022-11-05
ice40: Fix path delay definitions

Riesi:
I tried to simulate an ice40 synthesized Verilog design with path delays in iverilog. After finding the correct flags, iverilog threw some error when reading `ice40/cells_sim.v`.

I did some digging on how path delays are defined in Verilog and came to the following conclusion.
Parallel connections do not allow matching different bit widths. A full connection has to be used instead.
This PR allows iverilog to parse the simulation library with hardware path delays enabled.

#### Section 30.4.2 - Simple module paths 
`The operator *> establishes a full connection between source and destination. The operator => establishes a parallel connection between source and destination.`
#### Section 30.4.5 - Full connection and parallel connection paths
Full connection:
` In a full connection, every bit in the source shall connect to every bit in the destination. The module path source need not have the same number of bits as the module path destination.`
Parallel conncetion:
`Parallel module paths can be created only between sources and destinations that contain the same number of bits.`

# How to reproduce:
```
iverilog \
		-gspecify \
		-D 'NO_ICE40_DEFAULT_ASSIGNMENTS=1' \
		-D 'ICE40_HX=1' \
		$(yosys-config --datdir/ice40/cells_sim.v)
```
Errors:
```
/usr/share/yosys/ice40/cells_sim.v:1677: error: For a parallel connection the source/destination width must match found (1/16).
/usr/share/yosys/ice40/cells_sim.v:1677: error: For a parallel connection the source/destination width must match found (1/16).
/usr/share/yosys/ice40/cells_sim.v:1813: error: For a parallel connection the source/destination width must match found (1/16).
/usr/share/yosys/ice40/cells_sim.v:1677: error: For a parallel connection the source/destination width must match found (1/16).
/usr/share/yosys/ice40/cells_sim.v:2085: error: For a parallel connection the source/destination width must match found (1/16).
/usr/share/yosys/ice40/cells_sim.v:1677: error: For a parallel connection the source/destination width must match found (1/16).
/usr/share/yosys/ice40/cells_sim.v:1949: error: For a parallel connection the source/destination width must match found (1/16).
```
Similar errors occur also when replacing `ICE40_HX` with `ICE40_LP` or `ICE40_U`, so these have also been fixed.

### DIFF
--- a/techlibs/ice40/cells_sim.v
+++ b/techlibs/ice40/cells_sim.v
@@ -1674,7 +1674,7 @@ module SB_RAM40_4K (
 		// https://github.com/YosysHQ/icestorm/blob/95949315364f8d9b0c693386aefadf44b28e2cf6/icefuzz/timings_hx1k.txt#L400
 		$setup(WE, posedge WCLK, 133);
 		// https://github.com/YosysHQ/icestorm/blob/95949315364f8d9b0c693386aefadf44b28e2cf6/icefuzz/timings_hx1k.txt#L401
-		(posedge RCLK => (RDATA : 16'bx)) = 2146;
+		(posedge RCLK *> (RDATA : 16'bx)) = 2146;
 	endspecify
 `endif
 `ifdef ICE40_LP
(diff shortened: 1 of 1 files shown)


Riesi:
anything I can do to help resolve this issue otherwise, if my PR isnt the way forward?

whitequark (approved):
Looks reasonable to me

PR #3551, opened 2022-11-12
Support for arrays with swapped ranges within structs

### LINKED ISSUE #3550: Little endian indexing of packed multidimensional arrays is incorrectly assumed to be big endian
### Version

Yosys 0.22+1 (git sha1 c4a52b1b028, gcc 12.2.1 -O2 -fexceptions -fstack-protector-strong -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -fPIC -Os)

Linux

### Reproduction Steps

I have added a test to tests/svtypes/struct_array.sv to demonstrate that indexing misbehaves - see diff below:
[struct_array_le.diff.gz](https://github.com/YosysHQ/yosys/files/9983714/struct_array_le.diff.gz)

The asserts have been verified with Verilator.
Note that iverilog has an off-by-one error in little endian array slices, and can thus not be used for verification.

Replacing the `initial` block for assigments to `s3` with the output of sv2v makes the test pass:

```
	initial begin
		s3 = 1'sb0;
		s3[24+:16] = 16'h1234;
		s3[56+:8] = 8'h42;
		s3[72+:8] = 1'sb1;
		s3[73-:2] = 1'sb0;
		s3[15-:16] = 1'sb1;
		s3[1:0] = 1'sb0;
	end
```

### Expected Behavior

The augmented array test passes.

### Actual Behavior

The array test fails.

daglem:
This also corrects the implementation of C type arrays within structs.

Fixes #3550

### DIFF
--- a/frontends/ast/ast.h
+++ b/frontends/ast/ast.h
@@ -203,7 +203,7 @@ namespace AST
 
 		// if this is a multirange memory then this vector contains offset and length of each dimension
 		std::vector<int> multirange_dimensions;
-		std::vector<bool> multirange_swapped; // true if range is swapped, not used for structs
+		std::vector<bool> multirange_swapped; // true if range is swapped
 
 		// this is set by simplify and used during RTLIL generation
 		AstNode *id2ast;
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -285,10 +285,16 @@ static void save_struct_array_width(AstNode *node, int width)
 
 }
 
+static void save_struct_range_swapped(AstNode *node, bool range_swapped)
+{
+	node->multirange_swapped.push_back(range_swapped);
+
+}
+
 static int get_struct_array_width(AstNode *node)
 {
 	// the stride for the array, 1 if not an array
-	return (node->multirange_dimensions.empty() ? 1 : node->multirange_dimensions.back());
+	return (node->multirange_dimensions.size() != 2 ? 1 : node->multirange_dimensions[1]);
 
 }
 
@@ -321,18 +327,22 @@ static int size_packed_struct(AstNode *snode, int base_offset)
 					if (node->children[1]->type == AST_RANGE) {
 						// unpacked array e.g. bit [63:0] a [0:3]
 						auto rnode = node->children[1];
-						int array_count = range_width(node, rnode);
-						if (array_count == 1) {
-							// C-type array size e.g. bit [63:0] a [4]
-							array_count = rnode->range_left;
-						}
+						// C-type array size e.g. bit [63:0] a [4]
+						bool c_type = rnode->children.size() == 1;
+						int array_count = c_type ? rnode->range_left : range_width(node, rnode);
+						save_struct_array_width(node, array_count);
+						save_struct_range_swapped(node, rnode->range_swapped || c_type);
 						save_struct_array_width(node, width);
+						save_struct_range_swapped(node, node->children[0]->range_swapped);
 						width *= array_count;
 					}
 					else {
 						// array element must be single bit for a packed array
 						struct_array_packing_error(node);
 					}
+				} else {
+					save_struct_array_width(node, width);
+					save_struct_range_swapped(node, node->children[0]->range_swapped);
 				}
 				// range nodes are now redundant
 				for (AstNode *child : node->children)
@@ -347,8 +357,11 @@ static int size_packed_struct(AstNode *snode, int base_offset)
 					struct_array_packing_error(node);
 				}
 				int array_count = range_width(node, rnode->children[0]);
+				save_struct_array_width(node, array_count);
+				save_struct_range_swapped(node, rnode->children[0]->range_swapped);
 				width = range_width(node, rnode->children[1]);
 				save_struct_array_width(node, width);
+				save_struct_range_swapped(node, rnode->children[1]->range_swapped);
 				width *= array_count;
 				// range nodes are now redundant
 				for (AstNode *child : node->children)
@@ -428,8 +441,16 @@ static AstNode *offset_indexed_range(int offset, int stride, AstNode *left_expr,
 	return new AstNode(AST_RANGE, left, right);
 }
 
-static AstNode *make_struct_index_range(AstNode *node, AstNode *rnode, int stride, int offset)
+static AstNode *make_struct_index_range(AstNode *node, AstNode *rnode, int stride, int offset, AstNode *member_node)
 {
+	if (member_node->multirange_swapped[0]) {
+		// The struct item has swapped range; swap index into the struct accordingly.
+		int msb = member_node->multirange_dimensions[0] - 1;
+		for (auto &expr : rnode->children) {
+			expr = new AstNode(AST_SUB, node_int(msb), expr);
+		}
+	}
+
 	// generate a range node to perform either bit or array indexing
 	if (rnode->children.size() == 1) {
 		// index e.g. s.a[i]
@@ -447,6 +468,7 @@ static AstNode *make_struct_index_range(AstNode *node, AstNode *rnode, int strid
 static AstNode *slice_range(AstNode *rnode, AstNode *snode)
 {
 	// apply the bit slice indicated by snode to the range rnode
+	// TODO: Check for swapped indexes - see make_struct_index_range
 	log_assert(rnode->type==AST_RANGE);
 	auto left  = rnode->children[0];
 	auto right = rnode->children[1];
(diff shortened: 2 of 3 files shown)


udif:
> Is this something that is explicitly supported by any other tools? If not, I think it would be best to also reject this in Yosys.

Unsurprisingly, Cadence Incisive simulator and Aldec Riviera Pro agrees with you, and reject an unpacked array in a packed struct.

daglem:
Unfortunately I didn't check the spec on that, I simply expanded on (incorrect) code in simplify.cc

I agree that unpacked arrays in packed structs should be forbidden by Yosys, however this is a separate issue to this PR, IMO.

I haven't made any changes to the Yosys parser here, so would simply removing the tests for the unpacked arrays be sufficient?

jix:
I agree that support for unpacked arrays in packed structs is a separate issue, especially after realizing that it might have been intentional as there is no support for unpacked structs at all.

Given that this PR does also change the implementation for unpacked arrays in packed structs, I think it's best to keep the tests, but enclosed in ``` `ifndef VERIFIC``` together with the already existing comment explaining the non-standard Yosys behavior.

While reviewing the changes, I did some more testing and noticed that there is a disagreement for the following case between this PR and the verific frontend:

```systemverilog
	struct packed {
		bit [0:7] [0:7] a;	// 8 element packed array of bytes
		bit [0:15] b;		// filler for non-zero offset
	} s3_b;

	initial begin
		s3_b = '0;

		s3_b.a[5:6] = 16'h1234;
		s3_b.a[2] = 8'h42;

		s3_b.a[0] = '1;
		s3_b.a[0][0:1] = '0;

		s3_b.b = '1;
		s3_b.b[14:15] = '0;
	end

	always_comb assert(s3_b==80'h3F00_4200_0012_3400_FFFC);
```

This passes with verific but not with this PR. My understanding is that this should pass, but it might be a good idea to double check the spec.

daglem:
According to my understanding of the spec, the test should pass, and Verilator agrees with this. Icarus Verilog fails the test, but is clearly wrong whichever way you look at it (I have made a PR for that as well).

@udif I don't have access to any commercial tools, perhaps you can confirm using your tools?

@jix I'm not sure if I understand you correctly - does the test (which is part of this PR) unconditionally fail with Verific, or does it only fail if the rest of the PR is included?

jix:
There are two completely separate issues:

1. The test cases in this PR below the `// Note that the tests below for unpacked arrays ...` comment fail unconditionally with the Verific frontend due to the non-standard use of unpacked arrays in packed structs. Given that this is currently supported by Yosys I would still keep the tests until/unless a decision and code changes are made to not support this. To not break the tests for Yosys builds with the Verific frontend they should be wrapped in an ifndef. All the other tests added by this PR pass with both the Verific frontend and the Yosys frontend with this PR.
2. While reviewing the code changes, it wasn't directly clear to me how multiple dimensions are handled, so to better understand what's going on, I tried further test cases not among the ones added by this PR. While doing so I discovered the one in my above comment using `s3_b`. It uses LSB-first for both ranges unlike the existing `s3` test case which uses LSB-first for the first range and MSB-first for the second. Since this is all packed, the `s3_b` test case works fine with Verific, but the Yosys frontend with this PR applied produces a different result, one that I think is still incorrect.

I hope that makes things more clear, then before when I was leaving comments as I was testing.

udif:
> According to my understanding of the spec, the test should pass, and Verilator agrees with this. Icarus Verilog fails the test, but is clearly wrong whichever way you look at it (I have made a PR for that as well).
> 
> @udif I don't have access to any commercial tools, perhaps you can confirm using your tools?
> 
> @jix I'm not sure if I understand you correctly - does the test (which is part of this PR) unconditionally fail with Verific, or does it only fail if the rest of the PR is included?



> @udif I don't have access to any commercial tools, perhaps you can confirm using your tools?

My tools are at work, but EDA playground says the code snippet from above passes with Aldec Riviera Pro:
```
module t;
struct packed {
		bit [0:7] [0:7] a;	// 8 element packed array of bytes
		bit [0:15] b;		// filler for non-zero offset
	} s3_b;

	initial begin
		s3_b = '0;

		s3_b.a[5:6] = 16'h1234;
		s3_b.a[2] = 8'h42;

		s3_b.a[0] = '1;
		s3_b.a[0][0:1] = '0;

		s3_b.b = '1;
		s3_b.b[14:15] = '0;
	end

	always_comb assert(s3_b==80'h3F00_4200_0012_3400_FFFC);
endmodule
```

daglem:
@jix It's all clear to me now, I simply didn't catch that you had written a new test :see_no_evil: 

I think you may have found yet another issue. If I'm guessing correctly, `bit [7:0] [0:7] a;` won't work correctly either, with or without this PR.

I'll see whether I can fix this issue as well.

daglem:
@jix Should be fixed now. I had actually left a TODO in simplify.cc for this, which I had already forgotten about :see_no_evil: 

I added your test, I only changed the slicing and the test bit pattern to match the other tests.

daglem:
@jix Can this be merged now?

jix:
Before realizing that other parts of our frontend currently limits packed arrays in structs to two dimensions, I was concerned about this code not working correctly with more than two dimensions. Given that the number of dimensions is not limited by SV itself, this restriction might be lifted at some point. To avoid this silently breaking at that point, can you add an assertion that checks for at most two dimensions when processing the `AST_MULTIRANGE` node together with a comment stating that the slice handling code only supports up to two dimensions currently?

Otherwise this looks good to me.

Just for future reference, the following tests should pass when that restriction is lifted:

```
	struct packed {
		bit [0:7] [0:1] [0:3] a;
		bit [0:15] b;		// filler for non-zero offset
	} s3_lll;

	initial begin
		s3_lll = '0;

		s3_lll.a[5:6] = 16'h1234;
		s3_lll.a[2] = 8'h42;

		s3_lll.a[0] = '1;
		s3_lll.a[0][1][2:3] = '0;

		s3_lll.b = '1;
		s3_lll.b[14:15] = '0;
	end

	always_comb assert(s3_lll==80'hFC00_4200_0012_3400_FFFC);

	struct packed {
		bit [0:7] [1:0] [0:3] a;
		bit [0:15] b;		// filler for non-zero offset
	} s3_lbl;

	initial begin
		s3_lbl = '0;

		s3_lbl.a[5:6] = 16'h1234;
		s3_lbl.a[2] = 8'h42;

		s3_lbl.a[0] = '1;
		s3_lbl.a[0][0][2:3] = '0;

		s3_lbl.b = '1;
		s3_lbl.b[14:15] = '0;
	end

	always_comb assert(s3_lbl==80'hFC00_4200_0012_3400_FFFC);

	struct packed {
		bit [0:7] [0:1] [3:0] a;
		bit [0:15] b;		// filler for non-zero offset
	} s3_llb;

	initial begin
		s3_llb = '0;

		s3_llb.a[5:6] = 16'h1234;
		s3_llb.a[2] = 8'h42;

		s3_llb.a[0] = '1;
		s3_llb.a[0][1][1:0] = '0;

		s3_llb.b = '1;
		s3_llb.b[14:15] = '0;
	end

	always_comb assert(s3_llb==80'hFC00_4200_0012_3400_FFFC);
```

daglem:
I am aware that the limitation of two dimensions is not a SystemVerilog limitation, and so I actually made an effort to simplify any later work on this by using the existing vector `multirange_swapped` to store information on range swapping per dimension.

As far as I can tell, the error handling you call for is mostly in place already, by calls to `struct_array_packing_error` in `simplify.cc`, although the function name and error message are misleading. I have now renamed the function to `struct_array_dimension_error`, fixed the error message, added comments, and finally added a new check which catches e.g. `bit [7:0][7:0] a[2];`. This would cause incorrect behavior earlier, and actually a segfault with this PR (now fixed).

jix:
> As far as I can tell, the error handling you call for is mostly in place already, by calls to `struct_array_packing_error` in `simplify.cc`, although the function name and error message are misleading.

What I meant is that in addition to the error handling via `struct_array_packing_error`, the code in `slice_range` could have an assert (not a user facing error message), so that when the restriction in struct_array_packing_error is lifted in the future, it will be clear that `slice_range` and the code it calls will also need to be updated to support more dimensions.

Especially due to the use of `multirange_swapped`, which does handle any number of dimensions, it might not be clear otherwise which places need to be updated to support more dimensions. For example when reviewing this code I briefly assumed this was supposed to handle more dimensions already, then noticed that it likely wouldn't do that correctly and only when I tried running tests, I saw that more dimensions lead to the `struct_array_packing_error` error earlier. Such an assertion would have made immediately clear that this isn't supported yet by `slice_range`, so I wouldn't have wondered how this was supposed to work. I think such an assertion might also be a helpful debugging aid for anyone working on this parts of the frontend in the future.

daglem:
I believe it's not only `slice_range`  which would have to be rewritten, but also `make_struct_index_range` and `make_struct_member_range`. But that is for someone to dive into in the future, and is not addressed by this PR.

Following your advice, I have added asserts in a few places where I believe they could be helpful, and I hope this is can now finally be concluded!

PR #3552, opened 2022-11-13
Correct interpretation of SystemVerilog C-style array dimensions

daglem:
IEEE Std 1800™-2017 7.4.2 specifies that [size] is the same as [0:size-1].

### DIFF
--- a/frontends/verilog/verilog_parser.y
+++ b/frontends/verilog/verilog_parser.y
@@ -229,9 +229,9 @@ static AstNode *checkRange(AstNode *type_node, AstNode *range_node)
 static void rewriteRange(AstNode *rangeNode)
 {
 	if (rangeNode->type == AST_RANGE && rangeNode->children.size() == 1) {
-		// SV array size [n], rewrite as [n-1:0]
-		rangeNode->children[0] = new AstNode(AST_SUB, rangeNode->children[0], AstNode::mkconst_int(1, true));
-		rangeNode->children.push_back(AstNode::mkconst_int(0, false));
+		// SV array size [n], rewrite as [0:n-1]
+		rangeNode->children.push_back(new AstNode(AST_SUB, rangeNode->children[0], AstNode::mkconst_int(1, true)));
+		rangeNode->children[0] = AstNode::mkconst_int(0, false);
 	}
 }
 


jix:
This change looks good to me, but it's not clear to me whether this makes an observable difference for anything currently supported by yosys's frontend. Do you have an example that requires this fix? It should be merged in any case, so the implementation matches the specification, but it would be good to know whether I missed something obvious here.

daglem:
I found this bug when comparing a pristine SystemVerilog design with the same design preprocessed by sv2v, using equiv_induct. At the very least, the issue causes noise if you wish to check for *other* issues in a similar way.

I don't think it's very likely that one would be find any real world example where this makes a difference, however it certainly *can*. I guess one could write a test e.g. based on the following statement from section 7.6 in the spec: "For example, if array A is declared as int A[7:0] and array B is declared as int B[1:8], the
assignment A = B; will assign element B[1] to element A[7], and so on." I believe that changing B to int B[7] in this example would reveal the issue. Should I try to add such a test to the PR?

daglem:
Perhaps it's also possible to make a test based on 6.24.3 Bit-stream casting, even though IIRC support for type casting is quite limited in Yosys.

jix:
Assigning arrays is what I tried, but that doesn't seem to be supported. Neither is bit-stream casting. If you have any example you can share where you end up with different RTLIL before and after this change, even if not a minimized test case, that would be great. I'm not super familiar with the frontend, and having an example makes it easier to spot potential issues when reviewing changes.

I'm pretty sure the change in this PR could not break anything new, though, so if you can't reproduce it or when it would be too much work I'm fine with merging this as is.

daglem:
I must have confused the use of equiv_induct with #3551, I found this issue using read_verilog -dump_vlog2. I am not able to find any construct in Yosys which can reveal the difference, and I guess that's why I am not able to show any difference in RTLIL either. The last point where I can see the difference is with -dump_vlog2. However this is the same with e.g. bit v[1:0] vs. bit v[0:1], as far as I can tell :man_shrugging: 

I am a bit confused about the test you pasted above. Which PR is causing trouble with that - this PR, #3552, or both?

jix:
Ah, that's where the edit went 😅 No this test should have only gone to #3551. I edited the comment on #3551 to remove part of it when I realized I hadn't properly tested it and then after properly testing it wanted to undo that and must have switched to the wrong tab. Sorry for all that confusion.

PR #3577, opened 2022-12-07
Deprecate manual

KrystalDelusion:
Working towards deprecating/removing the old manual now that the ReadtheDocs is available.


mmicko:
Makefile needs to be cleaned as well. Targets update-manual and manual fully removed and  3rd line of clean  target

PR #3605, opened 2022-12-30
stat: ensure area is included in json output

gadfort:
I noticed that the area is not captured by the json output from the `stat` command.
This adds the area information into the json output just like the regular text output.


nakengelhardt (approved):

gadfort:
@nakengelhardt thanks for approving. This PR appears to be failing one test, but in looking at it, it seems to be unrelated (the master branch also seems to have this issue). Thanks.

PR #3641, opened 2023-01-22
Handle struct members of union type

daglem:


tests/svtypes/union_simple.sv
```diff
@@ -48,14 +48,24 @@ module top;
...
 	assign ir1 = 32'h0AA01EB7;          //	lui t4,0xAA01
+	assign s1.ir = ir1;
+
 	always_comb begin
 		assert(ir1.u.opcode == 'h37);
 		assert(ir1.r.opcode == 'h37);
 		assert(ir1.u.rd == 'd29);
 		assert(ir1.r.rd == 'd29);
 		assert(ir1.u.imm == 'hAA01);
+		assert(s1.ir.u.imm == 'hAA01);
```
zachjs, line 68: Could we test accesses to the other variants of `s1.ir`?

daglem: OK, I have now added tests for all fields in s1.

zachjs (approved):

PR #3651, opened 2023-01-29
Corrected handling of nested typedefs of struct/union

### LINKED ISSUE #3555: Issue using typedef with packed struct in packed union
### Version

0.22

macOS

### Reproduction Steps

Try to compile this SV code:

```
typedef struct packed {
  logic [7:0] a;
  logic [7:0] b;
} Foo;

typedef Foo Bar;

typedef union packed {
  Bar bar;
  Foo foo;
} FooBar;
```

### Expected Behavior

It should work fine. I tried it in VCS with some extra asserts and it compiles:

```
module StaticAssert;
  if ($bits(Foo) != 16)
    $error("Foo is not the right length");
  if ($bits(Bar) != 16)
    $error("Bar is not the right length");
  if ($bits(FooBar) != 16)
    $error("FooBar is not the right length");
endmodule
```

### Actual Behavior

You get this error:

    ERROR: member bar of a packed union has 1 bits, expecting 16

If you change the second `typedef` to just be a copy of the structure then it works fine.

daglem:
Fixes #3555

### DIFF
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -1380,6 +1380,60 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
 		break;
 
 	case AST_STRUCT_ITEM:
+		if (is_custom_type) {
+			log_assert(children.size() == 1);
+			log_assert(children[0]->type == AST_WIRETYPE);
+			auto type_name = children[0]->str;
+			if (!current_scope.count(type_name)) {
+				log_file_error(filename, location.first_line, "Unknown identifier `%s' used as type name\n", type_name.c_str());
+			}
+			AstNode *resolved_type_node = current_scope.at(type_name);
+			if (resolved_type_node->type != AST_TYPEDEF)
+				log_file_error(filename, location.first_line, "`%s' does not name a type\n", type_name.c_str());
+			log_assert(resolved_type_node->children.size() == 1);
+			AstNode *template_node = resolved_type_node->children[0];
+
+			// Ensure typedef itself is fully simplified
+			while (template_node->simplify(const_fold, at_zero, in_lvalue, stage, width_hint, sign_hint, in_param)) {};
+
+			// Remove type reference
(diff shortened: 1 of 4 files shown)


zachjs (changes requested):
Thank you for your contribution! This change unfortunately causes a regression due to a long-standing deficiency in `copyTypeDefinition`: it substitutes blindly without awareness of scoping or shadowing.

Before this change, `blk.y` appropriately has width 5. After, the shadowing of `W` causes it to incorrectly have width 10.
```SystemVerilog
module top;
    localparam W = 5;
    typedef logic [W-1:0] T;
    T x;
    if (1) begin : blk
        localparam W = 10;
        typedef T U;
        typedef logic [W-1:0] V;
        U y;
        V z;
    end
endmodule
```

I think it's important to avoid this regression in fixing the original issue. Possible routes may involve full handling of `AST_WIRETYPE` in struct/union resolution or adding some scoping logic to `copyTypeDefinition`.

daglem:
Oh. Thanks for reviewing and avoiding the regression!

Could you add your scoping example to the tests?

daglem:
I *finally* got this working - countless hours spent for a small improvement.

I don't think `copyTypeDefinition` can be made to work, since IUC that would require folding of constants from within `verilog_parser.y`.

I have thus removed `copyTypedefinition` in favor of resolving struct/union types in `simplify.cc`. To make this work, it was necessary to differentiate between structs/unions within typedefs (must be expanded) and in wires (must be packed). With `copyTypeDefinition`, only the latter was needed.

daglem:
@zachjs Could you please have a look at this again?

I just amended the commit to check for an instance in `simplify.c` in a less intrusive way - the diff is quite small now.

clairexen:
@zachjs I don't feel knowledgeable enough about this code path to review this. Cab you have another look?

zachjs (approved):
This is a good step in the right direction. Thank you! I will merge this shortly _after_ the next release is cut.

PR #3663, opened 2023-02-06
gowin: Add new types of oscillator

uis246:


gatecat:
Just as a future note to the gowin folks, it's worth considering a script like https://github.com/YosysHQ/yosys/blob/master/techlibs/xilinx/cells_xtra.py to import all the primitives from the vendor tool, so you don't need a new manual patch & PR for each one you add.

mmicko:
ping @pepijndevos @yrabbit

yrabbit:
No objection. 

Worth merging and I think this will be the last commit about adding new primitives - I'm really hoping for a future big addition of all GOWIN primitives at once as @gatecat advises.

PR #1784, opened 2020-03-19
[DRAFT] Merge DFF cells connected to BRAM output ports into BRAM cell, preserving its functionality

dh73:
Derived from Litex VexRiscv+Ethernet SoC, registered outputs w/wo CE and/or RST functionallity in BRAM cells, can be merged into that BRAM, reducing the number of flip-flop cells that the synthesis infers and thus, increasing QoR:

__Before xilinx_bram pass, 36 FDRE.__
```
=== blockram ===
     FDRE                           36
     LUT2                           37
     RAMB36E1                        1
```

__After xilinx_bram pass, 0 FDRE.__
```
=== blockram ===
     LUT2                            1
     RAMB36E1                        1
```
In the latter case, the BRAM DOx_REG (where x is A or B) is set to 1, and the flops with CE or RST functionallity, are connected to the corresponding BRAM input.

---

There exist also test cases in `tests/arch/xilinx/xilinx_bram.ys` that demonstrate:

* [1] Vivado's suggested Verilog for BRAM with CE and RST.
* [2] Test case [1] increasing address and data width (matching shuffled DOA/DOB bits).
* [3] Reduced test case extracted from Litex VexRiscv+Ethernet SoC (2 BRAMS).
* [4] Original test case extracted from Litex VexRiscv+Ethernet SoC (4 BRAMS).

All of the pattern match code is heavily inherited from `xilinx_dsp.pmg`.

__TODO__:
Process other BRAMs that are not `RAMB18E1` and `RAMB36E1`.

PR #3668, opened 2023-02-10
ecp5: Added output register inference for bram

rowanG077:
This PR adds output register inference specifically for ECP5.


whitequark:
Should this not be done as a part of BRAM mapping?

rowanG077:
Would extending `memory_dff` be oke? That's fine by me as well if that is preferred.

whitequark:
> Would extending `memory_dff` be oke?

I think it's `memory_bram` but yes, that's what I would suggest.

rowanG077:
`memory_bram` is not used at all for `synth_ecp5` unless I'm blind.

nakengelhardt:
`memory_bram` is deprecated, `memory_libmap` is the replacement.

whitequark:
> But I'm not sure if it needs to be done as an integral part of memory mapping or if FFs could be merged in after the fact.

Right. I should be more precise in my request: I think it makes more sense to implement this functionality as an architecture-independent pass with some architecture-specific settings (we have a few of these already) rather than a ECP5-specific pass.

rowanG077:
For now I added a new `portoption` called `pipeline_outreg <optional_stages>` which can be used be a mapper to merge in FFs. I'm not sure if that's a good approach either since I'm not that familiar with the code. But I guess how far I'll get with this.

gatecat:
jfyi, there was some previous discussion on a similar feature for xilinx fpgas in #1784

nakengelhardt:
> > But I'm not sure if it needs to be done as an integral part of memory mapping or if FFs could be merged in after the fact.
> 
> Right. I should be more precise in my request: I think it makes more sense to implement this functionality as an architecture-independent pass with some architecture-specific settings (we have a few of these already) rather than a ECP5-specific pass.

Ah! Yes, agreed. It'll be more efficient to handle the relevant FFs while still in coarse grain representation, and that'll also make it easier to handle who gets priority on absorbing a flop if multiple inference mechanisms could use it (such as the issue we had with DSP inference "stealing" the memory read registers) by just changing the order of the passes.

gatecat:
The only case I know of more than one extra pipeline stage is the Xilinx URAM, where there is a second pipeline stage intended for pipelining the ECC decoder. But we have no way of specifying to infer ECC RAM anyway at the moment... 
Screenshot from 2023-02-13 10-22-29

PR #3669, opened 2023-02-10
tests: Fix path of yosys invocation in xprop tests

jix:
For now xprop test failures are still expected and ignored, but without this change, they did not even run unless the yosys build was in path.


jix:
Seems to make the tests take significantly longer overall, I'll have to adjust some parameters to make them faster. Also the comparison with iverilog seems to fail for all tests, which I don't see locally, I'll add more debug output for failing tests to diagnose this.

Issue #3384, opened 2022-06-18
ABC9 crashes on executing ABC9_OPS pass 


benlcb:
## Steps to reproduce the issue

I'm trying to synthesize the [VTR benchmark](https://github.com/verilog-to-routing/vtr-verilog-to-routing/tree/master/vtr_flow/benchmarks/verilog). I've added all missing ram definitions in these files here:

[vtr_benchmark_files.zip](https://github.com/YosysHQ/yosys/files/8932912/vtr_benchmark_files.zip)

I'm starting the synthesis like this:
```
yosys -l yosys.log boundtop.ys
```

boundtop.ys:
``` 
read_verilog -nomem2reg -yydebug boundtop.v
synth_xilinx -dff -flatten -noiopad -edif ch_intrinsics.edif
write_verilog -noexpr -norename boundtop_syn.v
```

Yosys Version: Yosys 0.17+76 (git sha1 d9bb10ba5, gcc 4.8.5 -fPIC -Os)

## Expected behavior

Yosys-ABC9 should finish synthesizeing the design, like Vivado and Yosys (non-abc9).
The benchmakrs for which Yosys-ABC9 crashes are:

boundtop.v
mkDelayWorker32B.v
or1200.v
spree.v

## Actual behavior

Yosys-ABC9 crashes with the following message:
```
...
Successfully finished Verilog frontend.

2.44.16. Executing ABC9_OPS pass (helper functions for ABC9).
terminate called after throwing an instance of 'std::out_of_range'
  what():  vector::_M_range_check
Aborted (core dumped)
```

Full log included in .zip file [vtr_benchmark_files.zip](https://github.com/YosysHQ/yosys/files/8932912/vtr_benchmark_files.zip).

Ravenslofty:
My apologies in getting round to this so late.

There are indeed ABC9 bugs that I will get around, but there's an issue in your code I'm going to fix to reduce other problems down the line (and also because it's probably what you meant).

```
Warning: Replacing memory \myBlockram with list of registers. See boundtop.v:1656, boundtop.v:1654
```

`-nomem2reg` is the wrong thing to use to resolve this. Yosys here says "I don't know how to map this into a memory", and indeed using blocking assignments are not recognised as a memory pattern (we'd like to change this, but one would likely need to rewrite `read_verilog` for that).

```verilog
    reg [DATA_WIDTH-1:0] myBlockram [MEM_DEPTH-1:0];

    always@(posedge clk) begin //Port 1
[...]

PR #3670, opened 2023-02-11
ABC9: Cell Port Bug Patch

benlcb:
Fixes a bug that causes `2.44.16. Executing ABC9_OPS pass` to crash. 

Within ABC9: if a cell's ports weren't fully specified (e.g. 8 bit port, but only 4 bits were set) and those ports had a setup time requirement (aka arrival time) then an exception would occur when trying to set those arrival times.

I added a test file which exhibits this error, but it is not minimal. 

My run looks like this, with `synth_xilinx` being the relevant part of course: 
```
read_verilog /path/to/design/des50.v
synth_xilinx -dff -flatten -noiopad -abc9 -edif des50.edif
write_verilog -noexpr -norename /path/to/output/des50_syn.v
```

Follow up: 
There's also the question about why only 4 bits of the 8 bit port were set, as this concerns the RAMB36E1's write-enable port -- this should be populated by Yosys since it is doing RAM inference. 
--> Fixing this will also make the test case obsolete for the bug above, but I'm not sure how to address this.


gatecat:
A smaller test case than 11k lines of verilog would be appreciated here.

whitequark:
Note, you can use the `bugpoint` pass to minimize testcases.

eddiehung:
@benlcb Hope you don't mind I rebased on master and simplified the testcase.

@gatecat @Ravenslofty Please take another look! I fixed a third breakage in the xaiger writer too.

eddiehung:
Looks like this could resolve https://github.com/YosysHQ/yosys/issues/3384 too (it does for me).

Ravenslofty (approved):

PR #3675, opened 2023-02-15
Support for data and array queries on struct/union item expressions

### LINKED ISSUE #3604: Incorrect return value from `$bits` function
### Version

Yosys 0.24+10 (git sha1 3ebc50dee, gcc 9.3.0-22 -fPIC -Os)

Linux

### Reproduction Steps

While processing a single element of `typedef struct packed` by a `$bits` function, it returns size of a whole structure.
Example design:
```verilog
module top(output int o);

	typedef struct packed {
		logic [1:0][15:0] v_2x16;
		logic [1:0] 	  v_2bit;
	} struct_2el;

	struct_2el array;

	assign o = $bits(array.v_2x16[0]);
endmodule
```
Executed commands:
```bash
$ yosys
yosys> read_verilog -sv -dump_vlog2 top.sv
```

It should return a size of single `[15:0]` element of `v_2x16`.

### Expected Behavior

Yosys should inffer a value equal to a size of an expression passed to `$bits`:
```verilog
assign o = 16;
```

### Actual Behavior

Yosys inffers a value equal to the size of a whole structure:
```verilog
assign o = 34;
```

daglem:
For now, $bits, $left, $right, $low, $high, and $size are supported.

Fixes #3604

### DIFF
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -2077,6 +2077,14 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
 					auto range = make_struct_member_range(this, item_node);
 					newNode = new AstNode(AST_IDENTIFIER, range);
 					newNode->str = sname;
+					// save type and original number of dimensions for $size() etc.
+					newNode->attributes[ID::wiretype] = item_node->clone();
+					if (!item_node->multirange_dimensions.empty() && children.size() > 0) {
+						if (children[0]->type == AST_RANGE)
+							newNode->integer = 1;
+						else if (children[0]->type == AST_MULTIRANGE)
+							newNode->integer = children[0]->children.size();
+					}
 					newNode->basic_prep = true;
 					if (item_node->is_signed)
 						newNode = new AstNode(AST_TO_SIGNED, newNode);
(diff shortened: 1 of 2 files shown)


jix:
Will have to look at the code later, but there are a few issues I spotted while quickly running the tests using the Verific frontend as comparison:

Using `assert property` (i.e. a concurrent assertion) requires a clock (e.g. being inside a clocked always block) but Yosys's frontend doesn't actually support those so it's treated the same as a plain `assert` (immediate assertion) which would be sufficient here and doesn't require a clock. But even for an immediate `assert` the SV spec (and Verific) restricts the use to places where procedural statements are allowed, so they also need to be wrapped in an `always @*` block.

In some places the tests use the undefined `t` instead of `s.t` which seems unintentional (Yosys's frontend also errors on that if ``` `default_nettype none``` is used).

Additionally Verific doesn't like `$size({3{s.x}})` and complains `VERIFIC-ERROR [VERI-1267] struct_sizebits.sv:29: argument of 'size' should be an identifier` but I don't see any such restriction in the standard, so I'd wrap this one in `` `ifndef VERIFIC``.

daglem:
Thank you for your thorough review of the tests - the issues you identified there should be fixed now.

Just FYI: The use of `assert property` was lifted from `ast/sizebits.sv`.

PR #3682, opened 2023-02-19
Out of bounds checking for struct/union members

daglem:
Currently, only constant indices are checked.

### DIFF
--- a/frontends/ast/ast.cc
+++ b/frontends/ast/ast.cc
@@ -524,7 +524,13 @@ void AstNode::dumpVlog(FILE *f, std::string indent) const
 		break;
 
 	case AST_IDENTIFIER:
-		fprintf(f, "%s", id2vl(str).c_str());
+		{
+			AST::AstNode *member_node = AST::get_struct_member(this);
+			if (member_node)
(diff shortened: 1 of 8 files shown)


daglem:
@jix This makes the tests you suggested in #3573 work.

I'm not sure how out of bounds checking should/could be done for dynamic indices - can a wire chunk representing a struct member somehow be broken out before the "shifter cell" in `genrtlil.c` is created?

I noticed that there is also a "replace dynamic ranges" in `simplify.cc` - this seems similar, but seems to be restricted to assignments.

daglem:
@jix Struct/union members are now represented with wire chunks, and indexing is performed within these chunks.

daglem:
@jix I just realized that `read_verilog` `-dump_ast2` and `-dump_vlog2` obviously don't take the wire chunk offset into consideration. Since this is a regression, I converted the PR to a draft for now.

daglem:
@jix I think I'm happy with this now. No really intrusive changes, the chunks are visible in `-dump_ast2` and `-dump_vlog`, and designs with dynamic indexing members which are part of bigger structs/unions can use less resources.

PR #3684, opened 2023-02-23
Makefile: fix GIT_REV extraction if Yosys is built as submodule

whitequark:


mmicko:
@whitequark This breaks `sudo make install` and you get something like

```
fatal: detected dubious ownership in repository at '/xxxxxxxxxx/src/yosys'
To add an exception for this directory, call:
```

whitequark:
> This breaks `sudo make install` and you get something like

I cannot reproduce this. Could you please try `GIT_DIR=$(YOSYS_SRC)/.git git rev-parse HEAD` instead of `git -C $(YOSYS_SRC) rev-parse HEAD`? That should work, I think.

jix:
On my system I cannot reproduce this with `sudo` alone anymore as git checks the `SUDO_UID` environment variable now and also considers the repository trustworthy if owned by the user having that UID. AFAICT and IIRC the check of `SUDO_UID` got introduced just before we made the current changes to the Makefile but hadn't made it into a git release or at least not into a widely distributed git release at that point (while the permission check itself already had). Overall I'm not sure how likely it is to still encounter a version of git that has the permission check without the check for `SUDO_UID`, but keeping some workaround for this seems like the safe choice.

I can still reproduce the issue with `sudo env SUDO_UID= make install` when using the parent commit of this PR and can confirm that the current version of this PR does work when invoked like this. To me explicitly setting `GIT_DIR` seems to be a much better workaround than misusing `git ls-remote`, .

While looking up the timeline of the Makefile changes I also saw that the use of `cat` was deliberate, since supposedly `$(file < ...)` does not work with the older make in Ubuntu 18.04 (see https://github.com/YosysHQ/yosys/pull/3138#issuecomment-1136988711), although I don't know whether we actually still want to support that.

mmicko:
@whitequark to reproduce I first did `su` and login, so continue as root. Went to `yosys` dir and then run `git rev-parse --short=9 HEAD`, even with update you did later I get same error.  And same as @jix I do not get this when running `sudo git rev-parse --short=9 HEAD`. Could be issue with sudo in past was just for certain git versions, but not sure.

whitequark:
> Could be issue with sudo in past was just for certain git versions, but not sure.

So what is the git version you are using?

mmicko:
@whitequark git version 2.39.1

nakengelhardt:
> While looking up the timeline of the Makefile changes I also saw that the use of `cat` was deliberate, since supposedly `$(file < ...)` does not work with the older make in Ubuntu 18.04 (see [#3138 (comment)](https://github.com/YosysHQ/yosys/pull/3138#issuecomment-1136988711)), although I don't know whether we actually still want to support that.

macOS still has Make 3.81 and indeed `$(file < ...)` does not work, I get `Building kernel/version_.cc`. We do need to keep the `$(shell cat ...)`.

whitequark:
Fixed.

whitequark:
Thanks! Now YoWASP builds will have the proper sha1 in them, so if you get bug reports you'll know which commit it was built from at least. That said, I currently only publish release builds unless there's a critical bugfix someone needs urgently.

PR #3688, opened 2023-02-27
gatemate: Enable register initialization

pu-cc:
We managed to implement register initialization through specific configuration of CPEs in the P&R bitstream. This commit enables both 0 and 1 initialization with `dfflegalize`.


nakengelhardt:
Great, that'll get rid of all the issues from #3321!

PR #3692, opened 2023-02-28
stat: pass down quiet arg

nakengelhardt:


racodond:
OK. Thanks. Let me know when there is a "master" version available.

PR #3694, opened 2023-03-02
Handling of attributes for struct / union variables

daglem:
(* nowrshmsk *) on a struct / union variable now affects bit slice assignments to members of the struct / union.

### DIFF
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -557,7 +557,7 @@ static int get_max_offset(AstNode *node)
 	return node->range_left;
 }
 
-static AstNode *make_packed_struct(AstNode *template_node, std::string &name)
+static AstNode *make_packed_struct(AstNode *template_node, std::string &name, decltype(AstNode::attributes) &attributes)
 {
 	// create a wire for the packed struct
 	auto wnode = new AstNode(AST_WIRE);
@@ -565,6 +565,9 @@ static AstNode *make_packed_struct(AstNode *template_node, std::string &name)
 	wnode->is_logic = true;
 	wnode->range_valid = true;
 	wnode->is_signed = template_node->is_signed;
+	for (auto &pair : attributes) {
+		wnode->attributes[pair.first] = pair.second->clone();
+	}
 	int offset = get_max_offset(template_node);
 	auto range = make_range(offset, 0);
 	wnode->children.push_back(range);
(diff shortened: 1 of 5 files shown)


frontends/ast/simplify.cc
```diff
@@ -2734,6 +2731,20 @@ bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage,
...
 
+		int stride = 1;
+		AST::AstNode *member_node = get_struct_member(children[0]);
+		if (member_node) {
+			// Clamp chunk to range of member within struct/union.
+			log_assert(!source_offset && !children[0]->id2ast->range_swapped);
+			source_width = member_node->range_left - member_node->range_right + 1;
+
+			int dims = range->type == AST_MULTIRANGE ? GetSize(range->children) : 1;
+			stride = source_width;
```
zachjs, line 2742: Can you add a comment explaining what's going on here with `stride`? My understanding is that this is an optimization because we know that `shift_expr` can only be multiples of `stride`, where `stride` is the number of entries in the slice of the LHS member. You can probably describe this better than I can.

tests/svtypes/struct_dynamic_range.sv
```diff
@@ -24,7 +24,7 @@ module range_case(
     input logic [2:0] addr_o,
     output logic [7:0] data_o
 );
-    // (* nowrshmsk = 1 *)
+    (* nowrshmsk = 1 *)
```
zachjs, line 27: Does this ensure the attributes are propagated and influence the generated RTLIL? Could we check, e.g., that one uses two `$shift` cells and the other doesn't? They may other ways to test this too.

zachjs (changes requested):
Overall this looks good! Two minor requests in the review comments.

frontends/ast/simplify.cc
daglem: Yes, you are correct. I have made an attempt at explaining this now.

Interestingly, for the cases I've tried, ABC seems to be so smart that it can work out the same result without any CASE optimization, but I think it's nice to be able to generate a much smaller AST.

P.S. Before I noticed your review, I also fixed a mistake in the code above.

tests/svtypes/struct_dynamic_range.sv
daglem: I have checked that it takes effect by running `yosys -p 'read_verilog -sv -dump_ast2 struct_dynamic_range.sv'`.

I'm not sure how you'd make an automatic test for that - by using `-dump-rtlil` and checking for a certain amount of `$shiftx` and `switch` in the result?

The tests were really made as a simple test for indexing of struct members, to ensure that this works with and without `(* nowrshmsk *)`, not considering whether that actually makes any difference :slightly_smiling_face: 

BTW I found that I couldn't parameterize the attribute value using a module parameter. I guess this means that structs are prematurely simplified, so that nothing about them can currently be parameterized?

daglem: I have now added checks for the generated RTLIL, this should hopefully cover what you wanted to verify.

PR #3696, opened 2023-03-04
add all gowin sim models

pepijndevos:
I copied the simulation models from the Gowin IDE and transplanted the annotations and timing specifications we've added.

Probably broke all the things, but at least got the tedious bit out of the way.

### DIFF
--- a/techlibs/gowin/cells_sim.v
+++ b/techlibs/gowin/cells_sim.v
@@ -1,632 +1,391 @@
-(* abc9_lut=1 *)
-module LUT1(output F, input I0);
-	parameter [1:0] INIT = 0;
-	specify
-		(I0 => F) = (555, 902);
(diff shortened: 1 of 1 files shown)


gatecat:
Copying the ports from module headers automatically (which is basically just an unavoidable fact) is fine with me as we do for xilinx, nexus, etc but I'm less keen on copying the entire contents of the vendor sim models which are (a) very much copyrighted and (b) probably cause problems for verification in Yosys compared to the sim models we write ourselves.

pepijndevos:
Then I'm not clear what exactly you expect me to do...

gatecat:
Something like this script... https://github.com/YosysHQ/yosys/blob/master/techlibs/xilinx/cells_xtra.py

PR #3704, opened 2023-03-14
verific: Fix enum_values support and signed attribute values

jix:
This uses the same constant parsing for enum_values and for attributes and extends it to handle signed values as those are used for enums that implicitly use the int type.


mmicko (approved):
This is fine, but if you can add the test in verific tests that would be great

PR #3705, opened 2023-03-17
Insert IO buffers for ECP5

mmicko:


gatecat:
Can we have this default disabled for now while it's fleshed out?

We also need `(* iopad_external_pin *)` on any primitive pins that connect directly to top level IO and shouldn't have an IO buffer put inbetween (the TRELLIS_IO primitive, SERDES, etc).

mmicko:
Sure, will rename "-noiopad" to "-iopad" so it is easy to to change back. Also will add attribute on those that I am aware of.

techlibs/ecp5/cells_sim.v
```diff
@@ -378,6 +378,7 @@ endmodule
 // ---------------------------------------
 (* keep *)
 module TRELLIS_IO(
+	(* iopad_external_pin *)
```
gatecat, line 381: This is also needed on the following pins at a minimum:
 - `TCK, TMS, TDI, TDO` of `JTAGG`
 - `CH0_HDINP, CH1_HDINP, CH0_HDINN, CH1_HDINN` of `DCUA`
 - `REFCLKP`, `REFCLKN` of `EXTREFB`

PR #3706, opened 2023-03-17
MachXO2 unification of TRELLIS primitives with ECP5

mmicko:
Note this is minimal effort unification, enough to provide proper handling and simulation of these primitives, in future, they can fully share simulation models for these.

https://github.com/YosysHQ/yosys/pull/3705 is needed since cells_io.vh from ECP5 is being used.

### DIFF
--- a/techlibs/machxo2/Makefile.inc
+++ b/techlibs/machxo2/Makefile.inc
@@ -1,6 +1,7 @@
 
 OBJS += techlibs/machxo2/synth_machxo2.o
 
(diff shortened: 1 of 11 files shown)


gatecat (approved):
LGTM, not sure what order you want things merged in though.

mmicko:
Guess, this one first, then prjtrellis-db, prjtrellis (I will update DB link before that) and then nextpnr (so I can update prjtrellis version)

PR #3709, opened 2023-03-21
Update Xilinx cell definitions, fixes #3699

### LINKED ISSUE #3699: Xilinx IBUF missing parameter CCIO_EN in cells_sim.v
### Version

Yosys 0.9+4276

Linux

### Reproduction Steps

When creating a Verilog netlist using Vivado 2022.2, the IBUF primitives are instantiated with the `CCIO_EN` parameter:

```
 IBUF #(
    .CCIO_EN("TRUE")) 
    \a_IBUF[0]_inst 
       (.I(a[0]),
        .O(a_IBUF[0]));
```
This parameter is missing from the IBUF model in cells_sim.v

This behavior seems to be new, as Vivado 2020.2 would not output this parameter.

This was tested with a 7-series part (xc7a35tcpg236-1).  For context, I am using this tcl script to generate the netlist in Vivado <https://github.com/chipsalliance/f4pga-xc-fasm2bels/blob/master/tests/create_golden_file.tcl>; however, this seems to occur with various scripts I have that use `write_verilog` in Vivado 2022.2.

### Expected Behavior

yosys should be able to read in Vivado-produced netlists.

### Actual Behavior

```
ERROR: Module `IBUF' referenced in module `add32' in cell `rst_IBUF_inst' does not have a parameter named 'CCIO_EN'.
```

mmicko:


gatecat:
I don't think this actually fixes #3699, which is about a cell that's in `cells_sim.v` not `cells_xtra.v`

mmicko:
Right,  I will update PR

PR #3716, opened 2023-03-28
Fix missing brackets around ``else`` in genrtlil

kamilrakoczy:
I believe this was intended behavior of this code.


frontends/ast/genrtlil.cc
```diff
@@ -844,15 +844,16 @@ void AstNode::detectSignWidthWorker(int &width_hint, bool &sign_hint, bool *foun
...
-				range = children[0];
+			} else {
+				if (id_ast->children[0]->type != AST_CONSTANT)
+					while (id_ast->simplify(true, false, false, 1, -1, false, true)) { }
+				if (id_ast->children[0]->type == AST_CONSTANT)
+					this_width = id_ast->children[0]->bits.size();
+				else
+					log_file_error(filename, location.first_line, "Failed to detect width for parameter %s!\n", str.c_str());
+				if (children.size() != 0)
+					range = children[0];
```
clairexen, line 855: This last ``if (children.size() != 0) range = children[0];`` should not be moved into the else-block. It handles the case where the identifier is used with a bit slice, and that step should not condition on how the parameter or localparam is declared.

clairexen (changes requested):

PR #3719, opened 2023-03-31
machxo2: Initial support for carry chains (CCU2D)

mmicko:
Keeping it off by default for now, includes simulation model.
Also includes lutram to TRELLIS_DPR16X4.


gatecat:
Can you add a couple of arch tests to also run with this enabled?

PR #3722, opened 2023-04-05
Touch on ast/simplify

povik:
This is supposed to untangle the simplify code a bit by removing three of its arguments (in_lvalue/in_param/at_zero), and making few other changes.

I expect the at_zero removal to be less objectionable than the in_param/in_lvalue removal, since the latter (with the way it's being replaced) requires the code to be sprinkled with calls to `fixup_hierarchy_flags` and `set_attribute` (new methods). That may not be as bad if we (1) don't require the in_param/in_lvalue values of the old code to be perfectly reproduced (which the change here does reproduce, at least as exercised by `make test`); (2) move to a coding style in the simplify code that doesn't rely on explicitly modifying the children list of an existing node as much. See the commits and let me know what you think.

This should be passing `make test` at each commit.

### DIFF
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -99,7 +99,7 @@ std::string AstNode::process_format_str(const std::string &sformat, int next_arg
 								cformat, str.c_str());
 
 					node_arg = children[next_arg++];
-					while (node_arg->simplify(true, false, false, stage, width_hint, sign_hint, false)) { }
+					while (node_arg->simplify(true, false, stage, width_hint, sign_hint, false)) { }
 					if (node_arg->type != AST_CONSTANT)
 						input_error("Failed to evaluate system task `%s' with non-constant argument.\n", str.c_str());
 					break;
@@ -192,7 +192,7 @@ void AstNode::annotateTypedEnums(AstNode *template_node)
 		log_assert(current_scope.count(enum_type) == 1);
 		AstNode *enum_node = current_scope.at(enum_type);
 		log_assert(enum_node->type == AST_ENUM);
-		while (enum_node->simplify(true, false, false, 1, -1, false, true)) { }
+		while (enum_node->simplify(true, false, 1, -1, false, true)) { }
 		//get width from 1st enum item:
 		log_assert(enum_node->children.size() >= 1);
 		AstNode *enum_item0 = enum_node->children[0];
@@ -823,8 +823,8 @@ static bool try_determine_range_width(AstNode *range, int &result_width)
 	AstNode *left_at_zero_ast = range->children[0]->clone_at_zero();
 	AstNode *right_at_zero_ast = range->children[1]->clone_at_zero();
 
-	while (left_at_zero_ast->simplify(true, false, false, 1, -1, false, false)) {}
-	while (right_at_zero_ast->simplify(true, false, false, 1, -1, false, false)) {}
+	while (left_at_zero_ast->simplify(true, false, 1, -1, false, false)) {}
+	while (right_at_zero_ast->simplify(true, false, 1, -1, false, false)) {}
 
 	bool ok = false;
 	if (left_at_zero_ast->type == AST_CONSTANT
@@ -909,10 +909,8 @@ static void check_auto_nosync(AstNode *node)
 //
 // this function also does all name resolving and sets the id2ast member of all
 // nodes that link to a different node using names and lexical scoping.
-bool AstNode::simplify(bool const_fold, bool at_zero, bool in_lvalue, int stage, int width_hint, bool sign_hint, bool in_param)
+bool AstNode::simplify(bool const_fold, bool in_lvalue, int stage, int width_hint, bool sign_hint, bool in_param)
 {
-	log_assert(!at_zero);
-
 	static int recursion_counter = 0;
 	static bool deep_recursion_warning = false;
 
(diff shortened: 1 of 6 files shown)


frontends/ast/simplify.cc
```diff
@@ -1093,7 +1186,9 @@ bool AstNode::simplify(bool const_fold, bool in_lvalue, int stage, int width_hin
 
 	// in certain cases a function must be evaluated constant. this is what in_param controls.
 	if (type == AST_PARAMETER || type == AST_LOCALPARAM || type == AST_DEFPARAM || type == AST_PARASET || type == AST_PREFIX)
-		in_param = true;
+		in_param_ = true;
+	log_assert(in_param == in_param_);
+	log_assert(in_lvalue == in_lvalue_);
```
povik: I want to point out these asserts (comparing the flags on the node with the flags passed in to simplify) are upheld through `make test`.

zachjs:
Thank you for your submission! I think this offers some nice cleanup overall, though I'd like to hear from @jix and/or @clairexen before moving forward. I think we're far more likely to get this merged if it each commit is split into a separate PR. So far, I have only skimmed a bit and provided some initial thoughts.

* The first commit seems straightforwardly good and mergable as-is.
* This seems like a nice improvement, but it is a sweeping stylistic change that A) requires broader agreement; and B) is essentially orthogonal to the rest of the PR.
* This seems like a nice strategy. You mention assuming `id2ast` is valid, but I'm not sure how we benefit from making this assumption. Could `clone_at_zero` look up identifiers in `current_scope` as was done before? Given the subtree has no scoping, I don't see any harm in using the scope as a fallback.
* I like splitting up obviating the parameter and actually removing it. However, I would really like to see the _prior_ commit remove the code that _uses_ `at_zero`, rather than making it unreachable there and removing it here. This would make it easier to compare the new "at zero" logic with what it replaces, both now and in the future.
* I haven't had the chance to look closely at the remaining three commits, but they may be trickier and/or more contentious.

clairexen:
I agree with everything zach said, and I'm fine with merging the first few simple PRs first, and then discussing the remaining commits as individual PRs.

povik:
Hi Zach, Claire,

dropped the last three commits, so now there are only those Zach commented on. Also did rebase.

>  You mention assuming `id2ast` is valid, but I'm not sure how we benefit from making this assumption. Could `clone_at_zero` look up identifiers in `current_scope` as was done before? Given the subtree has no scoping, I don't see any harm in using the scope as a fallback.

I think it just seemed favorable to rely on an attribute of the AST node in the node's method, rather than be coupled to the simplify() process through the `current_scope` variable. Bit surprisingly, when I now tried to use `current_scope` directly to look up identifiers in `clone_at_zero`, I started failing some tests, even though that's akin to what the original at_zero simplify code does. I didn't investigate, and changed it to use `id2ast` and fall back on `current_scope` for now -- that's what's in the PR now.

> I like splitting up obviating the parameter and actually removing it. However, I would really like to see the _prior_ commit remove the code that _uses_ `at_zero`, rather than making it unreachable there and removing it here. This would make it easier to compare the new "at zero" logic with what it replaces, both now and in the future.

Changed it to be that way. Prior commit replaces the code conditioned on `at_zero` with an assertion that the flag is false, latter commit removes the flag.

frontends/ast/simplify.cc
```diff
@@ -775,6 +775,33 @@ static IdentUsage always_asgn_before_use(const AstNode *node, const std::string
...
+	}
+
+	AstNode *left_at_zero_ast = range->children[0]->clone();
+	AstNode *right_at_zero_ast = range->children[1]->clone();
+
+	while (left_at_zero_ast->simplify(true, true, false, 1, -1, false, false)) {}
+	while (right_at_zero_ast->simplify(true, true, false, 1, -1, false, false)) {}
+
+	bool ok = false;
+	if (left_at_zero_ast->type == AST_CONSTANT \
```
zachjs, line 837: AFAIK, this backslash isn't required, and isn't used elsewhere in this file.

frontends/ast/simplify.cc
```diff
@@ -41,6 +41,13 @@ YOSYS_NAMESPACE_BEGIN
 using namespace AST;
 using namespace AST_INTERNAL;
 
+void AstNode::input_error(const char *format, ...) const
+{
+	va_list ap;
+	va_start(ap, format);
+	logv_file_error(filename, location.first_line, format, ap);
+}
```
zachjs, line 49: I'm partial to defining this ins `ast.cc` instead.

frontends/ast/simplify.cc
```diff
@@ -780,6 +780,44 @@ static IdentUsage always_asgn_before_use(const AstNode *node, const std::string
...
+	case AST_IDENTIFIER:
+		if (id2ast)
+			pointee = id2ast;
+		else if (current_scope.count(str))
+			pointee = current_scope[str];
+		else
+			break; 
+		
+		if (pointee->type != AST_WIRE && \
+				pointee->type != AST_AUTOWIRE && \
```
zachjs, line 799: Let's drop the backslashes.

zachjs (approved):
I'm fine with merging these 4 commits as a batch. I plan to merge this shortly _after_ the next release is cut. I have a few very minor nits from my second review. I'm happy to fix them if you 1) don't object; and 2) don't have the chance address them in the next few days.

Thanks again for your contribution! These are great simplifications.

povik:
Hi Zach,

thanks for taking another look at the changes! I happen to be AFK for the next week but feel free to merge it with those minor adjustments.

PR #3732, opened 2023-04-17
Remove a statement without effect.

hzeller:
The return value of the min(...) call is never used. Looks like some leftover from some previous implementation.


hzeller:
CC: @jix

Issue #3151, opened 2022-01-08
Installing plugins in /usr/share violates FHS


DanielG:
yosys currently loads shared library plugins from `/usr/share/yosys/plugins` (in the default configuration). Installing architecture-dependent files in share however violates the [Filesystem Hirarchy Standard (Section 4.11)](https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html#usrshareArchitectureindependentData). I came upon this problem as part of packaging work on a yosys plugin in Debian.

While this can be worked around in distribution specific patches I believe a better approach would be to fix this upstream by adding a concept of libdir to yosys/yosys-config, preferably loading from there and falling back to share if nothing is found there.

I'm happy to deliver patches if this change is welcome.

See also #2545 which is about allowing user defined plugins locations, I'm proposing to just another default location for plugins instead.

widlarizer:
Hi, sorry for the lack of activity. Is this still relevant? Doesn't `yosys-config --datdir` implement what you need for plugins by giving the share directory location?

Issue #2545, opened 2021-01-17
Allow loading plugins "by name" from more than one source (alt: add a search path for plugins)


thoughtpolice:
When yosys is invoked with the flag `-m foo` or the command `plugin -i foo` is run &mdash; that is, with a non-absolute path to a dynamically loadable library &mdash; the loader will look for a file named `foo.so` by default in (among other places) `/usr/share/yosys/plugins`, where `$PREFIX/share/yosys` is also known as `YOSYS_DATDIR`.

The intention of this design is that this single location is global and shared across the system; tools like **[yosys-ghdl](https://github.com/ghdl/ghdl-yosys-plugin)** or **[yosys-bluespec](https://github.com/thoughtpolice/yosys-bluespec)** should by default be installed into such a location by invoking `cp ... $(yosys-config --datdir)/plugins` or whatever after compilation. Package managers should put the `.so` files in this directory, and so forth.

This is mostly true, except when using a package manager like **[Nix](https://nixos.org)**. To be very brief, Nix installs a package into a directory, and then marks that directory as read-only. This means that plugins cannot, later on, install their `.so` files into the same data directory of the yosys package. Nix users, therefore, can't simply install a plugin and then invoke a command like `yosys -m bluespec -p 'read_bluespec design.bsv'` &mdash; the location of the plugin file will exist in another directory.

---

To work around this as the Yosys maintainer for Nix, I've added a patch to Yosys that effectively makes yosys instead search a *list* of directories when looking for a "named" plugin. This patch **[is very simple](https://github.com/NixOS/nixpkgs/blob/2660037f8cdf6bb69bacbf3d9d2f9f72d056c8d6/pkgs/development/compilers/yosys/plugin-search-dirs.patch)**. It allows you to add a set of directories to the environment variable `NIX_YOSYS_PLUGIN_DIRS` separated by `:`, and Yosys will search them. So invoking

```bash
$ export NIX_YOSYS_PLUGIN_DIRS=/d1:/d2:...
$ yosys -m bluespec
```

will try to load `bluespec.so` from the following locations, in this order:

```
YOSYS_DATDIR/plugins/bluespec.so
/d1/bluespec.so
/d2/bluespec.so
...
```

(Note this environment variable is private and undocumented. It is populated automatically by the Nix infrastructure, so users should be unaware of it. It only exists as a means to an end to make the Yosys UX for Nix users behave similarly to non-Nix users.)

---

[...]

PR #3749, opened 2023-05-03
Updated the `plugin` command to better handle paths

lethalbit:
This PR updates the `plugin -i` command to better deal with some edge cases and inconsistencies with how it behaved.

The first update is that it now ensures that the plugin does not end with `.so` before trying to append it. Previously when running `plugin -i nya.so` it would try to look for `nya.so.so` in the `YOSYS_SHARE/plugins` directory, that has been fixed.

The second update is that it now checks for the `.py` extension even without Python support being enabled, and produces an error message accordingly.

The final update is that the filename path is now also run through `rewrite_path`, as is done with a large portion of the other file inputs in Yosys.

This PR does not address #2545 nor #3151 yet, I wanted to get some basic restructuring done first to make implementation a little nicer.

### DIFF
--- a/passes/cmds/plugin.cc
+++ b/passes/cmds/plugin.cc
@@ -21,12 +21,12 @@
 
 #ifdef YOSYS_ENABLE_PLUGINS
 #  include <dlfcn.h>
-#  include <boost/filesystem.hpp>
-#  include <boost/algorithm/string/predicate.hpp>
 #endif
 
 #ifdef WITH_PYTHON
+#  include <boost/algorithm/string/predicate.hpp>
 #  include <Python.h>
+#  include <boost/filesystem.hpp>
 #endif
 
 YOSYS_NAMESPACE_BEGIN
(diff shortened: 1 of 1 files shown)


mmicko:
LGTM, but there are still situations where checking if plugin is already loaded could be problematic (already is like that so you did not broke anything), guess storing base name of file would be better than original filename.

lethalbit:
@mmicko would you like me to implement the normalization for checking to see if the plugin is loaded or not?

My general idea for that is to just get the stem from the file name so if you load something like `+/plugins/foo.bar.so` it would just register it as `foo.bar.so`, although this could potentially cause problems if there are plugins with ambiguous stems. In which case some unique plugin identifier would be needed, or would that not be a problem?

mmicko:
Hi @lethalbit, my plan is to merge this right after release (have to check with others but should be these days), just to have more time to test all in real use after merge, since now tools like eqy are relaying on plugin mechanism.

mmicko (approved):

lethalbit:
Okay! Sounds good to me~

Issue #3594, opened 2022-12-19
Print backtrace on abort/assert


nakengelhardt:
### Feature Description

Currently it only prints the line where the assert was hit - it might help for some bug reports if it also prints more information about the context.

arcturusannamalai:
**libunwind** can be adapted do this in debug build of Yosys; however since Yosys is a command-by-command interpreter this is sufficient isn't it ? Unlike having function calls etc. its a straight line interpreter and it just fails on most recent command so I don't see a major need for it yet.

arcturusannamalai:
rough cut implementation to print a stack of commands executed to point of failure https://github.com/YosysHQ/yosys/pull/3757

arcturusannamalai:
Updated to actually print the stack trace; see commit detail

arcturusannamalai:
@nakengelhardt @whitequark - any idea on the PR ?

PR #3757, opened 2023-05-09
[YOSYS][Issue 3594] Print backtrace on abort/assert #3594

arcturusannamalai:
ERROR: No such command: prox (type help for a command overview) Error while executing script:
	Running script on file demo.ys
---------------------------------
   1 read_verilog ./tests/lut/map_not.v
   2 opt
   3 proc_clean
-->4 prox
---------------------------------

### DIFF
--- a/kernel/log.cc
+++ b/kernel/log.cc
@@ -354,6 +354,21 @@ static void logv_error_with_prefix(const char *prefix,
 
 	log_check_expected();
 
+	int lineno = 1;
+	if ( !log_line_number.empty() ) {
+	  log("Error while executing script:\n\t");
+	  log( log_line_number.begin()->c_str() );
+	  log("\n");
+          log_line_number.erase(log_line_number.begin());
+	  log("---------------------------------\n");
+	  for(const std::string& script_line : log_line_number) {
+	    const char* pfx = (lineno == log_line_number.size()) ? "-->" : "   ";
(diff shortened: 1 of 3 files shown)


mmicko:
Please split these two features in separate PRs  ( create new one for logging script line , keep backtrace in this)

arcturusannamalai:
OK got it.

arcturusannamalai:
Update done.

arcturusannamalai:
The test is failing perhaps because of the message, 
"Backtrace not available on this platform,"
[1948](https://github.com/YosysHQ/yosys/actions/runs/5006307984/jobs/8971410479?pr=3757#step:10:1949)

kernel/driver.cc
```diff
@@ -174,6 +202,8 @@ extern "C" {
 
 void yosys_atexit()
 {
+  
+  yosys_print_trace();
```
whitequark, line 206: `yosyos_atexit` does not seem to me like the right place to print a backtrace, why this place in particular?

arcturusannamalai:
I officially abandon this PR.

widlarizer:
No problem. In case anybody is interested in reworking this: I think we'd first need a stronger delineation between errors that imply a bug to be reported to us, and errors that imply the user should change something. Unfortunately we may be using errors and asserts to mean both of those so that should be looked into first

PR #3767, opened 2023-05-17
Assign wires an smtoffset

KrystalDelusion:
Wires weren't being assigned an smtoffset value so when generating a yosys witness trace it would also use an offset of 0.
Not sure if this has any other effects, but it fixes the bug I was having.
@jix could you take a look at this?

### DIFF
--- a/backends/smt2/smt2.cc
+++ b/backends/smt2/smt2.cc
@@ -626,9 +626,12 @@ struct Smt2Worker
 				}
 
 				bool init_only = cell->type.in(ID($anyconst), ID($anyinit), ID($allconst));
-				for (auto chunk : cell->getPort(QY).chunks())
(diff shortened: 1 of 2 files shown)


jix:
I'm pretty sure that you did find the root cause with the missing `smtoffset`, but I think your fix only worked incidentally and doesn't result in the correct `smtoffset` in general. I pushed a new commit with what I think should fix this, more details are in the commit message.

While verifying this by running the SBY testsuite with `vcd_sim` on by default, I ran into another small bug in `sim` for which I also pushed a commit here.

Can you check whether the new `smtoffset` computation still fixes the bug for you?

KrystalDelusion:
Thanks!  I had a feeling my quick fix was missing something, and your change still fixes the bug I had with the yosys witness.

PR #3770, opened 2023-05-18
Standard compliance for tests/verilog/block_labels.ys

dwRchyngqxs:
genvar declaration cannot take an initial value when declared as a module_or_generate_item_declaration. Correct this test so that it doesn't fail unexpectedly if Yosys aligns with the standard.


zachjs (approved):
I agree with your interpretation of the standard. I agree it seems worthwhile to fix this test.

Issue #3743, opened 2023-04-26
LUTs not topologically sorted after write_verilog


cgouert:
### Version

Yosys 0.9+2406 (git sha1 334ec5fa, clang 10.0.0-4ubuntu1 -fPIC -Os)

### On which OS did this happen?

Linux

### Reproduction Steps

Given this verilog program (adder.v)

```
module adder(G11,G12,G14);
input [7:0] G11,G12;
output [7:0] G14;
assign G14 = G11+G12;
endmodule
```

Yosys script: `read_verilog adder.v; synth; abc -lut 6; write_verilog -noexpr -v adder_out.v` 

LUTs are not topologically sorted in adder_out.v

However, executing `read_verilog adder.v; synth; abc -g simple,-MUX; write_verilog -noexpr -v adder_out.v`  yields topologically sorted logic gates.

cc: @jimouris

### Expected Behavior

I would expect the LUTs to be topologically sorted in the generated Verilog.

### Actual Behavior

LUTs are not topologically sorted when dumped to a Verilog netlist.

whitequark:
There's no guarantee that the generated Verilog is topologically sorted. What you see when using it with `abc` is a coincidence.

cgouert:
Thank you for your reply; I can also replicate this behavior with other circuits, but this may be a coincidence as you mentioned. I know that Yosys has the capability to topologically sort circuits with the torder command  (which prints a topological ordering of the nodes/gates). Is there any way to combine this somehow with the write_verilog command to reflect this in the generated netlist?

whitequark:
Not that I know of.

jimouris:
It seems that it is not a coincidence: https://github.com/berkeley-abc/abc/issues/60 and `abc` sorts the gates. It seems that it's a bug on their end but can also be resolved on yosys' side with the output of the `torder` command.

whitequark:
Sorry, I was unclear. By "coincidence" I meant that, as far as I know, Yosys makes no guarantee that the Verilog output (or any other output) will be topologically sorted after running `abc`.

arcturusannamalai:
my attempt at a solution https://github.com/YosysHQ/yosys/pull/3774

PR #3774, opened 2023-05-21
[YOSYS][Issue-3743] enable toposort onto Module Cells  #3743

arcturusannamalai:
and reuse for TORDER and WRITE_VERILOG for issue  #3743

e.g.
read_verilog adder.v; synth; abc -lut 6; write_verilog -noexpr -toposort -v adder_out_topo.v

### DIFF
--- a/backends/verilog/verilog_backend.cc
+++ b/backends/verilog/verilog_backend.cc
@@ -40,7 +40,7 @@ int auto_name_counter, auto_name_offset, auto_name_digits, extmem_counter;
 std::map<RTLIL::IdString, int> auto_name_map;
 std::set<RTLIL::IdString> reg_wires;
 std::string auto_prefix, extmem_prefix;
-
+bool toposort;
 RTLIL::Module *active_module;
 dict<RTLIL::SigBit, RTLIL::State> active_initdata;
 SigMap active_sigmap;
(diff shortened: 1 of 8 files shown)


passes/techmap/abc9_ops.cc
```diff
@@ -1202,7 +1202,7 @@ void reintegrate(RTLIL::Module *module, bool dff_mode)
 	}
 
 	std::vector<Cell*> boxes;
-	for (auto cell : module->cells().to_vector()) {
+	for (auto cell : module->cells()) {
```
eddiehung, line 1205: Note that `to_vector` creates a copy of the cell list, necessary because the body of the `for` loop is modifying this very same list.

backends/verilog/verilog_backend.cc
```diff
@@ -2345,6 +2353,7 @@ struct VerilogBackend : public Backend {
 				continue;
 			}
 			log("Dumping module `%s'.\n", module->name.c_str());
+			if (toposort) module->run_toposort_cells();
```
eddiehung, line 2356: In my opinion, a backend should have no side effects on the design.

Ravenslofty: Agreed: sometimes it can be useful to output to multiple formats, and there shouldn't be order-dependency - we have enough bug reports about chaos in Yosys as it is.

arcturusannamalai:
RIP

PR #3779, opened 2023-05-25
Take DESTDIR variable into account for yosys-config

Coloquinte:
This allows plugin compilation to work with user installs (for local developments, or when a user does not have sudo rights)


Coloquinte:
Closing because DESTDIR is used for packaging and shouldn't change the generated package --> local installations should only use PREFIX, not DESTDIR

PR #3780, opened 2023-05-25
intel_alm: re-enable carry chains for ABC9

Ravenslofty:
The issue of whitebox carry chains having multiple fanout seems to have been fixed by berkeley-abc/abc#139, so we can safely revert 34a08750fa1a490be09411b07f64f4236eff234e to re-enable whitebox optimisation of carry chains.


gatecat (approved):
LGTM

Issue #2616, opened 2021-02-27
CXXRTL: failed backend invocation leaves empty output file


whitequark:
## Steps to reproduce the issue

```
cat >repro.v <<END
module a; endmodule
module top; a a(); endmodule
END
yosys repro.v -o repro.cc
```

## Expected behavior

`repro.cc` is either not created, or removed after being created, to play nice with build systems.

## Actual behavior

Empty `repro.cc` is created, breaking downstream build if it is the sole target in the dependency graph.

## Notes

This might be affecting more backends than CXXRTL, CXXRTL is just particularly picky about the input.

arcturusannamalai:
this happens because of https://github.com/YosysHQ/yosys/issues/3762

whitequark:
I don't actually see how that's caused by not having `post_execute()`.

PR #3781, opened 2023-05-25
[Backend] fix empty files when exception is thrown issue #2616

arcturusannamalai:
enable list of files planned in execution which will be deleted if exception thrown
fix issue #2616

### DIFF
--- a/kernel/register.cc
+++ b/kernel/register.cc
@@ -649,16 +649,36 @@ void Backend::execute(std::vector<std::string> args, RTLIL::Design *design)
 {
 	std::ostream *f = NULL;
 	auto state = pre_execute();
-	execute(f, std::string(), args, design);
-	post_execute(state);
-	if (f != &std::cout)
-		delete f;
+	bool restore_log_cmd_error_throw = log_cmd_error_throw;
+	log_cmd_error_throw = true;
+	try {
+	  execute(f, std::string(), args, design);
+	  post_execute(state);
+	} catch(log_cmd_error_exception& _) {
+	  if (f && (f != &std::cout)) {
+	    delete f;
+	    f = NULL;
+
+	    for(auto& itr : filenames) {
+	      if ( itr == "<stdout>" ) continue;
+	      int code = std::remove(itr.c_str());
+	      if ( code )
+		log("\t Failed to deleted file %s - with exitcode %d\n",itr.c_str(),code);
+	    }
+	  }
+	  log_cmd_error_throw = restore_log_cmd_error_throw;
+	  if ( log_cmd_error_throw )
+	    throw log_cmd_error_exception();
+	}
+
+	log_cmd_error_throw = restore_log_cmd_error_throw;
+	if (f && (f != &std::cout)) {
+	  delete f;
+	}
 }
 
 void Backend::extra_args(std::ostream *&f, std::string &filename, std::vector<std::string> args, size_t argidx, bool bin_output)
 {
-	bool called_with_fp = f != NULL;
-
 	for (; argidx < args.size(); argidx++)
 	{
 		std::string arg = args[argidx];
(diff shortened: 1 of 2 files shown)


whitequark (commented):
I haven't read the code in detail but the overall approach seems reasonable to me.

arcturusannamalai:
@whitequark - could someone please review/merge this PR ? thanks

kernel/register.cc
```diff
@@ -696,18 +718,18 @@ void Backend::extra_args(std::ostream *&f, std::string &filename, std::vector<st
 				delete ff;
 				log_cmd_error("Can't open output file `%s' for writing: %s\n", filename.c_str(), strerror(errno));
 			}
+			filenames.push_back(filename);
 			f = ff;
 		}
 	}
 
-	if (called_with_fp)
-		args.push_back(filename);
```
jix, line 704: Why was this removed? This actually changes behavior in potentially breaking ways and seems unrelated to the actual changes done to remove empty files on cmd errors.

kernel/register.cc
```diff
@@ -649,16 +649,36 @@ void Backend::execute(std::vector<std::string> args, RTLIL::Design *design)
...
+	  }
+	  log_cmd_error_throw = restore_log_cmd_error_throw;
+	  if ( log_cmd_error_throw )
+	    throw log_cmd_error_exception();
+	}
+
+	log_cmd_error_throw = restore_log_cmd_error_throw;
+	if (f && (f != &std::cout)) {
+	  delete f;
+	}
```
jix, line 677: Please keep indentation consistent and void mixing spaces and tabs.

kernel/register.cc
```diff
@@ -668,8 +688,9 @@ void Backend::extra_args(std::ostream *&f, std::string &filename, std::vector<st
 		if (f != NULL)
 			cmd_error(args, argidx, "Extra filename argument in direct file mode.");
 
-		if (arg == "-") {
+		if (arg == "-") {		  
 			filename = "<stdout>";
+			filenames.push_back(filename);
```
jix, line 693: Why is `"<stdout>"` pushed here? You're ignoring it above and there are no other uses for `filenames`

kernel/register.cc
```diff
@@ -696,18 +718,18 @@ void Backend::extra_args(std::ostream *&f, std::string &filename, std::vector<st
...
 
-	if (called_with_fp)
-		args.push_back(filename);
 	args[0] = pass_name;
 	// cmd_log_args(args);
 
 	if (f == NULL) {
 		filename = "<stdout>";
 		f = &std::cout;
+		filenames.push_back(filename);
```
jix, line 732: Same as above

kernel/register.cc
```diff
@@ -649,16 +649,36 @@ void Backend::execute(std::vector<std::string> args, RTLIL::Design *design)
...
+	try {
+	  execute(f, std::string(), args, design);
+	  post_execute(state);
+	} catch(log_cmd_error_exception& _) {
+	  if (f && (f != &std::cout)) {
+	    delete f;
+	    f = NULL;
+
+	    for(auto& itr : filenames) {
+	      if ( itr == "<stdout>" ) continue;
```
jix, line 663: See my question about pushing `"stdout"` below

kernel/register.cc
```diff
@@ -668,8 +688,9 @@ void Backend::extra_args(std::ostream *&f, std::string &filename, std::vector<st
 		if (f != NULL)
 			cmd_error(args, argidx, "Extra filename argument in direct file mode.");
 
-		if (arg == "-") {
+		if (arg == "-") {
```
jix, line 691: Avoid adding trailing whitespace

jix (changes requested):

kernel/register.cc
arcturusannamalai: ok let me redo this; thanks

kernel/register.cc
arcturusannamalai: reverted it

PR #3789, opened 2023-05-29
intel_alm: re-enable 8x40-bit M10K support

Ravenslofty:
This is effectively a revert of #3124, which was a misunderstanding of how the M10K works - I had assumed that one port is a 20-bit write and one port is a 20-bit read, but in fact both ports have a 20-bit data input and a 20-bit data output, so this can function as a 8x40-bit SDP.

I've plumbed support into nextpnr as YosysHQ/nextpnr#1169, so let's enable it in Yosys as well.


Ravenslofty:
The CI failure seems spurious?

mmicko:
@Ravenslofty yes, on macOS it fails from time to time, just check one on linux

Ravenslofty:
Well, given that it was the only one to fail, I'm going to merge this.

PR #3791, opened 2023-05-31
show -colorattr: extend colors to arrows when wires have attribute

nakengelhardt:
This updates the show command to use similar logic for the `-color <color> <selection>` and `-colorattr <attr>` options. Both will now color the cells, wires, and connection arrows for the designated elements.

Also adds a `-viewer none` option to disable opening the viewer (this is already possible indirectly because specifying `-format <foo>` disables the viewer).

Also fixes a bug where the `:e` and `:w` compass port options would be added twice on some nodes (graphviz doesn't mind the redundancy so this didn't change anything functionally).

### DIFF
--- a/passes/cmds/show.cc
+++ b/passes/cmds/show.cc
@@ -106,7 +106,7 @@ struct ShowWorker
 	{
 		std::string color = findColor(conn);
 		if (!color.empty()) return color;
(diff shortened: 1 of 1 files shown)


clairexen (approved):
LGTM

Issue #525, opened 2018-04-04
yosys-smtbmc: Error when using "--stbv" option


blazra:
I am using symbiyosys and when I add --stbv option to the smtbmc engine it crashes. Without the option it seems to work fine. (I am not sure what the option does, I was just trying it out, but this doesn't seem to be correct behavior)
With yices it says:
>File "/usr/local/bin/yosys-smtbmc", line 392, in <module>
smt.write(line)
File "/usr/local/bin/../share/yosys/python3/smtio.py", line 409, in write
self.p_write(stmt + "\n", True)
File "/usr/local/bin/../share/yosys/python3/smtio.py", line 293, in p_write
if flush: self.p.stdin.flush()
BrokenPipeError: [Errno 32] Broken pipe

and with z3:
>Solver Error: (error "line 564 column 64: invalid declaration, named expression already defined with this name testbench_h uut")

Complete outputs:
With yices:

> SBY 15:21:14 [trigger_rules] Removing direcory 'trigger_rules'.
SBY 15:21:14 [trigger_rules] Copy 'trigger_rules.vhd' to 'trigger_rules/src/trigger_rules.vhd'.
SBY 15:21:14 [trigger_rules] Copy 'formal_tb.sv' to 'trigger_rules/src/formal_tb.sv'.
SBY 15:21:14 [trigger_rules] engine_0: smtbmc --stbv yices
SBY 15:21:14 [trigger_rules] base: starting process "cd trigger_rules/src; yosys -ql ../model/design.log ../model/design.ys"
SBY 15:21:14 [trigger_rules] base: -- (c) Copyright 1999 - 2018 Verific Design Automation Inc. All rights reserved
SBY 15:21:14 [trigger_rules] base: -- Compilation time was Thu Mar 29 21:26:25 2018
SBY 15:21:14 [trigger_rules] base: -- This program will expire on Sat Jul  7 21:26:25 2018
SBY 15:21:24 [trigger_rules] base: finished (returncode=0)
SBY 15:21:24 [trigger_rules] smt2_stbv: starting process "cd trigger_rules/model; yosys -ql design_smt2_stbv.log design_smt2_stbv.ys"
SBY 15:21:24 [trigger_rules] smt2_stbv: -- (c) Copyright 1999 - 2018 Verific Design Automation Inc. All rights reserved
SBY 15:21:24 [trigger_rules] smt2_stbv: -- Compilation time was Thu Mar 29 21:26:25 2018
[...]

PR #3797, opened 2023-06-07
smt2: if abits == 0, use a (_ BitVec 1) anyway and force #b0

### LINKED ISSUE #2577: ERROR: Assert `GetSize(subexpr) == 1' failed in backends/smt2/smt2.cc:410.
When running the following yosys script

```
read_ilang design.il
dffunmap
stat
write_smt2 -wires design_smt2.smt2
```
I get the following assertion error:
```
ERROR: Assert `GetSize(subexpr) == 1' failed in backends/smt2/smt2.cc:410. 
```

I minimized my testcase to the following using `bugpoint` (and a manual call of `opt` after `bugpoint` was done)
```
# Generated by Yosys 0.9+3755 (git sha1 UNKNOWN, g++ 7.5.0 -fPIC -Os)
autoidx 3073
attribute \keep 1
attribute \nmigen.hierarchy "top.dut.last_fifo.fifo"
attribute \generator "nMigen"
module \fifo
  attribute $bugpoint 1
  wire width 3 output 1 $auto$bugpoint.cc:253:simplify_something$306
  cell $mem \storage
    parameter \ABITS 0
    parameter \INIT 3'000
    parameter \MEMID "\\storage"
    parameter \OFFSET 0
    parameter \RD_CLK_ENABLE 1'0
    parameter \RD_CLK_POLARITY 1'1
    parameter \RD_PORTS 1
    parameter \RD_TRANSPARENT 1'1
    parameter \SIZE 1
    parameter \WIDTH 3
    parameter \WR_CLK_ENABLE 1'1
    parameter \WR_CLK_POLARITY 1'1
    parameter \WR_PORTS 1
    connect \RD_ADDR { }
    connect \RD_CLK 1'x
    connect \RD_DATA $auto$bugpoint.cc:253:simplify_something$306
    connect \RD_EN 1'x
    connect \WR_ADDR { }
    connect \WR_CLK 1'x
    connect \WR_DATA 3'xxx
    connect \WR_EN 3'xxx
  end
end
```

[Here](https://github.com/YosysHQ/yosys/files/5937343/design.il.txt) you can find the full design aswell.

charlottia:
Fixes #2577.

I first explored trying to do away with the address line related stuff entirely here, changing the state from `(Array (_ BitVec <abits>) (_ BitVec <width>))` to just `(_ BitVec <width>)` where abits == 0, drop the unnecessary `(select ...)` and `(store ...)` etc., but then ran into trouble where smtbmc interacts with related entities.

It would be neater to use a single-member sort for the key (is there a "Unit" equivalent?) instead of `(_ BitVec 1)` where we only use the `#b0` index and not `#b1`, but this does get things working with minimally invasive changes.

Tested with regular stat, `--stdt`, and `--stbv`; it looks like there's another latent stbv bug (probably similar to #525) I noted while testing.  I think it's because we `makebits()` for the "#final" memstate variant at both lines 1245 and 801.


charlottia:
> I'm not sure if changing this to `max(1, ceil_log2(mem->size))` would make it work again or produce broken smt2 output (I only checked that it doesn't hit an assertion so far), although in that case that might point to a general issue with handling SIZE != 1 << ABITS memories in the smt2 backend.

This is kinda essentially what this PR does anyway — I'm forcing a minimum of 1 (which means it uses a 1-length bitvector for the address line) and then always using `0` for the index. 

I've investigated trying this instead, and you know what? It works perfectly, and does exactly what I was trying to do here but with that much less code.

Because `abits == 1`, the generated logic uses a `(_ BitVec 1)` for the address line, same as I do. Without anything extra, we'd still assert, but because we call `addr_sig.extend_u0(abits)` in all cases before attempting `get_bv(addr_sig)`, the signal gets padded with a constant `S0`, meaning the logic for the address line calculation is written out as `#b0`:

https://github.com/YosysHQ/yosys/blob/8b2a0010216f9a15c09bd2f4dc63691949b126df/backends/smt2/smt2.cc#L375-L381

I've tested this with my test cases, and the generated logic is the same as this PR currently is (modulo slightly different comments), and it works correctly:

```smt
(define-fun |in_fifo_m:R0A storage| ((state |in_fifo_s|)) (_ BitVec 1) #b0) ; 1'0
; ...
(define-fun |in_fifo_m:W0A storage| ((state |in_fifo_s|)) (_ BitVec 1) #b0) ; 1'0
```

Given how simple this fix is, and how the code seems to already anticipate(/expect!) it with the zero-padding already being done, I'm inclined to suggest this is probably _the_ fix.

I'm not sure if `clean_zerowidth` angle would end up being more maintainable going forward — likewise, I can't foretell the implications of the `memory_map` call and how it might interact with various modes of `write_smt2` (i.e. default vs `stbv` vs `stdv`).

> I haven't run into any zero-width signals/memories so far (is that even possible with the verilog frontend?) so I might also be missing some downsides of my proposed alternatives.

I'm a bit of a noob so take my observations with a grain of salt, but quite possibly not! The ultimate cause of this is a `memory [...] size 1`, which has no address given it only has a single entry. It is therefore barely a memory and probably has no real representation in input Verilog; it's being generated from a FIFO in Amaranth and so things would change a lot in how it all gets synthesised once I bump up the size there.

charlottia:
I've force-pushed this PR to have the suggested fix; it's at least strictly better than what it was.

PR #3813, opened 2023-06-21
ast/simplify: Remove in_lvalue/in_param simplify() parameters

povik:
Split off from PR #3722, ~~but it's still on top of the other PR's commits to save on rebase work. So, only the last three commits, starting with "[ast/simplify: Make tweaks in advance of big in_lvalue/in_param change](https://github.com/YosysHQ/yosys/commit/31b7b2e41c53ab59266e69f613f3b948565299dc)" should be the subject of this PR.~~

As is, the change might be objectionable since it requires the code to be sprinkled with calls to fixup_hierarchy_flags and set_attribute (new methods). That may not be as bad if we (1) don't require the in_param/in_lvalue values of the old code to be perfectly reproduced (which the change here does reproduce, at least as exercised by make test); (2) move to a coding style in the simplify code that doesn't rely on explicitly modifying the children list of an existing node as much. See the commits and let me know what you think.

This should be passing make test at each commit.

### DIFF
--- a/frontends/ast/ast.cc
+++ b/frontends/ast/ast.cc
@@ -1078,7 +1078,7 @@ static RTLIL::Module *process_module(RTLIL::Design *design, AstNode *ast, bool d
 		// simplify this module or interface using the current design as context
 		// for lookup up ports and wires within cells
 		set_simplify_design_context(design);
-		while (ast->simplify(!flag_noopt, false, 0, -1, false, false)) { }
+		while (ast->simplify(!flag_noopt, 0, -1, false)) { }
 		set_simplify_design_context(nullptr);
 
 		if (flag_dump_ast2) {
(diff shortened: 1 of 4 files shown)


povik:
Rebased, shaked off the commits that were merged in PR #3722

zachjs (approved):
I like this change. The only downside I see is that one might easily forget to call `fixup_hierarchy_flags()` when working on new code that modifies `children`, but I think this overhead will be small compared to the benefit. Thank you!

povik:
Rebased, but as it is now those old way/new way equivalence asserts between the second and third commit fail at some point in the test suite. Maybe I will investigate. (Those asserts are removed in the final commit so it's an option to merge this as is.)

povik:
> Rebased, but as it is now those old way/new way equivalence asserts between the second and third commit fail at some point in the test suite.

Should be resolved now (as expected, it was due to interaction with other AST changes in the meantime).

zachjs (approved):
Thanks for fixing the semantic conflicts in the intermediate commit! This still looks good to me.

povik:
If that's the case, let's do it.

PR #3817, opened 2023-06-23
check: Also check for conflicts with constant drivers

jix:
Apparently we didn't check for conflicts with constant driver at all.

This adds checks for 0,1 and x drivers conflicting with other drivers. It doesn't check for z drivers as the already existing checks also didn't count inout ports as drivers. I'm not certain whether there are any cases where we do want to allow conflicts with x drivers, though. (@clairexen any idea?)


clairexen:
> I'm not certain whether there are any cases where we do want to allow conflicts with x drivers, though. (@clairexen any idea?)

Nothing off the top of my head, no.

PR #3822, opened 2023-06-25
feat(parser): RFC for concurrent assertions

tianrui-wei:
This commit adds initial experiemental support for concurrent SVA in the
open source parser

Signed-off-by: Tianrui Wei <tianrui@tianruiwei.com>


tianrui-wei:
Hi,

This is an RFC PR for collecting feedback in adding concurrent SVA assertion support for the open source verilog parser, as well as performing transformations for the backend formal tools/synthesis targeting FPGAs. It'd be great if people could share how/what things should be added, as well as their opinion on coding style etc in general.

Thanks,
Tianrui

Makefile
```diff
@@ -10,7 +10,7 @@ CONFIG := clang
 
 # features (the more the better)
 ENABLE_TCL := 1
-ENABLE_ABC := 1
+ENABLE_ABC := 0
```
Ravenslofty, line 13: I don't think this should have been included.

Makefile
```diff
@@ -31,7 +31,7 @@ ENABLE_PYOSYS := 0
 # other configuration flags
 ENABLE_GCOV := 0
 ENABLE_GPROF := 0
-ENABLE_DEBUG := 0
+ENABLE_DEBUG := 1
```
Ravenslofty, line 34: Likewise, I don't think this should have been included.

Ravenslofty:
I think you should remove your Makefile changes from the commit; if you need to make local changes to your Makefile, put them in `Makefile.conf` instead.

tianrui-wei:
Thank you @Ravenslofty for reviewing! I've already reverted changes to Makefile :)

Ravenslofty:
You should amend the original commit instead of making a new commit.

tianrui-wei:
@Ravenslofty Thanks for the feedback! Already fixed!

tianrui-wei:
Hi All,

I was trying the PR locally and it seems to be working reasonably well for the intended use case, I was wondering if you could review and hopefully merge this PR in preparation for implementing a larger subset?

Thanks,
Tianrui

PR #3824, opened 2023-06-28
cxxrtl: add `$divfloor`.

charlottia:
Does what it says on the tin.

I'm not sure if just subtracting one in the negative-result non-zero-remainder case is the "most correct" way of representing this operation, but it produces the result!


whitequark:
Thank you!

PR #3826, opened 2023-06-28
memory_libmap: print message about attributes forcing ram kind

nakengelhardt:
This gives additional context that helps explain why mapping failed if the requested ram kind does not exist on the target architecture.


whitequark (approved):
Nice!

PR #3840, opened 2023-07-10
cellaigs: Generate models for integer comparison cells

povik:
Add the case of $lt, $le, $gt, $ge to the code generating AIGs.

### DIFF
--- a/kernel/cellaigs.cc
+++ b/kernel/cellaigs.cc
@@ -385,6 +385,26 @@ Aig::Aig(Cell *cell)
 		goto optimize;
 	}
 
+	if (cell->type.in({ID($lt), ID($gt), ID($le), ID($ge)}))
+	{
+		int width = std::max(GetSize(cell->getPort(ID::A)),
+							 GetSize(cell->getPort(ID::B))) + 1;
+		vector<int> A = mk.inport_vec(ID::A, width);
+		vector<int> B = mk.inport_vec(ID::B, width);
+
+		if (cell->type.in({ID($gt), ID($ge)}))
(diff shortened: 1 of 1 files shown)


kernel/cellaigs.cc
```diff
@@ -385,6 +385,29 @@ Aig::Aig(Cell *cell)
 		goto optimize;
 	}
 
+	if (cell->type.in({ID($lt), ID($gt), ID($le), ID($ge)}))
+	{
+		int width = std::max(
+			GetSize(cell->getPort(ID::A)) + cell->getParam(ID::A_SIGNED).as_int(false),
+			GetSize(cell->getPort(ID::B)) + cell->getParam(ID::B_SIGNED).as_int(false)
```
jix, line 392: AFAICT the `+ cell->getParam(ID::?_SIGNED).as_int(false)` are unnecessary (and confused me at first, as I didn't see how it would help handle the signed case) as the `+1` below is enough to handle the signed and unsigned case in a uniform way. I checked this using

```
read_verilog -sv <<EOF
    module top(input [3:0] a, input [3:0] b);
        wire y1 = $signed(a) <= $signed(b);
        wire y2 = $signed(a) <= $signed(b);
        always @* assert (y1 == y2);
    endmodule
EOF

hierarchy -top top; proc
aigmap */y1 %ci2
sat -prove-asserts -verify
```

as well as some variations of that using other ops and the unsigned case, which might also be useful tests to add.

povik: Agreed, not sure what the reasoning was there. Pushed a new commit.

PR #3841, opened 2023-07-10
tests: Extend aigmap.ys with SAT comparison

povik:
For the test to pass this depends on PRs #3838 (for `$_NMUX_`) and #3840 (for `$lt` and friends).

From the commit:

> Extend the aigmap.ys test with SAT-based comparison of the original cells and their AIG implementations.
>
> This tests both the usual cells and the single-bit Yosys gates.

As I have found out after having written this, the lowercase cells can also be tested with `test_cell -aigmap`, though that does not seem to check that the cells are actually mapped onto an and-inverter network (and not kept as they are).

### DIFF
--- a/tests/techmap/aigmap.ys
+++ b/tests/techmap/aigmap.ys
@@ -1,3 +1,118 @@
+read_verilog -icells <<EOF
+module test();
+`define CELL_AY(typ)  \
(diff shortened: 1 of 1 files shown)


povik:
(Fixed typo)

PR #3854, opened 2023-07-19
kernel/mem: Assert ABITS is not below wide_log2

povik:
Later in the check() code we check the bottom wide_log2 bits on the address port are zeroed out. If the address port is too narrow, we crash due to out of bounds access. Explicitly assert the address port is wide enough, so we don't crash on input such as

    read_rtlil <<EOF
    module \top
      wire input 1 \clk

      memory width 8 size 2 \mem

      cell $memwr $auto$:1:$8
        parameter \PRIORITY 1'0
        parameter \CLK_POLARITY 1'1
        parameter \CLK_ENABLE 1'1
        parameter \MEMID "\\mem"
        parameter \ABITS 1'0
        parameter \WIDTH 6'010000
        connect \DATA 16'0000000000000000
        connect \ADDR { }
        connect \EN 16'0000000000000000
        connect \CLK \clk
      end
    end
    EOF
    memory

### DIFF
--- a/kernel/mem.cc
+++ b/kernel/mem.cc
@@ -451,6 +451,7 @@ void Mem::check() {
 		log_assert(GetSize(port.en) == 1);
 		log_assert(GetSize(port.arst) == 1);
 		log_assert(GetSize(port.srst) == 1);
+		log_assert(GetSize(port.addr) >= port.wide_log2);
 		log_assert(GetSize(port.data) == (width << port.wide_log2));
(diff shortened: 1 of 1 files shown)


Ravenslofty:
A piece of personal philosophy is that if an assert *ever* fires, it should be immediately replaced with a more helpful call to `log_error` that explains the problem. Assertions are not friendly for an end-user.

PR #3859, opened 2023-07-22
abc, abc9_exe: respect `-q` when built with linked ABC

whitequark:
This is mostly important for YoWASP builds, since those do not have a way to build with external ABC (I prototyped it but for some reason ABC always segfaults when built as an independent Wasm binary...)

### DIFF
--- a/passes/techmap/abc.cc
+++ b/passes/techmap/abc.cc
@@ -1091,10 +1091,17 @@ void abc_module(RTLIL::Design *design, RTLIL::Module *current_module, std::strin
 		buffer = stringf("\"%s\" -s -f %s/abc.script 2>&1", exe_file.c_str(), tempdir_name.c_str());
 		log("Running ABC command: %s\n", replace_tempdir(buffer, tempdir_name, show_tempdir).c_str());
 
-#ifndef YOSYS_LINK_ABC
 		abc_output_filter filt(tempdir_name, show_tempdir);
+#ifndef YOSYS_LINK_ABC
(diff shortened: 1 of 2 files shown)


whitequark:
For context, right now using `yowasp-yosys -q` results in this:

```
$ yowasp-yosys -q test.v -p synth_ice40
ABC command line: "source /tmp/yosys-abc-000000/abc.script".

+ read_blif /tmp/yosys-abc-000000/input.blif
+ read_lut /tmp/yosys-abc-000000/lutdefs.txt
+ strash
+ &get -n
+ &fraig -x
+ &put
+ scorr
Warning: The network is combinational (run "fraig" or "fraig_sweep").
+ dc2
+ dretime
+ strash
+ dch -f
+ if
+ mfs2
+ lutpack -S 1
+ dress /tmp/yosys-abc-000000/input.blif
Total number of equiv classes                =       2.
Participating nodes from both networks       =       2.
Participating nodes from the first network   =       1. (  50.00 % of nodes)
Participating nodes from the second network  =       1. (  50.00 % of nodes)
Node pairs (any polarity)                    =       1. (  50.00 % of names can be moved)
Node pairs (same polarity)                   =       1. (  50.00 % of names can be moved)
Total runtime =     0.00 sec
+ write_blif /tmp/yosys-abc-000000/output.blif
```

This is quite annoying when the build runs as a part of e.g. the Glasgow software stack, and it can also be really confusing/scary to novice users of Amaranth. Since both Amaranth and ChipFlow now suggest using YoWASP builds of the tools, this needed to be fixed.

PR #3860, opened 2023-07-23
abc, abc9_exe: fix build on WASI (and others with `const* stdout`).

whitequark:
It turns out that https://github.com/YosysHQ/yosys/pull/3859 didn't actually build on WASI (or musl). My bad for not testing it more thoroughly; I forgot that `stdout`/`stderr` can be `const`.

C does not guarantee that stdout/stderr can be reassigned. Most platforms do make them assignable, however musl and WASI that is based on musl do not. WASI does not have `dup2()`; instead it has its own non-portable version of it that can only assign to previously allocated fds.

Update the stream redirection code so that it does the right thing on WASI and other platforms.


mwkmwkmwk (approved):
absolutely horrible, but I don't see a way to improve on this

PR #3866, opened 2023-07-27
ice40: Implement simple SB_HFOSC and SB_LFOSC for simulation

byteit101:
This isn't a full implementation (notably ignoring trim and startup being unstable), but good enough to simulate designs that use it as a simple clock

### DIFF
--- a/techlibs/ice40/cells_sim.v
+++ b/techlibs/ice40/cells_sim.v
@@ -2678,6 +2678,37 @@ module SB_HFOSC(
 );
 parameter TRIM_EN = "0b0";
 parameter CLKHF_DIV = "0b00";
+
+`ifndef BLACKBOX
+	// 96mhz hidden clock
+	reg hidden_clock = 0;
+
+	// all 4 options for the clock divider
+	reg [3:0] divisions = 0;
+
+	always begin
+		while (CLKHFPU) begin
+			#5208 hidden_clock = 0;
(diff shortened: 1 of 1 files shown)


techlibs/ice40/cells_sim.v
```diff
@@ -2678,6 +2678,37 @@ module SB_HFOSC(
...
+	// TODO: simulate the 100us of "not yet ready"/power stability early on
+
+	always @(posedge hidden_clock) begin
+		divisions <= divisions + 1;
+	end
+
+	localparam index = CLKHF_DIV == "0b00" ? 0 : 
+				   CLKHF_DIV == "0b01" ? 1 : 
+				   CLKHF_DIV == "0b10" ? 2 : 
+				   CLKHF_DIV == "0b11" ? 3 : 0;
```
whitequark, line 2707: Not sure this is right--is it illegal to pass a non-string here? What do the vendor simulation models do?

byteit101: I must admit I was confused they were strings too when I saw them in the vendor documentation. I haven't installed the proprietary toolchains, so can't comment on them

whitequark: I have to ask you to install the toolchain; I feel that having subtly incorrect simulation models is worse than not having them at all.

Issue #3833, opened 2023-07-06
part-selects (`$shift`, `$shiftx`) create huge MUX-trees


phsauter:
### Version

Yosys 0.28 (git sha1 0d6f4b068, gcc 11.2.0 -fPIC -Os)

### On which OS did this happen?

Linux

### Reproduction Steps

[shift_issue.zip](https://github.com/YosysHQ/yosys/files/11967121/shift_issue.zip)
```
make all
```
Then check the reports in output and compare them.

### Expected Behavior

The two modules should be about the same size after synthesis as they do the same thing.

### Actual Behavior

The module using part-select is about twice as large as the other one.

---

#### Background

During synthesis of our (soon-to-be) open-source [SoC Iguana](https://github.com/pulp-platform/iguana), we noticed that indexed part-selects like `array[i*33+:33]` leads to a very large number of muxes during techmap.
A large amount of them get optimized away later on but:

- During techmap the ballooned number of mux cells increases the memory footprint substantially
- Not all of them get optimized away, increasing the total cell area significantly

In the worst modules this leads to a roughly 4x-ish larger module than what we would expect.

#### Example

Attached I have a rather simple example resulting from converting [the register-interface demux module](https://github.com/pulp-platform/register_interface/blob/master/src/reg_demux.sv) using [sv2v](https://github.com/zachjs/sv2v).
The verilog file has two modules, they should be identical in function but in one I replaced the problematic part-select with another construct.
The original module has a roughly 2x larger area than the manually rewritten one.
(it should be noted that a normal invocation of `techmap` would produce even worse results but the way `techmap` is called right now aids in seeing the problem).

To generate the netlists and reports to see the problem, run:

```bash
make all
```

Per default this will download the [IHP13-PDK](https://github.com/IHP-GmbH/IHP-Open-PDK), you may change the used `.lib` to any other, it should not matter.

#### What we think happens

As far as we can tell the problem seems to be an inherent problem of how `$shift` and `$shiftx` work. They work with a generic shift amount, by which the data can be shifted, it essentially looses the information about how the parts in a part-select are grouped.
Together with their implementation in `techmap.v`, using a logarithmic barrel shifter, seems to result in yosys building large mux trees and then failing to optimize them completely later on.

So in an expression like `array[i*33+:33]`, I know that always 33 bit belong together and any other shift value than a multiple of 33 is impossible. 
This information seems to be lost and it just shifts by a generic amount, implemented using log-shifters.
[...]

PR #3873, opened 2023-08-02
Extend the `peepopt` shiftmul matcher

povik:
This extends the shiftmul matcher in two useful ways:

 * It deals with cases where the shift amount is zero-padded from below because `opt_expr` ran before and applied `opt.opt_expr.mul_low_zeros`.
 * It can deal with left shifts implementing a demuxer (through the addition of a mirrored matcher).

Related to #3833.

### DIFF
--- a/passes/pmgen/peepopt.cc
+++ b/passes/pmgen/peepopt.cc
@@ -26,7 +26,6 @@ PRIVATE_NAMESPACE_BEGIN
 bool did_something;
 
 #include "passes/pmgen/peepopt_pm.h"
-#include "generate.h"
 
 struct PeepoptPass : public Pass {
 	PeepoptPass() : Pass("peepopt", "collection of peephole optimizers") { }
@@ -41,31 +40,14 @@ struct PeepoptPass : public Pass {
 	}
 	void execute(std::vector<std::string> args, RTLIL::Design *design) override
 	{
-		std::string genmode;
-
 		log_header(design, "Executing PEEPOPT pass (run peephole optimizers).\n");
 
 		size_t argidx;
 		for (argidx = 1; argidx < args.size(); argidx++)
 		{
-			if (args[argidx] == "-generate" && argidx+1 < args.size()) {
-				genmode = args[++argidx];
-				continue;
-			}
 			break;
 		}
 		extra_args(args, argidx, design);
-
-		if (!genmode.empty())
-		{
-			if (genmode == "shiftmul")
-				GENERATE_PATTERN(peepopt_pm, shiftmul_right);
-			else if (genmode == "muldiv")
-				GENERATE_PATTERN(peepopt_pm, muldiv);
-			else
-				log_abort();
-			return;
-		}
 
 		for (auto module : design->selected_modules())
 		{
--- a/passes/pmgen/peepopt.cc
+++ b/passes/pmgen/peepopt.cc
@@ -59,7 +59,7 @@ struct PeepoptPass : public Pass {
 		if (!genmode.empty())
 		{
 			if (genmode == "shiftmul")
-				GENERATE_PATTERN(peepopt_pm, shiftmul);
+				GENERATE_PATTERN(peepopt_pm, shiftmul_right);
 			else if (genmode == "muldiv")
 				GENERATE_PATTERN(peepopt_pm, muldiv);
 			else
@@ -79,7 +79,8 @@ struct PeepoptPass : public Pass {
 
 				pm.setup(module->selected_cells());
 
-				pm.run_shiftmul();
+				pm.run_shiftmul_right();
+				pm.run_shiftmul_left();
 				pm.run_muldiv();
 			}
 		}
--- a/passes/pmgen/peepopt.cc
+++ b/passes/pmgen/peepopt.cc
@@ -24,8 +24,6 @@ USING_YOSYS_NAMESPACE
 PRIVATE_NAMESPACE_BEGIN
 
 bool did_something;
-dict<SigBit, State> initbits;
-pool<SigBit> rminitbits;
 
 #include "passes/pmgen/peepopt_pm.h"
 #include "generate.h"
@@ -60,9 +58,6 @@ struct PeepoptPass : public Pass {
 
 		if (!genmode.empty())
 		{
-			initbits.clear();
-			rminitbits.clear();
-
 			if (genmode == "shiftmul")
 				GENERATE_PATTERN(peepopt_pm, shiftmul);
 			else if (genmode == "muldiv")
@@ -79,47 +74,13 @@ struct PeepoptPass : public Pass {
 			while (did_something)
 			{
 				did_something = false;
-				initbits.clear();
-				rminitbits.clear();
 
 				peepopt_pm pm(module);
-
-				for (auto w : module->wires()) {
-					auto it = w->attributes.find(ID::init);
-					if (it != w->attributes.end()) {
-						SigSpec sig = pm.sigmap(w);
-						Const val = it->second;
-						int len = std::min(GetSize(sig), GetSize(val));
-						for (int i = 0; i < len; i++) {
-							if (sig[i].wire == nullptr)
-								continue;
-							if (val[i] != State::S0 && val[i] != State::S1)
-								continue;
-							initbits[sig[i]] = val[i];
-						}
-					}
-				}
 
 				pm.setup(module->selected_cells());
 
 				pm.run_shiftmul();
 				pm.run_muldiv();
-
-				for (auto w : module->wires()) {
-					auto it = w->attributes.find(ID::init);
-					if (it != w->attributes.end()) {
-						SigSpec sig = pm.sigmap(w);
-						Const &val = it->second;
-						int len = std::min(GetSize(sig), GetSize(val));
-						for (int i = 0; i < len; i++) {
-							if (rminitbits.count(sig[i]))
-								val[i] = State::Sx;
-						}
-					}
-				}
-
-				initbits.clear();
-				rminitbits.clear();
 			}
 		}
 	}
--- a/passes/pmgen/peepopt.cc
+++ b/passes/pmgen/peepopt.cc
@@ -37,6 +37,17 @@ struct PeepoptPass : public Pass {
 		log("\n");
 		log("This pass applies a collection of peephole optimizers to the current design.\n");
 		log("\n");
+		log("This pass employs the following rules:\n");
+		log("\n");
+		log("   * muldiv - Replace (A*B)/B with A\n");
+		log("\n");
+		log("   * shiftmul - Replace A>>(B*C) with A'>>(B<<K) where C and K are constants\n");
+		log("                and A' is derived from A by appropriately inserting padding\n");
+		log("                into the signal. (right variant)\n");
+		log("\n");
+		log("                Analogously, replace A<<(B*C) with appropriate selection of\n");
+		log("                output bits from A<<(B<<K). (left variant)\n");
+		log("\n");
 	}
 	void execute(std::vector<std::string> args, RTLIL::Design *design) override
 	{
--- a/passes/pmgen/peepopt_shiftmul_left.pmg
+++ b/passes/pmgen/peepopt_shiftmul_left.pmg
@@ -1,0 +1,160 @@
+pattern shiftmul_left
+//
+// Optimize mul+shift pairs that result from expressions such as foo[s*W+:W]
+//
+
+match shift
+	select shift->type.in($shift, $shiftx, $shl)
+	select shift->type.in($shl) || param(shift, \B_SIGNED).as_bool()
+	filter !port(shift, \B).empty()
+endmatch
+
+match neg
+	if shift->type.in($shift, $shiftx)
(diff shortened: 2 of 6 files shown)


povik:
Force pushed. There was a confusion in the commit message about whether it's the left or right shift matcher we are adding.

phsauter:
As we already started to discuss in #3833, this currently only handles: `foo[s*W+:W]`, where `W` is a constant.

Since yesterday I went through different scenarios/cases I could see happening and thought about how to handle them.
For this I will generally refer to `foo` as an array of multi-bit elements as this is probably the most common origin of such a construct.

- `foo[s*W+:W1]` where `W1<W`. This is the case when only a few bits are selected from some array of elements or similar. This case is trivial as one can just add more padding in-between output bits but currently It seems to be that `shift_amount` must exactly match the shifter output-width for it to then apply the transformation.
- `foo[s*W+:W2]` where `W2>W`. My guess is that this case would not occur very often but it covers the case where one would like to select something from two consecutive elements from an array of elements. Personally I think this case should be noted but does not need to be handled right now. (the transformation here should however work basically the same except that we need to use `clog2(W2)` instead.
- `foo[s*W+O+:W1]` where `O` is a constant. This case describes the selection of a few bits from one element of the array. The offset gives the starting position of the bits. As I understand it, your proposal was to handle this separately in a `shiftadd` `peepot`, so that we then have two shifts, one being directly attached to a `$mul`, making it possible to use `shiftmul`. I think this is a good idea but it does have one minor caveat, lets again distinguish two options:
    1. We place the constant-offset shift in-front of the soon-to-be `shiftmul` block. This always works but it has the slight disadvantage that multiple accesses to different bits from the same element, may now be converted into `shiftmul` blocks, each having a different offset at its input and we are going to rely on later optimization passes to optimize this for us. If it is caught before `techmap` I would say this is a non-problem. If it is only caught after, then we create a lot of `$MUX` cells for nothing, hurting performance and increasing the memory footprint.
    2. We place the constant-offset shift after the `shiftmul` block. This would eliminate the minor caveat from (1) but it only works if `W>(O+W1)`, which should be the case if the code and all previous transformations are reasonable but we cannot guarantee it. Meaning such a transformation could not be done independently and would have to be called as part of `shiftmul`, since we need to know `W` from the `mul`.
- `foo[s*W+O+:W1]` where `O=V*var`. I do not expect this to occur often (if at all). This represents the case where first an element is selected from an array and then some starting point is selected, from which `W1` bits are used. Such a scenario would likely also require a more holistic approach as we now essentially have two nested shifts and we would have to find out which is larger and perform that first. It is also possible that some earlier part might actually already split this into two shifts for us (though I doubt it). Again, I think this is uncommon enough to make a note of it but ignore it.


Tl;Dr: assuming `foo[s*W+:W1]`, we should check which of `W` and `W1` is larger and add padding accordingly, `W=W1` is an arbitrary restriction.
Constant offsets can be generally dealt with by adding another shift in-front, then this pass can take care of the strided access. This may be suboptimal for performance.
More complex cases are possible but likely rare. They would likely require a monolithic approach.

povik:
>   * `foo[s*W+O+:W1]` where `O` is a constant. This case describes the selection of a few bits from one element of the array. The offset gives the starting position of the bits. As I understand it, your proposal was to handle this separately in a `shiftadd` `peepot`, so that we then have two shifts, one being directly attached to a `$mul`, making it possible to use `shiftmul`. I think this is a good idea but it does have one minor caveat, lets again distinguish two options:
>       
>     1. We place the constant-offset shift in-front of the soon-to-be `shiftmul` block. This always works but it has the slight disadvantage that multiple accesses to different bits from the same element, may now be converted into `shiftmul` blocks, each having a different offset at its input and we are going to rely on later optimization passes to optimize this for us. If it is caught before `techmap` I would say this is a non-problem. If it is only caught after, then we create a lot of `$MUX` cells for nothing, hurting performance and increasing the memory footprint.

I am not sure there's an issue there, and if there is overhead, maybe the situation can be improved by only leaving those signals connected to the shift cell that can actually be selected, and rounding up `W1` (instead of `W`) to the next power-of-2. Though at that point maybe we should just be leaving `$bmux` behind, which didn't exist when the transformation was originally written.

phsauter:
> Well but then at some point if `W2` is large and `W` is small you could be better off expressing this as `W` independent shifts, instead of one block select with a lot of replication in the input signal. That's what I am slightly curious about how other compilers handle.

Makes sense, I mentally didn't make `W2` large enough :-)
I am going to try it today or tomorrow and will report back.

povik:
> A block-mux halves the bits it works on in every stage, so its cost should be something like `block_size*2*log2(#blocks)`.

I don't think that's true since if each stage works on a different amount of bits, and that amount doubles going to earlier stages, the final cost must come down to something like `block_size*#blocks*2 - 1` (for power-of-two #blocks). Once you prune the trivial and unused muxes in the implementation of the shift I think the cost will be the same.

> One big exception is once `W2` is larger than `W` because for the bmux `block_size=max(W2,W)`, while for the shifter it doesn't really change anything.

Agreed on shift having a lower cost (in terms of elementary muxes) for the case `W2>W`.

> Personally I think there is a pretty strong case for transforming it to a $bmux anyway...

I agree. Not so much because of any overhead consideration but for clarity and to make it easier for later passes to reason about what's going on in the circuit.

phsauter:
By the way I am currently working on a shiftadd.pmg that should be able to deal with any `$add` and `$sub` cells.
Funnily enough dealing with `foo[s*W-O+:W1]` (note the minus) is actually easier since we can just prepend some fill-bits.

The more common `foo[s*W+O+:W1]` is a bit of a pain if you want to make it as generic as possible since the `+O` would make it so a previous shift makes the lowest `O` bits unreachable, this is fine in all reasonable cases but there is nothing stopping `foo[s*W]` from going into the negative just enough to still reach these now lost bits. So the transformation wouldn't be 100% the same anymore.

Currently I just limited it so that one value must be constant (`O`) and the other must be unsigned. This might not be ideal if it starts chaining multiple operations though but for now it should work. I still have some bugs though, meaning they are not quite functionally equivalent for all the test cases I am using.
I will push them as soon as they are, then you can decide if you want them in here as well.

phsauter:
> the final cost must come down to something like `block_size*#blocks*2 - 1` (for power-of-two #blocks)

Yes you are correct, it even says that on the paper in front of me, I just didn't pay attention while typing it (well I didn't have the -1 but I ignored it on purpose since its just a constant).

povik:
> By the way I am currently working on a shiftadd.pmg...

Good to know!

> Currently I just limited it so that one value must be constant (`O`) and the other must be unsigned.

Yeah, I don't see any other reasonable way to do it. `wreduce` will convert cell inputs to unsigned if it deduces that it's okay, so relying on the input being unsigned in your situation should not be too restrictive, hopefully.

> I still have some bugs though, meaning they are not quite functionally equivalent for all the test cases I am using.
I will push them as soon as they are, then you can decide if you want them in here as well.

I think it's best if you open a separate PR (a draft one if you don't feel it's ready for merging), and then we can discuss the shiftadd rule there. I don't have merging powers here but those that have will be happy we split things.

phsauter:
I will be opening a new PR, in the last user-meeting they said they prefer more small PRs since they are easier to review, exactly what you are saying as well.

Will you try to use `$bmux/$demux` now or are you still going for pure `$shift/shiftx` transformations for this PR?

I guess the second approach is probably cleaner for now, the other one should probably also be added as a separate PR.
So that in the end we have a flow that does `shiftadd`, then `shiftbmux` (if `W2/W` ratio is favorable) and then finally `shiftmul` that can take the remaining ones.

povik:
> Will you try to use `$bmux/$demux` now or are you still going for pure `$shift/shiftx` transformations for this PR?
> 
> I guess the second approach is probably cleaner for now, the other one should probably also be added as a separate PR.

Yeah, I think I will make the `$bmux/$demux` change into a separate PR.

>  So that in the end we have a flow that does `shiftadd`, then `shiftbmux` (if `W2/W` ratio is favorable) and then finally `shiftmul` that can take the remaining ones.

Agreed as a prospect if what you mean by `shiftmul` here is the transformation to produce `W` independent shifters for high `W2/W`, though I would rather name it differently -- call `shiftmul` the transformation producing bmuxes (for continuity), and name a separate rule producing the `W` independent shifters. If what you meant by `shiftmul` here is the original shiftmul transformation, then I don't think we want that at the same time as the bmux transformation, since the cost of the circuitry they produce is essentially the same (once you prune the mux network in the implementation of the shifter).

povik:
Err, I wrote something bit confusing earlier...

> > One big exception is once `W2` is larger than `W` because for the bmux `block_size=max(W2,W)`, while for the shifter it doesn't really change anything.
> 
> Agreed on shift having a lower cost (in terms of elementary muxes) for the case `W2>W`.

I was thinking the *untransformed* shifter has a lower cost than the block-mux, but that is without accounting for the cost of the `$mul`... Once you shiftmul-transform the shifter to remove the `$mul` then the cost of the shifter and block-mux should be the same.

povik:
So as I see it we have three options we need to decide between when stumbling on a shiftmul pattern: Transform to `$bmux`, transform to `W` independent shifters, do nothing.

phsauter:
Your naming scheme is fine as well, I just think its important to keep the original (ie `shift+mul -> shift`) to make sure people can decide not to use the `shift+mul -> bmux` transformation if that makes any trouble for them (some potential later pass that relies on the `$shiftx` cells specifically.

> So as I see it we have three options we need to decide between when stumbling on a shiftmul pattern: Transform to $bmux, transform to W independent shifters, do nothing.

Yes exactly, except that we don't really decide. Each option is just a separate `peepopt` that checks for the respective `W2/W` ratio.

passes/pmgen/peepopt.cc
phsauter, line 1: It looks like this cleans up the removal of the dffmux pattern in a0e99a9f3f0a52e3a6f2000791d93b54ce116928
Originally added in b8774ae849cb6aa54a852a245f8634afaac1eb76

The changes (removals) concerning init look fine to me.

passes/pmgen/peepopt.cc
```diff
@@ -60,11 +58,8 @@ struct PeepoptPass : public Pass {
 
 		if (!genmode.empty())
 		{
-			initbits.clear();
-			rminitbits.clear();
-
 			if (genmode == "shiftmul")
-				GENERATE_PATTERN(peepopt_pm, shiftmul);
+				GENERATE_PATTERN(peepopt_pm, shiftmul_right);
```
phsauter, line 62:
```suggestion
				log_error("Abort in %s:%d: peepopt-shiftmul does not implement a generate option\n", __FILE__, __LINE__)
```

Doesn't this require a generate block in the pattern to actually work?

Running it ends with:
```
ERROR: pmgen generator is stuck: 10000 iterations with no matching module generated.
```

I think the most relevant PR is #1298.

phsauter:
The two actual code changes should make it pass the current tests.
The generate change is more a suggestion that also applies to `muldiv`.

Personally I also think we should add `log_header()` for all pmg-pases that describe the transformation in 1-2 sentences.

passes/pmgen/peepopt_shiftmul_left.pmg
```diff
@@ -0,0 +1,159 @@
...
+// Optimize mul+shift pairs that result from expressions such as foo[s*W+:W]
+//
+
+match shift
+	select shift->type.in($shift, $shiftx, $shl)
+	select shift->type.in($shl) || param(shift, \B_SIGNED).as_bool()
+	filter !port(shift, \B).empty()
+endmatch
+
+match neg
```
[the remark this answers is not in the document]
povik: Ha! Of course

passes/pmgen/peepopt_shiftmul_left.pmg
```diff
@@ -0,0 +1,159 @@
...
+code shift_amount log2scale
+	if (neg) {
+		// case of `$shift`, `$shiftx`
+		shift_amount = port(neg, \A);
+		if (!param(neg, \A_SIGNED).as_bool())
+			shift_amount.append(State::S0);
+	} else {
+		// case of `$shl`
+		shift_amount = port(shift, \B);
+		if (!param(neg, \B_SIGNED).as_bool())
```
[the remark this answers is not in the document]
povik: Good catch

passes/pmgen/peepopt.cc
povik: Yes, can add that to the commit message

passes/pmgen/peepopt.cc
povik: I don't know, I never looked into the generate stuff. Then maybe we should remove the generate flag altogether, I don't think muldiv has any generate blocks either.

povik:
Thanks. I think we should mainly expand the help message, not sure if the patterns each warrant their `log_header`

phsauter:
> Thanks. I think we should mainly expand the help message, not sure if the patterns each warrant their `log_header`

yeah that makes more sense, help it is!

povik:
Rebased, squashed the fixup into its parent commit, last three commits are new.

PR #3875, opened 2023-08-04
Optimization of nowrshmsk

daglem:
Generation of minimal AST_CASE for variable slices on the form dst[i*stride +: width] = src.

Corrections of corner cases, now correctly tested by tests/verilog/dynamic_range_lhs.v

### DIFF
--- a/frontends/ast/ast.h
+++ b/frontends/ast/ast.h
@@ -321,6 +321,9 @@ namespace AST
 		static AstNode *mkconst_str(const std::vector<RTLIL::State> &v);
 		static AstNode *mkconst_str(const std::string &str);
 
+		// helper function to create an AST node for a temporary register
+		AstNode *mktemp_logic(const std::string &name, AstNode *mod, bool nosync, int range_left, int range_right, bool is_signed);
+
 		// helper function for creating sign-extended const objects
 		RTLIL::Const bitsAsConst(int width, bool is_signed);
 		RTLIL::Const bitsAsConst(int width = -1);
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -35,11 +35,24 @@
 #include <stdarg.h>
 #include <stdlib.h>
 #include <math.h>
+// For std::gcd in C++17
+// #include <numeric>
 
 YOSYS_NAMESPACE_BEGIN
 
 using namespace AST;
 using namespace AST_INTERNAL;
+
+// gcd computed by Euclidian division.
+// To be replaced by C++17 std::gcd
+template<class I> I gcd(I a, I b) {
+	while (b != 0) {
+		I tmp = b;
+		b = a%b;
+		a = tmp;
+	}
+	return std::abs(a);
+}
 
 void AstNode::set_in_lvalue_flag(bool flag, bool no_descend)
 {
@@ -2818,27 +2831,12 @@ bool AstNode::simplify(bool const_fold, int stage, int width_hint, bool sign_hin
 		if (!children[0]->id2ast->range_valid)
 			goto skip_dynamic_range_lvalue_expansion;
 
-		int source_width = children[0]->id2ast->range_left - children[0]->id2ast->range_right + 1;
-		int source_offset = children[0]->id2ast->range_right;
-		int result_width = 1;
-		int stride = 1;
 		AST::AstNode *member_node = get_struct_member(children[0]);
-		if (member_node) {
-			// Clamp chunk to range of member within struct/union.
-			log_assert(!source_offset && !children[0]->id2ast->range_swapped);
-			source_width = member_node->range_left - member_node->range_right + 1;
-
-			// When the (* nowrshmsk *) attribute is set, a CASE block is generated below
-			// to select the indexed bit slice. When a multirange array is indexed, the
-			// start of each possible slice is separated by the bit stride of the last
-			// index dimension, and we can optimize the CASE block accordingly.
-			// The dimension of the original array expression is saved in the 'integer' field.
-			int dims = children[0]->integer;
-			stride = source_width;
-			for (int dim = 0; dim < dims; dim++) {
-				stride /= get_struct_range_width(member_node, dim);
-			}
-		}
+		int wire_width = member_node ?
+			member_node->range_left - member_node->range_right + 1 :
+			children[0]->id2ast->range_left - children[0]->id2ast->range_right + 1;
+		int wire_offset = children[0]->id2ast->range_right;
+		int result_width = 1;
 
 		AstNode *shift_expr = NULL;
 		AstNode *range = children[0]->children[0];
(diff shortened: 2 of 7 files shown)


phsauter:
This PR seems to reduce the Quality-of-Results for `nowrshmsk`, the attached zip-file can serve as a base to investigate the issue.

[shift_issue3.zip](https://github.com/YosysHQ/yosys/files/12279345/shift_issue3.zip)

yosys 0.32:
- reg_demux: area: 72k; normalized runtime: 1 
- reg_demux_pow2: area: 37k; normalized runtime: 0.25
- reg_demux_nowrshmsk_typedef: area: 33k; normalized runtime: 0.25 
- _(reg_demux_nowrshmsk_mdim_arr: area: 36k; normalized runtime: 17.5)_
_mdim_arr only added for completeness, yosys 0.32 can't handle it properly_

PR #3875:
- reg_demux: area: 75k; normalized runtime: 0.9
- reg_demux_pow2: area: 37k; normalized runtime: 0.13
- reg_demux_nowrshmsk_typedef: area: 51k; normalized runtime: 0.15 
- reg_demux_nowrshmsk_mdim_arr: area: 51k; normalized runtime: 0.15

daglem:
Good find!

The resource increase seems to be caused by the simplification of `non_opt_range` in `verilog_parser.y`, affecting *reads* from `out_rsp_i`. Perhaps some optimization pass relies on the exact pattern `(@selfsz@((index)*(width)))+(0))` for the range width?

I'll see if I can avoid changing `non_opt_range`, instead extracting stride from the original AST.

BTW please don't use unpacked arrays within packed structs; this is not allowed by the LRM. The following (correct) typedef is functionally equivalent:

```
	typedef struct packed {
	    logic [12:0][116:0] port;
	} out_req_t;
```

phsauter:
If you look at the reports in the attached zip, you can see that your solution is actually smaller right up to running `abc`, even if you look at the rtlil and verilog dumps during the synthesis process, they are usually almost identical.

So I see three possibilities:
1. The optimization potential cannot be detected anymore by `abc` (its likely that different `abc` scripts would then find it
2. The `abc` call does not pass the circuit the same way, maybe because of the attribute
3. The optimization potential is actually lost for some reason (I think thats unlikely)

Sadly I am somewhat out of time for the next 2-3 weeks (I will still try to work on it, just less).

[shift_issue_pr3875.zip](https://github.com/YosysHQ/yosys/files/12289653/shift_issue_pr3875.zip)

daglem:
> If you look at the reports in the attached zip, you can see that your solution is actually smaller right up to running `abc`, even if you look at the rtlil and verilog dumps during the synthesis process, they are usually almost identical.

Yeah. I found that the real (current) optimization was always the simplification of the AST_CASE using stride, seemingly the removal of multiplication of index by stride will in certain scenarios only do harm ATM (I guess some optimization pass removes the multiplication in any case).
 
As a pragmatic measure I've disabled the removal of the multiplication for now, and at the same time I generalized the AST_CASE simplification to handle variable slices on the form dst[i*stride +: width] = src

> Sadly I am somewhat out of time for the next 2-3 weeks (I will still try to work on it, just less).

If you like, you can probably use this PR together with #3877 right away for testing of the SoC design. When you continue working on optimization of the shift approach, the results from `nowrshmsk` + `nordshift` can hopefully serve as a comparison baseline.

daglem:
@povik I did a cleanup by squashing the commits.

As a torture test I've also run `make test` after temporarily setting `bool use_case_method = true;` in `simplify.cc`.

This resulted in two failures:
* tests/svtypes/struct_dynamic_range.ys
  - expected, since this test explicitly checks counts of mul, shift, and shiftx cells.
* tests/various/dynamic_part_select/forloop_select.v
  - inexplicably (to me) fails to prove equivalence with forloop_select_gate.v

Running `../../yosys -p "read_verilog -dump_vlog2 ./dynamic_part_select/forloop_select.v"` yields code which AFAICT should yield identical results to `forloop_select_gate.v`, considering the constraints implied by the stride in each `case`. Setting `stride = 1` in `simplify.cc` makes the test pass, but I am at a loss as to what the logical difference could be.

To confuse things further, the failure summary of `sat` in `dynamic_part_select.log` seems to show impossible combinations of `ctrl`, `din`, and `dout` for both `gold` and `gate`?!?

Do you have any idea what can be wrong here, either in the generated optimized CASE, or in the equivalence test?

daglem:
The only hypothesis I've been able to come up with regarding the failure for `tests/various/dynamic_part_select/forloop_select.v` is that `ctrl*sel` in `dout[(ctrl*sel)+:SLICE] <= din;` can overflow, yielding an index which is not covered by the generated CASE.

Given that `ctrl` and `sel` are 4 and 5 bits, respectively, e.g. `3*15 = 45` (6 bits) would assumedly, due to SystemVerilog expression size determination rules, be truncated to 5 bits, yielding `13`, while the generated CASE only covers `0` and `15`.

frontends/ast/simplify.cc
```diff
@@ -2875,12 +2863,76 @@ bool AstNode::simplify(bool const_fold, bool in_lvalue, int stage, int width_hin
...
+			// long long bitno_div = std::gcd(stride, shift_mod);
+			long long bitno_div = boost::integer::gcd((long long)stride, shift_mod);
+
 			did_something = true;
 			newNode = new AstNode(AST_CASE, shift_expr);
-			for (int i = 0; i < source_width; i += stride) {
+			for (int i = 1 - result_width; i < source_width; i++) {
+                                // Out of range indexes are handled in genrtlil.cc
 				int start_bit = source_offset + i;
-				int end_bit = std::min(start_bit+result_width,source_width) - 1;
```
povik, line 2882: I think the earlier code here was confused since `start_bit+result_width` had the `source_offset` included but was compared to `source_width` which had not. So this led to incorrect behavior on wires with non-zero `source_offset`, right?

frontends/ast/simplify.cc
```diff
@@ -2875,12 +2873,76 @@ bool AstNode::simplify(bool const_fold, bool in_lvalue, int stage, int width_hin
...
-				AstNode *cond = new AstNode(AST_COND, mkconst_int(start_bit, true));
+				int end_bit = start_bit + result_width - 1;
+                                // Check whether the current index can be generated by shift_expr.
+				if (start_bit < min_offset || start_bit > max_offset)
+					continue;
+				if (start_bit%bitno_div != 0 || (stride == 0 && start_bit != 0))
+					continue;
+
+				AstNode *cond = new AstNode(AST_COND, mkconst_int(start_bit, case_sign_hint, case_width_hint));
```
povik, line 2945: So here by forcing the sign and width of `shift_expr` on the `start_bit` constant in the case, can we make it overflow? In either case what's the purpose?

frontends/ast/simplify.cc
daglem: > Btw shouldn't we rename all those `source_` variables? It's the target of the assignment, so it's confusing.

I agree! I've been thinking about changing that, however I didn't dare to change the "source_" naming which originates with `source_width`in the shift and mask code.

Should I change all `source_offset` and `source_width` to e.g. `dst_offset` and `dst_width`?

frontends/ast/simplify.cc
daglem: Yes, it can overflow. Not considering this can lead to incorrect behavior, as I found here: https://github.com/YosysHQ/yosys/pull/3875#issuecomment-1813840459

frontends/ast/simplify.cc
povik: Yes, go ahead. I would suggest `wire_offset` and `wire_width`, but really anything will be an improvement it seems.

daglem:
All tests pass now with a forced `use_case_method = true`, except for the expected failure of `tests/svtypes/struct_dynamic_range.ys`

The possible overflow of `i*stride` required a different take on this in order to optimize the generated CASE. Lots of hours for a seemingly small change, as always :see_no_evil: 

In the end I had to add an internal `gcd` function to make the CI run - sorry about all the force pushes.

IMO this is ready to be merged now, with a possible change of `source_` to `dst_` or similar.

frontends/ast/simplify.cc
povik: OK, so we consider the overflow in figuring out the reachable values of `shift_expr`, but shouldn't we then take the overflown value for the bit-select on the target wire?

frontends/ast/simplify.cc
daglem: OK, changed to `wire_offset` and `wire_width` now.

frontends/ast/simplify.cc
povik: I think I see now. So we are not relying on `mkconst_int(start_bit, case_sign_hint, case_width_hint)` wrapping around the constant since at that point it should never be outside the [`min_offset`,`max_offset`] range. If that's the case I suggest we have it as `mkconst_int(start_bit, true)` so it's clearer.

frontends/ast/simplify.cc
```diff
@@ -2867,12 +2865,76 @@ bool AstNode::simplify(bool const_fold, int stage, int width_hint, bool sign_hin
...
+			//   bitno mod gcd(stride, shift_mod) = 0
+			//
+			// long long is at least 64 bits in C++11
+			int max_bits = case_width_hint;
+			long long max_offset = (1ll << (max_bits - case_sign_hint)) - 1;
+			long long min_offset = case_sign_hint ? -(1ll << (max_bits - 1)) : 0;
+			long long shift_mod = 1ll << (case_width_hint - case_sign_hint);
+			// std::gcd requires C++17
+			// long long bitno_div = std::gcd(stride, shift_mod);
+			long long bitno_div = gcd((long long)stride, shift_mod);
```
povik, line 2923: I feel like this undoes much of the optimisation when `bitno_div` can only be power-of-2, not that I don't see why it's done to address the overflown cases.

frontends/ast/simplify.cc
daglem: That would unfortunately make `tests/verilog/dynamic_range_lhs.v` fail again.
The width is required, since the bit pattern for a negative number depends on the width.

E.g. here is the diff between the working and non-working generated CASE in the first failure in the test above, generated with `../../yosys -DALT=1 -DSPAN=1 -DLEFT=-3 -DRIGHT=0 -p "read_verilog -dump_vlog2 ./dynamic_range_lhs.v"`:
```
--- case-ok.txt	2023-11-20 21:24:03.486296307 +0100
+++ case-not-ok.txt	2023-11-20 21:25:05.460245527 +0100
@@ -46,15 +46,15 @@
             1:
               begin
                 case ((sel1)*(sel2))
-                  2'b 00:
+                  0:
                     out_u[0:0] = data;
                 endcase
                 case (/** AST_TO_SIGNED **/)
-                  2'b 10:
+                  -2:
                     out_s[-2:-2] = data;
-                  2'b 11:
+                  -1:
                     out_s[-1:-1] = data;
-                  2'b 00:
+                  0:
                     out_s[0:0] = data;
                 endcase
               end
```

frontends/ast/simplify.cc
daglem: Well, it's not ideal, although power-of-2 will perfectly optimize common (most?) cases (e.g. `gcd(8, 256) = 8`), and partially optimize others (e.g. `gcd(12, 256) = 4`).

Since we know the value of `stride`, it might be possible to identify cases where the index expression cannot possibly overflow, and fully optimize those cases. I'll look into it.

daglem: How about now? :sweat_smile: 

All tests still pass with a forced `use_case_method = true`, and the results from running the tests in `shift_issue3.zip` above are:
```
   Chip area for module '\reg_demux_nowrshmsk_mdim_arr': 19213.891200
   Chip area for module '\reg_demux_nowrshmsk_typedef': 19115.271000
   Chip area for module '\reg_demux_pow2': 5513.961600
   Chip area for module '\reg_demux': 21857.661000
```

As far as I can tell, the PR now only brings improvements, without any regressions.

frontends/ast/ast.h
```diff
@@ -321,6 +321,9 @@ namespace AST
 		static AstNode *mkconst_str(const std::vector<RTLIL::State> &v);
 		static AstNode *mkconst_str(const std::string &str);
 
+		// helper function to create an AST node for a temporary register
+		AstNode *mktemp_logic(const std::string &name, AstNode *mod, bool nosync, int range_left, int range_right, bool is_signed);
```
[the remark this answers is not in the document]
daglem: It can't be static, as long as it's encoding the filename and location in `wire->str`.

Is that a problem? If so, I guess I could make a separate (non-static) function to generate the name, and pass the result of that into `mktemp_logic`, but I'm not sure I see the benefit.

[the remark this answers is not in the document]
povik: Ah sorry, missed that. No, it's not a problem per se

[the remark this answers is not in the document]
daglem: FYI I've also thought about enhancing the function later, adding a parameter to decide whether the name should be local (no name decoration) or global (name decorated as it is now).

I believe the function could then be used to replace most of the current code to instantiate temporaries, for which I could make a PR.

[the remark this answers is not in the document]
daglem: @povik As promised did some more tests with shift_issue3.zip (including reg_demux_noshift in the Makefile). I applied #3877 to reproduce increased chip area when an unconditional temporary was introduced for the rhs.

Without any changes (no temporary in this case):
```   Chip area for module '\reg_demux_noshift': 18689.907600```
Forcing use of the temporary, which is currently assigned by AST_ASSIGN_EQ in the current block, instantiating a `(* nosync  *)` register:
```   Chip area for module '\reg_demux_noshift': 18815.101200```
Forcing use of the temporary using AST_ASSIGN at the module level instead, instantiating a wire, since it could possibly be argued that this should use less resources than a register in this case (it did the contrary):
```   Chip area for module '\reg_demux_noshift': 18896.749200```

As it stands for this test case, I'm hesitant to make any further changes.

Perhaps someone wants to investigate why introducing a temporary makes any difference at all (e.g. should it have been completely optimized away in `simplify.cc` / `genrtlil.cc` in this case, since it's assigned from a simple 117 bit wire?), however this is outside the scope for this PR.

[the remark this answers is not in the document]
povik: Thanks! I experimented with this, and as far as I can tell the difference in area is due to the chaotic nature of the synthesis flow, not because of any systematic effect of introducing the temporary. When that's the case I think we want to make the temporary non-conditional, just so we spare a bit of control flow in the congested `simplify.cc` code.

povik (approved):

povik:
The CI here is stuck because of the bug introduced in #4061 before it was fixed with #4067. @daglem Can you please rebase the branch? Or if you give me the go-ahead, I can push to your branch with a rebased version.

frontends/ast/ast.h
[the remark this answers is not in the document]
daglem: OK, done.

povik:
Thanks! Let's wait for CI to finish then merge.

daglem:
> The CI here is stuck because of the bug introduced in #4061 before it was fixed with #4067. @daglem Can you please rebase the branch? Or if you give me the go-ahead, I can push to your branch with a rebased version.

Done!

PR #3889, opened 2023-08-14
cellaigs: Drop initializer list in call to `IdString::in`

povik:
Remove superfluous curly braces in call to IdString::in to address a compilation error (reproduced below) under GCC 9 and earlier.

kernel/cellaigs.cc:395:18: error: call to member function 'in' is ambiguous if (cell->type.in({ID($gt), ID($ge)}))
~~~~~~~~~~~^~
./kernel/rtlil.h:383:8: note: candidate function
bool in(const std::string &rhs) const { return *this == rhs; } ^
./kernel/rtlil.h:384:8: note: candidate function
bool in(const pool &rhs) const { return rhs.co...
^

Should fix #3885.

Issue #3885, opened 2023-08-11
compilation error at the very first with Centos 7


msaideroglu:
### Version

2829cd9ca

### On which OS did this happen?

Linux

### Reproduction Steps

git clone https://github.com/YosysHQ/yosys.git
make

### Expected Behavior

The tool is builded successfully.

### Actual Behavior

it gives this error with all gcc 7 8 9 versions provided by devtoolset:

[  0%] Building kernel/version_2829cd9ca.cc
[  0%] Building kernel/version_2829cd9ca.o
[  0%] Building kernel/driver.o
[  0%] Building techlibs/common/simlib_help.inc
[  0%] Building techlibs/common/simcells_help.inc
[  0%] Building kernel/register.o
[  1%] Building kernel/rtlil.o
[  1%] Building kernel/log.o
[  1%] Building kernel/calc.o
[  2%] Building kernel/yosys.o
[  2%] Building kernel/binding.o
[  2%] Building kernel/cellaigs.o
kernel/cellaigs.cc:395:18: error: call to member function 'in' is ambiguous
                if (cell->type.in({ID($gt), ID($ge)}))
                    ~~~~~~~~~~~^~
./kernel/rtlil.h:383:8: note: candidate function
                bool in(const std::string &rhs) const { return *this == rhs; }
                     ^
./kernel/rtlil.h:384:8: note: candidate function
                bool in(const pool<IdString> &rhs) const { return rhs.co...
                     ^
./kernel/rtlil.h:381:8: note: candidate function not viable: cannot convert
      initializer list argument to 'const Yosys::RTLIL::IdString'
                bool in(const IdString &rhs) const { return *this == rhs; }
                     ^
./kernel/rtlil.h:382:8: note: candidate function not viable: cannot convert
      initializer list argument to 'const char *'
                bool in(const char *rhs) const { return *this == rhs; }
                     ^
./kernel/rtlil.h:374:8: note: candidate function not viable: requires 0
      arguments, but 1 was provided
                bool in(Args... args) const {
                     ^
kernel/cellaigs.cc:398:40: error: call to member function 'in' is ambiguous
                int carry = mk.bool_node(!cell->type.in({ID($le), ID($ge)}));
                                          ~~~~~~~~~~~^~
./kernel/rtlil.h:383:8: note: candidate function
                bool in(const std::string &rhs) const { return *this == rhs; }
                     ^
./kernel/rtlil.h:384:8: note: candidate function
[...]

PR #3891, opened 2023-08-14
Add GCC 9 to CI builds

povik:
Bug #3889 was a case of build failure under GCC 9 which wasn't replicated by any of the other compilers. Add GCC 9 to the test build matrix.

Test run here: https://github.com/povik/yosys/actions/runs/5857751293/job/15880288410

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -25,6 +25,9 @@ jobs:
           - os: { id: ubuntu-20.04, name: focal }
             compiler: 'gcc-10'
             cpp_std: 'c++11'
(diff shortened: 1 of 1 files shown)


povik:
Huh, the test builds succeeded despite not having #3889 merged in. Something's off.

povik:
I guess there was more to the build environment in #3885 than just GCC 9? Reading the CI workflow script I can't imagine it not actually compiling with GCC 9.

mmicko:
@povik Debian/Ubuntu is updating gcc packages if there are bugfixes, so it could be it is Centos version specific.

PR #3892, opened 2023-08-15
abc: Exposes dont_use flag in ABC

QuantamHD:
ABC's read_lib command has a dont_use
cell list that is configurable by the user.

This PR exposes that option to Yosys.

See
https://github.com/berkeley-abc/abc/blob/5405d4787a86615de76afccf1356676fb5c56c1c/src/map/scl/scl.c#L285 for documentation on this option.


QuantamHD:
@mmicko With the latest import your doing of ABC this would be a good feature to have.

mmicko:
Thanks @QuantamHD for this, it is one of the update reasons

Issue #3893, opened 2023-08-16
Suboptimal default synthesis of `$bmux`, `$shiftx`


povik:
### Version

Yosys 0.23 (git sha1 7ce5011c24b, clang++ 11.1.0 -fPIC -Os)

### On which OS did this happen?

macOS

### Reproduction Steps

Working on #3880 I noticed `$bmux` and `$shiftx` are mapped onto gates suboptimally by the default synthesis script. The mapping does not make use of undef bits.

Consider the following example, where we have instances of bmux and shiftx which we synthesize with `techmap; abc;` (modelling what `synth` does, producing equivalent result here) and separately with `techmap; opt_expr -mux_undef; abc`.

```
read_verilog -icells <<EOF
module shiftx(input [9:0] data, input [5:0] sel, output [1:0] out);
	$shiftx #(
		.A_WIDTH(10),
		.A_SIGNED(0),
		.B_WIDTH(6),
		.B_SIGNED(0),
		.Y_WIDTH(2),
	) shiftx(
		.A(data),
		.B(sel),
		.Y(out)
	);
endmodule

module bmux(input [5:0] data, input [5:0] sel, output [2:0] out);
	$bmux #(
		.S_WIDTH(6),
		.WIDTH(3)
	) bmux(
		.A({{186{1'bx}}, data}),
		.S(sel),
		.Y(out)
	);
endmodule
EOF

copy bmux bmux2
copy shiftx shiftx2

select bmux shiftx
techmap; abc

select bmux2 shiftx2
techmap; opt_expr -mux_undef; abc

select -clear
stat
```

### Expected Behavior

The default synthesis script makes full use of the undef bits for reduced gate count.

### Actual Behavior

There's a gate count gap between what `techmap; abc;` (equivalent to `synth` here) and `techmap; opt_expr -mux_undef; abc;` produce. As far as I can tell ABC has no concept of undef bits, and when Yosys hands over the circuit it replaces all instances of undef bits with zero. With explicit insertion of `opt_expr -mux_undef` the undef bits are exploited to eliminate some of the muxes.

The FPGA scripts I looked into (at least `synth_ice40`) don't seem to suffer from the same thing because they happen to call `opt_expr -mux_undef` ahead of `abc` in the `map_ffs` block.

Output from the script above:

```
=== bmux ===

   Number of wires:                 10
   Number of wire bits:             22
   Number of public wires:           3
   Number of public wire bits:      15
   Number of memories:               0
[...]

PR #3894, opened 2023-08-16
abc: Warn about replacing undef bits

povik:
A suggestion inspired by #3893


nakengelhardt:
I don't think this should be a warning; we only use warnings in Yosys when there's a real problem with the user code. This feels like the kind of "Warning: I was able to optimize your design somewhat" messages that lead people to completely ignore warnings. It would fit better as a regular log message IMO.

povik:
OK, I was expecting some pushback on this. We don't want to turn this into *insert vendor tool here*, after all. Changed it to an ordinary log message.

povik:
Though I feel we could make the decision that proper synthesis scripts should never leave undef bits for ABC to find, at which point the warning would be justified.

PR #3900, opened 2023-08-23
Create unified synth_lattice

mmicko:
[x] ecp5_gsr -> lattice_gsr
[x] opt_lut_ins - use lattice instead of ecp5 (old param still supported)
[x] move synth_machxo2 to synth_lattice
[x] update machxo2 tests
[x] create wrappers for machxo3, machxo3d, and ecp5 (commented for now)
[x] create script to extract blackboxes from Diamond files
[ ] abc9 support for xo2/xo3/xo3d
[x] enable more primitives/blackboxes supported by nextpnr


Ravenslofty (approved):

techlibs/lattice/synth_lattice.cc
```diff
@@ -0,0 +1,574 @@
...
+		RTLIL::constpad["synth_lattice.abc9.W"] = "300";
+	}
+
+	void help() override
+	{
+		//   |---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|
+		log("\n");
+		log("    synth_lattice [options]\n");
+		log("\n");
+		log("This command runs synthesis for Lattice FPGAs.\n");
```
gatecat, line 44: can you add a clarification like "(excluding iCE40)"

what's the plan for nexus support here?

techlibs/lattice/synth_lattice.cc
```diff
@@ -0,0 +1,574 @@
...
+		args[0] = "synth_lattice";
+		args.insert(args.begin()+1, std::string());
+		args.insert(args.begin()+1, std::string());
+		args[1] = "-family";
+		args[2] = "xo3";
+		Pass::call(design, args);
+	}
+} SynthMachXO3Pass;
+
+struct SynthMachXO3DPass : public Pass
```
gatecat, line 559: Do we need this `synth_machxo3d` shim? There was never a `synth_machxo3d` pass to be backwards compatible with in the first place...

techlibs/lattice/synth_lattice.cc
```diff
@@ -0,0 +1,574 @@
...
+		log("    -asyncprld\n");
+		log("        use async PRLD mode to implement ALDFF (EXPERIMENTAL)\n");
+		log("\n");
+		log("    -abc2\n");
+		log("        run two passes of 'abc' for slightly improved logic density\n");
+		log("\n");
+		log("    -abc9\n");
+		log("        use new ABC9 flow (EXPERIMENTAL)\n");
+		log("\n");
+		log("    -vpr\n");
```
gatecat, line 119: can we remove this? there's never been any meaningful VPR support for lattice fpgas, and I think VPR has moved on enough that whatever we output would probably be no use anyway

techlibs/lattice/synth_lattice.cc
```diff
@@ -0,0 +1,574 @@
...
+		if (check_label("check"))
+		{
+			run("autoname");
+			run("hierarchy -check");
+			run("stat");
+			run("check -noinit");
+			run("blackbox =A:whitebox");
+		}
+
+		if (check_label("blif"))
```
gatecat, line 481: we could actually remove `blif` output altogether

techlibs/lattice/synth_lattice.cc
mmicko: ice40 is for sure out of scope here (so I will add message), nexus is possible to be added but I did not investigated it too much, but it looks it should not be a problem. Would rather create new PR of adding that later

techlibs/lattice/synth_lattice.cc
mmicko: synth_machxo3 and synth_machxo3d are more here for user convenience then for backward compatibility

techlibs/lattice/synth_lattice.cc
mmicko: sure vpr and blif can be removed it was added to be fully compatible with synth_ecp5

techlibs/lattice/synth_lattice.cc
gatecat: I don't think they're needed, we don't have `synth_xc7` and `synth_ultrascale` for example

techlibs/lattice/synth_lattice.cc
gatecat: sounds good - this is one of the places where I'd rather see an easy opportunity for cleanup taken rather than 100% backwards compatibility

techlibs/lattice/synth_lattice.cc
mmicko: in that case I would remove synth_machxo2 as well since parameters are anyway not fully compatible and it would require user to change old scripts anyway

gatecat: yes, given the relatively low number of users of the current machxo2 flow that sounds reasonable

gatecat (approved):

PR #3915, opened 2023-09-04
sim: Add print support

povik:
Ping @jix since they expressed interest in this support.

Ping @kivikakk, @mwkmwkmwk should they want to double check the triggering semantics.


whitequark (approved):

povik:
Changed this from being a draft since it could be ready for all I know.

PR #3918, opened 2023-09-05
ast: Substitute rvalues when parsing out print arguments

### LINKED ISSUE #3917: `$print` cells converted from `$display` tasks don't sample signals according to Verilog semantics
### Version

Yosys 0.32+76 (git sha1 73cb4977b2e, clang++ 11.1.0 -fPIC -Os) with patch for print support in sim command

macOS

### Reproduction Steps

Take the following script

```
read_verilog <<EOF
module top(clk);
input wire clk;

reg [7:0] counter = 0;
always @(posedge clk) begin
    counter = counter + 1;
end

reg [7:0] save;
always @(posedge clk) begin
    save = counter;
    $display("counter %x save %x", counter, save);
end
endmodule
EOF
prep
sim -q -clock clk
```

### Expected Behavior

`counter` and `save` have the same value as far as the print is concerned, i.e. the following gets printed

```
counter 00 save 00
counter 01 save 01
counter 02 save 02
counter 03 save 03
counter 04 save 04
counter 05 save 05
...
```

This is what happens with iverilog.

### Actual Behavior

Log output

```
counter 01 save 00
counter 02 save 01
counter 03 save 02
counter 04 save 03
counter 05 save 04
counter 06 save 05
...
```

Design after `prep`

output

povik:
Apply the local substitutions stemming from process context when parsing out format arguments to `$display` or other statements.

This is to fix #3917.

### DIFF
--- a/frontends/ast/genrtlil.cc
+++ b/frontends/ast/genrtlil.cc
@@ -759,7 +759,7 @@ struct AST_INTERNAL::ProcessGenerator
 						arg.realtime = true;
 					} else {
 						arg.type = VerilogFmtArg::INTEGER;
-						arg.sig = node->genRTLIL();
+						arg.sig = node->genWidthRTLIL(-1, false, &subst_rvalue_map.stdmap());
 						arg.signed_ = is_signed;
 					}
 					args.push_back(arg);


frontends/ast/genrtlil.cc
```diff
@@ -759,7 +759,7 @@ struct AST_INTERNAL::ProcessGenerator
 						arg.realtime = true;
 					} else {
 						arg.type = VerilogFmtArg::INTEGER;
-						arg.sig = node->genRTLIL();
+						arg.sig = node->genWidthRTLIL(-1, false, &subst_rvalue_map.stdmap());
```
povik, line 762: This is all that seems to need change. It introduces an extra call to `detectSignWidthWorker` from within `genWidthRTLIL` which hopefully is benign.

whitequark (approved):

PR #3920, opened 2023-09-06
sv: support assignments within expressions

### LINKED ISSUE #3870: SystemVerilog procedural assignments within expressions
### Version

“unexpected TOK_INCREMENT” when SystemVerilog source contains ++

Linux

### Reproduction Steps

yosys -read -sv [hamming/hamming_dec.sv](https://github.com/dshekhalev/FEC/blob/main/rtl/hamming/hamming_parameters.svh)

### Expected Behavior

……

### Actual Behavior

-- Parsing `hamming/hamming_dec.sv' using frontend ` -sv' --

1. Executing Verilog-2005 frontend: hamming/hamming_dec.sv
Parsing SystemVerilog input from `hamming/hamming_dec.sv' to AST representation.
hamming/hamming_parameters.svh:57: ERROR: syntax error, unexpected TOK_INCREMENT

zachjs:
- Add support for assignments within expressions, e.g., `x[y++] = z;` or `x = (y *= 2) - 1;`. The logic is handled entirely within the parser by injecting statements into the current procedural block.
- Add support for pre-increment/decrement statements, which are behaviorally equivalent to post-increment/decrement statements.
- Fix non-standard attribute position used for post-increment/decrement statements.

I am not particularly attached to this implementation. I had initially anticipated adding a new AST node type and elaborating them in `simplify.cc`, but sticking the statements into the current procedure ended up working quite naturally.

Closes #3870.

### DIFF
--- a/frontends/verilog/verilog_parser.y
+++ b/frontends/verilog/verilog_parser.y
@@ -325,7 +325,7 @@ static AstNode *addIncOrDecExpr(AstNode *lhs, dict<IdString, AstNode*> *attr, AS
 	log_assert(stmt->type == AST_ASSIGN_EQ);
 	AstNode *expr = stmt->children[0]->clone();
 	if (undo) {
-		AstNode *minus_one = AstNode::mkconst_int(-1, true);
+		AstNode *minus_one = AstNode::mkconst_int(-1, true, 1);
 		expr = new AstNode(op, expr, minus_one);
 	}
 	SET_AST_NODE_LOC(expr, begin, end);
--- a/frontends/verilog/verilog_parser.y
+++ b/frontends/verilog/verilog_parser.y
@@ -292,6 +292,61 @@ static void rewriteGenForDeclInit(AstNode *loop)
 	substitute(incr);
 }
 
+static void ensureAsgnExprAllowed()
+{
+	if (!sv_mode)
+		frontend_verilog_yyerror("Assignments within expressions are only supported in SystemVerilog mode.");
+	if (ast_stack.back()->type != AST_BLOCK)
+		frontend_verilog_yyerror("Assignments within expressions are only permitted within procedures.");
+}
+
+// add a pre/post-increment/decrement statement
+static const AstNode *addIncOrDecStmt(AstNode *lhs, dict<IdString, AstNode*> *attr, AST::AstNodeType op, YYLTYPE begin, YYLTYPE end)
+{
+	AstNode *one = AstNode::mkconst_int(1, true);
+	AstNode *rhs = new AstNode(op, lhs->clone(), one);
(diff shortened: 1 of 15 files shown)


frontends/verilog/verilog_parser.y
```diff
@@ -2610,17 +2665,15 @@ simple_behavioral_stmt:
...
-	attr lvalue TOK_INCREMENT {
-		AstNode *node = new AstNode(AST_ASSIGN_EQ, $2, new AstNode(AST_ADD, $2->clone(), AstNode::mkconst_int(1, true)));
-		ast_stack.back()->children.push_back(node);
-		SET_AST_NODE_LOC(node, @2, @3);
-		append_attr(node, $1);
+	attr lvalue attr inc_or_dec_op {
+		// The position 1 attr to avoid shift/reduce conflicts with the
+		// other productions. We reject attributes in that position.
+		if (!$1->empty())
+			frontend_verilog_yyerror("Attributes are not allowed on this size of the lvalue in an inc_or_dec_expression!");
```
povik, line 2672: As far as I can tell these attributes would be allowed by the standard, since attributes can be prefix to a statement and this is a self-standing statement, no?

frontends/verilog/verilog_parser.y
```diff
@@ -292,6 +292,61 @@ static void rewriteGenForDeclInit(AstNode *loop)
...
+	SET_AST_NODE_LOC(stmt, begin, end);
+	if (attr != nullptr)
+		append_attr(stmt, attr);
+	ast_stack.back()->children.push_back(stmt);
+	return stmt;
+}
+
+// create a pre/post-increment/decrement expression, and add the corresponding statement
+static AstNode *addIncOrDecExpr(AstNode *lhs, dict<IdString, AstNode*> *attr, AST::AstNodeType op, YYLTYPE begin, YYLTYPE end, bool undo)
+{
```
povik, line 318: I think we want to recover the sizing information here, or do something other than the plus one minus one dance (maybe an assignment to a dummy wire?). Consider the following:
```
  w = {1, r++};
```

povik:
Other than the two comments I went through all of it and it looks good.

frontends/verilog/verilog_parser.y
zachjs: Good point! I'll pushing a fix for this shortly.

frontends/verilog/verilog_parser.y
zachjs: This is a great catch! I should have set the width of `minus_one` explicitly to prevent it from affecting the width of the expression. This logic seems simpler than the alternatives I considered: A) adding a temporary wire; or B) actually doing the increment afterwards. That said, my preference here likely comes from having done it the same way in sv2v.

PR #3924, opened 2023-09-06
multpass -- create Booth Encoded multipliers for

andyfox-rushc:
multpass generates booth encoded multipliers for $mul cell. Example of use:

read_verilog umultiply5_rtl.v
opt
wreduce
opt
multpass
alumacc
maccmap
opt
techmap -map adder_map_file.v
techmap -map ./techmap.v
dfflibmap -liberty NangateOpenCellLibrary_typical.lib
abc -liberty NangateOpenCellLibrary_typical.lib
stat -liberty NangateOpenCellLibrary_typical.lib
write_verilog -norename booth_final.v


andyfox-rushc:
If this pass is acceptable then if possible I would like to make it an option in the generic "synth" command (eg synth -mult) so it can be invoked in generic flows used by some folks. I experimented with invoking the pass from techmap.v but problem is that is a bit late if alumacc gets called. But I think that should be subject of a different PR. This one is just a standalone multiplier inferencing/optimization pass (multpass).

povik:
Just a superficial thought: Putting `pass` into the name seems redundant. How about calling it `booth`, `boothmul` or `boothmult`?

povik:
I think this could be thoroughly tested with `test_cell -script map_booth.ys $mul` where `map_booth.ys` is a script with a single command calling your pass. If you want I can push a commit onto your branch (seemingly GitHub would allow me to do that) adding this to the Yosys test suite, so that it would run as part of the tests on the PR.

andyfox-rushc:
Yes please do push your commit and i will dig later today. Thanks for the suggestions ..this is my first go at a pass.
 
one thing I observed is that the alumacc pass maps $mul so to get the multpass to be invoked it must be called before alumacc.
 
another thing is that I have Verilog files for the booth multipliers (ie parameterized verilog with generate statements) which I would also like to somehow get in the yosys examples.
 
thanks again for the suggestions

povik:
OK, I added the test. It has found some failing edge cases.

There is misevaluation in the signed case, illustrated with the script below:

```
read_rtlil <<EOF
module \gold
  wire width 5 input 1 \A
  wire width 7 input 2 \B
  wire width 8 output 3 \Y
  cell $mul \UUT
    parameter \A_SIGNED 1
    parameter \A_WIDTH 5
    parameter \B_SIGNED 1
    parameter \B_WIDTH 7
    parameter \Y_WIDTH 8
    connect \A \A
    connect \B \B
    connect \Y \Y
  end
end
EOF

eval -set A 00001 -set B 1110000 w:Y # \Y = 8'11110000.
multpass
eval -set A 00001 -set B 1110000 w:Y # \Y = 8'01110000.
```

Another is undriven wires in the implementation of e.g. an unsigned 7-bit times 7-bit multiplier, reproduced with the script below:

```
read_rtlil <<EOF
module \gold
  wire width 7 input 1 \A
  wire width 7 input 2 \B
  wire width 8 output 3 \Y
  cell $mul \UUT
    parameter \A_SIGNED 0
    parameter \A_WIDTH 7
    parameter \B_SIGNED 0
    parameter \B_WIDTH 7
    parameter \Y_WIDTH 8
    connect \A \A
    connect \B \B
    connect \Y \Y
  end
end
EOF

multpass
check
```

povik:
You can invoke the test on its own with something like `cd tests/techmap; ../../yosys booth.ys`.

> one thing I observed is that the alumacc pass maps $mul so to get the multpass to be invoked it must be called before alumacc.
> another thing is that I have Verilog files for the booth multipliers (ie parameterized verilog with generate statements) which I would also like to somehow get in the yosys examples.

Couldn't you supply your Verilog files to `techmap` to be the implementation of `$macc`, so that there's no need for an extra pass and adjusting the `synth` scripts?

povik:
Btw, you seem to be generating single-bit `$pos` cells, which really don't do anything and are in Yosys just to help the Verilog frontend express what the input source code was more closely. What's the reasoning for them in your pass?

andyfox-rushc:
Just a quick update. Thank you I isolated and reproduced the issue (you are right weird sized input/output combos edge cases are causing problems). That is a pretty cool verification script/approach. Thank you. Before doing anything else I will attempt fixes here.

On the verilog versus C++ module generator. I wanted to try doing a few things I could not easily express in verilog (the generate statement and gen vars are pretty restrictive) and was very impressed with what I saw in Yosys for building up components and was intrigued by the non EDA network approach (not cells/instances/nets/pins/ports but more compiler like). On the $pos I used it for wiring stuff together, much like we do in hdl processing with buffers (which get optimized out). 

More when I have something. Thanks again !

povik:
Happy to be of help!

You can use `module->connect()` to connect wires directly, so that for example instead of

```
connect_sigSpecToWire(op_wire, SigSpec(Z, cpa_ix, 1));
```

you can simply do

```
module->connect(op_wire, SigSpec(Z, cpa_ix, 1));
```

By the way, you (and anyone else interested) are invited to join the Yosys team on Mondays 14:00 UTC for a videoconference where PRs are discussed. Link to join: https://meet.jit.si/yosyshq-slack-devel-discuss

andyfox-rushc:
Small thing -- fixed the edge cases and the test_cell passes. Now will go back over your comments and do a little more clean up (renaming pass, removal of connect_sigSpecToWire etc).

andyfox-rushc:
observed a build failure in msbuild flow. this related to incorrect use of variable array subscripts (ok for 1 dimensional arrays but technically illegal for 2 or more). So  converted by 2-d arrays to 1-d arrays. Am curious about how to force a rebuild on the msbuild work flow from within this gui. It seems very powerful to me to be able to check across multiple compilers. (I don't have access to msbuild locally so it was great to see the check from the auto build on my pull request).

povik:
The test builds are conditioned on maintainers approval each time the code changes when the PR author is a first-time contributor, or some similar criteria. Anyway it looks like the MS build goes through now!

povik:
No, spoke too soon. Only one of the MS jobs succeeds.

andyfox-rushc:
No worries, now I can see how to reproduce, will fix here. Unfortunately will doubtless see a few more message on this screen. Thanks for the helpful feedback.

andyfox-rushc:
Finally got a candidate pull request (I think) - will attempt to sit on the Monday PR meeting and see if this is something we can get in. Thanks again, Martin, for the constructive help.

techlibs/common/synth.cc
```diff
@@ -17,17 +17,16 @@
  *
  */
 
-#include "kernel/register.h"
```
povik, line 20: There are unrelated whitespace changes but as we discussed in this week's Yosys dev call we can live with it and take it up including those. I went over this file and the only non-formatting change is the addition of the `-booth` flag

povik (approved):
There's a lot that can be simplified and cleaned up if someone more familiar with Yosys APIs does a pass over the code (I volunteer) but I am inclined to leave that for a later PR and merge this as is.

Issue #3923, opened 2023-09-06
commit 9f8e039a4 causes tests/arch/nexus to segfault on Fedora builds


gsomlo:
### Version

0.32+2 and later

### On which OS did this happen?

Linux

### Reproduction Steps

On an up-to-date f38 machine, as a regular user, replicate the `mock` ("official RPM") build environment:

```
export CXXFLAGS='-Wp,-D_GLIBCXX_ASSERTIONS'

git clone https://github.com/yosyshq/yosys
cd yosys; git checkout 83b1a57eed27604fa1a45193ee6894a5e0764971

make mrproper; make config-gcc; make -j10 ABCEXTERNAL=/usr/bin/abc all test
```

***NOTE:*** `ABCEXTERNAL` is in fact Fedora's packaged version of `yosyshq/abc` v.0.31 (still current as of yosys v.0.33)

### Expected Behavior

yosy builds, tests complete without error

### Actual Behavior

```
...
make[1]: Leaving directory '/home/somlo/yosys/tests/arch/intel_alm'
cd tests/arch/nexus && bash run-test.sh "-S 314159265359"
make[1]: Entering directory '/home/somlo/yosys/tests/arch/nexus'
/usr/include/c++/13/bits/basic_string.h:1305: std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::reference std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::front() [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>; reference = char&]: Assertion '!empty()' failed.
/usr/include/c++/13/bits/basic_string.h:1305: std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::reference std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::front() [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>; reference = char&]: Assertion '!empty()' failed.
/usr/include/c++/13/bits/basic_string.h:1305: std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::reference std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::front() [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>; reference = char&]: Assertion '!empty()' failed.
[...]

PR #3925, opened 2023-09-09
ci: Enable extra libstdc++ assertions

povik:


povik:
With this, #3923 seems to be reproduced. Thanks to @gsomlo for figuring out the key library flag.

PR #3926, opened 2023-09-11
Update ABC to latest

mmicko:


mmicko:
@jix If you could please check if this affects anything sby/eqy related

whitequark:
@mmicko  This breaks the Wasm build: https://github.com/YoWASP/yosys/actions/runs/6166481665/job/16736046078

mmicko:
@whitequark ah, sorry for that, not sure I will have time for a fix soon, but if you can provide patch it would be great

whitequark:
@mmicko Can we revert until this is fixed?

whitequark:
Actually that looks like an easy fix, patch incoming

PR #3928, opened 2023-09-12
mem: Fix index confusion in write port merging

povik:
Fix mistaking the read-port and write-port indices for each other when we are adding the partial transparency emulation to be able to merge two write ports.


mwkmwkmwk (approved):
good find, thanks

PR #3933, opened 2023-09-15
Alphabetize kernel headers to be installed, include some missing required ones for plugins

timkpaine:
fixes https://github.com/chipsalliance/synlig/pull/1972 https://github.com/dau-dev/tools/issues/6

Screenshot 2023-09-15 at 15 05 29


povik:
It's worthwhile to point out the headers being added, these seem to be:
```
6a7
> $(eval $(call add_include_file,kernel/cost.h))
8a10,11
> $(eval $(call add_include_file,kernel/ffmerge.h))
> $(eval $(call add_include_file,kernel/fmt.h))
20a24
> $(eval $(call add_include_file,kernel/timinginfo.h))
```
Ideally the change to sort the headers and to add new ones would be split. FWIW with the change the order is not fully alphabetical either (e.g. `kernel/modtools.h` above `kernel/mem.h`).

timkpaine:
I'm happy to alphabetize the remaining, let me know! Given that these headers were omitted by mistake, and should have been added here when the files themselves were added, I don't think there's a ton of value in separate commits (e.g. the git blame is already separated from the original commit)

povik:
Thanks! I think we will merge this as-is (dev call where PRs are discussed is in two hours). I see the value in separate commits in easier review and making it obvious what the user-facing changes are (which comes down to the diff I extracted above).

PR #3938, opened 2023-09-18
`booth` cleanup

povik:
Following up to #3924, this PR will to some degree clean up the code of the Booth pass.

### DIFF
--- a/passes/techmap/booth.cc
+++ b/passes/techmap/booth.cc
@@ -66,33 +66,6 @@ struct BoothPassWorker {
 
 	BoothPassWorker(RTLIL::Module *module) : module(module), sigmap(module) { booth_counter = 0; }
 
-	// Helper routines for building architecture subcomponents
-
-	RTLIL::Wire *mk_wireFromSigSpec(const SigSpec &v)
-	{
-
-		auto g = module->addCell(NEW_ID, ID($pos));
-		Wire *ret = module->addWire(NEW_ID, 1);
-		g->setPort(ID::A, v);
-		g->setPort(ID::Y, ret);
-		g->setParam(ID::A_WIDTH, 1);
-		g->setParam(ID::Y_WIDTH, 1);
-		g->setParam(ID::A_SIGNED, false);
-		return ret;
-	}
-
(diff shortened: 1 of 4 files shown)


povik:
@andyfox-rushc Take a look, feel free to ask and/or object.

passes/techmap/booth.cc
```diff
@@ -797,32 +764,23 @@ struct BoothPassWorker {
...
-
 		static int cpa_id;
 		cpa_id++;
 
-		RTLIL::Wire *carry = nullptr;
-
-		log_assert(s_vec.size() == c_vec.size());
+		log_assert(c_vec.size() == s_vec.size());
+		// TODO: doesn't pass
+		//log_assert(result.size() == s_vec.size() + 2);
```
povik, line 659: @andyfox-rushc Here I moved the `result.size() == s_vec.size() + 2` condition down from the "axioms" comment to an actual assert, but it wasn't passing. Should this be upheld?

andyfox-rushc: Hi, Martin,
My comment is wrong and the axiom you refer to does not need to be upheld. The axiom for the unsigned case should be result.size() ==s_vec.size(). 
In this unsigned case an array of CSA (Wallace) trees is used: 1 per result bit.  Each CSA tree produces a sum and carry. These are summed up by the CPA. The final carry out is ignored in the unsigned case. My error comes from the use of the same structure for signed multiplication, in which case the multiplier is not padded out and one partial product removed which leads to the different axiom for the size. But again my comment was wrong for the unsigned case.

povik: Thanks for the analysis! I wanted to make sure there's not an issue lurking there.

andyfox-rushc: PS Martin, sorry for being quiet. I did quickly go through all your changes which look good to me. I like the addition of the full adder helper creator to RTLIL.  The simplifications to the cell selection and the extension of the arguments to fit the right size (I was adding weird buffers), all seem to make the code much clearer to me. I see these all got safely merged in today. All looks great to me. Thanks.

PR #3939, opened 2023-09-18
rtlil: Speeds up Yosys by 17%

QuantamHD:
This PR speeds up by roughly 17% across a wide spectrum of designs tested at Google. Particularly for the mux generation pass.

### DIFF
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -4031,16 +4031,20 @@ void RTLIL::SigSpec::replace(const RTLIL::SigSpec &pattern, const RTLIL::SigSpec
 	unpack();
 	other->unpack();
 
+	dict<RTLIL::SigBit, int> pattern_to_with;
 	for (int i = 0; i < GetSize(pattern.bits_); i++) {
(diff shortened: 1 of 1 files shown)


QuantamHD:
# Before

image

# After

image

As you can see from the above the proc_mux pass is pretty much gone from the profile.

kernel/rtlil.cc
```diff
@@ -4031,16 +4031,20 @@ void RTLIL::SigSpec::replace(const RTLIL::SigSpec &pattern, const RTLIL::SigSpec
 	unpack();
 	other->unpack();
 
+	dict<RTLIL::SigBit, int> pattern_to_with;
```
povik, line 4034: I guess if a dict is what it takes, then a dict is what it takes.

povik (approved):

kernel/rtlil.cc
QuantamHD: It surprised a few of my coworkers as well. That's why real measurement is so important to optimization

QuantamHD:
@povik Is this ready to merge?

PR #3940, opened 2023-09-19
rtlil: Speeds up string decoding by 30%

QuantamHD:
This change represents about a 2% speed up of Yosys as a whole.

Written by @rmlarsen

### DIFF
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -27,6 +27,7 @@
 
 #include <string.h>
 #include <algorithm>
+#include <array>
 
 YOSYS_NAMESPACE_BEGIN
 
@@ -200,17 +201,30 @@ const pool<IdString> &RTLIL::builtin_ff_cell_types() {
 	return res;
 }
 
-RTLIL::Const::Const(const std::string &str)
-{
-	flags = RTLIL::CONST_FLAG_STRING;
-	bits.reserve(str.size() * 8);
-	for (int i = str.size()-1; i >= 0; i--) {
-		unsigned char ch = str[i];
-		for (int j = 0; j < 8; j++) {
-			bits.push_back((ch & 1) != 0 ? State::S1 : State::S0);
-			ch = ch >> 1;
-		}
-	}
+PRIVATE_NAMESPACE_BEGIN
+
+using CharToBitsLUT = std::array<std::array<RTLIL::State, 8>, 256>;
+
+const CharToBitsLUT build_CharToBitsLUT() {
+        CharToBitsLUT LUT;
+        for (int i = 0; i < 256; ++i) {
+          unsigned char ch = i;
+          for (int j = 0; j < 8; j++) {
+            LUT[i][j] = (ch & 1) != 0 ? State::S1 : State::S0;
+            ch = ch >> 1;
+          }
+        }
+        return LUT;
+}
+PRIVATE_NAMESPACE_END
+
+RTLIL::Const::Const(const std::string &str) : flags(RTLIL::CONST_FLAG_STRING) {
+        static const CharToBitsLUT LUT = build_CharToBitsLUT();
+        bits.reserve(str.size() * 8);
+        for (int i = str.size() - 1; i >= 0; --i) {
+          const unsigned char ch = str[i];
+          bits.insert(bits.end(), LUT[ch].begin(), LUT[ch].end());
+        }
 }
 
 RTLIL::Const::Const(int val, int width)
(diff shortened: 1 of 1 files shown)


rmlarsen:
Did you intend to include the change to decode_string here? Or is that going to be a separate PR?

QuantamHD:
Good catch. I added it back.

kernel/rtlil.cc
```diff
@@ -200,17 +201,30 @@ const pool<IdString> &RTLIL::builtin_ff_cell_types() {
...
+        return LUT;
+}
+PRIVATE_NAMESPACE_END
+
+RTLIL::Const::Const(const std::string &str) : flags(RTLIL::CONST_FLAG_STRING) {
+        static const CharToBitsLUT LUT = build_CharToBitsLUT();
+        bits.reserve(str.size() * 8);
+        for (int i = str.size() - 1; i >= 0; --i) {
+          const unsigned char ch = str[i];
+          bits.insert(bits.end(), LUT[ch].begin(), LUT[ch].end());
```
povik, line 226: I am bit surprised by this being faster -- would you perhaps have a figure for this change alone?

kernel/rtlil.cc
```diff
@@ -311,20 +325,30 @@ RTLIL::Const RTLIL::Const::from_string(const std::string &str)
...
+            if (bits[j] == RTLIL::State::S1) {
+              ch |= 1 << (j - start);
+            }
+          }
+          if (ch != 0) s.append({ch});
+        };
+
+        int i = n_over_8 * 8;
+        if (i < n) {
+          fill_byte(i, n);
```
povik, line 346: I must say I really dislike the closure here. What about something like:
```
    int cap = GetSize(bits); // where the current byte ends
    for (int i = GetSize(bits) / 8 * 8; i >= 0; i -= 8) {
        for (int j = i; j < cap; j++) {
            ...
        }
        cap = j;
    }

kernel/rtlil.cc
rmlarsen: This in itself is about 20% faster.  Decode_string is about 30% faster.

kernel/rtlil.cc
rmlarsen: Having a helper function for filling each byte is more readable to me, but I suppose it's a matter of what style you are used to read. This is a pretty common idiom in Google code. But we can certainly change it.

kernel/rtlil.cc
rmlarsen: Before:
image

After:
image

kernel/rtlil.cc
povik: Please do change it. While at it, please fix the patch to use tabs like in the rest of the file.

povik: Also to consider is splitting the patches: we are ready to merge the `decode_string()` change but we are not so sure about the lookup table in the constructor.

povik:
We are curious about the timings. Couple of questions: when you say this speeds up Yosys as a whole by 3 %, do you mean some representative ASIC flow you are running on your end? Would you have some typical timings for the `read_verilog` / `hierarchy` steps on their own, ideally absolute? We assume those steps are where the significant speedup is, e.g. in constructing the `Const` values for all the `src` attributes, et cetera.

kernel/rtlil.cc
rmlarsen: @QuantamHD do you want to update / split the PR or should I just file a new one?

rmlarsen:
Hi Martin,

Thanks for the comments. Response below.

> We are curious about the timings. Couple of questions: when you say this speeds up Yosys as a whole by 2 %, do you mean some representative ASIC flow you are running on your end? Would you have some typical timings for the `read_verilog` / `hierarchy` steps on their own, ideally absolute? We assume those steps are where the significant speedup is, e.g. in constructing the `Const` values for all the `src` attributes, et cetera.

Yes, the 2% is the reduction in overall time of the synthesis flow for a representative circuit block from a Google TPU that I'm working on. I can provide you with a flame graph for example, since that probably gives more context, or a wider slice of the Tree view.

povik:
Hello Rasmus,

Yes, anything you could provide to inform our discussion would be helpful.

rmlarsen:
@povik Will do. Notice that @QuantamHD is OOO until tomorrow. I'd like to coordinate with him on how to proceed in terms of splitting, so give us a few days. Thanks.

rmlarsen:
Ah, good to know. Thanks! Would be nice to join the call if I can at some point.

povik:
Link here: https://meet.jit.si/yosyshq-slack-devel-discuss

It's scheduled to be an hour long, occurs every week.

rmlarsen:
@povik I moved the `decode_string()`  change to https://github.com/YosysHQ/yosys/pull/3959. I got rid of the lambda and after a few tries found a variant that was a bit faster too.

QuantamHD:
@rmlarsen Should we close this then?

rmlarsen:
@QuantamHD sure. We can upstream the `Const` constructor separately, if the Yosys authors find it acceptable.

PR #3943, opened 2023-09-19
Set src attribute for verific with full info

mmicko:
This makes src attribute completely same as in Yosys verilog frontend and gives full information about source position.

Location for Yosys verilog frontent is created here :
https://github.com/YosysHQ/yosys/blob/master/frontends/ast/ast.cc#L985


nakengelhardt:
Thanks! `src` attribute format has caused issues with mcy in the past, can you check that mcy (and mcy-gui) still runs fine with this? I expect it will since it's the same format as `read_verilog`, but best to make sure.

nakengelhardt (approved):

PR #3945, opened 2023-09-19
Speed up the autoname pass by 3x. 

rmlarsen:
This is accomplished by only constucting IdString objects from plain strings that have a higher score.

Before:
image

After:
image


rmlarsen:
Hang on. I think this will be even faster if we keep them as plain strings until after we pull them out of the proposed_*_names map. Let me up this.

rmlarsen:
> Hang on. I think this will be even faster if we keep them as plain strings until after we pull them out of the proposed_*_names maps. Let me update this.

Yup, that makes this a 3x instead of a 2x speedup.

rmlarsen:
OK, this is ready for review.

whitequark (approved):

PR #3946, opened 2023-09-20
Speed up TopoSort by 2.7-3.3x.

rmlarsen:
The main implementation in `TopoSort::sort_worker` is 11-12x faster. Overall, the complete sequence of building the graph and sorting is about 2.7-3.3x faster for the circuit I am working on. The overall impact in e.g. the replace_const_cells optimization pass is a ~35% speedup. End-to-end impact on our synthesis flow is about 3.3%.

Flame graph, zoomed on replace_const_cells, before:
image

Flame graph, zoomed on replace_const_cells, after:
image

Topdown view before:
image

Topdown view after:
image


rmlarsen:
What are your thoughts on this PR? I could back out the unrelated formatting changes in utils.h, if so desired.

Ravenslofty (approved):
seems good to me.

rmlarsen:
OK, I got rid of the needless formatting changes in `kernel/utils.h.`  PTAL.

Ravenslofty (approved):

rmlarsen:
It looks like one of the xilinx tests fail. Investingating.

rmlarsen:
Fixed. BTW: I realized that this code is brittle, since the sort order in `map` and `set` for some uses of TopoSort depend on the sort order of pointers. This is generally a bad code smell, since it tends to make code non-deterministic.

rmlarsen:
OK, It finally passed.  :-)  PTAL.

rmlarsen:
Trimmed a little more fat off my implementation of `node()`. Speeds everything up by another 10%.

Ravenslofty:
In case you're wondering what's blocking this and #3975, I'd like somebody else to look over and approve it before I merge it, because I'm unfamiliar with the RTLIL internals; perhaps @mmicko?

rmlarsen:
> In case you're wondering what's blocking this and #3975, I'd like somebody else to look over and approve it before I merge it, because I'm unfamiliar with the RTLIL internals; perhaps @mmicko?

Yeah, no worries. It's always good to have more eyes on it.

povik (approved):
The code got less readable, unfortunately, but I suppose that's the price we ought to pay. Thanks @rmlarsen!

povik:
We could opt for a non-recursive algorithm ([a code example I have at hand](https://github.com/povik/toymap/blob/38cdfec69e7298f874dc2a559a6ea1b73ff188c7/toymap.cc#L813-L839)), but that might not be producing the sort order code users expect, especially if they supply the `C` template argument.

PR #3955, opened 2023-09-25
show: use dot for wire aliases instead of BUF

nakengelhardt:
The "BUF" display is confusing as it is easy to misinterpret as a `$_BUF_` cell.


povik (approved):
The result even looks better to me.

povik:
Probably best to mention this in the changelog since people can be used to reading the show graphs a certain way.

PR #3959, opened 2023-09-28
Speed up RTLIL::Const::decode_string by 1.7x.

rmlarsen:
This replaces part of https://github.com/YosysHQ/yosys/pull/3940. The main source of the speedup is from implementing the extraction of the data from all but the last n % 8 bits as a fixed trip-count loop, which the compiler can unroll. 

Benchmark measurements for a synthesis run of a representative circuit:

Before:
image

After:
image

### DIFF
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -321,7 +321,7 @@ std::string RTLIL::Const::decode_string() const
 	if (i < n) {
 		char ch = 0;
 		for (int j = 0; j < (n - i); j++) {
-			if (bits[j + i] == RTLIL::State::S1) {
+			if (bits[i + j] == RTLIL::State::S1) {
 				ch |= 1 << j;
 			}
 		}
@@ -332,7 +332,7 @@ std::string RTLIL::Const::decode_string() const
 	for (; i >= 0; i -= 8) {
 		char ch = 0;
 		for (int j = 0; j < 8; j++) {
-			if (bits[j + i] == RTLIL::State::S1) {
+			if (bits[i + j] == RTLIL::State::S1) {
 				ch |= 1 << j;
 			}
 		}
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -313,18 +313,33 @@ RTLIL::Const RTLIL::Const::from_string(const std::string &str)
 
 std::string RTLIL::Const::decode_string() const
 {
-	std::string string;
-	string.reserve(GetSize(bits)/8);
-	for (int i = 0; i < GetSize(bits); i += 8) {
+	const int n = GetSize(bits);
+	const int n_over_8 = n / 8;
+	std::string s;
+	s.reserve(n_over_8);
+	int i = n_over_8 * 8;
+	if (i < n) {
 		char ch = 0;
-		for (int j = 0; j < 8 && i + j < int (bits.size()); j++)
-			if (bits[i + j] == RTLIL::State::S1)
+		for (int j = 0; j < (n - i); j++) {
+			if (bits[j + i] == RTLIL::State::S1) {
 				ch |= 1 << j;
+			}
+		}
 		if (ch != 0)
-			string.append({ch});
+			s.append({ch});
 	}
-	std::reverse(string.begin(), string.end());
-	return string;
+	i -= 8;
+	for (; i >= 0; i -= 8) {
+		char ch = 0;
+		for (int j = 0; j < 8; j++) {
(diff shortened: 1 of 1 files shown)


rmlarsen:
FWIW: Here is an expanded bottom up view from the profile. You can see that almost the calls to `decode_string()` for our circuit original in `Yosys::simplemap_mux`.

image

kernel/rtlil.cc
```diff
@@ -4044,7 +4059,7 @@ void RTLIL::SigSpec::replace(const RTLIL::SigSpec &pattern, const RTLIL::SigSpec
 			other->bits_[j] = with.bits_[it->second];
 		}
 	}
-
```
povik, line 4047: Nit: extra newline change

povik (approved):

povik:
> FWIW: Here is an expanded bottom up view from the profile. You can see that almost all the calls to `decode_string()` for our circuit originate in `Yosys::simplemap_mux`.

If that's the case, an idea:
```
diff --git a/passes/techmap/simplemap.cc b/passes/techmap/simplemap.cc
index 11692b715..741bf73c6 100644
--- a/passes/techmap/simplemap.cc
+++ b/passes/techmap/simplemap.cc
@@ -265,7 +265,7 @@ void simplemap_mux(RTLIL::Module *module, RTLIL::Cell *cell)
 
        for (int i = 0; i < GetSize(sig_y); i++) {
                RTLIL::Cell *gate = module->addCell(NEW_ID, ID($_MUX_));
-               gate->add_strpool_attribute(ID::src, cell->get_strpool_attribute(ID::src));
+               gate->attributes[ID::src] = cell->attributes[ID::src];
                gate->setPort(ID::A, sig_a[i]);
                gate->setPort(ID::B, sig_b[i]);
                gate->setPort(ID::S, cell->getPort(ID::S));

```

There's probably a lot of other places we could do the same.

rmlarsen:
> > FWIW: Here is an expanded bottom up view from the profile. You can see that almost all the calls to `decode_string()` for our circuit originate in `Yosys::simplemap_mux`.
> 
> If that's the case, an idea:
> 
> ```
> diff --git a/passes/techmap/simplemap.cc b/passes/techmap/simplemap.cc
> index 11692b715..741bf73c6 100644
> --- a/passes/techmap/simplemap.cc
> +++ b/passes/techmap/simplemap.cc
> @@ -265,7 +265,7 @@ void simplemap_mux(RTLIL::Module *module, RTLIL::Cell *cell)
>  
>         for (int i = 0; i < GetSize(sig_y); i++) {
>                 RTLIL::Cell *gate = module->addCell(NEW_ID, ID($_MUX_));
> -               gate->add_strpool_attribute(ID::src, cell->get_strpool_attribute(ID::src));
> +               gate->attributes[ID::src] = cell->attributes[ID::src];
>                 gate->setPort(ID::A, sig_a[i]);
>                 gate->setPort(ID::B, sig_b[i]);
>                 gate->setPort(ID::S, cell->getPort(ID::S));
> ```
> 
> There's probably a lot of other places we could do the same.

Ah! I'll experiment with this and send a separate CL, if it yields a measurable improvement.

kernel/rtlil.cc
rmlarsen: Hmmm looks like my editor change an \tab\n to  \n. I can't even seem to put it back. Can we include this tiny cleanup, please?

povik: Sure, I think worse have been merged.

rmlarsen: Thanks :-)

kernel/rtlil.cc
```diff
@@ -313,18 +313,33 @@ RTLIL::Const RTLIL::Const::from_string(const std::string &str)
...
+	const int n_over_8 = n / 8;
+	std::string s;
+	s.reserve(n_over_8);
+	int i = n_over_8 * 8;
+	if (i < n) {
 		char ch = 0;
-		for (int j = 0; j < 8 && i + j < int (bits.size()); j++)
-			if (bits[i + j] == RTLIL::State::S1)
+		for (int j = 0; j < (n - i); j++) {
+			if (bits[j + i] == RTLIL::State::S1) {
```
Ravenslofty, line 324: maybe a nit, but must you really change the order of `i` and `j` here?

rmlarsen: @Ravenslofty Thanks for catching.

rmlarsen:
@povik your suggestions gives a major speedup (about 12% overall for my flow). Will send a CL tomorrow.

PR #3963, opened 2023-09-28
Improved `sim` handling of `$display`/`$print` 

jix:
This contains 3 changes that improve support for `$display` calls / `$print` cells in the context of simulating formal verification counterexample and cover traces using `sim` (used by SBY with `fst on` or `vcd_sim on`, will become the default in the future):

## 0e92cc1
Adds printed output to the JSON summary that SBY parses. This includes source and hierarchy information of the `$print` cell which might be useful even for other `sim` uses. I still have to add functionality to SBY to actually do anything with this data.

## 275c3ce
Prior to this, `$display` calls inside initial statements were not emitted as `$print` cells and generated an error if they were not using constant values. During FV it's common to have constant-but-solver-determined values (`rand const reg` or `$anyconst`) and outputting those in a structured way using $display in an initial block is something I very much want to do. This change still produces an elaboration time output when the values are constant but downgrades the error otherwise to an info message for calls that output the formatted string. In both cases it emits a $print cell so that the output during simulation is complete. It remains an error to use non-constant inputs for formatting functions that return the formatted string.

## 0c4939d
The current handling of `%m` statically inserts the module name, while the SV spec says that the hierarchical name should be output. Printing the hierarchical name is very useful when inserting debug logging to a module instantiated several times in the design. Since the hierarchical name is not known at the time the format string is processed, this requires actually keeping it as a format arg. To support it during `sim`, I added an argument to `render()` so `sim` can supply the correct hierarchical name. For the other elaboration time calls to `render()` I keep using the module name as the hierarchical name isn't statically known in that context.

Additionally `render()` assumed that the time is always `0` as it was only used for constant elaboration time calls before #3915, but since this isn't the case anymore I added another argument so that `sim` can supply the correct simulation time.

What is still missing, and the main reason this is a draft PR, is hierarchical `%m` support for CXXRTL. As I haven't really used CXXRTL so far I'm not sure what needs to be done and wanted to get some feedback on the general approach before I figure out what to do there. I'd also want to add some tests for the added functionality.


jix:
That's what I get for not trying a local build after rebasing...

whitequark (approved):

daglem:
FYI: In anticipation of this to be merged, I made a PR in [sv-tests](https://github.com/chipsalliance/sv-tests) to invoke the simulation pass in Yosys: [#5498](https://github.com/chipsalliance/sv-tests/pull/5498).

Merging of the current PR will then better reflect on the capabilites of Yosys in sv-tests, and make more working tests available for further development of Yosys. Yes, I have tested this :smile:

The only issue I can see with the current PR right off the bat is that a `$display` in an `initial` block is now invoked three times (jix/fmt-hiername merged into master after fixing a small conflict). Try e.g. this example: https://github.com/chipsalliance/sv-tests/blob/master/tests/chapter-7/arrays/packed/querying-functions/size.sv

kernel/fmt.cc
```diff
@@ -734,7 +761,9 @@ std::string Fmt::render() const
 					buf = part.sig.as_const().decode_string();
 				} else if (part.type == FmtPart::TIME) {
 					// We only render() during initial, so time is always zero.
```
whitequark, line 763: This comment no longer applies.

kernel/fmt.cc
```diff
@@ -551,6 +560,23 @@ std::vector<VerilogFmtArg> Fmt::emit_verilog() const
 				break;
 			}
 
+			case FmtPart::HIERNAME: {
+				VerilogFmtArg arg;
+				arg.type = VerilogFmtArg::HIERNAME;
+				args.push_back(arg);
+				fmt.str += '%';
+				if (part.plus)
```
whitequark, line 568: Are these modifiers actually valid for `%m`?

frontends/ast/simplify.cc
```diff
@@ -156,6 +156,9 @@ Fmt AstNode::processFormat(int stage, bool sformat_like, int default_base, size_
 			arg.type = VerilogFmtArg::INTEGER;
 			arg.sig = node_arg->bitsAsConst();
 			arg.signed_ = node_arg->is_signed;
+		} else if (!required) {
+			log_file_info(filename, location.first_line, "Skipping system task `%s' with non-constant argument at position %zu.\n", str.c_str(), index + 1);
```
whitequark, line 160: On second thought, I'm not sure we should just skip things we can't evaluate. If the designer doesn't want it to be evaluated they have `ifndef SYNTHESIS`, no?

kernel/fmt.cc
jix: AFAICT the spec is not clear on this and IIRC there was at least some inconsistency between other implementations. In places where the spec is unclear I think it would make sense to support whatever other commonly used open-source tools like iverilog support as long as that's a reasonable interpretation of the spec (although I don't remember what exactly iverilog does here and in general I'd also be fine with not supporting any modifiers for `%m`).

whitequark: > In places where the spec is unclear I think it would make sense to support whatever other commonly used open-source tools like iverilog support as long as that's a reasonable interpretation of the spec (although I don't remember what exactly iverilog does here and in general I'd also be fine with not supporting any modifiers for %m).

I agree with this stance. I just want to make sure that whichever modifiers we implement here actually match other tools, if we do support them.

PR #3965, opened 2023-09-28
check: Un-sigmap the `check` command

povik:
I suspect there are hidden issues with module connections being made in the reverse order. That is, with some passes making the connections in the reverse order and other passes expecting the RHS of connections to be the driver.

The change here modifies the `check` command not to rely on sigmap, instead considering the connections explicitly. So it catches cases like

```
module top;
	wire a, b, c;
	assign a = b;
	assign a = c;
endmodule
```

which are not caught otherwise.

### DIFF
--- a/passes/cmds/check.cc
+++ b/passes/cmds/check.cc
@@ -100,7 +100,6 @@ struct CheckPass : public Pass {
 		{
 			log("Checking module %s...\n", log_id(module));
 
-			SigMap sigmap(module);
 			dict<SigBit, vector<string>> wire_drivers;
 			dict<SigBit, int> wire_drivers_count;
 			pool<SigBit> used_wires;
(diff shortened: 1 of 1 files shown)


povik:
From the Jenkins log:
```
make[1]: Entering directory '/var/lib/jenkins/workspace/Yosys-PR/tests/techmap'
Warning: multiple conflicting drivers for top.\o:
    assignment of $not$<<EOT:3$1100_Y
    assignment of $abc$1186$o
Warning: multiple conflicting drivers for top.\w:
    assignment of $and$<<EOT:2$1099_Y
    assignment of $abc$1186$w
ERROR: Found 2 problems in 'check -assert'.
make[1]: *** [run-test.mk:6: abc9.ys] Error 1
```

I guess the suspicion was right?

whitequark:
IIRC there were some additional issues related to not using sigmap there. Talk to Claire maybe?

clairexen:
i'm all for adding that functionality to `check`, but can we add the following two options, instead of changing the default behavior?

```
    -alias
        Interpret wire-to-wire connections as directionless aliases.
        (This is currently the default behavior.)

    -conn
        Interpret wire-to-wire connections as directional buffers.
        (This may be made the default behavior in the future.)
```

Note: You can define `sigmap` as following to turn `sigmap()` into a no-op when in `conn_mode`:

```
    SigMap sigmap(conn_mode ? nullptr : module);
```

povik:
Another case currently not caught by `check` (stumbled on by @mattvenn):

```
module top;
	wire a;
	assign a = 0;
	assign a = 1;
endmodule
```

With the change here it's caught:
```
Checking module top...
Warning: multiple conflicting drivers for top.\a:
    assignment of 1'0
    assignment of 1'1
Found and reported 1 problems.
```

PR #3967, opened 2023-09-29
Add "buffered-normalized mode", add $buf cell type, and add "bufnorm" command

clairexen:
This PR adds "buffered-normalized mode" to RTLIL. In that mode there's a 1:1 relationship between cell ouput ports and wires, and each wire is annotated with a reference to the cell and cell port driving it.

Once in "buffered-normalized mode" calling `design->bufNormalize()` (or `module->bufNormalize()`) will ensure that the above condition is met, by adding additional intermediate wires and buffer cells as needed.

This PR also adds a `$buf` "coarse-grain" buffer cell type, similar in behavior to `$pos`, but techmap/simplemap/etc is not mapping this buffer type to an array of single-bit buffers. `bufNormalize()` is adding buffers of that new `$buf` type. Note that `bufNormalize()` is very efficient and only looks at the parts of the design that have been changed since the last time `bufNormalize()` has been run, i.e. it does not require a complete scan over the design.

This PR also adds the `bufnorm` command for getting the design into and out of "buffered-normalized mode", with more control over the exact type of network of buffers that is being created.


backends/rtlil/rtlil_backend.cc
```diff
@@ -115,13 +115,17 @@ void RTLIL_BACKEND::dump_sigspec(std::ostream &f, const RTLIL::SigSpec &sig, boo
...
+void RTLIL_BACKEND::dump_wire(std::ostream &f, std::string indent, const RTLIL::Wire *wire, bool flag_d)
 {
 	for (auto &it : wire->attributes) {
 		f << stringf("%s" "attribute %s ", indent.c_str(), it.first.c_str());
 		dump_const(f, it.second);
 		f << stringf("\n");
 	}
+	if (flag_d && wire->driverCell) {
+		f << stringf("%s" "driver %s %s\n", indent.c_str(),
+				wire->driverCell->name.c_str(), wire->driverPort.c_str());
```
povik, line 127: Since it should be always possible to reconstruct this data, why not dump it as a comment in the RTLIL output? (Does RTLIL support comments?) And if we do dump it as a comment, maybe we could make it non-optional and drop the argument to `dump_wire`.

povik: ...and drop the other new dump arguments and the option to the pass.

povik:
> This PR also adds a `$buf` "coarse-grain" buffer cell type, similar in behavior to `$pos`, but techmap/simplemap/etc is not mapping this buffer type to an array of single-bit buffers.

FWIW that doesn't seem to be the behavior for `$pos` cells either

https://github.com/YosysHQ/yosys/blob/7d30f716e82ff4782b653cb2c448062f7878c308/passes/techmap/simplemap.cc#L45-L53

widlarizer:
Since `setPort` may `bufNormQueue.insert`, a `Cell*` may survive in `bufNormQueue` beyond when the underlying cell is removed. Repro: `read_rtlil bug.proc.il; bufnorm -update; opt; bufnorm -update` segfaults on my linux machine) where `bug.proc.il` is
```
autoidx 16

module \bug

  wire width 32 $procmux$14_Y

  wire $procmux$15_CMP

  wire $auto$bugpoint.cc:258:simplify_something$1

  wire width 32 output 1 $auto$bugpoint.cc:258:simplify_something$3

  wire width 30 $auto$bugpoint.cc:258:simplify_something$4

  wire width 32 $auto$bugpoint.cc:258:simplify_something$5

  wire $delete_wire$6

  wire width 32 \pc$next

  cell $mux $procmux$14
    parameter \WIDTH 32
    connect \Y $procmux$14_Y
    connect \S $procmux$15_CMP
    connect \B 0
    connect \A 32'x
  end

  cell $dff $7
    parameter \WIDTH 32
    parameter \CLK_POLARITY 1
    connect \Q $auto$bugpoint.cc:258:simplify_something$3
    connect \D \pc$next
    connect \CLK $auto$bugpoint.cc:258:simplify_something$1
  end

  connect $procmux$15_CMP $delete_wire$6
  connect \pc$next $procmux$14_Y
end
```

whitequark:
oh hey, a new user of bugpoint :D

povik (approved):
Apart from the `bufnorm` command (which I haven't reviewed so I at least marked it experimental) this looks good to me. There are some outstanding issues (e.g. what Emil points out) but they don't affect all use cases. I am in favor of merging this as is to get the ball moving

povik:
Rebased to pull in updated CI

PR #3968, opened 2023-09-29
Speed up simplemap_mux by 9.6x.

rmlarsen:
This is accomplished by directly inserting the cell source attribute in the new object's 'attributes' map instead of creating a new attribute pool and then copying that. Based on a suggestion by Martin Poviser in a comment on https://github.com/YosysHQ/yosys/pull/3959

Topdown view before:
image

Topdown view after:
image

### DIFF
--- a/passes/techmap/simplemap.cc
+++ b/passes/techmap/simplemap.cc
@@ -18,10 +18,10 @@
  */
 
 #include "simplemap.h"
-#include "kernel/ff.h"
 #include "kernel/sigtools.h"
-#include <stdio.h>
+#include "kernel/ff.h"
 #include <stdlib.h>
+#include <stdio.h>
 #include <string.h>
 
 USING_YOSYS_NAMESPACE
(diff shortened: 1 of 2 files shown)


rmlarsen:
Formatting PR in https://github.com/YosysHQ/yosys/pull/3969

Ravenslofty:
I'm inclined to believe this should not be reformatted; it's a two line patch with 250 lines of unrelated changes, which makes it much more difficult to review.

Perhaps we can start by splitting "the patch" as a commit from the reformatting.

povik:
> It's a two line patch with 250 lines of unrelated changes, which makes it much more difficult to review.

FWIW it's more than a two-line patch since Rasmus seems to have gone through other occurrences of the same pattern. Of course that not being obvious is even more of an argument for splitting the formatting from other changes.

Rasmus, I understand there may be frustration from on one hand being told not to include any extra whitespace changes in your patches and on the other hand seeing pushback on #3969. I feel like how we approach formatting is something that would benefit from a bit of discussion among Yosys maintainers.

povik:
Not sure what the CI failure is about:
```
 make[1]: *** [macc.sh] Error 255
```
Let's see if I can trigger a re-run.

rmlarsen:
> > It's a two line patch with 250 lines of unrelated changes, which makes it much more difficult to review.
> 
> FWIW it's more than a two-line patch since Rasmus seems to have gone through other occurrences of the same pattern. Of course that not being obvious is even more of an argument for splitting the formatting from other changes.
> 
> Rasmus, I understand there may be frustration from on one hand being told not to include any extra whitespace changes in your patches and on the other hand seeing pushback on #3969. I feel like how we approach formatting is something that would benefit from a bit of discussion among Yosys maintainers.

Yup, I'd appreciate some clear guidance, once way or the other.

rmlarsen:
FWIW, contributing to Yosys is nowhere close to my day job, so I am trying to find the most friction-less way to land these improvements.

rmlarsen:
Let me back the formatting changes out of this PR for now.

povik:
So it seems the CI failure is not a fluke. I wonder what could have changed to be at fault here.

rmlarsen:
@rmlarsen let's see if it still fails after I roll back the formatting changes. Hang on...

rmlarsen:
@povik PTAL

rmlarsen:
Hmm the test still fails on masOC. I'm afraid I don't understand how to debug this properly.

whitequark:
Also on Linux, with a similar error: https://github.com/YosysHQ/yosys/actions/runs/6356727207/job/17266798637

rmlarsen:
@whitequark thanks. I'm able to replicate here, after installing iverilog.

rmlarsen:
Reverted changes to techmap.cc, which had no significant impact on timing anyway. PTAL.

rmlarsen:
OK, it did have a small impact on timing. The speedup drops from 11.6x to 9.6x. I can live with that. :-)

Ravenslofty (approved):
I'm happy to merge this if someone else is.

povik:
Go ahead!

PR #3969, opened 2023-09-29
Format files in yosys/passes/techmap

rmlarsen:
The files were formatted in accordance with yosys/.clang-format using the command:

 % clang-format -i -style=file:./.clang-format passes/techmap/*.{cc,h}

### DIFF
--- a/passes/techmap/abc.cc
+++ b/passes/techmap/abc.cc
@@ -30,7 +30,8 @@
 // http://en.wikipedia.org/wiki/Topological_sorting
 
 #define ABC_COMMAND_LIB "strash; &get -n; &fraig -x; &put; scorr; dc2; dretime; strash; &get -n; &dch -f; &nf {D}; &put"
-#define ABC_COMMAND_CTR "strash; &get -n; &fraig -x; &put; scorr; dc2; dretime; strash; &get -n; &dch -f; &nf {D}; &put; buffer; upsize {D}; dnsize {D}; stime -p"
+#define ABC_COMMAND_CTR                                                                                                                              \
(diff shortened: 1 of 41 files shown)


whitequark (changes requested):
My usual questions with these changes are:
- How do we keep the code formatted?
- What happens to `git blame`?

Especially the second.

Absent a satisfactory answer I wouldn't want to see it happen.

Ravenslofty:
(There are also my personal opinions about enforcing formatting being something of a barrier to entry (nevermind an accessibility issue to some), and heaven knows that newcomers to Yosys are few enough already...)

whitequark:
> There are also my personal opinions about enforcing formatting being something of a barrier to entry

I agree! Failing CI because clang-format failed is so absurdly disrespectful to your contributors that unless I *really* want to get my commit in I would just not bother probably.

rmlarsen:
Oh, I agree! I just assumed that the presence of the `.clang-format`  file was an indicator that you put some value on enforcing a unified code formatting (as much as I find the existing formatting with 8 space tabs bizarre and ugly.)

If that's not the case, I'm more than happy to back out formatting changes to all my PRs.

rmlarsen:
BTW: There are ways to tell git blame to ignore certain PRs. I believe this is supported by github.( It is unfortunatelt not supported by gitlab where we host Eigen  :-( .) See https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revs-fileltfilegt

I agree that this change by itself does not make sense. It seems that the Yosys maintainers need to make a decision regarding the enforcement of formatting.

PR #3975, opened 2023-10-02
Speed up OptMergePass by 1.75x.

rmlarsen:
The main speedup comes from switching away from using SHA1 hash to using `std::hash<std::string>` for fingerprinting cells. There is no need to use an expensive cryptographic hash for fingerprinting in this context.

Flamegraph of benchmark timings from synthesis of a representative circuit:

Before:
image

After:
image

### DIFF
--- a/kernel/celltypes.h
+++ b/kernel/celltypes.h
@@ -56,7 +56,7 @@ struct CellTypes
 		setup_stdcells_mem();
 	}
 
-	void setup_type(RTLIL::IdString type, const pool<RTLIL::IdString> &inputs, const pool<RTLIL::IdString> &outputs, bool is_evaluable = false)
+	void setup_type(const RTLIL::IdString& type, const pool<RTLIL::IdString> &inputs, const pool<RTLIL::IdString> &outputs, bool is_evaluable = false)
 	{
 		CellType ct = {type, inputs, outputs, is_evaluable};
 		cell_types[ct.type] = ct;
@@ -298,24 +298,24 @@ struct CellTypes
 		cell_types.clear();
 	}
 
-	bool cell_known(RTLIL::IdString type) const
+	bool cell_known(const RTLIL::IdString& type) const
 	{
 		return cell_types.count(type) != 0;
 	}
 
-	bool cell_output(RTLIL::IdString type, RTLIL::IdString port) const
+	bool cell_output(const RTLIL::IdString& type, const RTLIL::IdString& port) const
 	{
 		auto it = cell_types.find(type);
 		return it != cell_types.end() && it->second.outputs.count(port) != 0;
 	}
 
-	bool cell_input(RTLIL::IdString type, RTLIL::IdString port) const
+	bool cell_input(const RTLIL::IdString& type, const RTLIL::IdString& port) const
 	{
 		auto it = cell_types.find(type);
 		return it != cell_types.end() && it->second.inputs.count(port) != 0;
 	}
 
-	bool cell_evaluable(RTLIL::IdString type) const
+	bool cell_evaluable(const RTLIL::IdString& type) const
 	{
 		auto it = cell_types.find(type);
 		return it != cell_types.end() && it->second.is_evaluable;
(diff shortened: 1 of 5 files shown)


whitequark:
You could probably use an even cheaper hash than SHA1.

kernel/celltypes.h
```diff
@@ -56,7 +56,7 @@ struct CellTypes
 		setup_stdcells_mem();
 	}
 
-	void setup_type(RTLIL::IdString type, const pool<RTLIL::IdString> &inputs, const pool<RTLIL::IdString> &outputs, bool is_evaluable = false)
+	void setup_type(const RTLIL::IdString& type, const pool<RTLIL::IdString> &inputs, const pool<RTLIL::IdString> &outputs, bool is_evaluable = false)
```
povik, line 59: Are you sure those const references are worth it? AFAIK `IdString` is just an int.

kernel/rtlil.h
```diff
@@ -308,10 +308,14 @@ namespace RTLIL
 		bool operator!=(const char *rhs) const { return strcmp(c_str(), rhs) != 0; }
 
 		char operator[](size_t i) const {
-			const char *p = c_str();
+                        const char *p = c_str();
+#ifndef NDEBUG
```
Ravenslofty, line 312: Is checking that a series of bytes aren't zero really that expensive? I feel a little unnerved hiding away consistency checks like this.

passes/opt/opt_merge.cc
```diff
@@ -268,13 +267,13 @@ struct OptMergeWorker
 			}
 
 			did_something = false;
-			dict<std::string, RTLIL::Cell*> sharemap;
+                        std::unordered_map<FingerPrint, RTLIL::Cell*> sharemap;
```
Ravenslofty, line 270: I think this should remain `dict`; the Yosys codebase tends to avoid STL associative containers due to them having subtly different behaviour across platforms.

Ravenslofty (changes requested):
see the comments I made, plus a more general "IdString should be passed by value".

kernel/rtlil.h
rmlarsen: While this function was a modest contributor to the overall time, it was visible in the profile, and the change made it several times faster / easily inlineable.

kernel/celltypes.h
rmlarsen: You are right. Something I did here got rid of a lot of calls to the (very expensive) `IdString` constructor that takes a string. Let me go back and figure out what. Maybe some implicit conversion somewhere. 

I'm thinking that it might be a good idea to mark that constructor as `explicit` (which is part of the Google style guide, BTW), since that is a performance pitfall in fairly innocent looking code.

povik: That sounds like something definitely worth investigating.

rmlarsen:
@Ravenslofty is there a setting to auto-squash my commits here? I noticed my commits didn't get squashed for the PRs you merged yesterday.

Ravenslofty:
There is, but I have the habit of "rebase and merge" to not lose history; if you want your commits squashed I can do that with my next merge.

povik:
Btw @Ravenslofty we were discussing in the dev JF that doing merge commits is better since they have the PR number, which is otherwise lost.

rmlarsen:
OK, I'll leave that up to you. I'm just used to us squashing for Eigen.

povik:
FWIW squash commits should have the PR number and are ok too, it's just the rebase that loses it.

passes/opt/opt_merge.cc
rmlarsen: Done.

kernel/celltypes.h
rmlarsen: Reverted. I'll investigate this separately.

rmlarsen:
PTAL

Ravenslofty (approved):
Yeah, this looks much better.

rmlarsen:
> Yeah, this looks much better.

I just ran the measurements again, and it actually ended up slightly faster too. Go figure... :-)

PR #3977, opened 2023-10-03
gowin: fix abc9 attributes and specify blocks

Ravenslofty:
A certain past wolfenby wrote a bunch of sequential paths in `specify` blocks to describe asynchronous sets/resets within `abc9_box` cells, which require combinational paths. This horrifically broke ABC9 flows, and I apologise.

While I was here, I took the liberty of turning some fully-synchronous cells into `abc9_flop` rather than `abc9_box`, which will likely help sequential synthesis (if somebody asks me nicely to implement that).

cc @pepijndevos @yrabbit

### DIFF
--- a/techlibs/gowin/cells_sim.v
+++ b/techlibs/gowin/cells_sim.v
@@ -197,7 +197,7 @@ module DFFE (output reg Q, input D, CLK, CE);
   end
 endmodule // DFFE (positive clock edge; clock enable)
 
-(* abc9_box, lib_whitebox *)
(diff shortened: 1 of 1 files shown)


pepijndevos (approved):

Ravenslofty:
@mmicko did I break something? Jenkins complains but the error is somewhere inside sby?

mmicko:
@Ravenslofty it is fine, just please rebase on master, there is fix already for error

PR #3991, opened 2023-10-07
alumacc: merge independent of sign

adrianparvino:


Ravenslofty:
I think I'd want an equivalence test written for this to make sure that this transformation is correct.

Ravenslofty (approved):

Ravenslofty:
As it turns out, the test case also needed to cover the merging of comparisons with arithmetic, as #6063 shows.

PR #3992, opened 2023-10-07
write_verilog: avoid emitting empty cases.

mwkmwkmwk:
The Verilog grammar does not allow an empty case.  Most synthesis tools are quite permissive about this, but Quartus is not.  This causes problems for amaranth with Quartus (see amaranth-lang/amaranth#931).


mcclure:
Can confirm this fixed https://github.com/amaranth-lang/amaranth/issues/931 (the compile error at least)

PR #3994, opened 2023-10-09
cxxrtl: export wire attributes through the C API

whitequark:
Among other things this allows retrieving source locations for wires in the simulation, e.g.:

```c
void print_location_cb(void *, const char *name, struct cxxrtl_object *object, size_t parts) {
	printf("%s:\n", name);
	cxxrtl_attr_set attrs = object->attrs;
	if (cxxrtl_attr_type(attrs, "src") == CXXRTL_ATTR_STRING) {
		const char *src = cxxrtl_attr_get_string(attrs, "src");
		while (src) {
			char buf[256] = {0};
			const char *next = strchr(src, '|');
			if (!next) {
				printf("\t%s\n", src);
				src = NULL;
			} else {
				strncpy(buf, src, next - src);
				printf("\t%s\n", buf);
				src = next + 1;
			}
		}
	} else {
		printf("(missing!)\n");
	}
}

int main() {
	cxxrtl_handle top = cxxrtl_create_at(cxxrtl_design_create(), "toplevel");
	cxxrtl_reset(top);

	printf("Source location information for this simulation:\n");
	cxxrtl_enum(top, NULL, print_location_cb);

	return 0;
}
```

### DIFF
--- a/backends/cxxrtl/cxxrtl_capi.h
+++ b/backends/cxxrtl/cxxrtl_capi.h
@@ -249,6 +249,15 @@ struct cxxrtl_object {
 	// this field to NULL.
 	struct _cxxrtl_outline *outline;
 
+	// Opaque reference to an attribute set.
+	//
+	// See the documentation of `cxxrtl_attr_set` for details. When creating a `cxxrtl_object`, set
+	// this field to NULL.
+	//
+	// The lifetime of the pointers returned by `cxxrtl_attr_*` family of functions is the same as
+	// the lifetime of this structure.
+	struct _cxxrtl_attr_set *attrs;
+
 	// More description fields may be added in the future, but the existing ones will never change.
 };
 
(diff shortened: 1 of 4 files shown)


backends/cxxrtl/cxxrtl_capi.h
```diff
@@ -304,6 +313,62 @@ typedef struct _cxxrtl_outline *cxxrtl_outline;
...
+
+// Retrieve a string valued attribute from an attribute set. The returned string is zero-terminated.
+//
+// This function asserts that `cxxrtl_attr_type(attrs, name) == CXXRTL_ATTR_STRING`. If assertions
+// are disabled, returns NULL if the attribute is missing or has an incorrect type.
+const char *cxxrtl_attr_get_string(cxxrtl_attr_set attrs, const char *name);
+
+// Retrieve a string valued attribute from an attribute set.
+//
+// This function asserts that `cxxrtl_attr_type(attrs, name) == CXXRTL_ATTR_STRING`. If assertions
```
charlottia, line 368:
```suggestion
// Retrieve a double valued attribute from an attribute set.
//
// This function asserts that `cxxrtl_attr_type(attrs, name) == CXXRTL_ATTR_DOUBLE`. If assertions
```

whitequark: Thanks!

PR #4000, opened 2023-10-10
alumacc: alternative cmp unification implementation

adrianparvino:
`$lt` and related are sometimes inhibited from unifying with `$sub` as when `B < A`, then the `$sub` cell is unable to swap inputs, but `$lt` does. This resolves this problem by instead considering `sig_alu` as an unordered pair mapping.

Addendum:
`$add` and `$sub` might still be unable to unify with `share -aggressive` but changing the behavior [1] causes [2] to fail, so I'm keeping it to `cmp` only.

[1] https://github.com/YosysHQ/yosys/commit/ded586c0485484996ea85f3eb596321582631345#diff-f5a16f8ab735a667cabcdd605b428eeaa92f223f044ef01a91d52141d404633bL334-L336
[2] https://github.com/YosysHQ/yosys/blob/ded586c/tests/opt/opt_expr.ys#L27

### DIFF
--- a/passes/techmap/alumacc.cc
+++ b/passes/techmap/alumacc.cc
@@ -405,11 +405,6 @@ struct AlumaccWorker
 			RTLIL::SigSpec B = sigmap(cell->getPort(ID::B));
 			RTLIL::SigSpec Y = sigmap(cell->getPort(ID::Y));
 
-			if (B < A && GetSize(B)) {
-				cmp_less = !cmp_less;
-				std::swap(A, B);
-			}
-
 			alunode_t *n = nullptr;
 
 			for (auto node : sig_alu[RTLIL::SigSig(A, B)])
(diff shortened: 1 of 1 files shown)


adrianparvino:
This fails as reordering no longer happens. https://github.com/YosysHQ/yosys/blob/ded586c/tests/opt/opt_expr.ys#L27

passes/techmap/alumacc.cc
```diff
@@ -331,9 +331,6 @@ struct AlumaccWorker
 			if (GetSize(C) > 1)
 				goto next_macc;
 
-			if (!subtract_b && B < A && GetSize(B))
-				std::swap(A, B);
-
```
adrianparvino, line 336: In hindsight, this can remain, as `cmp` unification now is order-independent. This also allows the tests to pass.

povik:
I've rebased and pushed a small fix

povik (approved):

PR #4003, opened 2023-10-12
quicklogic: Fix pp3 `dffs` test

povik:
Fix name confusion which was making the test look into the vendor's cell blackbox rather than into the synthesis results.


povik:
Closing and opening to trigger a CI rerun

PR #4019, opened 2023-11-02
ice40, ecp5: enable ABC9 by default

Ravenslofty:
It's been [four years](https://github.com/YosysHQ/yosys/pull/1544) since ABC9 was last proposed to be used by default. The arguments made back then still apply: results are superior compared to the default ABC flow, and we seem to have gotten most of the bugs fixed or worked around.

So, I'm once again asking for ~~your financial support~~ ABC9 to be made the default on iCE40 and ECP5. I think using these two architectures as a baseline for wide adoption should quickly expose any other bugs in ABC9 that can then be fixed, given how popular they are over the others.

### DIFF
--- a/techlibs/ecp5/synth_ecp5.cc
+++ b/techlibs/ecp5/synth_ecp5.cc
@@ -93,8 +93,8 @@ struct SynthEcp5Pass : public ScriptPass
 		log("    -abc2\n");
 		log("        run two passes of 'abc' for slightly improved logic density\n");
 		log("\n");
-		log("    -abc9\n");
-		log("        use new ABC9 flow (EXPERIMENTAL)\n");
+		log("    -noabc9\n");
+		log("        disable use of new ABC9 flow\n");
 		log("\n");
 		log("    -vpr\n");
 		log("        generate an output netlist (and BLIF file) suitable for VPR\n");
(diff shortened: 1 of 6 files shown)


whitequark (approved):
Based on discussion with @Ravenslofty I am in support of this change.

Ravenslofty:
Well, here goes nothing.

nakengelhardt:
I feel like this is a big enough change that it should have been discussed before merging...

Ravenslofty:
Would you be willing to humour me?

- It's past the beginning of the month, so the commercial builds of Yosys use the stable ABC flow.
- There's a built-in option in this patch to revert back to that ABC flow, for emergencies.
- ABC9 has been experimental for...five years now, and the only way for us to tell how stable it is in the real world is for it to get out there and receive bug reports.
- I'm quite happy for this to be reverted at the end of the month, but I'm a little tired of the circle of "it's experimental because we don't know how it behaves in the real world" and "we don't know how it behaves in the real world because it's experimental".

Ultimately, as the de-facto maintainer of ABC9, responsibility for fixing the ensuing bug reports falls on my shoulders, and I would rather have this out there and then get reverted so bugs can be found and fixed than the present "useful feature is not enabled by default and scares users with a big `(EXPERIMENTAL)`."

nakengelhardt:
And this was so urgent that you could not wait until Monday to discuss these points at the dev JF? The release for this month is not out yet.

Ravenslofty:
Okay, I've reverted it.

PR #4021, opened 2023-11-07
Change `booth` architecture for improved delay, similar signed/unsigned results

povik:
~~Implement Wallace tree summation as an option in `booth`. This significantly improves the propagation delay of the multiplier, at least for some dimensions of inputs and outputs. So far the Wallace tree option is only supported with the signed multiplier architecture, and is implemented in the most crude way possible. There's likely space for improvement.~~ Redo the selection of architectures: Have one classic Booth architecture for both signed and unsigned multiplication. Move what was formerly the signed architecture under a new `-lowpower` option. Also improve the balance of the Wallace tree generated.

With these changes, the Booth multiplier no longer has worse timing on MachXO3 than the baseline multiplier emitted by `maccmap`. Since Booth can use a lot less area, opt for it as the default on MachXO3, making that architecture the early adopter. (MachXO3 has no hard logic for multipliers.)

### DIFF
--- a/kernel/rtlil.h
+++ b/kernel/rtlil.h
@@ -921,6 +921,9 @@ public:
 	RTLIL::SigSpec extract(int offset, int length = 1) const;
 	RTLIL::SigSpec extract_end(int offset) const { return extract(offset, width_ - offset); }
 
+	RTLIL::SigBit lsb() const { log_assert(width_); return (*this)[0]; };
+	RTLIL::SigBit msb() const { log_assert(width_); return (*this)[width_ - 1]; };
+
 	void append(const RTLIL::SigSpec &signal);
 	inline void append(Wire *wire) { append(RTLIL::SigSpec(wire)); }
 	inline void append(const RTLIL::SigChunk &chunk) { append(RTLIL::SigSpec(chunk)); }
--- a/passes/techmap/booth.cc
+++ b/passes/techmap/booth.cc
@@ -307,7 +307,7 @@ struct BoothPassWorker {
 			      SigSpec Y, // multiplier
 			      SigSpec Z)
 	{ // result
-		int x_sz = X.size(), z_sz = Z.size();
+		int z_sz = Z.size();
 
 		SigSpec one_int, two_int, s_int, sb_int;
 		int encoder_count = 0;
@@ -376,7 +376,7 @@ struct BoothPassWorker {
 		for (int i = 0; i < encoder_count + 1; i++)
 			aligned_pp[i].extend_u0(z_sz);
 
-		AlignPP(x_sz, z_sz, ppij_int, aligned_pp);
+		AlignPP(z_sz, ppij_int, aligned_pp);
 
 		// Debug: dump out aligned partial products.
 		// Later on yosys will clean up unused constants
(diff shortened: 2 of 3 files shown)


andyfox-rushc:
Hi, Martin,

You are right: Wallace Trees are the traditional way of doing the partial product summation in Booth multipliers and should be used.  

The original code has two flows: unsigned (which uses the classic Booth encoding with Wallace Tree -- ie CSA tree -- implementation) and signed (which uses a low power algorithm & does not use a Wallace Tree; I was interested in learning the new low power multiplier architectures:  A Low Power Radix-4 Booth Multipliers with Pre-Encoded Mechanism, IEEE Access  https://ieeexplore.ieee.org/document/9121226, so slipped it in on the signed flow). 

The unsigned flow is the original "classic" Booth with Wallace tree. The unsigned flow can be modified to handle signed numbers by removing the "padding" of the multiplier which assures a zero at the end of the multiplier number (and hence kills the sign extension).

 //
 // step case. If multiplier ends on a boundary
// then add an extra booth encoder bounded by
// zeroes to ensure unsigned works.
//
RTLIL::Wire *y0_wire;
RTLIL::Wire *y1_wire;
RTLIL::Wire *y2_wire;

This weirdness is explained in Appendix A of  Gary Bewick's thesis (http://i.stanford.edu/pub/cstr/reports/csl/tr/94/617/CSL-TR-94-617.pdf). The concept is stick a zero at the end of the multiplier to kill sign extension.

So if I was to redo this I would give the multiplier a power parameter and use that to gate the architecture (rather than the signed/unsigned choice). My guess is that ASIC folks might want the lowest power choice.

For low power I would choose one of the "neg first" low power algorithms.

For non lower power I would use the traditional Booth -> Reduced Partial Products -> Wallace Tree -> CPA -> Result 
(Again, the unsigned flow is the traditional booth and to make it work for signed it is a matter of conditionally removing the "padding" -- for signed remove the extra gates and include sign extension).

You are absolutely right to observe that summation of partial products is best done with Wallace Trees. Thanks for the interesting note.

Andy

povik:
Hi, Andy,

thanks for getting back to me and for the clarifications. I think I saw then forgot the unsigned flow builds a Wallace tree already. Taking it into account the `-wallace` option doesn't make that much sense.

> The unsigned flow can be modified to handle signed numbers by removing the "padding" of the multiplier which assures a zero at the end of the multiplier number (and hence kills the sign extension).

That's good to know, thanks!

> So if I was to redo this I would give the multiplier a power parameter and use that to gate the architecture (rather than the signed/unsigned choice).

Yes, this is the behavior we want to move to. Given that the classic and low-power architectures can have vastly different parameters (at least the delay in FPGA implementations, which is what I am focused on right now), we shouldn't be switching between one and the other based on whether the multiplier is signed or unsigned. We want a similar baseline signed and unsigned architecture (for which it is useful we can turn the unsigned architecture into signed like you have pointed out), then we can gate the alternative low-power architecture behind a command option.

By the way, I have been reading the reference you provide (A Low Power Radix-4 Booth Multipliers with Pre-Encoded Mechanism, IEEE Access https://ieeexplore.ieee.org/document/9121226), and it seems what they propose there isn't exactly what we have implemented as the signed flow. In the reference they talk about a "pre-encoding mechanism" with clock-driven latches. For those we have no counterpart in our purely-combinational implementation. We should account for that in the help message of the pass, say that it's a modified architecture from the one in the reference perhaps, or use some other reference.

Martin

povik:
> > So if I was to redo this I would give the multiplier a power parameter and use that to gate the architecture (rather than the signed/unsigned choice).
> 
> Yes, this is the behavior we want to move to.

I made this change. Also as far as I could tell the Wallace tree generated in the unsigned flow was unbalanced, so I redid that part.

PR #4026, opened 2023-11-11
Fix printf formats

uis246:
Just what compiler found. Uses proper types like %zu for size_t and %PRIx64 for uint64_t in hexadecimal.

### DIFF
--- a/passes/memory/memory_libmap.cc
+++ b/passes/memory/memory_libmap.cc
@@ -690,7 +690,7 @@ bool apply_clock(MemConfig &cfg, const PortVariant &def, SigBit clk, bool clk_po
 
 // Perform write port assignment, validating clock options as we go.
 void MemMapping::assign_wr_ports() {
-	log_reject(stringf("Assigning write ports... (candidate configs: %lu)", cfgs.size()));
+	log_reject(stringf("Assigning write ports... (candidate configs: %zu)", (size_t) cfgs.size()));
 	for (auto &port: mem.wr_ports) {
 		if (!port.clk_enable) {
 			// Async write ports not supported.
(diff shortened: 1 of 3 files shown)


passes/memory/memory_libmap.cc
```diff
@@ -690,7 +690,7 @@ bool apply_clock(MemConfig &cfg, const PortVariant &def, SigBit clk, bool clk_po
 
 // Perform write port assignment, validating clock options as we go.
 void MemMapping::assign_wr_ports() {
-	log_reject(stringf("Assigning write ports... (candidate configs: %lu)", cfgs.size()));
+	log_reject(stringf("Assigning write ports... (candidate configs: %zu)", cfgs.size()));
```
povik, line 693: As I am reading into the standard `std::vector<T>::size()` returns an implementation-defined size type. Do we need to both use `%zu` and add an explicit cast to `size_t` for this to be portable?

uis246: Can implementation-defined size type be larger than architecture-defined size type? Does C++ standard require vector to be completely in memory? If yes, then cast to size_t is safe, if not... 64-bit size_type on 32-bit CPU with 32-bit size_t... Should be unlikely enough. Even current state is better, than the one before.

nakengelhardt: From a practical perspective, I would argue that if you have more than 4 billion possible configuration options to choose between to map your memory instance, getting the number truncated in a debug log message is the least of your problems... we could cast to `uint64_t` just in case, but I'm happy with anything as long as we don't cast to `char`.

uis246: I will just cast to size_t

uis246:
Did I randomy request review?

nakengelhardt:
Nah, it's a [github feature](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) where it automatically requests review when any file listed in CODEOWNERS is modified. (Also, IIRC you need at least triage permissions on a repo to request reviews.)

povik:
@uis246 Can I force-push to your branch with the `size_t` casts so we can merge this?

uis246:
Yes. Sorry I forgot about this.

PR #4028, opened 2023-11-13
synth_lattice: Optionally do constant comparisons in soft logic

povik:
Here we are adding a `-cmp2softlogic` option to `synth_lattice` to implement constant comparisons (arithmetic comparisons of a signal to a constant) in soft logic. Take e.g. the following:

```
read_verilog <<EOF
module top(input [15:0] a, output y);
	assign y = a >= 16'h7700;
endmodule
EOF
synth_lattice -family xo3 [-cmp2softlogic]
```

With the soft logic option, this takes up
```
   Number of cells:                  3
     LUT4                            3
```
while ordinarily it would take

```
   Number of cells:                 14
     CCU2D                           8
     LUT4                            6
```

The lower usage is due to two factors:

 * Once the comparisons are lowered to hard carry chains they won't be optimized, but here e.g. the lower 8 bits of the signal need not be compared at all, which is optimized away in the soft logic implementation. We could, and probaly should, extend `opt_expr` to optimize away the lower bits of the comparison on the `$ge` and other cells if the constant allows that, probably conditioned on the `-fine` mode. That would optimize the comparison before it gets implemented with the carry chains, so it would close the usage gap here somewhat.

* Optimized or not optimized, simple LUT implementation of the comparison takes up less resources anyway, by (I think) a factor of three. This is at the cost of higher delay since we are not using the carry paths.


whitequark:
> Here we are adding a `-cmp2softlogic` option to `synth_lattice` to implement constant comparisons (arithmetic comparisons of a signal to a constant) in soft logic. Take e.g. the following:

I already implemented this in [cmp2lut.v](https://github.com/YosysHQ/yosys/blob/master/techlibs/common/cmp2lut.v), which should already be applied to iCE40. Is there an issue with that techmap?

Ravenslofty:
The issue is that `cmp2lut.v` only works for comparisons that fit within a single LUT; this works across multiple LUTs.

povik:
As far as I can tell, in the Lattice flow, the `$lut` that's generated by `cmp2lut.v` gets techmapped to Yosys gates and eaten up by `abc` and mapped to LUTs again, so it doesn't need to map onto a single physical LUT. Of course the intermediate LUT data grow exponentially in size of the signal to be compared, so that is the issue with using it for wide comparisons. We might want to supersede `cmp2lut.v` with something like `cmp2softlogic.v`, even if by default we don't enable it for comparisons wider than the architecture LUT width.

whitequark:
That makes sense.

PR #4044, opened 2023-11-27
sim: Evaluate all cells once

povik:
If a cell starts with undefined inputs and the value of those inputs is never changed, `sim` code will never evaluate that cell. This means its output will be left at undefined, but in some edge cases such cell can evaluate to a defined value. To address those cases evaluate all cells at least once.


povik:
One way this can be hit in practice is when some of the `memory_*` passes emit `$eq` cells comparing zero-sized signals, and those are not removed by `opt_expr` before `sim` is ran.

passes/sat/sim.cc
```diff
@@ -287,16 +287,18 @@ struct SimInstance
 
 			if (mod != nullptr) {
 				dirty_children.insert(new SimInstance(shared, scope + "." + RTLIL::unescape_id(cell->name), mod, cell, this));
-			}
-
-			for (auto &port : cell->connections()) {
+			} else {
+				for (auto &port : cell->connections())
```
KrystalDelusion, line 291: What's the reasoning for moving this into an else block?  Is marking the bits dirty redundant if the child is already marked dirty?

jix:
In addition to a response to @KrystalDelusion question, it would be helpful to have a concrete reproducer for the issue this is trying to fix. There have been some subtle issues with sim initialization and update order in the past, so having as much context as possible would be helpful for reviewing anything related to that.

povik:
> it would be helpful to have a concrete reproducer

I think an `$eq` of two zero-width signals will be mishandled without this fix. Unfortunately I have lost all other context.

PR #4049, opened 2023-12-03
Enable bram for Gowin

pepijndevos:
This should work now thanks to @yrabbit  https://github.com/YosysHQ/apicula/pull/202


mmicko:
@pepijndevos Is this ready to be merged now ?

pepijndevos:
I think yes but @yrabbit is working on a more extensive bram inference PR I think so worth checking with him what the preferred path would be.

PR #4050, opened 2023-12-04
quicklogic: Generate `bram_types_sim.v` at build time

povik:
The source tree has a checked-in version of `bram_types_sim.v` even though that's a file that is automatically generated. Check out the file from version control, and instead add rules to generate the file on-demand.

By making sure the generated file is up-to-date, this fixes

    ../share/quicklogic/qlf_k6n10f/dsp_sim.v:0: ERROR: Can't find object for defparam `RAM_INIT`!

raised by

    synth_quicklogic -family qlf_k6n10f -bram_types


techlibs/quicklogic/Makefile.inc
```diff
@@ -1,6 +1,9 @@
 %_pm.h: passes/pmgen/pmgen.py %.pmg
 	$(P) mkdir -p pmgen && $(PYTHON_EXECUTABLE) $< -o $@ -p $(subst _pm.h,,$(notdir $@)) $(filter-out $<,$^)
 
+techlibs/quicklogic/qlf_k6n10f/bram_types_sim.v: techlibs/quicklogic/qlf_k6n10f/generate_bram_types_sim.py
+	$(P) python3 $^ $@
```
mmicko, line 5: Use $(PYTHON_EXECUTABLE) here instead of python3

povik: Pushed a change

mmicko (approved):
LGTM, just merge when CI is happy

PR #4045, opened 2023-11-27
Upstream QuickLogic k6n10f flow

povik:
- [x] Look into libmap-based inference not setting the right attributes for TDP36K per-configuration renaming in QuickLogic’s master
- [x] Sort out the licensing issue
- [x] create a draft PR from Martin's branch
- [x] Add the tests for the base arch cells
- [x] Add small simulation tests for memory
  - [x] for expected code patterns, check they infer
  - [x] and simulation check that they work for some selected addresses
- [x] Add small simulation tests for DSP
  - [x] for expected code patterns, check they infer
  - [x] and simulation check that they work for some selected values

PR #4051, opened 2023-12-05
Add WASI CI build

mmicko:
Added script from https://github.com/YosysHQ/yosys/pull/4045, was not able to reproduce a bug but it more looks like order of execution issue in passes/pmgen directory. Let see if this CI is happy.

### DIFF
--- a/.github/workflows/wasi.yml
+++ b/.github/workflows/wasi.yml
@@ -1,0 +1,30 @@
+name: WASI Build
+
+on: [push, pull_request]
+
+jobs:
+  wasi:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Build
+        run: |
+          WASI_SDK=wasi-sdk-19.0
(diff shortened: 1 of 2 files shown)


whitequark (approved):
LGTM

povik:
I am not that familiar with GitHub Actions but I wonder if splitting the SDK fetch from the other commands and putting it above the source checkout would cause GitHub to sometimes cache the environment with the unpacked SDK tree

mmicko:
No caching enabled here so no issues. Spliting into multiple steps does not help since there is no reuse. Leaving this to  be decided on Mondays meeting.

povik:
Well, I meant caching by the GitHub infrastructure itself, since it sees there are no variable inputs to the workflow up to the point of the source checkout. Though it's probably not that clever to deduce the link to the SDK hasn't changed content

whitequark:
> Well, I meant caching by the GitHub infrastructure itself, since it sees there are no variable inputs to the workflow up to the point of the source checkout. Though it's probably not that clever to deduce the link to the SDK hasn't changed content

GitHub does no such thing. We could make it cache explicitly but there's no point since the SDK is downloaded from GitHub's infrastructure in the first place. There would be no meaningful difference in speed, it would just waste some space.

PR #4052, opened 2023-12-05
Remove redundant (but also altered) rule

mmicko:


povik:
Ha, as it happens I just posted a fix too. Anyway I think mine is the proper fix, since we still need to make sure the `techlibs/quicklogic` directory exists in the build tree at the time we are generating the header.

(Only after I learned about what `VPATH` does in Makefiles did the linked YoWASP build failure in #4045 start to make sense.)

povik:
The other PR: #4053

mmicko:
Agree. Already asked @whitequark to take a look at your PR.

whitequark:
What the heck is `VPATH`.

whitequark:
Okay, I see, that's cursed.

PR #4053, opened 2023-12-05
pmgen: Have a single make pattern

povik:
Remove duplicate %.pmg -> %_pm.h pattern. One of the duplicates overrode the other, and in some conditions there were build races as to whether the target directory for the generated header would exist. Instead have a single rule which is properly generalized.


whitequark:
@mmicko One unfortunate issue is that, I think, the tagged release wouldn't build in YoWASP without a custom patch or something...

povik:
You could roll the dice again on the race :|

povik:
The one CI failure appears to be a fluke. Merging.

PR #4055, opened 2023-12-06
sim: Print hierarchy for failed assertions

povik:
Take e.g.

```
module top();
	a a_();
endmodule

module a();
	 b b_();
endmodule

module b();
	always @(*)
		assert(0);
endmodule
```

as input to sim (`read_verilog -sv sample.sv; hierarchy -top top; proc; sim -assert`).

You will now get

```
4. Executing SIM pass (simulate the circuit).
Simulating cycle 0.
Simulating cycle 1.
Failed assertion $assert$sample.sv:10$5 at
    sample.sv:10.13-11.12
  in instance b_ of module b at
    sample.sv:6.5-6.9
  in instance a_ of module a at
    sample.sv:2.4-2.8
ERROR: Assertion top.a_.b_.$assert$sample.sv:10$5 (sample.sv:10.13-11.12) failed.
```

### DIFF
--- a/passes/sat/sim.cc
+++ b/passes/sat/sim.cc
@@ -775,6 +775,30 @@ struct SimInstance
 		return did_something;
 	}
 
+	static void log_source(RTLIL::AttrObject *src)
+	{
+		for (auto src : src->get_strpool_attribute(ID::src))
+			log("    %s\n", src.c_str());
+	}
+
+	void log_cell_w_hierarchy(std::string opening_verbiage, RTLIL::Cell *cell)
+	{
+		log_assert(cell->module == module);
+		bool has_src = cell->has_attribute(ID::src);
+		log("%s %s%s\n", opening_verbiage.c_str(),
+			log_id(cell), has_src ? " at" : "");
+		log_source(cell);
+
+		struct SimInstance *sim = this;
+		while (sim->instance) {
+			has_src = sim->instance->has_attribute(ID::src);
+			log("  in instance %s of module %s%s\n", log_id(sim->instance),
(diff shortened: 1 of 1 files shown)


kernel/rtlil.cc
```diff
@@ -483,6 +483,58 @@ pool<string> RTLIL::AttrObject::get_strpool_attribute(const RTLIL::IdString &id)
...
+	}
+	set_string_attribute(id, attrval);
+}
+
+vector<string> RTLIL::AttrObject::get_strlist_attribute(const RTLIL::IdString &id) const
+{
+	return split_tokens(get_string_attribute(id), "|");
+}
+
+void RTLIL::AttrObject::append_src_attribute(const RTLIL::AttrObject &other, bool prepend)
```
povik, line 513: I haven't tested this helper yet. This one and the helpers above, which can be used to manipulate `src` as an ordered list, we might want to introduce in an effort to establish a defined ordered for the `src` entries.

povik: (This really doesn't have a good reason to be in this changeset, I am going to remove this.)

whitequark: >  This one and the helpers above, which can be used to manipulate src as an ordered list, we might want to introduce in an effort to establish a defined order for the src entries.

This would be my concern for merging this PR, yes. Right now src is considered a "pool" attribute, right?

povik: Yes, that’s how the codebase treats the attribute now.

We were considering to change this, and I have written those helpers, but I removed them from this PR. I made it use the usual strpool method instead.

PR #4061, opened 2023-12-09
cxxrtl: Remove redundant divmod

merryhime:
Self-explanatory.

### DIFF
--- a/backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
+++ b/backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
@@ -576,6 +576,37 @@ struct value : public expr_base<value<Bits>> {
 		result.data[result.chunks - 1] &= result.msb_mask;
 		return result;
 	}
+
+	std::pair<value<Bits>, value<Bits>> udivmod(value<Bits> divisor) const {
+		value<Bits> quotient;
+		value<Bits> dividend = *this;
+		if (dividend.ucmp(divisor))
+			return {/*quotient=*/value<Bits>{0u}, /*remainder=*/dividend};
+		uint32_t divisor_shift = dividend.ctlz() - divisor.ctlz();
+		divisor = divisor.shl(value<Bits>{divisor_shift});
+		for (size_t step = 0; step <= divisor_shift; step++) {
+			quotient = quotient.shl(value<Bits>{1u});
+			if (!dividend.ucmp(divisor)) {
+				dividend = dividend.sub(divisor);
+				quotient.set_bit(0, true);
+			}
+			divisor = divisor.shr(value<Bits>{1u});
+		}
+		return {quotient, /*remainder=*/dividend};
+	}
+
+	std::pair<value<Bits>, value<Bits>> sdivmod(const value<Bits> &other) const {
+		value<Bits + 1> quotient;
+		value<Bits + 1> remainder;
+		value<Bits + 1> dividend = sext<Bits + 1>();
+		value<Bits + 1> divisor = other.template sext<Bits + 1>();
+		if (dividend.is_neg()) dividend = dividend.neg();
+		if (divisor.is_neg()) divisor = divisor.neg();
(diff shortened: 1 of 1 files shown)


backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
```diff
@@ -585,78 +585,6 @@ struct value : public expr_base<value<Bits>> {
...
-						blk[k] = subtractBuf[k];
-					}
-				}
-			}
-		}
-
-		std::copy(blk, blk + origLen, data);
-	}
-
 	static chunk::type getShiftedBlock(const value<Bits> &num, size_t x, size_t y) {
```
whitequark: This is now unused.

backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
```diff
@@ -849,9 +763,9 @@ std::ostream &operator<<(std::ostream &os, const value_formatted<Bits> &vf)
 			if (val.is_zero())
 				buf += '0';
 			while (!val.is_zero()) {
-				value<Bits> quotient;
-				val.divideWithRemainder(value<Bits>{10u}, quotient);
-				buf += '0' + val.template trunc<(Bits > 4 ? 4 : Bits)>().val().template get<uint8_t>();
+				value<Bits> quotient, remainder;
+				std::tie(quotient, remainder) = val.template divmod_uu<Bits>(value<Bits>{10u});
```
whitequark, line 767: The dependency here is a bit in the wrong direction: `divmod_uu` (a Yosys cell implementation) should depend on something in `value<>` (the generic library), not the other way around.

I would also like to have a `divmod` method on `value<>` to use elsewhere.

whitequark (changes requested):
Thanks!

backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
merryhime: Done. Location of `value<>::divmod` is due to reference to `value<32>` and `value<1>`, thus necessitating the definition to come after `expr_base`. Open to suggestions on a better location.

backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
```diff
@@ -775,6 +691,25 @@ struct expr_base {
 	}
 };
 
+template<size_t Bits>
+std::pair<value<Bits>, value<Bits>> value<Bits>::divmod(value<Bits> divisor) const {
```
whitequark, line 695:
```suggestion
std::pair<value<Bits>, value<Bits>> value<Bits>::divmod(value<Bits> divisor) const {
	value<1> one;
	one.data[0] = 1;
```

backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
```diff
@@ -585,83 +577,7 @@ struct value : public expr_base<value<Bits>> {
...
-
-		std::copy(blk, blk + origLen, data);
-	}
-
-	static chunk::type getShiftedBlock(const value<Bits> &num, size_t x, size_t y) {
-		chunk::type part1 = (x == 0 || y == 0) ? 0 : (num.data[x - 1] >> (chunk::bits - y));
-		chunk::type part2 = (x == num.chunks) ? 0 : (num.data[x] << y);
-		return part1 | part2;
-	}
+	std::pair<value<Bits>, value<Bits>> divmod(value<Bits>) const;
```
whitequark, line 580: I think this should be `udivmod`. There's also an `sdivmod` implementation buried below (in `divmod_ss`); it would be useful to have it on `value<>` too for `cxxrtl::time`.

whitequark (approved):
Thank you!

povik:
I think this change has made the `tests/fmt/always_comb_tb.cc` test hang, making CI jobs on master stuck. I am looking into it.

povik:
Hang appears to be fixed by #4067

PR #4063, opened 2023-12-10
cxxrtl: Use the base name of the interface file for the include directive

Henkru:
This pull request addresses an issue where an incorrect include file name is generated for the `.cc` file when using the `-header` flag and the filename includes folders.

For example:
```verilog
// top.v
module top ();
endmodule
```

```bash
$ mkdir -p build && yosys -p "read_verilog top.v; write_cxxrtl -header build/top.cc"
```

The following files are created: `build/top.h` and `build/top.cc`. The `top.cc` has the following include directive to include the `top.h`:

```c++
#include "build/top.h"
```

It should instead be:

```c++
#include "top.h"
```

### DIFF
--- a/backends/cxxrtl/cxxrtl_backend.cc
+++ b/backends/cxxrtl/cxxrtl_backend.cc
@@ -630,18 +630,16 @@ std::string escape_cxx_string(const std::string &input)
 
 std::string basename(const std::string &filepath)
 {
-#ifndef _WIN32
-	const std::string dir_seps = "/";
-#else
+#ifdef _WIN32
 	const std::string dir_seps = "\\/";
+#else
+	const std::string dir_seps = "/";
 #endif
 	size_t sep_pos = filepath.find_last_of(dir_seps);
-	if (sep_pos != std::string::npos) {
+	if (sep_pos != std::string::npos)
 		return filepath.substr(sep_pos + 1);
-	}
-	else {
+	else
 		return filepath;
-	}
 }
 
 template<class T>
--- a/backends/cxxrtl/cxxrtl_backend.cc
+++ b/backends/cxxrtl/cxxrtl_backend.cc
@@ -628,6 +628,22 @@ std::string escape_cxx_string(const std::string &input)
 	return output;
 }
 
+std::string basename(const std::string &filepath)
+{
+#ifndef _WIN32
+	const std::string dir_seps = "/";
+#else
+	const std::string dir_seps = "\\/";
+#endif
+	size_t sep_pos = filepath.find_last_of(dir_seps);
+	if (sep_pos != std::string::npos && sep_pos + 1 < filepath.length()) {
+		return filepath.substr(sep_pos + 1);
+	}
+	else {
+		return filepath;
+	}
+}
+
 template<class T>
 std::string get_hdl_name(T *object)
 {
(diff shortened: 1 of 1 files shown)


backends/cxxrtl/cxxrtl_backend.cc
```diff
@@ -3577,13 +3578,25 @@ struct CxxrtlBackend : public Backend {
...
-				              worker.intf_filename.c_str(), strerror(errno));
+				              worker.intf_filepath.c_str(), strerror(errno));
 
 			worker.intf_f = &intf_f;
+
+			// get the base name of the interface file path
+#ifndef _WIN32
+			const std::string sep = "/";
+#else
+			const std::string sep = "\\";
```
whitequark, line 3593: Both `\\` and `/` are valid separators on Win32.

I suggest making a `basename` function splitting on either and then using it in the `include` directive alone.

Henkru: Thanks for the feedback. I amended the code to take that approach.

backends/cxxrtl/cxxrtl_backend.cc
```diff
@@ -628,6 +628,22 @@ std::string escape_cxx_string(const std::string &input)
...
 
+std::string basename(const std::string &filepath)
+{
+#ifndef _WIN32
+	const std::string dir_seps = "/";
+#else
+	const std::string dir_seps = "\\/";
+#endif
+	size_t sep_pos = filepath.find_last_of(dir_seps);
+	if (sep_pos != std::string::npos && sep_pos + 1 < filepath.length()) {
```
whitequark, line 639: Is the second part of the condition doing anything?

whitequark (approved):

backends/cxxrtl/cxxrtl_backend.cc
```diff
@@ -628,6 +628,22 @@ std::string escape_cxx_string(const std::string &input)
...
 
+std::string basename(const std::string &filepath)
+{
+#ifndef _WIN32
+	const std::string dir_seps = "/";
+#else
+	const std::string dir_seps = "\\/";
+#endif
+	size_t sep_pos = filepath.find_last_of(dir_seps);
+	if (sep_pos != std::string::npos) {
```
povik, line 639: Formatting nit: no curly braces when there's a single statement, also for the else clause below

backends/cxxrtl/cxxrtl_backend.cc
```diff
@@ -628,6 +628,22 @@ std::string escape_cxx_string(const std::string &input)
 	return output;
 }
 
+std::string basename(const std::string &filepath)
+{
+#ifndef _WIN32
```
povik, line 633: Nit: maybe more natural to have the conditional the other way: say what's the separator for `_WIN32`, then handle the rest

backends/cxxrtl/cxxrtl_backend.cc
Henkru: Updated

backends/cxxrtl/cxxrtl_backend.cc
Henkru: Updated

povik (approved):

Issue #4064, opened 2023-12-11
Frontend/AST: signed assign to indexed part-select


phsauter:
### Version

Yosys 0.36+13 (git sha1 2858c33f6, clang 10.0.0-4ubuntu1 -fPIC -Os)

### On which OS did this happen?

Linux

### Reproduction Steps

The zip contains the source RTL, yosys output and a testbench.
[sign_part_assign.zip](https://github.com/YosysHQ/yosys/files/13638840/sign_part_assign.zip)

Execute the commands:
```
read_verilog sign_part_assign.v; proc; clean -purge; write_verilog -noattr sign_part_assign.yosys.v;
```
Compare the output `data_o` using the testbench or compare by inspection (its only a few lines to think through).

From inspecting the debug output from `read_verilog -debug` I expect the problem to be in the AST simplification step.

### Expected Behavior

Source RTL:
```verilog
always @(*) begin
    data_ref_o[offset_i+:4] = 4'b1111;
    data_o[offset_i+:4]     = 1'sb1;
end
```
The RHS of the second expression should be sign extended to match LHS width as per Sec 5.5.3 of IEEE Std 1364-2005 (Verilog 2005).
> The following are the steps for evaluating an assignment:
— Determine the size of the right-hand side by the standard assignment size determination rules (see 5.4).
— If needed, extend the size of the right-hand side, performing sign extension if, and only if, the type of the right-hand side is signed.

### Actual Behavior

The LHS extension to match RHS is done without considering the sign bit. So it produces essentially:
```verilog
always @(*) begin
    data_ref_o[offset_i+:4] = 4'b1111;
    data_o[offset_i+:4]     = 4'b0001; // unsigned from signed 1'sb1
end
```

phsauter:
#4065 does not seem to fix this, it still produces:
```verilog
assign _01_ = { 29'h00000000, offset_i } + 32'd0;
assign _05_ = $signed({ 1'h0, _01_ }) - $signed(32'd0);
assign _03_ = - $signed(_05_);
assign data_o = $signed(_03_) < 0 ? 1'h1 << - _03_ : 1'h1 >> _03_;
```
note the last line where it is still `1'h1` (unsigned one bit).

daglem:
Here is the output which could have been generated if only a few PRs were to be merged :wink:

```
/* Generated by Yosys 0.36+8 (git sha1 74af59a70, gcc 13.2.1 -fPIC -Os) */

[...]

PR #4065, opened 2023-12-11
Respect the sign of the right operand of AST_SHIFT and AST_SHIFTX

daglem:
The $shift and $shiftx cells perform a left logical shift if the second operand is negative. This change passes the sign of the second operand of AST_SHIFT and AST_SHIFTX into $shift and $shiftx cells, respectively.

### DIFF
--- a/frontends/ast/genrtlil.cc
+++ b/frontends/ast/genrtlil.cc
@@ -1740,7 +1740,8 @@ RTLIL::SigSpec AstNode::genRTLIL(int width_hint, bool sign_hint)
 			if (width_hint < 0)
 				detectSignWidth(width_hint, sign_hint);
 			RTLIL::SigSpec left = children[0]->genRTLIL(width_hint, sign_hint);
-			RTLIL::SigSpec right = children[1]->genRTLIL();
+			// for $shift and $shiftx, the second operand can be negative
+			RTLIL::SigSpec right = children[1]->genRTLIL(-1, type == AST_SHIFT || type == AST_SHIFTX);
 			int width = width_hint > 0 ? width_hint : left.size();
 			is_signed = children[0]->is_signed;
 			return binop2rtlil(this, type_name, width, left, right);


povik:
As I understand this change it's not so much that the second operand to `AST_SHIFT`/`AST_SHIFTX` couldn't be negative otherwise, it's that it causes the second operand to be determined to be signed in more cases. In any case those AST nodes are for internal use only and we are free to tune their definition.

povik (approved):

daglem:
> As I understand this change it's not so much that the second operand to `AST_SHIFT`/`AST_SHIFTX` couldn't be negative otherwise, it's that it causes the second operand to be determined to be signed in more cases. In any case those AST nodes are for internal use only and we are free to tune their definition.

Thanks for merging!

Before this change, the second operand was converted to unsigned before being passed to the `$shift` / `$shiftx` cell. The change facilitates use of AST_SHIFTX in #3877 in favor of current complicated "fake_ast" code in `genrtlil.cc`, and I suspect that without the change, it would not be possible to use AST_SHIFT without using temporaries as is currently done in the bit slice assignment in `simplify.cc`. I'll make a PR to fix #4064 shortly (on top of #3875), using AST_SHIFT without the use of temporaries.

povik:
> Before this change, the second operand was converted to unsigned before being passed to the `$shift` / `$shiftx` cell.

Well, that depends on the operand, doesn't it? `binop2rtlil` which emits the cells themselves will use the operand's AST node's `->is_signed`, and the `is_signed` attribute is determined as a function of both the `sign_hint` (which we now flipped on), and what kind of expression that second operand was.

daglem:
> > Before this change, the second operand was converted to unsigned before being passed to the `$shift` / `$shiftx` cell.
> 
> Well, that depends on the operand, doesn't it? `binop2rtlil` which emits the cells themselves will use the operand's AST node's `->is_signed`, and the `is_signed` attribute is determined as a function of both the `sign_hint` (which we now flipped on), and what kind of expression that second operand was.

I don't think that's correct. As far as I understand, the previous call to `children[1]->genRTLIL()` would always make children[1] unsigned (`is_signed = false`), since `sign_hint` in the call to `genRTLIL()` must be `true` for the node to end up signed. Then, the sign of `children[1]` was used for the second operand to the `$shift`/ `$shiftx`, which would then also be unsigned. I verified this by looking at the generated RTLIL.

povik:
> since `sign_hint` in the call to `genRTLIL()` must be `true` for the node to end up signed.

Reading more into the code, I think you are right. At least not even `$signed()` will make the operand signed if the `sign_hint` is false. Maybe a wire could force signedness.

daglem:
> > since `sign_hint` in the call to `genRTLIL()` must be `true` for the node to end up signed.
> 
> Reading more into the code, I think you are right. At least not even `$signed()` will make the operand signed if the `sign_hint` is false. Maybe a wire could force signedness.

Yes, that's what I suspect is happening for the current use of `AST_SHIFT` in the bit slice assignment code in `simplify.cc`.

PR #4066, opened 2023-12-11
Uncloak array expressions generated by read_verilog -dump_vlog2

daglem:
Explicit conversion of AST_TO_SIGNED, AST_TO_UNSIGNED, and AST_CAST_SIZE makes it possible to reason about simplified array expressions.

### DIFF
--- a/frontends/ast/ast.cc
+++ b/frontends/ast/ast.cc
@@ -658,8 +658,17 @@ void AstNode::dumpVlog(FILE *f, std::string indent) const
 	if (0) { case AST_NEG:         txt = "-";  }
 	if (0) { case AST_LOGIC_NOT:   txt = "!";  }
 	if (0) { case AST_SELFSZ:      txt = "@selfsz@";  }
+	if (0) { case AST_TO_SIGNED:   txt = "signed'";  }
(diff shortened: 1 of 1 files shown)


frontends/ast/ast.cc
```diff
@@ -658,11 +658,20 @@ void AstNode::dumpVlog(FILE *f, std::string indent) const
 	if (0) { case AST_NEG:         txt = "-";  }
 	if (0) { case AST_LOGIC_NOT:   txt = "!";  }
 	if (0) { case AST_SELFSZ:      txt = "@selfsz@";  }
+	if (0) { case AST_TO_SIGNED:   txt = "signed'";  }
+	if (0) { case AST_TO_UNSIGNED: txt = "unsigned'";  }
```
povik, line 662: Maybe it would be best if we used `$signed` and `$unsigned` here since that's part of base Verilog. Though I am not that familiar with the standards to say if it's *exactly* the same thing as the SystemVerilog casts, nor do I know if there's something we can print out for `AST_CASE_SIZE` that's part of Verilog. Of course for a dump output what we print out here doesn't matter much.

povik (approved):

PR #4069, opened 2023-12-12
Simplify and correct array slice assignment

### LINKED ISSUE #4064: Frontend/AST: signed assign to indexed part-select

daglem:
Simplify and correct AST for array slice assignment

Corrects sign extension of the right hand side, and hopefully
makes the code simpler to understand.

Fixes #4064

### DIFF
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -2966,96 +2966,67 @@ bool AstNode::simplify(bool const_fold, int stage, int width_hint, bool sign_hin
 			}
 		} else {
 			// mask and shift operations
-
-			AstNode *wire_mask = new AstNode(AST_WIRE, new AstNode(AST_RANGE, mkconst_int(wire_width-1, true), mkconst_int(0, true)));
-			wire_mask->str = stringf("$bitselwrite$mask$%s:%d$%d", RTLIL::encode_filename(filename).c_str(), location.first_line, autoidx++);
-			wire_mask->set_attribute(ID::nosync, AstNode::mkconst_int(1, false));
-			wire_mask->is_logic = true;
-			while (wire_mask->simplify(true, 1, -1, false)) { }
-			current_ast_mod->children.push_back(wire_mask);
-
-			AstNode *wire_data = new AstNode(AST_WIRE, new AstNode(AST_RANGE, mkconst_int(wire_width-1, true), mkconst_int(0, true)));
-			wire_data->str = stringf("$bitselwrite$data$%s:%d$%d", RTLIL::encode_filename(filename).c_str(), location.first_line, autoidx++);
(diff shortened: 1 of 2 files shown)


frontends/ast/simplify.cc
```diff
@@ -2966,96 +2966,67 @@ bool AstNode::simplify(bool const_fold, int stage, int width_hint, bool sign_hin
...
+			shift_val = new AstNode(AST_NEG, shift_val);
 
-			AstNode *t;
-
-			t = mkconst_bits(std::vector<RTLIL::State>(result_width, State::S1), false);
-			t = new AstNode(AST_SHIFT, t, shamt->clone());
-			t = new AstNode(AST_ASSIGN_EQ, ref_mask->clone(), t);
-			newNode->children.push_back(t);
-
-			t = new AstNode(AST_BIT_AND, mkconst_bits(std::vector<RTLIL::State>(result_width, State::S1), false), children[1]->clone());
```
povik, line 3050: This is where the issue #4064 was, right? The rvalue is masked, but not sign-extended.

povik (approved):

povik:
> and hopefully makes the code simpler to understand.

Much better now. Thanks!

frontends/ast/simplify.cc
daglem: Correct!

PR #4070, opened 2023-12-13
cxxrtl: Add unit tests for runtime

merryhime:
Proposed testing strategy for the cxxrtl runtime.
This PR also fixes a few bugs found with randomized unit testing.

Use of `patchelf` is requried in order to override `libyosys.so` SONAME. The SONAME provides an absolute path into system library directory, which is not what we want here as we would like to test the currently built library and not the installed system library.

### DIFF
--- a/backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
+++ b/backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
@@ -510,8 +510,8 @@ struct value : public expr_base<value<Bits>> {
 		size_t count = 0;
 		for (size_t n = 0; n < chunks; n++) {
 			chunk::type x = data[chunks - 1 - n];
-			// First add to `count` as if the chunk is zero
-			count += (n == 0 ? Bits % chunk::bits : chunk::bits);
+			constexpr size_t first_chunk_bits = Bits % chunk::bits != 0 ? Bits % chunk::bits : chunk::bits;
+			count += (n == 0 ? first_chunk_bits : chunk::bits);
 			// If the chunk isn't zero, correct the `count` value and return
 			if (x != 0) {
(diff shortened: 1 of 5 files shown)


whitequark (changes requested):
I would rather not use gtest. I think the gtest stuff in `test/units` is kind of cursed and not really used, and having to use `patchelf` makes me sad. I think I would personally be happy with just a series of `assert` s

merryhime:
Revisit.

povik:
Ah, adding the gtest stuff to CI is the subject of the last commit

backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
```diff
@@ -444,12 +444,13 @@ struct value : public expr_base<value<Bits>> {
...
 		}
 		if (Signed && is_neg()) {
-			size_t top_chunk_idx  = (Bits - shift_bits) / chunk::bits;
-			size_t top_chunk_bits = (Bits - shift_bits) % chunk::bits;
+			size_t top_chunk_idx  = amount.data[0] > Bits ? 0 : (Bits - amount.data[0]) / chunk::bits;
+			size_t top_chunk_bits = amount.data[0] > Bits ? 0 : (Bits - amount.data[0]) % chunk::bits;
 			for (size_t n = top_chunk_idx + 1; n < chunks; n++)
 				result.data[n] = chunk::mask;
-			if (shift_bits != 0)
+			if (amount.data[0] != 0)
```
povik, line 451: Maybe we could make this check `top_chunk_idx < chunks`

povik:
I reviewed the fixes, they all look good, I will cherry-pick them into a PR of their own, so that they are not blocked on resolving what the testing should be.

PR #4072, opened 2023-12-13
cxxrtl: Add simple tests for cxxrtl::value from cxxrtl runtime

merryhime:
Superseeds #4070.


povik (approved):
I approve this as far as the fixes to the value operations themselves go.

whitequark (approved):
Thanks a lot! I'm quite happy with this PR.

PR #4073, opened 2023-12-13
cxxrtl: Fixes to shift operations, `ctlz`

povik:
This is split off from #4070 only containing @merryhime's fixes to some of the value operations. I reviewed the changes and they look good.


povik:
Closing in favor of #4072

PR #4084, opened 2023-12-19
$scopeinfo support

jix:
This PR adds `$scopeinfo` cells to preserve information about the hierarchy during flattening, in particular attributes including `src`.

This is still missing a kernel header containing utilities for easily querying `$scopeinfo` data.

### DIFF
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -1769,6 +1769,15 @@ namespace {
 				return;
 			}
 
+			if (cell->type == ID($scopeinfo)) {
+				param(ID::TYPE);
+				check_expected();
+				std::string scope_type = cell->getParam(ID::TYPE).decode_string();
+				if (scope_type != "module" && scope_type != "struct")
+					error(__LINE__);
+				return;
+			}
+
 			if (cell->type == ID($_BUF_))    { port(ID::A,1); port(ID::Y,1); check_expected(); return; }
 			if (cell->type == ID($_NOT_))    { port(ID::A,1); port(ID::Y,1); check_expected(); return; }
 			if (cell->type == ID($_AND_))    { port(ID::A,1); port(ID::B,1); port(ID::Y,1); check_expected(); return; }
--- a/passes/techmap/flatten.cc
+++ b/passes/techmap/flatten.cc
@@ -46,24 +46,6 @@ IdString map_name(RTLIL::Cell *cell, T *object)
 	return cell->module->uniquify(concat_name(cell, object->name));
 }
 
-template<class T>
-void map_attributes(RTLIL::Cell *cell, T *object, IdString orig_object_name)
-{
-	if (object->has_attribute(ID::src))
-		object->add_strpool_attribute(ID::src, cell->get_strpool_attribute(ID::src));
-
-	// Preserve original names via the hdlname attribute, but only for objects with a fully public name.
-	if (cell->name[0] == '\\' && (object->has_attribute(ID::hdlname) || orig_object_name[0] == '\\')) {
-		std::vector<std::string> hierarchy;
-		if (object->has_attribute(ID::hdlname))
-			hierarchy = object->get_hdlname_attribute();
-		else
-			hierarchy.push_back(orig_object_name.str().substr(1));
-		hierarchy.insert(hierarchy.begin(), cell->name.str().substr(1));
-		object->set_hdlname_attribute(hierarchy);
-	}
-}
-
 void map_sigspec(const dict<RTLIL::Wire*, RTLIL::Wire*> &map, RTLIL::SigSpec &sig, RTLIL::Module *into = nullptr)
 {
 	vector<SigChunk> chunks = sig;
@@ -76,6 +58,54 @@ void map_sigspec(const dict<RTLIL::Wire*, RTLIL::Wire*> &map, RTLIL::SigSpec &si
 struct FlattenWorker
 {
 	bool ignore_wb = false;
+	bool create_scopeinfo = true;
+	bool create_scopename = false;
+
+	template<class T>
+	void map_attributes(RTLIL::Cell *cell, T *object, IdString orig_object_name)
+	{
+		if (!create_scopeinfo && object->has_attribute(ID::src))
+			object->add_strpool_attribute(ID::src, cell->get_strpool_attribute(ID::src));
+
+		// Preserve original names via the hdlname attribute, but only for objects with a fully public name.
+		// If the '-scopename' option is used, also preserve the containing scope of private objects if their scope is fully public.
+		if (cell->name[0] == '\\') {
+			if (object->has_attribute(ID::hdlname) || orig_object_name[0] == '\\') {
+				std::string new_hdlname;
+
+				if (cell->has_attribute(ID::hdlname)) {
+					new_hdlname = cell->get_string_attribute(ID(hdlname));
+				} else {
+					log_assert(!cell->name.empty());
+					new_hdlname = cell->name.c_str() + 1;
+				}
+				new_hdlname += ' ';
+
+				if (object->has_attribute(ID::hdlname)) {
+					new_hdlname += object->get_string_attribute(ID(hdlname));
+				} else {
+					log_assert(!orig_object_name.empty());
+					new_hdlname += orig_object_name.c_str() + 1;
+				}
+				object->set_string_attribute(ID(hdlname), new_hdlname);
+			} else if (object->has_attribute(ID(scopename))) {
+				std::string new_scopename;
+
+				if (cell->has_attribute(ID::hdlname)) {
+					new_scopename = cell->get_string_attribute(ID(hdlname));
+				} else {
+					log_assert(!cell->name.empty());
+					new_scopename = cell->name.c_str() + 1;
+				}
+				new_scopename += ' ';
+				new_scopename += object->get_string_attribute(ID(scopename));
+				object->set_string_attribute(ID(scopename), new_scopename);
+			} else if (create_scopename) {
+				log_assert(!cell->name.empty());
+				object->set_string_attribute(ID(scopename), cell->name.c_str() + 1);
+			}
+		}
+	}
 
 	void flatten_cell(RTLIL::Design *design, RTLIL::Module *module, RTLIL::Cell *cell, RTLIL::Module *tpl, SigMap &sigmap, std::vector<RTLIL::Cell*> &new_cells)
 	{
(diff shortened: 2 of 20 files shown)


passes/techmap/flatten.cc
```diff
@@ -275,6 +295,14 @@ struct FlattenPass : public Pass {
 		log("    -wb\n");
 		log("        Ignore the 'whitebox' attribute on cell implementations.\n");
 		log("\n");
+		log("    -noscopeinfo\n");
+		log("        Do not create '$scopeinfo' cells that preserve attributes of cells and\n");
+		log("        modules that were removed during flattening. With this option the\n");
+		log("        'src' attributes is handled like it was before '$scopeinfo' was\n");
```
povik, line 301: Nit: typos here, either "attribute is handled" or "attributes are handled", "separated"

passes/techmap/flatten.cc
```diff
@@ -275,6 +295,14 @@ struct FlattenPass : public Pass {
...
+		log("        introduced, i.e. multiple 'src' locations are combined separted by\n");
+		log("        whitespace. Without this option these 'src' locations can be found\n");
+		log("        via the 'cell_src' and 'module_src' attribute of '$scopeinfo' cells.\n");
```
povik, line 304: I would suggest not referring to how this was handled in the past, just explain source locations would be collected from different points on the hierarchy and combined without distinction on the cell post flattening.

kernel/rtlil.cc
```diff
@@ -1747,6 +1747,12 @@ namespace {
 				return;
 			}
 
+			if (cell->type == ID($scopeinfo)) {
+				param(ID::TYPE);
```
povik, line 1751: I wonder if the type shouldn't be an attribute as well, but I don't have much of an argument in favor of it right now.

whitequark: I would like to not have to worry about invalid types in the debugger frontend, meaning it should be verified in `check()`. That requires it to be a parameter.

passes/techmap/flatten.cc
```diff
@@ -220,7 +221,26 @@ struct FlattenWorker
...
 		}
 
+		RTLIL::Cell *scopeinfo = nullptr;
+		RTLIL::IdString cell_name = cell->name;
+
+		if (create_scopeinfo)
+		{
+			// The $scopeinfo's name will be changed below after removing the flattened cell
+			scopeinfo = module->addCell(NEW_ID, ID($scopeinfo));
+			scopeinfo->setParam(ID::TYPE, RTLIL::Const("cell"));
```
povik, line 231: You have probably thought this out already and I am curious to hear it. Should this be `cell` or `module`?

povik: (Also sorry for bikeshedding but the only other occurences of enums on cell parameters within Yosys I have seen were on some vendor cells, where the values were fully uppercase. Also e.g. in select syntax we would have the parameter name and its value next to each other. For reasons of consistence we could say the values for an enum-style parameter should be uppercase.)

povik (approved):
Looks good to me

passes/techmap/flatten.cc
whitequark: > You have probably thought this out already and I am curious to hear it. Should this be cell or module?

It refers to a module instantiation (a `module` was written by the designer at some point) so I would say "module" and not "cell".

> (Also sorry for bikeshedding but the only other occurences of enums on cell parameters within Yosys I have seen were on some vendor cells, where the values were fully uppercase. Also e.g. in select syntax we would have the parameter name and its value next to each other. For reasons of consistence we could say the values for an enum-style parameter should be uppercase.)

I think it's consistent enough with lowercase. In Verilog and RTLIL both the module is spelled as `module`, and when extending this I think we might have other scopes such as `begin` or maybe `struct`. We aren't writing ALGOL to have `BEGIN` :)

povik: > I think it's consistent enough with lowercase. In Verilog and RTLIL both the module is spelled as module, and when extending this I think we might have other scopes such as begin or maybe struct. We aren't writing ALGOL to have BEGIN :)

No contest

whitequark (approved):

kernel/rtlil.cc
jix: If I'm not mistaken, we're so far not using attributes for anything that's unconditionally required to be present on a cell, which was why I agreed with @whitequark's suggestion on slack to use a parameter here.

I'm not sure if we should limit the available scope types to a fixed set, leave it open to allow arbitrary user defined scope types, or a hybrid where e.g. user defined scope types share a prefix identifying them as such (similar to private and public names).

In the current PR I only check for presence of the parameter, as I wanted to discuss this first.

whitequark: The scope types are a contract between the compiler and the debugger. The debugger must understand every scope type to properly present them in the interface, so the list of scope types should be closed (with potential for custom ones). This is how DWARF works for example.

passes/techmap/flatten.cc
jix: I couldn't decide between "cell" and "module" as which one is more appropriate depends on whether you want to indicate what caused the presence of the scope i.e. the entity that has the same hierarchical name as the `$scopeinfo` cell, in which case it would be "cell" (or "instance" if we're using verilog not RTLIL terminology). If we're indicating what defines the contents of the scope then "module" is clearly the correct choice.

Since both approaches make sense to me and changing it is trivial at this point, I went with an arbitrary choice.

Given the feedback so far, I'd change this to "module".

whitequark: I am not opposed to `instance`, though I feel it might be a little too ambiguous.

kernel/rtlil.cc
jix: Do you already have a list of scope types besides "module" that should be initially allowed?

I'm leaning towards having the list of scope types be completely closed initially, i.e. deferring how we handle custom scope types until an actual need for them arises. I can think of several ways, but wouldn't be able to make a good decision without a concrete use case in mind.

whitequark: I would say `module` and `struct`.

passes/techmap/flatten.cc
nakengelhardt: (Also FYI strpool uses `|` not whitespace as separator).

nakengelhardt:
Currently `opt_merge` considers all `$scopeinfo` cells identical and will remove all but one of them. Should we special-case `$scopeinfo` there too, or should we perhaps have a global ignorelist in yosys_celltypes that all optimization passes can check against so as to not touch those cells?

whitequark:
> or should we perhaps have a global ignorelist in yosys_celltypes that all optimization passes can check against so as to not touch those cells?

I do want `opt_merge` to merge the `$debugprobe` cells (using a special treatment) so this would not benefit that effort.

povik:
> Currently opt_merge considers all $scopeinfo cells identical and will remove all but one of them. Should we special-case $scopeinfo there too, or should we perhaps have a global ignorelist in yosys_celltypes that all optimization passes can check against so as to not touch those cells?

Huh, I thought there was some logic to consider cells equal only if the attributes match (except for src I suppose). Or was that only for wires?

passes/techmap/flatten.cc
jix: Right, I must have been thinking of `hdlname`, I'll update the description.

jix:
> or should we perhaps have a global ignorelist in yosys_celltypes that all optimization passes can check against so as to not touch those cells?

I think that makes sense if we think that list will grow or can already start out containing other cells too. If it's only going to be `$scopeinfo` I'd say we add code specific for that and can move to an ignore list later when there are more cells that should be treated the same way.

jix:
I'm currently working on getting this PR ready (as wall as a followup PR with code to build a $scopeinfo aware hierarchical index for use in yosys passes) and I'm not sure how the various backend passes should treat `$scopeinfo`. In particular backend passes that are used to transfer synthesis results to other tools.

I'm assuming that the current state of this PR is likely to break existing synthesis setups with other tools potentially failing on unexpected $scopeinfo cells. I can also imagine the $scopeinfo data could eventually be of use for some of the tools that would currently fail due to its presence.

Since I use yosys almost exclusively for FV, I could use some input on how to handle this, though.

whitequark:
Personally, I would be OK with every backend except for CXXRTL and JSON ignoring `$scopeinfo` entirely, since it is a Yosys internal cell that will definitely break such other tools. I don't know, however, how to deal with people who do want e.g. `$scopeinfo` in their structural Verilog output.

Maybe we could handle that when we encounter it?

jix:
That also makes sense to me, but for JSON there is the issue that on one hand the JSON backend is intended to include all of the RTLIL netlist but also that nextpnr which reads that JSON output will fail with `ERROR: Unable to place cell 'soc.cpu', no BELs remaining to implement cell type '$scopeinfo'` with this PR.

A way to work around this would be to have the corresponding synth passes explicitly delete `$scopeinfo` cells when using the `-json` option until nextpnr at least knows to ignore those cells. This would still break `synth_…; write_json output.json` when used with nextpnr, but in that case you can manually insert a `delete */t:$scopeinfo` between `synth_…` and `write_json` and it seems like all the nextpnr documentation and examples use `-json` anyway.

Does that sound acceptable?

Alternatives would be to get nextpnr to ignore these cells before merging this PR or to ignore `$scopeinfo` by default also for the JSON backend.

whitequark:
> Alternatives would be to get nextpnr to ignore these cells before merging this PR

I think we can do this.

gatecat:
I am happy to add this if we really want to avoid removing these cells in Yosys - however, despite the lack of any firm compatibility guarantees between Yosys and nextpnr, I do try hard to avoid situations like this (not least because they make bisecting more annoying) and it's been some time since we last had to introduce one.

whitequark:
I think we can just not emit the `$scopeinfo` cells anywhere but in CXXRTL and RTLIL then.

povik:
Now that I think of it, we want `$scopeinfo` as part of the JSON output for more informed logging on part of nextpnr! E.g. when explaining the critical path.

povik:
Well, at least in the future, when the synthesis flow will stop losing source attribution because of a roundtrip through ABC...

whitequark:
> Now that I think of it, we want `$scopeinfo` as part of the JSON output for more informed logging on part of nextpnr! E.g. when explaining the critical path.

Yes, which is why I proposed to have it in JSON initially.

povik:
> We will have the exact same problem with $probe cells, so the solution for $scopeinfo should be extensible to that cell at least.
>
> Similarly with any $check cells that end up in a synthesis flow. (It might be meaningful to synthesize those--eventually!)

As I see it, `$probe` and `$check` can simply be stripped at some point (as would `$print`), this is brought up in #4122, but `$scopeinfo` we want to get to the P&R tool for logging.

whitequark:
Right, even if I do want `$print` in hardware I don't want nextpnr to have the logic for it.

gatecat:
Either way, I'll update nextpnr to ignore `$scopeinfo` because that is never going to hurt (and in the future, extend the frontend to populate nextpnr's own hierarchy structures with the info from it, although these haven't been touched for a while and might need some changes too).

jix:
I think `$probe` and `$scopeinfo` are different from `$check` in that they would start appearing without any changes in the input source, so failing when they are present breaks existing flows with existing inputs whereas when anything that doesn't already support `$assert` fails on `$check` that would not be a regression. (And all the supported existing FV use cases would automatically support `$check` when `clk2fflogic` and `async2sync` do.) For this PR I'm more concerned about regressions, and think improvements on how `$assert` or `$check` are handled can be done separately.

My current plan for `$scopeinfo` would be: I add checks to ignore it to all backends besides RTLIL, CXXRTL and JSON. For now I'd explicitly check for `$scopeinfo`, but it will be easy to extend it with `$probe` or to eventually move it to some better cell-classification infrastructure should that appear as part of #4122 or otherwise. IMO the important part for now is identifying all the places where such checks are needed. Changing the exact check used later on when required for new cells or to support new use cases is not a big deal, IMO.

I don't have any opinion, though, on whether to add something temporary and/or optional to the JSON backend and/or to the relevant synth passes to ignore `$scopeinfo` for now, to keep compatibility with older nextpnr versions. I also don't think I'm in the best position to make a decision on that, but I can implement whatever consensus there is as part of this PR. (But I'm not sure there is one at this point?)

gatecat:
My suggestion is:

 - in the short term, we strip `$scopeinfo` from the JSON
 - in a couple of months, once there's been a nextpnr release that's at least able to ignore it, and support in nextpnr that actually makes use of it, we can consider to start outputting it.

whitequark:
> My suggestion is:
> 
> * in the short term, we strip `$scopeinfo` from the JSON
> * in a couple of months, once there's been a nextpnr release that's at least able to ignore it, and support in nextpnr that actually makes use of it, we can consider to start outputting it.

SGTM

nakengelhardt (approved):

whitequark:
I've been looking into integrating this with CXXRTL. I've actually hit a small amount of trouble since I'm not entirely sure how to expose both module and cell attributes, but this is CXXRTL-side. I will probably be done with it by the end of next week, but I do not mind having this PR merged before then.

PR #4087, opened 2023-12-20
lattice: Fix mapping onto DP8KC for data width 1 or 2

povik:
This appears to be the right wiring of `DP8KC` per the MachXO3 simulation model I have at hand.

Some points:

 * I don't know if this applies to all `DP8KC`primitives across the Lattice families.

 * Reading into the simulation model, it seems one cannot map a memory with write port of width 18 and read port of width 2 onto `DP8KC`. This is because `DIB1`, which would be used for the 10th data bit input, gets ignored. If this is true, or even if it's just a bug of the simulation model, we may want to disable this configuration in libmap inference of `$__PDPW8KC_`. **UPDATE:** done in the second commit

### DIFF
--- a/techlibs/lattice/brams_8kc.txt
+++ b/techlibs/lattice/brams_8kc.txt
@@ -32,6 +32,10 @@ ram block $__PDPW8KC_ {
 	cost 64;
 	init no_undef;
 	port sr "R" {
+		# width 2 cannot be supported because of quirks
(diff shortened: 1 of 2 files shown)


mmicko (approved):
LGTM, also regarding open points, we are sure this same primitive is used on all XO2/3/3D models, and their simulation models and mappings are same.

PR #4089, opened 2023-12-21
Automatically call `hierarchy` after reading command line inputs

whitequark:
Literally nobody who has ever used Yosys, in the entire history of Yosys, has wanted to see a `$abstract\foo` anywhere in the netlist. I have been encountering this ussue for several years now. Enough.


nakengelhardt:
I disagree with this change, it would make it impossible to load certain files as command line arguments, so you couldn't write scripts that work on any valid input without knowing the file name any more.

There are many passes that do not work correctly until you run `hierarchy -top foo`, just like there are many passes that don't work until you run `proc`. Introducing more workarounds that make it sometimes ok to skip at the expense of breaking certain inputs doesn't improve the situation IMO, the solution here is to make it clear that users should always run `hierarchy` and `proc` immediately unless they know why they are not doing it. (The long-term solution is to make yosys aware of the different design states and the guarantees associated with them, and we have plans in this regard, hopefully there will be some movement on those next year.)

whitequark:
> There are many passes that do not work correctly until you run `hierarchy -top foo`, just like there are many passes that don't work until you run `proc`.

Are there *any* passes that work correctly while there are `$abstract` modules in the design?

The more general problem here is that the change that delayed module instantiation until `hierarchy` has seriously regressed Yosys CLI's usability with Verilog inputs. Do you not care about that?

If this change is rejected I will probably look into updating the Verilog frontend to not create `$abstract` modules if there are no parameters in the AST. I think the current situation is unacceptable as the impact on the CLI's usability is too high.

whitequark:
> I disagree with this change, it would make it impossible to load certain files as command line arguments, so you couldn't write scripts that work on any valid input without knowing the file name any more.

Having thought about the reasoning here I agree that the change I proposed is wrong.

> The long-term solution is to make yosys aware of the different design states and the guarantees associated with them, and we have plans in this regard, hopefully there will be some movement on those next year.

I would be disappointed to have to wait a year or more for something so basic. This pull request now implements awareness of the design state where there are abstract modules in the design, and automatically runs `hierarchy` in case the pass does not support it.

In principle the same could be done for `proc` but I have not made that change for two reasons:
- The requirement to run `proc` has been in Yosys since the first release, so it had not break anyone's scripts or muscle memory.
- There is existing functionality to handle the existence of processes, and the change would be a lot more invasive, with a much smaller benefit.

I do think that something similar should be happening for processes (and `write_cxxrtl` does work that way) but I also think the tradeoff is not right.

whitequark:
> Are there any passes that work correctly while there are $abstract modules in the design?

(This got an answer in form of the `abstract_modules_ok = true;` statements added in the final patch.)

PR #4090, opened 2023-12-21
write_verilog: emit `casez` as `if/elif/else` whenever possible

whitequark:
This results in a very significant improvement in readability when emitting Verilog from Amaranth code, and a slight improvement when emitting Verilog from Yosys-parsed Verilog code. Not sure about Verific.

### DIFF
--- a/backends/verilog/verilog_backend.cc
+++ b/backends/verilog/verilog_backend.cc
@@ -376,7 +376,7 @@ void dump_sigspec(std::ostream &f, const RTLIL::SigSpec &sig)
 	}
 }
 
-void dump_attributes(std::ostream &f, std::string indent, dict<RTLIL::IdString, RTLIL::Const> &attributes, char term = '\n', bool modattr = false, bool regattr = false, bool as_comment = false)
+void dump_attributes(std::ostream &f, std::string indent, dict<RTLIL::IdString, RTLIL::Const> &attributes, std::string term = "\n", bool modattr = false, bool regattr = false, bool as_comment = false)
 {
 	if (noattr)
 		return;
@@ -392,13 +392,13 @@ void dump_attributes(std::ostream &f, std::string indent, dict<RTLIL::IdString, 
 			f << stringf(" 1 ");
 		else
 			dump_const(f, it->second, -1, 0, false, as_comment);
-		f << stringf(" %s%c", as_comment ? "*/" : "*)", term);
+		f << stringf(" %s%s", as_comment ? "*/" : "*)", term.c_str());
 	}
 }
 
 void dump_wire(std::ostream &f, std::string indent, RTLIL::Wire *wire)
 {
-	dump_attributes(f, indent, wire->attributes, '\n', /*modattr=*/false, /*regattr=*/reg_wires.count(wire->name));
+	dump_attributes(f, indent, wire->attributes, "\n", /*modattr=*/false, /*regattr=*/reg_wires.count(wire->name));
 #if 0
 	if (wire->port_input && !wire->port_output)
 		f << stringf("%s" "input %s", indent.c_str(), reg_wires.count(wire->name) ? "reg " : "");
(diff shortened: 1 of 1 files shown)


povik:
Before
```
    (* src = "test.v:5.3-8.14" *)
    casez (a)
      /* src = "test.v:5.7-5.8" */
      1'h1:
          _0_ = 8'h55;
      /* src = "test.v:7.3-7.7" */
      default:
          _0_ = 8'haa;
    endcase
```
After
```
    (* src = "test.v:5.7-5.8" *)
    if (a)
      _0_ = 8'h55;
    else
      _0_ = 8'haa;
```

backends/verilog/verilog_backend.cc
```diff
@@ -1983,14 +1983,64 @@ void dump_proc_switch(std::ostream &f, std::string indent, RTLIL::SwitchRule *sw
...
+						f << "\n" << indent;
+					dump_attributes(f, "", (*it)->attributes, "\n" + indent);
+				}
+				f << stringf("if (");
+				dump_sigspec(f, *sig_it);
+				f << stringf(")\n");
+			}
+			dump_case_body(f, indent, *it);
+			if ((*it)->compare.empty())
+			  break;
```
povik, line 2031: Nit: whitespace

backends/verilog/verilog_backend.cc
```diff
@@ -1983,14 +1983,64 @@ void dump_proc_switch(std::ostream &f, std::string indent, RTLIL::SwitchRule *sw
...
+			}
+			if (compare[case_index] != State::S1)
+				is_cond = false;
+			for (int bit_index = 0; bit_index < compare.size(); bit_index++)
+				if (bit_index != case_index && compare[bit_index] != State::Sa)
+					is_cond = false;
+		} else is_cond = false;
+	}
+
+	if (is_cond) {
```
povik, line 2005: Nit: I would prefer splitting this somehow. E.g. have a `dump_proc_switch_ifstring` returning false if the process cannot be handled the `if/else` way, in which case we would carry on with the `casez` emission in `dump_proc_switch`

povik (approved):

whitequark:
@povik All fixed, thank you

povik:
Say you have an
```
    casez (sel)
      2'b1?: ...
      2'b?1: ...
      default: ...
    endcase
```
and `sel[0]` can have an x-bit, does what we emit with this change have different semantics?

povik:
To illustrate the change better, this takes
```
module top(a, b, c, out);
	input wire a, b, c;
	output reg [3:0] out;
	always @(*) begin
		casez ({c, b, a})
			3'b??1: out = 0;
			3'b?1?: out = 1;
			3'b1??: out = 2;
			default: out = 3;
		endcase
	end
endmodule
```
and emits
```
  always @* begin
    if (\$auto$verilog_backend.cc:2239:dump_module$2 ) begin end
    _0_ = _1_;
    (* src = "test.v:0.0-0.0" *)
    if (a)
      _1_ = 4'h0;
    else 
    (* src = "test.v:0.0-0.0" *)
    if (b)
      _1_ = 4'h1;
    else 
    (* src = "test.v:0.0-0.0" *)
    if (c)
      _1_ = 4'h2;
    else
      _1_ = 4'h3;
  end
  always @* begin
      out <= _0_;
  end
```

whitequark:
> does what we emit with this change have different semantics?

I'm going to counter this with: Do we care? Is `'x` in this context a part of a contract we state anywhere or are expected to uphold?

povik:
> Do we care? Is 'x in this context a part of a contract we state anywhere or are expected to uphold?

That's exactly what we were discussing on the dev JF. (I used the comments here as a scratchpad for our discussion.) Anyway I think we agreed on merging this, only please rebase this since there's conflict with #4111.

whitequark:
> That's exactly what we were discussing on the dev JF.

Thanks for dealing with this! I was too sick to attend today's dev JF but I will be present on the next one.

I'll rebase.

jix:
More specifically, AFAICT the SV spec can have different 'x behavior for both variants, but they agree on all synthesizable cases (i.e. excluding behavior that's specified for 'x but not a sound approximation of the '0 and '1 cases, thus requiring the synthesis incompatible ability to tell non-constant 'x apart from '0 or '1).

whitequark:
Updated. GitHub seems to be having trouble right now, with lots of 500 returned.

whitequark:
Updated. GitHub seems to be having trouble right now, with lots of 500 returned.

PR #4091, opened 2023-12-21
meta: Fix ASAN Static Order Initialization Errors

QuantamHD:
Changes static init variables into functions to ensure their correct initalization order.

Also removes destruct guard by ensuring their construction occurs before any IdString.

### DIFF
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -30,13 +30,18 @@
 
 YOSYS_NAMESPACE_BEGIN
 
-bool RTLIL::IdString::destruct_guard_ok = false;
-RTLIL::IdString::destruct_guard_t RTLIL::IdString::destruct_guard;
-std::vector<char*> RTLIL::IdString::global_id_storage_;
-dict<char*, int, hash_cstr_ops> RTLIL::IdString::global_id_index_;
 #ifndef YOSYS_NO_IDS_REFCNT
-std::vector<int> RTLIL::IdString::global_refcount_storage_;
-std::vector<int> RTLIL::IdString::global_free_idx_list_;
+std::vector<int> &RTLIL::IdString::global_refcount_storage()
+{
+	static std::vector<int> *global_refcount_storage = new std::vector<int>();
+	return *global_refcount_storage;
+}
+
+std::vector<int> &RTLIL::IdString::global_free_idx_list()
+{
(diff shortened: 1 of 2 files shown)


QuantamHD:
image
image

clairexen:
The std::vector default constructor is constexpr (since C++20 I think), and we should probably make the hashlib default constructors constexpr as well..

QuantamHD:
My understanding was that Yosys is supposed to compile with c++11. Are you proposing making that requirement c++20?

QuantamHD:
I'm okay either way, but I do need to get this one resolved sooner rather than later.

clairexen:
I would assume that the code generated by virtually every optimizing compiler would have static initialization for empty std::vector, even in c++11 mode.

What I'm proposing is to build with -std=c++20 for ASAN. Can you confirm that the ASAN issues for the std::vector cases are resolved when building with -std=c++20 (and using a C++20 std library)?

QuantamHD:
The ASAN issue was found in c++20 mode.

clairexen:
I can't reproduce this. I've built with `SANITIZER=address` and used the following command line

    ASAN_OPTIONS="check_initialization_order=true:print_stats=true:atexit=true" ./yosys -p 'read_verilog tests/simple/fiedler-cooley.v; synth -auto-top'

With this I got no ASAN errors at all. This is with

    Yosys 0.36+58 (git sha1 ea7818d31, clang 18.0.0 -fPIC -Os -O1 -fno-omit-frame-pointer -fno-optimize-sibling-calls -fsanitize=address)

I've also created this small demo to help discussing the meaning of that asan check wrt std::vector and relevant compiler options:
https://github.com/clairexen/handicraft/tree/handicraft-2023/asan-initfiasc-demo

clairexen:
With https://github.com/YosysHQ/yosys/pull/4097 I get no errors for `ASAN_OPTIONS=strict_init_order=true`, no `constinit` required. This is the `Makefile.conf` that I used:

```
SANITIZER = address
CXXSTD = c++20
LDFLAGS += -stdlib=libc++
CXXFLAGS += -stdlib=libc++
```

clairexen:
Merged #4097

QuantamHD:
Thanks @clairexen! Sorry for the late reply. Looks like this should work

Issue #2060, opened 2020-05-15
Follow the XDG Base Directory Specification


piegamesde:
## Steps to reproduce the issue

Yosys should follow the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).

## Expected behavior

Please don't do that. Put it into `$XDG_DATA_HOME/yosys/` or `$XDG_CACHE_HOME/yosys/` instead.

## Actual behavior

I just found a `.yosys_history` file while cleaning up my home directory.

whitequark:
Yosys isn't an X11 or a desktop application.

Although in some cases (like Rust's [Cargo tool](https://github.com/rust-lang/cargo/issues/1734)) you can argue that the existing `~/.cargo` directory benefits from being split into configuration and cache directories so that existing tools don't e.g. unnecessarily backup the cache, Yosys isn't one of such cases: it only writes a history file.

piegamesde:
> Yosys isn't an X11 or a desktop application.

The XDG specifications grew out of their name a while ago and can be applied to all kinds of software, not only graphical/desktop/Xorg applications.

> it only writes a history file

That's great, because it will be easy to change. As software grows and gets older, more of those files  start appearing and migrating a big code base to the XDG specification afterwards is way harder than growing when appropriate code is already there. (See the Gradle folks, the `~/.gradle` is a few gigabytes of a mess and I doubt they'll ever make the jump.)

So yes, there may be less urge to do so for Yosys compared to other applications, but don't think it would hurt and there are still enough benefits to it to make it worth the effort.

whitequark:
What are the actual benefits in case of Yosys, besides "my `$HOME` is cleaner" which is entirely subjective, and "it'll cause more problems later" which is a slippery slope fallacy?

piegamesde:
Having all application data at only three locations makes it easier to manage. For example, it is more convenient to clean up files of removed software when it's not hidden between user data.

But yes, to me "my $HOME is cleaner" is the biggest point and all my arguments more or less boil down to this.

whitequark:
> The XDG specifications grew out of their name a while ago and can be applied to all kinds of software, not only graphical/desktop/Xorg applications.

I wasn't aware of the full breadth of software that uses XDG Base Directories these days, including `ps`, `nano`, `htop`, `hg` and so on. So I agree that with that context, it makes a lot of sense for Yosys to support them as well.

PR #4093, opened 2023-12-21
Follow the XDG Base Directory Specification

hakan-demirli:
Closes https://github.com/YosysHQ/yosys/issues/2060

### DIFF
--- a/kernel/driver.cc
+++ b/kernel/driver.cc
@@ -244,7 +244,11 @@ int main(int argc, char **argv)
 	bool mode_q = false;
 
 #if defined(YOSYS_ENABLE_READLINE) || defined(YOSYS_ENABLE_EDITLINE)
-	if (getenv("HOME") != NULL) {
+	if (getenv("XDG_DATA_HOME") != NULL && getenv("XDG_DATA_HOME")[0] != '\0') {
+		yosys_history_file = stringf("%s/yosys/.yosys_history", getenv("XDG_DATA_HOME"));
+		read_history(yosys_history_file.c_str());
+		yosys_history_offset = where_history();
+	} else if (getenv("HOME") != NULL) {
 		yosys_history_file = stringf("%s/.yosys_history", getenv("HOME"));
 		read_history(yosys_history_file.c_str());
 		yosys_history_offset = where_history();
--- a/kernel/driver.cc
+++ b/kernel/driver.cc
@@ -244,8 +244,10 @@ int main(int argc, char **argv)
 	bool mode_q = false;
 
 #if defined(YOSYS_ENABLE_READLINE) || defined(YOSYS_ENABLE_EDITLINE)
-	if (getenv("XDG_STATE_HOME") != NULL && getenv("XDG_STATE_HOME")[0] != '\0') {
-		yosys_history_file = stringf("%s/.local/state/.yosys_history", getenv("HOME"));
+	if (getenv("XDG_STATE_HOME") == NULL || getenv("XDG_STATE_HOME")[0] == '\0') {
+		if (getenv("HOME") != NULL) {
+			yosys_history_file = stringf("%s/.local/state/.yosys_history", getenv("HOME"));
+		}
 	} else {
 		yosys_history_file = stringf("%s/.yosys_history", getenv("XDG_STATE_HOME"));
 	}
--- a/kernel/driver.cc
+++ b/kernel/driver.cc
@@ -247,12 +247,14 @@ int main(int argc, char **argv)
 	if (getenv("XDG_STATE_HOME") == NULL || getenv("XDG_STATE_HOME")[0] == '\0') {
 		if (getenv("HOME") != NULL) {
 			yosys_history_file = stringf("%s/.local/state/.yosys_history", getenv("HOME"));
+			read_history(yosys_history_file.c_str());
+			yosys_history_offset = where_history();
 		}
 	} else {
 		yosys_history_file = stringf("%s/.yosys_history", getenv("XDG_STATE_HOME"));
+		read_history(yosys_history_file.c_str());
+		yosys_history_offset = where_history();
 	}
-	read_history(yosys_history_file.c_str());
-	yosys_history_offset = where_history();
 #endif
 
 	if (argc == 2 && (!strcmp(argv[1], "-h") || !strcmp(argv[1], "-help") || !strcmp(argv[1], "--help")))
--- a/kernel/driver.cc
+++ b/kernel/driver.cc
@@ -53,6 +53,78 @@
 #  include <unistd.h>
 #endif
 
+#include <sys/stat.h> // _stat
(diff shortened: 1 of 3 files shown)


whitequark (changes requested):
I think this isn't an accurate implementation.

1. The XDG base directories specification does not require `XDG_DATA_HOME` to be set and it is usually not set. So this would do nothing most of the time.
2. I think `XDG_DATA_HOME` isn't the right place to put `.yosys_history`. If you install Yosys with a prefix of `~/.local` (which is a reasonable thing to do if you don't want to put it into `/usr/local` for some reason), it puts the files that usually go to `/usr/share` there, and history shouldn't be mixed with those. I think either `XDG_CACHE_HOME` or `XDG_CONFIG_HOME` should be used for this. (I did a quick search and on my system history files are placed in all three of data, config, and cache.)

hakan-demirli:
Yea, XDG_CACHE_HOME is more suitable for the history file. 
As for setting the XDG variables I didn't want to break things by changing the location for the users who are not explicitly using XDG. Even though it is not spec compliant.

Is it OK to just obey the spec? Or should I create an extra env variable like other projects do: YOSYS_HISTORY_FILE. Since, users can set it to an XDG compliant location themselves.

nakengelhardt:
Reading the spec, wouldn't `XDG_STATE_HOME` be the appropriate directory?
> The $XDG_STATE_HOME contains state data that should persist between (application) restarts, but that is not important or portable enough to the user that it should be stored in $XDG_DATA_HOME. It may contain:
> - actions history (logs, history, recently used files, …)
> - current state of the application that can be reused on a restart (view, layout, open files, undo history, …)

I don't think there's a point to having an extra variable, the people who care about where the files land can set the XDG variables (we should perhaps note that somewhere in the docs, but I'm not sure where would be a good place). The fallback is mostly for the people who don't really care and just want things to work, so I think what the spec says works fine there:
>If $XDG_STATE_HOME is either not set or empty, a default equal to $HOME/.local/state should be used.

I would be happy to change the fallback location to that, the one-time loss of contents of the history file is not a big deal IMO.

hakan-demirli:
I accidentally requested review for the previous commit. I am sorry about that. It's my first time using PR on GitHub.

As requested fallback location is `$HOME/.local/state` and XDG_STATE_HOME is used if it exists.

kernel/driver.cc
```diff
@@ -244,11 +244,15 @@ int main(int argc, char **argv)
...
-		read_history(yosys_history_file.c_str());
-		yosys_history_offset = where_history();
+	if (getenv("XDG_STATE_HOME") == NULL || getenv("XDG_STATE_HOME")[0] == '\0') {
+		if (getenv("HOME") != NULL) {
+			yosys_history_file = stringf("%s/.local/state/.yosys_history", getenv("HOME"));
+		}
+	} else {
+		yosys_history_file = stringf("%s/.yosys_history", getenv("XDG_STATE_HOME"));
 	}
+	read_history(yosys_history_file.c_str());
```
whitequark, line 254: I think this will fail if neither `HOME` nor `XDG_STATE_HOME` are set. In that case you should not read the history file at all.

kernel/driver.cc
```diff
@@ -244,8 +244,14 @@ int main(int argc, char **argv)
 	bool mode_q = false;
 
 #if defined(YOSYS_ENABLE_READLINE) || defined(YOSYS_ENABLE_EDITLINE)
-	if (getenv("HOME") != NULL) {
-		yosys_history_file = stringf("%s/.yosys_history", getenv("HOME"));
+	if (getenv("XDG_STATE_HOME") == NULL || getenv("XDG_STATE_HOME")[0] == '\0') {
+		if (getenv("HOME") != NULL) {
+			yosys_history_file = stringf("%s/.local/state/.yosys_history", getenv("HOME"));
```
whitequark, line 249: I think this should probably not be a dotfile when it's under `~/.local/state`:

```suggestion
			yosys_history_file = stringf("%s/.local/state/yosys_history", getenv("HOME"));
```

whitequark: Also, what happens if there is no `~/.local/state` directory? It should probably be created if it doesn't exist.

whitequark: It seems that `~/.local/state` is often nonexistent. If we are creating a directory here already, it's best to create `~/.local/state/yosys/` and then put a file named `history` there.

mmicko:
@hakan-demirli please note that we build mingw build as main windows build (for oss-cad-suite) and for it `_mkdir` is in `direct.h`, so chec for `_WIN32` is enough for that one, and `sys/stat.h` and `sys/types.h` are both  available on mingw

hakan-demirli:
I will clean the code since there are redundancies and unnecessary includes which are also coming from yosys.h. However it functionally works for all combinations of HOME and XDG_STATE_DIR. I will also test it with mingw. 

However I have two questions.
- Should I put newly created functions inside `yosys.h` with an inline attribute? Or like a regular function with an implementation in   `yosys.cc`?
- Boost libraries are listed as a dependency, but they are scarcely used. Why is that? 90 lines of code to create a directory tree is quite a tech burden imo. I have noticed similar patterns throughout the repo where one lines of cpp17 or boost code could clean tens of lines of code.

whitequark:
> * Boost libraries are listed as a dependency, but they are scarcely used.

Boost isn't a hard Yosys dependency. (There are some features that only work with Boost, but the base Yosys application can be built without Boost.)

kernel/driver.cc
```diff
@@ -244,11 +244,35 @@ int main(int argc, char **argv)
...
+
+    if (!state_dir.empty()) {
+        std::string yosys_dir = state_dir + "/yosys";
+        create_directory(yosys_dir);
+
+        yosys_history_file = yosys_dir + "/history";
+        read_history(yosys_history_file.c_str());
+        yosys_history_offset = where_history();
+    } else {
+        log("state_dir is empty. No history file will be created.");
```
whitequark, line 274: Maybe print what `state_dir` is? As a downstream user of Yosys I would have zero clue what this means.

whitequark (approved):
LGTM

mmicko:
Have tested windows build and it creates directory correctly. However there is issue with log messages, they are never displayed. 
Note that logger is initialized later in main function so actual  `log` call does not do anything. So solution is to replace it with `printf` or `fprintf` but then it will not appear in logs at all, or to keep it as `log` but move this complete block after, that will make it also more visible to users.
Also log messages are missing `\n` so that needs fixing.

`
"Directory to put history file does not exist. If you are on Windows either $HOMEDRIVE or $HOMEPATH is empty."
`

maybe it should be (not sure about Windows part needed ).

`
"Unable to determine directory to put history file in.\n If you are on Windows either $HOMEDRIVE or $HOMEPATH is empty.\n"
`

PR #4100, opened 2024-01-01
Add support for packed multidimensional arrays

### LINKED ISSUE #667: Feature request: Support for multidimensional wires
This works:

`  input [1:0] a;`

but this does not work:

`  input [4:0][1:0] a;`

This may be a duplicate issue, but I am reporting all the things I find which can be supported in the future.

### LINKED ISSUE #3776: [ERROR] Synthesis RTL with unpack/pack array internal signal
### Version

Yosys 0.28+6

Linux

### Reproduction Steps

i tried to synthesis myRTL with pack array internal signal **wire  [1:0][1:0] mysignal ;**

I can't use command **read -sv myRTL.sv** to read this file. when i run this command i get the following error **ERROR: syntax error, unexpected '[', expecting TOK_ID or # **

daglem:
* Generalization of dimensions metadata (also simplifies $size et al.)
* Parsing and elaboration of multidimensional packed arrays
* Implementation of $increment, $dimensions and $unpacked_dimensions
* Resolve multiple dimensions defined in stages with typedef

Fixes #667
Fixes #3776

### DIFF
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -3603,10 +3603,11 @@ skip_dynamic_range_lvalue_expansion:;
 				goto apply_newNode;
 			}
 
-			if (str == "\\$increment" || str == "\\$size" || str == "\\$bits" || str == "\\$high" || str == "\\$low" || str == "\\$left" || str == "\\$right")
+			if (str == "\\$dimensions" || str == "\\$unpacked_dimensions" ||
+			    str == "\\$increment" || str == "\\$size" || str == "\\$bits" || str == "\\$high" || str == "\\$low" || str == "\\$left" || str == "\\$right")
 			{
 				int dim = 1;
-				if (str == "\\$bits") {
+				if (str == "\\$dimensions" || str == "\\$unpacked_dimensions" || str == "\\$bits") {
 					if (children.size() != 1)
 						input_error("System function %s got %d arguments, expected 1.\n",
 								RTLIL::unescape_id(str).c_str(), int(children.size()));
@@ -3625,6 +3626,7 @@ skip_dynamic_range_lvalue_expansion:;
 				AstNode *buf = children[0]->clone();
 				int mem_depth = 1;
 				int result, high = 0, low = 0, left = 0, right = 0, width = 1; // defaults for a simple wire
+				int expr_dimensions = 0, expr_unpacked_dimensions = 0;
 				AstNode *id_ast = NULL;
 
 				buf->detectSignWidth(width_hint, sign_hint);
@@ -3650,6 +3652,10 @@ skip_dynamic_range_lvalue_expansion:;
 						// or if the second argument is out of range, then 'x shall be returned."
 						if (dim < 1 || dim > dims)
 							input_error("Dimension %d out of range in `%s', as it only has %d dimensions!\n", dim, id_ast->str.c_str(), dims);
+
+						expr_dimensions = dims - dim + 1;
+						expr_unpacked_dimensions = std::max(id_ast->unpacked_dimensions - dim + 1, 0);
+
 						right = low  = id_ast->dimensions[dim - 1].range_right;
 						left  = high = low + id_ast->dimensions[dim - 1].range_width - 1;
 						if (id_ast->dimensions[dim - 1].range_swapped) {
@@ -3662,9 +3668,16 @@ skip_dynamic_range_lvalue_expansion:;
 					width = high - low + 1;
 				} else {
 					width = width_hint;
+					right = low  = 0;
+					left  = high = width - 1;
+					expr_dimensions = 1;
 				}
 				delete buf;
-				if (str == "\\$high")
+				if (str == "\\$dimensions")
+					result = expr_dimensions;
+				else if (str == "\\$unpacked_dimensions")
+					result = expr_unpacked_dimensions;
+				else if (str == "\\$high")
 					result = high;
 				else if (str == "\\$low")
 					result = low;
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -381,8 +381,7 @@ static int size_packed_struct(AstNode *snode, int base_offset)
 	snode->range_right = base_offset;
 	snode->range_left = base_offset + width - 1;
 	snode->range_valid = true;
-	if (snode->dimensions.empty())
-		snode->dimensions.push_back({ 0, width, false });
+	snode->dimensions.push_back({ 0, width, false });
 
 	return width;
 }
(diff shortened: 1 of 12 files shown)


tests/simple/memory.v
```diff
@@ -44,6 +44,7 @@ input [2:0] bit;
 output reg y1, y2;
 output y3, y4;
 
+(* nomem2reg *)
 reg [7:0] mem1 [3:0];
```
povik, line 48: Is adding the attribute related to the other changes here?

tests/svtypes/struct_sizebits.sv
```diff
@@ -25,6 +25,18 @@ struct packed {
 //wire [$bits({s.x, s.x})-1:0]xx_bits;
 
 always_comb begin
+	assert ($dimensions(s) == 1);
+	assert ($dimensions(s.t) == 1);
+	assert ($dimensions(s.x) == 1);
+`ifndef VERIFIC
+	assert ($dimensions({3{s.x}}) == 1);
```
povik, line 32: Huh, is there an issue with Verific?

tests/simple/memory.v
daglem: Not really, but while testing the branch I found that `mem1` was automatically converted to registers, while the point here seems to be to test both memory (`mem1`) and memory converted to registers (`mem2`).

Something has probably changed along the way here wrt. the heuristics for automatic conversion to registers, possibly with the change from `memory_bram` to `memory_libmap`?

tests/svtypes/struct_sizebits.sv
daglem: While reviewing #3675, @jix found that Verific failed on `assert ($size({3{s.x}}) == 3*4);`, so I just assumed it would also fail on this (I don't have Verific to test).

povik: Just tested it out, and indeed

> VERIFIC-ERROR [VERI-1267] test.sv:31: argument of 'dimensions' should be an identifier

tests/simple/memory.v
povik: It's a nit, but it's best to separate changes like this into commits of their own.

> Something has probably changed along the way here wrt. the heuristics for automatic conversion to registers, possibly with the change from memory_bram to memory_libmap?

FWIW the `mem2reg`/`nomem2reg` attribute is understood by the frontend, once we are operating on the netlist they are not picked up by anything. Right now I don't understand the nature of the test here and the frontend heuristics for `mem2reg` conversion to say if we want the attribute here.

povik:
I only gave this a superficial drive-by look, I don't understand the part of the frontend this is touching. We will probably need to wait for @zachjs to give this a proper review, if he has the chance.

tests/simple/memory.v
daglem: > It's a nit, but it's best to separate changes like this into commits of their own.

Sure, I'll split this change out.
 
> > Something has probably changed along the way here wrt. the heuristics for automatic conversion to registers, possibly with the change from memory_bram to memory_libmap?
> 
> FWIW the `mem2reg`/`nomem2reg` attribute is understood by the frontend, once we are operating on the netlist they are not picked up by anything. Right now I don't understand the nature of the test here and the frontend heuristics for `mem2reg` conversion to say if we want the attribute here.

Right. I checked why `mem1` is converted to registers by default - it's because all the (non-initial) assignments to `mem1` are made using constant indices (`MEM2REG_FL_CONST_LHS` is set in `simplify.cc`).

I'm pretty sure the `(* nomem2reg *)` attribute on `mem1` makes sense there, since otherwise the test of `mem1` is superfluous (covered by the test of `mem2`).

I checked the timeline of this. Since `(* mem2reg *)` was added to the test before `MEM2REG_FL_CONST_LHS` was introduced in `simplify.cc`, it seems clear that adding `(* nomem2reg *)` to `mem1` is necessary to make the test work according to the original intention:
```
$ git show -s --format="%h %as %s" 19dba2561e tests/simple/memory.v
19dba2561 2013-11-20 Implemented part/bit select on memory read
$ git show -s --format="%h %as %s" 7cfae2c52 frontends/ast/simplify.cc
7cfae2c52 2019-03-01 Use mem2reg on memories that only have constant-index write ports
````

File: frontends/ast/simplify.cc [hunk 3]
```diff
@@ -1902,89 +1865,70 @@ bool AstNode::simplify(bool const_fold, int stage, int width_hint, bool sign_hin
...
-				goto apply_newNode;
-			}
-			delete children[1];
-			children.pop_back();
+			// Pretend it's just a wire in order to resolve the type in the code block above.
+			AstNodeType param_type = type;
+			type = AST_WIRE;
+			AstNode *expr = children[0];
+			children.erase(children.begin());
+			while (is_custom_type && simplify(const_fold, stage, width_hint, sign_hint)) {};
```
zachjs, line 1926: If this type substitution pattern proliferates, let's factor it out. You can leave it for now.

zachjs (commented):
This is a great change that both extends the frontend and improves its overall code quality. Thank you!

My most substantive request is that we add a few more tests, especially for the "nested" typedefs logic. I think we find a way to break anything in the frontend that isn't fully tested. If you can't do this (you've already done a lot!), I'd be happy to. In any event, it shouldn't prevent this from getting merged. I think we can merge it right _after_ the next release.

daglem:
> This is a great change that both extends the frontend and improves its overall code quality. Thank you!
> 
> My most substantive request is that we add a few more tests, especially for the "nested" typedefs logic. I think we find a way to break anything in the frontend that isn't fully tested. If you can't do this (you've already done a lot!), I'd be happy to. In any event, it shouldn't prevent this from getting merged. I think we can merge it right _after_ the next release.

It would be great if you could add the tests you find relevant, thank you!

I am aware of one construct where nested typedefs will yield incorrect results; arrays of structs/unions. However this is material for another PR.

frontends/verilog/verilog_parser.y
```diff
@@ -197,9 +197,20 @@ static AstNode *checkRange(AstNode *type_node, AstNode *range_node)
...
+		bool valid = true;
+		if (range_node->type == AST_RANGE) {
+			valid = range_node->children.size() == 2;
+		} else {  // AST_MULTIRANGE
+			for (auto child : range_node->children) {
+				valid = valid && child->children.size() == 2;
+			}
+		}
+		if (!valid)
+			frontend_verilog_yyerror("wire/reg/logic packed dimension must be of the form: [<expr>:<expr>], [<expr>+:<expr>], or [<expr>-:<expr>]");
```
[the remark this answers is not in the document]
daglem: OK, fixed. Note that `[<expr>]` would not be correct here, since `checkRange` is only used to check packed ranges, and `[<expr>]` is only allowed for unpacked ranges in the LRM.

I first renamed `checkRange` to `checkPackedRange` in order to make this clearer, but decided against it in order to keep the diff small. This could always be changed later.

frontends/ast/simplify.cc
```diff
@@ -5360,6 +5276,17 @@ void AstNode::meminfo(int &mem_width, int &mem_size, int &addr_bits)
...
+	/*
+	mem_size = 1;
+	mem_width = 1;
+	for (int i = 0; i < unpacked_dimensions; i++) {
+		mem_size *= dimensions[i].range_width;
+	}
+	for (int i = unpacked_dimensions; i < GetSize(dimensions); i++) {
+		mem_width *= dimensions[i].range_width;
+	}
+	*/
```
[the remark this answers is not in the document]
daglem: It's probably going to have to be used at some point (arrays of structs), but I've removed it for now.

frontends/ast/simplify.cc
```diff
@@ -4936,6 +4849,9 @@ void AstNode::mem2reg_as_needed_pass1(dict<AstNode*, pool<std::string>> &mem2reg
 	{
 		AstNode *mem = id2ast;
 
+		if (integer < (unsigned)mem->unpacked_dimensions)
+			input_error("Insufficient number of array indices for %s.\n", log_id(str));
```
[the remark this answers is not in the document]
daglem: This is really a shortcoming in Yosys, which currently treats anything with unpacked ranges as "memory", where the memory word consists of the packed part, and only indexing which ultimately selects the packed part is allowed (contrary to the LRM).

I'd rather not add tests for this shortcoming; I think it would be better to consider what could possibly be done about this in the future. Perhaps automatic `mem2reg` conversion of "memory" which has unpacked parts selected?

frontends/ast/simplify.cc
```diff
@@ -3609,10 +3582,11 @@ skip_dynamic_range_lvalue_expansion:;
 				AstNode *buf = children[0]->clone();
 				int mem_depth = 1;
 				int result, high = 0, low = 0, left = 0, right = 0, width = 1; // defaults for a simple wire
+				int expr_dimensions = 0, expr_unpacked_dimensions = 0;
 				AstNode *id_ast = NULL;
 
 				// Is this needed?
-				//while (buf->simplify(true, false, stage, width_hint, sign_hint, false)) { }
+				//while (buf->simplify(true, stage, width_hint, sign_hint)) { }
```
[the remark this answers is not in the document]
daglem: No idea whether it can be needed in some situation, really - I only adapted the commented out code to earlier changes in Yosys.

In any case I've removed the comment now.

frontends/ast/simplify.cc
```diff
@@ -3622,114 +3596,53 @@ skip_dynamic_range_lvalue_expansion:;
...
-										}
-									} else if ((dim > dims+1) || (dim < 0))
-										input_error("Dimension %d out of range in `%s', as it only has dimensions 1..%d!\n", dim, buf->str.c_str(), dims+1);
-								}
-							} else {
-								input_error("Unknown memory depth AST type in `%s'!\n", buf->str.c_str());
-							}
+						int dims = GetSize(id_ast->dimensions);
+						if (dim < 1 || dim > dims)
+							input_error("Dimension %d out of range in `%s', as it only has %d dimensions!\n", dim, id_ast->str.c_str(), dims);
```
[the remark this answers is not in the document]
daglem: I actually noticed this (old) incorrect behavior as well, however I had to stop somewhere for now.
TODO added.

frontends/ast/simplify.cc
```diff
@@ -3158,6 +3125,8 @@ skip_dynamic_range_lvalue_expansion:;
 	// found right-hand side identifier for memory -> replace with memory read port
 	if (stage > 1 && type == AST_IDENTIFIER && id2ast != NULL && id2ast->type == AST_MEMORY && !in_lvalue &&
 			children.size() == 1 && children[0]->type == AST_RANGE && children[0]->children.size() == 1) {
+		if (integer < (unsigned)id2ast->unpacked_dimensions)
+			input_error("Insufficient number of array indices for %s.\n", log_id(str));
```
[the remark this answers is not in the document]
daglem: It doesn't, however it is old behavior, and incorrect at that - see comment above.

frontends/ast/simplify.cc
daglem: OK, I'll keep it in mind.

zachjs (approved):

PR #4104, opened 2024-01-04
Correct hierarchical path names for structs and unions

daglem:


whitequark (approved):
This is a welcome improvement! I would like a second pair of eyes on the diff though.

zachjs (approved):
This seems right. Thanks for fixing this!

PR #4105, opened 2024-01-05
cxxrtl: add a representation of simulation timestamps

whitequark:
While the VCD format separates the timescale and the timestep (likely to allow representing the timestep with a small integer type), time in CXXRTL is represented using a uniform 96-bit number, which allows for a ±100 year range at femtosecond resolution.

The implementation uses `value<96>`, which provides fast arithmetic and comparison operations, as well as conversion to/from a more common representation of integer seconds plus femtoseconds.

This isn't yet used anywhere, but will be used in the upcoming record/replay debugging feature.

### DIFF
--- a/backends/cxxrtl/runtime/cxxrtl/cxxrtl_time.h
+++ b/backends/cxxrtl/runtime/cxxrtl/cxxrtl_time.h
@@ -1,0 +1,229 @@
+/*
+ *  yosys -- Yosys Open SYnthesis Suite
+ *
+ *  Copyright (C) 2023  Catherine <whitequark@whitequark.org>
+ *
+ *  Permission to use, copy, modify, and/or distribute this software for any
+ *  purpose with or without fee is hereby granted.
+ *
+ *  THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ *  WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ *  MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ *  ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ *  WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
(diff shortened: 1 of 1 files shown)


backends/cxxrtl/runtime/cxxrtl/cxxrtl_time.h
```diff
@@ -0,0 +1,229 @@
...
+	bool is_zero() const {
+		return raw.is_zero();
+	}
+
+	// Extracts the sign of the value.
+	bool is_negative() const {
+		return raw.is_neg();
+	}
+
+	// Extracts the absolute number of whole seconds. Negative if the value is negative.
```
povik, line 71: What does 'absolute' mean here if the number can go negative?

whitequark: Oh, that is an editing error. Ditto below.

backends/cxxrtl/runtime/cxxrtl/cxxrtl_time.h
```diff
@@ -0,0 +1,229 @@
...
+
+#include <cinttypes>
+#include <string>
+
+#include <cxxrtl/cxxrtl.h>
+
+namespace cxxrtl {
+
+// A timestamp or a difference in time, stored as a 96-bit number of femtoseconds (10e-15 s). The dynamic range and
+// resolution of this format can represent any VCD timestamp within 136 years, without the need for a timescale.
```
povik, line 30: At first I was confused here. The 136 year limit doesn't follow from storing femtoseconds in a 96-bit number, but in storing seconds in a 32-bit number as is done elsewhere in the code. The 96-bit femtoseconds have a ~2512308 year limit.

povik: A time point outside the ~136 year window can overflow if converted to a string and then parsed back, but I don't know if we care.

povik (commented):
Bit late to the party, looks good to me other than the overflow comment

backends/cxxrtl/runtime/cxxrtl/cxxrtl_time.h
whitequark: I guess converting `time::maximum()` would overflow.

whitequark:
Thanks for the review! Follow-up at https://github.com/YosysHQ/yosys/pull/4108

PR #4106, opened 2024-01-05
cxxrtl: add a way to observe state changes during the commit step

whitequark:
The commit observer is a structure containing a callback that is invoked whenever the `commit()` method changes a wire or a memory. This allows code external to the compiled netlist to react to changes in the design state in a very efficient way. One example of how this feature can be used is an efficient implementation of record/replay.

Note that the VCD writer does not benefit from this feature because it must be able to react to changes in any debug items and not just those that contain design state.

<hr>

There is a performance penalty associating with translating wire/memory identity (`base` argument) to some sort of persistent identifier in an upper layer using this feature. The penalty, specific to this translation, can be as high as 25% compared to an uninstrumented design. At a later point, the argument to the observer may end up being switched from a base pointer to a permanently assigned integer, but there are challenges associated with that design and the current one is much simpler while still providing adequate performance.

### DIFF
--- a/backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
+++ b/backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
@@ -837,6 +837,29 @@ std::ostream &operator<<(std::ostream &os, const value_formatted<Bits> &vf)
 	return os;
 }
 
+// An object that can be passed to a `commit()` method in order to produce a replay log of every
+// state change in the simulation.
+struct observer {
+	// Called when a `commit()` method for a wire is about to update the `chunks` chunks at `base`
+	// with `chunks` chunks at `value` that have a different bit pattern. It is guaranteed that
+	// `chunks` is equal to the wire chunk count and `base` points to the first chunk.
+	virtual void on_commit(size_t chunks, const chunk_t *base, const chunk_t *value) = 0;
+
+	// Called when a `commit()` method for a memory is about to update the `chunks` chunks at
+	// `&base[chunks * index]` with `chunks` chunks at `value` that have a different bit pattern.
+	// It is guaranteed that `chunks` covers is equal to the memory element chunk count and `base`
+	// points to the first chunk of the first element of the memory.
+	virtual void on_commit(size_t chunks, const chunk_t *base, const chunk_t *value, size_t index) = 0;
+};
+
+// The `null_observer` class has the same interface as `observer`, but has no invocation overhead,
+// since its methods are final and have no implementation. This allows the observer feature to be
+// zero-cost when not in use.
+struct null_observer final: observer {
+	void on_commit(size_t chunks, const chunk_t *base, const chunk_t *value) override {}
+	void on_commit(size_t chunks, const chunk_t *base, const chunk_t *value, size_t index) override {}
+};
+
 template<size_t Bits>
 struct wire {
 	static constexpr size_t bits = Bits;
@@ -871,8 +894,14 @@ struct wire {
 		next.template set<IntegerT>(other);
 	}
 
-	bool commit() {
+	// This method intentionally takes a mandatory argument (to make it more difficult to misuse in
+	// black box implementations, leading to missed observer events). It is generic over its argument
+	// to make sure the `on_commit` call is devirtualized. This is somewhat awkward but lets us keep
+	// a single implementation for both this method and the one in `memory`.
+	template<class ObserverT>
+	bool commit(ObserverT &observer) {
 		if (curr != next) {
+			observer.on_commit(curr.chunks, curr.data, next.data);
 			curr = next;
 			return true;
 		}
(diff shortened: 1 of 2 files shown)


backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
```diff
@@ -837,6 +837,29 @@ std::ostream &operator<<(std::ostream &os, const value_formatted<Bits> &vf)
...
+// state change in the simulation.
+struct observer {
+	// Called when a `commit()` method for a wire is about to update the `chunks` chunks at `base`
+	// with `chunks` chunks at `value` that have a different bit pattern. It is guaranteed that
+	// `chunks` is equal to the wire chunk count and `base` points to the first chunk.
+	virtual void on_commit(size_t chunks, const chunk_t *base, const chunk_t *value) = 0;
+
+	// Called when a `commit()` method for a memory is about to update the `chunks` chunks at
+	// `&base[chunks * index]` with `chunks` chunks at `value` that have a different bit pattern.
+	// It is guaranteed that `chunks` covers is equal to the memory element chunk count and `base`
```
povik, line 850: Nit: 'covers' is extra there?

backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
```diff
@@ -837,6 +837,29 @@ std::ostream &operator<<(std::ostream &os, const value_formatted<Bits> &vf)
 	return os;
 }
 
+// An object that can be passed to a `commit()` method in order to produce a replay log of every
+// state change in the simulation.
+struct observer {
+	// Called when a `commit()` method for a wire is about to update the `chunks` chunks at `base`
+	// with `chunks` chunks at `value` that have a different bit pattern. It is guaranteed that
+	// `chunks` is equal to the wire chunk count and `base` points to the first chunk.
+	virtual void on_commit(size_t chunks, const chunk_t *base, const chunk_t *value) = 0;
```
povik, line 846: So this would be a pointer to `wire` except that the wire type is parametrized, so we take `chunk_t *` to the wire's data instead?

povik: (Regarding the `base` argument)

whitequark: Correct. In practice there are two reasons this argument is there:
- It provides the identity of the wire (which can be correlated via debug items to its name, etc.)
- It provides the value before the commit.

backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
whitequark: https://github.com/YosysHQ/yosys/pull/4109

backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
```diff
@@ -871,8 +894,14 @@ struct wire {
 		next.template set<IntegerT>(other);
 	}
 
-	bool commit() {
+	// This method intentionally takes a mandatory argument (to make it more difficult to misuse in
+	// black box implementations, leading to missed observer events). It is generic over its argument
```
povik, line 898: > (to make it more difficult to misuse in black box implementations, leading to missed observer events)

I like this choice.

whitequark: I will most likely completely remove user-defined black boxes as they are right now, but in the interim this works.

backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
povik: I guess as an option we could pass in the number of bits instead of number of chunks, providing more infromation. This is a trade-off, I assume the number of bits can be looked up through debug items anyway, and passing in the number of chunks optimizes this more towards fast recording of state.

backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
povik: FWIW I have used user-defined black boxes and would be sad to see them go

backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
whitequark: Yes. This interface is solely concerned with the storage representation.

backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
whitequark: It's very difficult, bordering on impossible, to combine user-defined black boxes and record/replay debugging. They are not being completely removed though but rather replaced with something that is behaving more predictably.

The replacement will restrict the black boxes to only have synthesizable logic and also wires that are treated as-if they were toplevel ports, and provide a way to retrieve these wires in a way that is more convenient than individually plucking them from the debug items map.

The idea is that you put as much logic as you can into the synthesizable part (e.g. an UART receiver), and then, from the outside (from the simulation point of view, not necessarily C++ code point of view), you will be able to check a flag indicating there's a character received, and print it on screen. In most cases, these pseudo toplevel ports should be structured as FIFOs but there is freedom for the application to do whatever it wants, allowing implementing things like memories or caches as well.

This also avoids the race condition issue that has been plaguing the black box implementation since it was added.

The actual interface for instantiating these new-style black boxes isn't finalized yet.

<hr>

The above covers the use of black boxes for I/O. There is another valid use of black boxes, which is to make a simulation faster by replacing a block with the equivalent native code. I might continue to enable that use by removing the black box instantiation mechanism and simply emitting a normal module where `eval` is user-defined. `commit` would not be overridable at all, nor would instantiation be customizable.

PR #4109, opened 2024-01-05
cxxrtl: fix comment wording. NFC

whitequark:


povik (approved):
I stumble a bit over `chunks` chunks when reading those comments but anyway I read through the new text once, looks good to me.

PR #4111, opened 2024-01-06
write_verilog: Handle edge case with non-pruned processes

povik:
A minor thing I stumbled on while reviewing #4090

This change only matters for processes that weren't processed by `proc_rmdead` for which follow-up cases after a default case are treated differently in Verilog and RTLIL semantics.

### DIFF
--- a/backends/verilog/verilog_backend.cc
+++ b/backends/verilog/verilog_backend.cc
@@ -1988,12 +1988,10 @@ void dump_proc_switch(std::ostream &f, std::string indent, RTLIL::SwitchRule *sw
 	dump_sigspec(f, sw->signal);
 	f << stringf(")\n");
 
-	bool got_default = false;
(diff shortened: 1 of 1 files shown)


backends/verilog/verilog_backend.cc
```diff
@@ -2006,6 +2004,14 @@ void dump_proc_switch(std::ostream &f, std::string indent, RTLIL::SwitchRule *sw
 		}
 		f << stringf(":\n");
 		dump_case_body(f, indent + "    ", *it);
+
```
whitequark, line 2007: Nit: Now that the variable is no longer used to carry state across iterations I think it's best to inline it.

whitequark (approved):

backends/verilog/verilog_backend.cc
povik: Thanks!

PR #4118, opened 2024-01-08
rtlil: Fix handling of connections on wire deletion

povik:
Ought to fix #4082

The issue here is we replace removed wires in assignments with x-bits on both the LHS and RHS of the assignment. This triggers an assertion later on because we cannot assign to a constant.

PR #4123, opened 2024-01-09
opt_clean: Add commentary, remove dead code

povik:
(This is a by-product of investigating the hang from #4118.)

See the commits.

### DIFF
--- a/passes/opt/opt_clean.cc
+++ b/passes/opt/opt_clean.cc
@@ -240,6 +240,7 @@ int count_nontrivial_wire_attrs(RTLIL::Wire *w)
 	return count;
 }
 
+// Should we pick `s2` over `s1` to represent a signal?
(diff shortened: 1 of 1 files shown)


passes/opt/opt_clean.cc
```diff
@@ -418,6 +438,9 @@ bool rmunused_module_signals(RTLIL::Module *module, bool purge_mode, bool verbos
 			goto delete_this_wire;
 		} else
 		if (!used_signals.check_any(s2)) {
+			// this path shouldn't be possible: this wire is used directly (otherwise it would get cleaned up above), and indirectly
+			// used wires are a superset of those used directly
+			log_assert(false);
```
povik, line 443: I pushed the commits piecemeal so there's a CI job with this assertion in place.

povik:
Rebased and dropped the final patch removing the assertion. To play it safe we plan to keep the assertion in for a release or two.

PR #4125, opened 2024-01-09
read_blif: Represent sequential elements with gate cells

### LINKED ISSUE #4117: The write_blif command not working properly.
### Version

Final Version

Linux

### Reproduction Steps

I am currently trying to read a blif file with Yosys, run the ABC9 pass on it, and write it as a blif file. I encountered a double definition of a signal when trying to read the blif with ABC. The same thing happens when trying to read the blif using Yosys. Yosys cannot read a blif file generated by itself. I used different blif files and got the same results. This is a and_latch module as an example.

```
.model and_latch
.inputs and_latch^clock and_latch^a_in and_latch^b_in
.outputs and_latch^out

.names gnd
.names unconn
.names vcc
1

.latch and_latch^bAND~0^lAND~1 and_latch^out_FF re and_latch^clock 3

.names and_latch^a_in and_latch^b_in and_latch^bAND~0^lAND~1
11 1

.names and_latch^out_FF and_latch^out
1 1

.end
```
Here is my terminal log for your reference

```
yosys> read_blif and_latch.blif 
1. Executing BLIF frontend.

yosys> write_blif output.blif

2. Executing BLIF backend.

yosys> read_blif output.blif 

3. Executing BLIF frontend.
ERROR: Duplicate definition of module and_latch in line 3!
```

Here are some other blif files for your reference.
[adder.blif.txt](https://github.com/YosysHQ/yosys/files/13856229/adder.blif.txt)
[sv.blif.txt](https://github.com/YosysHQ/yosys/files/13856238/sv.blif.txt)

### Expected Behavior

The expected behavior is writing a blif file that can be read by Yosys and ABC.

### Actual Behavior

The ```write_blif``` command is saving the blif file in the wrong way that cannot be re-read.

povik:
When reading the BLIF input, represent the native sequential elements with fine-grained cells like `$_FF_` instead of the coarse-grained cells like `$ff` which we were using up to now.

There are two reasons for this:

 * The sequential elements in BLIF are always single-bit, so the gate cells are a better fit.

 * This makes it symmetrical to the BLIF backend which only understands the fine-grained cells, and only translates those to the native BLIF features.

Fix #4117.

### DIFF
--- a/frontends/blif/blifparse.cc
+++ b/frontends/blif/blifparse.cc
@@ -352,17 +352,17 @@ void parse_blif(RTLIL::Design *design, std::istream &f, IdString dff_name, bool 
 					goto no_latch_clock;
 
 				if (!strcmp(edge, "re"))
-					cell = module->addDff(NEW_ID, blif_wire(clock), blif_wire(d), blif_wire(q));
+					cell = module->addDffGate(NEW_ID, blif_wire(clock), blif_wire(d), blif_wire(q));
 				else if (!strcmp(edge, "fe"))
-					cell = module->addDff(NEW_ID, blif_wire(clock), blif_wire(d), blif_wire(q), false);
+					cell = module->addDffGate(NEW_ID, blif_wire(clock), blif_wire(d), blif_wire(q), false);
 				else if (!strcmp(edge, "ah"))
-					cell = module->addDlatch(NEW_ID, blif_wire(clock), blif_wire(d), blif_wire(q));
+					cell = module->addDlatchGate(NEW_ID, blif_wire(clock), blif_wire(d), blif_wire(q));
 				else if (!strcmp(edge, "al"))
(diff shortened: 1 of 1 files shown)


KrystalDelusion:
I noticed a similar problem with write/read_aiger where it reads in `$ff` cells, but when the write throws an error due to unsupported `$ff` cells and needed a call to techmap to work.

nakengelhardt:
Hmm, I would think that is expected behavior rather than a problem - reading in files happens at the beginning of a flow (usually coarse grain), whereas writing happens at the end (often fine grain, but not necessarily). There is always an easy way to get to fine grain cells if you require them (techmap), but there is no good way to get back to coarse grain cells for any operation that only works on those. Especially if we want to move to more defined design stages where a pass can rely on having only coarse-grain cells or only fine-grain cells in the design instead of having to handle every weird state that is possible to construct in RTLIL, having all frontends only produce coarse grain cells might be a good principle to enforce.

PR #4128, opened 2024-01-11
Add `$check` cell to represent assertions with messages

whitequark:
Relates to https://github.com/YosysHQ/yosys/pull/4143.


jix:
Just leaving a note that `async2sync` and `clk2fflogic` will require support to lower `$check` into `$assert`/`$assume`/etc before merging this to avoid breaking our formal verification flows. (I added the check-sby label which should cause jenkins CI runs to also run the SBY tests exercising the FV flows.)

whitequark:
> Just leaving a note that `async2sync` and `clk2fflogic` will require support to lower `$check` into `$assert`/`$assume`/etc before merging this

You mentioned that you would be able to do this--is that still the case?

backends/verilog/verilog_backend.cc
```diff
@@ -1918,9 +1955,13 @@ void dump_sync_print(std::ostream &f, std::string indent, const RTLIL::SigSpec &
 	for (auto cell : cells) {
 		f << stringf("%s" "  if (", indent.c_str());
 		dump_sigspec(f, cell->getPort(ID::EN));
-		f << stringf(")\n");
+		f << stringf(") begin\n");
 
 		dump_cell_expr_print(f, indent + "    ", cell);
```
povik, line 1960: For `$check` I think we want `if (!A)` in front of the print statement

povik:
Just occurred to me, so noting it down: we need to visit `set_keep_assert` in `hierarchy.c` in this PR

backends/verilog/verilog_backend.cc
whitequark: Oh yeah, good catch!

povik: `if (!A)` is good for `assert`, what about the other flavors?

whitequark: I think they're all good with `if (!A)`

whitequark: Wait, no. `A` is emitted as a part of the `assert`/etc itself. It doesn't need to be added the second time. (Which is the whole point of having it.)

whitequark:
> we need to visit `set_keep_assert` in `hierarchy.c` in this PR

Done

backends/verilog/verilog_backend.cc
povik: I am confused. It is part of the `assert`, but e.g. if we roundtrip Verilog through Yosys we don't want to print `assert property ... failed` each cycle the assertion is enabled.

whitequark: Ohhh, you meant for the `$print` part! Yes, absolutely.

whitequark: Check out my latest update, where I think I've handled it.

whitequark:
From my point of view this PR is done, though of course it breaks formal flows, and the tests that are looking for formal flow cells don't pass.

jix:
My branch with the necessary yosys changes to keep the FV flows working by handling `$check` cells in `async2sync` and `clk2fflogic` and by ensuring that all yosys tests call either as required is at https://github.com/jix/yosys/tree/check-fv-compat (which contains a cherry-pick of this PR's commit, such that the tests pass before and after `$check` is added). It passes all yosys tests apart from `tests/verilog/conflict_assert.ys` which we discussed during the dev JF (I didn't look more into that yet).

The corresponding changes on the SBY side are at https://github.com/YosysHQ/sby/pull/259 which keeps all SBY tests passing when using the yosys from my `check-fv-compat` branch.

whitequark:
Fantastic work! Thank you @jix. I'll take a look at the final test failure, probably, tomorr.w

whitequark:
I've updated the PR to not emit an assertion message in the Verilog frontend and to ignore empty message in the Verilog backend; and to check for non-unique names of procedural assertions. I think that's all!

whitequark:
There are a few failures of the kind:

```
ERROR: Failed to import cell $assert$static_cast_simple.sv:35$6 (type $check) to SAT database.
```

jix:
Ah I missed the tests that end up using `read` instead of `read_verilog` as the build I was using has verific enabled which doesn't yet emit `$check` cells. I'll go over the remaining tests with verific disabled later today.

jix:
The Jenkins failure is due to Yosys PRs using the latest nightly oss-cad-suite to obtain SBY and all its dependencies while I merged the corresponding SBY changes only today. It should pass when rerun tomorrow.

whitequark:
@jix @povik Any further review comments, or are we basically done with this patch?

povik:
@whitequark I can have a final look-over tomorrow.

whitequark:
Thanks, that would be great.

frontends/ast/simplify.cc
```diff
@@ -3039,97 +3039,6 @@ bool AstNode::simplify(bool const_fold, int stage, int width_hint, bool sign_hin
 	}
 skip_dynamic_range_lvalue_expansion:;
 
-	if (stage > 1 && (type == AST_ASSERT || type == AST_ASSUME || type == AST_LIVE || type == AST_FAIR || type == AST_COVER) && current_block != NULL)
```
povik, line 3042: Am I happy we got to remove something from `simplify()`...

passes/sat/async2sync.cc
```diff
@@ -41,31 +41,87 @@ struct Async2syncPass : public Pass {
...
+						SigBit sig_trg = cell->getPort(ID::TRG);
+						Wire *sig_en_q = module->addWire(NEW_ID);
+						Wire *sig_args_q = module->addWire(NEW_ID, GetSize(sig_args));
+						sig_en_q->attributes.emplace(ID::init, State::S0);
+						module->addDff(NEW_ID, sig_trg, sig_en, sig_en_q, trg_polarity, cell->get_src_attribute());
+						module->addDff(NEW_ID, sig_trg, sig_args, sig_args_q, trg_polarity, cell->get_src_attribute());
+						cell->setPort(ID::EN, sig_en_q);
+						cell->setPort(ID::ARGS, sig_args_q);
+						if (cell->type == ID($check)) {
+							SigBit sig_a = cell->type == ID($check) ? cell->getPort(ID::A) : State::S0;
```
povik, line 109: Forgotten ternary here

passes/sat/async2sync.cc
```diff
@@ -41,31 +41,87 @@ struct Async2syncPass : public Pass {
...
+							Wire *sig_a_q = cell->type == ID($check) ? module->addWire(NEW_ID) : nullptr;
+							sig_a_q->attributes.emplace(ID::init, State::S1);
```
povik, line 111: The `S1` is fine for assert, but at least for `cover` we want `S0` I think

povik: Though on second thought, `EN` being zero in initstate makes the `A` state irrelevant

povik: So maybe we can drop the initstate for `A` altogether?

jix: It shouldn't matter because the `EN` signal created above will be `0` in the initial state. The main reason I'm setting an initial value at all is to avoid creating uninitialized FFs with private names.

povik: > avoid creating uninitialized FFs with private names.

I assume that matters for formal flows.

passes/sat/clk2fflogic.cc
```diff
@@ -194,79 +199,134 @@ struct Clk2fflogicPass : public Pass {
...
+						SigBit sig_en = cell->getPort(ID::EN);
+						cell->setPort(ID::EN, module->And(NEW_ID, sig_en, initstate));
 					} else {
-						// $sr.
-						log("Replacing %s.%s (%s): SET=%s, CLR=%s, Q=%s\n",
-								log_id(module), log_id(cell), log_id(cell->type),
-								log_signal(ff.sig_set), log_signal(ff.sig_clr), log_signal(ff.sig_q));
+						SigBit sig_en = cell->getPort(ID::EN);
+						SigSpec sig_args = cell->getPort(ID::ARGS);
+						bool trg_polarity = cell->getParam(ID(TRG_POLARITY)).as_bool();
```
povik, line 226: Polarity is per trigger signal, but we reduce it to a single boolean here

passes/sat/clk2fflogic.cc
```diff
@@ -194,79 +199,134 @@ struct Clk2fflogicPass : public Pass {
...
+						SigSpec sig_trg = cell->getPort(ID::TRG);
+
+						SigSpec sig_trg_sampled;
+
+						for (auto const &bit : sig_trg)
+							sig_trg_sampled.append(sample_control_edge(module, bit, trg_polarity, false));
```
povik, line 232: We don't do anything with `sig_trg_sampled`, I think we want to reduce-OR it, then AND it with `sig_en_sampled`

passes/sat/async2sync.cc
jix: All such signals will be turned into public signals to make sure there's a stable name for everything solver chosen (to keep the engines and sim compatible), so adding such signals when there's not actually anything relevant for the solver to solve adds noise and makes it harder to debug anything related to private solver chosen signals that actually do influence assertions or public signals.

povik: I thought as much. Thanks for clarifying!

passes/sat/clk2fflogic.cc
jix: Yeah that's what I intended to do, I must have got distracted while I was in the middle of that. I'm a bit worried that this wasn't caught by the SBY tests. There are clearly not enough tests that use `clk2fflogic` and in general I can't think of any existing SBY projects I have that would exercise `clk2fflogic` without also requiring verific (and thus not generating `$check` cells).

passes/sat/clk2fflogic.cc
jix: Good catch, I completely missed that.

passes/cmds/chformal.cc
```diff
@@ -23,6 +23,52 @@
...
+}
+
+static void set_formal_flavor(RTLIL::Cell *cell, RTLIL::IdString flavor)
+{
+	if (cell->type != ID($check)) {
+		cell->type = flavor;
+		return;
+	}
+
+	if (cell->type == ID($assert))
```
povik, line 53: This looks like it should dispatch on `flavor` not `cell->type`

passes/cmds/chformal.cc
```diff
@@ -278,21 +352,68 @@ struct ChformalPass : public Pass {
...
 				for (auto cell : constr_cells)
-					module->addCover(NEW_ID_SUFFIX("coverenable"),
-						cell->getPort(ID::EN), State::S1, cell->get_src_attribute());
+				{
+					if (cell->type == ID($check)) {
+						Cell *cover = module->addCell(NEW_ID_SUFFIX("coverenable"), ID($check));
+						cover->attributes = cell->attributes;
+						cover->parameters = cell->parameters;
+						for (auto conn : cell->connections())
+							cover->setPort(conn.first, conn.first == ID::A ? State::S1 : conn.second);
```
povik, line 361: This is a minor thing, but if we do `chformal -lower; chformal -coverenable` we get the enable signal of the assertion on the `A` input of the created `$cover`, but if we do `chformal -coverenable; chformal -lower` we get it on `EN` of `$cover`

passes/cmds/chformal.cc
```diff
@@ -278,21 +352,68 @@ struct ChformalPass : public Pass {
...
+
+					plain_cell->attributes = cell->attributes;
+
+					SigBit id_a = cell->getPort(ID::A);
+					SigBit id_en = cell->getPort(ID::EN);
+
+					plain_cell->setPort(ID::A, id_a);
+					plain_cell->setPort(ID::EN, id_en);
+
+					SigBit combined_en = module->And(NEW_ID, id_a, id_en);
```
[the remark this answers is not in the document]
jix: Yeah `A` here should be inverted for assert and assume. While `live` and `fair` are a form of assert and assume, they can only be violated for infinite traces where `A` remains low forever, so printing a message whenever there is progress in keeping the assertion/assumption valid seems like the only useful thing.

frontends/ast/simplify.cc
whitequark: Me too.

passes/cmds/chformal.cc
```diff
@@ -216,6 +274,16 @@ struct ChformalPass : public Pass {
 				}
 
 				for (auto cell : constr_cells)
+				{
+					if (is_triggered_check_cell(cell)) {
+						cell->setPort(ID::TRG, SigSpec());
```
povik, line 279: This looks wrong. We are simply losing the triggers for `-early`

povik (approved):
I went through all of it, only skipped some of the CXXRTL bits, and other than the things I commented on nothing jumped at me.

passes/cmds/chformal.cc
jix: Discarding the triggers is intended and corresponds to the behavior `chformal -early` has prior to these changes (where it would discard a FF no matter what signal the FF is using as a clock) or with these changes after running `async2sync`. It's only valid for a design with a single clock and no async signals, in which case the trigger is the clock and removing the trigger makes it active one clock cycle earlier. For any other kind of design you'd need to run `async2sync` or `clk2fflogic` before running `chformal -early`.

I'm not at all sure now, though, whether my current handling of the `EN` signal is correct and would want to add some tests that better characterize the behavior of it before and after this PR.

povik: We are definitely missing some disclaimers in the help message then.

passes/cmds/chformal.cc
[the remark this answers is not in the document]
jix: For `live` and `fair` you would want to print the fact that they were active every time they were active so I think it makes sense to also print the message in that instance as it might contain relevant signals as arguments.

For `cover` when using SV syntax to manually print a message you would also get that message every time it is reached.

I think in some cases it makes sense to print only the first time for cover, but I'd want that to be user selectable and I think the mode where it prints every time a cover is reached should be the default.

passes/cmds/chformal.cc
jix: I now have a test (that I'll add to this PR) for checking that `async2sync` and `chformal -early` commute and the code in the PR successfully passes that test.

passes/cmds/chformal.cc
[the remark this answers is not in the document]
jix: This is during explicit conversion from `$check` cells to `$assert` etc using `chformal -lower` where I currently do add a `$print` cell when the `$check` cell has a message. This is unrelated to the verilog frontend or backend and only about not losing messages that may be present on `$check` cells during this conversion.

passes/cmds/chformal.cc
jix: I'll change it to be consistent

passes/cmds/chformal.cc
povik: > It's only valid for a design with a single clock and no async signals, in which case the trigger is the clock and removing the trigger makes it active one clock cycle earlier. For any other kind of design you'd need to run async2sync or clk2fflogic before running chformal -early.

If that's the case, let's at least bail here if `TRG_WIDTH > 1`

jix (approved):
This should address everything @povik brought up during review and also adds some tests that would have detected the issues with the now fixed clk2fflogic code and now also passes all tests on CI including the SBY tests.

whitequark:
I'll do the honors.

passes/sat/clk2fflogic.cc
```diff
@@ -194,79 +199,137 @@ struct Clk2fflogicPass : public Pass {
...
+							sig_trg_sampled.append(sample_control_edge(module, bit, trg_polarity[GetSize(sig_trg_sampled)] == State::S1, false));
+						SigSpec sig_args_sampled = sample_data(module, sig_args, Const(State::S0, GetSize(sig_args)), false, false).sampled;
+						SigBit sig_en_sampled = sample_data(module, sig_en, State::S0, false, false).sampled;
+
+						SigBit sig_trg_combined = module->ReduceOr(NEW_ID, sig_trg_sampled);
+
+						cell->setPort(ID::EN, module->And(NEW_ID, sig_en_sampled, sig_trg_combined));
+						cell->setPort(ID::ARGS, sig_args_sampled);
+						if (cell->type == ID($check)) {
+							SigBit sig_a_sampled = sample_data(module, sig_en, State::S1, false, false).sampled;
```
povik, line 241: Should be sampling `sig_a`

PR #4129, opened 2024-01-11
Pass `$display` et al. in `initial` blocks on to `sim`

daglem:
This cherry picks the completed parts from #3963, cherry picks commits by @whitequark correcting the Verilog backend output of the `$print` cell and implementing initial `$print` cells in CXXRTL, and adds the `-nodisplay` option to `read_verilog` to optionally suppress elaboration output from `$display` et al.

Fixes https://github.com/chipsalliance/sv-tests/issues/5082

### DIFF
--- a/frontends/ast/genrtlil.cc
+++ b/frontends/ast/genrtlil.cc
@@ -718,7 +718,7 @@ struct AST_INTERNAL::ProcessGenerator
 					}
 				}
 				cell->parameters[ID::TRG_WIDTH] = triggers.size();
-				cell->parameters[ID::TRG_ENABLE] = !triggers.empty();
+				cell->parameters[ID::TRG_ENABLE] = (always->type == AST_INITIAL) || !triggers.empty();
 				cell->parameters[ID::TRG_POLARITY] = polarity;
 				cell->parameters[ID::PRIORITY] = --last_print_priority;
 				cell->setPort(ID::TRG, triggers);
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -1058,33 +1058,31 @@ bool AstNode::simplify(bool const_fold, int stage, int width_hint, bool sign_hin
 	{
 		if (!current_always) {
 			log_file_warning(filename, location.first_line, "System task `%s' outside initial or always block is unsupported.\n", str.c_str());
-		} else if (current_always->type == AST_INITIAL) {
-			int default_base = 10;
-			if (str.back() == 'b')
-				default_base = 2;
-			else if (str.back() == 'o')
-				default_base = 8;
-			else if (str.back() == 'h')
-				default_base = 16;
-
-			// when $display()/$write() functions are used in an initial block, print them during synthesis
-			Fmt fmt = processFormat(stage, /*sformat_like=*/false, default_base, /*first_arg_at=*/0, /*may_fail=*/true);
-			if (str.substr(0, 8) == "$display")
-				fmt.append_string("\n");
-			log("%s", fmt.render().c_str());
-			for (auto node : children)
-				while (node->simplify(true, stage, -1, false)) {}
-			return false;
+			delete_children();
+			str = std::string();
 		} else {
-			// when $display()/$write() functions are used in an always block, simplify the expressions and
-			// convert them to a special cell later in genrtlil
+			// simplify the expressions and convert them to a special cell later in genrtlil
 			for (auto node : children)
 				while (node->simplify(true, stage, -1, false)) {}
+
+			if (current_always->type == AST_INITIAL && !flag_nodisplay && stage == 2) {
+				int default_base = 10;
+				if (str.back() == 'b')
+					default_base = 2;
+				else if (str.back() == 'o')
+					default_base = 8;
+				else if (str.back() == 'h')
+					default_base = 16;
+
+				// when $display()/$write() functions are used in an initial block, print them during synthesis
+				Fmt fmt = processFormat(stage, /*sformat_like=*/false, default_base, /*first_arg_at=*/0, /*may_fail=*/true);
+				if (str.substr(0, 8) == "$display")
+					fmt.append_string("\n");
+				log("%s", fmt.render().c_str());
+			}
+
 			return false;
 		}
-
-		delete_children();
-		str = std::string();
 	}
 
 	// activate const folding if this is anything that must be evaluated statically (ranges, parameters, attributes, etc.)
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -145,7 +145,7 @@ void AstNode::fixup_hierarchy_flags(bool force_descend)
 
 // Process a format string and arguments for $display, $write, $sprintf, etc
 
-Fmt AstNode::processFormat(int stage, bool sformat_like, int default_base, size_t first_arg_at) {
+Fmt AstNode::processFormat(int stage, bool sformat_like, int default_base, size_t first_arg_at, bool may_fail) {
 	std::vector<VerilogFmtArg> args;
 	for (size_t index = first_arg_at; index < children.size(); index++) {
 		AstNode *node_arg = children[index];
@@ -169,6 +169,9 @@ Fmt AstNode::processFormat(int stage, bool sformat_like, int default_base, size_
 			arg.type = VerilogFmtArg::INTEGER;
 			arg.sig = node_arg->bitsAsConst();
 			arg.signed_ = node_arg->is_signed;
+		} else if (may_fail) {
+			log_file_info(filename, location.first_line, "Skipping system task `%s' with non-constant argument at position %zu.\n", str.c_str(), index + 1);
+			return Fmt();
 		} else {
 			log_file_error(filename, location.first_line, "Failed to evaluate system task `%s' with non-constant argument at position %zu.\n", str.c_str(), index + 1);
 		}
@@ -1065,10 +1068,13 @@ bool AstNode::simplify(bool const_fold, int stage, int width_hint, bool sign_hin
 				default_base = 16;
 
 			// when $display()/$write() functions are used in an initial block, print them during synthesis
-			Fmt fmt = processFormat(stage, /*sformat_like=*/false, default_base);
+			Fmt fmt = processFormat(stage, /*sformat_like=*/false, default_base, /*first_arg_at=*/0, /*may_fail=*/true);
 			if (str.substr(0, 8) == "$display")
 				fmt.append_string("\n");
 			log("%s", fmt.render().c_str());
+			for (auto node : children)
+				while (node->simplify(true, stage, -1, false)) {}
+			return false;
 		} else {
 			// when $display()/$write() functions are used in an always block, simplify the expressions and
 			// convert them to a special cell later in genrtlil
--- a/frontends/verilog/verilog_frontend.cc
+++ b/frontends/verilog/verilog_frontend.cc
@@ -100,6 +100,10 @@ struct VerilogFrontend : public Frontend {
 		log("    -assert-assumes\n");
 		log("        treat all assume() statements like assert() statements\n");
 		log("\n");
+		log("    -nodisplay\n");
+		log("        suppress output from display system tasks ($display et. al).\n");
+		log("        This does not affect the output from a later 'sim' command.\n");
+		log("\n");
 		log("    -debug\n");
 		log("        alias for -dump_ast1 -dump_ast2 -dump_vlog1 -dump_vlog2 -yydebug\n");
 		log("\n");
@@ -235,6 +239,7 @@ struct VerilogFrontend : public Frontend {
 	}
 	void execute(std::istream *&f, std::string filename, std::vector<std::string> args, RTLIL::Design *design) override
 	{
+		bool flag_nodisplay = false;
 		bool flag_dump_ast1 = false;
 		bool flag_dump_ast2 = false;
 		bool flag_no_dump_ptr = false;
@@ -306,6 +311,10 @@ struct VerilogFrontend : public Frontend {
 			}
 			if (arg == "-assert-assumes") {
 				assert_assumes_mode = true;
+				continue;
+			}
+			if (arg == "-nodisplay") {
+				flag_nodisplay = true;
 				continue;
 			}
 			if (arg == "-debug") {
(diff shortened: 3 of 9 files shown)


frontends/ast/genrtlil.cc
```diff
@@ -717,6 +717,10 @@ struct AST_INTERNAL::ProcessGenerator
 						polarity.bits.push_back(RTLIL::S0);
 					}
 				}
+				if (polarity.bits.empty()) {
+					// TRG_POLARITY must be at least one bit, otherwise the "verilog" backend outputs .TRG_POLARITY({0{1'b0}})
```
whitequark, line 721: Is that a problem? I think the cell might fail check when reading it.

whitequark:
I'm not confident that this should be merged in its current form. I'm currently working on modifying the frontend to emit `$check` cells, and there is intersection with what this PR is doing (namely, generation of RTLIL for `initial assert (...);` and later conversion back to Verilog). I think `$check` and `$print` cells should be handled uniformly in this aspect.

frontends/ast/genrtlil.cc
daglem: tests/simple/string_format.v fails, since iverilog doesn't like the `{0{1'b0}}` part.

whitequark: Actually, I have a different question first: when does the output `.TRG_POLARITY({0{1'b0}})` ever happen? That sounds like a bug in and of itself.

daglem:
> I'm not confident that this should be merged in its current form. I'm currently working on modifying the frontend to emit `$check` cells, and there is intersection with what this PR is doing (namely, generation of RTLIL for `initial assert (...);` and later conversion back to Verilog). I think `$check` and `$print` cells should be handled uniformly in this aspect.

Since the changes in the frontend are so small, could you still consider merging this, and improve on it / rewrite it later?
I'd love to have this working now. And perfect is the enemy of good! :wink:

whitequark:
> Since the changes in the frontend are so small, could you still consider merging this, and improve on it / rewrite it later?

No.

frontends/ast/genrtlil.cc
daglem: As far as I understand `tests/tools/autotest.sh`, it is generated by the `verilog` backend. This is the output in `tests/simple/string_format.out/string_format_syn0.v` without the change above:
```
/* Generated by Yosys 0.36+79 (git sha1 97d655a20, gcc 13.2.1 -fPIC -Os) */

module string_format_top();
  \$print  #(
    .ARGS_WIDTH(32'd168),
    .FORMAT("A: {168:> c}\n"),
    .PRIORITY(32'd4294967295),
    .TRG_ENABLE(32'd0),
    .TRG_POLARITY({0{1'b0}}),
    .TRG_WIDTH(32'd0)
  ) _0_ (
    .ARGS(168'h736f6d657468696e6720696e746572657374696e67),
    .EN(1'h1),
    .TRG()
  );
  \$print  #(
    .ARGS_WIDTH(32'd168),
    .FORMAT("B: {168:> c}\n"),
    .PRIORITY(32'd4294967294),
    .TRG_ENABLE(32'd0),
    .TRG_POLARITY({0{1'b0}}),
    .TRG_WIDTH(32'd0)
  ) _1_ (
    .ARGS(168'h736f6d657468696e6720696e746572657374696e67),
    .EN(1'h1),
    .TRG()
  );
endmodule
```

daglem:
> > Since the changes in the frontend are so small, could you still consider merging this, and improve on it / rewrite it later?
> 
> No. In fact, upon further review, I think #3963 has issues that should be resolved first.

Sure, but I cherry picked only the first two commits, not the third (unfinished) one. Are there really any controversial changes in the current PR?

frontends/ast/genrtlil.cc
whitequark: Okay, so that's a bug in the Verilog backend where it does not output `$print` cells in `initial` blocks correctly. It should be fixed; any time the Verilog backend outputs a bare `$print` cell (without being explicitly requested to) is a bug in the backend.

frontends/verilog/verilog_frontend.cc
```diff
@@ -100,6 +100,10 @@ struct VerilogFrontend : public Frontend {
 		log("    -assert-assumes\n");
 		log("        treat all assume() statements like assert() statements\n");
 		log("\n");
+		log("    -nodisplay\n");
```
whitequark, line 103: `-noprint` is a better name for the option.

frontends/ast/simplify.cc
```diff
@@ -1058,33 +1058,30 @@ bool AstNode::simplify(bool const_fold, int stage, int width_hint, bool sign_hin
...
+				// when $display()/$write() functions are used in an initial block, print them during synthesis
+				Fmt fmt = processFormat(stage, /*sformat_like=*/false, default_base, /*first_arg_at=*/0, /*required=*/false);
+				if (str.substr(0, 8) == "$display")
+					fmt.append_string("\n");
+				log("%s", fmt.render().c_str());
+			}
+
+			// simplify the expressions and convert them to a special cell later in genrtlil
 			for (auto node : children)
 				while (node->simplify(true, stage, -1, false)) {}
```
whitequark, line 1082: I think you want to simplify before formatting, to maximize the chances that formatting succeeds.

whitequark:
> Are there really any controversial changes in the current PR?

I thought there is at first, but on closer look it seems like the cherry picked commits are OK functionality-wise. If you rename the variable in https://github.com/YosysHQ/yosys/pull/4129/commits/97d655a20aa27d1892d5d214e66da621f57cae9a from `required` to `may_fail` I'm fine with having it.

https://github.com/YosysHQ/yosys/pull/4129/commits/4e4d7028a5c6c9775b7d5189d4bcbac561e5ec2e should definitely not be merged.

For https://github.com/YosysHQ/yosys/pull/4129/commits/07927443bf34f741816ad99c5d39e92f2b922784, I've added inline comments.

whitequark:
Note that this comment:

> [4e4d702](https://github.com/YosysHQ/yosys/commit/4e4d7028a5c6c9775b7d5189d4bcbac561e5ec2e) should definitely not be merged.

directly relates to this comment of mine:

> I'm currently working on modifying the frontend to emit $check cells, and there is intersection with what this PR is doing (namely, generation of RTLIL for initial assert (...); and later conversion back to Verilog). I think $check and $print cells should be handled uniformly in this aspect.

So more work will need to be done on `$print` in `initial` blocks no matter what. (I'm OK merging this PR excluding the problematic commit in the meantime.)

frontends/verilog/verilog_frontend.cc
daglem: Depends on whether you think of `$print` cells or SystemVerilog "display tasks", I guess?
I can change it to `-noprint`, of course.

whitequark: There is not just `$display` but also `$write` and friends. What they do is print to the terminal, so I think `-noprint` accurately describes what the option does and is less ambiguous than `-nodisplay`.

daglem: Just a thought, as there are no `$print` cells involved during elaboration, but only "display tasks", do you still want the option changed from `-nodisplay` to `-noprint`?

whitequark: Yes, because it's not about `$print` cells, but covering `$write` as well as `$display`.

frontends/ast/simplify.cc
daglem: OK, done.

whitequark:
I've narrowed down the cause of the issues with `.TRG_POLARITY` and currently working on a fix.

daglem:
> I've narrowed down the cause of the issues with `.TRG_POLARITY` and currently working on a fix.

Great!

whitequark:
You can cherry pick https://github.com/YosysHQ/yosys/commit/cfbaf18ec045aee05ba24d3e3ea7d15cfb5d68a2 to have a fix the `.TRG_POLARITY` issue in this PR.

whitequark:
> > I'm currently working on modifying the frontend to emit $check cells, and there is intersection with what this PR is doing (namely, generation of RTLIL for initial assert (...); and later conversion back to Verilog). I think $check and $print cells should be handled uniformly in this aspect.

There is actually still a bug in the Verilog backend even after my fix: this testcase

```verilog
module string_format_top;
	parameter STR = "something interesting";
	initial begin
		$display("A: %s", STR);
		$display("B: %0s", STR);
	end
endmodule
```

is transformed via `-b verilog` to this code:

```verilog
(* hdlname = "\\string_format_top" *)
(* top =  1  *)
(* src = "tests/simple/string_format.v:1.1-7.10" *)
module string_format_top();
  (* src = "tests/simple/string_format.v:4.3-4.26" *)
  wire _0_;
  (* src = "tests/simple/string_format.v:5.3-5.27" *)
  wire _1_;
  always @*
    if (1'h1)
      $write("A: %s\n", $unsigned(168'h736f6d657468696e6720696e746572657374696e67));
  always @*
    if (1'h1)
      $write("B: %s\n", $unsigned(168'h736f6d657468696e6720696e746572657374696e67));
  assign _1_ = 1'h1;
  assign _0_ = 1'h1;
endmodule
```

This is very obviously wrong. However, since this particular bit is entirely unrelated to what this PR is doing, it doesn't need to block this PR.

frontends/verilog/verilog_frontend.cc
daglem: Sure, my point is that `$display*`, `$strobe*`, `$write*`, and `$monitor*` are all called "display tasks" in the LRM.

whitequark: Oh, I see your point now. I agree it doesn't need changing in this case.

daglem:
> You can cherry pick [cfbaf18](https://github.com/YosysHQ/yosys/commit/cfbaf18ec045aee05ba24d3e3ea7d15cfb5d68a2) to have a fix the `.TRG_POLARITY` issue in this PR.

For some reason I'm not able to cherry pick that - is there some git incantation I have to cast? I have synced my local repository, and I was able to cherry pick the out-of-tree commits above earlier.

whitequark:
I don't think you pushed it.

daglem:
> I don't think you pushed it.

Yeah, I just found out :sweat_smile:

frontends/ast/simplify.cc
```diff
@@ -145,7 +145,7 @@ void AstNode::fixup_hierarchy_flags(bool force_descend)
 
 // Process a format string and arguments for $display, $write, $sprintf, etc
 
-Fmt AstNode::processFormat(int stage, bool sformat_like, int default_base, size_t first_arg_at) {
+Fmt AstNode::processFormat(int stage, bool sformat_like, int default_base, size_t first_arg_at, bool required) {
```
whitequark, line 148: Please rename this parameter to `may_fail`. (The function would be inverted in this case.)

whitequark (changes requested):
Looks good, just needs one change I called out inline.

frontends/ast/simplify.cc
daglem: Should I edit the commit, or add a new one?

whitequark: I prefer editing the commit.

whitequark:
Please also cherry pick 55793309fd7f07f970d9a7292034d03558eae618 since this fixes the bug I described in https://github.com/YosysHQ/yosys/pull/4129#issuecomment-1886869495.

frontends/ast/simplify.cc
daglem: Done.

daglem:
> Please also cherry pick [5579330](https://github.com/YosysHQ/yosys/commit/55793309fd7f07f970d9a7292034d03558eae618) since this fixes the bug I described in [#4129 (comment)](https://github.com/YosysHQ/yosys/pull/4129#issuecomment-1886869495).

Great, done!

frontends/ast/simplify.cc
```diff
@@ -1055,30 +1058,31 @@ bool AstNode::simplify(bool const_fold, int stage, int width_hint, bool sign_hin
...
+				int default_base = 10;
+				if (str.back() == 'b')
+					default_base = 2;
+				else if (str.back() == 'o')
+					default_base = 8;
+				else if (str.back() == 'h')
+					default_base = 16;
+
+				// when $display()/$write() functions are used in an initial block, print them during synthesis
+				Fmt fmt = processFormat(stage, /*sformat_like=*/false, default_base, /*first_arg_at=*/0, /*required=*/false);
```
whitequark, line 1078: You forgot to update the parameter here:

```suggestion
				Fmt fmt = processFormat(stage, /*sformat_like=*/false, default_base, /*first_arg_at=*/0, /*may_fail=*/true);
```

daglem: Thanks! I edited it in the original commit, but brought it right back when resolving the conflict in the next commit :see_no_evil:

whitequark (approved):
LGTM

daglem:
> LGTM

Thanks for all the help!

BTW awesome work on the `$print` cell, this really brings major improvements to Yosys!

whitequark:
Thanks! It was motivated by Amaranth but I implemented it (or at least started that work) in a way that benefits everyone. I do have to say that implementing SystemVerilog semantics correctly took a *lot* of time and effort...

daglem:
> Thanks! It was motivated by Amaranth but I implemented it (or at least started that work) in a way that benefits everyone. I do have to say that implementing SystemVerilog semantics correctly took a _lot_ of time and effort...

I can only imagine! Give yourself a pat on the back, knowing that this would quite possibly never have materialized without your efforts.

daglem:
It seems like [5579330](https://github.com/YosysHQ/yosys/commit/55793309fd7f07f970d9a7292034d03558eae618) causes tests/fmt/display_lm_tb.cc to fail.

whitequark:
That happens because CXXRTL does not (yet) support `initial` statements with code. It's not clear how or whether to implement that (it would have to be exposed to the record/replay code and the C interface, for one).

daglem:
> That happens because CXXRTL does not (yet) support `initial` statements with code. It's not clear how or whether to implement that (it would have to be exposed to the record/replay code and the C interface, for one).

Can I simply temporarily comment out the `initial $display` line in the test, seeing that CXXRTL isn't ready for that yet?

whitequark:
> Edit: It seems like the `always $display` line doesn't work either.

That seems to be a bug in CXXRTL.

whitequark:
> I think it's a broken testbench.

The testbench expects both the synthesis log and the simulation log to contain the output. By commenting out `initial $display` it is no longer contained in the synthesis log, so the test fails.

daglem:
> > I think it's a broken testbench.
> 
> The testbench expects both the synthesis log and the simulation log to contain the output. By commenting out `initial $display` it is no longer contained in the synthesis log, so the test fails.

Yes, I just found a fix for this - both removing the `initial` line *and* changing a line in run-test.sh to
`../../yosys -p "read_verilog display_lm.v; hierarchy -auto-top; proc; sim" >yosys-display_lm.log` fixes it.

whitequark:
I'm not sure that's really the right fix? I'm actually looking at implementing `initial $print` in CXXRTL right now.

whitequark:
Cherry-pick 0dd012816.

daglem:
Color me impressed! :sweat_smile:

daglem:
> Please also cherry pick [5579330](https://github.com/YosysHQ/yosys/commit/55793309fd7f07f970d9a7292034d03558eae618) since this fixes the bug I described in [#4129 (comment)](https://github.com/YosysHQ/yosys/pull/4129#issuecomment-1886869495).

@whitequark Whoops, this commit also completely removes `intitial $display` output from simulation.

whitequark:
What is "simulation" here?

daglem:
> What is "simulation" here?

In this context, the `sim` command. See #4131

PR #4130, opened 2024-01-12
hierarchy: Without a known top module, derive all deferred modules

jix:
This fixes hierarchy when used with cell libraries that were loaded with -defer and also makes more of the hierarchy visible to the auto-top heuristic.


whitequark (approved):
I like this change in behavior

PR #4131, opened 2024-01-14
Restore sim output from initial $display

daglem:
Restores `sim` output from `initial $display`, which was inadvertedly removed in #4129.

### DIFF
--- a/passes/sat/sim.cc
+++ b/passes/sat/sim.cc
@@ -185,6 +185,7 @@ struct SimInstance
 
 	struct print_state_t
 	{
+		bool initial_done;
 		Const past_trg;
 		Const past_en;
 		Const past_args;
@@ -350,6 +351,7 @@ struct SimInstance
 				print.past_trg = Const(State::Sx, cell->getPort(ID::TRG).size());
 				print.past_args = Const(State::Sx, cell->getPort(ID::ARGS).size());
 				print.past_en = State::Sx;
+				print.initial_done = false;
 			}
 		}
 
@@ -852,13 +854,14 @@ struct SimInstance
 			bool triggered = false;
 
 			Const trg = get_state(cell->getPort(ID::TRG));
+			bool trg_en = cell->getParam(ID::TRG_ENABLE).as_bool();
 			Const en = get_state(cell->getPort(ID::EN));
 			Const args = get_state(cell->getPort(ID::ARGS));
 
 			if (!en.as_bool())
 				goto update_print;
 
-			if (cell->getParam(ID::TRG_ENABLE).as_bool()) {
+			if (trg.size() > 0 && trg_en) {
 				Const trg_pol = cell->getParam(ID::TRG_POLARITY);
 				for (int i = 0; i < trg.size(); i++) {
 					bool pol = trg_pol[i] == State::S1;
@@ -868,7 +871,12 @@ struct SimInstance
 					if (!pol && curr == State::S0 && past == State::S1)
 						triggered = true;
 				}
+			} else if (trg_en) {
+				// initial $print (TRG width = 0, TRG_ENABLE = true)
+				if (!print.initial_done && en != print.past_en)
+					triggered = true;
 			} else {
+				// always @(*) $print
 				if (args != print.past_args || en != print.past_en)
 					triggered = true;
 			}
@@ -889,6 +897,7 @@ struct SimInstance
 			print.past_trg = trg;
 			print.past_en = en;
 			print.past_args = args;
+			print.initial_done = true;
 		}
 
 		if (check_assertions)


daglem:
@whitequark Is this the correct fix?

daglem:
The fix is based on the amended documentation for TRG in https://github.com/YosysHQ/yosys/pull/4129/commits/1159e487217096cde0aa7a19ca4db095b19cf01f
"If the **width of this port is zero** and ``\TRG_ENABLE`` is true, the cell is triggered during initial evaluation (time zero) only."

daglem:
FYI no ill effects in sv-tests either.

whitequark (approved):

whitequark (changes requested):
Actually, no; this would make the `$print` cell trigger on each cycle of `sim`, not just the initial one.

daglem:
> Actually, no; this would make the `$print` cell trigger on each cycle of `sim`, not just the initial one.

As far as I can tell, it works as intended. E.g. running `./yosys -p "read_verilog -nodisplay -sv high.sv; proc; sim"` on this input

```
module top ();

bit [7:0] arr;

initial begin
	$display(":assert: (%d == 8)", $size(arr));
end

endmodule
```

yields

```
3. Executing SIM pass (simulate the circuit).
Simulating cycle 0.
:assert: (          8 == 8)
Simulating cycle 1.
Simulating cycle 2.
Simulating cycle 3.
Simulating cycle 4.
Simulating cycle 5.
Simulating cycle 6.
Simulating cycle 7.
Simulating cycle 8.
Simulating cycle 9.
Simulating cycle 10.
Simulating cycle 11.
Simulating cycle 12.
Simulating cycle 13.
Simulating cycle 14.
Simulating cycle 15.
Simulating cycle 16.
Simulating cycle 17.
Simulating cycle 18.
Simulating cycle 19.
Simulating cycle 20.
Simulating cycle 21.
Simulating cycle 22.
Simulating cycle 23.
Simulating cycle 24.
Simulating cycle 25.
Simulating cycle 26.
Simulating cycle 27.
Simulating cycle 28.
Simulating cycle 29.
Simulating cycle 30.
Simulating cycle 31.
Simulating cycle 32.
Simulating cycle 33.
Simulating cycle 34.
Simulating cycle 35.
Simulating cycle 36.
Simulating cycle 37.
Simulating cycle 38.
Simulating cycle 39.
Simulating cycle 40.
```

whitequark (approved):
OK. I don't understand the sim pass very well.

povik:
> Actually, no; this would make the $print cell trigger on each cycle of sim, not just the initial one.

Well, on each cycle in which the arguments change, since the initial prints are now being handled as if `TRG_ENABLE` was false, so as an `always @(*) $display`.

Consider
```
module top(input wire clk);
reg [7:0] t = 0;
always @(posedge clk) t <= t + 1;
initial begin
	$display("%x", t);
end
endmodule
```
simulated with  `./yosys -p "read_verilog -nodisplay test.v; proc; sim -clock clk"`.

You get
```
3. Executing SIM pass (simulate the circuit).
Simulating cycle 0.
00
Simulating cycle 1.
Simulating cycle 2.
01
Simulating cycle 3.
Simulating cycle 4.
02
Simulating cycle 5.
Simulating cycle 6.
03
Simulating cycle 7.
Simulating cycle 8.
...
```

daglem:
> > Actually, no; this would make the $print cell trigger on each cycle of sim, not just the initial one.
> 
> Well, on each cycle in which the arguments change, since the initial prints are now being handled as if `TRG_ENABLE` was false, so as an `always @(*) $display`.
> 

@povik I tried to remedy this now - can you please have a look?

povik:
Still doesn't do what we want in all cases, consider:

```
module top(input wire clk);
reg [7:0] t = 0;
always @(posedge clk) t <= t + 1;
initial begin
	if (t % 2)
		$display("%x", t);
end
endmodule
```

this will make the `EN` port of the `$print` cell change value, which will repeatedly trigger the cell.

povik:
I suggest we establish a general way to query if we are in the initial cycle of a simulation. As far as I can tell, there's the `set_initstate_outputs` method that's already called to set whether the next cycle is the initial one. Maybe we should generalize that, add a method e.g. `set_cycle_initial` which in turn calls `set_initstate_outputs` and updates the "is initial cycle" flag. We would then change the existing call sites from `set_initstate_outputs` to `set_cycle_initial`.

What do you say @jix? Are we free to conflate the `$initstate` output value with a flag that says "this is the initial cycle of the simulation" which would influence things like the print cells?

povik (approved):
I think we should eventually clean this up so that each print cell doesn't maintain it's own ad hoc "initial" flag.

daglem:
> I think we should eventually clean this up so that each print cell doesn't maintain it's own ad hoc "initial" flag.

I was just about to write that I propose this as an interim solution :sweat_smile: 
All the cases above work now - it would be nice if this could be merged while you discuss other possibilites.

povik:
> All the cases above work now - it would be nice if this could be merged while you discuss other possibilites.

I agree. Let's wait for CI to be happy then merge.

PR #4134, opened 2024-01-16
CXXRTL: Allow capturing `$print` cell output

whitequark:
See individual commits for detailed explanation and rationale. Although there are a few changes that are strictly speaking breaking, I don't think any are controversial. I tried to keep backwards compatibility where it made sense, e.g. `-print-output` is preserved.

### DIFF
--- a/backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
+++ b/backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
@@ -865,19 +865,12 @@ struct observer {
 	// Called when the `commit()` method for a wire is about to update the `chunks` chunks at `base` with `chunks` chunks
 	// at `value` that have a different bit pattern. It is guaranteed that `chunks` is equal to the wire chunk count and
 	// `base` points to the first chunk.
-	virtual void on_update(size_t chunks, const chunk_t *base, const chunk_t *value) = 0;
+	void on_update(size_t chunks, const chunk_t *base, const chunk_t *value) {}
 
 	// Called when the `commit()` method for a memory is about to update the `chunks` chunks at `&base[chunks * index]`
 	// with `chunks` chunks at `value` that have a different bit pattern. It is guaranteed that `chunks` is equal to
 	// the memory element chunk count and `base` points to the first chunk of the first element of the memory.
-	virtual void on_update(size_t chunks, const chunk_t *base, const chunk_t *value, size_t index) = 0;
-};
-
-// The `null_observer` class has the same interface as `observer`, but has no invocation overhead, since its methods
-// are final and have no implementation. This allows the observer feature to be zero-cost when not in use.
-struct null_observer final: observer {
-	void on_update(size_t chunks, const chunk_t *base, const chunk_t *value) override {}
-	void on_update(size_t chunks, const chunk_t *base, const chunk_t *value, size_t index) override {}
+	void on_update(size_t chunks, const chunk_t *base, const chunk_t *value, size_t index) {}
 };
 
 template<size_t Bits>
@@ -916,8 +909,7 @@ struct wire {
 
 	// This method intentionally takes a mandatory argument (to make it more difficult to misuse in
 	// black box implementations, leading to missed observer events). It is generic over its argument
-	// to make sure the `on_update` call is devirtualized. This is somewhat awkward but lets us keep
-	// a single implementation for both this method and the one in `memory`.
+	// to allow the `on_update` method to be non-virtual.
 	template<class ObserverT>
 	bool commit(ObserverT &observer) {
 		if (curr != next) {
--- a/backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
+++ b/backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
@@ -39,6 +39,7 @@
 #include <memory>
 #include <functional>
 #include <sstream>
+#include <iostream>
 
 // `cxxrtl::debug_item` has to inherit from `cxxrtl_object` to satisfy strict aliasing requirements.
 #include <cxxrtl/capi/cxxrtl_capi.h>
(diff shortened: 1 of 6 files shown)


backends/cxxrtl/runtime/cxxrtl/cxxrtl_replay.h
```diff
@@ -556,22 +556,20 @@ class recorder {
 	bool record_incremental(ModuleT &module) {
 		assert(streaming);
 
-		struct : public observer {
```
povik, line 559: I take it user-defined observers are now not supposed to derive from the generic `observer` struct.

kernel/fmt.cc
```diff
@@ -569,82 +569,61 @@ std::vector<VerilogFmtArg> Fmt::emit_verilog() const
...
-			}
-
-			default: log_abort();
+			case FmtPart::STRING:    os << "STRING";    break;
+			case FmtPart::INTEGER:   os << "INTEGER";   break;
+			case FmtPart::CHARACTER: os << "CHARACTER"; break;
+			case FmtPart::TIME:      os << "TIME";      break;
+		}
+		os << ", ";
+		// escape_cxx_string() is defined in cxxrtl_backend.cc
```
povik, line 608: This isn't the case (at least within the scope of the commit)

povik, line 608: Stray comment? We are re-declaring `escape_cxx_string` within `fmt.cc`, but this comment suggests we want to link to the declaration from`cxxrtl_backend.cc` (which is invisible because of private namespace)

backends/cxxrtl/runtime/cxxrtl/cxxrtl_replay.h
whitequark: You can do either. I removed the base class in this particular struct so that I can capture using an aggregate initializer; this is purely a consequence of a stylistic choice elsewhere.

backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
```diff
@@ -891,8 +892,21 @@ struct fmt_part {
...
+	virtual double realtime() const { return time(); }
+
+	// Called when a `$print` cell is triggered.
+	virtual void on_print(const std::string &output) { std::cout << output; }
+};
+
 // An object that can be passed to a `commit()` method in order to produce a replay log of every state change in
-// the simulation.
+// the simulation. Unlike `performer`, `observer` does not use virtual calls as their overhead is unacceptable, and
+// a comparatively heavyweight template-based solution is justified.
```
povik, line 909: Heavyweight as in compile-time heavyweight? Complexity-wise the templating doesn't seem that heavy.

povik:
For `$check` I expect we will have a mechanism for the `performer` to ask for the message to be formatted, so that the formatting is conditional and up to the performer's decision, do we want the same for `$print`s for consistence?

backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
whitequark: It requires you to have the `commit` method in the header, for one, so that it's visible to any potential call sites which need an observer. That is quite invasive compared to what was there before, which is a single virtual call (which we still have and which has a wrapper delegating to this template.)

There continues to be nothing particularly compile-time heavyweight in CXXRTL. Parsing STL includes takes about as much time as expanding a CXXRTL netlist for a RISC-V CPU.

kernel/fmt.cc
whitequark: Yeah, I forgot about the private namespace for a moment.

whitequark:
> For `$check` I expect we will have a mechanism for the `performer` to ask for the message to be formatted, so that the formatting is conditional and up to the performer's decision, do we want the same for `$print`s for consistence?

I don't think so. Because `on_print` and `on_check` are virtual, the only real option you have is a `std::function` which would result in an unavoidable heap allocation per cell per `eval()`, which is unacceptable. I think there should be a `on_check_fail(std::string name, std::string message, const metadata_map &attributes)` as well as `on_check_pass(std::string name, const metadata_map &attributes)`. This avoids the allocation.

whitequark:
And yeah, `on_print` should be amended to have the `on_print(const std::string &output, const metadata_map &attributes)` signature; I forgot about the attributes.

backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
povik: I wonder if there can be a heavily-instrumented design where the overhead of the virtual call per every enabled `$check` starts to matter, in any case it shouldn't matter for a typical design.

povik:
> I think there should be a on_check_fail(std::string name, std::string message, const metadata_map &attributes) as well as on_check_pass(std::string name, const metadata_map &attributes). This avoids the allocation.

OK, so we make the decision we only format for failing checks. If you want something else you can always duplicate the `$check` with a separate `$print` easily in preprocessing.

povik (approved):
Looks good!

backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
whitequark: This seems like something the indirect call predictor should cover well. For `commit()` though I have benchmarked it and the current implementation is just about the best I can do while still using pointers as identifiers (assigning identifiers persistently in the module would give additional 15% or so, which is desirable, but it's also kind of hard to do).

whitequark:
> If you want something else you can always duplicate the `$check` with a separate `$print` easily in preprocessing.

Yes. Or we could simply change the logic. I do want to have concrete use cases before committing to something particularly complex. For the simple pass/fail separation we do have a use case at ChipFlow so I'm OK adding that.

PR #4135, opened 2024-01-16
Fix verific clocking when no driver exist

mmicko:


nakengelhardt (approved):

tests/verific/clocking.ys
```diff
@@ -0,0 +1,12 @@
...
+always @(*) assert(foo);
+endmodule
+EOT
+
+verific -import test
+prep
+
+select -assert-count 1 t:$assert
+select -assert-count 1 t:$not
+select -assert-none t:$assert t:$not %% t:* %D
```
nakengelhardt, line 12: I think we only want to test that the $assert cell exists, we don't want to constrain the specific implementation choice of checker logic beyond that, so I would remove lines 11-12 from the test.

nakengelhardt:
For reference this was caused by an update in Verific introduced with the Dec 1 release.

Before the change, the netlist from verific looked like this:
Oct23_SW_Release

After the change, it is optimized down to:
Dec23_SW_Release

tests/verific/clocking.ys
```diff
@@ -0,0 +1,10 @@
+read -sv <<EOT
+module test(input clk, foo);
```
nakengelhardt, line 2: Oh, the `clk` input is unnecessary for this test, it was just there for debugging so I could switch quickly between `always @(*)` and `always @(posedge clk)`

mmicko: Updated

Issue #4098, opened 2023-12-31
Gowin BSRAM inference -- We need B suffix (such as DPB instead of current DP)


chili-chips:
### Version

35

### On which OS did this happen?

Linux

### Reproduction Steps

https://github.com/YosysHQ/apicula/issues/208#issuecomment-1872635263

### Expected Behavior

It's been a while since Gowin has added B suffix to their BSRAM primitives, to differentiate them from S(hadow) SRAM. 
_nextpnr-gowin_ (project Apicula) has updated the naming. However, Yosys seems to still be using the original nomenclature, creating a disconnect.

We need BSRAM to be inferred as DP**B**, not DP.

### Actual Behavior

BSRAM is inferred as DP instead of DPB.

yrabbit:
A patch that purely mechanically corrects memory synthesis. Now there will be less swearing when creating images, but this should not make them work.
[b.path.txt](https://github.com/YosysHQ/yosys/files/13868571/b.path.txt)


A good image is with direct use of the DPB primitive.

https://github.com/YosysHQ/yosys/assets/6075465/6746b0c4-dba3-40c1-9d1a-c1f99a81ebbe



Not a very good image - when we infer.

https://github.com/YosysHQ/yosys/assets/6075465/200afeaa-22e3-4b53-a685-7002d40971cb

And now a little of my stupid reasoning as a newbie.
When we use a primitive directly, we set its parameters (data width, address width, write mode, read mode, etc.), and also explicitly connect signals to the inputs of the primitive - that is, we decide what to connect to ClockEnable and what to WriteEnable.

``` verilog
    DPB mem(
        .DOB({dummy, read_data}),
        .DIA({{8{gnd}}, write_data}),
        .DIB({16{gnd}}),
        .ADA({write_ad, gnd, gnd, gnd}),
        .ADB({ read_ad, gnd, gnd, gnd}),
        .CLKA(write_clk),
        .CLKB(clk),
        .OCEA(vcc),
        .OCEB(vcc),
        .CEA(write_ce),
        .CEB(vcc),
        .WREA(vcc),
        .WREB(read_wre),
        .BLKSELA(3'b000),
        .BLKSELB(3'b000),
        .RESETA(write_reset),
        .RESETB(reset)
    );
    defparam mem.READ_MODE0 = 1'b1;
    defparam mem.READ_MODE1 = 1'b1;
    defparam mem.WRITE_MODE0 = 2'b01;
[...]

PR #4137, opened 2024-01-19
gowin: fix the BRAM mapping.

yrabbit:
The primitives used have been corrected and changes have been made to the set of signals. It has been empirically established that it is necessary to set the OCEx signal to 1 when using READ_MODE=0.

Addresses issue https://github.com/YosysHQ/yosys/issues/4098

Thanks Saviour Owolabi for the help.


mmicko (approved):
LGTM

PR #4139, opened 2024-01-19
Rename TIME to VLOG_TIME in formatting code

whitequark:
See commit messages for detailed justification.

Amaranth would probably want to use a different formatting operation that uses absolute `cxxrtl::time` rather than Verilog's timescale-relative `$time`/`$realtime`.


backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
```diff
@@ -1001,16 +1001,16 @@ struct fmt_part {
 	bool signed_; // = false;
 	bool plus; // = false;
 
-	// TIME type
+	// VLOG_TIME type
 	bool realtime; // = false;
```
povik, line 1005: This is only weakly related to this PR but still: isn't `realtime` a confusing name for this? I know the Verilog function is called `$realtime`, but I thought this had something do to with real time until I peeked into the spec just now. Could be e.g. `nonquantized`

povik (commented):
I am not sure about this. I understand the motivation but I am not convinced having a suitably generalized "time" support in formatting isn't a better option. Could we parse out the `timescale` from Verilog in the frontend and imprint it in the RTLIL format specifier for each occurrence of `$time`/`$realtime`? As I understand it, at least with CXXRTL here, if an user wants up-to-spec handling of `$time`/`$realtime`, they are under obligation to supply performer `vlog_time`/`vlog_realtime` returning values in the right scale.

Since we have the CXXRTL time type introduced recently, what about having a method returning that on the performer, and covering all the use cases this way?

backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
whitequark: > isn't `realtime` a confusing name for this? I know the Verilog function is called `$realtime`, but I thought this had something do to with real time until I peeked into the spec just now. Could be e.g. `nonquantized`

It is. I think I misunderstand what it does when I added it a few years ago when I first wrote the Fmt prototype.

povik:
> Whenever someone who is interested in that offers a patch we can switch to a single unified time support.

Fair enough, though are you willing to break the API once that happens?

whitequark:
> are you willing to break the API once that happens?

Yeah. It's not a part of the C API so the API contract states that it can be changed without notice. (I would want this to be done before exposing time in the C API.)

PR #4150, opened 2024-01-23
write_verilog: don't emit code with dangling else related to wrong condition

whitequark:
It is embarrassing to have written it, but also, three independent reviewers missed it too :D

See https://github.com/amaranth-lang/amaranth/issues/1049.

Issue introduced in #4090

### DIFF
--- a/backends/verilog/verilog_backend.cc
+++ b/backends/verilog/verilog_backend.cc
@@ -1949,13 +1949,8 @@ void dump_conn(std::ostream &f, std::string indent, const RTLIL::SigSpec &left, 
 
 void dump_proc_switch(std::ostream &f, std::string indent, RTLIL::SwitchRule *sw);
 
-void dump_case_body(std::ostream &f, std::string indent, RTLIL::CaseRule *cs, bool omit_trailing_begin = false)
+void dump_case_actions(std::ostream &f, std::string indent, RTLIL::CaseRule *cs)
 {
-	int number_of_stmts = cs->switches.size() + cs->actions.size();
-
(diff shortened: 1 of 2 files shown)


whitequark:
I can't reproduce the test failure locally for some reason. That's odd.

povik:
Oh no

povik:
In any case I feel like we need some tests here.

povik:
> I can't reproduce the test failure locally for some reason. That's odd.

That looks like the issue we saw yesterday due to iverilog updating, but since your PR is based on #4148 the CI here should have been pinned to a good iverilog revision already. @mmicko ideas?

povik:
@whitequark OK with you if I push to your branch? I would rebase (to fix CI), and maybe add some test.

whitequark:
Go ahead please.

povik:
Rebased

backends/verilog/verilog_backend.cc
```diff
@@ -1996,36 +1982,52 @@ bool dump_proc_switch_ifelse(std::ostream &f, std::string indent, RTLIL::SwitchR
 		}
 	}
 
+	dump_attributes(f, indent, sw->attributes);
```
povik, line 1985: So this is something we were missing before, unrelated to the fix

povik (approved):
Looks fine

backends/verilog/verilog_backend.cc
whitequark: We were dumping slightly different attributes (for case rules and not switch itself), the switch is probably better.

whitequark:
Thanks @povik!

Issue #2746, opened 2021-05-06
In CXXRTL edge eval is before calculating value


korneliuszo:
I have code: https://github.com/korneliuszo/nhci/tree/aa6cbe84b3df7d755972e31baf6ef20d714c9a23

and generated code doesn't trigger posedge in spi.v
	always @(posedge IOWR)
	begin
			IOWR_data = D_in;
	end

IOWR here needs calculation, when i delete it in 
https://github.com/korneliuszo/nhci/blob/aa6cbe84b3df7d755972e31baf6ef20d714c9a23/fw/top.v#L69
It latches value.

In generated code there is:
bool p_sim::eval() {
	bool converged = true;
	bool posedge_i_flatten_5c_top_2e__24_or_24_top_2e_v_3a_69_24_48__Y = this->posedge_i_flatten_5c_top_2e__24_or_24_top_2e_v_3a_69_24_48__Y();

and then
	i_flatten_5c_top_2e__24_or_24_top_2e_v_3a_69_24_48__Y = or_uu<1>(p_top_2e_IOWR, not_u<1>(p_top_2e_CONFIGURED));

so as i understand edge doesn't have a chance to be detected

korneliuszo:
Attaching unit test
[2746.tar.gz](https://github.com/YosysHQ/yosys/files/6433790/2746.tar.gz)
Should return 1 but returns 0.
After moving:
	bool posedge_i_and_24_top_2e_v_3a_13_24_1__Y = this->posedge_i_and_24_top_2e_v_3a_13_24_1__Y();
after
	p_C = i_and_24_top_2e_v_3a_13_24_1__Y;
it works

modwizcode:
So I've taken a look at this issue. First I'd like to cover a few issues in the design that are likely to bring trouble in general, and especially so combined with things like cxxrtl (and real hardware for that matter).

First of all, I would recommend never mixing blocking `=` assignments and non-blocking `<=` in a single block. In general I don't recommend blocking assignments at all (especially not in clocked blocks), they tend to infer latches easily and mostly make things harder. This is especially true when working with cxxrtl.

Secondly, doing `@(posedge signal)` when that signal is not a clock is also likely to produce bad behaviors. In this case I seem to notice you both look for a low-high transition using the `IOWR_ev` registers and (for `IOWR`) you seem to be trying to also use the posedge. 

[...]

PR #4162, opened 2024-01-25
sim: Bring $print trigger/sampling semantics in line with FFs

jix:
While trying to add more tests to get better coverage for the code I added for #4128 (which had quite a few bugs not uncovered by our current tests) I noticed that `sim`'s sampling behavior of enable and arguments for `$print` with a trigger does not match how we treat edge triggered FFs. In particular that makes it impossible to use `sim` to test that `clk2fflogic` doesn't change the semantics of a design that should be supported by `clk2fflogic`.

I think it could also lead to quite confusing behavior when for Verilog inputs, inside an edge triggered process there is an assignment to a register next to a print of the same expression that is assigned to the register. ~~I haven't yet tried to reproduce that exact scenario (I've focused on the more complex `clk2fflogic` testcases) but I think the previous behavior could show a mismatch of what is printed and what assignment is made (or even whether an assignment is made) while the new behavior should match.~~ In simple cases and basically for everything you would intentionally write as part of a design, it makes no difference though, you actually need to have an enable or input signal changing simultaneously with the trigger signal, which you would want to generally avoid as it's a race condition. FV still needs to provide consistent behavior across all engines and simulations within the same flow, though, and for code written as part of a formal test bench there might even be valid uses of this.

I briefly looked at what cxxrtl does, but with #2746 (and maybe related issues) AFAICT it currently doesn't support the testcases that I need to exercise clk2fflogic. ~~I do plan to perform some simpler tests to make sure that sim and cxxrtl agree on what is printed for those.~~ All test cases where this change makes a difference would run into #2746 and I think even with that fixed cxxrtl might consider those to be race conditions and not want to specify the exact behavior. 

@povik in any case I wanted to get some early feedback from you, in case I'm missing something that prompted the current design and would break with my changes

I'm not sure about the `trg_on_gclk` attribute, but while testing `clk2fflogic` I realized I need some way to have a `$print` without trigger produce output at most once every gclk cycle while also outputting every gclk cycle it is enabled, even when no enable or argument changed (essentially to emulate a one full update per fixed cycle simulation).

The alternative would be to agree on a way to explicitly represent that by some parametrization of the `$print` cell, but given that the exact behavior of a trigger-less side effect is inherently coupled to the simulation strategy (e.g. also when there are glitches) and the fact that `sim` isn't a fixed cycle simulation but it tries to emulate one for the FV flows, I think using an attribute to switch between two behaviors that I'd consider to be in the scope of the allowed implementation defined behaviors for a trigger-less `$print` cell is not the worst solution.

### DIFF
--- a/passes/sat/sim.cc
+++ b/passes/sat/sim.cc
@@ -813,7 +813,7 @@ struct SimInstance
 		}
 	}
 
-	void update_ph3(bool check_assertions)
+	void update_ph3(bool gclk_trigger)
 	{
 		for (auto &it : ff_database)
 		{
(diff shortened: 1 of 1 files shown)


whitequark:
I think I'm OK with the attribute.

jix:
For something like

```sv
always @(*)
    derived_from_clock <= !clk;

always @(posedge clk) begin
    if (counter >= 1 && counter <= 10) begin
        $display("snapshot is %0d;", snapshot);
    end;

    if (counter == 2 || counter == 5) begin
        snapshot <= counter + derived_from_clock;
        $display("snapshot <= %0d;", counter + derived_from_clock);
    end
end
```

the current sim code actually assigns a different value to `snapshot` than it prints, even though both are in the same block. This is fixed by this PR so that the value printed matches the value assigned. What you get with SV simulation actually depends on whether you use `derived_from_clock <= !clk` or `derived_from_clock = !clk` / directly use `!clk`, which is a difference Yosys cannot and should not capture, but in both cases the assigned and printed value is consistent during a simulation. For FV where we need to define the behavior also for all race conditions, we already made the choice for FFs (in `sim` and `clk2fflogic`) so I think bringing the behavior for `$print` and other side effects in line with that choice shouldn't cause new problems elsewhere.

passes/sat/sim.cc
```diff
@@ -858,49 +858,53 @@ struct SimInstance
...
-
-			if (trg.size() > 0 && trg_en) {
-				Const trg_pol = cell->getParam(ID::TRG_POLARITY);
-				for (int i = 0; i < trg.size(); i++) {
-					bool pol = trg_pol[i] == State::S1;
-					State curr = trg[i], past = print.past_trg[i];
-					if (pol && curr == State::S1 && past == State::S0)
+			bool sampled = trg_en && trg.size() > 0;
+
+			if (sampled ? print.past_en.as_bool() : en.as_bool()) {
```
povik, line 863: This got a bit hard to read, I think we should rewrite this (not necessarily in this PR), so that we have:
```
    bool triggered = /* is the cell triggered (not necessarily enabled)? */
    bool enabled = /* EN appropriately sampled */
    if (triggered && enabled)
      /* text rendering */
```

povik:
> I'm not sure about the `trg_on_gclk` attribute, but while testing `clk2fflogic` I realized I need some way to have a `$print` without trigger produce output at most once every gclk cycle while also outputting every gclk cycle it is enabled, even when no enable or argument changed (essentially to emulate a one full update per fixed cycle simulation).

Is the attribute a debugging aid, or would it be set by `clk2fflogic` ordinarily?

povik:
As, I think the attribute would be there so we can lower a cell with a clock trigger, that's e.g. enabled for a few consecutive cycles without its arguments changing, to a trigger-less `$print`, right?

povik (approved):

jix:
> Ah, I think the attribute would be there so we can lower a cell with a clock trigger, that's e.g. enabled for a few consecutive cycles without its arguments changing, to a trigger-less `$print`, right?

That's the exact use case

povik:
> In simple cases and basically for everything you would intentionally write as part of a design, it makes no difference though, you actually need to have an enable or input signal changing simultaneously with the trigger signal, which you would want to generally avoid as it's a race condition.

I think this could make a difference for a practical design if it combines posedge and negedge triggering

jix:
> I think this could make a difference for a practical design if it combines posedge and negedge triggering

I don't see how that would make a difference. Do you have a concrete example in mind?

povik:
Sure!

```
read_verilog <<EOF
module top(clk);
	input wire clk;
	reg [7:0] a = 0;
	reg [7:0] b = 0;

	always @(posedge clk)
		a <= a + 1;

	always @(negedge clk)
		b <= b + 1;

	always @(posedge clk)
		if (a == 4)
			$display("%d", b);
endmodule
EOF

proc
sim -clock clk -q
```

Prints 3 on master, 4 on your branch

PR #4163, opened 2024-01-25
write_verilog: Making sure BUF cells are converted to expressions.

QuantamHD:
These were previously not being converted correctly leading to yosys internal cells being written to my netlist.

### DIFF
--- a/backends/verilog/verilog_backend.cc
+++ b/backends/verilog/verilog_backend.cc
@@ -1058,7 +1058,7 @@ bool dump_cell_expr(std::ostream &f, std::string indent, RTLIL::Cell *cell)
 		f << stringf("%s" "assign ", indent.c_str());
 		dump_sigspec(f, cell->getPort(ID::Y));
 		f << stringf(" = ");
-		dump_attributes(f, "", cell->attributes, ' ');
+		dump_attributes(f, "", cell->attributes, " ");
 		dump_cell_expr_port(f, cell, "A", false);
 		f << stringf(";\n");
 		return true;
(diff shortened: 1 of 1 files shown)


backends/verilog/verilog_backend.cc
```diff
@@ -1053,6 +1053,16 @@ bool dump_cell_expr(std::ostream &f, std::string indent, RTLIL::Cell *cell)
 		f << stringf(";\n");
 		return true;
 	}
+	
+	if (cell->type == ID($_BUF_)) {
+		f << stringf("%s" "assign ", indent.c_str());
+		dump_sigspec(f, cell->getPort(ID::Y));
+		f << stringf(" = ");
+		dump_attributes(f, "", cell->attributes, " ");
```
povik, line 1061: Will these attributes get attached to anything? For e.g. `$_NOT_` we dump the attributes as a suffix to the `~` operator, so they get attached to that, but here I don't know if they are of any significance.

povik (approved):
Looks good, the one comment is more of a call for discussion with anyone who wants to chime in

backends/verilog/verilog_backend.cc
povik: `assign a = (* foo *) b;` turns out to be invalid Verilog. Since it doesn't look like there's any place else appropriate that we can put the attribute, we will simply need to drop it. (`opt_clean` would drop the attribute anyway when it cleans up `$_BUF_` cells.)

povik (changes requested):
As is, this can generate invalid Verilog. Please remove the `dump_attributes` call then we can merge this.

povik (approved):
Thanks!

PR #4165, opened 2024-01-26
peepopt: handle offset too large in `shiftadd`

### LINKED ISSUE #4160: BUG：Issue in Yosys Synthesis: 'std::length_error' Leads to Termination
### Version

yosys 0.35+56

Linux

### Reproduction Steps

Hello,

While undergoing the synthesis procedure using Yosys, I encountered a complication when processing the design file (rtl.v). The execution of the read_verilog rtl.v command resulted in an error, as indicated in the provided screenshot. This issue arose during the optimization and synthesis phases of the design using Yosys tools. Specifically, the error pertained to a 'std::length_error,' causing the termination of the synthesis process.
image

Synthesis processes as follows:
read_verilog rtl.v         
synth
write_verilog syn_yosys.v

Please find attached the code of RTL .
Thank you in advance for your attention to this matter.
I look forward to hearing from you regarding this issue.
[yosys_project_01.zip](https://github.com/YosysHQ/yosys/files/14045610/yosys_project_01.zip)

### Expected Behavior

synthesis success

### Actual Behavior

synthesis fail

phsauter:
Fixes #4160 

If the offset is larger than the signal itself, meaning the signal is completely shifted out,
it tried to extract a negative amount of bits from the old signal.

This RTL pattern is suspicious since it is a complicated way of arriving at a constant value, so we warn the user.

### DIFF
--- a/passes/pmgen/peepopt_shiftadd.pmg
+++ b/passes/pmgen/peepopt_shiftadd.pmg
@@ -103,8 +103,18 @@ code
 		new_a.append(old_a);
 	} else {
 		// data >> (...+c) transformed to data[MAX:c] >> (...)
-		new_a.append(old_a.extract_end(offset));
+		if(offset < GetSize(old_a)) // some signal bits left?
+			new_a.append(old_a.extract_end(offset));
+		else {
+			if(shift->type.in($shiftx))
+				log_warning("at %s: result of indexed part-selection is always constant (selecting from '%s' with index '%s + %d')\n", \
+							shift->get_src_attribute().c_str(), log_signal(old_a), log_signal(var_signal), offset);
+			else
+				log_warning("at %s: result of shift operation is always constant (shifting '%s' by '%s + %d'-bits)\n", \
+							shift->get_src_attribute().c_str(), log_signal(old_a), log_signal(var_signal), offset);
+		}
 
+		// is it fine to leave new_a empty? (size 0)
 	}
 
 	SigSpec new_b = {var_signal, SigSpec(State::S0, log2scale)};
--- a/passes/pmgen/peepopt_shiftadd.pmg
+++ b/passes/pmgen/peepopt_shiftadd.pmg
@@ -103,15 +103,19 @@ code
 		new_a.append(old_a);
 	} else {
 		// data >> (...+c) transformed to data[MAX:c] >> (...)
-		if(offset < GetSize(old_a)) // some signal bits left?
+		if(offset < GetSize(old_a)) { // some signal bits left?
 			new_a.append(old_a.extract_end(offset));
-		else {
+		} else {
+			// warn user in case data is empty (no bits left)
+			std::string location = shift->get_src_attribute();
+			if (location.empty())
+				location = shift->name.str();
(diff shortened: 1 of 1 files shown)


passes/pmgen/peepopt_shiftadd.pmg
```diff
@@ -103,8 +103,18 @@ code
...
+		else {
+			if(shift->type.in($shiftx))
+				log_warning("at %s: result of indexed part-selection is always constant (selecting from '%s' with index '%s + %d')\n", \
+							shift->get_src_attribute().c_str(), log_signal(old_a), log_signal(var_signal), offset);
+			else
+				log_warning("at %s: result of shift operation is always constant (shifting '%s' by '%s + %d'-bits)\n", \
+							shift->get_src_attribute().c_str(), log_signal(old_a), log_signal(var_signal), offset);
+		}
+
+		// is it fine to leave new_a empty? (size 0)
```
povik, line 117: Zero-width ports should be fine. Please remove the comment and we can go ahead with merging this.

phsauter:
Removed the unnecessary comment, should be ready to merge.

passes/pmgen/peepopt_shiftadd.pmg
```diff
@@ -103,8 +103,16 @@ code
 		new_a.append(old_a);
 	} else {
 		// data >> (...+c) transformed to data[MAX:c] >> (...)
-		new_a.append(old_a.extract_end(offset));
-
+		if(offset < GetSize(old_a)) // some signal bits left?
+			new_a.append(old_a.extract_end(offset));
```
povik, line 107: Best to put the block here in curly braces if the else clause has curly braces.

passes/pmgen/peepopt_shiftadd.pmg
```diff
@@ -103,8 +103,16 @@ code
...
+		else {
+			if(shift->type.in($shiftx))
+				log_warning("at %s: result of indexed part-selection is always constant (selecting from '%s' with index '%s + %d')\n", \
```
povik, line 110: `shift->get_src_attribute()` is not guaranteed not to be empty, please add minimal handling of this case. At least for example
```
    std::string location = shift->get_src_attribute();
    if (location.empty())
        location = shift->name;
```
then use `location` in the phrasing. I think I saw other places in Yosys handle this the same way.

povik (changes requested):
Sorry @phsauter, missed some things when I looked at it last. See the comments.

passes/pmgen/peepopt_shiftadd.pmg
phsauter: deleted comment and made it explicit that it should be empty

passes/pmgen/peepopt_shiftadd.pmg
phsauter: Handled this case, the warning is not the cleanest if the src-attribute is missing but I think it is acceptable, especially because without it a 'good' warning is hard to do anyway.

povik (approved):

PR #1654, opened 2020-01-24
verific: unflatten struct ports

eddiehung:
Fixes YosysHQ/SymbiYosys#69.

Previously:
```
$ ./yosys -Qp "verific -sv bug69.sv; verific -import -all; dump i:in*"

-- Running command `verific -sv bug69.sv; verific -import -all; dump i:in*' --

1. Executing VERIFIC (loading SystemVerilog and VHDL designs using Verific).
Built with Verific Dec19_SW_Release, released at Tue Dec 31 23:22:23 2019.
VERIFIC-COMMENT [VERI-1482] Analyzing Verilog file 'bug69.sv'

2. Executing VERIFIC (loading SystemVerilog and VHDL designs using Verific).
Built with Verific Dec19_SW_Release, released at Tue Dec 31 23:22:23 2019.
Running hier_tree::ElaborateAll().
VERIFIC-INFO [VERI-1018] bug69.sv:6: compiling module 'mcve'
Importing module mcve.
Importing module \$unit_bug69_sv.


  attribute \src "bug69.sv:6"
  wire width 2 input 1 \in
```

Now:
```
$ ./yosys -Qp "verific -sv bug69.sv; verific -import -all; dump i:in*"

-- Running command `verific -sv bug69.sv; verific -import -all; dump i:in*' --

1. Executing VERIFIC (loading SystemVerilog and VHDL designs using Verific).
Built with Verific Dec19_SW_Release, released at Tue Dec 31 23:22:23 2019.
VERIFIC-COMMENT [VERI-1482] Analyzing Verilog file 'bug69.sv'

2. Executing VERIFIC (loading SystemVerilog and VHDL designs using Verific).
Built with Verific Dec19_SW_Release, released at Tue Dec 31 23:22:23 2019.
Running hier_tree::ElaborateAll().
VERIFIC-INFO [VERI-1018] bug69.sv:6: compiling module 'mcve'
Importing module mcve.
Importing module \$unit_bug69_sv.


  attribute \src "bug69.sv:6"
  wire input 2 \in.other

  attribute \src "bug69.sv:6"
  wire input 1 \in.element
```

PR #4172, opened 2024-01-30
Revert change introduced in #1654

mmicko:


gatecat:
Is it worth making this configurable? I can see reasons why the previous behaviour might be wanted, too.

mmicko:
@gatecat yap, we are already changing that

PR #4174, opened 2024-01-30
Add API to overwrite existing pass from plugin

clairexen:


povik:
Code looks good but the Python binding interacts badly with it:
```
[ 99%] Building kernel/python_wrappers.o
kernel/python_wrappers.cc:417:16: error: 'replace_existing_pass' marked 'override' but does not override any member functions
                virtual bool replace_existing_pass() override;
                             ^
kernel/python_wrappers.cc:417:16: warning: 'YOSYS_PYTHON::Pass::replace_existing_pass' hides overloaded virtual function [-Woverloaded-virtual]
./kernel/register.h:73:15: note: hidden overloaded virtual function 'Yosys::Pass::replace_existing_pass' declared here: different qualifiers ('const' vs unqualified)
        virtual bool replace_existing_pass() const { return false; }
                     ^
1 warning and 1 error generated.
```

povik:
It's not pretty but I pushed a fix to the binding generator to handle this.

The below loads:

```
import libyosys as ys

class StatPass(ys.Pass):
    def __init__(self):
        super().__init__("stat", "")

    def py_replace_existing_pass(self):
        return True

    def py_execute(self, args, design):
        pass

p = StatPass()
```

PR #4175, opened 2024-02-01
Improve `opt_merge` performance

povik:
See the commits. This replaces the hashing implementation, and makes it ignore cells holding memory initialization data.

Ping @rmlarsen since they were interested in `opt_merge` performance before.

### DIFF
--- a/passes/opt/opt_merge.cc
+++ b/passes/opt/opt_merge.cc
@@ -105,7 +105,7 @@ struct OptMergeWorker
 			}
 			conn_hash = mkhash_xorshift64(a.hash()) + acc;
 		} else {
-			for (auto conn : cell->connections())
+			for (const auto &conn : cell->connections())
 			if (!cell->output(conn.first))
 				conn_hash += mkhash_xorshift64(mkhash(conn.first.hash(), assign_map(conn.second).hash()));
 
@@ -114,7 +114,7 @@ struct OptMergeWorker
 		}
 
 		uint64_t param_hash = 0;
-		for (auto &it : cell->parameters)
+		for (const auto &it : cell->parameters)
 			param_hash += mkhash_xorshift64(mkhash(it.first.hash(), it.second.hash()));
 
 		return conn_hash + mkhash_xorshift64(param_hash + cell->type.hash());
--- a/passes/opt/opt_merge.cc
+++ b/passes/opt/opt_merge.cc
@@ -223,14 +223,14 @@ struct OptMergeWorker
 		while (did_something)
 		{
 			std::vector<RTLIL::Cell*> cells;
-			cells.reserve(module->cells_.size());
-			for (auto &it : module->cells_) {
-				if (!design->selected(module, it.second))
+			cells.reserve(module->cells().size());
+			for (auto cell : module->cells()) {
+				if (!design->selected(module, cell))
 					continue;
-				if (mode_keepdc && has_dont_care_initval(it.second))
+				if (mode_keepdc && has_dont_care_initval(cell))
 					continue;
-				if (ct.cell_known(it.second->type) || (mode_share_all && it.second->known()))
-					cells.push_back(it.second);
+				if (ct.cell_known(cell->type) || (mode_share_all && cell->known()))
+					cells.push_back(cell);
 			}
 
 			did_something = false;
--- a/passes/opt/opt_merge.cc
+++ b/passes/opt/opt_merge.cc
@@ -227,6 +227,11 @@ struct OptMergeWorker
 			for (auto cell : module->cells()) {
 				if (!design->selected(module, cell))
 					continue;
+				if (cell->type.in(ID($meminit), ID($meminit_v2), ID($mem), ID($mem_v2))) {
+					// Ignore those for performance: meminit can have an excessively large port,
+					// mem can have an excessively large parameter holding the init data
+					continue;
+				}
 				if (mode_keepdc && has_dont_care_initval(cell))
 					continue;
 				if (ct.cell_known(cell->type) || (mode_share_all && cell->known()))
--- a/passes/opt/opt_merge.cc
+++ b/passes/opt/opt_merge.cc
@@ -83,7 +83,8 @@ struct OptMergeWorker
 
 		if (cell->type.in(ID($and), ID($or), ID($xor), ID($xnor), ID($add), ID($mul),
 				ID($logic_and), ID($logic_or), ID($_AND_), ID($_OR_), ID($_XOR_))) {
-			conn_hash = assign_map(cell->getPort(ID::A)).hash() + assign_map(cell->getPort(ID::B)).hash();
+			conn_hash = hashlib::mkhash_xorshift64(assign_map(cell->getPort(ID::A)).hash()) \
+							+ hashlib::mkhash_xorshift64(assign_map(cell->getPort(ID::B)).hash());
 		} else if (cell->type.in(ID($reduce_xor), ID($reduce_xnor))) {
 			SigSpec a = assign_map(cell->getPort(ID::A));
 			a.sort();
@@ -100,13 +101,13 @@ struct OptMergeWorker
 			SigSpec s = assign_map(cell->getPort(ID::S));
 			for (int i = 0; i < cell->getParam(ID::S_WIDTH).as_int(); i++) {
 				SigSpec b_window = b.extract(i * width, width);
-				acc += mkhash(s[i].hash(), b_window.hash());
+				acc += mkhash_xorshift64(mkhash(s[i].hash(), b_window.hash()));
 			}
-			conn_hash = mkhash(a.hash(), acc);
+			conn_hash = mkhash_xorshift64(a.hash()) + acc;
 		} else {
 			for (auto conn : cell->connections())
 			if (!cell->output(conn.first))
-				conn_hash += mkhash(conn.first.hash(), assign_map(conn.second).hash());
+				conn_hash += mkhash_xorshift64(mkhash(conn.first.hash(), assign_map(conn.second).hash()));
 
 			if (RTLIL::builtin_ff_cell_types().count(cell->type))
 				conn_hash += initvals(cell->getPort(ID::Q)).hash();
@@ -114,9 +115,9 @@ struct OptMergeWorker
 
 		uint64_t param_hash = 0;
 		for (auto &it : cell->parameters)
-			param_hash += mkhash(it.first.hash(), it.second.hash());
+			param_hash += mkhash_xorshift64(mkhash(it.first.hash(), it.second.hash()));
 
-		return mkhash(conn_hash, param_hash) + cell->type.hash();
+		return conn_hash + mkhash_xorshift64(param_hash + cell->type.hash());
 	}
 
 	bool compare_cell_parameters_and_connections(const RTLIL::Cell *cell1, const RTLIL::Cell *cell2)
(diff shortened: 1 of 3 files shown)


rmlarsen:
Cool. I'll run out benchmarks on it when I get a chance.

rmlarsen:
For our circuit, this reduces the time spent in OptMerge from ~25.5s to ~21.8s, so a ~17% speedup.

Flame graph before:
image

Flame graph after:
image

rmlarsen:
@povik yeah, that is odd. Is the `mkhash` function not mixing the hashes sufficiently?

passes/opt/opt_merge.cc
```diff
@@ -77,78 +77,46 @@ struct OptMergeWorker
...
-					s += "{" + chunk.wire->name.str() + " " +
-							int_to_hash_string(chunk.offset) + " " +
-							int_to_hash_string(chunk.width) + "}";
-				else
-					s += RTLIL::Const(chunk.data).as_string();
-			}
-			hash_conn_strings.push_back(s + "\n");
+			conn_hash = mkhash(a.hash(), acc);
+		} else {
+			for (auto conn : cell->connections())
```
rmlarsen, line 107: If you use `const auto&` instead of `auto` here, it shaves off another couple of seconds on my benchmark, which is significant.

rmlarsen:
@povik yeah `mkhash` creates a simple 32 bit hash:

```
// The XOR version of DJB2
inline unsigned int mkhash(unsigned int a, unsigned int b) {
	return ((a << 5) + a) ^ b;
}
```

You end up combining a lot of those potentially. So the hashing is definitely weaker.

QuantamHD:
I wonder if pulling in xxhash or cityhash would be an improvement here.

rmlarsen:
BTW: This code is really weird. `unsigned int` is 32 bit per the C++ standard. https://github.com/YosysHQ/yosys/blob/1b73b5beb7e407089c0b5db56a2d25e6be64c911/kernel/hashlib.h#L39

passes/opt/opt_merge.cc
rmlarsen, line 107: I don't quite grok what's going on here, but it seems quite different from the original hashing of connections. Could this be the source of more collisions?

povik:
> BTW: This code is really weird. `unsigned int` is 32 bit per the C++ standard.
> 
> https://github.com/YosysHQ/yosys/blob/1b73b5beb7e407089c0b5db56a2d25e6be64c911/kernel/hashlib.h#L39

Right, that doesn't make much sense.

povik:
Turns out the collisions are due to a combination of `mkhash` being a weak mixing primitive, and `Const` hashing not working at all (`hash()` for those returns a constant due to a coding error). Let me collect my changes and push.

nakengelhardt:
> BTW: This code is really weird. `unsigned int` is 32 bit per the C++ standard.

Wait, what? ILP64 may not be popular but since when does the C++ standard ban it?

povik:
> > BTW: This code is really weird. `unsigned int` is 32 bit per the C++ standard.
> 
> Wait, what? ILP64 may not be popular but since when does the C++ standard ban it?

Second closer look at https://en.cppreference.com/w/cpp/language/types indeed confirms the standard allows int be 64-bit, the page just doesn't list it as an option in the "common data models" table. I will revert the `mkshift_xorshift` change.

povik:
> I will revert the `mkhash_xorshift` change.

On second thought those ILP64 being so rare there's no point keeping a special branch for them, if the default branch is basically fine too. I think I will keep the change in.

nakengelhardt:
It's currently the only way to get around the limitation on design size, which people do run into somewhat regularly. https://github.com/YosysHQ/yosys/blob/d00843d436f4ebaed19c4a49f6e5dc78960fdf00/kernel/hashlib.h#L203-L204

I don't know if anyone has taken this path in practice, but from past conversations in issues I suspect some people patched yosys locally to replace "int" with "long". We might want to go that way long-term too... we've stayed away from it so far out of fear of introducing more subtle issues that our tests won't be able to uncover, but given how we keep finding that the hash wasn't working properly in the first place maybe we should just go for it.

rmlarsen:
@nakengelhardt @povik if this is indeed something people struggle with in the Yosys, I'd highly recommend going with the standard types int32_t and int64_t etc. We don't have to suffer these indignities anymore.

nakengelhardt:
> @nakengelhardt @povik if this is indeed something people struggle with in the Yosys, I'd highly recommend going with the standard types int32_t and int64_t etc. We don't have to suffer these indignities anymore.

How does that help, what benefit do you get from requiring a precise size rather than a minimum size here?

rmlarsen:
> > @nakengelhardt @povik if this is indeed something people struggle with in the Yosys, I'd highly recommend going with the standard types int32_t and int64_t etc. We don't have to suffer these indignities anymore.
> 
> How does that help, what benefit do you get from requiring a precise size rather than a minimum size here?

Well defined semantics of your code, perhaps?

povik:
But it doesn't seem like the lack of defined semantics is the issue here, or something anyone is struggling with. It seems like the underspecification of `int` is a feature right now, there's `sizeof(int) == 4` which is the well-tested case, and `sizeof(int) == 8` for adventurers who are hitting limits otherwise. Of course eventually moving to `X` with fixed `sizeof(X) == 8` in the code where people are hitting limits is desirable.

rmlarsen:
> But it doesn't seem like the lack of defined semantics is the issue here, or something anyone is struggling with. It seems like the underspecification of `int` is a feature right now, there's `sizeof(int) == 4` which is the well-tested case, and `sizeof(int) == 8` for adventurers who are hitting limits otherwise. Of course eventually moving to `X` with fixed `sizeof(X) == 8` in the code where people are hitting limits is desirable.

Sure. But in this instance, the code is less readable and has different behavior on different platforms, which should be avoided unless absolutely necessary IMHO.

nakengelhardt:
Yes, we do need configurability of the hash size for different use cases. The current way of doing it has just fallen out of favor since (or possibly before) this code was first written. Ideally we would have 64b hash be the default on 64b architectures, but we need to maintain the option of 32b hashes for smaller/slower architecture targets where the performance hit matters (people running yosys on an older raspi are still a thing, and then there's that proof-of-concept someone made of a softcore doing synthesis to partially reconfigure the FPGA it is running on, which we want to enable just for the heck of it). So there's no way around it that any caller of the function has to work for all sizes anyway.

rmlarsen:
> Yes, we do need configurability of the hash size for different use cases. The current way of doing it has just fallen out of favor since (or possibly before) this code was first written. Ideally we would have 64b hash be the default on 64b architectures, but we need to maintain the option of 32b hashes for smaller/slower architecture targets where the performance hit matters (people running yosys on an older raspi are still a thing, and then there's that proof-of-concept someone made of a softcore doing synthesis to partially reconfigure the FPGA it is running on, which we want to enable just for the heck of it). So there's no way around it that any caller of the function has to work for all sizes anyway.

I see. [This is C++, so this is usually done with overload, not by adding a runtime check for `sizeof`, but there appears to be a lot of C code in Yosys.]

QuantamHD:
What's the purpose of this custom hash function in the first place? It seems like adopting an existing function that works on char* streams would both be platform independent, and faster than what we've done here.

rmlarsen:
> What's the purpose of this custom hash function in the first place? It seems like adopting an existing function that works on char* streams would both be platform independent, and faster than what we've done here.

Moreover, in the code in question, it is also being used to combine hashes for which this hash function is very poor. There should probably be a separate hash function for that purpose. IFAICT, the DJB2 is a very simple hashing for strings.

kernel/rtlil.h
```diff
@@ -712,7 +712,7 @@ struct RTLIL::Const
 	inline unsigned int hash() const {
 		unsigned int h = mkhash_init;
 		for (auto b : bits)
-			mkhash(h, b);
+			h = mkhash(h, b);
```
rmlarsen, line 715: good catch!

povik:
DJB2 didn't make the leaderboard? Notwithstanding we are probably using it wrong in the changed `opt_merge` code and in few other places in `hashlib.h`.

whitequark:
> Ideally we would have 64b hash be the default on 64b architectures, but we need to maintain the option of 32b hashes for smaller/slower architecture targets where the performance hit matters (people running yosys on an older raspi are still a thing, and then there's that proof-of-concept someone made of a softcore doing synthesis to partially reconfigure the FPGA it is running on, which we want to enable just for the heck of it).

Has anyone actually benchmarked this and found that the hash function is an issue? Or is this conjecture?

rmlarsen:
> > Ideally we would have 64b hash be the default on 64b architectures, but we need to maintain the option of 32b hashes for smaller/slower architecture targets where the performance hit matters (people running yosys on an older raspi are still a thing, and then there's that proof-of-concept someone made of a softcore doing synthesis to partially reconfigure the FPGA it is running on, which we want to enable just for the heck of it).
> 
> Has anyone actually benchmarked this and found that the hash function is an issue? Or is this conjecture?

It is not a hotspot in my benchmarks, but this was measured on Intel Skylake x86_64. The main issue in this context was the poor hashing, which caused a significant number of hash collisions. But I think Martin fixed the primary source of that (lack of hashing of constants).

povik:
I assume Catherine's question was about hashing in general all over Yosys code, and the associated `dict<>` and `pool<>` operations which consume the hashes. Was that what you looked at, Rasmus, or did you mean `opt_merge` specifically?

> But I think Martin fixed the primary source of that (lack of hashing of constants).

FWIW the poor properties of `mkhash` contributed the most in my testing, which went away when I inserted `mkhash_xorshift64` in few key places in the PR here. It may not be the best approach but it's at least something we can improve on.

rmlarsen:
> I assume Catherine's question was about hashing in general all over Yosys code, and the associated `dict<>` and `pool<>` operations which consume the hashes. Was that what you looked at, Rasmus, or did you mean `opt_merge` specifically?
> 
> > But I think Martin fixed the primary source of that (lack of hashing of constants).
> 
> FWIW the poor properties of `mkhash` contributed the most in my testing, which went away when I inserted `mkhash_xorshift64` in few key places in the PR here. It may not be the best approach but it's at least something we can improve on.

Ah yes. Hashing in general still dominates the profile:

image

rmlarsen:
> > I assume Catherine's question was about hashing in general all over Yosys code, and the associated `dict<>` and `pool<>` operations which consume the hashes. Was that what you looked at, Rasmus, or did you mean `opt_merge` specifically?
> > > But I think Martin fixed the primary source of that (lack of hashing of constants).
> > 
> > 
> > FWIW the poor properties of `mkhash` contributed the most in my testing, which went away when I inserted `mkhash_xorshift64` in few key places in the PR here. It may not be the best approach but it's at least something we can improve on.
> 
> Ah yes. Hashing in general still dominates the profile:
> 
> image
> 
> FWIW, we turned on more ABC passes since my original PRs, so ignore those. But in yosys proper, the largest time sinks are hashing related.

I believe the hot spots in my profile are related to the non-coherent layout of the hash tables in yosys, not time spent in the hash function itself. Replacing the hash table implementation with something like the swisstables from Abseil would likely give a significant speedup. I have not had the time to experiment with this, as it is a rather invasive change.  https://abseil.io/about/design/swisstables

rmlarsen:
@povik here is the flame graph for most of the Yosys cost. It's a mix of OptMergePass/OptExprPass/CleanupPass/OptMuxTreePass.

image

povik:
Superseded by Emil's work in #4677 making opt_merge use a new hashing interface

PR #4176, opened 2024-02-01
Improve `opt_expr` performance

povik:
See the commits. This avoids having to populate `dict<RTLIL::Cell*, std::set<RTLIL::SigBit>> cell_to_inbit;`, and it saves repeated conversions of type IDs to strings (36 times per cell for one iteration of the loop in `opt_expr`). Allocating a string for the IDs was significant enough to stand out when watching where `opt_expr` spends its time.

### DIFF
--- a/passes/opt/opt_expr.cc
+++ b/passes/opt/opt_expr.cc
@@ -409,6 +409,9 @@ void replace_const_cells(RTLIL::Design *design, RTLIL::Module *module, bool cons
 		}
 	}
 
+	CellTypes ct_memcells;
+	ct_memcells.setup_stdcells_mem();
+
 	if (!noclkinv)
 	for (auto cell : module->cells())
 	if (design->selected(module, cell)) {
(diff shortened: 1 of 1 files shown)


widlarizer:
With `-noopt` hacked into `proc` as called by `techmap` in the first part of the failing test, before `opt_expr` would be called `check` can be called instead and it does reveal a logic loop

widlarizer:
Since

https://github.com/YosysHQ/yosys/blob/1e401c3e04b813bb8f9091d326f047ce5d58906e/techlibs/xilinx/cells_sim.v#L3900

and

https://github.com/YosysHQ/yosys/blob/1e401c3e04b813bb8f9091d326f047ce5d58906e/techlibs/xilinx/cells_sim.v#L3726

and `Pd` is constructed from expressions that combinationally depend on `Z`, a comb loop is apparently generated. However, it is unreachable, since one of the two statements I linked depends on different values of the parameter `PREG`. The inactive statements only get resolved by opt, which is what the failing test actually tests for

widlarizer:
I think if you're okay with it we should just supress the warning with `-expect`. See commit de799783 for how we got to the current version of `techlibs/xilinx/cells_sim.v`

povik:
Ah so it was the new warning which was tripping up the tests. I am fine with suppressing it to pass the test, but I think we should discuss whether we want the warning in the first place.

In the meantime we have changed `check` to allow coarse loops, and those would trigger the warning too. So we need to account for that in the wording.

widlarizer (approved):

PR #4178, opened 2024-02-01
check: Extend loop handling

povik:
* Add `-allow-loops` mode

 * Print loops in-order and with source attribute, if available


povik:
Some of the changes here will go in with #4184, the `-allow-loops` change will need an extra PR later.

PR #4179, opened 2024-02-01
sim: Add `-assert-cover` option to assert covers were reached

povik:
A feature for small automated tests around `sim`


povik:
As pointed out by @jix this doesn't distinguish between cover cells within the same module at different points in the hierarchy.

PR #4183, opened 2024-02-05
CI Improvements

KrystalDelusion:
Refactor linux/mac tests into unified build and test stages.  Builds are performed out-of-tree, compressing and uploading the compiled artifacts to be used in the test stage.  The included test suite is run independently from the docs tests.  Additional smoke tests for compiler/cpp standard are done separately with a new `compile-only` target.  In all, this should provide a much cleaner idea at a glance of what part of the infrastructure is failing.
Also updates to use latest OS and compiler.


povik:
I like this change. I saw people get confused by the fact that only some of the "Build and run tests" CI jobs fail, which made it seem as if the test failure was compiler-dependent when in fact it was not. So if this change splits the building and testing into jobs of their own, that will help present the CI results better.

mmicko:
It would be better if we could reuse `compiler testing` tasks to also produce artifacts for testing (just for two that needs to be used), looking into that also noticed macOS tests are missing

KrystalDelusion:
> It would be better if we could reuse `compiler testing` tasks to also produce artifacts for testing (just for two that needs to be used)

I had it setup like that initially (157a94a), but I noticed under [Details](https://github.com/YosysHQ/yosys/actions/runs/7795342458/job/21258370528) that it skipped running the tests because the `C++20` build was failing, even though the job creating the artifact worked fine.  I did some research and it doesn't seem like it allows specifying a specific job from a matrix as a prerequisite, so I added 533fea6 to split the test builds from the compiler test matrix.

> looking into that also noticed macOS tests are missing

Added back in now.

mmicko (approved):
This all looks great now.

Since those gcc and clang versions do exist for ubuntu-22.04 as well, can we use that one (ubuntu-latest will get promoted to ubuntu-24.04 as it arrives), but 20.04 could be deprecated soon, so if that works please do change it). Maybe just gcc-10 would need to be left on 20.04.

Small nitpick is missing empty line at action.yml at the end.

KrystalDelusion:
Surprisingly gcc-10 has builds for 24.04, but clang-13 and earlier stop at 22.04 so we will need to update our clang target if we want to use `ubuntu-latest` for the `cpp_std` test matrix.

mmicko:
I have cleaned few things. Main change is that trigger for action should be all branches (non-filtered), VisualStudio action is updated to latest to get rid of warning (NodeJS version) and for regular build/test we always should use latest, so that change is made for ubuntu.
The biggest problem left is, I think, number of jobs that are running. Current state is 18 jobs to do all the builds, and with this PR it rises to 30 jobs. Be aware that when someone make a PR and that is on branch on YosysHQ/yosys it will run 60 jobs and that will take a while. We need to filter this list in compiler tests, since that one is mostly affecting number.
Also I would suggest that for macos (since those are slower and less available machines) we have just compiler test for c++17 in test-compile, since build is already checking c++11

KrystalDelusion:
Maybe we could split tests so that only the cpp standards run on PRs as a smoke test, and maybe the full range of compilers run periodically (maybe weekly or something)

PR #4189, opened 2024-02-08
hashlib: Add missing `stdint.h` include

povik:
We use `uint32_t` `uint64_t` etc. so add an explicit include.


Ravenslofty:
you're welcome :p

PR #4190, opened 2024-02-08
Follow the XDG Base Directory Specification

mmicko:
Expands https://github.com/YosysHQ/yosys/pull/4093 PR by moving code and display messages only if used with debug parameter, since yosys is fully functional without setting variables, and also history is only useful in interactive mode, and logs could spam users using it for scripting.


nakengelhardt:
As a side effect, if `-g` is not used and it can't create the file, this now prints `<suppressed ~1 debug messages>` after the first command that you run. This is a bit misleading since it happened before the first command, should we call `log_suppressed()` before, e.g. in `yosys_setup()`? That would print it before the first prompt which is more suitable IMO:
```
   <header snipped>

 Yosys 0.37+90 (git sha1 2797d6756, clang 15.0.0 -fPIC -Os)

<suppressed ~1 debug messages>

yosys>

nakengelhardt:
Actually, maybe `driver.cc:593` (after `load_plugin`) would be an even better place to put it, that way it accommodates plugins using `log_debug` during registration as well.

PR #4192, opened 2024-02-08
fmt: Extend string handling for SystemVerilog 

povik:
This makes for a distinction between a string argument from a quoted literal, and a string argument from a variable or other expression, which is something that's required for the SV spec.

### DIFF
--- a/frontends/ast/genrtlil.cc
+++ b/frontends/ast/genrtlil.cc
@@ -771,8 +771,6 @@ struct AST_INTERNAL::ProcessGenerator
 					if (node->type == AST_CONSTANT && node->is_string) {
 						arg.type = VerilogFmtArg::STRING;
 						arg.str = node->bitsAsConst().decode_string();
-						// and in case this will be used as an argument...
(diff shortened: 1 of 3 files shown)


frontends/ast/genrtlil.cc
```diff
@@ -771,8 +771,6 @@ struct AST_INTERNAL::ProcessGenerator
 					if (node->type == AST_CONSTANT && node->is_string) {
 						arg.type = VerilogFmtArg::STRING;
 						arg.str = node->bitsAsConst().decode_string();
-						// and in case this will be used as an argument...
```
widlarizer, line 774: Shouldn't an equivalent change be made in the other STRING emitting code in `AstNode::processFormat` in `simplify.cc`?

widlarizer:
In case you want to revive this, this is what I'd suggest:

- [ ] attach test case if this functionality is exposed, clarify what this is preparing the codebase for if not
- [ ] fix or clarify relation to `AstNode::processFormat`
- [ ] rebase

PR #4197, opened 2024-02-10
stat: Add sequential area output to stat -liberty

QuantamHD:
Checks to see if a cell is of type ff in the liberty, and keeps track of an additional area value.

```
   Chip area for module '\addr': 92.280720
   Sequential area for module '\addr': 38.814720
```

### DIFF
--- a/passes/cmds/stat.cc
+++ b/passes/cmds/stat.cc
@@ -256,7 +256,7 @@ struct statdata_t
 		if (area != 0) {
 			log("\n");
 			log("   Chip area for %smodule '%s': %f\n", (top_mod) ? "top " : "", mod_name.c_str(), area);
-			log("   Sequential area for %smodule '%s': %f\n", (top_mod) ? "top " : "", mod_name.c_str(), sequential_area);
+			log("     of which used for sequential elements: %f (%.2f%%)\n", sequential_area, 100.0*sequential_area/area);
 		}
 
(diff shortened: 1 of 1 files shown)


mmicko (commented):
Code looks fine, and I also checked liberty files documentation to check if there is anything missing, but all looks fine.
However, I would suggest renames to better represent structures.

cell_properties - > cell_area (so back as it was)
cell_data_t -> cell_area_t
is_flip_flop -> is_sequential

QuantamHD:
Done.

mmicko (approved):
Looks good to me now

passes/cmds/stat.cc
```diff
@@ -244,6 +256,7 @@ struct statdata_t
 		if (area != 0) {
 			log("\n");
 			log("   Chip area for %smodule '%s': %f\n", (top_mod) ? "top " : "", mod_name.c_str(), area);
+			log("   Sequential area for %smodule '%s': %f\n", (top_mod) ? "top " : "", mod_name.c_str(), sequential_area);
```
nakengelhardt, line 259: I think the text could make it clearer that these two are not cumulative, maybe something like this:
```suggestion
			log("     of which used for sequential logic: %f (%.2f%%)\n", sequential_area, 100.0*sequential_area/area);
```

PR #4201, opened 2024-02-11
verilog: indirect AST_CONCAT and AST_TO_UNSIGNED port connections

### LINKED ISSUE #4158: Assert `arg->is_signed == sig.as_wire()->is_signed' failed
### Version

Yosys 0.37 (git sha1 a5c7f69ed, gcc 10.2.1-6 -fPIC -Os)

Linux

### Reproduction Steps

get test case
[chirp_driver_bug.txt](https://github.com/YosysHQ/yosys/files/14042077/chirp_driver_bug.txt)
and rename as chirp_driver_bug.v

At the yosys prompt
`read_verilog chirp_driver_bug.v`

### Expected Behavior

```
1. Executing Verilog-2005 frontend: chirp_driver_bug.v
Parsing Verilog input from `chirp_driver_bug.v' to AST representation.
Generating RTLIL representation for module `\chirp_driver_bug'.
```

I can get this behavior by removing the useless curly-braces on line 19.

### Actual Behavior

```
1. Executing Verilog-2005 frontend: chirp_driver_bug.v
Parsing Verilog input from `chirp_driver_bug.v' to AST representation.
Generating RTLIL representation for module `\chirp_driver_bug'.
ERROR: Assert `arg->is_signed == sig.as_wire()->is_signed' failed in /redacted/yosys/frontends/ast/genrtlil.cc:2067.
```

zachjs:
- AST_CONCAT and AST_TO_UNSIGNED are always unsigned, but may generate RTLIL that exclusively reference a signed wire.
- AST_CONCAT may also contain a memory write.

Fixes #4158.

### DIFF
--- a/frontends/ast/genrtlil.cc
+++ b/frontends/ast/genrtlil.cc
@@ -2124,11 +2124,9 @@ RTLIL::SigSpec AstNode::genRTLIL(int width_hint, bool sign_hint)
 						if (sig.is_wire()) {
 							// if the resulting SigSpec is a wire, its
 							// signedness should match that of the AstNode
-							if (arg->type == AST_IDENTIFIER && arg->id2ast && arg->id2ast->is_signed && !arg->is_signed)
-								// fully-sliced signed wire will be resolved
-								// once the module becomes available
-								log_assert(attributes.count(ID::reprocess_after));
-							else
+							// unless this instantiation depends on module
+							// information that isn't available yet
+							if (!attributes.count(ID::reprocess_after))
 								log_assert(arg->is_signed == sig.as_wire()->is_signed);
 						} else if (arg->is_signed) {
 							// non-trivial signed nodes are indirected through
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -1292,6 +1292,12 @@ bool AstNode::simplify(bool const_fold, int stage, int width_hint, bool sign_hin
 				else if (contains_unbased_unsized(value))
 					// unbased unsized literals extend to width of the context
 					lookup_suggested = true;
+				else if (value->type == AST_TO_UNSIGNED)
+					// inner expression may be signed by default
+					lookup_suggested = true;
+				else if (value->type == AST_CONCAT && value->children.size() == 1)
+					// concat of a single expression is equivalent to $unsigned
+					lookup_suggested = true;
 			}
 		}
 
--- a/tests/simple/memwr_port_connection.sv
+++ b/tests/simple/memwr_port_connection.sv
@@ -5,9 +5,11 @@ module producer(
 endmodule
 
 module top(
-    output logic [3:0] out
+    output logic [3:0] out0, out1
 );
-    logic [3:0] v[0:0];
-    producer p(v[0]);
-    assign out = v[0];
+    logic [3:0] v[1:0];
+    producer p0(v[0]);
+    producer p1({v[1]});
+    assign out0 = v[0];
+    assign out1 = v[1];
 endmodule
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -5,6 +5,12 @@ List of major changes and improvements between releases
 Yosys 0.38 .. Yosys 0.39-dev
 --------------------------
 
+ * Verilog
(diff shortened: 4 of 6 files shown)


frontends/ast/simplify.cc
```diff
@@ -1292,6 +1292,12 @@ bool AstNode::simplify(bool const_fold, int stage, int width_hint, bool sign_hin
 				else if (contains_unbased_unsized(value))
 					// unbased unsized literals extend to width of the context
 					lookup_suggested = true;
+				else if (value->type == AST_TO_UNSIGNED)
+					// inner expression may be signed by default
+					lookup_suggested = true;
+				else if (value->type == AST_CONCAT && value->children.size() == 1)
+					// concat of a single expression is equivalent to $unsigned
+					lookup_suggested = true;
```
povik, line 1300: OK, so suggesting a lookup here will make us indirect all port connections that are not connections to a plain identifier.

frontends/ast/genrtlil.cc
```diff
@@ -2124,11 +2124,9 @@ RTLIL::SigSpec AstNode::genRTLIL(int width_hint, bool sign_hint)
...
 							// signedness should match that of the AstNode
-							if (arg->type == AST_IDENTIFIER && arg->id2ast && arg->id2ast->is_signed && !arg->is_signed)
-								// fully-sliced signed wire will be resolved
-								// once the module becomes available
-								log_assert(attributes.count(ID::reprocess_after));
-							else
+							// unless this instantiation depends on module
+							// information that isn't available yet
+							if (!attributes.count(ID::reprocess_after))
 								log_assert(arg->is_signed == sig.as_wire()->is_signed);
```
povik, line 2130: So what we are checking here is: does the wire we are connecting have an `is_signed` flag matching the signedness of the expression in the connection?

tests/simple/memwr_port_connection.sv
```diff
@@ -5,9 +5,11 @@ module producer(
...
 module top(
-    output logic [3:0] out
+    output logic [3:0] out0, out1
 );
-    logic [3:0] v[0:0];
-    producer p(v[0]);
-    assign out = v[0];
+    logic [3:0] v[1:0];
+    producer p0(v[0]);
+    producer p1({v[1]});
```
povik, line 12: I am not clear on how the code change interacts with this example. We seem to be addressing signedness mismatches, but there's not a signed variable in sight here.

povik: Ah, I think this has to do with the fact that memory writes need to be indirected, for reasons unrelated to signedness, but if wrapped in an `AST_CONCAT` they did not trigger the indirection previously.

povik: But if that's the case, shouldn't we trigger the indirection in all cases of `AST_CONCAT`, not only those with `children.size() == 1`?

frontends/ast/simplify.cc
zachjs: Correct. Is this problematic? Should I clarify the surrounding code?

frontends/ast/genrtlil.cc
zachjs: That's correct. Should I clarify this comment?

tests/simple/memwr_port_connection.sv
zachjs: You're right! I'll address this.

frontends/ast/genrtlil.cc
povik: No need I think, I am primarily checking my understanding of what happens here, and maybe leaving some notes for anyone else who would look at the PR.

frontends/ast/simplify.cc
povik: Well, the intent wasn't really obvious the first moment I looked at this `if (type == AST_CELL) {` branch, so any clarification would be nice. Though I wouldn't consider it required for the PR.

povik (approved):

povik:
@zachjs I've rebased your changes but now there's a CI failure. Do you have capacity to investigate?

ShinyKate:
@povik any idea of the status on this one?

povik:
@ShinyKate I've lost track of the status. I see @widlarizer picking the PR up

PR #4210, opened 2024-02-13
read_verilog: correctly format `hdlname` attribute value

whitequark:
The leading slash is not a part of the attribute as it only concerns public values.

No idea why this fails tests. @povik, do you know?

### DIFF
--- a/frontends/ast/ast.cc
+++ b/frontends/ast/ast.cc
@@ -1820,7 +1820,7 @@ std::string AstModule::derive_common(RTLIL::Design *design, const dict<RTLIL::Id
 
 	AstNode *new_ast = ast->clone();
 	if (!new_ast->attributes.count(ID::hdlname))
-		new_ast->set_attribute(ID::hdlname, AstNode::mkconst_str(stripped_name));
+		new_ast->set_attribute(ID::hdlname, AstNode::mkconst_str(stripped_name.substr(1)));
 
 	para_counter = 0;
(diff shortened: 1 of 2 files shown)


povik:
> No idea why this fails tests. @povik, do you know?

Not by looking at the logs, but will try reproduce locally.

povik:
@whitequark See `AstNode::try_pop_module_prefix()`

whitequark:
> See `AstNode::try_pop_module_prefix()`

Do you know what the correct behavior would be? I tried patching that and had no luck.

povik:
@whitequark Pushed what I think we want there. Also amended `uniquify` which seems to have been inserting backslashes.

Looking there's e.g. `yw.h` with some hdlname logic, it would be good to get a confirmation from @jix that this change won't break anything on the formal side. Though if anything this brings the native Verilog frontend in line with Verific, so I reckon there must be low chance of that.

povik:
> Also amended uniquify which seems to have been inserting backslashes.

`log_id` doesn't put the backslash in, so the original code was fine. Forced pushed to drop that commit.

jix:
The hierarchical names in yw files include the initial backslash (so that it can also support private names without ambiguity or escaping, at least in non-flattened designs) and I have been under the assumption that it's not included and not supposed to be included in the hdlname attribute, thus any code that deals with yw and hdlname should translate as appropriate, if it doesn't that's a bug.

PR #4211, opened 2024-02-14
clk2fflogic: Fix handling of $check cells

jix:
Fixes a bug in the handling of the recently introduced $check cells. Both $check and $print cells in clk2fflogic are handled by the same code and the existing tests for that were only using $print cells. This missed a bug where the additional A signal of $check cells that is not present on $print cells was dropped due to a typo, rendering $check cells non-functional.

Also updates the tests to explicitly cover both cell types such that they would have detected the now fixed bug.


povik (approved):

povik:
Great we got some tests to cover this!

PR #4215, opened 2024-02-16
Address race in `xprop` tests

povik:
Ought to fix the CI failures which prompted iverilog pinning in #4148 / ddfd867.

### DIFF
--- a/tests/xprop/test.py
+++ b/tests/xprop/test.py
@@ -271,7 +271,7 @@ if "prepare" in steps:
 
         for pattern in patterns:
             print(
-                f'    gclk = 1; #0; A[0] = 1\'b{pattern[-1]}; #0; A = {input_width}\'b{pattern}; #5; gclk = 0; $display("%b %b", A, Y); #5',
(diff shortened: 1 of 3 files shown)


tests/xprop/test.py
```diff
@@ -270,8 +270,9 @@ def vcdextract(signals, on_change, file, output, limit=None):
         print("initial begin", file=tb_file)
 
         for pattern in patterns:
+            # A[0] might be the clock which requires special sequencing
             print(
-                f'    gclk = 1; #0; A[0] = 1\'b{pattern[-1]}; #0; A = {input_width}\'b{pattern}; #5; gclk = 0; $display("%b %b", A, Y); #5',
+                f'    #0; gclk = 1; #0; A[0] = 1\'b{pattern[-1]}; #0; A = {input_width}\'b{pattern}; #5; gclk = 0; $display("%b %b", A, Y); #5',
```
povik, line 275: Here we generate Verilog code to be put into an `initial` block. We seem to need the `#0` to defer the first `gclk = 1;` statement, so by the time it executes the initial state has been fully propagated through continuous assignments spread around the module. I couldn't find a part of the Verilog specification which would comment on that (specifically what I am interested in is whether the `initial` blocks and assignment of register initial values are put on the active queue at the same time, or one after the other).

PR #4218, opened 2024-02-17
proc_rom: don't assert on big actionless switch.

kivikakk:
See the test case.  PROC_ROM will consider this for evaluation, even though -- without any actions -- lhs is empty (but still "uniform"). A zero-width memory is constructed, which later fails check with:

```
ERROR: Assert `width != 0' failed in kernel/mem.cc:518.
```

Ensure we don't proceed if there's nothing to encode.

---

Alternatively, do we want to reject this switch as meaningless? This came up in an Amaranth (`188eb8d4`) project; I've yet to investigate deeply why it's produced this, but either way Yosys probably doesn't want to give the assert it currently does here.

### DIFF
--- a/passes/proc/proc_rom.cc
+++ b/passes/proc/proc_rom.cc
@@ -66,6 +66,11 @@ struct RomWorker
 			}
 		}
 
+		if (lhs.empty()) {
+			log_debug("rejecting switch: lhs empty\n");
+			return;
+		}
+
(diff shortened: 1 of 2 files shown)


whitequark:
> Alternatively, do we want to reject this switch as meaningless?

RTLIL does allow empty switches.

whitequark (approved):

PR #4219, opened 2024-02-18
dfflibmap: Add a -dont_use flag to ignore cells

rovinski:
This is an alternative to setting the `dont_use` property in lib. This brings `dfflibmap` in parity with the abc pass for `-dont_use`.

This feature is extremely useful in the case where Liberty files are not easily modifiable, such as when they are stored on a read-only network drive.

This is my first PR here, so I am very open to feedback!

### DIFF
--- a/passes/techmap/dfflibmap.cc
+++ b/passes/techmap/dfflibmap.cc
@@ -124,7 +124,7 @@ static bool parse_pin(LibertyAst *cell, LibertyAst *attr, std::string &pin_name,
 
 static int glob_match(const char *pattern, const char *string) {
 	#ifdef _WIN32
-	return PathMatchSpec(string, pattern);
+	return PathMatchSpecA(string, pattern);
 	#else
 	return fnmatch(pattern, string, 0) == 0;
 	#endif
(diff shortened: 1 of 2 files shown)


rovinski:
I (hopefully) fixed compilation on Windows and also added a simple test case.

QuantamHD:
@nakengelhardt This one would be useful to us, and @rovinski  is a known collaborator with the OpenROAD project..

rovinski:
Windows build issue resolved - all tests pass now.

passes/techmap/dfflibmap.cc
```diff
@@ -115,7 +122,15 @@ static bool parse_pin(LibertyAst *cell, LibertyAst *attr, std::string &pin_name,
 	return false;
 }
 
-static void find_cell(LibertyAst *ast, IdString cell_type, bool clkpol, bool has_reset, bool rstpol, bool rstval)
+static int glob_match(const char *pattern, const char *string) {
```
povik, line 125: Please use `patmatch` from `kernel/yosys.cc` for this, if possible

rovinski: I didn't see that, thanks. Opened #4223.

PR #4223, opened 2024-02-20
dfflibmap: use patmatch() from kernel/yosys.cc

rovinski:
Replace OS matching functions with yosys kernel function.

Addresses feedback in https://github.com/YosysHQ/yosys/pull/4219#discussion_r1495765618 @povik


povik (commented):
Looks good, please also pick a revert of 20dbc860 into the PR if you don't mind.

rovinski:
Understood, thanks!

PR #4228, opened 2024-02-22
synth: Introduce `-extra-map` for amending techmap

### LINKED ISSUE #4220: Have a way to inject extra techmapping rules when calling `synth`
### Feature Description

A great power of customization would be exposed if we had a way to inject user-written "techmap" rules into the internal `synth` script. This should take the form of a command option to `synth`.

povik:
Closes #4220

### DIFF
--- a/techlibs/common/synth.cc
+++ b/techlibs/common/synth.cc
@@ -88,6 +88,10 @@ struct SynthPass : public ScriptPass {
 		log("        read/write collision\" (same result as setting the no_rw_check\n");
 		log("        attribute on all memories).\n");
 		log("\n");
+		log("    -inject filename\n");
+		log("        inject rules from the given file to complement the default\n");
+		log("        mapping library in the `techmap` step. this option can be\n");
+		log("        repeated.\n");
 		log("\n");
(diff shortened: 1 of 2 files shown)


povik:
Quoting the help message for the new option:

```
    -inject filename
        inject rules from the given file to complement the default
        mapping library in the `techmap` step. this option can be
        repeated.
```

povik:
TODO: rename to `-extra-map`

techlibs/common/synth.cc
```diff
@@ -261,7 +270,17 @@ struct SynthPass : public ScriptPass {
 			run("opt -fast -full");
 			run("memory_map");
 			run("opt -full");
-			run("techmap");
+			if (help_mode) {
+				run("techmap", "                  (unless -inject)");	
+				run("techmap -map +/techmap.v -map <inject>", "  (if -inject)");
```
nakengelhardt, line 275:
```suggestion
				run("techmap", "                  (unless -extra-map)");	
				run("techmap -map +/techmap.v -map <inject>", "  (if -extra-map)");
```

povik: Thanks!

PR #4232, opened 2024-02-23
opt_mem, memory_*: Refuse to operate in presence of processes

povik:
Processes can contain `MemWriteAction` entries which are invisible to most passes operating on memories but which will be lowered to write ports later on by `proc_memwr`. For that reason we can get corrupted RTLIL if we sequence the memory passes before `proc`. Address that by making the affected memory passes ignore modules with processes.

Also correct the log header printed by `memoy_memx`.

### DIFF
--- a/passes/memory/memory_collect.cc
+++ b/passes/memory/memory_collect.cc
@@ -39,6 +39,9 @@ struct MemoryCollectPass : public Pass {
 		log_header(design, "Executing MEMORY_COLLECT pass (generating $mem cells).\n");
 		extra_args(args, 1, design);
 		for (auto module : design->selected_modules()) {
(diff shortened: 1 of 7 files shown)


povik:
This is what would happen before this fix:
```
$ yosys -p "read_verilog ../picorv32/picorv32.v; hierarchy -top picorv32; memory; prep"

...

4.2.12. Executing OPT_EXPR pass (perform const folding).
Optimizing module picorv32.
<suppressed ~826 debug messages>
ERROR: Assert `memories.count(it.second->parameters.at(ID::MEMID).decode_string())' failed in kernel/rtlil.cc:1986.
```

PR #4233, opened 2024-02-23
Apply patch for Makefile and begin supporting nix flakes build system

RCoeurjoly:
To build yosys in nixpkgs (https://github.com/NixOS/nixpkgs/blob/nixos-23.11/pkgs/development/compilers/yosys/default.nix), a patch is applied to Makefile. I applied that same patch to the Makefile and was surprised that all the pipelines pass (https://github.com/RCoeurjoly/yosys/actions).

This PR would fix https://github.com/YosysHQ/yosys/issues/2011.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -215,7 +215,7 @@ ABC_ARCHFLAGS += "-DABC_NO_RLIMIT"
 endif
 
 ifeq ($(CONFIG),clang)
-CXX = clang
+CXX = clang++
 LD = clang++
 CXXFLAGS += -std=$(CXXSTD) -Os
 ABCMKARGS += ARCHFLAGS="-DABC_USE_STDINT_H -Wno-c++11-narrowing $(ABC_ARCHFLAGS)"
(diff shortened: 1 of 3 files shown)


whitequark:
> This PR would fix #2011.

No it wouldn't, because of https://github.com/YosysHQ/yosys/pull/4233/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L241.

povik:
Hah, I happen to have written that patch quite some time ago.

whitequark:
I would also suggest splitting the PR in two: one patching the Makefile, and another introducing Nix files. The former shouldn't be too controversial nowadays (I hope) while the latter will definitely need a discussion on the next dev meeting, mainly because it should be verified on CI that the flake still builds, and it's unclear who's going to be responsible for that in YosysHQ.

RCoeurjoly:
Sounds reasonable. Let me close this PR and open two separate PR.

PR #4235, opened 2024-02-24
Add autowires in genblk/for expension

ylm:
Improves Yosys' handling of implicit wires in blocks by appropriately prefixing them as if they were declared in that context.
Previously, Yosys would expand the blocks without verifying the presence of implicit wires which would prevent prefixing. This could lead to improper connectivity.


frontends/ast/simplify.cc
```diff
@@ -4585,6 +4586,93 @@ void AstNode::expand_genblock(const std::string &prefix)
 			}
 			break;
 
+		case AST_IDENTIFIER:
```
zachjs, line 4589: It looks like this duplicates two large chunks of non-trivial code. The first comes from earlier in `expand_genblock`, and the second from the existing implicit wire logic in `simplify`. I worry that code duplication will hurt the maintainability of the frontend, especially in an area that can be finicky (as you know). Do you think it would be possible to avoid this duplication, potentially with appropriate helper functions?

tests/verilog/genblk_wire.ys
```diff
@@ -0,0 +1,16 @@
+#logger -expect warning "Identifier `\\genblk1[0].x' is implicitly declared." 1
+#logger -expect warning "Identifier `\\genblk1[1].x' is implicitly declared." 1
```
zachjs, line 2: Why are these commented out?

Should they be the following instead?
```
logger -expect warning "Identifier `\\genblk1\[0\]\.x' is implicitly declared\." 1
logger -expect warning "Identifier `\\genblk1\[1\]\.x' is implicitly declared\." 1
```

zachjs (commented):
I apologize for not taking a look at this sooner. Thanks for working on this issue and including a test!

tests/verilog/genblk_wire.ys
ylm: I couldn't get the regex working at the time. I'll try those and update accordingly.

frontends/ast/simplify.cc
ylm: Those are indeed duplicate of two chunks of code. Making wrappers would make sense. Will take a look whenever I can.

tests/verilog/genblk_wire.ys
ylm: This form doesn't work and returns an error:
```
-- Executing script file `genblk_wire.ys' --
Added regex 'Identifier `\\genblk1\[0\]\.x' is implicitly declared\.' for warnings to expected warning list.
ERROR: Error in regex expression 'Identifier `\\genblk1\[0\]\.x' is implicitly declared\.' !
```

ylm: According to https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html#tag_09_03_05
```
.[\
    The <period>, <left-square-bracket>, and <backslash> shall be special except when used in a bracket expression (see [RE Bracket Expression](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html#tag_09_03_05)). An expression containing a '[' that is unescaped and is not part of a bracket expression produces undefined results.
```
Hence only the left square bracket needs to be escaped, not the right one. Will fix accordingly.

widlarizer: We use egrep for which slash escaping square brackets isn't perfectly standardized. Or something like that. Changing the `-std` option here affects the result: https://godbolt.org/z/eYsYc6Esf

widlarizer (approved):

PR #4243, opened 2024-02-27
Add YosysHQ/abc as a submodule located in abc.

RCoeurjoly:
Add YosysHQ/abc as a submodule located in abc. Apart from cloning with submodules, no other modifications are made to github actions.

This would enable building yosys with abc fetched as tar.gz, as pointed out by @mmicko in  https://github.com/YosysHQ/yosys/pull/4238#issuecomment-1966067326

Before starting the compilation of abc, we check four cases:

1. abc directory does not exist. The compilation is stopped with the following error.

> Error: The 'abc' directory does not exist.
> Initialize the submodule: Run 'git submodule update --init' to set up 'abc' as a submodule.

2. abc is a clone of abc (abc/.gitcommit content is Format). The compilation is stopped with the following error.

> Error: 'abc' is not configured as a git submodule.
> To resolve this:
> 1. Back up your changes: Save any modifications from the 'abc' directory to another location.
> 2. Remove the existing 'abc' directory: Delete the 'abc' directory and all its contents.
> 3. Initialize the submodule: Run 'git submodule update --init' to set up 'abc' as a submodule.
> 4. Reapply your changes: Move your saved changes back to the 'abc' directory, if necessary.

3. abc comes from a tarball (abc/.gitcommit content is different than $Format:%h$). The compilation can continue.

> 'abc' comes from a tarball. Continuing.

4. abc is a git submodule (checked with `git submodule status`). The compilation can continue.

> 'abc' is a git submodule. Continuing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,7 +18,6 @@ __pycache__
 /coverage.info
 /coverage_html
 /Makefile.conf
-/abc
 /viz.js
 /yosys
 /yosys.exe
@@ -46,3 +45,4 @@ __pycache__
 /tests/unit/bintest/
 /tests/unit/objtest/
 /tests/ystests
+/build
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "abc"]
+	path = abc
+	url = https://github.com/YosysHQ/abc
--- a/Makefile
+++ b/Makefile
@@ -160,15 +160,6 @@ OBJS = kernel/version_$(GIT_REV).o
 bumpversion:
 	sed -i "/^YOSYS_VER := / s/+[0-9][0-9]*$$/+`git log --oneline c1ad377.. | wc -l`/;" Makefile
 
-# set 'ABCREV = default' to use abc/ as it is
-#
-# Note: If you do ABC development, make sure that 'abc' in this directory
-# is just a symlink to your actual ABC working directory, as 'make mrproper'
-# will remove the 'abc' directory and you do not want to accidentally
-# delete your work on ABC..
-ABCREV = 237d813
-ABCPULL = 1
-ABCURL ?= https://github.com/YosysHQ/abc
 ABCMKARGS = CC="$(CXX)" CXX="$(CXX)" ABC_USE_LIBSTDCXX=1 ABC_USE_NAMESPACE=abc VERBOSE=$(Q)
 
 # set ABCEXTERNAL = <abc-command> to use an external ABC instance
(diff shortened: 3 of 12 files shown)


mmicko:
@RCoeurjoly please note that WASI is intentionally built in build directory since that exercise out of tree build that is used in YoWASP distribution. That is then very useful since it points out if changes done in Makefiles affect out of tree builds.

RCoeurjoly:
The steps to build with a tar.gz abc would be:

```
rm abc
wget http://www.eecs.berkeley.edu/~alanmi/abc/abc51205.tar.gz
tar -xvzf abc51205.tar.gz -C abc  --strip-components=1
make ENABLE_ABC_SUBMODULE=0
```

RCoeurjoly:
> @RCoeurjoly please note that WASI is intentionally built in build directory since that exercise out of tree build that is used in YoWASP distribution. That is then very useful since it points out if changes done in Makefiles affect out of tree builds.

Oh I see.

mmicko:
Also abc.tar.gz that is used is one from https://github.com/YosysHQ/yosys/releases This is more for situations where people are downloading specific Yosys version and corresponding abc.tar.gz file and building them on isolated environments.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "abc_submodule"]
+	path = abc_submodule
```
whitequark, line 2: I don't really like this. If someone downloads a `.tar.gz` from the GitHub archive, the `abc` folder for the submodule is going to be just empty. There's no need for them to have different names.

Makefile
```diff
@@ -783,41 +775,28 @@ $(PROGRAM_PREFIX)yosys-config: misc/yosys-config.in
...
-endif
-	$(Q) rm -f abc/abc-[0-9a-f]*
-	$(Q) $(MAKE) -C abc $(S) $(ABCMKARGS) $(if $(filter %.a,$@),PROG="abc-$(ABCREV)",PROG="abc-$(ABCREV)$(EXE)") MSG_PREFIX="$(eval P_OFFSET = 5)$(call P_SHOW)$(eval P_OFFSET = 10) ABC: " $(if $(filter %.a,$@),libabc-$(ABCREV).a)
+	$(Q) $(MAKE) -C abc $(S) $(ABCMKARGS) $(if $(filter %.a,$@),PROG="abc",PROG="abc$(EXE)") MSG_PREFIX="$(eval P_OFFSET = 5)$(call P_SHOW)$(eval P_OFFSET = 10) ABC: " $(if $(filter %.a,$@),libabc.a)
 
 ifeq ($(ABCREV),default)
-.PHONY: abc/abc-$(ABCREV)$(EXE)
-.PHONY: abc/libabc-$(ABCREV).a
+.PHONY: abc/abc$(EXE)
+.PHONY: abc/libabc.a
```
whitequark, line 792: `$(ABCREV)` was used to make sure abc actually gets rebuilt when you change the ABC checkout. Is this still the case? If not you'll need some other way to rebuild ABC when that happens.

.gitmodules
RCoeurjoly: The reasoning behind that is that we want to support both submodule and downloading a tar.gz, and abc should always be located in abc. So I create a symbolic link from abc_submodule to abc.

If we locate the submodule in abc, using a tar.gz would be more difficult and messy, I think.

Also, abc_submodule is a terrible name, suggestions welcome.

whitequark: > If we locate the submodule in abc, using a tar.gz would be more difficult and messy, I think.

This is *solely* relevant for the case where you download a git archive from github, where the submodule is empty. So the abc folder is free to use then.

RCoeurjoly: So would you recommend changing the submodule path?

whitequark: Yeah, put it in `abc` unless there's some other reason that hasn't been brought up yet.

RCoeurjoly: But if the submodule is located in abc, abc will not be empty, and extracting the tar.gz into abc would not make much sense. Am I missing something?

whitequark: Yes, you are. If you download the .tar.gz archive from GitHub, no submodules are included, and the directory is empty. (If GitHub included submodules in the .tar.gz download none of this would've been necessary in the first place.)

whitequark: Yeah. I think the only thing you actually need to watch out for is that there isn't necessarily a valid git repository under `yosys/abc`.

techlibs/quicklogic/Makefile.inc
```diff
@@ -1,4 +1,4 @@
-techlibs/quicklogic/qlf_k6n10f/bram_types_sim.v: techlibs/quicklogic/qlf_k6n10f/generate_bram_types_sim.py
```
povik, line 1: I am not sure we want this. Like all generated files this should be generated in the out-of-tree build directory so that the source tree stays clean. What problem is this fixing?

RCoeurjoly: That's what I thought too, but then I saw this:
techlibs/quicklogic/qlf_k6n10f/.gitignore:1:/bram_types_sim.v
So I guess  bram_types_sim.v is supposed to be generated in techlibs/quicklogic/qlf_k6n10f/

whitequark: Pretty sure this broke the Wasm build before, even.

povik: > So I guess bram_types_sim.v is supposed to be generated in techlibs/quicklogic/qlf_k6n10f/

I wouldn't take that to imply `bram_types_sim.v` must be generated there, it's just that it can be unless you are doing an out-of-tree build. In the case of out-of-tree I would expect it to be placed at the same path but relative to the build directory. If there's an issue I think we should fix the rule that takes `bram_types_sim.v` for its input, not this rule that produces it.

Makefile
```diff
@@ -776,15 +776,15 @@ $(PROGRAM_PREFIX)yosys-config: misc/yosys-config.in
 
 ABC_SOURCES := $(wildcard $(YOSYS_SRC)/abc/*)
 
-abc/abc$(EXE) abc/libabc.a: $(ABC_SOURCES)
+$(YOSYS_SRC)/abc/abc$(EXE) $(YOSYS_SRC)/abc/libabc.a: $(ABC_SOURCES)
```
whitequark, line 779: An out-of-tree build of Yosys should also do an out-of-tree build of abc.

techlibs/quicklogic/Makefile.inc
RCoeurjoly: Done, now generated files get generated in the build dir

povik: Great!

Makefile
```diff
@@ -568,7 +559,7 @@ define add_gen_share_file
 EXTRA_TARGETS += $(subst //,/,$(1)/$(notdir $(2)))
 $(subst //,/,$(1)/$(notdir $(2))): $(2)
 	$$(P) mkdir -p $(1)
-	$$(Q) cp $(2) $(subst //,/,$(1)/$(notdir $(2)))
+	$$(Q) cp "$(CURDIR)"/$(2) $(subst //,/,$(1)/$(notdir $(2)))
```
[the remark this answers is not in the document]
povik: Shouldn't this work without `$(CURDIR)` either way, since `cwd` would be the build directory when these commands are executed?

Makefile
RCoeurjoly: Should the abc executable be located in build in that case?

whitequark: Yep, along with all other files generated during the build. Running `git clean -dxf` in the Yosys or abc checkout should do nothing. (Modulo the odd case of abc being not actually a checked-out submodule.)

Makefile
```diff
@@ -783,41 +774,17 @@ $(PROGRAM_PREFIX)yosys-config: misc/yosys-config.in
...
-
-$(PROGRAM_PREFIX)yosys-abc$(EXE): abc/abc-$(ABCREV)$(EXE)
-	$(P) cp abc/abc-$(ABCREV)$(EXE) $(PROGRAM_PREFIX)yosys-abc$(EXE)
-
-$(PROGRAM_PREFIX)yosys-libabc.a: abc/libabc-$(ABCREV).a
-	$(P) cp abc/libabc-$(ABCREV).a $(PROGRAM_PREFIX)yosys-libabc.a
+	$(Q) $(MAKE) -C $(YOSYS_SRC)/abc $(S) $(ABCMKARGS) $(if $(filter %.a,$@),PROG="abc",PROG="abc$(EXE)") MSG_PREFIX="$(eval P_OFFSET = 5)$(call P_SHOW)$(eval P_OFFSET = 10) ABC: " $(if $(filter %.a,$@),libabc.a)
+
+$(PROGRAM_PREFIX)yosys-abc$(EXE): $(YOSYS_SRC)/abc/abc$(EXE)
+	$(P) cp $(YOSYS_SRC)/abc/abc$(EXE) $(PROGRAM_PREFIX)yosys-abc$(EXE)
```
povik, line 784: Nit: Best to do `$(P) cp $< $@` which makes it obvious the rule inputs/outputs match what we do inside. Especially given that this makefile can sometimes use `VPATH`, in which case `$<` would be resolved including any `VPATH` location.

povik: In any case we still need to make the abc build out-of-tree, as I understand it this builds it in-tree, then copies the resultant binary out of tree.

.gitignore
```diff
@@ -46,3 +45,4 @@ __pycache__
 /tests/unit/bintest/
 /tests/unit/objtest/
 /tests/ystests
+build
```
whitequark, line 48: Should this not be `/build`? This would ignore every file or directory called `build` in the entire tree, which might unpleasantly surprise someone.

Makefile
```diff
@@ -789,41 +780,19 @@ $(PROGRAM_PREFIX)yosys-config: misc/yosys-config.in
 			-e 's#@BINDIR@#$(strip $(BINDIR))#;' -e 's#@DATDIR@#$(strip $(DATDIR))#;' < $< > $(PROGRAM_PREFIX)yosys-config
 	$(Q) chmod +x $(PROGRAM_PREFIX)yosys-config
 
-abc/abc-$(ABCREV)$(EXE) abc/libabc-$(ABCREV).a:
+ABC_SOURCES := $(wildcard $(YOSYS_SRC)/abc/*)
+
+
+# This is a relative path, that is, we would like to build abc in $ROOT/abc/abc if compiling from root or $ROOT/build/abc/abc if compiling from $ROOT/build
```
whitequark, line 786: This seems a bit of a confusing comment, as it appears to imply there's no support for out-of-tree builds.

RCoeurjoly:
> Finally!!! Great job!

Thanks! It was a team effort.
Also note that since you approved of the changes, I made a change to address @povik proposal here https://github.com/YosysHQ/yosys/pull/4243#discussion_r1508766314.

You may want to review that also.

whitequark:
This broke the YoWASP build after all: https://github.com/YoWASP/yosys/actions/runs/9010585766/job/24756882140#step:7:657

```
'abc' comes from a tarball. Continuing.
mkdir -p abc && make -C abc -f "" ABCSRC="/home/runner/work/yosys/yosys/yosys-src/abc"  CC="ccache clang" CXX="ccache clang" ABC_USE_LIBSTDCXX=1 ABC_USE_NAMESPACE=abc VERBOSE= AR="llvm-ar" RANLIB="llvm-ranlib" ARCHFLAGS="-target wasm32-wasi --sysroot /home/runner/work/yosys/yosys/wasi-sdk-19.0/share/wasi-sysroot  -D_WASI_EMULATED_PROCESS_CLOCKS -DABC_USE_STDINT_H -DABC_NO_DYNAMIC_LINKING -DABC_NO_RLIMIT" OPTFLAGS="-Os" "ABC_USE_NO_READLINE=1" "ABC_USE_NO_PTHREADS=1" PROG="abc" MSG_PREFIX="-> ABC: " libabc.a
make: the '-f' option requires a non-empty string argument
make: Entering an unknown directory
Usage: make [options] [target] ...
```

whitequark:
So it turns out the YoWASP build was broken because I didn't check out submodules recursively, and the really confusing error message comes from the combination of an improper check for the presence of the submodule combined with the use of `realpath`.

PR #4247, opened 2024-02-29
Tweaked the representation of nets and buffers in the Show command graph.

zapta:
Two tweaks to the generated graph.

1. Nets changed from a diamond shapes to plain text with the net name. This provides the visual load of the nets while preserving their names.

2. Buffers changed from a period shape to a plain text with the label "[BUF]", providing hint what they are while keeping the visual load low

Before
before

After
after

Dot files
[Archive.zip](https://github.com/YosysHQ/yosys/files/14445442/Archive.zip)

### DIFF
--- a/passes/cmds/show.cc
+++ b/passes/cmds/show.cc
@@ -429,7 +429,7 @@ struct ShowWorker
 
 		std::map<std::string, std::string> wires_on_demand;
 		for (auto wire : module->selected_wires()) {
-			const char *shape = "diamond";
+			const char *shape = "plaintext";
 			if (wire->port_input || wire->port_output)
 				shape = "octagon";
 			if (wire->name.isPublic()) {
@@ -584,7 +584,7 @@ struct ShowWorker
 				} else {
 					net_conn_map[right_node].in.insert({stringf("x%d", single_idx_count), GetSize(conn.first)});
 					net_conn_map[left_node].out.insert({stringf("x%d", single_idx_count), GetSize(conn.first)});
-					fprintf(f, "x%d [shape=point, %s];\n", single_idx_count++, findColor(conn).c_str());
+					fprintf(f, "x%d [shape=plaintext, label=\"[BUF]\", %s];\n", single_idx_count++, findColor(conn).c_str());
 				}
 			}
 		}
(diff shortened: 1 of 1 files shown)


whitequark:
Personally, I find it unconvincing that the new representation is any better (or any worse) than the old one. Both have issues in slightly different ways.

povik:
> 2\. Buffers changed from a period shape to a plain text with the label "[BUF]", providing hint what they are while keeping the visual load low

This makes buffers more visible, why? I suspect this is because it looks better when used with `-colors`, but it makes the output more cluttered. Just recently we removed an explicit `BUF` label, and this would change it back.

zapta:
@povik, the rationale behind the "[BUF]" is to provide to readers an indication what that node mean.  Below are the three version of the buffer. The "[BUF]" one provide a balance between providing the information while keep low clutter.

buf1

buf2

buf3

whitequark:
> is to provide to readers an indication what that node mean

Yosys developers already know what the node means though.

zapta:
@whitequark, yes, it targets new readers, eliminating the question "what does this dot mean?", without much burden on readers that already know.

zapta:
This yosys application suggests that the diagram is also targeting users and learners. I hope that is still the case.  

image

nakengelhardt:
The reason I changed the "BUF" nodes to a dot was that there were a lot of issues filed by confused users that asked how to make yosys remove the buffers (that didn't really exist in the netlist, but were just a visual representation for two connected wires) and I got tired of answering them. I haven't seen any new questions like that filed since the change, so I feel quite confident in saying that the dot shape is less confusing to users.

zapta:
Thanks @nakengelhardt.  I will restore the 'point' shape and will submit a new clean PR.

nakengelhardt:
For the plaintext I guess it can be an option, I personally dislike the wide space between the place where the arrow points and the actual text, especially for short wire names, but I can see that it's a taste question. The default should remain the diamond but I would be ok with a generic option to override it, something like `show -wireshape foo`.

zapta:
Sounds good. Thanks.

zapta:
Hi @nakengelhardt, I have the -wireshape code working, but before I send a PR, I would like to ask you one more question.

You mentioned that you dislike the space in the plaintext nodes and I found a way to eliminate it (see below). Would you consider  having this is the new default (and only) behavior, resulting with simpler code, or should I go ahead with the --wireshape PR?

Original/default:
_default


With -wireshape=plaintext:
_plaintext

 BTW, https://edotor.net/ is a great site to experiment with .dot files.

zapta:
Opening for the ongoing discussion.

Issue #4244, opened 2024-02-27
Cannot build yosys in Ubuntu following instructions in README


RCoeurjoly:
### Version

N/A

### On which OS did this happen?

Linux

### Reproduction Steps

```
git clone git@github.com:YosysHQ/yosys.git
cd yosys
sudo apt-get install build-essential clang bison flex \
	libreadline-dev gawk tcl-dev libffi-dev git \
	graphviz xdot pkg-config python3 libboost-system-dev \
	libboost-python-dev libboost-filesystem-dev zlib1g-dev
make config-clang
make
```

### Expected Behavior

Build successful.

### Actual Behavior

```
In file included from kernel/satgen.cc:20:
In file included from ./kernel/satgen.h:23:
In file included from ./kernel/rtlil.h:20:
./kernel/yosys.h:42:10: fatal error: 'map' file not found
#include <map>
         ^~~~~
In file included from kernel/calc.cc:24:
./kernel/yosys.h:42:10: fatal error: 'map' file not found
#include <map>
         ^~~~~
In file included from kernel/rtlil.cc:20:
./kernel/yosys.h:42:10: fatal error: In file included from kernel/cellaigs.cc:'map' file not found20
:
In file included from ./kernel/cellaigs.h:23:
./kernel/yosys.h:42:10: fatal error: 'map' file not found
#include <map>#include <map>

         ^~~~~         ^~~~~

In file included from kernel/log.cc:20:
./kernel/yosys.h:42:10: fatal error: 'map' file not found
#include <map>
         ^~~~~
In file included from kernel/driver.cc:20:
./kernel/yosys.h:42:10: fatal error: 'map' file not found
#include <map>
         ^~~~~
In file included from kernel/celledges.cc:20:
In file included from ./kernel/celledges.h:23:
./kernel/yosys.h:42:10: fatal error: 'map' file not found
#include <map>
         ^~~~~
In file included from kernel/binding.cc:20:
In file included from kernel/binding.h:23:
In file included from ./kernel/rtlil.h:20:
./kernel/yosys.h:42:10: fatal error: 'map' file not found
#include <map>
         ^~~~~
[  5%] Building kernel/json.o
In file included from kernel/yosys.cc:20:
./kernel/yosys.h:42:10: fatal error: 'map' file not found
#include <map>
         ^~~~~
1 error generated.
In file included from kernel/qcsat.cc:20:
In file included from ./kernel/qcsat.h:23:
[...]

PR #4258, opened 2024-03-05
Fix CXX variables in Makefile

RCoeurjoly:
Thanks to the comment by @KrystalDelusion in https://github.com/YosysHQ/yosys/issues/4244#issuecomment-1975402037, the solution was found.

This PR fixes https://github.com/YosysHQ/yosys/issues/4244.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -215,7 +215,7 @@ ABC_ARCHFLAGS += "-DABC_NO_RLIMIT"
 endif
 
 ifeq ($(CONFIG),clang)
-CXX = clang++
+CXX ?= clang++
 CXXFLAGS += -std=$(CXXSTD) -Os
 ABCMKARGS += ARCHFLAGS="-DABC_USE_STDINT_H -Wno-c++11-narrowing $(ABC_ARCHFLAGS)"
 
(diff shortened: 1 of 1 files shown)


RCoeurjoly:
the emcc pipeline fails, it seems to be related to the CXX variable

KrystalDelusion:
TIL make automatically provides a default value of `CXX = g++`, which can be seen as it is g++ returning the error.  I suspect `make config-clang; make` would also run g++ in this case despite the implication of using the clang config; which (in hindsight) is probably why this was setup the way it was.  Rather than changing the the CXX assignments, it might make more sense to assign `CONFIG=none` by default and leave config-gcc/config-clang etc to override as previously intended.

@RCoeurjoly can you check if #4270 also fixes your issue?  You'll want to delete `Makefile.conf` first and then go straight to `make` (no call to `make config-clang`)

RCoeurjoly:
> TIL make automatically provides a default value of `CXX = g++`, which can be seen as it is g++ returning the error. I suspect `make config-clang; make` would also run g++ in this case despite the implication of using the clang config; which (in hindsight) is probably why this was setup the way it was. Rather than changing the the CXX assignments, it might make more sense to assign `CONFIG=none` by default and leave config-gcc/config-clang etc to override as previously intended.
> 
> @RCoeurjoly can you check if #4270 also fixes your issue? You'll want to delete `Makefile.conf` first and then go straight to `make` (no call to `make config-clang`)

I confirm that PR 4270 fixes this issue.

KrystalDelusion:
Awesome.  I'll close this one in favour of that then.

Issue #3426, opened 2022-07-28
opt_clean reverses some connections


georgerennie:
## Steps to reproduce the issue

```
read_verilog <<EOT
module top(
	input wire in,
	output wire out1, out2
);

wire int1, int2;
assign int1 = in;
assign out1 = in;
assign out2 = int2;

endmodule
EOT

dump
opt_clean
dump
```

## Expected behavior

Both dump outputs should contain
```
  connect \int1 \in
  connect \out1 \in
  connect \out2 \int2
```

## Actual behavior

Second dump output instead contains
```
  connect \int2 \out2
  connect \int1 \in
  connect \out1 \in
```

Notably, the connection between out2 and int2 has been reversed. I played around with connecting a few different types of signals and only saw this occur for an undriven signal driving an output wire.

nakengelhardt:
Does this cause a problem? Semantically it seems identical to me, two undriven wires aliasing each other.

georgerennie:
Not necessarily but it can lead to some slightly confusing behaviour, e.g. Sby reports the output net as being undriven, rather than the source (I am not at my computer right now to check which yosys command that corresponds to). I'm happy for the issue to be closed if its not considered an issue.

georgerennie:
Not necessarily but it can lead to some slightly confusing behaviour, e.g.
Sby reports the output net as being undriven, rather than the source (I am
not at my computer right now to check which yosys command that corresponds
to). I'm happy for the issue to be closed if its not considered an issue.

KatCe:
We have built a pass that analyses data flows in designs. The reversing of connections prevents our pass from being able to determine the data flow direction between these two signals.

nakengelhardt:
There won't be any data flowing through an undriven wire, though, so does that actually have relevant effects on the result? As a workaround, you could use something like `setundef -undriven -undef` before the first call to `opt_clean` to drive those wires with Xs.

KatCe:
We saw it happening with driven wires after optimizations.

povik:
Based on knowledge of the `opt_clean` code I can confirm that when you have a pair of connected wires, it can happen that the driver is switched from one of the wires to the other, and the direction of the connection is reversed.

nakengelhardt:
But in that case all users of the signal should also be connected to the chosen wire, no?

[...]

PR #4267, opened 2024-03-07
opt_clean: Add `-coarse` mode

povik:
```
    -coarse
        perform a fast, conservative clean-up in which no public wires are
        removed and wire usage is analyzed at a coarse level without
        distinguishing individual bits
```

### DIFF
--- a/passes/opt/opt_clean.cc
+++ b/passes/opt/opt_clean.cc
@@ -31,6 +31,156 @@ PRIVATE_NAMESPACE_BEGIN
 
 using RTLIL::id2cstr;
 
+bool should_be_kept(Design *d, Cell *cell)
+{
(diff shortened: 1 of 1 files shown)


passes/opt/opt_clean.cc
```diff
@@ -31,6 +31,156 @@ PRIVATE_NAMESPACE_BEGIN
 
 using RTLIL::id2cstr;
 
+bool should_be_kept(Design *d, Cell *cell)
```
povik, line 34: Before we merge (if we merge), we should get rid of this `should_be_kept` function and use `keep_cache` and the other code that's already there.

povik:
This mode preserves the direction of connections (see issue #3426).

PR #4270, opened 2024-03-07
Change default CONFIG to none

KrystalDelusion:
Default to using provided CXX (either via environment variable or make).

Some of the `ifeq ($(CONFIG), ..)` checks might also be worth changing to check `$(CXX)` instead to, e.g. use a clang sanitizer if one is selected without config-clang.


Makefile
```diff
@@ -362,6 +363,10 @@ EXE = .exe
 
 else ifneq ($(CONFIG),none)
 $(error Invalid CONFIG setting '$(CONFIG)'. Valid values: clang, gcc, emcc, mxe, msys2-32, msys2-64)
+else
+$(info Using provided C++ compiler, '$(CXX)'.)
```
whitequark, line 367: Is this going to be printed on every `make` invocation? That's distracting.

KrystalDelusion: Good point (although it would "only" be every `make` invocation without a value for `CONFIG`).  I think it's still useful for debug purposes but it now requires `WHICH_CXX` to be defined so it is opt-in.

mmicko:
This changes default behavior since now CXX is detected and on Linux it gets defaulted to g++, and current Makefile is enforcing clang by default.

Suggest removing `-Wno-c++11-narrowing` from Makefile since that will make `ABCMKARGS` behave same on gcc and clang config paths. (best checking building abc by removing abc directory before build)

whitequark:
> This changes default behavior since now CXX is detected and on Linux it gets defaulted to g++, and current Makefile is enforcing clang by default.

To be honest it was pretty weird that clang would be picked by default even on Linux (I *strongly prefer* clang but it's still weird).

KrystalDelusion:
> This changes default behavior since now CXX is detected and on Linux it gets defaulted to g++, and current Makefile is enforcing clang by default.

There's nowhere that actually says *what* the default compiler is, just that running `make config-*` was necessary for a specific compiler so I don't think that's a problem.  And as someone who prefers clang I would always run `make config-clang` on a fresh install, having assumed until making this PR that gcc would be the default.

PR #4283, opened 2024-03-15
abc9: Experiment with importing structural choices

povik:
This is a bit of code on top of #4282 to have the structural choices that abc9 comes up reimported back into Yosys, for inspection and possibly other uses.

Minimal example where one sees it in action:
```
read_verilog <<EOF
(* top *)
module top(a, b, y);
	input wire [1:0] a;
	input wire [1:0] b;
	output wire [1:0] y;
	assign y = a * b;
endmodule
EOF
opt_clean
techmap
splitnets -ports
abc9 -script +&synch2;,&ps;;
show
```

graph

### DIFF
--- a/frontends/aiger/aigerparse.cc
+++ b/frontends/aiger/aigerparse.cc
@@ -505,6 +505,26 @@ void AigerReader::parse_xaiger()
 				boxes.emplace_back(cell);
 			}
 		}
+		else if (c == 'q') {
+			f.ignore(sizeof(uint32_t));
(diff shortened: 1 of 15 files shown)


povik:
Let me leave a note that the imported `$__choice` nodes represent equivalences up to a negation. (One can find out the value of the referenced nodes under all PIs being zero to recover polarities.)

povik:
I won't be pursuing this further.

PR #4297, opened 2024-03-23
chore: fix master branch refs

chenrui333:
I could not find exact change wrt master branch to main branch, but since this already happened, just rename all the master branch references to main.

also relates to https://github.com/Homebrew/homebrew-core/pull/166979
- https://github.com/Homebrew/homebrew-core/issues/166965


KrystalDelusion:
There isn't a commit that marks the change in default branch name, it was purely on the github side.  d73f71e was the last commit prior to the change.

There is also an existing PR to update references in the docs, #4292.  Yours has a number of whitespace fixes in the affected files, which is great, but if that's the only difference I would rather you were consistent and did it to *all* of the docs source files.  And probably in a separate PR.

chenrui333:
yeah, I can exclude the whitespace changes from the PR if that is the concern in here. (I can also incorporate the pr changes, I did not know that one's existence)

chenrui333:
@KrystalDelusion should be good now.

KrystalDelusion:
Thanks. I'll close the other one in favour of this one then.

KrystalDelusion (approved):

chenrui333:
the test failure should be unrelated, right?

KrystalDelusion:
Looks like it's unrelated and transient.  Re run passed.

PR #4301, opened 2024-03-28
Add several new formatting features used by Amaranth to translate Python format strings

whitequark:
This is a work-in-progress PR at the moment.
- [x] `H` format type (uppercase hex digits)
- [x] `c` format type (Unicode code point emitted as UTF-8)
- [x] `=` justification (align right, but after the padding)
- [x] arbitrary fill character
- [x] ` ` sign mode (either `-` or ` `, as opposed to either `-` or `+`)
- [x] `_` option (insert an underscore between each 3 decimal, 4 binary/octal/hex digits)
- [x] `#` option (insert `0x`/`0o`/`0b` after sign and before padding)
- [x] signed hex numbers
- [x] padding interacts poorly with base
- [x] consider combination of `0` padding and `_`
- [x] test integration with Amaranth

### DIFF
--- a/backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
+++ b/backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
@@ -1034,6 +1034,7 @@ struct fmt_part {
 	unsigned base; // = 10;
 	bool signed_; // = false;
 	bool plus; // = false;
+	bool hex_upper; // = false;
 
 	// VLOG_TIME type
 	bool realtime; // = false;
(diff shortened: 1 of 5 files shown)


whitequark:
@mmicko I need to have these in the next Yosys release to meet the current Amaranth release schedule. The PR should be done by the 1st or very shortly after.

povik:
This looks like it overlaps or supersedes #4192.

whitequark:
@povik I *think* there's actually no overlap? Note `VerilogFmtArg` vs `FmtPart`.

povik:
Right you are

PR #4303, opened 2024-03-28
Infrastructure to run a Sat solver as a command

Coloquinte:
As discussed on the slack channel, here is a patch for a solver interface using an arbitrary command (I used kissat for my test).
It also removes some unused code that was supposed to register arbitrary solvers.

As the moment, it is only usable for experimental purposes from the C++. I didn't implement any direct access from the command line (sat, equiv, ...)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -632,6 +632,7 @@ $(eval $(call add_include_file,kernel/yosys.h))
 $(eval $(call add_include_file,kernel/yw.h))
 $(eval $(call add_include_file,libs/ezsat/ezsat.h))
 $(eval $(call add_include_file,libs/ezsat/ezminisat.h))
+$(eval $(call add_include_file,libs/ezsat/ezcommand.h))
 ifeq ($(ENABLE_ZLIB),1)
 $(eval $(call add_include_file,libs/fst/fstapi.h))
 endif
@@ -672,6 +673,7 @@ OBJS += libs/json11/json11.o
 
 OBJS += libs/ezsat/ezsat.o
 OBJS += libs/ezsat/ezminisat.o
+OBJS += libs/ezsat/ezcommand.o
 
 OBJS += libs/minisat/Options.o
 OBJS += libs/minisat/SimpSolver.o
--- a/kernel/register.cc
+++ b/kernel/register.cc
@@ -979,18 +979,6 @@ struct EchoPass : public Pass {
 	}
 } EchoPass;
 
-SatSolver *yosys_satsolver_list;
-SatSolver *yosys_satsolver;
-
-struct MinisatSatSolver : public SatSolver {
-	MinisatSatSolver() : SatSolver("minisat") {
-		yosys_satsolver = this;
-	}
-	ezSAT *create() override {
-		return new ezMiniSAT();
-	}
-} MinisatSatSolver;
-
 struct LicensePass : public Pass {
 	LicensePass() : Pass("license", "print license terms") { }
 	void help() override
--- a/libs/ezsat/ezcommand.cc
+++ b/libs/ezsat/ezcommand.cc
@@ -9,19 +9,20 @@ ezSATCommand::~ezSATCommand() {}
 
 bool ezSATCommand::solver(const std::vector<int> &modelExpressions, std::vector<bool> &modelValues, const std::vector<int> &assumptions)
 {
-	if (!assumptions.empty()) {
-		Yosys::log_error("Assumptions are not supported yet by command-based Sat solver\n");
-	}
 	const std::string tempdir_name = Yosys::make_temp_dir(Yosys::get_base_tmpdir() + "/yosys-sat-XXXXXX");
 	const std::string cnf_filename = Yosys::stringf("%s/problem.cnf", tempdir_name.c_str());
 	const std::string sat_command = Yosys::stringf("%s %s", command.c_str(), cnf_filename.c_str());
 	FILE *dimacs = fopen(cnf_filename.c_str(), "w");
-	printDIMACS(dimacs);
-	fclose(dimacs);
 
 	std::vector<int> modelIdx;
 	for (auto id : modelExpressions)
 		modelIdx.push_back(bind(id));
+	std::vector<std::vector<int>> extraClauses;
+	for (auto id : assumptions)
+		extraClauses.push_back({bind(id)});
+
+	printDIMACS(dimacs, false, extraClauses);
+	fclose(dimacs);
 
 	bool status_sat = false;
 	bool status_unsat = false;
--- a/libs/ezsat/ezcommand.cc
+++ b/libs/ezsat/ezcommand.cc
@@ -9,6 +9,7 @@ ezSATCommand::~ezSATCommand() {}
 
 bool ezSATCommand::solver(const std::vector<int> &modelExpressions, std::vector<bool> &modelValues, const std::vector<int> &assumptions)
 {
+#if !defined(YOSYS_DISABLE_SPAWN)
 	const std::string tempdir_name = Yosys::make_temp_dir(Yosys::get_base_tmpdir() + "/yosys-sat-XXXXXX");
 	const std::string cnf_filename = Yosys::stringf("%s/problem.cnf", tempdir_name.c_str());
 	const std::string sat_command = Yosys::stringf("%s %s", command.c_str(), cnf_filename.c_str());
@@ -79,4 +80,7 @@ bool ezSATCommand::solver(const std::vector<int> &modelExpressions, std::vector<
 		modelValues[i] = (values.at(idx - 1) == refvalue);
 	}
 	return true;
+#else
+	Yosys::log_error("SAT solver command not available in this build!\n");
+#endif
 }
--- a/libs/ezsat/ezcommand.cc
+++ b/libs/ezsat/ezcommand.cc
@@ -1,0 +1,81 @@
+
+#include "ezcommand.h"
+
+#include "../../kernel/yosys.h"
+
+ezSATCommand::ezSATCommand(const std::string &cmd) : command(cmd) {}
+
+ezSATCommand::~ezSATCommand() {}
+
+bool ezSATCommand::solver(const std::vector<int> &modelExpressions, std::vector<bool> &modelValues, const std::vector<int> &assumptions)
+{
+	if (!assumptions.empty()) {
+		Yosys::log_error("Assumptions are not supported yet by command-based Sat solver\n");
+	}
+	const std::string tempdir_name = Yosys::make_temp_dir(Yosys::get_base_tmpdir() + "/yosys-sat-XXXXXX");
+	const std::string cnf_filename = Yosys::stringf("%s/problem.cnf", tempdir_name.c_str());
+	const std::string sat_command = Yosys::stringf("%s %s", command.c_str(), cnf_filename.c_str());
+	FILE *dimacs = fopen(cnf_filename.c_str(), "w");
+	printDIMACS(dimacs);
+	fclose(dimacs);
+
+	std::vector<int> modelIdx;
+	for (auto id : modelExpressions)
+		modelIdx.push_back(bind(id));
+
+	bool status_sat = false;
+	bool status_unsat = false;
+	std::vector<bool> values;
+
+	auto line_callback = [&](const std::string &line) {
+		if (line.empty()) {
+			return;
+		}
+		if (line[0] == 's') {
+			if (line.substr(0, 5) == "s SAT") {
+				status_sat = true;
+			}
+			if (line.substr(0, 7) == "s UNSAT") {
+				status_unsat = true;
+			}
+			return;
+		}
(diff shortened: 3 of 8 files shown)


kernel/register.cc
```diff
@@ -979,18 +979,6 @@ struct EchoPass : public Pass {
 	}
 } EchoPass;
 
-SatSolver *yosys_satsolver_list;
-SatSolver *yosys_satsolver;
-
-struct MinisatSatSolver : public SatSolver {
```
povik, line 985: I am not sure we want this removed. This allowed for a plugin to register a different SAT solver, though I am not sure `yosys_satsolver_list` gets used anywhere, so that the other-than-default solver would ever get selected.

povik: Looks like we are missing a command `satselect`, `solverselect` or similar which would update the `yosys_satsolver` pointer, but the overall approach of these `SatSolver` descendants appears sane. On the contrary, instead of removing this, we should arrange for an `ezcommand` solver to be instantiable through this.

Coloquinte: Thanks for your explanation, now I see the benefit. I thought it was entirely unused.
I can just remove this commit and the next one: ezSATCommand would not be accessible from the global solver yet, but would at least be available from C++ plugins.

povik: OK

libs/ezsat/ezcommand.cc
```diff
@@ -0,0 +1,86 @@
...
+				bool val = lit >= 0;
+				int ind = lit >= 0 ? lit - 1 : -lit - 1;
+				if (Yosys::GetSize(values) <= ind) {
+					values.resize(ind + 1);
+				}
+				values[ind] = val;
+			}
+		}
+	};
+	Yosys::run_command(sat_command, line_callback);
```
povik, line 61: You are not inspecting the return value. Also, if I can ask you to make the change, make it so that with the `DISABLE_SPAWN` option `run_command` becomes an empty stub always returning an error value instead of disappearing.  This means we could get rid of the macro conditionals here. @whitequark any objections?

kernel/register.cc
povik: That's a good point. I will defer to @jix's judgement on this. Should we e.g. always override the default solver with the one from a plugin, if available?

libs/ezsat/ezcommand.cc
```diff
@@ -0,0 +1,86 @@
...
+	FILE *dimacs = fopen(cnf_filename.c_str(), "w");
+
+	std::vector<int> modelIdx;
+	for (auto id : modelExpressions)
+		modelIdx.push_back(bind(id));
+	std::vector<std::vector<int>> extraClauses;
+	for (auto id : assumptions)
+		extraClauses.push_back({bind(id)});
+
+	printDIMACS(dimacs, false, extraClauses);
```
povik, line 25: Would there be issues with appending the extra clauses to the file from here, instead of introducing a new argument to `printDIMACS`? I see `printDIMACS` finds out the maximum term or clause number to arrange for uniform padding, but is that a requirement of the format or just a convenience for inspection?

Makefile
```diff
@@ -672,6 +673,7 @@ OBJS += libs/json11/json11.o
 
 OBJS += libs/ezsat/ezsat.o
 OBJS += libs/ezsat/ezminisat.o
+OBJS += libs/ezsat/ezcommand.o
```
povik, line 676: This can be a bit of bikeshedding but I suggest `ezcmdline` or `ezextern` for the name.

libs/ezsat/ezcommand.cc
Coloquinte: I preferred to extend printDIMACS rather than duplicate the writing code. You'd prefer not to change the API?

Makefile
Coloquinte: Fine by me. ezcmdlinesat for the file and ezCmdLineSAT for the class to keep it consistent with the rest?

libs/ezsat/ezcommand.cc
Coloquinte: I suggested an implementation in the last commit

kernel/yosys.cc
```diff
@@ -318,8 +318,10 @@ int run_command(const std::string &command, std::function<void(const std::string
 #else
 	return WEXITSTATUS(ret);
 #endif
-}
+#else
+	log_cmd_error("Command execution is disabled in this build.\n");
```
povik, line 322: Make it a `log_warn` please since `log_cmd_error` throws an exception if I am not mistaken, and the call site might not be prepared to handle that. Also rephrase to "Shell command execution ...".

passes/cmds/show.cc
```diff
@@ -758,7 +758,7 @@ struct ShowPass : public Pass {
 		std::vector<std::pair<std::string, RTLIL::Selection>> color_selections;
 		std::vector<std::pair<std::string, RTLIL::Selection>> label_selections;
 
-#if defined(_WIN32) || defined(YOSYS_DISABLE_SPAWN)
```
povik, line 761: I don't think we want to remove this one conditional.

passes/cmds/viz.cc
```diff
@@ -892,7 +892,7 @@ struct VizPass : public Pass {
 		log_header(design, "Generating Graphviz representation of design.\n");
 		log_push();
 
-#if defined(_WIN32) || defined(YOSYS_DISABLE_SPAWN)
```
povik, line 895: Same as above, don't remove this one conditional

kernel/fstdata.cc
```diff
@@ -29,7 +29,6 @@ static std::string file_base_name(std::string const & path)
 
 FstData::FstData(std::string filename) : ctx(nullptr)
 {
-	#if !defined(YOSYS_DISABLE_SPAWN)
```
Coloquinte, line 32: The behaviour changes here, but I think there would be an error later when trying to open the fst file

povik: Yes, this is probably fine to remove.

kernel/register.cc
```diff
@@ -249,9 +248,6 @@ void Pass::call(RTLIL::Design *design, std::string command)
 		if (retCode != 0)
 			log_cmd_error("Shell command returned error code %d.\n", retCode);
 		return;
-#else
-		log_cmd_error("Shell is not available.\n");
```
Coloquinte, line 253: I'll actually leave this one

passes/cmds/show.cc
Coloquinte: Fixed

passes/cmds/viz.cc
Coloquinte: Fixed

Makefile
Coloquinte: I renamed as described above

povik: Sorry to bikeshed but I think it's better to have `ezcmdline`, since the fact that it's an interface to a SAT solver is clear from the context, and `ezCmdlineSAT` since cmdline is a single word as used

libs/ezsat/ezcommand.cc
povik: It depends on how much code would be duplicated in practice. How much lines would you estimate that to be?

kernel/register.cc
povik: Good call

kernel/yosys.cc
```diff
@@ -318,8 +318,11 @@ int run_command(const std::string &command, std::function<void(const std::string
 #else
 	return WEXITSTATUS(ret);
 #endif
-}
+#else
+	log_warning("Shell command execution is disabled in this build.\n");
```
whitequark, line 322: This should definitely not be a warning but an error.

whitequark (changes requested):

kernel/yosys.cc
povik: It signifies an error through the return value, the warning is a courtesy to help the user understand the cause of the error.

The outside code needs handling of an error anyway for platforms with spawning enabled, and relying on that has the advantage that the handling may be log_cmd_error, which is better for user experience. We can’t do blanket log_cmd_error from within the run_command function.

libs/ezsat/ezcommand.cc
Coloquinte: Not that many actually, less than 10

libs/ezsat/ezcmdline.cc
```diff
@@ -0,0 +1,84 @@
...
+				bool val = lit >= 0;
+				int ind = lit >= 0 ? lit - 1 : -lit - 1;
+				if (Yosys::GetSize(values) <= ind) {
+					values.resize(ind + 1);
+				}
+				values[ind] = val;
+			}
+		}
+	};
+	if (Yosys::run_command(sat_command, line_callback) != 0) {
```
jix, line 60: Many solvers follow the convention that they exit with status 10 for satisfiable and with status 20 for unsatisfiable instances, so those exit code shouldn't be treated as an error if they match result parsed from the log output.

kernel/register.cc
jix: It would be nice to have a way to override the default solver, that would also make it easier to test this PR. I don't want to change the default without an explicit command though (`sat -select-solver ...` could be added, there is precedent for subcommand like options and it would avoid cluttering the command list, but maybe also make it a bit hard to discover)

jix (changes requested):
On top of the comments I left inline, there's no easy way to actually test this without manually patching the default or writing a plugin making direct use of it. This should come with either a way to change the default solver or an example plugin with a test case.

libs/ezsat/ezcommand.cc
Coloquinte: As pointed out by Jix, the number of clauses is in the header, so we cannot actually add this afterwards

kernel/yosys.cc
Coloquinte: I removed it from the PR

gussmith23:
@Coloquinte we're trying to resolve this PR (merging or closing) -- can you recall what was left to be done on this on your side?

gussmith23:
@jix @whitequark @povik 

Do any of you have time for a final review of this?

I have rebased this PR (no conflicts to resolve) and added a test. No other modifications have been made, but I'm fairly certain all existing comments except the below comment have been resolved:

https://github.com/YosysHQ/yosys/pull/4303#discussion_r1555719968

@jix if you feel this feature is necessary for merging, i can take a crack at it!

gussmith23:
FYI: the test makes a very simple dummy SAT solver (parses dimacs w/ awk, but rejects all but two input problems, those being the simple SAT case of a single variable and the simple UNSAT case of a variable and its negation). It uses this solver from a simple plugin, using the functionality added in this PR. No functionality of the PR was changed to support this.

libs/ezsat/ezcmdline.cc
```diff
@@ -0,0 +1,89 @@
...
+	int return_code = Yosys::run_command(sat_command, line_callback);
+	if (return_code != 0 && return_code != 10 && return_code != 20) {
+		Yosys::log_cmd_error("Shell command failed!\n");
+	}
+
+	modelValues.clear();
+	modelValues.resize(modelIdx.size());
+
+	if (!status_sat && !status_unsat) {
+		solverTimoutStatus = true;
```
hexratcc: typo: `solverTimeoutStatus`

libs/ezsat/ezcmdline.cc
```diff
@@ -0,0 +1,89 @@
...
+	FILE *dimacs = fopen(cnf_filename.c_str(), "w");
+
+	std::vector<int> modelIdx;
+	for (auto id : modelExpressions)
+		modelIdx.push_back(bind(id));
+	std::vector<std::vector<int>> extraClauses;
+	for (auto id : assumptions)
+		extraClauses.push_back({bind(id)});
+
+	printDIMACS(dimacs, false, extraClauses);
```
hexratcc, line 28: Would like to see some kind of check for `dimacs` being null, or some form of check in `printDIMACS`.

gussmith23: Done, I added checks in both places just for completeness's sake, though lmk if you prefer just one

libs/ezsat/ezcmdline.cc
gussmith23: Fixed

gussmith23:
@Goubermouche Addressed your comments!

> I agree with @jix regarding `-select-solver`, but I'm wondering how ergonomic it would be for scripts that run multiple SAT operations. Having to pass `-select-solver kissat` to every `sat`, `equiv`, `miter`, etc. command could get repetitive. Would adding a global command like `select_sat` be viable? 

By @nakengelhardt's suggestion, I opted to use the scratchpad to control the SAT solver selection, but we could add a simple wrapper command that sets the scratchpad value. so something like `scratchpad -set sat.solver minisat` will set the solver. The resolution order is currently: command line flag > scratchpad > global default (`yosys_satsolver`).

PR #4304, opened 2024-03-28
extract: add arbitrary port width matching

phsauter:
The goal is to improve extract and make it easier to use in scenarios similar to the ones described in the [extract example](https://yosyshq.readthedocs.io/projects/yosys/en/latest/using_yosys/synthesis/extract.html).

The final goal is to extract something like a multiplier followed by an arbitrary amount of variable-width adders.

ToDo:
- [ ] minimum port widths per port (in user port matching)
- [ ] avoid expanding ports to max width with `'Z`
- [ ] document in `macc` examples

### DIFF
--- a/passes/techmap/extract.cc
+++ b/passes/techmap/extract.cc
@@ -148,7 +148,7 @@ struct bit_ref_t {
 	int bit;
 };
 
-bool module2graph(SubCircuit::Graph &graph, RTLIL::Module *mod, bool constports, RTLIL::Design *sel = nullptr,
+bool module2graph(SubCircuit::Graph &graph, RTLIL::Module *mod, bool constports, int min_port_width = -1, RTLIL::Design *sel = nullptr,
 		int max_fanout = -1, std::set<std::pair<RTLIL::IdString, RTLIL::IdString>> *split = nullptr)
 {
 	SigMap sigmap(mod);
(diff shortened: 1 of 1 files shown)


passes/techmap/extract.cc
```diff
@@ -427,6 +438,10 @@ struct ExtractPass : public Pass {
 		log("    -mine_max_fanout <num>\n");
 		log("        don't consider internal signals with more than <num> connections\n");
 		log("\n");
+		log("    -min_port_width <num>\n");
+		log("        match all subcircuits with port-sizes between needle port widths and this number\n");
```
povik, line 442: So "this number" is lower than the needle port widths?

phsauter: Yes, this number defines the lower bounds of the needle port width (it applies to all ports).  
Usually the port widths must match exactly, this exposes an existing feature of the algorithm.
So: hay_port_width ∈ [min_width; needle_port_width]

povik: Yes, I am asking due to the phrasing, which suggested to me `<num>` is an upper bound, not a lower bound. Not sure what we should change it to though.

phsauter: maybe something like:
```
match all subcircuits with port widths from <num> up to the needle port width given in the map.
```

PR #4305, opened 2024-03-29
cxxrtl: Fix sdivmod

merryhime:
`x = x.neg();` results in the subsequent `x.is_neg()` to always be false.
Ditto for the `dividend.is_neg() != divisor.is_neg()` test.
Fix this.


whitequark (approved):
Thanks!

PR #4311, opened 2024-04-02
cleanup: extra_args and argidx

widlarizer:
This PR simplifies and unifies boilerplate by removing unused iteration over pass args which have been left in as placeholders. It also removes commented out arg checks that I think aren't relevant to the pass at hand and have only been copied from another pass.

Incidentally, by setting argidx correctly it also prevents the warning opt_demorgan would unconditionally throw: `Warning: Selection "opt_demorgan" did not match any module.`

### DIFF
--- a/passes/cmds/autoname.cc
+++ b/passes/cmds/autoname.cc
@@ -101,15 +101,7 @@ struct AutonamePass : public Pass {
 	}
 	void execute(std::vector<std::string> args, RTLIL::Design *design) override
 	{
-		size_t argidx;
-		for (argidx = 1; argidx < args.size(); argidx++)
-		{
-			// if (args[argidx] == "-foo") {
-			// 	foo = true;
-			// 	continue;
-			// }
-			break;
-		}
+		extra_args(args, 1, design);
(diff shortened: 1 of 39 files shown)


whitequark:
> Incidentally, by setting argidx correctly it also prevents the warning opt_demorgan would unconditionally throw: `Warning: Selection "opt_demorgan" did not match any module.`

This seems like a bug fix that should be submitted separately.

widlarizer:
@whitequark true, opened #4313

@povik It can be expanded at any point by copying from another pass. Or we can make an example pass to make contributing new passes easier. It doesn't seem necessary at all to have this boilerplate in some passes and not others, without explicit TODOs about what should be added. I mean it's a cosmetic change, but I think letting unused code and variables hang around can be distracting

povik:
Right, it's a cosmetic thing and a matter of taste. Let's leave it up to the dev JF, and if my position will be in the minority let's merge it.

widlarizer:
Eh, I think we ultimately have bigger fries to fish

PR #4312, opened 2024-04-02
kernel: Avoid including files outside include guards

jix:
This adjusts the way the headers kernel/{yosys,rtlil,register,log}.h include each other to avoid the need of including headers outside of include guards as well as avoiding the inclusion of rtlil.h in the middle of yosys.h with rtlil.h depending on the prefix of yosys.h, and the suffix of yosys.h depending on rtlil.h.

To do this I moved some of the declaration in yosys.h into a new header yosys_common.h. I'm not sure if that is strictly necessary.

Including any of these files still results in the declarations of all these headers being included, so this shouldn't be a breaking change for any passes or external plugins.

My main motivation for this is that ccls's (clang based language server) include guard handling gets confused by the previous way the includes were done. It often ends up treating the include guard as a generic disabled preprocessor conditional, breaking navigation and highlighting for the core RTLIL data structures.

Additionally I think avoiding cyclic includes in the middle of header files that depend on includes being outside of include guards will also be less confusing for developers reading the code, not only for tools like ccls.


nakengelhardt (approved):
Confirming that all the linty plugins still compile after this change, the earlier problem was just my own local misconfiguration.

PR #4313, opened 2024-04-03
opt_demorgan: fix extra args warning

widlarizer:
By setting argidx correctly this PR prevents the warning opt_demorgan would unconditionally throw: `Warning: Selection "opt_demorgan" did not match any module.`


whitequark (approved):
Thanks!

PR #4328, opened 2024-04-10
drivertools: WIP

jix:
I'll add more description later


jix:
I did some minor cleanup of the commit history and moved the temporary example pass as well as parts of the functional backend that go beyond drivertools into a separate branch at `https://github.com/jix/yosys/tree/functional`.

PR #4332, opened 2024-04-12
Add docs generation from cells help output

KrystalDelusion:
Generate rst files for the cells library, similar to how the command reference is generated.  Python pre-parse allows for a bit more flexibility.

ReadtheDocs preview available at https://yosyshq.readthedocs.io/projects/yosys/en/docs-preview-cellhelp/


jix:
A few things I noticed, including some nits that might not need to be addressed:
* On the top-level "Internal cell library" page, there is still the "TODO brief overview of internal cell library"
* On the "Unary operators" page:
	* it says "Note that [$reduce_or](https://yosyshq.readthedocs.io/projects/yosys/en/docs-preview-cellhelp/cell/word_unary.html#unary.$reduce_or) and [$reduce_bool](https://yosyshq.readthedocs.io/projects/yosys/en/docs-preview-cellhelp/cell/word_unary.html#unary.$reduce_bool) actually represent the same logic function. But the HDL frontends generate them in different situations". AFAICT verific doesn't generate either, so maybe this should be more specific and say that "read_verilog" does this?
	* It says $pos is a buffer, which currently is true, but when we add a $buf cell, it should only be used for arithmetic unary +, which is the same as a buffer except when modeling x-propagation where unary + will be treated as an arithmetic operation in that any single x input bit will produce all x outputs. Not sure if anything about the documentation should be changed now though, as it correctly documents how things are treated currently.
* On the "Binary operators" page:
	* is $modfloor really known as “remainder”? I know that terminology of the different rounding behaviors is a big mess, but if there's any correlation I'd expect it to be the other way around with modulo always being non-negative for a positive divisor and remainder corresponding to the truncating variant. I do like that the page lists examples and that the cells state the invariants that hold, so no matter the expected terminology there shouldn't be any confusion.
	* the descriptions for $shift and $shiftx seem to be swapped
* On the "Multiplexers" page:
	* $bmux and $bwmux have no description, I think it should at least mention that $bmux is a wide mux that instead of a one-hot in the case of $pmux, uses a binary encoding / an address to select an input. For $bwmux it should mention that it is a bitwise mux and maybe mention the equivalent logic term
* I didn't fully review the "Memories" page as I know that there are some subtleties that I don't quite remember and that I'd have to look up in the yosys sources.
* I didn't review the "Finite state machines", "Coarse arithmetics", "Arbitrary logic functions" and "Specify rules" pages as I'm not familiar enough with the corresponding cell types.
* The "Formal verification cells" page is still all TODOs

KrystalDelusion:
The `$modfloor` thing really threw me as well, because remainder should never be negative; but there *are* languages that follow the given semantics (Python, for example, gives the same results when using the `%` operator)...  Will change it to at least be more specific about 'other languages' to hopefully be less ambiguous.

widlarizer:
This also added unused CellType members, which is a bit risky to expose before implementing. I discovered it because I was looking at adding `is_symmetrical`. I'll avoid touching it if this is something you're working on, let me just suggest that we could move from `std::vector` to `std::set` and then set members by checking intersections:

```cpp
for (auto type : unary_ops)
	setup_type(type, {ID::A}, {ID::Y}, true, symmetrical_ops.count(type));
```

KrystalDelusion:
@widlarizer it was more a proof of concept because I don't actually know which cells have those properties, just that they were useful properties to report 😅 Feel free to change it by adding others or commenting out the unused ones or whatever, just make sure that [kernel/register.cc write_cell_rst](https://github.com/YosysHQ/yosys/blob/main/kernel/register.cc#L897) is updated to match 😄

PR #4334, opened 2024-04-12
Strip compilation date from doc outputs

KrystalDelusion:
- [x] Fix `yosys-abc` usage changing for every compile.
- [x] Fix `test*.log` changing for every compile.
- [x] Fix `test*.log` changing for every verific release.
- [x] Resolve ambiguity on source/temp by renaming to source/generated (i.e. the files should not be included in the Yosys repo, but they are needed for publishing the docs).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -985,16 +985,27 @@ docs/gen_images:
 	$(Q) $(MAKE) -C docs images
 
 DOCS_GUIDELINE_FILES := GettingStarted CodingStyle
-docs/guidelines:
+docs/guidelines docs/source/temp:
 	$(Q) mkdir -p docs/source/temp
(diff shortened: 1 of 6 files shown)


mmicko (changes requested):
`    :end-before: UC Berkeley` needs to be removed
otherwise output is not generated :
```
docs/source/appendix/auxprogs.rst:39: WARNING: end-before pattern not found: UC Berkeley
```

Are logs actually used anywhere ? 

I have added ignore file in tree so it should not affect builds.

https://github.com/YosysHQ/yosys/commit/e2cfcbcf25afc9062d0f85748ab908c7ced71d81

KrystalDelusion:
The log files are not currently being used, but there is a todo in the page on extensions to use them instead of the inline code sections currently there.

``` rst
.. todo:: update to use :file:`/code_examples/extensions/test*.log`
```

PR #4340, opened 2024-04-15
add support for using ABCs library merging when providing multiple liberty files

gadfort:
Changes:
- adds the `-m` flag to the ABC script when multiple liberty files are specified to indicate to ABC that is should merge the libraries, this solves the issue of only the last liberty file specified getting used for synthesis.

Requirements:
- needs the abc fork to be synced with the main abc to get the new flag.

### DIFF
--- a/passes/techmap/abc.cc
+++ b/passes/techmap/abc.cc
@@ -804,8 +804,10 @@ void abc_module(RTLIL::Design *design, RTLIL::Module *current_module, std::strin
 		for (std::string dont_use_cell : dont_use_cells) {
 			dont_use_args += stringf("-X \"%s\" ", dont_use_cell.c_str());
 		}
+		bool first_lib = true;
 		for (std::string liberty_file : liberty_files) {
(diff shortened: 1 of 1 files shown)


QuantamHD:
@povik This one is also useful to Google. Any chance the Yosys team could propritize it?

povik:
I am testing this change against current ABC. When I pass two Liberty files, the cells from the first one get prefixed with `L1_` while the cells from the second get no prefix. That looks weird.

I saw there's the `read_lib -p` option, but we don't use it with neither of the `read_lib` invocations.

povik:
> I've added a PR to fix that https://github.com/berkeley-abc/abc/pull/287.

I will watch that one get merged so we update our fork after that.

> We shouldn't use the -p since that is what is supposed to add the prefixes to the names, which is not the behavior we want.

OK, I thought so.

gadfort:
@povik should have been merged now.

povik:
Let's go ahead and merge this even if the vendored ABC isn't at the right version yet. There's no harm.

PR #4343, opened 2024-04-18
ABC: read_lib args and placeholders

phsauter:
Adds a way to give read_lib additional arguments. 

This can for example be used to define `-S <slew> -G <gain>` so ABC will calculate its internal delay model from the standard cells liberty data. Without this it uses a unit delay model, which may produce worse results.  
An example use:
```
abc -liberty "$tech_cells" -D $period -script "$script" -liberty_args "-S 20 -G 3" -showtmp
```

I changed it so the placeholders like `{D}` are also replaced in user provided scripts. This seems reasonable to me as OpenLane, OpenROAD-flow-scripts etc currently use the same placeholders but have to replace it themself.  
Additionally, there is a new placeholder called `{tmpdir}`, it contains the path to the yosys-abc directory currently being used. This allows users to temporarily store information from their user script.

### DIFF
--- a/passes/techmap/abc.cc
+++ b/passes/techmap/abc.cc
@@ -707,7 +707,8 @@ void abc_module(RTLIL::Design *design, RTLIL::Module *current_module, std::strin
 		std::vector<std::string> &liberty_files, std::vector<std::string> &genlib_files, std::string constr_file,
 		bool cleanup, vector<int> lut_costs, bool dff_mode, std::string clk_str, bool keepff, std::string delay_target,
 		std::string sop_inputs, std::string sop_products, std::string lutin_shared, bool fast_mode,
-		const std::vector<RTLIL::Cell*> &cells, bool show_tempdir, bool sop_mode, bool abc_dress, std::vector<std::string> &dont_use_cells)
+		const std::vector<RTLIL::Cell*> &cells, bool show_tempdir, bool sop_mode, bool abc_dress, std::vector<std::string> &dont_use_cells,
+		std::string liberty_args)
 {
 	module = current_module;
 	map_autoidx = autoidx++;
@@ -806,7 +807,7 @@ void abc_module(RTLIL::Design *design, RTLIL::Module *current_module, std::strin
 		}
 		bool first_lib = true;
 		for (std::string liberty_file : liberty_files) {
-			abc_script += stringf("read_lib %s %s -w \"%s\" ; ", dont_use_args.c_str(), first_lib ? "" : "-m", liberty_file.c_str());
+			abc_script += stringf("read_lib %s %s -w %s \"%s\" ; ", dont_use_args.c_str(), first_lib ? "" : "-m", liberty_args.c_str(), liberty_file.c_str());
 			first_lib = false;
 		}
 		for (std::string liberty_file : genlib_files)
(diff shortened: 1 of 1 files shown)


passes/techmap/abc.cc
```diff
@@ -849,6 +861,7 @@ void abc_module(RTLIL::Design *design, RTLIL::Module *current_module, std::strin
 		for (size_t pos = abc_script.find("dretime;"); pos != std::string::npos; pos = abc_script.find("dretime;", pos+1))
 			abc_script = abc_script.substr(0, pos) + "dretime; retime -o {D};" + abc_script.substr(pos+8);
 
+	// replace placeholders in abc.script and user.script
```
povik, line 864: Please factor out the substitution operation so that it's not open-coded many times

passes/techmap/abc.cc
```diff
@@ -819,17 +820,28 @@ void abc_module(RTLIL::Design *design, RTLIL::Module *current_module, std::strin
...
 				else if (script_file[i] == ',')
 					abc_script += " ";
 				else
 					abc_script += script_file[i];
-		} else
-			abc_script += stringf("source %s", script_file.c_str());
+		} else {
+			// read in user script
+			std::ifstream ifs(script_file.c_str());
+			if(ifs.is_open()) {
```
povik, line 837: Please fix whitespace/indentation in this block

passes/techmap/abc.cc
```diff
@@ -876,6 +908,13 @@ void abc_module(RTLIL::Design *design, RTLIL::Module *current_module, std::strin
 	fprintf(f, "%s\n", abc_script.c_str());
 	fclose(f);
 
+	buffer = stringf("%s/user.script", tempdir_name.c_str());
```
povik, line 911: Do we need the `abc.script`/`user.script` separation? Why not write both to the same file?

passes/techmap/abc.cc
```diff
@@ -1743,6 +1782,10 @@ struct AbcPass : public Pass {
 				liberty_files.push_back(args[++argidx]);
 				continue;
 			}
+			if (arg == "-liberty_args" && argidx+1 < args.size()) {
```
povik, line 1785: Document this option

passes/techmap/abc.cc
phsauter: The `abc.script` and `user.script` separation is how it currently works. It basically separates the setup (loading the network and library and later writing it out) from the actual optimization and mapping script.

A different approach (and the one I would like to move in) is to give an option to not use the `abc.script` part and instead expect it to be part of the custom `user.script` and then just replace the placeholders in `user.script`.

This would give full freedom in what people do with ABC, going as far as invoking multiple ABC scripts in parallel and selecting one based one some criteria before returning to Yosys.

ShinyKate:
@povik anything needed from our side on this?

donn:
Also fwiw neither [OpenLane](https://github.com/The-OpenROAD-Project/OpenLane/blob/b729aa514c499a24467377dd9910fa78858ebadf/scripts/yosys/synth.tcl) nor [its successor](https://github.com/librelane/librelane/blob/4dc53448d98f3688e0096873a88200b04d268c6c/librelane/scripts/pyosys/synthesize.py#L408) manually substitute `{D}`. We were under the impression ABC would be the one subbing in `{D}`, not Yosys…

KrystalDelusion:
@donn from what I can tell the `-D` option never makes it to abc (except when replacing `{D}` in the abc.script) so I don't think that is possible and it's probably just ignoring the `{D}` 😬

donn:
…well, quite serendipitous that I was on the developer call where this PR was discussed then 🙃 thank you so much for checking

phsauter:
I have time to work more on tools again, yay!  
Since `abc.cc` changed quite a bit in the meantime I would suggest the following:
- Close this PR
- Restart and port #4592 and continue work on that

I think in general there are a few things that would be nice to have:
- Customize which exe and how it is called
- Support multiple script to ease organisation and reuse
- Replace placeholders in all scripts
- Expose some extra information via placeholders (eg tmpdir)

All of this is done by the other PR.  
The other option is to close both and implement the above mentioned things more piece-wise.  
What is your opinion?

PR #4350, opened 2024-04-22
rtlil: Add packed `extract` implementation for `SigSpec`

jix:
Previously `extract` on a `SigSpec` would always unpack it. Since a significant amount of `SigSpec`s have one or few chunks, it's worth having a dedicated implementation.

This is especially true, since the RTLIL frontend calls into this for every `wire [lhs:rhs]` slice, making this `extract` take up 40% when profiling `read_rtlil` with one of the largest coarse grained RTLIL designs I had on hand.

With this change the `read_rtlil` profile looks like I would expect it to look like, but I noticed that a lot of the other core RTLIL methods also are a bit too eager with unpacking or implementing `SigChunk`/`Const` overloads that just convert to a single chunk `SigSpec` and forward to the implementation for that, when a direct implementation would avoid temporary std::vector allocations. While not relevant for `read_rtlil`, to me it looks like there might be a few easy overall performance gains to be had by addressing this more generally.

### DIFF
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -4435,8 +4435,39 @@ RTLIL::SigSpec RTLIL::SigSpec::extract(int offset, int length) const
 	log_assert(offset >= 0);
 	log_assert(length >= 0);
 	log_assert(offset + length <= width_);
-	unpack();
+
 	cover("kernel.rtlil.sigspec.extract_pos");
+
+	if (packed())
+	{
+		if (chunks_.size() == 1)
+			return chunks_[0].extract(offset, length);
+
+		SigSpec extracted;
+		int end = offset + length;
+		int chunk_end = 0;
+
+		for (auto const &chunk : chunks_)
+		{
+			int chunk_begin = chunk_end;
+			chunk_end += chunk.width;
+			int extract_begin = std::max(chunk_begin, offset);
+			int extract_end = std::min(chunk_end, end);
+			if (extract_begin >= extract_end)
+				continue;
+			int extract_offset = extract_begin - chunk_begin;
+			int extract_len = extract_end - extract_begin;
+			if (extract_offset == 0 && extract_len == chunk.width)
+				extracted.chunks_.emplace_back(chunk);
+			else
+				extracted.chunks_.emplace_back(
+					chunk.extract(extract_offset, extract_len));
+		}
+
+		extracted.width_ = length;
+		return extracted;
+	}
+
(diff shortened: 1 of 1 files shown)


kernel/rtlil.cc
```diff
@@ -4435,8 +4435,39 @@ RTLIL::SigSpec RTLIL::SigSpec::extract(int offset, int length) const
...
+			if (extract_begin >= extract_end)
+				continue;
+			int extract_offset = extract_begin - chunk_begin;
+			int extract_len = extract_end - extract_begin;
+			if (extract_offset == 0 && extract_len == chunk.width)
+				extracted.chunks_.emplace_back(chunk);
+			else
+				extracted.chunks_.emplace_back(
+					chunk.extract(extract_offset, extract_len));
+		}
```
povik, line 4465: I feel like this is more complicated then it needs to be. What about this?

```
SigSpec extracted;
extracted.width_ = length;

auto it = chunks_.begin();
for (; it != chunks_.end() && offset; offset -= it->width, it++) {
    if (offset < it->width) {
        extracted.chunks_.emplace_back(it->extract(offset, it->width - offset));
        length -= it->width - offset;
        break;
    }
}
for (; it != chunks_.end() && length; length -= it->width, it++) {
    if (length >= it->width) {
        extracted.chunks_.emplace_back(*it);
    } else {
        extracted.chunks_.emplace_back(it->extract(0, length));
        break;
    }
}

return extracted;
```
Not that I have compile-tested it.

povik: The `it != chunks_.end()` checks should be superfluous even.

povik:
> While not relevant for read_rtlil, to me it looks like there might be a few easy overall performance gains to be had by addressing this more generally.

I also have the impression from looking at some profilesm and knowing how a lot of operations can trigger unpacking/packing.

kernel/rtlil.cc
jix: I don't really care about the specific implementation, the way I wrote it was just the first thing I thought of where I could quickly convince myself of its correctness and when it made this disappear from the profile I was looking at I didn't spend any time thinking of alternatives.

povik: Sure, but now that we know the perfomance implication I propose the above.

povik: and it's missing an `it++;` above the first break

jix:
Your new implementation looks good to me

povik:
Thanks!

RCoeurjoly (approved):
looks good to me. I checked with valgrind and test defvalue_ref.sv and no memory issues

povik:
Let me merge it then

PR #4351, opened 2024-04-22
Bump abc

povik:
This pulls:

- &mfs fix: YosysHQ/abc#30
- &if workaround: YosysHQ/abc#31
- out-of-tree build support (ahead of switching abc to be a submodule)
- other upstream changes (including the one required for #4340)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -166,7 +166,7 @@ bumpversion:
 # is just a symlink to your actual ABC working directory, as 'make mrproper'
 # will remove the 'abc' directory and you do not want to accidentally
 # delete your work on ABC..
-ABCREV = b502f00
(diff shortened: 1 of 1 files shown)


povik:
That pesky WASM build -- good thing we added it to the CI. Next time I push for defining `int system(const char *arg) { return -1; }` instead of patching every occurrence of `system()` in ABC source.

whitequark:
> That pesky WASM build -- good thing we added it to the CI.

It's really easy to build it locally by the way--you can check out yowasp/yosys and run `./build.sh`, it even downloads the toolchain for you. Just point the yosys-src submodule to whichever you like.

PR #4362, opened 2024-04-29
export define marco to qtcreator.config

offline3:
without defining marco in qtcreator.config, some code will be regarded as a comment by qt creator


mmicko (approved):
Works fine now, thanks.

PR #4364, opened 2024-04-29
Reduce action spam when a PR source branch is on YosysHQ/yosys

KrystalDelusion:
Introduce [fkirc/skip-duplicate-actions](https://github.com/marketplace/actions/skip-duplicate-actions) to skip duplicate action runs.
Also remove deprecated `emcc.yml`.


mmicko (changes requested):
This is fine regarding functionality.

Please just rebase/resolve conflicts since there were changes in files. Note that this configuration change was needed for abc as submodule https://github.com/YosysHQ/yosys/pull/4243/commits/09b9aa83c9ec7c1e7cf64ae267d5947c4b962b9a 

Also emcc.yml could be fully removed as agreed on meeting so that can be part of this PR as well.

mmicko (approved):

PR #4365, opened 2024-05-02
techmap: add dynamic cell type test

widlarizer:
Is this behavior expected? A foo of width 4 gets techmapped to something named LCU_10. I expected the WIDTH parameter to be inferred. Are port sizes allowed to be mismatched when techmapping?

### DIFF
--- a/tests/techmap/techmap_chtype.ys
+++ b/tests/techmap/techmap_chtype.ys
@@ -1,0 +1,33 @@
+read_verilog <<EOT
+
+(* techmap_celltype="foo" *)
+module _80_lcu_primitive(P, G, CI, CO);
+	parameter WIDTH = 10;
+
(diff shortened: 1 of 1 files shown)


povik:
I bet this is because you are missing a call to ‘wreduce’. I suggest you do ‘prep’ in between reading the input and techmapping.

povik:
Actually no, this has nothing to do with wreduce. The WIDTH won’t get inferred, it needs to be an explicit parameter value on the cell, otherwise the default takes effect.

widlarizer:
Makes perfect sense, fixed

povik:
I had to change the base branch since `techmap-chtype` I happened to already merge. Please rebase accordingly

povik (approved):
This needs rebasing otherwise looks good

PR #4367, opened 2024-05-03
intel_alm: drop quartus support

Ravenslofty:
I have some spare cycles to work on this, so let's get some things out of the way while I can.

The biggest issue with supporting Quartus is that it imposes awkward requirements on memories that cannot easily be satisfied:
- memories must be grouped together with "logical RAM names", which is incredibly unclear and not very helpful.
- the Quartus cells have parameters to support memory initialisation, but it ignores them, forcing you to use their proprietary .mif format (formats like .hex get converted to .mif, so .mif might as well be used directly). Yosys does not have the infrastructure to output a bunch of files like this.

This means that Arria V and Cyclone 10GX support gets removed too, but with no easy way to test for regressions, they're extremely vulnerable to bitrot, and I'd rather explicitly remove support than let it quietly break.


gatecat (approved):

whitequark:
> * Yosys does not have the infrastructure to output a bunch of files like this.

This is actually something we've discussed on a JF a few months ago (see the notes) and it's something I'm still interested in.

Issue #4383, opened 2024-05-09
Build error: `make: *** [Makefile:810: abc/abc] Error 2`


dopamane:
### Version

Yosys 0.37+1 (git sha1 37a6c9a09, clang 10.0.0-4ubuntu1 -fPIC -Os)

### On which OS did this happen?

Linux

### Reproduction Steps

```
git pull
make -j8
```

### Expected Behavior

Build without error.

### Actual Behavior

```
...
[ 99%] Building techlibs/quicklogic/ql_dsp_macc.o
[ 99%] Building abc/abc
make: the '-f' option requires a non-empty string argument
Usage: make [options] [target] ...
Options:
  -b, -m                      Ignored for compatibility.
  -B, --always-make           Unconditionally make all targets.
  -C DIRECTORY, --directory=DIRECTORY
                              Change to DIRECTORY before doing anything.
  -d                          Print lots of debugging information.
  --debug[=FLAGS]             Print various types of debugging information.
  -e, --environment-overrides
                              Environment variables override makefiles.
  --eval=STRING               Evaluate STRING as a makefile statement.
  -f FILE, --file=FILE, --makefile=FILE
                              Read FILE as a makefile.
  -h, --help                  Print this message and exit.
  -i, --ignore-errors         Ignore errors from recipes.
  -I DIRECTORY, --include-dir=DIRECTORY
                              Search DIRECTORY for included makefiles.
  -j [N], --jobs[=N]          Allow N jobs at once; infinite jobs with no arg.
  -k, --keep-going            Keep going when some targets can't be made.
  -l [N], --load-average[=N], --max-load[=N]
                              Don't start multiple jobs unless load is below N.
  -L, --check-symlink-times   Use the latest mtime between symlinks and target.
  -n, --just-print, --dry-run, --recon
                              Don't actually run any recipe; just print them.
  -o FILE, --old-file=FILE, --assume-old=FILE
                              Consider FILE to be very old and don't remake it.
  -O[TYPE], --output-sync[=TYPE]
                              Synchronize output of parallel jobs by TYPE.
  -p, --print-data-base       Print make's internal database.
  -q, --question              Run no recipe; exit status says if up to date.
  -r, --no-builtin-rules      Disable the built-in implicit rules.
  -R, --no-builtin-variables  Disable the built-in variable settings.
  -s, --silent, --quiet       Don't echo recipes.
  -S, --no-keep-going, --stop
                              Turns off -k.
[...]

PR #4384, opened 2024-05-09
Fix `check-git-abc` Makefile target

dopamane:

https://github.com/YosysHQ/yosys/issues/4383

Add the condition that if `git submodule status abc` results in a string starting with `-` then the git submodule is uninitialized.

Clone the repo, run `make` or `make check-git-abc`, examine instructions to initialize the submodule, `git submodule update --init`, re-run `make` or `make check-git-abc` and there should not be any error.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -783,14 +783,14 @@ $(PROGRAM_PREFIX)yosys-config: misc/yosys-config.in
 .PHONY: check-git-abc
 
 check-git-abc:
-	@if [ ! -d "$(YOSYS_SRC)/abc" ]; then \
+	@if [ ! -d "$(YOSYS_SRC)/abc" ] || [ -z "$(git submodule status abc | grep '^-' )" ]; then \
 	    echo "Error: The 'abc' directory does not exist."; \
 			echo "Initialize the submodule: Run 'git submodule update --init' to set up 'abc' as a submodule."; \
 	    exit 1; \
 	elif git -C "$(YOSYS_SRC)" submodule status abc 2>/dev/null | grep -q '^ '; then \
 	    echo "'abc' is a git submodule. Continuing."; \
 	    exit 0; \
-	elif ([ -f "$(YOSYS_SRC)/abc/.gitcommit" ] && grep -q '\$$Format:%h\$$' "$(YOSYS_SRC)/abc/.gitcommit") || [ -z "$(git submodule status abc | grep '^-' )" ]; then \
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -123,6 +123,8 @@ Makefile.
 	$ vi Makefile            # ..or..
 	$ vi Makefile.conf
 
+Initialize the `abc` submodule with `git submodule update --init`.
```
whitequark, line 126: You also need to update the submodule each time you pull, so this isn't enough.

Makefile
```diff
@@ -790,7 +790,7 @@ check-git-abc:
 	elif git -C "$(YOSYS_SRC)" submodule status abc 2>/dev/null | grep -q '^ '; then \
 	    echo "'abc' is a git submodule. Continuing."; \
 	    exit 0; \
-	elif [ -f "$(YOSYS_SRC)/abc/.gitcommit" ] && grep -q '\$$Format:%h\$$' "$(YOSYS_SRC)/abc/.gitcommit"; then \
+	elif ([ -f "$(YOSYS_SRC)/abc/.gitcommit" ] && grep -q '\$$Format:%h\$$' "$(YOSYS_SRC)/abc/.gitcommit") || [ -z "$(git submodule status abc | grep '^-' )" ]; then \
```
dopamane, line 793: -z or -n?

README.md
dopamane: How about "After cloning or pulling, intialize the `abc` submodule with `git submodule update --init`?

Makefile
dopamane: -z

dopamane: Weird... on a fresh clone running `make check-git-abc` returns the expected error message. However, running `make`, still results in the error message seen in the issue statement.

mmicko:
@dopamane thanks for changes suggestions.  Please note there are couple of cases we need to handle here
1. no "abc" directory ( ask for submodule init)
2. there is "abc" directory but it is empty ( ask for submodule init)
3. there is existing "abc" from before the change and there is .gitcommit (with "$Format:%h$")  ask for it to be removed and initialized
4. there is existing "abc" with .gitcommit containing  git revision number (this is fine it is unpacked abc tarball)

Last one is special case we need to handle for users not using git but  building all from source tarballs.
Meta code that should cover this

```
if abc dir exists 
   if abc/.gitcommit is not "$Format:%h$"
          all fine abc tarball, just exit
   else if abc submodule status is "-" 
          using old abc checkout, remove and initialize
   else 
         all fine
else 
   ask for submodule init
 
```

mmicko:
@dopamane Could you please take a look at https://github.com/YosysHQ/yosys/pull/4386 and if this one works fine for you.

PR #4387, opened 2024-05-09
docs: Document $lut and $sop

widlarizer:
`$lut` is described with an expression. I tried that for `$sop` but it wasn't readable, so there is an example instead


KrystalDelusion (approved):
LGTM.

Issue #4385, opened 2024-05-09
Documentation is unreadable if the system theme is dark


whitequark:
E.g. [here](https://yosyshq.readthedocs.io/projects/yosys/en/latest/using_yosys/synthesis/memory.html#example):

![Screenshot_20240509_070157](https://github.com/YosysHQ/yosys/assets/54771/7b29daf5-d0b9-49d1-8546-45f6090c6af4)

I've [fixed this problem before](https://pypi.org/project/sphinxcontrib-yowasp-wavedrom/) so let me know if you want assistance in doing this in a sphinx-compatibel way without JS.

jix:
On my [personal blog](https://jix.one/proving-50-year-old-sorting-networks-optimal-part-1/#identifying-sorting-networks) I'm handling diagrams with some CSS that inverts the brightness but keeps the hues roughly the same, might be useful here too:

```css
filter: url("data:image/svg+xml,<svg xmlns='http%3A//www.w3.org/2000/svg'><filter id='f'><feColorMatrix color-interpolation-filters='sRGB' type='matrix' values='1.47 -1.73 -0.467 0 0.867 -0.733 0.467 -0.467 0 0.867 -0.667 -1.07 1.07 0 0.867 0 0 0 1.0 0'></feColorMatrix></filter></svg>#f");
```

The `color-interpolation-filters='sRGB'` is needed to get consistent behavior across browsers even though it's arguably wrong. The exact transformation matrix also depends on the foreground and background colors used in the theme (mostly because mapping `#000` in diagrams made for light themes to `#fff` in dark themes is too bright and it should be mapped to the body text color instead).

whitequark:
Oh, that's way more advanced than what I did! Very clever. How's the performance like?

jix:
I haven't noticed any negative performance impact of this, including on mobile browsers, but my blog also isn't the most demanding page in general. AFAICT from chrome's rendering dev tools the SVGs are still rendered once (with the filter applied) into tiles/layers shared with the surrounding content, so unless it's used for animated or otherwise dynamic content I wouldn't expect this to have a significant impact on performance. I didn't find a way to easily check how Firefox behaves here.

jix:
[...]

PR #4391, opened 2024-05-10
Dark theme images for docs

KrystalDelusion:
#4385

New `invert-helper` css class uses svg color filter matrix from @jix (via https://github.com/YosysHQ/yosys/issues/4385#issuecomment-2102187927) to invert brightness for images when using dark theme.

Make docs locally, check images in both light and dark mode, using both the forced light/dark mode selector and the auto mode with browser/OS selected light/dark mode.


jix (approved):
tested it locally, lgtm

PR #4392, opened 2024-05-10
test-docs should not ignore docs changes

KrystalDelusion:
https://github.com/YosysHQ/yosys/actions/runs/9038879272 skipped test-docs because `pre_job` ignores changes to docs, but it should not.

Add a different skip check for the docs that doesn't ignore the docs directory.

### DIFF
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -62,10 +62,6 @@ jobs:
         run: |
           ./yosys-config || true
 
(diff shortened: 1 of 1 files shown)


mmicko (approved):
LGTM. Was looking if we could optimize jobs by putting test-docs as part of test-verific script, it would be faster that way, but also in case of error could be confusing, so lets leave it like this

PR #4396, opened 2024-05-14
Add note on docs to clarify verific support

KrystalDelusion:
Having a verific license does not provide access to the verific frontend.  This helps to make that clearer.

Add page detailing how to build yosys+verific.


docs/source/yosys_internals/extending_yosys/build_verific.rst
```diff
@@ -93,6 +92,10 @@ Yosys builds:
 Partially supported builds
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+.. todo:: still unclear on the purpose of this section.
+
+   Are these the configurations we have tested as being able to compile?
```
nakengelhardt, line 97: Yes, exactly. We know they compile, we think they will work correctly, but we don't work with those configurations ourselves so we don't know for sure, and if someone does run into an issue we can't guarantee we'll be able to fix it.

docs/source/yosys_internals/extending_yosys/build_verific.rst
```diff
@@ -0,0 +1,153 @@
...
+
+   The YosysHQ specific extensions are only available with the TabbyCAD suite.
+
+Required Verific features
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The following features, along with their corresponding Yosys build parameters,
+are required for the Yosys-Verific patch:
+
+* RTL elaboration with
```
nakengelhardt, line 83: @KrystalDelusion This line somehow got interpreted as a description and put in `<dt>` tags, which makes it show as bold(ish?) in browser. Can you see if you can get rid of that?

PR #4404, opened 2024-05-21
box_derive: New command to derive modules for boxes

povik:
In certain uses cases you want to derive modules for blackbox instances, which are otherwise ignored by the `hierarchy` process. Here we add a new command, `box_derive`, by which one can ask to have modules derived for a selected set of blackbox instances.

I can think of two use cases:

 * Functional modelling of blackboxes (the abc9 integration has an `abc9_ops -prep_hier` command to a similar end, we can merge those commands eventually)
 * Deriving of one of the possible implementations for a blackbox (where the blackbox is a blackbox because an implementation hasn't been selected yet)

An ~~incomplete~~ test is attached.


povik:
```

    bbox_derive [-base <base_module>] [-naming_attr <attr>] [selection]

As part of the assembly of the design hierarchy done by the 'hierarchy' command,
specializations of parametric modules are derived on demand: for each choice of
parameter values appearing in the design, a copy of the parametric module is
derived which is specialized to that choice.

This derivation process ignores blackboxes and whiteboxes. To supplement, this
'bbox_derive' command can be used to request the derivation of modules based on
blackbox or whitebox instances appearing in the design, which is desirable in
certain use cases. Only the selected cells are considered as the instances that
steer the derivation process.

    -base <base_module>
        instead of deriving the module that directly corresponds to each blackbox
        instance, derive a specialization of <base_module> (this option applies
        to all selected blackbox cells)

    -naming_attr <attr>
        once a specialization is derived, use the value of the module attribute
        <attr> for a name which should be used for the derived module (this
        replaces the internal Yosys naming scheme in which the names of derived
        modules start with '$paramod$')

```

whitequark (commented):
Bikeshed: should this not be called just `blackbox_derive`? We have `blackbox` already, so it'd be inconsistent otherwise.

tests/various/bbox_derive.ys
```diff
@@ -0,0 +1,37 @@
...
+
+	bb #(.X(1)) bb1(.d(d), .q(q3));
+	bb #(.X(3)) bb2(.d(d), .q(q4));
+
+	cc #(.X(1)) cc1(.d(d), .q(q5));
+	cc #(.X(1)) cc2(.d(d), .q(q6));
+endmodule
+EOF
+
+bbox_derive -naming_attr final_name top
```
povik, line 37: The PR is a draft because I am not sure how to test `aa1` etc. exists when those modules are whiteboxes (they get ignored by the selection syntax I think).

whitequark:
`box_derive` seems nice, considering you also handle whiteboxes.

tests/various/box_derive.ys
```diff
@@ -0,0 +1,47 @@
...
+EOF
+
+box_derive -naming_attr final_name top
+select -assert-mod-count 1 =bb1
+select -assert-mod-count 0 =bb2
+select -assert-mod-count 1 =bb3
+
+select -assert-mod-count 1 =cc1
+select -assert-mod-count 0 =cc2
+select -assert-mod-count 0 =cc3
```
nakengelhardt, line 44: Why no asserts on `aa1` and `aa2`?

passes/cmds/select.cc
```diff
@@ -1065,6 +1065,10 @@ struct SelectPass : public Pass {
 		log("        selection is non-empty. i.e. produce an error if no object or module\n");
 		log("        matching the selection is found.\n");
 		log("\n");
+		log("    -assert-mod-count N\n");
+		log("        do not modify the current selection. instead assert that the given\n");
+		log("        selection contains exactly N modules.\n");
```
nakengelhardt, line 1070: Should we add a note that this will count partially selected modules, and not check that the module is fully selected? Right now it's ambiguous from the description what exactly the "contains" relationship is.

nakengelhardt (approved):

passes/cmds/select.cc
povik: I think we should. Let's do

> do not modify the current selection. instead assert that the given selection contains exactly N modules (partially or fully selected)

Issue #4405, opened 2024-05-21
CXXRTL: Information loss when outputs are aliased


noeamiot:
### Version

Yosys 0.39

### On which OS did this happen?

Linux

### Reproduction Steps

For the following `and_ti.v` file:
```
module top(input a0, input a1, input b0, input b1, input z, output c0, output c1);
    wire a0b0, a0b1, a1b0, a1b1;
    wire a0b1z, a1b0z;
    wire culprit;

    assign a0b0 = a0 & b0;
    assign a0b1 = a0 & b1;
    assign a1b0 = a1 & b0;
    assign a1b1 = a1 & b1;

    //assign a1b0 = a1b0 ^ z;
    //assign a0b1 = a0b1 ^ z;
    assign a0b1z = a0b1 ^ z;
    assign a1b0z = a1b0 ^ z;
    
    //assign c0 = a0b0 ^ (a0b1 ^ z);
    //assign c1 = a1b1 ^ (a1b0 ^ z);
    assign culprit = a1b1 ^ a1b0z; 
    assign c0 = culprit;
    assign c1 = a0b0 ^ a0b1z;
endmodule
```

The workflow used is simple:
```
read_verilog and_ti.v 
write_cxxrtl -O6 -g4 -header cxxrtl_and_ti.cpp
```
Gives the following debug information for the wires `c0` and `c1` :
```
items->add(path + "culprit", debug_item(debug_eval_outline, p_culprit, 0), metadata_map({
	{ "src", "and_ti.v:4.10-4.17" },
}));
...
items->add(path + "c1", debug_item(p_c1, 0, debug_item::OUTPUT|debug_item::DRIVEN_COMB), metadata_map({
	{ "src", "and_ti.v:1.79-1.81" },
}));
items->add(path + "c0", debug_item(debug_eval_outline, p_culprit, 0), metadata_map({
	{ "src", "and_ti.v:4.10-4.17" },
}));
```

There are other combinations of HDL or `-O` flags that can lead to this behavior, sometimes on an outline like here, sometimes on normal wires.

### Expected Behavior

There exists a way to retrieve the information that `c0` is an output wire like for `c1`.

### Actual Behavior

There is no way to know that `c0` is an output because it is aliased to another wire that is not an output.

whitequark:
Thanks for the report. Yeah, ports should probably just not be aliasable.

whitequark:
Do you think you could write a patch?

noeamiot:
Sure !

PR #4406, opened 2024-05-21
cxxrtl: Prevent wires with input or output ports from being aliased

noeamiot:
As explained in #4405, having outputs or inputs aliased causes a loss of information in `debug_info`.

To prevent this behavior, we stop assigning debug information at the beginning of the `debug_alias` option corresponding logic if the wire is an input and/or an output.

One implementation choice here was to choose, for an input that is tied to a constant, whether we should generate:
```
items->add(path, "a0", "src\000sand_ti.v:1.18-1.20\000", p_a0, 0, debug_item::INPUT|debug_item::DRIVEN_COMB);
```
or
```
static const value<1> const_p_a0 = value<1>{0u};
items->add(path, "a0", "src\000sand_ti.v:1.18-1.20\000", const_p_a0);
```

This patch is an implementation for the first one, because we can still query the constant later in the first whereas we would loose the `INPUT` and `DRIVEN_COMB` flags in the second one.

### DIFF
--- a/backends/cxxrtl/cxxrtl_backend.cc
+++ b/backends/cxxrtl/cxxrtl_backend.cc
@@ -3263,6 +3263,7 @@ struct CxxrtlWorker {
 						debug_wire_type = wire_type; // wire is a member
 
 					if (!debug_alias) continue;
+					if (wire->port_input || wire->port_output) continue; // preserve input/output metadata in flags
 					const RTLIL::Wire *it = wire;
 					while (flow.is_inlinable(it)) {
 						log_assert(flow.wire_comb_defs[it].size() == 1);


backends/cxxrtl/cxxrtl_backend.cc
```diff
@@ -3263,6 +3263,7 @@ struct CxxrtlWorker {
 						debug_wire_type = wire_type; // wire is a member
 
 					if (!debug_alias) continue;
+					if (wire->port_input || wire->port_output) continue; // no aliases for inputs and outputs
```
whitequark, line 3266: I think we could go with a more descriptive comment here, something like this:

```suggestion
					if (wire->port_input || wire->port_output) continue; // preserve input/output metadata in flags
```

whitequark (approved):
LGTM, thanks!

PR #4409, opened 2024-05-22
Only build yosys-abc usage docs when ABCEXTERNAL is not set

marysaka:
Since https://github.com/YosysHQ/yosys/pull/4334, when building docs with `ABCEXTERNAL` set, the rule for yosys-abc docs usage would attempt to build in-tree ABC.

This fix this regression by only including yosys-abc in `DOCS_USAGE_STDERR` when `ABCEXTERNAL` is not set.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -946,7 +946,13 @@ define DOC_USAGE_STDERR
 docs/source/generated/$(1): $(PROGRAM_PREFIX)$(1) docs/source/generated
 	-$(Q) ./$$< --help 2> $$@
 endef
-DOCS_USAGE_STDERR := yosys-config yosys-filterlib yosys-abc
+DOCS_USAGE_STDERR := yosys-config yosys-filterlib
+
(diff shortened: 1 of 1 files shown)


Ravenslofty (approved):

KrystalDelusion:
How does this affect the [auxiliary programs](https://yosyshq.readthedocs.io/projects/yosys/en/latest/appendix/auxprogs.html) docs page?

marysaka:
> How does this affect the [auxiliary programs](https://yosyshq.readthedocs.io/projects/yosys/en/latest/appendix/auxprogs.html) docs page?

Before #4334, it would generate an empty usage silently for `yogsys-abc`.

With my changes it match previous behavior

Here is the pdf side generated with `DOC_TARGET=latexpdf` for 0.41 + this patch:
image

KrystalDelusion (approved):

PR #4412, opened 2024-05-23
ci: Skip duplicate actions on nix build

KrystalDelusion:
nix test builds run on both push and pull request, resuling in duplicated runs for PRs from this repo.

Move nix build to extra-builds.yml and use the conditional skipping.  This also skips commits that only change docs.

@RCoeurjoly can you confirm that it is okay to skip these tests if a) a duplicate is already running (e.g. push triggers before pull request) and/or b) only docs files have changed?


RCoeurjoly:
yes, that looks reasonable to me.

RCoeurjoly (approved):

mmicko (approved):

Issue #4402, opened 2024-05-18
Synthesis issue with taking the negative sign of 0


WeneneW:
### Version

Yosys 0.39+165

### On which OS did this happen?

Linux

### Reproduction Steps

Consider the following code. When the input is `wire0 = 6'b111101`, it is a negative number less than `forvar14`. Therefore, `reg16` takes the value of `$unsigned((-forvar9))`, which is `-0`, and the output `y` is 0. However, when using Yosys synthesis, the output result is 1.

```
module top  (y, clk, wire0);
  output wire y;
  input wire clk;

  input wire signed [5:0] wire0;

  reg reg16 = (1'h0);

  reg signed [2:0]forvar14 = (1'h0);

  reg [2:0] forvar10 = (1'h0);
  reg [2:0] forvar9 = (1'h0);

  assign y = reg16;
  always
    @(posedge clk) begin
       for (forvar14 = (1'h0); (forvar14 < (1'h1)); forvar14 = (forvar14 + (1'h1)))
        begin
       reg16 = ($signed((wire0 <= forvar14)) ? $unsigned((-forvar9)) : (^~$signed(forvar10)));
       end
    end
endmodule

```

testbench
```
`include "syn_yosys.v"

`timescale 1ns / 1ps

module testbench;

  reg clk;
  reg signed [5:0] wire0;
  wire y;


  top uut (
    .y(y),
    .clk(clk),
    .wire0(wire0)
  );


  initial begin

    clk = 0;
    wire0 = 6'b111101; 


    forever #5 clk = ~clk;
  end


  initial begin

    #100;
    $display("y = %d", y);
    $finish;
  end

  initial
    begin
      $dumpfile("wave_yosys.vcd");
      $dumpvars(0, testbench);
    end
endmodule

```
Here is the directory structure of the attachment. The `syn_yosys.v` file is the synthesized file. 

In the `identity` directory, execute the following commands:
```
iverilog -o identity_main identity_testbench.v > icarus_stderr.log 2>&1
vvp -n identity_main -lxt2 > vvp_identity.log
```

In the `syn_yosys` directory, execute the following commands:
```
iverilog -o yosys_main yosys_testbench.v > icarus_stderr.log 2>&1
vvp -n yosys_main -lxt2 > vvp_yosys.log
```
to reproduce the issue.
[yosys_5_18.zip](https://github.com/YosysHQ/yosys/files/15365896/yosys_5_18.zip)

```markdown
📂 identity
[...]

PR #4416, opened 2024-05-26
fix(#4402):missing sign while for loop iteration variable is signed

YoYo-0513:
**Reseaon**
 issue #4402 

**RCA**:
Like the code blow, iteration variable `forvar14` is signed. However, the iteration variable lost the signed information during the ast simplify. For example, the value of `forvar14` should be `$signed(3'h0)` instead of `3'h0` in the first loop cycle.
```
reg signed [2:0] forvar14 = (1'h0);
for (forvar14 = (1'h0); (forvar14 < (1'h1)); forvar14 = (forvar14 + (1'h1)))  begin
      .......
end
```
**Fixes**:
1. backends/verilog/verilog_backend.cc
    Signed should be writed while dump verilog;

2. frontends/ast/genrtlil.cc
    Before this change, children[0] will use the children[1]'s `width_hint` and `sign_hint` when `genRTLIL`, because they use the same variables and  children[1]'s will overide children[0]'s.

3. frontends/ast/simplify.cc
    Record iteration variable is signed or not;

**Test**
A case was provided in #4402 , use PR branch to build a new yosys, read the 'rtl.v' and dump the verilog after synthesis. Then follow the reproduce steps in issue #4402 , result will meet expectations.

### DIFF
--- a/frontends/ast/genrtlil.cc
+++ b/frontends/ast/genrtlil.cc
@@ -1840,8 +1840,8 @@ RTLIL::SigSpec AstNode::genRTLIL(int width_hint, bool sign_hint)
 			int width = max(width_hint, 1);
 			width_hint = -1, sign_hint = true;
 			children[0]->detectSignWidthWorker(width_hint, sign_hint);
-			children[1]->detectSignWidthWorker(width_hint, sign_hint);
 			RTLIL::SigSpec left = children[0]->genRTLIL(width_hint, sign_hint);
+			children[1]->detectSignWidthWorker(width_hint, sign_hint);
 			RTLIL::SigSpec right = children[1]->genRTLIL(width_hint, sign_hint);
 			RTLIL::SigSpec sig = binop2rtlil(this, type_name, width, left, right);
 			return sig;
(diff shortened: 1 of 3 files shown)


frontends/ast/genrtlil.cc
```diff
@@ -1840,8 +1840,8 @@ RTLIL::SigSpec AstNode::genRTLIL(int width_hint, bool sign_hint)
 			int width = max(width_hint, 1);
 			width_hint = -1, sign_hint = true;
 			children[0]->detectSignWidthWorker(width_hint, sign_hint);
-			children[1]->detectSignWidthWorker(width_hint, sign_hint);
 			RTLIL::SigSpec left = children[0]->genRTLIL(width_hint, sign_hint);
+			children[1]->detectSignWidthWorker(width_hint, sign_hint);
```
zachjs, line 1844: I'm a bit surprised at by this change. It seems like we probably need to account for the sign and width of both sides of the comparison before generating the RTLIL for _either_ side. The existing code should produce the Verilog semantics we're looking for: the comparison is done on signed operands with _both_ are signed, and using the maximum of the widths of each side. I'm sure I'm missing something here. Can you help me understand?

zachjs (commented):
Thank you for submitting these 3 bug fixes! Do you think you could also include (perhaps 3 separate) test cases for these bug fixes?

* See `tests/verilog/const_arst.ys` for an existing test that exercises `write_verilog`, then `read_verilog`, and confirms the two versions match (yours could likely be simpler for checking this particular bug).
* There are plenty of existing examples testing Verilog expression semantics in `tests/verilog/` and `tests/various/`, which could be applicable for your other two changes.

PR #4417, opened 2024-05-26
cxxrtl: don't emit invalid code on unconnected outputs.

kivikakk:
Unconnected sync outputs cause `dump_cell_sync` to emit an invalid assignment in the instantiating module, namely `value<0>() = …;`.

Ensure the LHS is not empty before emitting.


whitequark (approved):
Thanks.

PR #4425, opened 2024-06-03
sigmap: comments

widlarizer:
This PR adds developer comments to sigmap and mfp


kernel/sigtools.h
```diff
@@ -302,6 +312,7 @@ struct SigMap
 
 	inline void add(Wire *wire) { return add(RTLIL::SigSpec(wire)); }
 
+	// Lazy version of add (doesn't promote if const)
```
jix, line 315: I think this comment is misleading, this is usually used as a representative lookup that returns the representative by updating the passed argument in-place. I'm not quite sure where the `add` overload that takes just a single SigBit is used or useful for, but the other `add` overloads don't do anything similar.

jix (changes requested):

kernel/sigtools.h
widlarizer: Changed to `// Modify bit to its representative`, does that make sense?

jix (approved):

PR #4433, opened 2024-06-07
cxxrtl: capi: don't use deprecated invocation.

kivikakk:
The C API uses the deprecated invocation format.

Also correct the deprecation message, and don't suggest a value for a parameter the caller already has.

Assuming neither this nor #4417 gets a response, I'll stop submitting CXXRTL fixes.


whitequark (approved):

whitequark:
There's a merge conflict.

whitequark:
> Assuming neither this nor #4417 gets a response, I'll stop submitting CXXRTL fixes.

It's not clear to me why but I've not received a notification for either PR. (I'm subscribed to all events in this repository, I'm a code owner for CXXRTL, and I checked my GitHub settings and haven't found anything there that would prevent it.)

kivikakk:
That might've been something on my end; should be addressed now. (I was also confused by the lack of notifications while trying to find out where #4417 went, only to find it was merged!)

Conflict has been remedied.

backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
```diff
@@ -1582,7 +1582,7 @@ struct module {
 
 	// Compatibility method.
 #if __has_attribute(deprecated)
-	__attribute__((deprecated("Use `debug_info(&items, /*scopes=*/nullptr, path);` instead. (`path` could be \"top \".)")))
```
whitequark, line 1585: This particular line was previously fixed in https://github.com/YosysHQ/yosys/pull/4432/files. (I don't really have an opinion on which message is better; this is a comment on the merge conflict.)

kivikakk: Yepyep — I resolved the conflict by preferring mine (as they modified the base of the deprecation message the same way), since the `` `path` could be "top ". `` bit seems confusing at best.

(At least, the first time I saw this message, it seemed to imply I'd need to add some kind of path, otherwise why suggest a value for one? But callers of the deprecated API already have a `path` value, since the signature is indeed `void debug_info(debug_items &items, std::string path)`, so it seems best to just remove the suggestion.)

whitequark: Sounds good.

whitequark:
Thanks!

kivikakk:
Thank you!

PR #4434, opened 2024-06-07
Fix memory leak in verific file parsing.

mikesinouye:
Running AddressSanitizer on yosys when using verific uncovered a memory leak in a `file_names` Array, as the char* copy created by `strdup()` is not freed.

Manually iterate over the array and free the memory, similar to other instances in the `verific.cc` file.


mmicko:
Thanks

PR #4437, opened 2024-06-10
rtlil.cc: Fix #4427

### LINKED ISSUE #4427: Yosys Verilog Parsing Error: Unable to Synthesize After Reading File
### Version

yosys 0.41+126

Linux

### Reproduction Steps

Hello,
I encountered an error while using Yosys to read a Verilog file. The steps and the error message are as follows:
截图 2024-06-04 21-27-09

After this error, the parsing process is interrupted. I have checked the Verilog file for syntax errors but couldn't find any obvious issues. I am using the latest version of Yosys.

Attached is the Verilog file (design.v) that triggers the error.
Thank you in advance for your attention to this matter.
I look forward to hearing from you regarding this issue.
[design_file.zip](https://github.com/user-attachments/files/15552274/design_file.zip)

### Expected Behavior

synthesis success

### Actual Behavior

synthesis fail

KrystalDelusion:

If a `RTLIL::Const` is composed of multiple strings, such as when using a ternary expression to select between two strings of different lengths, zero padding for the strings needs to be maintained.

Only leading (and trailing) null characters should be dropped from the decoded string, rather than all null characters.

### DIFF
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -325,8 +325,7 @@ std::string RTLIL::Const::decode_string() const
 				ch |= 1 << j;
 			}
 		}
-		if (ch != 0)
-			s.append({ch});
+		s.append({ch});
 	}
 	i -= 8;
 	for (; i >= 0; i -= 8) {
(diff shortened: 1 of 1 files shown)


KrystalDelusion:
Actually I think while this does prevent the assertion in #4427, it leads to an incorrect result.  The problem was that the "t" would get width extended to match the length of "e...", and then concatenated to the end of "Q...", so that one side of the assertion is "Q...t\0" while the other is "Q...\0t".  I changed the left to match the right, but I think the left side was actually the correct one and it is the right side that needs to be changed...

ShinyKate:
Just adding you as assignee so it's pulled from the open PR list!

PR #4439, opened 2024-06-10
liberty: Support for IO liberty files for verification

gatecat:
This makes two changes to allow reading in IO liberty files, for e.g. verification or simulation purposes where the Verilog simulation models are too non-synthesisable for Yosys to be able to handle them:

 - support for the `three_state` attribute which specifies a condition on which an output is tristated, converted into a `$tribuf`
 - `inout`s with a missing function aren't skipped, as the library we are using uses them for power pins


jix (approved):
The code changes LGTM and improved liberty support for FV use cases is something I'm also indirectly interested in.

PR #4443, opened 2024-06-10
cxxxrtl: fix use of format specifiers in test

maribu:
#### What are the reasons/motivation for this change?

Correct C++ code

#### Explain how this is achieved.

A few instances of incorrect (and non-portable) use of format specifiers have been fixed.

#### If applicable, please suggest to reviewers how they can test the change.

Compilation with `-Werror=format` of the changed file should now succeed.


whitequark (approved):
Thanks!

Issue #4445, opened 2024-06-11
Performance Issue: Synthesis Takes Too Long to Complete


LoSyTe:
### Version

yosys 0.41+126

### On which OS did this happen?

Linux

### Reproduction Steps

Hello,
I encountered a performance issue while using Yosys to synthesize a Verilog file. The specific details are as follows:

I used the Verilog file design.v, and when running the Yosys synthesis, it has been over 30 minutes without completion. My system configuration is as follows: 128G RAM, and a high-performance CPU. However, the synthesis process seems to be stuck in an infinite loop or some kind of performance bottleneck.

To rule out issues with the file itself, I have checked the Verilog file for syntax errors but did not find any obvious problems. Additionally, I am using the latest version of Yosys.

Attached is the Verilog file (design.v) that triggers this issue. I hope to get the community's help and attention.

After this error, the parsing process is interrupted. I have checked the Verilog file for syntax errors but couldn't find any obvious issues. I am using the latest version of Yosys.

Attached is the Verilog file (design.v) that triggers the error.
Thank you in advance for your attention to this matter.
I look forward to hearing from you regarding this issue.
[design_file.zip](https://github.com/user-attachments/files/15786422/design_file.zip)

### Expected Behavior

Synthesis completes successfully within a reasonable time

### Actual Behavior

Synthesis takes too long and does not complete

whitequark:
You should minimize the issue. It is a waste of everyone's time to file non-minimized fuzzer-discovered issues and personally I am leaning towards closing such issues outright.

georgerennie:
I would agree with the previous comment on minimization. If you haven't seen it before https://blog.regehr.org/archives/2037 provides good guidelines on reporting fuzzed bugs. Additionally it would have been helpful to include the synthesis script used. I have assumed it was just `synth` as that causes a hang for me too.

If you aren't familiar with it, Yosys includes a testcase minimizer in the form of the `bugpoint` command. Running
`yosys -p "read_verilog design.v; bugpoint -command \"synth\" -runner \"timeout 1\" -fast; write_verilog bug.v"` results in the following testcase after only a few minutes (I've tidied the one shown here with `clean -purge; write_verilog -noattr bug.v`)

```verilog
/* Generated by Yosys 0.40+45 (git sha1 dd2195543, g++ 13.2.1 -fPIC -Os) */

module module11(_0_);
  output [3:0] _0_;
  wire [3:0] _0_;
  wire _1_;
[...]

PR #4448, opened 2024-06-12
peepopt shiftadd: Only match for sufficiently small constant widths

georgerennie:
This addresses issue #4445. There are two issues from that, one being that casting a constant of arbitrary size to an `int` can overflow, and the other being that the `shiftadd` peephole optimization creates logic that can be very large for large shift constants.

This is a simple fix to both that just disables the optimization for constants over a certain bitwidth. I have arbitrarily chosen 24 bits here but am happy to change that or discuss a more fine-grained solution than just disabling the optimization. I suspect most user code will not have offsets outside of this range.

### DIFF
--- a/passes/pmgen/peepopt_shiftadd.pmg
+++ b/passes/pmgen/peepopt_shiftadd.pmg
@@ -53,6 +53,11 @@ match add
 	select port(add, constport).is_fully_const()
 	define <IdString> varport (constport == \A ? \B : \A)
 
+	// only optimize for constants up to a fixed width. this prevents cases
+	// with a blowup in internal term size and prevents larger constants being
+	// casted to int incorrectly
+	select (GetSize(port(add, constport)) <= 24)
+
 	// if a value of var is able to wrap the output, the transformation might give wrong results
 	// an addition/substraction can at most flip one more bit than the largest operand (the carry bit)
 	// as long as the output can show this bit, no wrap should occur (assuming all signed-ness make sense)


phsauter:
As mentioned in the issue:
Verilog-2005 specifies in 4.3.1 the following:
The smallest allowed upper limit on the size of a vector (number of wires/bits) is 65536 ($2^{16}$).

So a $2^{24}$ limit is certainly a good general choice.  
Since this is only an optimization and it works without it, it would also be legal to select a smaller upper limit, this would be a good idea if we think it could degrade performance (of the circuit or the synthesis process). 

In this particular case I could see it degrading the synthesis process in particular during `techmap`.  
Shift operations of the form `data >> ... -c` are transformed into a padded data instead. This is fine in isolation but then during `techmap` it seems to first build the full barrel shifter and only then optimize it using the constant input signals (padding of data). With a shift of $2^{24}/24-1$* this would create tens of millions of `$_MUX_` cells and then optimize them away.  
I don't know how big of an impact this has on the runtime and peak memory.  

The 'nice' way to solve this would probably involve mapping shift operations first to `$mux` instead and then during the simple-mapping of `$mux` it is possible to check if `A==B` and if this is the case, skip creating this `$_MUX_` cell and instead just connect the wires.  
The simpler way is to measure the performance impact of a max-size shift, its currently running.


\* something in `techmap` checks the expression width and limits it to 24-bit apparently, hence a smaller limit applies or we run into an error:
```
ERROR: Expression width 50331679 exceeds implementation limit of 16777216!
```

widlarizer (approved):
Since #4455, shift size is sufficiently limited, and this safety limit is safe to implement

Issue #4413, opened 2024-05-24
Yosys right shift error


Noah-S-E:
### Version

Yosys 0.39+165

### On which OS did this happen?

Linux

### Reproduction Steps

Consider the following code: 
`shift_amount` has a bit width of `3` bits, and when it is `3'b111`, the value of `shift_result_lsb` should be `1`. This means that `shift_result_lsb` can take the value `1`. However, after synthesis with Yosys, the value of `shift_result_lsb` is fixed at `1'b0`.
```verilog
module combinational_logic_module (
    // --- Outputs ---
    output wire       shift_result_lsb,         // The LSB of the final shifted result
    output wire [2:0] shift_amount,             // The 3-bit amount for the shift operation
    output wire [2:0] truncated_add_result,     // The truncated result of the addition

    // --- Inputs ---
    input  wire       clock,                    // Clock input (currently unused)
    input  wire       control_input             // Single-bit control input
);

    // --- Internal Constant Definitions ---
    // Using localparam for better readability and maintainability
    localparam [7:0] ADDITION_CONSTANT = 8'hA0; // 8'b10100000, or 160 in decimal
    localparam [7:0] VALUE_TO_SHIFT    = 8'h9C; // 8'b10011100, or 156 in decimal

    // --- Combinational Logic Implementation ---

    // Step 1: Add the 8-bit constant to the 1-bit input.
    //         The $unsigned cast ensures an unsigned addition. The result is implicitly
    //         truncated to match the 3-bit width of 'truncated_add_result'.
    //         e.g., 10100000 (160) + 1 = 10100001 (161). Truncated to 3 bits, this is 3'b001.
    assign truncated_add_result = $unsigned(ADDITION_CONSTANT + control_input);

    // Step 2: Subtract 1 from the truncated result of the previous step to get the shift amount.
    //         This is a 3-bit subtraction, which will underflow if the result is negative (e.g., 000 - 1 = 111).
    assign shift_amount = truncated_add_result - 1'b1;

    // Step 3: Right-shift the 'VALUE_TO_SHIFT' by the calculated 'shift_amount'.
    //         The output 'shift_result_lsb' is the least significant bit (bit 0) of the shifted value.
    assign shift_result_lsb = (VALUE_TO_SHIFT >> shift_amount);

endmodule

```

The synthesized `syn_yosys.v` is as follows: (Note the `assign y = 1'h0;`)

```
[...]

PR #4452, opened 2024-06-13
peepopt: avoid shift-amount underflow

phsauter:
This solves issue #4413. `shiftadd` did not consider underflows for unsigned shift-amounts.  
Specifically the issue arises if peepopt receives an expression like 
```verilog
wire [6:0] shift;
wire [7:0] data, shift2;
assign shift2 = shift - 1'b1;
assign out = data >> shift2;
```
Here an _intended_ underflow occurs if `shift<1`. (by _intended_ I mean it is what the current network implies, no matter what the designer might have wanted).  
This adds a filter for the case described and adds the bug to the tests.


widlarizer (approved):
The added test does fail prior to this PR, so it covers the added logic, which seems sound

PR #4460, opened 2024-06-17
Make C++17 compiler required

mmicko:
Team decided it is time to update minimal requirement for C++ compiler to be at least C++17 compliant.

Change in header file (that is used everywhere) makes sure that requirement is enforced.

Also updated workflow to run now tests with new minimal requirements.

### DIFF
--- a/.github/workflows/test-compile.yml
+++ b/.github/workflows/test-compile.yml
@@ -64,11 +64,11 @@ jobs:
           $CXX --version
 
       # minimum standard
-      - name: Build C++11
+      - name: Build C++17
         shell: bash
         run: |
           make config-$CC_SHORT
-          make -j$procs CXXSTD=c++11 compile-only
+          make -j$procs CXXSTD=c++17 compile-only
 
       # maximum standard, only on newest compilers
       - name: Build C++20
--- a/Makefile
+++ b/Makefile
@@ -90,7 +90,7 @@ all: top-all
 YOSYS_SRC := $(dir $(firstword $(MAKEFILE_LIST)))
 VPATH := $(YOSYS_SRC)
 
-CXXSTD ?= c++11
+CXXSTD ?= c++17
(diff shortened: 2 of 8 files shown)


povik:
These Ubuntu 20.04 commands may be at odds with the C++17 requirement: https://github.com/YosysHQ/yosys/pull/4460/files#diff-7678ac5f848b6f19279c04c8df2c8547cc23e1d2875678f826e929f6153318d1L93-L100

whitequark:
I wonder if I should upgrade CXXRTL to C++17 as well then...

widlarizer:
> These Ubuntu 20.04 commands may be at odds with the C++17 requirement: https://github.com/YosysHQ/yosys/pull/4460/files#diff-7678ac5f848b6f19279c04c8df2c8547cc23e1d2875678f826e929f6153318d1L93-L100

We're good on Ubuntu 20.04. Tested in docker with the apt install command you linked to:
```
root@19879717f761:/# clang --version
clang version 10.0.0-4ubuntu1
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
root@19879717f761:/# cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=20.04
DISTRIB_CODENAME=focal
DISTRIB_DESCRIPTION="Ubuntu 20.04.6 LTS"
root@19879717f761:/# clang -std=c++17 a.cpp
/usr/bin/ld: /usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../x86_64-linux-gnu/crt1.o: in function `_start':
(.text+0x24): undefined reference to `main'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
root@19879717f761:/# clang -std=c++1237 a.cpp
error: invalid value 'c++1237' in '-std=c++1237'
```

povik:
Good. Thanks for the confirmation

KrystalDelusion (approved):

povik (approved):

PR #4465, opened 2024-06-20
Test build against upstream ABC

povik:

This is a canary PR to see if Yosys passes tests when built against upstream ABC.


whitequark:
```
make[1]: *** No rule to make target 'src/map/if/acd/ac_wrapper.o', needed by 'libabc.a'.  Stop.
```

Not the WASI failure I expected...

PR #4467, opened 2024-06-21
rtlil: Fix `addShiftx` for signed shifts

povik:

This fixes the helper creating invalid RTLIL when called with `is_signed` set to true.

See here from the commit message:

> Only the `B` input (the shift amount) can be marked as signed on a `$shiftx` cell. Adapt the helper accordingly and prevent it from creating invalid RTLIL when called with `is_signed` set. Previously it would mark both `A` and `B` as signed.
 
There's no way to test this without writing external code, I think.


widlarizer (approved):
This matches the constraint set by the internal cell checker

Issue #3713, opened 2023-03-25
Synthesis results can change based on comments or blank lines in Verilog source


algofoogle:
### Version

Yosys 0.22 (git sha1 f109fa3d4c5, gcc 8.3.1 -fPIC -Os)

### On which OS did this happen?

Linux

### Reproduction Steps

# tl;dr

If you add just the right number of comments-only header lines (e.g. license text) or blank lines to the top of a Verilog file being run through yosys, you can find that in some cases it will change the synthesis result even though the rest of the file is unchanged.

This emerged through OpenLane, and I'm too new to yosys to know how to reproduce this completely with yosys alone, but hopefully I can still provide enough information or be guided to do so.

# Background

I encountered this using [efabless/caravel_user_project](https://github.com/efabless/caravel_user_project) (latest tag `mpw-9a`), via OpenLane, and reported it in https://github.com/The-OpenROAD-Project/OpenLane/issues/1697. Here's an example simple design intended for the sky130A PDK that demonstrates the issue:

[MYTEST.tar.gz](https://github.com/YosysHQ/yosys/files/11068620/MYTEST.tar.gz) - Contents:
```
verilog/rtl/MYTEST.v          # Top design; instantiates MYTEST_sub.
verilog/rtl/MYTEST_sub.v      # The file where blank lines change synthesis results!!
openlane/MYTEST/config.json   # Uses the above 2 Verilog files.
```

Using OpenLane with Caravel, we'd run `make MYTEST` and grab the synthesis result from `openlane/MYTEST/runs/MYTEST/*/results/synthesis/MYTEST.v`.

If you run the above `make MYTEST`, you'll get a certain result in `results/synthesis/MYTEST.v`.

If you add 20 blank lines to the top of the source `MYTEST_sub.v` file and run the make again, you'll get a **different** result in `results/synthesis/MYTEST.v`.

If you add 10 more blank lines (30 total) and run the make again, you'll get the **SAME** result as for only 20 blank lines.

If you add 10 more (40 total), you'll get a **DIFFERENT** result yet again.

Results for these 4 cases in my OpenLane environment are attached: 
[MYTEST_synth.tar.gz](https://github.com/YosysHQ/yosys/files/11068647/MYTEST_synth.tar.gz)

### Expected Behavior

Whitespace that doesn't actually alter the meaning of the Verilog code (especially adding blank lines or comment-only lines) should not change the synthesis result, allowing a consistent result regardless of whether comments (or blank lines) are used or not.

[...]

PR #4468, opened 2024-06-25
remove sorts from some optimisation passes

Ravenslofty:
I was given a test case that is an example of #3713, where changing the name of a variable changed synthesis results. I tracked that down to a handful of `design->sort()` calls which sort the wires in the design alphabetically; removing them leads to identical synthesis results for the changed names.

This isn't really a *fix* for #3713, but it does alleviate some of the pain.

### DIFF
--- a/passes/opt/opt.cc
+++ b/passes/opt/opt.cc
@@ -182,7 +182,6 @@ struct OptPass : public Pass {
 		}
 
 		design->optimize();
-		design->sort();
 		design->check();
 
 		log_header(design, fast_mode ? "Finished fast OPT passes.\n" : "Finished OPT passes. (There is nothing left to do.)\n");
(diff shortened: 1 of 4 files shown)


widlarizer:
But there must have been some intention behind these sorts. Avoiding reordering parts of a design before vs after a command seems just about as valid a UX requirement as "a newline doesn't change how my stuff synthesizes"

Ravenslofty:
Well, based on how the Anlogic latch test produces very different synthesis results on macOS vs Linux, perhaps one of those reasons was to mitigate platform differences.

But from my experiments, Yosys is generally pretty good at preserving the order of things in the network anyway.

widlarizer:
I don't believe #3713 is as severe as unknown changes to the risk of platform dependent results. Have we previously identified the sources of platform dependent behavior inside yosys so that I could review this PR by verifying whether any apply here?

Issue #4458, opened 2024-06-17
Yosys Synthesis Error: Hash Table Exceeded Maximum Size


LoSyTe:
### Version

yosys 0.41+126

### On which OS did this happen?

Linux

### Reproduction Steps

Hello,
I encountered another issue while using Yosys to synthesize a Verilog file. The specific details are as follows:
I used the Verilog file rtl.v, and during the Yosys synthesis, a yosys_stderr.log file was generated with the following content:

“terminate called after throwing an instance of 'std::length_error'
  what():  hash table exceeded maximum size.”

The synthesis process was interrupted due to this error. I suspect this is because the design file is too large, causing the hash table to exceed its maximum size limit. To rule out issues with the design file itself, I have checked the structure of the Verilog file but did not find any obvious problems. Additionally, I am using the latest version of Yosys.

Attached is the Verilog file (rtl.v) that triggers this issue. I hope to get the community's help and attention.
[design_file.zip](https://github.com/user-attachments/files/15866575/design_file.zip)

### Expected Behavior

synthesis success

### Actual Behavior

synthesis fail

whitequark:
You should minimize the reproducer further.

mtuomi:
I have previously fixed this locally by adding large prime like 7718116163 at the end of zero_and_some_primes array in hashlib.h (inside hashtable_size function). I couldn't reproduce the problem with your design_file.zip. I've seen this only with really massive designs with hundreds of millions of gates.

Kmanfi:
Seen also same  error with huge designs. Haven’t made an issue here coz privacy of those designs.
I confirm: By adding additional prime will fix the issue, as explained in previous comment.

georgerennie:
The reason you are hitting this limit seems to be because of the large structures from your large constant shifts interacting badly with peepopt shiftadd, the same thing that caused issues in #4445. Please minimize and look into these cases yourself first.

If you try running with patch #4455, this issue isn't hit as it doesn't generate these massive intermediate nodes.

whitequark:
Thanks, the new reproducer is vastly more useful.

georgerennie:
Could you double check you have the patch from #4455 applied correctly? For me on latest main (1288166f7ac03456b6beb0acff5d4fc84f63f57c) without it I can reproduce your issue still with the new rtl, but when I cherry pick that patch synthesis completes quickly without issue.

I applied the patch by doing something along the lines of
```
git fetch
git checkout main
[...]

PR #4471, opened 2024-07-01
hashlib: Add some more primes

georgerennie:
Add some primes as suggested in comments in #4458. This allows larger hashtables to be allocated for very big designs.
The new primes match the geometric ratio of ~1.25 used for the others and don't overflow 4 byte ints.


jix:
463830273 and 1132398207 aren't primes. I've used the following sage math code to compute the extended the sequence:

```sage
a = [23]
def next_or_same_prime(n):
    # this matches the nextprime OEIS uses here
    # sage's next_prime always returns a greater prime and requires an integer argument
    return next_prime(ceil(n) - 1)
while a[-1] < (1<<31):
    a.append(next_or_same_prime(a[-1] + (a[-1] + 1) / 4))
a.pop() # remove the last prime that overflows a 32-bit signed int
print(a)
```

```
[23, 29, 37, 47, 59, 79, 101, 127, 163, 211, 269, 337, 431, 541, 677, 853, 1069, 1361, 1709, 2137, 2677, 3347, 4201, 5261, 6577, 8231, 10289, 12889, 16127, 20161, 25219, 31531, 39419, 49277, 61603, 77017, 96281, 120371, 150473, 188107, 235159, 293957, 367453, 459317, 574157, 717697, 897133, 1121423, 1401791, 1752239, 2190299, 2737937, 3422429, 4278037, 5347553, 6684443, 8355563, 10444457, 13055587, 16319519, 20399411, 25499291, 31874149, 39842687, 49803361, 62254207, 77817767, 97272239, 121590311, 151987889, 189984863, 237481091, 296851369, 371064217, 463830313, 579787991, 724735009, 905918777, 1132398479, 1415498113, 1769372713]
```

georgerennie:
Oh you are totally right, I'm not even quite sure how I managed to mess that one up, I was obviously not fully awake yesterday.

widlarizer (approved):
All added numbers are primes, and adding more than is necessary isn't a problem

PR #4472, opened 2024-07-01
Let makefile run without git

### LINKED ISSUE #4470: Can't build from 0.42 tarball
### Version

The 0.42 tarball (can't build, so can't give yosys --version otuput)

Linux

### Reproduction Steps

```
tar -xzf yosys-yosys-0.42.tar.gz
cd yosys-yosys-0.42/
LANG=C make
```

### Expected Behavior

yosys builds (like it does for me when using the 0.40 or 0.42 tarballs).

### Actual Behavior

yosys fails to build, giving this erroer message:
```
Initialize the submodule: Run 'git submodule update --init' to set up 'abc' as a submodule.
make: *** [Makefile:748: check-git-abc] Error 1
```

KrystalDelusion:

Fix #4470

Test if current directory is a git repository, if it's not suggest calling `make wget-abc`.
Add `wget-abc` target, which attempts to download and extract the corresponding ABC release archive.

Check "Build without abc or git" should pass.

### DIFF
--- a/.github/workflows/extra-builds.yml
+++ b/.github/workflows/extra-builds.yml
@@ -120,7 +120,13 @@ jobs:
 
       - name: Setup environment
         uses: ./.github/actions/setup-build-env
-      
+
+      - name: Check abc
+        shell: bash
+        run: |
+          ! make check-git-abc
+          make wget-abc
+
       - name: Build
         shell: bash
         run: |
--- a/.github/workflows/extra-builds.yml
+++ b/.github/workflows/extra-builds.yml
@@ -125,7 +125,7 @@ jobs:
         shell: bash
         run: |
           ! make check-git-abc
-          make wget-abc
+          make curl-abc
 
       - name: Build
         shell: bash
--- a/.github/workflows/extra-builds.yml
+++ b/.github/workflows/extra-builds.yml
@@ -108,6 +108,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           sparse-checkout: |
+            .github/actions
             .
             backends
             frontends
--- a/.github/workflows/extra-builds.yml
+++ b/.github/workflows/extra-builds.yml
@@ -95,3 +95,34 @@ jobs:
         with:
           install_url: https://releases.nixos.org/nix/nix-2.18.1/install
       - run: nix build .?submodules=1
+
+  sparse-build:
+    name: "Build without abc or git" # emulate build from release tarball
+    needs: pre_job
+    if: needs.pre_job.outputs.should_skip != 'true'
+    runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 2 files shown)


Ravenslofty (approved):

KrystalDelusion:
"Build without abc or git" check may fail if/when the target ABC revision has been updated since the last release.  Not quite sure what to do in that case.  In theory the `wget-abc` target should only be used when coming from a release tarball, so ignoring the test failure might be fine?  Leaving this as a draft for now in case someone can come up with a better solution.

@mmicko @RCoeurjoly any ideas?

RCoeurjoly:
> "Build without abc or git" check may fail if/when the target ABC revision has been updated since the last release. Not quite sure what to do in that case. In theory the `wget-abc` target should only be used when coming from a release tarball, so ignoring the test failure might be fine? Leaving this as a draft for now in case someone can come up with a better solution.
> 
> @mmicko @RCoeurjoly any ideas?

I don´t understand that failure case. Could you expand on how that failure would be triggered?
I don´t see any problems with the PR as it stands

mmicko:
From implementation side this all looks fine. It is more decision about which use cases are supported and which require some more user involvement. Also fact that that for macOS for example curl is coming bundled with OS and not wget would require one additional target to use curl to achieve same goal.

whitequark (commented):
Why not upload complete tarballs with abc and attach them to releases instead? This can be done automatically via Actions. Then you could simply error out if `abc` is empty, perhaps with a message that includes the correct URL for the release artifact.

KrystalDelusion:
> I don´t understand that failure case. Could you expand on how that failure would be triggered?

As an example `make curl-abc` matches on the last release (0.43) which had [abc @ 237d813](https://github.com/YosysHQ/abc/tree/237d81397fcc85dd3894bf1a449d2955cd3df02d), but building for 0.43+22 is expecting the newer [abc @ 28d955c](https://github.com/YosysHQ/abc/tree/28d955ca97a1c4be3aed4062aec0241a734fac5d).  That change in abc could(?) lead to a build failure.  This _should_ only happen during github runs of the `sparse-build` test, which I added to test that building is possible without git.  So I can foresee an edge case where that one test is failing pre-release, and it hides something that has broken that support which we then don't notice until post-release (when we would expect the test to pass again).

> Why not upload complete tarballs with abc and attach them to releases instead? This can be done automatically via Actions. Then you could simply error out if abc is empty, perhaps with a message that includes the correct URL for the release artifact.

Uhhh, I didn't really look into changing the releases because I don't have admin access to the repository (with the assumption that I would need that access to be able to change it), whereas this I could throw together on my own...

KrystalDelusion:
Closing in preference of #4668

PR #4474, opened 2024-07-03
Add PolarFire FPGA support

tony-min-1:
- Adding new support for PolarFire FPGA as `synth_mchp` pass

- The overall flow is very similar to synth_xilinx. 
- Added DSP and RAM inference based on PolarFire [macro library guide](https://coredocs.s3.amazonaws.com/Libero/2021_2/Tool/pf_mlg.pdf) and [user fabric guide](https://ww1.microchip.com/downloads/aemDocuments/documents/FPGA/ProductDocuments/UserGuides/PolarFire_FPGA_PolarFire_SoC_FPGA_Fabric_UG_VD.pdf)
  - DSP inference achieved using `pmgen`, RAM inference uses `memory_libmap` pass
- Rest of simulation model taken from existing SmartFusion2 support because it has similar architecture to PolarFire

- Added some basic verilog tests. Ran equivalence check on the synthesized output Verilog

Credits to Damon Ma for adding the Verilog tests + code cleanups


techlibs/mchp/cells_sim.v
```diff
@@ -0,0 +1,848 @@
...
+		if (En == 1) begin
+			if (Set == 0)
+				Q <= 1;
+			else
+				Q <= D;
+		end
+	end
+
+	specify
+		$setup(D , posedge CLK &&& En && Set, 0); // neg setup not supported?
```
Ravenslofty, line 179: this is something of an implementation limitation; a setup constraints is internally a buffer with a specified delay, but ABC9 will not let you specify negative delays for things.

Ravenslofty (changes requested):
So, there's a lot of code here. Here are some general thoughts before going over this with a fine-toothed comb.

- Yosys always spells out the name of vendors and families; `synth_lattice`, `synth_ice40`. so, the `mchp` and `pf` abbreviation should be `microchip` and `polarfire`.
- architecture-specific tests go in `tests/arch/XXX`, rather than `techlibs/XXX/tests`. I suggest referencing the generic testbench files in `tests/arch/common`, and also looking at how synthesis testbenches are idiomatically formed.

techlibs/mchp/cells_sim.v
```diff
@@ -166,15 +166,20 @@ module MCHP_SYNC_SET_DFF(
 	input En,
 	output Q);
 	parameter [0:0] INIT = 1'b0; // unused
+
+	reg q_ff;
```
Ravenslofty, line 170: wouldn't it be easier to just make the output a `reg` instead?

tony-min-1: good point, updated

tony-min-1:
> So, there's a lot of code here. Here are some general thoughts before going over this with a fine-toothed comb.
> 
> * Yosys always spells out the name of vendors and families; `synth_lattice`, `synth_ice40`. so, the `mchp` and `pf` abbreviation should be `microchip` and `polarfire`.
> * architecture-specific tests go in `tests/arch/XXX`, rather than `techlibs/XXX/tests`. I suggest referencing the generic testbench files in `tests/arch/common`, and also looking at how synthesis testbenches are idiomatically formed.

- changed all references of `mchp` to `microchip` and `pf` to `polarfire`
- Moved all tests into `tests/arch/microchip` and setup the tests in same way as other architecture specific tests
  - enabled microchip tests in Makefile

techlibs/microchip/cells_sim.v
```diff
@@ -0,0 +1,863 @@
...
+	input A
+);
+	parameter [1:0] INIT = 2'h0;
+	assign Y = INIT >> A;
+	specify
+		(A => Y) = 127;
+	endspecify
+endmodule
+
+(* abc9_lut=2 *)
```
Ravenslofty, line 54: This doesn't look right to me, though I'm unfamiliar with the PolarFire. If you have a LUT4 architecture I would expect everything to have an area of 1, which implicitly tries to make ABC9 use LUT4s for as much logic as possible. Does it really require twice as much area for a LUT2 than a LUT1, and three times as much area for a LUT3?

techlibs/microchip/cells_sim.v
```diff
@@ -0,0 +1,863 @@
...
+	output Y
+);
+	assign Y = A & B & C;
+endmodule
+
+module AND4 (
+	input A, B, C, D,
+	output Y
+);
+	assign Y = A & B & C & D;
```
Ravenslofty, line 39: Is there dedicated hardware for this, or are you expecting this to map to soft logic?

techlibs/microchip/cells_sim.v
```diff
@@ -0,0 +1,863 @@
...
+endmodule
+
+module SLE (
+	output Q,
+	input ADn,
+	input ALn,
+	(* clkbuf_sink *)
+	input CLK,
+	input D,
+	input LAT,
```
Ravenslofty, line 222: I assume this is taken from the vendor model - is `LAT` really an input, and not a parameter?

techlibs/microchip/microchip_dffopt.cc
```diff
@@ -0,0 +1,340 @@
...
+					// Merge CE LUT and D LUT into one.  If it cannot be done, nothing to do about this FF.
+					if (!merge_lut(lut_d_post_ce, lut_d, lut_ce, true, sig_Q, max_lut_size))
+						continue;
+
+					// If this gets rid of a CE LUT, it's worth it.  If not, it still may be worth it, if we can remove set/reset
+					// as well.
+					if (it_CE != bit_to_lut.end())
+						worthy_post_ce = true;
+				} else if (sig_CE.data != State::S1) {
+					// Strange.  Should not happen in a reasonable flow, so bail.
```
Ravenslofty, line 220: Maybe a `log_assert` on this to fail noisily if this occurs?

techlibs/microchip/synth_microchip.cc
```diff
@@ -0,0 +1,541 @@
...
+	SynthMicrochipPass() : ScriptPass("synth_microchip", "synthesis for Microchip FPGAs") {}
+
+	void help() override
+	{
+		//   |---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|
+		log("\n");
+		log("    synth_microchip [options]\n");
+		log("\n");
+		log("This command runs synthesis for Microchip FPGAs. Operating on\n");
+		log("partly selected designs is not supported (you must submit a fully-selected \n");
```
Ravenslofty, line 37: Generally we imply this and check it with an assert; it makes the help a little easier to understand.

techlibs/microchip/synth_microchip.cc
```diff
@@ -0,0 +1,541 @@
...
+		log("    -top <module>\n");
+		log("        use the specified module as the top module\n");
+		log("\n");
+		log("    -family <family>\n");
+		log("        Run synthesis for the specified Microchip architecture. \n");
+		log("        Generate the synthesis netlist for the specified family.\n");
+		log("        supported values:\n");
+		log("        - polarfire: PolarFire\n");
+		log("\n");
+		log("    -edif <file>\n");
```
Ravenslofty, line 50: I assume you're outputting to Libero for testing; what format does that normally use?

techlibs/microchip/synth_microchip.cc
```diff
@@ -0,0 +1,541 @@
...
+		log("\n");
+		log("    -dff\n");
+		log("        Run 'abc'/'abc9' with -dff option\n");
+		log("\n");
+		log("    -retime\n");
+		log("        Run 'abc' with '-D 1' option to enable flip-flop retiming.\n");
+		log("        implies -dff.\n");
+		log("\n");
+		log("    -abc9\n");
+		log("        Use new ABC9 flow (EXPERIMENTAL)\n");
```
Ravenslofty, line 93: I'm known for having Opinions here, but I think ABC9 should be the default, because it gives much better synthesis results, despite being somewhat unreliable; flows using classic ABC mapping make Yosys look much worse than it is.

techlibs/microchip/synth_microchip.cc
```diff
@@ -0,0 +1,541 @@
...
+		log("\n");
+		log("    -noclkbuf\n");
+		log("        Disable automatic clock buffer insertion\n");
+		log("\n");
+		log("    -run <from_label>:<to_label>\n");
+		log("        Only run the commands between the labels (see below). an empty\n");
+		log("        'from_label' is synonymous to 'begin', and empty 'to_label' is\n");
+		log("        synonymous to the end of the command list.\n");
+		log("\n");
+		log("    -flatten\n");
```
Ravenslofty, line 79: This is generally the default, and flows have a `-noflatten` option instead. Xilinx is the odd one out for not flattening by default.

tests/arch/microchip/Registers.v
```diff
@@ -0,0 +1,38 @@
...
+THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+
+module Registers(
```
Ravenslofty, line 19: It's true that we don't have an "async dff plus enable" common test, but these should still be covered indirectly through `tests/arch/common/{a,}dffs.v`, right?

tests/arch/microchip/carryout.v
```diff
@@ -0,0 +1,34 @@
...
+module carryout (cout,out,a, b,c);
+parameter n = 6;
+parameter k = 2;
+output reg [k*(n+1)-1:0] out;
+output reg cout;
+input [n:0] a;
+input [n:0] b;
+input [n-1:0] c;
+
+	always @(a,b,c)
```
Ravenslofty, line 28: I think Yosys will suggest `always @*` here, so you might as well fix it

tests/arch/microchip/large_mult.v
```diff
@@ -0,0 +1,25 @@
...
+THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+
+module large_mult(
```
Ravenslofty, line 19: Use `tests/arch/common/mul.v`

tests/arch/microchip/ram_SDP.v
```diff
@@ -0,0 +1,39 @@
...
+THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+
+module ram_SDP(data,waddr,we,clk,q);
```
Ravenslofty, line 19: The test semantics aren't exactly the same, but this looks like `sync_ram_sdp` from `tests/arch/common/blockram.v`

tests/arch/microchip/ram_TDP.v
```diff
@@ -0,0 +1,56 @@
...
+THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+
+module ram_TDP (clka,clkb,wea,addra,dataina,qa,web,addrb,datainb,qb);
```
Ravenslofty, line 19: `sync_ram_tdp` from `tests/arch/common/blockram.v`

tests/arch/microchip/simple_ram.v
```diff
@@ -0,0 +1,37 @@
...
+input [9:0] addr;
+output [19:0] dout;
+
+
+reg [9:0] addr_reg;
+reg [19:0] mem [0:1023] ;
+assign dout = mem[addr_reg];
+
+always@(posedge clk) begin
+	addr_reg <= addr;
```
Ravenslofty, line 32: That's...an odd way to write this. I'm a little confused about the intended semantics here.

tests/arch/microchip/widemux.v
```diff
@@ -0,0 +1,31 @@
...
+THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+
+module widemux(
```
Ravenslofty, line 19: `mux8` from `tests/arch/common/mux.v`

Ravenslofty (changes requested):
I think all the DSP testcases should be merged together to avoid too much litter.

But here's a first pass review.

techlibs/microchip/cells_sim.v
tony-min-1: yes, it should just be an area of 1 for all LUTs in this architecture.

techlibs/microchip/cells_sim.v
tony-min-1: This is from the macro library, I would expect this to map to a LUT4.

techlibs/microchip/cells_sim.v
tony-min-1: yes, this is a static configuration port tied to 0 or 1 (same for ADn and SD)

techlibs/microchip/synth_microchip.cc
tony-min-1: Will remove this line. There is already a check for fully selected design

techlibs/microchip/cells_sim.v
Ravenslofty: but as implemented, it won't; you will need to techmap it into soft logic, or alternatively directly into a LUT.

techlibs/microchip/cells_sim.v
Ravenslofty: then I would assume it should be a parameter instead, no? as written, you could manually instantiate one of these and then drive the `LAT` input from soft logic. I assume Libero errors out if you try to do this, but it seems like a strange decision.

tests/arch/microchip/simple_ram.v
tony-min-1: it should be functionally same as `sync_ram_sp` from `tests/arch/common/blockram.v` . I've changed to use `sync_ram_sp` instead.

tests/arch/microchip/widemux.v
tony-min-1: mux4 generated using `case` is a little different. Having a bit of trouble inferring MX4 macro

tests/arch/microchip/Registers.v
tony-min-1: I've added in `tests/arch/common/(a)dffs.v` as tests

techlibs/microchip/synth_microchip.cc
tony-min-1: There was very minimum testing through Libero. We only performed equivalence checks of post synthesis netlist (Verilog) against original netlist. 

I did try to push a very simple design through Libero. The format I used was Verilog for the input synthesized netlist.

But yeah, I guess it's probably a good idea to add -vlog option to dump Verilog.

techlibs/microchip/cells_sim.v
tony-min-1: No, these have to be static ports.

Libero will error out when I connect these static configuration ports to soft logic.

Ravenslofty: Mmm, okay.

techlibs/microchip/cells_sim.v
tony-min-1: Libero actually recognizes these macros. You can manually instantiate these cells.

tony-min-1:
> I think all the DSP testcases should be merged together to avoid too much litter.
> 
> But here's a first pass review.

Thanks for the feedbacks, I have made the following changes:

- Organize similar tests together in single file to reduce clutter
- Re-use existing tests from `tests/arch/common` 
- Fix c++20 compile error
- use abc9 by default
- flatten hierarchy by default
- add vlog option to dump Verilog
- various cleanups

techlibs/microchip/synth_microchip.cc
```diff
@@ -0,0 +1,542 @@
...
+		log("        do not flatten design before synthesis\n");
+		log("\n");
+		log("    -dff\n");
+		log("        Run 'abc'/'abc9' with -dff option\n");
+		log("\n");
+		log("    -retime\n");
+		log("        Run 'abc' with '-D 1' option to enable flip-flop retiming.\n");
+		log("        implies -dff.\n");
+		log("\n");
+		log("    -abc\n");
```
Ravenslofty, line 90: this is conventionally `-noabc9`.

techlibs/microchip/synth_microchip.cc
```diff
@@ -0,0 +1,542 @@
...
+		}
+
+		if (check_label("map_luts")) {
+			run("opt_expr -mux_undef -noclkinv");
+			if (help_mode)
+				run("abc -luts 2:2,3,6:5[,10,20] [-dff] [-D 1]", "(option for '-nowidelut', '-dff', '-retime')");
+			else if (abc9) {
+
+				std::string abc9_opts;
+				// for the if command in abc to specify wire delay between adjects LUTs (default = 0)
```
Ravenslofty, line 484:
```suggestion
				// for the if command in abc to specify wire delay between adjacent LUTs (default = 0)
```

tests/arch/microchip/widemux.v
Ravenslofty: what's the trouble, may I ask?

Ravenslofty (changes requested):
last few things I see. maybe the arch-specific tests could be inlined into the yosys scripts that use them; that would cut down the PR line count a bit more.

tests/arch/microchip/widemux.v
tony-min-1: So I'm interested in `$_MUX4_` cells. To discover this cell structure, I use `muxcover` to pack `$_MUX_` trees into `$_MUX4_`. With the `case` statement, you get `$pmux` instead of a tree of `$_MUX_`.  I couldn't find a way to cleanly convert `$pmux` into `$mux` tree. I have tried using `pmuxtree` and `techmap -map +/pmux2mux.v` but there is always additional logic (i.e. `$eq`) driving the select control of `$mux`, which I assume makes it difficult for `muxcover` to pack into `$_MUX4_`

Ravenslofty: Then the answer is that you're calling `muxcover` too early. The reason you're not getting any `$_MUX4_` cells is because you have no `$_MUX_` cells which you get from `techmap`ping `$mux` cells.

tony-min-1: before `muxcover`, I call `simplemap t:$mux` to do the conversion from `$mux` to `$_MUX_` cells. And before `simplemap t:$mux` is where I experimented with the `$pmux` -> `$mux` conversions.

tony-min-1:
- Inlined all tests into .ys script.
- Realized that PolarFire DFF does not support init. Added switch to remove init attribute, this was done the same way as in `SmartFusion2` pass.

tony-min-1:
Is there any other changes we need to make?

Ravenslofty (approved):

PR #4479, opened 2024-07-06
Gowin. Add an energy saving primitive

yrabbit:
We add a BANDGAP primitive used to turn off power to OSC, PLL and other things on some GOWIN chips.

We also mark this primitive and GSR as keep.

### DIFF
--- a/techlibs/gowin/cells_sim.v
+++ b/techlibs/gowin/cells_sim.v
@@ -867,8 +867,12 @@ module ODDRC(D0, D1, CLEAR, TX, CLK, Q0, Q1);
 	parameter INIT = 0;
 endmodule
 
+(* blackbox, keep *)
 module GSR (input GSRI);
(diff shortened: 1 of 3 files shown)


widlarizer:
This is how I'm reading the code: you introduce into `cells_sim.v` primitives with `(* blackbox, keep *)`. I'd like to understand this first - it's a blackbox since their behavior can't be modeled in Verilog, and unlike others, it's keep, because they have no outputs to be used but cause side effects of hard IP power gating. Am I getting that right?

yrabbit:
That's right. The thing is that if I don’t specify “keep”, then such primitives disappear on the path to nextpnr along with all the functionality (hidden, as you correctly noted).
There may be another way, but a simple "keep" attribute does the job for me.

widlarizer (approved):

widlarizer:
Matches SUG283

PR #4487, opened 2024-07-12
Reproducible git hash

widlarizer:

See https://github.com/YosysHQ/abc/pull/35

abc submodule is bumped. abc .gitcommit checks for `Format:%h` are replaced with checks for either  `Format:%h` or  `Format:%H`. Yosys .gitcommit is changed also to ` Format:%H`. The yosys .gitcommit check for `Format:%h` is replaced with a check for a substring check for `Format:`

Modify .gitcommit files at will to simulate tarballs instead of git checkouts and observe build success and errors to match intended behavior

### DIFF
--- a/.gitcommit
+++ b/.gitcommit
@@ -1,1 +1,1 @@
-$Format:%h$
+$Format:%H$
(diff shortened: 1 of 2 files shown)


mmicko (approved):
LGTM

PR #4489, opened 2024-07-14
Gowin. Add the DCS primitive

yrabbit:
Not so much adding the primitive itself, but only its DCS_MODE parameter, without which an error occurs.

### DIFF
--- a/techlibs/gowin/cells_sim.v
+++ b/techlibs/gowin/cells_sim.v
@@ -1901,3 +1901,14 @@ output OSCOUT;
 parameter FREQ_DIV = 100;
 parameter REGULATOR_EN = 1'b0;
 endmodule
+
+(* blackbox *)
+module DCS (CLK0, CLK1, CLK2, CLK3, CLKSEL, SELFORCE, CLKOUT);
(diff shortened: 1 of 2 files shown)


widlarizer:
Under what conditions does an error occur? Why is that?

yrabbit:
You try to use a DCS primitive with a parameter specified.
Why? Apparently because the parameters must be described, which is what this PR actually corrects - it describes the parameter.

``` verilog
DCS dcs(
    .CLK0(1'b1),
    .CLK1(clk),
    .CLK2(1'b1),
    .CLK3(1'b1),
    .CLKSEL({1'b0, 1'b0, key, 1'b0}),
    .SELFORCE(1'b1),
    .CLKOUT(clk1)
);
defparam dcs.DCS_MODE="CLK1_GND";
```

``` 
2.3.2. Analyzing design hierarchy..
Top module:  \top
ERROR: Module `DCS' referenced in module `top' in cell `dcs' does not have a parameter named 'DCS_MODE'.
```

widlarizer:
Got it, looks good, matches UG286

widlarizer (approved):

PR #4490, opened 2024-07-14
sigspec: decrease size with tagged union of bits and chunks

widlarizer:

small is fast

This PR reduces `sizeof(SigSpec)` from 64 bytes to 40 bytes. SigSpec is always strictly either in packed mode (`bits_` are empty) or unpacked (`chunks_` are empty). A vector is three pointers. This PR puts `bits_` and `chunks_` into a union tagged with `bool packed_`. Union UB outside of rtlil.cc and rtlil.h is avoided since already the current implementation of SigSpec has `bits_` and `chunks_` as private members. This brings lower runtime memory usage and improve performance with improved cache locality

- [ ] TODO: measure impact

Manually verify that all direct `chunks_` accesses happen only in branches where `packed_` is true, and `bits_ `where false. Run benchmarks, build with sanitizer, etc

### DIFF
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -3791,6 +3791,8 @@ RTLIL::SigSpec::SigSpec(std::initializer_list<RTLIL::SigSpec> parts)
 	width_ = 0;
 	hash_ = 0;
 
+	packed_ = true;
+	(void)new (&this->chunks_) std::vector<SigChunk>();
 	log_assert(parts.size() > 0);
 	auto ie = parts.begin();
 	auto it = ie + parts.size() - 1;
@@ -3802,6 +3804,8 @@ RTLIL::SigSpec::SigSpec(const RTLIL::Const &value)
 {
 	cover("kernel.rtlil.sigspec.init.const");
 
+	packed_ = true;
+	(void)new (&this->chunks_) std::vector<SigChunk>();
 	if (GetSize(value) != 0) {
 		chunks_.emplace_back(value);
 		width_ = chunks_.back().width;
@@ -3816,6 +3820,8 @@ RTLIL::SigSpec::SigSpec(RTLIL::Const &&value)
 {
 	cover("kernel.rtlil.sigspec.init.const.move");
 
+	packed_ = true;
+	(void)new (&this->chunks_) std::vector<SigChunk>();
 	if (GetSize(value) != 0) {
 		chunks_.emplace_back(std::move(value));
 		width_ = chunks_.back().width;
@@ -3830,6 +3836,8 @@ RTLIL::SigSpec::SigSpec(const RTLIL::SigChunk &chunk)
 {
 	cover("kernel.rtlil.sigspec.init.chunk");
 
+	packed_ = true;
+	(void)new (&this->chunks_) std::vector<SigChunk>();
 	if (chunk.width != 0) {
 		chunks_.emplace_back(chunk);
 		width_ = chunks_.back().width;
@@ -3844,6 +3852,8 @@ RTLIL::SigSpec::SigSpec(RTLIL::SigChunk &&chunk)
 {
 	cover("kernel.rtlil.sigspec.init.chunk.move");
 
+	packed_ = true;
+	(void)new (&this->chunks_) std::vector<SigChunk>();
 	if (chunk.width != 0) {
 		chunks_.emplace_back(std::move(chunk));
 		width_ = chunks_.back().width;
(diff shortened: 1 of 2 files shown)


kernel/rtlil.cc
```diff
@@ -3982,23 +4012,47 @@ RTLIL::SigSpec::SigSpec(bool bit)
...
+	that->packed_ = false;
+	(void)new (&that->bits_) std::vector<SigBit>();
+}
+
 void RTLIL::SigSpec::pack() const
 {
 	RTLIL::SigSpec *that = (RTLIL::SigSpec*)this;
 
-	if (that->bits_.empty())
+	if (packed_ || that->bits_.empty())
```
povik, line 4049: Shouldn't we be removing the `bits_.empty()` check? We need to guarantee accesses to `chunks_` are valid once `pack()` returns, so we need to do the switch in full no matter if `bits_` are empty.

kernel/rtlil.cc
```diff
@@ -4016,31 +4070,36 @@ void RTLIL::SigSpec::pack() const
...
+
 	check();
 }
 
 void RTLIL::SigSpec::unpack() const
 {
 	RTLIL::SigSpec *that = (RTLIL::SigSpec*)this;
 
-	if (that->chunks_.empty())
+	if (!packed_ || that->chunks_.empty())
```
povik, line 4088: Analogously here

kernel/rtlil.h
```diff
@@ -842,20 +842,25 @@ struct RTLIL::SigSpec
 {
 private:
 	int width_;
+	bool packed_;
 	unsigned long hash_;
-	std::vector<RTLIL::SigChunk> chunks_; // LSB at index 0
-	std::vector<RTLIL::SigBit> bits_; // LSB at index 0
+	union {
+		std::vector<RTLIL::SigChunk> chunks_; // LSB at index 0
+		std::vector<RTLIL::SigBit> bits_; // LSB at index 0
```
povik, line 849: Could we do this with `std::variant`?

kernel/rtlil.cc
```diff
@@ -3982,23 +4012,47 @@ RTLIL::SigSpec::SigSpec(bool bit)
...
 
+	packed_ = false;
+	(void)new (&this->bits_) std::vector<SigBit>();
 	width_ = 0;
 	hash_ = 0;
 	append(SigBit(bit));
 	check();
 }
 
+void RTLIL::SigSpec::switch_to_packed() const
```
povik, line 4023: From the naming alone it would appear this does the same thing as `pack()`, I suggest we name it `switch_union_to_chunks` or similar.

kernel/rtlil.h
widlarizer: Good idea, this is a good small but hot case to try it on

widlarizer:
Turns out this gives a 5-10% performance *penalty* across the board, even with thin LTO. Profiling should show where this is coming from. Could be constructor/destructor or swap calls on pack/unpack. Time to read what STL methods compile down to I guess

povik:
We can add a 24 byte padding to the struct to check if the penalty is due to cache alignment effects.

widlarizer:
I tried that earlier - adding `alignas(64)` to this made perf and memory consumption significantly worse actually

povik:
@widlarizer Disregarding any switch to a union, please see what kind of a penalty we get from switching to
```
void RTLIL::SigSpec::unpack() const 
{
	RTLIL::SigSpec *that = (RTLIL::SigSpec*)this;

	if (that->chunks_.empty())
		return;

	cover("kernel.rtlil.sigspec.convert.unpack");
	log_assert(that->bits_.empty());

	std::vector<RTLIL::SigChunk> old_chunks;
	old_chunks.swap(that->chunks_);

	that->bits_.reserve(that->width_);
	for (auto &c : old_chunks)
		for (int i = 0; i < c.width; i++)
			that->bits_.emplace_back(c, i);

	that->hash_ = 0;
}
```
(Note the swap in place of a clear on `that->chunks_`.)

That should inform what we do next with this PR.

povik:
> @widlarizer Disregarding any switch to a union, please see what kind of a penalty we get from switching to

@widlarizer reports that the `unpack()` change alone didn't bring a noticable performance hit, so we can't explain the 5-10 % union penalty on the basis that we are losing the `chunks_` capacity.

widlarizer:
5.88% of the 22k cache misses happen in `malloc_consolidate`  for baseline (1166238c0) but 9.20% of the 18k for the union sigspec with some fixes (6e3bf76ea). The sigspec cache misalignment theory lives on, I'd say.

PR #4494, opened 2024-07-17
Makefile: do install `bitpattern.h`

povik:

`bitpattern.h` is useful to plugins which process HDLs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -579,6 +579,7 @@ S =
 endif
 
 $(eval $(call add_include_file,kernel/binding.h))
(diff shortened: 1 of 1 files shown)


widlarizer:
Since this suggests you have the context for working with this header, now would be a good time to add some comments about its functionality. The header is short enough that a single comment about the ideas it implements could suffice

povik:
It implements a pool of bit patterns, within Yosys it's used to check whether the cases in a case statement are exhaustive. If you have already looked at the change, could you review it?

PR #4495, opened 2024-07-18
check: Skip detailed edge modeling if costly

povik:

On real world designs like the Black Parrot, `check` can get stuck on some cells because of the edge data explosion. See e.g. this shift cell

```
attribute \src "../black-parrot/external/HardFloat/source/HardFloat_primitives.v:103.10-103.17"
  cell $shr $auto$slang_frontend.cc:657:operator()$20570
    parameter \Y_WIDTH 4097
    parameter \B_WIDTH 12
    parameter \B_SIGNED 0
    parameter \A_WIDTH 4097
    parameter \A_SIGNED 1
    connect \Y $auto$slang_frontend.cc:665:operator()$20571
    connect \B $auto$slang_frontend.cc:604:operator()$20569
    connect \A 4097'10000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
  end
```

For those cells for which the edge data is available and can exhibit a quadratic explosion, we check for the product of the input and output port widths, and if the product is over a threshold, we fall back to coarse connectivity modeling (we assume all output port bits are connected too all input port bits, like we do for cells for which the edge data is unavailable today).

At the same time, we introduce a new option to force the detailed modeling:
```
    -force-detailed-loop-check
        for the detection of combinatorial loops, use a detailed connectivity
        model for all internal cells for which it is available. This disables
        falling back to a simpler overapproximating model for those cells for
        which the detailed model is expected costly.
```

See this session on this somewhat synthetic example:
```
module top(input wire [15:0] i, output wire [31:0] y);
	assign y = {16'hbf39, ~y[31:15]} >> i; 
endmodule
```

```
yosys> read_verilog loop.v
...
yosys> check

2. Executing CHECK pass (checking for obvious problems).
Checking module top...
Warning: found logic loop in module top:
    cell $shr$loop.v:2$2 ($shr) source: loop.v:2.13-2.39
      (cell's internal connectivity overapproximated; loop may be a false positive)
    cell $not$loop.v:2$1 ($not) source: loop.v:2.24-2.33
      A[0] --> Y[0]
Warning: found logic loop in module top:
    cell $shr$loop.v:2$2 ($shr) source: loop.v:2.13-2.39
      (cell's internal connectivity overapproximated; loop may be a false positive)
    cell $not$loop.v:2$1 ($not) source: loop.v:2.24-2.33
      A[1] --> Y[1]

...

Warning: found logic loop in module top:
    cell $shr$loop.v:2$2 ($shr) source: loop.v:2.13-2.39
      (cell's internal connectivity overapproximated; loop may be a false positive)
    cell $not$loop.v:2$1 ($not) source: loop.v:2.24-2.33
      A[16] --> Y[16]
Found and reported 17 problems.
Consider re-running with '-force-detailed-loop-check' to rule out false positives.

yosys> check -force-detailed-loop-check

3. Executing CHECK pass (checking for obvious problems).
Checking module top...
Found and reported 0 problems.
```

### DIFF
--- a/passes/cmds/check.cc
+++ b/passes/cmds/check.cc
@@ -316,10 +316,10 @@ struct CheckPass : public Pass {
 							SigBit edge_to = sigmap(cell->getPort(to_port))[to_bit];
 
 							if (edge_from == from && edge_to == to && nhits++ < HITS_LIMIT)
-								message += stringf("        %s[%d] --> %s[%d]\n", log_id(from_port), from_bit,
+								message += stringf("      %s[%d] --> %s[%d]\n", log_id(from_port), from_bit,
 												   log_id(to_port), to_bit);
 							if (nhits == HITS_LIMIT)
-								message += "        ...\n";
+								message += "      ...\n";
 						}
 					};
 
--- a/passes/cmds/check.cc
+++ b/passes/cmds/check.cc
@@ -59,6 +59,12 @@ struct CheckPass : public Pass {
 		log("    -assert\n");
 		log("        produce a runtime error if any problems are found in the current design\n");
 		log("\n");
+		log("    -force-detailed-loop-check\n");
+		log("        for the detection of combinatorial loops, use a detailed connectivity\n");
+		log("        model for all internal cells for which it is available. This disables\n");
+		log("        falling back to a simpler overapproximating model for those cells for\n");
+		log("        which the detailed model is expected costly.\n");
+		log("\n");
 	}
 	void execute(std::vector<std::string> args, RTLIL::Design *design) override
 	{
@@ -68,6 +74,8 @@ struct CheckPass : public Pass {
 		bool mapped = false;
 		bool allow_tbuf = false;
 		bool assert_mode = false;
+		bool force_detailed_loop_check = false;
+		bool suggest_detail = false;
 
 		size_t argidx;
 		for (argidx = 1; argidx < args.size(); argidx++) {
@@ -89,6 +97,10 @@ struct CheckPass : public Pass {
 			}
 			if (args[argidx] == "-assert") {
 				assert_mode = true;
+				continue;
+			}
+			if (args[argidx] == "-force-detailed-loop-check") {
+				force_detailed_loop_check = true;
 				continue;
 			}
 			break;
@@ -154,9 +166,10 @@ struct CheckPass : public Pass {
 			struct CircuitEdgesDatabase : AbstractCellEdgesDatabase {
 				TopoSort<std::pair<RTLIL::IdString, int>> &topo;
 				SigMap sigmap;
+				bool force_detail;
 
-				CircuitEdgesDatabase(TopoSort<std::pair<RTLIL::IdString, int>> &topo, SigMap &sigmap)
-					: topo(topo), sigmap(sigmap) {}
+				CircuitEdgesDatabase(TopoSort<std::pair<RTLIL::IdString, int>> &topo, SigMap &sigmap, bool force_detail)
+					: topo(topo), sigmap(sigmap), force_detail(force_detail) {}
 
 				void add_edge(RTLIL::Cell *cell, RTLIL::IdString from_port, int from_bit,
 							  RTLIL::IdString to_port, int to_bit, int) override {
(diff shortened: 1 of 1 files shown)


rowanG077:
I would like an option that will automagically fall back to detailed loop checking when it was found a course loop. I expect that a loop is probably isolated to a relatively tiny subset of the netlist so detailed checking should be cheap.

povik:
> I expect that a loop is mostly isolated to a relatively tiny subset of the netlist so detailed checking should be cheap.

Where the new behavior matters the most is in cases where a single cell can explode to a prohibitive number of edges (like in the Black Parrot). In such cases it's a bit misleading to suggest `-force-detailed-loop-check` as it will get stuck, so we may want to adjust the wording or remove the suggestion.

> As it stands designs which yosys now reports as fine will be wrongly flagged as having logic loops with this PR as it is.

We could adjust the threshold to make sure the change doesn't affect realistic designs which were checkable before. So that the only designs affected are those on which `check` was stuck or close-to-stuck before.

> Sorry for this drive by comment 😅. If this is infeasible or just a lot of work ignore me.

Not at all. It's not infeasible though I am a bit worried about the code complexity for a little gain.

PR #4497, opened 2024-07-18
bitpattern: comments

widlarizer:
Since `kernel/bitpattern.h` is meant to be installed in #4494, its behavior is now exposed to external project developers. This PR adds comments describing the file's functionality and intended use case

### DIFF
--- a/kernel/bitpattern.h
+++ b/kernel/bitpattern.h
@@ -25,6 +25,18 @@
 
 YOSYS_NAMESPACE_BEGIN
 
+/**
+ * This file implements BitPatternPool for efficiently storing and querying
+ * sets of fixed-width 2-valued logic constants compressed as "bit patterns".
+ * A bit pattern can have don't cares on one or more bit positions (State::Sa).
+ *
+ * In terms of logic synthesis:
+ * A BitPatternPool is a sum of products (SOP).
+ * BitPatternPool::bits_t is a cube.
+ *
+ * BitPatternPool does not permit adding new patterns, only removing.
+ * Its intended use case is in analysing cases in case/match constructs in HDL.
+ */
 struct BitPatternPool
 {
 	int width;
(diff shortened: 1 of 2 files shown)


kernel/bitpattern.h
```diff
@@ -25,6 +25,15 @@
 
 YOSYS_NAMESPACE_BEGIN
 
+/**
+ * This file implements BitPatternPool for efficiently storing
+ * sets of fixed-width 2-valued logic constants compressed as "bit patterns"
+ * represented as bits_t. A bit pattern is a signal in 3-valued logic:
+ * 0, 1, a ("any" or "don't care").
```
povik, line 32: I wouldn't say a bit pattern is a signal. How about

>  This file implements BitPatternPool for efficiently storing and querying sets of fixed-width 2-valued logic constants compressed as "bit patterns". A bit pattern can have don't-cares on one or more bit positions.

kernel/bitpattern.h
```diff
@@ -87,6 +103,12 @@ struct BitPatternPool
 		return bits;
 	}
 
+	/**
+	 * Do two patterns match?
+	 * A pattern matches another when all their bits match.
+	 * A bit matches another if their values are equal.
+	 * A bit also matches another if either of their values is "any"
```
povik, line 110: Can this be summarized as "two patterns match if their intersection is non-empty"?

kernel/bitpattern.h
```diff
@@ -106,6 +135,14 @@ struct BitPatternPool
 		return false;
 	}
 
+	/**
+	 * Does sig match a pattern in the pool such that
```
povik, line 139: Can this be summarized as

> Are all combinations represented by the given signal contained in the set?

Is it true that `pool({aaa}).has_all(01a) == false`? That doesn't match what I think this function does.

povik: As a side note, the interface is a bit awkward in the sense that all the methods accept `SigSpec` but in fact they crash if the signal isn't convertable to `Const` (except for the constructor). Ideally we would change them to take `Const` instead.

widlarizer: yep the description and example was wrong, fixed and covered with unit test

PR #4499, opened 2024-07-18
ast: don't suggest use in external projects

widlarizer:
Some external projects, like synlig, depend on `ast.h` which is valid but isn't generally necessary or advisable. The Verific frontend doesn't use it. This PR removes the suggestion that external yosys frontends other than the open source Verilog frontend can use this ast functionality

### DIFF
--- a/frontends/ast/ast.cc
+++ b/frontends/ast/ast.cc
@@ -20,10 +20,10 @@
  *
  *  This is the AST frontend library.
  *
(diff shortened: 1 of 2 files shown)


povik (approved):
Yes, we want to make it clear we are not maintaining `frontends/ast` as an external-facing API. We urge existing projects who depend on it to fork it and maintain their own version, and we don't suggest use in new projects.

PR #4502, opened 2024-07-22
Release build configuration improvements

widlarizer:
This PR sets the build optimization level to `-O3` for yosys. It doesn't change abc build flags. Clang now uses `-Og` on `ENABLE_DEBUG=1`.

## New Makefile variables

`ENABLE_LTO` - primary way of disabling LTO
`OPT_LEVEL` - allows `-O` opt level override
`GCC_LTO` - allows gcc-specific LTO flags override
`CLANG_LTO` - allows clang-specific LTO flags override

## clang LTO

For clang builds, which are the preferred release builds, it adds `lld`, and uses its thin LTO. Yosys is a small executable and stripped binary size changes are not significant (15 -> 19MB). In practice, this change brings about 5-10% improvements to the synthesis of various designs. This PR, when clean built on 24 threads with clang, increases the build time from 48 to 61 seconds and raises memory usage at build time from about 800MB to 3.3GB.

## gcc LTO

I don't understand gcc LTO, using `-flto=auto` *lowered* build time memory usage and made the resulting binary *slower*. This is why I'm not enabling LTO for gcc

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ SANITIZER =
 
 # Prefer using ENABLE_DEBUG over setting these
 OPT_LEVEL := -O3
-GCC_LTO := -flto=auto
+GCC_LTO :=
 CLANG_LTO := -flto=thin
 
 PROGRAM_PREFIX :=
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,7 @@ ENABLE_PYOSYS := 0
 ENABLE_GCOV := 0
 ENABLE_GPROF := 0
 ENABLE_DEBUG := 0
+ENABLE_LTO := 1
 ENABLE_CCACHE := 0
 # sccache is not always a drop-in replacement for ccache in practice
 ENABLE_SCCACHE := 0
@@ -52,8 +53,11 @@ SANITIZER =
 # SANITIZER = undefined
 # SANITIZER = cfi
 
-# Prefer using ENABLE_DEBUG over setting this
+# Prefer using ENABLE_DEBUG over setting these
 OPT_LEVEL := -O3
+GCC_LTO := -flto=auto
+CLANG_LTO := -flto=thin
+
 PROGRAM_PREFIX :=
 
 OS := $(shell uname -s)
@@ -208,10 +212,15 @@ ifeq ($(OS), OpenBSD)
 ABC_ARCHFLAGS += "-DABC_NO_RLIMIT"
 endif
 
+# This gets overridden later.
+LTOFLAGS := $(GCC_LTO)
+
 ifeq ($(CONFIG),clang)
 CXX = clang++
 CXXFLAGS += -std=$(CXXSTD) $(OPT_LEVEL)
+LINKFLAGS += -fuse-ld=lld
 ABCMKARGS += ARCHFLAGS="-DABC_USE_STDINT_H $(ABC_ARCHFLAGS)"
+LTOFLAGS := $(CLANG_LTO)
 
 ifneq ($(SANITIZER),)
 $(info [Clang Sanitizer] $(SANITIZER))
@@ -227,6 +236,7 @@ endif
 ifneq ($(findstring cfi,$(SANITIZER)),)
 CXXFLAGS += -flto
 LINKFLAGS += -flto
+LTOFLAGS =
 endif
 endif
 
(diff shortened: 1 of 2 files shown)


widlarizer:
There might be unmet considerations for lld on CI and platforms like WASI

whitequark:
The YoWASP build is already using full LTO because of code size concerns. This causes an extreme rise in link time: it takes 10 *minutes* to link the final binary, due to Wasm-specific code generation concerns. This LTO enablement should probably not be done by default because it takes so long and because `ccache` does nothing to alleviate it: each rebuild is going to be 10 minutes plus however long compiling the C files takes.

You have removed `ENABLE_NDEBUG`. I believe this will mean that asserts are always a part of the build, is this intentional?

Makefile
```diff
@@ -436,12 +437,8 @@ CXXFLAGS += -pg
 LINKFLAGS += -pg
 endif
 
-ifeq ($(ENABLE_NDEBUG),1)
-CXXFLAGS := -O3 -DNDEBUG $(filter-out -Os -ggdb,$(CXXFLAGS))
-endif
-
 ifeq ($(ENABLE_DEBUG),1)
-CXXFLAGS := -Og -DDEBUG $(filter-out -Os,$(CXXFLAGS))
+CXXFLAGS := -Og -DDEBUG $(filter-out $(OPT_LEVEL),$(CXXFLAGS))
```
whitequark, line 458: The whole `$(filter-out $(some_flag),$(CXXFLAGS))` thing is ridiculous and ought to be gone.

whitequark:
Is `-O3` vs `-Os` what brings the perf improvement, or is it `-flto=thin`? Having both in the same PR makes it harder to understand the cause(s) of performance changes.

widlarizer:
`-O3` brings 1-10% synthesis time reduction, thin LTO brings 0-1%. Thin LTO additionally allows us to be reasonable about keeping code out of header files and keep compilation times down

widlarizer:
> You have removed `ENABLE_NDEBUG`

Yes, this is also intentional. I'm on board with @jix idea of switching to always-on asserts vs debug asserts that can be disabled, but since some of our current asserts are "load-bearing" and shouldn't ever be disabled, we treat them all as one protected bag right now. If somebody really wanted them gone, they can still directly add `-DNDEBUG` with flag overrides, but we're removing the suggestion that that's "supported" usage by having a switch in the Makefile. Sorry for mixing things together, this is partially a response to https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2110

widlarizer:
I should also mention that with clang and lld, adding thin LTO to `-O3` actually saved 5 seconds on the clean build in my testing

whitequark:
> I'm on board with @jix idea of switching to always-on asserts vs debug asserts that can be disabled, but since some of our current asserts are "load-bearing" and shouldn't ever be disabled, we treat them all as one protected bag right now.

I agree with this treatment, just wanted to check if it was intentional since nothing in the commit log indicates it was.

KrystalDelusion:
I'm not sure what (if anything) actually uses the Dockerfile, certainly not any of the ci builds.  It looks like `aminya/setup-cpp` used in the `.github/workflows/test-compile.yml` correctly sets up clang with lld for both linux/mac, but `.github/workflows/test-build.yml` is failing the mac build because it just uses the pre-installed stuff sans lld.  Would suggest copying the CPP setup step from test-compile.

The Verific runner also seems to be missing lld, best to get @mmicko to install that.  The nix flake I don't know anything about, but @RCoeurjoly might be able to help?

mmicko:
@KrystalDelusion is right Dockerfile is not used by CI it is for user convenience.
Regarding lld, it should be fine now, I have added it to CIs docker image and restarted runners so latest will be used.

RCoeurjoly:
> I'm not sure what (if anything) actually uses the Dockerfile, certainly not any of the ci builds. It looks like `aminya/setup-cpp` used in the `.github/workflows/test-compile.yml` correctly sets up clang with lld for both linux/mac, but `.github/workflows/test-build.yml` is failing the mac build because it just uses the pre-installed stuff sans lld. Would suggest copying the CPP setup step from test-compile.
> 
> The Verific runner also seems to be missing lld, best to get @mmicko to install that. The nix flake I don't know anything about, but @RCoeurjoly might be able to help?

Pushed a commit to fix the nix build. Tested locally and it works. Some further improvements could be made, like not duplicating the dependencies in both the package and the devShell, but that can be done in another PR, I think.

Makefile
mmicko: In this case it is better to override OPT_LEVEL with -Og value then to filter-out

widlarizer:
@mmicko we're still getting `clang: error: invalid linker name in argument '-fuse-ld=lld'` on macos after a rerun

mmicko:
For macos you need altering yml file since that runs on github piblic servers. I have only fixed it for private CI.

KrystalDelusion (approved):

mmicko (approved):

widlarizer:
@whitequark I think I misunderstood your previous comment, can you chime in more concretely about what you suggest for WASI builds? You're saying we shouldn't use thin LTO by default but that you build yowasp with full LTO?

PR #4503, opened 2024-07-23
Guess VHDL frontend for both *.vhd and *vhdl files

RCoeurjoly:
*.vhdl files, and not only *.vhd files, seem to be valid VHDL extensions.
For example, GitHub assigns VHDL to both .vhd and .vhdl, when searching for VHDL files in GHDL:
https://github.com/search?q=repo%3Aghdl%2Fghdl++language%3AVHDL&type=code

Tested locally the following:

```
cp tests/sva/basic04.vhd tests/sva/basic04.vhdl
./yosys tests/sva/basic04.vhdl
```

We get:

> -- Parsing `tests/sva/basic04.vhdl' using frontend ` -vhdl' --

Checking that we did not brake *.vhd

`./yosys tests/sva/basic04.vhd`

We also get:

> -- Parsing `tests/sva/basic04.vhd' using frontend ` -vhdl' --

### DIFF
--- a/kernel/yosys.cc
+++ b/kernel/yosys.cc
@@ -1133,7 +1133,9 @@ bool run_frontend(std::string filename, std::string command, RTLIL::Design *desi
 			command = " -vlog2k";
 		else if (filename_trim.size() > 2 && filename_trim.compare(filename_trim.size()-3, std::string::npos, ".sv") == 0)
 			command = " -sv";
-		else if (filename_trim.size() > 3 && filename_trim.compare(filename_trim.size()-4, std::string::npos, ".vhd") == 0)
+		else if ((filename_trim.size() > 3 && filename_trim.compare(filename_trim.size()-4, std::string::npos, ".vhd") == 0)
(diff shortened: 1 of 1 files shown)


whitequark (approved):
Looks good, but also can we do something about that awful conditional?

RCoeurjoly:
> Looks good, but also can we do something about that awful conditional?

You mean the whole command == "auto" block, right? If that's the case, what do you think of last commit?

Also tested locally with vhd, vhdl, sv, il, v.

whitequark:
Vastly better, thanks.

[We're only up to C++17, right?](https://en.cppreference.com/w/cpp/string/basic_string/ends_with)

RCoeurjoly:
> Vastly better, thanks.
> 
> [We're only up to C++17, right?](https://en.cppreference.com/w/cpp/string/basic_string/ends_with)

Yes: https://github.com/YosysHQ/yosys/pull/4460

PR #4507, opened 2024-07-25
Do not map `$eqx` `$nex` `$bweqx`

povik:

If someone ingests Verilog code using the `===` operator, and tries to synthesize it, I would like for that to be caught by `check -mapped` rather than silently processed as an `==` operator.

We make `simplemap` and `techmap` ignore `$eqx` `$nex` `$bweqx`, instead those cells are left in the netlist unmapped.

Along the way I make `simplemap` generate its help message based on the actual mappers registered (I think notwithstanding the `$eqx` change that help message grew out of sync.)


whitequark:
> If someone ingests Verilog code using the `===` operator, and tries to synthesize it, I would like for that to be caught by `check -mapped` rather than silently processed as an `==` operator.

Wouldn't this break existing code?

PR #4510, opened 2024-07-26
Update interactive_investigation.rst

JamesTimothyMeech:
The text starting at line 118 refers to proc twice but it should refer to opt and then to proc.


povik:
Thanks!

PR #4512, opened 2024-07-26
Increase maximum hashtable size to support larger designs.

mikesinouye:
Increase the maximum supported Yosys design size. My design ran into this limitation during `synth`.

Increase the maximum hashtable size to a value which is still below the int32 limit.


povik:
This happens to overlap with #4471

widlarizer:
Yep closing as duplicate of #4471 which we've just merged

PR #4515, opened 2024-07-30
Run nix build also on macos. Build with more logs

RCoeurjoly:
This is achieved by having a matrix build, where we build with nix under both ubuntu-latest and macos-latest.

We could also test that yosys builds locally with nix on macOS. Unfortunately I don't have the machine required to do that, so any help is 

Tested the action on my fork and it works (https://github.com/RCoeurjoly/yosys/actions/runs/10169394792).


widlarizer (approved):
This small change clearly only affects the nix flake build and does what it's supposed to

PR #4516, opened 2024-07-31
Represent string constants as strings

widlarizer:
Currently, a `RTLIL::Const` representing a string will use `std::vector<RTLIL::State>`, encoding the string into a vector of individual bits. This increases the memory consumption of string attributes, such as `src` and `hdlname`, by a factor of 8.

As of this PR, the structure now holds data represented with ~~an `std::variant<bitvectype, std::string>` named `backing`~~ a union. The `bits` vector is no longer accessible and instead `std::vector<RTLIL::State>& bits()` has to be used. When a Const is constructed from a string, it is represented with the string alternative. When its `bits()` method is used, the string is encoded into the bit vector alternative as prior to this PR, see `Const::bitvectorize()`. This is similar to how `RTLIL::SigSpec` "unpacks" into bits.

In the initial implementation, when `bits()` were being accessed, I asserted that the Const isn't represented with a string. This found some places in Yosys that access string attributes bit-by-bit. As a result, this PR comes with bonus content:

+ `kernel/cellaigs.cc`
+ `frontends/ast/ast.cc`
+ `Const::pretty_fmt`
+ cell checker `is_param` that doesn't try return bits

It's possible to split these now necessary changes out to a separate PR, but it will be heavy on git conflicts. Please advise.

This PR does not affect constant `SigChunk`s which also have an `std::vector<RTLIL::State>` each.

- [X] clean my desk (TODO, log_debug, did_something_hook, unwrap `std::get` from `assert_get_bits` etc)
- [x] ~~add some way of debugging how much unpacking we're doing~~ requires unrelated changes in followup PR

Please comment what memory usage changes you see with this PR. I've observed memory savings of 0-20% depending on the design with negligible runtime improvements. The open source Verilog frontend often dominates memory usage, so flows that use Verific may more readily show these improvements.

### DIFF
--- a/backends/cxxrtl/cxxrtl_backend.cc
+++ b/backends/cxxrtl/cxxrtl_backend.cc
@@ -328,7 +328,7 @@ struct FlowGraph {
 					node_comb_defs[node].insert(chunk.wire);
 				}
 			}
-		for (auto bit : sig.bits())
+		for (auto bit : sig)
 			bit_has_state[bit] |= is_ff;
 		// Only comb defs of an entire wire in the right order can be inlined.
 		if (!is_ff && sig.is_wire()) {
@@ -864,7 +864,7 @@ struct CxxrtlWorker {
 		if (!module->has_attribute(ID(cxxrtl_template)))
 			return {};
 
-		if (module->attributes.at(ID(cxxrtl_template)).flags != RTLIL::CONST_FLAG_STRING)
+		if (!(module->attributes.at(ID(cxxrtl_template)).flags & RTLIL::CONST_FLAG_STRING))
 			log_cmd_error("Attribute `cxxrtl_template' of module `%s' is not a string.\n", log_id(module));
 
 		std::vector<std::string> param_names = split_by(module->get_string_attribute(ID(cxxrtl_template)), " \t");
@@ -1665,15 +1665,15 @@ struct CxxrtlWorker {
 							switch (bit) {
 								case RTLIL::S0:
 								case RTLIL::S1:
-									compare_mask.bits.push_back(RTLIL::S1);
-									compare_value.bits.push_back(bit);
+									compare_mask.bits().push_back(RTLIL::S1);
+									compare_value.bits().push_back(bit);
 									break;
 
 								case RTLIL::Sx:
 								case RTLIL::Sz:
 								case RTLIL::Sa:
-									compare_mask.bits.push_back(RTLIL::S0);
-									compare_value.bits.push_back(RTLIL::S0);
+									compare_mask.bits().push_back(RTLIL::S0);
+									compare_value.bits().push_back(RTLIL::S0);
 									break;
 
 								default:
@@ -3028,7 +3028,7 @@ struct CxxrtlWorker {
 								if (init == RTLIL::Const()) {
 									init = RTLIL::Const(State::Sx, GetSize(bit.wire));
 								}
-								init[bit.offset] = port.init_value[i];
+								init.bits()[bit.offset] = port.init_value[i];
 							}
 						}
 					}
--- a/frontends/ast/ast.cc
+++ b/frontends/ast/ast.cc
@@ -951,15 +951,7 @@ RTLIL::Const AstNode::asAttrConst() const
 {
 	log_assert(type == AST_CONSTANT);
 
-	RTLIL::Const val;
-	val.bits = bits;
-
-	if (is_string) {
-		val.flags |= RTLIL::CONST_FLAG_STRING;
-		log_assert(val.decode_string() == str);
-	}
-
-	return val;
+	return is_string ? RTLIL::Const(str) : RTLIL::Const(bits);
 }
 
 RTLIL::Const AstNode::asParaConst() const
@@ -1005,7 +997,7 @@ uint64_t AstNode::asInt(bool is_signed)
 		uint64_t ret = 0;
 
 		for (int i = 0; i < 64; i++)
-			if (v.bits.at(i) == RTLIL::State::S1)
+			if (v.at(i) == RTLIL::State::S1)
 				ret |= uint64_t(1) << i;
 
 		return ret;
@@ -1023,15 +1015,15 @@ double AstNode::asReal(bool is_signed)
 	{
 		RTLIL::Const val(bits);
 
-		bool is_negative = is_signed && !val.bits.empty() && val.bits.back() == RTLIL::State::S1;
+		bool is_negative = is_signed && !val.empty() && val.back() == RTLIL::State::S1;
 		if (is_negative)
-			val = const_neg(val, val, false, false, val.bits.size());
+			val = const_neg(val, val, false, false, val.size());
 
 		double v = 0;
-		for (size_t i = 0; i < val.bits.size(); i++)
+		for (size_t i = 0; i < val.size(); i++)
 			// IEEE Std 1800-2012 Par 6.12.2: Individual bits that are x or z in
 			// the net or the variable shall be treated as zero upon conversion.
-			if (val.bits.at(i) == RTLIL::State::S1)
+			if (val.at(i) == RTLIL::State::S1)
 				v += exp2(i);
 		if (is_negative)
 			v *= -1;
@@ -1054,15 +1046,15 @@ RTLIL::Const AstNode::realAsConst(int width)
 #else
 	if (!std::isfinite(v)) {
 #endif
-		result.bits = std::vector<RTLIL::State>(width, RTLIL::State::Sx);
+		result = std::vector<RTLIL::State>(width, RTLIL::State::Sx);
 	} else {
 		bool is_negative = v < 0;
 		if (is_negative)
 			v *= -1;
 		for (int i = 0; i < width; i++, v /= 2)
-			result.bits.push_back((fmod(floor(v), 2) != 0) ? RTLIL::State::S1 : RTLIL::State::S0);
+			result.bits().push_back((fmod(floor(v), 2) != 0) ? RTLIL::State::S1 : RTLIL::State::S0);
 		if (is_negative)
-			result = const_neg(result, result, false, false, result.bits.size());
+			result = const_neg(result, result, false, false, result.size());
 	}
 	return result;
 }
@@ -1767,16 +1759,7 @@ static std::string serialize_param_value(const RTLIL::Const &val) {
 		res.push_back('r');
 	res += stringf("%d", GetSize(val));
 	res.push_back('\'');
-	for (int i = GetSize(val) - 1; i >= 0; i--) {
-		switch (val.bits[i]) {
-			case RTLIL::State::S0: res.push_back('0'); break;
-			case RTLIL::State::S1: res.push_back('1'); break;
-			case RTLIL::State::Sx: res.push_back('x'); break;
-			case RTLIL::State::Sz: res.push_back('z'); break;
-			case RTLIL::State::Sa: res.push_back('?'); break;
-			case RTLIL::State::Sm: res.push_back('m'); break;
-		}
-	}
+	res.append(val.as_string("?"));
 	return res;
 }
 
@@ -1868,7 +1851,7 @@ std::string AstModule::derive_common(RTLIL::Design *design, const dict<RTLIL::Id
 		} else if ((it->second.flags & RTLIL::CONST_FLAG_STRING) != 0)
 			child->children[0] = AstNode::mkconst_str(it->second.decode_string());
 		else
-			child->children[0] = AstNode::mkconst_bits(it->second.bits, (it->second.flags & RTLIL::CONST_FLAG_SIGNED) != 0);
+			child->children[0] = AstNode::mkconst_bits(it->second.to_bits(), (it->second.flags & RTLIL::CONST_FLAG_SIGNED) != 0);
 		rewritten.insert(it->first);
 	}
 
@@ -1881,7 +1864,7 @@ std::string AstModule::derive_common(RTLIL::Design *design, const dict<RTLIL::Id
 			if ((param.second.flags & RTLIL::CONST_FLAG_STRING) != 0)
 				defparam->children.push_back(AstNode::mkconst_str(param.second.decode_string()));
 			else
-				defparam->children.push_back(AstNode::mkconst_bits(param.second.bits, (param.second.flags & RTLIL::CONST_FLAG_SIGNED) != 0));
+				defparam->children.push_back(AstNode::mkconst_bits(param.second.to_bits(), (param.second.flags & RTLIL::CONST_FLAG_SIGNED) != 0));
 			new_ast->children.push_back(defparam);
 		}
 
--- a/frontends/verific/verific.cc
+++ b/frontends/verific/verific.cc
@@ -236,23 +236,6 @@ RTLIL::IdString VerificImporter::new_verific_id(Verific::DesignObj *obj)
 	return s;
 }
 
-RTLIL::Const mkconst_str(const std::string &str)
-{
-	RTLIL::Const val;
-	std::vector<RTLIL::State> data;
-	data.reserve(str.size() * 8);
-	for (size_t i = 0; i < str.size(); i++) {
-		unsigned char ch = str[str.size() - i - 1];
-		for (int j = 0; j < 8; j++) {
-			data.push_back((ch & 1) ? State::S1 : State::S0);
-			ch = ch >> 1;
-		}
-	}
-	val.bits = data;
-	val.flags |= RTLIL::CONST_FLAG_STRING;
-	return val;
-}
-
 static const RTLIL::Const extract_vhdl_boolean(std::string &val)
 {
 	if (val == "false")
@@ -295,7 +278,7 @@ static const RTLIL::Const extract_vhdl_char(std::string &val)
 
 static const RTLIL::Const extract_real_value(std::string &val)
 {
-	RTLIL::Const c = mkconst_str(val);
+	RTLIL::Const c(val);
 	c.flags |= RTLIL::CONST_FLAG_REAL;
 	return c;
 }
@@ -333,7 +316,7 @@ static const  RTLIL::Const extract_vhdl_const(const char *value, bool output_sig
 	} else if (val == "true") {
 		c = RTLIL::Const::from_string("1");
 	} else {
-		c = mkconst_str(val);
+		c = RTLIL::Const(val);
 		log_warning("encoding value '%s' as string.\n", value);
 	}
 	if (is_signed)
@@ -364,7 +347,7 @@ static const  RTLIL::Const extract_verilog_const(const char *value, bool allow_s
 	} else if (allow_string) {
 		c = RTLIL::Const(val);
 	} else {
-		c = mkconst_str(val);
+		c = RTLIL::Const(val);
 		log_warning("encoding value '%s' as string.\n", value);
 	}
 	if (is_signed)
@@ -1634,7 +1617,7 @@ void VerificImporter::import_netlist(RTLIL::Design *design, Netlist *nl, std::ma
 						if (*ascii_initdata == 0)
 							break;
 						if (*ascii_initdata == '0' || *ascii_initdata == '1') {
-							initval[bit_idx] = (*ascii_initdata == '0') ? State::S0 : State::S1;
+							initval.bits()[bit_idx] = (*ascii_initdata == '0') ? State::S0 : State::S1;
 							initval_valid = true;
 						}
 						ascii_initdata++;
@@ -1756,9 +1739,9 @@ void VerificImporter::import_netlist(RTLIL::Design *design, Netlist *nl, std::ma
 
 					if (init_nets.count(net)) {
 						if (init_nets.at(net) == '0')
-							initval.bits.at(bitidx) = State::S0;
+							initval.bits().at(bitidx) = State::S0;
 						if (init_nets.at(net) == '1')
-							initval.bits.at(bitidx) = State::S1;
+							initval.bits().at(bitidx) = State::S1;
 						initval_valid = true;
 						init_nets.erase(net);
 					}
@@ -1832,12 +1815,12 @@ void VerificImporter::import_netlist(RTLIL::Design *design, Netlist *nl, std::ma
 			initval = bit.wire->attributes.at(ID::init);
 
 		while (GetSize(initval) < GetSize(bit.wire))
-			initval.bits.push_back(State::Sx);
+			initval.bits().push_back(State::Sx);
 
 		if (it.second == '0')
-			initval.bits.at(bit.offset) = State::S0;
+			initval.bits().at(bit.offset) = State::S0;
 		if (it.second == '1')
-			initval.bits.at(bit.offset) = State::S1;
+			initval.bits().at(bit.offset) = State::S1;
 
 		bit.wire->attributes[ID::init] = initval;
 	}
@@ -2024,7 +2007,7 @@ void VerificImporter::import_netlist(RTLIL::Design *design, Netlist *nl, std::ma
 			}
 
 			Const qx_init = Const(State::S1, width);
-			qx_init.bits.resize(2 * width, State::S0);
+			qx_init.bits().resize(2 * width, State::S0);
 
 			clocking.addDff(new_verific_id(inst), sig_dx, sig_qx, qx_init);
 			module->addXnor(new_verific_id(inst), sig_dx, sig_qx, sig_ox);
@@ -2295,7 +2278,7 @@ void VerificImporter::import_netlist(RTLIL::Design *design, Netlist *nl, std::ma
 					continue;
 
 				if (non_ff_bits.count(SigBit(wire, i)))
-					initval[i] = State::Sx;
+					initval.bits()[i] = State::Sx;
 			}
 
 			if (wire->port_input) {
@@ -2482,7 +2465,7 @@ Cell *VerificClocking::addDff(IdString name, SigSpec sig_d, SigSpec sig_q, Const
 				if (c.wire && c.wire->attributes.count(ID::init)) {
 					Const val = c.wire->attributes.at(ID::init);
 					for (int i = 0; i < GetSize(c); i++)
-						initval[offset+i] = val[c.offset+i];
+						initval.bits()[offset+i] = val[c.offset+i];
 				}
 				offset += GetSize(c);
 			}
@@ -2553,7 +2536,7 @@ Cell *VerificClocking::addAldff(IdString name, RTLIL::SigSpec sig_aload, RTLIL::
 			if (c.wire && c.wire->attributes.count(ID::init)) {
 				Const val = c.wire->attributes.at(ID::init);
 				for (int i = 0; i < GetSize(c); i++)
-					initval[offset+i] = val[c.offset+i];
+					initval.bits()[offset+i] = val[c.offset+i];
 			}
 			offset += GetSize(c);
 		}
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -202,25 +202,34 @@ const pool<IdString> &RTLIL::builtin_ff_cell_types() {
 	return res;
 }
 
+#define check(condition) log_assert(condition && "malformed Const union")
+
+Const::bitvectype& Const::get_bits() const {
+	check(is_bits());
+	return *get_if_bits();
+}
+
+std::string& Const::get_str() const {
+	check(is_str());
+	return *get_if_str();
+}
+
 RTLIL::Const::Const(const std::string &str)
 {
 	flags = RTLIL::CONST_FLAG_STRING;
-	bits.reserve(str.size() * 8);
-	for (int i = str.size()-1; i >= 0; i--) {
-		unsigned char ch = str[i];
-		for (int j = 0; j < 8; j++) {
-			bits.push_back((ch & 1) != 0 ? State::S1 : State::S0);
-			ch = ch >> 1;
-		}
-	}
+	new ((void*)&str_) std::string(str);
+	tag = backing_tag::string;
 }
 
 RTLIL::Const::Const(long long val, int width)
 {
 	flags = RTLIL::CONST_FLAG_NONE;
-	bits.reserve(width);
+	new ((void*)&bits_) bitvectype();
+	tag = backing_tag::bits;
+	bitvectype& bv = get_bits();
+	bv.reserve(width);
 	for (int i = 0; i < width; i++) {
-		bits.push_back((val & 1) != 0 ? State::S1 : State::S0);
+		bv.push_back((val & 1) != 0 ? State::S1 : State::S0);
 		val = val >> 1;
 	}
 }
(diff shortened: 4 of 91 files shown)


kernel/rtlil.cc
```diff
@@ -203,93 +203,171 @@ const pool<IdString> &RTLIL::builtin_ff_cell_types() {
...
-	if (is_signed && bits.back() == State::S1)
-		for (size_t i = bits.size(); i < 32; i++)
+	if (is_signed && bv.back() == State::S1)
+		for (size_t i = bv.size(); i < 32; i++)
 			ret |= 1 << i;
 	return ret;
 }
 
-std::string RTLIL::Const::as_string() const
+std::string RTLIL::Const::as_string(std::string any) const
```
povik, line 358: What's up with this `any` parameter? This looks unrelated to the backing change

kernel/rtlil.cc
```diff
@@ -344,9 +429,11 @@ std::string RTLIL::Const::decode_string() const
 
 bool RTLIL::Const::is_fully_zero() const
 {
+	bitvectorize();
+	bitvectype& bv = assert_get_bits("Const::decode_string");
```
povik, line 433: Why not just call `bits()` instead of repeating `bitvectorize(); bitvectype& bv = assert_get_bits(...)` everywhere?

kernel/rtlil.h
```diff
@@ -673,19 +687,68 @@ struct RTLIL::Const
...
+		bitvectorize();
+		return assert_get_bits("Const::operator[]").at(index);
+	}
+	inline const RTLIL::State &operator[](int index) const {
+		bitvectorize();
+		return assert_get_bits("const Const::operator[]").at(index);
+	}
+	inline bitvectype::iterator begin() {
+		bitvectorize();
+		return assert_get_bits("Const bit iterator begin()").begin();
```
povik, line 746: I think this is not constant, you can make changes through this iterator

kernel/rtlil.h
```diff
@@ -695,25 +758,37 @@ struct RTLIL::Const
...
 	}
 
 	inline unsigned int hash() const {
 		unsigned int h = mkhash_init;
-		for (auto b : bits)
-			h = mkhash(h, b);
+
+		if(auto str = std::get_if<std::string>(&backing)) {
+			for (auto c : *str)
+				h = mkhash(h, c);
```
povik, line 786: If you have two Consts, one uses the string backing but the other uses the bit vector backing, this doesn't seem to guarantee for their hashes to be the same when their bit content is the same

kernel/rtlil.cc
```diff
@@ -203,93 +203,171 @@ const pool<IdString> &RTLIL::builtin_ff_cell_types() {
...
+		unsigned char ch = str[i];
+		for (int j = 0; j < 8; j++) {
+			b.push_back((ch & 1) != 0 ? State::S1 : State::S0);
+			ch = ch >> 1;
+		}
+	}
+	return b;
+}
+
+std::string RTLIL::Const::pretty_fmt() const {
```
povik, line 320: What are these new functions for?

kernel/rtlil.cc
```diff
@@ -203,93 +203,171 @@ const pool<IdString> &RTLIL::builtin_ff_cell_types() {
...
+			return bv[i] < other_bv[i];
 	return false;
 }
 
 bool RTLIL::Const::operator ==(const RTLIL::Const &other) const
 {
-	return bits == other.bits;
+	const char* ctx = "operator==";
+	if (std::get_if<std::string>(&backing) != std::get_if<std::string>(&other.backing))
+		return decode_string() == other.decode_string();
```
povik, line 276: `decode_string()` can be equal for two Const even if their content isn't. Consider `Const(1, 1)` vs `Const(1, 8)`

kernel/rtlil.h
widlarizer: Yeah, I meant to refer to `RTLIL::Const`, not the C++ `const` keyword

kernel/rtlil.cc
```diff
@@ -203,93 +203,171 @@ const pool<IdString> &RTLIL::builtin_ff_cell_types() {
...
+		if (bv[i] != other_bv[i])
+			return bv[i] < other_bv[i];
 	return false;
 }
 
 bool RTLIL::Const::operator ==(const RTLIL::Const &other) const
 {
-	return bits == other.bits;
+	const char* ctx = "operator==";
+	if (std::get_if<std::string>(&backing) != std::get_if<std::string>(&other.backing))
```
povik, line 275: This is a confusing conditional. The only way the two `get_if` can return the same pointer is if they both return null, or if the Const is being compared against itself, right?

kernel/rtlil.h
povik: OK, then we may drop Const from there as it's superfluous

kernel/rtlil.cc
povik: Ah you have a TODO to get rid of this

backends/cxxrtl/cxxrtl_backend.cc
```diff
@@ -864,7 +864,7 @@ struct CxxrtlWorker {
 		if (!module->has_attribute(ID(cxxrtl_template)))
 			return {};
 
-		if (module->attributes.at(ID(cxxrtl_template)).flags != RTLIL::CONST_FLAG_STRING)
```
povik, line 867: Here we are giving the flag a semantical significance, I think we may want to remove this check

whitequark: I'm going to reject this change.

whitequark: As far as I'm concerned the `STRING` flag has semantic significance and CXXRTL treats it as such in many places (e.g. the types are going to be different and this will change the ABI). I don't really care if RTLIL claims there's no significance to it, fix RTLIL then.

povik: With this PR we are removing the semantical significance of `STRING`, instead it serves to signify the backing used, which can change merely by using the wrong accessor on the Const object. We may not go through with it, but that's the plan. As an alternative we can add an extra flag which would signify the backing.

FWIW the `==` and `!=` operators on Const disregard the flags as of today, so we have an inconsistence either way.

kernel/rtlil.cc
```diff
@@ -1034,6 +1135,14 @@ namespace {
 					cell->name.c_str(), cell->type.c_str(), __FILE__, linenr, buf.str().c_str());
 		}
 
+		void is_param(const RTLIL::IdString& name)
```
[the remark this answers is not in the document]
widlarizer: It's related, having a version of this function that doesn't have a return value prevents unpacking every single string attribute to bits. Without this, the PR wouldn't have any positive impact because once you ran check, all Const would unpack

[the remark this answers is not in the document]
povik: That's for a parameter not an attribute though

backends/cxxrtl/cxxrtl_backend.cc
whitequark: CXXRTL relies on the semantic significance so I'll not accept this PR until there is a way forward for it.

povik: > With this PR we are removing the semantical significance of STRING, instead it serves to signify the backing used, which can change merely by using the wrong accessor on the Const object.

Looking at the code change that may not be true (but it's what I thought we were doing). So the backing seems decided by the state of the `std::variant<>` field, in which case I need to take back what I said about the `STRING` flag.

We still have a somewhat inconsistent state with the comparison operators ignoring the flags, but that may be exposed through to the user in some places and I am not sure we can change it.

whitequark: Note also that this conditional isn't the only (nor the most important) place where this is done; the attribute set dumping code is the primary one. In particular the results of this transformation are used for ChipFlow tools and we rely on it being done correctly.

whitequark: > We still have a somewhat inconsistent state with the comparison operators ignoring the flags, but that may be exposed through to the user in some places and I am not sure we can change it.

I guess this happens because in Verilog the string compares equal to the corresponding constant, and since Yosys is mainly used with Verilog nobody objected to this. I think that a good way to fix this issue woudl be to emit a warning (while comparing unequal) so that any such desired reliance can be found and retained with an explicit conversion.

kernel/rtlil.cc
widlarizer: Yes, it's unrelated refactoring to adjacent code. I'm giving as_string a parameter that sets the representation for don't-care / any bits to deduplicate code that previously used the `bits()` interface on some strings. I think currently it only deduplicates it with `serialize_param_value` in `ast.cc` but there's other places where you could use this method now

povik: Ideally this would be a separate commit then. Also this would be easier to review if the `s/.bits/.bits()` change was a commit of its own, done upfront before we start making changes to the backing within `Const`.

kernel/rtlil.cc
widlarizer: To move code in from elsewhere for deduplication and code reuse. This one is used in `Aig::Aig(Cell *cell)` now

kernel/rtlil.cc
widlarizer: Good catch, forgot to cast to bool

kernel/rtlil.cc
widlarizer: Right, previously, it was `name += stringf(":%d", p.second.as_int());` which would unpack string constants. We want to avoid that in all scenarios. Unpacking to bits is a safety measure, only because I can't guarantee it can't happen in some use case

povik: > which would unpack string constants. We want to avoid that in all scenarios.

Not here, these are the `WIDTH` parameters which are unlikely to have string backing in the first place

povik: We don't care about backing for parameters at all, at least not yet, they are almost never strings (the only exception I can think of is `FLAVOR` on `$check`; and parameters on user modules of course)

backends/cxxrtl/cxxrtl_backend.cc
widlarizer: Note: I wrote that not having seen the prior comments (github pls). I'll respond to it more properly soon when I have enough time to understand the situation some more

kernel/rtlil.cc
widlarizer: Ah true, attributes are not covered by this. WIDTH parameters specifically still use `param` rather than `is_param` because their value is used. I still don't like enforcing non-bit-ness when we're only checking for param presence

povik: > WIDTH parameters specifically still use `param` rather than `is_param` because their value is used.

I was commenting on the cellaigs code which does `name += stringf(":%d", p.second.as_int());` which touches on the `WIDTH` params (`A_WIDTH` or `B_WIDTH`).

> I still don't like enforcing bitvec backing when we're only checking for param presence

Regarding the `is_param` vs `param` situation -- there aren't any gains to expect there, on the other hand any change we can shed off from this PR will make the review easier, which is important given the wide-reaching nature of it, for that reason I think it's better we drop that change.

backends/cxxrtl/cxxrtl_backend.cc
povik: > I'll respond to it more properly soon when I have enough time to understand the situation some more

The key takeaway is this: not to upset users we need the choice of backing for Const to be independent of whether the `STRING` flag is set, which I think is already the case with this PR.

povik:
> Please comment what memory usage changes you see with this PR.

On the sky130hd/jpeg design from OpenROAD-flow-scripts, dumping the netlist just after the `techmap` command, I see the following difference in memory usage when loading in the RTLIL dump in a fresh Yosys process:
| | with `src` attributes | `src` stripped |
|---|---|---|
| baseline | 291 MB | 180 MB |
| patched | 221 MB | 172 MB |

kernel/rtlil.cc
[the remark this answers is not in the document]
widlarizer: Ok, removed

kernel/rtlil.h
widlarizer: This wrapper is removed and we lean on exceptions in the unexpected case of variant access problems instead

povik:
I have tried the following patch to hashlib on top of the PR, and updated the table:

```
diff --git a/kernel/hashlib.h b/kernel/hashlib.h
index 6b880f3a6..0e839e73e 100644
--- a/kernel/hashlib.h
+++ b/kernel/hashlib.h
@@ -190,7 +190,7 @@ inline int hashtable_size(int min_size)
 {
        // Primes as generated by https://oeis.org/A175953
        static std::vector<int> zero_and_some_primes = {
-               0, 23, 29, 37, 47, 59, 79, 101, 127, 163, 211, 269, 337, 431, 541, 677,
+               0, 3, 23, 29, 37, 47, 59, 79, 101, 127, 163, 211, 269, 337, 431, 541, 677,
                853, 1069, 1361, 1709, 2137, 2677, 3347, 4201, 5261, 6577, 8231, 10289,
                12889, 16127, 20161, 25219, 31531, 39419, 49277, 61603, 77017, 96281,
                120371, 150473, 188107, 235159, 293957, 367453, 459317, 574157, 717697,
```

widlarizer:
@povik the prime list change adds runtime improvements up to 3% on top of this with a 0.5% regression in one case

backends/cxxrtl/cxxrtl_backend.cc
widlarizer: Yep

frontends/verific/verific.cc
```diff
@@ -238,19 +238,7 @@ RTLIL::IdString VerificImporter::new_verific_id(Verific::DesignObj *obj)
 
 RTLIL::Const mkconst_str(const std::string &str)
```
povik, line 239: We should remove this helper altogether

kernel/rtlil.cc
povik: Still applies

kernel/rtlil.cc
```diff
@@ -203,93 +203,205 @@ const pool<IdString> &RTLIL::builtin_ff_cell_types() {
...
-	if (bits.size() != other.bits.size())
-		return bits.size() < other.bits.size();
-	for (size_t i = 0; i < bits.size(); i++)
-		if (bits[i] != other.bits[i])
-			return bits[i] < other.bits[i];
+	if (is_str() && other.is_str())
+		return str_ < other.str_;
+
+	// If either operand isn't a string, we have to construct a copy
+	// to bitvectorize without unpacking either operand
```
povik, line 308: This would be solved if we had a constant bit iterator which didn't unpack

kernel/rtlil.h
```diff
@@ -22,6 +22,8 @@
 
 #include "kernel/yosys_common.h"
 #include "kernel/yosys.h"
+#include <variant>
```
povik, line 25: Include no longer needed

kernel/rtlil.h
```diff
@@ -657,38 +667,111 @@ namespace RTLIL
...
+	// Do not access the union or tag even in Const methods unless necessary
+	mutable backing_tag tag;
+	union {
+		mutable bitvectype bits_;
+		mutable std::string str_;
+	};
+
+	// Use these utilities instead
+	bool is_bits() const { return tag == backing_tag::bits; }
+	bool is_str() const { return tag == backing_tag::string; }
```
povik, line 683: Mark those private, possibly rename those, otherwise someone at one point will take those to be semantically significant (e.g. mistake them for a `flags & CONST_FLAG_STRING != 0` check)

I suggest `bits_backed()` and `str_backed()`

kernel/rtlil.h
```diff
@@ -657,38 +667,111 @@ namespace RTLIL
...
 
-	Const() : flags(RTLIL::CONST_FLAG_NONE) {}
+	bitvectype* get_if_bits() const { return is_bits() ? &bits_ : NULL; }
+	std::string* get_if_str() const { return is_str() ? &str_ : NULL; }
+
+	void check(bool condition) const {
+		// TODO debug assert or throw
+		// log_assert(condition && "malformed Const union");
+	}
```
povik, line 691: I am not convinced we want this helper. This is worse than `log_assert` at the call site unless we start inspecting backtrace here

kernel/rtlil.h
```diff
@@ -657,38 +667,111 @@ namespace RTLIL
...
 	bool operator <(const RTLIL::Const &other) const;
 	bool operator ==(const RTLIL::Const &other) const;
 	bool operator !=(const RTLIL::Const &other) const;
 
+	const std::vector<RTLIL::State>& bits() const;
+	std::vector<RTLIL::State>& bits();
 	bool as_bool() const;
 	int as_int(bool is_signed = false) const;
-	std::string as_string() const;
+	std::string as_string(std::string any = "-") const;
```
povik, line 715: Better served with a `const char *` argument I think

passes/cmds/setundef.cc
```diff
@@ -243,6 +243,8 @@ struct SetundefPass : public Pass {
 			{
 				for (auto *cell : module->selected_cells()) {
 					for (auto &parameter : cell->parameters) {
+						if (parameter.second.flags & RTLIL::CONST_FLAG_STRING)
+							continue;
```
povik, line 247: I know `parameter.second.flags & RTLIL::CONST_FLAG_STRING` implies there shouldn't be any undef bits, but this makes the code harder to read for little gain

kernel/rtlil.h
```diff
@@ -657,38 +667,111 @@ namespace RTLIL
...
+	short flags;
+	private:
+	typedef std::vector<RTLIL::State> bitvectype;
+	enum class backing_tag: bool { bits, string };
+	// Do not access the union or tag even in Const methods unless necessary
+	mutable backing_tag tag;
+	union {
+		mutable bitvectype bits_;
+		mutable std::string str_;
+	};
```
povik, line 679: What was the issue with `std::variant` again?

kernel/rtlil.cc
```diff
@@ -365,6 +365,20 @@ std::vector<RTLIL::State> RTLIL::Const::to_bits() const
 	return b;
 }
 
+std::string RTLIL::Const::pretty_fmt() const {
+	if (get_if_str())
```
povik, line 369: This leaks the backing

kernel/rtlil.cc
```diff
@@ -365,6 +365,20 @@ std::vector<RTLIL::State> RTLIL::Const::to_bits() const
...
+		return decode_string();
+	else
+		return std::to_string(as_int());
+}
+
+std::string RTLIL::Const::pretty_fmt_undef() const {
+	if (get_if_str())
```
povik, line 376: Here too

passes/opt/opt_merge.cc
```diff
@@ -140,8 +140,13 @@ struct OptMergeWorker
 			hash_conn_strings.push_back(s + "\n");
 		}
 
-		for (auto &it : cell->parameters)
-			hash_conn_strings.push_back("P " + it.first.str() + "=" + it.second.as_string() + "\n");
```
povik, line 144: I don't understand why we want this change. The original is more readable

kernel/rtlil.h
```diff
@@ -715,6 +715,9 @@ struct RTLIL::Const
 	std::string as_string(std::string any = "-") const;
 	static Const from_string(const std::string &str);
 	std::vector<RTLIL::State> to_bits() const;
+	std::string pretty_fmt() const;
+	std::string pretty_fmt_undef() const;
```
povik, line 719: These helpers both have a single call site, it's not obvious what they do, I suggest we remove them

kernel/rtlil.cc
povik: Overridden in a subsequent commit

kernel/rtlil.h
widlarizer: They are private

povik: Ah the indentation threw me off, we should fix it

kernel/rtlil.h
widlarizer: How about this hack:

```diff
+#define check(condition) log_assert(condition && "malformed Const union")
 struct RTLIL::Const
 {
        short flags;
@@ -685,11 +686,6 @@ struct RTLIL::Const
        bitvectype* get_if_bits() const { return is_bits() ? &bits_ : NULL; }
        std::string* get_if_str() const { return is_str() ? &str_ : NULL; }

-       void check(bool condition) const {
-               // TODO debug assert or throw
-               // log_assert(condition && "malformed Const union");
-       }
-
        bitvectype& get_bits() const { check(is_bits()); return *get_if_bits(); }
        std::string& get_str() const { check(is_str()); return *get_if_str(); }
        public:
@@ -816,6 +812,7 @@ struct RTLIL::Const
                return h;
        }
 };
+#undef check
```

Thoughts? I previously verified with benchmarks that keeping the assert is free. Those asserts will make everything so verbose

povik: I'm fine with it

passes/cmds/setundef.cc
widlarizer: If this isn't done, then bits() is iterated over. In my current implementation, that means unpacking. I don't think this is a readability problem, maybe I'm blind to that with things I wrote myself, but "skip these because they're not interesting anyway" is a common pattern that's easy to recognize for me

povik: Yes, but we don't care about unpacking for parameters

passes/opt/opt_merge.cc
widlarizer: This is a functional change that hashes fewer characters. However, since we're not concerned with string parameters, only string attributes, and this only works on parameters, the change wouldn't be used, so I guess I'll put it back as you suggest

kernel/rtlil.cc
widlarizer: Yes, it's used for pretty printing, which isn't supposed to be some kind of correctness-dependent process, but as a compact string representation readable by a human

kernel/rtlil.cc
widlarizer: Do you expect a runtime or readability improvement from having two iterators? Or do you think it's okay to never unpack?

kernel/rtlil.cc
povik: There's no guarantee someone will not be relying on the output not changing if the underlying value hasn't changed. Case in point: you have used it this way for hashing in opt_merge.cc

widlarizer: Good point, I didn't consider that the same string packed and unpacked then has to be equal in hashing input as well

kernel/rtlil.cc
povik: I think it's okay ot never unpack in the comparison operator, but I feel better about having a non-unpacking constant iterator, and using it to implement all operations where we care about not inadvertently unpacking, than about constructing temporary copies which we let unpack. There may be a minimal performance benefit too.

The constant iterator would also solve the semantical question, if comparisons were implemented on top of the same iterator no matter the backing.

widlarizer: as discussed elsewhere, I'm adding a const iterator that doesn't unpack

kernel/rtlil.cc
```diff
@@ -203,93 +203,198 @@ const pool<IdString> &RTLIL::builtin_ff_cell_types() {
...
+
+bool RTLIL::Const::operator<(const RTLIL::Const &other) const
 {
-	if (bits.size() != other.bits.size())
-		return bits.size() < other.bits.size();
-	for (size_t i = 0; i < bits.size(); i++)
-		if (bits[i] != other.bits[i])
-			return bits[i] < other.bits[i];
+	if (is_str() && other.is_str())
+		return str_ < other.str_;
```
povik, line 305: The comparison still leaks backing

kernel/rtlil.h
```diff
@@ -657,38 +665,205 @@ namespace RTLIL
...
 	bool operator ==(const RTLIL::Const &other) const;
 	bool operator !=(const RTLIL::Const &other) const;
 
+	std::vector<RTLIL::State>& bits();
 	bool as_bool() const;
 	int as_int(bool is_signed = false) const;
-	std::string as_string() const;
+	std::string as_string(const char* any = "-") const;
 	static Const from_string(const std::string &str);
+	std::vector<RTLIL::State> to_bits() const;
```
povik, line 715: We have `to_bits()` and `bits().to_vec()` which AFAICT do the same thing

kernel/rtlil.cc
```diff
@@ -203,93 +203,198 @@ const pool<IdString> &RTLIL::builtin_ff_cell_types() {
...
+
+std::vector<RTLIL::State> RTLIL::Const::to_bits() const
+{
+	if (auto bv = get_if_bits()) {
+		return *bv;
+	}
+	auto& str = get_str();
+	bitvectype b;
+	b.reserve(str.size() * 8);
+	for (int i = str.size()-1; i >= 0; i--) {
```
povik, line 351: This could be implemented on top of the constant iterator

kernel/rtlil.h
```diff
@@ -657,38 +665,205 @@ namespace RTLIL
...
+			using pointer = const State*;
+			using reference = State;
+
+			const_iterator(const Const& c, size_t i) : parent(c), idx(i) {}
+
+			State operator*() const {
+				if (auto bv = parent.get_if_bits())
+					return (*bv)[idx];
+
+				bool bit = (parent.get_str()[idx / 8] & (1 << (7 - (idx % 8))));
```
povik, line 798: The indexing here is reversed both for the character selection and the bit in a character

kernel/rtlil.h
```diff
@@ -657,38 +665,205 @@ namespace RTLIL
...
+
+			const_iterator operator+(int add) {
+				return const_iterator(parent, idx + add);
+			}
+			int operator-(const const_iterator& other) {
+				return idx - other.idx;
+			}
+
+			bool operator==(const const_iterator& other) const {
+				return idx == other.idx;
```
povik, line 830: `&& &parent == &other.parent`, or are comparisons on iterators from different parents undefined? Not sure here what is mandated

kernel/rtlil.h
```diff
@@ -657,35 +665,205 @@ namespace RTLIL
...
+		}
+		std::vector<State> to_vec() const {
+			std::vector<State> v;
+			for (auto bit : *this)
+				v.push_back(bit);
+			return v;
+		}
+	};
+
+	const const_bit_view bits() const {
```
povik, line 864: I'm a bit worried naming the methods the same will lead to confusion. What about not having the `bits() const` method at all, and instead retain `begin()` and `end()` only? This will mean there's only one way to iterate over a Const object, which is `for (State bit : constval)`

kernel/rtlil.h
widlarizer: Strange, I actually tested it before, but very manually

kernel/rtlil.h
widlarizer: 1. What do you mean by retain? Const doesn't have a `begin()` method on main
2. You mean we should remove the iterator that can mutate Const? I'm pretty sure there's code that relies on that

povik: > 1. What do you mean by retain? Const doesn't have a `begin()` method on main

Not on main but in this PR it does.

> 2\. You mean we should remove the iterator that can mutate Const? I'm pretty sure there's code that relies on that

No, the other one, basically get rid of `const_bit_view` but keep `begin()` and `end()` on the Const itself as the way of accessing what is now `const_bit_view::const_iterator`

widlarizer: So there will be two ways of iterating Const. One is `(auto a : c)` and the other is `(auto a : c.bits())`. Both give you the same values, but the second one is mutable and makes performance worse? It's doable, I suppose

povik: Yes, that's what I am proposing

kernel/rtlil.cc
```diff
@@ -203,93 +203,157 @@ const pool<IdString> &RTLIL::builtin_ff_cell_types() {
...
 bool RTLIL::Const::operator <(const RTLIL::Const &other) const
 {
-	if (bits.size() != other.bits.size())
-		return bits.size() < other.bits.size();
-	for (size_t i = 0; i < bits.size(); i++)
-		if (bits[i] != other.bits[i])
-			return bits[i] < other.bits[i];
+	const char* ctx = "operator<";
+	if (std::get_if<std::string>(&backing) != std::get_if<std::string>(&other.backing))
+		return decode_string() < other.decode_string();
```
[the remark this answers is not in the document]
widlarizer: Should be fixed now

frontends/ast/ast.cc
```diff
@@ -1749,16 +1741,7 @@ static std::string serialize_param_value(const RTLIL::Const &val) {
 		res.push_back('r');
 	res += stringf("%d", GetSize(val));
 	res.push_back('\'');
-	for (int i = GetSize(val) - 1; i >= 0; i--) {
```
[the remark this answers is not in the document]
widlarizer: I still think refactoring into as_string is the correct thing to do when we have the opportunity

povik (commented):
Looks good apart from the extra cxxrtl change and the extra assert

kernel/rtlil.cc
```diff
@@ -200,96 +200,196 @@ const pool<IdString> &RTLIL::builtin_ff_cell_types() {
...
 
 bool RTLIL::Const::operator !=(const RTLIL::Const &other) const
 {
-	return bits != other.bits;
+	return !(*this == other);
+}
+
+std::vector<RTLIL::State>& RTLIL::Const::bits()
+{
+	log_assert(is_bits());
```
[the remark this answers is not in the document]
widlarizer: It's redundant now, but it wasn't before

kernel/rtlil.cc
```diff
@@ -3730,7 +3913,7 @@ RTLIL::SigChunk::SigChunk(const RTLIL::SigBit &bit)
 	wire = bit.wire;
 	offset = 0;
 	if (wire == NULL)
-		data = RTLIL::Const(bit.data).bits;
+		data = RTLIL::Const(bit.data).to_bits();
```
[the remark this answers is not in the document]
widlarizer: nice catch

kernel/rtlil.cc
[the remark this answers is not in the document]
povik: As I understand it would crash on the following
```
Const c("blah"s);
c.bits();
```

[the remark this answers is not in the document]
widlarizer: true, I misread the called code

povik (approved):
I haven't looked at the new unit testing but the Const change itself looks fine with the points from the last review resolved

whitequark (commented):
Great work!

widlarizer:
just some fixes from the functional merges

widlarizer:
I validated the memory savings to be 0-30%.

ORFS peak:

| design | 5c9b2df68 main [MB] | 2dd0ae31f PR [MB] |
| --- | --- | --- |
|asap7_ibex|224.52|194.96|
|sky130hd_gcd|57.75|57.75|
|sky130hd_ibex|223.20|159.32|
|sky130hd_jpeg|726.52|613.00|

In the flow I dumped out the RTLIL before synthesis and outside of ORFS loaded the RTLIL and ran just the basic `synth` command:

| design | 5c9b2df68 main [MB] | 2dd0ae31f PR [MB] |
| --- | --- | --- |
|asap7_ibex|175.88|122.42|
|sky130hd_gcd|45.56|45.56|
|sky130hd_ibex|175.76|123.02|
|sky130hd_jpeg|85.14|75.63|

As for runtime, there was negligible improvement (0-1%)

KrystalDelusion:
This PR introduced a lot of warnings along the lines of
`comparison of integers of different signs: 'size_t' (aka 'unsigned long') and 'int' [-Wsign-compare]`, I'm guessing because the previous method of checking string length used the `.size()` method of a `std::vector` which returns a `size_t`, but the `RTLIL::const` `.size()` method returns it as an `int` (and seems to implicitly convert the underlying `.size()` `size_t` value)

widlarizer:
True. I brought up size types on a dev jf and we decided to stick with int for Yosys data structures so I guess I'll make `Const::size()` consistent with that then

PR #4520, opened 2024-08-04
Stop unconditionally building abc

KrystalDelusion:
abc builds unconditional because `check-git-abc` is a phony prerequisite and therefore always runs, and since it always runs it will always trigger abc to rebuild.

Convert `check-git-abc` to an order-only prerequisite.  It still runs as before, but no longer triggers yosys-abc to rebuild when it does.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -805,7 +805,7 @@ check-git-abc:
 		exit 1; \
 	fi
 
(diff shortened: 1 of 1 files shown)


widlarizer (approved):
Works as expected

mmicko (approved):
LGTM

PR #4523, opened 2024-08-05
Makefile: no LTO and lld by default

widlarizer:
On mac, there's problems with lld presence, compatibility, and detection with brew vs nixos. Disable lld and LTO by default and inform users in the README about the benefit of enabling both.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,17 @@ unless `CXX` is assigned in the call to make, e.g.
 
   $ make CXX=$CXX
 
+The Makefile has many variables influencing the build process. These can be
+adjusted by modifying the Makefile.conf file which is created at the
+`make config-...` step (see above), or they can be set by passing an option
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -116,6 +116,16 @@ unless `CXX` is assigned in the call to make, e.g.
 
   $ make CXX=$CXX
 
+The Makefile has many option variables with default values. These can be used
+by editing Makefile.conf that `make config-...` creates (see above), or by
```
povik, line 120: Alternative phrasing: The Makefile has many variables influencing the build process. These can be adjusted by modifying the Makefile.conf file which is created at the `make config-...` step (see above), or they can be set by passing an option to the `make` command directly.

povik (approved):
I can confirm this does build with the default toolchain on macOS.

PR #4525, opened 2024-08-06
peepopt: Add formal opt to rewrite latches to ffs in clock gates

georgerennie:

Common clock gating patterns use a latch, which after `clk2fflogic` leaves a combinational path between the enable input and clock output. This can cause comb loops even though there is no true path due to an AND gate gating the output when the latch is transparent. This can be optimized into an equivalent circuit using a flip-flop that no longer has this comb path.

This optimization is suitable for formal to prevent false comb loops, but should not be used for synthesis where the latch is an intentional choice.

```verilog
// Latch style:
always @* if (!clk_i) latched_en = en;
assign gated_clk_o = latched_en & clk_i;

// Flip-flop style:
always @(posedge clk) flopped_en = en;
assign gated_clk_o = flopped_en & clk_i;
```

A peephole optimizer detects and rewrites the standard latch into and gate pattern.

I have checked that this works correctly with the clock gating rtl from [ibex](https://github.com/lowRISC/ibex/blob/master/syn/rtl/prim_clock_gating.v) and [cv32e40p](https://github.com/openhwgroup/cv32e40p/blob/master/bhv/cv32e40p_sim_clock_gate.sv).

This is draft for now as I'm not sure what command it is best to put this in, seeing as it is only applicable from formal. Should it just be a check in `clk2fflogic` instead of a pmgen optimizer? I have temporarily added it to peepopt.

### DIFF
--- a/passes/pmgen/Makefile.inc
+++ b/passes/pmgen/Makefile.inc
@@ -57,6 +57,7 @@ PEEPOPT_PATTERN  = passes/pmgen/peepopt_shiftmul_right.pmg
 PEEPOPT_PATTERN += passes/pmgen/peepopt_shiftmul_left.pmg
 PEEPOPT_PATTERN += passes/pmgen/peepopt_shiftadd.pmg
 PEEPOPT_PATTERN += passes/pmgen/peepopt_muldiv.pmg
+PEEPOPT_PATTERN += passes/pmgen/peepopt_formal_clockgateff.pmg
 
 passes/pmgen/peepopt_pm.h: passes/pmgen/pmgen.py $(PEEPOPT_PATTERN)
 	$(P) mkdir -p passes/pmgen && $(PYTHON_EXECUTABLE) $< -o $@ -p peepopt $(filter-out $<,$^)
--- a/passes/pmgen/peepopt.cc
+++ b/passes/pmgen/peepopt.cc
@@ -40,7 +40,7 @@ struct PeepoptPass : public Pass {
 		log("\n");
 		log("This pass applies a collection of peephole optimizers to the current design.\n");
 		log("\n");
-		log("This pass employs the following rules:\n");
+		log("This pass employs the following rules by default:\n");
 		log("\n");
 		log("   * muldiv - Replace (A*B)/B with A\n");
 		log("\n");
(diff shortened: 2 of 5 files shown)


povik:
Should we make this conditional on `-formal`? That sounds reasonable to me

povik:
As in `peepopt -formal`

passes/pmgen/peepopt_formal_clock_gate.pmg
```diff
@@ -0,0 +1,60 @@
...
+	select latch->type == $dlatch
+	select param(latch, \WIDTH) == 1
+	select param(latch, \EN_POLARITY).as_bool() == false
+	set clk port(latch, \EN)
+	set en port(latch, \D)
+	set latched_en port(latch, \Q)
+endmatch
+
+match and_gate
+	if latch
```
povik, line 33: Since `latch` is not optional this may be superfluous

povik (approved):
As far as the pattern matcher and transformation logic go, this looks good to me. I don't think the two circuits are equivalent under synthesis semantics, but noone is suggesting using this in synthesis flows anyway

passes/pmgen/peepopt_formal_clock_gate.pmg
georgerennie: ahh yeah makes sense

georgerennie:
Yeah I think `-formal` could work. One thing is that I suppose it would be nice to run this at the start of `clk2fflogic` but I would want to guard against arbitrary formal focused peephole optimizations being run there, so maybe a more specific `-formal-clk` or something would be appropriate.

I think the circuits should actually be equivalent under synthesis semantics too. The latch should be equivalent to the following:
```verilog
always @(posedge clk)
    flopped_en <= en;
// Use stored value when enable (!clk_i) is low, otherwise pass through transparently
assign latched_en = !clk_i ? en : flopped_en;

assign gated_clk_o = latched_en & clk_i;
```
Whenever `clk_i` is low, the latch passes through its value, but this value is ANDed with `clk_i`, so is thus `0`. This means that the case where `latched_en = en` is unobservable at the output so `latched_en` can be optimized to `flopped_en`, giving the transformed circuit.

This being said, this would obviously be an area increase for synthesis and the whole thing I think is not that relevant anyway as if you are using a clock gate for synthesis you would instantiate a clock gating primitive (which combines the latch and AND) from a cell library directly instead of trying to infer it.

povik:
> This means that the case where latched_en = en is unobservable at the output so latched_en can be optimized to flopped_en, giving the transformed circuit.

That ignores the question of glitches on `gated_clk_o` on the rising edge of `clk_i`. I will ask some people who know better but a synthesis tool may guarantee the latched version will be glitch-free, while for the flopped version there's no hope.

georgerennie:
Ahh yes you are right I had forgotten that glitch. In the latch case I think it could glitch too (e.g. if the latch to and gate distance was very long). I think this type of thing is why people use the integrated clock gating cells which contain the latch and gate together and come with their own setup/hold requirements. Either way I don't propose we enable it for anything but formal so it doesn't matter much.

PR #4528, opened 2024-08-07
Bump ABC

povik:

ABC has new features (timing for `&nf` blackboxes; improved LUT mapping); also I will push for our fork to converge with upstream -- for that let's pull changes from upstream to our fork first.


povik:
```
6.2.2. Re-integrating ABC results.
../tools/autotest.sh: line 171: 48993 Segmentation fault: 11  "$toolsdir"/../../yosys -b "verilog $backend_opts" -o ${bn}_syn${test_count}.v "$@"
```

Huh, does ABC now write out BLIFs which make Yosys crash?

povik:
The failures were due to `&nf` losing PI/PO names. Upstream fix is at berkeley-abc/abc#317

povik:
This should be good to go. The content is:

 * https://github.com/YosysHQ/abc/commit/bf64a922536d6020f8a521c664826848bb0ecb54
 * berkeley-abc/abc#317
 * berkeley-abc/abc#318

PR #4536, opened 2024-08-13
Functional Backend

aiju:
Add a new intermediate representation to be used for writing backends for functional language-shaped targets, conversion code from RTLIL, and three example backends (C++, SMTLIB and Rosette).


povik:
Is this preserving undefined bits, or is this more of a synthesis target like CXXRTL?

aiju:
> Is this preserving undefined bits, or is this more of a synthesis target like CXXRTL?

Currently 'x' bits are treated as 'dont care' (mostly set to 0), which I think matches the existing formal infrastructure. The idea is that users can use xprop if they want 'x' behaviour.

povik:
FWIW a write-up on how this differs from CXXRTL (apart from this being language-agnostic) would help me understand what this is. I know you all are busy though.

aiju:
> FWIW a write-up on how this differs from CXXRTL (apart from this being language-agnostic) would help me understand what this is. I know you all are busy though.

I don't actually know how CXXRTL works in detail but the key point is that this is intended to match formal semantics and I think CXXRTL is a mix of simulation and synthesis semantics.

georgerennie:
> Currently 'x' bits are treated as 'dont care' (mostly set to 0), which I think matches the existing formal infrastructure

Is this just for the C++ target or also for smtlib/rosette? For C++ which requires concrete values this obviously makes sense to set to 0, but if being used with smtlib/rosette for verification surely you want to leave `x`s as symbolic variables that can take the value 0 or 1 (pseudo primary inputs) to verify all allowed behaviours?

KrystalDelusion:
smtlib/rosette (currently) pull down to 0

aiju:
> > Currently 'x' bits are treated as 'dont care' (mostly set to 0), which I think matches the existing formal infrastructure
> 
> Is this just for the C++ target or also for smtlib/rosette? For C++ which requires concrete values this obviously makes sense to set to 0, but if being used with smtlib/rosette for verification surely you want to leave `x`s as symbolic variables that can take the value 0 or 1 (pseudo primary inputs, the same as `$anyseq`) to verify all allowed behaviours?
> 
> Existing formal infrastructure by default models `x`s in this way for this reason (although doesn't model x-prop semantics unless `xprop` is used)

Oh I didn't know that's how it works. If it's just constants then that's probably not too difficult to add as an option. I bounced off x-semantics because adding x-propagation seemed too complicated and enforcing 2-value logic seemed like the easiest solution.

KrystalDelusion (approved):
Missing smtlib and rosette tests, but that can be fixed later.

mmicko (approved):

backends/functional/cxx_runtime/sim.h
```diff
@@ -0,0 +1,418 @@
...
+#include <array>
+#include <cassert>
+#include <string>
+#include <iostream>
+#include <algorithm>
+
+template<size_t n>
+class Signal {
+    template<size_t m> friend class Signal;
+    std::array<bool, n> _bits;
```
povik, line 33: Btw why this instead of `std::bitset<n>`?

backends/functional/test_generic.cc
```diff
@@ -0,0 +1,154 @@
...
+				insert_concatenated(low, values);
+			}
+		}
+	}
+
+};
+
+struct FunctionalTestGeneric : public Pass
+{
+	FunctionalTestGeneric() : Pass("test_generic", "test the generic compute graph") {}
```
povik, line 119: I suggest we rename this to `test_cgraph` or similar. This has the potential to confuse people

kernel/functional.cc
```diff
@@ -0,0 +1,853 @@
...
+	for(const auto &[name, state]: _states)
+		if(state.has_next_value())
+			toposort.process(state.next_value().id());
+	for(const auto &[name, output]: _outputs)
+		if(output.has_value())
+			toposort.process(output.value().id());
+	// any nodes untouched by this point are dead code and will be removed by permute
+    _graph.permute(perm);
+    if(scc) log_error("The design contains combinational loops. This is not supported by the functional backend. "
+		"Try `scc -select; simplemap; select -clear` to avoid this error.\n");
```
povik, line 759: This might be an outstanding TODO but `simplemap` will leave around gate cells which are unsupported by the functional backend as of yet

backends/functional/cxx_runtime/sim.h
aiju: i didn't know about `std::bitset` :) i guess rewriting it to use that would simplify the code a bit, but would still need to implement addition, etc

backends/functional/test_generic.cc
aiju: yea i'm happy to rename this, also `test_generic` needs some cleanup that i forgot to do...

kernel/functional.cc
aiju: oh yeah gate-level cells was on my TODO list but i didn't get around to it for the deliverable

PR #4537, opened 2024-08-13
Liberty parsing cleanup

povik:

N/A


widlarizer (approved):
This is a refactoring PR which also adds regression test for yosys-filterlib with builtin rules. Custom rule files aren't covered, which is fine, since they don't seem much loved anyway

povik:
Thanks for reviewing!

PR #4538, opened 2024-08-14
verific: Add bottom and top bound properties to wire

RCoeurjoly:
### What are the reasons/motivation for this change?
Having bound properties for wires would help detect underflow and overflow in cases where the user is more restrictive than the RTLIL representation.
For example, the following VHDL

`signal a : INTEGER range 0 to 6;
`
is represented in RTLIL as:
`wire width 3 input 1 \a
`because to represent decimal 6 you need at least 3 bits.
However, decimal 7 can also be represented, which would be an overflow as per the user specified range, and that user constraint would get lost without this attribute.
This patch helps ensure that those user constraints are maintained and accessible in the RTLIL representation.
### Explain how this is achieved.
With the Verific API, we get the bounds information and create attributes (named `bottom_bound` and `top_bound`).
Those attributes are of the same bit width as the wire they belong to.
If at least one the bound is negative, both constants are flagged as signed.
#### Supporting changes
printattrs.cc: Updated the logic to print attributes for signed constants.
RTLIL::Const: The constructor was modified to use `long long` instead of `int`. This is necessary because to support VHDL 2019 integers with bit widths up to 64, the Verific API’s `TypeRange::GetScalarRangeLeftBound()` method returns a `long long`.
### Discarded options
We decided that bound attributes for enumeration types will not be added in this PR.

### DIFF
--- a/frontends/verific/verific.cc
+++ b/frontends/verific/verific.cc
@@ -450,6 +450,19 @@ void VerificImporter::import_attributes(dict<RTLIL::IdString, RTLIL::Const> &att
 		auto type_range = nl->GetTypeRange(obj->Name());
 		if (!type_range)
 			return;
+		if (type_range->IsTypeScalar()) {
+			const long long bottom_bound = type_range->GetScalarRangeLeftBound();
+			const long long top_bound = type_range->GetScalarRangeRightBound();
+			const unsigned bit_width = type_range->NumElements();
+			RTLIL::Const bottom_const(bottom_bound, bit_width);
+			RTLIL::Const top_const(top_bound, bit_width);
+			if (bottom_bound < 0 || top_bound < 0) {
+				bottom_const.flags |= RTLIL::CONST_FLAG_SIGNED;
+				top_const.flags |= RTLIL::CONST_FLAG_SIGNED;
+			}
+			attributes.emplace(ID(bottom_bound), bottom_const);
+			attributes.emplace(ID(top_bound), top_const);
+		}
 		if (!type_range->IsTypeEnum())
 			return;
 #ifdef VERIFIC_VHDL_SUPPORT
--- a/tests/verific/bounds.ys
+++ b/tests/verific/bounds.ys
@@ -1,0 +1,6 @@
+read -vhdl bounds.vhd
+verific -import work
+select -assert-count 1 a:bottom_bound=5'bs11011
+select -assert-count 1 a:top_bound=5'bs01010
+select -assert-count 1 a:bottom_bound=5'bs11010
+select -assert-count 1 a:top_bound=5'bs01011
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -213,7 +213,7 @@ RTLIL::Const::Const(const std::string &str)
 	}
 }
 
-RTLIL::Const::Const(int val, int width)
+RTLIL::Const::Const(long long val, int width)
 {
 	flags = RTLIL::CONST_FLAG_NONE;
 	bits.reserve(width);
--- a/kernel/rtlil.h
+++ b/kernel/rtlil.h
@@ -662,7 +662,7 @@ struct RTLIL::Const
 
 	Const() : flags(RTLIL::CONST_FLAG_NONE) {}
 	Const(const std::string &str);
-	Const(int val, int width = 32);
+	Const(long long val, int width = 32);
 	Const(RTLIL::State bit, int width = 1);
 	Const(const std::vector<RTLIL::State> &bits) : bits(bits) { flags = CONST_FLAG_NONE; }
 	Const(const std::vector<bool> &bits);
--- a/passes/cmds/printattrs.cc
+++ b/passes/cmds/printattrs.cc
@@ -42,7 +42,7 @@ struct PrintAttrsPass : public Pass {
 	static void log_const(const RTLIL::IdString &s, const RTLIL::Const &x, const unsigned int indent) {
 		if (x.flags == RTLIL::CONST_FLAG_STRING)
 			log("%s(* %s=\"%s\" *)\n", get_indent_str(indent).c_str(), log_id(s), x.decode_string().c_str());
-		else if (x.flags == RTLIL::CONST_FLAG_NONE)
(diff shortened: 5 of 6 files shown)


frontends/verific/verific.cc
```diff
@@ -450,8 +450,30 @@ void VerificImporter::import_attributes(dict<RTLIL::IdString, RTLIL::Const> &att
...
+		  }
+		  
+		  const std::bitset<sizeof(long long) * CHAR_BIT> binary_right(right_bound);
+		  log_assert(bit_width <= binary_right.size());
+		  std::vector<bool> bits_right(bit_width);
+		  for (size_t i = 0; i < bit_width; ++i) {
+		    bits_right[i] = binary_right[i];
+		  }
+		  
+		  attributes.emplace(ID(left_bound), RTLIL::Const(bits_left));
```
povik, line 472: How do we tell from RTLIL if the value should be interpreted signed or unsigned?

frontends/verific/verific.cc
```diff
@@ -450,8 +450,30 @@ void VerificImporter::import_attributes(dict<RTLIL::IdString, RTLIL::Const> &att
 		auto type_range = nl->GetTypeRange(obj->Name());
 		if (!type_range)
 			return;
+		if (type_range->IsTypeScalar()) {
+		  const long long left_bound = type_range->GetScalarRangeLeftBound();
```
povik, line 454: I suggest bottom and top bound, or some other naming. I assume it's always the case `left_bound <= right_bound`

frontends/verific/verific.cc
RCoeurjoly: Good point. Actually, in a first iteration, it was automatically detected because all the bounds were 32 bit integers.
Now that they can have arbitrary width, we have to specify signedness.
Last commit adds a `is_bound_signed` just for that.

frontends/verific/verific.cc
RCoeurjoly: Actually, if we put a range such as:
        `a : in INTEGER range 15 to -15;`
Verific gives the following warning:

> VERIFIC-WARNING [VHDL-1200] negative_integer_range.vhdl:7: range is empty (null range)

But continues.
This results in a completely incoherent RTLIL:

  ```
attribute \bottom_bound 2'11
  attribute \is_bound_signed 1'1
  attribute \src "negative_integer_range.vhdl:7.9-7.10"
  attribute \top_bound 2'01
```

Which corresponds with a [-1, 1] decimal range.
This range has nothing to do with the original (wrong) range nor with the Verific warning.

Anyway, I changed to `bottom_bound` and `top_bound` in last commit. Thanks!

File: frontends/verific/verific.cc (hunk 1 above)
Comment (povik) [reply frontends/verific/verific.cc]: We could use the signed flag on Const

Comment (RCoeurjoly) [reply frontends/verific/verific.cc]: The issue with using the signed flag on Const is that info gets lost if we `write_rtlil` and then `read_rtlil`.
If we read the attributes in the same yosys session there is no issue, but I think we should also support exporting to RTLIL

Comment (povik) [reply frontends/verific/verific.cc]: > The issue with using the signed flag on Const is that info gets lost if we `write_rtlil` and then `read_rtlil`.

I opened an issue #4539

**Review by nakengelhardt — CHANGES_REQUESTED** [review]
Please make your formatting conform to the surrounding code/[the yosys coding style](https://github.com/YosysHQ/yosys/blob/main/guidelines/CodingStyle).

The tests do not seem to test anything, what is their intent?

File: frontends/verific/verific.cc [hunk 3]
```diff
@@ -450,8 +450,42 @@ void VerificImporter::import_attributes(dict<RTLIL::IdString, RTLIL::Const> &att
...
+		  
+		  const std::bitset<sizeof(long long) * CHAR_BIT> binary_top(top_bound);
+		  log_assert(bit_width <= binary_top.size());
+		  std::vector<bool> bits_top(bit_width);
+		  for (size_t i = 0; i < bit_width; ++i) {
+		    bits_top[i] = binary_top[i];
+		  }
+
+		  RTLIL::Const bottom_const(bits_bottom);
+		  RTLIL::Const top_const(bits_top);
```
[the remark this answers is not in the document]
RCoeurjoly: Done. Note that to do that we need to fix the resulting ambiguity with some constructors overloads.

RCoeurjoly:
> Please make your formatting conform to the surrounding code/[the yosys coding style](https://github.com/YosysHQ/yosys/blob/main/guidelines/CodingStyle).
> 
> The tests do not seem to test anything, what is their intent?

Formatting done.

The tests were meant to exercise the `log_assert`s, but now that there are none, we could remove them.

RCoeurjoly:
Coverted to draft because most likely we want this [PR](https://github.com/YosysHQ/yosys/pull/4550) to be merged first so that `is_bound_signed` attribute is not necessary.

frontends/verific/verific.cc
```diff
@@ -450,6 +450,18 @@ void VerificImporter::import_attributes(dict<RTLIL::IdString, RTLIL::Const> &att
 		auto type_range = nl->GetTypeRange(obj->Name());
 		if (!type_range)
 			return;
+		if (type_range->IsTypeScalar()) {
+			const long long bottom_bound = type_range->GetScalarRangeLeftBound();
+			const long long top_bound = type_range->GetScalarRangeRightBound();
+			RTLIL::Const bottom_const(bottom_bound, 64);
+			RTLIL::Const top_const(top_bound, 64);
```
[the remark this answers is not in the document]
RCoeurjoly: Done in last commit. Thanks!

povik (approved):
I can't speak to how the Verific interface is used, but otherwise it looks good to me

tests/verific/bounds.ys
```diff
@@ -0,0 +1,6 @@
+read -vhdl bounds.vhd
+verific -import work
+select -assert-count 1 a:bottom_bound=5'bs11011
+select -assert-count 1 a:top_bound=5'bs01010
+select -assert-count 1 a:bottom_bound=5'bs11010
+select -assert-count 1 a:top_bound=5'bs01011
```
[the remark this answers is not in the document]
povik: > I think still the test is better than just checking the existence of `bottom_bound` and `top_bound` as it also takes the length and value into account.

Sure, I fully agree

nakengelhardt (approved):

PR #4540, opened 2024-08-15
Replace test-compile (ubuntu-22.04, clang-11)

KrystalDelusion:
[test-compile (ubuntu-22.04, clang-11)](https://github.com/YosysHQ/yosys/actions/runs/10374983466/job/28755517074?pr=4537#logs) is broken as of 14/08/2024 since clang-11 and clang-12 were dropped from ubuntu-22.04.

Replace with ubuntu-20.04, which also has earlier versions of clang so we can test clang-10 instead.


mmicko (approved):
Thanks. Let's wait for CI to confirm all is fine

PR #4541, opened 2024-08-15
Resolve (some) compiler warnings

KrystalDelusion:
The test-compile.yml action ends up with a lot of warnings, e.g. [latest main run](https://github.com/YosysHQ/yosys/actions/runs/10406437503) has 47 warnings.  Many of these warnings are simple things like uninitialized/unused variables and deprecated system functions.

Assign default values to variables, add `ifdef` guards to debug variables, replace sprintf with snprintf, and replace `realloc` in the minisat library with `reallocarray` when using GCC.

This should be passing all tests (and with fewer warnings), but there might be some weird edge case compilations that I'm not aware of

### DIFF
--- a/kernel/fmt.cc
+++ b/kernel/fmt.cc
@@ -832,7 +832,7 @@ std::string Fmt::render() const
 						}
 					} else log_abort();
 					if (part.justify == FmtPart::NUMERIC && part.group && part.padding == '0') {
-						int group_size = part.base == 10 ? 3 : 4;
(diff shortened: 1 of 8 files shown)


widlarizer:
This touches fstapi which has been vendored from gtkwave. In the meantime, there seems to have been duplicated work on problems with it within yosys and in gtkwave. I see @mmicko has been involved with this before. What is the motivation for not tracking gtkwave upstream fstapi closely?

KrystalDelusion:
That is a good question

PR #4542, opened 2024-08-15
Local readthedocs

KrystalDelusion:
[Yosys ReadtheDocs](https://yosyshq.readthedocs.io/projects/yosys/en/latest/) is currently built from a separate repo, [yosys-cmd-ref](https://github.com/YosysHQ-Docs/yosys-cmd-ref).  This would occasionally lead to problems with people looking at the wrong repository for the documentation source, since yosys-cmd-ref is automatically updated from this repo with the verific runner.  This was being done since a portion of the documentation requires yosys to be compiled in order to generate, which would take too long for the ReadtheDocs builder and would generate without verific compatibility which also changes the generated documentation.

Because the verific runner would only update documentation from the main branch, the current setup also makes it more difficult to preview documentation changes for, e.g., yosys PRs.  The yosys-cmd-ref repo also ends up with a lot of code-generated images and such which get committed/updated from time to time.

This PR provides the necessary framework to allow ReadtheDocs to be built from the Yosys repository directly, making use of the [rtds-action](https://github.com/dfm/rtds-action) to create build artifacts and trigger ReadtheDocs to build using those artifacts; negating the need for a second repository.

[These docs](https://tyrtd.readthedocs.io/en/latest/) were built from [my fork](https://github.com/KrystalDelusion/yosys/tree/rtd) of this repo with all of the expected artifacts (compiled on the standard git runner without verific).  And [another example](https://tyrtd--2.org.readthedocs.build/en/2/) demonstrating how [pr previews](https://github.com/KrystalDelusion/yosys/pull/2) could work with this.

If this PR is passing all checks, it should be safe to merge into main.  The current status of the ReadtheDocs integration is that the verific runner builds and uploads the artifact needed, and can trigger the ReadtheDocs webhook to build, but ReadtheDocs rejects building because krys/rtd is not a branch that it recognises as needing to publish.  Once this gets merged into main, I suspect that the verific runner will run the merged commit, triggering the webhook from the main branch, which will then update the 'latest' version of the ReadtheDocs.  If for some reason that breaks, manually rebuilding on ReadtheDocs will continue to use the yosys-cmd-ref repo.

### DIFF
--- a/.github/workflows/test-verific.yml
+++ b/.github/workflows/test-verific.yml
@@ -47,30 +47,6 @@ jobs:
         run: |
           make install DESTDIR=${GITHUB_WORKSPACE}/.local PREFIX=
 
-      - name: Checkout Documentation
-        if: ${{ github.ref == 'refs/heads/main' }}
-        uses: actions/checkout@v4
-        with:
-          path: 'yosys-cmd-ref'
-          repository: 'YosysHQ-Docs/yosys-cmd-ref'
-          fetch-depth: 0
-          token: ${{ secrets.CI_DOCS_UPDATE_PAT }}
-          persist-credentials: true
-
-      - name: Update documentation
-        if: ${{ github.ref == 'refs/heads/main' }}
-        run: |
-          make docs
-          rm -rf docs/build
-          cd yosys-cmd-ref
-          rm -rf *
-          git checkout README.md
-          cp -R ../docs/* .
-          rm -rf util/__pycache__
-          git add -A .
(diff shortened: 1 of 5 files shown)


mmicko (approved):
Looks good. 
Reminder for me TODO after release:
 - remove access secrets on this repo
 - yosys-cmd-ref - probably we can just archive this one
 - also disable readthedocs old webhooks that are not used any more
 - disable yosyshq-ci user access to yosys-cmd-ref

PR #4543, opened 2024-08-16
libs/fst: Update from upstream

KrystalDelusion:
https://github.com/YosysHQ/yosys/pull/4541#issuecomment-2293151370
i.e. why don't we track upstream fstapi

Add and run shell script (based on minisat lib) to clone and copy relevant files.

Need to check if there are any changes lost that we need to patch back in.


KrystalDelusion:
Windows build (or rather, vs build) now completes successfully, but it doesn't run any tests, and I'm not sure if we have any tests for fst.

mmicko:
Can confirm that tests are passing on Windows machine when build done by mingw.

PR #4544, opened 2024-08-16
Fixed #321, support for reduce ops in share command

### LINKED ISSUE #321: share -force with reduce ops.
Yosys will exit on share.cc at line 720 when
c1->type == "$reduce_or"
Might fail also with other reduce operators. I also noticed there is special handling for $memrd but not for $memwr.

canaknesil:
The share command aborts if the cell type is a reduce operation. This fix prevents the abort by supporting `$reduce_and/or/xor/xnor/bool`.

Added the following test that would fail before the fix:

```
read_verilog <<EOF
module top(a, b, s, y);
	input wire [7:0] a;
	input wire [7:0] b;
	input wire s;
	output wire y;
	assign y = s ? |a : |b;
endmodule
EOF

share -force
```

### DIFF
--- a/passes/opt/share.cc
+++ b/passes/opt/share.cc
@@ -1486,6 +1486,12 @@ struct SharePass : public Pass {
 		config.generic_cbin_ops.insert(ID($logic_and));
 		config.generic_cbin_ops.insert(ID($logic_or));
 
+		config.generic_uni_ops.insert(ID($reduce_and));
+		config.generic_uni_ops.insert(ID($reduce_or));
+		config.generic_uni_ops.insert(ID($reduce_xor));
+		config.generic_uni_ops.insert(ID($reduce_xnor));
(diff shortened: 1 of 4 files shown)


widlarizer:
The FSM tests, that do use `share`, fail but only on macOS. Strange...

Tagging in a comment so we get a link here on github: this fixes #321

widlarizer:
The reason it only failed on macos is because the seed is by default random and each runner ran FSM tests with a different seed. `./run-test.sh -S 358331205987550812` in `tests/fsm` fails in case 15 on this commit, but not on its parent commit

jix:
I'm not sure whether the handling of `generic_uni_ops` in `make_supercell` is appropriate for the reduction operators. In particular it's not clear to me that the way cells of different widths can be merged and how zero and sign extension are handled for that can be done the same way for all of the reduction operators.

Issue #4453, opened 2024-06-13
IDES4_MEM gowin module is missing


LaneaLucy:
### Version

Yosys 0.42+12 (git sha1 62bff3a20, g++ 11.4.0-1ubuntu1~22.04 -fPIC -Os)

### On which OS did this happen?

Linux

### Reproduction Steps

try to synthesize the output of litex

to get the verilog file:
python3 -m litex_boards.targets.sipeed_tang_primer_20k --build

here a example yosys script
```
verilog_defaults -push
verilog_defaults -add -defer
read_verilog /home/lanealucy/litex/pythondata-cpu-vexriscv/pythondata_cpu_vexriscv/verilog/VexRiscv.v
read_verilog /home/lanealucy/litex/build/sipeed_tang_primer_20k/gateware/sipeed_tang_primer_20k.v
verilog_defaults -pop
attrmap -tocase keep -imap keep="true" keep=1 -imap keep="false" keep=0 -remove keep=0

synth_gowin  -top sipeed_tang_primer_20k
write_json  sipeed_tang_primer_20k.json
``` 

### Expected Behavior

synthesize 

### Actual Behavior

```
 /----------------------------------------------------------------------------\
 |  yosys -- Yosys Open SYnthesis Suite                                       |
 |  Copyright (C) 2012 - 2024  Claire Xenia Wolf <claire@yosyshq.com>         |
 |  Distributed under an ISC-like license, type "license" to see terms        |
 \----------------------------------------------------------------------------/
 Yosys 0.42+12 (git sha1 62bff3a20, g++ 11.4.0-1ubuntu1~22.04 -fPIC -Os)

-- Executing script file `sipeed_tang_primer_20k.ys' --

1. Executing Verilog-2005 frontend: /home/lanealucy/litex/pythondata-cpu-vexriscv/pythondata_cpu_vexriscv/verilog/VexRiscv.v
Parsing Verilog input from `/home/lanealucy/litex/pythondata-cpu-vexriscv/pythondata_cpu_vexriscv/verilog/VexRiscv.v' to AST representation.
Storing AST representation for module `$abstract\VexRiscv'.
Storing AST representation for module `$abstract\DataCache'.
Storing AST representation for module `$abstract\InstructionCache'.
Successfully finished Verilog frontend.

2. Executing Verilog-2005 frontend: /home/lanealucy/litex/build/sipeed_tang_primer_20k/gateware/sipeed_tang_primer_20k.v
[...]

PR #4545, opened 2024-08-17
Adding missing Gowin cells

leviathanch:
Two memory cell defines were missing for running through a Litex design

This PR fixes the open issue in https://github.com/YosysHQ/yosys/issues/4453


widlarizer:
Thanks for contributing. @yrabbit should this be a blackbox? What about the other "empty" cells in `techlibs/gowin/cells_sim.v`?

For style and convenience: The main branch in your fork should only ever be behind the main branch on the upstream (YosysHQ/yosys) and work intended for PRs should be from different branches. This is because if you use your main branch to open a PR, then you can't use it for another PR, and you have no clean main branch to start a new branch from. Also, when the PR gets merged, and you push new commits onto your forked main, those commits will show up in this PR as well after merging, which is a bit confusing. Instead of merging branches into your PR branch, it's better to rebase on top of the main branch. It's not strictly necessary to keep a PR source branch perfectly rebased on the destination branch when there's no likely conflicts.

Either I'll squash the commits here before merging, or you can reset your fork's main branch to upstream and open a new PR with a new branch

yrabbit:
To start on the second part - I think we're definitely getting to a point where the descriptions of Gowin primitives in Yosus need to be reworked - from removing the overlap between cells_xtra.v and cells_sim.v to revising attributes and adding actual simulation wherever we can.
The stage when the main task was to add support for primitives is coming to an end and the quality of this support needs to be improved. 

And on the first part - I believe that IO (even SERDES) can be described in verilog for simulation purposes, but my opinion is not worth much - better ask the founders of Yosys ;).

leviathanch:
OK, I will create a branch and make a new PR.
I've already emailed with @yrabbit a few days ago for getting apicula working on my setup here and I agree that we should think about a better solution, but for now that's the quickest fix for getting synthesis working on my system.

widlarizer:
This PR has been moved to #4547

PR #4547, opened 2024-08-18
Add DQS and related primitives to Gowin tech files

leviathanch:
Without OSER4_MEM, IDES4_MEM and DQS the synthesis of my Rocket Chip design for my Sipeed Tang FPGA fails.

Two memory cell defines were missing for running through a Litex design

This PR fixes the open issue in https://github.com/YosysHQ/yosys/issues/4453


widlarizer (approved):
Matches UG289E

Issue #4539, opened 2024-08-15
Preserve `SIGNED` flag on attribute values through RTLIL export/import


povik:
### Feature Description

Looking inside `rtlil_backend.cc` (1eaf4e07) one can see the `SIGNED` flag is not preserved for attribute values.

For cell parameters it's preserved already. That probably has to do with proper sign extension when hierarchy is elaborated, while for attributes the need to preserve it hasn't been as pressing.

PR #4550, opened 2024-08-19
Ensure signed constants are correctly parsed and exported in RTLIL. Add a test to check parsing and exporting

RCoeurjoly:

Fix https://github.com/YosysHQ/yosys/issues/4539

Updated rtlil_lexer.l to recognize constants with an optional s for signedness.
Modified rtlil_parser.y to set RTLIL::CONST_FLAG_SIGNED when appropriate.
Adjusted rtlil_backend.cc to correctly export signed constants.
Added a test (rtlil_signed_attribute.ys) for round-trip verification of signed constants.

I have included a test doing round trip test. I assume the rest of tests could catch relevant bugs.

### DIFF
--- a/.github/workflows/test-verific.yml
+++ b/.github/workflows/test-verific.yml
@@ -112,6 +112,7 @@ jobs:
             docs/source/code_examples
 
       - name: Trigger RTDs build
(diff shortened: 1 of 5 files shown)


whitequark (approved):
LGTM

PR #4552, opened 2024-08-19
docs: Only trigger RTDs on main

KrystalDelusion:
PRs from forks don't have access to secrets and will error when trying to trigger ReadtheDocs, e.g. https://github.com/YosysHQ/yosys/actions/runs/10456408410/job/28954216463?pr=4550

Add back the "if ref == main" check from previous docs building.

The "Trigger RTDs build" step of this pr's action should be skipped.

### DIFF
--- a/.github/workflows/test-verific.yml
+++ b/.github/workflows/test-verific.yml
@@ -112,6 +112,7 @@ jobs:
             docs/source/code_examples
 
       - name: Trigger RTDs build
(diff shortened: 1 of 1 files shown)


KrystalDelusion:
Ideally we would have some way to run on other branches that we want to generate docs (such as for previews), but for now we can use the previous behaviour to prevent errors.

mmicko (approved):
This is as well what we had before change.

PR #4553, opened 2024-08-19
-y flag for libyosys Python scripts

donn:

This adds a Python equivalent to the `-c` option, where scripts importing `libyosys` can be used. 

Most of the requisite work for this was already done to enable Python passes a couple years back, so this is a relatively small changeset, and would add another option for supporting libyosys-based scripts (that does not involve PYTHONPATH shenanigans.)

Adding a new opt, `-y` (`-p` and `-P` were already taken,) and running `PyRun_SimpleFile` on `scriptfile`. Similar to `-c` for Tcl, any unprocessed arguments are passed to the script in the form of `sys.argv`.

```sh
./yosys -y <(echo "
import libyosys as ys

d = ys.Design()
ys.run_pass('help', d)
print(sys.path)
print(sys.argv)") -- test argv
```

### DIFF
--- a/kernel/driver.cc
+++ b/kernel/driver.cc
@@ -305,12 +305,12 @@ int main(int argc, char **argv)
 		printf("        execute the commands in the tcl script file (see 'help tcl' for details)\n");
 		printf("\n");
 		printf("    -C\n");
-		printf("        enters TCL interatcive shell mode\n");
+		printf("        enters TCL interactive shell mode\n");
 #endif
 #ifdef WITH_PYTHON
 		printf("\n");
 		printf("    -y python_scriptfile\n");
-		printf("        execute the python script");
+		printf("        execute a python script with libyosys available as a built-in module\n");
 #endif
 		printf("\n");
 		printf("    -p command\n");
--- a/kernel/driver.cc
+++ b/kernel/driver.cc
@@ -665,13 +665,19 @@ int main(int argc, char **argv)
 			PyObject_SetAttrString(sys, "argv", new_argv);
 			Py_DECREF(old_argv);
 
-			PyRun_SimpleString(("import os;sys.path.insert(0, os.path.dirname(os.path.abspath(\""+scriptfile+"\")))").c_str());
+			PyObject *py_path = PyUnicode_FromString(scriptfile.c_str());
+			PyObject_SetAttrString(sys, "_yosys_script_path", py_path);
+			Py_DECREF(py_path);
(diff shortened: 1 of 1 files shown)


donn:
Any feedback on this? (The Verific test is failing because of security reasons- I'm a first-time contributor)

KrystalDelusion:
> Any feedback on this? (The Verific test is failing because of security reasons- I'm a first-time contributor)

That should be fixed by #4552, you may need to merge/rebase on latest main to get the fix though

donn:
Rebased on the wrong upstream for a second there but should be fine now :)

donn:
Any feedback?

kernel/driver.cc
```diff
@@ -636,7 +651,30 @@ int main(int argc, char **argv)
...
+			PyObject *new_argv = PyList_New(argc - optind + 1);
+			PyList_SetItem(new_argv, 0, PyUnicode_FromString(scriptfile.c_str()));
+			for (int i = optind; i < argc; ++i)
+				PyList_SetItem(new_argv, i - optind + 1, PyUnicode_FromString(argv[i]));
+
+			PyObject *old_argv = PyObject_GetAttrString(sys, "argv");
+			PyObject_SetAttrString(sys, "argv", new_argv);
+			Py_DECREF(old_argv);
+
+			PyRun_SimpleString(("import os;sys.path.insert(0, os.path.dirname(os.path.abspath(\""+scriptfile+"\")))").c_str());
```
jix, line 668: I think the escaping issues I mentioned on the call could be avoided using something like this instead, similar how argv is already handled:

```
  PyObject *py_path = PyUnicode_FromString(scriptfile.c_str());
  PyObject_SetAttrString(sys, "_yosys_script_path", py_path);
  Py_DECREF(py_path);
  PyRun_SimpleString("import os, sys; sys.path.insert(0, path.dirname(os.path.abspath(sys._yosys_script_path)))");
```

I tested this with a small stand-alone program and had to import both `os` and `sys` in the python snippet.

kernel/driver.cc
```diff
@@ -636,7 +651,30 @@ int main(int argc, char **argv)
...
+
+			FILE *scriptfp = fopen(scriptfile.c_str(), "r");
```
jix, line 670: This also needs a check whether `scripfp` is a nullptr in case the file isn't present (or couldn't be opened for other reasons).

jix (changes requested):

PR #4554, opened 2024-08-19
internal_stats: init, report current memory consumption on linux and mac

widlarizer:
To find out static memory usage right after RTLIL generation and at other points in the flow, currently the best practice is to dump and reload and inspect peak memory usage. This PR adds the experimental `devstat` command that prints in a JSON format the current memory usage. It is intended to be expanded further for information about the efficiency of RTLIL object representation and anything else that's relevant to yosys developers but not users.

On linux, yosys inspects its `/proc/self/status` file. This may not be present on some Linux distributions, but it should be handled gracefully should that file not exist.

On macOS, I use [task_info](https://developer.apple.com/documentation/kernel/1537934-task_info)

- [ ] human readable format when `-json` isn't used

### DIFF
--- a/passes/cmds/internal_stats.cc
+++ b/passes/cmds/internal_stats.cc
@@ -1,0 +1,120 @@
+/*
+ *  yosys -- Yosys Open SYnthesis Suite
+ *
+ *  Copyright (C) 2012  Claire Xenia Wolf <claire@yosyshq.com>
+ *
+ *  Permission to use, copy, modify, and/or distribute this software for any
+ *  purpose with or without fee is hereby granted, provided that the above
+ *  copyright notice and this permission notice appear in all copies.
+ *
+ *  THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ *  WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ *  MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ *  ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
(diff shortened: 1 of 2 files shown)


passes/cmds/internal_stats.cc
```diff
@@ -0,0 +1,120 @@
...
+		if(!json_mode)
+			log_header(design, "Printing internal statistics.\n");
+
+		log_experimental("internal_stats");
+
+		if (json_mode) {
+			log("{\n");
+			log("   \"creator\": %s,\n", json11::Json(yosys_version_str).dump().c_str());
+			std::stringstream invocation;
+			std::copy(args.begin(), args.end(), std::ostream_iterator<std::string>(invocation, " "));
```
povik, line 103: I think it's best to serialize that as a list

passes/cmds/internal_stats.cc
```diff
@@ -0,0 +1,120 @@
...
+	return {};
+#endif
+}
+
+struct InternalStatsPass : public Pass {
+	InternalStatsPass() : Pass("internal_stats", "print some internal statistics") { }
+	void help() override
+	{
+		//   |---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|
+		log("\n"); // TODO
```
povik, line 77: Help missing, not something we should let into a release

povik (approved):
Tested on macOS, we can get this into shape as we go, it's useful already

passes/cmds/internal_stats.cc
widlarizer: I just copy pasted that from the json backend

passes/cmds/internal_stats.cc
KrystalDelusion: TIL empty help messages causes docs to fail (granted it is technically a warning, but warnings are errors in our ReadtheDocs setup)
`/home/docs/checkouts/readthedocs.org/user_builds/yosys/checkouts/latest/docs/source/cmd/internal_stats.rst:21: WARNING: Literal block expected; none found. [docutils]`

PR #4555, opened 2024-08-19
Improve nested list assignments

leviathanch:
I've encountered some issues when trying to synthesize the most recent Rocket Core with Yosys,
those changes fix those issues and now all that's left is to solve the issues with automatic logic
inside an always block


whitequark:
It looks like there are multiple unrelated changes in the same PR, can you separate them into one change per commit?

zachjs:
Also, can you please look into adding some test cases for these features? Without them, we might accidentally break them in the future.

frontends/verilog/Makefile.inc
```diff
@@ -6,7 +6,7 @@ GENFILES += frontends/verilog/verilog_lexer.cc
 
 frontends/verilog/verilog_parser.tab.cc: frontends/verilog/verilog_parser.y
 	$(Q) mkdir -p $(dir $@)
-	$(P) $(BISON) -Wall -Werror -o $@ -d -r all -b frontends/verilog/verilog_parser $<
+	$(P) $(BISON) --debug  -Wall -Werror -o $@ -d -r all -b frontends/verilog/verilog_parser $<
```
widlarizer, line 9: Was this left in by accident?

leviathanch: Yeah, oops, I'll fix that, one moment

leviathanch: Fixed, rebased and push -f

Please review again

widlarizer:
Will reopen when test provided

PR #4559, opened 2024-08-22
Break tests with a small hash function perturbation

widlarizer:

Test failures due to hashing refactoring have uncovered `dfflibmap.ys` to fail to prove the first step base case in the first `equiv_opt` call. This PR demonstrates that it is sufficient to slightly change the hashing function to break that test. Famously, go lang standard library and abseil randomize hash functions to prevent users on depending on the ordering of items in unordered maps like dicts and sets. In Yosys, the DJB2 hash function hasn't been touched in 10 years. This draft PR will be closed when this problem is resolved and the hash function has a mechanism when it's randomized. Until then, it's a reproducer.

There is no way to write an "incorrect" hash function that isn't meant to be cryptographic, since hash functions are only meant to be used for fast path handling of data. A `return 1;` hash function is just as valid as an actual hash function, though it make tests run for too long. Therefore this PR should never make correctness tests fail.

Every time you hash, xorshift the result, and xor with a fudge constant. For example, you can reproduce this with a fudge constant of 123456 or 42069, but not 3.

### DIFF
--- a/kernel/hashlib.h
+++ b/kernel/hashlib.h
@@ -24,11 +24,6 @@ namespace hashlib {
 const int hashtable_size_trigger = 2;
 const int hashtable_size_factor = 3;
 
-// The XOR version of DJB2
-inline unsigned int mkhash(unsigned int a, unsigned int b) {
-	return ((a << 5) + a) ^ b;
-}
-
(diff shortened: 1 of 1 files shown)


widlarizer:
Adding a command line option to do this as a part of #4524

PR #4563, opened 2024-08-22
Better docs builds

KrystalDelusion:

- The current setup for rtds_action doesn't work locally.
- There are also cases where ReadtheDocs will build before the docs artifact is ready and upload an incomplete version of the documentation.

- Gate rtds_action behind checks for both `READTHEDOCS` and `GITHUB_TOKEN` environment variables.  This also has some changes to (better) support building on mac and other small formatting fixes.
- Update `.readthedocs.yaml` to fail if sphinx raises any warnings, notably because it's missing files.  This also involves fixing the couple of warnings that were still popping up.  I also ended up splitting the `prepare-docs` job from the `test-verific` in order to better control when the job would run.  There is an extra change so that todos appear in preview builds but not in release builds.

Run `make docs` locally.  Check [docs-previewtest tag](https://github.com/YosysHQ/yosys/releases/tag/docs-previewtest) (https://yosyshq.readthedocs.io/projects/yosys/en/docs-previewtest/).

### DIFF
--- a/.github/workflows/prepare-docs.yml
+++ b/.github/workflows/prepare-docs.yml
@@ -1,0 +1,52 @@
+name: Build docs artifact with Verific
+
+on: push
+
+jobs:
+  prepare-docs:
+    # docs builds are needed for anything on main, any tagged versions, and any tag
+    # or branch starting with docs-preview
+    if: ${{ github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/docs-preview') || startsWith(github.ref, 'refs/tags/') }}
+    runs-on: [self-hosted, linux, x64, fast]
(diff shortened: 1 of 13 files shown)


KrystalDelusion:
As a note, I was running `export "LC_ALL=C" make docs PYTHON=python`.  The `LC_ALL` was to resolve an error about an unsupported locale when using sphinx, and the `PYTHON=python` was because I was getting an error when trying to `python3`/`pip3 install` about it being externally managed and suggesting using a venv instead, whereas `python`/`pip` didn't... Not sure if either of those are because mac, or specific to my partner's machine/setup.

mmicko (approved):

PR #4564, opened 2024-08-23
Fix nix devShell warning

RCoeurjoly:
Fix "bash: warning: setlocale: LC_ALL: cannot change locale (en_US.UT…F-8)" warning on nix devShell

Quality of life.
We set the locale correctly in the nix devShell.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -41,6 +41,7 @@
         packages.default = yosys;
         defaultPackage = yosys;
         devShell = pkgs.mkShell {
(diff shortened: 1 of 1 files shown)


widlarizer:
Can you explain the problem some more and point to standard nix practice?

RCoeurjoly:
> Can you explain the problem some more and point to standard nix practice?

After reconsidering, this issue is better handled locally by each user setting LOCALE_ARCHIVE with `nix profile`, so there's no need for a global change. I've added documentation on how to set this with `nix profile` (since `nix-env` is deprecated): https://wiki.nixos.org/wiki/Locales.

Closing the PR.

PR #4566, opened 2024-08-26
Add -blackbox flag to cutpoint pass

RCoeurjoly:
Black/white boxes cannot be processed by `sby`, so they must be substituted by `$anyseq` cells.
Adding a `-blackbox` flag to `cutpoint`.
If that flag is set, we find all modules that are black/white boxes, add their module names to the `select` pattern, and remove their black/whitebox attribute.

### DIFF
--- a/passes/sat/cutpoint.cc
+++ b/passes/sat/cutpoint.cc
@@ -72,7 +72,6 @@ struct CutpointPass : public Pass {
 				}
 			}
 			std::ostringstream pattern;
-
 			for (size_t i = 0; i < blackbox_module_names.size(); ++i) {
 				pattern << blackbox_module_names[i].str();
 				if (i != blackbox_module_names.size() - 1) {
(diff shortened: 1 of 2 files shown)


nakengelhardt (changes requested):
This approach won't work. `extra_args` can throw a message about invalid arguments, and the design must not be modified in any way prior to that. And given that the select argument is a stack-based sequence of instructions, appending something to the user-provided argument can have quite unpredictable and difficult-to-debug consequences. Please make a separate loop in the body of the pass that handles modifying the blackboxes directly.

RCoeurjoly:
> This approach won't work. `extra_args` can throw a message about invalid arguments, and the design must not be modified in any way prior to that. And given that the select argument is a stack-based sequence of instructions, appending something to the user-provided argument can have quite unpredictable and difficult-to-debug consequences. Please make a separate loop in the body of the pass that handles modifying the blackboxes directly.

Yes, I think the approach followed in https://github.com/YosysHQ/yosys/pull/4565 could be more succesful.
It takes care of not breaking passes that need to filter out blackboxes and at the same time enables blackbox selection when needed.

PR #4567, opened 2024-08-27
cxxrtl: use octal encoding of non-printables.

kivikakk:
`"\x0a"` is a perfectly valid escape sequence, but unfortunately `"\x0ac"` is equivalent to `"\xac"`, and not `"\x0a" "c"` as we might expect --- *any* number of hexadecimal characters after the `\x` is accepted. This can be hit pretty easily if a newline is present in a format string.

`"\x{...}"` syntax is only available as of C++23, so use octal format instead; a maximum of 3 digits following the backslash is accepted.

The alternative would be to render every escape like `" "\x0a" "`, but it seems more effort that way.

cc @whitequark.

### DIFF
--- a/backends/cxxrtl/cxxrtl_backend.cc
+++ b/backends/cxxrtl/cxxrtl_backend.cc
@@ -616,7 +616,7 @@ std::string escape_c_string(const std::string &input)
 				output.push_back('\\');
 			output.push_back(c);
 		} else {
-			char l = c & 0x3, m = (c >> 3) & 0x3, h = (c >> 6) & 0x3;
+			char l = c & 0x7, m = (c >> 3) & 0x7, h = (c >> 6) & 0x3;
 			output.append("\\");
 			output.push_back('0' + h);
 			output.push_back('0' + m);
--- a/kernel/fmt.cc
+++ b/kernel/fmt.cc
@@ -634,10 +634,11 @@ std::string escape_cxx_string(const std::string &input)
 				output.push_back('\\');
 			output.push_back(c);
 		} else {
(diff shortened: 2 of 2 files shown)


whitequark:
Oh oops, I could swear I've fixed this--I remember [making a poll](https://mastodon.social/@whitequark/112403120959273105) recently where basically nobody got it right. (Including me, evidently. I have no idea what that was about then...)

Thanks for fixing this!

whitequark (approved):

kivikakk:
Oh! Yeah, I didn't even see that either; amusing that we ended up writing almost exactly the same thing — but that one looks subtly wrong? The low and middle parts only have two significant bits each (`& 0x3`), but they should have three (`& 0x7`). I might add a fix for that to this PR?

whitequark:
> but that one looks subtly wrong?

Oh. <.<

> The low and middle parts only have two significant bits each (`& 0x3`), but they should have three (`& 0x7`). I might add a fix for that to this PR?

Please do. <.<

(What an embarrassing mistake...)

kivikakk:
Hehehe, nah, very easy to do. (Just thinking about "3 bits" and I nearly wrote the _exact_ same.) I did a bunch of manual tests just to make sure I wasn't making a subtly incorrect thing even subtly-er worse; for a bit I was outputting _absolute_ garbage due to signed integer behaviour <_>

kivikakk:
Thanks both!

whitequark:
Oh, apologies, this has completely slipped from my mind.

kivikakk:
Mine too ^^;

Issue #4560, opened 2024-08-22
Yosys is sensitive to the hash function


widlarizer:
### Version

All

### On which OS did this happen?

Linux

### Reproduction Steps

See #4559

### Expected Behavior

Tests pass, no matter the hash function

### Actual Behavior

`tests/techmap/dfflibmap.ys` fails when the hash function is changed

georgerennie:
Hunting this one down, ive minimized a reproducer to where yosys reads in the `dffsr` model in techmapping.

on the following script
```bash
read_verilog <<EOT
module dffsr(input CLK, D, CLEAR, PRESET, output reg Q);

always @(posedge CLK, posedge CLEAR, posedge PRESET)
	if (CLEAR)
		Q <= 0;
	else if (PRESET)
		Q <= 1;
	else
		Q <= D;

endmodule
EOT

proc -noopt

opt_expr -mux_bool
clean -purge
write_verilog -noattr
```

mainline yosys produces the following:
```verilog
assign _0_ = CLEAR ? 1'h0 : PRESET;
always @(posedge CLK, posedge _0_, posedge CLEAR)
  if (CLEAR) Q <= 1'b0;
  else if (_0_) Q <= 1'b1;
  else Q <= D;
```

whereas yosys with the new hash produces:
```verilog
assign _0_ = PRESET ? 1'h0 : CLEAR;
always @(posedge CLK, posedge PRESET, posedge _0_)
  if (_0_) Q <= 1'b0;
  else if (PRESET) Q <= 1'b1;
  else Q <= D;
```

In the first, clear takes priority over preset (as specified in the original verilog), whereas somehow in the version with the new hash the priority has been inverted and preset takes priority over clear. This change is introduced by `proc -noopt` (i added the further optimizations after just to make the verilog clearer but confirmed the behaviour appears before them).

georgerennie:
`proc_arst; proc_dff` seems sufficient.

From what I can tell, running `proc_arst` generates the following:
```
process $proc$<<EOT:3$1
  assign $0\Q[0:0] \Q
  assign $0\Q[0:0] \D
  sync high \CLEAR
    update \Q 1'0
  sync high \PRESET
    update \Q 1'1
  sync posedge \CLK
    update \Q $0\Q[0:0]
end
```

[...]

PR #4568, opened 2024-08-27
proc_dff: process sync rules in reverse input order

widlarizer:

It's assumed that sync rules are in some kind of priority ordering. See [here](https://github.com/YosysHQ/yosys/issues/4560#issuecomment-2312512074)

Since there is no "insertion ordered map" in STL, I provide a minimal one of my own

The tests now pass including dfflibmap.ys even when our hash function is modified as in #4559 and #4524. More investigation is needed about proc_dff and into this particular test and its correctness

### DIFF
--- a/passes/proc/proc_dff.cc
+++ b/passes/proc/proc_dff.cc
@@ -54,7 +54,7 @@ RTLIL::SigSpec find_any_lvalue(const RTLIL::Process *proc)
 }
 
 /**
- * Container storing key,std::set(value) pairs in insertion order
+ * Container storing key,std::set(value) pairs in reverse insertion order
  * and provides a linear search.
  */
 template<typename Key, typename Value>
@@ -71,8 +71,8 @@ public:
 			return backing.back().second;
 		}
 	}
-	typename std::vector<entry>::iterator begin() { return backing.begin(); }
-	typename std::vector<entry>::iterator end() { return backing.end(); }
+	typename std::vector<entry>::reverse_iterator begin() { return backing.rbegin(); }
+	typename std::vector<entry>::reverse_iterator end() { return backing.rend(); }
 	typename std::vector<entry>::size_type size() { return backing.size(); }
 	void clear() { return backing.clear(); }
 };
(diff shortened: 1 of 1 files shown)


whitequark:
I wonder if this should be clarified in the RTLIL guide in the manual.

widlarizer:
I think so, I checked `docs/source/yosys_internals/formats/rtlil_rep.rst` and it's not described there. I'd describe it if I had sufficient faith in my mental model.

Also as @georgerennie pointed out the reverse order would be the correct one, the CI only confirms this

georgerennie:
I discussed this with @widlarizer by dm, but I think this solution doesn't fully capture the priority for the sync rules. As the syncrules are still being aggregated by the assign expression for their condition, later syncrules automatically get the same priority as earlier syncrules that assign the same value, rather than their correct priority. I assume this is meant to be an optimization to merge together conditions leading to the same value being loaded, but some more care needs to be taken to preserve priorities.

Consider the following example
```
read_verilog <<EOT
module dffsr(input CLK, D, A, B, C, output reg Q);

always @(posedge CLK, posedge A, posedge B, posedge C)
	if (A)         Q <= '0;
	else if (B) Q <= '1;
	else if (C) Q <= '0;
	else          Q <= D;
endmodule
EOT

proc -noopt

opt -mux_bool
write_verilog -noattr
```
This produces
```verilog
assign _0_ = | { C, A };
assign _1_ = _0_ ? 1'h0 : B;
always @(posedge CLK, posedge _1_, posedge _0_)
  if (_0_) Q <= 1'b0;
  else if (_1_) Q <= 1'b1;
  else Q <= D;
```

The 0 case has taken the trigger `|{C, A}`, whereas really it should be something like `|{A, C && !B)`. This means that if the inputs `A = 0, B = 1, C = 1` are applied, in the first circuit `Q = 1` whereas in the processed version `Q = 0`

I think a possible solution is to process the sync rules one by one in the order they appear from `proc_arst`, guarding each condition with all the previous sync rule conditions, and taking the disjunction of conditions that lead to the same value.

With this, you would therefore get for each sync rule `A -> Q <= '0`, `B && !A -> Q <= '1`, `C && !B && !A -> Q <= '0`. You can then merge the conditions leading to zero, so you get `|{A, C && !B && !A} -> Q <= '0` and `B && !A -> Q <= '1`.

widlarizer:
Superseded by #4569

PR #4569, opened 2024-08-28
proc_dff: respect sync rule priorities when generating complex dffsrs

### LINKED ISSUE #4560: Yosys is sensitive to the hash function

georgerennie:
This fixes #4560, where previously the order that sync rules were processed in depended on the order they were pulled out of a std::map. This PR changes this to process them in the order they are found in, respecting the priorities among the async signals.

This is an alternative to #4568 that should fix the issues highlighted there (I've tested the example I left in the replies there and it seems to give the correct result now).

It still makes the optimizations made before, such as merging together control signals if all sync rules cause the DFF to take the same value.

- [x] Before merging this I want to add a few tests for this behaviour. edit: done


widlarizer (approved):
The first two cases in the test file fail on main, meaning the test covers the added functionality. This also fixes the test failure at #4524

PR #4576, opened 2024-09-03
write_xaiger: Get by without endianness helpers

povik:

A clean-up

Avoid any `bswap` helpers or the detection of host endianness when there's a portable way to write out a number in big endian: e.g. for 32-bit `x` you can do an equivalent of `put(x >> 24); put(x >> 16); put(x >> 8); put(x);`


widlarizer (approved):
Looks good

PR #4577, opened 2024-09-03
fsmdata: fix ctrl_out insertion

widlarizer:
Fix what seems to be a copy pasting error. This insert method's first arg is supposed to be an iterator into a valid position of this, but instead it's using an iterator into another vector. I went through all occurrences of `.insert\(.*,.*,.*` in the codebase and didn't find more of this kind of problem


KrystalDelusion:
Superseded by #4516

PR #4583, opened 2024-09-05
clockgate: centralize clock enables out of FFs

widlarizer:

According to several users, large power savings in ASICs can be achieved by centralizing clock gating per clock domain rather than routing CLK and EN into many FFs.

The new `clock_gate` command allows the user to specify a rising edge clock gating cell ("ICG") with `-pos` and a falling one with `-neg`. Within each selected module, flip flops driven with the same clock and enable signals of the same polarities are grouped. If this group has more flops than the threshold set by `-min_net_size N`, they are converted via `kernel/ff.h` to versions withouth clock enable, and corrsepondingly, clock gating cells are emitted to create one GCLK net per flop group. 

Since clock gates often have DFT ports, those are *assumed* to be required tied low. They can be listed with repeated `-tie_lo` pass args. 

It's also *assumed* that ICGs have active-high CE pins. This is based on the fact that there doesn't seem to be a provision in the liberty file format to describe that polarity either way. This pass may in the future be adapted to search for usable cells in a provided `.lib` file.

`clock_gate -pos pdk_icg ce:clkin:clkout -tie_lo scanen`

Run the included test which doesn't test anything but shows off the design. Use `-min_net_size` to show that you can eliminate small cases.

- [x] help string
- [x] proper test

### DIFF
--- a/passes/techmap/clockgate.cc
+++ b/passes/techmap/clockgate.cc
@@ -73,20 +73,20 @@ struct ClockgatePass : public Pass {
 	// if the number of FFs associated with it is sufficent
 	struct ClkNetInfo {
 		// Original, ungated clock into enabled FF
-		Wire* clk_net;
+		SigBit clk_bit;
 		// Original clock enable into enabled FF
-		Wire* ce_net;
+		SigBit ce_bit;
 		bool pol_clk;
 		bool pol_ce;
 		unsigned int hash() const {
-			auto t = std::make_tuple(clk_net, ce_net, pol_clk, pol_ce);
+			auto t = std::make_tuple(clk_bit, ce_bit, pol_clk, pol_ce);
 			unsigned int h = mkhash_init;
 			h = mkhash(h, hash_ops<decltype(t)>::hash(t));
 			return h;
 		}
 		bool operator==(const ClkNetInfo& other) const {
-			return (clk_net == other.clk_net) &&
-			       (ce_net == other.ce_net) &&
+			return (clk_bit == other.clk_bit) &&
+			       (ce_bit == other.ce_bit) &&
 			       (pol_clk == other.pol_clk) &&
 			       (pol_ce == other.pol_ce);
 		}
@@ -100,8 +100,8 @@ struct ClockgatePass : public Pass {
 	};
 
 	ClkNetInfo clk_info_from_ff(FfData& ff) {
-		Wire* clk = ff.sig_clk.as_wire();
-		Wire* ce = ff.sig_ce.as_wire();
+		SigBit clk = ff.sig_clk[0];
+		SigBit ce = ff.sig_ce[0];
 		ClkNetInfo info{clk, ce, ff.pol_clk, ff.pol_ce};
 		return info;
 	}
@@ -147,7 +147,12 @@ struct ClockgatePass : public Pass {
 
 				FfData ff(nullptr, cell);
 				// It would be odd to get constants, but we better handle it
-				if (ff.has_ce && ff.sig_clk.is_wire() && ff.sig_ce.is_wire()) {
+				if (ff.has_ce) {
+					if (!ff.sig_clk.is_bit() || !ff.sig_ce.is_bit())
+						continue;
+					if (!ff.sig_clk[0].is_wire() || !ff.sig_ce[0].is_wire())
+						continue;
+
 					ce_ffs.insert(cell);
 
 					ClkNetInfo info = clk_info_from_ff(ff);
@@ -159,7 +164,6 @@ struct ClockgatePass : public Pass {
 			}
 
 			for (auto& clk_net : clk_nets) {
-				log_debug("checking clk net %s\n", clk_net.first.clk_net->name.c_str());
 				auto& clk = clk_net.first;
 				auto& gclk = clk_net.second;
 
@@ -176,10 +180,9 @@ struct ClockgatePass : public Pass {
 				if (!matching_icg_desc)
 					continue;
 
-				log_debug("building ICG for clk net %s\n", clk_net.first.clk_net->name.c_str());
 				Cell* icg = module->addCell(NEW_ID, matching_icg_desc->name);
-				icg->setPort(matching_icg_desc->ce_pin, clk.ce_net);
-				icg->setPort(matching_icg_desc->clk_in_pin, clk.clk_net);
+				icg->setPort(matching_icg_desc->ce_pin, clk.ce_bit);
+				icg->setPort(matching_icg_desc->clk_in_pin, clk.clk_bit);
 				gclk.new_net = module->addWire(NEW_ID);
 				icg->setPort(matching_icg_desc->clk_out_pin, gclk.new_net);
 				// Tie low DFT ports like scan chain enable
@@ -187,7 +190,7 @@ struct ClockgatePass : public Pass {
 					icg->setPort(port, Const(0));
 				// Fix CE polarity if needed
 				if (!clk.pol_ce) {
-					SigBit ce_fixed_pol = module->NotGate(NEW_ID, clk.ce_net);
+					SigBit ce_fixed_pol = module->NotGate(NEW_ID, clk.ce_bit);
 					icg->setPort(matching_icg_desc->ce_pin, ce_fixed_pol);
 				}
 			}
--- a/passes/techmap/clockgate.cc
+++ b/passes/techmap/clockgate.cc
@@ -44,10 +44,9 @@ struct ClockgatePass : public Pass {
 		log("\n");
 		log("    clockgate [options] [selection]\n");
 		log("\n");
-		log("Creates gated clock nets for sets of FFs with clock enable\n");
-		log("sharing a clock and replaces the FFs with versions without\n");
-		log("clock enable inputs. Intended to reduce power consumption\n");
-		log("in ASIC designs.\n");
+		log("This pass transforms each set of FFs sharing the same clock and\n");
+		log("enable signal into a clock-gating cell and a set of enable-less FFs.\n");
+		log("Primarily a power-saving transformation on ASIC designs.\n");
 		log("\n");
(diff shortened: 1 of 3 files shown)


passes/techmap/clock_gate.cc
```diff
@@ -0,0 +1,205 @@
+#include "kernel/yosys.h"
+#include "kernel/ff.h"
+#include <optional>
+
+USING_YOSYS_NAMESPACE
+PRIVATE_NAMESPACE_BEGIN
+
+struct ClockGateCell {
+	std::string name;
```
povik, line 9: We can use `IdString` for all the fields here

passes/techmap/clock_gate.cc
```diff
@@ -0,0 +1,205 @@
...
+	std::string ce_pin;
+	std::string clk_in_pin;
+	std::string clk_out_pin;
+
+	ClockGateCell(std::string& name, std::string& str) : name(RTLIL::escape_id(name)) {
+		char delimiter = ':';
+		size_t pos1 = str.find(delimiter);
+		if (pos1 == std::string::npos)
+			log_assert(false && "Not enough ports in descriptor string");
```
povik, line 18: It would be nice to have those be `log_cmd_error` as long as we are careful where from this constructor is called. I guess we should pull the parsing into a method of its own when it calls `log_cmd_error`.

passes/techmap/clock_gate.cc
```diff
@@ -0,0 +1,205 @@
...
+	// One ICG will be generated per ClkNetInfo
+	// if the number of FFs associated with it is sufficent
+	struct ClkNetInfo {
+		// Original, ungated clock into enabled FF
+		Wire* clk_net;
+		// Original clock enable into enabled FF
+		Wire* ce_net;
+		bool pol_clk;
+		bool pol_ce;
+		unsigned int hash() const {
```
povik, line 55: Can we build a tuple and defer to the implicit hashing on that?

passes/techmap/clock_gate.cc
widlarizer: to avoid reconstructing them every time? True I didn't consider that and only wrote what's most ergonomical

passes/techmap/clock_gate.cc
```diff
@@ -0,0 +1,205 @@
...
+
+		int gated_flop_count = 0;
+		for (auto module : design->selected_whole_modules()) {
+			sigmap.set(module);
+			initvals.set(&sigmap, module);
+			for (auto cell : module->cells()) {
+				if (!RTLIL::builtin_ff_cell_types().count(cell->type))
+					continue;
+
+				FfData ff(&initvals, cell);
```
povik, line 126: I think we can pass `nullptr` for initvals when we are not interested in it (which we shouldn't be here)

passes/techmap/clock_gate.cc
```diff
@@ -0,0 +1,205 @@
...
+				if (ff.has_ce) {
+					ce_ffs.insert(cell);
+
+					ClkNetInfo info = clk_info_from_ff(ff);
+					auto it = clk_nets.find(info);
```
povik, line 131: An alternative: two dicts instead of `GClkNetInfo`, that would let us simplify the code here to `net_sizes[info]++;`

passes/techmap/clock_gate.cc
```diff
@@ -0,0 +1,205 @@
...
+	struct GClkNetInfo {
+		// How many CE FFs on this CLK net have we seen?
+		int net_size;
+		// After ICG generation, we have new gated CLK signals
+		Wire* new_net;
+	};
+
+	ClkNetInfo clk_info_from_ff(FfData& ff) {
+		Wire* clk = ff.sig_clk.as_wire();
+		Wire* ce = ff.sig_ce.as_wire();
```
povik, line 80: There's no guarantee these two are single-bit wires (when not the `as_wire()` conversion will crash). The solution is for the `ClkNetInfo` fields to be sigbits

passes/techmap/clock_gate.cc
widlarizer: That seems like the same amount of boilerplate but more construction of hashlib-internal dynamic arrays basically, is it worth it?

passes/techmap/clock_gate.cc
povik: Not really, it's just slightly more yosys-idiomatic

passes/techmap/clock_gate.cc
widlarizer: I considered this but forgot to put down a TODO huh

passes/techmap/clock_gate.cc
povik: Not sure, go with what you think is right

passes/techmap/clock_gate.cc
```diff
@@ -0,0 +1,205 @@
...
+		clk_in_pin = str.substr(pos1 + 1, pos2 - (pos1 + 1));
+		clk_in_pin = RTLIL::escape_id(clk_in_pin);
+
+		clk_out_pin = str.substr(pos2 + 1, str.size() - (pos2 + 1));
+		clk_out_pin = RTLIL::escape_id(clk_out_pin);
+	}
+};
+
+struct ClockgatePass : public Pass {
+	ClockgatePass() : Pass("clock_gate", "extract clock gating out of flip flops") { }
```
povik, line 38: Let's rename this to `clockgate`: if we look at the list of Yosys commands, underscores are used to group commands together but not really as a word separator

mole99:
Hi @widlarizer, this looks very interesting! Thanks for your work 🙌
Does this also work with FFs that don't have a CE input, but rather a MUX2 on the D input?
That would be similar in functionality to what Lighter currently does and could potentially replace it: https://github.com/AUCOHL/Lighter

povik:
> Does this also work with FFs that don't have a CE input, but rather a MUX2 on the D input?

Those should be converted to an FF with a clock enable input by the application of `opt_dff` earlier in the flow.

mole99:
That's great, thanks!

povik:
Sure thing!

passes/techmap/clock_gate.cc
widlarizer: I don't like the result:
```cpp
auto t = std::make_tuple(clk_net, ce_net, pol_clk, pol_ce);
unsigned int h = mkhash_init;
h = mkhash(h, hash_ops<std::tuple<Yosys::RTLIL::Wire *, Yosys::RTLIL::Wire *, bool, bool>>::hash(t));
return h;
```

widlarizer: I forgot I can just `h = mkhash(h, hash_ops<decltype(t)>::hash(t));`

povik: > hash_ops<decltype(t)>::hash(t)

Or just use the `mkhash` helper (there are two functions named `mkhash`)

I assume you are shaping this into a hash-in hash-out pattern for #4524, otherwise there's no reason to return the tuple hash directly.

passes/techmap/clock_gate.cc
widlarizer: Since there is one set of information per unique clknetinfo, I'd like to keep it in one record

passes/techmap/clock_gate.cc
widlarizer: Yep this is temporary

passes/techmap/clockgate.cc
```diff
@@ -0,0 +1,230 @@
...
+		for (auto module : design->selected_whole_modules()) {
+			sigmap.set(module);
+			initvals.set(&sigmap, module);
+			for (auto cell : module->cells()) {
+				if (!RTLIL::builtin_ff_cell_types().count(cell->type))
+					continue;
+
+				FfData ff(&initvals, cell);
+				// It would be odd to get constants, but we better handle it
+				if (ff.has_ce && ff.sig_clk.is_wire() && ff.sig_ce.is_wire()) {
```
povik, line 155: This disqualifies cases where the enable signal is a single bit taken out of a multi-bit wire. See what `is_wire()` checks for

widlarizer:
Related work:

- [Lighter](https://github.com/AUCOHL/Lighter) - Lighter is very simple. Using it for a different PDK requires rewriting the verilog techmapping file. It indiscriminately extracts clock gating out of each clock enabled flop. I assume it then deduplicates them with some opt command, but this can worsen area if you have many enable and clk pairs that are each only used by one FF cell with clock enable. This pass at the moment needs PDK-specific flags, which is annoying. It can be extended in the future to find correct ICGs in a Liberty file.
- [IMMS clock gating with yosys](https://fpga-ignite.github.io/presentations-pdf/presentation14.pdf) - this is closed source as far as I can tell

passes/techmap/clockgate.cc
```diff
@@ -0,0 +1,228 @@
...
+	ClockgatePass() : Pass("clockgate", "extract clock gating out of flip flops") { }
+	void help() override {
+		//   |---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|
+		log("\n");
+		log("    clockgate [options] [selection]\n");
+		log("\n");
+		log("Creates gated clock nets for sets of FFs with clock enable\n");
+		log("sharing a clock and replaces the FFs with versions without\n");
+		log("clock enable inputs. Intended to reduce power consumption\n");
+		log("in ASIC designs.\n");
```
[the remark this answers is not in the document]
widlarizer: It does many sets at once, typically. Hence the plural, even if it's awkward, is more correct. It's good that you found a way to add mentioning the ICG, I'll def adapt that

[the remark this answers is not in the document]
povik: In that case let me suggest "This pass transforms each set of FFs"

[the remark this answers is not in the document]
widlarizer: Adapted

PR #4588, opened 2024-09-08
Merge of function black boxing

kendroe:

I added a new feature to allow Verilog functions to be black boxed.  In the generated smt2 file, an euf symbol is used instead
of expanding the function definition.  Check out my Linkedin blog here https://www.linkedin.com/feed/update/urn:li:activity:7191433053066448896/.  This blog links to earlier blogs describing the changes.

Many changes to code generation.

I expect to do a lot of cleanup before these changes are accepted.

There are lots of debugging prints that should be removed.  It would be good for someone with lots of experience with the internals of Yosys to review.


kernel/rtlil.h
```diff
@@ -1501,6 +1509,31 @@ struct RTLIL::Module : public RTLIL::AttrObject
...
+	~Function();
+
+public:
+	// do not simply copy wires
+	Function(RTLIL::Function &other) = delete;
+	void operator=(RTLIL::Function &other) = delete;
+
+	RTLIL::Module *module;
+	RTLIL::IdString name;
+	AST::AstNode *def;
```
povik, line 1530: As far as I can tell this new `Function` object is here so that the ASTs of functions can be looked up. What is the AST used for?

This is adding a new type of member to modules (next to wires, cells, processes and memories). This means these now have to be considered all over the Yosys codebase anywhere we are manipulating modules and they should be saved/restored on writing out an RTLIL file.

That's a substantial addition in complexity to serve one feature. Preferably we encode what we need in parameters of the new cell type `$fun` instead (if we need a new cell type in the first place).

povik (commented):
Indeed there a lot of left-over debug prints, but let's focus on the fundamental approach first. I have only skimmed the changes, but is the basic idea that function calls are deferred, that is they are converted to `$fun` cells in the frontend, and the user has the option to either blackbox those or replace them with the function's body later on?

jix:
As explained in the review, making changes to core RTLIL data structures is not the way to extend yosys with functionality like this.

PR #4589, opened 2024-09-08
smt2: fix bitwuzla invocation

mattyoung101:

Bitwuzla no longer shares options in common with Boolector. It now requires "--lang smt2" instead of "--smt2", and always runs incrementally by default (see: [Bitwuzla news](https://github.com/bitwuzla/bitwuzla/blob/69e4be28d86ef9e0cebee2e846378e56b6da978a/NEWS.md?plain=1#L163))

I found this because I'm using Bitwuzla 0.5.0-dev-main@69e4be28-dirty (through the AUR `bitwuzla-git` package), and it currently does not run correctly through eqy. 

`smtio.py` is modified to invoke Bitwuzla correctly, taking into account the fact it uses incremental solving by default 

This should be testable by using any sby tools with Bitwuzla. I was using eqy and confirmed it worked with a very small testbench of mine (it's a bit hard to share because it's a Yosys plugin that generates the design on the fly).

### DIFF
--- a/backends/smt2/smtio.py
+++ b/backends/smt2/smtio.py
@@ -226,12 +226,21 @@ class SmtIo:
                 print('timeout option is not supported for mathsat.')
                 sys.exit(1)
 
-        if self.solver in ["boolector", "bitwuzla"]:
+        if self.solver == "boolector":
             if self.noincr:
                 self.popen_vargs = [self.solver, '--smt2'] + self.solver_opts
             else:
                 self.popen_vargs = [self.solver, '--smt2', '-i'] + self.solver_opts
             self.unroll = True
+            if self.timeout != 0:
+                print('timeout option is not supported for %s.' % self.solver)
+                sys.exit(1)
+
+        if self.solver == "bitwuzla":
+            self.popen_vargs = [self.solver, '--lang', 'smt2'] + self.solver_opts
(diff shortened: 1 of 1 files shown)


donn:
See https://github.com/YosysHQ/oss-cad-suite-build/issues/87#issuecomment-1927552563

mattyoung101:
Thanks for the link, unfortunately missed it when searching the issue tracker before making this PR. Do we know if the upstream Bitwuzla performance regression has been fixed? As I understand it, that was the main reason blocking oss-cad-suite from upgrading to a newer Bitwuzla?

donn:
Honestly don't know, you and I are in the same boat- you could try running @jix's benchmark (it's downloadable)

KrystalDelusion:
I clicked the button to run CI, although I suspect that the [test-verific](https://github.com/YosysHQ/yosys/actions/runs/10761115033) run which does run SBY tests will either fail because it's using the older version of Bitwuzla used in OSS CAD suite, or just pass without testing Bitwuzla specifically 😅

mattyoung101:
@KrystalDelusion Looks like the CI pipeline has passed, although that probably does mean that Bitwuzla is not tested directly by the Verific tests 😅

PR #4595, opened 2024-09-11
internal_stats: astnode (sizeof)

widlarizer:

Requested by @povik. We'd like to see how much of the memory usage of Yosys between passes is directly due to `AstNode`s. This doesn't count overhead of initialized vectors and such.

Adding a counter adjusted on `AstNode` construction/destruction

`yosys -p "read_verilog tests/sim/dffe.v; internal_stats -json"`


povik (approved):

Ravenslofty:
This should have used the scratchpad...

povik:
Wouldn’t that be inefficient accessing the scratchpad with every construction/destruction?

Ravenslofty:
I'm not proposing `astnodes` is a scratchpad variable, I'm proposing that `astnodes_count()` is a scratchpad variable.

povik:
But scratchpad is a static key-value store. We would need to periodically update it, or extend what scratchpad is.

In any case we can go back and redo this.

PR #4596, opened 2024-09-11
reduce OS ifdefs with OS_PATH_SEP

widlarizer:
See https://github.com/YosysHQ/yosys/pull/4592#issuecomment-2344397223

### DIFF
--- a/backends/cxxrtl/cxxrtl_backend.cc
+++ b/backends/cxxrtl/cxxrtl_backend.cc
@@ -639,12 +639,7 @@ std::string escape_cxx_string(const std::string &input)
 
 std::string basename(const std::string &filepath)
 {
-#ifdef _WIN32
-	const std::string dir_seps = "\\/";
-#else
-	const std::string dir_seps = "/";
-#endif
-	size_t sep_pos = filepath.find_last_of(dir_seps);
+	size_t sep_pos = filepath.find_last_of(OS_PATH_SEP);
 	if (sep_pos != std::string::npos)
 		return filepath.substr(sep_pos + 1);
 	else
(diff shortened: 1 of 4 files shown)


whitequark (changes requested):
I liked the idea of factoring out the code rather than the constants a lot more.

kernel/yosys.cc
```diff
@@ -540,6 +540,25 @@ std::string escape_filename_spaces(const std::string& filename)
...
+	size_t sep_pos = path.find_last_of(OS_PATH_SEP);
+	if (sep_pos != std::string::npos)
+		return path.substr(sep_pos + 1);
+	else
+		return path;
+}
+
+// Includes OS_PATH_SEP at the end
+std::string parent_from_file_path(std::string path) {
+	return path.substr(0, path.find_last_of(OS_PATH_SEP)+1);
```
whitequark, line 559: What happens if `path.find_last_of(OS_PATH_SEP)` returns `std::string::npos`?

whitequark: To me it looks like it'll return `""`, which does not include `OS_PATH_SEP` at the end. This either needs a documentation change or an assertion.

KrystalDelusion:
Is there a reason not to use `std::filesystem::path`?  It already provides `.filename()` and `.parent_path()`

phsauter:
> Is there a reason not to use `std::filesystem::path`? It already provides `.filename()` and `.parent_path()`

From #4592:
> Note that WASI does support the filesystem (otherwise Yosys and nextpnr could not work at all), but doesn't support std::filesystem specifically for reasons related to threading.

whitequark:
I've actually fixed the threading issue with the latest WASI SDK release, so you should be able to use `std::filesystem` now if you upgrade the SDK.

KrystalDelusion (approved):
Tested local_include.sh under yowasp-yosys and that is working as expected (pre-PR (circa Yosys 0.56) and with PR).

whitequark:
Thanks!

PR #4605, opened 2024-09-18
read_liberty: Optionally import unit delay arcs

povik:

Avoiding timing path segmentation on explicitly instantiated or pre-mapped standard cells. When the SC mapper is configured for the unit delay model, with this new option we can supply complementary arcs for those cells which are viewed as boxes

Under each cell and pin, we look at `timing()` sections with `timing_type` set to `combinational`, and for each arc found, we emit a `$specify2` cell with fixed delay of 1000 which matches the unit delay inside ABC.

### DIFF
--- a/tests/liberty/unit_delay.ys
+++ b/tests/liberty/unit_delay.ys
@@ -1,0 +1,3 @@
+# Nothing gets imported: the file lacks timing data
+read_liberty -wb -unit_delay normal.lib
+select -assert-none =*/t:$specify*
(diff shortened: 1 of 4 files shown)


tests/liberty/unit_delay.ys
```diff
@@ -0,0 +1,3 @@
+# Nothing gets imported: the file lacks timing data
+read_liberty -wb -unit_delay normal.lib
+select -assert-none =*/t:$specify*
```
povik, line 3: This is an almost empty test since we don't have any `.lib` files checked in with populated timing sections. We can find a new one provided the licensing is ok

PR #4607, opened 2024-09-19
quicklogic: Avoid carry chains in division mapping

povik:

The default mapping rules for division-like operations (div/divfloor/ mod/modfloor) invoke subtractions which can get mapped to carry chains in FPGA flows. Optimizations across carry chains are weak, so in practice this ends up too costly compared to implementing the division purely in soft logic.

For this reason arrange for `techmap.v` ignoring division operations under `-D NODIV`, and use this mode in `synth_quicklogic` to avoid carry chains for divisions.

### DIFF
--- a/techlibs/common/techmap.v
+++ b/techlibs/common/techmap.v
@@ -304,6 +304,7 @@ endmodule
 // Divide and Modulo
 // --------------------------------------------------------
 
+`ifndef NODIV
 module \$__div_mod_u (A, B, Y, R);
 	parameter WIDTH = 1;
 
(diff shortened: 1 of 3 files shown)


povik:
```
module top(input wire [10:0] a, input wire [10:0] b, output wire [10:0] y);
	assign y = a / b;
endmodule
```
before
```
   Number of cells:                704
     $lut                          308
     adder_carry                   396
```
after
```
   Number of cells:                125
     $lut                          125
```

nakengelhardt (approved):

povik:
When we merge this we should make a note to carry over (no pun intended) the change across all FPGA architectures provided we confirm this doesn't make the delay worse. I am not sure we should focus on trying to come up with a better architecture for division as it doesn't seem like an RTL operator in common use.

Ravenslofty:
Perhaps include a testcase; something like the synthesis tests which check the number and type of cells.

PR #4608, opened 2024-09-19
rtlil: add Const::compress helper function

phsauter:
If one has a Sigspec that is fully constant it is currently non-trivial to get its value.  
The main blocking point here is that the SigSpec may be wider than 32bit (an integer).  
However, this doesn't mean the value stored in it is necessarily this large/small.

Eg something like `33'd4` or `33'd-2` cannot be easily converted to an int.

This adds a `compress(is_signed)` function.  
It tries to minimize the number of bits used for the representation of the number. 
It does so by removing the leading bits while taking signedness into consideration.

With this function, the above can be successfully converted to an integer by first calling compress.

Edit:
Here is a possible application.
https://github.com/povik/yosys-slang/blob/28cea56840a6bc8a4b6398d36609500d5960ad79/src/builder.cc#L134  
With `compress` it is easy to check if the actual value has a larger representation than 24-bits instead of checking the SigSpec, which again may not use the minimum width.  
I am fairly convinced there are places in the Yosys codebase itself this would be helpful, finding them is a bit difficult.

### DIFF
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -286,20 +286,22 @@ void RTLIL::Const::compress(bool is_signed)
 
 	// back to front (MSB to LSB)
 	RTLIL::State leading_bit;
-	if(is_signed)
+	if (is_signed)
 		leading_bit = (bits.back() == RTLIL::State::Sx) ? RTLIL::State::S0 : bits.back();
 	else
 		leading_bit = RTLIL::State::S0;
 
-    size_t idx = bits.size();
-    while (idx > 0 && bits[idx -1] == leading_bit) {
-        --idx;
-    }
+	size_t idx = bits.size();
+	while (idx > 0 && bits[idx -1] == leading_bit) {
+		idx--;
+	}
 
 	// signed needs one leading bit
 	if (is_signed && idx < bits.size()) {
-        ++idx;
-    }
+		idx++;
+	}
+	// must be at least one bit
+	idx = (idx == 0) ? 1 : idx;
 
 	bits.erase(bits.begin() + idx, bits.end());
 }
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -280,6 +280,30 @@ int RTLIL::Const::as_int(bool is_signed) const
 	return ret;
 }
 
+void RTLIL::Const::compress(bool is_signed)
+{
+	if (bits.empty()) return;
+
+	// back to front (MSB to LSB)
+	RTLIL::State leading_bit;
+	if(is_signed)
+		leading_bit = (bits.back() == RTLIL::State::Sx) ? RTLIL::State::S0 : bits.back();
+	else
+		leading_bit = RTLIL::State::S0;
+
+    size_t idx = bits.size();
+    while (idx > 0 && bits[idx -1] == leading_bit) {
+        --idx;
+    }
+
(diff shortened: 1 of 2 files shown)


kernel/rtlil.cc
```diff
@@ -280,6 +280,30 @@ int RTLIL::Const::as_int(bool is_signed) const
...
+	else
+		leading_bit = RTLIL::State::S0;
+
+    size_t idx = bits.size();
+    while (idx > 0 && bits[idx -1] == leading_bit) {
+        --idx;
+    }
+
+	// signed needs one leading bit
+	if (is_signed && idx < bits.size()) {
```
povik, line 300: I think this check needs to be `idx + 1 < bits.size()`, take e.g. `4'bxxxx`

kernel/rtlil.cc
```diff
@@ -280,6 +280,30 @@ int RTLIL::Const::as_int(bool is_signed) const
...
+	if (bits.empty()) return;
+
+	// back to front (MSB to LSB)
+	RTLIL::State leading_bit;
+	if(is_signed)
+		leading_bit = (bits.back() == RTLIL::State::Sx) ? RTLIL::State::S0 : bits.back();
+	else
+		leading_bit = RTLIL::State::S0;
+
+    size_t idx = bits.size();
```
povik, line 294: Nit: please make the whitespacing consistent

povik (changes requested):
Thanks! Some details need fixing

kernel/rtlil.cc
povik: Looking at it with fresh eyes the check looks ok, nevermind

phsauter: In the case of `4'bxxxx`:  
Starts with `idx=4`, the bit `bits[4-1]` does not match the leading bit (which is never `X`).  
So we have `idx=4` which means `idx < bits.size()` is false and then it erases the vector from end to end which should do nothing.

But the case with unsigned all zeros is actually a problem because it will create a `idx=0` at the end which would delete the entire vector. -> fixed

povik (approved):

kernel/rtlil.cc
```diff
@@ -280,6 +280,32 @@ int RTLIL::Const::as_int(bool is_signed) const
 	return ret;
 }
 
+void RTLIL::Const::compress(bool is_signed)
```
widlarizer, line 283: What's the motivation for not respecting/setting `flags 
& CONST_FLAG_SIGNED` ?

povik: I am not sure we want to defer to that. That flag isn't set to a useful value in all contexts, especially when you convert a `SigSpec` with `as_const()`, which is the use case which prompted writing the helper.

phsauter:
Should I also add a `.as_int_compress(bool is_signed)` to make it even easier to get a constant integer?

povik:
For that one to be useful you need to know when you can trust the result and when not, so maybe it should return a pair of `int, bool`. If it can fail, then we may call it `try_int`.

phsauter:
If it is larger than 32-bit then it fails as it is not representable but yes, std::optional is a good option there.

phsauter:
Does this solution look reasonable?

widlarizer:
As far as I can tell, it does

PR #4613, opened 2024-09-24
log: Never silence `log_cmd_error`

povik:

Some errors can get hidden by `-v N`, see The-OpenROAD-Project/OpenROAD-flow-scripts#2375

This makes the user almost always worse off. The only exception I can think of is someone scripting around Yosys commands, checking for their exit status, and handling failures gracefully (for the few failure modes where that can be done). I've never seen such a Yosys script.

Add extra handling to arrange for `log_cmd_error` never being silenced by the command line `-v N` option. Similar path for `log_error` exists already.

With test.tcl:

```
yosys -import
hierarchy -top nonexistent
```

Before

```
$ ./yosys -v 3 test.tcl
1. Executing HIERARCHY pass (managing design hierarchy).
ERROR: TCL interpreter returned an error: Yosys command produced an error
```

After

```
$ ./yosys -v 3 test.tcl
1. Executing HIERARCHY pass (managing design hierarchy).
ERROR: Module `nonexistent' not found!
ERROR: TCL interpreter returned an error: Yosys command produced an error
```

### DIFF
--- a/kernel/log.cc
+++ b/kernel/log.cc
@@ -459,8 +459,21 @@ void log_cmd_error(const char *format, ...)
 
 	if (log_cmd_error_throw) {
 		log_last_error = vstringf(format, ap);
+
+		// Make sure the error message gets through any selective silencing
+		// of log output
+		bool pop_errfile = false;
+		if (log_errfile != NULL) {
+			log_files.push_back(log_errfile);
+			pop_errfile = true;
+		}
+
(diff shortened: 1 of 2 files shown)


Ravenslofty:
I think this should have a testcase, but I'm not entirely sure how it would be implemented; maybe with a custom `run_test.sh`?

Issue #4610, opened 2024-09-24
Yosys Synthesis Fails with std::out_of_range Error in OPT_DEMORGAN Pass


PerryLogic:
### Version

yosys 0.41+126

### On which OS did this happen?

Linux

### Reproduction Steps

Hello,

I encountered a crash issue while using Yosys to synthesize a Verilog file. The specific details are as follows:

When running Yosys on the attached Verilog file (rtl.v), the synthesis process crashes during the OPT_DEMORGAN pass. The error message indicates a std::out_of_range exception, which occurs when the tool tries to inspect a $reduce_or cell and push inverters through reduction.

To rule out issues with the file itself, I have checked the Verilog file for syntax errors but did not find any obvious problems. Additionally, I am using the following version of Yosys:Yosys 0.41+126 (git sha1 855ac285f49, g++ 11.4.0-1ubuntu1~22.04 -fPIC -Os)

Here is the synthesis command I used:
yosys -p "
read_verilog rtl.v
hierarchy; proc; opt_clean -purge; opt_expr -mux_bool; opt_dff -nodffe; opt_reduce -full; opt_demorgan
write_verilog syn_yosys.v"

Here is a snippet of the error log:
![image](https://github.com/user-attachments/assets/7779ebb8-0741-4d6e-a0ec-ac0037487ce0)
Attached is the Verilog file (rtl.v) that triggers this issue. I hope to get the community's help and attention to resolve this bug.
[yosys_project.zip](https://github.com/user-attachments/files/17108935/yosys_project.zip)

### Expected Behavior

Yosys completes the synthesis without throwing any exceptions.

### Actual Behavior

Yosys with a std::out_of_range error during the OPT_DEMORGAN pass.

Ravenslofty:
`opt_lut_ins` doesn't do anything here, they can be removed.

LoSyTe:
> `opt_lut_ins` doesn't do anything here, they can be removed.

Thank you for your reply! I have removed the opt_lut_ins pass, but the issue still persists. The current pass sequence is:hierarchy; proc; opt_clean; opt_dff; opt_demorgan; opt_merge; opt_reduce; opt_dff; opt_expr; opt_muxtree; opt_share; opt_dff; opt_clean; opt_reduce; opt_demorgan;

georgerennie:
```
read_ilang <<EOT
autoidx 1
module \top
  wire output 1 \Y
  cell $reduce_or $reduce_or$rtl.v:29$20
[...]

PR #4614, opened 2024-09-25
opt_reduce: keep at least one input to $reduce_or/and cells

georgerennie:
When all inputs to a `$reduce_or` (resp. `$reduce_and`) are `0` (resp `1`), `opt_reduce` would remove them all leaving a zero width input to the cell. This can sometimes be poorly handled by other passes as in #4610, so this PR checks for that case and drives the cell by a constant.

I haven't chosen to fully replace the cell as the logic in that function would turn e.g. `|{a, 1'b1, 1'b0}` into `|{1'b1}` rather than `1'b1`, so this is in keeping for the `1'b0` case. `opt_expr` easily eliminates this cell due to the constant input.

### DIFF
--- a/passes/opt/opt_reduce.cc
+++ b/passes/opt/opt_reduce.cc
@@ -89,6 +89,9 @@ struct OptReduceWorker
 		RTLIL::SigSpec new_sig_a(new_sig_a_bits);
 		new_sig_a.sort_and_unify();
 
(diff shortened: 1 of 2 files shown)


widlarizer (approved):
Test reproduces error described in issue, fix looks good

Issue #4580, opened 2024-09-04
Get a proper command line parser


widlarizer:
### Feature Description

Currently, we use a homemade `getopt` in `kernel/driver.cc`. We have lots of single letter options so to add a feature I want a full word option lke `--option=value`, which we don't support currently, which is probably why we have many single letter options in the first place.

Let's move to a boring, normal solution like [cxxopts](https://github.com/jarro2783/cxxopts) (C++11) or [argparse](https://github.com/p-ranav/argparse) (C++17). Our current demands will be met by just about anything, so we have to think about what we want from command line argument parsing in the future to pick something

widlarizer:
This will have to respect https://github.com/YosysHQ/yosys/pull/3973 and other possible quirks

Apteryks:
Currently cxxopts provided by the system cannot be easily used because of strange include paths:
```
[  2%] Building kernel/celledges.o
kernel/driver.cc:22:10: fatal error: libs/cxxopts/include/cxxopts.hpp: No such file or directory
   22 | #include "libs/cxxopts/include/cxxopts.hpp"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
make: *** [Makefile:758: kernel/driver.o] Error 1
make: *** Waiting for unfinished jobs....
```

I guess this is bundled as a submodule? It'd be cleaner if the 3rdparty stuff was kept outside the sources and the build system considered their location, to keep the normal 
```
#include <cxxopts.hpp>
```

In the source, which would work with a system provided cxxopts library.

widlarizer:
Yes, it's a submodule. Here's why:

- we would lose control over whatever cxxopts the user is going to get from their package manager
- therefore yosys of the same version may then behave in various ways
- it would also be more difficult for users to set up a build environment
- cxxopts is a single header file, there's miniscule or negative disk space savings on relying on system package managers
- build systems with package management support, such as cmake, are an alternative to system packages. Yosys isn't a cmake project
- languages with integrated build systems with package management support, such as cargo for rust, are another one. Yosys is written in C++

If you're encountering this error, is this a result of deliberately desiring to use a system packaged cxxopts? It's not the only submodule. Submodules aren't generally a roadblack for building yosys for package distribution based on what I've seen on AUR and nixpkgs. If you're building yosys from the [release](https://github.com/YosysHQ/yosys/releases) sources, the yosys.tar.gz file actually contains all submodules properly vendored.

[...]

PR #4616, opened 2024-09-26
driver: replace getopt with cxxopts, replace -B, clean up help

widlarizer:
Implements #4580.

In this PR I'm adding the [cxxopts](https://github.com/jarro2783/cxxopts) header-only library as a submodule. This allows me to introduce long option aliases such as `--quiet` for `-q`. Since `-B` was undocummented and is assumed not broadly used, I'm replacing it with `--perffile`. Behaviour of `--` is retained as implemented in #3973. The behavior of `yosys -help` being special cased not to be equivalent to `yosys -h -e -l -p` is also retained. Help string indentation isn't retained since we're letting cxxopts generate it. Argument names like `logfile` are converted for consistency into `<logfile>` which is then used as is broadly the POSIX/GNU convention. Long options don't have to match the first letter of the short option if it's more logical that way, for example, `-m` is `--plugin`, I also changed the help string to `load the specified <plugin> module` since it was colliding with the concept of the HDL top module. The HDL top module selection option is now `-r` and `--top` and its help is `elaborate the specified HDL <top> module`. There's more minor differences in wording

The error reporting the usage of an unimplemented option such as `-z` now mentions that option. Before: `Run 'yosys -h' for help.` After:

```
Error parsing options: Option 'z' does not exist
Run 'yosys --help' for help.
```

From this PR onward, it should  decision whether a flag gets to hog the top level namespace since we now have an alternative. On the other hand, it should make it easier to experiment by adding long-only options.

- [ ] Manually test on Windows

### DIFF
--- a/kernel/driver.cc
+++ b/kernel/driver.cc
@@ -19,6 +19,8 @@
 
 #include "kernel/yosys.h"
 #include "libs/sha1/sha1.h"
+#include "libs/cxxopts/include/cxxopts.hpp"
+#include <iostream>
 
 #ifdef YOSYS_ENABLE_READLINE
 #  include <readline/readline.h>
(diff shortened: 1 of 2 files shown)


kernel/driver.cc
```diff
@@ -250,292 +204,231 @@ int main(int argc, char **argv)
...
+		("v,verbose", "print log headers up to <level> to the console. " \
+                      "Implies -q for everything except the 'End of script.' message.",
+			cxxopts::value<int>(), "<level>")
+		("t,timestamp", "annotate all log messages with a time stamp")
+		("d,detailed-timing", "print more detailed timing stats at exit")
+		("l,logfile", "write log messages to <logfile>",
+			cxxopts::value<std::vector<std::string>>(), "<logfile>")
+		("L,line-buffered-logfile", "like -l but open <logfile> in line buffered mode",
+			cxxopts::value<std::vector<std::string>>(), "<logfile>")
+		("o,outfile", "write the design to <outfile> on exit",
```
povik, line 250: I think this one and the next one doesn't belong to logging

kernel/driver.cc
```diff
@@ -250,292 +204,231 @@ int main(int argc, char **argv)
...
-		printf("commands in a script file instead of specifying input and output files on the\n");
-		printf("command line.\n");
-		printf("\n");
-		printf("When no commands, script files or input files are specified on the command\n");
-		printf("line, yosys automatically enters the interactive command mode. Use the 'help'\n");
-		printf("command to get information on the individual commands.\n");
-		printf("\n");
+	cxxopts::Options options(argv[0], "Yosys Open SYnthesis Suite");
+
+	options.add_options("functional")
```
povik, line 209: Suggestion: "operation" for the name of the section, and include all the options from the logging section which set the purpose of calling the binary, which is `-H`, `-h`, `-V`

kernel/driver.cc
```diff
@@ -250,292 +204,231 @@ int main(int argc, char **argv)
...
+			cxxopts::value<std::vector<std::string>>(), "<regex>")
+		("E,deps-file", "write a Makefile dependencies file <depsfile> with input and output file names",
+			cxxopts::value<std::string>(), "<depsfile>")
+		("V,version", "print version information and exit")
+	;
+	options.add_options("developer")
+		("X,trace", "enable tracing of core data structure changes. for debugging")
+		("M,randomize-pointers", "will slightly randomize allocated pointer addresses. for debugging")
+		("A,abort", "will call abort() at the end of the script. for debugging")
+		("r,top", "elaborate the specified HDL <top> module",
```
povik, line 270: This one slipped by into the developer section

mmicko:
Please switch to https link for submodule, could create issues later if not

widlarizer:
dang, ok

PR #4619, opened 2024-09-29
Allow whitespace in `tee` command paths

malmeloo:

Currently, `tee -o "test file"` creates a file literally named `"test file"`. This essentially makes it impossible to indicate files/paths that include whitespace without adding the literal quotes.

Uses `void rewrite_filename(std::string &filename)` to rewrite the path. As a nice side effect, ~/tilde paths are now expanded as well.

Just try `tee -o "test file"` before and after.


malmeloo:
Not entirely sure why that one job is failing, but it seems unrelated to my changes?

KrystalDelusion (approved):

Issue #4617, opened 2024-09-27
VCD file parsing error in sim pass with GHDL-generated VCDs due to whitespace handling


RCoeurjoly:
### Version

Yosys 0.45+139 (git sha1 e7fc1b0cc, g++ 13.2.0 -fPIC -O3)

### On which OS did this happen?

Linux

### Reproduction Steps

Given the following VHDL design (`vector_assign.vhd`):

```
library IEEE;
use IEEE.STD_LOGIC_1164.ALL;

entity vector_assign is
    Port ( 
        a : in  std_ulogic_vector(3 downto 0);
        b : out std_ulogic_vector(3 downto 0)
    );
end vector_assign;

architecture Behavioral of vector_assign is
begin
    b <= a;
end Behavioral;
```

And the following test bench (`tb.vhd`):

```
library IEEE;
use IEEE.STD_LOGIC_1164.ALL;

entity tb is
end tb;

architecture Behavioral of tb is
    component vector_assign
        Port (
            a : in  std_ulogic_vector(3 downto 0);
            b : out std_ulogic_vector(3 downto 0)
        );
    end component;

    signal a : std_ulogic_vector(3 downto 0) := (others => '0');
    signal b : std_ulogic_vector(3 downto 0);

begin
    uut: vector_assign
        Port map (
            a => a,
            b => b
        );

    stimulus: process
    begin
        a <= "0001";
        wait for 10 ns;
        wait;
    end process;
end Behavioral;

```

When we create a VCD file with the following script:

```
ghdl -a vector_assign.vhd

ghdl -a tb.vhd

ghdl -e tb

ghdl -r tb --vcd=ghdl_simulation.vcd
```

If we elaborate the VHDL design (`vector_assign.vhd`) and load the GHDL generated VCD with the following yosys script:

```
verific -vhdl2019 vector_assign.vhd
verific -import vector_assign
sim -vcd yosys_simulation.vcd -n 1
sim -r ghdl_simulation.vcd -scope tb.uut
```

We get the following error:
> 3. Executing SIM pass (simulate the circuit).
> Exec: vcd2fst tests/verific/ghdl_simulation.vcd /tmp/nix-shell.eo1ehB/converted_ghdl_simulation.fst
> Warning: Unable to find wire tb.uut.a in input file.
> Warning: Unable to find wire tb.uut.b in input file.
> <suppressed ~4 debug messages>
> ERROR: Unable to find required 'tb.uut.a' signal in file

[...]

PR #4620, opened 2024-09-30
Fix: handle VCD variable references with and without whitespace

RCoeurjoly:

Fixes https://github.com/YosysHQ/yosys/issues/4617.

We detect if a variable reference in the VCD file contains a space between the signal name and the bit range. If no space is detected, the bit or part-select is parsed and removed from the signal name.

In addition to the added test case, I processed the [neorv32](https://github.com/stnolting/neorv32) design with its GHDL-generated VCD file, and the `sim` pass did not report any issues. It might be helpful to test with other real-world examples, but I believe the current testing is sufficient to verify the fix


mmicko (approved):
Just needed to be rebased on vcd2fst integration when that one is merged. 
addr variable could use a better name as well.

Would wait till after the release before merging this, so we can try out more designs and check for regressions

RCoeurjoly:
> Just needed to be rebased on vcd2fst integration when that one is merged. addr variable could use a better name as well.
> 
> Would wait till after the release before merging this, so we can try out more designs and check for regressions

Rebased on main. Changed `addr` for `index_or_range`.

PR #4624, opened 2024-10-01
cxxrtl: test stream operator

widlarizer:
Through fast and loose regex action in #4516 I broke the stream operator for `cxxrtl::value` but the tests didn't discover this. This PR addresses that


whitequark:
Isn't that PR unmerged?

widlarizer:
Yeah @povik discovered the problem in code review and I fixed it before merging. I was surprised we didn't cover it with a test, hence this PR

PR #4626, opened 2024-10-02
select: Add new `t:@<name>` syntax

povik:

Sample use case: you identify buffer cells in a standard cell library with the following command (requires #4625)

```
cellmatch -lut_attrs *
select -set buffers a:lut=2'b10 %m
```

now you want to see the buffer cells instanced in a design. This new syntax allows you to do

```
stat t:@buffers
```

### DIFF
--- a/passes/cmds/select.cc
+++ b/passes/cmds/select.cc
@@ -876,9 +876,20 @@ static void select_stmt(RTLIL::Design *design, std::string arg, bool disable_emp
 					sel.selected_members[mod->name].insert(cell->name);
 		} else
 		if (arg_memb.compare(0, 2, "t:") == 0) {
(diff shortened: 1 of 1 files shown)


povik:
To select instances with types starting with `@` one can still do `t:\@foo`, I checked

widlarizer:
I'm at my wit's end as to why this wouldn't work, so I pushed the reproducer directly as a test

```
select -list t:@buffers
# yosys prints top/buf
select -assert-count 0 top/not t:@buffers
# yosys prints:
# ERROR: Assertion failed: selection contains 2 elements instead of the asserted 0: top/not t:@buffers
# Selection contains:
# top/not
# top/buf
```

widlarizer (approved):

PR #4627, opened 2024-10-02
Assume x values for missing signal data in FST

RCoeurjoly:
Handle cases where signal data is missing in FST files during simulation. Previously, when a signal ID was not found in past_data, it would cause an error. This patch ensures that missing signal data is assumed to be undefined ('x'), which aligns with the behaviour of `gtkwave`. Below is a screenshot of how `gtkwave` interprets `tests/sim/simple_assign.vcd`.
gtkwave_simple_assign

Instead of logging an error when the signal ID is not found, the `valueOf` now returns a string filled with 'x' characters, where the length of the string corresponds to the signal’s width.
### Note
I considered adding a warning like:
`log_warning("Signal id %d not found, assuming x\n", (int)signal);`
but decided against it to avoid extra verbosity. Open to adding it if needed.

### DIFF
--- a/kernel/fstdata.cc
+++ b/kernel/fstdata.cc
@@ -258,7 +258,8 @@ void FstData::reconstructAllAtTimes(std::vector<fstHandle> &signal, uint64_t sta
 
 std::string FstData::valueOf(fstHandle signal)
 {
-	if (past_data.find(signal) == past_data.end())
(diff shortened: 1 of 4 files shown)


mmicko (approved):
Think this we can safely add after release

Issue #4631, opened 2024-10-04
docs images: Race between latex pdf build and "tidy"


DanielG:
Hi,

In Debian we've been seeing unspecific `! I can't find file `*.aux'.` errors when building the docs, but only in some environments and only with -j>1. See [#1052902](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1052902) and [#1069424](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1069424).

While reading the Makefiles for the 100th time it finally hit me. The problem is literally right there on the first damn line: https://github.com/YosysHQ/yosys/blob/main/docs/source/_images/Makefile#L1:

```
all: examples all_tex tidy
# [...]

.PHONY: clean tidy
tidy:
	rm -f  **/*.log **/*.aux

clean: tidy
	rm -rf code_examples
	rm -f **/*.pdf **/*.svg
```

The use of the `tidy` target seems to assume make will execute the `all` target dependencies in order. It doesn't. This cleanup would have to be done as part of the `%.pdf: %.tex` target to be safe. IMO it's really bad build manners to remove the log files though. They can be quite useful for figuring out which tex file failed to build in a -j build.

The tidy target should simply be removed.

Thanks,
--Daniel

KrystalDelusion:
Does it work as expected with latest main (after merging #4632)?

PR #4632, opened 2024-10-05
docs: avoid concurrency issues when generating images in parallel

tarikgraba:
This should resolve issue #4631 when building the doc images in parallel.

I have removed the `tidy` target and I just delete log and aux file after building each pdf.

### DIFF
--- a/docs/source/_images/Makefile
+++ b/docs/source/_images/Makefile
@@ -1,4 +1,4 @@
-all: examples all_tex tidy
+all: examples all_tex
 
 # set a fake time in pdf generation to prevent unnecessary differences in output
 FAKETIME := TZ='Z' faketime -f '2022-01-01 00:00:00 x0,001'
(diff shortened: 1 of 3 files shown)


docs/source/_images/Makefile
```diff
@@ -31,14 +31,13 @@ all_tex: $(TEX_FILES:.tex=.pdf) $(TEX_FILES:.tex=.svg)
 
 %.pdf: %.tex
 	cd $(@D) && $(FAKETIME) pdflatex $(<F) --interaction=nonstopmode
+	rm -f $(*).aux $(*).log
```
KrystalDelusion, line 34: I think the rm was there because of how the docs website was being done and is no longer necessary. It would be better to leave this out entirely and just rely on the gitignore instead of removing the log files.

KrystalDelusion (commented):
Thanks for doing this!  I always run make docs without a -j flag and have been meaning to look into it for ages but kept forgetting.

tarikgraba:
I have reverted my previous commit and just removed the `tidy` target from the main target as I understand it is not needed any more (aux/log are ignored in `docs/.gitignore`)

PR #4639, opened 2024-10-07
Explictly #include <variant> for std::variant usage.

mikesinouye:


povik (approved):
`std::variant` is referenced in the template specialization of hash_ops

PR #4643, opened 2024-10-09
wheels: fix missing yosys-abc/share directory

donn:

Due to an oversight on my part, #4534 does not copy either `yosys-abc` or the `share` directory. While Yosys technically still runs, it is unusable for anything non-trivial.

* `misc/__init__.py`:
  * checks if there's a `yosys-abc` in the same directory - if yes, sets the variable `sys._pyosys_abc`
  * checks if there's a `share` in the same directory - if yes, sets the variable `sys._pyosys_share_dirname`
* `yosys.cc::init_share_dirname`: check for `sys._pyosys_share_dirname` at the highest priority if Python is enabled
* `yosys.cc::init_abc_executable_name`: check for `sys._pyosys_abc`, use it at at the highest priority if Python is enabled
* `Makefile`: add new target, `share`, to only create the extra targets
* `setup.py`: compile libyosys.so, yosys-abc and share, and copy them all as part of the pyosys build
* `test/arch/ecp5/add_sub.py`: ported `add_sub.ys` to Python to act as a test for the share directory and abc with Python wheels, used in CI

```
pip install -i https://test.pypi.org/simple/ pyosys==0.46.0
```

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -110,7 +110,7 @@ jobs:
             MACOSX_DEPLOYMENT_TARGET=11
             makeFlags='BOOST_PYTHON_LIB=./boost/pfx/lib/libboost_python*.a CONFIG=clang'
           CIBW_BEFORE_BUILD: bash ./.github/workflows/wheels/cibw_before_build.sh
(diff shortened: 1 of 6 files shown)


mmicko (approved):
Synth passes works as expected now on isolated environment. Thanks

PR #4650, opened 2024-10-09
Tribuf propagate

Rodrigodd:

I am working on a non-synthesizable project, that make extensive use of tri-state signals, which. This change allows to eliminating all tristate usage in the design, by flattening the module hierarchy, detecting and propagating all tribuf cells, and  https://github.com/emu-russia/dmgcpu/pull/292


Rodrigodd:
Sorry, misclick the publish button.

PR #4657, opened 2024-10-11
cxxrtl: fix handling of 0-bit variables in `vcd_writer.sample()`.

jfng:
Before this commit, `vcd_writer.sample()` emits vector values for 0-bit variables.

For example, the following RTLIL:

```
module \top
  wire width 0 \foo
end
```

would result in the following VCD:

```
$timescale 1 us $end
$var wire 0 ! foo $end
$enddefinitions $end
#0
b !
```

However, according to IEEE 1364-2001, the syntax for vector values is:

20241011_22h37m15s_grim

After this commit, `foo` is no longer sampled in the VCD:

```
$timescale 1 us $end
$var wire 0 ! foo $end
$enddefinitions $end
#0
```

### DIFF
--- a/backends/cxxrtl/runtime/cxxrtl/cxxrtl_vcd.h
+++ b/backends/cxxrtl/runtime/cxxrtl/cxxrtl_vcd.h
@@ -127,6 +127,8 @@ class vcd_writer {
 			bool bit_curr = var.curr[bit / (8 * sizeof(chunk_t))] & (1 << (bit % (8 * sizeof(chunk_t))));
 			buffer += (bit_curr ? '1' : '0');
 		}
+		if (var.width == 0)
+			buffer += '0';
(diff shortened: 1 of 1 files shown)


jfng:
Actually, it may be better to emit a scalar value of 0 instead. If a VCD variable has no samples, GTKWave gives it the type "real" and the value "nan", which is confusing.

whitequark:
> Actually, it may be better to emit a scalar value of 0 instead.

Yeah, sounds reasonable.

jfng:
I changed it to emit a vector value of 0 instead (eg. `b0 !`).

Now, both Surfer and GTKWave show a 0. Otherwise, GTKWave would show an "X" for a scalar value of a variable that isn't 1-bit.

whitequark (approved):

Issue #4665, opened 2024-10-14
misc/yosys-config.in uses an non-portable test(1) operator


thorpej:
### Version

Yosys 0.45

### On which OS did this happen?

BSD

### Reproduction Steps

Building yosys 0.45 on NetBSD fails due to an non-portable test(1) operator used in misc/yosys-config.in.

        if [ "$1" == "--build" ]; then

NetBSD's pkgsrc system has a patch for this.  This issue is to track the upstreaming of that patch.  A pull request with the fix is forthcoming.

### Expected Behavior

The script should run.

### Actual Behavior

The script errors out.

thorpej:
Pull request filed for this issue: https://github.com/YosysHQ/yosys/pull/4666

PR #4666, opened 2024-10-14
misc/yosys-config.in: don't use the non-portable '==' operator with test(1)

thorpej:
The '==' operator is non-standard and does not work with all versions of test(1) or sh(1).

The fix is simple: use the standard '=' operator instead.

Ensure yosys builds.

Resolves issue #4665.

### DIFF
--- a/misc/yosys-config.in
+++ b/misc/yosys-config.in
@@ -44,7 +44,7 @@ if [ $# -eq 0 ]; then
 	help
 fi
 
-if [ "$1" == "--build" ]; then
+if [ "$1" = "--build" ]; then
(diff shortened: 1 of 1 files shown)


widlarizer:
The file is shebanged with `#/usr/bin/env bash` and bash on different platforms should be consistent, I'm not really sure how does `sh` come into play here, but since `=` and `==` is [strictly equivalent](https://www.gnu.org/software/bash/manual/html_node/Bash-Conditional-Expressions.html) in bash, this PR is good to go

widlarizer (approved):

PR #4668, opened 2024-10-16
actions: vendor sources with submodules for releases

widlarizer:
We like submodules for SBOM reasons and prefer them over in-repo vendored libraries. However this creates a burden on users with needs for gitless workflows who have to take our (yosys) source tarball, drop in our other (yosys-abc) source tarball, manually check submodules versions at the git tag, download those, and only then build yosys. An alternate solution is to add Makefile targets like curl-abc #4472 but that needs per-submodule work. This PR adds a workflow which on main commits creates a source tarball artifact including submodule sources

- [x] instruct users of vendored dependencies that licenses in `abc/` and `libs/*` apply to those components
- [x] eliminate zeroes that prevent `libs/cxxopts` from being visible

### DIFF
--- a/.github/workflows/source-vendor.yml
+++ b/.github/workflows/source-vendor.yml
@@ -1,0 +1,33 @@
+name: Create source archive with vendored dependencies
+
+on: [push, workflow_dispatch]
+
+jobs:
+  vendor-sources:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository with submodules
+        uses: actions/checkout@v4
+        with:
+          submodules: 'recursive'
+
+      - name: Create clean tarball
+        run: |
+          git archive --format=tar HEAD -o yosys-src-vendored.tar
+          git submodule foreach '
+            git archive --format=tar --prefix="${sm_path}/" HEAD --output=${toplevel}/vendor-${name}.tar
+          '
+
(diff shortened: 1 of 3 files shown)


.github/workflows/source-vendor.yml
```diff
@@ -0,0 +1,33 @@
...
+          done
+
+          gzip yosys-src-vendored.tar
+
+      - name: Store tarball artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: vendored-sources
+          path: yosys-src-vendored.tar.gz
+          retention-days: 1
```
KrystalDelusion, line 33: Is the idea here to replace the current ABC tarball with this combined tarball for inclusion in release?

widlarizer: I don't know, but the availability of an all-inclusive tarball is kind of orthogonal to whether there is still a need to provide yosys-abc seperately in the yosys repo

KrystalDelusion: I guess the replacement is less what I was asking, and more is retention-days: 1 because the intention is for it to be added to releases, as opposed to being generally available for the most recent commit

widlarizer: Yes, 1 because the intention is to be added to releases. My intention is to be able to run it manually actually, so:

- [ ]  re-add workflow_dispatch

KrystalDelusion (approved):

PR #4672, opened 2024-10-16
driver: fix special args passing to tcl and python

widlarizer:
Fixes #4669. To test the python side this needs the changes from #4671 but even with that I get a segfault with `valgrind ./yosys -y foo` starting with this:

```
Invalid read of size 8
   at 0x4D033A3: PyList_New.constprop.0 (in /nix/store/sxr2igfkwhxbagri49b8krmcqz168sim-python3-3.11.8/lib/libpython3.11.so.1.0)
   by 0x4CEB408: PyImport_Import (in /nix/store/sxr2igfkwhxbagri49b8krmcqz168sim-python3-3.11.8/lib/libpython3.11.so.1.0)
   by 0x4CEB6DF: PyImport_ImportModule (in /nix/store/sxr2igfkwhxbagri49b8krmcqz168sim-python3-3.11.8/lib/libpython3.11.so.1.0)
   by 0xDA234D: Yosys::init_share_dirname() (kernel/yosys.cc:1017)
   by 0xDA0DCF: Yosys::yosys_setup() (kernel/yosys.cc:550)
   by 0xCC1FCC: main (kernel/driver.cc:497)
 Address 0x10 is not stack'd, malloc'd or (recently) free'd
```

@donn any ideas?

PR #4675, opened 2024-10-17
yosys: fix pyosys initialization segfault

widlarizer:
see https://github.com/YosysHQ/yosys/pull/4672#issuecomment-2420128412

### DIFF
--- a/kernel/yosys.cc
+++ b/kernel/yosys.cc
@@ -547,12 +547,6 @@ void yosys_setup()
 	if(already_setup)
 		return;
 	already_setup = true;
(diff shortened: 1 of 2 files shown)


donn:
@widlarizer Is there an ETA for this being merged (I feel really bad about it bec its my fault)

widlarizer:
Pyosys is a dubiously designed feature and not part of the primary use case so it's a low impact partial breakage so it's not a big problem. I'll just go ahead and merge it without approval since it's your patch and I approve of it

PR #4677, opened 2024-10-18
opt_merge: hashing performance and correctness

widlarizer:
This is a direct remake of #4175 sans the 64-bit hash value. I'm making use of the interface in #4524 and requiring that PR (and containing its commits at the moment). Instead of xorshifts, values are sorted, though a final xorshift is included as a part of the fudge (`--hash-seed=N`) mechanism.

Additionally, I discovered opt_merge behaves incorrectly in the case of hash collisions. This suggests that this PR might in rare cases bring improvements in quality of results for flows that use opt_merge, since prior to it, hash collisions would inhibit merging. I modified the `sharemap` from a `dict<hash_t, Cell*>` to an equivalent `std::unordered_multimap` so that multiple cells can be associated with the same hash. This can't be a separate change since this bug actually broke the build just by changing how the hashes are constructed.

Sorry for the spam to code owners due to being based on the above mentioned wide-reaching PR #4524, I don't have a way of removing you from the reviewer list. The diff for this PR is not going to be very readable on github either until that's merged

### DIFF
--- a/passes/opt/opt_merge.cc
+++ b/passes/opt/opt_merge.cc
@@ -154,7 +154,7 @@ struct OptMergeWorker
 
 	bool compare_cell_parameters_and_connections(const RTLIL::Cell *cell1, const RTLIL::Cell *cell2) const
 	{
-		log_assert(cell1 != cell2);
+		if (cell1 == cell2) return true;
 		if (cell1->type != cell2->type) return false;
 
 		if (cell1->parameters != cell2->parameters)
@@ -313,6 +313,8 @@ struct OptMergeWorker
 					if (cell->has_keep_attr()) {
 						if (other_cell->has_keep_attr())
 							continue;
+						known_cells.erase(other_cell);
+						known_cells.insert(cell);
 						std::swap(other_cell, cell);
 					}
 
--- a/passes/opt/opt_merge.cc
+++ b/passes/opt/opt_merge.cc
@@ -26,6 +26,8 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <set>
+#include <unordered_map>
+#include <array>
 
 
 USING_YOSYS_NAMESPACE
@@ -41,6 +43,22 @@ struct OptMergeWorker
 
 	CellTypes ct;
 	int total_count;
+
+	static vector<pair<SigBit, SigSpec>> sorted_pmux_in(const dict<RTLIL::IdString, RTLIL::SigSpec> &conn)
+	{
+		SigSpec sig_s = conn.at(ID::S);
+		SigSpec sig_b = conn.at(ID::B);
+
+		int s_width = GetSize(sig_s);
+		int width = GetSize(sig_b) / s_width;
+
+		vector<pair<SigBit, SigSpec>> sb_pairs;
+		for (int i = 0; i < s_width; i++)
+			sb_pairs.push_back(pair<SigBit, SigSpec>(sig_s[i], sig_b.extract(i*width, width)));
+
+		std::sort(sb_pairs.begin(), sb_pairs.end());
+		return sb_pairs;
+	}
 
 	static void sort_pmux_conn(dict<RTLIL::IdString, RTLIL::SigSpec> &conn)
 	{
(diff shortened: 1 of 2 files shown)


widlarizer:
The current implementation regresses opt_merge runtime many times over. I'll see if `std::unordered_multimap`.
Currently, `std::unordered_multimap`, as I use it

+ maps from `hash_t` to `Cell*`
+ hashes `hash_t` with `std::hash` (unnecessary)
+ compares `Cell*`s with pointer equality

Instead it should:

+ map from `Cell*` to `Cell*`
+ hash with `hash_cell_function`
+ compare with `compare_cell_parameters_and_connections`

widlarizer:
`pool` fails to bring a clear advantage over `std::unordered_set`:

```
ibex pool
Elapsed time: 0:42.16[h:]min:sec. CPU time: user 41.87 sys 0.17 (99%). Peak memory: 164800KB.

ibex std::unordered_set
Elapsed time: 0:42.85[h:]min:sec. CPU time: user 42.56 sys 0.16 (99%). Peak memory: 164076KB.

ibex pool
Elapsed time: 0:49.41[h:]min:sec. CPU time: user 48.76 sys 0.49 (99%). Peak memory: 634672KB.

ibex std::unordered_set
Elapsed time: 0:47.62[h:]min:sec. CPU time: user 46.95 sys 0.51 (99%). Peak memory: 629852KB.
```

rmlarsen:
@widlarizer nice work!

widlarizer:
I checked for regression on the current rebased head (57ca09ee) and we're good on that

rmlarsen:
Switching to a `std::` datastructure is much appreciated!

passes/opt/opt_merge.cc
```diff
@@ -275,36 +307,34 @@ struct OptMergeWorker
...
-						}
-
+				auto [cell_in_map, inserted] = known_cells.insert(cell);
+				if (!inserted) {
+					// We've failed to insert since we already have an equivalent cell
+					Cell* other_cell = *cell_in_map;
+					if (cell->has_keep_attr()) {
+						if (other_cell->has_keep_attr())
+							continue;
+						std::swap(other_cell, cell);
```
povik, line 317: When this happens I think we need to update known_cells so that it holds the cell which survives

povik: Maybe this has been always a bug?

passes/opt/opt_merge.cc
```diff
@@ -275,36 +307,34 @@ struct OptMergeWorker
...
+						if (cell->output(it.first)) {
+							RTLIL::SigSpec other_sig = other_cell->getPort(it.first);
+							log_debug("    Redirecting output %s: %s = %s\n", it.first.c_str(),
+									log_signal(it.second), log_signal(other_sig));
+							Const init = initvals(other_sig);
+							initvals.remove_init(it.second);
+							initvals.remove_init(other_sig);
+							module->connect(RTLIL::SigSig(it.second, other_sig));
+							assign_map.add(it.second, other_sig);
+							initvals.set_init(other_sig, init);
```
povik, line 332: I don't understand this initvals dance but I see it was pre-existing

passes/opt/opt_merge.cc
```diff
@@ -65,93 +83,77 @@ struct OptMergeWorker
...
-			if (assign_map(alt_conn.at(ID::A)) < assign_map(alt_conn.at(ID::B))) {
-				alt_conn[ID::A] = conn->at(ID::B);
-				alt_conn[ID::B] = conn->at(ID::A);
+			std::array<RTLIL::SigSpec, 2> inputs = {
+				assign_map(cell->getPort(ID::A)),
+				assign_map(cell->getPort(ID::B))
+			};
+			std::sort(inputs.begin(), inputs.end());
+			h = hash_ops<std::array<RTLIL::SigSpec, 2>>::hash_into(inputs, h);
+			h = assign_map(cell->getPort(ID::Y)).hash_into(h);
```
povik, line 98: Output shouldn't be hashed here

povik (changes requested):

passes/opt/opt_merge.cc
widlarizer: Nice catch! After the loop you can add
```
for (auto cell : known_cells) {
	log("known_cells %s\n", log_id(cell));
}
```
and trip valgrind on `tests/opt/opt_merge_keep.ys`... that's no good

passes/opt/opt_merge.cc
widlarizer: My bad, but I gotta say, the idea that there might be zero tests for the basic functionality of opt_merge just hasn't crossed my mind... I'll have to add some. Better late than never

passes/opt/opt_merge.cc
widlarizer: I thought this might be to avoid `FfInitVals` or `SigMap` getting notified with their monitor about the connection, but they don't use `Monitor`s so it remains a mystery

KrystalDelusion (approved):
I've only checked the fixes from @povik's requested changes, but seems fine to me.  If possible, do you have any test cases for the correctness fixes that you mentioned?  The test script added at the moment passes on latest main (and I assume given the name is just testing the basics, not the edge cases where things fall over with the current implementation)

widlarizer:
@KrystalDelusion The "many collisions case" was something I discovered when I broke the hashing function and caused everything to collide. Previously, if you had cells `a`, `b` and `c`, and they all incidentally got the same hash, and `a`!=`b` and `b`=`c`, then `a` would become the representative cell for that hash value, `b` would be checked against it and found not equal, `a` would stay as the representative, and `c` would also be checked against `a` and found not equivalent. This means that although there are two mergable cells (`b` and `c`), just because they both collide with a third non-equivalent, they wouldn't get merged. This meant that the correct functionality of the pass relied on properties of the hashing and ordering of the cell iteration, which is no good.

Testing this would require adding a flag to the pass that forces this unfortunate case, since we can't really cause it to happen externally. For the hashing done at the core of this pass, it would override the hashing function to return the same result for anything

PR #4678, opened 2024-10-21
Start Tcl design inspection methods

povik:

Allow inspection and altering of the design from Tcl scripts for use-cases which are too specific to write a Yosys command for.

This is in addition to the existing ways of obtaining information about a design by

    [yosys tee -q -s result.json stat -json]

and variations thereof.

See sample usage here: https://github.com/povik/OpenROAD-flow-scripts/blob/72b084f68275aaefec903fbc8e4f8613679e52e0/flow/scripts/synth_preamble.tcl#L122

Another use case of the new API is replacing this Python script called from Tcl to process a JSON export with a direct Tcl implementation: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/master/flow/scripts/mem_dump.py (Cc @oharboe, feel free to comment)

We introduce a set of Tcl functions prefixed with `rtlil::` to query attributes and parameters.

Tcl bindings in other projects sometimes use text-formatted pointers as the Tcl-side handles, but for a RTLIL binding we can consistently use IDs of objects to refer to them from the Tcl side.

### DIFF
--- a/kernel/yosys.cc
+++ b/kernel/yosys.cc
@@ -829,11 +829,309 @@ static int tcl_yosys_cmd(ClientData, Tcl_Interp *interp, int argc, const char *a
 	return TCL_OK;
 }
 
+#define FLAG(name) \
+	if (!strcmp(argv[i], "-" #name)) { \
+		name##_flag = true; \
+		continue; \
+	} \
+
+#define ERROR(str) \
+	{ \
+		Tcl_SetResult(interp, (char *)(str), TCL_STATIC); \
+		return TCL_ERROR; \
+	}
+
+static int tcl_get_attr(ClientData, Tcl_Interp *interp, int argc, const char *argv[])
+{
+	int i;
+	bool mod_flag = false, string_flag = false, int_flag = false, bool_flag = false;
+	for (i = 1; i < argc; i++) {
+		FLAG(mod)
+		FLAG(string)
+		FLAG(int)
+		FLAG(bool)
+		break;
+	}
+
+	if ((mod_flag && i != argc - 2) ||
+			(!mod_flag && i != argc - 3) ||
+			(string_flag + int_flag + bool_flag > 1))
+		ERROR("bad usage: expected \"get_attr -mod [-string|-int|-bool] <module> <attrname>\""
+			  " or \"get_attr [-string|-int|-bool] <module> <identifier> <attrname>\"")
+
+	IdString mod_id, obj_id, attr_id;
+	mod_id = RTLIL::escape_id(argv[i++]);
+	if (!mod_flag)
+		obj_id = RTLIL::escape_id(argv[i++]);
+	attr_id = RTLIL::escape_id(argv[i++]);
+
+	RTLIL::Module *mod = yosys_design->module(mod_id);
+	if (!mod)
+		ERROR("module not found")
+
+	RTLIL::AttrObject *obj = nullptr;
+	if (mod_flag) {
+		obj = mod;
+	} else {
+		obj = mod->wire(obj_id);
+		if (!obj)
+			obj = mod->memories.at(obj_id, nullptr);
+		if (!obj)
+			obj = mod->cell(obj_id);
+		if (!obj)
+			obj = mod->processes.at(obj_id, nullptr);
+	}
+
+	if (!obj)
+		ERROR("object not found")
+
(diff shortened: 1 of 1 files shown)


kernel/tclapi.cc
```diff
@@ -0,0 +1,449 @@
...
+
+		RTLIL::Const &value = obj->attributes.at(attr_id);
+		if (value.size() > 32)
+			ERROR("value too large")
+
+		// FIXME: 32'hffffffff will return as negative despite is_signed=false
+		Tcl_SetResult(interp, (char *) std::to_string(value.as_int()).c_str(), TCL_VOLATILE);
+	} else if (bool_flag) {
+		bool value = obj->get_bool_attribute(attr_id);
+		Tcl_SetResult(interp, (char *) std::to_string(value).c_str(), TCL_VOLATILE);
```
jix, line 212: I think using `Tcl_SetObjResult(interp, Tcl_NewIntObj(value))` here and elsewhere would be nicer and avoid the roundtrip to a string representation.

kernel/tclapi.cc
```diff
@@ -0,0 +1,449 @@
...
+		FLAG(int)
+		FLAG(bool)
+		break;
+	}
+
+	if ((mod_flag && i != argc - 2) ||
+			(!mod_flag && i != argc - 3) ||
+			(string_flag + int_flag + bool_flag > 1))
+		ERROR("bad usage: expected \"get_attr -mod [-string|-int|-bool] <module> <attrname>\""
+			  " or \"get_attr [-string|-int|-bool] <module> <identifier> <attrname>\"")
```
jix, line 170: It might be nice to support an optional `-default <default_value>` (which will return the default value unchanged, not reinterpret it as the selected type).

One potential issue with adding more flags (whether it's this `-default` or something else) is that before doing so `get_attr -mod -default some_attr` would look up `some_attr` in a module `-default` but after adding that option it would be missing arguments. Maybe we should require using `{\-default}` (which already works due to escape_id) and/or `--` and interpret all `-` strings as flags/options (producing an error if unknown) until `--` is found. This is also an issue just for the existing flags, i.e. you have to use `[rtlil::get_attr -mod -string {\-mod} some_attr]` and can't use `[rtlil::get_attr -mod -string -mod some_attr]` to access `module \-mod (...` . I'm not sure what's the convention for tcl commands wrt to disambiguating options and strings starting with `-`.

kernel/tclapi.cc
```diff
@@ -0,0 +1,449 @@
...
+int yosys_tcl_iterp_init(Tcl_Interp *interp)
+{
+	if (Tcl_Init(interp)!=TCL_OK)
+		log_warning("Tcl_Init() call failed - %s\n",Tcl_ErrnoMsg(Tcl_GetErrno()));
+	Tcl_CreateCommand(interp, "yosys", tcl_yosys_cmd, NULL, NULL);
+	Tcl_CreateCommand(interp, "rtlil::get_attr", tcl_get_attr, NULL, NULL);
+	Tcl_CreateCommand(interp, "rtlil::has_attr", tcl_has_attr, NULL, NULL);
+	Tcl_CreateCommand(interp, "rtlil::set_attr", tcl_set_attr, NULL, NULL);
+	Tcl_CreateCommand(interp, "rtlil::get_param", tcl_get_param, NULL, NULL);
+	Tcl_CreateCommand(interp, "rtlil::set_param", tcl_set_param, NULL, NULL);
```
jix, line 423: I'd also expect `rtlil::unset_attr` as well as `rtlil::attr_list` to get a complete set of operations (same for param which is also missing `rtlil::has_param`).

Maybe also useful would be something like `rtlil::attr_array` which would produce all attributes as a key-value array such that you can do `array set object_attrs [rtlil::attr_array some_mod some_object]; puts "the value of keep is $object_attrs(\\keep)"` and use other tcl array functions. I don't actually use tcl enough to know how useful this would be, though.

jix (commented):
Overall this lgtm, I'd suggest making a decision on flag / option parsing before merging a first version, and also fixing the unsigned MSB set int handling, both to avoid backwards incompatible changes. The other points I commented on can be addressed in a follow up if the goal is to get something usable merged quickly.

povik:
Pulled fixes from @jix. Thanks!

PR #4680, opened 2024-10-22
Makefile: add -rdynamic on macOS to fix plugins with LTO

widlarizer:
Prior to this change, some symbols would be missing from the yosys executable if built on macOS with LTO. This would break plugins on dlopen. By adding `-rdynamic` to the linker flags, we can export all symbols even if they're unused due to LTO-enabled inlining or otherwise. You can check the exported symbols by building yosys with LTO on macOS and running `objdump --macho --exports-trie yosys`.


widlarizer:
Oh well

PR #4683, opened 2024-10-24
Respect $SOURCE_DATE_EPOCH in generate_bram_types_sim.py

keszybz:

I'm working on build reproducibility of Fedora packages, and this patch fixes an issue observed in test rebuilds.

The variable is set in build environments to allow the build to be reproducible (identical result in independent builds), see https://reproducible-builds.org/docs/source-date-epoch/. The code snippet is based on the sample from that page too.

Generate the output without $SOURCE_DATE_EPOCH set, which should result in a file with a current timestamp, and then again with SOURCE_DATE_EPOCH set, which should result in the specified timestamp.


keszybz:
Downstream pull request: https://src.fedoraproject.org/rpms/yosys/pull-request/5

techlibs/quicklogic/qlf_k6n10f/generate_bram_types_sim.py
```diff
@@ -1,10 +1,17 @@
...
 def generate(filename):
+	build_date = datetime.fromtimestamp(
+		int(os.environ.get('SOURCE_DATE_EPOCH', time.time())),
+		tz=timezone.utc,
+	)
+
 	with open(filename, "w") as f:
 		f.write("// **AUTOGENERATED FILE** **DO NOT EDIT**\n")
-		f.write(f"// Generated by {sys.argv[0]} at {datetime.now(timezone.utc)}\n")
+		f.write(f"// Generated by {sys.argv[0]} at {build_date}\n")
```
whitequark, line 14: Do we even need the build date here? The file modification time indicates it already...

keszybz: I don't think it's *particularly* useful, but meh, I don't care either way, as long as the file doesn't change between rebuilds ;)

widlarizer: I think you can go ahead and remove the generation time outright. It's an outlier among the other "Generated by" string formatting in the codebase

keszybz: Updated.

widlarizer (approved):

ldoolitt:
Can "we" not forget about this pull request, please?

widlarizer:
Thanks for the reminder and sorry for the delay, I was used to a bitbucket-based project where everybody has merge permissions only blocked on requiring an approval

PR #4684, opened 2024-10-25
qwp: remove

Ravenslofty:
This must be among the more obscure passes in yosys; a quadratic wirelength placer. nextpnr has had a placer of this type for [almost six years](https://github.com/YosysHQ/nextpnr/pull/219). 

Yosys is the wrong place for it, because it has no concept of the physical requirements of a target architecture.

yeet.


povik (approved):
We can always go back and recover this from the version control history

PR #4685, opened 2024-10-26
aiger2: Fix open-coded constants

povik:

A minor fix (to fix usage of Index out of the aiger backend)


Ravenslofty (approved):
I think we can Just Merge It

Issue #4573, opened 2024-08-30
Problems with 3 bit add not passing eqy


oharboe:
### Version

Yosys 0.43 (git sha1 ead4718e5, g++ 13.2.0-23ubuntu4 -fPIC -Os)

### On which OS did this happen?

Linux

### Reproduction Steps

I realize this is not the shortest possible test-case for this problem, but it is a standalone test-case with everything needed to reproduce it. Thanks to @hovind for figuring this out. This was discovered while working with OpenROAD on ASAP7.

To reproduce:

untar https://drive.google.com/file/d/1x1WbP7VeAsBBjP4zuXUaUmnka_AikNCp/view?usp=sharing

Then run:

```
./run-me-adder-asap7-base.sh
```

Output:

```
26. Executing Verilog backend.
Dumping module `\adder'.

End of script. Logfile hash: f8c4509b61, CPU: user 1.16s system 0.05s, MEM: 132.97 MB peak
Yosys 0.43 (git sha1 ead4718e5, g++ 13.2.0-23ubuntu4 -fPIC -Os)
Time spent: 24% 6x read_liberty (0 sec), 23% 2x stat (0 sec), ...
+ eqy -d ./logs/asap7/adder/base/4_eqy_output --force ./objects/asap7/adder/base/1_2_test.eqy
+ tee ./logs/asap7/adder/base/1_2_test.txt
EQY  2:05:25 [./logs/asap7/adder/base/4_eqy_output] read_gold: starting process "yosys -ql ./logs/asap7/adder/base/4_eqy_output/gold.log ./logs/asap7/adder/base/4_eqy_output/gold.ys"
EQY  2:05:25 [./logs/asap7/adder/base/4_eqy_output] read_gold: finished (returncode=0)
EQY  2:05:25 [./logs/asap7/adder/base/4_eqy_output] read_gate: starting process "yosys -ql ./logs/asap7/adder/base/4_eqy_output/gate.log ./logs/asap7/adder/base/4_eqy_output/gate.ys"
EQY  2:05:25 [./logs/asap7/adder/base/4_eqy_output] read_gate: finished (returncode=0)
EQY  2:05:25 [./logs/asap7/adder/base/4_eqy_output] combine: starting process "yosys -ql ./logs/asap7/adder/base/4_eqy_output/combine.log ./logs/asap7/adder/base/4_eqy_output/combine.ys"
[...]

PR #3882, opened 2023-08-07
extract_fa: Fix handling inversion of an FA xor output

jix:
Fixes what seems to be a missing inversion of the xor output when the heueristic decides to use an inverted FA.

Issue #3879, opened 2023-08-07
LEC failed after yosys synthesis


Telpukhov:
### Version

up to 0.32

### On which OS did this happen?

Linux

### Reproduction Steps

Yosys makes wrong full adder extraction in the case of adding multiple bits. Minimal example of such behaviour is the sum of 4 bits. Initial RTL:

```
module gcd(I, D);

  output [2:0] I;
  input [3:0] D;

  assign I = D[0]+D[1]+D[2]+D[3];
endmodule
```

Simple script for Yosys, with extract_fa command:

```
read_verilog gcd.v
synth  -top gcd -flatten
extract_fa

opt -purge
show -stretch -prefix gcd -format dot
write_verilog res.v
```

I've analysed output file and graphical representation. It have obvious inequivalence to the initial RTL. But to be consistent, I've made a script for equivalence check: 

```
read_verilog gcd.v

hierarchy -top gcd
proc
opt 
opt
async2sync
opt
splitnets -ports
opt
flatten
opt
select gcd
rename -hide
    
                   
design -stash gold


read_verilog res.v

hierarchy -top gcd
proc
opt
opt
async2sync
opt
splitnets -ports
opt
flatten
opt
select gcd
rename -hide

design -stash gate


design -copy-from gold -as gold gcd
design -copy-from gate -as gate gcd

miter -equiv -flatten -ignore_gold_x gold gate miter
hierarchy -top miter
opt
select miter
hierarchy -top miter
rename -hide

sat -verify -tempinduct -prove trigger 0 -set-init-undef -set-def-inputs -seq 5

```
 
Interesting part is that on LINUX builds I get this problem and wrong netlist. But then I tried to run this testcase on Windows pre-built binaries, and everything was fine - output netlist was correct, and equivalence was prooved. So in next sections I will submit Windows results like "Expected Behaviour", and Linux results like "Actual Behaviour".

Analysing wrong results you can verify that I[0] should be XOR of D[0], D[1], D[2], D[3]. But in resulted netlist we get I[0] as XOR of ~D[0], ~D[1], D[2], ~D[3]. It gets mismatch already on test vector (0, 0, 0, 0). Equivalence script expectedly indicates a mismatch.

[...]

PR #4691, opened 2024-10-30
extract_fa: Fix `xor3`/`xnor3` inversion bug

hovind:
#### _What are the reasons/motivation for this change?_

The `extract_fa` pass seems to have issues when an odd number of inputs are inverted, this is previously reported in https://github.com/YosysHQ/yosys/issues/3879 and https://github.com/YosysHQ/yosys/issues/4573.

_I think_ @eddiehung solved this in the `xor2`/`xnor2` case in https://github.com/YosysHQ/yosys/pull/1297, but the `xor3`/`xnor3` case lacks the equivalent logic. 

#### _Explain how this is achieved._

Similarly to how it is handled in the `xor2`/`xnor2` case, we invert the `xor3`/`xnor3` output if the `majority3` has an odd number of inverted inputs.

#### _If applicable, please suggest to reviewers how they can test the change._

I have tried to make this break using @jix's test case from https://github.com/YosysHQ/yosys/pull/3882.

~I don't have much confidence in my understanding of this code, and I hope that some discussion here will help clarify some things for me.~

~94215584e9bbc648c7c6184a4239d3be7df65db3 should perhaps be in a separate pull request (or probably just deleted). I left it in so I could ask if somebody could help clarify to me what the deleted code segment achieves.~

~84d0c8fbfc66bcd2a5ff54bd00a9cfba500ed248 I thought initially this was a bug fix, but I'm not sure `invert_xy` and `f2i.inv_a ^ f2i.inv_b` can be true at the same time?~

Thanks for the amazing tool!

### DIFF
--- a/passes/techmap/extract_fa.cc
+++ b/passes/techmap/extract_fa.cc
@@ -412,14 +412,15 @@ struct ExtractFaWorker
 					facache[fakey] = make_tuple(X, Y, cell);
 				}
 
+				bool invert_y = f3i.inv_a ^ f3i.inv_b ^ f3i.inv_c;
 				if (func3.at(key).count(xor3_func)) {
-					SigBit YY = invert_xy ? module->NotGate(NEW_ID, Y) : Y;
+					SigBit YY = invert_xy ^ invert_y ? module->NotGate(NEW_ID, Y) : Y;
 					for (auto bit : func3.at(key).at(xor3_func))
 						assign_new_driver(bit, YY);
 				}
 
 				if (func3.at(key).count(xnor3_func)) {
-					SigBit YY = invert_xy ? Y : module->NotGate(NEW_ID, Y);
+					SigBit YY = invert_xy ^ invert_y ? Y : module->NotGate(NEW_ID, Y);
 					for (auto bit : func3.at(key).at(xnor3_func))
 						assign_new_driver(bit, YY);
 				}
--- a/tests/various/bug3879.ys
+++ b/tests/various/bug3879.ys
@@ -1,0 +1,29 @@
+read_verilog <<EOF
+module gcd(I, D);
+
+  output [2:0] I;
+  input [3:0] D;
+
+  assign I = D[0]+D[1]+D[2]+D[3];
+endmodule
+EOF
+design -save input
+
+prep
+
+design -stash gold
+
+design -load input
+
+synth  -top gcd -flatten
+
+extract_fa -v
+
+design -stash gate
+
+design -copy-from gold -as gold gcd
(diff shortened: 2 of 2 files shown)


oharboe:
@povik Thoughts?

povik:
> @povik Thoughts?

I think I hope we will be able to obsolete this code soon in favor of something more readable.

povik (approved):

passes/techmap/extract_fa.cc
```diff
@@ -412,14 +412,15 @@ struct ExtractFaWorker
 					facache[fakey] = make_tuple(X, Y, cell);
 				}
 
+				bool invert_y = f3i.inv_a ^ f3i.inv_b ^ f3i.inv_c;
 				if (func3.at(key).count(xor3_func)) {
-					SigBit YY = invert_xy ? module->NotGate(NEW_ID, Y) : Y;
+					SigBit YY = invert_xy ^ invert_y ? module->NotGate(NEW_ID, Y) : Y;
```
povik, line 417: This makes sense: at this point we are looking to implement xor3, respectivelly xnor3 in terms of the signals held in A, B, C.

Y is the Y output of a (possibly cached) full-adder cell over A, B, C, with inversions inserted as indicated by f3i.inv_a/b/c, with the added caveat that if invert_xy is true, all of the inputs are complemented on top of that.

With that being said the logic around here seems correct to decide whether we need to use Y, or Y complemented to get the xor3/xnor3 result.

povik:
Thanks for the fix and for the included test case.

whitequark:
I have a better idea for fixing this pass: it's called `rm`. It's not useful!

povik:
> I have a better idea for fixing this pass: it's called rm. It's not useful!

Our ideas are actually one and the same since what I hope will happen is that we will be able to move this reponsibility into a generic cell mapper. I am not sure what you mean about this pass not being useful since there happen to be plenty users using it to map to standard cells, and saving area/wiring in the process.

whitequark:
> I am not sure what you mean about this pass not being useful since there happen to be plenty users using it to map to standard cells, and saving area/wiring in the process.

It's solving a subgraph logic equivalence problem, which isn't a great one to find yourself solving. If you are finding yourself solving it you should fix your flow so that it doesn't lower your adders down to gates first, not to do this.

povik:
I disagree it's so clear-cut, the cell mapper essentially does the same job for all other cells, it only needs to be extended slightly to handle this multi-output case.

hovind:
> Thanks for the fix and for the included test case.

Thanks for reviewing and pushing the change forward!

PR #4699, opened 2024-11-04
pyosys: catch boost::python::error_already_set

georgerennie:
This catches exceptions from internal passes, printing them in a readable manner where the user would otherwise see an unspecified boost exception.

e.g. consider the following pass in `pass.py`:
```python
import libyosys as ys

class MyPass(ys.Pass):
    def __init__(self):
        super().__init__("my_pass", "")

    def py_help(self):
        raise Exception("broken")

p = MyPass()
```

Previously, when invoked with `yosys -m pass.py -p "help my_pass"`, yosys would error with

```
-- Running command `help my_pass' --
terminate called after throwing an instance of 'boost::python::error_already_set'
```

but with this change it instead shows

```
-- Running command `help my_pass' --
  File "<...>/pass.py", line 8, in py_help
    raise Exception("broken")
Exception: broken
ERROR: Python interpreter encountered an exception.
```

### DIFF
--- a/misc/py_wrap_generator.py
+++ b/misc/py_wrap_generator.py
@@ -1569,10 +1569,15 @@ class WFunction:
 		return_stmt = "return " if self.ret_type.name != "void" else ""
 
 		text += ")\n\t\t{"
-		text += "\n\t\t\tif (boost::python::override py_" + self.alias + " = this->get_override(\"py_" + self.alias + "\"))"
-		text += f"\n\t\t\t\t{return_stmt}" + call_string
-		text += "\n\t\t\telse"
+		text += "\n\t\t\tif (boost::python::override py_" + self.alias + " = this->get_override(\"py_" + self.alias + "\")) {"
+		text += "\n\t\t\t\ttry {"
+		text += f"\n\t\t\t\t\t{return_stmt}" + call_string
+		text += "\n\t\t\t\t} catch (boost::python::error_already_set &) {"
+		text += "\n\t\t\t\t\tlog_python_exception_as_error();"
+		text += "\n\t\t\t\t}"
+		text += "\n\t\t\t} else {"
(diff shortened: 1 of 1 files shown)


povik:
I think we should catch it here instead: https://github.com/YosysHQ/yosys/blob/b2d78589e2bf7f37cb89eeb85f859be4b17d9e8b/misc/py_wrap_generator.py#L1573

For two reasons:
 * the Yosys internals are not written to have exceptions propagating, so we should catch it as soon as we leave Python
 * this will cover uses which don't go through main(), i.e. libyosys, or calling Python passes from Python

povik:
To illustrate it a little better, the generated code looks like

```
		void py_execute(boost::python::list args, Design* design)
		{
			if (boost::python::override py_execute = this->get_override("py_execute"))
				py_execute(args, design); // xxx
			else
				Pass::py_execute(args, design);
		}
```
and `// xxx` marks the spot for catching the exception

georgerennie:
Ahh sure, the main reason I hadn't done that was because I didn't want to get too into the mess of `py_wrap_generator` :P but I ended up doing that anyway so im happy to change it so its done that way

georgerennie:
Have moved it into the generator. The generated c++ diff looks as follows with the other `py_*` functions taking the same form as `py_help`
```diff
43a44,48
> 	[[noreturn]] static void log_python_exception_as_error() {
> 		PyErr_Print();
> 		log_error("Python interpreter encountered an exception.\n");
> 	}
> 
447,449c452,458
< 			if (boost::python::override py_help = this->get_override("py_help"))
< 				py_help();
< 			else
---
> 			if (boost::python::override py_help = this->get_override("py_help")) {
> 				try {
> 					py_help();
> 				} catch (boost::python::error_already_set &) {
> 					log_python_exception_as_error();
> 				}
> 			} else {
450a460
> 			}
```

povik (approved):
Codewise looks good to me

PR #4705, opened 2024-11-05
Emphasise that read_verilog doesn't lint

KrystalDelusion:
Users are often confused and/or surprised when the `read_verilog` frontend runs into syntactical edge cases that it doesn't handle well.

Mention Verilator (and `verilator --lint-only`) in a few places and reiterate that `read_verilog` does not lint, and you should lint your code before trying to use it.

Preview available on [readthedocs](https://yosyshq.readthedocs.io/projects/yosys/en/docs-preview-lintonly/introduction.html#things-you-can-t-do).  Note that preview builds show todos, which are normally hidden.

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,6 +260,16 @@ The command ``prep`` provides a good default word-level synthesis script, as
 used in SMT-based formal verification.
 
 
+Additional information
+======================
+
+The ``read_verilog`` command, used by default when calling ``read`` with Verilog
+source input, does not perform any syntax checking.  You should instead lint
+your source with another tool such as
+[Verilator](https://www.veripool.org/verilator/) first, e.g. by calling
+``verilator --lint-only``.
+
+
 Unsupported Verilog-2005 Features
 =================================
 
--- a/README.md
+++ b/README.md
@@ -264,8 +264,8 @@ Additional information
 ======================
 
 The ``read_verilog`` command, used by default when calling ``read`` with Verilog
-source input, does not perform any syntax checking.  You should instead lint
-your source with another tool such as
+source input, does not perform syntax checking.  You should instead lint your
+source with another tool such as
 [Verilator](https://www.veripool.org/verilator/) first, e.g. by calling
 ``verilator --lint-only``.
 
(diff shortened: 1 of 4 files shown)


docs/source/code_examples/fifo/fifo.v
widlarizer, line 1: What's the relation to the docs change? Is this to make verilator linting happy?

widlarizer (approved):

docs/source/code_examples/fifo/fifo.v
KrystalDelusion: Yeah verilator gave lint warnings so I figured it was easy enough to fix them (and this wasn't a guided tutorial of how to fix it yourself)

KrystalDelusion: Although it does make the first image slightly more complex 
image
(left is preview, right is latest)

docs/source/code_examples/fifo/fifo.v
```diff
@@ -5,15 +5,15 @@ module addr_gen
...
+	initial addr = 0;
 
 	// async reset
 	// increment address when enabled
 	always @(posedge clk or posedge rst)
 		if (rst)
 			addr <= 0;
 		else if (en) begin
-			if (addr == MAX_DATA-1)
+			if ({'0, addr} == MAX_DATA-1)
```
nakengelhardt, line 16: I would not include this change, it's fine to get some linting messages from verilator IMO (gives people some idea of what to expect from verilator too).

README.md
ldoolitt, line 1: > does not perform any syntax checking

Is an exaggeration.  Maybe: "does not perform robust syntax checking"?

nakengelhardt: I see it more as a statement of intent. `read_verilog` doesn't try to check if the input is correct or not, it just proceeds assuming it is correct. If that turns out to be a wrong assumption it will error out, but often not in a way that is particularly useful for the user for finding the problem. Given the history of this code and the alternatives, we think it would be a waste of effort to change that.

PR #4719, opened 2024-11-07
add select option for write_json

AdamLee7:
there is already codes of select_module, but the write_json cmd don't receive the option.
now we can use the cmd like this:
```
select -module top;
write_json -selected xxx.json
```

### DIFF
--- a/backends/json/json.cc
+++ b/backends/json/json.cc
@@ -353,6 +353,9 @@ struct JsonBackend : public Backend {
 		log("        emit 32-bit or smaller fully-defined parameter values directly\n");
 		log("        as JSON numbers (for compatibility with old parsers)\n");
 		log("\n");
+		log("    -selected\n");
+		log("        output only select module\n");
+		log("\n");
 		log("\n");
 		log("The general syntax of the JSON output created by this command is as follows:\n");
 		log("\n");
@@ -597,6 +600,7 @@ struct JsonBackend : public Backend {
 	{
 		bool aig_mode = false;
 		bool compat_int_mode = false;
+		bool use_selection = false;
 
 		size_t argidx;
 		for (argidx = 1; argidx < args.size(); argidx++)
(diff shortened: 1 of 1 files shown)


KrystalDelusion:
Could you please add a test?  Loading a test design, `write_json`, reset design, `read_json`, then asserting the unselected portion is not included would be good.

AdamLee7:
[yosys_case.zip](https://github.com/user-attachments/files/17674214/yosys_case.zip)
this is a simple case, if use the -selected option,  write_json only output the  top module like synth_only_top.json

KrystalDelusion:
Sorry, I meant add a test to the `tests` directory with a commit on this PR. Something like `tests/aiger/neg.ys`:
```
...
write_aiger -map neg.out/neg.map neg.out/neg.aig

design -reset
read_aiger -wideports -map neg.out/neg.map neg.out/neg.aig
select -assert-count 1 i:a
select -assert-count 1 i:b
select -assert-count 1 o:c
select -assert-count 3 x:* s:64 %i
```

nakengelhardt:
This functionality already exists, it is the command `json -o xxx.json top`. I believe the two were deliberately separated, though I'm not familiar with the details.

nakengelhardt:
We discussed this in the dev meeting, the `json` command was for convenience (mostly for debugging) and by analogy with `dump`, but in general we do want `-selected` options on all backends (including this one), and we don't want pass variants like `dump` and `json` for other backends.

AdamLee7:
> This functionality already exists, it is the command `json -o xxx.json top`. I believe the two were deliberately separated, though I'm not familiar with the details of why.

this works, thanks, i'll closed this pr

nakengelhardt:
Nono, as I mentioned in my follow-up comment after asking around about the reasons for why it was this way, we do want to have this PR so that all the backends have the same consistent interface. It'll be easier for users to discover how to do this with -selected.

PR #4720, opened 2024-11-07
bufnorm: preserve constant bits driving wires

georgerennie:
Fixes an issue where constant driving bits in connections would be replaced with `Sx` by bufnorm


povik (commented):
I think we haven't done a proper review of `bufnorm` (with the excuse that it's marked experimental) so it's fine by me if we fast-track and merge

Issue #4713, opened 2024-11-06
Reduce compiler warnings when building yosys


ldoolitt:
### Feature Description

Building current main-branch yosys (specifically commit 2de9f00) with g++ 12.2.0, I get 45 warnings:

-  22 x maybe-uninitialized 
- 19 x sign-compare
- 4 x unused-but-set-variable

I can eliminate most (18) of the maybe-uninitialized with this two-line patch.
```
diff --git a/kernel/drivertools.h b/kernel/drivertools.h
index 079701c35..39d4bd7a7 100644
--- a/kernel/drivertools.h
+++ b/kernel/drivertools.h
@@ -364,7 +364,7 @@ public:
 
        unsigned int hash() const
        {
-               unsigned int inner;
+               unsigned int inner = 0;
                switch (type_)
                {
                        case DriveType::NONE:
@@ -912,7 +912,7 @@ public:
 
        unsigned int hash() const
        {
-               unsigned int inner;
+               unsigned int inner = 0;
                switch (type_)
                {
                        case DriveType::NONE:
```

Do y'all want a pull request?  Should I add a default: assert() to the two switch() statements?

KrystalDelusion:
The sign compares have been raised and should be in progress for fix. PR for this would be appreciated :) `type_` is assigned `DriveType::NONE` by default, so there _shouldn't_ be a need to assert on default, but it wouldn't hurt to check just in case.

PR #4721, opened 2024-11-08
kernel/drivertools.h: avoid maybe-uninitialized compile warnings

ldoolitt:
Initialize "unsigned int inner" in hash() functions
Includes a log_assert() that might help catch corrupted data structures or future incomplete modification of DriveType definition

See Issue #4713, Reduce compiler warnings when building yosys

### DIFF
--- a/kernel/drivertools.h
+++ b/kernel/drivertools.h
@@ -364,7 +364,7 @@ public:
 
 	unsigned int hash() const
 	{
-		unsigned int inner;
+		unsigned int inner = 0;
 		switch (type_)
 		{
 			case DriveType::NONE:
@@ -384,6 +384,9 @@ public:
 				break;
 			case DriveType::MULTIPLE:
 				inner = multiple_.hash();
+				break;
+			default:
+				log_assert(0);
 				break;
 		}
 		return mkhash((unsigned int)type_, inner);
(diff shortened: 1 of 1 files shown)


KrystalDelusion (approved):

KrystalDelusion:
Huh, this actually doesn't affect the number of warnings in the git actions.  I wonder if this is because we only test gcc-10 and gcc (which is apparently gcc-11) where your issue mentions gcc-12.  Could indicate that we could/should declare a newer gcc target.

kernel/drivertools.h
```diff
@@ -385,6 +385,9 @@ struct DriveBit
 			case DriveType::MULTIPLE:
 				inner = multiple_.hash();
 				break;
+			default:
+				log_assert(0);
```
KrystalDelusion, line 389: I think instead of `log_assert(0)` it might be better to `log_abort()`

ldoolitt:
> Huh, this actually doesn't affect the number of warnings in the git actions. I wonder if this is because we only test gcc-10 and gcc (which is apparently gcc-11) where your issue mentions gcc-12. Could indicate that we could/should declare a newer gcc target.

gcc-12 is default in Debian Stable, so not exactly bleeding-edge.  I'm starting to test code with gcc-14 in Debian Testing, which is likely to ship (as the new Stable) next Spring.

PR #4727, opened 2024-11-12
bufnorm: preserve constant bits when mapping back to connections

georgerennie:
Turns out I missed another case where constant bits were not being handled correctly in #4720. Have fixed this case and refactored the logic for it into a lambda


povik (approved):
For the same reason as in #4720 I am happy to rubber-stamp this

Issue #4732, opened 2024-11-12
Module parameters are not affected by setundef pass


kamilrakoczy:
### Version

Yosys 0.47+22 (git sha1 cef87cc17, g++ 12.2.0-14 -fPIC -O3)

### On which OS did this happen?

Linux

### Reproduction Steps

Process SystemVerilog code (a.sv):

```
module foo #(parameter [1:0] a) (output [1:0] o);
	assign o = a;
endmodule

module top(output [1:0] o);
	foo #(2'b0x) foo(o);
endmodule
```
with following commands:

```
read_verilog -sv a.sv
setundef -zero -params
write_json out.json
```

### Expected Behavior

File `out.json` should contain entry: `"parameters": { "$1": "00" }` responsible for `foo`'s `a` parameter.

### Actual Behavior

File `out.json` contains entry: `"parameters": { "$1": "0x" }` responsible for `foo`'s `a` parameter.

PR #4733, opened 2024-11-12
Fix setting bits of parameters in setundef pass

kamilrakoczy:

This PR fixes setting bits of parameters in `setundef` pass.

Currently we are looping through module parameters using `for (auto bit : parameter.second)`.
It is equivalent to:
```
for( auto _bit = parameter.second.begin(); _bit != parameter.second.end(); _bit++) {
    auto bit = *_bit;
    ...
}
```
where `auto` is resolved to [`RTLIL::Const`](https://github.com/YosysHQ/yosys/blob/915df16c84f4595b6ad5f198de0f7aaf43896723/kernel/rtlil.h#L1677C2-L1677C49). 

Iterator that we are using is defined in [`Const` class](https://github.com/YosysHQ/yosys/blob/915df16c84f4595b6ad5f198de0f7aaf43896723/kernel/rtlil.h#L713) which iterates through [`State` enum](https://github.com/YosysHQ/yosys/blob/6d4f056a353faeaa69dc23bb30c1ea05f38aae80/kernel/rtlil.h#L30). 

`operator *` of iterator returns [copy of the state](https://github.com/YosysHQ/yosys/blob/915df16c84f4595b6ad5f198de0f7aaf43896723/kernel/rtlil.cc#L543).

This PR changes to use reference instead.

I attached example test case that showcase the problem.

Fixes: https://github.com/YosysHQ/yosys/issues/4732

### DIFF
--- a/passes/cmds/setundef.cc
+++ b/passes/cmds/setundef.cc
@@ -243,7 +243,7 @@ struct SetundefPass : public Pass {
 			{
 				for (auto *cell : module->selected_cells()) {
 					for (auto &parameter : cell->parameters) {
-						for (auto bit : parameter.second) {
+						for (auto &bit : parameter.second.bits()) {
 							if (bit > RTLIL::State::S1)
 								bit = worker.next_bit();
 						}
--- a/tests/various/setundef.sv
+++ b/tests/various/setundef.sv
@@ -1,0 +1,10 @@
+module foo #(parameter [1:0] a) (output [1:0] o);
+	assign o = a;
+endmodule
+
+module top(output [1:0] o);
+	foo #(2'b0x) foo(o);
+	always_comb begin
+		assert(o == 2'b00);
+	end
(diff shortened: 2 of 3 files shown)


passes/cmds/setundef.cc
```diff
@@ -243,7 +243,7 @@ struct SetundefPass : public Pass {
 			{
 				for (auto *cell : module->selected_cells()) {
 					for (auto &parameter : cell->parameters) {
-						for (auto bit : parameter.second) {
+						for (auto &bit : parameter.second.bits()) {
```
KrystalDelusion, line 246: This is what the line was prior to #4516 
@widlarizer can you check this?

widlarizer: Thanks for the ping, it's a case just like #4951 where I broke something and didn't notice partly due to missing test coverage

PR #4734, opened 2024-11-12
Fix bug when setting Verific runtime string  flags.

mikesinouye:
The setting of Verific runtime string flags does not work within Yosys as it encapsulates the value in quotes.

Temp string s that is preexisting is correctly passed into Verific RuntimeFlags call. 

Attempting to change the value of a Verific string RuntimeFlag will fail before this change.


povik (approved):
Looking at the different `verific -cfg` modes it's clear passing in the stripped string was the intention, and the correct thing to do

mikesinouye:
Thanks Martin

povik:
Sure Mike, I am not the maintainer of the Verific integration but here it's clear enough

PR #4736, opened 2024-11-13
portarcs: New command to derive propagation arcs

povik:

To preserve boundaries of arithmetic operators we need to synthesize them separately and box them for synthesis of the surrounding logic. We need to characterize the propagation delay of this box so the technology mapper is aware of it, this characterization step is where this new command comes in.

Help sample:
```

    portarcs [options] [selection]

This command characterizes the combinational content of selected modules and
derives timing arcs going from module inputs to module outputs representing the
propagation delay of the module.

    -draw
        plot the computed delay table to the terminal

    -icells
        assign unit delay to gates from the internal Yosys cell library

    -write
        write the computed arcs back into the module as $specify2 instances

```

### DIFF
--- a/kernel/timinginfo.h
+++ b/kernel/timinginfo.h
@@ -37,6 +37,14 @@ struct TimingInfo
 		bool operator==(const NameBit& nb) const { return nb.name == name && nb.offset == offset; }
 		bool operator!=(const NameBit& nb) const { return !operator==(nb); }
 		unsigned int hash() const { return mkhash_add(name.hash(), offset); }
+		std::optional<SigBit> get_connection(RTLIL::Cell *cell) {
+			if (!cell->hasPort(name))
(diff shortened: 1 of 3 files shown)


povik:
Sample plot of a mapped `$alu` (columns inputs, rows outputs):

Screenshot 2024-11-13 at 16 24 48

Ravenslofty:
To review #4737 I should probably look at what this does.

My first question is: where's the timing data coming from? The `timinginfo.h` stuff deals with Verilog specify rules, and this is intended for use in ASIC flows, right? Does this assume the gates have already been STA'd and turned into specify rules for Yosys to parse?

Ravenslofty:
This is maybe a little big to Just Merge It, but I don't think there's much to discuss/dispute, so whatever.

PR #4737, opened 2024-11-13
Support `abc9_box` on ordinary modules in abc_new

povik:

To preserve boundaries of arithmetic operators, we want to isolate them into a new module which will become part of the design hierarchy. Ordinarily this would mean segmenting the timing paths for the technology mapper within ABC, but with new changes to `abc_new` we support having those be abc9 boxes with their propagation arcs described to the mapper.

Internally this calls the new `portarcs` command (PR #4736) once the module to be boxed has been technology-mapped.

In addition this PR allows applying different ABC scripts per module.


povik:
Rebased to resolve merge conflict

PR #4739, opened 2024-11-13
Include stdlib.h for `abort()`

hzeller:
Simple cleanup notices by newer toolchains.


hzeller:
Anything else to do ?

povik:
Thanks!

PR #4740, opened 2024-11-13
Add `muldiv_c` and `muxadd` peepopts

akashlevy:

This PR adds two peep-hole optimizations that improve netlist quality:

1. `muldiv_c`: `y = (a * b_const) / c_const   ===>   a * eval(b_const / c_const)` if `b_const % c_const == 0`. This implements basic constant propagation for multipliers and dividers with divisible constants. 
2. `muxadd`: `y = s ? (a + b) : a   ===>   y = a + (s ? b : 0)`. This provides useful restructuring for improving logic optimization. `s ? b : 0` can be optimized to a bitwise AND in subsequent optimizations.

- [x] `passes/pmgen/peepopt_muldiv_c.pmg`: `muldiv_c` peepopt
- [x] `passes/pmgen/peepopt_muxadd.pmg`: `muxadd` peepopt
- [x] `passes/pmgen/peepopt.cc`: Add to docs and run peepopts during `peepopt` pass
- [x] `passes/pmgen/Makefile.inc`: Add the new sources

- [ ] YosysHQ to review source code and provide feedback/edits as necessary
- [ ] YosysHQ to construct test plan of 15-20 small-medium test cases per peepopt
- [ ] Silimate to review test plan and sign off
- [ ] YosysHQ to write test cases according to test plan and add to regression
- [ ] YosysHQ to provide formal equivalence checking scripts with `equiv_opt` or FOSS `eqy`
- [ ] Silimate to sign off on implementation

### DIFF
--- a/passes/pmgen/peepopt_muldiv_c.pmg
+++ b/passes/pmgen/peepopt_muldiv_c.pmg
@@ -1,0 +1,76 @@
+pattern muldiv_c
+//
+// Authored by Akash Levy of Silimate, Inc. under ISC license.
+// Transforms mul->div into const->mul when b and c are divisible constants:
+// y = (a * b_const) / c_const   ===>   a * eval(b_const / c_const)
+//
+
+state <SigSpec> a b_const mul_y
+
+match mul
+	// Select multiplier
+	select mul->type == $mul
+endmatch
+
+code a b_const mul_y
+	// Get multiplier signals
+	a = port(mul, \A);
+	b_const = port(mul, \B);
+	mul_y = port(mul, \Y);
+
+	// Fanout of each multiplier Y bit should be 1 (no bit-split)
+	for (auto bit : mul_y)
+		if (nusers(bit) != 2)
+			reject;
+
+	// A and B can be interchanged
+	branch;
+	std::swap(a, b_const);
+endcode
+
+match div
+	// Select div of form (a * b_const) / c_const
+	select div->type == $div
+
+	// Check that b_const and c_const is constant
+	filter b_const.is_fully_const()
+	filter port(div, \B).is_fully_const()
+endmatch
+
+code
+	// Get div signals
+	SigSpec div_a = port(div, \A);
+	SigSpec c_const = port(div, \B);
+	SigSpec div_y = port(div, \Y);
+
+	// Get offset of multiplier result chunk in divider
+	int offset = GetSize(div_a) - GetSize(mul_y);
+
+	// Get properties and values of b_const and c_const
+	int b_const_width = mul->getParam(ID::B_WIDTH).as_int();
+	bool b_const_signed = mul->getParam(ID::B_SIGNED).as_bool();
+	bool c_const_signed = div->getParam(ID::B_SIGNED).as_bool();
+	int b_const_int = b_const.as_int(b_const_signed);
+	int c_const_int = c_const.as_int(c_const_signed);
+	int b_const_int_shifted = b_const_int << offset;
+
+	// Check that there are only zeros before offset
+	if (offset < 0 || !div_a.extract(0, offset).is_fully_zero())
+		reject;
+
+	// Check that b is divisible by c
+	if (b_const_int_shifted % c_const_int != 0)
+		reject;
+
+	// Rewire to only keep multiplier
+	mul->setPort(\B, Const(b_const_int_shifted / c_const_int, b_const_width));
+	mul->setPort(\Y, div_y);
+
+	// Remove divider
+	autoremove(div);
+
+	// Log, fixup, accept
+	log("muldiv_const pattern in %s: mul=%s, div=%s\n", log_id(module), log_id(mul), log_id(div));
+	mul->fixup_parameters();
+	accept;
+endcode
--- a/passes/pmgen/peepopt_muxadd.pmg
+++ b/passes/pmgen/peepopt_muxadd.pmg
@@ -1,0 +1,57 @@
+pattern muxadd
+//
+// Authored by Akash Levy of Silimate, Inc. under ISC license.
+// Transforms add->mux into mux->add:
+// y = s ? (a + b) : a   ===>   y = a + (s ? b : 0)
+//
+
+state <SigSpec> add_a add_b add_y
+
+match add
+	// Select adder
+	select add->type == $add
+endmatch
+
+code add_y add_a add_b
+	// Get adder signals
+	add_a = port(add, \A);
+	add_b = port(add, \B);
+	add_y = port(add, \Y);
+
+	// Fanout of each adder Y bit should be 1 (no bit-split)
+	for (auto bit : add_y)
+		if (nusers(bit) != 2)
+			reject;
+
+	// A and B can be interchanged
+	branch;
+	std::swap(add_a, add_b);
+endcode
+
+match mux
+	// Select mux of form s ? (a + b) : a, allow leading 0s when A_WIDTH != Y_WIDTH
+	select mux->type == $mux
+	index <SigSpec> port(mux, \A) === SigSpec({Const(State::S0, GetSize(add_y)-GetSize(add_a)), add_a})
+	index <SigSpec> port(mux, \B) === add_y
+endmatch
+
+code
+	// Get mux signal
+	SigSpec mux_y = port(mux, \Y);
+
+	// Create new mid wire
+	SigSpec mid = module->addWire(NEW_ID, GetSize(add_b));
+
+	// Rewire
+	mux->setPort(\A, Const(State::S0, GetSize(add_b)));
+	mux->setPort(\B, add_b);
+	mux->setPort(\Y, mid);
+	add->setPort(\B, mid);
+	add->setPort(\Y, mux_y);
+
+	// Log, fixup, accept
+	log("muxadd pattern in %s: mux=%s, add=%s\n", log_id(module), log_id(mux), log_id(add));
+	add->fixup_parameters();
+	mux->fixup_parameters();
+	accept;
+endcode
--- a/passes/pmgen/Makefile.inc
+++ b/passes/pmgen/Makefile.inc
@@ -57,6 +57,8 @@ PEEPOPT_PATTERN  = passes/pmgen/peepopt_shiftmul_right.pmg
 PEEPOPT_PATTERN += passes/pmgen/peepopt_shiftmul_left.pmg
 PEEPOPT_PATTERN += passes/pmgen/peepopt_shiftadd.pmg
 PEEPOPT_PATTERN += passes/pmgen/peepopt_muldiv.pmg
+PEEPOPT_PATTERN += passes/pmgen/peepopt_muldiv_c.pmg
+PEEPOPT_PATTERN += passes/pmgen/peepopt_muxadd.pmg
 PEEPOPT_PATTERN += passes/pmgen/peepopt_formal_clockgateff.pmg
 
 passes/pmgen/peepopt_pm.h: passes/pmgen/pmgen.py $(PEEPOPT_PATTERN)
--- a/passes/pmgen/peepopt.cc
+++ b/passes/pmgen/peepopt.cc
@@ -42,7 +42,11 @@ struct PeepoptPass : public Pass {
 		log("\n");
 		log("This pass employs the following rules by default:\n");
 		log("\n");
+		log("   * muxadd - Replace S?(A+B):A with A+(S?B:0)\n");
+		log("\n");
 		log("   * muldiv - Replace (A*B)/B with A\n");
+		log("\n");
+		log("   * muldiv_c - Replace (A*B)/C with A*(B/C) when C is a const divisible by B.\n");
 		log("\n");
 		log("   * shiftmul - Replace A>>(B*C) with A'>>(B<<K) where C and K are constants\n");
 		log("                and A' is derived from A by appropriately inserting padding\n");
@@ -105,6 +109,8 @@ struct PeepoptPass : public Pass {
 					pm.run_shiftmul_right();
 					pm.run_shiftmul_left();
 					pm.run_muldiv();
+					pm.run_muldiv_c();
+					pm.run_muxadd();
 				}
 			}
 		}


passes/pmgen/peepopt_muxadd.pmg
```diff
@@ -0,0 +1,57 @@
...
+// Authored by Akash Levy of Silimate, Inc. under ISC license.
+// Transforms add->mux into mux->add:
+// y = s ? (a + b) : a   ===>   y = a + (s ? b : 0)
+//
+
+state <SigSpec> add_a add_b add_y
+
+match add
+	// Select adder
+	select add->type == $add
```
povik, line 12:
```suggestion
	select add->type == $add
	choice <IdString> A {\A, \B}
        define <IdString> B (A == \A ? \B : \A)
        set add_y port(add, \Y)
        set add_a port(add, A)
        set add_b port(add, B)
        set add_a_signed param(add, (A == \A) ? \A_SIGNED : \B_SIGNED))
        set add_a_id A
```

and replace the branch-and-swap pattern further below

Requires new state
```
state <bool> add_a_signed
state <IdString> add_a_id
```

passes/pmgen/peepopt_muxadd.pmg
```diff
@@ -0,0 +1,57 @@
...
+	SigSpec mux_y = port(mux, \Y);
+
+	// Create new mid wire
+	SigSpec mid = module->addWire(NEW_ID, GetSize(add_b));
+
+	// Rewire
+	mux->setPort(\A, Const(State::S0, GetSize(add_b)));
+	mux->setPort(\B, add_b);
+	mux->setPort(\Y, mid);
+	add->setPort(\B, mid);
```
povik, line 49: Wrong side of the adder if the swap above was hit (we need to use `add_a_id` instead)

povik (changes requested):
Hi Akash, I looked over the muxadd pattern

passes/pmgen/peepopt_muxadd.pmg
povik: Err, more like `add_b_id`

passes/pmgen/peepopt_muxadd.pmg
```diff
@@ -0,0 +1,57 @@
...
+
+	// A and B can be interchanged
+	branch;
+	std::swap(add_a, add_b);
+endcode
+
+match mux
+	// Select mux of form s ? (a + b) : a, allow leading 0s when A_WIDTH != Y_WIDTH
+	select mux->type == $mux
+	index <SigSpec> port(mux, \A) === SigSpec({Const(State::S0, GetSize(add_y)-GetSize(add_a)), add_a})
```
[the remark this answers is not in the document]
povik: Also I am not sure this doesn't crash on `add_y < add_a` which I don't think we rule out anywhere

akashlevy:
Thanks for the feedback on `muxadd`. I'll take a look today

passes/pmgen/peepopt_muldiv_c.pmg
```diff
@@ -0,0 +1,76 @@
...
+	for (auto bit : mul_y)
+		if (nusers(bit) != 2)
+			reject;
+
+	// A and B can be interchanged
+	branch;
+	std::swap(a, b_const);
+endcode
+
+match div
```
povik, line 31: No check that the divider and multiplier are connected

passes/pmgen/peepopt_muxadd.pmg
```diff
@@ -0,0 +1,57 @@
...
+	for (auto bit : add_y)
+		if (nusers(bit) != 2)
+			reject;
+
+	// A and B can be interchanged
+	branch;
+	std::swap(add_a, add_b);
+endcode
+
+match mux
```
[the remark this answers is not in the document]
akashlevy: This is added now

passes/pmgen/peepopt_muxadd.pmg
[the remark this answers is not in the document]
akashlevy: This should be resolved using `SigSpec`'s `extend_u0`. Please resolve if this looks good

passes/pmgen/peepopt_muxadd.pmg
akashlevy: This didn't compile... we adjusted this with `state <IdString> add_a_id`

akashlevy: Please resolve if this looks ok now

passes/pmgen/peepopt_muldiv_c.pmg
akashlevy: I just pulled in https://github.com/akashlevy/yosys/pull/2 which should resolve this. Let me know if this looks good

passes/pmgen/peepopt_muldiv_c.pmg
```diff
@@ -0,0 +1,83 @@
...
+	// Get offset of multiplier result chunk in divider
+	int offset = GetSize(div_a) - GetSize(mul_y);
+
+	// Get properties and values of b_const and c_const
+	int b_const_width = mul->getParam(ID::B_WIDTH).as_int();
+	bool b_const_signed = mul->getParam(ID::B_SIGNED).as_bool();
+	bool c_const_signed = div->getParam(ID::B_SIGNED).as_bool();
+	int b_const_int = b_const.as_int(b_const_signed);
+	int c_const_int = c_const.as_int(c_const_signed);
+	int b_const_int_shifted = b_const_int << offset;
```
povik, line 55: The `as_int()` conversion may overflow if the operands are too wide. We should catch this case and bail

passes/pmgen/peepopt_muldiv_c.pmg
```diff
@@ -0,0 +1,83 @@
...
+	SigSpec div_a = port(div, \A);
+	SigSpec c_const = port(div, \B);
+	SigSpec div_y = port(div, \Y);
+
+	// Get offset of multiplier result chunk in divider
+	int offset = GetSize(div_a) - GetSize(mul_y);
+
+	// Get properties and values of b_const and c_const
+	int b_const_width = mul->getParam(ID::B_WIDTH).as_int();
+	bool b_const_signed = mul->getParam(ID::B_SIGNED).as_bool();
```
povik, line 51: `b_const` may be coming from the A port. It's an RTLIL invariant that `A_SIGNED` equals `B_SIGNED`, so the code is fine, but we should leave a comment so it doesn't look like a bug

passes/pmgen/peepopt_muldiv_c.pmg
```diff
@@ -0,0 +1,83 @@
...
+		reject;
+
+	// Check that every single output bit of the multiplier is connected consecutively to the div operator (offset accounted for)
+	for (int i = 0; i < mul_y.size(); i++) {
+		if (sigmap(mul_y[i]) != sigmap(div_a[i + offset])) {
+			reject;
+		}
+	}
+
+	// Rewire to only keep multiplier
```
povik, line 72: No check for overflows on the original multiplier output

passes/pmgen/peepopt_muldiv_c.pmg
```diff
@@ -0,0 +1,83 @@
...
+	mul->setPort(\B, Const(b_const_int_shifted / c_const_int, b_const_width));
```
povik, line 73: Because `b_const_int_shifted` is shifted up, there's no guarantee it fits into `b_const_width`

Also this might need special care wrt signedness. `B_SIGNED` on the multiplier may be false but `b_const_int_shifted / c_const_int` can be negative due to the divider. I wouldn't be opposed to requiring matching signedness on the divider and multiplier to make the transformation a little easier to reason about

povik (changes requested):

widlarizer:
I've added an implementation of muldiv_c tests approximately as specified, see comments in the spec doc. Please pull and resolve the failures and feel free to ask for more context

passes/pmgen/pmgen.py
```diff
@@ -429,6 +429,13 @@ def process_pmgfile(f, filename):
     print("  }", file=f)
     print("", file=f)
 
+    print("  SigSpec remove_bottom_padding(SigSpec sig) {", file=f)
```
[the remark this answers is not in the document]
alaindargelas: @povik , https://github.com/YosysHQ/yosys/blob/main/passes/pmgen/peepopt.cc does not contain any other helper code. 
All the helper code is auto-generated here:
https://github.com/YosysHQ/yosys/blob/7a362f1f7496be9ece7cabd852dc0e2b5b64e91c/passes/pmgen/pmgen.py#L424
I followed the same pattern. Why would we want to deviate from the exiting pattern for helper code?

passes/pmgen/peepopt_muxadd.pmg
```diff
@@ -0,0 +1,72 @@
...
+	index <SigSpec> port(mux, \B) === add_y
+endmatch
+
+code
+	// Get mux signal
+	SigSpec mux_y = port(mux, \Y);
+	SigSpec mux_a = port(mux, \A);
+	SigSpec mux_b = port(mux, \B);
+	// Prevent multiple branches to edit the same pair of mux/add
+	if (mux_b == add_y || mux_a == add_y) {
```
alaindargelas, line 54: @povik, I had to add this if statement, else the code block was executed 4 times with this test:
log -header "Test basic s?(a+b):a pattern with intermediate var gets transformed (a,b module inputs)"

Is there a better way to prevent multiple branches to execute their final code transformation when a branch has performed successfully the transform? Shall I use the "finish" macro? But I don't see it used in any example.

tests/peepopt/muldiv_c.ys
```diff
@@ -0,0 +1,258 @@
...
+module top(
+	input signed [3:0] a,
+	output signed [7:0] y,
+);
+    wire signed [7:0] mul;
+	assign mul = a * 4'sd6;
+	assign y = mul / 8'sd3;
+endmodule
+EOT
+equiv_opt -assert peepopt
```
alaindargelas, line 12: @widlarizer please convert all equiv_opt to the miter/sat technique.
equiv_opt does not converge if internal nets are removed/replaced, where miter/sat does.
See example in the file in this PR:
tests/peepopt/multdiv_c.ys

passes/pmgen/peepopt_muldiv_c.pmg
alaindargelas: Fixed

passes/pmgen/peepopt_muldiv_c.pmg
alaindargelas: Fixed

tests/peepopt/muldiv_c.ys
georgerennie: equiv_opt uses the same sat solver invocations under the hood so this shouldn't be correct. equiv_opt should be preferred over manual sat invocations wherever possible as it is much less prone to errors where the wrong arguments are selected (e.g. see #4629). You will see that if you use the miter based commands you suggest for these tests they still fail, so I suspect the peepopt is being applied incorrectly in those cases.

tests/peepopt/multdiv_c.ys
```diff
@@ -0,0 +1,49 @@
+log -header "Test simple positive case"
+log -push
+design -reset
+read_verilog <<EOF
+module top (
+  input wire  [31:0] a,
+  output wire  [63:0] y
```
georgerennie, line 7: proving equivalence of large multipliers with a plain sat solver is generally gonna be slow so if possible id suggest here just using slightly smaller bitwidths, or maybe consider putting these two tests in separate test files that can be run in parallel.

tests/peepopt/muldiv_c.ys
```diff
@@ -0,0 +1,258 @@
...
+	output signed [7:0] y,
+);
+    wire signed [6:0] mul;
+	assign mul = a * 4'sd6;
+	assign y = mul / 8'sd3;
+endmodule
+EOT
+equiv_opt -assert peepopt
+design -load postopt
+write_verilog
```
georgerennie, line 167: Remove debugging statement

tests/peepopt/muldiv_c.ys
widlarizer: I've ran into this symptom as well, so I've switched temporarily to some hack, and when I removed it again, there was no problem. This would match with what George is saying, I've been happy with `equiv_opt -assert`

tests/peepopt/muldiv_c.ys
widlarizer: oops, I checked I didn't leave any show or dump but not write_verilog

tests/peepopt/multdiv_c.ys
alaindargelas: @georgerennie This case takes <30 seconds as-is, I'll reduce the size.

tests/peepopt/muldiv_c.ys
alaindargelas: @georgerennie the problem with equiv_opt is that it tries to prove internal signals that don't exist anymore after the peepopt. The miter/sat invocation only operates on primary outputs and passes.
Your statement "You will see that if you use the miter based commands you suggest for these tests they still fail" is incorrect. I ran all the PRs I submitted for the test I edited. They pass. I even injected errors to make sure the miter/sat catches them and it did.
Can you provide a proof that any of the tests submitted actually pass with the equiv_opt? For me they all fail on internal signals. Is there any option for it to consider only the primary outputs? Else these transforms all change the internal structure too much to be proven.

tests/peepopt/multdiv_c.ys
widlarizer: 10-30 seconds is extreme. If you can exercise the same codepaths in 300ms instead by reducing the bit widths, there's no reason not to do so. Or you can duplicate the test and then only run formal on the easy version and only check an effect of doing/not doing the transform in the hard version. I didn't find any issues stretching the pass with 4 bit multiplication and 8 bit division

tests/peepopt/muldiv_c.ys
```diff
@@ -0,0 +1,257 @@
...
+module top(
+	input signed [3:0] a,
+	output signed [7:0] y,
+);
+    wire signed [7:0] mul;
+	assign mul = a * 4'sd4;
+	assign y = mul / 8'sd0;
+endmodule
+EOT
+equiv_opt peepopt
```
georgerennie, line 133: missing -assert

tests/peepopt/muldiv_c.ys
georgerennie: `equiv_opt` will match on internal nets, but only ones that exist and have a driver on both the pre and post optimized netlists. Optimization passes should not change the semantics of these nets (with the exception of refining undefined values) and im pretty sure these peephole obey that.

> Can you provide a proof that any of the tests submitted actually pass with the equiv_opt?

On https://github.com/georgerennie/yosys/tree/george/muldiv_c_equiv_opt I have taken the latest commit of this PR (1c774f4) and rewritten `multdiv_c.ys` to use `equiv_opt` instead of manual commands. This passes on my machine (and in less time, 19s vs 24s).

On that branch I have also commented out all the currently failing tests from `muldiv_c.ys`. All remaining tests pass using `equiv_opt`, and if you uncomment others you will see that they still fail using the miter/sat invocation, I suspect because the optimization is being applied incorrectly. I have put comments starting with `# GEORGE` in that file with some other notes.

georgerennie: Ahh I see now that I missed the muxadd testcases. These do indeed fail with `equiv_opt`, but I believe that is a fault of the pass, e.g.
```verilog
wire [3:0] ab = a + b;
assign y = s ? ab : a;
```
is rewritten to something like
```verilog
wire [3:0] _0_ = a + (s ? ab : a);
wire [3:0] ab = _0_;
assign y = _0_;
```

This is not a valid rewrite application as it changes the meaning of `ab`, even though you only apply the rewrite in cases where `ab` has no other fanout. This is problematic for usecases like formal verification where suddenly an internal net shows incorrect values in a trace. To get around this you should probably just leave the original `ab = a + b` in the circuit. If we are doing synthesis it will be trivially cleaned up by `opt_clean` but it may be useful for other users if it remains.

tests/peepopt/multdiv_c.ys
alaindargelas: Fixed, below a second for each test (2 in total).
I also changed back to equiv_opt as this transform keep the logical wires intact and equiv_opt indeed works here.

passes/pmgen/peepopt_muldiv_c.pmg
alaindargelas: Fixed

passes/pmgen/peepopt_muldiv_c.pmg
alaindargelas: @povik should I change to: mul->setPort(\B, Const(b_const_int_shifted / c_const_int, b_const_width + offset));

passes/pmgen/peepopt_muxadd.pmg
```diff
@@ -0,0 +1,68 @@
...
+	add_y = port(add, \Y);
+	add_a_ext = SigSpec(add_a);
+	add_a_ext.extend_u0(GetSize(add_y), param(add, \A_SIGNED).as_bool());
+	// Fanout of each adder Y bit should be 1 (no bit-split)
+	if (nusers(add_y) != 2)
+		reject;
+
+	// A and B can be interchanged
+	branch;
+	std::swap(add_a, add_b);
```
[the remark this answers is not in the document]
alaindargelas: Fixed

passes/pmgen/peepopt_muxadd.pmg
```diff
@@ -0,0 +1,68 @@
...
+	set add_a_id A
+endmatch
+
+code add_y add_a add_b add_a_ext
+	// Get adder signals
+	add_a = port(add, \A);
+	add_b = port(add, \B);
+	add_y = port(add, \Y);
+	add_a_ext = SigSpec(add_a);
+	add_a_ext.extend_u0(GetSize(add_y), param(add, \A_SIGNED).as_bool());
```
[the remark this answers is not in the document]
alaindargelas: @povik, please verify: 
```
       << add_a_ext = SigSpec(add_a);
       << add_a_ext.extend_u0(GetSize(add_y), param(add, \A_SIGNED).as_bool());
Changed to:
       >> add_a_ext = SigSpec(port(add, add_a_id));
       >> add_a_ext.extend_u0(GetSize(add_y), add_a_signed.as_bool());
```

passes/pmgen/pmgen.py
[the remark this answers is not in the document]
povik: This new helper is not as generic as the existing built-in helpers of the generator. The generator is used in other places and also for plugins, if we introduce a new function we need to think of backward compatiblity.

passes/pmgen/peepopt_muldiv_c.pmg
povik: Thanks

passes/pmgen/peepopt_muldiv_c.pmg
povik: I think `b_const_int << offset` can still overflow

passes/pmgen/peepopt_muldiv_c.pmg
```diff
@@ -48,12 +48,26 @@ code
...
 	bool c_const_signed = div->getParam(ID::B_SIGNED).as_bool();
 	int b_const_int = b_const.as_int(b_const_signed);
 	int c_const_int = c_const.as_int(c_const_signed);
 	int b_const_int_shifted = b_const_int << offset;
 
+	// Integer values should be lesser than 64 bits
+	if (mul->getParam(ID::B_WIDTH).size() > 64)
+		reject;
+	if (b_const.size() > 64)
+		reject;
```
povik, line 62: I keep forgetting this but I think the typical size of int (and the standard mandated minimum) is 32 bits, also let's check `b_const.size() + offset`

passes/pmgen/peepopt_muldiv_c.pmg
povik: That would work but I suggest constructing the Const with 32 bits (the constructor default) and then calling compress(/signedness/) to fit to the constant. The reason for this is: unless `wreduce` is sequenced after `peepopt` I am not sure any pass would shrink the operand before we bitblast the multiplier

passes/pmgen/peepopt_muxadd.pmg
povik: Blacklist works for this but isn't the underlying cause that you have both: the `branch; std::swap()` code and `choice <IdString> A {\A, \B}` on the adder match block? I may not be looking at the right revision you seem to have both

tests/peepopt/muldiv_c.ys
povik: I think renaming the signal (to a new private name) is also fine when it no longer has the same meaning and we know there are no external fanouts. From a practical standpoint It's desirable for passes to be verifiable with equiv_opt because that's the go-to tool for tracking down correctness bugs over full designs and flows.

passes/pmgen/peepopt_muxadd.pmg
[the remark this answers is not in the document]
povik: This looks good

tests/peepopt/muldiv_c.ys
georgerennie: Yeah this works too and I would be happy with that

alaindargelas:
@widlarizer tests merged and multdiv removed

passes/pmgen/pmgen.py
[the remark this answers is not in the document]
alaindargelas: Moved to peepopt.cc

passes/pmgen/peepopt_muldiv_c.pmg
alaindargelas: Implemented here: https://github.com/akashlevy/yosys/pull/14/commits/8b08f81d192b085d200ed59b4d352c0b26a10b80
and 
https://github.com/akashlevy/yosys/pull/14/commits/8687e5f4fb2cce36c61df1097d48e5f336277a02

passes/pmgen/peepopt_muldiv_c.pmg
alaindargelas: Done here: https://github.com/akashlevy/yosys/pull/13

tests/peepopt/muldiv_c.ys
alaindargelas: Fixed here: https://github.com/akashlevy/yosys/pull/15

passes/pmgen/peepopt_muxadd.pmg
alaindargelas: Correct removed the bogus branch, fixed here: https://github.com/akashlevy/yosys/pull/15

passes/pmgen/peepopt_muldiv_c.pmg
alaindargelas: Not sure how to proceed, a test specifically designed for this condition will help.

povik: As long as you test `b_const_int.size() + offset <= 31` you should be fine

povik: Err, `b_const.size() + offset`

akashlevy:
@povik @georgerennie @widlarizer Can we split each test into a different file so that we can run each test individually? Right now, the script stops every time a test fails, so we can't easily tell how many tests are passing/failing.

Additionally, now that we have a (somewhat) workable pass, it would be great if you can analyze any test failures and ensure that the test is not the problem.

widlarizer:
It's a little bit more convenient to keep it in one file for long term maintanance. These are regression tests so they're meant to all pass in sequence and stopping on the first failure is fine. For development you could copy and chop them up and create a temporary directory with them that isn't tracked or is temporarily tracked under version control

widlarizer:
The only failing case I see is a missed optimization: `(a*(-6))/3 -> a*(-2)` which would be valid according to my testing. I think this peepopt should handle this case as well

alaindargelas:
@widlarizer I made a fix here: https://github.com/akashlevy/yosys/pull/19, but I also have to increase one constant to actually overflow. Since we are looking at the actual value, not the declared size: 4'd4 only takes 3 bits, 4'd8 takes 4 bits and indeed overflow and get caught.

passes/pmgen/peepopt_muldiv_c.pmg
```diff
@@ -0,0 +1,124 @@
...
+	if (c_const_int == 0)
+		reject;
+	b_const_actual = b_const_int_shifted;
+	b_const_actual.compress(b_const_signed);
+
+	const_ratio = b_const_int_shifted / c_const_int;
+	const_ratio.compress(b_const_signed | c_const_signed);
+
+	// Integer values should be lesser than 64 bits
+	// This is because we are using C++ types, and int is 64 bits
```
povik, line 92: It's not except for obscure ILP64 platforms. See https://en.cppreference.com/w/cpp/language/types

akashlevy: So should the check be 32 bit?

povik: Yes, please make it 32 bits

akashlevy: Ok. Will fix this. I think a future modification could be to use the `BigInteger` library or at the very least `long long` to get this to 64 bits. But there's no good way to get an RTLIL const as a `long long` or `BigInteger` at the moment

akashlevy:
@widlarizer @povik @georgerennie I just rebased. All of the tests appear to be passing, and I think we have addressed all your review comments. Can you confirm that everything looks ok on your end now?

alaindargelas:
Note:
For test tests/arch/quicklogic/qlf_k6n10f/dsp.ys
The muxadd pass prevents Quicklogic inference to map to a QL_DSP2_MULTACC.
Once transformed by this peepopt, the particular inference pattern probably does not match anymore the structure.
PR: https://github.com/akashlevy/yosys/pull/22

passes/pmgen/peepopt_muldiv_c.pmg
alaindargelas: @povik is this fix acceptable?

KrystalDelusion:
FYI that peepopt has moved from `passes/pmgen/peepopt.cc` to `passes/opt/peepopt.cc` upstream, hence the merge conflict(s).  This will need to be rebased on main (or otherwise merge in the changes).

akashlevy:
Thanks we can move this to the new dir structure soon

akashlevy:
That sounds good to me! I'll get to splitting this in the next couple of days/weeks

akashlevy:
We've decided to just upstream `muldiv_c` since `muxadd` seems to be more Silimate-specific. Expect that soon

akashlevy:
Created #5080, closing this now

PR #4741, opened 2024-11-14
Add `splitfanout` pass

akashlevy:

This PR adds a `splitfanout` pass that splits cells with fanout > 1 into copies, each with fanout 1. It is effectively the opposite of `opt_merge` and is useful for (1) enabling optimizations that cannot work when there is fanout, (2) splitting high fanout cells/nets that might cause problems in PD.

The pass can operate on selections and has a `-limit` flag to selectively operate on cells with fanout < specified limit. It does not operate on "bit-split" cells (word-level cells where different bits of the word have different/non-contiguous destinations), so `splitcells` should be run first if fanout splitting is desired for such cells.

The pass works by building a database of drivers and loads, then operating in reverse topological order on selected cells. It checks that each cell is not "bit-split", then makes a copy to drive each load. The original cell is used to drive the final load.

- [x] `passes/cmds/splitfanout.cc`: `splitfanout` pass
- [x] `passes/cmds/Makefile.inc`: Add the new source

- [x] YosysHQ to review source code and provide feedback/edits as necessary
- [x] Silimate to resolve feedback/edits
- [x] YosysHQ to construct test plan of 15-20 small-medium test cases per peepopt
- [x] Silimate to review test plan and sign off
- [x] YosysHQ to write test cases according to test plan and add to regression
- [x] YosysHQ to provide formal equivalence checking scripts with `equiv_opt`, `miter`, or FOSS `eqy`
- [x] Silimate to sign off on test case implementation

### DIFF
--- a/passes/cmds/splitfanout.cc
+++ b/passes/cmds/splitfanout.cc
@@ -1,0 +1,244 @@
+/*
+ *  yosys -- Yosys Open SYnthesis Suite
+ *
+ *  Copyright (C) 2012  Claire Xenia Wolf <claire@yosyshq.com>
+ *                2024  Akash Levy        <akash@silimate.com>
+ *
+ *  Permission to use, copy, modify, and/or distribute this software for any
+ *  purpose with or without fee is hereby granted, provided that the above
+ *  copyright notice and this permission notice appear in all copies.
+ *
+ *  THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ *  WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ *  MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ *  ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ *  WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ *  ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ *  OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *
+ */
+
+#include "kernel/yosys.h"
+#include "kernel/sigtools.h"
+#include "kernel/utils.h"
+
+USING_YOSYS_NAMESPACE
+PRIVATE_NAMESPACE_BEGIN
+
+struct SplitfanoutWorker
+{
+	Module *module;
+	SigMap sigmap;
+	dict<SigBit, tuple<IdString,IdString,int>> bit_drivers_db;
+	dict<SigBit, pool<tuple<IdString,IdString,int>>> bit_users_db;
+	TopoSort<IdString, RTLIL::sort_by_id_str> toposort;
+
+	SplitfanoutWorker(Module *module) : module(module), sigmap(module)
+	{
+		// Add nodes to topological sorter for all selected cells
+		log("Making toposort nodes for module %s...\n", log_id(module));
+		for (auto cell : module->selected_cells())
+			toposort.node(cell->name);
+
+		// Build bit_drivers_db
+		log("Building bit_drivers_db...\n");
+		for (auto cell : module->cells()) {
+			for (auto conn : cell->connections()) {
+				if (!cell->output(conn.first)) continue;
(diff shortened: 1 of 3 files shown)


passes/cmds/splitfanout.cc
```diff
@@ -0,0 +1,244 @@
...
+
+USING_YOSYS_NAMESPACE
+PRIVATE_NAMESPACE_BEGIN
+
+struct SplitfanoutWorker
+{
+	Module *module;
+	SigMap sigmap;
+	dict<SigBit, tuple<IdString,IdString,int>> bit_drivers_db;
+	dict<SigBit, pool<tuple<IdString,IdString,int>>> bit_users_db;
```
widlarizer, line 33: This construction is copied from `splitcells.cc`, it would be best to leave a comment that marks it as such for the sake of a future rework touching both instances of this pattern. I think some rework is possible to reduce the usage of these indices and make them more readable, but for the most part, if the runtime and memory consumption of the current implementation of `splitfanout` is sufficient for your use case then there isn't any immediate need for that.

Some of the possible improvements:
+ In buffer-normalized mode (see `bufnorm.cc`), `bit_drivers_db` wouldn't be needed
+ The construction of `toposort`, construction of the indices, and final cell duplication could be done with fewer passes over the cells. Though this would also make the pass harder to modify, read, maintain

passes/cmds/splitfanout.cc
```diff
@@ -0,0 +1,244 @@
...
+	TopoSort<IdString, RTLIL::sort_by_id_str> toposort;
+
+	SplitfanoutWorker(Module *module) : module(module), sigmap(module)
+	{
+		// Add nodes to topological sorter for all selected cells
+		log("Making toposort nodes for module %s...\n", log_id(module));
```
widlarizer, line 39: Generally, most of the `log` statements should be demoted to `log_debug`. For example, `opt_clean` only logs how many cells and wires it removed. Listing every cell that's ineligible for splitting is very verbose

passes/cmds/splitfanout.cc
```diff
@@ -0,0 +1,244 @@
...
+			log("Skipping %s cell %s/%s with high fanout %d.\n", log_id(cell->type), log_id(module), log_id(cell), GetSize(bit_users)-1);
+			return 0;
+		}
+
+		// Iterate over bit users and create a new cell for each one
+		log("Splitting %s cell %s/%s into %d copies based on fanout\n", log_id(cell->type), log_id(module), log_id(cell), GetSize(bit_users)-1);
+		int foi = 0;
+		cell->unsetPort(outport);
+		int num_new_cells = GetSize(bit_users)-1;
+		int bit_user_i = num_new_cells;
```
widlarizer, line 133: What's the motivation for counting down here?

passes/cmds/splitfanout.cc
```diff
@@ -0,0 +1,244 @@
...
+		log("This command operates only on cells with 1 output and no 'bit split' on that\n");
+		log("output.\n");
+		log("\n");
+		log("    -limit n\n");
+		log("        max fanout to split.\n");
+		log("\n");
+	}
+	void execute(std::vector<std::string> args, RTLIL::Design *design) override
+	{
+		int limit = -1;
```
widlarizer, line 208: Maybe consider `std::optional<int>`

widlarizer (commented):
I'm not sure why this functionality is needed, but the implementation approach seems solid and it's not redundant with existing passes

passes/cmds/splitfanout.cc
akashlevy: Thanks for noting these. We'll keep these in mind if performance issues arise in the future.

passes/cmds/splitfanout.cc
akashlevy: The original reason was that in our fork, we used `bit_user_i` as a suffix for the new cell's name. But that doesn't apply here, as we just use `NEW_ID` for the new cell's name. It could be changed to increment if that is preferred

passes/cmds/splitfanout.cc
akashlevy: Using `-1` is a pretty common convention

passes/cmds/splitfanout.cc
akashlevy: Agreed. I will reduce the verbosity

passes/cmds/splitfanout.cc
akashlevy: I will add comments for the parts copied from `splitcells`

akashlevy:
Thanks for the tests. They are well-written. I sign off on them.

akashlevy:
Feel free to merge this if it looks good to YosysHQ folks!

Issue #3717, opened 2023-03-29
The `hierarchy -auto-top` command doesn't mark the top level module correctly


tilk:
### Version

0.27

### On which OS did this happen?

Linux

### Reproduction Steps

Take the following Verilog file:

```verilog
module m(input a, output b);
  n x(a, b);
endmodule

module n(input a, output b);
  assign b = !a;
endmodule
```

Load it into Yosys and run `hierarchy -auto-top`, for example:

```
yosys -p "hierarchy -auto-top" test.v
```

Edit: the remark below turned out to be wrong. The results were correct by accident.

> The interesting part is, the problem can be worked around by running certain commands before `hierarchy`, for example:
>
> ```
> yosys -p "opt_clean; hierarchy -auto-top" test.v
> ```
>
> Or:
>
> ```
> yosys -p "write_verilog; hierarchy -auto-top" test.v
> ```

### Expected Behavior

Module `m` is detected as a top-level module:

```
2.3. Analyzing design hierarchy..
Top module:  \m
```

### Actual Behavior

Module `n` is incorrectly detected as a top-level module:

```
2.3. Analyzing design hierarchy..
Top module:  \n
```

This happens also for larger, real-world examples.

tilk:
The issue can be fixed by adding `design->sort()` to `Hierarchy::execute()` in `passes/hierarchy/hierarchy.cc`. As I'm unfamiliar with Yosys's design, I don't have any idea why this works.

arcturusannamalai:
problem with the code https://github.com/YosysHQ/yosys/blob/f93671eb8567f5548844e7065c808a5b4a32481d/passes/hierarchy/hierarchy.cc#L669 seems to be it does only one pass and picks the module containing the maximum instantiated cells not necessarily the top module.

the criterion for top module should be a combination of a high (not necessarily max) number of instances of modules as well as no self-instantiation (i.e. excluding test bench perhaps)

jix:
@arcturusannamalai That's not what `find_top_mod_score` does, the score is the maximum depth of any contained instance:

https://github.com/YosysHQ/yosys/blob/f93671eb8567f5548844e7065c808a5b4a32481d/passes/hierarchy/hierarchy.cc#L683

That is the correct metric to use here and taking the module with the highest score does result in a module that is not instantiated anywhere else (as otherwise the instantiating module would have a higher score).

[...]

PR #4742, opened 2024-11-14
Print a note about finding attribute (* top *) in hierarchy

nakengelhardt:

As suggested in https://github.com/YosysHQ/yosys/issues/3717#issuecomment-2460672212, make it more obvious to the user if the top module was selected due to an attribute overriding the heuristic.

Print a log message. If multiple modules are marked `(* top *)`, the message gets repeated multiple times, matching how the code works (the last attribute encountered while iterating over `design->modules()` overrides any previously seen attributes).


whitequark:
Thanks, this is quite useful!

povik (approved):

PR #4744, opened 2024-11-14
clockgate: add -liberty

widlarizer:
Integrated clock gating cells (ICGs) are ASIC cells designed as part of PDKs and they are described in Liberty files. To avoid requiring per-PDK hacks to PDK-agnostic scripts, instead of requiring in the invocation a specification of a concrete ICG cell name and its interface, this PR allows the user to use `-liberty <filename>` argument for the `clockgate` command and automatically selects the most suitable ICGs described in `<filename>`. A suitable cell is one with matching clock polarity, a reasonable interface based on the [Liberty Reference Manual](https://people.eecs.berkeley.edu/~alanmi/publications/other/liberty07_03.pdf) (refer to pin attributes `clock_gate_...` and cell attribute `clock_gating_integrated_cell`). Most suitable cells are those with the fewest scan/test inputs to tie low and smallest area. Using existing current libparse, I use logic similar to that in `dfflibmap.cc` to find these cells and interfaces.

This PR also assumes all test inputs are to be tied low and all enables are active high. I've tested it manually with asap7 and sky130hd to behave as expected

- [x] test
- [x] help string

### DIFF
--- a/passes/techmap/clockgate.cc
+++ b/passes/techmap/clockgate.cc
@@ -1,5 +1,6 @@
 #include "kernel/yosys.h"
 #include "kernel/ff.h"
+#include "libparse.h"
 #include <optional>
 
 USING_YOSYS_NAMESPACE
@@ -10,6 +11,7 @@ struct ClockGateCell {
 	IdString ce_pin;
 	IdString clk_in_pin;
 	IdString clk_out_pin;
+	std::vector<IdString> tie_lo_pins;
 };
 
 ClockGateCell icg_from_arg(std::string& name, std::string& str) {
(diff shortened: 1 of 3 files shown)


passes/techmap/clockgate.cc
```diff
@@ -37,6 +39,161 @@ ClockGateCell icg_from_arg(std::string& name, std::string& str) {
...
+	for (auto cell : ast->children)
+	{
+		if (cell->id != "cell" || cell->args.size() != 1)
+			continue;
+
+		const LibertyAst *dn = cell->find("dont_use");
+		if (dn != nullptr && dn->value == "true")
+			continue;
+
+		// TODO Should we have user-specified dont_use_cells here?
```
povik, line 73: Yes, suporting the same dont_use argument which dfflibmap and abc supports is a good idea

passes/techmap/clockgate.cc
```diff
@@ -37,6 +39,161 @@ ClockGateCell icg_from_arg(std::string& name, std::string& str) {
...
+		auto cell_name = cell->args[0];
+		auto icg_kind = icg_kind_ast->value;
+		std::vector<std::string> kind_tags;
+		std::stringstream ss(icg_kind);
+		std::string tag;
+		while (std::getline(ss, tag, '_')) {
+			kind_tags.push_back(tag);
+		}
+		if ((kind_tags.size() < 2) || (kind_tags.size() > 4))
+			log_error("Malformed liberty file - invalid clock_gating_integrated_cell value %s in cell %s\n",
```
povik, line 88: Are the set of valid `clock_gating_integrated_cell` standardized? I am a bit worried by the different "malformed input" errors we throw here. I am not sure Liberty can't get extended over time, even if the current standard restricts the valid values to a closed set

passes/techmap/clockgate.cc
```diff
@@ -37,6 +39,161 @@ ClockGateCell icg_from_arg(std::string& name, std::string& str) {
...
+
+		// TODO Should we have user-specified dont_use_cells here?
+
+		const LibertyAst *icg_kind_ast = cell->find("clock_gating_integrated_cell");
+		if (icg_kind_ast == nullptr)
+			continue;
+
+		auto cell_name = cell->args[0];
+		auto icg_kind = icg_kind_ast->value;
+		std::vector<std::string> kind_tags;
```
povik, line 81: I would rather see something in the effect of
```
auto starts_with = [&](std::string prefix) {
    return icg_kind.compare(0, prefix.size(), prefix) == 0;
};

if (icg_kind == "latch_posedge" || starts_with("latch_posedge_")) {
    clk_pol = true;
} else if (icg_kind == "latch_negedge" || starts_with("latch_negedge_")) {
    clk_pol = false;
} else {
    log("Ignoring ICG primitive %s of kind '%s'\n", cell_name.c_str(), icg_kind.c_str());
}
```
to replace all of the icg_kind matching code.

passes/techmap/clockgate.cc
```diff
@@ -37,6 +39,161 @@ ClockGateCell icg_from_arg(std::string& name, std::string& str) {
...
+		}
+		log_debug("maybe valid icg: %s\n", cell_name.c_str());
+		ClockGateCell icg_interface;
+		icg_interface.name = RTLIL::escape_id(cell_name);
+
+		for (auto pin : cell->children) {
+			if (pin->id != "pin" || pin->args.size() != 1)
+				continue;
+
+			log("Check pin %s\n", pin->args[0].c_str());
```
povik, line 115: I think this one was left behind

passes/techmap/clockgate.cc
```diff
@@ -37,6 +39,161 @@ ClockGateCell icg_from_arg(std::string& name, std::string& str) {
...
+	if (f.fail())
+		log_cmd_error("Can't open liberty file `%s': %s\n", filename.c_str(), strerror(errno));
+	LibertyParser libparser(f);
+	f.close();
+	auto ast = libparser.ast;
+
+	// We will pick the most suitable ICG absed on tie_lo count and area
+	struct ICGRankable {
+		ClockGateCell icg;
+		double area;
```
povik, line 55: Let's move area into `ClockGateCell` even if it will be unfilled for non-liberty invocations of the command

passes/techmap/clockgate.cc
```diff
@@ -37,6 +39,161 @@ ClockGateCell icg_from_arg(std::string& name, std::string& str) {
...
+		double area = 0;
+		const LibertyAst *ar = cell->find("area");
+		if (ar != nullptr && !ar->value.empty())
+			area = atof(ar->value.c_str());
+
+		std::optional<ICGRankable>& icg_to_beat = clk_pol ? best_pos : best_neg;
+
+		bool winning = false;
+		if (icg_to_beat) {
+			log_debug("ties: %zu ? %zu\n", icg_to_beat->icg.tie_lo_pins.size(),
```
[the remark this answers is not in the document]
povik: This will simplify the code around here to
```
if (!icg_to_beat ||
    std::make_pair(...) < std::make_pair(...))
  icg_to_beat.emplace(new_icg);

passes/techmap/clockgate.cc
widlarizer: The spec says "A concatenation of up to four strings that describe the functionality of the
cell to the clock-gating software". If it's less than two, we're missing the none vs latch etc detail, or whether it's posedge or negedge. We can't go on without those and in fact it means the Liberty file has non-sensical cells in it. I don't think we should just ignore non-sensical cells

passes/techmap/clockgate.cc
widlarizer: That will simplify things even if it is less extensible, so sure why not

passes/techmap/clockgate.cc
widlarizer: I don't like this idea, ClockGateCell represents a cell's interface. The more types the merrier

passes/techmap/clockgate.cc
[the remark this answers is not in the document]
widlarizer: Oh neat, this didn't occur to me, thanks

widlarizer:
I think what we need is errors that are suppressable as a new category of logging event, in *addition* to the non-suppressable errors that we have now. I want to terminate and tell the user to fix their Liberty file, and still give them the power to just cope with it otherwise

passes/techmap/clockgate.cc
povik: I disagree, it makes the code less readable due to superfluous `.icg`, but I won't withhold approving the PR over it

widlarizer: solved with inheritance instead

KrystalDelusion:
Is there a reason this uses lines like `if (auto clk = pin->find("clock_gate_clock_pin")) {` and then not using the assigned variable for anything else?  Because doing so is giving compiler warnings

PR #4745, opened 2024-11-14
cxxrtl: strip `$paramod` from module name in scope info

whitequark:

The `$paramod\...\...` name for a module doesn't match the source code and isn't useful to the end user given that it's usually containing just a hash.

Module's `hdlname` attribute is used if present.


jsaenznoval:
I'm having the same problem. How could I fixed ?

PR #4746, opened 2024-11-15
synth_gatemate: Ensure proper mapping of multiplexers

pu-cc:

This PR addresses an issue where multiplexers may remain unmapped, leading to their representation as simple assignments in the final netlist instead of as technology primitives.

By running `simplemap` after `muxcover`, we ensure proper mapping of multiplexers into technology primitives.

### DIFF
--- a/techlibs/gatemate/synth_gatemate.cc
+++ b/techlibs/gatemate/synth_gatemate.cc
@@ -301,6 +301,7 @@ struct SynthGateMatePass : public ScriptPass
 			}
 			run("muxcover " + muxcover_args);
 			run("opt -full");
+			run("simplemap");
 			run("techmap -map +/gatemate/mux_map.v");
 		}
 


techlibs/gatemate/synth_gatemate.cc
```diff
@@ -301,6 +301,7 @@ struct SynthGateMatePass : public ScriptPass
 			}
 			run("muxcover " + muxcover_args);
 			run("opt -full");
+			run("simplemap");
```
povik, line 304: `simplemap` will break down `$mux` into individual `$_MUX_` slices so I would expect it sequenced before `muxcover` which expects `$_MUX_` cells for its input. Then there's `simplemap` in the `map_regs` section already ahead of `muxcover`

povik (commented):
It's not clear what the issue is and how an unmapped mux gets to the final netlist. Do you perhaps have a reproducer?

povik:
I was provided a reproducer by @mmicko

techlibs/gatemate/synth_gatemate.cc
mmicko: Problem here is that in this case after `opt -full` `$mux` cells would appear again, so `simplemap` is needed them to get them back to `$_MUX_`.

mmicko (approved):
This approach works fine, and there are potential improvements there, but let's leave that for future, since this unbreaks some use cases.

Issue #3845, opened 2023-07-13
Improve write_verilog speed


j2kun:
### Feature Description

Based on my measurements of `write_verilog` on large designs, it writes to disk at a rate of roughly 2 MiB/s. This results in some large (~GiB) designs taking ~20-30 minutes to write to disk.

Repro: https://gist.github.com/j2kun/43b15994ddc4874c8dda26eb8fafb230

Contains `model.v` and `script.ys` which flattens the model and writes to verilog.

povik:
It's the `-noexpr` mode that feels especially slow. Anyway, with the following patch,

```
diff --git a/backends/verilog/verilog_backend.cc b/backends/verilog/verilog_backend.cc
index 3da168960..5ff191a9c 100644
--- a/backends/verilog/verilog_backend.cc
+++ b/backends/verilog/verilog_backend.cc
@@ -37,7 +37,7 @@ PRIVATE_NAMESPACE_BEGIN
 
 bool verbose, norename, noattr, attr2comment, noexpr, nodec, nohex, nostr, extmem, defparam, decimal, siminit, systemverilog, simple_lhs, noparallelcase;
 int auto_name_counter, auto_name_offset, auto_name_digits, extmem_counter;
-std::map<RTLIL::IdString, int> auto_name_map;
+dict<RTLIL::IdString, int> auto_name_map;
 std::set<RTLIL::IdString> reg_wires;
 std::string auto_prefix, extmem_prefix;
 
@@ -129,7 +129,7 @@ std::string id(RTLIL::IdString internal_id, bool may_rename = true)
                break;
        }
 
-       const pool<string> keywords = {
+       static const pool<string> keywords = {
                // IEEE 1800-2017 Annex B
                "accept_on", "alias", "always", "always_comb", "always_ff", "always_latch", "and", "assert", "assign", "assume", "automatic", "before",
                "begin", "bind", "bins", "binsof", "bit", "break", "buf", "bufif0", "bufif1", "byte", "case", "casex", "casez", "cell", "chandle",

```

the write speed was, in my testing, raised from ~2.5 MB/s to ~34 MB/s.

j2kun:
What an amazingly compact patch for such a dramatic speedup!

PR #4753, opened 2024-11-17
`write_verilog`: Fix `O(N^2)` port dumping to `O(N)`

akashlevy:

`write_verilog` is incredibly slow for modules with lots of ports. We have some modules with thousands of ports and `write_verilog` can take hours to run. Related to issue #3845 perhaps.

This is because port dumping is `O(N^2)` runtime, where `N` is the number of ports in the design. This poor runtime behavior arises due to the nested `for` loop (see lines 2337-2349); each loop can iterate up to the number of ports in the design.

We reduce the runtime to `O(N)` by pre-building a `port_id` to `wire` vector, then using a single `for` loop.

Passes all tests.

### DIFF
--- a/backends/verilog/verilog_backend.cc
+++ b/backends/verilog/verilog_backend.cc
@@ -2332,7 +2332,6 @@ void dump_module(std::ostream &f, std::string indent, RTLIL::Module *module)
 
 	dump_attributes(f, indent, module->attributes, "\n", /*modattr=*/true);
 	f << stringf("%s" "module %s(", indent.c_str(), id(module->name, false).c_str());
-	bool keep_running = true;
 	int cnt = 0;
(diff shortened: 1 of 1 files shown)


zachjs (approved):

povik:
Fwiw the ports field on modules is holding the ordered port names too

akashlevy:
Should we use that instead? Iterate through port names and get `module->wire(port)`?

zachjs:
> Should we use that instead? Iterate through port names and get `module->wire(port)`?

That sounds like a nice improvement!

akashlevy:
I'm on it :)

zachjs:
Patch looks good to me.

povik:
Let’s wait for green CI and merge

PR #4754, opened 2024-11-17
Fix `editline` Makefile

akashlevy:

There are three issues with `editline` in the Makefile:

1. It includes `libbsd` and `ncurses` libraries, which are no longer needed
2. It needs `ENABLE_READLINE := 1` in the Makefile to do anything
3. It does not properly set `ABC_USE_NO_READLINE=1`, so `readline` still gets used for `abc`

Fully decouple `editline` from `readline` in the Makefile and reduce `LIBS` flags down to just `-ledit`

Test on a machine/container/VM with `editline` installed but no `readline`. Run `make ENABLE_READLINE=0 ENABLE_EDITLINE=1` before and after the change

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -396,12 +396,12 @@ ifeq ($(CONFIG),mxe)
 LIBS += -ltermcap
 endif
 else
-ifeq ($(ENABLE_EDITLINE),1)
-CXXFLAGS += -DYOSYS_ENABLE_EDITLINE
-LIBS += -ledit -ltinfo -lbsd
-else
 ABCMKARGS += "ABC_USE_NO_READLINE=1"
 endif
+
+ifeq ($(ENABLE_EDITLINE),1)
+CXXFLAGS += -DYOSYS_ENABLE_EDITLINE
+LIBS += -ledit
 endif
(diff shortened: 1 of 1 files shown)


mmicko (changes requested):
Hi. Thanks for looking into this.
There is one change here that affects how it actually works. 
Please note that ENABLE_READLINE and ENABLE_EDITLINE should be mutually exclusive, and by moving it outside else block now they are not. 
If currently existing block is changed to (without moving code) :
```
ifeq ($(ENABLE_EDITLINE),1)
CXXFLAGS += -DYOSYS_ENABLE_EDITLINE
LIBS += -ledit
endif
ABCMKARGS += "ABC_USE_NO_READLINE=1"
endif
```
then I think should be fine, so just update LIBS and change `else` to `endif` would make sure ABC_USE_NO_READLINE is set properly

mmicko:
Thanks. So when testing you actually must set ENABLE_READLINE to 0 and ENABLE_EDITLINE to 1, since readline is default set option

mmicko (approved):

PR #4755, opened 2024-11-17
Gowin: add GW2A and GW5A cells

pepijndevos:

Missing GW2A primitives.

Adding the relevant simlib files and rerunning the script.

`read_verilog -lib techlibs/gowin/cells_xtra.v`

The python script seems to produce a few syntax errors where lines end in commas sometimes, not sure why, but there were only a hand full of instances that I just fixed by hand.
It also appears the script doesn't deduplicate primitives that are shared between families, which are a lot of them, but blackboxes just get overwritten so it seems harmless.
All I wanted was a file with all the Gowin primitives so I didn't investigate further.

Oddly enough it also removed a few primitives, and this doesn't seem to be a bug, they actually disappeared after V1.9.8


widlarizer:
Hmm, why does this PR remove FLASH128K?

pepijndevos:
As I mentioned, they are not part of the latest version of the ide, supposedly because they were only featured in EOL chips.

widlarizer (approved):
Fair enough

PR #4612, opened 2024-09-24
opt_demorgan: skip zero width cells

georgerennie:
Fixes #4610

PR #4762, opened 2024-11-20
tests: replace read_ilang with read_rtlil

georgerennie:
#4612 was written before read_ilang was deprecated but merged after so caused test failures. This switches read_ilang to read_rtlil in its testcase.


KrystalDelusion:
Isn't that also what merge queues do?

KrystalDelusion:
I would also point out that read_ilang has been deprecated for circa 5 years, it's only just now been removed

PR #4765, opened 2024-11-21
read_rtlil: Warn on assigns after switches in case rules

georgerennie:

Previously `read_rtlil` would allow `assign` and `switch` rules to be freely mixed in RTLIL case rules, but internally yosys stores the two in two separate vectors and treats assigns as if they came before switches. This change modifies the parser to ~~reject~~ (edit: changed to emit a warning instead) RTLIL that doesn't obey this and updates the docs. `write_rtlil` will only write RTLIL obeying this property.

~~The `case_body` rule is split into two parts, firstly assignments and then switches.~~ edit: A warning is run in the assign rule if the current level already has switches assigned.

The `read_rtlil` commands in the test suite should give some confidence to this.

### DIFF
--- a/docs/source/appendix/rtlil_text.rst
+++ b/docs/source/appendix/rtlil_text.rst
@@ -242,7 +242,7 @@ Processes
 
 Declares a process, with zero or more attributes, with the given identifier in
 the enclosing module. The body of a process consists of zero or more
-assignments, exactly one switch, and zero or more syncs.
+assignments followed by zero or more switches and zero or more syncs.
 
 See :ref:`sec:rtlil_process` for an overview of processes.
 
@@ -250,7 +250,7 @@ See :ref:`sec:rtlil_process` for an overview of processes.
 
     <process>       ::= <attr-stmt>* <proc-stmt> <process-body> <proc-end-stmt>
     <proc-stmt>     ::= process <id> <eol>
-    <process-body>  ::= <assign-stmt>* <switch>? <assign-stmt>* <sync>*
+    <process-body>  ::= <assign-stmt>* <switch>* <sync>*
     <assign-stmt>   ::= assign <dest-sigspec> <src-sigspec> <eol>
     <dest-sigspec>  ::= <sigspec>
     <src-sigspec>   ::= <sigspec>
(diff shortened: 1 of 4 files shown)


whitequark:
> This change modifies the parser to reject RTLIL that doesn't obey this and updates the docs.

That's a breaking change to RTLIL. Are you certain it will not break other producers?

povik:
> Are you certain it will not break other producers?

I guess we can never be certain. Btw what other RTLIL producers do we know of? (Asking because of #4723)

I agree with making the documentation change. With the parser change I would be ok with playing it safe and parsing the malformed input too.

povik:
> maybe i should look at Yosys slang too?)

@georgerennie as a plugin it doesn’t produce textual RTLIL currently (it fills the in-memory structures)

whitequark:
> Given the unintuitive current behaviour, I would hope that other producers are not reliant on the fact that assigns are treated before switches.

What I'm worried about is a producer that knows that assigns and switches are treated separately, and emits them interspersed *because it knows that*. I.e. a correctly written producer.

I think emitting a warning in `read_rtlil` where an error would happen otherwise should catch all such cases, and in a year or two we could remove it entirely.

georgerennie:
> I think emitting a warning in read_rtlil where an error would happen otherwise should catch all such cases, and in a year or two we could remove it entirely.

I'm happy to do this too. Do we actually know of any textual RTLIL producers external to yosys other than Amaranth?

whitequark:
I recall @nakengelhardt bringing some up in a past meeting but I don't know which one it was (it may have been some academic thing?)

povik (approved):

docs/source/appendix/rtlil_text.rst
```diff
@@ -262,8 +262,8 @@ Switches
 Switches test a signal for equality against a list of cases. Each case specifies
 a comma-separated list of signals to check against. If there are no signals in
 the list, then the case is the default case. The body of a case consists of zero
-or more switches and assignments. Both switches and cases may have zero or more
-attributes.
+or more assignments followed by zero or more switches. Both switches and cases
```
KrystalDelusion, line 265: Just to make sure, not following this currently raises a warning with the intention that it will be enforced later?

KrystalDelusion (approved):

docs/source/appendix/rtlil_text.rst
georgerennie: Yes exactly

PR #4768, opened 2024-11-25
Refactor full_selection

KrystalDelusion:
A `full_selection` currently includes boxed modules, but `Design::selected_modules()` silently ignores boxes.  This has lead to a number of workarounds in the code for getting all unboxed selected modules.

- [x] Change `full_selection` to not include (black)boxes.  Non-full selections can contain blackboxes (but this will only happen with =patterns that ask for them explicitly).
- [x] Add `complete_selection` flag that has the previous meaning of `full_selection`, i.e. all modules including boxes for optimisation purposes and to simplify nested commands that expect boxed modules, like `abc9`.

Note:
`Design::selected_whole_modules()` is marked deprecated because I'm not convinced `attrmap` should be ignoring boxes, even though that is the current behaviour.

Optional (but probably for a separate pr):
- [ ] Update existing iterations over `Design::modules()` to use `Design::selected_modules()` where relevant.
- [ ] Update backends to be consistent w.r.t. selections (providing `-selected` where possible, otherwise working on the full design).

New tests are added in `tests/select` to check the selection semantics w.r.t boxes that have been changed here, but I guess double check that the semantics when applied to boxes match their stated intention in `help select`.


KrystalDelusion:
Rebase on v`0.51+17`

kernel/rtlil.h
```diff
@@ -1135,31 +1165,49 @@ struct RTLIL::SigSpec
 struct RTLIL::Selection
 {
 	bool full_selection;
+	bool selects_boxes;
+	bool complete_selection;
```
aiju, line 1169: maybe put a comment here explaining the difference between full_selection and complete_selection?

kernel/rtlil.h
```diff
@@ -1268,29 +1322,34 @@ struct RTLIL::Design
 	}
 
 	template<typename T1> void select(T1 *module) {
-		if (selection_stack.size() > 0) {
-			RTLIL::Selection &sel = selection_stack.back();
-			sel.select(module);
-		}
+		RTLIL::Selection &sel = selection();
```
aiju, line 1325: why is the check gone? is there an invariant now that the selection stack is never empty?

kernel/rtlil.h
```diff
@@ -1268,29 +1322,34 @@ struct RTLIL::Design
...
+		sel.select(module, member);
 	}
 
 
-	std::vector<RTLIL::Module*> selected_modules() const;
-	std::vector<RTLIL::Module*> selected_whole_modules() const;
-	std::vector<RTLIL::Module*> selected_whole_modules_warn(bool include_wb = false) const;
+	std::vector<RTLIL::Module*> selected_modules(RTLIL::SelectPartials partials = SELECT_ALL, RTLIL::SelectBoxes boxes = SB_UNBOXED_WARN) const;
+
+	std::vector<RTLIL::Module*> all_selected_modules() const { return selected_modules(SELECT_ALL, SB_ALL); }
```
aiju, line 1337: maybe document all the different variants here? it's not clear to me what the differences are

passes/cmds/select.cc
```diff
@@ -366,38 +425,49 @@ static void select_op_diff(RTLIL::Design *design, RTLIL::Selection &lhs, const R
...
+	if (lhs.empty())
+		return;
+
+	if (rhs.empty()) {
 		lhs.full_selection = false;
-		for (auto mod : design->modules())
-			lhs.selected_modules.insert(mod->name);
+		lhs.complete_selection = false;
+		lhs.selected_modules.clear();
+		lhs.selected_members.clear();
```
aiju, line 441: maybe turn that into a method since it appears multiple times?

aiju (approved):
mostly looks good, just a few places where i would like to see more documentation (see comments)

kernel/rtlil.h
KrystalDelusion: Previously an empty stack was an implied full selection, but now it's explicit; popping to an empty stack calls `Design::push_full_selection()` and `Design::check()` has a `log_assert(!selection_stack.empty());`.

mmicko:
This warning is repeated for each inlcude of rtlil.h, so for compiling almost any file in yosys
```
./kernel/rtlil.h:1186:3: warning: field 'full_selection' will be initialized after field 'selects_boxes' [-Wreorder-ctor]
 1186 |                 full_selection(full && !boxes), selects_boxes(boxes), complete_selection(full && boxes), current_design(design) { }
      |                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~  ~~~~~~~~~~~~~~~~~~~~  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                 selects_boxes(boxes)            complete_selection(full && boxes) full_selection(full && !boxes)

```

PR #4773, opened 2024-11-27
wrapcell: Add new command

povik:

Preserving arithmetic operator boundaries upto a backend tool requires putting them into a module of their own. This module must not be parametric, and should be given a human-readable name (so not the `$paramod$...` naming used by Yosys internally).

Combinations of `techmap -extern`, `box_derive`, `chtype` can be used for this but the sequencing grows awkward and requires custom techmap files.

A new command let's you wrap selected internal cells into individual modules, with the name generated from a template by filling in parameter values.

E.g.
```
wrapcell -name MUL_{A_WIDTH}_{B_WIDTH}_{Y_WIDTH} t:$mul  # ignores signedness
```

### DIFF
--- a/passes/cmds/wrapcell.cc
+++ b/passes/cmds/wrapcell.cc
@@ -1,0 +1,170 @@
+/*
+ *  yosys -- Yosys Open SYnthesis Suite
+ *
+ *  Copyright (C) 2024  Martin Povišer <povik@cutebit.org>
+ *
+ *  Permission to use, copy, modify, and/or distribute this software for any
+ *  purpose with or without fee is hereby granted, provided that the above
+ *  copyright notice and this permission notice appear in all copies.
+ *
+ *  THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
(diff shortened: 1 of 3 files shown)


passes/cmds/wrapcell.cc
```diff
@@ -0,0 +1,170 @@
...
+				subm->fixup_ports();
+
+				for (auto rule : attributes) {
+					if (rule.value_fmt.empty()) {
+						subm->set_bool_attribute(rule.name);
+					} else {
+						std::optional<std::string> value = format(rule.value_fmt, cell->parameters);
+
+						if (!value)
+							log_error("Formatting error when processing cell '%s' in module '%s'\n",
```
widlarizer, line 156: What this PR does currently works ok, but this pattern of log + log_error makes me wish for something like Result types in other languages. Ultimately what we want to do is return from `format` either a `Result(formatted_string)` or `Error(msg)` and if it's Error then return a `Error(msg with pass-level context)` from `execute`. [This](https://github.com/oktal/result) could kind of get it done, but there's no way to match on it, and `std::variant` is too annoying

widlarizer (approved):

PR #4775, opened 2024-11-27
techmap: Add `-dont_map` for selective disabling of rules

povik:

Preserving operator boundaries requires leaving some cells unmapped at first. That can be accomplished by commenting out the relevant section of the mapping rule file, but an explicit argument to `techmap` is a more elegant solution.

### DIFF
--- a/passes/techmap/techmap.cc
+++ b/passes/techmap/techmap.cc
@@ -1029,8 +1029,7 @@ struct TechmapPass : public Pass {
 		log("        '-nooverwrite' option set.\n");
 		log("\n");
 		log("    -dont_map <celltype>\n");
-		log("        ignore any mapping rules for the given cell type, that is leave it\n");
-		log("        unmapped.\n");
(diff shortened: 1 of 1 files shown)


passes/techmap/techmap.cc
```diff
@@ -1028,6 +1028,10 @@ struct TechmapPass : public Pass {
 		log("        map file. Note that the Verilog frontend is also called with the\n");
 		log("        '-nooverwrite' option set.\n");
 		log("\n");
+		log("    -dont_map <celltype>\n");
+		log("        ignore any mapping rules for the given cell type, that is leave it\n");
```
widlarizer, line 1032: Maybe re-order, something like "leave <celltype> unmapped by ignoring mapping rules for it"

widlarizer (approved):

passes/techmap/techmap.cc
povik: I've rephrased

povik: Thanks

PR #4779, opened 2024-11-28
Add a Han-Carlson option for `$lcu` mapping

widlarizer:
The Han-Carlson parallel-prefix adder (PPA) is an adder construction for ASIC use with an attractive tradeoff. We expect similar delay to Kogge-Stone PPA but better area. I have implemented it in the same procedural style as the current Kogge-Stone and Brent-Kung PPAs. I'm testing it by proving equivalence with `equiv_simple` for bit widths from 1 to 16. I have additionally replaced the test for Kogge-Stone with the same approach. To sweep a variable, full TCL scripting is necessary, which Yosys supports. This PR adds support for using TCL scripts in tests. The HC-PPA and KS-PPA tests take 120ms and 95ms on my Zen 4 machine, respectively.

I'm not really sure what to do with the priority integer at the beginning of the module name. Reviewers may test the tests by locally modifying the designs and watching the tests fail


tests/techmap/lcu_refined.v
```diff
@@ -0,0 +1,20 @@
+// Copied from techlibs/common/simlib.v
+// with this condition removed: (^{P, G, CI} !== 1'bx)
+module lcu (P, G, CI, CO);
```
povik, line 3: If we instantiate `$lcu` explicitly inside the gold module we can avoid `ppa.nomatch`

povik: requires `-icells` on read_verilog

tests/techmap/kogge-stone.tcl
```diff
@@ -0,0 +1,14 @@
...
+
+read_verilog +/choices/kogge-stone.v
+read_verilog lcu_refined.v
+design -save init
+
+for {set i 1} {$i <= 16} {incr i} {
+    design -load init
+    chparam -set WIDTH $i
+    yosys proc
+    equiv_make -blacklist ppa.nomatch lcu _80_lcu_kogge_stone equiv
```
povik, line 11: The fact that this matches on internal names makes me think we are using the wrong command to check equivalence here

povik (approved):
I left some comments but by me it's fine to merge as is

tests/techmap/kogge-stone.tcl
widlarizer: looks like the folklore is `opt_clean` before `equiv_make`

PR #4781, opened 2024-11-28
proc_dff: bit-granularity optimizations and refactoring

georgerennie:
`proc_dff` applies a few optimizations to the asynchronous reset rules it considers, but previously did this in a fairly adhoc way and on the granularity of whole signals. This PR tries to tidy it up to be more principled, separating out the collection of sync rules, the optimization of rules and the resulting generation of flip flops. As part of this I have added the ability to separately consider optimizations for different parts of the signal. This means that different bits of a signal can be lowered to different DFF types, preferring simpler types to `$aldff` and `$dffsr`.

Leaving this as a draft for now and can write more about what it does later but there are a lot of comments in the code if you are interested.

This breaks some tests at the moment that rely on the worse inference. I need to have a look at them and add some more tests. As this is a fairly large change that touches a core pass used by most users I want to be confident that things won't break.

As a quick example for how this can help, consider the below example. This was motivated by a pattern I saw with sv2v where a struct is lowered to one wire and may be partially reset even though the whole thing gets assigned at once.
```verilog
module top(
	input wire clk,
	input wire rst,
	output reg [7:0] a,
	input wire [7:0] b
);

	if (rst)
		a[3:0] <= '0;
	else
		a <= b;
end

endmodule
```
Before this used an `$aldff`, which fails `check` because of a comb path from the output back through the load data input. After this PR this instead uses a `$dffe` and an `$adff`.

Another example
```verilog
module top(input wire clk, input wire rsta, input wire rstb, input wire [7:0] b, output reg [7:0] a);

always @(posedge clk or posedge rsta or posedge rstb) begin
	if (rsta)
		a <= '0;
	else if (rstb)
		a <= {a[7:4], 4'b0};
	else
		a <= b;
end

endmodule
```
`proc; opt_dff` would produce a `$dffsr` before, but now produces one `$adff` and one `$adffe`.


georgerennie:
As a side note, I think there are other bits of `proc` that could do with a bit of tidying up and being adapted to cover more general patterns. Maybe I'll have a look at `proc_arst` at somepoint...

jix:
As far as I can tell, the only thing missing is some more testing with representative async reset verilog designs to rule out regressions. @georgerennie have you been using this PR or done any more testing using it in the meantime?

georgerennie:
I have been using a build of yosys with this with a number of test designs (quite a few converted from SV2V which is what this was meant to address). I didn't find issues with it yet, can add some more tests and I'll also have a look over the code again - it's been long enough since I wrote it that I think I am almost approaching it with the eyes of a fresh reviewer!

georgerennie:
There's a mild performance degradation (a few percent for just this pass maybe) - I think perf can be improved by precalculating the set of disjoint lvalues rather than going through and calculating them for each dff which requires a lot of sorting the sync rule lvalues. I am working on a patch for that but its probably easier to put it in a different pr.

povik:
> I believe Amaranth doesn't use sync processes so read_verilog and yosys-slang are probably the main interfaces affected by this.

Let me leave a note yosys-slang has retired the use of sync processes.

jix:
> I am working on a patch for that but its probably easier to put it in a different pr.

Does that mean that what you just pushed to this PR is ready to be merged? Or are you still in the process of reviewing it and adding tests? (It's fine if you need more time, I just want to avoid this PR sitting around for months again in case it actually is ready.)

georgerennie:
> Does that mean that what you just pushed to this PR is ready to be merged? Or are you still in the process of reviewing it and adding tests? (It's fine if you need more time, I just want to avoid this PR sitting around for months again in case it actually is ready.)

In the process of trying to improvev that I've found what I think is a better way of approaching the whole optimizations anyway that happens to improve `proc_dff` speed 4-7x on the designs I ran it on - therefore I think actually I will put it in this PR but there is no point reviewing until I do so. Will mark it draft until I push that to make this clear.

PR #4782, opened 2024-11-28
opt_expr: limit effort to reduce runtime on anomalous designs

widlarizer:
The first case of the ROM section of `tests/arch/ice40/memories.ys` takes 326 seconds to run with majority of this runtime spent sorting cells in `opt_expr`. This PR allows sorting to only fail a limited number of times on a given module before reaching fixpoint without trying to sort the cells. This cuts down the test case runtime to 14 seconds. The limit can be changed with the new `-effort` argument.

```
5.3.12. Executing OPT_EXPR pass (perform const folding).
Optimizing module sync_rom.
Couldn't topologically sort cells, optimizing module sync_rom may take a longer time.
Couldn't topologically sort cells, optimizing module sync_rom may take a longer time.
Couldn't topologically sort cells, optimizing module sync_rom may take a longer time.
Couldn't topologically sort cells, optimizing module sync_rom may take a longer time.
Couldn't topologically sort cells, optimizing module sync_rom may take a longer time.
Effort of 5 exceeded, no longer attempting toposort on module sync_rom.
<suppressed ~6151 debug messages>
```

### DIFF
--- a/passes/opt/opt_expr.cc
+++ b/passes/opt/opt_expr.cc
@@ -30,6 +30,7 @@ USING_YOSYS_NAMESPACE
 PRIVATE_NAMESPACE_BEGIN
 
 bool did_something;
+int sort_fails = 0;
 
 void replace_undriven(RTLIL::Module *module, const CellTypes &ct)
 {
@@ -393,7 +394,7 @@ int get_highest_hot_index(RTLIL::SigSpec signal)
 	return -1;
 }
 
-void replace_const_cells(RTLIL::Design *design, RTLIL::Module *module, bool consume_x, bool mux_undef, bool mux_bool, bool do_fine, bool keepdc, bool noclkinv)
+void replace_const_cells(RTLIL::Design *design, RTLIL::Module *module, bool consume_x, bool mux_undef, bool mux_bool, bool do_fine, bool keepdc, bool noclkinv, int effort)
 {
 	SigMap assign_map(module);
 	dict<RTLIL::SigSpec, RTLIL::SigSpec> invert_map;
@@ -512,10 +513,14 @@ void replace_const_cells(RTLIL::Design *design, RTLIL::Module *module, bool cons
 			cells.edge(cells.node(outbit_to_cell.at(bit)), r_index);
 	}
 
-	if (!cells.sort()) {
+	if (sort_fails < effort && !cells.sort()) {
 		// There might be a combinational loop, or there might be constants on the output of cells. 'check' may find out more.
 		// ...unless this is a coarse-grained cell loop, but not a bit loop, in which case it won't, and all is good.
 		log("Couldn't topologically sort cells, optimizing module %s may take a longer time.\n", log_id(module));
+		sort_fails++;
+		if (sort_fails >= effort)
+			log("Effort of %d exceeded, no longer attempting toposort on module %s.\n",
+				 effort, log_id(module));
 	}
 
 	for (auto cell : cells.sorted)
@@ -2253,6 +2258,10 @@ struct OptExprPass : public Pass {
 		log("        all result bits to be set to x. this behavior changes when 'a+0' is\n");
 		log("        replaced by 'a'. the -keepdc option disables all such optimizations.\n");
 		log("\n");
+		log("    -effort N\n");
+		log("        allow toposort to fail in N iterations on each module before giving up\n");
+		log("         on sorting for that module. Default value is 5\n");
+		log("\n");
 	}
 	void execute(std::vector<std::string> args, RTLIL::Design *design) override
 	{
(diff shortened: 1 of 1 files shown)


passes/opt/opt_expr.cc
```diff
@@ -512,10 +513,14 @@ void replace_const_cells(RTLIL::Design *design, RTLIL::Module *module, bool cons
...
 		// There might be a combinational loop, or there might be constants on the output of cells. 'check' may find out more.
 		// ...unless this is a coarse-grained cell loop, but not a bit loop, in which case it won't, and all is good.
 		log("Couldn't topologically sort cells, optimizing module %s may take a longer time.\n", log_id(module));
+		sort_fails++;
+		if (sort_fails >= effort)
+			log("Effort of %d exceeded, no longer attempting toposort on module %s.\n",
+				 effort, log_id(module));
 	}
 
 	for (auto cell : cells.sorted)
```
povik, line 526: Isn't this empty if you don't run `cells.sort()`? And if that's the case, then the true cause of the speed-up isn't not bothering to sort anymore but giving up entirely

widlarizer: Fixed by iterating over a list of references to a snapshot of `module->cells()`. This increases the runtime to 83 seconds, which I still prefer to 326 seconds. I should say, all runtimes I'm mentioning in this PR are measured with `ENABLE_DEBUG=1` for no particular reason other having that around in `Makefile.conf`

povik: We could cache the sorting result provided we remove cells we removed

widlarizer: I can't really foresee the consequences of doing that, though - for how many cycles would we cache etc

povik: For the lifetime of one `opt_expr` invocation. It should be ok as long as we zero out the position in the sorted array corresponding to any removed cells.

I am not sure if there's a case where opt_expr replaces a cell with two or more cells, that's a slight complication

KrystalDelusion (commented):
I'm struggling to follow the control flow here; my comments are based on the understanding that `iterator` is a lambda function, which gets called once with another (anonymous) lambda function.  But if that is the case then I'm not sure why `iterator` is a lambda at all if it only gets called once.  In fact, naming the inner anonymous function (the contents of the previous for loop over `cells.sorted`), moving lines 494-537 after it, dropping the outer lambda and calling the new lambda directly instead of the locally defined `replace_cell` seems to give the same results, takes very slightly less time, and is more readable.  Which probably also suggests that the inner lambda function could even be elevated to a non-lambda?

passes/opt/opt_expr.cc
```diff
@@ -490,35 +491,51 @@ void replace_const_cells(RTLIL::Design *design, RTLIL::Module *module, bool cons
...
+			if (design->selected(module, cell) && yosys_celltypes.cell_evaluable(cell->type)) {
+				const int r_index = cells.node(cell);
+				for (auto &conn : cell->connections())
+				if (yosys_celltypes.cell_input(cell->type, conn.first))
+				for (auto bit : assign_map(conn.second))
+				if (outbit_to_cell.count(bit))
+					cells.edge(cells.node(outbit_to_cell.at(bit)), r_index);
+			}
+
+			if (sort_fails < effort && !cells.sort()) {
```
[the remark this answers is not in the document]
widlarizer: Yeah, good catch, that's redundant, this must be left over from experimenting with the pass

widlarizer:
@KrystalDelusion I suppose I tried to be too clever by retaining the pre-existing code ordering to make it easier to review (for myself as well). It uncouples the design traversal without reordering the responsible code blocks, forcing a refactoring into separate functions or creating a worker class that holds all the silly state. When you're saying you named the anonymous function and moved the replacement code after it, do you mean that you nested a function within `replace_const_cells`? If so, I don't think that's standard C++.  Otherwise, you have to pass a bunch of variables in, or create a stateful worker (which is the correct thing to do)

KrystalDelusion:
By named I mean assigned to a variable (`auto l = ..`)

widlarizer:
Uh, anyway, I spent a couple hours refactoring `opt_expr`. It's at [emil/opt_expr-limit-effort-refactor](https://github.com/YosysHQ/yosys/tree/emil/opt_expr-limit-effort-refactor). I'm not sure if it's worth the diff now that I see what you meant. If so, I can pull it into this PR or a followup one but I have to attend to different tasks at the moment

KrystalDelusion:
Ah I see, although I think just the minimal refactor to avoid the `visitor` lambda is still an improvement in readability (I pushed to [krys/opt_expr-limit-effort](https://github.com/YosysHQ/yosys/tree/krys/opt_expr-limit-effort))

KrystalDelusion:
Also the CI failing is fixed in latest main, just needs to rebase

ShinyKate:
Ping reminder 👋

KrystalDelusion:
Rebased on latest main (v`0.57+125`) for the CI fix

widlarizer:
Sorry for leaving this PR in purgatory. The original justification does not apply as of I think #4783. This PR continues to be up in case something pops up again and we decide to have this emergency mechanism but there's no known need for it and I would like to avoid merging speculatively beneficial code

PR #4784, opened 2024-11-28
Reduce number of warnings

KrystalDelusion:
There are a number of easy to fix warnings:
1.  `comparison of integers of different signs: 'size_t' (aka 'unsigned long') and 'int' [-Wsign-compare]` 

1. `Const::size()` returns int, so change iterators that use it to `auto` instead of `size_t`. For cases where size is being explicitly cast to `int`, use the wrapper that we already have instead: `Yosys::GetSize()`.

Check what/if any warnings still show up at the bottom of the "Files changed" tab and in the "Compiler testing" section of the checks


KrystalDelusion:
@povik Do you want me to tidy up the warnings in the aiger2 code?

povik:
Yes please

widlarizer (approved):

PR #4787, opened 2024-12-01
booth: Map simple `$macc` instances too

povik:

As it stands `booth` ignores `$macc` cells, which means one cannot sequence `booth` after `alumacc`. Make the change to `booth` to handle at least the basic `$macc` cells which represent a simple product. The more advanced case in which `$macc` is a combined multiply-accumulate operation is left for later to be dealt with

### DIFF
--- a/tests/techmap/booth.ys
+++ b/tests/techmap/booth.ys
@@ -10,6 +10,19 @@ endmodule
 EOF
 booth
 sat -verify -set a 0 -set b 0 -prove y 0
-design -reset
 
+design -reset
 test_cell -s 1694091355 -n 100 -script booth_map_script.ys_ $mul
+
+design -reset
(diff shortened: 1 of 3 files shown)


tests/techmap/booth.ys
```diff
@@ -10,6 +10,19 @@ endmodule
...
+read_verilog <<EOF
+module top(a,b,y);
+input wire [4:0] a;
+input wire [5:0] b;
+output wire [6:0] y;
+assign y = a * b;
+endmodule
+EOF
+synth -run :fine
+# test compatibility with alumacc
```
widlarizer, line 27: I think that ideally at this point we would assert the presence of `$macc` to make sure we're testing what we want to test

widlarizer (approved):

PR #4794, opened 2024-12-03
Fix Cygwin compatibility

KrystalDelusion:
Cygwin builds currently fail with two errors, one due to an overzealous compatibility check in libs/fst; the other in kernel/mem.cc due to redefining `addr_t` which is a system type.

Drop `__CYGWIN__` from the mmap compatibility check in libs/fst.
Remove the `addr_t` redefinition, instead using `MemContents::addr_t` where needed.


Ravenslofty (approved):
Seems like we can Just Merge It.

PR #4796, opened 2024-12-03
qlf_k6n10f: leave unused clocks disconnected

Ravenslofty:
The K6N10F architecture expects unused clocks to be unconnected, however `memory_libmap` connects unused clocks to ground.

The memory mapping files are modified to disconnect the clocks if they have a constant driver.


widlarizer:
Context for closing this: we didn't get confirmation that this actually does match what the physical design tools expect

PR #4797, opened 2024-12-03
Allow multiple -liberty args in dfflibmap and clockgate

widlarizer:
`dfflibmap` and `clockgate` are the only passes which accept a `-liberty` flag but only once. This requires workarounds for PDKs like asap7 which come in multiple files sharing the same "header" (that is, anything but cells) and having distinct sets of cells. The other commands accepting any number of `-liberty` arguments are the abc commands and `stat`. `read_liberty` is a frontend and therefore fixed to one file per invocation.

In this PR, I'm extending libparse with `LibertyMergedCells` which steals `LibertyAst*` out of `LibertyParser`s before the AST is destroyed when `~LibertyAst` is executed. I'm extending tests for both affected commands to cover this functionality. There's a nice extra benefit of deduplicating some sanity checks from passes into `LibertyMergedCells`.

### DIFF
--- a/passes/techmap/clockgate.cc
+++ b/passes/techmap/clockgate.cc
@@ -291,18 +291,22 @@ struct ClockgatePass : public Pass {
 				auto name = args[++argidx];
 				auto rest = args[++argidx];
 				pos_icg_desc = icg_from_arg(name, rest);
+				continue;
 			}
 			if (args[argidx] == "-neg" && argidx+2 < args.size()) {
 				auto name = args[++argidx];
 				auto rest = args[++argidx];
 				neg_icg_desc = icg_from_arg(name, rest);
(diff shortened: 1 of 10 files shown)


Ravenslofty (approved):

KrystalDelusion:
> `read_liberty` is a frontend and therefore fixed to one file per invocation.

Is that true for all frontends (other than `verific` which is a special case)?  I'm sure I've used `read_verilog` with multiple files, but maybe that results in multiple calls?

povik:
> I'm sure I've used read_verilog with multiple files, but maybe that results in multiple calls?

I've used `read_liberty` with multiple files, I think there's external logic to repeatedly call the frontend

PR #4799, opened 2024-12-05
wrapcell: Optionally track unused outputs

povik:

Provide the option for `wrapcell` to generate wrapper modules with unused ports removed. In the motivating use case for this command which is preserving arithmetic operator boundaries this helps with QoR.

### DIFF
--- a/passes/cmds/wrapcell.cc
+++ b/passes/cmds/wrapcell.cc
@@ -18,12 +18,37 @@
  */
 #include "kernel/yosys.h"
 #include "kernel/celltypes.h"
+#include "kernel/sigtools.h"
 #include "backends/rtlil/rtlil_backend.h"
 
 USING_YOSYS_NAMESPACE
 PRIVATE_NAMESPACE_BEGIN
 
-std::optional<std::string> format(std::string fmt, const dict<IdString, Const> &parameters)
+bool has_fmt_field(std::string fmt, std::string field_name)
+{
+	auto it = fmt.begin();
+	while (it != fmt.end()) {
+		if (*it == '{') {
+			it++;
+			auto beg = it;
+			while (it != fmt.end() && *it != '}') it++;
+			if (it == fmt.end())
+				return false;
+
+			if (std::string(beg, it) == field_name)
+				return true;
+		}
+		it++;
+	}
+	return false;
+}
+
+struct ContextData {
+	std::string unused_outputs;
+};
+
+std::optional<std::string> format(std::string fmt, const dict<IdString, Const> &parameters,
+								  ContextData &context)
 {
 	std::stringstream result;
 
@@ -38,13 +63,19 @@ std::optional<std::string> format(std::string fmt, const dict<IdString, Const> &
 				return {};
 			}
 
-			auto id = RTLIL::escape_id(std::string(beg, it));
-			if (!parameters.count(id)) {
-				log("Parameter %s referenced in format string '%s' not found\n", log_id(id), fmt.c_str());
-				return {};
-			}
+			std::string param_name = {beg, it};
 
-			RTLIL_BACKEND::dump_const(result, parameters.at(id));
+			if (param_name == "%unused") {
+				result << context.unused_outputs;
+			} else {
+				auto id = RTLIL::escape_id(std::string(beg, it));
+				if (!parameters.count(id)) {
+					log("Parameter %s referenced in format string '%s' not found\n", log_id(id), fmt.c_str());
+					return {};
+				}
+
+				RTLIL_BACKEND::dump_const(result, parameters.at(id));
+			}
 		} else {
 			result << *it;
 		}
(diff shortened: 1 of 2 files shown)


passes/cmds/wrapcell.cc
```diff
@@ -54,6 +85,44 @@ std::optional<std::string> format(std::string fmt, const dict<IdString, Const> &
...
+			return stringf("%s[%d:%d]", port.c_str(), base + len - 1, base);
+	}
+
+	SigSpec sample(Cell *cell)
+	{
+		return cell->getPort(port).extract(base, len);
+	}
+};
+
+std::vector<Chunk> collect_chunks(std::vector<std::pair<IdString, int>> bits)
```
widlarizer, line 111: This deserves a comment or two about how it joins contiguous runs of bits. It's not a complex function but I'd rather not have to reverse engineer it to see that it's correct

passes/cmds/wrapcell.cc
```diff
@@ -18,12 +18,37 @@
...
+		it++;
+	}
+	return false;
+}
+
+struct ContextData {
+	std::string unused_outputs;
+};
+
+std::optional<std::string> format(std::string fmt, const dict<IdString, Const> &parameters,
```
widlarizer, line 50: Maybe mark `const`

passes/cmds/wrapcell.cc
```diff
@@ -114,43 +187,88 @@ struct WrapcellPass : Pass {
...
 
-				if (d->module(name)) {
-					cell->type = name;
-					cell->parameters.clear();
-					continue;
+				subm = d->addModule(name);
+				subcell = subm->addCell("$1", cell->type);
+				for (auto conn : cell->connections()) {
+					if (ct.cell_output(cell->type, conn.first)) {
+						subcell->setPort(conn.first, SigSpec(RTLIL::Sm, conn.second.size()));
```
widlarizer, line 245: I thought the marker state was only used for temporary testing, but it's load-bearing - when I comment out setPort, the replace calls later in the file fail. Probably worth adding some comments as well. Is it acting as a placeholder as you swap the bits later? Why do `unused_outputs` still get created if they're unused and not labeled `port_output`?

passes/cmds/wrapcell.cc
```diff
@@ -114,43 +187,88 @@ struct WrapcellPass : Pass {
...
-				Module *subm = d->addModule(name);
-				Cell *subcell = subm->addCell("$1", cell->type);
-				for (auto conn : cell->connections()) {
-					Wire *w = subm->addWire(conn.first, conn.second.size());
-					if (ct.cell_output(cell->type, w->name))
-						w->port_output = true;
-					else
-						w->port_input = true;
-					subcell->setPort(conn.first, w);
+				for (auto chunk : collect_chunks(unused_outputs)) {
```
widlarizer, line 259: Interestingly, this isn't covered by the test, since commenting out this for loop doesn't make the test fail

widlarizer (approved):
The pass seems to work well but more context on the implementation would be welcome

passes/cmds/wrapcell.cc
povik: You mean `ContextData`? Sure, let's do that

passes/cmds/wrapcell.cc
povik: > Is it acting as a placeholder as you swap the bits later?

Yes. The fact that marker bits are disallowed in an RTLIL check works to our benefit because if we forget to replace some of the bits we will get caught. I've added a comment

> Why do unused_outputs still get created if they're unused and not labeled port_output?

Output ports must be connected to a wire even if unused, so we need to create those internal wires and connect them to the wrapped cell

passes/cmds/wrapcell.cc
povik: Let's add `check -assert` which should catch removing this loop

povik: No, neither the `check` command nor the internal RTLIL checker catch a marker bit on an output connection:
```
attribute \bar "w3"
attribute \foo 1
module \OR_2_3_3_unused_Y[1]

  wire output 4 \Y[2]

  wire output 3 \Y[0]

  wire width 3 input 2 \B

  wire width 2 input 1 \A

  cell $or $1
    parameter \A_SIGNED 0
    parameter \A_WIDTH 2
    parameter \B_SIGNED 0
    parameter \B_WIDTH 3
    parameter \Y_WIDTH 3
    connect \Y { \Y[2] 1'm \Y[0] }
    connect \B \B
    connect \A \A
  end
end
```
That's something to fix outside of this PR.

passes/cmds/wrapcell.cc
widlarizer: Sorry, wrong line. I meant mark `Chunk::format` `const`

PR #4804, opened 2024-12-05
read_liberty: Revisit for abc9 whiteboxes

povik:

Make two changes which make it easier to use boxes loaded in by `read_liberty` for abc9 boxes: Attach the `simple_comb_cell` attribute on box definitions which are purely combinational and have simple input/output pins. Also redo the `-unit_delay` mode to: avoid reading the `timing_type` attribute (which is optional) and avoid needing to deal with arcs from an output pin to an output pin.

### DIFF
--- a/frontends/liberty/liberty.cc
+++ b/frontends/liberty/liberty.cc
@@ -572,6 +572,8 @@ struct LibertyFrontend : public Frontend {
 			for (auto &attr : attributes)
 				module->attributes[attr] = 1;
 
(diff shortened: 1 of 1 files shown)


povik:
Pending change: setting `abc9_box` instead of the `simple_comb_cell` attribute

povik:
> Pending change: setting `abc9_box` instead of the `simple_comb_cell` attribute

Done

widlarizer (approved):
Looks good

PR #4805, opened 2024-12-06
add hash method to scopeindex cursor

nakengelhardt:

Allow `Cursor` to be stored in a `dict`.


povik (approved):
Comparing against `operator==` looks okay

PR #4815, opened 2024-12-12
Verific frontend: fix `top_bound`/`bottom_bound` attributes

nakengelhardt:

Fixes two issues:
1. Erroneous `top_bound`/`bottom_bound` attributes appear on SystemVerilog wires
2. `top_bound`/`bottom_bound` attributes are the wrong size on VHDL null ranges, and the value of the bounds may be truncated. It's not possible to tell the size of the wire from the `type_range` object in these cases.

1. Only run on nets coming from VHDL
2. Pass in an optional width hint that `import_attributes` can use to determine the size of the bounds.

Expanded tests for many port types are included.

### DIFF
--- a/frontends/verific/verific.cc
+++ b/frontends/verific/verific.cc
@@ -407,7 +407,7 @@ static const std::string verific_unescape(const char *value)
 }
 #endif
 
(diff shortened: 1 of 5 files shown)


mmicko (approved):
LGTM

PR #4817, opened 2024-12-13
macc: Stop using the B port

povik:

The B port is for single-bit summands. These can just as well be represented as an additional summand on the A port (which supports summands of arbitrary width). An upcoming `$macc_v2` cell won't be special-casing single-bit summands in any way.

In preparation, make the following changes:

 * remove the `bit_ports` field from the `Macc` helper (instead add any single-bit summands to `ports` next to other summands)

 * leave `B` empty on cells emitted from `Macc::to_cell`

An included test makes sure we interpret the B port on existing RTLIL exports.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -869,6 +869,7 @@ endif
 SH_ABC_TEST_DIRS =
 SH_ABC_TEST_DIRS += tests/memories
 SH_ABC_TEST_DIRS += tests/aiger
+SH_ABC_TEST_DIRS += tests/alumacc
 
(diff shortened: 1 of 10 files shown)


widlarizer:
I think some theoretical merit was mentioned in the past to the B port when techmapping to primitives that only support single-bit summands. Since I have no actual evidence of that being used by anyone, it's just something to keep in mind. Otherwise I'm in favor of the idea

widlarizer (approved):

PR #4824, opened 2024-12-19
yw: fix unintensional recursion in hash_into

widlarizer:
`IdPath` is an `std::vector<IdString>` with extra methods and should be hashed as `std::vector<IdString>`. That was the intention behind this implementation but it ended up creating unintential recursion


widlarizer:
To trigger SBY tests, I had to push a commit to change the actions criteria. The SBY tests passed, so I force pushed it away again and we're good to merge

PR #4825, opened 2024-12-19
yosys_common: fix gcc warning in #error directive

widlarizer:
See [here](https://stackoverflow.com/a/37195204/8048373)


lethalbit (approved):
lgtm

PR #4827, opened 2024-12-23
Update ALU MULT mode in gowin to match nextpnr

aerkiaga:
`nextpnr` currently wires an ALU in `MULT` mode in a different fashion than expected. Instead of `I2` to `GND`, it is connected to `VCC`. This has been mentioned in https://github.com/YosysHQ/nextpnr/pull/1407. While fixing this in `nextpnr` would work, leaving it as-is is arguably a better option, since it can perform the same operations plus being able to take a second operand. This would have a profound impact if choosing to synthesize multiplication using this mode.

So, instead of fixing the wiring in `nextpnr`, this PR seeks to update the expected behavior of the cell to reflect the current wiring.

`yosys` doesn't use this mode yet, and therefore no tests or workflows should be impacted.

### DIFF
--- a/techlibs/gowin/cells_sim.v
+++ b/techlibs/gowin/cells_sim.v
@@ -1005,7 +1005,7 @@ always @* begin
 			C = I0;
 		end
 		MULT: begin
-			S = I0 & I1;
+			S = (I0 & I1) ^ I3;
 			C = I0 & I1;
 		end
 	endcase


widlarizer:
I don't see I2 mentioned in the [apicula docs for the ALU primitive](https://github.com/YosysHQ/apicula/wiki/ALU), could you explain where I can take a look to see how this affects the function?

aerkiaga:
@widlarizer the documentation I was referring to is [this](https://github.com/YosysHQ/apicula/blob/master/doc/alu.md). The first table describes connections in different ALU modes. The `MULT` mode does not correspond to what `nextpnr` actually does, which corresponds to the functionality described in the last paragraph of the document. This is a superset of what the "regular" connection pattern would do, and is much more useful in practice (e.g. for implementing a Dadda tree).

widlarizer (approved):
So by wiring I2 to a different constant, we gain an extra use for the I3 given the fixed LUT4 config right? Sounds like a cool trick and I get how this lines up with what nextpnr is already doing now

PR #4829, opened 2024-12-24
dfflibmap: Adds flop name preserving pass to match write_verilog

QuantamHD:
When using the dfflibmap pass it prevents write_verilog from adding proper flop names based on the Q output port. This makes it harder to read timing reports in OpenROAD. This code only works on internal flop types and does not work for cells mapped to standard cell modules.

See code in write_verilog:
```c++
if (!norename && cell->name[0] == '$' && RTLIL::builtin_ff_cell_types().count(cell->type) && cell->hasPort(ID::Q) && !cell->type.in(ID($ff), ID($_FF_)))
	{
		RTLIL::SigSpec sig = cell->getPort(ID::Q);
		if (GetSize(sig) != 1 || sig.is_fully_const())
			goto no_special_reg_name;

		RTLIL::Wire *wire = sig[0].wire;

		if (wire->name[0] != '\\')
			goto no_special_reg_name;

		std::string cell_name = wire->name.str();

		size_t pos = cell_name.find('[');
		if (pos != std::string::npos)
			cell_name = cell_name.substr(0, pos) + "_reg" + cell_name.substr(pos);
		else
			cell_name = cell_name + "_reg";

		if (wire->width != 1)
			cell_name += stringf("[%d]", wire->start_offset + sig[0].offset);

		if (active_module && active_module->count_id(cell_name) > 0)
				goto no_special_reg_name;

		return id(cell_name);
	}
```

If the user passes -infer_flop_names to dfflibmap then it will attempt to rename flops based on the Q input port of the flop. This follows a similar renaming as in write_verilog, but that only works on $DFF and $dff based cell.

Pass -infer_flop_names to dfflibmap

### DIFF
--- a/passes/techmap/dfflibmap.cc
+++ b/passes/techmap/dfflibmap.cc
@@ -17,9 +17,13 @@
  *
  */
 
+#include <string>
+#include <optional>
+
 #include "kernel/yosys.h"
 #include "kernel/sigtools.h"
 #include "libparse.h"
+#include "kernel/rtlil.h"
 #include <string.h>
 #include <errno.h>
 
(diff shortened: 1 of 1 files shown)


povik:
How does this compare to running the `rename -wire` command post `dfflibmap`?

```
    rename -wire [selection] [-suffix <suffix>]

Assign auto-generated names based on the wires they drive to all selected
cells with private names. Ignores cells driving privatly named wires.
By default, the cell is named after the wire with the cell type as suffix.
The -suffix option can be used to set the suffix to the given string instead.
```

QuantamHD:
hmm, let me check to see. I didn't know about that command.

povik:
Fwiw ORFS does use it in their script

QuantamHD:
Yeah that definitely works. It's a little bit different than what comes out of write_verilog, but IMO close enough. Thanks @povik

PR #4831, opened 2025-01-02
Approximately double the max hash table size

mikesinouye:


mikesinouye:
@povik Would you mind taking a look once you're back from break?

kernel/hashlib.h
```diff
@@ -337,13 +337,13 @@ inline int hashtable_size(int min_size)
 		25499291, 31874149, 39842687, 49803361, 62254207, 77817767, 97272239,
 		121590311, 151987889, 189984863, 237481091, 296851369, 371064217,
 		463830313, 579787991, 724735009, 905918777, 1132398479, 1415498113,
-		1769372713
+		1769372713, 2211715897, 2764644871, 2764644887, 3455806139
```
povik, line 340: 2764644871 not a prime and similar to the number above it. I don't think it matters much for performance (the factorization is 157 * 17609203) but just so it doesn't surprise anyone let's remove it

povik (changes requested):
Looks good apart from the spurious prime

kernel/hashlib.h
mikesinouye: I had mistakenly left that in when calculating the sequence, thanks for the catch.

povik (approved):

PR #4834, opened 2025-01-06
Memory-efficient zlib usage across Liberty file consumers

### LINKED ISSUE #4830: Support `gzip`ed Liberty files
### Feature Description

Happy New Year, YosysHQ team! Please keep up the excellent work you are doing!

It looks like (I haven't tested myself) that Yosys is already supporting gzip'ed Verilog files. It would be equally useful to support gzip'ed Liberty (`.lib.gz`) files, as having zipped Liberty files in the PDK setup would be a big help for space reduction.

See also https://github.com/efabless/openlane2/issues/627

widlarizer:
In ASIC flows, Liberty files provided in the PDK can be very large and are often compressed with gzip. Prior to this PR, the only pass that accepted `.lib.gz` was `read_liberty`, just like `read_verilog` accepts `.v.gz` etc. `-liberty filename` arguments to passes like `clockgate` and `dfflibmap` didn't use the decompression code.

The `decompress_gzip` function also has been reading the entirety of the uncompressed file into RAM (into a stringstream). This is now fixed by implementing an `gzip_istream` much like the pre-existing `gzip_ostream`. It has one byte of guaranteed `unget()` which is sufficient for our Liberty parsing. Using [gzstream](https://www.cs.unc.edu/Research/compgeom/gzstream/) was ruled out because it's LGPL.

The PR carries also some shuffling of things around which should have limited impact on plugin devs since the new `kernel/io.h` header is included in `kernel/yosys_common.h`.

This will resolve #4830 but only on the Yosys side. The Yosys `abc` pass only hands off the paths specified with `abc -liberty path` to `abc` but `abc` doesn't support `.lib.gz` which is surprising since it does have its own vendored zlib and wrapper around it used for aiger


kernel/gzip.cc
```diff
@@ -0,0 +1,135 @@
...
+	}
+}
+
+#endif // YOSYS_ENABLE_ZLIB
+
+
+// Takes a successfully opened ifstream. If it's gzipped, returns an istream. Otherwise,
+// returns the original ifstream, rewound to the start.
+std::istream& uncompressed(const std::string filename, std::ios_base::openmode mode) {
+	std::ifstream& f = *new std::ifstream();
```
povik, line 100: Based on how we operate on `f` it should be a pointer

povik (commented):
I've only reviewed `gzip.cc`/`gzip.h`, it looks fine except for the nits

kernel/gzip.cc
widlarizer: Yeah I was wondering about this as well as the reference in the signature. For me, the key thing about references is that they're guaranteed non-null. I really enjoy having that invariant. This code upholds it at creation since `new` throws an exception instead of ever returning null. But since deleting doesn't make it fall out of scope, it should probably go back to being handled with the conventional paranoia associated with a pointer.

widlarizer:
- [ ] Look at `stat -liberty`

KrystalDelusion:
> someone™ needs to build and test the Wasm port

I take it that's more than just the WASI test build included in CI?

KrystalDelusion:
```
.../yosys-src/tests/various$ yowasp-yosys stat.ys

 Yosys 0.51+41 (git sha1 980a0a15c, ccache clang 18.1.2 -O3 -flto -flto)

-- Executing script file `stat.ys' --

1. Executing RTLIL frontend.
Input filename: <<EOF
Added regex 'Chip area for module '\\top': 9.072000' to expected log messages list.
ERROR: File `../../tests/liberty/foundry_data/sg13g2_stdcell_typ_1p20V_25C.lib.filtered.gz' is a gzip file, but Yosys is compiled without zlib.
ERROR: Expected log pattern 'Chip area for module '\\top': 9.072000' not found !
```
Hm.  So it doesn't work in `yowasp-yosys`, but it also hasn't broken anything.  `yowasp-yosys tests/techmap/shiftx2mux.ys` passes (along with a few other random tests I tried), which should be exercising the temp file stuff which moved.  This should be fine to merge imo.

KrystalDelusion (approved):

PR #4837, opened 2025-01-08
write_json: add option to include $scopeinfo cells

nakengelhardt:

Linty want to use the JSON output to build interactive netlist visualization, including of flattened designs. The information carried by `$scopeinfo` cells is valuable for that purpose.

Add an argument to override the logic that currently skips `$scopeinfo`.

Testcase included


povik:
I think we can start emitting those unconditionally: https://github.com/YosysHQ/yosys/pull/4084#issuecomment-1887366907

whitequark (approved):
LGTM. Ideally nextpnr can benefit from these too.

PR #4839, opened 2025-01-10
Add option for a custom flatten block separator char

mikesinouye:
Enables elaborated instance name parity with the default behavior of a commercial synthesis tool.

New option in `flatten` for a custom block separator string.

flatten -separator /

### DIFF
--- a/passes/techmap/flatten.cc
+++ b/passes/techmap/flatten.cc
@@ -28,22 +28,22 @@
 USING_YOSYS_NAMESPACE
 PRIVATE_NAMESPACE_BEGIN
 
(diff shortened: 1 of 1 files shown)


mikesinouye:
Hey @povik, whenever you get a chance if you could weigh in on this change, thank you!

povik (approved):

povik:
I wonder if there are other places we should be offering this option for consistency, but I can't think of anything atm

gadfort:
@povik the synth command can take in -flatten I think it would be helpful if we could set it there too

gadfort:
@povik I didn't know that was an option, I don't see that in the docs, how do you use it?

PR #4841, opened 2025-01-10
abc: extract compressed Liberty files to temporary directory

widlarizer:
This PR will extend abc passes to use the builtin gzip capabilities of yosys to decompress `.lib.gz` files to the temporary directories used to interact with `abc`. This will allow `-liberty <filename>` arguments to all passes to support `.lib.gz` consistently, even if it's a bit clunky and wasteful. This is sketchy, because on many systems, `/tmp/` is backed by `tmpfs` which is purely backed by system memory, and the point of using `.lib.gz` is to handle very large Liberty files.

- [X] unify with `abc` and `abc9` ~~through inheritance~~ - added `passes/techmap/abc_prep.h`
- [ ] add tests

This code also adds the cleanup functionality missing from `abc_new` by holding a single per-run top-level directory which then contains the extracted Liberty files as well as the current per-module temporary directories and removing the top-level directory at the end unless `-nocleanup` is set.

The best way to test this would be to run `abc`, `abc9` and `abc_new` passes with asap7 .lib.gz files available for example [here](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/tree/master/flow/platforms/asap7/lib/CCS) with the `-nocleanup` and inspect the directories created as `_tmp_` under the current working directory and the actually invoked abc command

The PR is rooted in the branch for #4834. and until that's merged and this PR is rebased. it will have a spammy diff against main


QuantamHD:
It looks like ABC already has a zlib dep in tree. Might be easier to just add support for .lib.gz in ABC rather than complicating Yosys.

widlarizer:
Superseded by upstream support, nice https://github.com/berkeley-abc/abc/pull/362

PR #4842, opened 2025-01-13
enforce that hdlname/scopename is used consistently with public/private names

nakengelhardt:

A number of passes and utilities use code of the shape `if(foo->name.isPublic()) { foo->get_hdlname_attribute(); }` which doesn't work as expected if the implicit invariant to use `hdlname` on public objects and `scopename` on private objects isn't upheld. Since this invariant is easy to break accidentally, enforce it.

Add checks in `Module::check()` to make sure the invariant holds for Module, Wire, Cell, and Memory names. (Should processes also be checked?)

Run larger flows and examples, and rarely used passes, especially in combination with `flatten -scopename`.

### DIFF
--- a/frontends/verific/verific.cc
+++ b/frontends/verific/verific.cc
@@ -1464,7 +1464,8 @@ void VerificImporter::import_netlist(RTLIL::Design *design, Netlist *nl, std::ma
 		log("Importing module %s.\n", RTLIL::id2cstr(module->name));
 	}
 	import_attributes(module->attributes, nl, nl);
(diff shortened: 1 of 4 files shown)


nakengelhardt:
There are currently too many cases that don't respect this invariant, so we'll defer the checks for now. We should introduce a group of helpers for copying attributes, which get some context on whether the destination object replaces or duplicates the origin object, and can make sure to handle this kind of invariant under the hood without each pass having to know all the details of attribute semantics.

Issue #4844, opened 2025-01-14
UB in celledges shift handling


FSY369:
### Version

Yosys 0.48+47 (git sha1 cbb95cb51, clang++ 18.1.8 -fPIC -O3)

### On which OS did this happen?

Linux

### Reproduction Steps

1. Enter yosys in the terminal
2. read the rtl.v file, read_verilog rtl.v
3. synthesize the command, synth
4. synthesis failed

[rtl.v.log](https://github.com/user-attachments/files/18405090/rtl.v.log)

### Expected Behavior

Synthesize Success

### Actual Behavior

synthesis failed，2.5. Executing CHECK pass (checking for obvious problems).
Checking module module128...
Checking module module150...
浮点数例外 (核心已转储)

whitequark:
You are submitting a bug report produced using fuzzing tools. With such bug reports, it is often hard to tell whether the issue reported is an issue in Yosys or an issue with the test harness.

You [were instructed](https://github.com/YosysHQ/yosys/issues/4828) how to submit such bug reports properly and you are not following the instructions. As such I am closing your report as invalid.

FSY369:
Sorry, I uploaded the wrong file, here is the minimal test case, the problem occurs with the assignment of reg161 thus causing the synthesis to fail.

[rtl.v.log](https://github.com/user-attachments/files/18405287/rtl.v.log)

whitequark:
Thanks. For other readers, it seems that:

> 浮点数例外 (核心已转储)

means

> Floating point exception (core dumped)

whitequark:
I have reduced your testcase further using [C-Reduce](https://github.com/csmith-project/creduce). Here is the actually minimal testcase:

```verilog
module a(b, c, d);
	output b;
	input c;
	input d;
	always b <= $signed(f) << (0 ? 0 : g);
endmodule
```

I have done this by putting your code into `issue4844.v`, writing a shell script `issue4844.sh` that checks for a floating point error (code 136) being returned:

```sh
#!/bin/sh
yosys issue4844.v -p synth
[ $? -eq 136 ]
```

and running C-Reduce as:

```console
$ creduce --not-c issue4844.sh issue4844.v
```

whitequark:
What happens is that in the following code:

https://github.com/YosysHQ/yosys/blob/6225abec71fe2cdc6e1fd32b727d9d4b9e3319fc/kernel/celledges.cc#L250-L257

[...]

PR #4845, opened 2025-01-14
CI: Sanitize for undefined behavior

whitequark:

Yosys includes some amount of buggy code that invokes undefined behavior, e.g. https://github.com/YosysHQ/yosys/issues/4844. This should be fixed and stay fixed.

By running the Undefined Behavior Sanitizer on each commit/PR. UBSan does not have false positives and has a slim impact on performance, so this is done unconditionally.

Add `SANITIZER = undefined` to your `Makefile.conf`, then run `make clean && make && UBSAN_OPTIONS=halt_on_error=1 make test`.

### DIFF
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -56,6 +56,7 @@ jobs:
           mkdir build
           cd build
           make -f ../Makefile config-$CC
+          echo 'SANITIZER = undefined' >> Makefile.conf
           make -f ../Makefile -j$procs ENABLE_LTO=1
 
       - name: Log yosys-config output
@@ -82,6 +83,7 @@ jobs:
     if: needs.pre_job.outputs.should_skip != 'true'
     env:
       CC: clang
+      UBSAN_OPTIONS: halt_on_error=1
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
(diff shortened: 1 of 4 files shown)


whitequark:
We should probably exclude the whole `libs/*` subtree from UBSan checking since it's not like we're going to fix those bugs anyway. I think we should use `-fsanitize-ignorelist=FILE` where `FILE` contains:

```
src:libs/*
```

or maybe 

```
[alignment]
src:libs/fst/*
```

What should `FILE` be? I'm thinking `.sanitize-ignorelist` or something.

whitequark:
The CI does fail in an unfortunately opaque way (the ubsan output gets redirected to `tests/simple/memory.err`) but I've confirmed that this is indeed #4844:

```
kernel/celledges.cc:262:30: runtime error: shift exponent 32 is too large for 32-bit type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior kernel/celledges.cc:262:30 in 
```

KrystalDelusion:
There is a `Report errors` step in the tests yaml that runs `find tests/**/*.err -print -exec cat {} \;`.  It doesn't always work, but in this case it does give the ubsan output you're looking for:
https://github.com/YosysHQ/yosys/actions/runs/12762313299/job/35571164274?pr=4845#step:12:4995

whitequark:
Ah, that's handy.

povik:
The CI is green now, someone please review

KrystalDelusion:
I think
```
[alignment]
src:libs/fst/*
```
in `.sanitize-ignorelist` is good (with appropriate gating I guess?)

povik:
The one instance of UB has been easy to patch. I'd only start the exclusion list if issues with that library keep coming back.

KrystalDelusion:
Oh!  I missed that you patched the library 😅 lgtm

KrystalDelusion (approved):

PR #4849, opened 2025-01-14
hashlib: merge hash_ops with hash_top_ops for plugin compat

widlarizer:
When writing #4524 I made it so that a type can have an implementation for how to hash a single instance (`hash_top_ops`) distinct from one for how to hash an instance in the context of an on-going context (`hash_ops`). This PR implements @povik's idea to merge them together so that plugin code can still explicitly pass `hash_ptr_ops` into associative data structures like `dict`.

- [ ] adjust `docs/source/yosys_internals/hashing.rst` to match

### DIFF
--- a/kernel/hashlib.h
+++ b/kernel/hashlib.h
@@ -142,7 +142,7 @@ using Hasher = HasherDJB32;
 #define HASH_TOP_LOOP_FST [[nodiscard]] static inline Hasher hash
 #define HASH_TOP_LOOP_SND { \
 	Hasher h; \
-	h.eat(a); \
+	h = hash_into(a, h); \
 	return h; \
 }
 
@@ -239,11 +239,7 @@ struct hash_cstr_ops {
 			h.hash32(*(a++));
 		return h;
 	}
-	[[nodiscard]] static inline Hasher hash(const char *a) {
-		Hasher h;
-		h = hash_cstr_ops::hash_into(a, h);
-		return h;
-	}
+	HASH_TOP_LOOP_FST (const char *a) HASH_TOP_LOOP_SND
 };
 
 template <> struct hash_ops<char*> : hash_cstr_ops {};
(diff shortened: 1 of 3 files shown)


kernel/hashlib.h
```diff
@@ -147,15 +137,14 @@ class HasherDJB32 {
...
-	static inline bool cmp(const T &a, const T &b) {
-		return hash_ops<T>::cmp(a, b);
-	}
-	static inline Hasher hash(const T &a) {
-		return hash_ops<T>::hash_into(a, Hasher());
-	}
-};
+// Boilerplate compressor for trivially implementing
+// top-level hash method with hash_into
+#define HASH_TOP_LOOP_FST [[nodiscard]] static inline Hasher hash
```
povik, line 142: Why not put all of the function into one macro, so that you would write e.g. `DEFAULT_HASH_TOP(std::pair<P, Q> a)` to use it?

kernel/hashlib.h
```diff
@@ -245,6 +239,11 @@ struct hash_cstr_ops {
 			h.hash32(*(a++));
 		return h;
 	}
+	[[nodiscard]] static inline Hasher hash(const char *a) {
```
povik, line 242: Why can't we use the macro here?

kernel/hashlib.h
widlarizer: I tried doing this the simplest possible way I could think of, passing a type into a macro argument, and it wasn't valid preprocessor code for some reason. We can change this later

kernel/hashlib.h
widlarizer: instead of calling the polymorphic `h.eat(...)` we explicitly call the `hash_into` method of self

widlarizer: as for why we do *that*, it's because I'm paranoid about pointer conversions

povik: Can't we use hash_into in the macro helper?

PR #4852, opened 2025-01-16
Fix undefined type error

gsomlo:
Explicitly include <cstdint> in libs/json11/json11.cpp, to avoid:

    "error: ‘uint8_t’ does not name a type"

compile errors when building against TCL 9.0
(https://bugzilla.redhat.com/show_bug.cgi?id=2337817)

### DIFF
--- a/libs/json11/json11.cpp
+++ b/libs/json11/json11.cpp
@@ -24,6 +24,7 @@
 #include <cmath>
 #include <cstdlib>
 #include <cstdio>
+#include <cstdint>
 #include <limits>
 
 namespace json11 {


povik:
I don't see how this could be related to Tcl 9.0 but the change itself is desirable. Please update the commit message if you can to remove the mention of Tcl.

gsomlo:
On Thu, Jan 16, 2025 at 10:38:54AM -0800, Martin Povišer wrote:
> I don't see how this could be related to Tcl 9.0

Might be just a coincidence -- it only happens to me during
mock-building an RPM for fedora when the build environment
has been configured with the latest Tcl 9.0.

> but the change itself is desirable. Please update the commit
> message if you can to remove the mention of Tcl.

Done. Thanks for the quick feedback!

Best,
--Gabriel

povik (approved):

Issue #4843, opened 2025-01-13
support for flatten.separator in `synth`


gadfort:
The way you would use it is you would have
```
scratchpad -set flatten.separator /
```
or similar at the top of your script.

Then we need code inside flatten which would pick this up.

_Originally posted by @povik in https://github.com/YosysHQ/yosys/issues/4839#issuecomment-2587342058_

gadfort:
@povik I created an issue to track this.

PR #4856, opened 2025-01-18
add support for passing flatten -separator to flatten in synth

gadfort:

Closes https://github.com/YosysHQ/yosys/issues/4843

Since https://github.com/YosysHQ/yosys/pull/4839 added this flag to flatten, this change allows for the user to specify this for the synth command.

```
scratchpad -set flatten.separator "/"
synth -flatten
write_verilog test.v
```

### DIFF
--- a/passes/techmap/flatten.cc
+++ b/passes/techmap/flatten.cc
@@ -357,6 +357,9 @@ struct FlattenPass : public Pass {
 
 		FlattenWorker worker;
 
+		if (design->scratchpad.count("flatten.separator"))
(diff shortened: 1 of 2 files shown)


whitequark:
The scratchpad check should be in the `flatten` pass, not `synth`, shouldn't it?

gadfort:
@whitequark sure, I can move it there (makes more sense there)

gadfort:
Looks like the failures might be related to the switch from ubuntu22 to ubuntu24 on the github runners

povik (approved):

PR #4858, opened 2025-01-19
Update workflow(s) for ubuntu-latest == ubuntu-24.04

KrystalDelusion:
ubuntu-latest is now 24.04 instead of 22.04 and broke some things:
- test-compile for {os: ubuntu-latest, compiler: clang-14} failing cpp setup
- test-build for {os: ubuntu-latest} missing bzlib.h

Change clang-14 to clang-10 as oldest supported under ubuntu-latest (somehow clang-10 is now working under latest but clang-11 through clang-16 are not).
Also update newest clang to -19.
apt-get libbz2-dev.

### DIFF
--- a/.github/actions/setup-build-env/action.yml
+++ b/.github/actions/setup-build-env/action.yml
@@ -8,7 +8,7 @@ runs:
       shell: bash
       run: |
         sudo apt-get update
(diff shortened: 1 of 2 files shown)


KrystalDelusion:
[13fdd52](https://github.com/YosysHQ/yosys/pull/4858/commits/13fdd5246a437ef830061afc2129f6fd4d237191) fixes tests for ubuntu, but breaks macos.  Minimum fix to get CI passing is just put it behind a `if: runner.os == 'Linux'`, but I'm not sure where the library was coming from before and if this makes something else redundant or if it should be included in the general prerequisites.

mmicko (approved):

PR #4860, opened 2025-01-20
Add groups to command reference

KrystalDelusion:
The cmdref is currently just a big list of all commands and can be a bit difficult to navigate.

Refactor cmds to follow cells pattern (i.e. export to json and import with autodoc).
Group commands based on path, with fallback to name if path is not available (since it requires `source_location`).

[Preview build](https://yosyshq.readthedocs.io/projects/yosys/en/docs-preview-cmdref_groups/)


KrystalDelusion:
Rebased to latest main (v`0.49+3`) to fix CI

KrystalDelusion:
Rebased to latest main (v`0.50+59`) for pmgen restructure

KrystalDelusion:
Rebased on latest main (v`0.51+65`) to resolve merge conflicts

widlarizer:
There's a huge amount of TODOs shown directly on the linked page, I guess that's not on purpose right?

KrystalDelusion:
> There's a huge amount of TODOs shown directly on the linked page, I guess that's not on purpose right?

TODOs are normally hidden for published builds, but not previews or local builds.  There might be a couple new TODOs added in this PR, but it shouldn't be anything blocking; I've taken to using `.. TODO::` for blocking and `.. todo::` for non-blocking, but the difference is only visible in the source.  The root index also collates all of the TODOs into one convenient location, but again they won't show up on the published build.

KrystalDelusion:
Rebased on latest main (v`0.55+46`)

PR #4863, opened 2025-01-21
Fix for WindowsApps permission denied with run_command().

alchitry:
I've been integrating Yosys into an application ([Alchitry Labs](https://github.com/alchitry/Alchitry-Labs-V2)) and it fails to run on Windows when installed to WindowsApps. The original error was reported [here](https://forum.alchitry.com/t/yosys-error-cant-guess-frontend-for-input-file/1585/4?u=alchitry).

I found that my program could run `yosys` but when `yosys` would start `yosys-abc` it would simply return `Access is denied.`

After two days of pulling my hair out, I discovered that using [`CreateProcessA()`](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessa) instead of `popen()` or `system()` makes it work. I don't know exactly what the difference is but Windows must be passing something along that doesn't happen otherwise.

This is a Windows only problem and a Windows only solution so a lot was added in an `#ifdef` block. I couldn't find a way to make it work using the more standard functions.

### DIFF
--- a/kernel/yosys.cc
+++ b/kernel/yosys.cc
@@ -312,22 +312,23 @@ int run_command(const std::string &command, std::function<void(const std::string
 	sa.bInheritHandle = TRUE;
 	sa.lpSecurityDescriptor = NULL;
 
-	if (!CreatePipe(&stdout_read, &stdout_write, &sa, 0)) {
-		fprintf(stderr, "CreatePipe failed: %d\n", GetLastError());
-		return -1;
-	}
+	if (process_line) {
+		if (!CreatePipe(&stdout_read, &stdout_write, &sa, 0)) {
+			fprintf(stderr, "CreatePipe failed: %d\n", GetLastError());
+			return -1;
+		}
 
-	if (!SetHandleInformation(stdout_read, HANDLE_FLAG_INHERIT, 0)) {
-		fprintf(stderr, "SetHandleInformation failed: %d\n", GetLastError());
-		CloseHandle(stdout_read);
-		CloseHandle(stdout_write);
-		return -1;
-	}
+		if (!SetHandleInformation(stdout_read, HANDLE_FLAG_INHERIT, 0)) {
+			fprintf(stderr, "SetHandleInformation failed: %d\n", GetLastError());
+			CloseHandle(stdout_read);
+			CloseHandle(stdout_write);
+			return -1;
+		}
 
-	// Set STARTUPINFO to redirect stdout/stderr
-	si.hStdError = stdout_write;
-	si.hStdOutput = stdout_write;
-	si.dwFlags |= STARTF_USESTDHANDLES;
+		// Set STARTUPINFO to redirect stdout
+		si.hStdOutput = stdout_write;
+		si.dwFlags |= STARTF_USESTDHANDLES;
+	}
 
 	// Convert command to char* for CreateProcessA
 	char* command_cstr = const_cast<char*>(command.c_str());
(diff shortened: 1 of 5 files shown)


kernel/yosys.cc
```diff
@@ -295,6 +295,102 @@ bool patmatch(const char *pattern, const char *string)
...
+
+	if (!SetHandleInformation(stdout_read, HANDLE_FLAG_INHERIT, 0)) {
+		fprintf(stderr, "SetHandleInformation failed: %d\n", GetLastError());
+		CloseHandle(stdout_read);
+		CloseHandle(stdout_write);
+		return -1;
+	}
+
+	// Set STARTUPINFO to redirect stdout/stderr
+	si.hStdError = stdout_write;
```
whitequark, line 328: As far as I know, `popen` redirects stdout only, not stderr.

kernel/yosys.cc
```diff
@@ -295,6 +295,102 @@ bool patmatch(const char *pattern, const char *string)
...
+				line = line.substr(newlinePos + 1);
+				newlinePos = line.find('\n');
+			}
+		}
+
+		// Process any remaining partial line
+		if (!line.empty()) {
+			process_line(line);
+		}
+	}
```
whitequark, line 380: For the case where you are not reading from the pipe, will you not end up with the pipe buffer getting full and deadlocking both processes?

kernel/yosys.cc
alchitry: I believe you're correct.

kernel/yosys.cc
alchitry: I improved the code for this case.

alchitry:
I improved the handling for when process_line is null to be more like `system()`. I also only redirect stdout not stderr.

`system()` uses a shell to launch the process while `CreateProcess()` does not. Using the shell is likely causing the problems but it means that stuff like redirection doesn't work with `CreateProcess()`. The launch of `yosys-abc` uses `2>&1` so it might be more accurate to also redirect stderr.

alchitry:
I updated the code so it only uses the new function when needed. This preserves the shell behavior everywhere else.

whitequark:
> The launch of `yosys-abc` uses `2>&1` so it might be more accurate to also redirect stderr. It could be conditionally redirected if the command ends with this.

Without the use of a shell, wouldn't this argument (however the redirection is done) be passed to `yosys-abc` itself, causing issues?

> Using the shell is likely causing the problems

Can you please confirm that running `cmd.exe` via `CreateProcess` is also causing the error you're observing?

whitequark:
> I found that my program could run `yosys` but when `yosys` would start `yosys-abc` it would simply return `Access is denied.`

Are you aware that you can link abc into the main Yosys executable, avoiding the use of `popen` entirely? On reflection, I think this might be the better way to go to solve this issue.

alchitry:
That would likely fix it. Can it be done using the oss-cad-suite-build? I'm using this to build the tools for windows/linux/mac.

whitequark:
> Can it be done using the oss-cad-suite-build?

Sorry, I'm not familiar with oss-cad-suite-build. I maintain [YoWASP](https://yowasp.org) tools and that's been my primary deployment target since 2020.

alchitry:
I believe I was able to add this by adding `echo 'LINK_ABC := 1' >> Makefile.conf` as line 11 of [yosys.sh](https://github.com/YosysHQ/oss-cad-suite-build/blob/main/default/scripts/yosys.sh) but this lead to a build error.

```
/usr/lib/gcc/x86_64-w64-mingw32/13.2.1/../../../../x86_64-w64-mingw32/bin/ld: yosys-libabc.a(sclLiberty.o): in function `Scl_LibertyGlobMatch':
/work/_builds/windows-x64/yosys/yosys/abc/src/map/scl/sclLiberty.c:86: undefined reference to `__imp_PathMatchSpecA'
collect2: error: ld returned 1 exit status
make: *** [Makefile:718: yosys.exe] Error 1
```

This lead me to line 86 of `sclLiberty.c` which conveniently had the comment of `// if the compiler complains, add "-lshlwapi"`

Adding `-lshlwapi` to `LIBS` in the makefile did the trick. I need to figure out a cleaner way to only add it to Windows builds but it's a good start.

whitequark:
> Adding `-lshlwapi` to `LIBS` in the makefile did the trick. I need to figure out a cleaner way to only add it to Windows builds but it's a good start.

That should probably be done as a part of the main Yosys makefile.

PR #4876, opened 2025-01-29
liberty expression lexer check if characters are found and use size if not for `pin()`

gadfort:

When the clocked_on expression contains an expression like: `D&!SE+SE&SI` the lexer segfaults when processing the SI pin because `find_first_of` returns `std::string::npos` causing the calls to `strsub` to throw an exception.

Check the return value and use the size of the string if those characters are not found.

I have a liberty file which was modified from one of the yosys testcases I can add, but I might need to guidance on how to integrate that as a test.
See attached (added the sdff cell for testing): [normal_scan_ff.zip](https://github.com/user-attachments/files/18594737/normal_scan_ff.zip)

### DIFF
--- a/passes/techmap/libparse.h
+++ b/passes/techmap/libparse.h
@@ -59,6 +59,10 @@ namespace Yosys
 
 			std::string pin() {
 				auto length = s.find_first_of("\t()'!^*& +|");
+				if (length == std::string::npos) {
+					// nothing found so use size of s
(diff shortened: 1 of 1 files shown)


gadfort:
@povik do you mind taking a look at this?

povik (approved):

povik:
Thanks!

PR #4879, opened 2025-01-30
Fixing undefined behaviours

### LINKED ISSUE #4844: UB in celledges shift handling

KrystalDelusion:

#4845 (and https://github.com/YosysHQ/yosys/issues/4844#issuecomment-2589140999) reveal undefined behaviors in the Yosys code.  This PR aims to resolve them.

- [x] kernel/celledges.cc
`b_width_capped` already exists for preventing arithmetic overflow, limiting the value of `b_width` to 30.  This just changes the left shifts to also use it. The caveat of incorrect results for extremely large values of `a_width` still applies, as does the improbability of that actually happening. This fixes #4844 (or at least, the floating point exception; the circuit still isn't valid but I think that's fine).
- [x] frontends/ast/simplify.cc
In the `nowrshmsk` handling, `log2(0)` returns -inf, which gives undefined behaviour when casting to an int.  So catch the case when the stride is 0 and just set the stride_width to 0.

Follow instructions for testing #4845 on this branch

### DIFF
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -2936,7 +2936,10 @@ bool AstNode::simplify(bool const_fold, int stage, int width_hint, bool sign_hin
 					lsb_expr->children[stride_ix]->detectSignWidth(stride_width, stride_sign);
 					max_width = std::max(i_width, stride_width);
 					// Stride width calculated from actual stride value.
-					stride_width = std::ceil(std::log2(std::abs(stride)));
+					if (stride == 0)
+						stride_width = 0;
+					else
+						stride_width = std::ceil(std::log2(std::abs(stride)));
 
 					if (i_width + stride_width > max_width) {
 						// For (truncated) i*stride to be within the range of dst, the following must hold:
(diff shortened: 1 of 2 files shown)


KrystalDelusion:
@povik the `b_width_capped` addition and previous left shift checks have you on the blame, so if you could sanity check that the celledges changes are okay I would appreciate it :)

whitequark:
I don't quite understand that code so I don't feel like I can sign off on it, apologies.

povik (approved):
Thanks again!

povik:
I think the simplify.cc fix is fine too

PR #4881, opened 2025-01-31
Move passes out of the passes/pmgen folder

KrystalDelusion:

In the Docs JF while reviewing #4860 it was brought up that architecture specific pmgen passes would be better served in the same folder as the rest of their architecture specific code.

@nakengelhardt suggested making this a separate PR for discussion at Dev JF in case people have any other passes they think should be moved at the same time.

Techlib pmgens are now in relevant techlibs/*.
`peepopt` also moved and is now in passes/opt.
`test_pmgen` is still in passes/pmgen.


KrystalDelusion:
(oops I figured out why the `mkdir -p $(dir $@)` was there)

PR #4887, opened 2025-02-09
Remove obsolete Makefile docs/images clean call

JasonBrave:

Every time when I run `make config-gcc` or `make clean` it would give an error of `docs/images` directory could not be found.

Remove Makefile call to `docs/images clean` from top level makefile, the new `docs/sources/_images` clean target would be called by `docs` Makefile clean target


KrystalDelusion (approved):
This is fine for merge, but I would also note that while the error message is printed, it is ignored by the top level `clean` target.

Issue #4883, opened 2025-02-02
Regression. Segmentation fault at commit 901935fbcec42cbf79594efd9cc2877877810395715


yrabbit:
### Version

Yosys 0.49+10 (git sha1 18a7c00382cd64d46b61ff6cafe80851ca29cb77, c++ 8.3 -O2 -fno-strict-aliasing -fPIC -O3)

### On which OS did this happen?

Linux, BSD

### Reproduction Steps

Compile the attached test design

``` shell
yosys -p "read_verilog ram.v; synth_gowin -json ram-synth.json"
```

test file:

[ram.v.txt](https://github.com/user-attachments/files/18630656/ram.v.txt)

### Expected Behavior

Error free compilation as it was up to "bad" commit https://github.com/YosysHQ/yosys/commit/901935fbcec42cbf79594efd9cc2877810395715

### Actual Behavior

Segmentation fault (core dumped)

```
2.8.14.9. Executing OPT_CLEAN pass (remove unused cells and wires).
Finding unused cells or wires in module \top..
Removed 4 unused cells and 4 unused wires.
<suppressed ~5 debug messages>

2.8.14.10. Executing MEMORY_COLLECT pass (generating $mem cells).

2.8.15. Executing OPT_CLEAN pass (remove unused cells and wires).
Finding unused cells or wires in module \top..

2.9. Executing MEMORY_LIBMAP pass (mapping memories to cells).
Segmentation fault (core dumped)
```

povik:
@widlarizer FYI

widlarizer:
Thanks for the quality report. The commit you bisected to seems to only expose the issue though since I can detect memory problems with valgrind on the reproducer with earlier versions as well
```
$ valgrind ./uut/0.45/bin/yosys -p "read_verilog garbage/ram.v; synth_gowin -json ram-synth.json" >/dev/null
==32735== Memcheck, a memory error detector
==32735== Copyright (C) 2002-2024, and GNU GPL'd, by Julian Seward et al.
==32735== Using Valgrind-3.24.0 and LibVEX; rerun with -h for copyright info
==32735== Command: ./uut/0.45/bin/yosys -p read_verilog\ garbage/ram.v;\ synth_gowin\ -json\ ram-synth.json
==32735==
==32735== Invalid read of size 4
[...]

PR #4890, opened 2025-02-10
Fix PortBit segfault

Anhijkt:
This is the fix for the [4883](https://github.com/YosysHQ/yosys/issues/4883) issue. I didn't delete the old code because I'm not sure that the new one wont fail on some corner cases. Although the new code passes all tests.

### DIFF
--- a/kernel/modtools.h
+++ b/kernel/modtools.h
@@ -305,9 +305,10 @@ struct ModWalker
 	struct PortBit
 	{
 		RTLIL::Cell *cell;
+		RTLIL::IdString cell_name;
 		RTLIL::IdString port;
 		int offset;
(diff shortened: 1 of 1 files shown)


widlarizer:
Thanks for the PR. `valgrind` still sees an invalid write, this hasn't fixed the root cause. I'll try to take a look later

widlarizer:
A PortBit should not outlive the cell it points to. Dereferencing it to remember its name is only hiding the root cause for sure which is that the cell is removed from the module but a PortBit keeps a dangling pointer to it. You pointed out that the Cell can not be copied - that's true, and as its assignment and copy code is deleted, it never gets copied. This code never copied it, it only copied a pointer to it, which is fine, as long as the lifetimes are all good, which we see they actually aren't

PR #4895, opened 2025-02-12
share: fix infinite loop in find_terminal_bits on $mux loop

### LINKED ISSUE #4260: `share` stuck on a combinational loop
### Version

Yosys 0.38+120 (git sha1 1e42b4f0f, clang++ 11.1.0 -fPIC -Os)

macOS

### Reproduction Steps

```
read_verilog <<EOF
module top(a, b, y);
	input wire a;
	input wire b;
	output wire y;

	wire j = k ? a : ~k;
	wire k = l ? ~j : b;
	wire l = j ? ~k : b;

	assign y = l;
endmodule
EOF
share
```

### Expected Behavior

`share` not stuck

### Actual Behavior

`share` stuck

widlarizer:
Previously, `share` assumed there is no combinational loop composed of `$mux` cells when collecting terminal bits. This PR uses the already implemented `visited_cells` to avoid running forever in the case of a `$mux` cell loop. Fixes #4260

### DIFF
--- a/passes/opt/share.cc
+++ b/passes/opt/share.cc
@@ -87,7 +87,7 @@ struct ShareWorker
 			queue_bits.clear();
 
 			for (auto &pbit : portbits) {
-				if (pbit.cell->type == ID($mux) || pbit.cell->type == ID($pmux)) {
+				if ((pbit.cell->type == ID($mux) || pbit.cell->type == ID($pmux)) && visited_cells.count(pbit.cell) == 0) {
 					pool<RTLIL::SigBit> bits = modwalker.sigmap(pbit.cell->getPort(ID::S)).to_sigbit_pool();
 					terminal_bits.insert(bits.begin(), bits.end());
 					queue_bits.insert(bits.begin(), bits.end());


widlarizer:
Hmm maybe this should be an assertion failure actually because it's like finding a loop with `check -assert`

widlarizer:
That doesn't work due to tripping on even something simple as this, so I guess we'll just silently ignore it as was my original plan
```
module \m
  wire width 8 output 1 \foo

  cell $mux $procmux$25
    parameter \WIDTH 8
    connect \A 8'x
    connect \B 8'x
    connect \S 1'x
    connect \Y \foo
  end

end
```

passes/opt/share.cc
```diff
@@ -87,7 +87,7 @@ struct ShareWorker
 			queue_bits.clear();
 
 			for (auto &pbit : portbits) {
-				if (pbit.cell->type == ID($mux) || pbit.cell->type == ID($pmux)) {
+				if ((pbit.cell->type == ID($mux) || pbit.cell->type == ID($pmux)) && visited_cells.count(pbit.cell) == 0) {
```
povik, line 90: I think we can add this check for all cells, not only muxes

Issue #4689, opened 2024-10-29
Segfault in XAIGER


Periodic1911:
### Version

Yosys 0.46+11 (git sha1 0200a7680, g++ 14.2.1 -march=x86-64 -mtune=generic -O2 -fno-plt -fexceptions -fstack-clash-protection -fcf-protection -fPIC -O3)

### On which OS did this happen?

Linux

### Reproduction Steps

Make a file `test.v` with:

```verilog
module test(input [1:0] a, b,
            input c,
            output [1:0] q); 

wire s;
assign s = c ? q[0] : q[1];
assign q = s ? a : b;

endmodule
```

And run `yosys -p "synth_ice40" test.v`

FWIW, it does not segfault when you swap the `assign` statements.

### Expected Behavior

Either an error because of a logic loop or a synthesized design. Not a segfault, in any case.

### Actual Behavior

Segfault in the XAIGER backend.

```
...

2.41.13.3. Executing XAIGER backend.
Segmentation fault (core dumped)
```

Full output: https://gist.github.com/Periodic1911/3618ee7299c6f9f69e6e32a9a547c4b9

PR #4898, opened 2025-02-13
write_xaiger: Detect and error on combinatorial loops

Anhijkt:
This is an attempt to fix stack overflow in [#4689](https://github.com/YosysHQ/yosys/issues/4689) issue.

This is achieved by finding cycles in XAIG. The code was taken from [aiger.cc](https://github.com/YosysHQ/yosys/commit/ac6fcb2547085d041a432a3182a3bfb2c3d2e6ea).

This fix currently passes all yosys test, so I would suggest to test it with large combinational circuits.

### DIFF
--- a/backends/aiger/xaiger.cc
+++ b/backends/aiger/xaiger.cc
@@ -53,6 +53,8 @@ struct XAigerWriter
 	dict<SigBit, float> arrival_times;
 
 	vector<pair<int, int>> aig_gates;
+	vector<SigBit> bit2aig_stack;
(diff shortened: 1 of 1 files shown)


KrystalDelusion (approved):
If the same code is being used in both `aiger.cc` and `xaiger.cc` I wonder if it wouldn't be better for them to share the code instead of repeating it.  But given how much of the code is already duplicated between the two I don't see any reason to block this PR on that basis.

PR #4901, opened 2025-02-14
Copy `abc` stuff for pyosys to enable use of the `abc` pass

akashlevy:

Pyosys is incapable of using the `abc` pass without this change when using `make install ENABLE_PYOSYS=1`, as the binary is missing.

Copy over `yosys-abc` binary and `share` directory to Pyosys install directory.

Try calling the `abc` pass in `pyosys` before and after the change when using `make install ENABLE_PYOSYS=1` with a virtual environment


KrystalDelusion:
Has this broken since #4643?  Or is that specific to building wheels, and if so does this break that?

akashlevy:
That is specific to building wheels. This fixes the same issue for `make install` target when using a Python virtual environment or installing directly to site packages

KrystalDelusion:
Okay, seems fine to me then.  @mmicko are you familiar with how the wheels are built?  Does this seem likely to break anything there or is it good to merge?

mmicko:
@donn could you please check if this is affecting  wheels build ?

donn:
@mmicko I'll test, but…

---

@akashlevy The wheels should already include both yosys-abc and share. How are you building them? https://github.com/YosysHQ/yosys/blob/38f858374c0ed83a6ac3ea66715263af015d7974/setup.py#L76

akashlevy:
I'm building with `make install` instead of building a wheel and loading that. I figure this should be a supported installation method as well

donn:
`make install ENABLE_PYOSYS=1` should install both `yosys-abc` and the Yosys `share` directory already is the thing. Wheels are kind of a special case. If `import libyosys` is not finding the globally installed yosys-abc and Yosys share, that's its own issue…

akashlevy:
Yes, I guess the real issue is that the search path does not get configured in the general case (which was what #4643 was attempting to fix I guess?)

For example, the workflow I use (which is pretty standard for Python development AFAIK) is to have a Python virtual environment called `.venv` and use `make install` to build the package into that. In this case, the current setup does not properly find `yosys-abc` and the `share` directory.

The best solution I could see is to just copy these resources right into the `pyosys` directory. That's guaranteed to work as that's the first place `libyosys` looks for them. I'm open to alternatives though

akashlevy:
If we think it makes most sense to use the wheel, then perhaps we should use a wheel-style installation in the Makefile when pyosys is enabled? Rather than copying stuff

akashlevy:
Here's what I have been using:
```
cd third_party/yosys && 
$(MAKE) install DESTDIR=../../.venv/ PREFIX= \
PYTHON_DESTDIR=`find ../../.venv/lib/python* -name "site-packages" | cut -c 13-` \
ENABLE_PYOSYS=1
```

Honestly, now that I look at this again, it does feel quite roundabout. I think it makes way more sense to just do `pip3 install .`

I propose we remove the legacy `make install` stuff that was manually copying `pyosys` into the Python site packages, and replace it with a wheel build when Pyosys is enabled. Does this seem reasonable?

KrystalDelusion:
> I propose we remove the legacy make install stuff that was manually copying pyosys into the Python site packages, and replace it with a wheel build when Pyosys is enabled. Does this seem reasonable?

That seems like the better approach here.

akashlevy:
@donn @mmicko @KrystalDelusion see now for approach using wheel

Makefile
```diff
@@ -957,6 +958,12 @@ unit-test: libyosys.so
 clean-unit-test:
 	@$(MAKE) -C $(UNITESTPATH) clean
 
+wheel: $(TARGETS)
+	python$(PYTHON_VERSION) -m pip wheel .
```
KrystalDelusion, line 962: What is the difference between `$(PYTHON_EXECUTABLE)` and `python$(PYTHON_VERSION)`?  I can see there is some older code that uses `python$(PYTHON_VERSION)`, but I think in general we should be using `$(PYTHON_EXECUTABLE)`.  Also because `$(PYTHON_VERSION)` is only declared if `ENABLE_PYOSYS` is, but the wheel and install-wheel targets don't have that limitation.

Makefile
```diff
@@ -976,8 +983,10 @@ ifeq ($(ENABLE_LIBYOSYS),1)
 	$(INSTALL_SUDO) cp libyosys.so $(DESTDIR)$(LIBDIR)/
 	$(INSTALL_SUDO) $(STRIP) -S $(DESTDIR)$(LIBDIR)/libyosys.so
 ifeq ($(ENABLE_PYOSYS),1)
-	$(INSTALL_SUDO) mkdir -p $(DESTDIR)$(PYTHON_DESTDIR)/$(subst -,_,$(PROGRAM_PREFIX))pyosys
-	$(INSTALL_SUDO) cp libyosys.so $(DESTDIR)$(PYTHON_DESTDIR)/$(subst -,_,$(PROGRAM_PREFIX))pyosys/libyosys.so
+	$(INSTALL_SUDO) @$(MAKE) install-wheel
+ifeq ($(ENABLE_ABC),1)
+	$(INSTALL_SUDO) cp yosys-abc $(DESTDIR)$(PYTHON_DESTDIR)/$(subst -,_,$(PROGRAM_PREFIX))pyosys/yosys-abc
```
KrystalDelusion, line 988: Isn't yosys-abc already included in the wheel?

akashlevy: Yep sorry I forgot to remove this

Makefile
akashlevy: Ok, I can change the `python$(PYTHON_VERSION)` to `$(PYTHON_EXECUTABLE)`

KrystalDelusion (changes requested):
Since the `wheel`/`install-wheel` targets are always available, it might be nice to automatically set `ENABLE_PYOSYS` if the user has explicitly called `make install-wheel`?  That or put those two targets inside a `ifeq ($(ENABLE_PYOSYS),1)` block.

akashlevy:
I don't know how to set a variable from within a target (is that even possible?)

Instead, I used the `ifeq ($(ENABLE_PYOSYS),1)` approach and throw an error if `ENABLE_PYOSYS` is not set with the `wheel` or `install_wheel` targets.

I also added a few more things to `clean` target that were missing for pyosys

KrystalDelusion:
> I don't know how to set a variable from within a target (is that even possible?)

I don't think you can set it within a target, and even if you did it would be too late to actually do what it needs to.  I think it is possible to check the target, but yeah the other way is much easier, and nice that you put the error on it too :)

akashlevy:
Thanks :)

PR #4902, opened 2025-02-14
`aldff`s do not get split by `splitcells` pass

akashlevy:

Currently, `aldff`s do not get split by splitcells; this PR fixes this.

Add `aldff` to the list of reg/latch cell types to be split.


akashlevy:
@povik @KrystalDelusion does this change look reasonable? Seems like a bug that `aldff` is missing

povik (approved):
Indeed an omission, the code does split on all appropriate ports already (AD, D, Q)

akashlevy:
Thanks @povik!

PR #4903, opened 2025-02-14
Fix macOS boost-python issues

akashlevy:

This PR resolves the issue below seen on my macOS Apple Silicon with a vanilla Yosys clone:
```
akashlevy@Akashs-MacBook-Pro yosys % make -j8 ENABLE_PYOSYS=1
$BREW_PREFIX is [/opt/homebrew/opt]
Makefile:343: *** BOOST_PYTHON_LIB could not be detected. Please define manually.  Stop.
```

There are two issues:
1. The `boost` include path is incorrect (`$(BREW_PREFIX)/boost/include/boost` instead of `$(BREW_PREFIX)/boost/include`). This means that the canonical form of Boost includes such as `#include <boost/...>` will not work
2. `boost-python3` library needs to be linked to detect `boost-python3` properly

You can work around these issues by manually setting the `BOOST_PYTHON_LIB`, but this PR makes it so that is not needed.

1. Use `$(BREW_PREFIX)/boost/include` as the Boost include path
2. Add `-L$(BREW_PREFIX)/boost-python3/lib` to the `LINKFLAGS` and include `LINKFLAGS` in `CHECK_BOOST_PYTHON`

This was tested on an Apple Silicon macOS (M2) with Pyosys enabled and was able to properly compile. I do not have access to an Intel macOS, it would be great if this can be tested against some builds as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -201,6 +201,8 @@ ifeq ($(ENABLE_PYTHON_CONFIG_EMBED),1)
 PYTHON_CONFIG := $(PYTHON_EXECUTABLE)-config --embed
 else
 PYTHON_CONFIG := $(PYTHON_EXECUTABLE)-config
+
+LINKFLAGS += $(shell $(PYTHON_CONFIG) --ldflags)
 endif
 
 PYTHON_DESTDIR := $(shell $(PYTHON_EXECUTABLE) -c "import site; print(site.getsitepackages()[-1]);")
@@ -331,7 +333,7 @@ endif
 
 ifeq ($(ENABLE_PYOSYS),1)
 # Detect name of boost_python library. Some distros use boost_python-py<version>, other boost_python<version>, some only use the major version number, some a concatenation of major and minor version numbers
-CHECK_BOOST_PYTHON = (echo "int main(int argc, char ** argv) {return 0;}" | $(CXX) -xc -o /dev/null $(shell $(PYTHON_CONFIG) --ldflags) $(LINKFLAGS) -l$(1) - > /dev/null 2>&1 && echo "-l$(1)")
+CHECK_BOOST_PYTHON = (echo "int main(int argc, char ** argv) {return 0;}" | $(CXX) -xc -o /dev/null $(LINKFLAGS) -l$(1) - > /dev/null 2>&1 && echo "-l$(1)")
 BOOST_PYTHON_LIB ?= $(shell \
 	$(call CHECK_BOOST_PYTHON,boost_python-py$(subst .,,$(PYTHON_VERSION))) || \
 	$(call CHECK_BOOST_PYTHON,boost_python-py$(PYTHON_MAJOR_VERSION)) || \
--- a/Makefile
+++ b/Makefile
@@ -201,8 +201,6 @@ ifeq ($(ENABLE_PYTHON_CONFIG_EMBED),1)
 PYTHON_CONFIG := $(PYTHON_EXECUTABLE)-config --embed
 else
 PYTHON_CONFIG := $(PYTHON_EXECUTABLE)-config
-
-LINKFLAGS += $(shell $(PYTHON_CONFIG) --ldflags)
 endif
 
 PYTHON_DESTDIR := $(shell $(PYTHON_EXECUTABLE) -c "import site; print(site.getsitepackages()[-1]);")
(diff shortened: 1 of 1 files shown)


KrystalDelusion:
Would you be willing to add a ci job for pyosys to the [`extra-builds.yml`](https://github.com/YosysHQ/yosys/blob/main/.github/workflows/extra-builds.yml)?  If you use a strategy matrix you can specify both `macos-latest` (which uses M1) and `macos-13` (which is Intel based); maybe throw in `ubuntu-latest` for well-roundedness.

akashlevy:
Certainly, I can try to do that when I get time

Makefile
```diff
@@ -331,7 +331,7 @@ endif
 
 ifeq ($(ENABLE_PYOSYS),1)
 # Detect name of boost_python library. Some distros use boost_python-py<version>, other boost_python<version>, some only use the major version number, some a concatenation of major and minor version numbers
-CHECK_BOOST_PYTHON = (echo "int main(int argc, char ** argv) {return 0;}" | $(CXX) -xc -o /dev/null $(shell $(PYTHON_CONFIG) --ldflags) -l$(1) - > /dev/null 2>&1 && echo "-l$(1)")
+CHECK_BOOST_PYTHON = (echo "int main(int argc, char ** argv) {return 0;}" | $(CXX) -xc -o /dev/null $(shell $(PYTHON_CONFIG) --ldflags) $(LINKFLAGS) -l$(1) - > /dev/null 2>&1 && echo "-l$(1)")
```
KrystalDelusion, line 334: I don't know if it matters, but I think this is sidestepping the whole point of `BOOST_PYTHON_LIB` and the `CHECK_BOOST_PYTHON` function, which is to find the boost library.  Maybe instead of putting `boost_python3/lib` in the `LINKFLAGS`, it would be better to add `-L$(BREW_PREFIX)/$(1)/lib` to the `CHECK_BOOST_PYTHON` function if `BREW_PREFIX` exists?   Then in 20 years when python4 releases this will still work :)

```Makefile
CHECK_BOOST_PYTHON = (echo "int main(int argc, char ** argv) {return 0;}" | $(CXX) -xc -o /dev/null $(shell $(PYTHON_CONFIG) --ldflags) -L/$(BREW_PREFIX)/$(1)/lib -l$(1) - > /dev/null 2>&1 && echo "-L/$(BREW_PREFIX)/$(1)/lib -l$(1)")
# ..
LIBS += $(shell $(PYTHON_CONFIG) --libs) $(filter-out -L%,$(BOOST_PYTHON_LIB)) -lboost_system -lboost_filesystem
# ..
LINKFLAGS += $(filter-out -l%,$(shell $(PYTHON_CONFIG) --ldflags) $(BOOST_PYTHON_LIB))
```

akashlevy: I don't see this as sidestepping the point of `CHECK_BOOST_PYTHON`. The `LINKFLAGS` variable only contains the uppercase `-L` flag (library directory), while `CHECK_BOOST_PYTHON` is designed to determine the lowercase `-l` flag (library name). Without adding `LINKFLAGS`, it doesn't know which directory to look for the library on macOS, and you'll always get the error: "BOOST_PYTHON_LIB could not be detected. Please define manually"

akashlevy: You can of course sidestep this problem by making sure that the `boost-python3` library directory is in the `LD_LIBRARY_PATH` (which happens automatically on Linux, but not for Homebrew), but I this solution allows this to happen automatically

akashlevy: > it would be better to add -L$(BREW_PREFIX)/$(1)/lib to the CHECK_BOOST_PYTHON function if BREW_PREFIX exists

This is basically what happens already, since the flag is only added when Homebrew is detected. I don't see how adding more `LINKFLAGS` can possibly hurt `CHECK_BOOST_PYTHON` being able to find the boost-python library name...

KrystalDelusion: > This is basically what happens already with this solution

`CHECK_BOOST_PYTHON` is used to check each of e.g.
```
boost_python-py3
boost_python-py311
boost_python3
boost_python311
```
Your solution says that they should be searching in `boost_python3/lib`, but only because you know that `boost_python3` is the correct name.  So either, brew should _always_ be using `-lboost_python3` from `-L$(BREW_PREFIX)/boost_python3/lib` and `CHECK_BOOST_PYTHON` is unnecessary under brew; or _both_ the `-l` and `-L` should be dynamic based on the python major/minor version, with the `-l` form being added to `LIBS` and the `-L` being added to `LINKFLAGS`.  Your solution is currently supporting the third option, which is that the library directory is always `-L$(BREW_PREFIX)/boost_python3/lib`, but the library name could be e.g. `boost_python311`; which could be possible but seems much less likely to me.

> I don't see how having more LINKFLAGS can possibly hurt CHECK_BOOST_PYTHON being able to find the boost-python library... and it certainly helps for macOS

I don't think there is anything wrong with adding `LINKFLAGS` to `CHECK_BOOST_PYTHON`, but then why not update `LIBS` and `LINKFLAGS` with `PYTHON_CONFIG` first, and then use `$(CXX) -xc -o /dev/null $(LINKFLAGS) $(LIBS) -l$(1) -`?

akashlevy: @KrystalDelusion In fact, the third option is what you see IRL on macOS:
```
akashlevy@Akashs-MacBook-Pro preqorsor % ls -l /opt/homebrew/opt/boost-python3/lib
total 14992
drwxr-xr-x  4 akashlevy  admin      128 Dec 11 16:00 cmake
-r--r--r--  1 akashlevy  admin   965568 Dec 11 16:00 libboost_numpy313.a
-r--r--r--  1 akashlevy  admin   115632 Jan 16 15:12 libboost_numpy313.dylib
-r--r--r--  1 akashlevy  admin  6197960 Dec 11 16:00 libboost_python313.a
-r--r--r--  1 akashlevy  admin   385440 Jan 16 15:12 libboost_python313.dylib
```

akashlevy: > why not update LIBS and LINKFLAGS with PYTHON_CONFIG first, and then use $(CXX) -xc -o /dev/null $(LINKFLAGS) $(LIBS) -l$(1) -?

I could do that, I was just trying to change as little of the existing script as possible :)

KrystalDelusion: Welp, I am happy to be proven wrong!  In that case yeah go ahead and put `LIBS` and `LINKFLAGS` first and I'll merge this :)

akashlevy: Done!

KrystalDelusion (approved):

akashlevy:
@KrystalDelusion The addition of ` $(LIBS)` into `CHECK_BOOST_PYTHON` actually causes it to fail somehow. I'm not sure why this happens, but I would just remove that.

PR #4904, opened 2025-02-14
share: add -pattern-limit to limit analysis effort

widlarizer:
The `share` pass uses SAT to find cells with same observability with regards to inputs of cells other than `$mux`. First, it expresses the observability conditions as basically a logical function over the S inputs of `$mux` cells in `find_cell_activation_patterns` by traversing contiguous `$mux` regions. This function is stored per cell as a sum of products in `activation_patterns_cache[cell]`. In some unfortunately shaped designs, the count of these products increases exponentially with the length of a chain of divering-converging muxes. This PR adds a user-configurable limit to how many "patterns" can be stored per cell. When this limit is exceeded, the cell is assumed to be always "active" in the sense that some of its input bits are always observable by outputs or cells other than multiplexers. This is always safe to do so as it can only inhibit merging cells. This limit is set to 1000 by default.

To test, you can observe the linear rise of memory usage beyond any reasonable bounds when running  `yosys -p "verific -sv slow-share.sv; prep; share"` on the following:

```sv
module test_case (
  input  wire         clk,
  input  wire         rst_clk,
  output wire [32:0]  rd [8:0],
  input  wire  [5:0]  wa [8:0],
  input  wire  [8:0]  wen
);

  logic [63:0][32:0] big;
  logic [63:0][32:0] wd;
  logic [5:0]        ra [8:0];

  always @(posedge clk) begin
      if (rst_clk) begin
        big <= '0;
      end
      else begin
        for (int i = 0; i < 9; i++) begin
          if (wen[i]) big[wa[i]] <= wd[wa[i]];
        end
      end
  end

  generate
    for (genvar i = 0; i < 9; i++) begin
      assign rd[i] = big[ra[i]];
    end
  endgenerate

endmodule
```

While this reproducer uses kind of a lot of wire bits, it is a completely reasonable design pattern. Pre-share statistics:

```
   Number of wires:                699
   Number of wire bits:          43561
   Number of public wires:          95
   Number of public wire bits:    2560
   Number of ports:                 21
   Number of port bits:            362
   Number of cells:                606
     $bmux                           9
     $dff                            1
     $mem_v2                         1
     $mux                          586
     $shl                            9
```

- [ ] add scratchpad support so that the limit can be adjusted even when `share` is used from `synth` commands

### DIFF
--- a/passes/opt/share.cc
+++ b/passes/opt/share.cc
@@ -856,18 +856,20 @@ struct ShareWorker
 	}
 
 	template<typename Iterator>
-	void insert_capped(pool<ssc_pair_t>& cache, Iterator begin_pattern, Iterator end_pattern)
+	bool insert_capped(pool<ssc_pair_t>& cache, Iterator begin_pattern, Iterator end_pattern)
 	{
 		if (cache.size() + std::distance(begin_pattern, end_pattern) > config.pattern_limit) {
 			cache.clear();
 			cache.insert(ssc_pair_t());
+			return false;
 		} else {
 			cache.insert(begin_pattern, end_pattern);
 		}
+		return true;
 	}
-	void insert_capped(pool<ssc_pair_t>& cache, ssc_pair_t pattern)
+	bool insert_capped(pool<ssc_pair_t>& cache, ssc_pair_t pattern)
 	{
-		insert_capped(cache, &pattern, &pattern + 1);
+		return insert_capped(cache, &pattern, &pattern + 1);
 	}
 
 	const pool<ssc_pair_t> &find_cell_activation_patterns(RTLIL::Cell *cell, const char *indent)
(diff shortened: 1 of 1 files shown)


mikesinouye:
Hi @widlarizer, are there any updates on this PR? My initial impression is that this approach will mitigate the issue for these types of designs, but 1000 is too high of a limit. Maybe n = 100 would work better?

However, I believe it's worth adding this fix now, and if there's additional improvement that can be done within the limit later that would be even better.

widlarizer:
@mikesinouye I'll try to push it forward. It's basically a hotfix, I'm not inclined to lowering it further since it might pessimize some designs. It's opened a can of worms on how to implement `share` in a more proper way though, see #4920 #4921

passes/opt/share.cc
```diff
@@ -853,6 +855,21 @@ struct ShareWorker
 		optimize_activation_patterns(patterns);
 	}
 
+	template<typename Iterator>
+	void insert_capped(pool<ssc_pair_t>& cache, Iterator begin_pattern, Iterator end_pattern)
+	{
+		if (cache.size() + std::distance(begin_pattern, end_pattern) > config.pattern_limit) {
+			cache.clear();
+			cache.insert(ssc_pair_t());
```
KrystalDelusion, line 863: Doesn't this lead to unnecessary work?  Once the cache contains the empty `ssc_pair_t()` then the result is always unshareable and we can skip the rest.  At the moment this could potentially result in the cache reaching the pattern limit many times over since each time it just resets the cache size to 1.  May be worth adding a boolean return for if the insertion succeeded, and then breaking out of loops if it returns false?  Testing locally with the toy example from the PR description that reduces runtime and memory usage by ~10%

widlarizer: I'm getting a better speedup than that:

```
Benchmark 1: ./uut/share-limit/bin/yosys -p "verific -sv garbage/slow-share.v; prep; share"
  Time (mean ± σ):      1.407 s ±  0.011 s    [User: 1.337 s, System: 0.064 s]
  Range (min … max):    1.393 s …  1.421 s    10 runs

Benchmark 2: ./uut/share-insert-return/bin/yosys -p "verific -sv garbage/slow-share.v; prep; share"
  Time (mean ± σ):     865.9 ms ±   3.8 ms    [User: 830.9 ms, System: 31.3 ms]
  Range (min … max):   859.5 ms … 873.8 ms    10 runs

Summary
  ./uut/share-insert-return/bin/yosys -p "verific -sv garbage/slow-share.v; prep; share" ran
    1.62 ± 0.01 times faster than ./uut/share-limit/bin/yosys -p "verific -sv garbage/slow-share.v; prep; share"
```

I'll have to check if it hasn't broken anything

KrystalDelusion (approved):

mikesinouye:
Thanks for driving this forward @widlarizer.

PR #4908, opened 2025-02-17
Cover all cases of submodule status in check-git-abc

### LINKED ISSUE #4888: abc not recognised as a submodule when using a different abc commit
### Version

0.48

Linux

### Reproduction Steps

1. Checkout Yosys
2. Call `git submodule update --init --remote` or `git -C abc checkout 8700bb` or any other command which changes the git commit of abc
3. Call `make check-git-abc`

### Expected Behavior

Should exit normally (status code 0)

### Actual Behavior

Exits with error (status code 1) and prints error message:
```
Error: 'abc' is not configured as a git submodule.
To resolve this:
1. Back up your changes: Save any modifications from the 'abc' directory to another location.
2. Remove the existing 'abc' directory: Delete the 'abc' directory and all its contents.
3. Initialize the submodule: Run 'git submodule update --init' to set up 'abc' as a submodule.
4. Reapply your changes: Move your saved changes back to the 'abc' directory, if necessary.
```

KrystalDelusion:

`git submodule status` can be prefixed with `[ -+U]`, of which we expect the space, and assume anything else is a `-`.  But seeing an error message that tells you abc is not initialized when it *is* initialized but pointing at a different commit is confusing.
Also closes #4888.

Handle the `[+U]` cases and give an actually related error message instead of the generic one.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -780,6 +780,10 @@ check-git-abc:
 		echo "Run 'git submodule update' to check out the correct version."; \
 		echo "Note: If testing a different version of abc, call `git commit abc` in the Yosys source directory to update the expected commit."; \
 		exit 1; \
+	elif git -C "$(YOSYS_SRC)" submodule status abc 2>/dev/null | grep -q '^U'; then \
+		echo "'abc' submodule has merge conflicts."; \
+		echo "Please resolve merge conflicts before continuing."; \
+		exit 1; \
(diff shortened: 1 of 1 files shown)


KrystalDelusion:
@widlarizer Is abc the only submodule that we care enough about being the exact version to raise an error when it's not?  Because the current suggestion to use `git submodule update` (with or without `--init`) will update all submodules, but we only check the status of abc.

widlarizer:
I think so, the only other submodule is libs/cxxopts and there's very little reason to bump libs/cxxopts and it's linked into yosys so the build error for when it's missing is fine

widlarizer (approved):

PR #4916, opened 2025-02-25
libs/fst: Fix stringop_overread warning under GCC

KrystalDelusion:

Fix the compiler warnings

    ‘ssize_t write(int, const void*, size_t)’ reading between 65536 and 4294967295 bytes from a region of size 16 [-Wstringop-overread]

Two of the three warnings were already fixed upstream, so we pull the fix and then patch the third.


povik (approved):
I've only reviewed the patch (being the bit that doesn't come from upstream), and it looks fine

Issue #4909, opened 2025-02-20
`splitcells` causes an assert failure


mattyoung101:
### Version

Yosys 0.50+7 (git sha1 9d3ab964b, ccache clang++ 19.1.7 -Og -fPIC)

### On which OS did this happen?

Linux

### Reproduction Steps

The following Verilog sample, `3510.v`, causes the `splitcells` command to fail with an assert failure.

```verilog
module top  (y, clk, wire2, wire1, wire0);
  output wire [(32'hb):(32'h0)] y;
  input wire [(1'h0):(1'h0)] clk;
  input wire signed [(2'h3):(1'h0)] wire2;
  input wire [(2'h3):(1'h0)] wire1;
  input wire [(2'h2):(1'h0)] wire0;
  wire [(2'h2):(1'h0)] wire4;
  reg signed [(2'h2):(1'h0)] reg7 = (1'h0);
  reg [(2'h3):(1'h0)] reg6 = (1'h0);
  reg [(2'h3):(1'h0)] reg5 = (1'h0);
  assign y = {wire4, reg7, reg6, reg5, (1'h0)};
  assign wire4 = (wire0[(2'h2):(1'h0)] ^~ $unsigned(wire1));
  always
    @(posedge clk) begin
      reg5 <= $signed(wire4);
    end
  always
    @(posedge clk) begin
      reg6 <= wire2;
    end
  always
    @(posedge clk) begin
      reg7 <= ({reg5} ? (~|reg5) : reg5);
    end
endmodule
```

Please let me know if the above sample is too long, and I will have a go at reducing it further.

Use the following script:
```
read_verilog 3510.v
prep
splitcells
```

This was discovered because I have been using the [Verismith](https://github.com/ymherklotz/verismith) fuzzer to find bugs in the Yosys pass I'm writing for my thesis, which incidentally discovered this bug, which I believe is an upstream issue.

### Expected Behavior

The `splitcells` command should not cause an assert failure.

### Actual Behavior

Yosys crashes with the following assert failure:

```
3. Executing SPLITCELLS pass (splitting up multi-bit cells).
Splitting $dff cell top/$procdff$14 into 2 slices:
  slice 1: $procdff$14[1:0] => \reg5 [1:0]
  slice 2: $procdff$14[2] => \reg5 [2]
Splitting $xnor cell top/$xnor$3510.v:13$1 into 2 slices:
  slice 1: $xnor$3510.v:13$1[1:0] => \wire4 [1:0]
[...]

PR #4922, opened 2025-02-28
splitcells: Fix the assertion bug caused by out-of-bound offset

Anhijkt:
#4909 
Before this change the offset was calculated from iterator variable and wasnt checked against GetSize(old_sig) bounds, which caused assertion error. This pr uses offset as an iteration variable which is bound checked in for loop.

### DIFF
--- a/passes/cmds/splitcells.cc
+++ b/passes/cmds/splitcells.cc
@@ -103,8 +103,7 @@ struct SplitcellsWorker
 
 				auto slice_signal = [&](SigSpec old_sig) -> SigSpec {
 					SigSpec new_sig;
-					for (int i = 0; i < GetSize(old_sig); i += GetSize(outsig)) {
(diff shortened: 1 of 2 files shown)


KrystalDelusion:
@mattyoung101 does this fix your issue?

mattyoung101:
Yep, I can confirm this fixes it. Thank you! We might want to amend this PR with a test case that covers the bugpoint reproducer in #4909 ?

KrystalDelusion (changes requested):
Please add a test case for this; adding a file such as `bug4909.ys` to the `/tests/various` directory should automatically get it picked up and running in CI.  Check out some of the existing files (such as `bug1496.ys`) for how you can include the RTLIL code directly in the Yosys script file.

KrystalDelusion (approved):

PR #4923, opened 2025-02-28
feat: Allow full constant wrapping for hilomap

KelvinChung2000:
There is no way to wrap the whole constant cell at once (based on my understanding). This is normally an unusual requirement for FPGA, but for a coarser grain device, bit level constant setting might not be available. 

By adding a new option to the `hilomap` which allows it to wrap a full constant by a cell. The original value will be stored into a parameter of the new cell, which is user-specified. 

Example module:
```verilog
module test();

  wire [31:0] in;
  wire [31:0] out;
  assign out = in + 16;

endmodule
```
Testing: `yosys -p "read_verilog test.v; hierarchy -auto-top; hilomap -wrap const O val; show"`

### DIFF
--- a/passes/techmap/hilomap.cc
+++ b/passes/techmap/hilomap.cc
@@ -26,29 +26,41 @@ PRIVATE_NAMESPACE_BEGIN
 
 static std::string hicell_celltype, hicell_portname;
 static std::string locell_celltype, locell_portname;
+static std::string fullcell_celltype, fullcell_portname, fullcell_paramname;
 static bool singleton_mode;
+static bool multi_bit;
 
 static RTLIL::Module *module;
 static RTLIL::SigBit last_hi, last_lo;
+static RTLIL::SigChunk value;
 
 void hilomap_worker(RTLIL::SigSpec &sig)
 {
-	for (auto &bit : sig) {
-		if (bit == RTLIL::State::S1 && !hicell_celltype.empty()) {
-			if (!singleton_mode || last_hi == RTLIL::State::Sm) {
-				last_hi = module->addWire(NEW_ID);
-				RTLIL::Cell *cell = module->addCell(NEW_ID, RTLIL::escape_id(hicell_celltype));
-				cell->setPort(RTLIL::escape_id(hicell_portname), last_hi);
+	if (multi_bit && sig.is_fully_const()){
+		value = module->addWire(NEW_ID, sig.size());
+		RTLIL::Cell *cell = module->addCell(NEW_ID, RTLIL::escape_id(fullcell_celltype));
+		cell->setParam(RTLIL::escape_id(fullcell_paramname), sig.as_const());
+		cell->setPort(RTLIL::escape_id(fullcell_portname), value);
+		sig = value;
+	}
+	else{
+		for (auto &bit : sig) {
+			if (bit == RTLIL::State::S1 && !hicell_celltype.empty()) {
+				if (!singleton_mode || last_hi == RTLIL::State::Sm) {
+					last_hi = module->addWire(NEW_ID);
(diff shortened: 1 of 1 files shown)


widlarizer:
Thanks for your contribution, this seems generally useful for possible other users with a similar use case and I don't see any other way to get constant mapping done with what we have already. Here is what I imagine this feature should look like to be merged:

- [ ] add a new pass named constmap that does what you want to do, as it seems distinct from the code and name of hilomap
- [ ] write a simple test

A pass named constmap will somewhat conflict with the CONSTMAP magic in techmap, but I think that's fine. Also, don't merge main into your feature branch, rebase the feature branch instead on top of main. You can reconstruct your feature branch like that by using a temporary branch rooted on main, `git cherry-pick` to move your commit onto there, deleting the original branch, recreating it on the temporary branch, and force pushing with `git push --force-with-lease`

KelvinChung2000:
Thanks for the feedback. I normally just use the sync fork button on GitHub. I will need to check how to set the default to rebase.  I will make the changes and get back to you.

KrystalDelusion:
> I don't know how to test whether the name of the output port is the expected one

```
select -assert-count 1 test/out1 %ci* r:value=16 %i
```

`test/out1 %ci*` is the input cone for out1, `%i` takes the intersection, so you're testing that the cell with `value=16` is in the input cone of `test/out1`.

tests/techmap/constmap.ys
```diff
@@ -0,0 +1,40 @@
...
+  wire [31:0] in;
+  wire [31:0] out1;
+  wire [31:0] out2;
+  assign out1 = in + 16;
+  assign out2 = in + 32;
+endmodule
+
+EOT
+
+constmap -cell const_cell O value
```
KrystalDelusion, line 35: For this case where there is a (blackbox) module provided, I wonder if it would make sense to test whether the module has the parameter being assigned.  Also because if you call `hierarchy -check` after running this you get an error:
```
ERROR: Module `const_cell' referenced in module `test' in cell `$auto$constmap.cc:36:constmap_worker$9' does not have a parameter named 'value'.
```
And most (all?) of the `synth_*` script passes run `hierarchy -check` before writing out the output so it would probably be expected for this pass to not fail there.

KrystalDelusion (changes requested):

tests/techmap/constmap.ys
KelvinChung2000: I think it is good that `hierarchy -check` fails should be expected since this tells the user they have used something that is not part of the primitive lib.

KrystalDelusion: That's not what `hierarchy -check` is for though.  `hierarchy -simcheck` has the added constraint of failing on blackbox instances which I think is the situation you are meaning.  If you read the `const_cell` module with `-noblackbox` instead of `-lib`, giving you a non-blackbox module without an implementation, that *should* fail `hierarchy -check`.  As should the first case where the `const_cell` module has no definition.  Adding the parameter (like you've now done) stops the error from appearing, but `const_cell` still isn't part of the primitive lib.

KelvinChung2000: So, should I add an assert at the pass to check if the cell is a known module and fail when the parameter is unavailable?

KrystalDelusion: Not an assert, it would be a `log_cmd_error` to indicate that the command is invalid for the current design/options/selection.  Or a `log_error` if you've already modified the design.  In a script the two are identical, but in an interactive prompt `log_cmd_error` aborts just the command instead of Yosys.

KrystalDelusion (approved):

PR #4926, opened 2025-03-03
Verilog file list suport

JasonBrave:

Most large Verilog/SystemVerilog IP use a file list/command '.f' file to hold list of Verilog file need to be compiled, this change make yosys able to directly parse file list.

Implement a `read_verilog_file_list` command, that read in file list, and calls `read_verilog` on each Verilog file in the file list. Currently the command only handles file path in file list file, in a future PR I will implement support for `+define+` and `+incdir+` in file list file.

### DIFF
--- a/frontends/verilog/verilog_frontend.cc
+++ b/frontends/verilog/verilog_frontend.cc
@@ -26,6 +26,8 @@
  *
  */
 
+#include <filesystem>
+
 #include "verilog_frontend.h"
 #include "preproc.h"
 #include "kernel/yosys.h"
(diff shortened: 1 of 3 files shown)


udif:
Please note that most, if not all tools, accept **any** command line argument in their file lists, not just filenames or +define/+incdir.
As far as I remember, the `slang` front end handles this by reading `-f` files recursively as part of the command line parser.

frontends/verilog/verilog_frontend.cc
```diff
@@ -672,6 +674,86 @@ struct VerilogDefines : public Pass {
...
+			}
+
+			bool is_sv = (std::filesystem::path(verilog_file_path).extension() == ".sv");
+
+			std::string command = "read_verilog";
+			if (is_sv) {
+				command += " -sv";
+			}
+			command = command + ' ' + verilog_file_path;
+			Pass::call(design, command);
```
KrystalDelusion, line 727: Why use the command version of `Pass::call` and not the args version?

KrystalDelusion: I would also highly recommend using the `-defer` argument to `read_verilog` like `read` does.

frontends/verilog/verilog_frontend.cc
```diff
@@ -672,6 +674,86 @@ struct VerilogDefines : public Pass {
...
+
+		std::string v_file_name;
+		while (std::getline(flist, v_file_name)) {
+			if (v_file_name.empty()) {
+				continue;
+			}
+
+			std::string verilog_file_path;
+			if (relative_to_file_list_path) {
+				verilog_file_path = file_list_parent_dir.string() + '/' + v_file_name;
```
KrystalDelusion, line 715: If you're using `<filesystem>` you should use the `/` operator for portability.  i.e. `verilog_file = file_list_parent_dir / v_file_name`.  It would also save on the converting it back into a path alter to check the extension, and you only convert to string when you pass it to `read_verilog`.

KrystalDelusion (commented):
If you're adding file list support for the verilog frontend, you should update the `read` pass in `/frontends/verific/verific.cc` to call it.

KrystalDelusion:
@udif The verific frontend for Yosys doesn't do this, so I don't think there's any reason that the read_verilog frontend should.

JasonBrave:
All comments addressed, please review @KrystalDelusion

frontends/verilog/verilog_frontend.cc
```diff
@@ -26,6 +26,8 @@
  *
  */
 
+#include <filesystem>
```
KrystalDelusion, line 29: This is causing the WASI build to fail, and from my understanding WASI doesn't _have_ a filesystem, or is agnostic of it; so I'm not sure if file list support makes sense on that platform and instead both the `#include` and the `parse_file_list`/`VerilogFileList` could be gated behind `#if !defined(__wasm)`.

frontends/verilog/verilog_frontend.cc
```diff
@@ -672,6 +674,84 @@ struct VerilogDefines : public Pass {
...
+		}
+		read_verilog_cmd.push_back(verilog_file_path.string());
+		Pass::call(design, read_verilog_cmd);
+	}
+
+	flist.close();
+}
+
+struct VerilogFileList : public Pass {
+	VerilogFileList() : Pass("read_verilog_file_list", "Parse a Verilog file list") {}
```
KrystalDelusion, line 714: > If you're adding file list support for the verilog frontend, you should update the read pass in /frontends/verific/verific.cc to call it.

frontends/verilog/verilog_frontend.cc
```diff
@@ -672,6 +676,86 @@ struct VerilogDefines : public Pass {
...
+	void help() override
+	{
+		//   |---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|
+		log("\n");
+		log("    read_verilog_file_list [options]\n");
+		log("\n");
+		log("Parse a Verilog file list, and pass the list of Verilog files to read_verilog command.\n");
+		log("\n");
+		log("    -F file_list_path\n");
+		log("        File list file contains list of Verilog files to be parsed, any path is treated relative to the file list file\n");
```
KrystalDelusion, line 728:
```suggestion
		log("        File list file contains list of Verilog files to be parsed, any path is\n");
		log("        treated relative to the file list file\n");
```

frontends/verilog/verilog_frontend.cc
```diff
@@ -672,6 +676,86 @@ struct VerilogDefines : public Pass {
...
+
+struct VerilogFileList : public Pass {
+	VerilogFileList() : Pass("read_verilog_file_list", "Parse a Verilog file list") {}
+	void help() override
+	{
+		//   |---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|
+		log("\n");
+		log("    read_verilog_file_list [options]\n");
+		log("\n");
+		log("Parse a Verilog file list, and pass the list of Verilog files to read_verilog command.\n");
```
KrystalDelusion, line 725:
```suggestion
		log("Parse a Verilog file list, and pass the list of Verilog files to read_verilog\n");
		log("command.\n");
```

frontends/verilog/verilog_frontend.cc
```diff
@@ -672,6 +676,86 @@ struct VerilogDefines : public Pass {
...
+		log("\n");
+		log("    -F file_list_path\n");
+		log("        File list file contains list of Verilog files to be parsed, any path is treated relative to the file list file\n");
+		log("\n");
+		log("    -f file_list_path\n");
+		log("        File list file contains list of Verilog files to be parsed, any path is treated relative to current working directroy\n");
```
KrystalDelusion, line 731:
```suggestion
		log("        File list file contains list of Verilog files to be parsed, any path is\n");
		log("        treated relative to current working directroy\n");
```

frontends/verilog/verilog_frontend.cc
KrystalDelusion: Help messages should break at 80 characters (which is why the `|---v---|` guides are there).

KrystalDelusion (changes requested):

KrystalDelusion (approved):

PR #4928, opened 2025-03-03
Add gatesi_mode to init gates under gates_mode in BLIF format

XutaxKamay:

Hello,

This is a proposal in order to give BLIF some extra information to initialize gates (specifically DFFs and Latches).

There's probably another way to get these values but since it was easier for me to parse the BLIF file this way in order to run my own simulation tool, I thought I might give a try to contribute to it.

I added `-gatesi` in order to indicate the initial values for the specific gates (mainly for DFFs and Latches outputs) in a non BLIF standard way.

With `write_blif` enabling both `-gates -gatesi` should be enough normally you should see something like this which indicates how to initialize the Q variable output:
```
.gate DFF 0 C=clk D=$abc$372906$auto$rtlil.cc:2985:MuxGate$372067 Q=\5.internal_registers[533]
.gate DFF 1 C=clk D=$abc$372906$auto$rtlil.cc:2985:MuxGate$372069 Q=\5.internal_registers[534]
.gate DFF 2 C=clk D=$abc$372906$auto$rtlil.cc:2985:MuxGate$372071 Q=\5.internal_registers[535]
.gate DFF 0 C=clk D=$abc$372906$auto$rtlil.cc:2985:MuxGate$372073 Q=\5.internal_registers[536]
```
instead of simply:
```
.gate DFF C=clk D=$abc$372906$auto$rtlil.cc:2985:MuxGate$372067 Q=\5.internal_registers[533]
.gate DFF C=clk D=$abc$372906$auto$rtlil.cc:2985:MuxGate$372069 Q=\5.internal_registers[534]
.gate DFF C=clk D=$abc$372906$auto$rtlil.cc:2985:MuxGate$372071 Q=\5.internal_registers[535]
.gate DFF C=clk D=$abc$372906$auto$rtlil.cc:2985:MuxGate$372073 Q=\5.internal_registers[536]
```

Of course, I might be able to understand why it wouldn't be appreciated to be accepted, but I would be grateful if possible since it would get into my package manager and not sitting in another fork, in the worst cases it doesn't modify the default behavior of yosys at all if not both are enabled.

Thank you.

### DIFF
--- a/backends/blif/blif.cc
+++ b/backends/blif/blif.cc
@@ -44,6 +44,7 @@ struct BlifDumperConfig
 	bool iattr_mode;
 	bool blackbox_mode;
 	bool noalias_mode;
+	bool gatesi_mode;
 
 	std::string buf_type, buf_in, buf_out;
 	std::map<RTLIL::IdString, std::pair<RTLIL::IdString, RTLIL::IdString>> unbuf_types;
@@ -51,7 +52,7 @@ struct BlifDumperConfig
 
 	BlifDumperConfig() : icells_mode(false), conn_mode(false), impltf_mode(false), gates_mode(false),
 			cname_mode(false), iname_mode(false), param_mode(false), attr_mode(false), iattr_mode(false),
-			blackbox_mode(false), noalias_mode(false) { }
+			blackbox_mode(false), noalias_mode(false), gatesi_mode(false) { }
 };
 
 struct BlifDumper
@@ -228,6 +229,8 @@ struct BlifDumper
 		{
 			if (cell->type == ID($scopeinfo))
 				continue;
+
+			auto gate_init = cell->hasPort(ID::Q) ? str_init(cell->getPort(ID::Q)) : std::string("");
 
 			if (config->unbuf_types.count(cell->type)) {
 				auto portnames = config->unbuf_types.at(cell->type);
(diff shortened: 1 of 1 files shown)


XutaxKamay:
> I don't think there's any problem with adding an extra mode like this. It's not the first non-standard feature. I think it's notable that this seems to be the first non-standard feature that can't then be parsed with `read_blif`. Without parsing support, it becomes difficult to test via roundtrip. With parsing support, we're affecting the parser, but that shouldn't cause trouble for parsing any .blif without this feature as I expect all parameter names to be reasonable and not start with a number, even if this is only a convention caused by users and developers being used to C and Verilog variable name restrictions. If you add support in read_blif and use it to write a test I'd be in favor of accepting this change


Alright, I'll write a read version aswell

XutaxKamay:
Alright, I think I've done it, can someone review it? (@widlarizer)

[test.blif.txt](https://github.com/user-attachments/files/24533042/test.blif.txt)
[test2.blif.txt](https://github.com/user-attachments/files/24533054/test2.blif.txt)

usage:
```yosys -p "read_blif test.blif; write_blif -gatesi test2.blif"```

Let me know if I need to change anything

widlarizer:
Man, I just wanted to force push an added test

XutaxKamay:
> Man, I just wanted to force push an added test

You want me to reopen another PR? Haha

whitequark:
You need to push some commits so that this branch isn't empty.

widlarizer:
That's the thing, I added a commit on top of the one here, pulled from XutaxKamay main specifically, and then force pushed to it since pushing and force pushing with lease wouldn't do anything. Of course, I don't force push without lease without reason, but like, I carefully checked I was moving around the commits I wanted

widlarizer:
When I was pushing my HEAD was absolutely not at the commit which was force pushed, I guess `git push XutaxKamay main --force` didn't care and pushed the old commit

widlarizer:
I guess that `git push XutaxKamay main` is not "push to branch main of remote XutaxKamay" but "push local branch main to branch main of remote XutaxKamay". But I'm not sure and I don't care enough to learn more. `git push XutaxKamay HEAD:main --force` did what I intended to do

whitequark:
> I guess that `git push XutaxKamay main` is not "push to branch main of remote XutaxKamay" but "push local branch main to branch main of remote XutaxKamay".

Yes.

XutaxKamay:
Ah sorry for the rebase

Issue #4927, opened 2025-03-03
Wrong error message mentioning defparam when overriding localparam


marzoul:
### Version

Yosys 0.50+56 (git sha1 9106d6b3b, g++ 14.2.1 -march=x86-64 -mtune=generic -O2 -fno-plt -fexceptions -fstack-clash-protection -fcf-protection -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -ffile-prefix-map=/home/prostboa/Dev/AUR/yosys-git/src=/usr/src/debug/yosys-git -flto=auto -fPIC -O3)

### On which OS did this happen?

Linux

### Reproduction Steps

Try to synthesize this (erroneous) design:

```verilog top.v

module comp #(
	localparam LP = 32
) (
	output wire [LP-1:0] odata
);

	assign odata = 0;

endmodule

module top #(
	parameter LP = 32
) (
	output wire [LP-1:0] oport
);

	wire [LP-1:0] owire;

	comp #(
		.LP(LP)
	 )
	comp_inst (
		.odata(owire)
	);

	assign oport = owire;

endmodule
```

Synthesize with this command:

```sh
yosys -l yosys.log -p "synth_xilinx -arch xc7 -top top ; write_json top.json" top.v
```

### Expected Behavior

The error message should indicate that overriding localparams is illegal, and indicate the name of the related parameter.
So the root cause of the error can be found even in a very large design.

### Actual Behavior

The error message mentions defparams, which is very misleading because defparam is a verilog keyword that is not used in the design.
The error message is extremely difficult to understand and does not indicate what parameter is related to the issue:

```
2.4.2. Executing AST frontend in derive mode using pre-parsed AST for module `\comp'.
Generating RTLIL representation for module `$paramod\comp'.
top.v:0: ERROR: Module name in defparam contains non-constant expressions!
```

KrystalDelusion:
This appears to be specific to `read_verilog -defer` (which is what gets called when you pass in top.v on the command line like that).  Running a regular read_verilog returns the expected:
```
ERROR: Module `comp' referenced in module `top' in cell `comp_inst' does not have a parameter named 'LP'.
```

marzoul:
I realize there was actually another issue in the reported error message.
`top.v:0: ERROR: Module name in defparam contains non-constant expressions!`
In cases where this error is appropriate of course, it would be extremely welcome to display a meaningful line number to so that the source of the issue can be found. But I don't know how to create such a reproducer.

[...]

PR #4930, opened 2025-03-06
Derive abstract cells in top module during hierarchy

KrystalDelusion:

Previously if hierarchy had a top module then it would not derive any other abstract cells, this means that `-check` could miss certain kinds of errors that it would normally detect, such as in #4927.

After finding (and potentially deriving) `top_mod`, find all of the cells included and derive them as well; similarly to how if there is no top module then *all* abstract modules are derived.  I'm just now realising this only works for one level of abstract cells, which means we probably need some kind of recursive approach instead so I'm leaving this as draft for now.

Includes `tests/various/hierarchy_reassign_localparam.ys` which fails under previous versions (instead giving the less helpful `ERROR: Module name in defparam contains non-constant expressions!`), but works when not using `read_verilog` without `-defer`, and/or when using `hierarchy -check` without `-top`.

### DIFF
--- a/passes/hierarchy/hierarchy.cc
+++ b/passes/hierarchy/hierarchy.cc
@@ -1066,6 +1066,17 @@ struct HierarchyPass : public Pass {
 					mod->attributes[ID::initial_top] = RTLIL::Const(1);
 				else
 					mod->attributes.erase(ID::initial_top);
+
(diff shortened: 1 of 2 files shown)


KrystalDelusion:
I seem to have found a can of worms.

PR #4938, opened 2025-03-14
Add .vscode to .gitignore

npe9:
This PR adds the .vscode directory to .gitignore to prevent it from being tracked in version control.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,3 +54,4 @@ __pycache__
 /venv
 /boost
 /ffi
+.vscode


KrystalDelusion:
Alternatively, add a .gitignore file to your .vscode directory, which just has a * and then git will ignore the directory without needing to change the top level file.

npe9:
Ahhh, clever, I will do that.

wjrforcyber:
> Alternatively, add a .gitignore file to your .vscode directory, which just has a * and then git will ignore the directory without needing to change the top level file.

I personal recommend  `.git/info/exclude` instead to manage all the editor/local project temp files/all local things etc in one place, or others who encounter with the similar issue will create another PR for this kind of ignore(which is unnecessary). [Reference](https://git-scm.com/docs/gitignore)

KrystalDelusion:
> I personally recommend .git/info/exclude instead to manage all the editor/local project temp files/all local things etc in one place

The reference you linked says only the editor specific files should be in the exclude, project specific files should be in the .gitignore, and that there are by design three files that control ignores because they mean different things. 

> or others who encounter with the similar issue will create another PR for this kind of ignore(which is unnecessary).

Unless I'm missing something the exclude file isn't tracked by the repository which means it will still be the user who needs to be aware of this feature of git in order to use it, and I can assure you there is nothing we can do to prevent people from opening a PR for it.

wjrforcyber:
> > I personally recommend .git/info/exclude instead to manage all the editor/local project temp files/all local things etc in one place
> 
> The reference you linked says only the editor specific files should be in the exclude, project specific files should be in the .gitignore, and that there are by design three files that control ignores because they mean different things. 
> 
> > or others who encounter with the similar issue will create another PR for this kind of ignore(which is unnecessary).
> 
> Unless I'm missing something the exclude file isn't tracked by the repository which means it will still be the user who needs to be aware of this feature of git in order to use it, and I can assure you there is nothing we can do to prevent people from opening a PR for it. 

Emmm, I'm just saying a way I do it, imagine there are different kinds of configuration folders of editor/local language server/etc(just local settings, not project related to simplify issue, I agree project generated should go to .gitignore), I just feel annoying keep adding gitingore in different configuration folder, and .vscode is a configuration folder, as you say git has different place for ignore, normally this kind of editor related folder should just be set in ignore of global configuration.(Since I like to control precisely I do repo by repo, so I give the 'exclude' option.)
As for PR, I thought people normally search for similar issue/PR before they make their own.

widlarizer:
Sounds like we're good to close this as unmerged? It'll still stay visible to people searching before making new PRs

PR #4939, opened 2025-03-14
mk: Fall back to cp if rsync is not available

DanielG:
Using rsync for such a simple job seems excessive. We're using this patch in Debian to avoid the build-dependency on `rsync`.

I'd actually just recommend switching to `cp -ru` (-u is "update") completely but I'm not sure what the rsync call is trying to accomplish exactly.

From `man cp`:
 
       -u, --update
              copy only when the SOURCE file is  newer  than  the  destination
              file or when the destination file is missing

--Daniel

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -114,6 +114,12 @@ BISON ?= bison
 STRIP ?= strip
 AWK ?= awk
 
+ifneq ($(shell :; command -v rsync),)
+RSYNC_CP ?= rsync -rc
+else
+RSYNC_CP ?= cp -ru
+endif
+
 ifeq ($(OS), Darwin)
 PLUGIN_LINKFLAGS += -undefined dynamic_lookup
 LINKFLAGS += -rdynamic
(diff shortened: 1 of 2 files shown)


KrystalDelusion:
The reason for rsync over cp was to allow checking for file update *without* relying on which is newer (off the top of my head I don't remember if it was hash or just file size).  Just checking which is newer results in always copying every cell/command help file even if they haven't changed, which in turn triggers sphinx to recheck and rebuild each file, and given the number of files that was triggering meant that rebuilding docs after changing one command took far longer than needed.

KrystalDelusion:
That being said, cells no longer generate .rst files directly and there is a(n almost complete) PR (#4860) which changes the cmd docs to the same, instead using autodoc and generated json files, making the .rst targets (and by extension the rsync use and this PR) no longer necessary

DanielG:
> Just checking which is newer results in always copying every cell/command help file even if they haven't changed, which in turn triggers sphinx to recheck and rebuild each file

I see.

I can't think of another tool to do that job other than a `find | xargs sh -c` oneliner ;-)

Making RSYNC_CP overridable and removing the cp fallback would also be perfectly fine for my purposes.

In the context of a distro buildsystem I'd rather have too many rebuilds than a invalidation issue since we almost never do unclean rebuilds anyway and when we do it's really, really painful if there's such a problem.

> making the .rst targets (and by extension the rsync use and this PR) no longer necessary

Either way works for me. If the rsync calls go away we'll see a conflict and fix it up so no worries.

LMK if you do still want the patch since we found some more rsync calls in docs/Makefile I'd still have to add in that case.

--Daniel

KrystalDelusion:
Hmm, the `rynsc` call in the `docs/source/_images/Makefile` was changed by me 7 months ago in a commit for macos safe builds that says
> Swap `cp -u` for `rsync -t`.

I don't _think_ that change was necessary for it to work on macos, and I'm not super sure why it is using the `-t` flag (which apparently preserves modification times)...

KrystalDelusion:
Ah.  Because `cp -u` doesn't exist on macos.  Cool and good.

PR #4940, opened 2025-03-14
mk: Avoid broken rebuild if program --help fails

DanielG:
Make will happily consider a target available if it's zero sized.

It's necessary to apply the `|| rm $@` pattern when writing output files with shell redirections to avoid problems with rebuilds here.

--Daniel

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1038,7 +1038,7 @@ docs/source/generated:
 
 # some commands return an error and print the usage text to stderr
 define DOC_USAGE_STDERR
-docs/source/generated/$(1): $(TARGETS) docs/source/generated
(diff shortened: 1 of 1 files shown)


KrystalDelusion:
The STDERR target(s) return an error code, which is why the line starts with -; does the || take that into account?  Given the failing CI I suspect not

DanielG:
Ah, I missed the rv=1 bit. I've added `FORCE` to always rerun instead.

IMO the proper fix would be for all commands to be normalized to not fail on --help and always print to stdout.

--Daniel

KrystalDelusion (approved):

Issue #4865, opened 2025-01-23
Yosys synth_ice40 -dsp fails on an assertion


zapta:
### Version

Yosys 0.49+1 (git sha1 3076803c9, aarch64-apple-darwin23.5-clang++ 18.1.8 -fPIC -O3)

### On which OS did this happen?

macOS

### Reproduction Steps

* Unzip the zip file below
* Source oss-cad-suite ``environment``
* Run build.sh

[yosys-ice40-dsp-error.zip](https://github.com/user-attachments/files/18524355/yosys-ice40-dsp-error.zip)

main.v
```
module main (

    input  clki,
    input  user_1,  // serial data in.
    output reg    user_2,  // serial data out

    // USB Pins (which should be statically driven if not being used).
    output usb_dp,
    output usb_dn,
    output usb_dp_pu
);

  // Prevent the host from trying to connect to this design.
  assign usb_dp = 1'b0;
  assign usb_dn = 1'b0;
  assign usb_dp_pu = 1'b0;

  // The size of each of the serial registers.
  // TODO: Why this fails with N = 40.
  //
  // localparam N = 38;
  localparam N = 40;
  // localparam N = 42;

  // The two words we multiply.
  reg [N-1:0] reg1;
  reg [N-1:0] reg2;

  always @(posedge clki) begin
    // Shift din into 2 x N words 
    reg1   <= {reg1[N-2:0], user_1};
    reg2   <= {reg1[N-2:0], reg1[N-1]};
    // Multiply the words, and output the parity of the result.
    user_2 <= ~^(reg1 * reg2);
  end

endmodule
```

Entire example:

### Expected Behavior

Yosys completes successfully and the designs uses the ICESTORM_DSP cells.

### Actual Behavior

Yosys fails on an assertion

```
⦗OSS CAD Suite⦘ ~/work$ ./build.sh 
+ yosys --version
Yosys 0.49+1 (git sha1 3076803c9, aarch64-apple-darwin23.5-clang++ 18.1.8 -fPIC -O3)
+ yosys -p 'synth_ice40 -top main -dsp -json _build/hardware.json' -q main.v
ERROR: Assert `nusers(O.extract_end(i)) <= 1' failed in ./passes/pmgen/ice40_dsp_pm.h:388.
⦗OSS CAD Suite⦘ ~/work$ 
```

widlarizer:
Reduced
```v
module main (
    input  clki,
    output reg user_2
);

[...]

PR #4942, opened 2025-03-15
ice40_dsp: fix log_assert issue

Anhijkt:
#4865 
This code iterates not from the beginning but from the end of output signal, leaving only bits that are currently in use with unused bits needed for offset.

### DIFF
--- a/tests/various/bug4865.ys
+++ b/tests/various/bug4865.ys
@@ -1,5 +1,6 @@
 read_rtlil << EOF
 
+
 autoidx 524
 
 attribute \top 1
@@ -10,8 +11,10 @@ module \main
   attribute \force_downto 1
   wire width 18 $verific$mult_4$garbage/usb.v:12$388.etc.blk.partial[0]
 
-  wire width 14 $delete_wire$514
+  wire width 12 $delete_wire$514
 
+  wire width 4 $test
+  
   attribute \module_not_derived 1
   cell \SB_MAC16 $verific$mult_4$garbage/usb.v:12$388.etc.sliceB[0].mul
     parameter \A_REG 1'0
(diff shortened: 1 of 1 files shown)


techlibs/ice40/ice40_dsp.pmg
```diff
@@ -46,17 +46,24 @@ code sigA sigB sigH
...
-	for (i = 0; i < GetSize(O); i++) {
-		if (nusers(O[i]) <= 1)
-			break;
-		sigH.append(O[i]);
-	}
+	for (i = GetSize(O) - 1; i > 0 && nusers(O[i]) <= 1; i--)
+		;
 	// This sigM could have no users if downstream sinks (e.g. $add) is
 	//   narrower than $mul result, for example
 	if (i == 0)
```
widlarizer, line 53: Is this bit still correct?

widlarizer: also this makes for a good opportunity to figure out what the comment referencing a non-existent `sigM` is actually supposed to say

techlibs/ice40/ice40_dsp.pmg
```diff
@@ -46,17 +46,24 @@ code sigA sigB sigH
...
 		reject;
 
-	log_assert(nusers(O.extract_end(i)) <= 1);
+	for (int j = 0, wire_width = 0; j <= i; j++) 
+		if (nusers(O[j]) == 0)
+			wire_width++;
+		else {
+			if (wire_width) { // add empty wires for bit offset if needed
+				sigH.append(module->addWire(NEW_ID, wire_width));
```
widlarizer, line 61: It seems odd to append to sigH a wire of `wire_width` width in each iteration. Is that the intended behavior?

tests/various/bug4865.ys
```diff
@@ -0,0 +1,67 @@
...
+    parameter \Y_WIDTH 2
+    connect \A $delete_wire$514
+    connect \B 2'x
+    connect \Y 2'x
+  end
+
+end
+EOF
+
+ice40_dsp
```
widlarizer, line 67: This doesn't trigger the assert on upstream main or your main, so this isn't a proper regression test

widlarizer (changes requested):

techlibs/ice40/ice40_dsp.pmg
Anhijkt: The reason for this is [this](https://github.com/YosysHQ/yosys/blob/59602740eea8662e30e245b267b59e47c158fa48/techlibs/ice40/ice40_dsp.cc#L191) line, as far as I understand, unused output ports on this type off cells should be connected with wires, so this code add them between wires of used ports.

PR #4946, opened 2025-03-18
cost: Add `$mem_v2`, `$macc_v2` estimates

povik:

Fill in blind spots of the keep_hierarchy command.

Add `$mem`, `$macc` estimation consistent with existing `$mul`, `$dff` formulas.

This script can exercise the new support:

```
design -reset
read_verilog <<EOF
module top(input clk, input [10:0] a, input [2:0] w, output [2:0] y);
	reg [3:0] m[60:0];

	always @(posedge clk) begin
		y <= m[a];
		m[a] <= w;
	end
endmodule
EOF
prep
keep_hierarchy -min_cost 20

design -reset
read_verilog <<EOF
module top(input [2:0] a, input [2:0] b, output [2:0] y);
	assign y = a * b;
endmodule
EOF
design -save r
keep_hierarchy -min_cost 20

design -load r
alumacc
opt_clean
keep_hierarchy -min_cost 20
```

### DIFF
--- a/kernel/consteval.h
+++ b/kernel/consteval.h
@@ -315,7 +315,7 @@ struct ConstEval
 			Macc macc;
 			macc.from_cell(cell);
 
-			for (auto &port : macc.ports) {
+			for (auto &port : macc.terms) {
 				if (!eval(port.in_a, undef, cell))
 					return false;
 				if (!eval(port.in_b, undef, cell))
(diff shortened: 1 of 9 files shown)


widlarizer (changes requested):
LGTM but the terminology change should be reflected in `docs/source/cell/word_arith.rst` as well

widlarizer (approved):
Actually the "port" concept that has been renamed to "term" is code-only and isn't referred to by the docs. The docs only refer to "port" as in cell port. The docs refer to the now-"term" as "summand" which is more explicit as to its role than "term" but longer. I would slightly prefer "term" being replaced with "summand" in the codebase for consistency but it's not a big deal

PR #4959, opened 2025-03-24
simplify: Skip AST_PRIMITIVE in AST_CELLARRAY

### LINKED ISSUE #4785: read_verilog: array of instances parsing assertion failure
### Version

Yosys 0.44 (git sha1 80ba43d26, clang++-18 18.1.8 -fPIC -O3)

Linux

### Reproduction Steps

```verilog
module test(in1, in2, out1);
  input in1, in2;
  output out1;

  nand  #2 t_nand[0:7](out1, in1, in2);
  nand  #2 x_nand[0:3](out1, in1, in2), y_nand[4:7](out1, in1, in2);
endmodule
```

### Expected Behavior

> IEEE Std 1364-2005:
>    7. Gate- and switch-level modeling
>      7.1 Gate and switch declaration syntax
>        7.1.5 The range specification
>          An array of instances shall have a continuous range. One instance identifier shall be
>          associated with only one range to declare an array of instances.
>          The range specification shall be optional. If no range specification is given, a single
>          instance shall be created.

The module body specified in reproduction steps was taken directly from the standard (page 78). It should be parsed just fine.

### Actual Behavior

```shell
 |  Copyright (C) 2012 - 2024  Claire Xenia Wolf <claire@yosyshq.com>         |
 Yosys 0.44 (git sha1 80ba43d26, clang++-18 18.1.8 -fPIC -O3)

yosys> read_verilog module.v 

1. Executing Verilog-2005 frontend: module.v
Parsing Verilog input from `module.v' to AST representation.
ERROR: Assert `new_cell->children.at(0)->type == AST_CELLTYPE' failed in frontends/ast/simplify.cc:2686.
```

### LINKED ISSUE #4875: async2sync fails on check cells with async reset
### Version

Yosys 0.49+3 (git sha1 9d4630484, g++ 14.2.1 -fPIC -O3)

Linux

### Reproduction Steps

The following SBY file declares an assertion that should only be triggered when `rst_n` is high within a block triggered asynchronously by `posedge clk` and `negedge rst_n`. This should be supported in `multiclock off` mode as `async2sync` should be able to handle a single async reset, converting it to a sync reset.

```
[options]
mode bmc

[engines]
smtbmc

[script]
read -sv test.sv
prep -top top

[file test.sv]
module top(input wire clk);
	always_ff @(posedge clk or negedge rst_n) begin
		if (rst_n) begin
			assert(1'b1);
		end
	end
endmodule
```

### Expected Behavior

SBY/Yosys accept the design and the proof passes. This happens fine with 9f27923782e0c9c8fb6298778929271eb06bcb02, a commit from just before `$check` cells were added. Although I haven't bisected to determine exactly when this issue was introduced, I assume its with `$check` cells as they are not able to be lowered correctly in this case.

### Actual Behavior

SBY reports an error from `async2sync`:
```
ERROR: $check cell _witness_.check_assert_test_sv_4_2 with TRG_WIDTH > 1 is not support by async2sync, use clk2fflogic.
```

This is because the trigger input is `{rst_n, clk}`. It would be good to be able to support this case otherwise embedding assertions within async reset blocks is broken.

KrystalDelusion:

Close #4785.

We have an input error about cell arrays of primitives not being supported, but the `AST_PRIMITIVE` is simplified before it can be checked, leading to an assertion error instead.

Don't simplify `AST_PRIMITIVE` in `AST_CELLARRAY`, similar to #3467.

Run `yosys tests/verilog/bug4785.ys` with/without this patch.

### DIFF
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -1758,7 +1758,7 @@ bool AstNode::simplify(bool const_fold, int stage, int width_hint, bool sign_hin
 			break;
 		if (type == AST_GENBLOCK)
 			break;
-		if (type == AST_CELLARRAY && children[i]->type == AST_CELL)
+		if (type == AST_CELLARRAY && (children[i]->type == AST_CELL || children[i]->type == AST_PRIMITIVE))
 			continue;
 		if (type == AST_BLOCK && !str.empty())
 			break;
@@ -2741,6 +2741,7 @@ bool AstNode::simplify(bool const_fold, int stage, int width_hint, bool sign_hin
 			if (new_cell->type == AST_PRIMITIVE) {
 				input_error("Cell arrays of primitives are currently not supported.\n");
 			} else {
+				this->dumpAst(NULL, "    ");
 				log_assert(new_cell->children.at(0)->type == AST_CELLTYPE);
 				new_cell->children.at(0)->str = stringf("$array:%d:%d:%s", i, num, new_cell->children.at(0)->str.c_str());
 			}
--- a/tests/verilog/bug4785.ys
+++ b/tests/verilog/bug4785.ys
@@ -1,0 +1,9 @@
+logger -expect error "Cell arrays of primitives are currently not supported" 1
+read_verilog <<EOT
+module test(in1, in2, out1);
+  input in1, in2;
+  output out1;
+
+  nand  #2 t_nand[0:7](out1, in1, in2);
+endmodule
+EOT


KrystalDelusion:
As pointed out in #4785, wrapping the primitive in a module, and then using an array of the wrapper module works fine.  So it may be possible to implement support for cell arrays of primitives, but I'm not sure how much work that is.  And there was already an error message for them being unsupported so 🤷 At least raising the proper error makes it clear we don't support it, rather than the mystery of an assertion error.

miquelt9:
@zachjs could you review this PR :pray:

zachjs (approved):

PR #4963, opened 2025-03-26
Trigger abc segfault for testing

widlarizer:
`abc` likes to crash irreproducibly. We'd like to gather core dumps or have more context for the crashes in other ways. This PR is triggers abc segfaults in CI


KrystalDelusion:
FWIW I don't think you need a PR, just a regular branch will run the same CI so long as it's on the YosysHQ/yosys repo

Issue #4968, opened 2025-03-28
Can't build from 0.51 tarball


spth:
### Version

0.51

### On which OS did this happen?

Linux

### Reproduction Steps

Same problem as https://github.com/YosysHQ/yosys/issues/4470, which was fixed in 0.47, but apparently is back in 0.51. Could you put some test in place for future releases, that check, that release source tarballs are actually buildable?

### Expected Behavior

yosys builds (like it does for me when using the 0.40, 0.42 and 0.47 tarballs).

### Actual Behavior

```
philipp@alecto:~/yosys-0.51$ make
[Makefile.conf] CONFIG := gcc
Initialize the submodule: Run 'git submodule update --init' to set up 'abc' as a submodule.
make: *** [Makefile:780: check-git-abc] Fehler 1
```

```
philipp@alecto:~/yosys-0.51$ LANG=C git submodule update --init --recursive
fatal: not a git repository (or any of the parent directories): .git
```

KrystalDelusion:
I just tried and it seems to be working fine?
```
dawn@holly:~/Downloads$ tar -xzf yosys.tar.gz -C yosys-0.51
dawn@holly:~/Downloads$ cd yosys-0.51
dawn@holly:~/Downloads/yosys-0.51$ make check-git-abc
'abc' comes from a tarball. Continuing.
```

It's the `check-git-abc` make target that gives the `"Initialize the submodule: Run 'git submodule update --init' to set up 'abc' as a submodule."` error, which should only happen if:
1. the `abc` directory exists,
2. `git submodule status` fails, and
3. there is no `.gitcommit` file in the `abc` directory.

If there *is* a `abc/.gitcommit`, and it contains `$Format:%H$`, then ABC comes from git, but `git submodule status` has failed so it can't be a submodule and it displays `"Error: 'abc' is not configured as a git submodule."`.

If `abc/.gitcommit` contains anything else, say `f2d68d590fa6f8fc32295a2edd79afc0d14a1414` like it does in the 0.51 release .tar.gz, then it gives the message `"'abc' comes from a tarball. Continuing."`, which is what it displays in the test I just did.

Are hidden files somehow being ignored when you extract the archive?

KrystalDelusion:
Oh, unless you downloaded the "Source code (tar.gz)" instead of the yosys.tar.gz file, which isn't a release tarball, it's just the source code.

spth:
[...]

PR #4969, opened 2025-03-28
Makefile: Test yosys git status in check-git-abc

KrystalDelusion:

As in #4968, `check-git-abc` is misleading if Yosys itself isn't a git repository. 

So check `git status` before suggesting `git` based solutions, providing alternative suggestions for using ABCEXTERNAL (which bypasses `check-git-abc`), or downloading release tar (noting that the 'Source code' archives won't work, which is probably how they ended up in this situtation).

Applying patch to release yosys.tar.gz and running `make check-git-abc` should be be unaffected (i.e. should still succeed).
Applying patch to release source code archive and running `make check-git-abc` (or removing contents of abc folder from yosys.tar.gz) should give new error message.
Applying patch to either archive, and deleting abc folder should give new error message.
All other cases should be unaffected.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -785,7 +785,7 @@ $(PROGRAM_PREFIX)yosys-config: misc/yosys-config.in $(YOSYS_SRC)/Makefile
 .PHONY: check-git-abc
 
 check-git-abc:
-	@if [ ! -d "$(YOSYS_SRC)/abc" ] && git -C "$(YOSYS_SRC)" status 2>/dev/null; then \
+	@if [ ! -d "$(YOSYS_SRC)/abc" ] && git -C "$(YOSYS_SRC)" status >/dev/null 2>&1; then \
 		echo "Error: The 'abc' directory does not exist."; \
 		echo "Initialize the submodule: Run 'git submodule update --init' to set up 'abc' as a submodule."; \
 		exit 1; \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -785,7 +785,7 @@ $(PROGRAM_PREFIX)yosys-config: misc/yosys-config.in $(YOSYS_SRC)/Makefile
 .PHONY: check-git-abc
 
 check-git-abc:
-	@if [ ! -d "$(YOSYS_SRC)/abc" ]; then \
+	@if [ ! -d "$(YOSYS_SRC)/abc" ] && git -C "$(YOSYS_SRC)" status 2>/dev/null; then \
```
mmicko, line 788: `status 2>&1 >/dev/null ` since we do not wish output to appear

Makefile
```diff
@@ -811,6 +811,12 @@ check-git-abc:
 		echo "3. Initialize the submodule: Run 'git submodule update --init' to set up 'abc' as a submodule."; \
 		echo "4. Reapply your changes: Move your saved changes back to the 'abc' directory, if necessary."; \
 		exit 1; \
+	elif ! git -C "$(YOSYS_SRC)" status 2>/dev/null; then \
```
mmicko, line 814: `status 2>&1 >/dev/null ` since we do not wish output to appear, same as above

mmicko (changes requested):

PR #4970, opened 2025-03-29
Gowin. BUGFIX. Restore MUXes.

yrabbit:
I mistakenly listed MUXes as unplanned (https://github.com/YosysHQ/yosys/pull/4966). This did not lead to catastrophic consequences - all apicula and uLinux examples compile and work, but the quality of synthesis was significantly reduced, which affected the compilation time badly.

Restoring the MUXes will bring the compilation time back to normal.

times

### DIFF
--- a/techlibs/gowin/cells_xtra.py
+++ b/techlibs/gowin/cells_xtra.py
@@ -26,8 +26,7 @@ _skip = { # These are already described, no need to extract them from the vendor
           'RAM16SDP1', 'RAM16SDP2', 'RAM16SDP4', 'rPLL', 'SDP',
           'SDPX9', 'SP', 'SPX9', 'TBUF', 'TLVDS_OBUF', 'VCC', 'DCS', 'EMCU',
           # These are not planned for implementation
-          'MUX2_MUX8', 'MUX2_MUX16', 'MUX2_MUX32', 'MUX4', 'MUX8', 'MUX16',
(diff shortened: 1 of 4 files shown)


gatecat:
Are you sure there isn't something else going on there? Just having  blackbox cells for wide muxes won't make Yosys use them.

yrabbit:
I'm reasonably sure: the picture shows running compilation of apicula examples with different versions of nextpnr and apicula, i.e. apicula is the same, nextpnr is the same. Only the yosys versions are different.

I realize from 0.15 to main there are many changes, but I tried this PR and the time is back to normal.



shot-3

gatecat:
Do you actually see these specific muxes in the Yosys output with/without this PR?

yrabbit:
No, I don't see them. So far I've managed to narrow it down to two suspects - if there is a description as a black box of either MUX4 and MUX8, no problems are observed. If both of them are missing, then compile time (actually the time of network routing in nextpnr) grows by an order of magnitude.

I'll keep looking, it's an interesting mystery. The good thing is that it is cured by very simply adding MUX4 to /usr/local/share/yosys/gowin/.... so I'll be able to advise users in case of anything:)

yrabbit:
Yes, I think the changes to yosys were a weird  coincidence, not the cause. So let's close this PR.

PR #4973, opened 2025-03-31
verific: set VHDL entity source location from another attribute if available

mmicko:
Source location coming from Verific is now pointing to architecture and not entity. Additional patch now adds new hidden attribute (starting with empty place) that adds source location of entity. With this change default behavior is not affected and it is possible to use Verific without our patch, but in case information exists it will be included in yosys netlist. In this case also architecture source location will be stored in "src_architecture" attribute.


mmicko:
Too complicated, using other solution

PR #4976, opened 2025-03-31
Support array ranges for identifiers in the Liberty parser.

Logikable:

Some pins may be in a list, and we need to refer to one specifically.

We reuse the range parsing code from a different part of the parser.

Try parsing a liberty file like:
```
pin("foo") {
  bar : baz[0] ;
}
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -889,6 +889,7 @@ SH_TEST_DIRS += tests/rpc
 SH_TEST_DIRS += tests/memfile
 SH_TEST_DIRS += tests/fmt
 SH_TEST_DIRS += tests/cxxrtl
+SH_TEST_DIRS += tests/liberty
 ifeq ($(ENABLE_FUNCTIONAL_TESTS),1)
 SH_TEST_DIRS += tests/functional
 endif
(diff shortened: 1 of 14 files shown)


mikesinouye:
Hi @widlarizer, this PR enables us to support a private PDK. Please review when you get a chance, thanks!

KrystalDelusion:
Are you able to add a small test to this for CI?

Logikable:
I tried adding a test to `tests/liberty` (not yet pushed to my branch). Are we sure these tests are being run as part of CI? When I run them locally, I get diffs against the goldens (e.g. `cd tests/liberty && ./run-test.sh`), so they should be failing.

georgerennie:
Yes it seems these have not been included in the makefile (by appending to `SH_TEST_DIRS` so aren't being run). Adding them I also see diffs but no error code being generated.

KrystalDelusion:
Yeah looks like the Makefile is missing a `SH_TEST_DIRS += tests/liberty` if you could include that in this PR too :)

widlarizer:
Weird, they haven't been listed in the Makefile prior to #4814 either

Logikable:
Added the existing `tests/liberty` tests to the Makefile, and fixed them (they were incorrect).
I also added a test for this PR.

KrystalDelusion (approved):

mikesinouye:
Hi @widlarizer, does everything look good on your end to merge this?

PR #4981, opened 2025-04-01
Improve lexer performance for read_liberty

jix:
This PR makes liberty parsing roughly twice as fast by

1. using a custom ifstream wrapper that implements buffering and the used methods in a way that is much more amenable to compiler optimizations than ifstream, and
2. extending the ifstream wrapper with arbitrary lookahead to avoid growing std::string instances byte by byte.

Here 1. is the biggest improvement, reducing runtime by around 40% and 2. adds saves another 10%.

### DIFF
--- a/kernel/yosys_common.h
+++ b/kernel/yosys_common.h
@@ -128,6 +128,12 @@
 #  error "C++17 or later compatible compiler is required"
 #endif
 
+#if defined(__has_cpp_attribute) && __has_cpp_attribute(gnu::cold)
+#  define YS_COLD [[gnu::cold]]
+#else
+#  define YS_COLD
+#endif
+
 #include "kernel/io.h"
 
 YOSYS_NAMESPACE_BEGIN
--- a/passes/techmap/libparse.cc
+++ b/passes/techmap/libparse.cc
@@ -77,6 +77,16 @@ int LibertyInputStream::get_cold()
 	return c;
 }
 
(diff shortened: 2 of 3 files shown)


mikesinouye:
This is very exciting, looking forward to any per-invocation caching opportunities as well.

widlarizer:
On reading gzipped liberty files with `read_liberty -ignore_miss_func [...]/OpenROAD-flow-scripts/flow/platforms/asap7/lib/CCS/*`, this keeps memory consumption on par and still speeds up by 45%

widlarizer:
As a sanity check, reading a 1GB empty liberty file like
```py
def print_library(n):
    print("library (x) {")
    print("\n" * n, end="")
    print("}")

n = 1_000_000_000
print_library(n)
```
Gives the following results (BL = baseline)
```
uncompressed
PR: CPU: user 1.74s system 0.11s, MEM: 24.45 MB peak
BL: CPU: user 4.81s system 0.11s, MEM: 24.58 MB peak
compressed
PR: CPU: user 3.08s system 0.00s, MEM: 24.45 MB peak
BL: CPU: user 6.25s system 0.00s, MEM: 24.47 MB peak
```

widlarizer (approved):
At first I was uncertain about the ranges used for appending, but it's safe due to how peek guarantees the buffer to be fine from one character before until the peeked character. The unget usage looks safe as well. There is no need to add tests as it's covered by other tests

PR #4986, opened 2025-04-03
Liberty file caching with new `libcache` command

jix:
This adds optional in-memory caching of parsed liberty files to speed up flows that repeatedly parse the same liberty files. To avoid increasing the memory overhead by default, the caching is disabled by default. The caching can be controlled globally or on a per path basis using the new `libcache` command, which also allows purging cached data.

The command to enable caching unconditionally is `libcache -enable -all`.

### DIFF
--- a/frontends/liberty/liberty.cc
+++ b/frontends/liberty/liberty.cc
@@ -539,7 +539,7 @@ struct LibertyFrontend : public Frontend {
 
 		log_header(design, "Executing Liberty frontend: %s\n", filename.c_str());
 
-		LibertyParser parser(*f);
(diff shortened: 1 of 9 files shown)


widlarizer:
FYI `tests/liberty` passes aren't getting run at the moment, see #4976

mikesinouye:
We are very excited to try this out. I observed 3x speedup in our combined `read_liberty` performance and 1.4x overall flow speedup for a small subblock from #4987, which is a great benefit to our iterative RTL development use case. Excited to see what additional benefit this can bring.

Issue #2648, opened 2021-03-09
ice40_dsp, extend_u0 and unextend incorrectly handle sign bits


sylefeb:
## Steps to reproduce the issue

On yosys 26e01a67db4135196b1d25ed89e9e6ceb536f4e3.

Using the three files in the gist https://gist.github.com/sylefeb/447c22fa39b25e7056ba4e4ecae92d7b
one can see that the results differ whether using the `-dsp` option or not. Without `-dsp` the result is:
`result = 117 (should be 117)`
With -dsp the result is:
`result = 5 (should be 117)`

## Expected behavior

The result with and without `-dsp` should match, the result is correct in this case without `-dsp`.

## Actual behavior

The result differs (5 vs 117).

## Analysis

It appears the problem stems from the combined behavior of `unextend` and `extend_u0` in `ice40_dsp.pgm` and `ice40_dsp.cc`. My understanding is that: 
- `unextend` removes leading bits from signals if they are the same (wire or const), including the bit that may represent a change of sign through e.g. a two's complement on an unsigned wire (and including the sign bit of a const signal, regardless of signedness).
- `extend_u0` assumes leading bits are always zeros when padding back an unsigned signal

## Suggested fix

I am not sure enough to propose this as a PR, but this works on my side in all tests so far:
- always preserve what represents a sign bit (including on unsigned signals) in `unextend`
- always give `extend_u0` a 'signed' signal (second parameter always true), such that the MSB is replicated during extension

This may be too conservative however.

Patch: https://github.com/sylefeb/fpga-binutils/blob/master/patches/yosys_patch_ice40_dsp.diff

## Remarks

- A similar issue is also fixed by the patch, which affects signed const signals loosing their sign bit in `unextend`.
- Other platforms (xilinx dsp) may be affected by a similar issue as they also use `unextend`.

emeb:
I also had an issue with connection of signed values to the SB_MAC16 cell and this patch appears to have corrected it.

bit-hack:
I think I have also encountered this issue myself.  I have observed incorrect and differing behavior when using the `-dsp` option to without it.  I compiled this code:
```v
module mdac12x8 (
    input                clk,
    input  signed [11:0] iVoice,
    input         [ 7:0] iEnv,
    output signed [15:0] oOut
);
  wire signed [15:0] lhs = { iVoice, 4'h0 };
  wire signed [15:0] rhs = { 8'h0, iEnv };

  reg signed [31:0] product;
[...]

PR #4992, opened 2025-04-05
ice40_dsp: fix const handling

Anhijkt:
ice40_dsp seems to incorrectly handle unsigned consts. The issue seems to be with unextend function. By searching previous issues I could find this one #2648.
Now unextend is called only on non constant inputs.
```
read_verilog << EOT
module top(input wire [14:0] a, output wire [31:0] b);
assign b = a*$unsigned(5'b01111);
endmodule
EOT

prep
ice40_dsp
write_rtlil
```

```
read_verilog << EOT
module top(input wire signed [14:0] a, output wire signed [31:0] b);
assign b = a*$signed(5'b01111);
endmodule
EOT

prep
ice40_dsp
write_rtlil
```

Gives two different connect \B values in \SB_MAC16 cell on pre-PR code, and the same value on post-PR code.

### DIFF
--- a/techlibs/ice40/ice40_dsp.pmg
+++ b/techlibs/ice40/ice40_dsp.pmg
@@ -23,7 +23,7 @@ match mul
 endmatch
 
 code sigA sigB sigH
-	auto unextend = [](const SigSpec &sig) {
+	auto unextend_signed = [](const SigSpec &sig) {
 		int i;
 		for (i = GetSize(sig)-1; i > 0; i--)
 			if (sig[i] != sig[i-1])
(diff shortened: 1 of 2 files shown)


widlarizer:
Please add a regression test. If you have the time, can you check `techlibs/xilinx/xilinx_dsp.pmg` too? It seems like either is the copy of the other, and they differ exactly on const handling. Maybe what xilinx_dsp.pmg does is somehow a more correct fix, or the other way around. Thanks for your contributions!

Anhijkt:
I tried to copy xilinx code, but it(and microchip_dsp.pmg) seems to fail on the same issue.  Also I mistaken this bug for const-only, turns out this is an issue with handling all unsigned signals (for example `a*$unsigned({7'b1010101, b})`), so I think there is other way to fix it.

Anhijkt:
Also I want to note that in first example, the issue is not just with `ice40_dsp`, theoretically unextended can handle `$unsigned(5'b01111)`  the problem is with `wreduce` in `prep`, that reducing it into `$unsigned(4'b1111)`.

Anhijkt:
The previous commit did not optimize unsigned zero signals unlike the original code. But it also differs from original when handling signals like `$unsigned({a[n], a[n], a})` where n - is the index of last bit of signal a. If original `unextend` would optimize it into  `$unsigned({a})` this code leave it unchanged. I dont know which implementation is more accurate, so I think I need feedback on this one.

widlarizer (approved):
I'm pretty sure the old behavior wasn't right. You can't unextend an unsigned `110` into a `10`, for example. The current version looks good. There are other DSP PMG passes where an unextend function is cobbled together but let's consider that out of scope for this PR and moved into #5059

Issue #4590, opened 2024-09-09
Synthesis with synth_xilinx crashes in TECHMAP pass


marzoul:
### Version

Yosys 0.45+106

### On which OS did this happen?

Linux

### Reproduction Steps

See attached archive. It contains the input verilog file, the Makefile, and the log of last Yosys run.
Just type `make` to reproduce.

The design has been reduced to as much as possible to reproduce the crash reliably.
The original design was VHDL, converted to Verilog with the tool ghdl, this is why the verilog is not easily readable.
Overall the datapath that remains is a series of FIFOs and a small read-only memory in parallel.
The issue is not reproduced with less memories hence the design is not that small. 

Complete version of Yosys as reported by `yosys --version` :
`Yosys 0.45+106 (git sha1 982fade0d, g++ 14.2.1 -march=x86-64 -mtune=generic -O2 -fno-plt -fexceptions -fstack-clash-protection -fcf-protection -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -ffile-prefix-map=/home/prostboa/Dev/AUR/yosys-git/src=/usr/src/debug/yosys-git -flto=auto -fPIC -O3)`

Yosys is freshly compiled from sources, from Archlinux AUR package yosys-git.

### Expected Behavior

Yosys is expected to finish the synthesis and emit the resulting json file without error.

### Actual Behavior

Yosys emits an error in TECHMAP pass and exits with error code zero (it should also be non-zero in such error case).

marzoul:
For whatever reason I could not attach the small archive (77kB).
Here is a public link : https://cloud.univ-grenoble-alpes.fr/s/YFwoXCHti7sTasM

georgerennie:
Minimized reproducer:
```
read_ilang <<EOT
autoidx 1
module \nnawaq_axi3
  wire output 1 \out
  cell $_MUX16_ \mux
    connect \A 1'x
    connect \B 1'x
    connect \C 1'x
    connect \D 1'x
    connect \E 1'x
    connect \F 1'x
    connect \G 1'x
    connect \H 1'x
    connect \I 1'x
    connect \J 1'x
    connect \K 1'x
    connect \L 1'x
    connect \M 1'x
    connect \N 1'x
    connect \O 1'x
    connect \P 1'x
    connect \S 1'x
    connect \T 1'x
    connect \U 1'x
    connect \V 1'x
    connect \Y \out
  end
end
EOT

[...]

PR #4994, opened 2025-04-06
Fix mux mapping for xilinx techno when all inputs are x

marzoul:
Fixes https://github.com/YosysHQ/yosys/issues/4590 where a bit-banged mux has all inputs at `x`.
This case can happen when the selection input signal is originally larger than the data source.
Mapping of underlying smaller muxes end up having to handle upper mux with all inputs at `x` which was not handled.
The root cause was identified by @KrystalDelusion.

To fix the crash, handling of the case where all inputs are `x` is explicitly handled, in that case the output is simply assigned to `x`.

Tested on the reproducer poster on https://github.com/YosysHQ/yosys/issues/4590


KrystalDelusion:
Thanks!

PR #4995, opened 2025-04-07
Add `NamedObject` to `py_wrap_generator.py`

akashlevy:

It seems that `pyosys` breaks with latest changes adding `NamedObject` class. Attempting to get the name of a module in `pyosys` no longer works.

Register `NamedObject` in `py_wrap_generator.py`

Load a simple design in `pyosys` and try to get the name of the top module before and after the change.


akashlevy:
Yes, this would be good to have eventually. It's pretty easy to break

PR #4997, opened 2025-04-07
techmap: map $alu to $fa instead of relying on extract_fa

widlarizer:
`extract_fa` is used in ASIC flows to find full and half adder cells by extracting any matching logic. It has been suggested that a better approach is to use full/half adders only when emitted from actual arithmetic. In most synthesis flows, cells that use an adder-like structure (addition, subtraction, comparison) get converted to `$alu`. For ASIC flows, they are then techmapped to a multi-bit `$lcu` in `techlibs/common/techmap.v`. `$lcu` cells need propagate (P) and generate (G) inputs which have to be generated by half adders.

Prior to this PR, `techmap.v` was generating these with explicit logic and relying on `extract_fa` to discover that it's half adders. As of this PR, half adders are instantiated directly as `$fa` cells with C inputs at constant 0. This is how `extract_fa` implements half adders already, and for example ORFS uses this constant C=0 to [map directly to half adders](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/master/flow/platforms/sky130hd/cells_adders_hd.v) in the PDK. 

This PR is best evaluated in the context of an ASIC flow as-is or with `extract_fa` removed

- [ ] add an equivalence test


widlarizer:
I originally wanted an equivalence test but in reality techmap.v is covered by a lot of equiv_opt tests

PR #5001, opened 2025-04-08
Update to latest ABC

mmicko:
Updates to latest ABC, also makes it less intrusive on ABC side for mingw support


widlarizer:
Thanks!

PR #5003, opened 2025-04-08
dump: add --sorted flag

widlarizer:
This is helpful for diffs when roundtripping. The sort is somewhat imperfect as sorting the connections uses SigSpec `operator<` which doesn't result in a nice ordering. However, this comes into play only if the connections aren't equivalent, like here: 
image
Therefore for getting a nice diff to check total complete equivalence excluding ordering even though `opt_clean` reordered your design (#3426) this is is still useful


widlarizer:
Let's do #5421 instead

PR #5004, opened 2025-04-08
opt_clean: handle undriven and x-bit driven bits consistently

widlarizer:
I believe driving the bit of a signal with a constant x-bit is exactly the same as not driving it at all. This PR removes this inconsistency by extending `SigPool`. As the `SigPool` is used to check the `assign_map`ped wire bits against unmapped bits, this creates an arbitrary distinction between driven and undriven. I have added a pass-specific mechanism by inheriting from `SigPool` that happens to work for this case. I don't know if `SigPool` is similarly misused elsewhere.

This PR shares its motivation with #5003

### DIFF
--- a/passes/opt/opt_clean.cc
+++ b/passes/opt/opt_clean.cc
@@ -31,6 +31,24 @@ PRIVATE_NAMESPACE_BEGIN
 
 using RTLIL::id2cstr;
 
+struct CleanerPool : SigPool
+{
+	bool check_all_def(const RTLIL::SigSpec &sig) const
+	{
+		for (auto &bit : sig) {
+			if (bit.wire != NULL && bits.count(bit) == 0)
+				return false;
+			if (bit.wire == NULL && bit.data == RTLIL::State::Sx)
+				return false;
+		}
+		return true;
+	}
+	bool check_def(const RTLIL::SigBit &bit) const
+	{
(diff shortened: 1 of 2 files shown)


widlarizer:
This doesn't create the ideal behavior and the implementation is hard to defend either. Looking into alternatives

KrystalDelusion:
Also given that `setundef` and `opt_expr` both refer to it as "replace undriven nets with undef (x) constants" in the help, it would be good to be consistent here (unless I'm misinterpreting, and this isn't the same).

jix:
I think in a synthesis context it makes more sense to treat constant `'z` as undriven and `'x` as driven but don't care, but also (optionally?) allow used-but-never-driven values to be converted into don't cares (producing a warning when that happens). Without that distinction, I'd be worried about what could happen to bidirectional signals.

For bufnorm I also had to add code to `opt_clean` to handle a buffered 'z signal as undriven, which moves us slightly in that direction. It doesn't really make sense from an EE perspective, but some way to mark a subset of the outputs bits of a `$buf` as undriven was an absolute requirement to make bufnorm work in the presence of bidirectional connections, and that seemed like the least risky choice.

While that was a somewhat forced step, I feel like we should have a discourse thread on how to make 'z, 'x and undriven more consistent and meaningful before deciding to merge changes to specific passes. Especially if there is a chance we will replace bufnorm with signorm which would allow us to undo this forced step.

PR #5008, opened 2025-04-09
wheels: fix builds, add linux aarch64 to ci

donn:

Essentially, something is attempting to build the Yosys EXE when you build libyosys.so now. With `ENABLE_PYTHON_CONFIG_EMBED=0`, the Yosys EXE will always fail to build.

Furthermore, because `ENABLE_PYOSYS` now attempts to build a wheel, building a wheel has become recursive.

This commit uses a supplementary set of libs for the EXE (EXE_LIBS) so it and libyosys.so can be built simultaneously, as well as a new Makefile flag, `ENABLE_WHEEL`, to prevent the aforementioned recursion.

I also enabled aarch64 Linux in the CI because it's publicly available now.

Create a venv and `pip3 install .`.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -14,16 +14,12 @@ jobs:
               runner: "ubuntu-22.04",
               archs: "x86_64",
             },
-            ## Aarch64 is disabled for now: GitHub is committing to EOY
-            ## for free aarch64 runners for open-source projects and
-            ## emulation times out:
-            ## https://github.com/orgs/community/discussions/19197#discussioncomment-10550689
-            # {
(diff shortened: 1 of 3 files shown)


donn:
Successful builds on all four platforms: https://github.com/donn/yosys/actions/runs/14352887468/job/40235544676

mmicko (approved):
LGTM. Wheels working, but still some issues building only pyosys that needs fixing.

georgerennie:
The install target needs gating behind `ENABLE_WHEEL` instead of `ENABLE_PYOSYS` too

mmicko (approved):

Issue #4964, opened 2025-03-26
commit c5fd96ebb vs. gcc15 vs. ppcle64 in Fedora


gsomlo:
### Version

commit c5fd96ebb

### On which OS did this happen?

Linux

### Reproduction Steps

I maintain yosys in Fedora. With each new release, the source RPM is built for a set of different architectures, including x86_64, aarch64, etc., including ppc64le.

While working on version 51, I noticed that `make test` fails, only on ppcle64, and only on fedora versions >= 42
(the latter most likely correlated with gcc version 15).

After a bisect, I found the first commit where this happens to be c5fd96ebb ("macc_v2: Start new cell"), authored by @povik 

Attached is the full ppcle64 fedora mock build log illustrating that problem. [c5fd96ebb0c9067927bd4df381af92783bb4b547.build.log.txt](https://github.com/user-attachments/files/19468432/c5fd96ebb0c9067927bd4df381af92783bb4b547.build.log.txt)

### Expected Behavior

`make test` succeeds on all architectures, including ppcle64

### Actual Behavior

`make test` fails on ppcle64, gcc15 (fedora 42 and later)

KrystalDelusion:
> (the latter most likely correlated with gcc version 15).

Ah, looks like we only have CI for gcc-13 or older.  And gcc-14 was failing when it was set to gcc-13 as the newest supported.  Should probably look into *why* that was failing.

gsomlo:
On Wed, Mar 26, 2025 at 03:15:53PM -0700, KrystalDelusion wrote:
>     (the latter most likely correlated with gcc version 15).
> 
> Ah, looks like we only have CI for gcc-13 or older. And gcc-14 was failing when
> it was set to gcc-13 as the newest supported. Should probably look into why
> that was failing.

The weird thing is that (for now) it only fails on
(gcc15 AND arch==ppcle64). I.e., it passes all tests
(even with gcc15) on x86_64.

On one hand, ppcle64 is the "tail wagging the dog" in
that I don't know anyone who might really care about
the intersection of yosys and that arch... :)

OTOH, just because it fails to fail on x86_64, doesn't
mean there isn't something subtly wrong and ppcle64 is
just an "early warning"...

KrystalDelusion:
Are you able to retrieve the `tests/arch/nanoxplore/meminit.log` file?  According to the log file you shared that should contain the failure
```
[...]

PR #5015, opened 2025-04-10
memory_libmap: Move MapWorker into loop

KrystalDelusion:

#4964 revealed that a potential segfault was introduced with #4892 because of scope issues with the MapWorker.  The segfault was only occurring on ppcle64 when compiled gcc15, but was able to be verified with Clang's address sanitizer (ASAN).  I'm guessing some particular combination of optimisations meant that the memory was being freed prior to the reassignment at the end of the loop, leading to the segfault.  And also for whatever reason this wasn't picked up by valgrind, only ASAN.

By moving the declaration of `worker` into the for loop it no longer needs to be reassigned at the end of the loop so the scoping is safe again.

There might be some performance reduction here, since the method from #4892 only re-indexed when the memory was emitted.  But given that the only time time it isn't emitted is when the memory uses the logic fallback, this is probably not a huge hit.  The other option would be to add some way to update the `MapWorker`, probably during the `MemMapping::emit()` step so that it doesn't need reassigning while still being able to have up to date indexing.

Build Yosys with `SANITIZER = address` and run `t_mem1.ys` from the `tests/arch/quicklogic/qlf_k6n10f` directory (test file can be generated by calling `python3 mem_gen.py`).  Without this change, the address sanitizer will report a `stack-out-of-scope` error.  With this change it does not.

### DIFF
--- a/passes/memory/memory_libmap.cc
+++ b/passes/memory/memory_libmap.cc
@@ -2232,10 +2232,10 @@ struct MemoryLibMapPass : public Pass {
 			if (module->has_processes_warn())
 				continue;
 
-			MapWorker worker(module);
 			auto mems = Mem::get_selected_memories(module);
 			for (auto &mem : mems)
 			{
+				MapWorker worker(module);
 				MemMapping map(worker, mem, lib, opts);
 				int idx = -1;
 				int best = map.logic_cost;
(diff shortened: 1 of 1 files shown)


widlarizer:
I would prefer if we knew why this fixes anything. I'll try to take a closer look. The asan trace doesn't indicate this to be the root cause either

KrystalDelusion:
> The asan trace doesn't indicate this to be the root cause either

My interpretation was that the `pool::hashtable` (which is what goes out of scope) belongs to the `mfp::database`, which belongs to the `SigMap::database`, which belongs to the `MapWorker::sigmap`, and that `MemMapping` having a reference to `&worker` is insufficient for preventing it going out of scope when `worker` gets reassigned.

widlarizer:
Oh, yeah somehow I forgot how everything works when I made that change originally, the worker was being reallocated on a stack that gets destroyed and rebuilt. My bad!

PR #5017, opened 2025-04-11
verific: bit blast RAM if using mem2reg or ram_style attribute

mmicko:
If there is certain user attribute set on RAM array signal we can blast RAM nets.
With this chage it will react on `mem2reg` attribute existance or `ram_style` value of `logic` or `register`.

### DIFF
--- a/frontends/verific/verific.cc
+++ b/frontends/verific/verific.cc
@@ -1446,6 +1446,30 @@ void VerificImporter::import_netlist(RTLIL::Design *design, Netlist *nl, std::ma
 		module_name = "\\" + sha1_if_contain_spaces(module_name);
 	}
 
+	{
+		Array ram_nets ;
+		MapIter mem_mi;
+		Net *mem_net;
+		FOREACH_NET_OF_NETLIST(nl, mem_mi, mem_net)
+		{
+			if (!mem_net->IsRamNet()) continue ;
+
+			if (mem_net->GetAtt("mem2reg")) {
+				ram_nets.Insert(mem_net) ;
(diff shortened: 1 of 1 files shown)


akashlevy:
@mmicko Does blasting the net prevent `$mem` inference? Or will it still create `$mem` cells, one for each bit in the memory word?

mmicko:
@akashlevy it will prevent $mem inference and will create registers + logic to implement it

nakengelhardt:
For context, this is a feature that has existed in `read_verilog` for a long time, now able to be used with `verific` as well.

The immediate motivation for the PR is for VHDL, where it's possible to describe memories that, with verific, produce `$mem_v2` cells that `memory_map` is not equipped to handle; e.g. with an asynchronous reset of the entire array. Analogous constructs in (System)Verilog would always produce FFs instead of memories even without the attribute in both `verific` and `read_verilog` (the blasting is invoked automatically). We want to eventually add support to `memory_map` to handle this case without user intervention (https://github.com/YosysHQ/yosys/issues/5022), but for now this gives at least a workaround to avoid the crash.

akashlevy:
Got it. The Verific frontend actually recognizes a similar attribute called `(* no_ram *)`, which has a similar effect to what you are doing here. The difference is that the Verific RAM net never gets inferred in the first place, so the RAM necessarily becomes flops.

I bring this up because at Silimate, we had tried a similar blasting trick and found some corner cases where it still infers a `$mem` on the blasted RAM net. This ended up causing issues in the `memory` pass later

PR #5021, opened 2025-04-14
memory_libmap: fix MapWorker memory allocation

widlarizer:
Alternative to #5015 without potential performance regression


widlarizer:
This makes the test asan-clean (except for some verilog frontend leaks that I don't care to figure out)

Issue #4921, opened 2025-02-28
share: limit SAT effort


widlarizer:
### Feature Description

The share pass is constructed on top of using a SAT solver to find pairs of cells that it can merge. The solver used is MiniSAT, which supports limiting the conflict and propagation budgets, which should act as deterministic timeouts, but it's used through the ezSAT wrapper. We should be able to give users better control over the SAT solver effort attempted by the pass if we figure out how to make use of the budgets. This would allow users to opportunistically run share on larger designs

Issue #4920, opened 2025-02-28
share: replace activation patterns with more SAT


widlarizer:
## Current implementation

Two cells of of the supported types can be shared if they are never both observable at once.
The function `find_cell_activation_patterns` constructs the observability conditions for a cell.
For the purposes of this pass:
+ A cell is observable if an output sigbit is observable.
+ A sigbit is observable if it's wired to a module output.
+ A sigbit is observable if it's wired to a module instance input.
+ A sigbit is observable if a cell propagates it.
+ A `$mux` or `$pmux` only propagates an input sigbit when the control signal has the correct value.
+ Other cells are modeled as always propagating all input sigbits to all output sigbits.

The term "activation patterns" is used to call a representation of the observability condition in disjunctive normal form (DNF). Constructing a DNF can lead to memory usage that's exponential with regards to the cell count, see #4904.

Let `obs(x)` mean "cell `x` is observable`. After the activation patterns are found, currently, pairs of cells `x`, `y` of the same shareable type are checked using a SAT solver for observability. First, it's checked that either cell is observable. Then, `obs(x)∧obs(y)` is solved for. If UNSAT, it's true that "x and y are never both observable at once" and may be shared.

`x` and `y` are replaced by a cell that accommodates either the computation of the value of `x` or `y` based on when an input multiplexer selects between the inputs to `x` and the inputs to `y`. The control signal `s` to this multiplexer is set to either implement `obs(x)` or to implement `obs(y)` based on a score heuristic.

## Proposed changes

Since the observability logic, currently expressed in DNF, is going into a solver, we may directly construct it within the solver and avoid constructing the DNF. This is the easy part.

If that is done, the `share` logic no longer has access to `obs(x)` and `obs(y)` in a practical way, which we need to construct `s`. But that was sub-optimal as well:

![Image](https://github.com/user-attachments/assets/62b5d7f6-1e58-4ee3-9d1b-9c67392e1ec4)

This is `share` ran on
```
assign o1 = s3 ? s4 ? (s0 ? a1 + b1: 1'b0) : 1'b0 : 1'b0;
assign o2 = s1 ? s2 ? (s0 ? 1'b0 : a2 + b2) : 1'b0 : 1'b0;
```

[...]

PR #5023, opened 2025-04-14
share: Restrict activation patterns to potentially relevant signals

jix:
In case the two sets of activation patterns are mutually exclusive without considering the logic feeding into the activation signals, an activation condition can only be relevant if present in both sets with opposite polarity.

This detects pattern-only mutual exclusion by running an additional SAT query before importing the input cone logic. If that is already UNSAT, we remove all non-relevant condition and re-simplify the remaining patterns.

In cases of pattern-only mutual exclusion, this will often produce much smaller selection logic and avoid the more costly SAT query that includes the input cones.

This effectively provides a limited version of the interpolant idea explained in #4921, but in a way that can be easily retrofitted to the current implementation of share, as this keeps the DNF based activation pattern approach (which also means this isn't a full replacement for #4921).

### DIFF
--- a/passes/opt/share.cc
+++ b/passes/opt/share.cc
@@ -1000,6 +1000,53 @@ struct ShareWorker
 		}
 	}
 
+	pool<std::pair<SigBit, State>> pattern_bits(const pool<ssc_pair_t> &activation_patterns)
+	{
+		pool<std::pair<SigBit, State>> bits;
+		for (auto const &pattern : activation_patterns) {
+			for (int i = 0; i < GetSize(pattern.second); ++i) {
+				SigBit bit = pattern.first[i];
+				State val = pattern.second[i];
+				bits.emplace(bit, val);
+			}
+		}
+		return bits;
+	}
+
+	void onesided_restrict_activiation_patterns(
+			pool<ssc_pair_t> &activation_patterns, const pool<std::pair<SigBit, State>> &other_bits)
+	{
+		pool<ssc_pair_t> new_activiation_patterns;
+
+		for (auto const &pattern : activation_patterns) {
+			ssc_pair_t new_pair;
+			for (int i = 0; i < GetSize(pattern.second); ++i) {
+				SigBit bit = pattern.first[i];
+				State val = pattern.second[i];
+				if (other_bits.count({bit, val == State::S0 ? State::S1 : State::S0})) {
+					new_pair.first.append(bit);
+					new_pair.second.append(val);
+				}
+			}
+			new_activiation_patterns.emplace(std::move(new_pair));
+		}
+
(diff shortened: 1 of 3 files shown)


passes/opt/share.cc
```diff
@@ -1299,45 +1346,57 @@ struct ShareWorker
...
-				int sub2 = qcsat.ez->expression(qcsat.ez->OpOr, other_cell_active);
 				if (!qcsat.ez->solve(sub2)) {
 					log("      According to the SAT solver the cell %s is never active. Sharing is pointless, we simply remove it.\n", log_id(other_cell));
 					cells_to_remove.insert(other_cell);
 					shareable_cells.erase(other_cell);
 					continue;
 				}
 
-				qcsat.ez->non_incremental();
+				if (pattern_only_solve) {
```
widlarizer, line 1368: Isn't this condition inverted? The "then" branch is pre-existing code, now used for when there is no solution found before importing the design. When I invert this, a simple example I cooked up starts working. It's not enough to pass tests/sat but it seems necessary

widlarizer (changes requested):

passes/opt/share.cc
widlarizer: Oh it actually is sufficient to pass the tests, I just got pranked by the `Expected error pattern 'Signal difference' found !!!` line again. It's so mean

widlarizer: No, the condition is correct because it's true if not mutually exclusive, so there has to be a different reason

widlarizer:
I'm looking into what's up with the pass as I get a regression even with a hand-written test that I only created to take a look at how this new functionality works in practice

jix:
I figured out how the test broke and am working on a fix for that, if your test case involves merging more than a single pair of cells that might be the same issue.

passes/opt/share.cc
jix: If `pattern_only_solve` is true, that means there was a counterexample to mutual exclusion when just considering the patterns. That's the case where we do want to fall back to the previous behavior, as they might still be mutually exclusive when we re-solve with the input cones imported, but in that case we can't optimize the activation patterns.

widlarizer (approved):
The idea as described in the commit message and implemented is sound. My test passes now as well but doesn't have to be added as the fix is covered by the new test. I synthesis on jpeg and ibex and saw it make improve the runtime slightly with no impact on QoR on those, not digging deeper atm though as this was just a quick smoke test

For the next release notes, this should be presented as a runtime optimization as well as a reduction in the timing impact this area-focused pass has

Also this is related to #4920 (interpolants) rather than the previously linked #4921 (limiting SAT effort)

PR #5025, opened 2025-04-15
satgen: Fix $macc_v2 x-prop

povik:

Fix the x-prop behavior of `$macc_v2` in satgen. The satgen extension from #4818 missed to account for `$macc_v2` having an extra `C` port.

Testing coupled to #5024


KrystalDelusion (approved):
Seems to work fine.  There are some problematic seeds (e.g. `-n 1 -s 1744844438`), but I suspect this is not unique to the _v2 or the changes here but rather trying to sat solve multiplication of non trivial widths.

PR #5031, opened 2025-04-16
stat: fix sequential area not being included in addition/multiplication

suisseWalter:
In Stat there is a separate calculation for the sequential area of a circuit. 
This output of stat has been incorrect if the design isn't flattened. 

We would like to be able to calculate accurate sequential area for non flattened designs. 

add the sequential area to the macro that implements the add and multiply operations. 

Check a report of a flattened and unflattened design. 
Before it had different sequential areas. Now the sequential area will be the same.

### DIFF
--- a/passes/cmds/stat.cc
+++ b/passes/cmds/stat.cc
@@ -40,7 +40,7 @@ struct statdata_t
 			X(num_ports) X(num_port_bits) X(num_memories) X(num_memory_bits) X(num_cells) \
 			X(num_processes)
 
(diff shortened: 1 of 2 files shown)


widlarizer:
Thanks for the contribution, can you extend `tests/various/stat.ys` to cover this behavior? The test should fail on behavior prior to this PR

suisseWalter:
I have now extended the testcase. 
It should cover most cases of the behaviour in the total area report. Not just this bug.

widlarizer:
Thanks!

PR #5034, opened 2025-04-16
liberty: fix tests

widlarizer:
`set -e` is the correct solution but `&&` prevents it from working. Doing `|| exit 1` on a for loop only checks the last iteration of the for loop, or something of that sort. `pipefail` added out of superstition

### DIFF
--- a/tests/liberty/run-test.sh
+++ b/tests/liberty/run-test.sh
@@ -1,16 +1,17 @@
 #!/usr/bin/env bash
-set -e
+set -eo pipefail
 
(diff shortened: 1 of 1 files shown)


widlarizer:
This fixes tests erroneously passing when something other than the diff failed, such as this:
```
run-test.sh: line 12: 83043 Segmentation fault      (core dumped) ../../yosys -q -s $x -l ${x%.ys}.log
Running options_test.ys..
...passed tests in tests/liberty
```

PR #5037, opened 2025-04-17
add flatten -nocleanup option

nakengelhardt:

Some (third-party) analysis passes work only on flattened modules, but still want to report back results for each module in the hierarchy.

Add an option to not delete the submodules when running flatten. (The functionality is already there, just needs a flag.)

Load a hierarchical design (with a top module), run `flatten -nocleanup`, observe the number of modules being more than 1, but no module instantiating another module.


nakengelhardt:
Confirmed working by customer.

PR #5039, opened 2025-04-18
gatemate: WRITE_THROUGH mode change

mmicko:
Removing "wrtrans" setting makes designs to work on actual hardware. Without this change LiteX cores are not working correctly and return bad chars over serial port.

Pinging @pu-cc

### DIFF
--- a/techlibs/gatemate/brams.txt
+++ b/techlibs/gatemate/brams.txt
@@ -34,7 +34,6 @@ ram block $__CC_BRAM_TDP_ {
 		}
 		portoption "WR_MODE" "WRITE_THROUGH" {
 			rdwr new;
-			wrtrans all new;
 		}
 		wrbe_separate;
 		optional_rw;


mmicko:
@mwkmwkmwk Since I am not sure if this is the actual solution, I was trying to narrow down this issue, and figured out the issue occurs when there is "read first" memory instantiated. Thing is that there are just two configuration modes for this FPGA, “NO_CHANGE” and “WRITE_THROUGH”. Documentation mentioned that read first can be implemented by a two clock cycle read and write implementation, where in 1st only read should be allowed and in 2nd write should be performed. Do you think this is proper solution in this case, and also do you have any suggestions for additional tests to perform on actual hardware ?

KrystalDelusion:
I believe that should be @wanda-phi if the `.mailmap` changes are anything to go by.

KrystalDelusion:
Based on the documentation for the gatemate memory
image

and `wrtrans`
> The wrtrans property is only allowed on write ports and defines behavior when another synchronous read port reads from the memory cell at the same time as the given port writes it. The values are:
>
> - old: the read port will get the old value of the cell
> - new: the read port will get the new value of the cell

I think the existing `wrtrans all new` is the correct setting, though I'm not sure how exactly that interacts with the optional output register.

KrystalDelusion:
My understanding is that `rdwr new` affects the value read on the same port, while `wrtrans all new` affects the value read from the other port.  i.e. if a write occurs on port A, and both A and B are reading (and from the same address), then `rdwr new` means that A read is the new value, while `wrtrans all new` means that B read is the new value.

KrystalDelusion:
If you're only running into the issue with read first memory, it means the `wrtrans` setting *is* correct, otherwise it would break write first memories.  Problem might be with how read first emulation is done, and removing the `wrtrans` setting is changing the emulation (or possibly affecting whether emulation is performed on `NO CHANGE` or `WRITE THROUGH` mode).  While this change might not break cases where write first transparency is desired, it will be adding extra unnecessary logic.

KrystalDelusion (approved):
Looking again at https://www.colognechip.com/docs/ug1001-gatemate1-primitives-library-latest.pdf, it does say:
> There is no internal conflict handling when accessing the same address at the same time

which might be why the `wrtrans all new` setting is incorrect on hardware.  Seems better to remove it so that emulation logic can be inserted to guarantee behaviour if needed.

pu-cc:
I can confirm @mmicko's findings, that removing `wrtrans all new` fixes the issues with read-first memory emulation. To be honest, I don't recall the details why it was in there in the first place. The fix doesn’t cause any other issues elsewhere. So, from our side it's fine if it's removed. Thank you!

mmicko:
Thanks for confirmation @pu-cc

PR #5044, opened 2025-04-21
Fixed the problem of not calling abc correctly when using libyosys.so

RonxBulld:
Fixed the problem of not calling abc correctly when using libyosys.so library (it will try to call yosys-abc as command line instead of calling abc::Abc_RealMain)

I tried to use libyosys.so and execute the abc instruction, which resulted in an unexpected error. Yosys tried to execute the yosys-abc program.

Simply add a check and append new switch options to ABCMKARGS.

Try to compile and generate libyosys.so and connect it, then use the C++ interface to call run_pass and execute abc related instructions.


widlarizer:
This PR turns on PIC for abc which might have a negative performance impact. If libyosys is built, it's built in addition to a yosys executable, and abc isn't built twice. This means that we can either
- break building the yosys executable and libyosys in a single build
- risk unnecessarily slowing down abc whenever we're linking abc into the yosys executable
I'll run some synthesis runtime benchmarks with LINK_ABC and PIC vs no PIC and get back to this

widlarizer (approved):
Disregard my previous comment, we already build yosys with `-fPIC` unconditionally, so this does make sense

Issue #5048, opened 2025-04-22
[CXXRTL] udivmod crashes when operand value exceeds logical width and ctlz() returns invalid result


sdjasj:
### Version

Yosys 0.52+32 (git sha1 9d3d0a433, clang++ 14.0.0-1ubuntu1.1 -fsanitize=address -fsanitize=undefined -fPIC -O3 -O1 -fno-omit-frame-pointer -fno-optimize-sibling-calls -fsanitize=address,undefined)

### On which OS did this happen?

Linux

### Reproduction Steps

Below is the Verilog code `test.v` that triggers the issue:

```verilog
`timescale 1ns/1ps
module a (in2, out2);

input wire in2;
output wire out2;

assign out2 = (7'b11100 / ({1'b1, in2}));

endmodule
```

The following command is used for conversion:

```bash
yosys -p "read_verilog test.v; write_cxxrtl test.cpp"
```

Then, the `main.cpp` file is written for simulation:

```cpp
#include <fstream>
#include <iostream>
#include <sstream>
#include <vector>
#include <string>
#include "test.cpp"

using namespace std;
using namespace cxxrtl_design;

int main() {
    std::unique_ptr<p_a> mod = std::make_unique<p_a>();

    mod->step();
    mod->p_in2 = cxxrtl::value<1>(10000u);
    cout << mod->p_in2.val() << std::endl; // 1'2710

    mod->step();
    cout << mod->p_out2.val() << std::endl;

    return 0;
}
```

The following command is used for compilation:

```bash
clang++ -w -g -O3 -std=c++14 -I $(yosys-config --datdir)/include/backends/cxxrtl/runtime main.cpp -o a.out
```

Running `./a.out` results in the following error:

```
a.out: /usr/local/share/yosys/include/backends/cxxrtl/runtime/cxxrtl/cxxrtl.h:612: std::pair<value<Bits>, value<Bits>> cxxrtl::value<7>::udivmod(value<Bits>) const [Bits = 7]: Assertion `divisor_shift >= 0' failed.
Aborted (core dumped)
```

I believe the issue is related to https://github.com/YosysHQ/yosys/issues/5042, where CXXRTL does not truncate out-of-range values during assignment, leading to an error during division:

```cpp
p_out2 = div_uu<7>(value<7>{0x1cu}, value<6>{0x1u}.concat(p_in2).val()).slice<0>().val();
```

[...]

PR #5053, opened 2025-04-23
fix udivmod crashes when operand value exceeds logical width

sdjasj:
This PR fixes https://github.com/YosysHQ/yosys/issues/5048, which causes an assertion failure during division when the actual value of the divisor exceeds its bit-width range.

The issue is resolved by modifying the `divmod_uu()` function to truncate the divisor to its declared bit-width before zero-extending it to the unified width, ensuring that the value passed to `udivmod()` remains within valid bounds:

```cpp
value<Bits> divisor = b.template trunc<BitsB>().template zext<Bits>();
```


whitequark:
Looks good, thanks!

whitequark (approved):

PR #5054, opened 2025-04-23
Simple lib estimation

suisseWalter:
On request of @phsauter  I'm working on making stat more expressive. 
The goal is to have more Information for unflattened designs. 
Especially regarding area. It is already possible to use a "liberty" file that helps to estimate the area after the techmap step. where the designs normally aren't yet flattened. With this file it's already possible to get a good area estimates pre optimization. But with the current stat command the resulting report doesn't contain a lot of useful information. 

In a later PR I will then also implement a way to get area estimates already before techmap. 

A sample report can be found here: [croc_chipexperiment.txt](https://github.com/user-attachments/files/19863246/croc_chipexperiment.txt)
A sample mock Liberty file can be found here : [experiment_mock.lib.txt](https://github.com/user-attachments/files/19847766/experiment_mock.lib.txt)
For the format I took inspiration from: https://github.com/pulp-platform/croc/pull/39. I haven't implemented all the data that is provided there. 

This PR isn't yet final. Many parts of the code aren't in a clean state. 
Regarding the generated report I'm looking for feedback on the following points:
- Is the naming of the columns clear?
- should I also do the hierarchical/local separation for wires/ports/memories/processes? We tend to yes. 
- Should I include the the differentiation between combinatorial and sequential area? should it go everywhere or only in the design hierarchy? we tend to no. 
- I started to differentiate between a cell and a instance, while in RTLIL everything is called a cell. For the user it makes a difference if something is a instance or a standard cell. If I keep this separation, what is a good way to do it? The current implementation depends on cells being in a lib file. Should it just be "all cells not defined in the design are stdcells" ? 
     - with the current idea the problem would be, that if one doesn't provide a lib file everything becomes a instance. Is this acceptable? 
- currently unknown area cells do not get propagated up. and the final report only includes the ones from the top-level module. Should it propagate everything up? We tend to a no, as instances also don't get propagated up.
- This doesn't yet have a json version. I will implement this as soon as everything else has been decided. As I assume that there are tools that are reading the json automatically, should there be a legacy mode for the json, such that these tools don't break? or should this be a new json mode with a different name, such that no scripts need to be changed?

We would like to have more detailed reporting of the area used by parts of the circuit. 
Parsing the modules hierarchically allows us to get the area of used modules. 

To test this have a design after Techmap and not yet (completely) flattened. 
Run stat `stat  -liberty <path/to/mock.lib>` where mock.lib can be the lib file from above.


widlarizer:
Hi, thanks for working on this. I like the general direction.

> In a later PR I will then also implement a way to get area estimates already before techmap.

In terms of gate equivalents? Keep in mind that Yosys is not just an ASIC toolchain. Best not to assume the target is a CMOS chip, it might be a an FPGA, a SAT solver, or an academic coarse-grain reconfigurable accelerator

- The naming is verbose, I feel like "Area including submodules" "Count excluding submodules" would do just as well
- Yes, let's make it consistent between cells and other entities
- I don't have strong feelings about seeing sequential vs combinational but I think that sequential elements are rarely affected by small changes for which one wants to see results in stat, especially since abc doesn't usually touch your flops. Separating combinational out makes changes in it more visible, I suppose. I don't see why not to keep it
- The output should make it clear when an area and cell count is real. The presence of an unknown module without liberty file coverage shouldn't break stat, but recursively everything including it in the module hierarchy should then be clearly marked for uncertainty in cell count and area as well if applicable. Unknown modules should show up as if they were cells as they're leaf nodes just like cells. Internal cells should behave the same way (invalidate submodules-included numbers throughout the hierarchy, contribute to cell counts)
- Yes, let's keep a legacy json mode unless it's convenient to extend the current json output such that there's just extra values in the json tree, which I guess isn't the case
- Sure, hierarchically ordering the modules would be best. Don't leave out modules outside the hierarchy

Additionally,

- Without any area data there's no reason to print columns for area
- Please prefer rebasing on top of main to keep the PR clean

suisseWalter:
> >     In a later PR I will then also implement a way to get area estimates already before techmap.
> 
> In terms of gate equivalents? Keep in mind that Yosys is not just an ASIC toolchain. Best not to assume the target is a CMOS chip, it might be a an FPGA, a SAT solver, or an academic coarse-grain reconfigurable accelerator

The goal is to have a file to describe the "area" of the cells. Similar to the liberty files used currently by stat. This way people can write their own description of "area". I'm open to discuss this also on slack(cwalter_at_ethz.ch), as it isn't the primary goal of this PR. 

Thanks for the feedback. I will implement it. 
> The naming is verbose, I feel like "Area including submodules" "Count excluding submodules" would do just as well

I currently use the word "Instances" to describe what you call submodules. Is submodules the prefered yosys naming and should I therefore switch over to using it everywhere?

widlarizer:
It's not yosys naming, but consider the hierarchy foo/bar/baz. If you look at stat and it says "baz: area including instances" then it may seem like it's referring to instances of baz, rather than instances in baz. I'm not sure how exactly to word it, I don't like introducing new terms but "submodule" at least seems unambiguous. I'm definitely open to alternate ideas

widlarizer:
We discussed this at the dev jour fixe (this is an event open to contributors [here](https://meet.jit.si/yosyshq-slack-devel-discuss) every Monday at 3PM central european time, feel free to pop in) and have reached some consensus for how we'd like to see this play out.
- Whenever possible the output should fit within 80 columns, as is yosys convention. This requires minimizing whitespace within reasonable bounds
- Hierarchical summary shouldn't have unused zeroed "local" columns
- The default view should match the current functionality, meaning no "hierarchical" columns. This will help keep it concise. The user story you're serving is "I'm synthesizing an ASIC and want to see where my area is coming from so that I can make it smaller". However, a common user story is "My synthesis flow targeting anything is creating this type of cell and I would like to see which pass created it". For this you don't care about including instances
- The "hierarchical" columns should then be hidden behind an option
- We agreed on "submodule" being nicely unambiguous but are definitely open to anybody chiming in with what terminology other tools may use in such circumstances
- The new functionality should have a matching json output mode to deter users from parsing the human-readable version

A possible way of formatting to reduce total output width at the cost of line count is this:
```
cell count
| area
| |
```
rather than
```
+------ cell count
| +---- area
||
```

suisseWalter:
[croc_chip_postTechmap.json](https://github.com/user-attachments/files/21728538/croc_chip_postTechmap.json)
[croc_chip_postTechmap.rpt.txt](https://github.com/user-attachments/files/21728539/croc_chip_postTechmap.rpt.txt)
[croc_chip_postTechmap_h.json](https://github.com/user-attachments/files/21728540/croc_chip_postTechmap_h.json)
[croc_chip_postTechmap_h.rpt.txt](https://github.com/user-attachments/files/21728542/croc_chip_postTechmap_h.rpt.txt)

I have now updated the Reports. with most of the suggestions. 
- the hierarchical reporting is now behind the -hierarchy option
- There is now a json option. (automatic if -hierarchy -json)
- improved design/labeling according to discussion. 
- area gets hidden automatically if no arrea can be calculated. 
- implemented the same for wires/ports/etc
- entries of "0" have been replaced with "-"
- the entire thing has been rebased on main. due to some issues this resulted in a force push, that cleared the history. 

I did not implement the following:
- have the modules listed in hierarchical order. It turns out this would require quite a lot of work to do it reliably with out of tree modules. I think this would only be a very minor improvement. 

With that I will take the PR out of draft mode.

KrystalDelusion:
There's a change in behavior here that is causing `tests/various/stat.ys` to fail.
Previously:
```
4. Printing statistics.

=== child ===

   Number of wires:                  5
   Number of wire bits:              5
   Number of public wires:           5
   Number of public wire bits:       5
   Number of ports:                  5
   Number of port bits:              5
     sg13g2_dfrbp_1                  1

   Chip area for module '\child': 47.174400
     of which used for sequential elements: 47.174400 (100.00%)

=== top ===

   Number of wires:                  3
   Number of wire bits:              3
   Number of ports:                  3
   Number of port bits:              3
   Number of cells:                  4
     child                           2
     sg13g2_and2_1                   2

   Area for cell type \child is unknown!

   Chip area for module '\top': 18.144000
     of which used for sequential elements: 0.000000 (0.00%)

=== design hierarchy ===

   top                               1
     child                           2

   Number of wires:                 13
   Number of wire bits:             13
   Number of public wires:          13
   Number of public wire bits:      13
   Number of ports:                 13
   Number of port bits:             13
   Number of cells:                  4
     sg13g2_and2_1                   2
     sg13g2_dfrbp_1                  2

   Chip area for top module '\top': 112.492800
     of which used for sequential elements: 94.348800 (83.87%)
```
In this PR:
```
4. Printing statistics.

=== child ===

        +----------Local Count including submodules.
        |        +-Local Area including submodules.
        |        | 
        5        - wires
        5        - wire bits
        5        - public wires
        5        - public wire bits
        5        - ports
        5        - port bits
        1   47.174 cells
        1   47.174   sg13g2_dfrbp_1

   Chip area for module '\child': 47.174400
     of which used for sequential elements: 47.174400 (100.00%)

=== top ===

        +----------Local Count including submodules.
        |        +-Local Area including submodules.
        |        | 
        3        - wires
        3        - wire bits
        3        - public wires
        3        - public wire bits
        3        - ports
        3        - port bits
        3   18.144 cells
        2   18.144   sg13g2_and2_1
        2        - submodules
        2        -   child

   Chip area for module '\top': 112.492800
     of which used for sequential elements: 94.348800 (83.87%)

=== design hierarchy ===

        +----------Local Count including submodules.
        |        +-Local Area including submodules.
        |        | 
        3   18.144 top
        1   47.174 child

        +----------Local Count including submodules.
        |        +-Local Area including submodules.
        |        | 
        3        - wires
        3        - wire bits
        3        - public wires
        3        - public wire bits
        3        - ports
        3        - port bits
        3   18.144 cells
        2   18.144   sg13g2_and2_1
        2        - submodules
        2        -   child

   Chip area for top module '\top': 112.492800
     of which used for sequential elements: 94.348800 (83.87%)

processed statistics
```

Previously the statistics for modules didn't include those for sub modules, except in the final "design hierarchy" section.  It looks like that has now changed so that the "chip area for module '\top'" includes the area of both child submodules.  But then the "Local Count" and "Local Area" don't include the child statistics in either the output for the top module or the design hierarchy.

For the Visual Studio build failure, I believe you should be using `snprintf` instead of `std::snprintf`; `yosys_common.h` has `#  define snprintf _snprintf` for windows compatibility.

passes/cmds/stat.cc
```diff
@@ -238,22 +327,119 @@ struct statdata_t
...
-		log("   Number of cells:             %6u\n", num_cells);
-		for (auto &it : num_cells_by_type)
-			if (it.second)
-				log("     %-26s %6u\n", log_id(it.first), it.second);
+		if (std::abs(value) < 1e-12)
+			return "       -";
+
+		char buf[16];
+
+		int len = std::snprintf(buf, sizeof(buf), "%.3f", value);
```
KrystalDelusion:
```suggestion
		int len = snprintf(buf, sizeof(buf), "%.3f", value);
```

passes/cmds/stat.cc
```diff
@@ -238,22 +327,119 @@ struct statdata_t
...
+		if (len > 0 && buf[len - 1] == '.')
+			--len;
+		buf[len] = '\0';
+
+		if (len <= 7) {
+			return std::string(8 - len, ' ') + std::string(buf);
+		}
+
+		// use scientific notation, this should always fit in 8 characters
+		std::snprintf(buf, sizeof(buf), "%8.3G", value);
```
KrystalDelusion:
```suggestion
		snprintf(buf, sizeof(buf), "%8.3G", value);
```

passes/cmds/stat.cc
```diff
@@ -238,22 +327,119 @@ struct statdata_t
...
+	{
+		if (print_area) {
+			if (print_hierarchical) {
+				log(" %8s-%8s-%8s-%8s-%s\n", "+", "--------", "--------", "--------", "Count including submodules.");
+				log(" %8s %8s-%8s-%8s-%s\n", "|", "+", "--------", "--------", "Area including submodules.");
+				log(" %8s %8s %8s-%8s-%s\n", "|", "|", "+", "--------", "Local count, excluding submodules.");
+				log(" %8s %8s %8s %8s-%s\n", "|", "|", "|", "+", "Local area, excluding submodules.");
+				log(" %8s %8s %8s %8s \n", "|", "|", "|", "|");
+			} else {
+				log(" %8s-%8s-%s\n", "+", "--------", "Local Count including submodules.");
```
KrystalDelusion: Despite the text saying it includes submodules, it matches the count/area when excluding submodules.

KrystalDelusion (changes requested):
It would be good to add a test (or modify the existing `tests/various/stat.ys`) to check that the local/global count/area matches the expected values with and without `-hierarchy`.

passes/cmds/stat.cc
```diff
@@ -520,6 +845,7 @@ struct StatPass : public Pass {
 		}
 
 		log("\n");
+		printf("processed statistics\n");
```
KrystalDelusion: This shouldn't be here

suisseWalter:
[croc_chip_postTechmap.json](https://github.com/user-attachments/files/21748273/croc_chip_postTechmap.json)
[croc_chip_postTechmap.rpt.txt](https://github.com/user-attachments/files/21748275/croc_chip_postTechmap.rpt.txt)
[croc_chip_postTechmap_h.json](https://github.com/user-attachments/files/21748276/croc_chip_postTechmap_h.json)
[croc_chip_postTechmap_h.rpt.txt](https://github.com/user-attachments/files/21748277/croc_chip_postTechmap_h.rpt.txt)

Both content errors have been fixed. 
The labeling is now correct and the design hierarchy now properly contains the values including submodules. 

I also need to fix the same thing for the json output.

passes/cmds/stat.cc
suisseWalter: fixed in latest version.

passes/cmds/stat.cc
suisseWalter: fixed in latest version.

passes/cmds/stat.cc
suisseWalter: fixed in latest version.

passes/cmds/stat.cc
suisseWalter: fixed in latest version.

suisseWalter:
[croc_chip_postTechmap.json](https://github.com/user-attachments/files/21754798/croc_chip_postTechmap.json)
[croc_chip_postTechmap.rpt.txt](https://github.com/user-attachments/files/21754801/croc_chip_postTechmap.rpt.txt)
[croc_chip_postTechmap_h.json](https://github.com/user-attachments/files/21754802/croc_chip_postTechmap_h.json)
[croc_chip_postTechmap_h.rpt.txt](https://github.com/user-attachments/files/21754803/croc_chip_postTechmap_h.rpt.txt)

Now the json is also updated.

suisseWalter:
The old testcase has been updated for the new formatting and a new testcase for the hierarchical view has been added.

passes/cmds/stat.cc
```diff
@@ -238,22 +327,132 @@ struct statdata_t
...
+			}
+		} else {
+			if (print_hierarchical) {
+				log(" %8s-%8s-%8s-%s\n", "+", "--------", "--------", "Count including submodules.");
+				log(" %8s %8s-%8s-%s\n", "|", "+", "--------", "Local count, excluding submodules.");
+				log(" %8s %8s \n", "|", "|");
+			} else if (print_global_only) {
+				log(" %8s-%8s-%s\n", "+", "--------", "Count including submodules.");
+				log(" %8s %8s \n", "|", "|");
+			} else {
```
KrystalDelusion: There's only one heading here but it still draws two lines
```suggestion
			} else if (print_global_only) {
				log(" %8s-%8s-%s\n", "+", "--------", "Count including submodules.");
				log(" %8s \n", "|");
			} else {
```

suisseWalter:
[croc_chip_postTechmap.json](https://github.com/user-attachments/files/21809254/croc_chip_postTechmap.json)
[croc_chip_postTechmap.rpt.txt](https://github.com/user-attachments/files/21809258/croc_chip_postTechmap.rpt.txt)
[croc_chip_postTechmap_h.json](https://github.com/user-attachments/files/21809260/croc_chip_postTechmap_h.json)
[croc_chip_postTechmap_h.rpt.txt](https://github.com/user-attachments/files/21809263/croc_chip_postTechmap_h.rpt.txt)

Updated the non hierarchical report to only show the local area. 

>Second thing (not a blocker but just a thought I had), looking at the following output, the module stats show both "including submodules" and "excluding submodules" even when there aren't any submodules. Would it be worthwhile checking if num_submodules == 0 and just giving the stats once?

There is a second case where this comes up. What if a module has no area? should the area column be printed?

I decided to print the columns anyway to keep a consistent formatting over the entire report. I feel like having consistent formatting improves the readability of the report.

KrystalDelusion:
> I decided to print the columns anyway to keep a consistent formatting over the entire report. I feel like having consistent formatting improves the readability of the report.

That makes sense!

KrystalDelusion (approved):

Issue #5026, opened 2025-04-15
Possibly incorrect non-blocking assignment in ecp5 cells_sim


secworks:
The cells_sim library for ECP5 contains an always comb (always(*)) select assignment that use non-blocking assignments. This makes verilator choke, and is normally an incorrect use of non-blocking assignments (AFAIK). If this is not intentional I can provide a fix.

https://github.com/YosysHQ/yosys/blob/19845be85c1fd2385dcfc07cdc39c2c6aae26a7f/techlibs/ecp5/cells_sim.v#L389

KrystalDelusion:
I don't think this is intentional, and most of the other `always @(*)` blocks use `=`.  At a glance it looks like some of the other techlibs have the same issue, gatemate and greenpak4, as well as some of the tests (in .ys files).  Would you be willing to fix those as well?

secworks:
I've created a PR for the issue in sim_cells: #5055.
Will look at the others and create PRs for those separately.

secworks:
I've also created a PR for greenpak4, see #5057.

KrystalDelusion:
Thanks for doing all of those, I'll go ahead and close this issue now that they're all merged :)

PR #5055, opened 2025-04-23
Changing non clocked alway assignment to blocking.

secworks:

Possibly an incorrect use of non-blocking assignment in a non-clocked always process.
see #5026 for discussion

Changing to blocking assignments.

Simulate and compare behavior.


gatecat (approved):

secworks:
See issue #5026.

PR #5056, opened 2025-04-23
Change to blocking assignments in non-clocked process.

secworks:

Use of non-blocking assignments in non-clocked processes.

Changed to blocking assignments.

Simulate and test with expected behavior.


secworks:
See issue #5026.

widlarizer (approved):

PR #5057, opened 2025-04-23
Change to use blocking assignments in non-clocked processes.

secworks:

Use of non-blocking assignments in non-clocked always processes.

Change to blocking assignments.

SImulate and checked expected behavior.


secworks:
See issue #5026.

widlarizer (approved):

Issue #4792, opened 2024-12-02
read_verilog gives multiple drivers to variables initialized both in declaration and initial


widlarizer:
### Version

`Yosys 0.47 (git sha1 647d61dd9, ccache clang++ 16.0.6 -fPIC -O3 -flto=thin)`

### On which OS did this happen?

Linux

### Reproduction Steps

`read_verilog a.v; proc -noopt; check`
```v
module a;
	integer i = 1'b1;
	initial begin
		i = i + 1'b0;
	end
endmodule
```

### Expected Behavior

The tool evaluates the assignments in some order (Verilog 2005 LRM 6.2.1). For SystemVerilog, the order has to be such that the declaration assignment is evaluated first (SV 2012 LRM 6.8).

### Actual Behavior

The tool evaluates both assignments with no concept of sequence, creating conflicting drivers for the variable and a logic loop. This could lead to incorrect synthesis and prevents efficient optimization (see #4782 #4783)

PR #5058, opened 2025-04-23
check: promote some problems to errors by default, add -permissive

widlarizer:
Some problems `check` discovers are too severe to keep as warnings by default.

Yosys is expected to continue to cover dodgy scenarios, such as combinational loops in flip flop models[[1]](#1). multiple drivers for a signal[[2]](#2), and undriven signals. When you mix a couple of these you can end up with nasty results[[3]](#3). For greater user experience, we should treat multiple drivers and possibly combinational loops as an error at any point we may encounter it, unless the user specifies otherwise. This is different from `check -assert` which makes every `check` warning an error, effectively.

The user should be able to ask for the old behavior by either
- directly using a flag like `check -permissive`
- indirectly using a flag like `synth -check-permissive` for any script pass running `check` (that's `proc` and `synth*`)
- maybe as a driver option like `yosys --check-permissive`

Alternatively, to avoid strongly customized scripts, we can keep old `check` behavior and add a flag like `check -smart` and use that in script passes. This way we still get the friendlier UX of `synth*` with less breakage.

A change to `check` like this could also be a step towards design-level invariant flags ("design has no loops", "design only includes fine-grained cells" etc) that have been brought up a couple of times by now

<a id="1">[1]</a> `tests/asicworld/code_hdl_models_d_ff_gates.v`, this should make the CI red on this draft
<a id="2">[2]</a> [opt_clean.cc](https://github.com/YosysHQ/yosys/blob/bf20bc0848e2089a4ca814c60471e82c7b666986/passes/opt/opt_clean.cc#L147C1-L147C5)
<a id="3">[3]</a> https://github.com/YosysHQ/yosys/issues/4979#issuecomment-2824113291

### DIFF
--- a/kernel/log.cc
+++ b/kernel/log.cc
@@ -400,6 +400,16 @@ void log_file_error(const string &filename, int lineno,
 	logv_file_error(filename, lineno, format, ap);
 }
 
+std::string fmt(const char *format, ...)
+{
+	std::string s;
+	va_list ap;
+	va_start(ap, format);
+	s = vstringf(format, ap);
+	va_end(ap);
+	return s;
+}
+
 void log(const char *format, ...)
 {
 	va_list ap;
--- a/kernel/log.h
+++ b/kernel/log.h
@@ -126,6 +126,7 @@ void logv_warning_noprefix(const char *format, va_list ap);
 [[noreturn]] void logv_error(const char *format, va_list ap);
 [[noreturn]] void logv_file_error(const string &filename, int lineno, const char *format, va_list ap);
 
+std::string fmt(const char *format, ...)  YS_ATTRIBUTE(format(printf, 1, 2));
 void log(const char *format, ...)  YS_ATTRIBUTE(format(printf, 1, 2));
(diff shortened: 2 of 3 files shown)


KrystalDelusion:
I think having the option of making more problems as errors, without making all problems as errors, is good.  I don't know if doing it by default is the best way (not that I think it shouldn't be the default, I'm just not sure how much breakage there would be vs benefit of avoiding later problems)

widlarizer:
One pattern I'm seeing twice in the failing tests already is a latch with a comb path from Q to D, that isn't ever activated all the way through:
`tests/various/dynamic_part_select/latch_002_gate.v`
`abc9_test009` and others in `tests/simple_abc9/abc9.v`

widlarizer:
The tests are passing except for quicklogic, but it really looks like we're not quite ready yet. We have to fix #5106 and #4792 first, or just work around #5106 in quicklogic sim files. If we only work around it though, we're going to be breaking more user designs to require permissive checks

Issue #5020, opened 2025-04-11
AddressSanitizer issues


KrystalDelusion:
### Version

0.52+10

### On which OS did this happen?

Linux

### Reproduction Steps

Build with `SANITIZER = address` then run `make test`.

Optionally use `ASAN_OPTIONS=halt_on_error=0` and comment out line 856 of Makefile (`MK_TEST_DIRS += tests/arch/quicklogic/qlf_k6n10f`).  That one seems to be unrecoverable with ASAN enabled even with halt_on_error disabled.  Run individually with `make makefile-tests/tests/arch/quicklogic/qlf_k6n10f`.

### Expected Behavior

No errors from ASAN.

### Actual Behavior

LeakSanitizer: detected memory leaks

- `in Yosys::AST::AstNode::clone() const frontends/ast/ast.cc:256:18` during 
     - `tests/verilog/typedef_*.ys`,
     - `tests/svtypes/typedef_struct_port.ys`, as well as an indirect leak in
     - `tests/various/struct_access.ys`.

- `in fstReaderIterBlocks2 libs/fst/fstapi.cc` during
   - `tests/sat/grom.ys`

- `in strdup` during
  - `tests/various/struct_access.ys`

- `in frontend_verilog_yylex frontends/verilog/verilog_lexer.l:434:20` during
  - `tests/verific/blackbox_ql.ys`
  - `tests/arch/quicklogic/qlf_k6n10f/*.ys`

- `in Verific::Strings::save` during
  - `tests/verific/setenv.ys`

and the AddressSanitizer: stack-use-after-scope from #4964 during `tests/arch/quicklogic/qlf_k6n10f/t_mem1.ys`.

PR #5060, opened 2025-04-24
rtlil: less sketchy lifetimes

widlarizer:
We have some asan findings noted in #5020 so I'm making a pass through the codebase and seeing what's ripe for a cleanup. So far I have replaced the destruct guard with a singleton pattern and demoted string interning data structures from globals into it, which doesn't resolve any known problems but seems like the right thing to do

- [ ] performance impact check

### DIFF
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -32,19 +32,6 @@
 
 YOSYS_NAMESPACE_BEGIN
 
-bool RTLIL::IdString::destruct_guard_ok = false;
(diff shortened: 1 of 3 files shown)


KrystalDelusion:
Thanks for looking into these @widlarizer

widlarizer:
I don't think this PR is worth the time put into it since my understanding is that the current mechanism is just as safe and fundamentally equivalent even if it looks messy. I built this on a wild goose chase and there's nothing to gain here beyond minor readability improvements

PR #5063, opened 2025-04-25
AstNode garbage collector for debugging

widlarizer:
Sanitizers can detect the presence of a memory safety violation, but not with much detail. I'm adding a debugging garbage collector for tracking lifetimes and provenance of AstNodes. AstNodes that aren't part of some AstModule AST are collected, optionally debugged, and destroyed, as they'd otherwise leak

Tracking is done by collecting an `std::set<AstNode*> nodes`. `AstNode` constructors register `this` into `nodes` and destructors deregister `this`. At a moment when all AstNodes should be "owned" by something point to them, they are tagged, and therefore, the set of untagged nodes should be empty. These nodes get killed to trip the sanitizer if they were being "owned" by anything that wasn't tagged. They can also be dumped. The code is off by macro, just like legacy debug features such as YOSYS_USE_STICKY_IDS. It's expected that the use case is only for debugging as it slows down the program to a large extent and on its own doesn't really dump anything. The PR is temporarily based on top of #5060 which makes the diff messy, sorry

In case you want to test the change, you could comment out some `delete` statement from ast.cc or simplify.cc, build yosys with `SANITIZER=address`, see the sanitizer detect a leak on yosys exit from an invocation where read_verilog is run, change the undef in ast.h to an ifdef, rebuild, and no longer see that leak

### DIFF
--- a/frontends/ast/ast.cc
+++ b/frontends/ast/ast.cc
@@ -290,8 +290,8 @@ void AstNode::delete_children()
 	for (auto &it : children)
 		if (Tagger::get().nodes.count(it))
 			delete it;
-		else
-			log("skip child %p\n", it);
+		// else
+		// 	log("skip child %p\n", it);
 #else
 	for (auto &it : children)
 		delete it;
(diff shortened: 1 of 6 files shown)


widlarizer:
No use after #5135

PR #5068, opened 2025-04-28
Updates to bugpoint

KrystalDelusion:

While working on a how-to guide for using `bugpoint` (especially in the context of being able to direct people who are making git issues) I ran into a few problems/pain points with the command.

- [x] Document the (already existing) `-wires` flag
- [x] Add an `-expect-return` option to differentiate e.g. SEGFAULT from other crashes
- [x] Add `-err-grep` option to check stderr (useful for non-yosys errors)
- [x] ~~Grep from stdout?~~
- [x] Add `raise_error` dev pass
- [x] `bugpoint` tests

Run `make makefile-tests/tests/bugpoint`.


whitequark (approved):
At a *very* cursory glance, looks fine.

widlarizer (approved):
Nice work. I would have expected for `raise_error` to have a primary mode of behavior to just raise an error rather than to raise it based on an attribute, but I can see how this was made to test bugpoint. What's the point of `-suffix`? That file isn't touched by the user, right? I'm used to only looking at the final file I `dump` after `bugpoint` is done

KrystalDelusion:
I added `-suffix` so that you could have multiple instances of `bugpoint` running in the same directory which (iirc) was for it to work with the existing test infrastructure.

KrystalDelusion:
> I would have expected for `raise_error` to have a primary mode of behavior to just raise an error rather than to raise it based on an attribute

If there is a use case for it to fall back to a default error instead of doing nothing then we can add that; `bugpoint` should handle that fine, but the `raise_error` tests would need to be updated since they rely on having a bunch of tests in the same file and only the last one crashes.  Or have it gated by a flag like `raise_error -always` or something.

widlarizer:
`-always` sounds good. We could call it a followup though to me it seems confusing without it

KrystalDelusion:
Rebased on main to get ASAN tests

Edit: seems github is having some issues with runners, will re-run failing builds later

Issue #5069, opened 2025-04-28
read_verilog: empty parameter value leaks memory


widlarizer:
### Version

main

### On which OS did this happen?

Linux

### Reproduction Steps

```
module leaky (
);
 	bar #(.paramname()) barinstance ();
endmodule

module ok (
);
 	bar #() barinstance (.portname());
endmodule
```
Build yosys with SANITIZER=address and `read_verilog bug.sv`

### Expected Behavior

Yosys runs and exits peacefully

### Actual Behavior

See quicklogic leaks in #5020

KrystalDelusion:
Is an empty parameter supposed to mean "use the default value"?

widlarizer:
That seems to be the intention, but I haven't checked in the standard

KrystalDelusion:
Changing this slightly to
```verilog
module bar ( input portname );
parameter paramname = 7;
endmodule

module leaky (
);
 	bar #(.paramname()) barinstance ();
endmodule

module ok (
);
 	bar #() barinstance (.portname());
endmodule

module reference (
	input a
);
	bar #(.paramname(3)) barinstance (.portname(a));
endmodule

```
and calling `hierarchy; dump`, `read_verilog` gives
```
attribute \hdlname "leaky"
attribute \src "bug.sv:5.1-8.10"
module \leaky

  attribute \module_not_derived 1
  attribute \src "bug.sv:7.23-7.37"
  cell \bar \barinstance
  end
end

attribute \hdlname "ok"
attribute \src "bug.sv:10.1-13.10"
module \ok

  attribute \module_not_derived 1
  attribute \src "bug.sv:12.11-12.36"
  cell \bar \barinstance
    connect \portname { }
  end
end

attribute \hdlname "reference"
attribute \src "bug.sv:15.1-19.10"
module \reference

  attribute \src "bug.sv:16.8-16.9"
  wire input 1 \a

  attribute \module_not_derived 1
  attribute \src "bug.sv:18.23-18.49"
  cell \bar \barinstance
    parameter signed \paramname 3
    connect \portname \a
  end
end
```

While `verific` gives
```
attribute \library "work"
attribute \hdlname "reference"
attribute \src "bug.sv:15.8-15.11"
module \reference

  wire $verific$barinstance$bug.sv:18$1

  attribute \src "bug.sv:16.8-16.9"
  wire input 1 \a

  attribute \src "bug.sv:18.23-18.49"
  cell \bar \barinstance
    parameter \paramname 3
[...]

PR #5070, opened 2025-04-28
quicklogic: workaround for empty parameter value memory leak

widlarizer:
See #5069


KrystalDelusion:
I'm disinclined to merge this, since it's changing the test to not trigger ASAN instead of actually fixing the problem.  Unless there is some reason we would want to include ASAN in CI and this is a blocker for doing so.

widlarizer:
I'll leave that up to you. The idea is not to close #5069 by doing this but to allow me to from time to time run the test suite with whatever sanitizer just to check, or to use the sanitizer to discover problems in sketchy code I'm working on myself without these

Issue #5074, opened 2025-04-29
CXXRTL: Missing sign extension before shift operation for signed values


sdjasj:
### Version

Yosys 0.52+117 (git sha1 bfe05965f, clang++ 14.0.0-1ubuntu1.1 -fsanitize=address -fsanitize=undefined -fPIC -O3 -O1 -fno-omit-frame-pointer -fno-optimize-sibling-calls -fsanitize=address,undefined)

### On which OS did this happen?

Linux

### Reproduction Steps

This is the original Verilog file `test.v`:
```verilog
`timescale 1ns/1ps
module cqsnqdxwkr (in2, out0);
reg signed [9:0] reg_7;
input wire in2;
output wire [12:0] out0;

    initial begin
        reg_7 = 848;
    end
  
    assign out0 = reg_7 >> in2;
endmodule
```

The following command is used to convert it into a CXXRTL file:
```
yosys -p "read_verilog test.v; write_cxxrtl test.cpp"
```

Then, the following `main.cpp` is written to perform simulation:
```cpp
#include <iostream>
#include "test.cpp"
    
using namespace std;
using namespace cxxrtl_design;

int main() {
    std::unique_ptr<p_cqsnqdxwkr> mod = std::make_unique<p_cqsnqdxwkr>();
    mod->p_in2 = cxxrtl::value<1>(1u);

    mod->step();

    cout << (mod->p_out0.get<uint32_t>()) << endl;

    return 0;
}
```

Compiled with:
```
clang++ -w -g -O3 -std=c++14 -I $(yosys-config --datdir)/include/backends/cxxrtl/runtime main.cpp -o a.out
```

Run:
```
./a.out
```

The output is:
```
424
```

I believe there is an issue with this result. Here's the analysis:

1. `reg_7` is a 10-bit signed number initialized to 848, which in binary is `10'b11_0101_0000`.

2.  
> The right operand is always treated as an unsigned number and has no effect on the signedness of the result. The result signedness is determined by the left-hand operand and the remainder of the expression.

According to the description on page 53 of the IEEE Std 1364™-2005 standard, the signedness of the expression `reg_7 >> in2` depends on the left operand `reg_7`. Since `reg_7` is signed, the result of the expression is also signed.

[...]

PR #5075, opened 2025-04-29
fix missing sign extension before shift operation for signed values

sdjasj:
This PR fixes the issue described in https://github.com/YosysHQ/yosys/issues/5074. Now, for expressions like `assign t = a op b`, when the width of the expression is larger than the width of `a` and `a` is signed, `a` will first be sign-extended before performing the shift operation. 

The fix is implemented by applying the above logic in the `shr_su` function：
```cpp
template<size_t BitsY, size_t BitsA, size_t BitsB>
CXXRTL_ALWAYS_INLINE
value<BitsY> shr_su(const value<BitsA> &a, const value<BitsB> &b) {
	if (BitsY > BitsA) {
		value<BitsY> extended_a = a.template scast<BitsY>();
		return extended_a.shr(b);
	}
	return a.shr(b).template scast<BitsY>();
}
```

After the fix, the CXXRTL produces results consistent with other simulators for the test case related to https://github.com/YosysHQ/yosys/issues/5074.

### DIFF
--- a/backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
+++ b/backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
@@ -1769,10 +1769,6 @@ value<BitsY> shr_uu(const value<BitsA> &a, const value<BitsB> &b) {
 template<size_t BitsY, size_t BitsA, size_t BitsB>
 CXXRTL_ALWAYS_INLINE
 value<BitsY> shr_su(const value<BitsA> &a, const value<BitsB> &b) {
-	if (BitsY > BitsA) {
-		value<BitsY> extended_a = a.template scast<BitsY>();
-		return extended_a.shr(b);
-	}
 	return a.template scast<BitsY>().shr(b);
 }
 
--- a/backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
+++ b/backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
@@ -1769,6 +1769,10 @@ value<BitsY> shr_uu(const value<BitsA> &a, const value<BitsB> &b) {
 template<size_t BitsY, size_t BitsA, size_t BitsB>
 CXXRTL_ALWAYS_INLINE
 value<BitsY> shr_su(const value<BitsA> &a, const value<BitsB> &b) {
+	if (BitsY > BitsA) {
+		value<BitsY> extended_a = a.template scast<BitsY>();
+		return extended_a.shr(b);
+	}
 	return a.shr(b).template scast<BitsY>();
 }
 
--- a/backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
+++ b/backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
@@ -1773,7 +1773,7 @@ value<BitsY> shr_su(const value<BitsA> &a, const value<BitsB> &b) {
 		value<BitsY> extended_a = a.template scast<BitsY>();
 		return extended_a.shr(b);
 	}
-	return a.shr(b).template scast<BitsY>();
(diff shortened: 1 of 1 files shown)


backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
```diff
@@ -1769,6 +1769,10 @@ value<BitsY> shr_uu(const value<BitsA> &a, const value<BitsB> &b) {
 template<size_t BitsY, size_t BitsA, size_t BitsB>
 CXXRTL_ALWAYS_INLINE
 value<BitsY> shr_su(const value<BitsA> &a, const value<BitsB> &b) {
+	if (BitsY > BitsA) {
+		value<BitsY> extended_a = a.template scast<BitsY>();
+		return extended_a.shr(b);
+	}
 	return a.shr(b).template scast<BitsY>();
```
whitequark, line 1776: Would the following one line change not work? It's been a while since I touched that code but `struct sext_cast` already includes the conditional I think.

```suggestion
	return a.template scast<BitsY>().shr(b);
```

sdjasj: > `struct sext_cast` already includes the conditional I think

Yes, I originally intended to make this change as well. However, I wasn't sure if BitsY could be smaller than BitsA, so I only handled the case where BitsY is greater than BitsA, while keeping the original functionality of the function intact. If BitsY is guaranteed to be greater than or equal to BitsA, then I think your modification is more elegant and appropriate.

whitequark: I don't think there are restrictions like that on BitsY (see rtlil.cc:1145) but `.template scast()` does handle the case where the result is smaller than `this`: see cxxrtl.h:321 (it uses enable_if to do so).

sdjasj: You are right, I didn't notice the role of `cxxrtl.h:321`. Your fix is more appropriate, thank you for the correction :).

whitequark (approved):
LGTM, thanks!

backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
```diff
@@ -1769,7 +1769,11 @@ value<BitsY> shr_uu(const value<BitsA> &a, const value<BitsB> &b) {
 template<size_t BitsY, size_t BitsA, size_t BitsB>
 CXXRTL_ALWAYS_INLINE
 value<BitsY> shr_su(const value<BitsA> &a, const value<BitsB> &b) {
-	return a.shr(b).template scast<BitsY>();
+	if (BitsY > BitsA) {
+		value<BitsY> extended_a = a.template scast<BitsY>();
+		return extended_a.shr(b);
+	}
```
whitequark, line 1775: Actually I think lines 1772-1775 are left over?

whitequark (changes requested):

backends/cxxrtl/runtime/cxxrtl/cxxrtl.h
sdjasj: Oh, they should indeed be removed, my mistake.

sdjasj: The unnecessary code lines have been deleted. I apologize for not being very familiar with the process of submitting PRs and for wasting your time.

sdjasj:
@whitequark Hi! Could you please take some time to check it? :) I think everything should be fine now.

whitequark (approved):

whitequark:
Thanks!

PR #5078, opened 2025-04-29
Change the implementation of log_debug in kernel/log.h

RonxBulld:
Change the implementation of log_debug in kernel/log.h from a macro function to a normal function.

Because when integrating Yosys in other applications, if you do not use the code like `using namespace Yosys;`, `log` will not be found correctly, which leads to the mandatory use of `using namespace` and potential namespace pollution problems.

It is just variable length parameter forwarding.

Just run the regression tests.


widlarizer (approved):
Agreed

Issue #5077, opened 2025-04-29
Flag to avoid writing Yosys version numbers into AIG files.


dinoruic:
### Feature Description

Hi all,

I'm using the Yosys's `write_aiger` function to generate AIGs. But Yosys writes its version number into the comment section which makes things non-reproducible every time there is a Yosys update.

Is it possible to add a flag to `write_aiger` that would simply leave the comment section empty?

Thanks!

mikesinouye:
Hi @widlarizer and @jix , @dinoruic works with us. Is it possible that the `-T` flag that omits the footer containing the version in the Yosys log file can also be respected in all of the passes, or that this flag can be added to the passes that do so? It seems that there are a few `Generated by <version>` footers per pass.

widlarizer:
hope the newly introduced `--no-version` works for you from now on

mikesinouye:
This is great! Thank you very much @widlarizer , I will integrate this with our setup.

PR #5086, opened 2025-05-05
driver: add --no-version to suppress writing Yosys version

widlarizer:
See #5077. In this PR, I add a global switch, and make printing of "Generated by" and "creator" lines by backends and other passes conditional on this switch

You may test this with invocations like this:
`yosys -p "read_verilog tests/sat/alu.v; proc; write_json" | grep creator`
`yosys --no-versions -p "read_verilog tests/sat/alu.v; proc; write_json" | grep creator`


KrystalDelusion:
I think the `backends/smt2/smtio.py` raises a question; should we suppress version *numbers* but keep the static "Generated by Yosys"?  In which case a simpler solution might be to replace the full `yosys_version_str` with just `"Yosys"`.  Suppressing the "Generated by" message entirely might still be desired, but I think that is a different thing.

PR #5095, opened 2025-05-06
rtlil: enable single-bit vector wires

widlarizer:
```
So, instead of: `wire foo;`
I'd like: `wire [0:0] foo;`.

### DIFF
--- a/backends/json/json.cc
+++ b/backends/json/json.cc
@@ -178,8 +178,10 @@ struct JsonWriter
 			f << stringf("          \"direction\": \"%s\",\n", w->port_input ? w->port_output ? "inout" : "input" : "output");
 			if (w->start_offset)
 				f << stringf("          \"offset\": %d,\n", w->start_offset);
-			if (w->upto)
+			if (w->width != 1 && w->upto)
 				f << stringf("          \"upto\": 1,\n");
+			if (w->width == 1 && w->sbvector)
+				f << stringf("          \"sbvector\": 1,\n");
 			if (w->is_signed)
 				f << stringf("          \"signed\": %d,\n", w->is_signed);
 			f << stringf("          \"bits\": %s\n", get_bits(w).c_str());
@@ -270,8 +272,10 @@ struct JsonWriter
 			f << stringf("          \"bits\": %s,\n", get_bits(w).c_str());
 			if (w->start_offset)
 				f << stringf("          \"offset\": %d,\n", w->start_offset);
-			if (w->upto)
+			if (w->width != 1 && w->upto)
 				f << stringf("          \"upto\": 1,\n");
+			if (w->width == 1 && w->sbvector)
+				f << stringf("          \"sbvector\": 1,\n");
 			if (w->is_signed)
 				f << stringf("          \"signed\": %d,\n", w->is_signed);
 			f << stringf("          \"attributes\": {");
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

**Review by whitequark — CHANGES_REQUESTED** [review]
I think this complicates RTLIL further without any clear (<del>or stated</del> justification was added after I posted the comment) benefit, and causes churn to RTLIL consumers.

**Comment by widlarizer** [comment]
Yeah I accidentally sent an unwritten description at first. Anyway, the distinction somehow does exist in the domain yosys is modeling (systemverilog). The design is with the full intention of the impact on downstream users being minimal as it's possible to entirely ignore this change without functional damage or build breakage

also cc @QuantamHD @mikesinouye

**Comment by whitequark** [comment]
> Anyway, the distinction somehow does exist in the domain yosys is modeling (systemverilog).

I was under impression that Yosys supports multiple languages (SV, VHDL, Amaranth, etc) rather than being beholden to being a SystemVerilog frontend. If it was the case that it is modeling SystemVerilog then it does a remarkably awful job of it since it doesn't preserve types, interfaces, enumerations, `logic` vs `wire`, does a haphazard job representing strings, and so on. Singling out a 1-bit vector change to be added as a core RTLIL concept rather than, say, an attribute is pretty odd given this context.

**Comment by whitequark** [comment]
I think this should be an attribute, similar to how SV enumerations are modelled using an attribute. Making it an attribute will reduce impact on downstream users since attributes will already roundtrip through any existing RTLIL consumer (including old versions of Yosys) while `sbvector` will not.

**Comment by povik** [comment]
I agree with Catherine's assessment and I'm also curious: What's the concrete use case where preserving this distinction is critical?

**Comment by mikesinouye** [comment]
> I agree with Catherine's assessment and I'm also curious: What's the concrete use case where preserving this distinction is critical?

We'd like to get the output netlist to have consistent naming conventions with the default behavior of other downstream tools. As an example, after ingesting the Yosys netlist into openroad later we'd like compatibility with already-written SDCs, floorplans, UPF, etc. generated by other tools.

In a Yosys flow, after running `splitnets -ports`, wires with [0:0] will be handled differently than single-bit wires. However, other tools preserve the array, so when they run their equivalent `splitnets -ports` pass, things are named according to array conventions causing mismatches as these wires were implictly converted to single bits in Yosys.

As added constraints, we are unsure which signals will be defined as [0:0] beforehand as these are typically parameterized, and we cannot edit the design or downstream artifacts.

In general we are happy with any approach to get this desired behavior, an alternative method to support this use case seems fine by us as well, but this PR's approach solves our issue.

**Comment by jix** [comment]
My reasoning for keeping this in Wire was that I consider this to be part of the same data as `offset` and I wanted to keep both at the same place. (I do think it was a mistake to have `offset` part of Wire given that it doesn't affect RTLIL semantics at all and is just there to pass through information from Verilog and VHDL.)

I thought we might be able to do that without causing any backwards compatibility issues but the a) current PR makes some decisions that change things beyond what I was originally thinking of and b) I forgot to consider external producers of textual RTLIL, which I don't want to break.

With that, I do think using an attribute here would be the best choice for now. I also think we can afford the characters to call it `single_bit_vector` so it's easier for users to figure out what it means.

**Comment by mikesinouye** [comment]
Great! Thank you for working on this. Is this ready to be merged?

**Comment by KrystalDelusion** [comment]
Uh, this seems to be leaking memory quite badly; e.g. `tests/arch/anlogic/add_sub.ys`: `SUMMARY: AddressSanitizer: 190742 byte(s) leaked in 1614 allocation(s).`

**Comment by KrystalDelusion** [comment]
```
Direct leak of 74016 byte(s) in 257 object(s) allocated from:
    #0 0x62a2815f390d in operator new(unsigned long) (/home/dawn/yosys/yosys+0xd0190d) (BuildId: aa7c142091acfdca999add6836667528edcd80c3)
    #1 0x62a281d07f72 in Yosys::AST::AstNode::mkconst_int(unsigned int, bool, int) /home/dawn/yosys/frontends/ast/ast.cc:830:18
    #2 0x62a281d3e9af in Yosys::AST::AstNode::simplify(bool, int, int, bool) /home/dawn/yosys/frontends/ast/simplify.cc:2091:43
    #3 0x62a281d34fdc in Yosys::AST::AstNode::simplify(bool, int, int, bool) /home/dawn/yosys/frontends/ast/simplify.cc:1198:18
    #4 0x62a281d30797 in Yosys::AST::AstNode::simplify(bool, int, int, bool) /home/dawn/yosys/frontends/ast/simplify.cc:937:10
    ...
```
Changing https://github.com/YosysHQ/yosys/blob/c21cd300a07d3b00cb12eccc0fa3f66c31c89bf0/frontends/ast/simplify.cc#L2090-L2091
to 
```
if (range_left == range_right && !attributes.count(ID::single_bit_vector))
	set_attribute(ID::single_bit_vector, mkconst_int(1, false));
```
seems to resolve it.

PR #5098, opened 2025-05-06
Add <optional> to haslib.h which uses std::optional

mikesinouye:


mikesinouye:
Hi @widlarizer and @jix, this is a fix for an issue we're seeing. PTAL when you get a chance, thanks

widlarizer (approved):

PR #5100, opened 2025-05-07
rename: add -move-to-cell option in -wire mode

jix:

For some use cases it is desirable to transfer names from wires to their driving cells. 

This adds a `-move-to-cell` option to `rename -wire` which doesn't add a suffix to the cell name but instead adds a suffix to the wire name, if necessary to avoid conflicts.

It also makes `rename -wire` use the correct HDL indices for wires with non-zero-indexing, whether using `-move-to-cell` or not.

By replacing `rename -wire` with `rename -wire -move-to-cell` in an existing flow.

### DIFF
--- a/passes/cmds/rename.cc
+++ b/passes/cmds/rename.cc
@@ -66,7 +66,7 @@ static std::string derive_name_from_src(const std::string &src, int counter)
 		return stringf("\\%s$%d", src_base.c_str(), counter);
 }
 
-static IdString derive_name_from_cell_output_wire(const RTLIL::Cell *cell, string suffix)
+static IdString derive_name_from_cell_output_wire(const RTLIL::Cell *cell, string suffix, bool move_to_cell)
 {
 	// Find output
 	const SigSpec *output = nullptr;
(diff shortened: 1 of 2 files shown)


jix:
@tspyrou @precisionmoon: This implements the renaming we were discussing. To make it easy to test this change in isolation, I rebased this PR branch on top of the Yosys commit that the ORFS master is currently using. Note that this adds a new `rename` variant, so testing it does require also changing `rename -wire` to `rename -wire -move-to-cell` in ORFS's `synth.tcl`.

widlarizer:
This is a pattern that I created with `rename -hide` that prevents the new mode from doing its job as it checks whatever is the directly driven wire without traversing the tree of aliases or querying a sigmap that would have a public name promoted. The purpose for this pass is to be able to refer cells post-synthesis, so I think we can assume such a shape won't happen, but I'm not 100% sure

image

passes/cmds/rename.cc
```diff
@@ -92,13 +92,23 @@ static IdString derive_name_from_cell_output_wire(const RTLIL::Cell *cell, strin
 
 		name += chunk.wire->name.str();
 		if (chunk.wire->width != chunk.width) {
-			name += "[";
-			if (chunk.width != 1)
-				name += std::to_string(chunk.offset + chunk.width) + ":";
-			name += std::to_string(chunk.offset) + "]";
+			int lhs = chunk.wire->to_hdl_index(chunk.offset + chunk.width - 1);
+			int rhs = chunk.wire->to_hdl_index(chunk.offset);
+			if (chunk.wire->upto)
```
widlarizer, line 97: On some level, I would expect printing to respect the uptoness of the original hdl objects, so I don't think we should be undoing it here with the swap

In case it's useful / becomes part of a test, here's the input for trying out this code path
```
read_verilog <<EOF
module top(input [0:7] a, input [0:7] b, output [0:4] o1, output [4:0] o2);
    wire [0:7] int1 = a + b;
    wire [0:7] int2 = a + b;
    assign o1[0:3] = int1[0:3];
    assign o2[3:0] = int2[4:7];
endmodule
EOF

proc -noopt
rename -hide w:int*
opt
rename -wire -move-to-cell
```

widlarizer (approved):

ShinyKate:
@widlarizer any more feedback for this?

precisionmoon:
image
I tried the new option on one design sky130hd / gcd.   There is no QoR impact and the resulting FF names look fine.  Is it safe to replace rename -wire with rename -wire -move-to-cell?

widlarizer:
Yes, feel free to use this in ORFS instead of the current `rename -wire` @precisionmoon

PR #5101, opened 2025-05-07
opt_expr: fix shift optimization with 

### LINKED ISSUE #5099: Incorrect shift optimization when shift amount ≥ `(1 << 32) - 1`
### Version

Yosys 0.53+3 (git sha1 6378ba10e, clang++ 14.0.0-1ubuntu1.1 -Og -fPIC -O1 -fno-omit-frame-pointer -fno-optimize-sibling-calls -fsanitize=address,undefined,leak)

Linux

### Reproduction Steps

The following code `test.v` triggers an error:

```
module top (in1, out40);

    input wire  [3:0] in1;
    output wire [6:5] out40;

    assign out40 = (in1[1:1] >> 32'hffffffffffff);
endmodule
```

Running the following command to optimize it:

```
yosys -p "read_verilog test.v; opt; write_verilog bug.v"
```

produces the optimized code `bug.v`:

```
/* Generated by Yosys 0.53+3 (git sha1 6378ba10e, clang++ 14.0.0-1ubuntu1.1 -Og -fPIC -O1 -fno-omit-frame-pointer -fno-optimize-sibling-calls -fsanitize=address,undefined,leak) */

(* cells_not_processed =  1  *)
(* src = "test.v:2.1-8.10" *)
module top(in1, out40);
  (* src = "test.v:3.23-3.26" *)
  input [3:0] in1;
  wire [3:0] in1;
  (* src = "test.v:4.23-4.28" *)
  output [6:5] out40;
  wire [6:5] out40;
  assign out40 = { in1[1], 1'h0 };
endmodule
```

In theory, `in1[1:1] >> 32'hffffffffffff` should be optimized to `2'h0`, but it is instead optimized to `{ in1[1], 1'h0 }`. This error only occurs when the shift amount is greater than or equal to `(1 << 32) - 1`; when the shift amount is less than `(1 << 32) - 1`, the optimization result is correct as `2'h0`.

### Expected Behavior

The expression `in1[1:1] >> 32'hffffffffffff` should be optimized to a constant `2'h0`, since the shift amount far exceeds the width of the operand, making the result definitively zero.

### Actual Behavior

Yosys incorrectly optimizes the expression to `{ in1[1], 1'h0 }`

georgerennie:
Fixes #5099 by limiting the constant shift amount to int max/min so they don't overflow. This could misoptimize in the case that the signal being shifted is in the range of `2^32` bits wide but that is incredibly unlikely (and plenty of other things likely break around there anyway).

In the process I add a few helper methods to the kernel for more safely calling `as_int`: `try_as_int` that returns an optional value only if the conversion is valid, `convertible_to_int` that returns true if the conversion can be done safely, and `as_int_saturating` that clamps values to int's max/min if they are too large/small.

Note that `as_int_compress` actually already does the same as `try_as_int` but it isn't used anywhere in Yosys and I think the naming is such that its function really isn't clear, so hopefully people would be more likely to see that `try_as_int` is an option over `as_int`.

### DIFF
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -380,6 +380,30 @@ int RTLIL::Const::as_int(bool is_signed) const
 	return ret;
 }
 
+bool RTLIL::Const::convertible_to_int(bool is_signed) const
+{
+	auto size = get_min_size(is_signed);
+	return (size > 0 && size <= 32);
+}
+
+std::optional<int> RTLIL::Const::try_as_int(bool is_signed) const
+{
+	if (!convertible_to_int(is_signed))
+		return std::nullopt;
+	return as_int(is_signed);
+}
+
+int RTLIL::Const::as_int_saturating(bool is_signed) const
+{
+	if (!convertible_to_int(is_signed)) {
+		const auto min_size = get_min_size(is_signed);
+		log_assert(min_size > 0);
+		const auto neg = get_bits().at(min_size - 1);
+		return neg ? std::numeric_limits<int>::min() : std::numeric_limits<int>::max();
+	}
+	return as_int(is_signed);
+}
+
 int RTLIL::Const::get_min_size(bool is_signed) const
 {
 	if (empty()) return 0;
@@ -5460,6 +5484,38 @@ int RTLIL::SigSpec::as_int(bool is_signed) const
 	if (width_)
 		return RTLIL::Const(chunks_[0].data).as_int(is_signed);
 	return 0;
+}
+
+bool RTLIL::SigSpec::convertible_to_int(bool is_signed) const
+{
+	cover("kernel.rtlil.sigspec.convertible_to_int");
+
+	pack();
+	if (!is_fully_const())
+		return false;
+
+	return RTLIL::Const(chunks_[0].data).convertible_to_int(is_signed);
+}
+
+std::optional<int> RTLIL::SigSpec::try_as_int(bool is_signed) const
+{
+	cover("kernel.rtlil.sigspec.try_as_int");
+
+	pack();
+	if (!is_fully_const())
+		return std::nullopt;
+
+	return RTLIL::Const(chunks_[0].data).try_as_int(is_signed);
+}
+
+int RTLIL::SigSpec::as_int_saturating(bool is_signed) const
+{
+	cover("kernel.rtlil.sigspec.try_as_int");
+
+	pack();
+	log_assert(is_fully_const() && GetSize(chunks_) <= 1);
+	log_assert(!empty());
+	return RTLIL::Const(chunks_[0].data).as_int_saturating(is_signed);
 }
 
 std::string RTLIL::SigSpec::as_string() const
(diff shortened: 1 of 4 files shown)


georgerennie:
I've added a fix for #5105 that came because I wasn't computing the sign bit correctly, as well as some more extensive testing of different configurations.

sdjasj:
Sure! Here's your message translated into English:

---

Hi, I think I’ve found another overflow issue — this one isn’t related to shift optimizations, but I find it kind of interesting :)

Consider the following code in `test.v`:

```verilog
module top (
    output wire [32:1] out1,
    output wire [32'h0fff_ffff:1] out2,
);
    wire signed wire_1 = 1'b1;
    assign out1 = wire_1 >>> 32'h8F7A7A7A;
    assign out2 = wire_1 >>> 36'hfffffff;
endmodule
```

Running:

```sh
yosys -p "read_verilog test.v; opt; write_verilog opt.v"
```

produces the following error, stating that the width exceeds an implementation limit:

```
test.v:7: ERROR: Expression width 268435455 exceeds implementation limit of 16777216!
```

However, if the width is set to `32'hffff_ffff` like this:

```verilog
module top (
    output wire [32:1] out1,
    output wire [32'hffff_ffff:1] out2,
);
    wire signed wire_1 = 1'b1;
    assign out1 = wire_1 >>> 32'h8F7A7A7A;
    assign out2 = wire_1 >>> 36'hfffffff;
endmodule
```

Then the optimization process completes successfully without errors, and produces the following `opt.v`:

```verilog
/* Generated by Yosys 0.53+3 (git sha1 f0c726ff1, clang++ 14.0.0-1ubuntu1.1 -fsanitize=address -fPIC -O3 -O1 -fno-omit-frame-pointer -fno-optimize-sibling-calls -fsanitize=address,undefined) */

(* cells_not_processed =  1  *)
(* src = "test.v:1.1-9.10" *)
module top(out1, out2);
  (* src = "test.v:2.24-2.28" *)
  output [32:1] out1;
  wire [32:1] out1;
  (* src = "test.v:3.35-3.39" *)
  output [-1:1] out2;
  wire [-1:1] out2;
  (* src = "test.v:5.17-5.23" *)
  wire wire_1;
  assign out1 = 32'd0;
  assign out2 = 3'h7;
  assign wire_1 = 1'h1;
endmodule
```

Note that the width of `out2` is incorrectly set to `[-1:1]`. I’m not sure whether this behavior is expected, or if it’s due to a lack of proper error handling.

georgerennie:
Hmm yeah given we do error on some cases there it could probably fixed to error on all unsupported widths, but I think it is low priority given we don't expect real designs to have signals with widths that overflow `int`.

sdjasj:
Haha, totally fair — I figured it was an edge case that no sane design would hit 😄

KrystalDelusion (approved):
> hopefully people would be more likely to see that `try_as_int` is an option over `as_int`

Might be good to add a comment on `as_int()` pointing to `try_as_int()` for unknown/potentially large values?

Looks good otherwise.

georgerennie:
> Might be good to add a comment on as_int() pointing to try_as_int() for unknown/potentially large values?

Sure, I've added some doc comments for each

PR #5108, opened 2025-05-09
Create a single-port URAM mapping to support memories 2048 x 144b

marzoul:

URAM mapping currently supports double-port 4096 x 72b mapping.
To implement a one-port 2048 x 144b memory, Yosys currently uses 2 URAM blocks, but there exists a mapping solution to use only 1 block.

For single-port user memories, the two "physical" URAM ports can be used as one single, wider port, performing either Write or Read (very similar to how the 2 ports of BRAMs can be used to implement twice larger ports).

One single URAM block can then implement 2048 x 144b.

I have not actually tested the implementation because I don't have the hardware at reach right now ; just performing benchmarking of synthesis. I confirm that the proposed changes make Yosys generate 1 URAM block for a memory 2048 x 144b.

Anyone with a large enough design that contains a compatible memory is welcome to help simulate and check on actual hardware :-)


KrystalDelusion:
> I confirm that the proposed changes make Yosys generate 1 URAM block for a memory 2048 x 144b.

Can you add a test for this to `tests/arch/xilinx`?  At the moment it seems the only `URAM288` tests we have are for priority ordering so the following may or may not work (without this PR it gives 2). 

```
# 2048 x 144b in one
design -reset
read_verilog priority_memory.v
chparam -set ABITS 11 -set WIDTH 144
synth_xilinx -family xcup -top sp_read_first -noiopad
select -assert-count 1 t:URAM288
```

marzoul:
Thank you for location of the tests, I was a bit lost in these.

The test with current `sp_read_first` will not give 1 block because its implementation allows Write AND Read to occur at same clock cycle. With the merged ports, only Write OR Read is possible at a given time.

So I will add a dedicated test.
I would also like to test that there is no extra logic around the block to ensure correct mapping model : currently there is one extra LUT2 that I would like to disappear.

marzoul:
The test is added.

(In the end I did not find a way to make Yosys _not_ generate an extra LUT - the one generated is actually not needed in some forms of the testcase. Removing that LUT would more broadly affect Yosys memory mapping so this thinking is outside the scope of the present PR.)

KrystalDelusion (approved):

PR #5113, opened 2025-05-10
simplify.cc: Fix mem leak

KrystalDelusion:

From #5020, LeakSanitizer: detected memory leaks in `Yosys::AST::AstNode::clone() const frontends/ast/ast.cc:256:18`.  Identified as being due to the redefinition of the wiretype attribute on the new AstNode when resolving recursive typedef.

Minimal reproducer:
```
read_verilog -sv <<EOF
typedef logic T;
typedef T [3:0] S;
module top;
    S s;
endmodule
EOF
```

Delete the wiretype attribute if it already exists, as is done just above.

Alternatively, the `set_attribute()` method should just do this automatically?  There are enough calls to it that are immediately preceded by either this exact check, the inverse (only adding the attribute if it isn't already), or deleting all children and making the check unnecessary, that this seems like it should just be a standard operation. I pushed [a branch that does this](https://github.com/YosysHQ/yosys/tree/krys/alt_ast_asan) to check CI, so will see if that passes.

Build with ASAN, run minimal reproducer above.

### DIFF
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -1932,6 +1932,8 @@ bool AstNode::simplify(bool const_fold, int stage, int width_hint, bool sign_hin
 			// Prepare replacement node.
 			newNode = template_node->clone();
 			newNode->str = str;
+			if (newNode->attributes.count(ID::wiretype))
+				delete newNode->attributes[ID::wiretype];
 			newNode->set_attribute(ID::wiretype, mkconst_str(resolved_type_node->str));
 			newNode->is_input = is_input;
 			newNode->is_output = is_output;


zachjs (approved):

widlarizer:
Nice, I must have missed this in #5062. About the systematic solution (freeing if attr already set): it's correct iff we can with confidence say that an `AstNode*` in attributes always owns what it points to. In such a case we could use `std::unique_ptr` instead of manually deleting anything. It *looks* like we could in fact do that but I'm not entirely sure. `wire->set_attribute(ID::enum_base_type, child->attributes[ID::enum_base_type]);` in `frontends/ast/simplify.cc` seems to violate this, but it looks like the only place to do so

widlarizer (approved):

KrystalDelusion:
@widlarizer do you want to bring it up at dev JF?

PR #5112, opened 2025-05-09
design.cc: Use on_shutdown method

KrystalDelusion:
_What are the reasons/motivation for this change?_

Fix #5088.

When building `WITH_PYTHON`, where a global list of modules is maintained, deleting a module also erases the entry in said global list.  This can lead to memory corruption if the global list is destructed before the module.

_Explain how this is achieved._

Using `on_shutdown()` instead means the module destructor is explicitly called before the global list can be destructed, preventing the issue.

Also add a comment to `Pass::~Pass()` to suggest the same for future passes that might try to use that (and see this commit/PR in the blame if they need a reason why).  

_If applicable, please suggest to reviewers how they can test the change._

PR #5114, opened 2025-05-10
design: ensure all_modules outlives pass when building WITH_PYTHON

widlarizer:
Alternative to #5112, without using `on_shutdown`. I haven't reproduced the issue yet so I'm not 100% sure this also fixes it but I think it should. The C++ implementation allocates statics with respect to accesses and deallocates in reverse order, so accessing it in the constructor guarantees `all_modules` outlives the `design` pass

### DIFF
--- a/passes/cmds/design.cc
+++ b/passes/cmds/design.cc
@@ -23,11 +23,17 @@
 
 YOSYS_NAMESPACE_BEGIN
 
+// TODO refactor such scattered program state
 std::map<std::string, RTLIL::Design*> saved_designs;
 std::vector<RTLIL::Design*> pushed_designs;
 
 struct DesignPass : public Pass {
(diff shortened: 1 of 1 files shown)


widlarizer:
BTW, `on_register` is similarly silly. It is used to assign to the global scratchpad `constpad` and calls `VerificExtensions::Reset();` in `verific`, but that's it. If we figure out the verific thing, we could stop using `on_register`, too, though we would have to remove direct access to `constpad` from plugin authors. Not that I know of any real need to stop using `on_register`

Xiretza:
This doesn't fix #5008, but I also can't find any trace of this `get_all_modules()` in the output binary.

widlarizer:
I think the "implementation" of the C++ language here should be informed by this added construction order dependency even if the return value is unused and causes no visible code change in the pass cnostructor. But if it's not getting the job done then that's too bad

PR #5116, opened 2025-05-12
Update fstlib

KrystalDelusion:

Memory leak detected with ASAN during `tests/sat/grom.ys`, with memory allocation `in fstReaderIterBlocks2 libs/fst/fstapi.cc`.
Problem occurs in `sim.cc` calling `throw dst_end_of_data_exception()` when the desired number of cycles has been reached; this means that the fst reader can't tidy up after itself and leads to memory leaks.
This doesn't happen when the `-stop` flag is used because the `Yosys::FstData` struct tracks the end time and skips the outer callback if the simulation has gone past the desired end time.

During investigation (while checking if there was a potential fix already upstream) I noticed that fstlib has been made available as a library separate from gtkwave.

- Update fstlib patches to use new code location, as well as adjusting where necessary for code refactoring.
- Move cycle checking into the inner callback along with the time checking; means that the outer callback no longer needs to throw an exception in order to stop checking further values, while still allowing the fst reader to finish reading and deallocate memory.
- Add test file for checking the number of simulation steps to confirm pre- and post-patch equivalence.

Run `tests/sim/sim_cycles.ys` under Yosys built with ASAN enabled.  Pre-patch passes all tests, but fails ASAN (See #5020).  Post-patch passes all tests *and* ASAN.

### DIFF
--- a/kernel/fstdata.cc
+++ b/kernel/fstdata.cc
@@ -206,6 +206,7 @@ static void reconstruct_clb_attimes(void *user_data, uint64_t pnt_time, fstHandl
 void FstData::reconstruct_callback_attimes(uint64_t pnt_time, fstHandle pnt_facidx, const unsigned char *pnt_value, uint32_t /* plen */)
 {
 	if (pnt_time > end_time || !pnt_value) return;
+	if (curr_cycle > last_cycle) return;
 	// if we are past the timestamp
 	bool is_clock = false;
 	if (!all_samples) {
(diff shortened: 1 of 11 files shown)


ig-raiderler:
I'm impressed by the modularity of this codebase. Makes it very maintainable.

mmicko (approved):
LGTM. Great that it is now delivered as separate library, for sure more usable that way.

KrystalDelusion:
Agreed.  Are there any of our patches that would be worth trying to open PRs for up-stream?

mmicko:
@KrystalDelusion for now they are quite small, but you are now maintainer of this :) so feel free to send PR to them upstream. ZLIB patch is for our build environment and also strict_alignment one, but others can be of interest for others.

KrystalDelusion:
I'm not actually sure which of them are c++ specific and which affect c as well, so I'll just leave them as patches for now.

PR #5118, opened 2025-05-13
Replaced `Yosys::RTLIL::Cell::module` with `Yosys::RTLIL::Cell::upscope_module`

RonxBulld:
Replaced `Yosys::RTLIL::Cell::module` with `Yosys::RTLIL::Cell::upscope_module` to avoid misunderstanding due to unfamiliarity with implementation details. And added the `Yosys::RTLIL::Cell::get_proto_module()` interface to obtain the module prototype corresponding to the instance.

`Yosys::RTLIL::Cell::module` is often misleading to developers who are not familiar with the implementation details. It is also a non-standard name, that is, its purpose cannot be inferred from the name alone. I recommend replacing the original name with the more clear `upscope_module`.

Careful renaming and refactoring, and added a convenient query method to get the module corresponding to the instance.

Just run the regression test once.

### DIFF
--- a/backends/cxxrtl/cxxrtl_backend.cc
+++ b/backends/cxxrtl/cxxrtl_backend.cc
@@ -224,7 +224,7 @@ bool is_effectful_cell(RTLIL::IdString type)
 
 bool is_cxxrtl_blackbox_cell(const RTLIL::Cell *cell)
 {
-	RTLIL::Module *cell_module = cell->module->design->module(cell->type);
+	RTLIL::Module *cell_module = cell->upscope_module->design->module(cell->type);
 	log_assert(cell_module != nullptr);
 	return cell_module->get_bool_attribute(ID(cxxrtl_blackbox));
 }
@@ -261,7 +261,7 @@ CxxrtlPortType cxxrtl_port_type(RTLIL::Module *module, RTLIL::IdString port)
 
 CxxrtlPortType cxxrtl_port_type(const RTLIL::Cell *cell, RTLIL::IdString port)
 {
-	RTLIL::Module *cell_module = cell->module->design->module(cell->type);
+	RTLIL::Module *cell_module = cell->upscope_module->design->module(cell->type);
 	if (cell_module == nullptr || !cell_module->get_bool_attribute(ID(cxxrtl_blackbox)))
 		return CxxrtlPortType::UNKNOWN;
 	return cxxrtl_port_type(cell_module, port);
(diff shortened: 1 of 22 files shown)


georgerennie:
This is a breaking change for downstream users and it seems to me that misunderstanding `module`, while annoying, is something that would be caught pretty quickly as soon as you try testing the code. I'm also not convinced this new name is actually more intuitive - upscope isn't a term used elsewhere and I wouldn't immediately understand it I think. It also seems to me there is only one module that a cell could have associated with it so I'm not sure its that unclear. 

Thanks for submitting this but I think I would be somewhat opposed to the change.

whitequark:
Was this PR based on a suggestion by AI tools?

RonxBulld:
> This is a breaking change for downstream users and it seems to me that misunderstanding `module`, while annoying, is something that would be caught pretty quickly as soon as you try testing the code. I'm also not convinced this new name is actually more intuitive - upscope isn't a term used elsewhere and I wouldn't immediately understand it I think.
> 
> Thanks for submitting this but I think I would be somewhat opposed to the change.

Maybe `upscope_module` is not a good name, but it at least describes the function of the field more clearly. If there is a better name, you are welcome to recommend it. Maybe the modification will bring short-term pain, but a good smelling code will bring long-term benefits to the project. :)

RonxBulld:
> I second @georgerennie's comment.
> 
> `Cell::get_proto_module()` could be a nice addition on its own.

Thank you for your recognition, and for the reasons I stated above.

RonxBulld:
> Was this PR based on a suggestion by AI tools?

Many of the ideas came to me about 5 years ago when I started using Yosys, but I have only recently started sharing them.

georgerennie:
> Maybe upscope_module is not a good name, but it at least describes the function of the field more clearly. If there is a better name, please recommend it to me. Maybe the modification will bring short-term pain, but a good smelling code will bring long-term benefits to the project. :)

The exact name is not my main issue, I would also reject a change to `parent_module`. Breaking API changes like this are not just short-term pain, downstream plugins are often not regularly updated/tested against current Yosys so breakages on our end may not be fixed for a long time or at all downstream. This means that it can cause pain for regular users just trying to build yosys+a plugin for a long time, so ideally we want to only make API breaking changes when there is a strong motivation to do so in terms of correctness/safety/performance issues (not the case here).

Additionally I think it very unlikely for a misunderstanding of this name not to be caught quickly - there are only two viable candidates for what could be the `module` for a cell, the one containing it and the one it instantiates. The latter should be implausible to people that understand RTLIL given cells are not required instantiate a module (and do not at all in the case of the in built cells). Even if someone does get these confused, there is no (or at least very very little) chance both meanings are the same module, so code relying on this would fairly obviously do the wrong thing when tested.

I'm going to close this for now because I don't think this is a change we are interested in.

Issue #5092, opened 2025-05-06
Support ifnone in Verilog specify blocks


FlinkbaumFAU:
### Version

Yosys 0.53+0 (git sha1 20921ad90, g++ 13.2.0 -fPIC -O3)

### On which OS did this happen?

Linux

## Yosys Build

Yosys was built on the latest main branch with the provided nix configuration. However, the same error also occured when using the packaged version from nixpkgs (0.52).

## Error Reproduction
The error occurs when trying to read the verilog file defining the standard cells in the IHP-Open-PDK (https://github.com/IHP-GmbH/IHP-Open-PDK) with 
```
read_verilog -lib <path-to-ihp-pdk>/ihp-sg13g2/libs.ref/sg13g2_stdcell/verilog/sg13g2_stdcell.v
```
The error reads:
```
ihp-sg13g2/libs.ref/sg13g2_stdcell/verilog/sg13g2_stdcell.v:36: ERROR: syntax error, unexpected '='
```
Since no error occurs when reading this file with other tools (i.e. Modelsim), I assume the syntax in the file is correct.

### Expected Behavior

Yosys should be able to parse the described file.

### Actual Behavior

Yosys crashes with
```
ihp-sg13g2/libs.ref/sg13g2_stdcell/verilog/sg13g2_stdcell.v:36: ERROR: syntax error, unexpected '='
```

georgerennie:
I suspect this is just that Yosys' specify syntax support is partially incomplete and it doesn't support `ifnone` because no one has needed/bothered to implement it

widlarizer:
To clarify: #5332 (fka #5121) doesn't close the original issue, since it doesn't add support for these langauge constructs in `read_verilog -specify` mode. It only allows the regular mode to ignore it. Also, sg13g2_stdcell.v will continue to be unreadable with `read_verilog -sv` due to I think the use of user-defined primitives, which isn't supported by Yosys

PR #5121, opened 2025-05-14
Add state_dependent_path_declaration so that `ifnone` can be parsed

FlinkbaumFAU:
The parser crashed when an `ifnone`-statement occured. See Bug #5092 .

The `state_dependen_path_declaration` from the specification was added to the parser file.

### DIFF
--- a/frontends/verilog/Makefile.inc
+++ b/frontends/verilog/Makefile.inc
@@ -6,7 +6,7 @@ GENFILES += frontends/verilog/verilog_lexer.cc
 
 frontends/verilog/verilog_parser.tab.cc: frontends/verilog/verilog_parser.y
 	$(Q) mkdir -p $(dir $@)
-	$(P) $(BISON) -Wall -Werror -o $@ -d -r all -b frontends/verilog/verilog_parser $<
+	$(P) $(BISON) -Wall -Wcex -Werror -o $@ -d -r all -b frontends/verilog/verilog_parser $<
 
 frontends/verilog/verilog_parser.tab.hh: frontends/verilog/verilog_parser.tab.cc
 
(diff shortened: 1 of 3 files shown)


frontends/verilog/Makefile.inc
```diff
@@ -6,7 +6,7 @@ GENFILES += frontends/verilog/verilog_lexer.cc
 
 frontends/verilog/verilog_parser.tab.cc: frontends/verilog/verilog_parser.y
 	$(Q) mkdir -p $(dir $@)
-	$(P) $(BISON) -Wall -Werror -o $@ -d -r all -b frontends/verilog/verilog_parser $<
+	$(P) $(BISON) -Wall -Wcex -Werror -o $@ -d -r all -b frontends/verilog/verilog_parser $<
```
KrystalDelusion, line 9: From https://www.gnu.org/software/bison/manual/html_node/Diagnostics.html:
> **cex**
> Provide counterexamples for conflicts. See [Generation of Counterexamples](https://www.gnu.org/software/bison/manual/html_node/Counterexamples.html). Counterexamples take time to compute. The option -Wcex should be used by the developer when working on the grammar; it hardly makes sense to use it in a CI.

I don't think this should be included.

frontends/verilog/verilog_parser.y
```diff
@@ -1541,18 +1541,60 @@ list_of_specparam_assignments:
...
+state_dependent_path_declaration:
+	TOK_IF '(' module_path_expression ')' simple_path_declaration
+	// | TOK_IF '(' module_path_expression ')' edge_sensitive_path_declaration
+	| TOK_IFNONE simple_path_declaration
+	;
+
+module_path_expression:
+	module_path_primary
+	// Flatten out unary_operator to avoid shift/reduce conflict
+	| '!' attr module_path_primary { delete $2; }
```
KrystalDelusion, line 1564: Based on the existing code, these need to be `freeattr($2);` instead to avoid a memory leak (not that I'm sure why they're being deleted).

KrystalDelusion (changes requested):

KrystalDelusion:
Also could you please add test(s) for the `ifnone` parsing?  There are some existing tests in `tests/various/specify.{v,ys}` that you can expand/reference.

mole99:
Hi, I really appreciate this fix. I just ran in the same issue as #5092 when I tried to read in the IHP stdcells as blackboxes.

@FlinkbaumFAU, do you still intend to finish it?

FlinkbaumFAU:
Hi,
sorry I have completely forgotten about this pull request, since I had already moved on by creating my own blackboxes with a python script.
I'll try to fix it this week. I suppose there will be others too with the same problem.

mole99:
Thanks a lot! Yes, it's just a matter of time until someone else stumbles across this issue :)

frontends/verilog/verilog_parser.y
widlarizer: All these rules are only used in `ignored_specify_item` which only applies when `-specify` is off so we discard attributes everything including these attributes that auto-allocate

PR #5127, opened 2025-05-17
Disable STRIP operations when appropriate.

RonxBulld:
It was always difficult to debug the logic in yosys when using it as a dynamic library, and it was further discovered that this was because STRIP was always used to remove symbol information in yosys when installing yosys, regardless of how ENABLE_DEBUG was set. I don't know if this is by design or just an unexpected behavior.

The STRIP variable is cleared when ENABLE_DEBUG is set to 1, and validity checks are performed everywhere the STRIP variable is used.

Build with ENABLE_DEBUG=1 and inspect with a debug symbol viewing tool.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -455,6 +455,7 @@ endif
 
 ifeq ($(ENABLE_DEBUG),1)
 CXXFLAGS := -Og -DDEBUG $(filter-out $(OPT_LEVEL),$(CXXFLAGS))
+STRIP :=
 endif
 
 ifeq ($(ENABLE_ABC),1)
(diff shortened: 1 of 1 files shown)


widlarizer (approved):
The core of the issue is that `make install` strips installed yosys and libyosys.so, regardless of `ENABLE_DEBUG`. This seems wrong, and this PR is a valid solution to this

I see a potential problem with somebody setting a custom STRIP variable regardless of if they set ENABLE_DEBUG, but also, I'm willing to bet, so I don't mind.

Also, beware: `strip -S` creates binaries that `file` reports as `not stripped`... terrifying

RonxBulld:
> The core of the issue is that `make install` strips installed yosys and libyosys.so, regardless of `ENABLE_DEBUG`. This seems wrong, and this PR is a valid solution to this
> 
> I see a potential problem with somebody setting a custom STRIP variable regardless of if they set ENABLE_DEBUG, but also, I'm willing to bet, so I don't mind.
> 
> Also, beware: `strip -S` creates binaries that `file` reports as `not stripped`... terrifying

You summed it up more clearly than I could.
Yes, I also noticed the "not stripped" problem you mentioned. The file tool does not seem to think that ELF without DWARF is "stripped".
For the other question, if the user executes "make" and simply leaves STRIP empty in the hope that STRIP will not be executed, it will cause another problem: the shell tries to execute a command called "-S".

widlarizer:
I thought of an evil alternative, by the way: if STRIP is set to `true`, the `true` shell builtin will be called, which does nothing and exits with 0. So stripping may be disabled by setting `STRIP=true` - by the user, or automatically when `ENABLE_DEBUG=1`. This is a neat solution but very confusing as well

PR #5128, opened 2025-05-17
Add keyword-based inputs struct constructor into Rosette backend 

gussmith23:

Currently, the Rosette code emitted by Yosys requires the user to ensure they order their module inputs correctly when constructing the "inputs" struct to be passed to the Rosette function generated for the module. This makes it difficult to use the backend in an automated way, and may lead to silent failures -- e.g., if we provide the right number of ports but in the wrong order, it won't be clear why our module isn't behaving as expected. (As an aside: even more confusingly, it seems like the backend may currently reverse the order of the ports!)

This change adds an optional helper function (guarded by a flag) that can construct the inputs struct using a keyword-based function. Users can thus explicitly specify each port by name. This allows users to be more confident that they are constructing the input struct correctly.

A new Rosette/Racket function is (optionally) written out which constructs an input struct.

I would appreciate feedback on how to test. I see that the Python tests currently run the generated Racket files. Here are two options:
1. Add the new `-keyword-constructor` flag in the existing tests, which will not actually cause the keyword constructors to be used, but it will cause them to be generated. This will at least test that we do not have syntax errors.
2. Add another set of nearly-identical Python tests that instantiate the input structs via the helper function, rather than the original constructor. 

TODO
- [ ] Add tests (see above, need guidance from maintainers)

### DIFF
--- a/backends/functional/smtlib_rosette.cc
+++ b/backends/functional/smtlib_rosette.cc
@@ -195,12 +195,13 @@ struct SmtrModule {
 	Functional::IR ir;
 	SmtrScope scope;
 	std::string name;
-	
+	std::optional<std::string> input_kw_name;
+
 	SmtrStruct input_struct;
 	SmtrStruct output_struct;
 	SmtrStruct state_struct;
 
-	SmtrModule(Module *module)
+	SmtrModule(Module *module, bool keyword_constructor)
 		: ir(Functional::IR::from_module(module))
 		, scope()
 		, name(scope.unique_name(module->name))
@@ -209,6 +210,10 @@ struct SmtrModule {
 		, state_struct(scope.unique_name(module->name.str() + "_State"), scope)
 	{
 		scope.reserve(name + "_initial");
+		if (keyword_constructor) {
+			input_kw_name = scope.unique_name(module->name.str() + "_inputs_kw");
+			scope.reserve(*input_kw_name);
+		}
 		for (auto input : ir.inputs())
 			input_struct.insert(input->name, input->sort);
 		for (auto output : ir.outputs())
@@ -264,6 +269,22 @@ struct SmtrModule {
 		w.pop();
 	}
 
+	void write_input_kw(SExprWriter &w)
+	{
+		w.push();
+		w.open(list("define"));
+		w.open(list(name + "_inputs_kw"));
+		for (auto name : input_struct.get_field_names()) {
+			w << "#:" + name << name;
+		}
+		w.close();
+		w.open(list(input_struct.name));
+		for (auto name : input_struct.get_field_names()) {
+			w << name;
+		}
+		w.pop();
+	}
+
 	void write(std::ostream &out)
 	{    
 		SExprWriter w(out);
@@ -271,6 +292,10 @@ struct SmtrModule {
 		input_struct.write_definition(w);
 		output_struct.write_definition(w);
 		state_struct.write_definition(w);
+
+		if (input_kw_name) {
+			write_input_kw(w);
+		}
 
 		write_eval(w);
 		write_initial(w);
@@ -289,12 +314,16 @@ struct FunctionalSmtrBackend : public Backend {
 		log("\n");
 		log("    -provides\n");
 		log("        include 'provide' statement(s) for loading output as a module\n");
+		log("    -keyword-constructor\n");
+		log("        provide a function which can construct inputs using keywords\n");
+		log("        \n");
 		log("\n");
 	}
 
 	void execute(std::ostream *&f, std::string filename, std::vector<std::string> args, RTLIL::Design *design) override
 	{
 		auto provides = false;
+		auto keyword_constructor = false;
 
 		log_header(design, "Executing Functional Rosette Backend.\n");
 
@@ -303,6 +332,8 @@ struct FunctionalSmtrBackend : public Backend {
 		{
 			if (args[argidx] == "-provides")
 				provides = true;
+			else if (args[argidx] == "-keyword-constructor")
+				keyword_constructor = true;
 			else
 				break;
 		}
@@ -315,7 +346,7 @@ struct FunctionalSmtrBackend : public Backend {
 
 		for (auto module : design->selected_modules()) {
 			log("Processing module `%s`.\n", module->name.c_str());
-			SmtrModule smtr(module);
+			SmtrModule smtr(module, keyword_constructor);
 			smtr.write(*f);
 		}
 	}
--- a/backends/functional/smtlib_rosette.cc
+++ b/backends/functional/smtlib_rosette.cc
@@ -114,6 +114,13 @@ public:
 		size_t i = field_names.at(name);
 		return list(fields[i].accessor, std::move(record));
 	}
+	std::vector<std::string> get_field_names()
+	{
+		std::vector<std::string> names;
+		for (auto field : fields)
+			names.push_back(field.name);
+		return names;
+	}
 };
 
 std::string smt_const(RTLIL::Const const &c) {


gussmith23:
I am making one more change to this and then it will be ready for review. Converting to draft until then.

gussmith23:
Ready for review.

KrystalDelusion:
> As an aside: even more confusingly, it seems like the backend may currently reverse the order of the ports!

Do you have an example of this?  I remember they were backwards at one point so I had to reverse them, but they should be in the right order now:
```verilog
module \$fa (A, B, C, X, Y);
input [WIDTH-1:0] A, B, C;
output [WIDTH-1:0] X, Y;
```
```racket
(struct _$fa_Inputs (A B C) #:transparent
  ; A (bitvector 1)
  ; B (bitvector 1)
  ; C (bitvector 1)
)
(struct _$fa_Outputs (X Y) #:transparent
  ; X (bitvector 1)
  ; Y (bitvector 1)
)
```
If there is any rearranging of ports I would suspect it's in Yosys or the frontend rather than the Rosette backend.

> Similarly on the outputs side, it would be nice to have a way to access outputs via their name.

As far as I understand Rosette/Racket, this should be possible with (e.g.) `_$fa_Outputs-X`, similarly to how inputs are (internally) referenced by name `(let ((A (_$fa_Inputs-A inputs)))`.  IIRC that's why they're marked `#:transparent` for both input and output, and also why there are comments for each of the port names/expected types, though I totally agree that being able to construct the input struct from named values would be useful (and the current state is mostly a reflection of me being unfamiliar with Racket).

> `access_by_base_name` feels unnecessary, but `access` doesn't work for my purposes because I can't actually get the IdString of the field name.

I think you want to iterate over `auto output : ir.outputs()` rather than `auto name : output_struct.get_field_names()`, because that'll give you the list of base names.  But again, I'm not sure how necessary this is since you can already access outputs by name.

> I would appreciate feedback on how to test.

I don't think adding the flag to the current tests is the right move, since that invites the possibility that the output stops being valid *without* the flag (though I realise this is already true for the `-provides` flag).  I also don't think it's necessary to test the flag for the full suite of cells.  I would say it's sufficient to add one test (or potentially a couple tests) specifically for checking that instantiation via the helper structs works.

gussmith23:
> > As an aside: even more confusingly, it seems like the backend may currently reverse the order of the ports!
> 
> Do you have an example of this? I remember they were backwards at one point so I had to reverse them, but they should be in the right order now:
> 
> ```verilog
> module \$fa (A, B, C, X, Y);
> input [WIDTH-1:0] A, B, C;
> output [WIDTH-1:0] X, Y;
> ```
> 
> ```racket
> (struct _$fa_Inputs (A B C) #:transparent
>   ; A (bitvector 1)
>   ; B (bitvector 1)
>   ; C (bitvector 1)
> )
> (struct _$fa_Outputs (X Y) #:transparent
>   ; X (bitvector 1)
>   ; Y (bitvector 1)
> )
> ```
> 
> If there is any rearranging of ports I would suspect it's in Yosys or the frontend rather than the Rosette backend.

Example: https://github.com/gussmith23/lakeroad/blob/50226012951d66190ea0e36a67c75035943b9d31/modules_for_importing/xilinx_ultrascale_plus/DSP48E2.v

```sh
yosys -p "read_verilog -sv DSP48E2.v; prep; clk2fflogic; write_functional_rosette -provides -assoc-list-helpers"
```

This should produce a file with the ports reversed; that's true on my machine at least. If it's not true on yours...something extra weird is going on!


> > Similarly on the outputs side, it would be nice to have a way to access outputs via their name.
> 
> As far as I understand Rosette/Racket, this should be possible with (e.g.) `_$fa_Outputs-X`, similarly to how inputs are (internally) referenced by name `(let ((A (_$fa_Inputs-A inputs)))`. IIRC that's why they're marked `#:transparent` for both input and output, and also why there are comments for each of the port names/expected types, though I totally agree that being able to construct the input struct from named values would be useful (and the current state is mostly a reflection of me being unfamiliar with Racket).

You're right that the accessors `_$fa_Outputs-X` are generated. However, they're hard to use programmatically. Imagine we have a module we're trying to use from Rosette, but we don't actually know its name/its ports at the time of writing the code. It would be natural, then, to take the name and ports list as strings. If we have an output port name `o` as a string `"o"`, how do we actually use the accessor `mod_Outputs-o`? We have to generate the symbol for the function using quasiquoting and then `eval` it -- see https://docs.racket-lang.org/guide/eval.html. It's possible, but it's very worth avoiding, especially when there's easy ways around it.

As an aside, I really do feel like this should be easier in Racket. I'm shocked that there's not a default way to look up a field on a struct if you know the field's name as a string. It could be that I'm totally wrong about this, though; maybe it's easy!

> > `access_by_base_name` feels unnecessary, but `access` doesn't work for my purposes because I can't actually get the IdString of the field name.
> 
> I think you want to iterate over `auto output : ir.outputs()` rather than `auto name : output_struct.get_field_names()`, because that'll give you the list of base names. But again, I'm not sure how necessary this is since you can already access outputs by name.

Thanks! Fixed.

> > I would appreciate feedback on how to test.
> 
> I don't think adding the flag to the current tests is the right move, since that invites the possibility that the output stops being valid _without_ the flag (though I realise this is already true for the `-provides` flag). I also don't think it's necessary to test the flag for the full suite of cells. I would say it's sufficient to add one test (or potentially a couple tests) specifically for checking that instantiation via the helper structs works. The current pytest framework may or may not be the best place to add that, but be aware that the functional tests are disabled by default and only run on the private runner during CI because of the additional requirements (though looking at it seems that even then the smtlib and rosette tests are skipped, which seems unideal...).

Any concrete suggestions on where I should put the tests? Should I not write them in pytest? Should I add them as "functional tests"? Are those the same things as the pytest tests?

KrystalDelusion:
> ...
> This should produce a file with the ports reversed; that's true on my machine at least. If it's not true on yours...something extra weird is going on!

Huh.  It's not just backwards, it's alphabetised too, but only from the functional backends (and not just Rosette).

> However, they're hard to use programmatically. Imagine we have a module we're trying to use from Rosette, but we don't actually know its name/its ports at the time of writing the code.

Ah right, that is true, particularly with the renaming/escaping that happens when it creates a unique name.  In that case it does make sense to have an output converter as well.

> Any concrete suggestions on where I should put the tests? Should I not write them in pytest? Should I add them as "functional tests"? Are those the same things as the pytest tests?

The `tests/functional` directory is gated by the make flag `ENABLE_FUNCTIONAL_TESTS`.  Currently the `tests/functional/run-test.sh` script is just a single call to `pytest -v -m "not smt and not rkt" "$@"`, which runs pytest, but skips the smt and rkt (Rosette) tests.  I *think* what happened is that the smt and rkt tests were disabled because they have extra prerequisites that are not currently installed on the normal git runner, and would also break regular calls to `make test` without them.  The `ENABLE_FUNCTIONAL_TESTS` was then added later, to limit running them to just the private runner which we also use to run the verific tests, meaning breaking `make test` is no longer a concern so the `-m "not smt and not rkt"` should be safe to remove.

Where you write the tests depends on if you prefer bash or python for scripting; if you prefer bash then you should put them in `tests/functional/run-test.sh`, if instead you prefer python then you're welcome to use the `pytest` infrastructure that already exists in that directory.  In either case, any extra `.v` or `.il` files you need can go in the same directory.

KrystalDelusion:
`hierarchy; write_functional_rosette` gives A B C.
`prep; write_functional_rosette` gives C B A...  Looks like it's `opt_clean` that sorts them, but I'm unclear on why this affects the functional backends but not the others.

gussmith23:
@KrystalDelusion I have updated this PR and it's ready for another review. I have also added testing by hooking into the existing tests, adding an arg which controls whether this feature is used.

gussmith23:
Failures from the last CI run look like this:
```
2025-07-01T22:31:26.7349954Z FAILED test_functional.py::test_rkt[not-6-12-S-True] - FileNotFoundError: [Errno 2] No such file or directory: 'racket'
2025-07-01T22:31:26.7350249Z FAILED test_functional.py::test_rkt[not-6-12-S-False] - FileNotFoundError: [Errno 2] No such file or directory: 'racket'
2025-07-01T22:31:26.7350541Z FAILED test_functional.py::test_rkt[not-6-12-U-True] - FileNotFoundError: [Errno 2] No such file or directory: 'racket'
2025-07-01T22:31:26.7350888Z FAILED test_functional.py::test_rkt[not-6-12-U-False] - FileNotFoundError: [Errno 2] No such file or directory: 'racket'
2025-07-01T22:31:26.7351214Z FAILED test_functional.py::test_rkt[not-32-16-S-True] - FileNotFoundError: [Errno 2] No such file or directory: 'racket'
...
```

@KrystalDelusion or whoever it may concern: it seems like the functional tests are getting run under the Verific tests -- is that expected? It seems like one of two options:
1. Verific testing workflow always ran the functional tests, and my changes broke them
2. Verific testing workflow didn't previously run the functional tests, and my changes somehow made them start running the functional tests

2 seems more likely to me, namely because the failure is so basic -- ie that Racket isn't installed in the environment. That implies to me that the Verific testing workflow isn't meant to run the functional tests.

KrystalDelusion:
I also believe it is the latter, and mentioned earlier that the smt and rkt tests were being skipped and we should enable them so uhh, that's probably on me for not checking if the prerequisites were installed 😛

gussmith23:
what would you like me to do here?

KrystalDelusion:
> what would you like me to do here?

Hopefully nothing, runner has just been updated and the test is re-running

tests/functional/simulate_rosette.py
```diff
@@ -0,0 +1 @@
+"""Python utilities for simulating Rosette code."""
```
KrystalDelusion: Unless I'm missing it, this file isn't used anywhere (and has nothing in it).

backends/functional/smtlib_rosette.cc
```diff
@@ -265,6 +310,12 @@ struct SmtrModule {
 		output_struct.write_definition(w);
 		state_struct.write_definition(w);
 
+		if (input_helper_name) {
+			if (!output_helper_name)
+				log_error("if keyword helpers are enabled, both input and output helper names are expected");
```
KrystalDelusion: Is it possible for this error to occur?

backends/functional/smtlib_rosette.cc
```diff
@@ -257,6 +265,43 @@ struct SmtrModule {
...
+			w.pop();
+			w.open(list("if", "assoc-result"));
+			w << list("cdr", "assoc-result");
+			w.open(list("begin"));
+			w << list("fprintf", list("current-error-port"), "\"%s not found in inputs\"");
+			w << "'not-found";
+			w.pop();
+		}
+		w.pop();
+		// Output struct keyword-based destructor.
```
KrystalDelusion: What makes this a destructor?

KrystalDelusion:
@gussmith23 Can you take a look at the comments I left when you have a chance?  Tests are running fine with parallelism enabled on CI now (shown on a [test merge](https://github.com/YosysHQ/yosys/actions/runs/16875222636) because the job can't be re-run anymore)

backends/functional/smtlib_rosette.cc
gussmith23: This is destructuring in the [Racket/functional sense](https://docs.racket-lang.org/destruct/index.html), not in the C++ sense. Instead of meaning "deleter" (which is what it means in C++), destructor here literally means de-`struct`-or, i.e. unwrapping a `struct` into some other object.

gussmith23: I can rephrase if you think it's helpful!

backends/functional/smtlib_rosette.cc
gussmith23: Not due to user input, no. This was more to prevent regression. I decided to just simplify this, though, and add a check directly in `write_assoc_list_helpers` to make sure the fields are set.

gussmith23:
@KrystalDelusion this should be ready for a final review aside from the following point:

In an earlier review, you had mentioned that I should add some other tests for edge cases/incorrect configuration. However, this modification to the Rosette backend only adds a single additional configuration option: the presence of a boolean flag indicating whether or not to use the assoc-list-based helpers. (I.e. things like the names of the helper functions are not configurable, though a previous version of the code probably made it look like they were). Given that that's the case, I'm not sure what other edge cases there are to test -- please let me know if you think of any when reviewing.

Otherwise, here are the changes since your last review:
* Removed a bunch of code formatting noise
* Merged with main
* Added a boolean that explicitly tracks whether the caller requested assoc list helpers (rather than using the presence/absence of the input helper name parameter as the indicator of whether the user requested the helpers)
* Misc change to the run-tests.sh script to make it runnable from anywhere

backends/functional/smtlib_rosette.cc
KrystalDelusion: Oh!  Huh.  Destructuring makes the difference apparent while still keeping the destruct part, maybe just "Output struct keyword-based destructuring"?

backends/functional/smtlib_rosette.cc
KrystalDelusion: I think `log_error` is for errors in the design or something wrong that the user did (or can fix).  For developer sanity checks I think it should be `log_assert` or `log_abort`, though I'm not sure that's actually specified anywhere and is more just Vibes.  Internally `log_assert` and `log_abort` still call `log_error`, but they include the file/line which feels more explicit about "this is a problem with the code, not your fault"

KrystalDelusion:
Yeah I'm not entirely sure what past Krystine was looking for with extra tests, present Krystine thinks this is fine (pending resolution of those last two issues)

gussmith23:
@KrystalDelusion fixed!

PR #5135, opened 2025-05-21
ast, read_verilog: ownership in AST, use C++ styles for parser and lexer

widlarizer:
The purpose of this PR is to clarify resource ownership (responsibility for construction and deletion) in the AST by distinguishing between owning pointers (`std::unique_ptr<AstNode>`) and non-owning pointers (`AstNode*`).

## Core idea

`struct AstNode` now holds these members:

```cpp
std::vector<std::unique_ptr<AstNode>> children;
std::map<RTLIL::IdString, std::unique_ptr<AstNode>> attributes;
```

There are no more `new AstNode` and corresponding `delete` statements present in the codebase. `std::unique_ptr<AstNode>` falling out of scope triggers `~AstNode`, destroying its children, too. Assigning `nullptr` to it behaves the same way. This should eliminate `AstNode` allocations being reported as leaking on `yosys` termination by `valgrind`.

## Damage

This implied returning `std::unique_ptr<AstNode>` from parse rules. Prior to this PR, the return data type for parse rules is auto-generated with the bison `%union` directive. The auto-generated code handling it relies on each union member type having an `T& operator=(T&);`, but `std::unique_ptr<...>` [deletes](https://en.cppreference.com/w/cpp/memory/unique_ptr/operator%3D) its own for good reason.

That's why I refactored the parser and lexer with their `C++` modes causing a greater code change.

Locations in AST and parser values now use the same custom data type with `std::shared_ptr<std::string> filename;`. This reduces filename copying in the frontend greatly. I'm seeing 6-20% memory usage improvements in `read_verilog` as a result, at a cost of a `1%` performance regression, which may be avoidable.

Superior location tracking without global state will allow proper column ranges in errors in a simple followup PR.

~~Somehow, syntax errors now don't quote characters, so you get `ERROR: syntax error, unexpected ;` instead of `ERROR: syntax error, unexpected ';'`.~~ fixed

## Testing

This PR has a high risk of creating new yosys segfaults in edge cases not hit by our test suite. More extensive testing will be needed. However, I would argue that crashes are more desirable than a mix of "everything somehow works out" and straight up silent data corruption.

## TODOs

- [x] Source locations are temporarily completely broken. This breaks tests
- [x] Extend source location diagnostics testing on main prior to merging this. See #5183
- [x] Can global variables be eliminated from the parser by moving them into the generated parser class?
- [x] Ensure there are no Makefile dependency errors
- [ ] ~~Extend ownership in ast and parser to attribute lists and strings. Containers already basically *are* owned pointers, but maybe should be still pointed to with `std::unique_ptr` to avoid accidental copies~~ strings done, attribute lists low prio followup
- [x] Play around with memory and address sanitizers to see if we're cutting down on the leaks - address sanitizer clean, memory sanitizer complains
- [ ] ~~Try reduce the perf regression likely caused by mandatory source location construction for all nodes~~
- [x] Fix the internal docs for the verilog frontend
- [x] ~~Workaround for older bison versions due to `parse.error` ?~~ Declare correct required bison version


widlarizer:
@whitequark I have a problem with flex headers not being available in the WASI build. Since the build runs on our regular linux CI image, it should have those installed if the linux bulid passes. I don't have experience with the WASI SDK. Does it not see system installed headers? Is there some reason even zlib and readline are disabled? Any suggestions appreciated. The run in question: https://github.com/YosysHQ/yosys/actions/runs/16206897063/job/45759169322

whitequark:
> Does it not see system installed headers?

No. Building for WASI uses a cross-compilation environment. Everything that you touch must be built for your target (WASI in this case) and installed in a cross-prefix. Doing things otherwise (particularly in C/C++) would be misguided: even if sometimes it works, that would be only by happy coincidence.

Readline will not work on WASI (for several reasons, including "WASI has no notion of escape codes" and "readline requires a working setjmp/longjmp which I currently cannot ship"). Zlib would not be too difficult to build but typically YoWASP builds are used with smaller netlists where it's just not really necessary. (Nobody asked for it in the entire time I maintained YoWASP).

Don't we have a pair of flex lexers in-tree already? I haven't checked the code/logs but I don't understand why flex headers would be necessary.

whitequark:
OK so it looks like it needs this header whenever you use the C++ mode. There are two options:
- Do the whole configure-make-makeinstall dance for flex after downloading the flex sources first.
- Copy this one header out of the flex installation and update it every 4 years or whenever they add a new field.

I have no real preference.

KrystalDelusion:
As a proof of concept, [WASI and VS builds can pass](https://github.com/YosysHQ/yosys/actions/runs/16308465218) if the `FlexLexer.h` file is given to them.  In practice, I think telling the user to `ln -s` a file if they want to build for WASI seems ill-advised, so hopefully there's a bit more work here before it gets merged.

widlarizer:
Doing things this way is the least likely to break, the header doesn't depend on anything and has no arch specific ifdefs. This way you will always get the header of the same version as the actual lexer generator expects. I can't guarantee that otherwise

I thought WASI builds get distributed from CI, but also, we are already effectively advising people to just figure out how they have to install dependencies on their system, the README example for some specific ubuntu distro is just illustrative

KrystalDelusion:
> This way you will always get the header of the same version as the actual lexer generator expects. I can't guarantee that otherwise

That's a good point actually.

> I thought WASI builds get distributed from CI

Uhhh, pass.  I think the WASI build on CI is mostly for checking it compiles, and it doesn't actually upload anything.  I think the main user is YoWASP, though *that* is probably predominantly distributed from CI builds.

whitequark:
> Uhhh, pass. I think the WASI build on CI is mostly for checking it compiles, and it doesn't actually upload anything. I think the main user is YoWASP, though _that_ is probably predominantly distributed from CI builds.

This is correct, although there are definitely non-YoWASP users (since the Emscripten build was deprecated and I know some people use it).

I'm not quite sure what you mean by "distributed from CI builds" specifically. Are you referring to the fact that people are unlikely to build their own? If so, yes, that's most likely the case (if people are building YoWASP binaries on their own they don't tell me).

whitequark:
I've added a flex build/install step to the CI workflow since it was faster to do than to keep discussing this.

whitequark:
Also, it looks like the MSVC build is failing too for the same reason, so this problem isn't exclusive to WASI.

whitequark:
The build passes locally with [act](https://github.com/nektos/act) so I'm going to assume it will pass on GitHub too.

edit: yep, [works](https://github.com/YosysHQ/yosys/actions/runs/16344844377/job/46176003852)

KrystalDelusion:
I added the commits to get the VS build to pass; I also double checked the header file includes the necessary license notice to redistribute it (it does).  This doesn't resolve the problem with the wheels building on Windows, but they are already disabled because of an issue with FFI, so it's not a blocker.

.github/workflows/wheels.yml
```diff
@@ -79,7 +79,7 @@ jobs:
       - if: ${{ matrix.os.family == 'windows' }}
         name: "[Windows] Flex/Bison"
         run: |
-          choco install winflexbison3
+          choco install winflexbison3 --source https://github.com/lexxmark/winflexbison/releases/download/v2.5.25/win_flex_bison-2.5.25.zip
```
KrystalDelusion: The `--source` here doesn't work for me (is that why you had the modified `.nupkg`?), I would probably get rid of this.

README.md
```diff
@@ -74,7 +74,7 @@ recommended) and some standard tools such as GNU Flex, GNU Bison, and GNU Make.
 TCL, readline and libffi are optional (see ``ENABLE_*`` settings in Makefile).
 Xdot (graphviz) is used by the ``show`` command in yosys to display schematics.
 
-For example on Ubuntu Linux 16.04 LTS the following commands will install all
+For example on Ubuntu Linux 22.04 LTS the following commands will install all
```
KrystalDelusion, line 83: The prerequisites are missing `libfl-dev`, as are the ones in the `getting_started/installation.rst`.

kernel/rtlil.cc
```diff
@@ -5704,16 +5700,11 @@ static void sigspec_parse_split(std::vector<std::string> &tokens, const std::str
...
-{
-	return 0;
-}
-
 bool RTLIL::SigSpec::parse(RTLIL::SigSpec &sig, RTLIL::Module *module, std::string str)
 {
 	cover("kernel.rtlil.sigspec.parse");
 
-	AST::current_filename = "input";
+	// AST::current_filename = "input";
```
KrystalDelusion: nit
```suggestion
```

passes/memory/memlib.h
```diff
@@ -100,7 +100,7 @@ enum class WrTransKind {
 
 struct WrTransDef {
 	WrTransTargetKind target_kind;
-	int target_group;
+	int target_group = 0;
```
KrystalDelusion, line 103: Is this related or just incidental tidy up?

frontends/verilog/preproc.h
```diff
@@ -71,7 +76,9 @@ frontend_verilog_preproc(std::istream                 &f,
                          std::string                   filename,
                          const define_map_t           &pre_defines,
                          define_map_t                 &global_defines_cache,
-                         const std::list<std::string> &include_dirs);
+                         const std::list<std::string> &include_dirs,
+						 VERILOG_FRONTEND::ParseState &parse_state,
```
KrystalDelusion: Inconsistent indentation

frontends/verilog/preproc.cc
```diff
@@ -749,7 +750,9 @@ frontend_verilog_preproc(std::istream                 &f,
                          std::string                   filename,
                          const define_map_t           &pre_defines,
                          define_map_t                 &global_defines_cache,
-                         const std::list<std::string> &include_dirs)
+                         const std::list<std::string> &include_dirs,
+						 ParseState                   &parse_state,
```
KrystalDelusion: Inconsistent indentation

frontends/verilog/verilog_lexer.l
```diff
@@ -41,78 +48,113 @@
...
 #define NON_KEYWORD() \
-	yylval->string = new std::string(std::string("\\") + yytext); \
-	return TOK_ID;
+	string_t val = std::make_unique<std::string>(std::string("\\") + YYText()); \
+	return parser::make_TOK_ID(std::move(val), out_loc);
 
-#define YY_INPUT(buf,result,max_size) \
-	result = readsome(*VERILOG_FRONTEND::lexin, buf, max_size)
+// #define YY_INPUT(buf,result,max_size) \
+// 	result = readsome(*extra->lexin, buf, max_size)
```
KrystalDelusion: nit
```suggestion
```

frontends/verilog/verilog_lexer.l
```diff
@@ -41,78 +48,113 @@
...
-	std::vector<std::string> fn_stack;
-	std::vector<int> ln_stack;
-	YYLTYPE real_location;
-	YYLTYPE old_location;
-}
+#undef YY_DECL
+#define YY_DECL parser::symbol_type VerilogLexer::nextToken()
+
+#undef yyterminate
+#define yyterminate() terminate()
```
KrystalDelusion: I'm not sure exactly how the warning relates to this line, but it should probably be checked.

.github/workflows/wheels.yml
widlarizer: Right, forgot to remove this when we found the code is disabled

passes/memory/memlib.h
widlarizer: Incidental fix for an uninitialized read

It was triggering the memory sanitizer, which I wanted to use to check this PR was okay. However, the memory sanitizer trips on too much stuff for me to handle at the moment, so it's not actually memory sanitizer clean. The main branch also has some problems across the board

KrystalDelusion: iirc I tried the memory sanitizer but was running into false positives from boost libraries because they weren't compiled with the sanitizer, and I gave up.  It would be nice to get it running but it seems like a lot more work than asan/ubsan

frontends/ast/genrtlil.cc
```diff
@@ -2186,30 +2179,28 @@ RTLIL::SigSpec AstNode::genRTLIL(int width_hint, bool sign_hint)
 
 	// use ProcessGenerator for always blocks
 	case AST_ALWAYS: {
-			AstNode *always = this->clone();
-			ProcessGenerator generator(always);
+			auto always = this->clone();
+			ProcessGenerator generator(std::move(always));
```
KrystalDelusion: The temporary assignment here feels unnecessary, wouldn't it be easier to pass `generator(this->clone())`? And again below

frontends/ast/simplify.cc
```diff
@@ -885,19 +886,28 @@ static void check_auto_nosync(AstNode *node)
...
+
+void AstNode::null_check()
+{
+	for (auto& child : children) {
+		// if (!child)
+		// 	VALGRIND_PRINTF_BACKTRACE("null child");
+		log_assert((bool) child);
+		child->null_check();
+	}
 }
```
KrystalDelusion: It doesn't look like this is used anywhere that isn't commented out

frontends/ast/simplify.cc
```diff
@@ -5193,34 +5170,37 @@ bool AstNode::mem2reg_as_needed_pass2(pool<AstNode*> &mem2reg_set, AstNode *mod,
...
+	log_assert(id2ast == nullptr || mem2reg_set.count(id2ast) == 0);
+
+	std::vector<AstNode*> children_list;
+	for (auto& child : children)
+		children_list.push_back(child.get());
 
-	auto children_list = children;
 	for (size_t i = 0; i < children_list.size(); i++)
 		if (children_list[i]->mem2reg_as_needed_pass2(mem2reg_set, mod, block, async_block))
 			did_something = true;
```
KrystalDelusion, line 5245: Is there a reason these are two for loops instead of one?

frontends/ast/ast.cc
```diff
@@ -1070,23 +1085,23 @@ RTLIL::Const AstNode::realAsConst(int width)
...
+	return stringf("%s:%d.%d-%d.%d", location.begin.filename->c_str(), location.begin.line, location.begin.column, location.end.line, location.end.column);
 }
 
 void AST::set_src_attr(RTLIL::AttrObject *obj, const AstNode *ast)
 {
 	obj->attributes[ID::src] = ast->loc_string();
 }
 
-static bool param_has_no_default(const AstNode *param) {
+static bool param_has_no_default(const std::unique_ptr<AstNode> &param) {
```
KrystalDelusion: Is there a reason this changed from `const AstNode*`?  It seems inconsistent with other code

frontends/ast/ast.cc
```diff
@@ -192,24 +190,24 @@ bool AstNode::get_bool_attribute(RTLIL::IdString id)
...
 
-	AstNode *attr = attributes.at(id);
-	if (attr->type != AST_CONSTANT)
-		attr->input_error("Attribute `%s' with non-constant value!\n", id.c_str());
+	AstNode& attr = *(attributes.at(id));
+	if (attr.type != AST_CONSTANT)
+		attr.input_error("Attribute `%s' with non-constant value!\n", id.c_str());
 
-	return attr->integer != 0;
+	return attr.integer != 0;
```
KrystalDelusion: Couldn't you have done `auto& attr = attributes.at(id);` instead and kept the `attr->`?

frontends/verilog/verilog_lexer.h
```diff
@@ -0,0 +1,48 @@
...
+
+YOSYS_NAMESPACE_BEGIN
+
+namespace VERILOG_FRONTEND {
+	using parser = frontend_verilog_yy::parser;
+	class VerilogLexer : public frontend_verilog_yyFlexLexer {
+		ParseState* extra;
+		ParseMode* mode;
+	public:
+		parser::location_type out_loc; // TODO private?
```
KrystalDelusion:
```suggestion
		parser::location_type out_loc;
	public:
```
Seems to compile fine

frontends/ast/simplify.cc
widlarizer: I hope it won't come up again but it might at which point it comes in handy. That's just the level of sketch of the AST layer we live with, but it's trivial to rewrite to taste when needed so I'm removing it 

image

frontends/ast/simplify.cc
widlarizer: We need to save these pointers before we actually run mem2reg_as_needed_pass2, because something changes about the relations between these nodes. We're getting a snapshot of the child list on purpose. The mem2reg.v test does fail if we don't

widlarizer:
This rebase fails because something about the import statement is broken

KrystalDelusion (approved):

KrystalDelusion:
@widlarizer somewhere along the way the `log_maybe_loc_(error|warn)` functions stopped accounting for the "maybe."  i.e. it was at one point
https://github.com/YosysHQ/yosys/blob/c8e0ac0c616732aac82f02da01413154c7bba843/frontends/verilog/const2ast.cc#L47-L65

But in the latest versions it just dereferences `*loc.begin.filename`, causing a segfault if there is no file source, such as below when parsing the option input to a constant
```
read_rtlil << EOT
module \top
  wire width 4 input 1 \addr
  wire width 8 output 2 \out
  connect \out 8'x
end
EOT
sat -prove out 8'01
```
(found via #6009)

Given that the `Location` and `Position` classes both have `to_string()` methods on them which do correctly check that the filename exists before using it, is it worth having an alternative to `log_file_error()` which takes a `Location` (or `Position`), or a `log_error_prefixed()` which can be used instead.  Or just something like
```cpp
void ConstParser::log_maybe_loc_error(std::string msg) {
	if (loc.begin.filename && !loc.begin.filename->empty())
		log_file_error(*loc.begin.filename, loc.begin.line, "%s", msg);
	else {
		log_error("%s", msg);
	}
}
```

PR #5137, opened 2025-05-21
Allow reading of gzipped files when not in NDEBUG

mikesinouye:


widlarizer:
Nice catch, sorry about that, I'll check if we're doing anything silly like this elsewhere

widlarizer (approved):

mikesinouye:
Great, thanks @widlarizer!

PR #5141, opened 2025-05-23
Accept (and ignore) SystemVerilog unique/priority if.

garytwong:
Add support to the `read_verilog -sv` parser to validate the `unique`, `unique0`, and `priority` keywords in contexts where they're legal according to 1800-2012, section 12.4.2.

This affects _only_ the grammar accepted; the behaviour of conditionals is _not_ changed.  (But accepting this syntax will provide scope for possible optimisations as future work; see below.)

Three test cases ("unique_if", "unique_if_else", and "unique_if_else_begin") verify that the keywords are accepted where legal, and rejected where illegal, as described in the final paragraph of 12.4.2.

Two reasons: to improve compatibility with SystemVerilog 1800-2012 (by accepting `unique if` and similar constructions permitted by the spec, which are currently rejected as syntax errors); and to allow future optimisations of conditionals (see below).

This is achieved by modifying the grammar to allow the `unique`/`unique0`/`priority` keywords before an `if`.  (Similar to the existing support when applied before `case`.)  Please note that the keywords are parsed, but *not yet* converted to attributes in the `AstNode`: this is deliberate, to make as small a change as possible with no semantic effects.  The parser does introduce a new `promoted_if` attribute on some `AstNode`s it builds, to distinguish them from the construction that could have been generated by an equivalent `case` statement: this is necessary to correctly reject (e.g.) the illegal `... else unique if ...` syntax.

`make test`; verify that the three newly added tests all succeed (correctly accepting the two valid snippets and observing the expected error for the invalid one); and that all other regression tests are unaffected.

_NB:_
If this PR and the general approach is acceptable, I'll plan to add the corresponding _semantic_ support by applying `full_case` and `parallel_case` attributes where appropriate, similarly to the current `case` implementation.  The approach I intend to take is basically flattening a series of `...else if...` branches into a single `AST_CASE/AST_COND.../AST_DEFAULT` construction, rather than the nested `AST_CASEs` currently generated.


KrystalDelusion (changes requested):
The logic here seems sound, even without the associated optimisation included.  I can't seem to find any tests that check the semantics of the `full_case` and `parallel_case` attributes, only that they are accepted by the frontend.  Would you be able to add something for this?

I will also note that the `verific` frontend ignores the `unique` and `priority` keywords on if and case (not that that means we can't or shouldn't add the support to `read_verilog`).

garytwong:
Thanks @KrystalDelusion!  I agree, verifying the resulting semantics would definitely make stronger test cases.

How do you feel about the three new tests added in commit 73e45d29?  They don't really do much at the moment (other than exercising the parser, and then succeeding because `priority`/`unique` is currently a no-op when applied to `if`), but will become more meaningful once the `priority`/`unique` `if` implementation I propose to add next is done.

(Or were you suggesting tests further into Yosys internals, checking the behaviour of `full_case` and `parallel_case` where they're generated by the current code?  I could add similar tests applying `case` statements instead of `if`, since `unique`/`priority` logic is already implemented there.  If that's what you mean, let me know and I'll add those tests as well or instead.)

KrystalDelusion (approved):

PR #4656, opened 2024-10-11
Prepend Verilog globals to module AST

jmi2k:
_What are the reasons/motivation for this change?_

Fixes #4653. Further AST and RTLIL stages seem to be order-sensitive, and appending globals to the module children list did not work.

_Explain how this is achieved._

Early during `AST::process()`, Verilog globals are prepended to the children list instead.

_If applicable, please suggest to reviewers how they can test the change._

#4653 contains a test case that triggers an assertion, but works fine after applying this change.

Issue #4653, opened 2024-10-10
SystemVerilog packed structs not supported


jmi2k:
### Version

Yosys 0.46+11 (git sha1 0200a7680, clang++ 14.0.0-1ubuntu1.1 -fPIC -O3)

### On which OS did this happen?

Linux

### Reproduction Steps

Minimal reproducible example `SoC.sv`:
```sv
typedef struct packed {
	logic y;
	logic x;
} Vec_2_B;

module Gpu (output logic mixed_sync);
endmodule

(* top *)
module Top(output Vec_2_B vga_sync);

	Gpu gpu(vga_sync.x);

endmodule
```

Build command:
```
yosys -p 'synth_ecp5 -abc2' SoC.sv
```

### Expected Behavior

The SV file is accepted without any error messages.

### Actual Behavior

An assertion fails:

```
Generating RTLIL representation for module `\Top'.
ERROR: Assert `!source_offset && !id2ast->range_swapped' failed in frontends/ast/genrtlil.cc:1604.
```

jmi2k:
Found an even simpler example, which indicates the issue manifests when using struct fields anywhere:

```sv
typedef struct packed {
	logic y;
	logic x;
} Vec_2_B;

(* top *)
module Top;

	Vec_2_B two_dee;
	wire foo = two_dee.x;

endmodule
```

jmi2k:
Another hint: this other example seems to work fine.

```sv
(* top *)
module Top;

	typedef struct packed {
		logic y;
		logic x;
	} Vec_2_B;

	Vec_2_B two_dee;
	wire foo = two_dee.x;

endmodule
```

jmi2k:
More hints: dumping the AST for both cases reveals only a single meaningful difference: the placement of the `AST_TYPEDEF` node inside the `AST_MODULE`. Note how, when the `typedef` is inside, it appears in the correct location (before everything else), but when it's outside it gets appended to the module contents.

`typedef` outside module:

```
Dumping AST before simplification:
    AST_MODULE <rtl/SoC.sv:7.1-12.10> [0x59dc00d2ad00] str='\Top'
      ATTR \top:
        AST_CONSTANT <rtl/SoC.sv:0.0-0.0> [0x59dc00d2ab90] bits='00000000000000000000000000000001'(32) range=[31:0] int=1 in_param
      AST_WIRE <rtl/SoC.sv:9.10-9.17> [0x59dc00d2b0c0] str='\two_dee' logic
[...]

PR #5143, opened 2025-05-26
SystemVerilog: Fix typedef struct in global space

KrystalDelusion:

When the `typedef struct` is inside the module, it is before the AST node that uses that type and so the custom type is fully resolved.

When the `typedef struct` is inside a package, the package is resolved before the module, so it is again fully resolved.

In the case from #4653, the `typedef struct` is in the global space, and ends up appended to the module so isn't fully resolved until _after_ the nodes that reference it are resolved, so indexing into it ends up with a range of `[-1:0]`, causing it to throw an assertion error.

By marking `struct`s with `is_custom_type`, the code will fully resolve the `typedef struct` before any nodes that reference it, thereby avoiding the assertion error and fixing #4653.

`tests/svtypes/typedef_struct_global.ys` should fail on main with an assertion error and pass with this fix.

### DIFF
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -1433,6 +1433,7 @@ bool AstNode::simplify(bool const_fold, int stage, int width_hint, bool sign_hin
 				current_ast_mod->children.push_back(wnode);
 			}
 			basic_prep = true;
+			is_custom_type = false;
 		}
 		break;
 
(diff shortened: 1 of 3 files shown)


zachjs (approved):

jmi2k:
Thanks a lot for this PR! I'll close #4656

PR #5148, opened 2025-05-26
Fix convertible_to_int handling of 32 bit unsigned ints with MSB set.

### LINKED ISSUE #5145: Yosys incorrectly optimizes single-bit >> wide shift into {a[n], 1'b0} instead of zeroing result
### Version

Yosys 0.53 (git sha1 53c22ab7c0ced80861c7536c5dae682c30fb5834, g++ 9.4.0-1ubuntu1~20.04.3 -fPIC -O3)

Linux

### Reproduction Steps

test.v:
module top (input wire [7:0] a, output wire [1:0] y);
    assign y = (a[2:2] >> 32'hffffffff);  
endmodule

yosys:
read_verilog test.v; proc; opt; synth; write_verilog;  syn_test.v;

### Expected Behavior

According to IEEE 1364-2005 Verilog standard:

If the shift amount is greater than or equal to the bit width of the value being shifted, the result shall be zero.

Here, a[2:2] is 1-bit wide, and the shift amount 32'hffffffff = 4294967295, which far exceeds the width. The expected result should be 2'b00;

### Actual Behavior

When performing a right-shift operation like a[2:2] >> 32'hffffffff (i.e., shifting a 1-bit value by an extremely large constant), Yosys synthesizes it into {a[2], 1'b0} rather than constant-zero, which is inconsistent with expected Verilog semantics.

/* Generated by Yosys 0.53 (git sha1 53c22ab7c0ced80861c7536c5dae682c30fb5834, g++ 9.4.0-1ubuntu1~20.04.3 -fPIC -O3) */

(* src = "rtl.v:1.1-3.10" *)
module top(a, y);
  (* src = "rtl.v:1.30-1.31" *)
  input [7:0] a;
  wire [7:0] a;
  (* src = "rtl.v:1.51-1.52" *)
  output [1:0] y;
  wire [1:0] y;
  assign y = { a[2], 1'h0 };
endmodule

georgerennie:

Fixes #5145 that was due to `convertible_to_int` not rejecting 32 bit unsigned ints with the MSB set. This slipped in in #5101 because I made the mistake of assuming that `as_int_compress` might be guarding its conversions correctly and just copied them, but it was not... The testcases I wrote then happened to pass spuriously because the widths were too large to show the buggy behaviour.

Add separate handling for the signed and unsigned cases. Also while I'm at it I update `as_int_compress` to just use `try_as_int`. There's no evidence anyone even uses this method but we might as well try to make it correct.

Copied some of the `opt_expr` testcases across to test for the particular overflow value that triggers this.

### DIFF
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -455,18 +455,7 @@ void RTLIL::Const::compress(bool is_signed)
 
 std::optional<int> RTLIL::Const::as_int_compress(bool is_signed) const
 {
-	auto size = get_min_size(is_signed);
-	if(size == 0 || size > 32)
-		return std::nullopt;
-
-	int32_t ret = 0;
-	for (auto i = 0; i < size && i < 32; i++)
-		if ((*this)[i] == State::S1)
-			ret |= 1 << i;
-	if (is_signed && (*this)[size-1] == State::S1)
-		for (auto i = size; i < 32; i++)
-			ret |= 1 << i;
-	return ret;
+	return try_as_int(is_signed);
 }
 
 std::string RTLIL::Const::as_string(const char* any) const


sdjasj:
I had noticed this issue before, but when I looked at the https://github.com/YosysHQ/yosys/pull/5101 I saw these comment lines in the modified code: https://github.com/YosysHQ/yosys/pull/5101/files#diff-a2f1fc0d8cf7d37b3202798fc55acfe317a1c83ee7183308861d4642dac440ddR1312-R1314. So I assumed that the abnormal behavior caused by `32'hffff_ffff` was expected...

georgerennie:
That comment still holds, such a bug would be triggered if the A input (the value being shifted) was on the order of 2^32 (or i guess 2^31?) bits wide, not with a value in that range (and thus a width of only 32 bits). This is very much my mistake for not double checking the fix with the actual testcase you had provided, rather than the modified version I created.

PR #5150, opened 2025-05-29
write_aiger: Add no-sort option

KrystalDelusion:

```
read_verilog << EOF
module top (input X, C, B, output A, Z, Y);

assign A = C;
assign Z = X;
assign Y = B & C;

endmodule
EOF
prep
aigmap
write_aiger -ascii -symbols
```
gives the following:
```
aag 4 3 0 3 1
2
4
6
4
8
6
8 4 2
i0 B
i1 C
i2 X
o0 A
o1 Y
o2 Z
c
```

As with #5133, the input and output ports are alphabetically sorted, rather than in the same order as the input design.

Add `-no-sort` option, which prevents sorting input/output bits.  Also use the `module->ports` to get inputs/outputs in order, separating the ports from the rest of the wires.

Test above script, adding `-no-sort` with the changes from this PR and get the following instead:
```
aag 4 3 0 3 1
2
4
6
4
2
8
8 6 4
i0 X
i1 C
i2 B
o0 A
o1 Z
o2 Y
c
```

### DIFF
--- a/backends/aiger/aiger.cc
+++ b/backends/aiger/aiger.cc
@@ -153,6 +153,9 @@ struct AigerWriter
 				sigmap.add(wire);
 
 		// handle ports
+		// provided the input_bits and output_bits don't get sorted they
+		// will be returned in reverse order, so add them in reverse to
+		// match
 		for (auto riter = module->ports.rbegin(); riter != module->ports.rend(); ++riter) {
 			auto *wire = module->wire(*riter);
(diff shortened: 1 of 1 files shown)


KrystalDelusion:
Note that I went with the solution of adding an option to prevent sorting rather than changing the default behaviour, since that has been the behaviour for the 8.5 years since the pass was added, so somebody might be expecting it to behave that way.

PR #5153, opened 2025-05-30
docs: several small documentation fixes.

garytwong:
Slight fixes for the manual.

Minor edits to several reST sources.

`make docs` and verify the built manual fixes errors; or just by inspection of the commit.


KrystalDelusion (approved):

docs/source/using_yosys/synthesis/abc.rst
```diff
@@ -176,5 +176,5 @@ implemented as whiteboxes too.
 Boxes are arguably the biggest advantage that ABC9 has over ABC: by being aware
 of carry chains and DSPs, it avoids optimising for a path that isn't the actual
 critical path, while the generally-longer paths result in ABC9 being able to
-reduce design area by mapping other logic to larger-but-slower cells.
+reduce design area by mapping other logic to smaller-but-slower cells.
```
Ravenslofty, line 179: Since I was the original author of this bit of text, I should mention this change is wrong; I really *did* mean larger cells.

If you look at the timings for, say, [iCE40HX](https://github.com/YosysHQ/yosys/blob/main/techlibs/ice40/cells_sim.v#L191-L198), you can see that I3 is the fastest input, followed by I2, and so on; so, to truly minimise delay, one wants to map logic to smaller LUT sizes where possible, but these have lower logic density than larger LUT sizes. Hence, one reduces design area by mapping non-critical logic to larger-but-slower LUTs.

Ravenslofty: see for example figure 5 of [the ABC9 paper](https://people.eecs.berkeley.edu/~alanmi/publications/2023/date23_gap.pdf).

KrystalDelusion: Fixed in 06db882

garytwong: Oh!!  Thank you for the explanation and the attention to spot it.  I appreciate the enlightenment.

(I had naively assumed the text in question was describing the kind of tradeoff where, say, one would want a look-ahead carry on the critical path but might tolerate a smaller-but-slower ripple carry elsewhere.  The subtleties of white box area recovery had gone over my head.)

Ravenslofty: This is a topic I can go into a lot of detail on, though the point of the document is to have some base I can point at before going into finer and more specific detail :p

PR #5154, opened 2025-05-30
read_verilog: fix -1 constant used to correct post increment/decrement

### LINKED ISSUE #5151: Incorrect handling of post decrement operation in `always_comb`
### Version

Yosys 0.51+46 (git sha1 8bde6ac17, g++ 12.2.0-14 -fPIC -O3)

Linux

### Reproduction Steps

```sv
module expr_postsub_comb (
    input logic [7:0] in_val_m2,
    input logic [7:0] sub_val_m2,
    output logic [7:0] out_diff_m2,
    output logic [7:0] var_out_m2
);
    logic [7:0] var_m2;
      always_comb begin
        var_m2 = in_val_m2;
        out_diff_m2 = (var_m2--) - sub_val_m2;
        var_out_m2 = var_m2;
      end
endmodule
```

sythesize with `yosys -p "read_verilog -sv expr_postsub_comb.sv ; proc; opt ; write_verilog -sv -noattr synth.sv"`

simulate with these values as input
```
  in_val_m2 = ff
  sub_val_m2 = 11
```

### Expected Behavior

output of `out_diff_m2` should be `238`

### Actual Behavior

output of `out_diff_m2` is in fact `236`

georgerennie:

Fixes #5151. See my comments there for more discussion about approaches for fixing this.

To handle post increment/decrement the frontend eagerly increments/decrements the variable before evaluating the expression, then in the expression undoes this by adding/subtracting `-1`. This wasn't working correctly as it did used a function to create `-1` that doesn't actually track negative values, instead tracking required sign extension for values interpreted as positive. We replace this with an explicit `-` in the AST.

Have added the testcase from #5151 and some basic other tests that the pre/post inc/dec operators are doing the right thing that should fail before this PR.

### DIFF
--- a/tests/verilog/incdec.ys
+++ b/tests/verilog/incdec.ys
@@ -1,0 +1,68 @@
+# From https://github.com/YosysHQ/yosys/issues/5151
+read_verilog -sv <<EOT
+module expr_postsub_comb (
+    input logic [7:0] in_val_m2,
+    input logic [7:0] sub_val_m2,
+    output logic [7:0] out_diff_m2,
+    output logic [7:0] var_out_m2
+);
+    logic [7:0] var_m2;
+    always_comb begin
+      var_m2 = in_val_m2;
+      out_diff_m2 = (var_m2--) - sub_val_m2;
+      var_out_m2 = var_m2;
+    end
+
+    always_comb assert(out_diff_m2 == in_val_m2 - sub_val_m2);
+endmodule
+EOT
+
+prep
+chformal -lower
+sat -prove-asserts -verify
+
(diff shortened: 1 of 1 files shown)


KrystalDelusion:
> Oh I've just seen that `asgn_expr.ys` should test for this, but it wasn't failing on this. I'm not quite sure why, perhaps due to the fact that the variables in use are `integer` not unsigned types.

Changing them to `logic [31:0]` causes the `sat -verify` to fail, so that seems likely.

KrystalDelusion (approved):

PR #5158, opened 2025-05-31
read_verilog/astsimplify: copy inout ports in and out of functions/tasks

### LINKED ISSUE #5157: `read_verilog`: `inout` parameters not copied out of tasks
This issue is related to #5151

### Version

Yosys 0.53+70 (git sha1 3ef4c91c3, g++ 12.2.0-14+deb12u1 -fPIC -O3)
And the #5154 PR 

Linux

### Reproduction Steps

```sv
module stmt_if_task (
    output logic [7:0] out_val_m6,
    input logic [7:0] in_val_m6,
    input bit condition_m6
);
    logic [7:0] var_m6;
    task automatic update_conditional_m6(input bit cond, inout logic [7:0] val);
        if (cond) begin
            val++;
        end else begin
            --val;
        end
    endtask
    always_comb begin
        var_m6 = in_val_m6;
        update_conditional_m6(condition_m6, var_m6);
        out_val_m6 = var_m6;
    end
endmodule
```

Simulate with these hexadecimal values as input
```txt
  in_val_m6 = 79 // or 121 in base10
  condition_m6 = 0
```

Synthesize with
```bash
yosys -p "read_verilog -sv stmt_if_task.sv ; proc; opt ; write_verilog -sv -noattr synth.sv"
```

### Expected Behavior

output of `out_val_m6` should be `120`

### Actual Behavior

output of `out_val_m6` is in fact `121`

georgerennie:

Fixes #5157

Previously, `inout` ports to tasks/functions were incorrectly handled as the code assumed ports were either input or output, so `inout` ports would only have the values copied in, not out when they are instantiated. This change makes it two separate branches that can both be true (and uses `->clone` methods so that both input and output get their own copies of the AST nodes if needed).

I've added the testcases from the issue as well as some basic tests using each of the port types for functions/tasks. These all fail before this change.

This probably conflicts with #5135 but I don't imagine its hard to fixup either depending on which gets merged first

### DIFF
--- a/tests/verilog/func_task_arg_copying.ys
+++ b/tests/verilog/func_task_arg_copying.ys
@@ -1,0 +1,132 @@
+# https://github.com/YosysHQ/yosys/issues/5157
+read_verilog -sv <<EOT
+module stmt_if_task (
+    output logic [7:0] out_val_m6,
+    input logic [7:0] in_val_m6,
+    input bit condition_m6
+);
+    logic [7:0] var_m6;
+    task automatic update_conditional_m6(input bit cond, inout logic [7:0] val);
+        if (cond) begin
+            val++;
+        end else begin
+            --val;
+        end
+    endtask
+    always_comb begin
+        var_m6 = in_val_m6;
+        update_conditional_m6(condition_m6, var_m6);
+        out_val_m6 = var_m6;
+    end
+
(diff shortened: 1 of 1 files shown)


KrystalDelusion (approved):
Appears to work as expected.  I also did a local build with ASAN just to double check (since CI isn't yet doing that) and no issues there.

PR #5159, opened 2025-05-31
Fix `proc` handling of whiteboxes

### LINKED ISSUE #5142: Warning: Ignoring boxed module $paramod\ALU\ALU_MODE=s32'00000000000000000000000000000010
### Version

Yosys 0.53+39 (git sha1 388955031, aarch64-apple-darwin23.5-clang++ 18.1.8 -fPIC -O3)

macOS

### Reproduction Steps

Transferring from discussions https://github.com/YosysHQ/yosys/discussions/5136

1. Create a file blinky.sv with the content below.

2. Run the command ``yosys -p "synth_gowin -top blinky -json _output.json" -q blinky.sv``

blinkey.sv:
```
module blinky #(
    // Num of click cycle per led toggle.
    parameter integer DIV = (27000000 / 2)
) (
    input        sys_clk,
    output logic led       // Active low
);

  // ---- Reset generator.

  logic [3:0] reset_counter = 0;
  logic sys_reset = 1;

  always_ff @(posedge sys_clk) begin
    if (reset_counter < 3) begin
      sys_reset <= 1;
      reset_counter <= reset_counter + 1;
    end else begin
      sys_reset <= 0;
    end
  end

  // ---- Blinker

  reg [31:0] blink_counter;

  always_ff @(posedge sys_clk) begin
    if (sys_reset) begin
      blink_counter <= 0;
      led <= 1;
    end else if (blink_counter < (DIV - 1)) begin
      blink_counter <= blink_counter + 1;
    end else begin
      blink_counter <= 0;
      led <= ~led;
    end
  end

endmodule
```

### Expected Behavior

Yosys completes with no warnings.

### Actual Behavior

The following warning is printed

```
Warning: Ignoring boxed module $paramod\ALU\ALU_MODE=s32'00000000000000000000000000000010
```

KrystalDelusion:

With the recent change to selections (#4768), a number of passes now give a warning when they ignore boxed modules, instead of doing so silently.  #5142 reveals that many of the `proc_*` passes ignore boxes, raising warnings when `abc9_ops` calls `proc` on the derived modules.  It seems like the major `proc_*` passes handled boxes but some of the newer optimizations used `selected_modules`, potentially without realizing that it was silently ignoring boxes.

All of the `proc_*` passes have now been updated to be consistent, universally accepting all selected modules, including boxes.

```c++
for (auto mod : design->all_selected_modules()) {
    // ...
    for (auto proc : mod->selected_processes()) {
        // ...
    }
    // ...
}
```

`opt_expr`, which is called at the end of `proc` by default, *also* emits a warning if the selection includes boxes.  However, unlike the `proc_*` passes, *none* of the `opt_*` passes operate on boxes.  Rather than changing `opt_expr` I instead changed `abc9_ops` to call `proc -noopt` and skip the `opt_expr` call entirely.

These two changes are sufficient to close #5142.

Following the instructions for #5142 should not produce any warnings with the changes from this PR, and none of the existing tests should break.

### DIFF
--- a/passes/proc/proc_arst.cc
+++ b/passes/proc/proc_arst.cc
@@ -288,43 +288,40 @@ struct ProcArstPass : public Pass {
 		extra_args(args, argidx, design);
 		pool<Wire*> delete_initattr_wires;
 
-		for (auto mod : design->modules())
-			if (design->selected(mod)) {
-				SigMap assign_map(mod);
-				for (auto &proc_it : mod->processes) {
-					if (!design->selected(mod, proc_it.second))
-						continue;
-					proc_arst(mod, proc_it.second, assign_map);
-					if (global_arst.empty() || mod->wire(global_arst) == nullptr)
-						continue;
-					std::vector<RTLIL::SigSig> arst_actions;
-					for (auto sync : proc_it.second->syncs)
-						if (sync->type == RTLIL::SyncType::STp || sync->type == RTLIL::SyncType::STn)
-							for (auto &act : sync->actions) {
-								RTLIL::SigSpec arst_sig, arst_val;
-								for (auto &chunk : act.first.chunks())
-									if (chunk.wire && chunk.wire->attributes.count(ID::init)) {
-										RTLIL::SigSpec value = chunk.wire->attributes.at(ID::init);
-										value.extend_u0(chunk.wire->width, false);
-										arst_sig.append(chunk);
-										arst_val.append(value.extract(chunk.offset, chunk.width));
-										delete_initattr_wires.insert(chunk.wire);
-									}
-								if (arst_sig.size()) {
-									log("Added global reset to process %s: %s <- %s\n",
-											proc_it.first.c_str(), log_signal(arst_sig), log_signal(arst_val));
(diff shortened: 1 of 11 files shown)


nakengelhardt:
I would like to document the rationale for why this is safe, i.e. why the situation that the call to `opt_expr` was added to handle can't occur in whiteboxes used by `abc9`. Admittedly I'm hazy on what that situation was, something related to constant propagation for memories? And so this can't happen here because abc doesn't understand memories, so you can't put them in box models anyway?

whitequark:
> Admittedly I'm hazy on what that situation was, something related to constant propagation for memories?

As far as I recall: The Verilog frontend emits memory initialization with addresses that may not be constants but are constant expressions, but the `$meminit` cell requires them to be constant. To handle this, the `opt_expr` call folds these constant expressions into constants. Otherwise synthesis just fails. (I think one memory write port per initialization value may also be involved, but this is outside of my expertise.)

nakengelhardt:
Thanks!

nakengelhardt (approved):

PR #5160, opened 2025-06-01
verilog: improve string literal matching speed (fixes #5076)

### LINKED ISSUE #5076: Parsing very long strings with read_verilog is much slower than alternatives
### Version

Yosys 0.50+14 (git sha1 077819572222094b30b3cc08e62132456095ca0d,  g++ 11.4.0 -fPIC -O3) 

Linux

### Reproduction Steps

Running 
```
yosys -p 'read -sv test.v'
```
against the attached `test.v` takes about 8 seconds on my system: 
```
Time spent: 99% 2x read_verilog (8 sec), 0% 1x read (0 sec)
```
comparing this to the equivalent file in `rtlil` gives 
```
Time spent: 100% 2x read_rtlil (0 sec)
```

Finally, reading the same file with yosys-slang's `read_slang` also reports:
```
Time spent: 59% 2x read_slang (0 sec), 22% 1x plugin (0 sec)
```

(Had to zip the input files for github to allow them)
[test.zip](https://github.com/user-attachments/files/19959661/test.zip)

### Expected Behavior

Parsing the rtlil and the verilog version should have roughly the same performance.

### Actual Behavior

The verilog input file takes drastically longer to parse.

garytwong:
Use a greedy regular expression to match input inside a string literal, so that flex can accumulate a longer match instead of invoking a rule for each individual character.

Before this change, matching long string literals was slow (see issue #5706).

A small regexp is added to match the body of string literals while in the `<STRING>` state; this allows the DFA to do the work of analysing the string so it doesn't have to suspend processing between characters.

From the `flex` manual:
> Another area where the user can increase a scanner's performance (and one that's easier to implement) arises from the fact that the longer the tokens matched, the faster the scanner will run.  This is because with long tokens the processing of most input characters takes place in the (short) inner scanning loop, and does not often have to go through the additional work of setting up the scanning environment (e.g., 'yytext') for the action.

Try the test case attached to #5076, or use the following shell script.  There should be a significant speedup (say 50-100x) after applying this change.

```
#!/bin/sh
a='Hello, world.   ' # 16 chars
b="$a$a$a$a$a$a$a$a$a$a$a$a$a$a$a$a" # 256 chars
c="$b$b$b$b$b$b$b$b$b$b$b$b$b$b$b$b" # 4096 chars
d="$c$c$c$c$c$c$c$c" # 32768 chars
echo "module big;" > big.v
echo "    initial if( 0 ) \$display( \"$d\" );" >> big.v
echo "    initial if( 0 ) \$display( \"$d\" );" >> big.v
echo "    initial if( 0 ) \$display( \"$d\" );" >> big.v
echo "    initial if( 0 ) \$display( \"$d\" );" >> big.v
echo "    initial if( 0 ) \$display( \"$d\" );" >> big.v
echo "    initial if( 0 ) \$display( \"$d\" );" >> big.v
echo "    initial if( 0 ) \$display( \"$d\" );" >> big.v
echo "    initial if( 0 ) \$display( \"$d\" );" >> big.v
echo "endmodule" >> big.v
${YOSYS:-yosys} -p "read_verilog big.v"
```

### DIFF
--- a/frontends/verilog/verilog_lexer.l
+++ b/frontends/verilog/verilog_lexer.l
@@ -336,7 +336,7 @@ TIME_SCALE_SUFFIX [munpf]?s
 }
 
 \"		{ BEGIN(STRING); }
-<STRING>\\.	{ yymore(); real_location = old_location; }
+<STRING>([^\"]|\\.)+	{ yymore(); real_location = old_location; }
 <STRING>\"	{
 	BEGIN(0);
 	char *yystr = strdup(yytext);
@@ -376,7 +376,6 @@ TIME_SCALE_SUFFIX [munpf]?s
 	free(yystr);
 	return TOK_STRING;
 }
-<STRING>.	{ yymore(); real_location = old_location; }
 
 and|nand|or|nor|xor|xnor|not|buf|bufif0|bufif1|notif0|notif1 {
 	yylval->string = new std::string(yytext);


zachjs (approved):

zachjs:
Out of an abundance of caution, let’s merge this just after the next Yosys release.

mmicko:
This one breaks one of projects I am using as test case. Have minimized this to :
Without this PR it works fine.

```
module top(input wire clk, input wire [7:0] scancode, output reg [7:0] a );
    always @(posedge clk)
    case (scancode)
        30: a = "1";
        31: a = "2";
        32: a = "3";
        33: a = "4";
        34: a = "5";
        35: a = "6";
        36: a = "7";
        37: a = "8";
        38: a = "9";
        39: a = "0";
        40: a = "\n";       // enter
        41: a = 27;         // esc
        42: a = 8;          // backspace
        43: a = 9;          // tab
        44: a = 32;         // space
        45: a = "-";        // -
        46: a = "=";        // =
        47: a = "[";        // [
        48: a = "]";        // ]
        49: a = "\\";       // \
        50: a = "#";        // non-use # ~
        51: a = ";";        // ;
        52: a = "'";        // '
        53: a = "`";        // `
        54: a = ",";        // ,
        55: a = ".";        // .
        56: a = "/";        // /
        57: ;               // caps lock
    endcase
endmodule

```

garytwong:
Thank you for the report and test case!  I have a patch fixing the problem for me in commit 40aa7eaf93ec922be75024ecedc855130381da83.  Could you please test whether that change works for you?

mmicko:
@garytwong yes, this does help. Thanks for looking into it.

PR #5162, opened 2025-06-04
Make attrmap able to alter memory attributes as well

mmicko:
It was not possible to alter memory attributes with attrmap. 

By using `selected_members` iterator instead we are covering all as we already do plus memories.

attrmap_apply was removed from processes block, but also used a chance to cleanup and use selected_processes to handle rest of changes needed

### DIFF
--- a/passes/techmap/attrmap.cc
+++ b/passes/techmap/attrmap.cc
@@ -263,16 +263,17 @@ struct AttrmapPass : public Pass {
 
 		if (modattr_mode)
 		{
-			for (auto module : design->selected_whole_modules())
+			for (auto module : design->all_selected_whole_modules())
 				attrmap_apply(stringf("%s", log_id(module)), actions, module->attributes);
 		}
 		else
 		{
-			for (auto module : design->selected_modules())
+			for (auto module : design->all_selected_modules())
 			{
-				for (auto wire : module->selected_members())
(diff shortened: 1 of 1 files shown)


KrystalDelusion:
Since we are touching this file we should also review the deprecation warning on it.  From #4768:
> Note:
> `Design::selected_whole_modules()` is marked deprecated because I'm not convinced `attrmap` should be ignoring boxes, even though that is the current behaviour.

`Design::selected_whole_modules()` previously iterated over all wholly selected modules while silently ignoring any boxed modules.  Most of the previous uses of this method were updated in #4768 to use the equivalent, `Design::selected_unboxed_whole_modules()`, but `attrmap` wasn't because I wasn't sure how it should handle boxes.

For comparison, `setattr` uses `Design::all_selected_modules()` which includes boxes. `wbflip` specifically says "Blackbox cells are not effected by this command" and uses `Design::selected_modules(RTLIL::SELECT_ALL, RTLIL::SB_EXCL_BB_ONLY)`.  `blackbox` uses `Design::selected_whole_modules_warn(true)` to include whiteboxes but warns on blackboxes and partially selected modules.

For the non `-modattr` mode it is using the default `Design::selected_modules()` which includes partially selected modules, and warns on any boxes in the selection.  If we end up including boxes for `-modattr` we may also want to change the other one to match.

PR #5173, opened 2025-06-07
Allows calling yosys_shutdown and then yosys_setup to restart.

RonxBulld:

When using yosys as a dynamic library, the yosys environment is sometimes opened and closed multiple times for different purposes, and the already_setup flag is not maintained correctly.

Simply maintain the already_setup and already_shutdown flags.

Only one regression test needs to be executed. I don't know if there is any place that takes advantage of the side effects of the previous commit changes (although that feature doesn't make common sense).

### DIFF
--- a/kernel/yosys.cc
+++ b/kernel/yosys.cc
@@ -187,12 +187,14 @@ int run_command(const std::string &command, std::function<void(const std::string
 #endif
 
 bool already_setup = false;
+bool already_shutdown = false;
 
 void yosys_setup()
 {
 	if(already_setup)
 		return;
 	already_setup = true;
+	already_shutdown = false;
 
 #ifdef WITH_PYTHON
 	// With Python 3.12, calling PyImport_AppendInittab on an already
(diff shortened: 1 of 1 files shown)


widlarizer (approved):

widlarizer:
Technically the PR seems valid regardless, but could you clarify your use case for totally restarting yosys? I would expect doing the equivalent of `design -clear` to be enough to reset the state sufficiently for most use cases

RonxBulld:
> Technically the PR seems valid regardless, but could you clarify your use case for totally restarting yosys? I would expect doing the equivalent of `design -clear` to be enough to reset the state sufficiently for most use cases

I'm very very sorry for being late. According to my design, Yosys will be a common component that serves multiple modules in the compilation process. Sometimes I want to dynamically have different Pass environments in multiple modules, which is more in line with semantic closures, but I currently have no plan to make these Passes into dynamic link libraries, so I consider using setup/shutdown to maintain a clean semantic environment.

PR #5177, opened 2025-06-13
rename: add -unescape

widlarizer:
ASIC flows with some commercial tools rely on the transformation of escaped Verilog identifiers into simple identifiers by replacing some characters with underscores. Most notably, `a[0]` gets renamed to `a_0_`. This PR implements a new mode for `rename` that implements this and ensures no selected public name has to be escaped by `write_verilog`. Conflicts in renaming are resolved by appending numbering suffixes.

The current implementation approach relies on the verilog backend exposing its escaping logic in a new header, `verilog_backend.h`. In addition to seeing the test passes, reviewers may add `write_verilog` to its end and observe the end result is such that no wire names are escaped by the Verilog backend

### DIFF
--- a/passes/cmds/rename.cc
+++ b/passes/cmds/rename.cc
@@ -20,6 +20,7 @@
 #include "kernel/register.h"
 #include "kernel/rtlil.h"
 #include "kernel/log.h"
+#include "backends/verilog/verilog_backend.h"
 
 USING_YOSYS_NAMESPACE
 PRIVATE_NAMESPACE_BEGIN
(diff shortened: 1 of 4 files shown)


widlarizer:
cc @mikesinouye

mikesinouye:
This is fantastic, thank you for your prompt action on this. I will experiment with this PR.

passes/cmds/rename.cc
```diff
@@ -252,6 +273,12 @@ struct RenamePass : public Pass {
 		log("can be used to change the random number generator seed from the default, but it\n");
 		log("must be non-zero.\n");
 		log("\n");
+		log("\n");
+		log("    rename -unescape [selection]\n");
+		log("\n");
+		log("Rename all selected public wires and cells that have to be escaped.\n");
+		log("Replaces characters with underscores or adds additional underscores and numbers.\n");
```
KrystalDelusion: I think either the option or the text should say that it is specific to Verilog.

KrystalDelusion (approved):

mikesinouye:
Thank you for developing this @widlarizer

PR #5180, opened 2025-06-14
Add ASAN to CI

KrystalDelusion:

We have UBSAN running on CI, but sometimes ASAN issues slip through as well (e.g. #5020, #5163).

Define the sanitizer as part of the test strategy matrix in the `test-build.yml` family of CI jobs.

There are still memory leaks being picked up in the Verific tests, but I added a (hopefully temporary) patch to `tests/verific/run-test.sh` to disable `halt_on_error` for the tests in that directory.  The sanitizer builds are running on the open source runner, so they shouldn't actually hit the Verific tests _anyway_, but better safe than sorry.

CI should pass, having run the test suite with ASAN.

### DIFF
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -40,6 +40,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
+        sanitizer: [undefined, address]
       fail-fast: false
     steps:
       - name: Checkout Yosys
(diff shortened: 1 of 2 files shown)


KrystalDelusion:
FYI this is failing CI because of a memory issue introduced by #5152

widlarizer:
I have this fixed in my big ownership PR (#5135), specifically [this bit](https://github.com/YosysHQ/yosys/pull/5135/commits/5af4e051251b354e0253a9eb2ba7ea95bbb077f5#diff-cfa391ea6199382f4c13a4d3f892e3eaa76ff9b24793ee31c045793afe71a550) but applying an equivalent to the relevant part to main isn't enough, see #5189

PR #5181, opened 2025-06-16
const2ast: prove tests cover error filenames

widlarizer:
The deletion of uses of the AST layer global state member `current_filename` should be caught by tests. I will close this draft PR when rebasing onto main makes the tests fail.

> we have mutation testing at home
mutation testing at home:


widlarizer:
They'll get removed in #5135 as they are definitely no longer needed there

PR #5182, opened 2025-06-16
kernel: treat zero width constant as zero

### LINKED ISSUE #5175: Regression in `opt_expr` in Yosys 0.54
### Version

Yosys 0.54 (git sha1 db72ec3bd, ccache clang 18.1.3 -O3 -flto -flto)

Linux

### Reproduction Steps

Reproducer: [video-rgb-input.il.txt](https://github.com/user-attachments/files/20663224/video-rgb-input.il.txt)

```
yosys video-rgb-input.il -p proc
```

### Expected Behavior

Works (which it did in 0.53).

### Actual Behavior

```
...

2.12. Executing OPT_EXPR pass (perform const folding).
Optimizing module top.cd_sync.clk_buf.
Optimizing module top.cd_sync.
Optimizing module top.U$5.U$7.
Optimizing module top.U$5.U$6.
Optimizing module top.U$5.U$5.
Optimizing module top.U$5.U$4.
Optimizing module top.U$5.dck_buffer.io.
Optimizing module top.U$5.dck_buffer.
Optimizing module top.U$5.b_buffer.io.
Optimizing module top.U$5.b_buffer.
Optimizing module top.U$5.g_buffer.io.
Optimizing module top.U$5.g_buffer.
Optimizing module top.U$5.r_buffer.io.
Optimizing module top.U$5.r_buffer.
Optimizing module top.U$5.
<suppressed ~7 debug messages>
ERROR: Assert `!empty()' failed in ../yosys-src/kernel/rtlil.cc:5532.
```

georgerennie:

Fixes #5175 

Zero width constants are special cased and treated as the constant value zero.

The testcase from #5175

### DIFF
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -384,7 +384,7 @@ bool RTLIL::Const::convertible_to_int(bool is_signed) const
 {
 	auto size = get_min_size(is_signed);
 
-	if (size <= 0)
+	if (size < 0)
 		return false;
 
 	// If it fits in 31 bits it is definitely convertible
@@ -5509,6 +5509,9 @@ bool RTLIL::SigSpec::convertible_to_int(bool is_signed) const
 	if (!is_fully_const())
 		return false;
 
+	if (empty())
+		return true;
+
 	return RTLIL::Const(chunks_[0].data).convertible_to_int(is_signed);
 }
 
@@ -5520,6 +5523,9 @@ std::optional<int> RTLIL::SigSpec::try_as_int(bool is_signed) const
 	if (!is_fully_const())
 		return std::nullopt;
 
+	if (empty())
+		return 0;
+
 	return RTLIL::Const(chunks_[0].data).try_as_int(is_signed);
 }
 
@@ -5529,7 +5535,10 @@ int RTLIL::SigSpec::as_int_saturating(bool is_signed) const
 
 	pack();
 	log_assert(is_fully_const() && GetSize(chunks_) <= 1);
-	log_assert(!empty());
+
+	if (empty())
+		return 0;
+
 	return RTLIL::Const(chunks_[0].data).as_int_saturating(is_signed);
 }
 


cr1901:
Can confirm this fixes [Sentinel CI failures](https://github.com/cr1901/sentinel/actions/runs/15371389250/job/44082887582?pr=52) when I test this patch locally. I'll just have to wait until a `yosys` with this PR makes its way to OSS-CAD-Suite/`yowasp` before I can make CI green.

georgerennie:
Maybe it would be useful to have a few amaranth designs in the test suite - verilog derived designs generally shouldn't end up with zero width cells.

KrystalDelusion:
Treating a disconnected port (zero width or not) as a constant instead of undefined feels counter intuitive to me...  Is there precedence for this elsewhere in Yosys, or is it just that it worked before so we shouldn't change it?

georgerennie:
Yeah I am equally kinda uncomfortable with it which is why I had put an assertion guarding against it in there in the first place - to me it seemed like something that wouldn't be super well defined and should be avoided. That being said, it does seem that amaranth codegen seems to rely on this behaviour so we should support some behaviour for it.

The existing `as_int` infrastructure will treat it as zero, and after discussing with @widlarizer and others I think it is fair to define it as zero, that is the only value that can be represented in [0, 2^0). `x`s tend to have two meanings in Yosys, either as a don't care or more specifically as a Verilog `x`. Verilog `x`s (and signals in general) can't be zero width, so I think its reasonable to assume if we have a zero width signal it doesnt correspond to a Verilog `x`, and if we take it to mean don't care then the only concrete value that don't care can actually correspond to for that width is `0` so `x` and `0` are equivalent. Its worth noting that the cells this matters for are not just ones with undefined inputs, but specifically with a zero width - a disconnected port would normally be defined with `'x` and a non-zero width instead afaik?

Also worth noting that this new `try_as_int`/`as_int_saturating` infrastructure is currently only being used in the shift optimization in `opt_expr`, so anywhere else in the codebase is still using `as_int` that treats it as zero.

I'm not sure any of this is 100% convincing, but after thinking about it for a bit it seems as reasonable approach as any to me and shouldnt change existing behaviour.

KrystalDelusion:
> after discussing with @widlarizer and others I think it is fair to define it as zero, that is the only value that can be represented in [0, 2^0)

Yeah actually that does seem reasonable

whitequark:
> Treating a disconnected port (zero width or not) as a constant instead of undefined feels counter intuitive to me...

I'm not sure I agree. A zero-width concatenation isn't the same as being disconnected, is it?

KrystalDelusion:
`connect \B {  }` is (also) used to describe a disconnected port

whitequark:
> `connect \B { }` is (also) used to describe a disconnected port

Oh, that's... cursed. I think this is an issue with the RTLIL design, but now that you bring it up I *also* think that Amaranth may be misusing RTLIL here.

whitequark:
@KrystalDelusion On discussion, I think this is wrong. There aren't any port bits that aren't being connected here. The port is also zero width:

```
    parameter \B_WIDTH 0
    connect \B {  }
```

whitequark:
Also, wouldn't a disconnected port just be missing from the connection list of a cell?

KrystalDelusion:
> Also, wouldn't a disconnected port just be missing from the connection list of a cell?

Nope
```
ERROR: Found error in internal cell \top.$85 ($shr) at kernel/rtlil.cc:1439:
  cell $shr $85
    parameter \Y_WIDTH 1
    parameter \B_WIDTH 0
    parameter \A_WIDTH 1
    parameter \B_SIGNED 0
    parameter \A_SIGNED 0
    connect \Y \Y
    connect \A \A
  end
```

whitequark:
> Nope

I don't see how your snippet suppots your conclusion. All it shows is that internal cells don't allow ports to be disconnected.

KrystalDelusion:
> I don't see how your snippet suppots your conclusion. All it shows is that internal cells don't allow ports to be disconnected.

The cell that started this PR/discussion was a $shr cell, I said that a connection to `{ }` was the way a disconnected port was rendered in RTLIL, you suggested that a disconnected port could be missing from the list of connections rather than being connected to `{ }`, I showed that a $shr cell (which is, once again, the cell that started this) missing a port connection was invalid RTLIL. The way that RTLIL describes a disconnected port (for internal cells and not) *is by connecting them to* `{ }`:
https://github.com/YosysHQ/yosys/blob/a0a77cd1d0b8710e8bbfb6f3188b5d43bdaf44fd/backends/rtlil/rtlil_backend.cc#L107-L119
That's what gets called when dumping a cell.

KrystalDelusion:
For what it's worth I was initially trying to show `read_RTLIL` of a cell with a missing port connection would be printed back out during `write_RTLIL` as `connect { }`, but I was mistaken and that isn't true for non-internal cells.

whitequark:
> For what it's worth I was initially trying to show `read_RTLIL` of a cell with a missing port connection would be printed back out during `write_RTLIL` as `connect { }`, but I was mistaken and that isn't true for non-internal cells.

Yes, that is what I'm trying to tell you.

gatecat:
Friendly ping, has this been forgotten about?

whitequark:
Disappointingly this wasn't a part of the last release..

KrystalDelusion:
> has this been forgotten about?

That's probably my fault.  I was getting frustrated with the discussion and had to stop engaging.

georgerennie:
Yeah I've been away a lot in the past few weeks and not particularly actively keeping track of things, I probably could have prodded it as well.

PR #5183, opened 2025-06-16
logger: add -expect types prefix-log, prefix-warning, prefix-error

widlarizer:
For #5135 I need to be able to check the correctness of filenames and line numbers in diagnostics. In this PR I add new types for `logger -expect` that also match not just on the content of the diagnostic but also on the prefix. I use this in new tests that cover some const2ast code. The intention is to resolve #5181 with this infrastructure soon

### DIFF
--- a/kernel/log.cc
+++ b/kernel/log.cc
@@ -44,6 +44,7 @@ std::vector<std::string> log_scratchpads;
 std::map<std::string, std::set<std::string>> log_hdump;
 std::vector<std::regex> log_warn_regexes, log_nowarn_regexes, log_werror_regexes;
 dict<std::string, LogExpectedItem> log_expect_log, log_expect_warning, log_expect_error;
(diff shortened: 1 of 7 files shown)


kernel/log.cc
```diff
@@ -266,9 +267,15 @@ static void logv_warning_with_prefix(const char *prefix,
 				log_error("%s",  message.c_str());
 
 		bool warning_match = false;
-		for (auto &item : log_expect_warning)
-			if (std::regex_search(message, item.second.pattern)) {
-				item.second.current_count++;
+		for (auto &[_, item] : log_expect_warning)
```
KrystalDelusion, line 270: I need to remember this pattern, it's so much nicer than the usual `.first` `.second`

KrystalDelusion (approved):
I haven't tested it, but the code seems sensible and the included tests are passing CI.  Not convinced by the filenames of the tests, but that might be personal preference rather than a blocker.

PR #5187, opened 2025-06-19
verilog: fix string literal regular expression

garytwong:
Fix the bug reported in #5160, where a closing quote on a string literal failed to terminate the string when preceded by an escaped backslash.
_Originally posted by @mmicko in https://github.com/YosysHQ/yosys/issues/5160#issuecomment-2983643084_

Properly escape the backslash in the regular expression in `verilog_lexer.l`.

See the included regression test `tests/verilog/bug5160.v`, or @mmicko's test case in the original report.

With this fix applied, both cases work properly.  Before the fix, parsing would fail with a spurious syntax error.

### DIFF
--- a/frontends/verilog/verilog_lexer.l
+++ b/frontends/verilog/verilog_lexer.l
@@ -336,7 +336,7 @@ TIME_SCALE_SUFFIX [munpf]?s
 }
 
 \"		{ BEGIN(STRING); }
-<STRING>([^\"]|\\.)+	{ yymore(); real_location = old_location; }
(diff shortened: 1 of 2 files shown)


garytwong:
While fixing this, I also noticed another bug related to escape sequences in string literals: yosys doesn't seem to process escaped newlines (as in string literal split over multiple lines, not the "\n" sequence) according to IEEE 1800-2012 sec 5.9.  (It looks like such strings are accepted, but the resulting value can be wrong.)

However, since that bug predates the recent changes, I'll plan to submit a new PR once I have time to investigate.

PR #5188, opened 2025-06-19
Makefile: fix hardcoded -soname for libyosys.so

Jiahui17:

I am using the Yosys API in my project through `libyosys.so`. However, the current Makefile always hardcodes `soname` to `$PREFIX/lib/yosys/libyosys.so`. Unless the `libyosys.so` is exactly in that location, the runtime linking will fail even if the `LD_LIBRARY_PATH` contains the shared lib.

This PR removes the hardcoded prefix.


ShinyKate:
@mmicko need anything for this?

Jiahui17:
Thanks!

PR #5190, opened 2025-06-20
dfflibmap: propagate negated next_state to output correctly

### LINKED ISSUE #5186: Invalid mapping of ASAP7 FFs
### Yosys version

Yosys 0.53 (git sha1 53c22ab7c, g++ 15.1.1 -fPIC -O3)

Linux

### Problem

As noticed by @povik, there's an [issue with how Yosys handles input vs output inversion](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3215#issuecomment-2960758173).
See the entire thread in https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3215 for more context.
I'm not very familiar with this, so I'm unable to provide a better description. 
Reporting this so it doesn't get lost.

widlarizer:
Fixes #5186. Prior to this PR, a negation in next_state was considered equivalent to an inverter on the output port. However, flops with set or reset need swapping.

- [x] add test covering original report
- [x] ensure flop kinds other than `_DFFSR_*` with set or reset are still mapped correctly as well

### DIFF
--- a/tests/techmap/dfflibmap.ys
+++ b/tests/techmap/dfflibmap.ys
@@ -81,3 +81,19 @@ clean
 select -assert-count 0 t:dffn
 select -assert-count 5 t:dffsr
 select -assert-count 1 t:dffe
+
+design -load orig
+dfflibmap -liberty dfflibmap.lib -liberty dfflibmap_dffsr_mixedpol.lib -dont_use dffsr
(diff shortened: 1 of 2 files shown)


povik:
Good to merge?

PR #5191, opened 2025-06-21
json: add -wide mode with word-level net indices

widlarizer:
For well-behaved designs, Yosys passes `bufnorm -nomode -buf -chain; splice -wires` can be safely used to normalize the design such that we eliminate `SigSig`s and slicing and concatenation in `SigSpec`s. By default, the json backend implements slicing/concatenating SigSpecs by representing all values as concatenations of bits. If a yosys json consumer tool is free to do the aforementioned normalization, this PR allows it to use word-level net indices instead

Example cell: notice `"connections"`

```json
        "$not$tests/sat/alu.v:50$10": {
          "hide_name": 1,
          "type": "$not",
          "parameters": {
            "A_SIGNED": "00000000000000000000000000000000",
            "A_WIDTH": "00000000000000000000000000001000",
            "Y_WIDTH": "00000000000000000000000000001000"
          },
          "attributes": {
            "src": "tests/sat/alu.v:50.18-50.20"
          },
          "port_directions": {
            "A": "input",
            "Y": "output"
          },
          "connections": {
            "A": 1250077441,
            "Y": -132126554
          }
        }
```
- [ ] add test
- [ ] help string
- [ ] enumerate nets instead of dumping `hashidx_`
- [ ] normalization probably depends on #5004 being available to work consistently

### DIFF
--- a/backends/json/json.cc
+++ b/backends/json/json.cc
@@ -35,6 +35,7 @@ struct JsonWriter
 	bool aig_mode;
 	bool compat_int_mode;
 	bool scopeinfo_mode;
+	bool wide_mode;
 
 	Design *design;
 	Module *module;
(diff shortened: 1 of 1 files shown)


widlarizer:
I'm no longer planning to continue this

PR #5197, opened 2025-06-26
OpenSTA verilog compatibility

widlarizer:
OpenSTA's `read_verilog` command supports a limited subset of Verilog. With this PR, yosys should be able to consistently present a design including derived blackboxes of internal cells made public with. The usefulness of these code changes is yet to be determined so it's a draft for now

I'm adding to `box_derive` a mode named `-apply` which actually remaps instances of parametric modules to be instances of the new derived ones. I'm also adding a `publish` pass that escapes internal cell types to become valid identifiers. @povik does the `-apply` mode seem sensible? 

- [x] add tests
- [ ] ~~discard cell parameters in `box_derive -apply`~~ I ended up not using this but for my use case I'm contributing it anyway since it's well defined and tested. Removing the parameters could be done in a followup if necessary


passes/cmds/publish.cc
```diff
@@ -0,0 +1,40 @@
...
+private:
+	static void publish(RTLIL::IdString& id) {
+		if (id.begins_with("$")) {
+			log_debug("publishing %s\n", id.c_str());
+			id = "\\" + id.str();
+			log_debug("published %s\n", id.c_str());
+		}
+	}
+public:
+	PublishPass() : Pass("publish", "publish private cell types") { }
```
povik: I'd say this is a general name for a niche application. What about `chtype -publish_icells`?

passes/cmds/box_derive.cc
```diff
@@ -51,6 +51,9 @@ struct BoxDerivePass : Pass {
 		log("        replaces the internal Yosys naming scheme in which the names of derived\n");
 		log("        modules start with '$paramod$')\n");
 		log("\n");
+		log("    -apply\n");
+		log("        use the derived modules\n");
```
povik, line 55: What about `-apply_derived_type`?

povik (commented):
I'd rename the new option otherwise the `box_derive` changes look good to me

widlarizer:
Ideally, I'd add 
```
design -reset
read_verilog <<EOT
module top(input A, input B, output Y);
assign Y = A ^ B;
endmodule
EOT

select -assert-count 1 t:$xor
select -assert-count 0 t:\$xor
chtype -publish_icells
dump
select -assert-count 0 t:$xor
select -assert-count 1 t:\$xor
```
or similar to `tests/techmap/techmap_chtype.ys`. However, #5414

PR #5198, opened 2025-06-26
add linecoverage command to generate lcov report from selection

nakengelhardt:

Creating a useful coverage metric for formal verification is hard, but a simple sanity check coverage metric is to look for any logic that does not feed into any assertions at all. (Similarly to line coverage in software, a line being covered doesn't tell you much about how good your tests are, but a line not being covered at all tells you something about how bad they are.) With the select command it's already easy to find these logic elements, but for better user access this information needs to be exported somehow.

This command takes a selection (the covered elements, for whatever idea of "covered" the user has - this doesn't implement the coverage metric) and dumps a report in `lcov` format. (The bare-bones lcov format is easy to create and plenty of tools understand it, and we already generate it in MCY as well.) Any line that is found on a src attribute of a design element not in the selection is marked as uncovered.

Line coverage is an imperfect metric and the provenance of netlist elements can be a bit vague as well, so there are several aspects to consider when constructing the selection and using this command:
1) The design needs to be flattened to see which portions of higher level modules are covered by assertions in submodules. We'll need to use `design -save` and `design -load` to undo these transformations in flows that don't flatten, e.g. for the `smtbmc` engine in SBY.
2) Several (sometimes structurally completely independent) design elements can have the same source location. This is particularly bad for `always` blocks, where all inferred registers will have the entirety of the block as source (cf. `reg1` and `reg2` in the testcase). Since "covered" is weak information but "uncovered" is strong information, I chose to have any uncovered element override the line's status.
3) There are several cells in the netlist that don't represent design logic that is expected to count for coverage - e.g. `$scopeinfo` cells, or `$assume` cells and their associated logic implementing the assumed property. I suggest always including any cells and wires that do not feed into any output signals in the selection, so that these are ignored.

Currently my suggested script for exporting structural assertion coverage would be this:
```
prep -flatten -top <top>
select -set covered t:$assert %ci*
select -set irrelevant o:* %ci* %n
linecoverage -lcov coverage.lcov @covered @irrelevant %u
```
If you have any designs that include formal properties, I would appreciate feedback if those selections produce output that matches your expectations of coverage.

### DIFF
--- a/passes/cmds/linecoverage.cc
+++ b/passes/cmds/linecoverage.cc
@@ -29,7 +29,7 @@ PRIVATE_NAMESPACE_BEGIN
 static const std::regex src_re("(.*):(\\d+)\\.(\\d+)-(\\d+)\\.(\\d+)");
 
 struct CoveragePass : public Pass {
-	CoveragePass() : Pass("linecoverage", "write coverage information to file") { }
+	CoveragePass() : Pass("linecoverage", "report coverage information") { }
 	void help() override
 	{
 		//   |---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|


passes/cmds/linecoverage.cc
```diff
@@ -0,0 +1,148 @@
...
+#include <regex>
+
+USING_YOSYS_NAMESPACE
+PRIVATE_NAMESPACE_BEGIN
+
+
+static const std::regex src_re("(.*):(\\d+)\\.(\\d+)-(\\d+)\\.(\\d+)");
+
+struct CoveragePass : public Pass {
+	CoveragePass() : Pass("linecoverage", "write coverage information to file") { }
```
KrystalDelusion: Minor nitpick, but it only writes if the `-lcov` option is provided, and unlike say `jny` it doesn't output to the terminal if no output file is given (which I think makes sense so we don't lock ourselves into the lcov format).  It may be more correct to say "find coverage information", or add a summary (all lines count vs uncovered lines count) and say "print coverage information".

iv2nl0b9v:
@nakengelhardt could you, please, add a test with some assertions? Also, would this approach work on non-synthesizeable assertions?

nakengelhardt:
@iv2nl0b9v I've updated the test to use an assertion and the suggested script; for more complex examples we might want to create a documentation page instead once we've nailed down the SBY integration.

As for "non-synthesizable" I'm not sure what kind of assertions you're thinking of exactly - Yosys as a synthesis tool by definition can't work with non-synthesizable code. This will work for any assertion that is represented in the Yosys internal format RTLIL, so if it isn't rejected by the frontend or at import you're good.

iv2nl0b9v:
I meant assertions that include the time component, for instance:
- `##n`: Cycle delay
- `s_eventually`: Eventually
- `s_always`: Always
- `s_nexttime`: Next time
- `s_until`, `s_until_with_overlap`

And repetition operators like `[*n]`, `[=n]`, `[->n]`. As far as I understand, they are non-synthesizable or may be hard to synthesize.

nakengelhardt:
[We do support most of these operators in principle when using the verific frontend](https://yosyshq.readthedocs.io/projects/sby/en/latest/verific.html#sequences), though there can be limitations if the sequence becomes too complex (the importer may bail out if the checker state machine gets too huge).

iv2nl0b9v:
Sounds good, thank you!

PR #5200, opened 2025-06-27
Use Booth multipliers for GateMate

mmicko:
To cover soft logic implementation of multipliers using now Booth multipliers due to better resource usage. This makes some larger designs to place and route easier. This affects only leftovers or all multipliers in case `-nomult` is provided.

@pu-cc please check if you see any issues with this.

### DIFF
--- a/techlibs/gatemate/synth_gatemate.cc
+++ b/techlibs/gatemate/synth_gatemate.cc
@@ -237,6 +237,7 @@ struct SynthGateMatePass : public ScriptPass
 
 		if (check_label("coarse"))
 		{
+			run("booth");
 			run("alumacc");
 			run("opt");
 			run("memory -nomap");


pu-cc:
Thanks for letting me know, but I got some issues e.g. with vexriscv in litex. Without the “booth” pass, it basically always works without any problems, so I would prefer to test this on more examples first.

povik:
>  I got some issues e.g. with vexriscv in litex. Without the “booth” pass, it basically always works without any problems

Are those issues with place and route, or correctness issues?

pu-cc:
Seems to be causing timing issues, so yes, place and route. But I am not yet working on @mmicko's  [improved_model2](https://github.com/YosysHQ/prjpeppercorn/tree/improved_model2) branch in prjpeppercorn.

mmicko:
This is bad performing due to timing so closing

PR #5202, opened 2025-06-30
Himbaechel for Xilinx uarch : Improve mapping of multiplexers

marzoul:

Implementation of multiplexers for uarch XIlinx is currently often suboptimal or excessively biased for minor speed gain at the cost of a significant area increase. Stress on nextpnr is also unnecessarily high with large proportion of muxf7/muxf8.

This contribution brings area reduction in personal test design (not that much mux-intensive) of 1.1% of SLICE_LUTX and 17.7% of SELMUX2_1 as reported by nextpnr.

- Add explicitly handling of A_WIDTH=1 for completeness
- mux2 uses one LUT3 instead of a hard mux (which did use LUTs anyway)
- mux4 uses one LUT4 instead of hard muxes (which did use LUTs anyway)
- mux8 uses only bottom half of a slice
- Add a mux12 for intermediate variant between mux8 and mux16
- For sizes larger than 16 inputs, instantiate the right mux size
- More comments about implementation choices
- More tests including with -widemux and -abc9, and more comments

Tests are included.

TODO : Find how to specify a range for inputs to do equivalence check for mux12 and mux20

### DIFF
--- a/techlibs/xilinx/cells_map.v
+++ b/techlibs/xilinx/cells_map.v
@@ -195,20 +195,18 @@ module \$__XILINX_SHIFTX (A, B, Y);
     else if (A_WIDTH < `MIN_MUX_INPUTS) begin
       wire _TECHMAP_FAIL_ = 1;
     end
-    else if (A_WIDTH == 2) begin
-      MUXF7 fpga_hard_mux (.I0(A[0]), .I1(A[1]), .S(B[0]), .O(Y));
+    else if (A_WIDTH == 1) begin
+      assign Y = A[0];
     end
+    // Use one LUT3 instead of a MUXF7 because a MUXF7 gets its inputs from 2 LUT6.O6 anyway
+    else if (A_WIDTH == 2) begin
+      assign Y = B[0] ? A[1] : A[0];
+    end
+    // Use one LUT6 instead of 2 LUT3 + MUXF7 because a MUXF7 gets its inputs from 2 LUT6.O6 anyway
     else if (A_WIDTH <= 4) begin
-      wire [4-1:0] Ax;
-      if (A_WIDTH == 4)
-        assign Ax = A;
-      else
-        // Rather than extend with 1'bx which gets flattened to 1'b0
-        // causing the "don't care" status to get lost, extend with
-        // the same driver of F7B.I0 so that we can optimise F7B away
-        // later
-        assign Ax = {A[1], A};
-      \$__XILINX_MUXF78 fpga_hard_mux (.I0(Ax[0]), .I1(Ax[2]), .I2(Ax[1]), .I3(Ax[3]), .S0(B[1]), .S1(B[0]), .O(Y));
+      wire [4-1:0] Ax = {{(4-A_WIDTH){1'bx}}, A};
+      assign Y = B[1] ? (B[0] ? Ax[3] : Ax[2])
+                      : (B[0] ? Ax[1] : Ax[0]);
     end
     // Note that the following decompositions are 'backwards' in that
     // the LSBs are placed on the hard resources, and the soft resources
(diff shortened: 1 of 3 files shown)


marzoul:
Sorry I accidentally included "himbaechel" in the original title, which is a term specific to nextpnr to designate the backend implementation that provides support for xilinx uarch. I corrected the issue title.

Ravenslofty:
(I have no experience with Xilinx architectural details, so I'd like your thoughts, Myrtle)

marzoul:
One thing I am not sure of, is how the module `__XILINX_SHIFTX` is supposed to behave in the cases where the number of mux inputs (size of input `A`) is actually smaller than the largest value that can be represented in the shift input `B`.

If that module is used for what is intended to be a SHIFT then we have to connect extra inputs to zero.
But if that module is used for a mux, it is not needed. This is why my implementation connects uses `x` as value for non-provided inputs.

Certainly some spec or doc would be added, related to how these modules should be used and the purpose of the various wrappers in several files, but I could not the full intent of the code.

Ravenslofty:
> One thing I am not sure of, is how the module `__XILINX_SHIFTX` is supposed to behave in the cases where the number of mux inputs (size of input `A`) is actually smaller than the largest value that can be represented in the shift input `B`.

Isn't that the point of [this](https://github.com/YosysHQ/yosys/blob/dbb977aa8b4fdc5f98b67259be6c2e0e21101896/techlibs/xilinx/mux_map.v#L60-L61)?

> If that module is used for what is intended to be a SHIFT then we have to connect extra inputs to zero.

but extra inputs being filled with zero is `$shift`, not `$shiftx`.

Ravenslofty:
I've had a look at how the Xilinx multiplexers work, and I think there's a missing piece of the puzzle here. I think the unwritten assumption is that the inputs to the multiplexer here are packed with the multiplexer itself; e.g. for a 2-bit mux being mapped into a MUXF7, the mux inputs come directly from the logic in question. That makes the mux cost nothing. Similarly for the MUXF78 type structures.

I think this PR makes sense in the present world where the xilinx uarch does not presently perform this optimisation, but if/when it did, we would then want to "optimise" it back to how it was originally. So...I'm really not sure about this.

marzoul:
> Isn't that the point of [this](https://github.com/YosysHQ/yosys/blob/dbb977aa8b4fdc5f98b67259be6c2e0e21101896/techlibs/xilinx/mux_map.v#L60-L61)?

Thanks for this note.
I'll investigate and make sure my implementation integrates into existing code.

> I think this PR makes sense in the present world where the xilinx uarch does not presently perform this optimisation, but if/when it did, we would then want to "optimise" it back to how it was originally. So...I'm really not sure about this.

I can provide more information on that observation.
- IMHO there will be no evolution towards giving these MUXFx dedicated slice inputs : the cost in switch boxes would be significant, extra configurable hard mux would need to be added between LUTs and these MUXFx to keep current functionality, which would increase device size and delays. They didn't even separate bits CI and AX in order to share expensive slice inputs.
- The 7-series family was replaced by the Ultrascale, which added MUXF9 in the same conditions : MUXF7 inputs come from LUT O6 outputs, hardwired (documentation is [UG574](https://docs.amd.com/r/en-US/ug574-ultrascale-clb)).
- The Ultrascale family is itself already replaced by Versal, in which all these MUXFx are no longer present (documentation is [AM005](https://docs.amd.com/r/en-US/am005-versal-clb)). From documentation, in LUTRAM section : "Wide function multiplexers (MUXF7, MUXF8, and MUXF9) are no longer implemented. Other LUTs are now used to implement wide multiplexing."

So it makes sense to improve the current code for currently supported technologies (7-series and Ultrascale).
Handling Versal family would require a complete rewrite anyway since there are seriously deep changes.

Note that better support for Ultrascale could be added as a follow-up to my proposal.
Currently Ultrascale model is constrainted to only what 7-series did support.

Ravenslofty:
> > I think this PR makes sense in the present world where the xilinx uarch does not presently perform this optimisation, but if/when it did, we would then want to "optimise" it back to how it was originally. So...I'm really not sure about this.
> 
> I can provide more information on that observation.
> 
>     * IMHO there will be no evolution towards giving these MUXFx dedicated slice inputs : the cost in switch boxes would be significant, extra configurable hard mux would need to be added between LUTs and these MUXFx to keep current functionality, which would increase device size and delays. They didn't even separate bits CI and AX in order to share expensive slice inputs.

You appear to have mistaken my commentary on "the nextpnr himbaechel uarch that targets Xilinx 7-Series" with "how Xilinx implement their FPGA". My point is that there is an inefficiency in nextpnr itself here, while the existing mapping was presumably written with Vivado or RapidWright in mind, which I assume packs the LUTs into the mux inputs.

>     * The 7-series family was replaced by the Ultrascale, which added MUXF9 in the same conditions : MUXF7 inputs come from LUT O6 outputs, hardwired (documentation is [UG574](https://docs.amd.com/r/en-US/ug574-ultrascale-clb)).

Which confirms my theory.

> So it makes sense to improve the current code for currently supported technologies (7-series and Ultrascale). Handling Versal family would require a complete rewrite anyway since there are seriously deep changes.

We agree there should be an improvement, but disagree on where that improvement should be. I believe it should be in nextpnr, not yosys.

> Note that better support for Ultrascale could be added as a follow-up to my proposal. Currently Ultrascale model is constrainted to only what 7-series did support.

You could always add that regardless of this PR.

Ravenslofty:
I think this PR is ultimately asking for an optimisation that should be performed in nextpnr, not yosys.

I think multiplexer mapping would be useful to think about in the context of Versal, but if one wants to add Versal as a target it would be best to start with more basic functionality. Feel free to email me at lofty@yosyshq.com if you want to discuss that.

marzoul:
Sorry I did not have time to answer right away.

> My point is that there is an inefficiency in nextpnr itself here

OK I didn't get that it was about nextpnr.
I get your point, I didn't expect that point of view, but why not.
This would revert my proposed changes for mux2 and mux4.

> I think this PR is ultimately asking for an optimisation that should be performed in nextpnr, not yosys.

No : my proposed changes for mux8 and mux12 are still valid improvements because current implementations already have LUTs + MUXFx so there is nothing better to expect from nextpnr. My patch brings the following reduction in LUT usage :
- mux8  : 4 -> 2 LUTs
- mux12 : 4 -> 3 LUTs

The case of muxes larger than 16 inputs should also be kept because, in case of inputs not multiple of 16, the upper sub-mux is extremely sub-optimal in current code and relies on subsequent logic opt in hope to recover (we don't recover much).

In an ideal world, I would _really_ like a Yosys option of some kind to choose a preferred implem for mux2 and mux4... precisely because nextpnr is not up the task of choosing. But I'm not sure what is the way preferred by current code maintainers to add such option, any hint would be appreciated.

> but if one wants to add Versal as a target it would be best to start with more basic functionality. Feel free to email me

I'm not on Versal, just precisely on 7-Series for now.
Depending on how my optims are received I will extend to improving Ultrascale+ support because the slice structure has many similarities (8-LUT slices and new muxf9).

marzoul:
So overall, I can revise my proposal, no problem.
Given that I am new in these parts of Yosys code I have more integration considerations to study before considering merging anyway. The topic really is worth it.

Ravenslofty:
> No : my proposed changes for mux8 and mux12 are still valid improvements because current implementations already have LUTs + MUXFx so there is nothing better to expect from nextpnr. My patch brings the following reduction in LUT usage :
> 
>     * mux8  : 4 -> 2 LUTs
> 
>     * mux12 : 4 -> 3 LUTs
> 
> 
> The case of muxes larger than 16 inputs should also be kept because, in case of inputs not multiple of 16, the upper sub-mux is extremely sub-optimal in current code and relies on subsequent logic opt in hope to recover (we don't recover much).

I think a patch for just these would be okay.
 
> In an ideal world, I would _really_ like a Yosys option of some kind to choose a preferred implem for mux2 and mux4... precisely because nextpnr is not up the task of choosing. But I'm not sure what is the way preferred by current code maintainers to add such option, any hint would be appreciated.

It's not obvious to me that yosys is a better place to decide that, at least, lacking things like delayed techmapping.

> > but if one wants to add Versal as a target it would be best to start with more basic functionality. Feel free to email me
> 
> I'm not on Versal, just precisely on 7-Series for now. Depending on how my optims are received I will extend to improving Ultrascale+ support because the slice structure has many similarities (8-LUT slices and new muxf9).

Sure, that would be reasonable.

PR #5203, opened 2025-06-30
CMake: Rebase From Main & Tests Working

nickrallison:

Got tests working with CMake after rebasing from main

1. Modified CMakeLists.txt files to be up-to-date with Makefiles, 
2. Updated flake.nix to add CMake to user development shell
3. Added tests to CMake to be ran with ctest
4. Added build script for build process to produce the same result as makefiles
    - In Future it may be worth getting CMake Presets functional

See build.sh, this file can likely be removed / placed elsewhere but it seemed reasonable to provide the build sequence used


KrystalDelusion:
Is rebasing from main necessary?  Because that's a whole lot of commits being included in the PR that aren't relevant, and are causing a whole lot of reviews to be requested. Also because you're not actually rebasing it, you're trying to merge the rebased changes back into the branch they came from (which is why it's so broken).

whitequark:
I think if it was rebased in a more skilled way the extra commits wouldn't be pulled in in first place.

KrystalDelusion:
I think the problem is because it's trying to apply commits from main to a branch (note the PR target)

nickrallison:
Yeah I'm sorry about that, based on a conversation in the Slack, it sounded as if rebasing was the way to go. I shall re-open after a merge instead

PR #5206, opened 2025-07-02
Gowin. BUGFIX. Fix multi-line descriptions.

yrabbit:
If let's say the enumeration of inputs took several lines, then all after the first one were ignored. Since the first line ended with a comma, an error was generated when trying to use the resulting file.

### DIFF
--- a/techlibs/gowin/cells_xtra.py
+++ b/techlibs/gowin/cells_xtra.py
@@ -11,6 +11,7 @@ import re
 class State(Enum):
     OUTSIDE = auto()
     IN_MODULE = auto()
+    IN_MODULE_MULTILINE = auto()
     IN_PARAMETER = auto()
 
 _skip = { # These are already described, no need to extract them from the vendor files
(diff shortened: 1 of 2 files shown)


widlarizer:
So the intention is just to reformat cells and support this reformat? Any particular reason you're also adding and changing some cells?

yrabbit:
> So the intention is just to reformat cells and support this reformat? Any particular reason you're also adding and changing some cells?

I'm working on Yosys/Nextpnr/Apicula support for the new Gowin family of chips - G5A series and the idea was to make the yosys files compile without errors :)



This current list of primitives is not very good in terms of syntax (see comma at the end):

https://github.com/YosysHQ/yosys/blob/99f7d79abb6a4514c141366237f510b1c1ec6679/techlibs/gowin/cells_xtra_gw5a.v#L1641-L1642



Changes in other cells are just a side effect of the fact that this file is generated from Gowin IDE files. The new family is supported by the new version of the IDE and the changes that Gowin has made to the primitives library are migrated to us, as these changes are massively related to silicon bug detection (like the failure to Read_Before_Write in DualPort mode*). 

And these silicon bug fixes will be implemented in nextpnr/apicula over time, but to do so we need to at least have an entire correct set of primitive descriptions for the specific current version of the Gowin IDE in use.

* Just an example, although such a fix did take place.

PR #5209, opened 2025-07-03
Start `opt_hier` to enable hierarchical optimization

povik:

Improve QoR in synthesis with preserved module boundaries.

Add `opt_hier` which propagates tied-together signals, tied-off signals, and unused signals both into and out of modules. The pass supports non-uniquified hierarchies in which case it propagates what's common to all instances.

See the attached test which verifies functional equivalence on few simple test sources.

### DIFF
--- a/passes/opt/opt_hier.cc
+++ b/passes/opt/opt_hier.cc
@@ -1,0 +1,470 @@
+/*
+ *  yosys -- Yosys Open SYnthesis Suite
+ *
+ *  Copyright (C)  Martin Povišer <povik@cutebit.org>
+ *
+ *  Permission to use, copy, modify, and/or distribute this software for any
+ *  purpose with or without fee is hereby granted, provided that the above
+ *  copyright notice and this permission notice appear in all copies.
+ *
+ *  THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ *  WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ *  MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ *  ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ *  WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ *  ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ *  OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *
+ */
+
+#include "kernel/register.h"
+#include "kernel/rtlil.h"
+#include "kernel/sigtools.h"
+
+USING_YOSYS_NAMESPACE
+PRIVATE_NAMESPACE_BEGIN
+
+// Used to propagate information out of a module
+struct ModuleIndex {
+	Module *module;
+	SigMap sigmap;
+	SigPool used;
+	dict<SigBit, SigBit> constant_outputs;
+	std::vector<SigSpec> tie_together_outputs;
+
+	ModuleIndex(Module *module)
+		: module(module), sigmap(module)
+	{
+		if (module->get_blackbox_attribute()) {
+			for (auto wire : module->wires()) {
+				for (auto bit : SigSpec(wire))
+					used.add(sigmap(bit));
+			}
+			return;
+		}
+
+		auto count_usage = [&](const SigSpec &signal) {
+			for (auto bit : signal)
+				used.add(sigmap(bit));
+		};
+		for (auto wire : module->wires()) {
+			if (wire->port_output) {
+				SigSpec wire1 = wire;
+				count_usage(wire1);
+			}
+		}
+		for (auto [_, process] : module->processes)
+			process->rewrite_sigspecs(count_usage);
+		for (auto cell : module->cells()) {
+			bool known = cell->known();
+			for (auto &conn : cell->connections()) {
+				if (!known || cell->input(conn.first))
+					count_usage(conn.second);
+			}
+		}
+
+
+		dict<SigBit, SigSpec> classes;
+		for (auto &pair : module->connections_) {
+			for (int i = 0; i < pair.first.size(); i++) {
+				if (pair.first[i].wire
+						&& pair.first[i].wire->port_output
+						&& !pair.first[i].wire->port_input) {
+					if (!pair.second[i].wire) {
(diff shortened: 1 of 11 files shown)


povik:
@KrystalDelusion I'm not sure what I need to do to fix the tests. I've changed the help message for `opt` and `synth` which looks related to the test failure.

KrystalDelusion:
It is using the help text, but only as a proxy for the list of sub-passes (usually the `help_script()` for `ScriptPass`es). The check is included for the macro commands to make sure that any mentions in text are updated to match.

In the case of `opt` you need to update [`docs/source/code_examples/macro_commands/opt.ys`](https://github.com/YosysHQ/yosys/blob/afcf125bc90917e1b693aa36d208cb21098624f0/docs/source/code_examples/macro_commands/opt.ys), and ideally add a heading to `opt.rst` for the new `opt_hier`.  There isn't a check for `synth`, so that one is fine.

KrystalDelusion:
If you don't mind, could you rebase on latest main so that CI runs tests with ASAN enabled?

I haven't looked at `opt_hier.cc`, but the rest of it looks fine.  I notice `synth -hieropt` is marked as EXPERIMENTAL, but `opt -hier` and `opt_hier` are not, is that intentional?

passes/opt/opt_hier.cc
```diff
@@ -0,0 +1,470 @@
...
+				if (!known || cell->input(conn.first))
+					count_usage(conn.second);
+			}
+		}
+
+
+		dict<SigBit, SigSpec> classes;
+		for (auto &pair : module->connections_) {
+			for (int i = 0; i < pair.first.size(); i++) {
+				if (pair.first[i].wire
```
widlarizer, line 70: Should there be a distinction between the first and second sigspec in a sigsig here? Connections are undirected and swapping the first and second should make no difference. Is there some code elsewhere this relies on?

passes/opt/opt_hier.cc
```diff
@@ -0,0 +1,470 @@
...
+			signal.replace(replacement_map);
+		};
+		module->rewrite_sigspecs(disconnect_rewrite);
+		for (auto chunk : disconnect_outputs.chunks()) {
+			log("Disconnected unused output terminal '%s' in module '%s'\n", log_signal(chunk), log_id(module));
+			did_something = true;
+			module->connect(chunk, SigSpec(RTLIL::Sx, chunk.size()));
+		}
+
+		// Connect constant inputs
```
widlarizer, line 355: generally having a long sequence of non-trivial operations that can be grouped with comments into distinct sections suggests they could be separated into functions. Consider breaking these up

povik:
> we could have a pass that duplicates instantiations, prjunnamed-style, to enable opt_hier to do its job more intensely

@widlarizer You may be thinking of `uniquify`.

passes/opt/opt_hier.cc
povik: > Connections are undirected and swapping the first and second should make no difference.

Has this question been settled? I can think of related threads https://github.com/YosysHQ/yosys/issues/3426 or https://github.com/YosysHQ/yosys/issues/3965.

I think part of the Yosys codebase considers connections undirected and another part directed. `opt_clean` normalizes the direction and that's why it doesn't come up as a problem as often.

povik:
> I notice synth -hieropt is marked as EXPERIMENTAL, but opt -hier and opt_hier are not, is that intentional?

@KrystalDelusion I should make this consistent. I'm all for not marking any of it as experimental. Let me know your thoughts.

passes/opt/opt_hier.cc
povik: > Is there some code elsewhere this relies on?

See https://github.com/YosysHQ/yosys/pull/3965#issuecomment-1752615754, there's e.g. `write_verilog` or `insbuf`

KrystalDelusion:
> > I notice synth -hieropt is marked as EXPERIMENTAL, but opt -hier and opt_hier are not, is that intentional?
> 
> @KrystalDelusion I should make this consistent. I'm all for not marking any of it as experimental. Let me know your thoughts.

Others might disagree, but for this to be a part of `opt` and not be experimental, I think it should be always-on instead of only being enabled with a flag.  Maybe experimental opt-in for a release or two just by virtue of being new and less tested in the real world and then remove the experimental flag and make it always-on?  That would also remove the need to update every `synth_*` pass with it, but might require some changes to `opt_hier` so that it has less overhead when it isn't needed.

povik:
> for this to be a part of opt and not be experimental, I think it should be always-on instead of only being enabled with a flag.

I'm not sure it can be enabled unconditionally as it changes the semantics of ports. Let's say you have a hierarchical design and a module `foo` which outputs the same signal on ports X and Y. `opt_hier` might rewrite an instantiation
```
foo bar(.X(x), .Y(y));
```
into
```
assign y = x;
foo bar(.X(x), .Y(unused));
```
In a later iteration `opt_hier` will see the `Y` output is unused and disconnect it, that is change the module such that it drives an x-state on the output.

These transformations can make simulation traces confusing, and stand in the way of correct application of design constraints. For that reason they might need to be opt-in.

povik:
I'm getting feedback that proprietary ASIC synthesizers rewrite ports freely as the default behavior though they also support to propagate SDCs accordingly.

nakengelhardt:
> Others might disagree, but for this to be a part of `opt` and not be experimental, I think it should be always-on instead of only being enabled with a flag.

Those two things are orthogonal IMO, we have plenty of options that only make sense in certain contexts but are not experimental (which given the noisy warning probably means something like "high likelihood that you will encounter bugs if you use this").

In this case, if there aren't any big caveats on the kind of design this'll work with, I think it doesn't need to be experimental. Immediately turning it on by default seems slightly too aggressive though, since it's likely to interfere with non-synthesis uses. Perhaps instead we could start with using the option by default in the `synth_*` passes where it's more likely to be a context where the user appreciates such optimization.

passes/opt/opt_hier.cc
widlarizer: After much pondering and experimentation and spec reading I see you're right

whitequark:
> These transformations can make simulation traces confusing, and stand in the way of correct application of design constraints. For that reason they might need to be opt-in.

The way to correctly apply design constraints is to do it on post-elaboration netlist before any optimization happens. That the current FOSS ecosystem cannot do this is a design flaw that should be rectified.

povik:
> The way to correctly apply design constraints is to do it on post-elaboration netlist before any optimization happens. That the current FOSS ecosystem cannot do this is a design flaw that should be rectified.

I think we can agree that would the best handling but in absence of support for this users might be relying on ports not getting optimized with Yosys as an anchor for constraints, as opposed to registers or nets which do get optimized. It might be fine to enable the optimization in synthesis as long as we provide a flag to disable it and/or add support for `(* keep *)` or `(* dont_touch *)` on ports.

povik:
I have dropped the experimental label and kept it optional.

PR #5219, opened 2025-07-10
dfflibmap: fix dffe inference when spaces used for ANDs in liberty next_state

widlarizer:
#4698 introduced inference of flops with enables by parsing `next_state` expressions. Contextually interpreting whitespace as an AND operator (vs as meaningless) was left as a TODO. `D EN` gets interpreted as a reference to a `DEN` variable, rather than equivalent `D&EN`. With the test change, `tests/techmap/dfflibmap.ys` fails, and this debug message explains why:

`Inference failed on expression '(DEN)|(IQ!EN)' in next_state attribute of cell 'dffe' because it does not evaluate to an enable flop - skipping.`

This PR expands compatibility with more PDKs by more strictly adhering to the Liberty file format.

More specifically: spaces are disambiguated between whitespace spaces and AND spaces. First, spaces are no longer filtered out in dfflibmap `parse_next_state`. Then in `LibertyExpression::parse`, I keep the space skipping loop at the start of expressions, because in that context, it can never be an AND. Later, in the tail loop, I first treat a sequence of spaces the same as a single space, and I consider the space equivalent to a proper AND, unless it's succeeded by a proper binop. If it is succeeded by a proper binop, it is skipped and treated as whitespace. This should consistenly disambiguate between ANDs and whitespace according to the standard. Also for dev purposes I've added an S-expression stringifier for LibertyExpression

- [x] add some more tests using the stringifier

### DIFF
--- a/tests/unit/techmap/libparseTest.cc
+++ b/tests/unit/techmap/libparseTest.cc
@@ -1,0 +1,88 @@
+#include <gtest/gtest.h>
+#include "passes/techmap/libparse.h"
+
+YOSYS_NAMESPACE_BEGIN
+
+namespace RTLIL {
+
+	class TechmapLibparseTest : public testing::Test {
+	protected:
+		TechmapLibparseTest() {
+			if (log_files.empty()) log_files.emplace_back(stdout);
+		}
+		void checkAll(std::initializer_list<std::string> expressions, std::string expected) {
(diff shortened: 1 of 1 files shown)


Ravenslofty (approved):

mikesinouye:
Verified this works for our case, thanks @widlarizer

Issue #5010, opened 2025-04-09
memory pass is not respecting (* keep *)


KelvinChung2000:
### Version

Yosys 0.51+85 (git sha1 1113c8c95, g++ 11.4.0-1ubuntu1~22.04 -fPIC -O3)

### On which OS did this happen?

Linux

### Reproduction Steps

With the following module:

```verilog
module comb_mem_d1 #(
    parameter WIDTH = 32,
    parameter SIZE = 16,
    parameter IDX_SIZE = 4
) (
   input wire                logic [IDX_SIZE-1:0] addr0,
   input wire                logic [ WIDTH-1:0] write_data,
   input wire                logic write_en,
   input wire                logic clk,
   input wire                logic reset,
   (* keep *)output                    logic [ WIDTH-1:0] read_data,
);

  (* keep *)logic [WIDTH-1:0] mem[SIZE-1:0];

  /* verilator lint_off WIDTH */
  assign read_data = mem[addr0];

  always_ff @(posedge clk) begin
    if (!reset && write_en)
      mem[addr0] <= write_data;
  end

  // Check for out of bounds access
  `ifdef VERILATOR
    always_comb begin
      if (addr0 >= SIZE)
        $error(
          "comb_mem_d1: Out of bounds access\n",
          "addr0: %0d\n", addr0,
          "SIZE: %0d", SIZE
        );
    end
  `endif
endmodule

module test();
  logic [3:0] addr0;
  logic [31:0] write_data;
  logic write_en;
  logic clk;
  logic reset;
  logic [31:0] read_data1;
  logic [31:0] read_data2;
  logic done;

  assign write_en = 1;

  (* keep=1 *)
  comb_mem_d1 #(
    .WIDTH(32),module comb_mem_d1 #(
    parameter WIDTH = 32,
    parameter SIZE = 16,
    parameter IDX_SIZE = 4
) (
   input wire                logic [IDX_SIZE-1:0] addr0,
   input wire                logic [ WIDTH-1:0] write_data,
   input wire                logic write_en,
   input wire                logic clk,
   input wire                logic reset,
   (* keep *)output                    logic [ WIDTH-1:0] read_data,
);

  (* keep *)logic [WIDTH-1:0] mem[SIZE-1:0];

  /* verilator lint_off WIDTH */
  assign read_data = mem[addr0];

  always_ff @(posedge clk) begin
    if (!reset && write_en)
      mem[addr0] <= write_data;
  end

endmodule

module test();
[...]

PR #5222, opened 2025-07-11
opt_mem: Adding external-init to opt_mem

KelvinChung2000:
The aim is to add a way to prevent memory that has never been written/initialised by the design itself, but might be initialised by the external (CPU/DMA writing to the memory directly in a SoC system).  

A new flag is added for the `memory` pass and `opt_mem` pass. At the stage of removing lanes, if the flag is presented, it will add the lane to `swizzle` and preserve the lane.

### DIFF
--- a/passes/memory/memory.cc
+++ b/passes/memory/memory.cc
@@ -31,7 +31,7 @@ struct MemoryPass : public Pass {
 	{
 		//   |---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|
 		log("\n");
-		log("    memory [-norom] [-nomap] [-nordff] [-nowiden] [-nosat] [-memx] [-no-rw-check] [-bram <bram_rules>] [selection]\n");
+		log("    memory [-external-init] [-norom] [-nomap] [-nordff] [-nowiden] [-nosat] [-memx] [-no-rw-check] [-bram <bram_rules>] [selection]\n");
 		log("\n");
 		log("This pass calls all the other memory_* passes in a useful order:\n");
 		log("\n");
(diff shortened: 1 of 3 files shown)


widlarizer:
Here are the ordinary approaches that should cover your use case without this code change:

If you're running synthesis and Yosys is inferring a `$mem` from your Verilog, it means it can objectively see that there is no other initialization going on than can be handled. I expect real world designs with some (maybe externally influenced) memory blocks to add those blocks after synthesis is done and connect them in place of module instance black boxes or to top-level memory buses

If you want to run formal or simulation on such a memory, you would hook up to your memory interface an extra model module that isn't part of your design otherwise

Let us know more about what you're doing if these don't cover it afterall

KelvinChung2000:
I am considering a memory that has two memory interfaces, but the synthesis/device can only see one of the memory interfaces. So even though nothing is written/into the memory in the design, it is still receiving content from the other interface. So one way to think about it is that the memory block is the system bus itself. 

I can attach a black box to the write interface of the memory block to prevent it from getting optimised away. But, the design I have is a high-level synthesis-generated design, and adding/modifying from that side is a bit more difficult than asking yosys not to do something. I have tried to add `keep` to the memory instances, but this led to #5010, which resulted in me implementing this

widlarizer:
In the example you have in #5010, what is the purpose of giving the definition of `module Mem` to yosys? If it is something you don't want to fully model in Verilog, then you have the freedom of just declaring `module Mem` as a blackbox. This is what I was suggesting, rather than connecting a blackbox to a write interface of an instantiated memory

KelvinChung2000:
I don't want to make the whole `module Mem` a black box if possible, as there are some logics within the module which might be redundant and can be optimised away, such as the handshaking it provides.

widlarizer:
Blackboxing just the memory part of the model would be the way to go. I don't think supporting quirks of Verilog generator tools should be within scope for Yosys. That isn't to say `keep` attributes shouldn't be handled differently as discussed in #5010, I suppose they should, but this PR doesn't deal with that

PR #5233, opened 2025-07-16
Assumptions for equiv_*

### LINKED ISSUE #5196: equiv_* doesn't respect $assume
### Version

main

Linux

### Reproduction Steps

Originally analyzed by @KrystalDelusion 
```tcl
read_verilog -sv <<EOT
module gold (input D, output Q);
assign Q = '0;
endmodule

module gate (input D, output Q);
assume property (D == '0);
assign Q = D;
endmodule
EOT

stat gate
chformal -lower
stat gate
async2sync

equiv_make gold gate equiv
#opt_clean -purge equiv
#show equiv
equiv_simple equiv
#equiv_induct equiv
equiv_status -assert equiv
```

Image

### Expected Behavior

Equivalence proven

### Actual Behavior

Equivalence unproven. This is surprising since equiv_* passes use SatGen infrastructure which should construct the corresponding operations for ezsat

KrystalDelusion:

Fix #5196.

- [x] Add `-set-assumes` option to `equiv_simple`.
- [x] Also the other `equiv_*` passes (just `equiv_induct`?).

Based on `-set-assumes` option from `sat` pass (also because that is an option rather than default behavior).

I initially wasn't convinced that the `$assume` cells would work with FFs because of the `seed_a` `next_seed_a` stuff going on, but I think actually the implementation I did of *not* checking through the FFs and instead only testing the inputs of the `$assume` is fine.  It probably could use some tidy up though since it is a bit of a mess just because of how it does the same thing for both the A and B input cones.

### DIFF
--- a/kernel/satgen.h
+++ b/kernel/satgen.h
@@ -64,7 +64,7 @@ struct ezSatPtr : public std::unique_ptr<ezSAT> {
 struct SatGen
 {
 	ezSAT *ez;
-	SigMap *sigmap;
+	const SigMap *sigmap;
 	std::string prefix;
 	SigPool initial_state;
 	std::map<std::string, RTLIL::SigSpec> asserts_a, asserts_en;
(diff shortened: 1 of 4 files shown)


widlarizer (approved):
Nice work!

PR #5248, opened 2025-07-22
Use unordered_map instead of dict for IdString's char* to index storage.

calewis:
`dict` is pretty slow when you don't ever need to iterate the container and the hash function for `char*` in `dict` hashes for every single byte in the string, likely doing significantly more work than `std::hash`.

## What are the reasons/motivation for this change? 

A private flatten benchmark shows that this change provides a 2-4x improvement in `IdString`s constructor:
Screenshot 2025-07-22 at 3 56 41 PM
Screenshot 2025-07-22 at 3 57 14 PM

### DIFF
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -35,7 +35,7 @@ YOSYS_NAMESPACE_BEGIN
 bool RTLIL::IdString::destruct_guard_ok = false;
 RTLIL::IdString::destruct_guard_t RTLIL::IdString::destruct_guard;
 std::vector<char*> RTLIL::IdString::global_id_storage_;
-dict<char*, int> RTLIL::IdString::global_id_index_;
+std::unordered_map<std::string_view, int> RTLIL::IdString::global_id_index_;
 #ifndef YOSYS_NO_IDS_REFCNT
 std::vector<int> RTLIL::IdString::global_refcount_storage_;
 std::vector<int> RTLIL::IdString::global_free_idx_list_;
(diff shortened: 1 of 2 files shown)


whitequark:
> and the hash function for `char*` in `dict` hashes for every single byte in the string, likely doing significantly more work than `std::hash`.

Should this not be improved instead? That sounds like it would improve more code paths than just this one.

QuantamHD:
It looks like dict<char*, ...> is only used for IdString's global cache. Everything else is probably using IdString directly so you would never see the characters.

https://github.com/search?q=repo%3AYosysHQ%2Fyosys%20dict%3Cchar*&type=code

calewis:
> Should this not be improved instead? That sounds like it would improve more code paths than just this one.

Even fixing the hash issue I think dict would be slower due to it's orderable nature.  That orderableness isn't needed in IdString.  

Also implementing a good hash is a lot more finicky than swapping containers.  I was comparing Yosys' hash with Abseil's and while Abseil's will hash up to 64 bytes at a time, the implementation is significantly more complicated.   You can see that libc++'s is nontrival as well https://github.com/llvm/llvm-project/blob/15715f4089878640405ac897723f24b30e5f8e03/libcxx/include/__functional/hash.h#L85.  I'd rather specialize the types in Yosys for std::hash than try to match the hash implementation in Yosys with libc++/Abseil.

whitequark:
> I'd rather specialize the types in Yosys for std::hash than try to match the hash implementation in Yosys with libc++/Abseil.

That's what I would suggest yeah.

PR #5255, opened 2025-07-28
Reapply "Add groups to command reference"

KrystalDelusion:

#4860 was reverted from main so that it could be reviewed more.

Reapply #4860, and make any other changes needed.


kernel/register.h
```diff
@@ -23,27 +23,62 @@
 #include "kernel/yosys_common.h"
 #include "kernel/yosys.h"
 
+#include <version>
+#if __cpp_lib_source_location == 201907L
```
povik: Before this was reverted it caused trouble linking a c++20 codebase to Yosys built under c++17 (showed up on yosys-slang https://github.com/YosysHQ/oss-cad-suite-build/actions/runs/16460314227/job/46526151437).

I believe in principle linking across standards should be fine but here we create an explicit ABI breakage.

KrystalDelusion: Would an explicit define to enable it be better than a compiler check here?  This feature is currently only used in the generated rst documentation, it shouldn't matter for regular builds of Yosys.

povik: I'm happy with any solution as long as the signature of `Pass::Pass()` isn't compiler dependent. An explicit define sounds good.

KrystalDelusion:
Rebased on 0.55+194

KrystalDelusion:
@DanielG FYI for building docs with Yosys >= v`0.56` since IIRC you're building Debian release with docs included.  This PR (re-)introduces changes to how the command reference is formatted (currently live on the [latest docs](https://yosyshq.readthedocs.io/projects/yosys/en/latest/cmd/index_passes.html)).  From the readme: 

> It is recommended to use the `ENABLE_HELP_SOURCE` make option for Yosys builds that will be used to build the documentation.  This option enables source location tracking for passes and improves the command reference through grouping related commands and allowing for the documentation to link to the corresponding source files.

The fallback behaviour without this make option set is that none of the passes will have the "Help text automatically generated from ..." note, and that *most* of them will end up in the "unknown" group (rendered on `cmd/index_other`).  Command groups with no associated passes (currently `cmd/index_passes_{hierarchy,techmap,sat}`) will display a warning like this:
image
It should still be usable, but may be confusing.

Note that the `ENABLE_HELP_SOURCE` option does still require a compiler with `source_location` support, but it tries both `<source_location>` and `<experimental/source_location>` so a reasonably modern gcc/clang should work fine.

DanielG:
Thanks for the heads up Krystal! I made a note in the BTS. Updating is a
hassle until Trixie is out the door :-)

--Daniel

PR #5261, opened 2025-07-31
docs: mention Yosys Discourse group and blog

mattyoung101:
Since the Slack has been deprecated, I thought it might be a good idea to direct users to the Discourse in the readme. I couldn't find any other direct links to it on the Yosys website or elsewhere. This is similar for the blog too.

Happy to adjust wording if needed.

My editor has removed some trailing whitespace later in the file, that is the reason for those diffs.


mlyoung101:
My fork is severely outdated, I'm going to redo this

PR #5262, opened 2025-07-31
docs: mention Yosys Discourse group and blog

mattyoung101:
Since the Slack has been deprecated, I thought it might be a good idea to direct users to the Discourse in the readme. I couldn't find any other direct links to it on the Yosys website or elsewhere. This is similar for the blog too.

Happy to adjust wording if needed.

My editor has removed some trailing whitespace later in the file, that is the reason for those diffs.

_(this is the same as #5261, but my fork was severely outdated, so I've redone the PR to clean things up)_

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,11 +1,8 @@
 contact_links:
-  - name: Discussions
-    url: https://github.com/YosysHQ/yosys/discussions
-    about: "Have a question? Ask it on our discussions page!"
-  - name: Community Slack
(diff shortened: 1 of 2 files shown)


KrystalDelusion:
Would you mind updating the links in the [issues popup](https://github.com/YosysHQ/yosys/issues/new/choose) too? https://github.com/YosysHQ/yosys/blob/main/.github/ISSUE_TEMPLATE/config.yml

mlyoung101:
No worries. I removed the Slack and replaced it with Discourse. I didn't actually know about the Discussions tab, and I wasn't sure if we wanted to centralise things to the Discourse or not, so I've also removed that and redirected everyone to the Discourse group.

Issue #4006, opened 2023-10-12
Import package error


sifferman:
### Version

Yosys 0.34+14 (git sha1 3e2279181, clang 10.0.0-4ubuntu1 -fPIC -Os)

### On which OS did this happen?

Linux

### Reproduction Steps

Thanks for this tool!

I found an error with package imports. (Similar to #2187)

I have a package `"my_pkg.sv"`, and two tests `"test1.sv"` and `"test2.sv"`:

```systemverilog
package my_pkg;
    localparam NUM = 1;
endpackage
```
```systemverilog
module test1 import my_pkg::*; (
    input a,
    output b
);
    assign b = (a ^ NUM);
endmodule
```
```systemverilog
module test2 (
    input a,
    output b
);
    import my_pkg::*;
    assign b = (a ^ NUM);
endmodule
```

This is my tcl script:

```shell
# both these cause and error:
read_verilog -sv my_pkg.sv testX.sv
# or...
read_verilog -sv my_pkg.sv
read_verilog -sv testX.sv
```

### Expected Behavior

This is valid syntax for VCS, Icarus, and Verilator

### Actual Behavior

Both files give this error:

```log
Executing Verilog-2005 frontend: test2.sv
Parsing SystemVerilog input from `test2.sv' to AST representation.
test2.sv: ERROR: syntax error, unexpected TOK_PACKAGESEP, expecting '(' or '['
```

zachjs:
For those still facing this issue, https://github.com/zachjs/sv2v supports these imports.

PR #5265, opened 2025-08-04
Fix package import

bhagwat-rahul:

Add support for a _newer_ system verilog feature, declaring common packages and importing, attempts to fix [issue](https://github.com/YosysHQ/yosys/issues/4006)

Added TOK_IMPORT token, import_stmt grammar rule, AST_IMPORT node type, and import processing logic that looks for packages in both current AST and design context to support cross-file imports.

Test with:-
`make -f run-test.mk verilog-package_import_separate.ys`
this tests package imports from different files


tests/verilog/package_import_separate.sv
```diff
@@ -0,0 +1,13 @@
...
+    ADDR_WIDTH = 4;
+    
+  localparam logic [2:0]
+    IDLE = 3'b000,
+    START = 3'b001,
+    DATA = 3'b010,
+    ODD_PARITY = 3'b011,
+    STOP = 3'b100,
+    DONE = 3'b101;
+endpackage
```
KrystalDelusion: Test files have an extra space at the end instead of an empty new line.

frontends/ast/simplify.cc
```diff
@@ -1103,6 +1103,74 @@ bool AstNode::simplify(bool const_fold, int stage, int width_hint, bool sign_hin
...
+					for (auto &design_child : current_ast->children) {
+						if (design_child->type == AST_PACKAGE) {
+							if (design_child->str == child->str) {
+								package_node = design_child;
+								break;
+							}
+						}
+					}
+				}
+
```
KrystalDelusion: Empty lines shouldn't have whitespace

tests/verilog/package_import_separate.ys
KrystalDelusion, line 1: This one is still missing an empty new line at the end

KrystalDelusion (changes requested):

bhagwat-rahul:
> It looks like this only supports wildcard imports, do you think you would be able to add support for the following:
> 
> ```
> import config_pkg::DATA_WIDTH;
> import config_pkg::IDLE;
> 
> module top;
>   logic [DATA_WIDTH-1:0] data;
>   logic [3:0] addr;
>   logic [2:0] state;
>   
>   always_comb begin
>     case (state)
>       IDLE: data = 8'h00;
>       default: data = 8'hFF;
>     endcase
>   end
> endmodule
> ```
> 
> It doesn't have to be in this PR, but I think it would make sense to also allow specific imports instead of just wildcard.
> 
> For this PR I think the only thing required is the whitespace issues.

will add support for this in a follow up pr

PR #5270, opened 2025-08-07
Reorder the port wire declarations to follow the same order of the port declarations

zhanghongce:

In our applications, we often dump Verilog using `write_verilog`, make some small changes, and then use `read_verilog` to read it back to yosys. We found that the port ordering would change in this process (even when we make no changes). This creates a lot of problems for us. We conjecture that this is because `write_verilog` does not dump port wire declarations according to the order as the port declaraction part. So we make this change and find it solves our problem and we hope this may help others as well.

We add a vector to remember the order of signals when dumping the port declarations with `module` and then first dump these port wires following the order and then dump the other wires.

You may dump some verilog and then read it back and see if the ports now follow the same order.


backends/verilog/verilog_backend.cc
```diff
@@ -2358,12 +2358,16 @@ void dump_module(std::ostream &f, std::string indent, RTLIL::Module *module)
 	dump_attributes(f, indent, module->attributes, "\n", /*modattr=*/true);
 	f << stringf("%s" "module %s(", indent.c_str(), id(module->name, false).c_str());
 	int cnt = 0;
+	
+	std::vector<Wire *> port_ordered;
```
jix: There is no need to create a vector of `Wire *` when `module->ports` has the vector of names.

backends/verilog/verilog_backend.cc
```diff
@@ -2374,8 +2378,16 @@ void dump_module(std::ostream &f, std::string indent, RTLIL::Module *module)
 		f << indent + "  " << "reg " << id(initial_id) << " = 0;\n";
 	}
 
-	for (auto w : module->wires())
+	// first dump input / output according to port_ordered;
+	for (auto w : port_ordered)
```
jix: you can use `for (auto port : module->ports)` and then `dump_wire(f, indent + "  ", module->wire(port));`

backends/verilog/verilog_backend.cc
```diff
@@ -2374,8 +2378,16 @@ void dump_module(std::ostream &f, std::string indent, RTLIL::Module *module)
...
 		dump_wire(f, indent + "  ", w);
+	
+	for (auto w : module->wires()) {
+		// avoid duplication
+		if (std::find(port_ordered.begin(), port_ordered.end(), w) != port_ordered.end())
```
jix: This does a linear scan over all ports for every single wire of a module, this can waste a lot of time on modules with 100s of ports. You can use `if (w->port_id)` instead as every wire already knows whether it is a port or not.

jix (changes requested):
> In our applications, we often dump Verilog using write_verilog, make some small changes, and then use read_verilog to read it back to yosys.

We advise against doing that, Verilog is not a suitable format if you want to write out a design and load it back into yosys, it's best to use the native RTLIL format or JSON format for this.

> We found that the port ordering would change in this process (even when we make no changes). This creates a lot of problems for us.

We make no guarantees about the order of module items that `write_verilog` produces. The order of ports in the module header is meaningful in Verilog, but the order of module items makes no difference for the kind of Verilog that yosys produces. If this causes issues, that is most likely a bug in whatever code is used to process the Verilog.

> So we make this change and find it solves our problem and we hope this may help others as well.

While I'd advice against depending on this for any automated processing, I do think that having the port declarations at the beginning of the output makes it more readable when manually inspecting the output, so I agree with the change you want to make, just for different reasons.

There is a performance issue with the current implementation that needs to be fixed before we can merge this, though, for details see the comments I left on the code changes.

backends/verilog/verilog_backend.cc
zhanghongce: Got it. Thanks!

backends/verilog/verilog_backend.cc
zhanghongce: Sure. Thanks!

backends/verilog/verilog_backend.cc
zhanghongce: Fixed.

zhanghongce:
Hi @jix  

Many thanks for your comments! I fixed these problems with a new commit. Could you please have a look and see if there is anything more I should take care of? 

Thank you for your time!

jix (approved):

PR #5273, opened 2025-08-07
equiv_simple: refactor

widlarizer:
Deduplication, splitting things up, renaming things, structs modeling the domain instead of infinite arg lists

Additional damage: sigmap in satgen is now a const pointer, which [nobody will mind](https://github.com/search?q=%2F%28%3F-i%29SatGen+%2F+-path%3Asatgen.h+-path%3Asatgen.cc+path%3A*.cc+-path%3Aequiv_*+-path%3Aeval*+-path%3Afreduce*+-path%3Amemory_*+-path%3Atest_*+-path%3Ashare.cc+-path%3Asat.cc+NOT+is%3Afork&type=code)

Feel free to suggest changes to the comments if I got anything wrong. Functionality should be completely unaffected

### DIFF
--- a/passes/equiv/equiv_simple.cc
+++ b/passes/equiv/equiv_simple.cc
@@ -213,7 +213,7 @@ struct EquivSimpleWorker
 			assume_seed.insert(model.sigmap(assume->getPort(ID::A)).as_bit());
 			assume_seed.insert(model.sigmap(assume->getPort(ID::EN)).as_bit());
 
-			for (auto& cone : std::array{cone_a, cone_b}) {
+			for (auto& cone : {cone_a, cone_b}) {
 				Cone assume_cone;
 				ConeFinder{model, dummy_next_seed, assume_cone, cone, &overlap_bits}
 					.find_input_cone(assume_seed);


KrystalDelusion:
> sigmap in satgen is now a const pointer

Isn't that a pointer to a const rather than a const pointer (`Sigmap * const sigmap`)?  Demonstrated by `setContext` changing where the pointer is pointing.  But also, what does having the sigmap be const achieve, is it that you can only use the `const` methods and not change the database itself?

KrystalDelusion:
That is a lot easier to follow now, and matches with my understanding of how the pass worked.

The VS build doesn't like the `std::array`, so I've pushed an attempted fix.

KrystalDelusion (approved):

widlarizer:
Thanks, the initializer list is neater too, and I did mean a pointer to const. The purpose of the constness is to constrain underconstrained code to make it clearer what the intention is. The primary problem I was dealing with by refactoring was the pattern where utility types carry objects some of which are mutable output or temporaries and others are just inputs, and I wanted to make that distinction. I intend this to make it easier for a reader to answer "so what does this do exactly?" by making it clear some of the things it's not doing (modifying the sigmap it received)

PR #5081, opened 2025-05-01
autoname: Check for potential overflow

KrystalDelusion:
_What are the reasons/motivation for this change?_
Fix #4983
While unlikely, a wire with a score greater than about `INT32_MAX/20000` will lead to an interger overflow when trying to rename users of that wire.

_Explain how this is achieved._
Call `log_error` if any of the wires have a score over `INT32_MAX/24000`.  Because the score is calculated with `(10000*score + new_name.size())*2`, if the new name is longer than 2000 characters long then there could still be an integer overflow, but this seems sufficiently unlikely that I thought checking each wire score once was better than checking each proposal, while still allowing for a fanout of almost 90000 bits from a single wire.

_If applicable, please suggest to reviewers how they can test the change._

Issue #4983, opened 2025-04-02
An integer overflow occurred during the `Executing AUTONAME pass`.


sdjasj:
### Version

Yosys 0.50+49 (git sha1 05c81b3f1, clang++ 14.0.0-1ubuntu1.1 -Og -fPIC -O1 -fno-omit-frame-pointer -fno-optimize-sibling-calls -fsanitize=address,undefined)

### On which OS did this happen?

Linux

### Reproduction Steps

The following is the Verilog code file `bug.v` that triggers the bug. I have already minimized it as much as possible.
```
module a(input b, input [83 : 10] c, d, e, output f);
    reg g;
    always @(*) begin
        f <= g;
        g = d / b;
        if (d / g) begin
            g = b ? b / e / c : 0;
        end
    end 
endmodule
```
Running the following script will trigger the bug.
```
yosys -p "read_verilog bug.v; synth_nexus"
```
I discovered this bug while testing Yosys using a fuzzing tool I am developing.  While reviewing historical issues, I found that [https://github.com/YosysHQ/yosys/issues/4509](https://github.com/YosysHQ/yosys/issues/4509) reported a similar problem. However, the code snippet that caused the error in that issue is different from the one in the current case, so I believe they might be different issues. Specifically, the code snippet that leads to the integer overflow is in `passes/cmds/autoname.cc:80:17`: 
```cpp
for (auto &it : proposed_wire_names) {
	if (best_score*2 < it.second.first)
		continue;
	IdString n = module->uniquify(IdString(it.second.second));
	log_debug("Rename wire %s in %s to %s.\n", log_id(it.first), log_id(module), log_id(n));
	module->rename(it.first, n);
}
```
The issue was triggered by the line `if (best_score*2 < it.second.first)`.

### Expected Behavior

The synthesis process should either succeed or fail due to an error; integer overflows and other abnormal behaviors (such as OOM) should not occur.

### Actual Behavior

The synthesis process hangs during `Executing AUTONAME pass`. After some time, it triggers a runtime error:  
`passes/cmds/autoname.cc:80:17: runtime error: signed integer overflow: 1136330020 * 2 cannot be represented in type 'int'`.  
Later, it results in:  `AddressSanitizer: Out of memory. The process has exhausted 65536MB for size class 32`.  
The detailed log is as follows:
```
2.49. Executing SETUNDEF pass (replace undef values with defined constants).

2.50. Executing HILOMAP pass (mapping to constant drivers).
[...]

PR #5277, opened 2025-08-08
autoname: Avoid integer overflow

KrystalDelusion:

> I wouldn't expect a pass like `autoname` to produce a fatal error. While that is certainly better than undefined behavior, so I'm not opposed to merging this as a first step, how difficult would it be to make `autoname` generate slightly less exhaustive names when they would otherwise become to long?

_Originally posted by @jix in https://github.com/YosysHQ/yosys/issues/5081#issuecomment-3152198240_

Alternative approach to #5081 which leaves the score as-is and only compares the name length as necessary. It also changes the wire score from `int` to `unsigned int` so that instead of overflowing if the score is over `INT32_MAX/20000`, the wire fanout now needs to be over `INT32_MAX`, and is never dependent on the name length.

This avoids the integer overflow and may be preferable, but the test case from #4983 still crashes.

Also includes a fix for `autoname` to support selection arguments, matching the behaviour in the help message.  And a fix for the printed number of renamed objects not matching the actual number of renamed objects.

### DIFF
--- a/tests/various/autoname.ys
+++ b/tests/various/autoname.ys
@@ -175,6 +175,8 @@ module \top
   end
 end
 EOT
+design -save order_test
+
 # wires are named for being cell outputs
 logger -expect log "Rename wire .d in top to or_Y" 1
 logger -expect log "Rename cell .name2 in top to or_Y_.or_B" 1
@@ -186,11 +188,18 @@ logger -check-expected
 logger -expect log "Rename cell .name in top to a_.__unknown_A" 1
 # another output wire
 logger -expect log "Rename wire .e in top to or_Y_.or_B_Y" 1
+logger -expect log "Renamed 4 objects" 1
+debug autoname
+logger -check-expected
+
+# don't rename prematurely (some objects should be named after $name2)
+design -load order_test
+
+# $c gets shortest name, since the cell driving it doesn't have known port
+# directions (otherwise a_$__unknown_A_Y)
+logger -expect log "Rename wire .c in top to or_Y_.or_B_A" 1
 # $name3 named for lowest fanout wire (otherwise a_$__unknown_A_Y_$and_A)
 logger -expect log "Rename cell .name3 in top to or_Y_.or_B_Y_.and_B" 1
-# $c gets shortest name, since the cell driving it doesn't have known port
-# directions
-logger -expect log "Rename wire .c in top to or_Y_.or_B_A" 1
-logger -expect log "Renamed 4 objects" 1
+logger -expect log "Renamed 6 objects" 1
 debug autoname
 logger -check-expected


KrystalDelusion:
Nix on macos failure should be resolved by #5276

widlarizer (approved):
I am confident that this fixes what it says it fixes and keeps everything else the same. But I found something strange - to try things out, I commented out the parts I didn't understand:
```diff
-       best_name.score *= 2;
+       // best_name.score *= 2;

        for (auto &it : proposed_cell_names) {
-               if (best_name < it.second)
-                       continue;
+               // if (best_name < it.second)
+               //      continue;
                IdString n = module->uniquify(IdString(it.second.name));
                log_debug("Rename cell %s in %s to %s.\n", log_id(it.first), log_id(module), log_id(n));
                module->rename(it.first, n);
        }

        for (auto &it : proposed_wire_names) {
-               if (best_name < it.second)
-                       continue;
+               // if (best_name < it.second)
+               //      continue;
```
This didn't make the test fail, which would be nice to fix. What I noticed is that before I made this change, the test log said this:
```
11. Executing AUTONAME pass.
Rename wire $e in top to or_Y_$or_B_Y.
Rename cell $name in top to a_$__unknown_A.
Rename cell $name3 in top to or_Y_$or_B_Y_$and_B.
Rename wire $c in top to or_Y_$or_B_A.
Renamed 6 objects in module top (3 iterations).
```
And with the change, it went to 4 objects instead which is what I would expect. This is in comparison to main, where it reports 8 objects if I try the new test on it (with all the logger checks disabled). Do you know why?

KrystalDelusion:
So the way I understand it is that we are trying to minimise the wire score (lower score means less re-use, therefore more specific and more unique names).  Each iteration only proposes names that come from already named cells/wires, which means there could be a "better" name available after we apply some of the proposed names.  

The current approach (which I haven't changed) is to check against double the best score.  If the best name is more than twice as good as the current proposed name, we delay checking until the next iteration.  Since the best score will always be assigned, the next iteration can now use that newly named object for naming other cells/wires.  In the worst case this can however mean that only 1 name gets applied in each iteration, and there is no logic to check if a given proposal could actually be affected by another proposal being applied (meaning that it can take extra iterations without producing a different result).  I think the way to do it would be that each wire/cell checks _all_ of the connections, recording the best connection _for that wire/cell_ separated into public/private scores, and then will only skip if there is a better private score and it isn't the best score available.

The object count not matching how many actually get renamed is because the function returns the number of _proposed_ names, rather than the number of renames.  So if there is more than 1 iteration the number reported will almost always be higher than the actual number (unless every proposal is applied in each iteration, and the remaining cells/wires can _only_ be named by cells/wires in the previous iteration), which is a mistake and should be fixed.

KrystalDelusion:
I added a new check, using the same circuit that was used in the previous final check but without pre-running `autoname t:$or`.  If the best score skip is disabled, the names are assigned differently and the check fails (and it takes one less iteration, but that doesn't get checked).  My installed Yosys (0.57+7) gives names consistent with the skip behaviour (provided I disable the other checks which are expected to fail without this PR).

widlarizer:
I see what you mean. Instead of this heuristic mess that's pretty scary about recomputation, do you think the proposed names could be a std::map so you always propagate autonames from the best name you haven't applied yet?

KrystalDelusion:
I'm not sure I follow exactly what you mean, but I'm sure there are many ways that would be better than the current approach.  I spun it out into a separate issue #5394 if you want to continue the conversation there.

PR #5280, opened 2025-08-12
Workflow adjustments

KrystalDelusion:

1. Currently (most) workflows trigger on both `push` and `pull_request` which often results in duplicated jobs when maintainers push commits to this repo that are also a part of a PR.  Concurrent skipping was enabled to reduce this duplicated work, but this has problems of its own and has resulted in a few PRs being merged to main and then failing.
2. Building and running the sanitizer checks (especially the address sanitizer) takes a long time, slowing down the whole process and potentially holding up job slots.
3. Jobs running on main are cancelled by further commits (such as multiple consecutive PR merges).  This can make it more difficult to diagnose when a failure was introduced.

1. Change (most) workflows to only trigger on pushes for the main branch.  If maintainers want to run tests without making a PR they can still accomplish this with the added `workflow_dispatch` by manually triggering tests to run on their branch.
2. Split sanitizers out from `test-build.yml` into a new `test-sanitizers.yml`.  This new workflow doesn't automatically run on PRs, but again can be triggered with `workflow_dispatch`.  This workflow is currently set to run on all pushes to main, but may in future be changed to run only on trigger, with the version bump triggering it to run, limiting to a max of once per day.
3. Only cancel outdated commits if they aren't on main.

### DIFF
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -42,6 +42,7 @@ jobs:
   build-yosys:
     name: Reusable build
     runs-on: ${{ matrix.os }}
+    # pre_job is a subset of pre_docs_job, so we can always build for pre_docs_job
     needs: pre_docs_job
     if: needs.pre_docs_job.outputs.should_skip != 'true'
     env:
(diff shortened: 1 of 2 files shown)


widlarizer:
Looks like something broke on main with this

PR #5281, opened 2025-08-12
STAT: Add parameterised cells 

suisseWalter:
THIS PR IS A EXTENSION OF #5054 . (I'm not sure how this should be presented best, should it just contain the changes after #5054? currently it is a branch of the #5054 branch)
In discussion with @phsauter and the PULP group there is a need to have a way to estimate the area earlier in the flow. 
This PR implements a way to estimate the area earlier in the flow. It allows for lib files that have been customised to contain tables to lookup the "area" by the size of the input/output of the cells. This is the same idea as for #5278 . 
The details are explained there. 

See the same idea as #5278. 
The user can use their own definition of area. 
There is a prepared lib files for area estimated using nand2 gates as equivalent. and then just using mathematical scaling. 

There are these three lib files as a example (the first one could also be replaced by a actual lib file from a pdk)
they can be loaded and should be usable with any design. I have used the [croc](https://github.com/pulp-platform/croc/tree/main) design from pulp.
These are the same files as used by #5278. 
IHP130 standard cells: [ihp130_stdcell.lib.txt](https://github.com/user-attachments/files/21711729/ihp130_stdcell.lib.txt)
Yosys internal cells (which are used after techmap): [nand2_based_internal.lib.txt](https://github.com/user-attachments/files/21711740/nand2_based_internal_delay.lib.txt)
Yosys internal cells (before techmap): [area_by_width.lib.txt](https://github.com/user-attachments/files/21714708/area_by_width.lib.txt)


passes/cmds/stat.cc
```diff
@@ -659,9 +745,66 @@ void read_liberty_cellarea(dict<IdString, cell_area_t> &cell_area, string libert
...
+			if (single_parameter_area.size() == 0)
+				printf("error: %s\n", sar->args[single_parameter_area.size() - 1].c_str());
+			// check if it is a double parameterised area
+		}
+		const LibertyAst *dar = cell->find("double_area_parameterised");
+		if (dar != nullptr) {
+			for (const auto &s : dar->args) {
+
+				// printf("value: %s\n",sar->value.c_str());
+				// printf("args1: %s\n",dar->args[0].c_str());
```
KrystalDelusion: These can be removed.

passes/cmds/stat.cc
```diff
@@ -206,6 +209,89 @@ struct statdata_t {
...
+							cell_area.at(cell_type).is_sequential = cell_data.is_sequential;
+
+						} else {
+							printf("too small single_parameter_area %s %d %f\n", cell_type.c_str(), max_width,
+							       cell_data.single_parameter_area.back());
+							cell_area.at(cell_type).area = cell_data.single_parameter_area.back();
+							cell_area.at(cell_type).is_sequential = cell_data.is_sequential;
+						}
+						// printf("single_paramter_extraction %s %d %f\n", cell_type.c_str(), max_width,
+						// cell_area.at(cell_type).area);
```
KrystalDelusion: This can also be removed.

passes/cmds/stat.cc
```diff
@@ -206,6 +209,89 @@ struct statdata_t {
...
+						if (!cell_area.count(cell_type)) {
+							cell_area[cell_type] = cell_data;
+						}
+						if (cell_data.single_parameter_area.size() > max_width - 1u) {
+							cell_area.at(cell_type).area = cell_data.single_parameter_area.at(max_width - 1);
+							cell_area.at(cell_type).is_sequential = cell_data.is_sequential;
+
+						} else {
+							printf("too small single_parameter_area %s %d %f\n", cell_type.c_str(), max_width,
+							       cell_data.single_parameter_area.back());
```
KrystalDelusion: Why are you using `printf()` instead of `log()`?  Are these intended for the user or were they for debugging and need removing?

KrystalDelusion (changes requested):
As with #5054, it would be good to add a test or two for the behavior you're adding, along with minimal versions of the .lib files you've got in the PR description that can be included in the repository (containing just a couple cell types to check a small design).

passes/cmds/stat.cc
suisseWalter: I'm honestly not sure if this should be a log statement or be removed. it happens in the following case:
you define a cell area for width 1-10 but the design has a cell with width 20. It will just use the biggest entry in the list. 
Should that be logged as a warning or error or not at all?

passes/cmds/stat.cc
suisseWalter: done

passes/cmds/stat.cc
suisseWalter: done

passes/cmds/stat.cc
suisseWalter: I have now implemented this as a warning.

suisseWalter:
Testcases have been added.

suisseWalter:
> Approved, pending changes to #5054.
> 
> Up to you if you want to combine the two PRs into one, or wait for #5054 to be merged and then rebase this on `main`.

Feel free to merge them together if you want to.

PR #5284, opened 2025-08-12
check: Limit detailed cell edge checking for $pmux and $bmux

### LINKED ISSUE #5282: `check` takes excessively long on a specific module from cva6
[CVA6](https://github.com/openhwgroup/cva6) is being used as a testcase for the adder overflow formal app, but it doesn't get past the initial stages because it gets stuck in `check` (early on in `prep`).

To reproduce, clone the repo, then set some environment variables used in the flist file:
```
export CVA6_REPO_DIR=`pwd`
export HPDCACHE_DIR=$CVA6_REPO_DIR/core/cache_subsystem/hpdcache
export TARGET_CFG=cv64a6_imafdc_sv39
```
Yosys script:
```
verific -f -sv ./core/Flist.cva6
hierarchy -top cva6
check
```

The formal app uses flatten but it's reproducible without, `check` gets stuck in the `csr_regfile(...)` module specifically in the non-flattened design. This module only has 1791 cells, so something pathological must be going on...

jix:
This limit detailed cell edge checking for $pmux and $bmux when their S and Y ports become too large. While these cells can't have a quadratic number of edges between A, B and Y, they do have a quadratic number of edges between S and Y.

Yosys getting stuck when running `check` on some real world designs, see #5282.

We already exclude other cell types that can have a quadratic number of cell edges from detailed analysis within `check`. This adds `$pmux` and `$bmux` to those cell types, but for them, we exclude ports A and B from the threshold computation, as those don't contribute to the quadratic behavior. 

Fixes #5282

### DIFF
--- a/passes/cmds/check.cc
+++ b/passes/cmds/check.cc
@@ -195,16 +195,23 @@ struct CheckPass : public Pass {
 					// in port widths are those for us to check.
 					if (!cell->type.in(
 							ID($add), ID($sub),
-							ID($shl), ID($shr), ID($sshl), ID($sshr), ID($shift), ID($shiftx)))
+							ID($shl), ID($shr), ID($sshl), ID($sshr), ID($shift), ID($shiftx),
+							ID($pmux), ID($bmux)))
 						return false;
 
 					int in_widths = 0, out_widths = 0;
 
-					for (auto &conn : cell->connections()) {
-						if (cell->input(conn.first))
-							in_widths += conn.second.size();
-						if (cell->output(conn.first))
-							out_widths += conn.second.size();
+					if (cell->type.in(ID($pmux), ID($bmux))) {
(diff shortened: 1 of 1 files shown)


povik:
@jix Do you know what this corresponds to in the design? It's curious the mux is too a large for the check but gets processed and bitblasted fine.

jix:
@povik  It's what the verific frontend generates for https://github.com/openhwgroup/cva6/blob/bcb0f7de8654c8ad0b260d53a2ce070152d10c9f/core/csr_regfile.sv#L1013-L1744. It ends up generating a `$pmux` with `parameter \WIDTH 3456` and `parameter \S_WIDTH 238`

Update: And a single run of `opt_reduce` cuts the `S_WIDTH` down to `2` and `opt` overall also reduces the `WIDTH` by a lot.

widlarizer:
Ok to merge?

PR #5285, opened 2025-08-12
abstract: Add -initstates option

jix:
This adds a new `-initstates <n>` option to `abstract` which causes the enable condition for the abstraction to remain high for `n` initial (global clock) time steps.

To support use cases of `abstract` where the enable condition is a fixed number of FV time steps, independent of any signals present in the target module. This is useful for scripts that are supposed to work across different designs and thus there may not be a known local enable condition present in the target module.

To obtain the signal for the enable condition, we add an `$initstate` and for values of `n > 1` we delay it by chaining `n - 1` `$ff` cells with an initial value of 1.

This PR includes tests that verify the correct operation. Testing with other modes besides `-state` shouldn't be necessary, since this required no mode specific changes.

### DIFF
--- a/passes/cmds/abstract.cc
+++ b/passes/cmds/abstract.cc
@@ -373,6 +373,12 @@ struct AbstractPass : public Pass {
 		log("        abstractions performed by either mode. This option is not supported in\n");
 		log("        the -init mode.\n");
 		log("\n");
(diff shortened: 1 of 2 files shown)


nakengelhardt (approved):
Tested on a customer example and works as expected 👍

Issue #5279, opened 2025-08-11
Yosys 0.56 timeout


eder-matheus:
### Version

Yosys 0.56 (git sha1 9c447ad9d, g++ 13.3.0-6ubuntu2~24.04 -fPIC -O3)

### On which OS did this happen?

Linux

### Reproduction Steps

[ariane133_issue.tar.gz](https://github.com/user-attachments/files/21723923/ariane133_issue.tar.gz)

`tar -xf ariane133_issue.tar.gz`
`cd ariane133_issue`
`./run-me-ariane133-nangate45-base.sh`

### Expected Behavior

Yosys run ends with no errors in a reasonable time.

### Actual Behavior

After updating Yosys to the 0.56 version in OpenROAD-flow-scripts, two designs are failing with a timeout during 

```
Yosys execution. All of them are stuck here:
[...]
[2025-08-07T15:03:30.126Z] 3.7.8. Executing FSM_MAP pass (mapping FSMs to basic logic).
[2025-08-07T15:03:30.126Z] 3.8. Executing OPT pass (performing simple optimizations).
[2025-08-07T15:03:30.126Z] 3.8.1. Executing OPT_EXPR pass (perform const folding).
[2025-08-07T15:03:30.126Z] 3.8.2. Executing OPT_MERGE pass (detect identical cells).
[2025-08-07T15:03:30.126Z] 3.8.3. Executing OPT_MUXTREE pass (detect dead branches in mux trees).
[2025-08-07T15:03:32.767Z] 3.8.4. Executing OPT_REDUCE pass (consolidate $*mux and $reduce_* inputs).
[2025-08-07T15:03:32.767Z] 3.8.5. Executing OPT_MERGE pass (detect identical cells).
[2025-08-07T15:03:32.767Z] 3.8.6. Executing OPT_DFF pass (perform DFF optimizations).
```

The designs are nangate45/ariane133 and nangate45/ariane136. The attached test case reproduces the error for nangate45/ariane133.

Let me know if I need to provide any extra information or a different format for the test case.

georgerennie:
A smaller example [5279.zip](https://github.com/user-attachments/files/21724473/5279.zip). Running `yosys -p "read_rtlil 5279.il; opt_dff"` takes about 1 second on my machine even though it only contains one dff. This is because the algorithm implemented in #5241 seems to have some performance issues.

[...]

PR #5287, opened 2025-08-12
opt_dff: fix timeout issue

Anhijkt:
https://github.com/YosysHQ/yosys/issues/5279
Keeping track on `new_patterns` insertions.

### DIFF
--- a/tests/opt/opt_dff-simplify.ys
+++ b/tests/opt/opt_dff-simplify.ys
@@ -1,8 +1,16 @@
-# 5287 issue
+# 5279 issue
 # Check only for complimentary patterns elimination
 
-read_rtlil opt_dff-simplify.il 
+read_rtlil opt_dff-simplify.il
+
+select -assert-count 0 t:$adffe
+select -assert-count 1 t:$adff
+select -assert-count 0 t:$ne
+
 opt_dff
+
+select -assert-count 1 t:$adffe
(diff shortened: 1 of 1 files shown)


georgerennie:
Thank you, that was a suprisingly simple fix! I can confirm that this seems to fix the perf issues from #5279.

Before merging this, could you add some tests to check that the optimization is applied as expected - the original PR didn't have these and it would be good to make sure we have some coverage of the optimization, particularly when there are a large number of patterns as was the case in #5279. Some of the existing `opt_dff` tests like `opt_dff_en.ys` can probably be used for inspiration.

widlarizer:
Maybe this is fine but that's a very large test input being submitted that is hard to review - that kind of test is great as a regression test (like the `bug*.ys` tests) but not what I expect @georgerennie meant when asking for something like `opt_dff_en.ys`

Anhijkt:
I think it makes sense to leave `opt_dff-simplify.il` as a performance test for this optimization, but it doesn't fully cover `simplify_patterns `. New test covers both complimentary and redundancy elimination parts.

widlarizer (approved):
Test looks good now and fails when the opt_dff changes from this PR and from #5241 are removed. As of this PR it also passes way faster. Currently we can't make time it takes to run a test a requirement for the test to pass so this is as good as it gets

Issue #4349, opened 2024-04-20
"ERROR: Assert `count_id(wire->name) == 0' failed in kernel/rtlil.cc:2143" when using synth_{ice40,ecp5} on simple design


cr1901:
### Version

Yosys 0.40+25 (git sha1 171577f90, sccache x86_64-w64-mingw32-g++ 13.2.0 -Os)

### On which OS did this happen?

Windows

### Reproduction Steps

1. Save this file as `bad.rtlil`:
    ```
    # Generated by Yosys 0.40+25 (git sha1 171577f90, sccache x86_64-w64-mingw32-g++ 13.2.0 -Os)
    autoidx 95
    attribute \src "C:/msys64/home/William/Projects/FPGA/amaranth/smolarith/src/smolarith/base10.py:347"
    attribute \generator "Amaranth"
    attribute \hdlname "top"
    attribute \top 1
    module \top
      attribute \src "b2tob1000.v:143.3-152.6"
      wire width 11 $0\$15[10:0]$34
      attribute \src "b2tob1000.v:173.3-189.6"
      wire width 20 $0\$18[19:0]$46
      attribute \src "b2tob1000.v:91.3-92.23"
      wire width 11 $0\d0_guess[10:0]
      attribute \src "b2tob1000.v:100.3-101.28"
      wire width 20 $0\outp__payload[19:0]
      attribute \src "b2tob1000.v:143.3-152.6"
      wire width 11 $1\$15[10:0]$35
      attribute \src "b2tob1000.v:173.3-189.6"
      wire width 20 $1\$18[19:0]$47
      attribute \src "b2tob1000.v:143.3-152.6"
      wire width 11 $2\$15[10:0]$36
      attribute \src "b2tob1000.v:173.3-189.6"
      wire width 20 $2\$18[19:0]$48
      attribute \src "b2tob1000.v:173.3-189.6"
      wire width 20 $3\$18[19:0]$49
      attribute \src "b2tob1000.v:76.17-76.128"
      wire width 11 $add$b2tob1000.v:76$7_Y
      attribute $bugpoint 1
      wire $auto$bugpoint.cc:258:simplify_something$59
      attribute $bugpoint 1
      wire $auto$bugpoint.cc:258:simplify_something$61
      attribute $bugpoint 1
      wire width 10 input 3 $auto$bugpoint.cc:258:simplify_something$63
      attribute $bugpoint 1
      wire width 10 $auto$bugpoint.cc:258:simplify_something$64
[...]

PR #5291, opened 2025-08-14
rename.cc: Fixup ports after -unescape

KrystalDelusion:

The following test fails with an assertion error:
```
read_verilog << EOT
module top(output \$e );
submod \a$ (\$e );
endmodule

module submod(output \a[0] );
    assign \a[0] = 0;
endmodule
EOT

rename -unescape
check
```

If a port is renamed, we need to call `module->fixup_ports()`.

`tests/various/rename_unescape.ys` has been updated with the above testcase.

### DIFF
--- a/passes/cmds/rename.cc
+++ b/passes/cmds/rename.cc
@@ -598,6 +598,8 @@ struct RenamePass : public Pass {
 
 				for (auto &it : new_cell_names)
 					module->rename(it.first, it.second);
+
+				module->fixup_ports();
 			}
 		}
 		else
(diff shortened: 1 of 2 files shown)


KrystalDelusion:
Not sure if this should be applied to all `rename` pathways at the end instead or if it's just the `-unescape` that is the problem.

widlarizer (approved):
Thanks! Test does fail on main and doesn't with this change, looks good. Where did you run into this problem?

KrystalDelusion:
> Where did you run into this problem?

I used `rename -unescape` as a workaround for #4349, and then found this error when I tried it on #5274.

Looked into the other `rename` options and it seems like `unescape` is the only one affected by this; all the other ones either only target private wires, or they check `wire->port_id == 0`.  Which I guess raises the question; _should_ `unescape` be renaming ports?

widlarizer:
My guess is that it should rename ports as well, but the mode (as added in #5177) is kind of a compatibility version of a compatibility mode so I'm not sure. Thoughts @mikesinouye ?

Regardless, I think we should fix the assertion and maybe handle port renaming differently later

mikesinouye:
We have not had this issue with ports, but in my mind it should fix them also.

PR #5300, opened 2025-08-18
wheels: bison 3.8 on almalinux + memory pressure easing

donn:

Recent Linux wheel builds are failing because Bison is out of date.

Recent Mac wheel builds are failing because of what appears to be an OOM issue.

- GNU Bison built and added to PATH on Linux
- Upstream flex (via homebrew) now correctly added to `PATH` - automatic adding behavior in the Makefile seems to be suppressed by something at this point
- cibw now builds and uses bison 3.8.2 explicitly on platforms with no or out-of-date bison — AlmaLinux 8 only goes up to Bison 3.0
- cibw environment now includes `OPTFLAGS=-O3` to avoid generating debug info for abc, saving space in memory during linking
- setup.py attempts to build `yosys-abc` independently first to avoid memory pressure from gigantic abc link step running in parallel with something else

`python3 -m venv venv && ./venv/bin/python3 -m pip install .`

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -4,7 +4,7 @@ name: Build Wheels for PyPI
 on:
   workflow_dispatch:
   schedule:
-    - cron: '0 10 * * 0'
+    - cron: "0 10 * * 0"
 
 jobs:
   build_wheels:
@@ -54,9 +54,6 @@ jobs:
           fetch-depth: 0
           submodules: true
           persist-credentials: false
-      - if: ${{ matrix.os.family == 'linux' }}
-        name: "[Linux] Set up QEMU"
-        uses: docker/setup-qemu-action@v3
       - uses: actions/setup-python@v5
       - name: Get Boost Source
         shell: bash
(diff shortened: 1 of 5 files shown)


donn:
Wheel CI running at https://github.com/donn/yosys/actions/runs/17050617394/job/48337090728

Ravenslofty:
The Mac x86_64 build fails with this:

```
2025-08-18T20:17:41.7581210Z   frontends/verilog/verilog_lexer.cc:3895:21: error: out-of-line definition of 'LexerInput' does not match any declaration in 'frontend_verilog_yyFlexLexer'
2025-08-18T20:17:41.7583690Z   size_t yyFlexLexer::LexerInput( char* buf, size_t /* max_size */ )
2025-08-18T20:17:41.7585370Z                       ^~~~~~~~~~
2025-08-18T20:17:41.7588050Z   frontends/verilog/verilog_lexer.cc:3924:19: error: out-of-line definition of 'LexerOutput' does not match any declaration in 'frontend_verilog_yyFlexLexer'
2025-08-18T20:17:41.7590520Z   void yyFlexLexer::LexerOutput( const char* buf, size_t size )
2025-08-18T20:17:41.7592110Z                     ^~~~~~~~~~~
2025-08-18T20:17:42.0239170Z   2 warnings and 2 errors generated.
2025-08-18T20:17:42.0344540Z   make: *** [frontends/verilog/verilog_lexer.o] Error 1
```

donn:
aarch64 as well. dammit. investigating

donn:
If it turns out to be that PATH in GITHUB_ENV doesn't update until after the step concludes I'm going to be slightly angry at myself

```bash
          echo "PATH=$(brew --prefix flex)/bin:$PATH" >> $GITHUB_ENV
          echo "PATH=$(brew --prefix bison)/bin:$PATH" >> $GITHUB_ENV
```

EDIT: Yeah, that was it… https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#setting-an-environment-variable

New CI at https://github.com/donn/yosys/actions/runs/17054242085

EDIT 2: Turns out this was always an issue, but the reason we did not notice is that the Yosys Makefile has a backup that adds flex/bison/gettext to PATH. For some reason that fails now. Might have something to do with me adding Bison to PATH but honestly I'm not even sure anymore.

PR #5301, opened 2025-08-18
Reapply "Workflow adjustments"

KrystalDelusion:

#5280 had two failing jobs CI on merge and was reverted.
1. The `test-cells` job was added in #5024 based on the old reusable build cache name (i.e. including sanitizer).  This was merged after #5280 was written, but it didn't trigger a merge conflict so I forgot to update the PR to catch that one.
2. `run-san` worked fine when it was part of the `test-build.yml` where it had separate build and test jobs, with the build job using an out-of-tree build.  When I moved it into `test-sanitizers.yml` I merged the build and test jobs, but didn't account for the out-of-tree build.
3. While investigating, I also noticed that the `iverilog` cache key has been broken since #5016.  It was changed to use shell expansion, but it wasn't running in a shell and was instead an action variable.

1. Drop the sanitizer from `test-cells`.
2. Change `run-san` to build Yosys in-tree.
3. Write `IVERILOG_GIT` to output instead of env, so we can expand it without worrying about environment variable vulnerabilities.

I tested the two failing jobs on my fork, since `test-sanitizers.yml` no longer exists on `main` and therefore github doesn't allow workflow triggers.  [Compiler testing (test-cells)](https://github.com/KrystalDelusion/yosys/actions/runs/17055215422), and [run-san](https://github.com/KrystalDelusion/yosys/actions/runs/17060126147).

### DIFF
--- a/passes/memory/memlib.cc
+++ b/passes/memory/memlib.cc
@@ -878,7 +878,7 @@ struct Parser {
 						}
 					}
 				}
-				var.clk_en = find_single_cap(pdef.clken, cram.options, portopts, "clken");
+				var.clk_en = find_single_cap(pdef.clken, cram.options, portopts, "clken") != nullptr;
 			}
 			const PortWidthDef *wdef = find_single_cap(pdef.width, cram.options, portopts, "width");
 			if (wdef) {
--- a/passes/memory/memlib.h
+++ b/passes/memory/memlib.h
@@ -109,7 +109,7 @@ struct PortVariant {
 	PortKind kind;
 	int clk_shared;
 	ClkPolKind clk_pol;
-	bool clk_en;
+	bool clk_en = false;
 	bool width_tied;
 	int min_wr_wide_log2;
 	int max_wr_wide_log2;


KrystalDelusion:
Converted to draft while investigating failing UB.  ~I can't reproduce locally with `Yosys 0.56+141 (git sha1 1cdf058df, ccache clang++ 18.1.3 -fPIC -O3 -O1 -fno-omit-frame-pointer -fno-optimize-sibling-calls -fsanitize=undefined,address)` on Ubuntu 24.04~, and the PR reverting #5280: https://github.com/YosysHQ/yosys/actions/runs/17049372046 also passed (with both `undefined` and `address` rather than `undefined,address`).

Helps if I set `halt_on_error=1`, in `tests/memlib/t_sync_small.log`:
```
passes/memory/memory_libmap.cc:1733:12: runtime error: load of value 68, which is not a valid value for type 'bool'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior passes/memory/memory_libmap.cc:1733:12 
```

KrystalDelusion:
I'm getting this at least as far back as v`0.55` with just `undefined`.  Not sure why it's passing on the Yosys CI, but not locally or on forked CI.

With `undefined,address` on ubuntu I only get `t_sync_small` failing.  With `undefined` I get quite a few:
```
$ find tests/**/*.log -exec grep -l "Sanitizer" {} \;
tests/memlib/t_async_big.log
tests/memlib/t_async_small.log
tests/memlib/t_init_lut_val2_any.log
tests/memlib/t_init_lut_val2_no_undef.log
tests/memlib/t_init_lut_val_any.log
tests/memlib/t_init_lut_val_no_undef.log
tests/memlib/t_init_lut_x_any.log
tests/memlib/t_init_lut_x_no_undef.log
tests/memlib/t_init_lut_x_none.log
tests/memlib/t_init_lut_x_zero.log
tests/memlib/t_init_lut_zeros_any.log
tests/memlib/t_init_lut_zeros_zero.log
tests/memlib/t_sync_big_lut.log
tests/memlib/t_sync_small.log
```
Comparing with the CI `run-san` linked above, the ubuntu job doesn't error until `t_sync_small`, but the macos job immediately fails on `t_async_big` (which is the first test in memlib/tests).  The list above also correlates *exactly* with the tests that expect a non-zero number of `RAM_LUT`.

Yosys 0.56 with `undefined` hits the same list of failures, but with `undefined,address` it hits *none*.

KrystalDelusion:
[Current CI on fork](https://github.com/KrystalDelusion/yosys/actions/runs/17058640817/job/48362323263) also doesn't run into it.  Trying [`run-san` with separated sanitizers](https://github.com/KrystalDelusion/yosys/actions/runs/17059181727) shows it failing `undefined,address` but not `undefined` so that's fun 🙃.  It does potentially mean that this has just *always* been a problem with `memory_libmap.cc` that was never caught rather than an introduced one.

passes/memory/memlib.cc
```diff
@@ -878,7 +878,7 @@ struct Parser {
 						}
 					}
 				}
-				var.clk_en = find_single_cap(pdef.clken, cram.options, portopts, "clken");
+				var.clk_en = find_single_cap(pdef.clken, cram.options, portopts, "clken") != nullptr;
```
nakengelhardt, line 881: I'm morbidly curious why this change is required 🤔 Is something about how `find_single_cap` is declared interfering with the implicit conversion to bool?

KrystalDelusion: The actual fix is assigning a default value to `PortVariant::clk_en`, this change was incidental while investigating but it made sense to me to leave it in since it helps with readability.

widlarizer:
~~I think this broke CI again~~ oh it was reverted already so it's the LLVM 21 problem

PR #5308, opened 2025-08-21
opt_muxtree: refactor

widlarizer:
I did some refactoring and added a bunch of clarifying comments with no intended functional impact to allow transparency for soundness and make future rewrites easier

cc @rocallahan

### DIFF
--- a/passes/opt/opt_muxtree.cc
+++ b/passes/opt/opt_muxtree.cc
@@ -427,14 +427,17 @@ struct OptMuxtreeWorker
 			if (root_enable_muxes.at(m))
 				continue;
 			else if (root_muxes.at(m)) {
+				// This leaf node of the current tree
+				// is the root of an input tree of the current tree
 				if (limits.recursions_left == 0) {
-					// Ran out of subtree depth, re-eval this subtree in the next re-run
+					// Ran out of subtree depth, re-eval this input tree in the next re-run
 					root_mux_rerun.insert(m);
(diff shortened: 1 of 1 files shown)


passes/opt/opt_muxtree.cc
```diff
@@ -303,84 +352,123 @@ struct OptMuxtreeWorker
...
 					root_enable_muxes.at(m) = true;
 					log_debug("      Removing pure flag from root mux %s.\n", log_id(mux2info[m].cell));
-				} else
-					eval_mux(knowledge, m, false, do_enable_ports, abort_count - 1);
-			} else
-				eval_mux(knowledge, m, do_replace_known, do_enable_ports, abort_count);
-			if (glob_abort_cnt == 0)
+				} else {
+					auto new_limits = limits.subtree();
+					// Since our knowledge includes assumption, we can't generally allow replacing based on it
```
jix: I was briefly confused by this, because at the top you added a comment that explains that internal nodes of a mux tree will only have a single user that also is a mux, and in that case the activation assumptions added during recursion would be valid assumptions for replacing signals throughout the entire subtree.

What this comment here doesn't mention is that we're in a `else if (root_muxes.at(m)) {` branch, and this code handles the case when we recurse across mux trees, i.e. it handles exactly those cases where there may be other users for which the assumptions don't hold. I think it would be helpful to include that detail here.

jix (approved):

PR #5311, opened 2025-08-22
Logging: Add log sink system to replace ostreams

Logikable:
This PR adds a LogSink class that aims to be more versatile than ostream sinks. See https://yosyshq.discourse.group/t/cleaning-up-log-sinks/86 for the discussion.

The old description can be found below:

~~We'd like to subscribe to only the most critical logs. Currently, there is no way to specify a log level threshold. This PR provides a stopgap solution.~~

~~In addition to the log_streams vector that new log sinks can push to, this PR adds a log_warning_streams vector that behaves similarly, but ostreams in this new vector only receive warning and error logs.~~

### DIFF
--- a/.github/actions/setup-build-env/action.yml
+++ b/.github/actions/setup-build-env/action.yml
@@ -8,7 +8,7 @@ runs:
       shell: bash
       run: |
         sudo apt-get update
-        sudo apt-get install gperf build-essential bison flex libfl-dev libreadline-dev gawk tcl-dev libffi-dev git graphviz xdot pkg-config python3 libboost-system-dev libboost-python-dev libboost-filesystem-dev zlib1g-dev libbz2-dev libgtest-dev libgmock-dev
+        sudo apt-get install -y gperf build-essential bison flex libfl-dev libreadline-dev gawk tcl-dev libffi-dev git graphviz xdot pkg-config python3 libboost-system-dev libboost-python-dev libboost-filesystem-dev zlib1g-dev libbz2-dev libgtest-dev libgmock-dev
 
     - name: Install macOS Dependencies
       if: runner.os == 'macOS'
--- a/.github/actions/setup-build-env/action.yml
+++ b/.github/actions/setup-build-env/action.yml
@@ -8,7 +8,7 @@ runs:
       shell: bash
       run: |
         sudo apt-get update
-        sudo apt-get install gperf build-essential bison flex libfl-dev libreadline-dev gawk tcl-dev libffi-dev git graphviz xdot pkg-config python3 libboost-system-dev libboost-python-dev libboost-filesystem-dev zlib1g-dev libbz2-dev libgtest-dev
+        sudo apt-get install gperf build-essential bison flex libfl-dev libreadline-dev gawk tcl-dev libffi-dev git graphviz xdot pkg-config python3 libboost-system-dev libboost-python-dev libboost-filesystem-dev zlib1g-dev libbz2-dev libgtest-dev libgmock-dev
 
     - name: Install macOS Dependencies
       if: runner.os == 'macOS'
--- a/.github/workflows/test-verific.yml
+++ b/.github/workflows/test-verific.yml
@@ -36,8 +36,7 @@ jobs:
           persist-credentials: false
           submodules: true
       - name: Runtime environment
-        run: |
-          echo "procs=$(nproc)" >> $GITHUB_ENV
+        uses: ./.github/actions/setup-build-env
 
       - name: Build Yosys
         run: |
--- a/kernel/log.cc
+++ b/kernel/log.cc
@@ -39,7 +39,7 @@
 YOSYS_NAMESPACE_BEGIN
 
 std::vector<FILE*> log_files;
-std::vector<std::ostream*> log_streams;
+std::vector<std::ostream*> log_streams, log_warning_streams;
 std::vector<std::string> log_scratchpads;
 std::map<std::string, std::set<std::string>> log_hdump;
 std::vector<std::regex> log_warn_regexes, log_nowarn_regexes, log_werror_regexes;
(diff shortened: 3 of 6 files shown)


mikesinouye:
Hey @widlarizer, please take a look when you get a chance, thank you!

widlarizer:
Seems like `ERROR` is predefined somewhere in Windows specific headers. You can rename the enum variand or `#ifdef #undef` it

Logikable:
Renamed, thanks :)

widlarizer:
I'm in favor of the LogSeverity struct and an overloaded log function seems like a decent way to safely retrofit it. Is there an intended followup that exposes `log_warning_streams` somehow? As a side note, please take care to respect established whitespace, C++ is indented with tabs

Logikable:
I've rewritten this PR in light of the log changes by @rocallahan. One notable consequence of this PR is that `log_formatted_warning` and `log_error_with_prefix` now call `log_formatted_string` instead of `log`. I've also switched to tabs.

Regarding the last comment (@widlarizer): `log_warning_streams` is already exposed the same way `log_streams` is exposed. Today, to subscribe a sink, you append an ostream to the global `Yosys::log_streams`; the same can be done with `log_warning_streams`.

widlarizer (changes requested):
There is no code in this PR appending to `log_warning_streams`

widlarizer:
Also, this breaks `tests/various/logger_cmd_error.sh`

Logikable:
Sorry -- I misunderstood. We'll be appending to `log_warning_streams` in a downstream change. I felt it better to contribute the public parts of our change upstream; is that ok?

Logikable:
gentle bump

widlarizer (changes requested):
It took some internal alignment on what we want to do in situations like this. I'm also going to be updating contributing guielines and PR/issue templates to match what we've arrived at. Let me explain on the two sides of this PR:

The problem with adding warning streams that are only appended to in your patches is that it is impossible to maintain for us something that we can't cover with tests locally. Our regular testing requires exposing features to regular Yosys users too. In this case, that could be done via a driver option, like we already have `-l` / `--logfile`. Please add such an option (`--warnfile`? no need for a single letter spelling) and tests that use it to cover that feature

As for the log.h change, it's okay to change public headers like this if we agree on this being worth whatever the benefit is. For that, it would be best to discuss what direction you're taking Yosys in your patches on [Discourse](https://yosyshq.discourse.group/) ahead of time. The fact you declare this a stopgap measure even suggests a lack of faith in this being the long-term right thing to do. I wouldn't block it on just this aspect, but even for this, I'd prefer first to hear more about the user story and your view on the alternate approaches we could be taking

Ultimately, my previous review still applies. If that is resolved, I don't mind merging this.

- [ ] expose warning streams in the driver (see above)

Logikable:
Thank you! I've created https://yosyshq.discourse.group/t/cleaning-up-log-sinks/86.

Logikable:
I've rewritten this PR to contain a more permanent solution. It's not perfect -- there are still places in log.cc where one API log call turns into multiple logv_string calls, and log sinks will see these as separate logs. That said, this was already an issue with the existing ostream system.

jix:
The issue with multiple log calls used for a single message only extended to the log message matching used as part of our test suite, since with the stream based approach there isn't really a user-visible concept of individual log messages, as there are just lines ending in newlines and nothing to indicate where individual log calls start and end.

As I mentioned in my reply to your discourse thread, we do also have existing log emitting code that splits single lines of single messages across multiple `log` calls at the source, so any plan for improving the logging needs to include a plan for what to do about those. My suggestion would be to add line buffering when translating from stream based log sources to message based log sinks.

Repeating what I wrote on discourse, it's not a requirement that all of this is addressed in a first PR, but what is a requirement is that we can see how the PR is a step towards something that makes sense for the project. And in cases where we do have some ideas for improvements we'd like to see, also to ensure it's aligned or at least not in conflict with that.

@mmicko: I also want to note that this PR now contains changes to the CI workflows and adds a new build time dependency. (cc @widlarizer): I feel like part of our contribution guidelines should include that these are things that should be noted in the PR description, done separately from code changes, and ideally also discussed before being submitted for review.

What is still missing to move forward with this is:

* Responding to the discourse thread, so we have an idea how the plan behind this relates to the ideas that we already have for logging and whether it is headed in the same direction or at least can be extended to support those use cases.
* As already explained by @widlarizer, exposing the added functionality from the CLI by adding a driver option.

Logikable:
Can you explain what the driver option would do given that this PR is no longer about adding a warning-log-specific sink, but a general programmatic sink system?

My understanding was that the driver option was helpful for primarily for testing and exposing a specific log sink usecase. This PR now contains tests that test the interface directly, and I'm not sure how to expose the LogSink class through a flag.

I feel that introducing a `--warnfile` flag may be best suited for a follow-up PR so as to keep PRs as minimal as possible.

widlarizer:
I understood what a separate warning log sink would do, but what's a general programmatic sink system? I don't understand the purpose of this PR. What is the change in yosys behavior you are adding and how is it observable to a user? In case you want to discuss this on a call you can find a link to our open dev JF meeting in CONTRIBUTING.md

Logikable:
We are looking to pipe logs to a cloud provider's logging system to be able to understand our usage.

We'd like to tag the logs that are going into this system with metadata (log level, the RTL source line the log is associated with, etc.) so that we can gain deeper insight on how we're using Yosys.

ostreams do not provide us with this metadata, and we'd prefer to avoid searching the log string or hooking into other parts of Yosys while it's running in order to get the metadata.

mmicko:
@Logikable before we go into more details with this, there were lot of changes on main meanwhile and gmock is for example already there, so changes in `.github` directory and `Makefile` itself are not needed anymore, so could you please rebase this PR ?
Also if you do not have time/resources to do that, just let me know so I can take care of such changes myself.

Logikable:
Thanks for finding this PR again and reaching out! I no longer have the capacity to work on this, so I appreciate that you're able to take it over.

mmicko:
@Logikable Thanks for reply. Will see then to continue working on separate branch for this but would reuse your existing commits, otherwise would be complicated since this is in your repo main branch.

mmicko:
Added current work to https://github.com/YosysHQ/yosys/tree/log-sink will keep this PR alive for until some progress is done and create a new one to point on this one.

mmicko:
Closing this as https://github.com/YosysHQ/yosys/pull/6115 address it

Issue #5205, opened 2025-06-30
Cross-OS Determinism


donn:
### Feature Description

Yosys is currently deterministic within the same OS: On any version of Linux, whether it's x86-64 or aarch64, for the same inputs, Yosys will always return the same result.

However, Synthesis results appear to be quite different across OSes: See this reproducible for more info, where I ran the same command `yosys -L$(uname).log ./cross-os.tcl` once on macOS and once on Linux-- the RTLIL was quite different even for a simple script:

```tcl
yosys -import
read_verilog ./picorv32.v
hierarchy -check -top picorv32 -nokeep_prints -nokeep_asserts
tribuf
synth -top picorv32
opt
opt_clean -purge
write_rtlil picorv32-[exec uname].rtlil
```

[ys-repro.zip](https://github.com/user-attachments/files/20988175/ys-repro.zip)

---

It would be really nice to have cross-OS determinism (ASIC flows typically diverge at Synthesis, radically changing the layout which makes issues harder to track when working on a Mac.) I might work on this but would like to know if this is already a known issue or something (can't find anything via search.)

KrystalDelusion:
> It would be really nice to have cross-OS determinism (ASIC flows typically diverge at Synthesis, radically changing the layout which makes issues harder to track when working on a Mac.) I might work on this but would like to know if this is already a known issue or something (can't find anything via search.)

I could be wrong, but I suspect this is largely due to the chaotic nature of ABC and a lack of input normalization (c.f. #3713, #4913, #4618, #5091).  If you are willing/able to investigate where exactly differences start appearing in the RTLIL design that would potentially be quite helpful.  My guess is that a lot of it is due to absolute path differences, which would lead to things like the following being quite variable:
```
attribute \src "/usr/bin/../share/yosys/ice40/abc9_model.v:14.11-19.3|/usr/bin/../share/yosys/ice40/cells_sim.v:229.30-229.38"
wire $techmap$auto$abc9_ops.cc:834:prep_xaiger$955613.carry.$logic_or$/usr/bin/../share/yosys/ice40/cells_sim.v:229$19810_Y
```

This would also show up on the same OS, for different install locations, or even on the same system when comparing a local Yosys executable vs an installed version of the same executable.  Without looking into it too much, I think you could potentially remove a lot of path differences even without changing Yosys by removing the src attribute `setattr -unset src`, and then scrambling names `rename -scramble-name` (which by default always uses a seed of 1).  This does however make the design completely unreadable.

[...]

Issue #5207, opened 2025-07-02
Functional backend error when bitwidths too wide


gussmith23:
### Version

Yosys 0.53+39 (git sha1 d8b27d41c, c++ 16.0.0 -fPIC -O3)

### On which OS did this happen?

macOS

### Reproduction Steps

Use the file: 
```sv
(* use_dsp = "yes" *) module top(
	input  [17:0] a,
	input  [17:0] b,
	input  [17:0] c,
	input  [17:0] d,
	output [17:0] out,
	input clk);

	logic  [35:0] stage0;
	logic  [35:0] stage1;

	always @(posedge clk) begin
	stage0 <= ((d + a) * b) + c;
	stage1 <= stage0;
	end

	assign out = stage1;
endmodule
```

Run the command:
```
 yosys -q p 'read_verilog -sv file.sv; hierarchy -top top; prep; clk2fflogic; write_functional_rosette -provides'
```

### Expected Behavior

No error, converts to Rosette

### Actual Behavior

ERROR: The design contains an undriven signal stage1. This is not supported by the functional backend. Call setundef with appropriate options to avoid this error. 

Note that shrinking the  registers to the correct size fixes the issue. However, I wouldn't expect this to be something that would cause a failure. Is there a way around this?

gussmith23:
Oh my gosh, I didn't see that the solution was right in the error message. It was being clipped in my terminal 🤦 I'll try out setundef.

georgerennie:
It's not entirely clear to me why that example should require setundef though, the entirety of stage1 should be well defined, it has no undef bits, just unused bits

nakengelhardt:
This made me realize that some of the important documentation about the expectations of the functional backend ended up only in a PDF we had to send to Sandia; the first page of this doc should really be added to the [functional IR docs page](https://yosyshq.readthedocs.io/projects/yosys/en/latest/yosys_internals/extending_yosys/functional_ir.html), ideally with the suggested yosys commands for achieving that design state consistently...

[Interface Document Functional Backend.pdf](https://github.com/user-attachments/files/21033438/Interface.Document.Functional.Backend.pdf)

The concrete problem here is that `wreduce` (which is part of `prep`) recognizes that half of the intermediary signals isn't actually used, and optimizes the bitwidth of all the operators, which leaves the upper half of `stage0` and `stage1` (and some privately-named intermediate signals) undriven. (And partially-undriven is not supported under the "no undriven signals" requirement.)

[...]

PR #5315, opened 2025-08-28
write_rtlil: don't sort

### LINKED ISSUE #3576: `write_rtlil` and `read_rtlil` won't recreate the same design
### Version

Yosys 0.24+1 (git sha1 1f6ac926a, gcc 12.2.1 -fPIC -Os)

Linux

### Reproduction Steps

```
#!/bin/bash
TEST_DIR=operators
INPUT=yosys/tests/simple/operators.v

mkdir -p $TEST_DIR

# read verilog, generate RTLIL and netlist
yosys \
-p "read_verilog -sv $INPUT" \
-p "write_rtlil $TEST_DIR/yosys_rtlil.txt" \
-p "synth_xilinx" \
-p "write_verilog $TEST_DIR/yosys_gate.v"

# read RTLIL, generate netlist
yosys \
-p "read_rtlil $TEST_DIR/yosys_rtlil.txt" \
-p "synth_xilinx" \
-p "write_verilog $TEST_DIR/yosys_rtlil_gate.v"
```

`operators/yosys_rtlil_gate.v` and `operators/yosys_gate.v` differ.

### Expected Behavior

It's expected, that when we generate a netlist using `read_verilog` , `synth_xilinx`, `write_verilog`, it will be the same as  (or equivalent to) the netlist that is generated from RTLIL, assuming that the RTLIL is of the same exact design, as can bee seen in the script above.

### Actual Behavior

The netlists differ, even though they were generated for the same exact test case.

widlarizer:
Fixes #3576. We need to fix it becuse the obvious, most default attempt at roundtripping through textual representation of the core IR shouldn't fail. But it does, which makes building reproducers and custom workflows insufferable.

I think that the purpose of the sort is to hide reproducibility bugs and cross-platform equivalence bugs in Yosys. I don't know of recent cases of the former and they're bugs, we shouldn't be proactively hiding them. For the latter, see #5205. These are real categories of bugs that probably happen, but hiding them in one place in a way that breaks roundtripping solves nothing as these will still occur inside of flows. See 43951099 for how the sort was added, replacing pre-existing sorting of the output rather than the design.

From #3576:

> Reloading the design changes some autoindex numbers and some pointer values

I think this was actually referring to roundtripping through Verilog. Reloading RTLIL doesn't change autoidx. Dependence on pointer values would be a bug and I don't expect we have a lot of these, I would even expect a sanitizer to report them as an error (pointer arithmetic etc) though I haven't checked. Until recently, we haven't been using `std::unordered_map` and friends, instead opting for custom replacements in hashlib that have well-defined iteration order. Since we have started using the `std` types, we've been careful not to rely on their iteration order.

For allowing RTLIL designs to be easily `diff`ed, see #5003

Related: #5150

- [x] Fix that the backend writes out things in insertion order (the usual iteration order is reverse insertion order)
- [x] Figure out how to test by comparing against canonical dumps when src attribute values depend on source path


KrystalDelusion:
I have used `dump` instead of `write_rtlil` when running into issues with reproducibility because that also avoids the sort, and I think not needing to remember that they are different would be an improvement.

widlarizer:
One of the CI failures is a functional backend test failing: `ERROR: The design contains an undriven signal $flatten\picorv32_i.$auto$wreduce.cc:514:run$3801. This is not supported by the functional backend. Call setundef with appropriate options to avoid this error.
`, can be reproduced with `pytest -k picorv -m "not smt and not rkt"` run in `tests/functional`

widlarizer:
The failing functional test is due to #5207

whitequark:
Is `-relativeshare` something likely to become a default?

widlarizer:
@whitequark I added `-relativeshare` just so that I can use a fully synthesized demo design and compare against a committed reference. I think that with `-noabc` Yosys actually is fully reproducible across platforms and if not we have to fix it either way. I don't think it would make sense to use in any other context

widlarizer:
I expect some users may dislike this change, since it will introduce frustrating disorder in even how cell ports are ordered. However, that is just a realistic view of what Yosys sees, so it's necessary for `write_rtlil` creating "Yosys save files". They can still do `write_rtlil -sort` to get the old behavior

widlarizer:
@whitequark Are you okay with the removal of the `design.sort()` in bugpoint? It's been there from the start, so I assume it was for consistency with write_rtlil or with some similar idea, but it might even hide some bugs, making them unbugpointable

jix (approved):
Making a pass over the diff, this lgtm

whitequark:
> Are you okay with the removal of the `design.sort()` in bugpoint? It's been there from the start, so I assume it was for consistency with write_rtlil or with some similar idea, but it might even hide some bugs, making them unbugpointable

I honestly don't remember, your call is at least as good as mine here.

PR #5317, opened 2025-08-30
libparse/filterlib: mark LibertyParser::error() as weak

donn:

Admittedly a niche use-case: I use `libparse.cc` as a standalone library. Problem is, the default `error()` methods exit the process altogether.

When FILTERLIB is defined (attempts to compile libparse more or less standalone,) mark the `LibertyParser::error()` as weak so utilities using libparse as a library can override its behavior.

As the code is quite performance-critical, I've elected to not modify it to raise an exception or have a callback or similar and simply allow for a link-time replacement.

EDIT: Virtual methods also can't work for two reasons:
- performance
- c++ doesn't do virtual dispatch in constructors and parse() is called in a constructor

There should be no functional change to Yosys from this at all. Doubly so since FILTERLIB is not normally defined.

### DIFF
--- a/passes/techmap/libparse.cc
+++ b/passes/techmap/libparse.cc
@@ -676,12 +676,15 @@ void LibertyParser::error(const std::string &str) const
 
 #else
 
+YS_ATTRIBUTE(weak)
 void LibertyParser::error() const
 {
 	fprintf(stderr, "Syntax error in liberty file on line %d.\n", line);
 	exit(1);
 }
 
+
+YS_ATTRIBUTE(weak)
 void LibertyParser::error(const std::string &str) const
 {
 	std::stringstream ss;


widlarizer:
I'm fine with the idea. If you rebase onto main the macOS CI failure should go away, but first you should use `YS_ATTRIBUTE` to avoid MSVC build errors. No idea how linking works on Windows and if there's an alternative etc

donn:
There is no well-documented alternative, no:

https://stackoverflow.com/questions/2290587/gcc-style-weak-linking-in-visual-studio

donn:
no matter how many times i rebase, the mac build error is not going away. im confused

widlarizer:
As of #5319 that CI error should be fixed, but you might get hit with #5320, I'm not sure. Sorry about that!

KrystalDelusion:
> As of #5319 that CI error should be fixed, but you might get hit with #5320, I'm not sure. Sorry about that!

The reusable build uses the default cxxstd (C++17); #5320 isn't currently impacting CI, but it may affect user builds (which is why I spun it out as a separate issue instead of addressing it in #5319 as well).

@donn #5319 had not yet been merged to main when you were rebasing yesterday; it should work fine if you do it now.

donn:
@KrystalDelusion Thanks! I went ahead and rebased one last time.

widlarizer (approved):

PR #5319, opened 2025-09-01
Fix failing macOS CI runs

KrystalDelusion:
_What are the reasons/motivation for this change?_

https://yosyshq.discourse.group/t/psa-llvm21-on-macos-is-failing-yosys-builds/76

_Explain how this is achieved._

Specify `llvm@20` in brewfile, and use the brewfile for macOS CI runs.
Also add macos-latest to test-compile.

Note that setting macOS compilers to specify clang-19 passes C++17, but fails C++20 (on both x86 and arm) despite passing both C++ versions under ubuntu, spun out to #5320.

_If applicable, please suggest to reviewers how they can test the change._

PR #5322, opened 2025-09-02
CI Fixes

KrystalDelusion:

While working on #5319 I discovered that macOS builds using C++20 seem to be broken in the compiler testing due to ambiguity of the `format` function.
Also the migration to windows-2025 has begun and the vs-build is targeting an SDK version which is no longer available.

Rename the offending function to avoid the ambiguity.
Update to a supported SDK version.

Follow instructions in #5319 to reproduce failure without this change.

Since windows-latest is rolling out the update to windows-2025 I also made sure to run the current job under windows-2022 to make sure we don't get failures if it's picked up by an un-updated runner: https://github.com/YosysHQ/yosys/actions/runs/17416640362/job/49446921683
And here is the windows-2025 job passing: https://github.com/YosysHQ/yosys/actions/runs/17416542448/job/49446214118?pr=5322

### DIFF
--- a/.github/workflows/extra-builds.yml
+++ b/.github/workflows/extra-builds.yml
@@ -51,7 +51,7 @@ jobs:
         uses: microsoft/setup-msbuild@v2
       - name: MSBuild
         working-directory: yosys-win32-vcxsrc-latest
-        run: msbuild YosysVS.sln /p:PlatformToolset=v142 /p:Configuration=Release /p:WindowsTargetPlatformVersion=10.0.17763.0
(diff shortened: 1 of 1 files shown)


KrystalDelusion:
Visual Studio build failure is unrelated, was [previously passing](https://github.com/YosysHQ/yosys/actions/runs/17389371865) when running under windows-2022 but now failing under windows-2025.  The updated windows-latest just started rolling out, so there's a good chance re-running the build once the rest finish that it will get picked up by a windows-2022 image and pass.

mmicko (approved):

Issue #5234, opened 2025-07-16
fminit causes std::out_of_range


hhe07:
### Version

0.54+15

### On which OS did this happen?

Linux

### Reproduction Steps

With the following SystemVerilog design:
```systemverilog
module counter(input clk, input rst, output logic is_full);


    logic [1:0] ctr;
    always @(posedge clk) begin
        if (rst) begin
            ctr <= 0;
        end else begin
            ctr <= ctr+1;
        end
    end
    assign is_full = (ctr == 2'b11);

endmodule
```
run the following yosys script:
```
read_verilog -sv -formal reset.sv

hierarchy -check -top counter
prep -top counter
fminit -seq rst 0,1
```
The following output is given:
```
SBY 14:39:48 [reset_task1] base: terminate called after throwing an instance of 'std::out_of_range'
SBY 14:39:48 [reset_task1] base: what():  vector::_M_range_check: __n (which is 1) >= this->size() (which is 1)
SBY 14:39:48 [reset_task1] base: finished (returncode=-6)
SBY 14:39:48 [reset_task1] base: task failed. ERROR
```

### Expected Behavior

Documentation of fminit is rather thin, but I would expect either an error about the sequence not being supported, or for the sequence to properly initialise.

### Actual Behavior

The aforementioned ``std::out_of_range`` is given. 

The error is not thrown if the sequence is 1,0, but re-appears if the sequence is 1,0,1 or similar. I think it's an issue from adding entries of the ``initdata`` vector in the ``fminit.cc`` pass, but am unsure about further details.

KrystalDelusion:
Looks like what's happening here is that the first value gets parsed as the width of the LHS, but any additional values (other than `0` and `~0`) are parsed as raw values, so an integer will get treated as a 32bit value.  Then when it tries to assign the 32-bit wide value to a 1-bit wide `rst` it throws.  As a work-around, you can either provide a width (`1'b1` instead of `1`), or use `~0`.

PR #5328, opened 2025-09-04
fix(parse): #5234 adjust width of rhs according to lhs

rhanqtl:

#5234 

Add code in `SigSpec::parse_rhs` to adjust the width of RHS.

Try different combinations of width and signedness of the arguments of `-seq`

### DIFF
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -5974,7 +5974,11 @@ bool RTLIL::SigSpec::parse_rhs(const RTLIL::SigSpec &lhs, RTLIL::SigSpec &sig, R
 		}
 	}
 
-	return parse(sig, module, str);
+	if (!parse(sig, module, str))
+		return false;
+	if (sig.width_ > lhs.width_)
+		sig.remove(lhs.width_, sig.width_ - lhs.width_);
+	return true;
 }
 
 RTLIL::CaseRule::~CaseRule()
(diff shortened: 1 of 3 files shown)


rhanqtl:
Seems that `SigSpec` does not provide APIs to check signedness, so the fix uses zero-extend.

widlarizer:
I think the issue can be fixed by only touching code paths that would previously crash if it was done like this. Tests do pass this way. Do you have an example where actually zero-extending the constant would be necessary? I don't have an example where it would be a *problem* but one might exist.

```diff
-       sig.extend_u0(lhs.width_);
+
+       if (lhs.width_ <= sig.width_)
+               sig.remove(lhs.width_, sig.width_ - lhs.width_);
+
```

widlarizer:
In the new test, if you use `fminit -set rst 1'b1` and run it on main, you'll see that rst has only the lowest bit driven. With the zero extension, all bits of rst are driven. With the trim-only change, the behavior is consistent with main, so I think that's the right way to fix this

widlarizer (changes requested):
see comments

rhanqtl:
@widlarizer Thanks for your reviews! Yes, trimming-only is enough to fix the crash in fminit. Still, as a core API, `parse_rhs` should probably return a bit-vector that matches the length of lhs. Please correct me If I misunderstood.

widlarizer:
If the parse_rhs doesn't know signedness, then it can't correctly extend to match the context of the lhs. If zero extension or sign extension is desired, which one is the right one is up to the code using `parse_rhs`, so it should extend it if necessary on its own

Issue #862, opened 2019-03-10
Can't put assert inside case statement


Edgeworth:
## Steps to reproduce the issue
test.v
```
module test(input wire A);
  localparam TEST = 1;
  always_comb begin
    case (A)
      TEST: assert(1);
    endcase
  end
endmodule
```
Repro command:
`yosys -p 'read -sv test.v'`

## Expected behavior
Output:
```
Parsing SystemVerilog input from `test.v' to AST representation.
Generating RTLIL representation for module `\test'.
Successfully finished Verilog frontend.
```

## Actual behavior
Fails to parse.
Output:
```
Parsing SystemVerilog input from `test.v' to AST representation.
test.v:5: ERROR: syntax error, unexpected TOK_ASSERT
```

Reverting e7a34d3  #699  seems to fix this.

cliffordwolf:
Oops! #864 should fix it. Can someone double-check I didn't break anything else this time around?  ;)

PR #5330, opened 2025-09-05
Fix handling of cases that look like sva labels again.

higuoxing:
Commit c8e0ac0 introduces a regression on handling case exprs that look like sva labels.  After some debugging, we shouldn't push the identifier ast node to the ast_stack, otherwise, we will get the following assertion failure:

```
➜  /tmp yosys -p 'read -sv a1.v'

 Yosys 0.57+1 (git sha1 baa61a146, clang++ 20.1.8 -fPIC -O3)

-- Running command `read -sv a1.v' --

1. Executing Verilog-2005 frontend: a1.v
Parsing SystemVerilog input from `a1.v' to AST representation.
ERROR: Assert `extra->ast_stack.size() == 1' failed in frontends/verilog/verilog_parser.y:709.
➜  /tmp cat a1.v
module test(input wire A);
  localparam TEST = 1;
  always_comb begin
    case (A)
      TEST: assert(1);
    endcase
  end
endmodule
```

We encountered this issue before but with a different error message[^1],

[^1]: https://github.com/YosysHQ/yosys/issues/862


widlarizer:
Thanks for the contribution. We didn't add a regression test back when first fixed which is how this resurfaced so I added one

widlarizer (approved):

PR #5332, opened 2025-09-08
Add state_dependent_path_declaration so that `ifnone` can be parsed (rebased)

widlarizer:
Rebased #5121 with an added minimal test and the `-Wcex` bison flag removed


widlarizer:
When built with `SANITIZER=address`, reading in this leaks an attribute list on yosys shutdown:
```v
module ff(input CLK, DI, output reg Q);
specify
	if (CLKMUX == "INV") (negedge CLK => (Q : DI)) = 0;
endspecify
endmodule
```
This isn't the case with the original proposal

EDIT: fixed

FlinkbaumFAU:
Thanks for reviving #5121

PR #5335, opened 2025-09-10
verilog_parser: Use unique_ptr<> to store the attribute list.

higuoxing:
We can get rid of the new / delete statements and free_attr() and we
don't need to manage the memory manually.

### DIFF
--- a/frontends/verilog/verilog_parser.y
+++ b/frontends/verilog/verilog_parser.y
@@ -69,9 +69,10 @@
 		struct ParseState {
 			int port_counter;
 			dict<std::string, int> port_stubs;
-			dict<IdString, std::unique_ptr<AstNode>> *attr_list, default_attr_list;
-			std::stack<dict<IdString, std::unique_ptr<AstNode>> *> attr_list_stack;
-			dict<IdString, std::unique_ptr<AstNode>> *albuf;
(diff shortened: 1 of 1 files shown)


higuoxing:
I did some investigation on removing the unique_ptr<> wrapper. It turns out that we cannot remove it for now. Because there're some logic depends on the nullness of the pointer:

```
if (attr_list)
    append_attr(ast_node, std::move(attr_list));
```

and we cannot replace it with

```
if (!attr_list.empty())
    append_attr(ast_node, std::move(attr_list));
```

widlarizer:
If what you were suggesting was replacing `unique_ptr<dict<...>>` with `dict<...>`, that would increase `sizeof(parser::value_type)` up from 16. I accidentally tested this with vector instead of dict, which increased the value type size to 32, a dict should be 48. This could be a performance or memory usage problem

higuoxing:
> If what you were suggesting was replacing `unique_ptr<dict<...>>` with `dict<...>`, that would increase `sizeof(parser::value_type)` up from 16. I accidentally tested this with vector instead of dict, which increased the value type size to 32, a dict should be 48. This could be a performance or memory usage problem

Ah, I didn't think of it before. Thanks for providing the information.

PR #5344, opened 2025-09-13
verilog_parser: replace manual AST node allocation with typed midrule actions

higuoxing:
Use Bison's typed midrule actions to construct AST_FCALL nodes with std::unique_ptr, replacing manual 'new' and extra->ast_stack management. This improves type safety, ensures proper ownership, and eliminates potential memory leaks.

Ref: https://www.gnu.org/software/bison/manual/html_node/Typed-Midrule-Actions.html


widlarizer (approved):
Makes sense, is already covered by existing tests, thanks

PR #5351, opened 2025-09-16
write_xaiger2: Fix output port mapping when opaque boxes are present

jix:

While working on adding inout / undirected connection support to bufnorm and trying to make sure those survive `abc_new` I ran into sporadic bugs where unrelated wires would end up getting connected. Eventually I was able to track it down to inconsistent numbering of aiger output ports between the actual logic in the aiger file and the generated port mapping file.

The changed code looked like an easy to make bug and making this change fixed it. I don't really understand the reasoning behind the way the aiger2 backend allocates ports, so I'm not sure this change is generally correct, but the observed inconsistencies disappeared.

I'm primarily looking for confirmation from @povik that this change is correct with respect to how this backend is designed. This should end up being exercised by a followup PR for the updated bufnorm handling.

### DIFF
--- a/backends/aiger2/aiger.cc
+++ b/backends/aiger2/aiger.cc
@@ -903,6 +903,16 @@ struct XAigerWriter : AigerWriter {
 	typedef std::pair<SigBit, HierCursor> HierBit;
 	std::vector<HierBit> pos;
 	std::vector<HierBit> pis;
+
+    // * The aiger output port sequence is COs (inputs to modeled boxes),
+    //   inputs to opaque boxes, then module outputs. COs going first is
+    //   required by abc.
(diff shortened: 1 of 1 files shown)


povik:
> If a module output is directly driven by a module input or an opaque box, the emission of the po statement in the mapping file is skipped.

Actually this can only happen for an opaque box, never a module input. Between two ports there will be a `$buf` which doesn't qualify as direct connection in the sense used here.

jix:
> I think this is right. I'd be good to also have the increment near L960 pulled out of the if condition for consistence.

Thanks for confirming this and also for the explanation. I'll merge this after updating it for consistency tomorrow.

PR #5358, opened 2025-09-17
help: fix memory leak for -dump-cells-json

georgerennie:
`dict::operator[]` inserts an empty element if needed (like `std::unordered_map`) so rather than separating the cases for whether the groups vector exists or not and creating a leaking vector we can just use `[]` directly, creating an empty vector if one doesn't exist yet and emplace name to the back of it.


KrystalDelusion (approved):
Thanks!  I think I knew that this was leaking at one point, but then I forgot about it because the help generation isn't covered in the CI san tests.  Testing `make docs/prep` with ASAN shows no more memory leak :)

PR #5362, opened 2025-09-18
Don't sort anything ever

widlarizer:
This PR adds one small commit on top of #5315. That commit removes `design->sort();` from all passes that unconditionally run it. I believe Yosys is deterministic with reproducible results with or without those. If abc isn't, then that's a more impactful problem than can be saved by sorting at any point in the flow.

The only test that ultimately has to be fixed with this change is `tests/arch/xilinx/dsp_cascade.ys` where the reordering of the design changes the count of some `FDRE` primitives received by synthesis. This change also greatly increases the runtime of `equiv_opt` in that one test despite the size of the SAT problem being exactly the same. Reduced reproducer for this slowdown:

```
design -reset
read_verilog <<EOT
module cascade(input clk, input [4:0] a, input [4:0] b, output reg [9:0] o);
reg [4:0] ar1, ar2, ar3, br1, br2, br3;
reg [9:0] m, n;
always @(posedge clk) begin
ar1 <= a;
ar2 <= ar1;
ar3 <= ar2;
br1 <= b;
br2 <= br1;
br3 <= br2;
m <= ar1 * br1;
n <= ar2 * br2 + m;
o <= ar3 * br3 + n;
end
endmodule
EOT
proc
design -save read

equiv_opt -assert -map +/xilinx/cells_sim.v synth_xilinx -noiopad
```

This PR **speeds up running synth on jpeg by 5%**, tested with 3 different values for `--hash-idx`. It also likely makes `--hash-idx` a more powerful hyper-optimization input. However it also makes the test suite take around 30% longer to complete on my machine due to the arbitrary change in oredering.


mmicko:
@widlarizer  Can we close this one since https://github.com/YosysHQ/yosys/pull/5621 handles it ?

widlarizer:
Technically we still sort in some backends, and the idea of a backend modifying the design in-place is upsetting and something I've addressed specifically in write_verilog and write_rtlil, but yeah I'm not reviving this PR on this branch

PR #5363, opened 2025-09-18
CI: Don't use self-hosted runner on forks

KrystalDelusion:

A few of the CI jobs use the YosysHQ self-hosted runner.  If someone makes a fork of Yosys (and enables actions), these jobs will be queued waiting for a self-hosted runner that probably doesn't exist (and if it does exist, it won't have the necessary configurations to be able to actually run Verific, which is generally what the YosysHQ self-hosted runner is used for).

Jobs targeting the self-hosted runner should only run if `github.repository == 'YosysHQ/Yosys'`.  This prevents the job from being queued on a fork entirely, without affecting PRs to this repo.

This PR is coming from a fork.  The self-hosted jobs should run for this PR, but not on the fork.

### DIFF
--- a/.github/workflows/prepare-docs.yml
+++ b/.github/workflows/prepare-docs.yml
@@ -26,7 +26,7 @@ jobs:
     # docs builds are needed for anything on main, any tagged versions, and any tag
     # or branch starting with docs-preview
     needs: check_docs_rebuild
-    if: ${{ needs.check_docs_rebuild.outputs.should_skip != 'true' }}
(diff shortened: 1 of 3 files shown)


widlarizer:
I think we also want this for `.github/workflows/update-flake-lock.yml`

mmicko (approved):
Looks good to me. Feel free to merge

KrystalDelusion:
@mmicko can you double check the changes for the other files look fine?  Specifically that `version.yml` now uses the same style of `if: github.repository` at the job level instead of the steps that bump/push the change, and that `wheels.yml` is okay to `build_wheels` but `upload_wheels` checks the repo.

PR #5368, opened 2025-09-20
Gowin. Renaming inputs of the DCS primitive.

yrabbit:
The dynamic clock selection (DCS) primitive has undergone changes with the release of the GW5A series—the CLK0,1,2,3 inputs are now CLKIN0,1,2,3, but only for GW5A series chips.

There are no functional changes, only renaming.

Here we are transferring the description of the DCS primitive from general to specialized files for each chip series.

We have also fixed a bug in the generation script that caused the loss of primitive parameters. Fortunately, this only affected the analog-to-digital converter, which has not yet been implemented.

Testing: in the apicula/examples directory, run `make clean; make`.

The entire compilation should complete without errors. This will test the pre-GW5A chips, DCS for the new series under development, and this PR is part of it.

### DIFF
--- a/techlibs/gowin/cells_sim.v
+++ b/techlibs/gowin/cells_sim.v
@@ -1959,14 +1959,6 @@ parameter REGULATOR_EN = 1'b0;
 endmodule
 
 (* blackbox *)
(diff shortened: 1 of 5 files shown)


yrabbit:
Pages from the Gowin IDE 1.9.1003 documentation. On the left for pre-GW5, on the right for GW5.
dcs

PR #5374, opened 2025-09-22
Fix building and running unit tests

mmicko:
This makes build on clean environment with  `ENABLE_LIBYOSYS=1 make unit-test -j9` working and running all the tests

### DIFF
--- a/.github/workflows/prepare-docs.yml
+++ b/.github/workflows/prepare-docs.yml
@@ -48,7 +48,7 @@ jobs:
           echo "ENABLE_VERIFIC_YOSYSHQ_EXTENSIONS := 1" >> Makefile.conf
           echo "ENABLE_CCACHE := 1" >> Makefile.conf
           echo "ENABLE_HELP_SOURCE := 1" >> Makefile.conf
-          make -j$procs ENABLE_LTO=1
+          make -j$procs
 
       - name: Prepare docs
         shell: bash
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -67,7 +67,7 @@ jobs:
           mkdir build
           cd build
           make -f ../Makefile config-$CC
-          make -f ../Makefile -j$procs ENABLE_LTO=1
+          make -f ../Makefile -j$procs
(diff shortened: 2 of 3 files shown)


tests/unit/kernel/rtlilTest.cc
```diff
@@ -385,7 +385,7 @@ namespace RTLIL {
 
 	TEST_P(WireRtlVsHdlIndexConversionTest, WireRtlVsHdlIndexConversion) {
 		std::unique_ptr<Module> mod = std::make_unique<Module>();
-		Wire *wire = mod->addWire(ID(test), 10);
+		Wire *wire = mod->addWire(NEW_ID, 10);
```
jix: This change hides an actual bug in the unit testing setup. We need to ensure that `IdString::ensure_prepopulated()` is called before running any test. There seems to be support for global set up and tear down that we could use: https://google.github.io/googletest/advanced.html#global-set-up-and-tear-down

KrystalDelusion:
[Clang sanitizers job was failing](https://github.com/YosysHQ/yosys/actions/runs/17926378024) due to `ENABLE_LTO=1` in the initial make, which was not included when trying to build `libyosys.so` and causing it to fail.  I pushed a fix and [it's re-running now](https://github.com/YosysHQ/yosys/actions/runs/17936184435/job/51002701098).

Also, is there a reason the unit tests only run in the sanitizer builds and not the regular ones?

jix:
@KrystalDelusion  During the Dev JF I pointed out that I consider it important to make sure that we run the unit tests with sanitizers. We didn't discuss whether we should also run them without, but I think it's actually fine to only run them with sanitizers as we're not shipping unit tests to end users so there isn't really the argument that we should also test a configuration that would end up in the release. There is an argument that we would get faster CI failures if we run the unit tests also without sanitizers, but I'm not sure how much impact it would have in practice.

mmicko:
It was not working at all on macOS due scripting in Makefile, so updated it, and running it again at https://github.com/YosysHQ/yosys/actions/runs/17939842441

PR #5380, opened 2025-09-23
CONTRIBUTING: add link to dev jf doc, replace slack with Discourse

widlarizer:
A view-only link to the dev jf doc is added instead of telling people to go through the abandoned public slack. I also added Discourse. Do we prefer Discourse over GH Discussions? If so, I'll also reorder those two

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -60,8 +60,8 @@ merge; please do not use these labels if you are not a maintainer.
 
 # Asking questions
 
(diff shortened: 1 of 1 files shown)


KrystalDelusion:
> Do we prefer Discourse over GH Discussions?

I think so, if only to keep more discussion in one place instead of spreading it around.  I think the distinction between user/developer is also cleaner on Discourse.

PR #5389, opened 2025-09-24
verific: New `-sva-continue-on-error` import option

jix:
This option allows you to process a design that includes unsupported SVA. Unsupported SVA gets imported as formal cells using 'x inputs and with the `unsupported_sva` attribute set. 

This enables limited processing of designs with unsupported SVA, e.g. to inspect the design hierarchy including all properties, or to explicitly select a subset of supported properties to perform FV with. To ensure that this can be safely used without the risk of silently ignored properties, `hierarchy -smtcheck` (used by SBY) will by error out if the user script leaves any unsupported properties around, making sure both parsing unsupported SVA and performing FV while ignoring unsupported SVA require separate opt-ins.

This mostly reuses existing internal debug functionality of verificsva to recover from unsupported SVA but adds additional logging error handling, emission of `unsupported_sva`-marked placeholder property cells and testing.

The added tests cover all new functionality. Since the added safeguards extend existing checks used by SBY that we can fully test within yosys, this requires no separate SBY changes or tests.

### DIFF
--- a/frontends/verific/verificsva.cc
+++ b/frontends/verific/verificsva.cc
@@ -124,6 +124,7 @@ struct SvaFsm
 {
 	Module *module;
 	VerificClocking clocking;
+	std::function<void (std::string)> parser_error;
 
 	SigBit trigger_sig = State::S1, disable_sig;
 	SigBit throughout_sig = State::S1;
@@ -148,6 +149,7 @@ struct SvaFsm
 		module = clking.module;
 		clocking = clking;
 		trigger_sig = trig;
+		parser_error = [](std::string msg){ log_error("%s", msg); };
 
 		startNode = createNode();
 		acceptNode = createNode();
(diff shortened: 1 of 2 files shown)


frontends/verific/verificsva.cc
nakengelhardt: This wasn't a requirement for this PR, I'm just curious since they encountered this earlier with a certain style of property: how difficult would it be to recover from this error as well?

passes/hierarchy/hierarchy.cc
```diff
@@ -1149,6 +1149,25 @@ struct HierarchyPass : public Pass {
...
+						continue;
+					if (!cell->has_attribute(ID(unsupported_sva)))
+						continue;
+
+					auto src = cell->get_src_attribute();
+
+					if (!src.empty())
+						src += ": ";
+
+					log_error("%sProperty `%s' in module `%s' uses unsupported SVA constructs. See frontend warnings for details, run `delete */a:unsupported_sva' to ignore.\n",
```
nakengelhardt: I would've recommended `chformal -remove a:unsupported_sva`, feels cleaner somehow, although I can't think of any difference it makes in practice...

jix: I agree that it looks cleaner so I pushed a commit to change it.

frontends/verific/verificsva.cc
jix: Not too difficult, I pushed another commit that implements this.

There is the issue that at this point we might have emitted a lot of logic already, so I also added code to roll back any cells and wires added while processing the failing property. Since SVA import only appends to RTLIL and the wires and cells collection are order preserving as long as no removals are performed, this was possible to do without instrumenting every single cell addition within verificsva.

PR #5391, opened 2025-09-25
Use to_bits() instead of bits() to avoid deprecation warning.

tomverbeure:

When I compile verific.cc, I get tons of warnings like this:

```
frontends/verific/verific.cc:1652:16: warning: 'bits' is deprecated: Don't use direct access to the internal std::vector<State>, that's an implementation detail. [-Wdeprecated-declarations]
 1652 |                                                         initval.bits()[bit_idx] = (*ascii_initdata == '0') ? State::S0 : State::S1;
      |                                                                 ^
./kernel/rtlil.h:888:4: note: 'bits' has been explicitly marked deprecated here
  888 |         [[deprecated("Don't use direct access to the internal std::vector<State>, that's an implementation detail.")]]
```

When I replace bits() by to_bits(), the warning goes away.

By replacing bits() by to_bits().

They need to run some tests with Verific enabled to make sure it works, but I compiled with my own Verific stub code instead of the real thing.

### DIFF
--- a/frontends/verific/verific.cc
+++ b/frontends/verific/verific.cc
@@ -1649,7 +1649,7 @@ void VerificImporter::import_netlist(RTLIL::Design *design, Netlist *nl, std::ma
 						if (*ascii_initdata == 0)
 							break;
 						if (*ascii_initdata == '0' || *ascii_initdata == '1') {
-							initval.bits()[bit_idx] = (*ascii_initdata == '0') ? State::S0 : State::S1;
+							initval.to_bits()[bit_idx] = (*ascii_initdata == '0') ? State::S0 : State::S1;
 							initval_valid = true;
 						}
 						ascii_initdata++;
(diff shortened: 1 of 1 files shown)


widlarizer:
Sorry for not keeping an eye on warnings in the verific builds and thanks for reporting this. However `to_bits()` returns a copy of the bit vector, not a reference like `bits()`. I'll put up a fixed version later

You can use `Const::set` instead, or construct and modify `std::vector<State>` that you then construct a `Const` from, or refactor with `Const::Builder` which basically only exposes `push_back(State)`. This might seem like a hassle but there's memory and runtime reasons for removing this direct bit vector access

mmicko:
This is done by https://github.com/YosysHQ/yosys/commit/2f8f421dee5d71a3c2c6ca0dbf09610b2b4d0e84 and additional sanity check in https://github.com/YosysHQ/yosys/commit/9570b395193691967491a23be8af3efd79d01866 (just in case something can gets wrong, but all tests passed without that as well)

Issue #5367, opened 2025-09-19
Issue with simlib.v, Unsupported: Verilog 1995 gate primitive: tran for Verilator


MrPoloGit:
### Version

0.57

### On which OS did this happen?

Linux

### Reproduction Steps

Issue from this commit
https://github.com/YosysHQ/yosys/commit/1251e92e3a988b33f2075118071f4454247041d6

1. Using Verilator 
```bash
Verilator 5.039 devel rev v5.038-8-gd89df33fc
```

2. Run folder of simlib.v
```bash
verilator --binary -f verilator.f --top-module simlib simlib.v
```

3. verilator.f
```bash

--timing
-j 0
-Wall
-Wno-fatal
--assert
--trace-fst
--trace-structs
--main-top-name "-"

--x-assign unique
--x-initial unique

-Werror-IMPLICIT
-Werror-USERERROR
-Werror-LATCH
```

### Expected Behavior

1. Parse all the Verilog files listed in verilator.f plus simlib.v.
2. Treat simlib as the top module.
3. Generate a C++ simulation model.
4. Link it into a runnable binary.
5. Able tp execute simulation.

### Actual Behavior

```bash
%Error-UNSUPPORTED: simlib.v:3229:6: Unsupported: Verilog 1995 gate primitive: tran
 3229 | tran connect[WIDTH-1:0] (A, B);
      |      ^~~~~~~
                    ... For error description see https://verilator.org/warn/UNSUPPORTED?v=5.039
%Error: Exiting due to 2 error(s)
```

Ravenslofty:
@jix do you know of any other ways of expressing that primitive? Verilator doesn't like it.

KrystalDelusion:
Can we do like `$pow` and stick the `$connect` cell in an `if(n)def`?

jix:
@Ravenslofty my answer to that is essentially no. I do know of another way that should work in principle, directly connecting both ports to the same wire using ANSI port declarations with an explicit expression:

```
module \$connect #(parameter WIDTH = 0) (inout .A(W), inout .B(W));
	wire [WIDTH-1:0] W;
endmodule
```

This isn't useful, though as neither Yosys's read_verilog, Icarus Verilog nor Verilator currently support that (Verific does, confirming my interpretation that the spec allows this).

[...]

PR #5393, opened 2025-09-25
Added ifndef SIMLIB_NOCONNECT

sifferman:

Fixes https://github.com/YosysHQ/yosys/issues/5367

I added in the `` `ifndef SIMLIB_NOCONNECT ``, as was mentioned in https://github.com/YosysHQ/yosys/issues/5367.

Now Verilator can correctly read the file with:

```bash
verilator -DSIMLIB_NOCONNECT -Wno-fatal --lint-only techlibs/common/simlib.v
```

Possibly the next step would be to create a test that verifies this never happens again. Maybe there could be a new file `".github/workflows/test-techlibs.yml"`, that runs Icarus and Verilator on each of the techlib cells_sim.v files.

### DIFF
--- a/techlibs/common/simlib.v
+++ b/techlibs/common/simlib.v
@@ -3219,6 +3219,8 @@ endmodule
 
 // --------------------------------------------------------
 //* group wire
+`ifndef SIMLIB_NOCONNECT
+
 module \$connect (A, B);
 
 parameter WIDTH = 0;
@@ -3230,6 +3232,7 @@ tran connect[WIDTH-1:0] (A, B);
 
 endmodule
 
+`endif
 // --------------------------------------------------------
 //* group wire
 module \$input_port (Y);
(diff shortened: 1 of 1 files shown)


jix:
I think for an automated test to verify that we never break verilator simulation flows again, we would need an additional define like `SIMLIB_VERILATOR_COMPAT` that implies `SIMLIB_NOCONNECT` and that we could extend for any future workaround that might be necessary for verilator compatibility.

If we do that, I wouldn't mind running it in CI, but I'm not sure if we'd want to do that as a separate workflow or integrate it into an existing workflow. @mmicko any thoughts?

jix:
I still want to add a `SIMLIB_VERILATOR_COMPAT` option to avoid future breaking changes in case there will be more incompatible constructs used, but I added the merge-before-release label to ensure we don't forget about merging at least some version of this

Issue #5398, opened 2025-09-29
`synth -hieropt` generates wrong code


sashimi-yzh:
### Version

Yosys 0.57+218 (git sha1 7ebd97216, clang++ 18.1.8 -fPIC -O3)

### On which OS did this happen?

Linux

### Reproduction Steps

```verilog
// CacheMeta.sv
module tag_2x4(	// @[src/main/CacheMeta.scala:19:16]
  input        R0_addr,
               R0_en,
               R0_clk,
  output [3:0] R0_data,
  input        W0_addr,
               W0_en,
               W0_clk,
  input  [3:0] W0_data
);

  reg [3:0] Memory[0:1];	// @[src/main/CacheMeta.scala:19:16]
  always @(posedge W0_clk) begin	// @[src/main/CacheMeta.scala:19:16]
    if (W0_en & 1'h1)	// @[src/main/CacheMeta.scala:19:16]
      Memory[W0_addr] <= W0_data;	// @[src/main/CacheMeta.scala:19:16]
  end // always @(posedge)
  assign R0_data = Memory[R0_addr];	// @[src/main/CacheMeta.scala:19:16]
endmodule

module CacheMeta(	// @[src/main/CacheMeta.scala:6:7]
  input        clock,	// @[src/main/CacheMeta.scala:6:7]
               reset,	// @[src/main/CacheMeta.scala:6:7]
  output [3:0] io_tagOut,	// @[src/main/CacheMeta.scala:11:14]
  input  [3:0] io_tagIn,	// @[src/main/CacheMeta.scala:11:14]
  input        io_ridx,	// @[src/main/CacheMeta.scala:11:14]
               io_widx,	// @[src/main/CacheMeta.scala:11:14]
               io_wen // @[src/main/CacheMeta.scala:11:14]
);

  tag_2x4 tag_ext (	// @[src/main/CacheMeta.scala:19:16]
    .R0_addr (io_ridx),
    .R0_en   (1'b1),	// @[src/main/CacheMeta.scala:6:7]
    .R0_clk  (clock),
    .R0_data (io_tagOut),
    .W0_addr (io_widx),
    .W0_en   (io_wen),
    .W0_clk  (clock),
    .W0_data (io_tagIn)
  );	// @[src/main/CacheMeta.scala:19:16]
endmodule
```

```tcl
# yosys.tcl
yosys -import
read_verilog -sv CacheMeta.sv
synth -top CacheMeta -hieropt
[...]

PR #5399, opened 2025-09-29
opt_hier: Fix two optimizations conflicting

povik:

Fix a conflict between the following two:

 * propagation of tied-together inputs in
 * propagation of unused inputs out

Tie-together optimization is made ignore unused inputs.

See included test case which is reduced from #5398


povik:
Thanks for the quick review

Issue #1579, opened 2019-12-16
verific front end core dump with sub module memory reference


mattvenn:
## Steps to reproduce the issue

* unpack attached directory
* cd verific-memory-module-issue
* sby -f test.sby

## Expected behavior

should be able to refer to another module's registers with . operator

## Actual behavior

`SBY 15:08:06 [test] Removing direcory 'test'.
SBY 15:08:06 [test] Copy 'test.v' to 'test/src/test.v'.
SBY 15:08:06 [test] Copy 'simple_mem.v' to 'test/src/simple_mem.v'.
SBY 15:08:06 [test] engine_0: smtbmc
SBY 15:08:06 [test] base: starting process "cd test/src; yosys -ql ../model/design.log ../model/design.ys"
SBY 15:08:06 [test] base: VERIFIC-WARNING [VERI-1927] test.v:13: port 'addr' remains unconnected for this instance
SBY 15:08:06 [test] base: VERIFIC-WARNING [VDB-1002] simple_mem.v:10: net 'mem' does not have a driver
SBY 15:08:06 [test] base: VERIFIC-WARNING [VDB-1002] test.v:16: net 'f_addr[7]' does not have a driver
SBY 15:08:06 [test] base: VERIFIC-WARNING [VDB-1002] test.v:17: net 'f_data[7]' does not have a driver
SBY 15:08:06 [test] base: VERIFIC-WARNING [VDB-1053] test.v:11: input port 'addr[7]' remains unconnected for this instance
SBY 15:08:06 [test] base: terminate called after throwing an instance of 'std::out_of_range'
SBY 15:08:06 [test] base: what():  dict::at()
SBY 15:08:06 [test] base: bash: line 1: 19786 Aborted                 (core dumped) yosys -ql ../model/design.log ../model/design.ys
SBY 15:08:06 [test] base: finished (returncode=134)
SBY 15:08:06 [test] base: job failed. ERROR.
SBY 15:08:06 [test] summary: Elapsed clock time [H:MM:SS (secs)]: 0:00:00 (0)
SBY 15:08:06 [test] summary: Elapsed process time [H:MM:SS (secs)]: 0:00:00 (0)
SBY 15:08:06 [test] DONE (ERROR, rc=16)
`

This goes away by un-commenting the flatten line in the .sby file

## tool versions

symbiotic-20191122A-symbiotic/
[...]

PR #5406, opened 2025-10-03
verific: print source location of problematic object on import error

nakengelhardt:
It's difficult to find out what caused import errors when using the verific frontend with just the (often autogenerated) name of the instance.

Extracted the source location formatting into its own function, and added a prefix to the error message if there is an object that can be pointed at as the likely culprit, and it has a source location.

I've found a reproducer of one of the import errors in #1579, would love to add more if anyone can figure out how to trigger the other errors...

### DIFF
--- a/frontends/verific/verific.cc
+++ b/frontends/verific/verific.cc
@@ -166,6 +166,25 @@ string get_full_netlist_name(Netlist *nl)
 	return nl->CellBaseName();
 }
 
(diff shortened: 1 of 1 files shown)


mmicko (approved):
LGTM

frontends/verific/verific.cc
mmicko, line 1998: Was first thinking if it should be added in this one as well, but think it is fine for skipping it in this case.

PR #5412, opened 2025-10-08
Add sdc command with smart flattening

widlarizer:
This PR adds the `sdc` pass which implements the uninterpreted call graph and basic smart flattening with `sdc -keep_hierarchy` [as described on Discourse](https://yosyshq.discourse.group/t/sdc-support-roadmap/95/2) by instrumenting an Tcl interpreter. The supported getter subset is `get_pins`, `get_nets` and `get_ports` that don't support wildcards or `-of_objects`. Non-getter commands are completely supported since they don't actually get interpreted. A simple test is included.


passes/cmds/sdc/sdc.cc
```diff
@@ -0,0 +1,785 @@
...
+		log("        Dump the referenced design objects.\n");
+		log("\n");
+		log("    -dump-graph\n");
+		log("        Dump the uninterpreted call graph.\n");
+		log("\n");
+		log("    -keep_hierarchy\n");
+		log("        Add keep_hierarchy attributes while retaining SDC validity.\n");
+		log("\n");
+	}
+	SdcPass() : Pass("sdc", "sniff at some SDC") { }
```
Ravenslofty: as funny as this is, perhaps you should change this before merging.

PR #5418, opened 2025-10-11
Gowin. Reduce the range of flip-flop types.

yrabbit:
UG303-1.0E_Arora Ⅴ Configurable Function Unit (CFU) User Guide.pdf specifies that the only flip-flop types supported in GW5 are DFFSE, DFFRE, DFFPE, and DFFCE.

However, the bit streams generated by the vendor IDE also contain DFF flip-flops, which are probably the result of optimisation, so we leave them in the list of permitted items, but add a flag that will allow the generation of completely correct output files, acceptable for further P& R using vendor tools (they will not allow the use of flip-flops other than the four specified in the netlist).

In the GW5 SemiDual Port BSRAM series, the primitive does not have RESETA and RESETB ports—they are replaced by the RESET port, so we separate the files for BSRAM generation, especially since in the future we may have to take into account other, as yet unexplored, differences in BSRAM.

The attached archive contains the text of a simple RISCV processor along with memory. The script compiles and shows the types of flip-flops used. The presence of a RESET pin instead of RESETA/RESETB can be seen in the generated files.
[tests.tar.gz](https://github.com/user-attachments/files/22863921/tests.tar.gz)

For comparison, the same design for Tangnano20k shows the use of DFFE, which is prohibited in any variant:
``` text
      209 submodules
      136   ALU
       38   DFFE
       35   DFFRE
```


yrabbit:
Selected excerpts from Gowin documentation
sdp
dffs

PR #5427, opened 2025-10-15
plugins: add search paths

donn:

https://github.com/YosysHQ/yosys/issues/2545 — On the nixpkgs side, there's been a hot-patch for this since 2021: https://github.com/NixOS/nixpkgs/commits/9720de3f1f7f8ad28cf3fafbf1349037ce2895d6/pkgs/development/compilers/yosys/plugin-search-dirs.patch

This uses the environment variable `YOSYS_PLUGIN_PATH` to provide multiple colon-delimited search paths for native plugins in a similar manner to `PATH` for executables and `PYTHONPATH` for Python modules.

I have updated `tests/various/plugin.sh` to test exactly this functionality.

---

Some notes:

- I did not test this on Windows. I don't know what the standard is over there.
- ~~Still testing in Nix, should have results shortly (after I open this PR)~~
   - Building fine https://github.com/fossi-foundation/nix-eda/pull/35/files
   
   ```diff
    -       --set NIX_YOSYS_PLUGIN_DIRS $out/share/yosys/plugins \
    +       --suffix YOSYS_PLUGIN_PATH : $out/share/yosys/plugins \
    ```
- I picked the name `YOSYS_PLUGIN_PATH` arbitrarily, but any other name would work.
- This is the second file in the codebase to use `std::filesystem` since Yosys is compiled under C++17 now. This shouldn't affect wasm because afaict plugins aren't available under wasm anyways…? But anyways, it makes a couple things a bit more idiomatic/neater. [EDIT: See @whitequark's post below for more context]

### DIFF
--- a/passes/cmds/plugin.cc
+++ b/passes/cmds/plugin.cc
@@ -22,6 +22,8 @@
 
 #ifdef YOSYS_ENABLE_PLUGINS
 #  include <dlfcn.h>
+#  include <filesystem>
+namespace fs = std::filesystem;
 #endif
 
 #ifdef YOSYS_ENABLE_PYTHON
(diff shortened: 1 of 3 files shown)


whitequark:
> * This is the second file in the codebase to use `std::filesystem` since Yosys is compiled under C++17 now. This shouldn't affect wasm because afaict plugins aren't available under wasm anyways…? But anyways, it makes a couple things a bit more idiomatic/neater.

I want to clarify that in the latest wasi-sdk, std::filesystem as well as threaded primitives like `<mutex>` should be fully supported even on single-thread targets. This was a major undertaking  for me and the result benefits the entire Wasm ecosystem including Yosys.

Wasm does have dynamic libraries available in principle, but I don't believe anybody has taken the effort to use them together with Yosys specifically. (I've modified plugins to be linked statically by hacking the Makefile where necessary...)

mmicko:
@donn regarding Windows compatibility it should work fine as is, it is just that on Windows for separator in path `;` is used instead of `:` since the other one is used when naming full paths using drive name as well (`D:\plugins` for example)

donn:
@mmicko Thanks for the tip, I went ahead and did that, though I based it on `preferred_separator` to help with Cygwin shenanigans.

mmicko (approved):

widlarizer:
Nice, with this we'll be down to zero patches needed in nixpkgs

mmicko:
@donn Did this patch https://github.com/YosysHQ/yosys/commit/db8c1878a0df1eaafaef698469bb91ac4118f285 to also make it work for mingw. This makes conversion of string safe and handles wchar_t to char conversion on Windows correctly.

donn:
@mmicko Thanks for the heads up! Very curious… `c_str()` does claim to return the path in native format:

https://en.cppreference.com/w/cpp/filesystem/path/native.html

PR #5430, opened 2025-10-16
sim: Make cycle width small as possible and configurable

mmicko:
Default size is made 2 since number must be even since we do add half cycle size on some places.

Since one test case did utilize time as well, needed to update it.

### DIFF
--- a/passes/sat/sim.cc
+++ b/passes/sat/sim.cc
@@ -2692,7 +2692,7 @@ struct SimPass : public Pass {
 	{
 		SimWorker worker;
 		int numcycles = 20;
-		int cycle_width = 2;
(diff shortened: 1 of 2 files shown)


jix:
Can we keep the default the same?

mmicko:
@jix Then it would require a way for a width to be defined and propagated from sby, and for that I was not sure on how many places it is actually used

jix:
I consider changing the default a significant regression, since this makes correlating timestamps and cycle counts require mental arithmetic which wasn't required before (very simple arithmetic, but still distracting when debugging a complicated FV failure).

I'll look into adding an option for this to SBY.

PR #5434, opened 2025-10-18
pyosys: uv for non-wheel builds, update instructions

donn:

@Ravenslofty noted that unlike the previous setup, users may have to manually manage a virtual environment to get the new Python build dependencies (pybind11 and cxxheaderparser,) and that is cumbersome.

While pybind11 can in fact be installed via package managers, it tends to be out of date more often than not. cxxheaderparser is not in any package manager that isn't Nix.

I considered just handling the ImportError to instruct the user on creating a venv but that's also just kind of cumbersome. 

- add `uv` to list of dependencies: saves builders from manually having to manage a venv for python build dependencies
  - when building wheels, pip automatically creates the environment with those dependencies, so no need for uv
  - when running simply `make ENABLE_PYOSYS=1`, this is not the case. people attempting to `pip3 install --upgrade pybind11 cxxheaderparser` to add it to their system packages will be met with a scare message about "breaking system packages"
  - if someone really prefers to provide their own dependencies (such as within Nix,) passing `PYOSYS_USE_UV=0` suppresses this behavior
- update installation instructions to drop boost and add tox instead

Naturally, uv is not required if you're not building Yosys with Python support.

On an Ubuntu 22.04 installation (Ubuntu 20.04 also in theory but the Bison on there is out of date so you have to build that on your own), running the instructions in Readme.md should be enough to allow `make ENABLE_PYOSYS=1` to build successfully without further intervention.

I validated that statement as follows:

```bash
docker run --rm -ti docker.io/ubuntu:22.04 bash -c "
    apt-get update &&
    DEBIAN_FRONTEND=noninteractive apt-get install -y git build-essential clang lld bison flex libfl-dev libreadline-dev gawk tcl-dev libffi-dev graphviz xdot pkg-config python3-dev zlib1g-dev curl &&
    curl -LsSf https://astral.sh/uv/install.sh | sh &&
    source \$HOME/.local/bin/env &&
    git clone --recursive --branch pyosys_tweaks https://github.com/donn/yosys &&
    cd yosys &&
    make ENABLE_PYOSYS=1 -j\$(nproc) &&
    ./yosys -H
"
```

---

* Also bec `macos-13` is being sunset by EOY, I updated the wheels CI script to use `macos-15` and `macos-15-intel` before I forget about them, and for consistency updated the other CI scripts as well. https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/


donn:
Rebased on `main` to pull in #5435

docs/source/getting_started/installation.rst
```diff
@@ -99,8 +99,7 @@ Installing all prerequisites for Ubuntu 22.04:
 
    sudo apt-get install gperf build-essential clang lld bison flex libfl-dev \
       libreadline-dev gawk tcl-dev libffi-dev git \
-      graphviz xdot pkg-config python3 libboost-system-dev \
-      libboost-python-dev libboost-filesystem-dev zlib1g-dev
+      graphviz xdot pkg-config python3-dev tox zlib1g-dev
```
mmicko: here tox is left

donn: argh, missed a spot. thanks

PR #5440, opened 2025-10-21
Gowin. Disable unsupported BSRAM mode in GW5A

yrabbit:
All supported (and planned to be supported) GW5A series chips do not have the mode 2: Read-before-Write write mode.

Here, we prohibit the generation of BSRAM with this mode.

To verify this, you can run the script from the attached archive.
The old version generated mode 2.
```shell
 ./go.sh
===============
  defparam \vram_mem.0.0 .WRITE_MODE0 = 0;
  defparam \vram_mem.0.0 .WRITE_MODE1 = 2;
  defparam \vram_mem.0.1 .WRITE_MODE0 = 0;
  defparam \vram_mem.0.1 .WRITE_MODE1 = 2;
  defparam \vram_mem.0.10 .WRITE_MODE0 = 0;
....
```
The new version should not do this.
[test.tar.gz](https://github.com/user-attachments/files/23013040/test.tar.gz)


yrabbit:
Excerpt from BSRAM documentation:
bsram-mode

Ravenslofty:
The PR as-is is fine, and I can merge it if you want me to, but I think there's a more concise way to do this: `memory_libmap` supports "defines", and so instead of switching out the default bram file for a gw5a one, you can gate read-first under a define.

yrabbit:
Let's merge it as is and I'll come back to a more concise way later - that way yosys will generate the correct design while I finish investigating another problematic element of BSRAM - a semidual port with a large data width - it doesn't work without a lot of auxiliary elements and I really don't want to lose my train of thought right now :)

PR #5449, opened 2025-10-29
Gowin. Fix GW5A ADCs.

yrabbit:
For these primitives, Gowin decided to use a different option for describing ports—directly in the module header, i.e.

``` verilog
module ADC(input CLK);
```

instead of
``` verilog
module ADC(CLK);
input CLK;
```

Since this one-time parser becomes too confusing, it is easier to simply add ADC descriptions as they are from a separate file, especially since these primitives are only available in the GW5A series.

Test:
``` shell
yosys -p "read_verilog top.v; synth_gowin -json top-synth.json -family gw5a"
```

The old version of Yosys simply won't compile the design due to the lack of port descriptions, while the new version will compile without errors.

[top.v.tar.gz](https://github.com/user-attachments/files/23200630/top.v.tar.gz)

### DIFF
--- a/techlibs/gowin/adc.v
+++ b/techlibs/gowin/adc.v
@@ -1,0 +1,260 @@
+/********ADC***********/
+//ADC for LRC,GW5AT-138K
+module ADCLRC (
+//Voltage signal source, /mV
+  input  ADCINBK2A,       //input from bank2 IO, adc_in_b is the reference
+  input  ADCINBK2B,       //input from bank2 IO, adc_in_b is the reference
+  input  ADCINBK3A,       //input from bank3 IO, adc_in_b is the reference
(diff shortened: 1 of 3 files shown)


mmicko:
Is `adc.v` separate file among gowin simulation files ? if so why there is a copy in this PR and should not .py script form a path, load it and include content ?

yrabbit:
No, Gowin ide does not have an adc.v file — it is just a piece of the same file that is parsed by the parser.

It's just that ADCs are described in a different style — with ports in the module() structure, although all other primitives are in the old style.
adc-ide



I want to say that it is not worth the effort to change the parser to process only ADCs — it is easier to separate the description into a separate file.

PR #5454, opened 2025-11-01
Shuffling and documenting prerequisites

KrystalDelusion:

While responding to a [Discourse topic](https://yosyshq.discourse.group/t/adding-gmock-to-test-dependencies/98) I remembered that I wanted to split the package dependencies to improve clarity and distinguish between requirements for building and testing (and to a lesser degree documenting).  I also discovered the [cache-apt-packages action](https://github.com/marketplace/actions/cache-apt-packages), and while setting up the build environment isn't a huge overhead, it is done very frequently and so caching it seems reasonable.  And iverilog was previously being installed/cached in two separate places, so we can reduce the redundancy there.

- Cache apt package installs
- Split (Linux) dependencies in the `setup-build-env` action
  - Only install packages required for the task, e.g. running the tests doesn't need all of the build prereqs
  - Move iverilog install to a separate action, called during `setup-build-env`
- Update existing documentation for prereqs
- Add basic documentation for tests (in particular for their prereqs)
- Drop boost-python (no longer needed since #5370)

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -25,13 +25,13 @@ jobs:
               archs: "aarch64",
             },
             {
-              name: "macOS 13",
+              name: "macOS 15 x64",
               family: "macos",
               runner: "macos-15-intel",
               archs: "x86_64",
             },
             {
-              name: "macOS 14",
+              name: "macOS 15 arm64",
               family: "macos",
               runner: "macos-15",
               archs: "arm64",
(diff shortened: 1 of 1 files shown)


mmicko:
Not to overload this PR too much, but think moving to macos-14 or macos-15-intel would be needed since old one is getting deprecated, there are already changes needed added in https://github.com/YosysHQ/yosys/pull/5434 but that one is on hold. For wheels I would keep what @donn added (so 15) but for compile and tests not sure, but looks like 15 is more available so better move to it.

donn:
yeah sorry i haven't had the time to get back to 5434 yet, will do when i can

KrystalDelusion:
That was actually on my to-do list for today :)
For the non-wheel CI jobs it's only `test-compile` which uses `macos-13` and that should just be a one-line change, since we already use `macos-latest` everywhere else.

mmicko (approved):

PR #5456, opened 2025-11-02
proc_mux: use assignment action locations for mux src

widlarizer:
Synthesis users often aim try to reduce the FPGA resource usage or chip area by tracking down what in their designs is causing cells to be emitted by looking at the src attributes Yosys emits. Prior to this PR, the multiplexers created by `proc_mux` to select the correct values from cases of Verilog switches would be labeled with the source range of the entire switch together with the source range of the entire case (see Note 1). After this PR, the src attribute for such a multiplexer covers exactly the assignments that it implements.

To do this, I'm introducing "RTLIL" state that can't be roundtripped through textual RTLIL, specifically, the `Const src` field of the new `struct SyncAction`. Generally the `proc` passes are considered inseparable from `read_verilog`, `proc` converts `Process`es into cells. I don't believe there is a use case where a user would like to print out RTLIL with processes in it and actually do anything with it. OpenROAD flow scripts does write out canonical pre-synthesis RTLIL but running proc ahead of this seems to be a valid change (see Note 2). If it's attempted, the RTLIL backend has a warning that action sources will be lost. When they are lost, the case src range is used, which I consider still an upgade over the current state of things. I'm also open to alternative approaches.

Before undrafting:

- [x] Fix verific
- [x] Evaluate performance impact and improve if needed
- [x] Write notes about how `proc_mux` works

Note 1: Also the latter of which was referring only to the end of the case. For example, the `$pmux` at `module tiny2` from the included `tests/proc/proc_mux_src.v`.
Note 2: See https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/b2918bafbddf77a231fdd5c530cf9d72c1737199/flow/scripts/synth_canonicalize.tcl which refers to https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/b2918bafbddf77a231fdd5c530cf9d72c1737199/flow/scripts/synth_preamble.tcl

### DIFF
--- a/passes/proc/proc_dff.cc
+++ b/passes/proc/proc_dff.cc
@@ -157,6 +157,20 @@ void gen_dff(RTLIL::Module *mod, RTLIL::SigSpec sig_in, RTLIL::Const val_rst, RT
 	log(".\n");
 }
 
+
+template <typename T>
+static void error_at_src(const RTLIL::Const* src, T msg) {
+	if (!src || src->empty()) {
+		log_error("%s\n", msg);
+	} else {
+		log_error("%s: %s\n", src->decode_string(), msg);
+	}
+}
+template <typename T>
+static void error_at_proc(const RTLIL::Process* proc, T msg) {
+	error_at_src(proc->attributes.count(ID::src) ? &proc->attributes.at(ID::src) : nullptr, msg);
+}
+
 void proc_dff(RTLIL::Module *mod, RTLIL::Process *proc, ConstEval &ce)
 {
 	while (1)
@@ -194,14 +208,15 @@ void proc_dff(RTLIL::Module *mod, RTLIL::Process *proc, ConstEval &ce)
 				async_rules.emplace_back(rstval, sync);
 			}
 			else if (sync->type == RTLIL::SyncType::STp || sync->type == RTLIL::SyncType::STn) {
-				if (sync_edge != NULL && sync_edge != sync)
-					log_error("Multiple edge sensitive events found for this signal!\n");
+				if (sync_edge != NULL && sync_edge != sync) {
+					error_at_src(&action.src, stringf("Multiple edge sensitive events found for signal %s", log_signal(sig)));
+				}
 				sig.replace(action.lhs, action.rhs, &insig);
 				sync_edge = sync;
 			}
 			else if (sync->type == RTLIL::SyncType::STa) {
 				if (sync_always != NULL && sync_always != sync)
-					log_error("Multiple always events found for this signal!\n");
+					error_at_proc(proc, stringf("Multiple always events found for signal %s", log_signal(sig)));
 				sig.replace(action.lhs, action.rhs, &insig);
 				sync_always = sync;
 			}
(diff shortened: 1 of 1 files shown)


whitequark (changes requested):
> I don't believe there is a use case where a user would like to print out RTLIL with processes in it and actually do anything with it.

Amaranth emits RTLIL with processes in it and would emit fine-grained `src` attributes if it was available. This functionality should be preserved in textual RTLIL.

widlarizer:
Oh, I thought Amaranth was emitting flops to avoid `proc`. I don't want to give preferential treatment to Verilog frontends. This might be best exposed with textual RTLIL v2, which is a good opportunity to get other improvements in. I'll put up a Discourse thread when I gather my thoughts. Exposing this in textual makes sense as a followup but I won't push on this PR until we have some idea what that might even look like

whitequark:
> Oh, I thought Amaranth was emitting flops to avoid `proc`.

Amaranth still uses processes for decision trees (and there is no reasonable way of avoiding that; we are not going to implement `proc_mux` in Amaranth backend, that would be a preposterous violation of layering and also really inefficient). It does not use sync rules; it basically only uses `proc` for the equivalent of `always @*` blocks. (It makes use of the behavior where a process without `sync always` rules still generates an `always @*` equivalent logic.)

We run `proc -nomux -norom` in the Verilog backend because we want to make use of the Verilog backend behavior of serializing switches as-is.

passes/proc/proc_mux.cc
```diff
@@ -28,27 +28,46 @@
...
+ * Actions are assignments in cases of switches.
+ * Snippets are non-overlapping slices of signals on the left-hand side
+ * of actions. For example, if you have these two actions for r[7:0]:
+ *   r = a; r[1:2] = 2'b0;
+ * you will arrive at three snippets:
+ *   r[0], r[2:1], r[7:3]
+ * For each snippet, multiplexers ($mux or $pmux) are emitted
+ * based on the full switch.
+ * $pmux are only emitted when legal based on whether the cases can be
+ * evaluated in parallel. Bad HDLs allow users to build cases that can't.
```
whitequark: I don't think there's anything bad about having to build a priority matcher (at least assuming you mean "evaluated in parallel" here in the same sense as `(*parallel_case*)`. Indeed, how else would you naturally express a priority matching construct if not with a ladder-shaped switch?

widlarizer: Yes, my bone to pick is that it's too easy to build them by accident. I'll just leave this out

widlarizer:
I'm happy with the PR's current scope, but before it hits main, there's more to be worked out. I have retargeted this PR into a larger feature branch that I'll be working on top of to improve the handling of source locations across Yosys. This will require fairly drastic changes that don't make sense to review all at once. The topic of this PR is a part of the motivation for those

docs/source/using_yosys/more_scripting/interactive_investigation.rst
```diff
@@ -300,7 +300,7 @@ example, calling :yoscrypt:`dump $2` after the :yoscrypt:`cd example` above:
 
 .. literalinclude:: /code_examples/show/example.out
    :language: RTLIL
-   :start-after: yosys [example]> dump
+   :start-after: Warning: Module example contains processes. Case action sources attributes will be lost.
```
KrystalDelusion, line 303: I haven't looked into the code beyond a very cursory glance to see where this warning is produced, but "Case action sources attributes" feels like a mouthful and maybe bad grammar? Specifically the "sources attributes"

widlarizer:
The PR is targeting a feature branch, we will figure out textual RTLIL representation for the added source locations at a later stage, when we have twine-based "fused locations", since that will also extend the textual RTLIL format
[sc-267]

PR #5460, opened 2025-11-03
timeest: Add top ports launching/sampling

povik:

Make `timeest` not ignore paths starting or ending on a module port.

Without the `-clk` option the command now considers all flops and all module ports for candidate startpoints or endpoints.

I have attached timeest.ys as a smoke test


povik:
Thanks for review!

PR #5467, opened 2025-11-05
libparse: support unquoted expressions

### LINKED ISSUE #5451: LibertyParser fails on latch enable with parens, but no enclosing double quotes
### Version

Yosys 0.58+89 (git sha1 75eff54b3, g++ 13.3.0-6ubuntu2~24.04 -fPIC -O3)

Linux

### Reproduction Steps

[hack.lib.txt](https://github.com/user-attachments/files/23246073/hack.lib.txt)
read_liberty hack.lib

### Expected Behavior

Successful parsing of expression

### Actual Behavior

The offending statement is:

```
    latch (IQ,IQN) {
      data_in : "D";
      /*
        | double quotes | parens | Example | yosys result | ORFS result |
	| ------------- | ------ | ------- | ------------ | ----------- |
	|      N        |   N    |   CLK   |    passes    |   passes    |
	|      N        |   Y    |  (CLK)  |    fails     |   passes    |
	|      Y        |   N    |  "CLK"  |    passes    |   passes    |
	|      Y        |   Y    | "(CLK)" |    passes    |   passes    |
      */
      enable : (CLK);
    }
```

which was effectively what I see in a proprietary Liberty file provided by an IP vendor.

```
% yosys

 Yosys 0.58+89 (git sha1 75eff54b3, g++ 13.3.0-6ubuntu2~24.04 -fPIC -O3)

yosys> read_liberty hack.lib

1. Executing Liberty frontend: hack.lib
ERROR: Syntax error in liberty file on line 35.
```

The same file goes through openroad's liberty reader cleanly.

widlarizer:
Fixes #5451. The [Liberty file format](https://people.eecs.berkeley.edu/~alanmi/publications/other/liberty07_03.pdf) isn't sufficiently consistent or explicit in how it defines its syntax. As a result, PDK authors enabled by proprietary tool vendors, expect some unreasonable things to work. Previously, we allowed expressions like `"A B"` to be interpreted as `"A&B"` in #5219 in the expression parsing code added in #4698 to support detecting DFFs with enables. We also previously had a separate special case for parsing identifiers like `A` as if they were correctly quoted like `"A"`. However, #5451 shows that PDK authors want general expressions like `(A)` to also be valid, so I infer they would also expect `A B '` to be just as valid as `"A & B'"`.

Note: the way our code is structured, boolean expression parsing is separate from liberty AST parsing. I make this distinction explicit in the following paragraph

This PR implements this by removing the special case from the AST lexer, and when AST-parsing values after colons, if there are multiple string tokens, concatenating them together with spaces. RHS of `next_state : D EN;` gets AST-lexed as string `D`, string `EN`, and then AST-parsed as `D EN`. When this expression is actually used in Yosys, it gets expr-parsed into the corresponding binop. Prior to this PR, expression parsing was only done in dfflibmap, since it was fine for filterlib to dump expr strings into .lib verbatim, and trivially translate into valid Verilog. However, because our AST lexer discards whitespace, which we then have to reintroduce to ensure the separate expression parsing works out, we have to use the expression parser in filterlib as well, and write it out in a Verilog format, implemented with the new method `LibertyExpression::vlog_str`.

See `tests/liberty/unquoted.lib` for the direct changes. Almost all of the `*.*.ok` diff is due to fact that dumping out parsed expressions with `filterlib -verilogsim` prints them in a normalized format.

### DIFF
--- a/passes/techmap/libparse.cc
+++ b/passes/techmap/libparse.cc
@@ -676,23 +676,18 @@ LibertyAst *LibertyParser::parse(bool top_level)
 					tok = lexer(str);
 				} else {
 					// Hack for when an expression string is unquoted
-					// std::cout << "consume_wrecked_str from :\n";
-					// std::cout << " weh " << str << "\n";
 					tok = consume_wrecked_str(tok, ast->value);
 				}
 			} else if (tok == '(') {
 				// Hack for when an expression string is unquoted and starts with
 				// parentheses
-				// tok = '';
-				// ast->value = "(";
-				// std::cout << "consume_wrecked_str from (\n";
 				tok = consume_wrecked_str(tok, ast->value);
 			}
 			while (tok == '+' || tok == '-' || tok == '*' || tok == '/' || tok == '!') {
 				ast->value += tok;
 				tok = lexer(str);
 				if (tok != 'v')
-					error("one");
+					error();
 				ast->value += str;
 				tok = lexer(str);
 			}
@@ -704,7 +699,7 @@ LibertyAst *LibertyParser::parse(bool top_level)
 			if ((tok == ';') || (tok == 'n'))
 				break;
 			else
-				error("two");
+				error();
 			continue;
 		}
 


maliberty:
I think everyone who has implemented a Liberty parser goes through this endless series of surprises about non-conforming syntax that is accepted by the main vendor defining the format.  Whatever they accept is de-facto the format unfortunately.

Issue #5246, opened 2025-07-22
FSM extraction for FSMs with asynchronous reset


georgerennie:
### Feature Description

It seems that `fsm` doesn't optimize FSMs if their reset is asynchronous.

E.g. consider the following FSM with an async reset
```verilog
module sequence_detector (
    input  wire clk,
    input  wire reset,
    output wire s1
);
    localparam [7:0] RST = 8'b10010010;
    localparam [7:0] S1 = 8'b01001000;
    localparam [7:0] S2 = 8'b11000111;

    reg [7:0] current_state, next_state;
    always @(posedge clk or posedge reset) begin
        if (reset) begin
            current_state <= RST;
        end else begin
            current_state <= next_state;
        end
    end

    always @(*) begin
        next_state = current_state;

        case (current_state)
            RST: next_state = S1;
            S1: next_state = S2;
            S2: next_state = S1;
            default: next_state = RST;
        endcase
    end

    assign s1 = next_state == S1;
endmodule
```

Looking at the output of `synth` on this, we get the following warning
```
Not marking sequence_detector.current_state as FSM state register:
    Circuit seems to be self-resetting.
```

If you change this to a sync reset, the FSM is detected and recoded instead.

If we run `fsm_detect` with `-ignore-self-reset` in `synth`, the FSM is now correctly detected and recoded, but I'm not sure if turning this on in the general case preserves correctness. Perhaps it would be better to instead augment `fsm_detect` with the ability to understand async resets as a present reset, as it seems `fsm_extract` is able to handle them correctly.

whitequark:
It also doesn't work for FSMs with initial value, making the whole FSM pass suite basically useless in practice.

PR #5472, opened 2025-11-06
fsm_detect: add adff detection

Anhijkt:
[5246](https://github.com/YosysHQ/yosys/issues/5246) 
This code checks if arst signal of fsm's state register depends on it being in invalid state.


whitequark:
Asking here since the people involved are probably the most up-to-date ones on FSM in Yosys: do we have detection of FSM with resets due to any of the recent changes?

PR #5473, opened 2025-11-07
Handle unsized params

### LINKED ISSUE #5465: SystemVerilog unsized literal '1 incorrectly evaluated as 1-bit value instead of filling all bits
### Version

Yosys 0.58+84 (git sha1 f5c9e122d, g++ 15.2.1 -march=x86-64 -mtune=generic -O2 -fno-plt -fexceptions -fstack-clash-protection -fcf-protection -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -fPIC -O3)

Linux

### Reproduction Steps

```sv
read_verilog -sv <<EOF
module top (
    output logic [3:0] o
);
   localparam logic [3:0] L1 = '1;
   assign o = L1;
   always_comb assert(o == 4'b1111);
endmodule
EOF

prep -top top
chformal -lower
select top
sat -prove-asserts -verify
```

### Expected Behavior

From what I understand of the IEEE-1800-2023 paragraph 5.7.1 Integer literal constants

> An unsized unsigned single-bit value can be specified by preceding the single-bit value with an apostrophe
( ' ), but without the base specifier. All bits of the unsized value shall be set to the value of the specified bit. 
> In a self-determined context, it shall have a width of 1 bit.
> '0, '1, 'X, 'x, 'Z, 'z // sets all bits to specified value

Then the check should succeed, as o should take the value 'b1111

### Actual Behavior

Instead it takes the value 'b0001

KrystalDelusion:

Fix #5465.

```
[read_verilog -sv|verific -sv|read_slang] << EOF
module set_param #(
    parameter logic [3:0] VALUE
) (
    output logic [3:0] out
);
    assign out = VALUE;
endmodule

module top (
);
    localparam [3:0] a = '1;
    wire [3:0] b;
    wire [3:0] c = '1;
    set_param #('1) sp (b);
    always @* begin
        assert(a == 4'b1111);
        assert(b == 4'b1111);
        assert(c == 4'b1111);
    end
endmodule
EOF
prep -top top -flatten
chformal -lower
select top
sat -prove-asserts -verify -show-all
```

Running the above script (with main) with `read_verilog` only passes the assertion on `c`, while both `verific` and `read_slang` pass all three.

For the `localparam [3:0] a = '1;` case we check the `is_unsized` field, using the same logic as the `AST_CASE_SIZE`: https://github.com/YosysHQ/yosys/blob/5d0847f6fbef980914134a6a98b2397ecb0a3189/frontends/ast/simplify.cc#L4394-L4403

`set_param #('1) sp (b);` is a bit more difficult, requiring a new `CONST_FLAG_UNSIZED` to preserve the `is_unsized` field from the AST in the `RTLIL::Const` struct.

Tests included and run by CI.

### DIFF
--- a/backends/rtlil/rtlil_backend.cc
+++ b/backends/rtlil/rtlil_backend.cc
@@ -61,7 +61,9 @@ void RTLIL_BACKEND::dump_const(std::ostream &f, const RTLIL::Const &data, int wi
 				return;
 			}
 		}
-		f << stringf("%d'", width);
+		if ((data.flags & RTLIL::CONST_FLAG_UNSIZED) == 0) {
+			f << stringf("%d'", width);
+		}
 		if (data.flags & RTLIL::CONST_FLAG_SIGNED) {
 			f << stringf("s");
 		}
@@ -172,9 +174,10 @@ void RTLIL_BACKEND::dump_cell(std::ostream &f, std::string indent, const RTLIL::
 	dump_attributes(f, indent, cell);
 	f << stringf("%s" "cell %s %s\n", indent, cell->type, cell->name);
 	for (const auto& [name, param] : reversed(cell->parameters)) {
-		f << stringf("%s  parameter%s%s %s ", indent,
+		f << stringf("%s  parameter%s%s%s %s ", indent,
 				(param.flags & RTLIL::CONST_FLAG_SIGNED) != 0 ? " signed" : "",
 				(param.flags & RTLIL::CONST_FLAG_REAL) != 0 ? " real" : "",
+				(param.flags & RTLIL::CONST_FLAG_UNSIZED) != 0 ? " unsized" : "",
 				name);
 		dump_const(f, param);
 		f << stringf("\n");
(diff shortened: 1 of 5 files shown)


widlarizer:
Could this be implemented without introducing unsized constants into RTLIL by canonicalizing constants within the frontend? How much context would be necessary to do that?

KrystalDelusion:
I might be missing something, but I think avoiding the unsized constants in RTLIL would instead require `RTLIL::Module` to track parameter types (or at least widths), which starts to get tricky when parameter types are determined by other parameters

KrystalDelusion:
Or refactor module derivation so that it may be called with the instantiating AST rather than a `const dict<RTLIL::IdString, RTLIL::Const> &parameters`, so that the parameters can be resolved during derivation.  Something similar to interfaces and the `IFExpander` struct may work instead of tracking parameter types in the `RTLIL::Module`, but it would still require some way to handle recursive types for things like
```systemverilog
#(
    parameter NBITS,
    parameter logic [NBITS-1:0] VALUE
)
```
including the case where derivation is skipped because there are parameters without default values.

PR #5492, opened 2025-11-19
pyosys: __getitem__ for supported classes

donn:

This line no longer worked: https://github.com/donn/nl2bench/blob/e1fcf29dab450dce6e87a84d5d3466f1ae296a6e/nl2bench/nl_parser.py#L142

745222fa3bf2ac570935ffd044c86fa2eb12d123 made it so the result of `.bits()` was not iterable in Python.

Furthermore, after fixing it, this line failed:

https://github.com/donn/nl2bench/blob/e1fcf29dab450dce6e87a84d5d3466f1ae296a6e/nl2bench/nl_parser.py#L87

This was I think a regression by myself.

- functions that have a const `[]` operator method now support `__getitem__` in Python
- fields of a pointer type now return a `reference_internal` instead of a `copy` because classes referenced to by pointers typically aren't copyable (e.g. RTLIL::Wire, RTLIL::Module, etc)
- removed duplicate pyosys test

A new test has been added. I also updated an existing test that is now failing…? Not sure who's maintaining the .ys version of this test though.


mmicko:
Seems that Design was also a copy, since checks were passing but should have failed.
No worries for .ys version that one is fine and passing but pyosys test was not actually testing anything in this case.
What was exact issue on macOS ?

mmicko:
Just tried on my macOS environment and it was running fine on it (commented line to skip test in case of macOS).

Please note that it could be due to environment settings I had to point `PYTHONPATH` to explicit location where plugin is installed in order for it to work, so it could be it was pulling some older version for you.

`export PYTHONPATH=/opt/homebrew/opt/python@3.14/Frameworks/Python.framework/Versions/3.14/lib/python3.14/site-packages`

donn:
~~Issue appears to persist and unfortunately it IS pointing at the right file.~~

~~This could be any number of things, from the Nix environment to the Darwin version to the libc++ version…~~

```
$ uname -a
Darwin Donns-Mac-mini-M4.local 25.1.0 Darwin Kernel Version 25.1.0: Mon Oct 20 19:32:56 PDT 2025; root:xnu-12377.41.6~2/RELEASE_ARM64_T8132 arm64 arm Darwin
$ c++ --version
clang version 19.1.7
Target: arm64-apple-darwin
Thread model: posix
InstalledDir: /nix/store/5hgabdyqh22l57cb3qgx10lcc40d5n9p-clang-19.1.7/bin
$ python3 ./tests/pyosys/test_ecp5_addsub.py | tee out.log
/Volumes/T7/Code/yosyshq/yosys-tree/main/venv/lib/python3.12/site-packages/pyosys/__init__.py
[…]
```

~~[out.log](https://github.com/user-attachments/files/23632676/out.log)~~

~~I'll revert it for now. I can investigate this in a later PR.~~

~~EDIT: To be exact, tests/arch/ecp5/add_sub.ys also fails with the same exact error on this machine.~~

EDIT 2: ABC was contaminated. False alarm.

mmicko (approved):

PR #5496, opened 2025-11-20
read_liberty: support loopy retention cells

### LINKED ISSUE #5491: Can't resolve wire name in proprietary Liberty file
### Version

Yosys 0.59+0 (git sha1 26b51148a, g++ 13.3.0-6ubuntu2~24.04 -fPIC -O3)

Linux

### Reproduction Steps

read_liberty cell_with_q2_issue.lib

```
library (cell_with_q2_issue) {
    delay_model : table_lookup;
    voltage_unit : 1V;
    current_unit : 1mA;
    leakage_power_unit : 1nW;
    time_unit : 1ns;
    capacitive_load_unit (1, pf);
    pulling_resistance_unit : 1kohm;
    input_threshold_pct_rise : 50;
    input_threshold_pct_fall : 50;
    output_threshold_pct_rise : 50;
    output_threshold_pct_fall : 50;
    slew_lower_threshold_pct_rise : 30;
    slew_upper_threshold_pct_rise : 70;
    slew_upper_threshold_pct_fall : 70;
    slew_lower_threshold_pct_fall : 30;
    cell ("cell_with_q2_issue") {
        ff (Q1,QN1) {
            clocked_on : "CK";
            next_state : "(D * !SE + SI * SE)";
            clear : "(((!B2B) * !Q2) + !RD)";
            preset : "((!B2B) * Q2)";
            clear_preset_var1 : "L";
            clear_preset_var2 : "H";
        }
        latch (Q2,QN2) {
            enable : "B1";
            data_in : "Q1";
        }
        pin (B1) {
            direction : input;
        }
        pin (B2B) {
            direction : input;
        }
        pin (CK) {
            clock : true;
            direction : input;
        }
        pin (D) {
            direction : input;
        }
        pin (Q) {
            direction : output;
            function : "Q1";
        }
        pin (RD) {
            direction : input;
        }
        pin (SE) {
            direction : input;
        }
        pin (SI) {
            direction : input;
        }
    }
}
```

### Expected Behavior

No parse error

### Actual Behavior

```
$ yosys

 Yosys 0.59+0 (git sha1 26b51148a, g++ 13.3.0-6ubuntu2~24.04 -fPIC -O3)

yosys> read_liberty cell_with_q2_issue.lib 

1. Executing Liberty frontend: cell_with_q2_issue.lib
ERROR: Can't resolve wire name Q2 in module \cell_with_q2_issue.
```

Note that I modified the error message to include the offending cell. Would be great if you could update the error message, as well. Otherwise, it's not clear which cell has the issue.

widlarizer:
Fixes #5491. Retention cells seem to be advanced PDK cells that allow retention of flop state, usually while the main power domain is off. Balloon retention cells loop the output of a DFFSR through a latch to its input. For this reason, to construct them correctly in `read_liberty`, we have to create `latch` and `ff` group state wires before we build the logic that models them.

Additionally, to test this change, this PR modifies `filterlib -verilogsim` to avoid emitting unsynthesizable verilog when possible. This is possible whenever both `clear_preset_var*` are defined as either low or high by emitting a process for either state variable separately and adjusting the precedence of clear and preset.

This allowed me to cover the new functionality with `tests/liberty/read_liberty.ys` which proves equivalence between the cell model for `retention.lib` created by `read_liberty` and by `filterlib -verilogsim`.

### DIFF
--- a/tests/liberty/dff.lib.verilogsim.ok
+++ b/tests/liberty/dff.lib.verilogsim.ok
@@ -5,11 +5,9 @@ module dff (D, CLK, Q);
   output Q;
   assign Q = IQ; // IQ
   always @(posedge CLK) begin
-      // D
       IQ <= D;
   end
   always @(posedge CLK) begin
-      // ~(D)
       IQN <= ~(D);
   end
 endmodule
--- a/tests/liberty/normal.lib.verilogsim.ok
+++ b/tests/liberty/normal.lib.verilogsim.ok
@@ -58,7 +58,6 @@ module dff (D, CLK, RESET, PRESET, Q, QN);
       IQ <= 1;
     end
     else begin
-      // D
       IQ <= D;
     end
   end
@@ -70,7 +69,6 @@ module dff (D, CLK, RESET, PRESET, Q, QN);
       IQN <= 1;
     end
     else begin
-      // ~(D)
       IQN <= ~(D);
     end
   end
--- a/tests/liberty/read_liberty.ys
+++ b/tests/liberty/read_liberty.ys
@@ -1,0 +1,9 @@
+read_liberty retention.lib
+rename retention_cell retention_cell_lib
+read_verilog retention.lib.verilogsim
+proc
+rename retention_cell retention_cell_vlog
+async2sync
+equiv_make retention_cell_lib retention_cell_vlog equiv
+equiv_induct equiv
+equiv_status -assert equiv
--- a/tests/liberty/retention.lib
+++ b/tests/liberty/retention.lib
@@ -1,0 +1,57 @@
+library (retention) {
+    delay_model : table_lookup;
+    voltage_unit : 1V;
+    current_unit : 1mA;
+    leakage_power_unit : 1nW;
+    time_unit : 1ns;
+    capacitive_load_unit (1, pf);
+    pulling_resistance_unit : 1kohm;
+    input_threshold_pct_rise : 50;
+    input_threshold_pct_fall : 50;
+    output_threshold_pct_rise : 50;
+    output_threshold_pct_fall : 50;
+    slew_lower_threshold_pct_rise : 30;
+    slew_upper_threshold_pct_rise : 70;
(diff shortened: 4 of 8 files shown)


frontends/liberty/liberty.cc
```diff
@@ -646,6 +656,13 @@ struct LibertyFrontend : public Frontend {
 			{
 				// some liberty files do not put ff/latch at the beginning of a cell
 				// try to find "ff" or "latch" and create FF/latch _before_ processing all other nodes
+				// but first, in case of balloon retention cells, we need all ff/latch output wires
```
gussmith23, line 659: will the phrase "balloon retention cells" make sense to people reading this code in the future? i'm wondering if this comment can be a bit more general (i.e., mention the fact that this is to support things with loops, if that's indeed what it's doing)

frontends/liberty/liberty.cc
```diff
@@ -646,6 +656,13 @@ struct LibertyFrontend : public Frontend {
...
+				// defined before we add ff/latch cells
+				for (auto node : cell->children)
+				{
+					if ((node->id == "ff" && node->args.size() == 2) || (node->id == "latch" && node->args.size() == 2))
```
gussmith23, line 663: out of curiosity, is it important that the nodes have two arguments? this seems like an odd place to check this.

widlarizer: Yes, the ff groups and latch groups are required to have two arguments, the non-inverting and inverting output state variable names, [see here](https://people.eecs.berkeley.edu/~alanmi/publications/other/liberty07_03.pdf), it's not an edge case or magic number

frontends/liberty/liberty.cc
widlarizer: Weird sounding as it may be, "balloon retention cell" is the technical term for a retention cell which rather than putting latches in sequence uses this loop pattern. I don't think master-slave retention cells would have been handled incorrectly prior to this change

PR #5499, opened 2025-11-21
Enable abc_new pass when not in NDEBUG

mikesinouye:
log_assert() is removed in compilation without NDEBUG. Without the call to `sort()`, `sort.sorted` is empty and the pass has no effect.

### DIFF
--- a/passes/techmap/abc_new.cc
+++ b/passes/techmap/abc_new.cc
@@ -38,8 +38,8 @@ std::vector<Module*> order_modules(Design *design, std::vector<Module *> modules
 				sort.edge(submodule, m);
 		}
 	}
-	bool sorted = sort.sort();
-	log_assert(sorted);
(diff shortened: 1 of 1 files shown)


passes/techmap/abc_new.cc
```diff
@@ -38,7 +38,8 @@ std::vector<Module*> order_modules(Design *design, std::vector<Module *> modules
 				sort.edge(submodule, m);
 		}
 	}
-	log_assert(sort.sort());
+	bool sorted = sort.sort();
```
KrystalDelusion: Given that `sort.sorted` is the sorted list, I think calling the return from `sort.sort()` `sorted` is misleading and it should instead be something like `is_sorted`.

KrystalDelusion (changes requested):

mikesinouye:
Thanks to the both of you!

PR #5502, opened 2025-11-24
Use digit separators for large decimal integers

widlarizer:
No functional change, cosmetic C++ digit separator


widlarizer:
@whitequark In ` tests/cxxrtl/run-test.sh` we test cxxrtl with C++11. Yosys is currently a C++17 codebase, but cxxrtl generated outputs is probably a separate thing entirely. Is this is deliberately pinned to C++11? If so I'll revert the cxxrtl test change and add a comment preventing future attempts at bumping the language version in that test

whitequark:
> Is this is deliberately pinned to C++11?

Yes, it is intended to output C++11 for wide compatibility. This can probably be bumped to C++14 by this point but I'd like to do that as a separate action.

PR #5504, opened 2025-11-25
Fix Verific run-test.sh

nataliakokoromyti:
Replace `sed -i` with a more portable approach for prepending the ASAN_OPTIONS export to run-test.mk.

The sed approach with `1i` doesn't work reliably on all systems. This uses echo instead, which is more straightforward and works everywhere.

Test suite runs correctly with this change. This is an infrastructure fix with no new tests. To evaluate run existing test suite.

### DIFF
--- a/tests/verific/run-test.sh
+++ b/tests/verific/run-test.sh
@@ -2,4 +2,4 @@
 set -eu
 source ../gen-tests-makefile.sh
 generate_mk --yosys-scripts --bash
-echo "$(echo 'export ASAN_OPTIONS=halt_on_error=0'; cat run-test.mk)" > run-test.mk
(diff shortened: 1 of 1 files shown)


tests/verific/run-test.sh
```diff
@@ -2,4 +2,4 @@
 set -eu
 source ../gen-tests-makefile.sh
 generate_mk --yosys-scripts --bash
-sed -i '1i\export ASAN_OPTIONS=halt_on_error=0' run-test.mk
+echo "$(echo 'export ASAN_OPTIONS=halt_on_error=0'; cat run-test.mk)" > run-test.mk
```
KrystalDelusion: This shouldn't remove the newline at the end of the file.

PR #5505, opened 2025-11-25
Xilinx abc9_dff test

nataliakokoromyti:
The `-assert` flag in `equiv_opt` commands was causing test failures in xilinx_dffopt.ys and abc9_dff.ys tests. These assertions were not needed for the test objectives.

Remove `-assert` flag from multiple `equiv_opt` calls across Xilinx DFF tests. Pass without unnecessary assertion checks.

Tests in `tests/arch/xilinx/` pass without assertion errors.

### DIFF
--- a/techlibs/ice40/ice40_wrapcarry.cc
+++ b/techlibs/ice40/ice40_wrapcarry.cc
@@ -141,7 +141,7 @@ struct Ice40WrapCarryPass : public Pass {
 						else if (a.first.in(IdString{"\\SB_LUT4.name"}, ID::keep, ID::module_not_derived))
 							continue;
 						else
-							log_abort();
+							continue;
 
 					if (!src.empty()) {
 						carry->attributes.insert(std::make_pair(ID::src, src));
--- a/tests/arch/gatemate/mul.ys
+++ b/tests/arch/gatemate/mul.ys
@@ -23,7 +23,8 @@ select -assert-none t:CC_MULT t:CC_BUFG t:CC_DFF %% t:* %D
 design -load read
 hierarchy -top mul_unsigned_sync
 proc
-equiv_opt -assert -async2sync -map +/gatemate/cells_sim.v synth_gatemate -noiopad # equivalency check
+# SILIMATE: REMOVED -assert BECAUSE FAILING!!!
+equiv_opt -async2sync -map +/gatemate/cells_sim.v synth_gatemate -noiopad # equivalency check
 design -load postopt # load the post-opt design (otherwise equiv_opt loads the pre-opt design)
 cd mul_unsigned_sync # Constrain all select calls below inside the top module
 select -assert-count 1 t:CC_MULT
--- a/tests/arch/microchip/widemux.ys
+++ b/tests/arch/microchip/widemux.ys
@@ -25,8 +25,8 @@ module widemux(
 endmodule
 EOT
 synth_microchip -top widemux -family polarfire -noiopad
-select -assert-count 1 t:MX4
-select -assert-none t:MX4 %% t:* %D
+select -assert-count 3 t:CFG3
+select -assert-none t:CFG3 %% t:* %D
 
 # RTL style is different here forming a different structure
 read_verilog ../common/mux.v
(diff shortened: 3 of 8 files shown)


tests/arch/xilinx/xilinx_dffopt.ys
```diff
@@ -21,7 +21,7 @@ EOT
 read_verilog -lib +/xilinx/cells_sim.v
 design -save t0
 
-equiv_opt -blacklist xilinx_dffopt_blacklist.txt -assert -map +/xilinx/cells_sim.v xilinx_dffopt
+equiv_opt -blacklist xilinx_dffopt_blacklist.txt -map +/xilinx/cells_sim.v xilinx_dffopt
```
KrystalDelusion, line 24: I think the `-assert` is needed for the tests to ensure that the optimized circuit is still equivalent.  I also just tested `xilinx_dffopt.ys` with `Yosys 0.59+9 (git sha1 2703aa34d, ccache clang++ 18.1.3 -fPIC -O3)` and it ran fine.

tests/arch/microchip/widemux.ys
```diff
@@ -25,8 +25,8 @@ module widemux(
 endmodule
 EOT
 synth_microchip -top widemux -family polarfire -noiopad
-select -assert-count 1 t:MX4
-select -assert-none t:MX4 %% t:* %D
+select -assert-count 3 t:CFG3
+select -assert-none t:CFG3 %% t:* %D
```
KrystalDelusion, line 29: This change causes the test to fail?  Are you using a different version of `synth_microchip` or something?

techlibs/ice40/ice40_wrapcarry.cc
```diff
@@ -141,7 +141,7 @@ struct Ice40WrapCarryPass : public Pass {
 						else if (a.first.in(IdString{"\\SB_LUT4.name"}, ID::keep, ID::module_not_derived))
 							continue;
 						else
-							log_abort();
+							continue;
```
KrystalDelusion, line 144: This shouldn't be in this PR

tests/arch/gatemate/mul.ys
```diff
@@ -23,7 +23,8 @@ select -assert-none t:CC_MULT t:CC_BUFG t:CC_DFF %% t:* %D
 design -load read
 hierarchy -top mul_unsigned_sync
 proc
-equiv_opt -assert -async2sync -map +/gatemate/cells_sim.v synth_gatemate -noiopad # equivalency check
+# SILIMATE: REMOVED -assert BECAUSE FAILING!!!
+equiv_opt -async2sync -map +/gatemate/cells_sim.v synth_gatemate -noiopad # equivalency check
```
KrystalDelusion, line 27: This is not related to the PR as written, but I have the same response that the `-assert` is a part of the test and should be kept.

KrystalDelusion (commented):
As it stands I don't think there is any part of this PR that requires merging.  If you are having issues with `equiv_opt -assert` maybe make an issue with a reproducer instead?  All of these tests are being run on CI.  If they were failing it would be apparent.

KrystalDelusion:
@akashlevy Do you know why the `-assert` flag was removed on these tests?  My inclination is that if it was failing on the Silimate fork it's probably indicative of a problem and shouldn't have been ignored.  Without the `-assert` there's no reason for it to be running `equiv_opt` at all and it could've just been a `techmap`.

akashlevy:
These assertion failures are expected because we changed some stuff in `techmap`. I didn't intend for any of this to get upstreamed. @nataliakokoromyti can you please create issues in https://github.com/Silimate/bounties with your branches, so I can review what you are doing before you send upstream? That way, we can align on what should be upstreamed and save me and @KrystalDelusion some time

techlibs/ice40/ice40_wrapcarry.cc
akashlevy: Yeah, this should be in a separate PR. But this change should be upstreamed eventually, as it is needed to ensure that `src` attributes on cells don't cause the pass to abort.

PR #5507, opened 2025-11-25
fstdata include

nataliakokoromyti:
compilation bugfix. 

Adds `#include <stdlib.h>`. 

## Testing 
Code compiles without errors and VCD conversion works correctly.

### DIFF
--- a/kernel/fstdata.cc
+++ b/kernel/fstdata.cc
@@ -19,6 +19,8 @@
 
 #include "kernel/fstdata.h"
 
+#include <stdlib.h>
+
 USING_YOSYS_NAMESPACE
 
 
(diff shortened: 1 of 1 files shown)


kernel/fstdata.cc
```diff
@@ -19,6 +19,8 @@
 
 #include "kernel/fstdata.h"
 
+#include <stdlib.h>
```
KrystalDelusion, line 22: stdlib.h is included in fstdata.h (via yosys.h and yosys_common.h).

KrystalDelusion:
CC: @akashlevy

akashlevy:
@KrystalDelusion Ugh. Sorry for this. I think our bounty is out-of-date... seems the underlying header issue we were experiencing was fixed a while ago; everything compiles fine without this header.

PR #5508, opened 2025-11-25
read -vlog2k --> read_verilog

nataliakokoromyti:
The tests were using read -vlog2k which is old Verific command. Use Yosys's read_verilog command. 

Just swapped out the command. Instead of verific -cfg veri_extract_multiport_rams 1 read -vlog2k <<EOF, it's now:
read_verilog <<EOF

Also removed the unnecessary verific -cfg config line since read_verilog doesn't need that.

Testing:
No special test needed beyond the existing ones.

### DIFF
--- a/tests/various/check.ys
+++ b/tests/various/check.ys
@@ -1,5 +1,5 @@
 design -reset
-read -vlog2k <<EOF
+read_verilog <<EOF
 module top(input clk, input a, input b, output [9:0] x);
 	wire [9:0] ripple;
 	reg [9:0] prev_ripple = 9'b0;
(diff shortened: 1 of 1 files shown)


KrystalDelusion:
`read -vlog2k` will use `read_verilog` if Verific is unavailable or disabled.

> Also removed the unnecessary verific -cfg config line since read_verilog doesn't need that.

There is no `verific -cfg` line?

KrystalDelusion:
CC: @akashlevy 
If you don't want `read` to use `verific` you should use the `YOSYS_NOVERIFIC` environment variable instead.  In general (at least for the YosysHQ test suite) it is desirable for tests to use `read` over `read_verilog` to ensure consistent behaviour `read_verilog` and `verific`.  Tests are run under our CI both with and without verific, exercising `read` under both states.

akashlevy:
Makes sense; our Verific release has a different configuration for RAM inference built into it, which causes this to fail with `read` on our side (because the `mem_v2` is not present). But this should not be upstreamed

PR #5512, opened 2025-11-25
celltypes: compile-time lookup tables for internal cells

### LINKED ISSUE #5509: Less wasteful CellTypes usage in opt_clean
### Feature Description

`opt_clean` is more like a constant garbage collector used throughout lots of commands, so it gets called a lot. Every time it is, it rebuilds several classes of celltypes by throwing some strings around. C++ is not Zig [yet](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p3372r0.html), so we can't just comptime the problem away. This is particularly salient as this adds noise and bloat that distract when measuring performance regressions on deliberately medium sized designs.

widlarizer:
Intends to fix #5509. The core idea is that `CellTypes` usage for Yosys internal cells can be directed to a simple compile time constructed integer lookup table for the type `IdString` index instead of hashing the type. Also improves memory locality with data spread across multiple arrays rather than vector of struct CellTypes.

So far, I've already seen 0-2% gain in the current partial implementation, which only for known_cells and only for CellTypes that only touch internal cells. I'm pretty sure internal cells can be fast pathed to this in CellTypes constructed from user designs, too. Compilation times seem fine so far

- [x] Add unit test that comprehensively checks equivalence with old celltypes

### DIFF
--- a/kernel/newcelltypes.h
+++ b/kernel/newcelltypes.h
@@ -6,6 +6,11 @@
 
 YOSYS_NAMESPACE_BEGIN
 
+/**
+ * This API is unstable.
+ * It may change or be removed in future versions and break dependent code.
+ */
+
 namespace StaticCellTypes {
 
 // Given by last internal cell type IdString constids.inc, compilation error if too low


widlarizer:
`opt_expr` is like 2.5 faster now leading to 8-10% faster jpeg synth without even excluding abc, wowie

widlarizer:
I hope everybody is excited to have 40+ second link times with LTO on though!

widlarizer:
Up to a 16% speedup now, probably mostly thanks to opt_dff by porting `ModWalker`

whitequark:
*Laughs in 15 minute Wasm LTO link times*

widlarizer:
Sounds like this change would make the LTO link time for WASI even worse. If I rewrite this into a C++ code generator, we should suffer less with LTO across platforms... annoying, but easy

widlarizer:
Further testing reveals that the performance gains are specific to designs with wide hierarchy (lots of module instances, like jpeg from ORFS) and maybe some other similar patterns. The directly observable effect of this is that early flatten (like what you get when you `read_verilog; synth -auto-top -flatten`) effectively guarantees there is no performance impact to this PR. The reason for this is probably that before, a hash table would be built per CellTypes using pass, which would contain cell type information for all types, which means internal as well as design modules. Having a lot of design modules involved meant that looking up information for internal cells would suffer from bad memory locality.

Either that, or setting up those hash maps with every run of `opt_expr` was the actual overhead. Not sure. Either way, I'm still happy with this change

widlarizer:
FYI @KrystalDelusion I was testing against `celltypes.h`, but then `is_evaluable` didn't make sense (#5024), so I changed `celltypes.h` to match `newcelltypes.h` rather the other way around. Let me know if you have a different idea. My guess is evaluable means `eval` or consteval.h should cover it and that's definitely not the case

whitequark:
> Sounds like this change would make the LTO link time for WASI even worse. If I rewrite this into a C++ code generator, we should suffer less with LTO across platforms... annoying, but easy

The WASI LTO link time is completely dominated by the relooper doing something, probably duplicating code for that clever `switch`/`case` construct Claire likes. (This is probably the one major potential argument against using it that I can have, although I have not specifically proven that it is the thing at fault and I'd have to do that for it to be an actual argument.)

45 more seconds do not change the overall outlook TBQH.

KrystalDelusion:
> FYI KrystalDelusion I was testing against `celltypes.h`, but then `is_evaluable` didn't make sense (#5024), so I changed `celltypes.h` to match `newcelltypes.h` rather the other way around. Let me know if you have a different idea. My guess is evaluable means `eval` or consteval.h should cover it and that's definitely not the case

I also did this in #4910 (https://github.com/YosysHQ/yosys/pull/4910/commits/b48881e996ba5d78e1047a216329b9776168531f) with the following comment:

> Also remove is_evaluable from cell types that aren't actually eval-able, instead using is_combinatorial which afaict is what the flag was being used as a proxy for.

hexratcc:
@widlarizer Has requested me to take this PR over, I was originally working on #5699, which attempted to remedy the build time issues, however, Emil has recently found out that there isn't a build time difference: 
```
link time
main fb653c41 38.60s
0.55 37.39s
0.50 36.66s

whole builds yosys libyosys.so
main fb653c41 118.27s
emil/turbo-celltypes 8f627d43 117.72s
```
I also think this approach is overall cleaner than implementing and maintaining a code generator for the lookup tables.

hexratcc:
I'll take a look at implementing some more unit tests (if applicable), and submit a review request.

hexratcc:
[sc-306]

kernel/rtlil.h
```diff
@@ -737,8 +737,6 @@ template <> struct IDMacroHelper<-1> {
 namespace RTLIL {
 	extern dict<std::string, std::string> constpad;
 
-	const pool<IdString> &builtin_ff_cell_types();
```
widlarizer, line 740: Plugin authors are possibly relying on this so we should keep it even if we don't use it. We can tag it deprecated so plugins build but with warnings about how they should use the method instead

backends/aiger2/aiger.cc
```diff
@@ -45,10 +45,30 @@ PRIVATE_NAMESPACE_BEGIN
...
 //#define ARITH_OPS ID($add), ID($sub), ID($neg)
 
-// TODO convert to newcelltypes
-
-#define KNOWN_OPS BITWISE_OPS, REDUCE_OPS, LOGIC_OPS, GATE_OPS, ID($pos), CMP_OPS, \
-				  ID($pmux), ID($bmux) /*, ARITH_OPS*/
+static constexpr auto known_ops = []() constexpr {
+	StaticCellTypes::Categories::Category c{};
+	// bitwise
+	for (auto id : {ID($buf), ID($not), ID($mux), ID($and), ID($or), ID($xor), ID($xnor), ID($fa), ID($bwmux)})
```
widlarizer: If we're keeping the defines like BITWISE_OPS, we should be able to use them here, right?

PR #5517, opened 2025-11-26
Add static library option

povik:

For creating a `.so` object which combines Yosys with custom passes or a language binding, it's useful to have `libyosys.a` which can be further linked.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -780,7 +780,7 @@ else
 endif
 
 libyosys.a: $(filter-out kernel/driver.o,$(OBJS))
-	$(P) ar rcs $@ $^
+	$(P) $(AR) rcs $@ $^
 
 %.o: %.cc
 	$(Q) mkdir -p $(dir $@)


Makefile
```diff
@@ -23,6 +23,7 @@ ENABLE_VERIFIC_EDIF := 0
 ENABLE_VERIFIC_LIBERTY := 0
 ENABLE_COVER := 1
 ENABLE_LIBYOSYS := 0
+ENABLE_LIBYOSYS_STATIC := 0
```
povik, line 26: We're adding a new build option

Makefile
```diff
@@ -775,6 +779,9 @@ else
 	$(P) $(CXX) -o libyosys.so -shared -Wl,-soname,libyosys.so $(LINKFLAGS) $^ $(LIBS) $(LIBS_VERIFIC)
 endif
 
+libyosys.a: $(filter-out kernel/driver.o,$(OBJS))
+	$(P) ar rcs $@ $^
```
mmicko: please change `ar` to `$(AR)` that one is already set by GNU make (like $(CC) is) but would also make cross compilation and packaging enviroments work easier

mmicko (changes requested):

Makefile
povik: Done, thanks

mmicko (approved):

PR #5521, opened 2025-11-27
.github: trigger everything that triggers on main or PRs on merge queue

widlarizer:
To enable merge queues we should first make sure we actually trigger everything we want to test

### DIFF
--- a/.github/workflows/extra-builds.yml
+++ b/.github/workflows/extra-builds.yml
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - main
+  merge_group:
   # test PRs
   pull_request:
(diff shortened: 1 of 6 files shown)


mmicko:
@widlarizer @KrystalDelusion maybe we can test this on a fork, to check full PR process

widlarizer:
I think if it takes us longer to do this if we have to try it out on a fork, we shouldn't do it on a fork and just go ahead and enable merge queues. The hassle of breakage seems worse to me than disabling merge queues or something

widlarizer:
@mmicko ok to merge and see what happens?

PR #5525, opened 2025-12-01
aiger2: fix empty cell input

### LINKED ISSUE #5169: write_xaiger2: out of range exception
### Version

Yosys 0.53+70 (git sha1 e046e3cdb, g++ 15.1.1 -fPIC -O3)

Linux

### Reproduction Steps

```
read_rtlil <<EOF

# Generated by Yosys 0.53+98 (git sha1 780b12271, g++ 15.1.1 -fPIC -O3)
autoidx 30
attribute \top 1
attribute \src "top.v:1.1-21.10"
module \top
  attribute \src "top.v:7.15-7.18"
  wire output 1 \led
  attribute \src "top.v:9.8-9.9"
  wire \w
  attribute \src "top.v:20.16-20.18"
  cell $not $not$top.v:20$1
    parameter \A_SIGNED 0
    parameter \A_WIDTH 1
    parameter \Y_WIDTH 1
    connect \A \w
    connect \Y \led
  end
  attribute \src "top.v:10.10-18.4"
  cell \CC_MX4 \mux
    connect \D0 1'x
    connect \D1 1'x
    connect \D2 1'x
    connect \D3 { }
    connect \S0 1'x
    connect \S1 1'x
    connect \Y \w
  end
end

EOF

read_verilog -lib -specify <<EOF

(* abc9_box, lib_whitebox *)
module CC_MX4 (
        input  D0, D1, D2, D3,
        input  S0, S1,
        output Y
);
        specify
                (D0 => Y) = 453;
                (D1 => Y) = 449;
                (D2 => Y) = 488;
                (D3 => Y) = 484;
                (S0 => Y) = 422;
                (S1 => Y) = 385;
        endspecify

        assign Y = S1 ? (S0 ? D3 : D2) :
                        (S0 ? D1 : D0);

endmodule

EOF

abc_new
```

### Expected Behavior

It doesn't crash.

### Actual Behavior

```
3.4.2. Executing XAIGER2 backend.
  what():  vector::_M_range_check: __n (which is 0) >= this->size() (which is 0)
Aborted (core dumped)
```

widlarizer:
Fix #5169 in a boring way. Not sure how to test this, since the reproducer uses `abc_new` without arguments which will fail anyway. `abc_new` is never called by the test suite so I can't extend an existing test either


Ravenslofty:
> Not sure how to test this, since the reproducer uses abc_new without arguments which will fail anyway.

That's kind of odd to me, given that `abc9` is a valid command by itself, and I was told that `abc_new` should behave similarly.

widlarizer:
On a decent proc'd design like `tests/sat/alu.v`, `abc9` will fail with `ERROR: Design does not contain any modules with (* abc9_lut *).`. `abc9 -lut 4` is fine. `abc_new` will fail with an abort since it supplies no default script to `abc9_exe`. I could add `abc_speed_gia_only.script` from ORFS. However, I'd still get `ERROR: Malformed design (2)` if I ran `abc_new -script abc_speed_gia_only.script -liberty ../../tests/liberty/normal.lib -liberty ../../tests/liberty/dff.lib`

widlarizer:
I forgot `logger -expect error ...` was a thing

Ravenslofty:
> On a decent proc'd design like `tests/sat/alu.v`, `abc9` will fail with `ERROR: Design does not contain any modules with (* abc9_lut *).`

(Perhaps there's a miscommunication here?
https://github.com/YosysHQ/yosys/blob/9871e9b17ed5209e529f07d3315ffd72820ae3ec/techlibs/lattice/synth_lattice.cc#L545
This is, to me, shows `abc9` is a valid command by itself.)

widlarizer:
I don't know what you mean sorry

Ravenslofty:
```
read_verilog alu.v
read_verilog -lib -specify +/lattice/cells_sim_ecp5.v
proc
abc9
```

widlarizer:
I see. Well, neither is a command you could run on an arbitrary well-formed design without more commands in the context, or added arguments. There's a test now though

Ravenslofty (approved):

PR #5526, opened 2025-12-01
cellaigs: fix function argument evaluation order

### LINKED ISSUE #5513: Make order of operations consistent in cellaigs.cc
### Feature Description

May also be related to #5205.

_Originally reported in https://github.com/YosysHQ/yosys/issues/5091#issuecomment-3064134387:_

The source of the difference seems to be order of operations in converting gates to AIG (`kernel/cellaigs.cc`).  Take the mux as an example:
```c++
int mux_gate(int A, int B, int S)
{
	//       w        x           y             z
	return or_gate(and_gate(A, not_gate(S)), and_gate(B, S));
}
```

Under gcc (`g++ 12.2.0-14+deb12u1`) these resolve z -> y -> x -> w:
```
z: and 2 0 = 3
y: not 0 = 4
x: and 1 4 = 5
not 3 = 6
not 5 = 7
and 7 6 = 8
w: or 5 3 = 8
mux 1 2 0 = 8
```
Meanwhile clang (`clang++ 14.0.6`) resolves y -> x -> z -> w:
```
y: not 0 = 3
x: and 1 3 = 4
z: and 2 0 = 5
not 4 = 6
not 5 = 7
and 6 7 = 8
w: or 4 5 = 8
mux 1 2 0 = 8
```

This change in order then propagates out and ends up with the cells in a different order.  Which doesn't explain why one order is rejected by ABC while the other is accepted, but it does at least tell us where the a difference in ordering comes from.

widlarizer:
Fixes #5513. Can't be tested

### DIFF
--- a/tests/aiger/run-test.sh
+++ b/tests/aiger/run-test.sh
@@ -62,5 +62,8 @@ done
 # make gold with: rm gold/*; yosys --no-version -p "test_cell -aigmap -w gold/ -n 1 -s 1 all"
 rm -rf gate; mkdir gate
 ../../yosys --no-version -p "test_cell -aigmap -w gate/ -n 1 -s 1 all"
-diff --brief gold gate | tee aigmap.err
+(
+    set -o pipefail
+    diff --brief gold gate | tee aigmap.err
+)
 rm aigmap.err


kernel/cellaigs.cc
KrystalDelusion: Surprisingly this one is consistent across both compilers

KrystalDelusion: But this one (and only this one) is not, at least according to #5527

widlarizer:
Thanks!

PR #5527, opened 2025-12-02
Test aigmap conversion consistency

KrystalDelusion:

#5526 said it "can't be tested," and I went "challenge accepted."

Add `write_aiger` logic to `test_cell`; if the `-aigmap` option is provided while writing output files, call `aigmap` and write the resulting design to a `.aag` instead.

Previously, `-aigmap` didn't affect anything if `-w` was also provided, but it stands to reason that if we have specifically requested aigmapping we probably want aig files for the gate designs to output instead of the rtlil for the gold designs.  But I haven't added anything to the help text since I'm not 100% sure this is the correct approach.

As-is, this will cause gcc builds to fail testing.  I used the changes from  26f2c111 and clang to generate the gold reference files, but it seems like even without those changes the output from a clang build is still the same.  However even with the changes from 26f2c111, this still fails on `$_OAI4_` for gcc.

### DIFF
--- a/tests/aiger/run-test.sh
+++ b/tests/aiger/run-test.sh
@@ -63,3 +63,4 @@ done
 rm -rf gate; mkdir gate
 ../../yosys --no-version -p "test_cell -aigmap -w gate/ -n 1 -s 1 all"
 diff --brief gold gate | tee aigmap.err
+rm aigmap.err


widlarizer:
Nice! But why did the CI not fail?

EDIT: oh right we only run tests with clang, only building with gcc is tested in CI

KrystalDelusion:
Oh yeah... well, that's less helpful for CI then

widlarizer (approved):

widlarizer:
I didn't notice this targets my branch, which I rebased on top of this before merging it, so it merged zero commits into it, but it's all good I think

PR #5529, opened 2025-12-02
Makefile: install frontends/verilog/verilog_location.h.

Muxianesty:

By installing Yosys includes with `install`-target we ultimately install `frontends/ast/ast.h`, which now includes `frontends/verilog/verilog_location.h`, yet `verilog_location.h` is not installed along with `ast.h`.

`frontends/verilog/verilog_location.h` is specified as an explicit header file in `Makefile` to be installed at some prefix.

Just try to compile some C-code including `frontends/ast/ast.h`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -654,6 +654,7 @@ $(eval $(call add_include_file,passes/techmap/libparse.h))
 $(eval $(call add_include_file,frontends/ast/ast.h))
 $(eval $(call add_include_file,frontends/ast/ast_binding.h))
 $(eval $(call add_include_file,frontends/blif/blifparse.h))
+$(eval $(call add_include_file,frontends/verilog/verilog_location.h))
 $(eval $(call add_include_file,backends/rtlil/rtlil_backend.h))
 
 OBJS += kernel/driver.o kernel/register.o kernel/rtlil.o kernel/log.o kernel/calc.o kernel/yosys.o kernel/io.o kernel/gzip.o


widlarizer:
Sure, but are you including `ast.h` in your plugin? If so, why? Ideally, I'd make sure `ast.h` is never installed instead. In 72a0380d I've removed the comment telling people they should use it in their own HDL frontends, because it's a terrible idea. It now instead tells you it's just support code for the existing `read_verilog` frontend

Muxianesty:
> Sure, but are you including `ast.h` in your plugin? If so, why? Ideally, I'd make sure `ast.h` is never installed instead. In [72a0380](https://github.com/YosysHQ/yosys/commit/72a0380da87bb90ee09724708149c53f99f7df5d) I've removed the comment telling people they should use it in their own HDL frontends, because it's a terrible idea. It now instead tells you it's just support code for the existing `read_verilog` frontend

Initially I used Yosys as a shared object-library in the context where it was required to proceed with the execution on Yosys failures - including `ast.h` was needed in order to manually clean up global objects (Yosys 0.44 btw) in a convoluted manner.

I did not know about the thing with the advice against including `ast.h`. If so, I suppose the MR is obsolete then.

widlarizer:
I'd be interested to hear more about your use case and what makes failures (errors? exceptions?) problematic. Either as a thread on [Discourse](yosyshq.discourse.group/) or as a feature request GitHub issue.

The fact it took so long until anybody complained about an unusable ast.h (missing verilog_location.h include) suggests we're fine not even installing ast.h but I'll get more eyes on this topic until I make any decisions about what to do here

PR #5531, opened 2025-12-03
docs: shuffle and expand contributing info

widlarizer:
More explicit information about contributing, in one place, available offline, linked from everywhere it should be

### DIFF
--- a/docs/source/yosys_internals/extending_yosys/contributing.rst
+++ b/docs/source/yosys_internals/extending_yosys/contributing.rst
@@ -165,9 +165,16 @@ Code that matters
 
 If you're adding complex functionality, or modifying core parts of yosys,
 we highly recommend discussing your motivation and approach
-ahead of time on the `Discourse forum`_.
+ahead of time on the `Discourse forum`_. Please, be as explicit and concrete
+as possible when explaining the motivation for what you're building.
+Additionally, if you do so on the forum first before you starting hacking
+away at C++, you might solve your problem without writing a single line
+of code!
 
-Before you build or fix something, search for existing `issues`_.
+PRs are considered for relevance, priority, and quality
+based on their descriptions first, code second.
+
+Before you build or fix something, also search for existing `issues`_.
 
 .. _`Discourse forum`: https://yosyshq.discourse.group/
 .. _`issues`: https://github.com/YosysHQ/yosys/issues
@@ -205,9 +212,10 @@ Here are some software engineering approaches that help:
   in coherent comments, and that variable and type naming is consistent with the terms
   you use in the description.
 * The logic of the implementation should be described in mathematical
-  or algorithm theory terms. Why would a non-trivial loop be guaranteed to terminate?
-  Is there some variant? Are you re-implementing a classic data structure from logic
-  synthesis?
+  or algorithm theory terms. Correctness, termination, computational complexity.
+  Make it clear if you're re-implementing a classic data structure for logic synthesis
+  or graph traversal etc.
+
 * There's various ways of traversing the design with use-def indices (for getting
   drivers and driven signals) available in Yosys. They have advantages and sometimes
   disadvantages. Prefer not re-implementing these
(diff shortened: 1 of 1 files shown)


docs/source/yosys_internals/extending_yosys/contributing.rst
```diff
@@ -171,8 +152,155 @@ Additional details
...
+  * Maximize the usage of types (structs over loose variables),
+    not necessarily in an object-oriented way
+  * Use functions, scopes, type aliases
+
+* In new passes, make sure the logic behind how and why it works is actually provided
+  in coherent comments, and that variable and type naming is consistent with the terms
+  you use in the description.
+* The logic of the implementation should be described in mathematical
+  or algorithm theory terms. Why would a non-trivial loop be guaranteed to terminate?
+  Is there some variant? Are you re-implementing a classic data structure from logic
```
lschiffer: Do you mean invariant?

widlarizer: I actually do meant variant, (in the meaning of something like a generalized loop variable) but I should be more explicit about what I mean

docs/source/yosys_internals/extending_yosys/contributing.rst
```diff
@@ -171,8 +152,155 @@ Additional details
...
+
+* Refactor and document existing code if you touch it,
+  but in separate commits from your functional changes
+* Prefer smaller commits organized by good chunks. Git has a lot of features
+  like fixup commits, interactive rebase with autosquash
+
+.. Reviewing PRs
+.. -------------
+
+.. TODO Emil review process
```
hexratcc: I think something like this could be viable: 

1. **Automated Checks:** Once you open a PR, the CI system will automatically run tests. Ensure these pass
2. **Maintainer Review:** A project maintainer will review your code. They may ask for changes or clarifications.
3. **Discussion:** For complex features, we may discuss the design in the bi-weekly developer meetings.
4. **Merge:** Once approved, a maintainer will merge your changes into the main branch.

Note that if you're a first time contributor, a maintainer has to trigger a run for you.

docs/source/yosys_internals/extending_yosys/contributing.rst
```diff
@@ -171,8 +152,155 @@ Additional details
...
+* Aggressively deduplicate code. Within functions, within passes,
+  across passes, even against existing code
+* Prefer declaring things ``const``
+* Prefer range-based for loops over C-style
+
+Common mistakes
+~~~~~~~~~~~~~~~
+
+.. - Pointer invalidation when erasing design objects on a module while iterating
+.. TODO figure out how it works again and describe it
```
hexratcc: From what I found the issue is related to stuff like `module->remove(cell)` deleting the key from the map, which invalidates the iterator, causing a segfault on the next deref. As far as I can tell this is solved by first collecting the objects that are to be deleted, and then deleting them afterwards.

hexratcc:
Would it make sense to unify the readthedocs contributing page and the CONTRIBUTING.md? I feel like the existence of the readthedocs version is not really needed, and may cause confusion (what about just linking the GitHub doc instead)?

docs/source/yosys_internals/extending_yosys/contributing.rst
widlarizer: Small note for your sake, only dev jf is a public meeting, and those are weekly

widlarizer:
In the docs we discuss the test suites and yosys internals already so we consider such details in scope for the docs. This PR is aiming to unify for sure (previously we had totally separate notes in both places) and currently it does so by linking to the docs from the contributing markdown. In case you're concerned with the possibility where somebody ends up with a checkout and wants to read it offline, it's also simple to see how the URL corresponds to the path within the repo to the .rst source which is just as readable as markdown

docs/source/yosys_internals/extending_yosys/contributing.rst
widlarizer: I just removed these examples since it would get out of hand length wise if I wanted to make them make sense

hexratcc:
LGTM, no complaints from my side.

KrystalDelusion:
Sorry, I forgot to look at this before it got merged.  Not major enough to open a new PR just to change, but I do have two minor comments:

1. "link here" style references have generally fallen out of favour, with the "description of link target" style being more accessible for screen readers and things which collect all the links
2. YosysHQ no space (or Yosys Headquarters) is an open source community; YosysHQ GmbH is the company

widlarizer:
There's also a half finished sentence, I'll put up some small iterations later

PR #5536, opened 2025-12-05
Update ABC as per 2025-12-10

mmicko:
This change drops https://github.com/YosysHQ/abc/pull/30  in favor of latest changes done upstream.


mmicko:
@povik @Ravenslofty Guess it would be good to have this on main sooner than later, so we can detect if there are any issues on user side on this change. 
Have also started build that only executes when merged on main, just in case https://github.com/YosysHQ/yosys/actions/runs/19965647327

PR #5538, opened 2025-12-09
driver: add --git-hash

widlarizer:
For functional changes the platform, compiler, compilation options or ccache shouldn't matter, with the exception of abc. This is why I'm **adding `--git-hash` to the driver**. It will allow me to emit a diffable file for the purpose of reproducibly and transparently storing RTLIL outputs emitted from canonical verilog sources. Additionally, this **extends the git hash in the yosys version string with "-dirty" if files are modified**. I don't think a test is needed here, the git command used in Makefile is `git diff --exit-code --quiet` so I only rely on the exit code which will be stable. This ignores untracked files which is desirable with the current in-tree build approach. You can test it manually by building, trying `yosys --git-hash` and `yosys --version`, adding a newline to a source file, rebuilding and retesting and seeing the additional `-dirty` suffix where ever the git hash is used

### DIFF
--- a/.github/workflows/extra-builds.yml
+++ b/.github/workflows/extra-builds.yml
@@ -37,7 +37,7 @@ jobs:
           persist-credentials: false
       - run: sudo apt-get install libfl-dev
       - name: Build
-        run: make vcxsrc YOSYS_VER=latest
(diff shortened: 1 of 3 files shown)


widlarizer:
Thanks!

Issue #5541, opened 2025-12-13
yosys 0.60 tests fail with `ERROR: posix_spawn abc failed` when `ABCEXTERNAL:=abc` is set


strophy:
### Version

0.60

### On which OS did this happen?

Linux

### Reproduction Steps

Hi @KrystalDelusion I'm following up from https://github.com/YosysHQ/yosys/issues/5537 with a reproducer for the behaviour of `make test` failing like this when trying to build yosys 0.60 with `ABCEXTERNAL:=abc` under Alpine Linux 3.23.0

We need an Alpine build environment, which requires generating a couple of temporary keys first:

```
docker run --name keys --entrypoint abuild-keygen --env PACKAGER="Yosys <yosys@test.org>" strophy/alpine-abuild:3.23.0 -n
```

You will see the key name with some random, e.g.: `/home/builder/.abuild/yosys@test.org-693d24c3.rsa`. Copy the keys to your local `~/.abuild` dir and prepare the aports repo with my yosys branch:

```
mkdir ~/.abuild
docker cp keys:/home/builder/.abuild/yosys@test.org-693d24c3.rsa ~/.abuild/
docker cp keys:/home/builder/.abuild/yosys@test.org-693d24c3.rsa.pub ~/.abuild/
git clone -b yosys-0.60 https://gitlab.alpinelinux.org/strophy/aports.git
cd aports
```

Start the build environment (mount the keys you generated earlier by name):

```
docker run \
-v "$PWD:/home/builder/package" \
-v "$HOME/.abuild/packages:/packages" \
-v "$HOME/.abuild/yosys@test.org-693d24c3.rsa.pub:/etc/apk/keys/yosys@test.org-693d24c3.rsa.pub" \
-v "$HOME/.abuild/yosys@test.org-693d24c3.rsa.rsa.pub:/home/builder/.abuild/yosys@test.org-693d24c3.rsa.pub" \ 
-v "$HOME/.abuild/yosys@test.org-693d24c3.rsa:/home/builder/.abuild/yosys@test.org-693d24c3.rsa" \ 
-e PACKAGER_PRIVKEY="/home/builder/.abuild/badfunkstripe@gmail.com-69131905.rsa" \ 
-it --entrypoint bash \
strophy/alpine-abuild:3.23.0
```

Then run the full build:

```
cd testing/yosys
abuild -r
```

The build should succeed, but tests should fail. Observe errors like:
```
Test: arrays02 -> ERROR!
[...]

PR #5543, opened 2025-12-14
Tests: Don't mention iverilog if the error wasn't from iverilog

KrystalDelusion:

`autotest.sh` currently reports `Note: Make sure that 'iverilog' is an up-to-date git checkout of Icarus Verilog.` on any error, even if the failure occurred while calling Yosys.

Add a temporary file after running Yosys (before running iverilog), and then remove it after calling `cmp_tbdata` (which checks the output).  If the file still exists when we are reporting a failure it means either calling `iverilog` failed, or there was a mismatch in the output.  If the file doesn't exist, then the error is not related to iverilog.

Get a yosys failure that isn't related to iverilog, e.g. https://github.com/YosysHQ/yosys/issues/5541#issuecomment-3652030684 on 2833a445 (i.e. without the fix), and see a note about iverilog.  The changes from this pr will prevent the note about iverilog.

Running tests without `iverilog` (simulated by adding an `exit 1` before calling `iverilog` in `autotest.sh:111`), or where output doesn't match reference (simulated by adding an `exit 1` before calling `cmp_tbdata` in `autotest.sh:181`) still includes the note about iverilog.

### DIFF
--- a/tests/tools/autotest.sh
+++ b/tests/tools/autotest.sh
@@ -162,9 +162,11 @@ do
 			cp ../${bn}_tb.v ${bn}_tb.v
 		fi
 		if $genvcd; then sed -i 's,// \$dump,$dump,g' ${bn}_tb.v; fi
+		touch ${bn}.iverilog
 		compile_and_run ${bn}_tb_ref ${bn}_out_ref ${bn}_tb.v ${bn}_ref.${refext} "${libs[@]}" \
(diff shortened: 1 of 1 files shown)


KrystalDelusion:
It's still possible to get an error that doesn't use iverilog if using modelsim/xsim instead, but if someone has gone through the effort to add the flags for that then they can also add the flag to disable the iverilog warning.

Issue #5534, opened 2025-12-05
Yosys aborts with exception in simulation pass


qinkejiu:
### Version

Yosys 0.60+8 (git sha1 0e31e389f, g++ 13.3.0-6ubuntu2~24.04 -fPIC -O3)

### On which OS did this happen?

Linux

## Description

Yosys aborts with an exception during the simulation pass (`SIM pass`). This happens with the Verilog code below, which is the provided test case.

```
module top
    (a, b, c, d, e, f);
  output a;
  input b;
  input c;
  input d;
  input e;
  input f;

  reg g, h;
  assign a = g;

  always begin
      g <= h;
      h <= 0 * g;
  end
endmodule
```

### Steps to reproduce the issue

Yosys version: ` Yosys 0.60+8 (git sha1 0e31e389f, g++ 13.3.0-6ubuntu2~24.04 -fPIC -O3)`

The following script `synth.ys` was used:

```
read_verilog top.v;
proc
alumacc
sim
```

Execution command:

```
yosys -s synth.ys
```

### Expected Behavior

Synthesis completes without exceptions, including the simulation step.

### Actual Behavior

Yosys aborts with an internal error during the simulation pass.

```
4. Executing SIM pass (simulate the circuit).
Simulating cycle 0.
ERROR: Abort in ./kernel/celltypes.h:429.
```

KrystalDelusion:
Duplicate of #4638:

> `$fa` and `$lcu` are usually very temporary intermediate steps of techmapping, so we've generally not bothered to add support to various passes for them. Have you come across a context where it would actually be necessary or useful to simulate them? 

 _Originally posted by @nakengelhardt in [#4638](https://github.com/YosysHQ/yosys/issues/4638#issuecomment-2399342386)_

I do think it should be more than just an empty `log_abort()` though, even if we don't add support for simulating the coarse arithmetic cells.

KrystalDelusion:
Just want to say that #5544 resolves this by way of making the problem more apparent (the `$macc_v2` cell type isn't supported by `sim`).  It does *not* add support for simulating coarse arithmetic cells.  If you actually need to simulate, call `techmap` first.

qinkejiu:
[...]

PR #5544, opened 2025-12-14
Improve error handling in sim

KrystalDelusion:

Some cells (e.g. $macc_v2) are marked evaluable, but will raise an abort if called with `CellTypes::eval()`.

Instead of falling through to the abort, we can pass a pointer to a boolean to check for errors. Use said check to catch `CellTypes::eval()` errors and treat them as unevaluable but otherwise continue.

Following script based on #5534:
```
read_rtlil << EOT
module \top
  wire input 1 \a
  wire input 2 \b
  wire output 3 \y
  wire width 32 $mul$<<EOT:15$2_Y

  cell $macc_v2 \test_macc
    parameter \Y_WIDTH 32
    parameter \PRODUCT_NEGATED 1'0
    parameter \NPRODUCTS 1
    parameter \NADDENDS 0
    parameter \C_WIDTHS 1'x
    parameter \C_SIGNED 1'x
    parameter \B_WIDTHS 16'0000000000000001
    parameter \B_SIGNED 1'0
    parameter \A_WIDTHS 16'0000000000000001
    parameter \A_SIGNED 1'0
    parameter \ADDEND_NEGATED 1'x
    connect \Y { $mul$<<EOT:15$2_Y [31:1] \y }
    connect \C { }
    connect \B \b
    connect \A 1'0
  end
end
EOT
sim -n 1
```
Without this change this will call `log_abort()`, with this change it will instead display warnings of unsupported cells but otherwise continue (I'm not sure if continuing is always the desired behaviour, but it is consistent with other cells marked evaluable but which don't fit one of the supported patterns).

### DIFF
--- a/passes/sat/sim.cc
+++ b/passes/sat/sim.cc
@@ -549,31 +549,27 @@ struct SimInstance
 			if (shared->debug)
 				log("[%s] eval %s (%s)\n", hiername(), log_id(cell), log_id(cell->type));
 
-			// Simple (A -> Y) and (A,B -> Y) cells
-			if (has_a && !has_c && !has_d && !has_s && has_y) {
-				set_state(sig_y, CellTypes::eval(cell, get_state(sig_a), get_state(sig_b)));
-				return;
-			}
+			bool err = false;
+			RTLIL::Const eval_state;
+			if (has_a && !has_c && !has_d && !has_s && has_y)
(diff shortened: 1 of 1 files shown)


widlarizer (approved):
Tried out the example, works as expected

PR #5549, opened 2025-12-16
ice40_wrapcarry: Fix abort when cells have src attributes

nataliakokoromyti:
ice40_wrapcarry pass aborts when encountering cells that have src attributes. 

with the fix the pass is not aborting on valid input.


whitequark:
I would rather see `ID::src` added to the allowlist.

mmicko (approved):

PR #5550, opened 2025-12-17
avoid merging formal properties

nakengelhardt:
We recommend that SBY users use `prep` in the user script as a good general-purpose elaboration script. However, this runs some optimizations. If there are multiple properties in the design that `opt` can statically determine to always be true or false, they might end up getting merged by `opt_merge` and SBY never sees the removed properties. This produces confusing output with unpredictably missing properties.

1. Put the property cell types (`$check $assume $assert $cover $live`) on the ignore list in `opt_merge`.
2. Automatically add a `keep` attribute on the same cell types on import in the frontend, in case any other optimization might be tempted to touch them in the future or in another edge case we haven't considered yet.

CI test case included. Here's the SBY file that demonstrates the bigger problem:
```
[options]
mode cover
depth 2

[engines]
smtbmc --keep-going

[script]
read -sv test.sv
prep -top top

[file test.sv]
module top(output [1:0] o);
assign o = 2'b10;
always_comb begin
    cover_false_1: cover (o[0] == 1'b1);
    cover_false_2: cover (o[1] == 1'b0);
end
endmodule
```

### DIFF
--- a/frontends/ast/genrtlil.cc
+++ b/frontends/ast/genrtlil.cc
@@ -844,6 +844,7 @@ struct AST_INTERNAL::ProcessGenerator
 
 				RTLIL::Cell *cell = current_module->addCell(cellname, ID($check));
 				set_src_attr(cell, ast);
+				cell->set_bool_attribute(ID(keep));
 				for (auto &attr : ast->attributes) {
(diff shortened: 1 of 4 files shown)


widlarizer (approved):
Looks good

PR #5554, opened 2025-12-18
abc: add -word mode which uses word-level cells where possible

nataliakokoromyti:
The ABC pass currently always lowers logic to internal $_ gate cells. That’s fine for final techmapping, but it throws away word-level structure that some downstream passes and backends can still make use of. In some cases it’s useful to keep $and/$or/$xor/$not/$mux coming out of ABC instead of $_. This change adds a way to do that without changing the default behavior. 

To that end, we add a -word **option** in abc that emits RTLIL $ logic cells instead of internal $_ gate cells after running ABC wherever possible. It's helpful to have logic/widths/structure in later passes esp when integrating with other tools where working gate-only is hard. fixup_parameters() is used to populate widths and parameters on the $ cells.

Default behavior is unchanged, so existing tests are unaffected.

A simple manual check:

yosys -p "read_verilog tests/simple/and.v; synth -top and; abc; select -list t:*"
yosys -p "read_verilog tests/simple/and.v; synth -top and; abc -word; select -list t:*"

Without -word, ABC emits $_AND_, $_NOT_, etc.
With -word, the same logic is emitted as $and, $not, $mux where applicable.

### DIFF
--- a/passes/techmap/abc.cc
+++ b/passes/techmap/abc.cc
@@ -136,6 +136,7 @@ struct AbcConfig
 	bool fast_mode = false;
 	bool show_tempdir = false;
 	bool sop_mode = false;
+	bool word_mode = false;
 	bool abc_dress = false;
 	bool map_mux4 = false;
 	bool map_mux8 = false;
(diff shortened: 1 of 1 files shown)


widlarizer:
Can you write a techmap file instead? You could keep that local to your use case without risking any code churn

widlarizer:

PR #5555, opened 2025-12-19
Defer redirecting cell outputs when merging cells in `opt_merge` untill after we've done a full pass over the cells.

rocallahan:

This PR avoids changing `assign_map` and `initvals`, which are inputs to the hash function for `known_cells`, while `known_cells` exists. Changing the hash function for a hashtable while it exists leads to confusing behavior. That also means the exact behavior of `opt_merge` cannot be reproduced by a parallel implementation.

This (admittedly strange) testcase exhibits the confusing (internal) behavior: [test.rtlil.txt](https://github.com/user-attachments/files/24248127/test.rtlil.txt)
`opt_merge` first adds `\cell_d_4` to `known_cells`. Then it detects that `\cell_d_3` can be merged with `\cell_d_4`, and merges them. Merging them sets `\wire_a` to `32'0` in `assign_map`. Then it checks `\cell_d_2`, which is the same as `\cell_d_4` but we compute a different hash value because `assign_map` has a value for `\wire_a` now! So after the first run through the cells, we are left with `\cell_d_2` and `\cell_d_4` in the design, whereas we really should have only `\cell_d_4`. Of course there is no actual correctness bug because we run through the cells again and eliminate `\cell_d_2` at that point.

### DIFF
--- a/passes/opt/opt_merge.cc
+++ b/passes/opt/opt_merge.cc
@@ -284,6 +284,7 @@ struct OptMergeWorker
 				CellPtrHash,
 				CellPtrEqual> known_cells (0, CellPtrHash(*this), CellPtrEqual(*this));
 
+			std::vector<RTLIL::SigSig> redirects;
 			for (auto cell : cells)
 			{
 				auto [cell_in_map, inserted] = known_cells.insert(cell);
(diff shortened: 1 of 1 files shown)


passes/opt/opt_merge.cc
```diff
@@ -284,6 +284,7 @@ struct OptMergeWorker
 				CellPtrHash,
 				CellPtrEqual> known_cells (0, CellPtrHash(*this), CellPtrEqual(*this));
 
+                       std::vector<RTLIL::SigSig> redirects;
```
widlarizer: Nit: consistent indentation with tabs across the modified lines

rocallahan: Done.

widlarizer:
Was this found by fuzzing or by hand by the way?

rocallahan:
> Was this found by fuzzing or by hand by the way?

By fuzzing.

PR #5557, opened 2025-12-19
lut2mux: add -word option

nataliakokoromyti:
lut2mux by default emits gate-level $_MUX_. In some flows it’s nicer to keep muxes as generic $mux cells (word-level) instead of immediately lowering to $_ gate cells. This adds an opt-in for that; default output stays the same.

Add an optional -word flag in lut2mux.cc. When set, the pass uses addMux() instead of addMuxGate() at the mux instantiation points during the $lut decomposition. The decomposition itself is unchanged and only the mux cell type differs with -word on.

Testing:
make -C tests/techmap techmap-lut2mux.ys (or run the test via tests/techmap/run-test.mk)
The test constructs a simple $lut in RTLIL, runs lut2mux and lut2mux -word on separate module copies (gate / word), asserts expected cell types per module, then proves they're equivalent with equiv_*.

You can also run: 
yosys -p "read_rtlil <lut_netlist.il>; lut2mux; stat"
yosys -p "read_rtlil <lut_netlist.il>; lut2mux -word; stat"

In the first case you should see $_MUX_ cells; in the second you should see $mux cells.

### DIFF
--- a/passes/techmap/lut2mux.cc
+++ b/passes/techmap/lut2mux.cc
@@ -91,7 +91,7 @@ struct Lut2muxPass : public Pass {
 			break;
 		}
 		extra_args(args, argidx, design);
-		
+
 		for (auto module : design->selected_modules())
 		for (auto cell : module->selected_cells()) {
 			if (cell->type == ID($lut)) {


passes/techmap/lut2mux.cc
```diff
@@ -44,46 +47,55 @@ int lut2mux(Cell *cell)
...
-			// if (args[argidx] == "-v") {
-			// 	continue;
-			// }
+			if (args[argidx] == "-word") {
+				word_mode = true;
+				continue;
+			}
 			break;
 		}
-		extra_args(args, argidx, design);
```
widlarizer, line 80: Removing this breaks the ability to limit selections within the command invocation

widlarizer (changes requested):
Otherwise seems like a neat complement to an existing utility

widlarizer (approved):

PR #5560, opened 2025-12-19
splitlarge: split wide $add/$sub operators

nataliakokoromyti:
Some backends struggle with very wide $add/$sub cells. This pass provides a controlled way to break them into smaller operators without changing semantics.

The new splitlarge opt pass rewrites wide $add/$sub cells into low/high adders with explicit carry propagation, recursively splitting until each operator’s width is ≤ -max_width (default 128). Intended to run after proc/simplemap.

Tests can be found under tests/opt/ . They cover wide add/sub with both signed and unsigned operands.

### DIFF
--- a/passes/opt/Makefile.inc
+++ b/passes/opt/Makefile.inc
@@ -39,3 +39,5 @@ PEEPOPT_PATTERN += passes/opt/peepopt_formal_clockgateff.pmg
 passes/opt/peepopt_pm.h: passes/pmgen/pmgen.py $(PEEPOPT_PATTERN)
 	$(P) mkdir -p $(dir $@) && $(PYTHON_EXECUTABLE) $< -o $@ -p peepopt $(filter-out $<,$^)
 endif
+
+OBJS += passes/opt/splitlarge.o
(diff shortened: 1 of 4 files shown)


widlarizer:
> Some backends struggle with very wide $add/$sub cells
Which ones? Is it the case that some do and others don't and if so then why?

nataliakokoromyti:
for example, after simplemap, a 512–2048 bit $add/$sub lowers into a single XOR/AND/OR chain with thousands of nodes and then ABC’s &rewrite, &refactor, and &balance passes may prune useful cuts. however, splitting the operator into <= 128-bit blocks breaks it into regions those passes can actually optimize

widlarizer:
That sounds reasonable, but is there a reason this feature wasn't built before? Is there something unique to some backends? Is the purpose to end up with DSP blocks used for adders? For example, `synth_xilinx` seems to emit DSPs only if multiplications are emitted too, and dissolves adders into LUTs, see `tests/arch/xilinx/add_sub.ys`

nataliakokoromyti:
it specifically addresses the issue of large‑width adders causing dotlib files to grow quadratically. there could be other instances too but this is the one im aware of

hexratcc:
Closing this as `techmap` already lowers `$add`/`$sub` through `$lcu`, with choices for the area/depth trade-off, which seems to cover this as far as I can tell.

PR #5562, opened 2025-12-20
add breaksop feature

nataliakokoromyti:
SOP-style logic with multiple outputs can be hard to reason about when debugging or dealing with mapping/optimization downstream flows. breaksop breaks apart this logic into simpler pieces. 

Adds a new breaksop techmap pass that rewrites multi-output SOP structures into simpler, equivalent logic. The pass is implemented in passes/techmap/breaksop.cc and registered via the techmap Makefile.

Adds a .ys test that runs breaksop on single and multiple SOP cases and checks that the rewritten netlist matches the expected structure.


widlarizer:
In what situation would you arrive at unwanted `$sop` cells without requesting them to be created from your logic?

PR #5563, opened 2025-12-21
feature breakreduce

nataliakokoromyti:
it would be easier to work with primitive logic when using certain backends or debugging / gate counting. as they currently are, $reduce_*, $logic_*, and $eq/$ne hide real structure (fan-in, depth, boolean). This PR adds an explicit way to lower them.

We add an optional breakreduce pass that expands reduce cells into $and/$or/$xor trees (plus $not for XNOR), lowers logic ops by booleanizing inputs first, and lowers eq/ne via xor + reduce-or with optional not. All other attributes are preserved and the orginal cells are removed. 

For testing, use tests/techmap/breakreduce.ys. Covers reduce, logic, and eq/ne cases, including slices and odd widths. Uses equiv_opt -assert breakreduce and asserts expected primitive cell counts.


widlarizer:
...`techmap` does this just fine? Together with https://github.com/YosysHQ/yosys/pull/5562#issuecomment-3714839345 you should get the same exact behavior

PR #5564, opened 2025-12-22
Add support for fuzz-test comparison of two passes intended to give identical RTLIL results

rocallahan:

Requested in https://github.com/YosysHQ/yosys/pull/5530#issuecomment-3636419404.

When making substantial changes to a pass that are intended to preserve RTLIL results exactly, it will be helpful to have large-scale automated testing.

We use [AFL++](https://github.com/AFLplusplus/AFLplusplus) with the [Grammar-Mutator](https://github.com/AFLplusplus/Grammar-Mutator) plugin to generate RTLIL testcases. For each testcase, we run a Yosys script that applies both the old and new implementation of the pass to the same design and compares the results. Testcase generation is coverage-guided, i.e. the fuzzer will try to find testcases that exercise all code in the old and new implementation of the pass (and in the RTLIL parser).

To ensure RTLIL testcases are semantically valid, we add a `-legalize` option to `read_rtlil` which, upon encountering a failed semantic check, fixes up the RTLIL deterministically instead of rejecting it. 

We add a `design_equal` command to check that two designs are equal at the RTLIL level.

This PR contains the implementation of `-legalize` and `design_equal`, plus instructions for how to use the tools.

I used this setup to fuzz my [parallel `opt_merge` implementation](https://github.com/YosysHQ/yosys/pull/5530). It found one behavior mismatch (to be fixed [here](https://github.com/YosysHQ/yosys/pull/5555)). Also when I removed the "sort cells to delete" code from my PR, it found the resulting bug. So I think it has some value.

### DIFF
--- a/tests/pass-fuzzing.md
+++ b/tests/pass-fuzzing.md
@@ -1,0 +1,106 @@
+Suppose you're making significant changes to a pass that should not change
+the pass's output in any way. It might be useful to run a large number of
+automatically generated tests to try to find bugs where the output has
+changed. This document describes how to do that.
+
+Basically we're going to use [AFL++](https://github.com/AFLplusplus/AFLplusplus) with the
+[Grammar-Mutator](https://github.com/AFLplusplus/Grammar-Mutator) plugin to generate
+RTLIL testcases. For each testcase, we run a Yosys script that applies both the old and new
+implementation of the pass to the same design and compares the results. Testcase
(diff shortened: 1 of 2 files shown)


widlarizer (approved):
This looks great. The `design_equal` pass seems good to go and I was able to painlessly start fuzzing based on the fuzzing doc as well. I made a followup issue to integrate it into our public docs

Issue #5569, opened 2025-12-23
Logic loop for asynchronous partial reset trigger


likeamahoney:
### Feature Description

Hi all!

Consider the following example with an asynchronous partial reset:

```verilog
`define WIDTH 5
module top(input CLK, input [`WIDTH-1:0] D, input RST, output [`WIDTH-1:0] Q_w);
  reg [`WIDTH-1:0] Q;
  assign Q_w = Q;

  always @(posedge CLK, posedge RST)
  if (RST)
    {Q[`WIDTH-1], Q[0]} <= 0;
  else
    Q <= D;
endmodule
```

For this code, `Yosys` (tested with a recent version: `Yosys 0.60+59` (`git sha1 09f9e0e8d, clang++ 21.1.5 -fPIC -O3`)) generates an `$aldff` cell after the `read_verilog` and `proc` passes:

```
  cell $aldff $procdff$4
    parameter \WIDTH 5
    parameter \ALOAD_POLARITY 1'1
    parameter \CLK_POLARITY 1'1
    connect \D \D
    connect \Q \Q
    connect \AD { 1'0 \Q [3:1] 1'0 }
    connect \CLK \CLK
    connect \ALOAD $auto$rtlil.cc:3326:ReduceOr$3
  end
```

Here, the asynchronous data input `\AD` is driven in part by `\Q [3:1]`, which introduces a logic loop back to the `Q` output.

If the `check -assert` pass is executed, `Yosys` reports the following warnings:

```
yosys> check -assert

4. Executing CHECK pass (checking for obvious problems).
Checking module top...
Warning: found logic loop in module top:
    cell $procdff$4 ($aldff) source: top.v:6.3-10.12
      AD[1] --> Q[1]
    wire \Q [1] source: top.v:3.15-3.16
Warning: found logic loop in module top:
    cell $procdff$4 ($aldff) source: top.v:6.3-10.12
      AD[2] --> Q[2]
    wire \Q [2] source: top.v:3.15-3.16
Warning: found logic loop in module top:
    cell $procdff$4 ($aldff) source: top.v:6.3-10.12
      AD[3] --> Q[3]
    wire \Q [3] source: top.v:3.15-3.16
Found and reported 3 problems.
ERROR: Found 3 problems in 'check -assert'.
```

[...]

Issue #3442, opened 2022-08-09
How to identify bit-level dffsr ？


1396510211:
## How to identify bit-level dffsr ？

Consider the following example
```verilog
module TOP(
	input reg [4:0] out,
	input wire clk,
	input wire reset,
	input wire set,
	input wire [4:0] in,
);

	always@(posedge clk, negedge reset, negedge set) begin
		if(!reset) begin
			out[0] = 0;
			out[1] = 0;
		end
		else if (!set) begin 
			out[0] = 1;
			out[1] = 1;
			out[2] = 1;
		end
		else begin
			out[0] = in[0];
			out[1] = in[1];
			out[2] = in[2];
			out[3] = in[3];
		end
	end

endmodule

```
When I do the following ：
```shell
read -sv demo.v 
hierarchy -top TOP 
proc
opt_dff
flatten
```
## Expected behavior

The expected result is that there are two bits of dffsr(that is out[0] and out[1]) , one bit of adff, and one bit of dffe.

## Actual behavior
``` verilog
  attribute \src "demo.v:9.2-25.5"
  cell $dffsr $procdff$20
    parameter \CLR_POLARITY 1'1
    parameter \SET_POLARITY 1'1
    parameter \CLK_POLARITY 1'1
    parameter \WIDTH 4
    connect \CLR $auto$proc_dff.cc:124:gen_dffsr_complex$19
    connect \SET $auto$proc_dff.cc:117:gen_dffsr_complex$17
    connect \CLK \clk
    connect \Q \out [3:0]
    connect \D \in [3:0]
  end
```

nakengelhardt:
The script you have posted does not at any point map to fine-grain cells, so I am not sure why you would expect that outcome? Mapping to fine-grain would involve a flow that runs `techmap` at some point.

1396510211:
When I execute the techmap pass :
```bash attribute \src "test.v:9.51-25.5"
  cell $_DFFSR_PPP_ $auto$90
    connect \Q \out [2]
    connect \D \in [2]
    connect \R $auto$19 [2]
    connect \S $auto$17 [2]
    connect \C \clk
  end

  attribute \src "test.v:9.51-25.5"
  cell $_DFFSR_PPP_ $auto$89
    connect \Q \out [1]
    connect \D \in [1]
    connect \R $auto$19 [1]
    connect \S $auto$17 [1]
    connect \C \clk
  end

  attribute \src "test.v:9.51-25.5"
[...]

Issue #344, opened 2017-05-13
Verilog hieararchical names with $readmemh


niels-moller:
I'm trying to initialize a memory block using $readmemh, and I don't want to do the initialization in the module defining the ram, but in my top module. Here's a complete but cut down example: 

```
module ram_unit(input clk, input rst,
		input cyc, input stb,
		input we, output ack,
		input [SIZE-1:0]  adr,
		input [63:0] 	  dat_write,
		output reg [63:0] dat_read);
   parameter  SIZE = 3; // Address size
   reg [63:0] mem [0:(1 << SIZE)-1];
   integer    i;

   // Always completes in a single cycle.
   assign ack = 1;

   // Written this way due to
   // https://github.com/cliffordwolf/yosys/issues/335
   always @(posedge clk)
     if (cyc & stb & ~we) begin
	dat_read <= mem[adr];
     end
   always @(posedge clk)
     if (cyc & stb & we) begin
	mem[adr] <= dat_write;
     end
endmodule

module main(input clk);
   parameter SIZE = 3;

   reg rst;
   reg [SIZE-1:0] adr;
   reg cyc;
   reg stb;
   reg we;
   wire ack;
   reg [63:0] dat_write;
   wire [63:0] dat_read;
          
   ram_unit #(SIZE) 
   ram(clk, rst, cyc, stb, we, ack,
       adr, dat_write, dat_read);

   initial begin
      $readmemh("/dev/null", ram.mem);
   end
endmodule
```

This is accepted by iverilog -Wall, but not by yosys:

```
$ yosys -q -f verilog -p synth_ice40 readmemh-1.vl 
ERROR: Failed to evaluate system function `\$readmemh' with non-memory 2nd argument at readmemh-1.vl:43.
```

As a work-around, I tried changing the initial block to read into a locally defined memory, and then copying it to the target memory, like this:

```
module ram_unit(input clk, input rst,
		input cyc, input stb,
		input we, output ack,
		input [SIZE-1:0]  adr,
		input [63:0] 	  dat_write,
		output reg [63:0] dat_read);
   parameter  SIZE = 3; // Address size
   reg [63:0] mem [0:(1 << SIZE)-1];
   integer    i;

   // Always completes in a single cycle.
   assign ack = 1;

   // Written this way due to
   // https://github.com/cliffordwolf/yosys/issues/335
   always @(posedge clk)
     if (cyc & stb & ~we) begin
	dat_read <= mem[adr];
     end
[...]

PR #5570, opened 2025-12-23
Add aldff splitting to avoid logic data loops

likeamahoney:
Hi all,

This PR implements a fix for issue https://github.com/YosysHQ/yosys/issues/5569.

The approach splits the output of the proc pass into multiple chunks, which are then mapped to different trigger types (`$dff`, `$dffe`, or `$aldff`) depending on whether the corresponding output bits are constant, self-dependent, or neither.

I also added a set of tests to validate this solution. I would appreciate it if you could point out any proc block patterns or cases that I may have missed.

### DIFF
--- a/passes/proc/proc_dff.cc
+++ b/passes/proc/proc_dff.cc
@@ -144,6 +144,38 @@ void gen_dff(RTLIL::Module *mod, RTLIL::SigSpec sig_in, RTLIL::Const val_rst, RT
 	log(".\n");
 }
 
+
+void gen_dffe(RTLIL::Module *mod, RTLIL::SigSpec sig_in, RTLIL::SigSpec sig_out,
+		bool clk_polarity, bool en_polarity, RTLIL::SigSpec clk, RTLIL::SigSpec *en, RTLIL::Process *proc)
+{
+	std::stringstream sstr;
+	sstr << "$procdffe$" << (autoidx++);
+
+	RTLIL::Cell *cell = mod->addCell(sstr.str(), clk.empty() ? ID($ff) : ID($dffe));
+	cell->attributes = proc->attributes;
+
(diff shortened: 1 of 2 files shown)


widlarizer:
First off: _sequential_ logic loops aren't expected to be a problem for any use case, they're a totally valid pattern, right?

About splitting flops: I'm convinced that synthesis flows benefit from later `opt_dff` runs to simplify things. I think multi-bit registers can't be optimized as well if it would require slicing (only constant bits get cut out I think) but every synthesis flow we maintain ends up slicing them all into fine-grained single-bit cells and running `opt_dff`, so for synthesis I don't see value in slicing within `proc` at all

widlarizer:
Also this is related to #3442 which I think we're yet to fully address

likeamahoney:
> First off: _sequential_ logic loops aren't expected to be a problem for any use case, they're a totally valid pattern, right?
> 
> About splitting flops: I'm convinced that synthesis flows benefit from later `opt_dff` runs to simplify things. I think multi-bit registers can't be optimized as well if it would require slicing (only constant bits get cut out I think) but every synthesis flow we maintain ends up slicing them all into fine-grained single-bit cells and running `opt_dff`, so for synthesis I don't see value in slicing within `proc` at all

Hi, Thank you for review! 

Given that this is related to [#344](https://github.com/YosysHQ/yosys/issues/3442) and that the underlying issue is not fully resolved yet, one possible interim approach could be to keep my fix behind an option and eventually move this logic into `opt_dff` itself, once there is a clearer consensus on how multi-bit flops should be handled in that context. If this change is not necessary, I can close this PR. Please let me know.

widlarizer:
Ultimately, I think this is already covered by #4781 which is a thorough proc_dff rework. I'll probably be transferring the logic from that PR into `opt_dff` quite soon

PR #5575, opened 2025-12-29
Update ABC as per 2025-12-29

mmicko:


mmicko:
@whitequark can you please take a look at WASI fix at https://github.com/YosysHQ/abc/commit/a47467b5cef16ebf93b551405eac2adff1feaf1d tried to make it small as possible so easier to maintain later, but wonder if we can clean something more before submitting this upstream.

whitequark:
Looks good other than pthread_exit--I think that function should be `noreturn`. (I'm not sure whether this ever matters, and my guess is it doesn't, but `noreturn` functions returning tends to cause incredibly obscure bugs.)

PR #5580, opened 2026-01-03
verific: add -set_ignore_translate_off option

dhvll:
This PR adds a `verific -set_ignore_translate_off` option that instructs Verific to ignore `translate_off/translate_on` pragmas during frontend elaboration.

Silimate’s Yosys fork includes an option to disable `translate_off/on` handling explicitly.

The change adds a simple option that invokes `veri_file::SetIgnoreTranslateOff(1)` when specified. The option is parsed as a boolean toggle, guarded under `VERIFIC_SYSTEMVERILOG_SUPPORT`, and does not affect default behavior unless explicitly enabled.

### DIFF
--- a/frontends/verific/verific.cc
+++ b/frontends/verific/verific.cc
@@ -3688,15 +3688,15 @@ struct VerificPass : public Pass {
 				veri_file::AddLOption(args[++argidx].c_str());
 				continue;
 			}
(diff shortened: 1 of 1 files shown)


mmicko:
Deprecated by https://github.com/YosysHQ/yosys/pull/5619

PR #5581, opened 2026-01-03
verific: add -set_relaxed_checking option

dhvll:
This PR adds a `verific -set_relaxed_checking` option that enables relaxed semantic checking in the Verific frontend.

Silimate’s Yosys fork includes an option to relax Verific’s checking rules to allow such designs to be processed.

Upstreaming this option provides users with an explicit, opt-in mechanism to enable relaxed checking while preserving default strict behavior.

The change introduces a boolean option that calls `VeriNode::SetRelaxedChecking(1)` when specified. The behavior is fully
guarded under `VERIFIC_SYSTEMVERILOG_SUPPORT` and only takes effect when the option is explicitly passed.

### DIFF
--- a/frontends/verific/verific.cc
+++ b/frontends/verific/verific.cc
@@ -3688,15 +3688,16 @@ struct VerificPass : public Pass {
 				veri_file::AddLOption(args[++argidx].c_str());
 				continue;
 			}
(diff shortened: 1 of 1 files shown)


mmicko:

PR #5582, opened 2026-01-03
verific: add -set_relaxed_file_ext_modes option

dhvll:
This PR adds a `verific -set_relaxed_file_ext_modes` option that relaxes
Verific’s file extension to language mapping.

Silimate’s Yosys fork includes an option that configures Verific to treat a broader set of file extensions as SystemVerilog.

Upstreaming this option allows users to explicitly enable this relaxed behavior without changing default file handling semantics.

When specified, the option updates Verific’s file extension modes to associate several common extensions (e.g. `.v`, `.vh`, `.sv`, `.svh`) with SystemVerilog parsing. The change is guarded under `VERIFIC_SYSTEMVERILOG_SUPPORT` and does not alter behavior unless the option is used.

### DIFF
--- a/frontends/verific/verific.cc
+++ b/frontends/verific/verific.cc
@@ -3688,23 +3688,23 @@ struct VerificPass : public Pass {
 				veri_file::AddLOption(args[++argidx].c_str());
 				continue;
 			}
-			if (args[argidx] == "-set_relaxed_file_ext_modes") {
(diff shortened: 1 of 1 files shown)


mmicko:
Deprecated by https://github.com/YosysHQ/yosys/pull/5619

PR #5583, opened 2026-01-03
verific: add -set_relaxed_file_libext_modes option

dhvll:
This PR adds a `verific -set_relaxed_file_libext_modes` option that relaxes Verific’s library file extension handling.

Silimate’s Yosys fork includes an option to extend Verific’s recognized library file extensions explicitly.

Upstreaming this option allows users to opt into broader library extension support while keeping default behavior unchanged.

The option updates Verific’s library extension list to include additional common HDL extensions. The behavior is guarded under
`VERIFIC_SYSTEMVERILOG_SUPPORT` and only takes effect when the option is explicitly specified.


mmicko:

PR #5584, opened 2026-01-03
verific: add -set_vhdl_default_library_path option

dhvll:
This PR adds a `verific -set_vhdl_default_library_path` option that allows users to specify default VHDL library search paths for the Verific frontend.

Silimate’s Yosys fork includes an option to configure Verific’s default VHDL library path from the command line.

Upstreaming this option enables better support for VHDL workflows without affecting existing behavior for users who do not rely on VHDL.

The option consumes the remaining command-line arguments as library paths and configures them via `vhdl_file::SetDefaultLibraryPath()`. Since the option consumes positional arguments, option parsing intentionally terminates after this flag. The behavior is guarded under `VERIFIC_SYSTEMVERILOG_SUPPORT` and `VERIFIC_VHDL_SUPPORT`.

### DIFF
--- a/frontends/verific/verific.cc
+++ b/frontends/verific/verific.cc
@@ -3688,6 +3688,14 @@ struct VerificPass : public Pass {
 				veri_file::AddLOption(args[++argidx].c_str());
 				continue;
 			}
+			if (GetSize(args) > argidx && args[argidx] == "-set_vhdl_default_library_path") {
+				for (argidx++; argidx < GetSize(args); argidx++) {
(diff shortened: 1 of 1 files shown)


mmicko (changes requested):
There is no sense for this to be under VERIFIC_SYSTEMVERILOG_SUPPORT check block, it only requires VHDL so it should only be able to be used/compiled if you have Verific with VHDL support.

dhvll:
Are these changes correct ?

mmicko:

PR #5585, opened 2026-01-03
verific: add -ignore_module option

dhvll:
This PR adds a `verific -ignore_module` option that allows users to specify module names that should be ignored during Verific frontend elaboration.

Silimate’s Yosys fork includes an option to explicitly list module names to be ignored by Verific.

Upstreaming this option provides a clear and explicit mechanism to exclude such modules without modifying source files.

The option consumes the remaining command-line arguments as module names and registers them via `veri_file::AddToIgnoredParsedModuleNames()`. Because the option consumes positional arguments, option parsing intentionally terminates after this flag. The behavior is guarded under `VERIFIC_SYSTEMVERILOG_SUPPORT`.

### DIFF
--- a/frontends/verific/verific.cc
+++ b/frontends/verific/verific.cc
@@ -3688,6 +3688,13 @@ struct VerificPass : public Pass {
 				veri_file::AddLOption(args[++argidx].c_str());
 				continue;
 			}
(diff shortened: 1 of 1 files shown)


mmicko:

PR #5586, opened 2026-01-04
verific: add -no_split_complex_ports option

dhvll:
This PR adds a new `verific -no_split_complex_ports` option that disables Verific’s default behavior of splitting complex ports during frontend elaboration.

Silimate’s Yosys fork includes an option to disable this behavior explicitly.

Upstreaming this option allows users to control Verific’s port handling behavior while keeping the default behavior unchanged.

The change introduces a boolean `verific_no_split_complex_ports` flag that is set when the `-no_split_complex_ports` option is passed to the `verific` frontend.

When enabled, the flag disables Verific’s complex port splitting behavior during elaboration. The implementation is guarded under
`VERIFIC_SYSTEMVERILOG_SUPPORT` and does not affect existing behavior unless the option is explicitly used.

### DIFF
--- a/backends/verilog/verilog_backend.cc
+++ b/backends/verilog/verilog_backend.cc
@@ -386,14 +386,31 @@ void dump_sigspec(std::ostream &f, const RTLIL::SigSpec &sig)
 		f << "{0{1'b0}}";
 		return;
 	}
+	if (sig.is_fully_const() && GetSize(sig) > 8192) {
+		f << stringf("{ ");
+		int i = 0;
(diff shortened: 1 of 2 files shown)


mmicko (changes requested):
This should not be inside VERIFIC_SYSTEMVERILOG_SUPPORT, it is valid option in any case since it port bus can be generated in case of VHDL as well.

dhvll:
Are these changes correct ?

mmicko:

PR #5591, opened 2026-01-07
Improve handling of empty switches

### LINKED ISSUE #5572: Invalid verilog generated for empty RTLIL `switch` with attributes
### Version

Yosys 0.54 (git sha1 db72ec3bd, g++ 15.2.1 -march=x86-64-v3 -mtune=haswell -O3 -fno-plt -fexceptions -fstack-clash-protection -fcf-protection -mpclmul -ffile-prefix-map=/startdir/src=/usr/src/debug/yosys -fPIC -O3)

Linux

### Reproduction Steps

```rtlil
attribute \top 1
module \top
  wire width 1 \sig
  wire width 1 \val

  process $2
    switch \sig [0]
      case 1'0
      case 1'1
      case
        assign \val [0] 1'1
    end
  end
end
```

Or any other exhaustive switch for which there are no expressions for "normal" cases, but there is some expression in the default (dead) case.

- `read_rtlil`
- `proc -nomux`
- `write_verilog`

### Expected Behavior

Valid verilog is generated

### Actual Behavior

```verilog
module top();
  reg \$auto$verilog_backend.cc:2359:dump_module$1  = 0;
  wire sig;
  wire val;
  always @* begin
    if (\$auto$verilog_backend.cc:2359:dump_module$1 ) begin end
    (* full_case = 32'd1 *)
    
  end
endmodule
```

See `(* full_scase = 32'd1 *)`. As it is an attribute it should be attached to something (e.g empty expression), so there should be a `;` before the `end` clause.

KrystalDelusion:

Close #5572 by resolving two problems:

1. Attributes on empty switches produce invalid verilog
2. `full_case` switches with all empty cases require two calls to `proc_clean`

1. Skip empty switches in `write_verilog`
2. Set `did_something = true` when clearing a `full_case` switch

tests/verilog/bug5572.ys and tests/proc/bug5572.ys from #5572 respectively for the two fixes.

### DIFF
--- a/tests/proc/bug5572.ys
+++ b/tests/proc/bug5572.ys
@@ -1,0 +1,19 @@
+read_rtlil << EOT
+attribute \top 1
+module \top
+  wire width 1 \sig
+  wire width 1 \val
+
+  process $2
+    switch \sig [0]
+      case 1'0
+      case 1'1
+      case
+        assign \val [0] 1'1
+    end
+  end
+end
+EOT
+proc_rmdead
+proc_clean
+select -assert-none p:*
--- a/tests/verilog/.gitignore
+++ b/tests/verilog/.gitignore
@@ -1,3 +1,4 @@
+/bug5572.v
 /const_arst.v
 /const_sr.v
 /doubleslash.v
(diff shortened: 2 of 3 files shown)


whitequark (approved):
LGTM, thanks for handling this!

PR #5592, opened 2026-01-07
More helpful error messages when loading Yosys Witness files with `yosys-smtbmc` 

### LINKED ISSUE #5503: `yosys-smtbmc` cannot load Yosys witness when design differs slightly; error message unhelpful
### Version

Yosys 0.56+171 (git sha1 6fdcdd41d, aarch64-apple-darwin23.5-clang++ 18.1.8 -fPIC -O3)

macOS

### Reproduction Steps

[witness_mismatch_demo.zip](https://github.com/user-attachments/files/23729799/witness_mismatch_demo.zip)

```
cd witness_mismatch_demo
./run.sh
```

This demo runs a toy two-stage SBY flow, where our goal is to produce a witness in stage 1 and use it in stage 2. The underlying design is the same, but stage 1 and 2 have different formal components (gated by `WAYPOINT`). 

`stage1.sby` defines `WAYPOINT`, adds one extra formal-only flop, forces `assert(0)`, and dumps `stage1/engine_0/trace.yw`.

`stage2.sby` uses the exact same DUT but *without* `WAYPOINT` (so the waypoint flop is absent) and replays `--yw src/trace.yw --check-witness`.

### Expected Behavior

One of two things could be expected here.

1. If we determine that Yosys/sby should support slight variations in the design between stages (which would make the tool more flexible, but would require some work to implement), then this should "just work" without any error.
2. BUT, even if we decide not to implement this, at bare minimum there should be a clear error here telling us that the design is different and so the witness can't be loaded.

### Actual Behavior

`yosys-smtbmc` aborts with an unhelpful error:
  ```
  File ".../yosys-smtbmc", line 795, in ywfile_signal
    assert pos == sig.width
  AssertionError
  ```
  then SBY prints “engine terminated without status.” No message explains the witness/design mismatch.

Again, there are really two issues posed in this issue:
* Should we have a way to load a witness file onto a design that differs slightly? (or does one exist?)
* When a design differs slightly from the witness, a more helpful error should be thrown.

Regarding the question of whether a witness file can be loaded onto a design that mismatches: I think there are likely nuances to this that I don't yet fully understand. It may be the case that this sometimes works -- I think it depends on what signals are present in the witness file and how each design's signals are named.

Not supporting witness files that aren't an exact match is probably a sane default. In that case, we should at least have better error messages. Additionally, we could also provide easy ways to transform a witness to be used for another design.

## Update 2025-11-25

I have a bit more clarity on why this occurs. The signals in witness files are stored using their names. Even if two files are largely the same (e.g. one file just has an extra SVA), it's very likely that many of the names in the file will not match due to uniquification/`autoidx`. Even if the change is small, it can lead to cascading effects that affect the uniquification of many signals in the design.

Some workarounds here could be:
- Write your two designs so that uniquification gives the same name to important signals (this may not be possible, and seems challenging based on my limited experimentation). 
- Provide a way to map between signals.
  - a simple translation script that rewrites witness files
  - an attribute on specific signals/constructs in the design that will force the use of specific names when writing to witness files

Originally, we thought that only _adding_ SVAs to the second file wouldn't lead to issues. However, this doesn't seem to be the case. If the witness from stage 1 includes `\witness` signals, then these signals are likely (due to uniquification) to have a different name in the stage 2 design. Some of our simpler demos wrote out witness files which didn't include `\witness` signals, which can be successfully used in stage 2 without error. So the question is: will any sufficiently complex example lead to `\witness` signals being included in the stage 1 output witness?

## Update 2025-12-01

Note: Jannis mentioned the correct way to do this would be to make one overall verification file with all assertions, and then disable assertions in each stage via scripting.

gussmith23:
Closes #5503.

#5503 points out that the error messages from `yosys-smtbmc` are not helpful when we load an invalid witness file. This PR adds more helpful error messages.

Adds more helpful error messages for a number of cases:
- Indexing a memory out of bounds (previously seems like it would have generated invalid code)
- Memory or signal not found (previously gave an unhelpful error message)
- Some bits of signal left uncovered by witness file (previously  gave an unhelpful error message)

I'm not as sure on the memory portion of the code -- would love someone with more experience with memories to take a look.

Test provided!

### DIFF
--- a/backends/smt2/smtbmc.py
+++ b/backends/smt2/smtbmc.py
@@ -735,6 +735,12 @@ def ywfile_signal(sig, step, mask=None):
 
     output = []
 
+    def ywfile_signal_error(reason, detail=None):
+        msg = f"Yosys witness signal mismatch for {sig.pretty()}: {reason}"
+        if detail:
+            msg += f" ({detail})"
+        raise ValueError(msg)
+
     if sig.path in smt_wires:
         for wire in smt_wires[sig.path]:
             width, offset = wire["width"], wire["offset"]
@@ -765,6 +771,12 @@ def ywfile_signal(sig, step, mask=None):
             for mem in smt_mems[sig.memory_path]:
                 width, size, bv = mem["width"], mem["size"], mem["statebv"]
 
+                if sig.memory_addr is not None and sig.memory_addr >= size:
+                    ywfile_signal_error(
+                        "memory address out of bounds",
+                        f"address={sig.memory_addr} size={size}",
+                    )
+
                 smt_expr = smt.net_expr(topmod, f"s{step}", mem["smtpath"])
 
                 if bv:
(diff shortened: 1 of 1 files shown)


hexratcc:
LGTM - One minor thought: might be worth using a custom exception instead of `ValueError` for easier downstream handling. Just a nitpick, though.

Issue #5574, opened 2025-12-29
optimization pass crashes with cmd error aborting


zhihaoxu1325:
### Version

Yosys 0.58+132 (git sha1 12cb8e951, g++ 11.4.0-1ubuntu1~22.04.2 -fPIC -O3)

### On which OS did this happen?

Linux

### Reproduction Steps

Run 
```
yosys -v 3 -p "                                                                                                                                                                          
read_verilog /tmp/test.v;
synth;
abc -script +resub,-K,8;,refactor,-z;,if,-K,4;,mfs2,-W,10;,mfs;,mfs2
"
```

### Expected Behavior

Normally optimization

### Actual Behavior

3.1. Extracting gate netlist of module `\fuzz_mwoqk' to `<abc-temp-dir>/input.blif'..
** cmd error: aborting 'source /tmp/yosys-abc-XXXXXX/abc.script'

[test.v.zip](https://github.com/user-attachments/files/24364921/test.v.zip)

KrystalDelusion:
Calling `abc` works fine, so the problem here is with your abc script.  However, this does reveal that that the multithreaded ABC doesn't appear to catch an abort and hangs instead.  @rocallahan do you think you could take a look at that?

Minimised reproducer, should give an ABC error (`Error: This command can only be applied to an AIG (run "strash").` when run by hand), instead hangs:

```
read_verilog << EOT
module fuzz_mwoqk (input i0, output o0);
  assign o0 = i0 ^ 1;
endmodule
EOT
synth
abc -script +resub,-K,8;
```

rocallahan:
PR #5593 looks good to me.

KrystalDelusion:
This is still affecting builds that use `READLINE=0`, but for a normal build this is resolved.

KrystalDelusion:
Fixed with #5669

PR #5593, opened 2026-01-07
abc: handle ABC script errors instead of hanging

RCoeurjoly:
Fixing https://github.com/YosysHQ/yosys/issues/5574
By checking if ABC has printed an error.
Credit for the test goes to @KrystalDelusion

### DIFF
--- a/tests/techmap/bug5574.ys
+++ b/tests/techmap/bug5574.ys
@@ -1,4 +1,7 @@
-logger -expect error "ABC: Error: This command can only be applied to an AIG" 1
+# On Linux, with a spawned abc, this message is the error
+# otherwise the error is the failure to load the output.blif
+logger -expect log "ABC: Error: This command can only be applied to an AIG" 1
+logger -expect error "ABC" 1
 read_verilog << EOT
 module fuzz_mwoqk (input i0, output o0);
   assign o0 = i0 ^ 1;


rocallahan:
This looks good. I think you just need commit https://github.com/rocallahan/yosys/commit/f4c38f0a16bba6b1c39460cee441c52a9ca243c6 to mark the test as expected to error out.

RCoeurjoly:
> This looks good. I think you just need commit [rocallahan@f4c38f0](https://github.com/rocallahan/yosys/commit/f4c38f0a16bba6b1c39460cee441c52a9ca243c6) to mark the test as expected to error out.

Thanks! That was it

RCoeurjoly:
The two jobs that are failing on this PR:

- **Run tests (macos-latest)**: fails on macOS, but I don’t see a clear error in the logs. I don’t have access to a macOS environment, so I can’t reproduce or debug this locally.

- **Try build docs (html)**: fails with “the self-hosted runner lost communication with the server” (https://github.com/YosysHQ/yosys/actions/runs/20811203375/job/59775670768?pr=5593), which looks like a runner/infrastructure issue. I also don’t have permissions to retry the job.

I don’t see a concrete action I can take from my side at the moment.

mmicko:
@RCoeurjoly When looking into errors best is to take a look at raw logs

Here is the relevant part when it fails on macOS 

```
2026-01-08T09:32:46.3945840Z 2.26. Executing CHECK pass (checking for obvious problems).
2026-01-08T09:32:46.3946070Z Checking module fuzz_mwoqk...
2026-01-08T09:32:46.3946250Z Found and reported 0 problems.
2026-01-08T09:32:46.3946360Z 
2026-01-08T09:32:46.3946460Z 3. Executing ABC pass (technology mapping using ABC).
2026-01-08T09:32:46.3946620Z 
2026-01-08T09:32:46.3946790Z 3.1. Extracting gate netlist of module `\fuzz_mwoqk' to `<abc-temp-dir>/input.blif'..
2026-01-08T09:32:46.3947040Z 
2026-01-08T09:32:46.3947080Z 3.1.1. Executed ABC.
2026-01-08T09:32:46.3947320Z Extracted 1 gates and 2 wires to a netlist network with 1 inputs and 1 outputs.
2026-01-08T09:32:46.3947620Z Running ABC script: <abc-temp-dir>/abc.script
2026-01-08T09:32:46.3947870Z ABC: ======== ABC command line "source <abc-temp-dir>/abc.script"
2026-01-08T09:32:46.3948130Z ABC: + read_blif <abc-temp-dir>/input.blif 
2026-01-08T09:32:46.3948500Z ABC: + read_library /var/folders/03/bcr7nd0x5lz0x5lkgq6vrh5w0000gn/T/yosys-abc-c7GyCl/stdcells.genlib 
2026-01-08T09:32:46.3948850Z ABC: + resub -K 8 
2026-01-08T09:32:46.3949050Z ABC: ** cmd error: aborting 'source <abc-temp-dir>/abc.script'
2026-01-08T09:32:46.3949360Z ABC: Error: This command can only be applied to an AIG (run "strash").
2026-01-08T09:32:46.3949830Z ERROR: Can't open ABC output file `/var/folders/03/bcr7nd0x5lz0x5lkgq6vrh5w0000gn/T/yosys-abc-c7aEoo/output.blif'.
2026-01-08T09:32:46.3950350Z ERROR: Expected error pattern 'ABC: Error: This command can only be applied to an AIG' not found !
2026-01-08T09:32:46.4046560Z Post job cleanup.
2026-01-08T09:32:46.4133910Z Post job cleanup.
```

KrystalDelusion:
For non-threaded abc usage (e.g. macOS, or Linux with `YOSYS_LINK_ABC`), the AIG error is a regular log rather than an error log.  We can still check it appears, but we can't use it as the expected error message.

KrystalDelusion:
@RCoeurjoly @rocallahan 

If ABC is built without readline, e.g. when building Yosys with `ENABLE_READLINE=0 ENABLE_EDITLINE=1` (making sure to `make clean-abc` first since it won't automatically rebuild for a makeflag change), the `read_until_abc_done` function doesn't get the `ERROR` *or* the `abc %02d> `.  That means that `bug5574.ys` still hangs even with this PR merged, and with #5651 merged *all* calls to abc hang.

I can resolve the hang introduced by #5651 with the following change, but it still hangs indefinitely if abc aborts.  Can one of you try to fix the error checking without readline?  Maybe checking stderr for `** cmd error: aborting 'source <abc-temp-dir>/abc.script'`?  In the meantime I am recommending that for releasing the next version of Yosys we disable parallel abc execution if Yosys is built without readline (though it's still not perfect, since an external abc could be missing readline even if Yosys isn't).

```diff
diff --git a/passes/techmap/abc.cc b/passes/techmap/abc.cc
index e9d02b85c..9aac03842 100644
--- a/passes/techmap/abc.cc
+++ b/passes/techmap/abc.cc
@@ -1154,6 +1154,7 @@ bool read_until_abc_done(abc_output_filter &filt, int fd, DeferredLogs &logs) {
 	std::string line;
 	char buf[1024];
 	bool seen_source_cmd = false;
+	bool seen_yosys_abc_start = false;
 	bool seen_yosys_abc_done = false;
 	while (true) {
 		int ret = read(fd, buf, sizeof(buf) - 1);
@@ -1190,11 +1191,17 @@ bool read_until_abc_done(abc_output_filter &filt, int fd, DeferredLogs &logs) {
 			filt.next_line(line);
 			if (is_prompt && rest.substr(0, 7) == "source ")
 				seen_source_cmd = true;
+			if ((is_prompt ? rest : line).substr(0, 15) == "YOSYS_ABC_START")
+				seen_yosys_abc_start = true;
 			if (line.substr(0, 14) == "YOSYS_ABC_DONE")
 				seen_yosys_abc_done = true;
 			line.clear();
 		}
 		line.append(start, end - start);
+		// Catch successful completion when we don't have proper prompts
+		// (e.g. ABC was built without readline)
+		if (!seen_source_cmd && seen_yosys_abc_start && seen_yosys_abc_done)
+			return true;
 	}
 }
 
@@ -1381,6 +1388,7 @@ void RunAbcState::run(ConcurrentStack<AbcProcess> &process_pool)
 			return;
 		}
 		std::string cmd = stringf(
+				"echo \"YOSYS_ABC_START\"\n"
 				"empty\n"
 				"source %s\n", tmp_script_name);
 		int ret = write(process.to_child_pipe, cmd.c_str(), cmd.size());
```

rocallahan:
I don't think we need `YOSYS_ABC_START`. We can just stop unconditionally when we see `YOSYS_ABC_DONE`. Of course we would still hang when there is an ABC error.

> Maybe checking stderr for ** cmd error: aborting 'source <abc-temp-dir>/abc.script'?

This can be done but it requires significant additional complexity because we have to read from both stdout OR stderr using `select()` or something similar. Duping stdout and stderr into a single fd would not work correctly because, due to stdout buffering, any token like `YOSYS_ABC_DONE` could be interrupted by output on stderr.

rocallahan:
The actual problem here is that ABC still generates the prompts, but without `ENABLE_READLINE` it doesn't flush stdout so Yosys never sees the prompts :-(.

rocallahan:
I think rather than make this ABC process-monitoring code even more complex, we should instead be more conservative about reusing ABC processes. PR #5671 makes us only reuse ABC processes if we're running the `yosys_abc_executable` and Yosys was built with `ENABLE_READLINE`.

Note that this is about reusing ABC processes, not parallelism itself. When this is disabled we can still have parallelism, but in large designs we might be spawning ABC thousands of times, which is slow.

KrystalDelusion:
> The actual problem here is that ABC still generates the prompts, but without `ENABLE_READLINE` it doesn't flush stdout so Yosys never sees the prompts :-(.

Yeah, I tried to do a hack passing the `process.to_child_pipe` into `read_until_abc_done()` and writing an `echo;\n` or `echo "message"\n` after reading a message (as well as trying before calling `read()`) to see if that would get it to flush but it didn't change anything.

PR #5594, opened 2026-01-08
Check for missing port in SDC code to work around compiler bug

rocallahan:
I am getting weird crashes on `main` in `tests/sdc/alu_sub.ys` which I traced to a null `Wire*` in `SdcObjects::constrained_ports`. The null `Wire*` is being set in the `SdcObjects` constructor. I don't understand what's going on here, so I added this check to detect the missing wire early ... and that made the crash go away. Compiler bug maybe? I have `Debian clang version 19.1.7 (3+build5)`, default build configuration.

Anyway this code seems fine to have.

### DIFF
--- a/passes/cmds/sdc/sdc.cc
+++ b/passes/cmds/sdc/sdc.cc
@@ -165,7 +165,12 @@ struct SdcObjects {
 		if (!top)
 			log_error("Top module couldn't be determined. Check 'top' attribute usage");
 		for (auto port : top->ports) {
-			design_ports.push_back(std::make_pair(port.str().substr(1), top->wire(port)));
+			RTLIL::Wire *wire = top->wire(port);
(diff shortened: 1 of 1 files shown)


widlarizer:
welp I'm thoroughly confused about how we'd end up with a module port that doesn't have a wire

rocallahan:
> welp I'm thoroughly confused about how we'd end up with a module port that doesn't have a wire

Right, I think this is a compiler bug ... or possibly there is some UB lurking. To figure out exactly what's going on I'd have to reduce the `sdc.cc` translation unit to a minimized testcase, which would be a lot of work.

rocallahan:
> Got it, if you add a comment about the intended futility of the check I'll merge it to get it out the way of fuzzing

Done.

PR #5605, opened 2026-01-13
Add opt_balance_tree pass

nataliakokoromyti:
This pass converts cascaded chains of arithmetic and logic cells ($add, $mul, $and, $or, $xor) into balanced binary trees to improve timing performance in hardware synthesis. For example, converting sth like `a + b + c + d + e + f` into into balanced binary trees (`((a + b) + (c + d)) + ((e + f) + ...)`) cuts the critical path depth roughly in half.

The pass does a BFS traversal starting from each cell to find chains of the same operation type ($add, $mul, $and, $or, $xor). For each chain:

  1. Walk backwards from the sink to collect all source signals, checking that we only traverse cells of the matching type
  2. Verify all operands have consistent signedness (mixed signed/unsigned chains are skipped)
  3. Recursively build a balanced binary tree by splitting sources at the midpoint
  4. Replace the original chain with the tree

Width calculations handle the fact that adders grow by 1 bit per level and multipliers sum their input widths. The pass preserves all cell attributes and properly handles port connections.

**Command-line flags**: `-arith` for just arithmetic ops, `-logic` for just logic ops, or default for all supported types.

## Tests

  The test suite has 30 cases covering:
  - Basic chains (4-input AND, 8-input ADD)
  - Chains with intermediate outputs and fanout
  - Mixed bit widths and signedness checking
  - Pre-existing partial trees
  - Wide operands (64+ bits)
  - Diamond patterns and reconvergent fanout

Each test uses `equiv_opt` to verify functional equivalence and `select` with traversal operators to check the actual tree depth.

Run with: `./yosys tests/opt/opt_balance_tree.ys

Original implementation by Akash Levy <akash@silimate.com> for Silimate. Upstreamed from https://github.com/Silimate/yosys

### DIFF
--- a/passes/opt/opt_balance_tree.cc
+++ b/passes/opt/opt_balance_tree.cc
@@ -340,6 +340,7 @@ struct OptBalanceTreePass : public Pass {
 	}
 	void execute(std::vector<std::string> args, RTLIL::Design *design) override {
 		log_header(design, "Executing OPT_BALANCE_TREE pass (cell cascades to trees).\n");
+		log_experimental("open_balance_tree");
 
 		// Handle arguments
 		size_t argidx;


widlarizer:
Note that there is no TESTS variable in Makefile so your example command will run all tests

QuantamHD:
This is a great change! Would love to try this on our test cases

passes/opt/opt_balance_tree.cc
```diff
@@ -0,0 +1,356 @@
...
+struct OptBalanceTreeWorker {
+	// Module and signal map
+	Module *module;
+	SigMap sigmap;
+
+	// Counts of each cell type that are getting balanced
+	dict<IdString, int> cell_count;
+
+	// Check if cell is of the right type and has matching input/output widths
+	bool is_right_type(Cell* cell, IdString cell_type) {
```
widlarizer, line 41: This is insufficient to prevent truncation. Truncation happens at various widths of Y for the various cell types and doesn't commute with the operations you're balancing. Counterexample where equivalence fails:
```
read_verilog <<EOF
module top (
  input wire [15:0] wide_a,
  input wire [7:0] med_b,
  input wire [3:0] small_c,
  input wire [1:0] tiny_d,
  input wire single_e,
  input wire single_f,
  output wire [16:0] wide_result
);
  wire [7:0] w = small_c + med_b + single_f;
  assign wide_result = wide_a + w + tiny_d + single_e;
endmodule
EOF
equiv_opt -assert opt_balance_tree
```
`w` is 8 bits wide, so you get that width of the Y port on the adder. This is as big as its larger input. However, addition has a natural Y port size that's one larger than the larger input, and the input discards that top bit in this example. Meanwhile, multipliers have a natural result size of the sum of the input operands. Moving these operands into a larger adder won't be equivalent since it won't do that discard. Also, elsewhere in the code, you're ensuring the signedness matches. It would be an opportunity to improve the result you get here since if for example there's no sign extension on an adder's output the signedness doesn't matter I'm pretty sure as a feature of two's complement

PR #5606, opened 2026-01-14
wreduce: queue traversal

nataliakokoromyti:
Implement iterative queue-based traversal in wreduce pass to propagate width reductions across dependent cells and wires. Previously, wreduce would process all cells once, then all wires once. The new algorithm maintains work queues for both cells and wires, processing them iteratively until no more reductions are possible. When a cell or wire is reduced, dependent cells and wires are added back to the queues for reprocessing.

Added test to verify that width reductions propagate through a chain of operations: (a + b)[3:0] + c, ensuring the first addition is reduced from 9 bits to 4 bits.

### DIFF
--- a/passes/opt/wreduce.cc
+++ b/passes/opt/wreduce.cc
@@ -55,6 +55,7 @@ struct WreduceWorker
 	ModIndex mi;
 
 	std::set<Cell*, IdString::compare_ptr_by_name<Cell>> work_queue_cells;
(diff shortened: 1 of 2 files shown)


nataliakokoromyti:
hey @widlarizer! do you have any feedback on this pr?

widlarizer:
From your description the benefit of this code change is unclear and the test passes on main. Is this a performance improvement?

widlarizer:

PR #5607, opened 2026-01-14
splitnets: add -ports_only and -top_only options

nataliakokoromyti:
Added two new options to the splitnets pass:

- `-ports_only`: Split only module ports, not internal signals. This is useful when you want to split ports while keeping internal signals as multi-bit wires. 

- `-top_only`: Apply splitting only at the top module level, not in submodules. This is helpful for hierarchical designs where you need split signals only at the top-level interface.

These options can be combined with existing flags:
- `splitnets -ports_only`: Split all ports in all modules
- `splitnets -ports_only -top_only`: Split ports only in top module
- `splitnets -ports -top_only`: Split both ports and nets only in top

Added tests that verify wire/port counts for all flag combinations.

### DIFF
--- a/tests/various/test_splitnets.tcl
+++ b/tests/various/test_splitnets.tcl
@@ -7,11 +7,11 @@ proc read_stats { file } {
     set ports 0
     set nets 0
     foreach line [split $result "\n"] {
-	# Match upstream yosys stat format - be permissive about whitespace
-	if [regexp {([0-9]+).*wires} $line -> n] {
+	# Match only lines ending with "wires" or "ports" (not "public wires", "wire bits", etc)
+	if [regexp {([0-9]+)[ \t]+wires[ \t]*$} $line -> n] {
 	    set nets [expr $nets + $n]
 	}
-	if [regexp {([0-9]+).*ports} $line -> n] {
+	if [regexp {([0-9]+)[ \t]+ports[ \t]*$} $line -> n] {
 	    set ports [expr $ports + $n]
 	}
     }


widlarizer:
I'd expect the selection syntax to be sufficient, making this PR unnecessary. Is it not?

nataliakokoromyti:
i believe -ports_only provides functionality that selection syntax cannot express right now. specifically, splitting
only wires where port_id != 0 while leaving internal wires (port_id == 0) as multi-bit. Is there currently an existing selection pattern that can distinguish port wires from internal wires within a module? 

for top_only, it can be replicated with select <modname>; splitnets -ports; select -clear, but you need to know top module's name, so the -top_only flag is convenient for scripting and automated flows bcs it automatically uses design->top_module()

widlarizer:
Yes, there is such selection syntax [documented](https://yosyshq.readthedocs.io/projects/yosys/en/latest/cmd/index_passes_status.html#select-modify-and-view-the-list-of-selected-objects). Your test passes on `main` with the following changes that remove the new features:

```diff
diff --git a/tests/various/test_splitnets.tcl b/tests/various/test_splitnets.tcl
index fe8ef6cb5..8e86a1554 100644
--- a/tests/various/test_splitnets.tcl
+++ b/tests/various/test_splitnets.tcl
@@ -33 +33,2 @@ design -save "pre"
-splitnets -ports_only -top_only
+splitnets -ports x:* A:top %i
+
@@ -41 +42 @@ design -load "pre"
-splitnets -ports_only
+splitnets -ports x:*
@@ -49 +50 @@ design -load "pre"
-splitnets -ports -top_only
+splitnets -ports A:top
```

PR #5609, opened 2026-01-15
Add Design::run_pass()

nataliakokoromyti:
Adds a new run_pass() method to the RTLIL::Design class, providing a convenient API for executing Yosys passes programmatically.

This is particularly useful for PyYosys users who want to run passes on a design object without needing to manually construct Pass::call() invocations. The method wraps Pass::call() with appropriate logging to maintain consistency with command-line pass execution.

Example usage (from Python):
    design = ys.Design()
    # ... build or load design ...
    design.run_pass("hierarchy")
    design.run_pass("proc")
    design.run_pass("opt")

Changes:
- kernel/rtlil.h: Add run_pass() method declaration
- kernel/rtlil.cc: Implement run_pass() method
- tests/unit/kernel/test_design_run_pass.cc: Add unit tests

### DIFF
--- a/tests/pyosys/test_design_run_pass.py
+++ b/tests/pyosys/test_design_run_pass.py
@@ -1,13 +1,20 @@
-from pathlib import Path
-
+﻿from pathlib import Path
 from pyosys import libyosys as ys
 
 __file_dir__ = Path(__file__).absolute().parent
-src = __file_dir__.parent / "simple" / "fiedler-cooley.v"
+add_sub = __file_dir__.parent / "arch" / "common" / "add_sub.v"
 
-design = ys.Design()
-design.run_pass(["read_verilog", str(src)])
-design.run_pass("hierarchy -top up3down5")
-design.run_pass(["proc"])
-design.run_pass("opt -full")
-design.run_pass("select -assert-mod-count 1 up3down5")
+base = ys.Design()
+base.run_pass(["read_verilog", str(add_sub)])
+base.run_pass("hierarchy -top top")
+base.run_pass(["proc"])
+base.run_pass("equiv_opt -assert -map +/ecp5/cells_sim.v synth_ecp5")
+
+postopt = ys.Design()
+postopt.run_pass("design -load postopt")
+postopt.run_pass(["cd", "top"])
+postopt.run_pass("select -assert-min 25 t:LUT4")
+postopt.run_pass("select -assert-max 26 t:LUT4")
+postopt.run_pass(["select", "-assert-count", "10", "t:PFUMX"])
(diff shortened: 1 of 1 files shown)


widlarizer:
Can you share an example of the increased ergonomy when used in Python? Specifically, how you'd call a pass before, and how this PR makes it neater?

donn:
Well before it'd be:

```python
ys.run_pass(cmd, d)
```

but with this it'd be:

```python
d.run_pass(cmd)
```

It's a minor change but I have forgotten the trailing `, d` more than I'd care to admit, which means I'm running with the wrong design. I personally go a bit further and also provide an overload for the `std::vector<std::string>` overload of `Pass::call` to skip having to do string concatenation to construct a command; this is how I hack it in without patching Yosys itself:

```python
def _Design_run_pass(self, *command):
    ys.Pass.call(self, command)


ys.Design.run_pass = _Design_run_pass  # type: ignore
```

widlarizer:
So what you're saying is that these convenience functions could be patched into the python part of pyosys without altering kernel headers? Because I would prefer that

donn:
What I demonstrated was strictly a hack for user scripts (though it works well enough.) The correct way of doing it without touching the kernel headers would be to modify the generator for pyosys/wrappers.cc such that those two lines are emitted:

```diff
 		{using namespace Yosys::RTLIL; py::class_<Design>(m, "Design")
+			.def("run_pass", [](const Design &s, std::vector<std::string> cmd) { Pass::call(cmd, &s); })
+			.def("run_pass", [](const Design &s, std::string cmd) { Pass::call(cmd, &s); })
```

…I don't have the bandwidth to do this in a clean way right now though…

widlarizer:
Thanks, I suggest the PR author puts that alternative up and tests it out to avoid the C++ boilerplate

pyosys/generator.py
```diff
@@ -701,6 +701,16 @@ def process_class(
 
         self.process_class_members(metadata, metadata, cls, basename)
 
+        if basename == "Design":
```
donn, line 704: …That's one way of doing it

nataliakokoromyti: will fix will fix, just testing

donn: if it works it works i mean, this file is hardly a model for best coding practices

tests/pyosys/test_design_run_pass.py
hexratcc, line 1: I'd consider adding an assertion to the test. Currently it only tests that the method doesn't crash, not that it actually works.

donn: @nataliakokoromyti I'd suggest making a copy of tests/pyosys/test_ecp5_addsub.py that uses a mixture of both new methods.

PR #5610, opened 2026-01-15
Add debugon pass for persistent debug logging

nataliakokoromyti:
Adds a new 'debugon' pass that enables debug log messages globally and persistently for the rest of the session. This complements the existing 'debug' pass which only enables debug logging for a single command.

Use case:
- 'debug cmd' - runs 'cmd' with debug logging, then disables it
- 'debugon' - enables debug logging for all subsequent commands

Implementation:
The pass simply increments log_force_debug, which is a global counter that forces debug messages to be displayed. Unlike the 'debug' pass, it does not decrement the counter after execution.

### DIFF
--- a/passes/cmds/trace.cc
+++ b/passes/cmds/trace.cc
@@ -115,14 +115,43 @@ struct DebugPass : public Pass {
 		log("\n");
 		log("Execute the specified command with debug log messages enabled\n");
 		log("\n");
+		log("    debug -on\n");
+		log("    debug -off\n");
+		log("\n");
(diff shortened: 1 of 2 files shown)


nataliakokoromyti:
does anyone know how i can fix the failing test issue? ive confirmed locally that the code runs correctly so maybe clearing the CI cache would help

widlarizer (approved):
Looks ok now, we'll see what a rerun of the failing job does, it's clearly not the fault of the code change

Issue #5194, opened 2025-06-23
$dffsr semantics as implemented differ from models and docs


widlarizer:
### Version

main

### On which OS did this happen?

Linux

### Reproduction Steps

[Liberty file format](https://people.eecs.berkeley.edu/~alanmi/publications/other/liberty07_03.pdf) allows the specification of the value internal variables attain when a DFFSR is being both set and reset with variables `clear_preset_var1` and `clear_preset_var2`. Yosys internal cells [canonically prioritize reset over set in all DFFSR](https://yosyshq.readthedocs.io/projects/yosys/en/latest/cell/gate_reg_ff.html#reg_ff.$_DFFSR_PPP_) and `dfflibmap` ignores the relevant liberty variables `clear_preset_var1` and `clear_preset_var2`.

Some examples of the interpretation of these cells:
- async2sync - reset has priority
- verilog, proc_dff - no assumption about s&r behavior
- dfflibmap - assumes s&r behavior is free to be replaced with any other (disrespects `clear_preset_var*`)
- read_liberty - incorrectly always presents liberty dffsr as "reset has priority" (disrespects `clear_preset_var*`)
- filterlib -verilogsim - correctly models liberty cell behavior (respects `clear_preset_var*`)

### Expected Behavior

Either:
- documentation declares dffsr cells to have an undefined output and state when `set&reset`
- or dfflibmap gets way more complex and worse at mapping because it only picks cells with matching `set&reset` behavior
- or we add more internal flop cell types or parameters to be able to specify this behavior

### Actual Behavior

Some passes secretly assume a different definition of `$dffsr` and related cells than is documented

widlarizer:
This is extra strange. When I say that `filterlib -verilogsim` models it correctly, it's that the verilog *looks* logically correct. However, it can't be converted to a `$dffsr` by `proc_dff`: `ERROR: Multiple edge sensitive events found for this signal!` 
This is the offending Verilog:
```v
module dffsr (D, CLK, CLEAR, PRESET, Q, QN);
  reg IQ, IQN;
  input D;
  input CLK;
  input CLEAR;
  input PRESET;
  output Q;
  assign Q = IQ; // "IQ"
  output QN;
  assign QN = IQN; // "IQN"
  always @(posedge CLK, posedge CLEAR, posedge PRESET) begin
    if ((CLEAR) && (PRESET)) begin
      IQ <= 0;
      IQN <= 0;
    end
    else if (CLEAR) begin
[...]

Issue #5018, opened 2025-04-11
equiv_opt falsely claims equivalence on designs with different unknown modules


widlarizer:
### Version

Yosys 0.51+41 (git sha1 0c689091e, ccache clang++ 16.0.6 -Og -fPIC)

### On which OS did this happen?

Linux

### Reproduction Steps

```
module gold(input X, output Y);
    bar bar(X, Y);
endmodule

module gate(input X, output Y);
    baz baz(X, Y);
endmodule
```

### Expected Behavior

Failure to prove equivalence or other error

### Actual Behavior

Only warnings are emitted and equivalence is claimed to be proven:
```
3. Executing EQUIV_INDUCT pass.
Found 1 unproven $equiv cells in module equiv:
Warning: No SAT model available for cell baz_gate (baz).
Warning: No SAT model available for cell bar_gold (bar).
  Proving existence of base case for step 1. (10 clauses over 5 variables)
  Proving induction step 1. (17 clauses over 8 variables)
  Proof for induction step holds. Entire workset of 1 cells proven!
Proved 1 previously unproven $equiv cells.

4. Executing EQUIV_STATUS pass.
Found 1 $equiv cells in equiv:
  Of those cells 1 are proven and 0 are unproven.
  Equivalence successfully proven!
```

widlarizer:
equiv_make is doing some wild sigmapping here

widlarizer:
This works correctly when either
- blackboxes are provided, and modules are instantiated by named connection list
- blackboxes are provided, modules are instantiated by ordered connection list, and `hierarchy` is called

At no point is a warning emitted if any of this isn't done. Because sigmap ends up picking a different representative, equiv_make does the following, assuming that those ports are inputs:

```
Changing input $2 of cell baz_gate (baz): \Y_gate -> \Y
Changing input $2 of cell bar_gold (bar): \Y_gold -> \Y
Setting undriven nets to undef: \Y_gold
Setting undriven nets to undef: \Y_gate
```

I don't think I'll be fixing this. It's just another case of the most annoying things about RTLIL and about just how DIY the pass sequencing is (we don't reject designs that have incomplete hierarchy because what if it's already working for somebody the way it is, and what if they plan on adding more module definitions later, etc). `equiv_*` are probably okay in the hands of developers and regular users are unlikely to hit this since that's not what `sby` or anything like that builds on

PR #5611, opened 2026-01-15
Add -nocells flag to equiv_make and equiv_opt

nataliakokoromyti:
Passes like `opt_dff` and `dfflegalize` can change cell structure (parameters, types, or names) while preserving wire-level functionality. The `-nocells` flag allows equivalence checking to skip cell matching and verify only wire equivalence, which is more appropriate when cell structure is intentionally modified. 

The `-nocells` flag is added to both `equiv_make` and `equiv_opt` passes. When set, `equiv_make` skips the `find_same_cells()` call, checking only wire equivalence.

For testing, I added `tests/various/equiv_nocells.ys`.

### DIFF
--- a/passes/equiv/equiv_make.cc
+++ b/passes/equiv/equiv_make.cc
@@ -34,6 +34,7 @@ struct EquivMakeWorker
 	vector<string> blacklists;
 	vector<string> encfiles;
 	bool make_assert;
+	bool nocells;
 
 	pool<IdString> blacklist_names;
 	dict<IdString, dict<Const, Const>> encdata;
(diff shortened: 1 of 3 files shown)


widlarizer:
Well, that sounds like `opt_dff` and `dfflegalize` bugs. I'd like to hear more about where `equiv_*` fails you first - it's not ergonomic (#5018) and our flip flop semantics have bugs (#5194, will be fixed soonish) but I am definitely using `equiv_*` in `tests/techmap/dfflibmap.ys` for example.

widlarizer:

PR #5612, opened 2026-01-16
verific: add explicit System Verilog 2017 option

mmicko:
Option "-sv2012" now properly set to SYSTEM_VERILOG_2012 and added option "-sv2017" to set to SYSTEM_VERILOG_2017

Option "-sv" will always use latest standard depending of current Verific support (when set to SYSTEM_VERILOG)

### DIFF
--- a/frontends/verific/verific.cc
+++ b/frontends/verific/verific.cc
@@ -3114,10 +3114,11 @@ struct VerificPass : public Pass {
 		//   |---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|
 		log("\n");
 #ifdef VERIFIC_SYSTEMVERILOG_SUPPORT
-		log("    verific {-vlog95|-vlog2k|-sv2005|-sv2009|-sv2012|-sv2017|-sv} <verilog-file>..\n");
+		log("    verific {-vlog95|-vlog2k|-sv2005|-sv2009|-sv2012|\n");
+		log("             -sv2017|-sv} <verilog-file>..\n");
 		log("\n");
 		log("Load the specified Verilog/SystemVerilog files into Verific.\n");
-		log("Note that -sv option will use parser for latest supported standard.\n");
+		log("Note that -sv option will use latest supported SystemVerilog standard.\n");
 		log("\n");
 		log("All files specified in one call to this command are one compilation unit.\n");
 		log("Files passed to different calls to this command are treated as belonging to\n");


frontends/verific/verific.cc
```diff
@@ -3114,9 +3114,10 @@ struct VerificPass : public Pass {
 		//   |---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|
 		log("\n");
 #ifdef VERIFIC_SYSTEMVERILOG_SUPPORT
-		log("    verific {-vlog95|-vlog2k|-sv2005|-sv2009|-sv2012|-sv} <verilog-file>..\n");
+		log("    verific {-vlog95|-vlog2k|-sv2005|-sv2009|-sv2012|-sv2017|-sv} <verilog-file>..\n");
 		log("\n");
 		log("Load the specified Verilog/SystemVerilog files into Verific.\n");
+		log("Note that -sv option will use parser for latest supported standard.\n");
```
KrystalDelusion: I think it is sufficient to say "will use latest supported" without mentioning parser, since this is the first description of the command. May be worth making "latest supported SystemVerilog standard" explicit, though given the vlog standards are older than all the sv ones it probably isn't necessary.

frontends/verific/verific.cc
```diff
@@ -3114,9 +3114,10 @@ struct VerificPass : public Pass {
 		//   |---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|---v---|
 		log("\n");
 #ifdef VERIFIC_SYSTEMVERILOG_SUPPORT
-		log("    verific {-vlog95|-vlog2k|-sv2005|-sv2009|-sv2012|-sv} <verilog-file>..\n");
+		log("    verific {-vlog95|-vlog2k|-sv2005|-sv2009|-sv2012|-sv2017|-sv} <verilog-file>..\n");
```
KrystalDelusion: This goes over the line length limit

frontends/verific/verific.cc
```diff
@@ -3161,7 +3162,7 @@ struct VerificPass : public Pass {
 #endif
 #ifdef VERIFIC_SYSTEMVERILOG_SUPPORT
 		log("    verific {-f|-F} [-vlog95|-vlog2k|-sv2005|-sv2009|\n");
-		log("                     -sv2012|-sv|-formal] <command-file>\n");
+		log("                     -sv2012|-sv2017|-sv|-formal] <command-file>\n");
```
KrystalDelusion, line 3166: Is there a reason we list `-formal` for command files but not normal files?  It works for both, right?  Might also be good to mention somewhere (if we don't already) that `-formal` uses the same parser as `-sv`.

frontends/verific/verific.cc
mmicko: Have updated this one as proposed

frontends/verific/verific.cc
mmicko: Thanks for noticing this

frontends/verific/verific.cc
mmicko: `-formal` is described in section above,  this is place where `-f/-F` option is described and language options are just enumerated

PR #4269, opened 2024-03-07
Avoid `module_not_derived` on internal cells in techmap result

povik:
With the following

```
read_verilog <<EOF
module top(a, b, y);
	input wire [7:0] a;
	input wire [7:0] b;
	output wire [7:0] y;
	assign y = a + b;
endmodule
EOF
prep
alumacc
techmap -max_iter 1
dump t:$lcu
```

you get

```
  attribute \module_not_derived 1
  attribute \src "test.v:5.13-5.18|.../yosys/share/techmap.v:286.27-286.69"
  cell $lcu $auto$alumacc.cc:485:replace_alu$2.lcu
    parameter \WIDTH 8
    connect \CO $auto$alumacc.cc:502:replace_alu$4
    connect \CI 1'0
    connect \G $techmap$auto$alumacc.cc:485:replace_alu$2.$and$/Users/pk/repos/yosys/share/techmap.v:286$84_Y
    connect \P $auto$alumacc.cc:501:replace_alu$3
  end
```

where the `module_not_derived` attribute is misplaced. It doesn't make sense on an internall cell and `hierarchy` won't clean it.

To address this, we want `read_verilog -icells` not to place the attribute on internal cells, and `techmap` to use the frontend in the `-icells` mode instead of fixing up the cell types after the fact.

PR #5614, opened 2026-01-19
Icells not derived 

gussmith23:
Duplicate of #4269 for testing purposes.


widlarizer:
I think there's two things that can be used to avoid PRs like this, even though I'm not sure if they're applicable here they might be convenient at some other point:
- to run CI you don't need to open a PR, if you want to run checks that don't run on all pushed commits to origin, you can request the checks. However I've never done this since github has an annoying interface and I didn't really need to
- to suggest a code change on somebody's fork, if they have PRs enabled, you can open a branch targeting their fork. Uh, this is annoying too because of github's interface, but I do it from time to time, specifically, targeting a fork branch from an origin branch requires manually editing the PR dialog URL

KrystalDelusion:
FYI that we generally prefer to rebase on main (fixing conflicts in the rebased commits they are relevant, and updating in place or with a follow up commit), rather than merge main into branches.

Also as a maintainer you should be able to edit (most) PRs yourself if you checkout the branch in question; it tells you on the page if "Maintainers are allowed to edit this pull request."  Installing `gh` makes this easy, just `gh pr checkout 4269` or whatever and it will automatically set the upstream location for fetching/pushing, or there are (e.g.) vscode plugins that can do it too.

gussmith23:
Ah, thanks to the both of you for this. 

> FYI that we generally prefer to rebase on main (fixing conflicts in the rebased commits they are relevant, and updating in place or with a follow up commit), rather than merge main into branches.

Will do in the future. Is this documented somewhere? Should it be?

Copy re: triggering CI -- I figured opening a PR wouldn't be the worst idea as I figured there's a chance I'd need it in PR form anyway.



> Also as a maintainer you should be able to edit (most) PRs yourself if you checkout the branch in question; it tells you on the page if "Maintainers are allowed to edit this pull request." Installing `gh` makes this easy, just `gh pr checkout 4269` or whatever and it will automatically set the upstream location for fetching/pushing, or there are (e.g.) vscode plugins that can do it too.

Ah, given that it was povik's own fork I assumed I couldn't push changes. I see the message re: being able to edit, but I had no idea this meant actually pushing commits to their fork. That's surprising but useful, thank you! I will rebase the old PR and push, and close this PR if and when that works.

gussmith23:
Closing for #4269, as I'm able to edit that PR directly.

PR #5615, opened 2026-01-19
Don't update `used_signals` for retained wires in `rmunused_module_signals`.

rocallahan:
These updates should not be necessary. In fact, if they were necessary, this code would be buggy, because the results would depend on the order in which wires are traversed: If wire A is retained, which causes an update to `used_signals`, which then causes wire B to be retained when it otherwise wouldn't be, then we would get different results depending on whether A is visited before B.

These updates will also make it difficult to process these wires in parallel.

### DIFF
--- a/passes/opt/opt_clean.cc
+++ b/passes/opt/opt_clean.cc
@@ -467,8 +467,6 @@ bool rmunused_module_signals(RTLIL::Module *module, bool purge_mode, bool verbos
 					wire->attributes.erase(ID::init);
 				else
 					wire->attributes.at(ID::init) = initval;
-				used_signals.add(new_conn.first);
-				used_signals.add(new_conn.second);
 				module->connect(new_conn);
 			}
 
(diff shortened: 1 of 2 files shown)


rocallahan:
FWIW this actually fixes a bug where `opt_clean` doesn't work properly sometimes. E.g. currently the testcase
```
module \test
  wire \wire_a
  wire \wire_f
  connect \wire_f \wire_a
end
```
after `opt_clean` becomes
```
module \test
  attribute \unused_bits "0"
  wire \wire_f
  attribute \unused_bits "0"
  wire \wire_a
  connect \wire_f \wire_a
end
```
... which seems wrong. In fact, running `opt_clean` again gives an empty module.

With this PR, a single `opt_clean` produces the empty module immediately.

widlarizer:
I agree that an opt pass when re-run should never have any more work to do as a general principle. This is not the first time I've noticed opt_clean behaving contrary to that

PR #5578, opened 2026-01-01
verific: add -optimization option

dhvll:
This PR adds a new `verific -optimization` option that enables Verific’s internal frontend optimizations during elaboration.

Silimate’s Yosys fork already uses this option to allow users to explicitly enable Verific post-elaboration and operator optimizations. Upstreaming this flag allows users to access the same functionality in a controlled, opt-in
manner without changing default behavior.

The change introduces a boolean `verific_opt` flag that is set when the `-optimization` option is passed to the `verific` frontend.

When enabled, Verific’s `PostElaborationProcess()` and `OperatorOptimization()` are invoked on each imported netlist. The behavior is fully guarded under `VERIFIC_SYSTEMVERILOG_SUPPORT` and does not alter existing behavior unless the flag is explicitly used.

#closes #5577

PR #5619, opened 2026-01-20
Add verific -set interface for Yosys-Verific settings

dhvll:
This PR introduces a verific -set command interface for managing Yosys-specific Verific options, similar to the existing verific -cfg for Verific runtime flags.

Implemented Features :
```
verific -set                    # List all settings with current/default values
verific -set <name>             # Show specific setting
verific -set <name> <value>     # Set a value
verific -set-reset              # Reset all to defaults
```
Usage Examples :
verific -set ignore_translate_off true
verific -set vlog_file_extensions ".v,.sv,.vh,.svh,.h,.inc"

ref: https://yosyshq.discourse.group/t/rfc-upstreaming-a-small-set-of-verific-frontend-options-from-silimate-fork/113

### DIFF
--- a/frontends/verific/verific.cc
+++ b/frontends/verific/verific.cc
@@ -114,6 +114,168 @@ int verific_sva_fsm_limit;
 vector<string> verific_incdirs, verific_libdirs, verific_libexts;
 #endif
 
+struct YosysVerificSettings {
+	enum class Type { BOOL, STRING, STRING_LIST };
+
+	struct Option {
+		Type type;
+		bool bool_value;
+		std::string string_value;
+		std::vector<std::string> string_list_value;
+		bool default_bool;
+		std::string default_string;
+		std::vector<std::string> default_string_list;
+		std::string description;
+		bool available; 
+	};
+
+	std::map<std::string, Option> options;
+
+	void init() {
+#ifdef VERIFIC_SYSTEMVERILOG_SUPPORT
+		// Bool option example: ignore_translate_off
+		// When true, translate_off/translate_on pragmas are ignored
+		{
+			Option opt;
+			opt.type = Type::BOOL;
+			opt.bool_value = false;
+			opt.default_bool = false;
+			opt.description = "Ignore translate_off/translate_on pragmas";
+			opt.available = true;
+			options["ignore_translate_off"] = opt;
+		}
+
+		// String-list option example: vlog_file_extensions
+		// Allows user to specify which file extensions should be treated as SystemVerilog
+		{
+			Option opt;
+			opt.type = Type::STRING_LIST;
+			opt.string_list_value = {".v", ".vh", ".sv", ".svh"};
+			opt.default_string_list = {".v", ".vh", ".sv", ".svh"};
+			opt.description = "Comma-separated Verilog/SystemVerilog file extensions";
+			opt.available = true;
+			options["vlog_file_extensions"] = opt;
+		}
+#endif
+	}
+
+	void reset() {
+		for (auto &it : options) {
+			auto &opt = it.second;
+			switch (opt.type) {
+				case Type::BOOL:
+					opt.bool_value = opt.default_bool;
+					break;
+				case Type::STRING:
+					opt.string_value = opt.default_string;
+					break;
+				case Type::STRING_LIST:
+					opt.string_list_value = opt.default_string_list;
+					break;
+			}
+		}
+	}
+
+	bool has_option(const std::string &name) const {
+		auto it = options.find(name);
+		return it != options.end() && it->second.available;
+	}
+
+	bool get_bool(const std::string &name) const {
+		auto it = options.find(name);
+		if (it == options.end() || it->second.type != Type::BOOL)
+			return false;
+		return it->second.bool_value;
+	}
+
+	std::string get_string(const std::string &name) const {
+		auto it = options.find(name);
+		if (it == options.end() || it->second.type != Type::STRING)
+			return "";
+		return it->second.string_value;
+	}
+
+	std::vector<std::string> get_string_list(const std::string &name) const {
+		auto it = options.find(name);
+		if (it == options.end() || it->second.type != Type::STRING_LIST)
+			return {};
+		return it->second.string_list_value;
+	}
+
+	void set_bool(const std::string &name, bool value) {
+		auto it = options.find(name);
+		if (it != options.end() && it->second.type == Type::BOOL)
+			it->second.bool_value = value;
+	}
+
+	void set_string(const std::string &name, const std::string &value) {
+		auto it = options.find(name);
+		if (it != options.end() && it->second.type == Type::STRING)
+			it->second.string_value = value;
+	}
+
+	void set_string_list(const std::string &name, const std::vector<std::string> &value) {
+		auto it = options.find(name);
+		if (it != options.end() && it->second.type == Type::STRING_LIST)
+			it->second.string_list_value = value;
+	}
+
+	// Apply a setting to Verific APIs when changed
+	void apply_setting(const std::string &name) {
+#ifdef VERIFIC_SYSTEMVERILOG_SUPPORT
+		if (name == "ignore_translate_off") {
+			veri_file::SetIgnoreTranslateOff(get_bool("ignore_translate_off") ? 1 : 0);
+			return;
+		}
+		if (name == "vlog_file_extensions") {
+			// Remove default .v extension and add user-specified ones
+			veri_file::RemoveFileExt(".v");
+			for (const auto &ext : get_string_list("vlog_file_extensions")) {
+				veri_file::AddFileExtMode(ext.c_str(), veri_file::SYSTEM_VERILOG);
+			}
+			return;
+		}
+#else
+		(void)name;
+#endif
+	}
+
+	// Parse comma-separated string into vector
+	static std::vector<std::string> parse_string_list(const std::string &value) {
+		std::vector<std::string> result;
+		size_t start = 0;
+		size_t end;
+		while ((end = value.find(',', start)) != std::string::npos) {
+			std::string item = value.substr(start, end - start);
+			// Trim whitespace
+			size_t first = item.find_first_not_of(" \t");
+			size_t last = item.find_last_not_of(" \t");
+			if (first != std::string::npos)
+				result.push_back(item.substr(first, last - first + 1));
+			start = end + 1;
+		}
+		std::string item = value.substr(start);
+		size_t first = item.find_first_not_of(" \t");
+		size_t last = item.find_last_not_of(" \t");
+		if (first != std::string::npos)
+			result.push_back(item.substr(first, last - first + 1));
+		return result;
+	}
+
+	// Format vector as comma-separated string
+	static std::string format_string_list(const std::vector<std::string> &list) {
+		std::string result;
+		for (size_t i = 0; i < list.size(); i++) {
+			if (i > 0) result += ",";
+			result += list[i];
+		}
+		return result;
+	}
+};
+
+static YosysVerificSettings yosys_verific_settings;
+static bool yosys_verific_settings_initialized = false;
+
 void msg_func(msg_type_t msg_type, const char *message_id, linefile_type linefile, const char *msg, va_list args)
 {
 	string message_prefix = stringf("VERIFIC-%s [%s] ",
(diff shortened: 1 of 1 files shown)


frontends/verific/verific.cc
```diff
@@ -114,6 +114,168 @@ int verific_sva_fsm_limit;
...
+	std::map<std::string, Option> options;
+
+	void init() {
+#ifdef VERIFIC_SYSTEMVERILOG_SUPPORT
+		// Bool option example: ignore_translate_off
+		// When true, translate_off/translate_on pragmas are ignored
+		{
+			Option opt;
+			opt.type = Type::BOOL;
+			opt.bool_value = false;
```
mmicko: think that defining `_value` is not needed, but better copy `_default` to `_value` in a loop for all options

frontends/verific/verific.cc
```diff
@@ -114,6 +114,168 @@ int verific_sva_fsm_limit;
...
+			opt.default_bool = false;
+			opt.description = "Ignore translate_off/translate_on pragmas";
+			opt.available = true;
```
mmicko: available flag is probably not needed, depending of ifdef section option would be added or not

frontends/verific/verific.cc
```diff
@@ -3345,6 +3507,34 @@ struct VerificPass : public Pass {
...
+		log("\n");
+		log("Get/set Yosys-Verific settings. These are Yosys-specific options that control\n");
+		log("how Verific integration behaves.\n");
+		log("\n");
+		log("Without arguments, lists all available settings and their current values.\n");
+		log("With one argument, shows the current value of the specified setting.\n");
+		log("With two arguments, sets the specified setting to the given value.\n");
+		log("\n");
+		log("Available settings:\n");
+#ifdef VERIFIC_SYSTEMVERILOG_SUPPORT
```
mmicko: idea of having meta data (including help message) in options is so you can just loop trough it and print it here.

frontends/verific/verific.cc
```diff
@@ -3345,6 +3507,34 @@ struct VerificPass : public Pass {
...
+		log("  ignore_translate_off   (bool)        Ignore translate_off/translate_on pragmas\n");
+		log("  vlog_file_extensions   (string-list) Verilog/SV file extensions for relaxed mode\n");
+#endif
+		log("\n");
+		log("For boolean settings, use 0/1, true/false, on/off, yes/no.\n");
+		log("For string-list settings, provide comma-separated values (e.g. \".v,.sv,.vh\").\n");
+		log("\n");
+		log("Example usage:\n");
```
mmicko: it is ok to remove example usage block

mmicko (changes requested):
Have to think if there is better way to store current/default data in general, but let's clean it first according to comments

frontends/verific/verific.cc
```diff
@@ -114,6 +114,163 @@ int verific_sva_fsm_limit;
...
+					opt.string_value = opt.default_string;
+					break;
+				case Type::STRING_LIST:
+					opt.string_list_value = opt.default_string_list;
+					break;
+			}
+		}
+	}
+
+	bool has_option(const std::string &name) const {
```
mmicko: just remove `has_option` completely and use `count` instead of it

frontends/verific/verific.cc
```diff
@@ -114,6 +114,163 @@ int verific_sva_fsm_limit;
 vector<string> verific_incdirs, verific_libdirs, verific_libexts;
 #endif
 
+struct YosysVerificSettings {
+	enum class Type { BOOL, STRING, STRING_LIST };
+
+	struct Option {
```
mmicko, line 139: Was thinking of this, how about using variant type ? :
Then you actually do not need a type then at all.
```
using OptionValue = std::variant<bool,std::string,std::vector<std::string>>;

struct Option {
	OptionValue value;
	OptionValue default_value;
	std::string description;
};
```

This should also make setters/getters to use templates and should simplify rest of code, like reset is only copy of default_value over value, also it gives chance of adding some more types (integers maybe for some settings) without affecting rest of code much

mmicko (changes requested):

frontends/verific/verific.cc
mmicko: Ah, sorry, maybe better keep type, that way we may prevent situations when we can not that easy determine desired type and also we can error out if wrong type is supplied.

dhvll: so should i just fix  >> just remove has_option completely and use count instead of it this changes right?

frontends/verific/verific.cc
mmicko: your latest change using only count is fine

dhvll: okay thanks
Few questions from my side.
1. Regarding my other open PRs should I close them or what should I do ? 
2. And do we need other features from those PRs?

mmicko: Those for options you have implemented here you can close marking that PRs are superseded by this one. For rest you can do the same when functionality of those are moved in this PR.  I think it is best to do that is when interface for `-set` is confirmed and then it would be a trivial job for you of moving it in here.

dhvll: okay thanks

dhvll:
hey @mmicko any update related to PR?

mmicko (approved):

mmicko:
@KrystalDelusion Would appreciate another pair of eyes, also on user/docs perspective not just code part.

frontends/verific/verific.cc
```diff
@@ -4303,6 +4486,105 @@ struct VerificPass : public Pass {
...
 		}
+
+		if (argidx < GetSize(args) && args[argidx] == "-set-reset")
+		{
+			yosys_verific_settings.reset();
+			log("All Yosys-Verific settings reset to defaults.\n");
+			goto check_error;
+		}
+
+		if (argidx < GetSize(args) && args[argidx] == "-set")
```
KrystalDelusion, line 4532: Being able to call `verific -set <opt1> <val1> -set <opt2> <val2>` might be a nice addition, but current behavior is consistent with `verific -cfg` so I don't think it's a requirement here.

frontends/verific/verific.cc
```diff
@@ -4303,6 +4486,105 @@ struct VerificPass : public Pass {
...
+			if (argidx+1 == GetSize(args)) {
+				log("Yosys-Verific settings:\n");
+				for (const auto &it : yosys_verific_settings.options) {
+					const auto &name = it.first;
+					const auto &opt = it.second;
+					switch (opt.type) {
+						case YosysVerificSettings::Type::BOOL:
+							log("  %s = %s (bool, default: %s)\n", name.c_str(),
+								opt.get<bool>() ? "true" : "false",
+								opt.get_default<bool>() ? "true" : "false");
```
KrystalDelusion: Inconsistent with `verific -cfg`, but I think the inclusion of a default is an improvement.

frontends/verific/verific.cc
```diff
@@ -3345,6 +3489,39 @@ struct VerificPass : public Pass {
...
+		log("    verific -set [<name> [<value>]]\n");
+		log("\n");
+		log("Get/set Yosys-Verific settings. These are Yosys-specific options that control\n");
+		log("how Verific integration behaves.\n");
+		log("\n");
+		log("Without arguments, lists all available settings and their current values.\n");
+		log("With one argument, shows the current value of the specified setting.\n");
+		log("With two arguments, sets the specified setting to the given value.\n");
+		log("\n");
+		log("Available settings:\n");
```
KrystalDelusion, line 3535: Do we have a sense of how many settings there will eventually be?  If it's under 10 I think this is fine, but more than that and we might be better off following the example of `verific -cfg` and only providing the list when called.

frontends/verific/verific.cc
```diff
@@ -4303,6 +4486,105 @@ struct VerificPass : public Pass {
...
+					log_cmd_error("Unknown Yosys-Verific setting '%s'.\n", name.c_str());
+				const auto &opt = yosys_verific_settings.options.at(name);
+				switch (opt.type) {
+					case YosysVerificSettings::Type::BOOL:
+						log("verific -set %s %s\n", name.c_str(),
+							opt.get<bool>() ? "true" : "false");
+						break;
+					case YosysVerificSettings::Type::STRING:
+						log("verific -set %s \"%s\"\n", name.c_str(),
+							opt.get_default<std::string>().c_str());
```
KrystalDelusion: I think this one is incorrect (it should be `opt.get` instead), but we don't have any `Type::STRING` options yet that would show it.

frontends/verific/verific.cc
```diff
@@ -114,6 +114,150 @@ int verific_sva_fsm_limit;
...
+	// Apply a setting to Verific APIs when changed
+	void apply_setting(const std::string &name) {
+#ifdef VERIFIC_SYSTEMVERILOG_SUPPORT
+		if (name == "ignore_translate_off") {
+			veri_file::SetIgnoreTranslateOff(get_option<bool>("ignore_translate_off") ? 1 : 0);
+			return;
+		}
+		if (name == "vlog_file_extensions") {
+			// Remove default .v extension and add user-specified ones
+			veri_file::RemoveFileExt(".v");
```
KrystalDelusion: Why do we remove `".v"` here before adding the user ones?  There is also a problem here where if we set (and apply) the setting, and then set it again we don't remove the previous extensions (nor does resetting it), requiring a full restart of Yosys to be able to remove an extension from being parsed as SystemVerilog.

frontends/verific/verific.cc
```diff
@@ -114,6 +114,150 @@ int verific_sva_fsm_limit;
...
+			options["ignore_translate_off"] = opt;
+		}
+
+		// String-list option example: vlog_file_extensions
+		// Allows user to specify which file extensions should be treated as SystemVerilog
+		{
+			Option opt;
+			opt.type = Type::STRING_LIST;
+			opt.default_set<std::vector<std::string>>({".v", ".vh", ".sv", ".svh"});
+			opt.description = "Comma-separated Verilog/SystemVerilog file extensions";
```
KrystalDelusion: I think we can just say SystemVerilog instead of Verilog/SystemVerilog since this is specifically setting them as being SystemVerilog. The other places where we use the phrasing of Verilog/SystemVerilog are where it applies to both.

KrystalDelusion (changes requested):
I would like the question about `RemoveFileExt` to be addressed before merging.  The rest are minor/non-blocking.

mmicko:
@dhvll just to let you know that I will be pushing some commits on your PR branch (that is a feature available on GitHub) to do some code changes. Will notify when I am done, so you can pull on your side as well. 
Will probably need to change when applying is performed, since `veri_file` and `vhdl_file` are reseted after import as well. Let's make these options work first and then we can see on adding other features if those make sense.

Also a question, do you have access to Verific library ? Asking since code from some of PRs did not compile do to wrong name of function, but also on one place it required Verific with special features enabled.

dhvll:
> @dhvll just to let you know that I will be pushing some commits on your PR branch (that is a feature available on GitHub) to do some code changes. Will notify when I am done, so you can pull on your side as well. Will probably need to change when applying is performed, since `veri_file` and `vhdl_file` are reseted after import as well. Let's make these options work first and then we can see on adding other features if those make sense.
> 
> Also a question, do you have access to Verific library ? Asking since code from some of PRs did not compile do to wrong name of function, but also on one place it required Verific with special features enabled.

I have pull the changes.
Regarding Verific access. I don't have direct access to the Verific library. The API calls (like SetIgnoreTranslateOff, RemoveFileExt, AddFileExtMode) were referenced from the existing codebase and Verific documentation available to me, but I haven't been able to compile-test against the actual library. Apologies for the incorrect function names happy to fix anything that doesn't compile once you identify the issues.

dhvll:
@mmicko what other changes should I focus on now ?

mmicko:
@dhvll Hi. Sorry I did not have much time for this PR. Thing that is left is when to apply certain settings and that is quite dependant from use case to use case. Also some of the settings would be lost after import of design, and it is not always desired way to do. But you having no access to Verific library and not able to test does not help on this. Will see to take care of this next week.

dhvll:
hey @mmicko any update from your side?

mmicko:
Closing related PRs. We still need some more explanations of usefulness of new  options 
https://github.com/YosysHQ/yosys/pull/5581
https://github.com/YosysHQ/yosys/pull/5583
https://github.com/YosysHQ/yosys/pull/5584
https://github.com/YosysHQ/yosys/pull/5585
https://github.com/YosysHQ/yosys/pull/5586 
https://github.com/YosysHQ/yosys/pull/5578

PR #5621, opened 2026-01-21
Remove `Design::sort()` calls from optimization passes

rocallahan:

These `sort()` calls in `opt_clean` and `opt` should be unnecessary, and they can have noticeable performance impact for very large designs. In a large example circuit I have with a flattened module of 3.5M cells, `Design::sort()` takes several seconds, and in typical flows `opt_clean` runs many times.

Any analysis or optimization that depends on objects being sorted by name or by IdString value is inherently fragile. In the former case, that means renaming objects could change the results. In the latter case, changing the IdString internal indices could change the results. As far as I know, the only reason why sorting might be needed for correctness is if the rest of the pass is nondeterministic and sorting restores determinism. But that is not the case for `opt_clean` as far as I can tell. For `opt`, if sorting was needed to fix nondeterminism in some subpass, the sorting should be done by that subpass.

Removing these calls causes two test failures. In `gowin/mux.ys` the `mux8` test starts failing. If I split that test out into its own file, it stops failing, so this test is inherently fragile. In this PR I split `gowin/mux.ys` into separate files to avoid the problem.

In `xilinx/dsp_cascade.ys` the "mno" `xc6s` test starts failing with 10 `FDRE`s instead of 5. I don't really know what's going on here, but as I discussed above I don't think anything should depend on sorting for correctness, so I've just adjusted the test expectation.

### DIFF
--- a/docs/source/code_examples/macro_commands/prep.ys
+++ b/docs/source/code_examples/macro_commands/prep.ys
@@ -17,6 +17,7 @@ coarse:
     opt_clean
     memory_collect
     opt -noff -keepdc -fast
+    sort
 
 check:
     stat
--- a/passes/opt/opt.cc
+++ b/passes/opt/opt.cc
@@ -193,7 +193,6 @@ struct OptPass : public Pass {
 		}
 
 		design->optimize();
-		design->sort();
 		design->check();
 
 		log_header(design, "Finished fast OPT passes.%s\n", fast_mode ? "" : " (There is nothing left to do.)");
--- a/passes/opt/opt_clean.cc
+++ b/passes/opt/opt_clean.cc
@@ -715,7 +715,6 @@ struct OptCleanPass : public Pass {
 			log("Removed %d unused cells and %d unused wires.\n", count_rm_cells, count_rm_wires);
 
 		design->optimize();
-		design->sort();
 		design->check();
 
 		keep_cache.reset();
@@ -780,7 +779,6 @@ struct CleanPass : public Pass {
 			log("Removed %d unused cells and %d unused wires.\n", count_rm_cells, count_rm_wires);
 
 		design->optimize();
-		design->sort();
 		design->check();
 
 		keep_cache.reset();
--- a/techlibs/common/prep.cc
+++ b/techlibs/common/prep.cc
@@ -211,6 +211,7 @@ struct PrepPass : public ScriptPass
 				run("memory_collect");
 			}
 			run(nokeepdc ? "opt -noff -fast" : "opt -noff -keepdc -fast");
+			run("sort");
 		}
 
 		if (check_label("check"))
--- a/techlibs/gowin/synth_gowin.cc
+++ b/techlibs/gowin/synth_gowin.cc
@@ -311,6 +311,7 @@ struct SynthGowinPass : public ScriptPass
 
 		if (check_label("map_luts"))
 		{
+			run("sort");
 			if (nowidelut && abc9) {
 				run("read_verilog -icells -lib -specify +/abc9_model.v");
 				run("abc9 -maxlut 4 -W 500");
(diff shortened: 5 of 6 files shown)


Ravenslofty:
> Removing these calls causes two test failures. In `gowin/mux.ys` the `mux8` test starts failing. If I split that test out into its own file, it stops failing, so this test is inherently fragile. In this PR I split `gowin/mux.ys` into separate files to avoid the problem.

"re-blessing" tests when changes occur is *fine*, but blaming the test feels like the wrong approach: can you be more specific with "starts failing"?

widlarizer:
There is #5362 already for this, but I apologize, I constantly had other things to focus on. I think this PR takes more care of the FPGA tests by splitting them up which should also help with solver runtimes being annoyingly unpredictable, so I'm happy to close mine in favor of this one.

Generally, re-blessing is bad for maintainability because person dealing with unrelated changes should not be required to have the clerical authority to do so. If there's chaos it should be re-jiggleable with `--hash-seed` or something and should be discovered by occasional tests of such stability in the test suite and removed/fixed. We shouldn't try to pin down something that is supposed to be okay if it changes because it feels like it

widlarizer:
I'm in favor of never forcing sorting, but I introduced the `sort` command earlier that can be added so that the user choice to have stable results is respected. Maybe we can even resolve the test failures with the `sort` command? There's a more general discussion to be had about the tradeoff of unstable tests and how our QA is done. Let's not block the PR on things like that, since re-blessing the FPGA test numbers for now doesn't actually change anything from the status quo of our test suite

rocallahan:
Regarding performance, FYI I have developed a parallel implementation of `opt_clean`. With that implementation, on my 3.5M cell circuit the `Design::sort()` alone is more than half of the runtime of `opt_clean`, which is unfortunate. I could parallelize the `sort()` itself, but that feels like unnecessary work and it is not easy to do it in a scalable way.

rocallahan:
There's a test failure in CI that I need to look into:
```
ERROR: The design contains an undriven signal $flatten\picorv32_i.$auto$wreduce.cc:514:run$3801. This is not supported by the functional backend. Call setundef with appropriate options to avoid this error.
=========================== short test summary info ============================
FAILED test_functional.py::test_cxx[picorv] - AssertionError: /work/_work/yosys/yosys/yosys failed
assert 1 == 0
 +  where 1 = CompletedProcess(args=[PosixPath('/work/_work/yosys/yosys/yosys'), '-Q', '-p', 'read_rtlil /tmp/pytest-of-user/pytest-1/test_cxx_picorv_0/rtlil.il ; clk2fflogic ; write_functional_cxx /tmp/pytest-of-user/pytest-1/test_cxx_picorv_0/my_module_functional_cxx.cc'], returncode=1).returncode
```

Instead of changing the tests, we could just move the `Design::sort()` calls into the passes that apparently depend on the design being sorted to produce those expected results. I still think those passes (or their tests) are buggy, but at least the workaround will be closer to the buggy code.

Ravenslofty:
> I still think those passes (or their tests) are buggy

One way I know where order dependence manifests is in `write_xaiger` where the netlist is output in topological order, but this results in ordering differences within a topological level; because ABC9 then maps LUTs structurally, this results in a changed mapping. `synth_gowin` uses ABC9, so it will exercise this code path, and result in chaos.

I'm not against your suggestion of manual sorting like that, but you will need to add it to all the `synth_*` passes to preserve current behaviour.

A different proposal is to make sorting opt-in, that FPGA flows can select the existing behaviour; I suspect ASIC flows would care less about it.

rocallahan:
I've updated the PR to remove the `sort()` calls from `opt` and `opt_clean` and add them to `prep`, `synth_gowin` and `synth_xilinx`. This passes tests for me locally and hopefully CI will agree. Maybe more `sort` calls should be added elsewhere but without failing tests to guide me, I'm not sure where exactly to add them. Any suggestions on how to proceed?

PR #5623, opened 2026-01-21
opt_dff restructure.

Goubermouche:
Restructure opt_dff pass for better clarity, add some concise documentation.

### DIFF
--- a/tests/unit/opt/optDffFindComplementaryPatternTest.cc
+++ b/tests/unit/opt/optDffFindComplementaryPatternTest.cc
@@ -1,5 +1,5 @@
 #include <gtest/gtest.h>
-#include "passes/opt/opt_dff_comp.h"
+#include "kernel/pattern.h"
 
 YOSYS_NAMESPACE_BEGIN
 


widlarizer:
I suggest using draft PRs for WIPs to signal the change isn't ready to be merged yet

hexratcc:
I've added some of the changes we've discussed on our 1on1, including the unit tests for the find comp lambda.

passes/opt/opt_dff_comp.h
```diff
@@ -0,0 +1,31 @@
+#ifndef OPT_DFF_COMP_H
+#define OPT_DFF_COMP_H
+
+#include "kernel/rtlil.h"
+#include <map>
+#include <optional>
+
+YOSYS_NAMESPACE_BEGIN
+
+typedef std::map<RTLIL::SigBit, bool> pattern_t;
```
widlarizer: This file should be documented to its algorithmic functionality similarly to `kernel/bitpattern.h` and moved to `kernel/` as well since it implements a general purpose data structure for logic synthesis

passes/opt/opt_dff_comp.h
```diff
@@ -0,0 +1,31 @@
...
+
+inline std::optional<RTLIL::SigBit> find_complementary_pattern_var(
+	const pattern_t& left,
+	const pattern_t& right
+) {
+	std::optional<RTLIL::SigBit> ret;
+	for (const auto &pt : left) {
+		if (right.count(pt.first) == 0)
```
widlarizer: Please describe in which case which return is hit in terms of the information the patterns represent

passes/opt/opt_dff.cc
```diff
@@ -173,31 +224,16 @@ struct OptDffWorker
 	void simplify_patterns(patterns_t& patterns)
```
widlarizer: Should be moved to the header / separate source file, it doesn't depend on `OptDffWorker` so it shouldn't be a problem

passes/opt/opt_dff.cc
```diff
@@ -173,31 +224,16 @@ struct OptDffWorker
...
-				else
-					if (ret)
-						return {};
-					else
-						ret = pt.first;
-			return ret;
-		};
 
-		// remove complimentary patterns
+		// Remove complimentary patterns
```
widlarizer: Please replace complimentary to complementary (multiple inherited occurrences in this file)

PR #5625, opened 2026-01-21
VCD Writer Bug Fix

stanminlee:
The VCD writer does not follow the IEEE format for VCD files for the time_number for $timescale

To do this, it is very simple, just add a "1" to conform to the standard. This is okay since the number can be a 1|10|100

Please reference IEEE 1800-2017 page 657 as evidence for this bug.

### DIFF
--- a/passes/sat/sim.cc
+++ b/passes/sat/sim.cc
@@ -2329,7 +2329,7 @@ struct VCDWriter : public OutputWriter
 		}
 
 		if (!worker->timescale.empty())
-			vcdfile << stringf("$timescale %s $end\n", worker->timescale);
+			vcdfile << stringf("$timescale 1%s $end\n", worker->timescale);
 
 		worker->top->write_output_header(
 			[this](IdString name) { vcdfile << stringf("$scope module %s $end\n", log_id(name)); },


mmicko (changes requested):
In this case it would be then expected from user to provide `-timescale ns` for example, but then it would generate `$timescale 1ns $end` (so number would be directly concatenated to unit.
Have still to check with team if there is any user of this option already there, but by my opinion is that it would be better to actually expect 2 parameters for this option, check if they are vaild (first must be number, 2nd must be valid unit) and generate `worker.timescale` based on those two values

KrystalDelusion:
It looks like this is already being done for some parts of the code (fst2tb)
https://github.com/YosysHQ/yosys/blob/1cd3fd50905c3359054e7f96d36184dd5b53ac1b/passes/sat/sim.cc#L2205

mmicko:
@KrystalDelusion @stanminlee Sorry all, my bad. This line is forming a verilog output, but anyway in VCD number and unit is concatenated, so adding parameter `-timescale 10ns` should give you already good output. 
By doing change from this PR we are limiting timescale to provide only unit, but with current behavior we can specify even more `-timescale 1ns`  or `-timescale 10ms`

KrystalDelusion:
> By doing change from this PR we are limiting timescale to provide only unit, but with current behavior we can specify even more `-timescale 1ns` or `-timescale 10ms`

That was my initial thought too and I wondered if this was an incorrect use of the option rather than a bug, is there anything that suggests it should just be the unit?

mmicko:
Timescale option is made so user provide number and unit at once, as it is anyway used by VCD file, so no further changes are needed for this to work.

PR #5628, opened 2026-01-23
Add `linux_perf` command to turn Linux perf recording on and off.

rocallahan:
This is extremely useful for profiling specific passes.

### DIFF
--- a/passes/cmds/Makefile.inc
+++ b/passes/cmds/Makefile.inc
@@ -37,6 +37,7 @@ OBJS += passes/cmds/chformal.o
 OBJS += passes/cmds/chtype.o
 OBJS += passes/cmds/blackbox.o
 OBJS += passes/cmds/ltp.o
+OBJS += passes/cmds/linux_perf.o
 ifeq ($(DISABLE_SPAWN),0)
 OBJS += passes/cmds/bugpoint.o
 endif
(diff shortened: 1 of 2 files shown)


widlarizer:
Though if I use the exact invocation in the help message I get perf record stuck in a loop of addr2line timing out. `perf record` doesn't allow me to provide a faster addr2line than the GNU one, to my knowledge. Relying on function pointers with `--call-graph fp` works but is best combined with adding `-fno-omit-frame-pointer` of course

KrystalDelusion:
The `linux_perf` command is currently under the ["Design modification" commands group](https://yosyshq.readthedocs.io/projects/yosys/en/latest/cmd/index_passes_cmds.html#linux-perf-turn-linux-perf-recording-off-or-on), I would probably suggest moving it to internal commands for developers by adding `internal();` to its constructor.  Also the formatting is broken, which might be fixed by indenting it, or updating it to use the `Pass::formatted_help()` method (which is still only [documented in a preview build](https://yosyshq.readthedocs.io/projects/yosys/en/docs-preview-contributing_help/yosys_internals/extending_yosys/documenting.html) >.>) and putting it in a `codeblock`.

rocallahan:
> Though if I use the exact invocation in the help message I get perf record stuck in a loop of addr2line timing out. `perf record` doesn't allow me to provide a faster addr2line than the GNU one, to my knowledge.

Yeah, it's a huge pain. I build my own `perf` from `perf-tools-next` with `BUILD_NONDISTRO=1` in `tools/perf/Makefile.perf`.

PR #5630, opened 2026-01-23
ast: Add support for array-to-array assignment

apullin:
This commit adds support for SystemVerilog array-to-array assignment operations that were previously unsupported:

1. Direct array assignment: `b = a;`
2. Array ternary expressions: `out = sel ? a : b;`

Both single-dimensional and multi-dimensional unpacked arrays are supported. The implementation expands these array operations during AST simplification into element-wise assignments.

Example of now-supported syntax:
```systemverilog
wire [7:0] state_regs[8];
wire [7:0] r[8];
wire [7:0] sel[8];
assign sel = condition ? state_regs : r;
```
I was generating some verilog with Google XLS, and this pattern came up, and I found that it was not supported by Yosys. 

Tests are added in `tests/svtypes/array_assign.sv`.                                                                                                                                         
                                                                                                                                                                                              
  This collides with PR #2331, but that PR has been up for > 5 years. This implementation also addresses the review feedback from that PR, including support for array ternary expressions    
  (`out = sel ? a : b`) and multi-dimensional unpacked arrays.

### DIFF
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -269,6 +269,83 @@ static int add_dimension(AstNode *node, AstNode *rnode)
 	node->input_error("Unpacked array in packed struct/union member %s\n", node->str);
 }
 
+// Check if node is an unexpanded array reference (AST_IDENTIFIER -> AST_MEMORY without indexing)
+static bool is_unexpanded_array_ref(AstNode *node)
+{
+	if (node->type != AST_IDENTIFIER)
+		return false;
+	if (node->id2ast == nullptr || node->id2ast->type != AST_MEMORY)
+		return false;
+	// No indexing children = whole array reference
+	return node->children.empty();
+}
+
+// Check if two memories have compatible unpacked dimensions for array assignment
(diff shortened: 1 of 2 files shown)


apullin:
> quite a lot

AFAIK, RTLIL just has to be a ton of entries, wires, ports - there's no bulk memory copy primitive, so combinational array assignment fundamentally requires N parallel paths.

But large code gen is an issue: arrays of ~100K elements take ~3s, and ~1M elements can take 60s+.

I added a warning for arrays over 10,000 elements.

whitequark:
I'm happy with a warning, and I don't think you can do better here, either.

apullin:
@widlarizer bump

apullin:
- Added the requested test case. Sure enough, it failed. Resolved by mapping array assignment expansion by element position (range-direction aware), instead of reusing raw numeric indices across LHS/RHS.
- Also added the procedural ternary array-assignment case.

apullin:
Yay. Can probably mark #2331 as superseded ?

PR #5638, opened 2026-01-26
linux_perf: mark internal, fix help formatting

widlarizer:
See [this comment](https://github.com/YosysHQ/yosys/pull/5628#issuecomment-3801211754) for context


KrystalDelusion:
I switched it over to use `formatted_help`, which also fixes the paragraph lines being too long.  I did just notice the codeblock is still overlength though.  Feel free to force push and override if you disagree with any of my changes @widlarizer.  I did have to sneak in the `log_help.cc` fix for codeblock formatting, but the example shell call is now formatted nicely on both command line and web.

widlarizer:
Thanks!

PR #5645, opened 2026-01-28
Upstream verific mixed sv vhdl

nataliakokoromyti:
Main currently treats -f as Verilog only not fully covering when you have mixed Verilog and VHDL source lists. This change enables mixed -f inputs while preserving the existing SystemVerilog-only behavior if no VHDL is present.

### DIFF
--- a/frontends/verific/verific.cc
+++ b/frontends/verific/verific.cc
@@ -3747,12 +3747,10 @@ struct VerificPass : public Pass {
 #ifdef VERIFIC_VHDL_SUPPORT
 			int i;
 			Array *file_names_sv = new Array(POINTER_HASH);
-			bool has_vhdl = false;
 			FOREACH_ARRAY_ITEM(file_names, i, filename) {
 				std::string filename_str = filename;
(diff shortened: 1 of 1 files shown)


frontends/verific/verific.cc
```diff
@@ -3744,10 +3744,42 @@ struct VerificPass : public Pass {
...
+					vhdl_file::SetDefaultLibraryPath((proc_share_dirname() + "verific/vhdl_vdbs_2019").c_str());
+					if (!vhdl_file::Analyze(filename, work.c_str(), vhdl_file::VHDL_2019)) {
+						verific_error_msg.clear();
+						log_cmd_error("Reading VHDL sources failed.\n");
+					}
+				} else {
+					file_names_sv->Insert(strdup(filename));
+				}
+			}
+			if (has_vhdl) {
```
mmicko: Why can't you use `AnalyzeMultipleFiles` for both cases ? Array used is the same and it contains only verilog files

mmicko (approved):
Thanks, this is now fine. Will see to add support to select VHDL standard, but this is good as is for now.

PR #5646, opened 2026-01-28
Dump module details when `design_equal` fails

rocallahan:
When `design_equal` detects that the designs are not equal, dump the offending modules in RTLIL if possible. This makes it much quicker and easier to understand the issue.


widlarizer (approved):
Ok with me

PR #5648, opened 2026-01-29
verific: fixed -sv2017 option and added ability to set VHDL standard

mmicko:
This one fixes option missed in https://github.com/YosysHQ/yosys/pull/5612 and also gives ability of setting used VHDL standard if there are such file imported as addition to https://github.com/YosysHQ/yosys/pull/5645


mmicko:
@nataliakokoromyti Please take a look if this is fine for your workflow. Default is VHDL 2008 since that is default when option -vhdl is used in other parts of verific.cc

nataliakokoromyti:
looks good to me, thanks!

PR #5649, opened 2026-01-29
Upstream rtlil dump hash

nataliakokoromyti:
Needed a quick way to get a module’s RTLIL as a string and a simple hash of it, so downstream code can compare modules without writing files.

Added two helpers on RTLIL::Module:
-rtlil_dump() sorts the module and returns the RTLIL dump as a string.
-rtlil_hash() hashes that string with std::hash<std::string> and returns an unsigned int.

### DIFF
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -29,8 +29,6 @@
 #include <string.h>
 #include <algorithm>
 #include <charconv>
-#include <codecvt>
-#include <locale>
 #include <optional>
 #include <string_view>
 #include <sstream>


widlarizer:
Unnecessary. You can dump a module with `RTLIL_BACKEND::dump_module` which takes an `ostream` so you can use [std::ostringstream](https://cplusplus.com/reference/sstream/ostringstream/) to avoid writing a file. I see no reason to hash the textual output either, if you actually need that, you can handle that part in your own code. And sorting in a dump function is actively dangerous

widlarizer:
If this is a Python related thing like in #5609 the same applies (C++ interfaces shouldn't be polluted by helpers that wrap the ostream for Python)

PR #5651, opened 2026-01-29
Handle ABC nonfatal "Error:" messages

rocallahan:
PR #5593 treats all ABC output starting with 'Error: ' as a fatal error. Unfortunately in ABC not all such messages indicate fatal errors so we don't want to always stop processing there.

So instead look for `** cmd error: aborting` in ABC's stderr. This message has been stable since 2005 so I think it's OK to rely on.

### DIFF
--- a/passes/techmap/abc.cc
+++ b/passes/techmap/abc.cc
@@ -188,10 +188,10 @@ struct AbcProcess
 		int status;
 		int ret = waitpid(pid, &status, 0);
 		if (ret != pid) {
-			log_error("waitpid(%d) failed", pid);
+			log_error("waitpid(%d) failed\n", pid);
 		}
 		if (!WIFEXITED(status) || WEXITSTATUS(status) != 0) {
-			log_error("ABC failed with status %X", status);
+			log_error("ABC failed with status %X\n", status);
 		}
 		if (from_child_pipe >= 0)
 			close(from_child_pipe);
@@ -203,12 +203,12 @@ std::optional<AbcProcess> spawn_abc(const char* abc_exe, DeferredLogs &logs) {
 	// fork()s.
 	int to_child_pipe[2];
 	if (pipe2(to_child_pipe, O_CLOEXEC) != 0) {
-		logs.log_error("pipe failed");
+		logs.log_error("pipe failed\n");
 		return std::nullopt;
 	}
 	int from_child_pipe[2];
 	if (pipe2(from_child_pipe, O_CLOEXEC) != 0) {
-		logs.log_error("pipe failed");
+		logs.log_error("pipe failed\n");
 		return std::nullopt;
 	}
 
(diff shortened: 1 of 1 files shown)


tests/techmap/bug5574.ys
```diff
@@ -1,6 +1,6 @@
 # On Linux, with a spawned abc, this message is the error
 # otherwise the error is the failure to load the output.blif
-logger -expect log "ABC: Error: This command can only be applied to an AIG" 1
+logger -expect log "ABC script failed" 1
```
KrystalDelusion: I suspect this message is specific to Linux with a spawned ABC and is why the tests are failing on macOS (to test locally you can build with `LINK_ABC=1`).  But also, the point of checking for the message that it was checking for is to ensure that the ABC error is printed.  I think logging "ABC script failed" without capturing the actual error is a downgrade and not something that should be merged.

KrystalDelusion (changes requested):
Not directly related to this PR (so feel free to address separately), but it looks like in #5266 you introduced a change to use two different tempdirs for ABC (`run_abc.tempdir_name` and `config.global_tempdir_name`); the current output filtering only replaces one of these directories with `<abc-temp-dir>` not both (I think it's the global tempdir name that's slipping through).  Could you please look into that as well?

tests/techmap/bug5574.ys
KrystalDelusion: Actually why does this error message get lost?  With `LINK_ABC=1` the cmd error is after the AIG error, and the modified code should still print `Error:` lines if it hasn't aborted yet

rocallahan:
> Could you please look into that as well?

Sure, I'll do that separately.

tests/techmap/bug5574.ys
rocallahan: With a spawned ABC on Linux, ABC prints `** cmd error: aborting` to stderr first and only then flushes stdout to send `Error: This command can only be applied to an AIG`. This PR currently loses whatever was buffered in stdout. But I have another idea...

rocallahan: Alright, I've addressed this by having the spawned-ABC code only stop processing ABC output when we've seen both the `Error:` message and the `** cmd error:` message` (whichever order they come in).

rocallahan: Hmm, mixing stdout and stderr on the same fd is probably not a good idea since strings like "Error:" could be split in the output if the buffering boundary lands on exactly the wrong place :-(.

rocallahan: Ok I've updated this. Bigger comment below.

rocallahan:
Ok, I've taken a different approach to detecting the end of the script run. I look at the ABC prompts (text of the form `abc NNN> ` where NNN is a number with at least two digits). First we expect to see an ABC prompt followed by "source ..." which is the sourcing of Yosys' controlling ABC script. When we see the next ABC prompt after that, the script must have finished one way or another.

KrystalDelusion (approved):
LGTM, thanks!

Logikable:
Thanks for this! Are we able to get this merged? We're looking to move downstream closer to head.

PR #5653, opened 2026-01-30
Upstreaming abc module parallelism

nataliakokoromyti:
The abc pass runs module snippets through ABC one at a time. We want to run those ABC jobs in parallel and control how many run at once. This change parallelizes ABC runs over BLIF snippets using a process pool and the user can specify how many processes should be in the pool. 

It splits the non‑DFF/clock‑domain flow into stages and runs the ABC step with a worker pool. Adds -max_threads and -reserved_cores (and defaults abc.max_threads / abc.reserved_cores) to control parallelism.

Adds tests/techmap/abc_parallel.ys.

### DIFF
--- a/passes/techmap/abc.cc
+++ b/passes/techmap/abc.cc
@@ -143,6 +143,8 @@ struct AbcConfig
 	bool markgroups = false;
 	pool<std::string> enabled_gates;
 	bool cmos_cost = false;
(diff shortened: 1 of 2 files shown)


widlarizer:
Superseded by https://github.com/YosysHQ/yosys/pull/6194

Issue #5654, opened 2026-01-30
sat: dff output changes without clock edge


widlarizer:
### Version

main

### On which OS did this happen?

Linux

### Reproduction Steps

```
read_verilog -sv -icells <<EOT
module top(input C, D, output Q);
$_DFF_P_ ff (.C(C), .D(D), .Q(Q));
assume property (~C);
assert property (Q !== 1'bx);
endmodule
EOT
stat
chformal -lower
stat
sat -show-all -enable_undef -set-assumes -set-init-def -verify -prove-asserts -seq 2
```
Found when fixing `tests/techmap/dfflibmap_formal.ys`

### Expected Behavior

QED

### Actual Behavior

When D goes from def to undef, Q goes from def to undef, despite C not changing.
```
  Time Signal Name                             Dec       Hex           Bin
  ---- ------------------------------- ----------- --------- -------------
  init \Q                                        0         0             0
  ---- ------------------------------- ----------- --------- -------------
     1 $auto$rtlil.cc:3337:Not$12                0         0             0
     1 $auto$rtlil.cc:3337:Not$7                 0         0             0
     1 $auto$rtlil.cc:3384:And$14                0         0             0
     1 $auto$rtlil.cc:3384:And$9                 0         0             0
     1 $nex$<<EOT:4$4_Y                          1         1             1
     1 $not$<<EOT:3$2_Y                          1         1             1
     1 \C                                        0         0             0
     1 \D                                       --        --             x
     1 \Q                                        0         0             0
  ---- ------------------------------- ----------- --------- -------------
     2 $auto$rtlil.cc:3337:Not$12                0         0             0
     2 $auto$rtlil.cc:3337:Not$7                 1         1             1
     2 $auto$rtlil.cc:3384:And$14                0         0             0
     2 $auto$rtlil.cc:3384:And$9                 1         1             1
     2 $nex$<<EOT:4$4_Y                          0         0             0
     2 $not$<<EOT:3$2_Y                          1         1             1
     2 \C                                        0         0             0
     2 \D                                        0         0             0
     2 \Q                                       --        --             x
```

widlarizer:
Without modeling undef, DFF's output can spontaneously update without a clock edge:
```
read_verilog -sv -icells <<EOT
module top(input C, D, output Q);
$_DFF_P_ ff (.C(C), .D(D), .Q(Q));
assume property (~C);
assert property (Q !== 1'b1);
endmodule
EOT
chformal -lower
[...]

PR #5655, opened 2026-01-30
Undefine set&reset behavior of $dffsr

### LINKED ISSUE #5194: $dffsr semantics as implemented differ from models and docs

widlarizer:
Fixes #5194

### DIFF
--- a/tests/techmap/dfflibmap_proc_formal.ys
+++ b/tests/techmap/dfflibmap_proc_formal.ys
@@ -78,7 +78,7 @@ delete top miter
 copy top_unmapped top
 dfflibmap -liberty dfflibmap_dffsr_mixedpol.lib top
 
-async2sync
+clk2fflogic
 flatten
 miter -equiv -make_assert -flatten top_unmapped top miter
 
@@ -92,7 +92,7 @@ delete top miter
 copy top_unmapped top
 dfflibmap -liberty dfflibmap_dffsr_not_next.lib top
 
-async2sync
+clk2fflogic
 flatten
 miter -equiv -make_assert -flatten top_unmapped top miter
 


frontends/liberty/liberty.cc
```diff
@@ -220,10 +225,24 @@ static void create_ff(RTLIL::Module *module, const LibertyAst *node)
...
+					log_error("Unexpected length of clear_preset_var* value %s in FF cell %s\n", child->value, name);
+				*var = child->value[0];
+			}
+
+	}
+	if (clear_preset_var1 == 'X' || clear_preset_var2 == 'X') {
+		if (!clear_preset_reported) {
+			log_warning("FF cell %s has well-defined clear&preset behavior, but Yosys models it as undefined\n", name);
+			clear_preset_reported = true;
+		}
```
gussmith23: I'm confused here -- it isn't clear to me why the condition `clear_preset_var1 == 'X' || clear_preset_var2 == 'X'` implies that the FF cell has well-defined clear&preset behavior. I would think that "well-defined clear&preset behavior" means that:
1. At least one of `clear_preset_var*` is set to something (because they can be unset, right?)
2. AND at least one of them is not set to `X` (because if they're both `X`, then they match Yosys semantics, if I understand correctly based on the "Yosys models it as undefined")

So that would be something like

`(clear_preset_var1 && *clear_preset_var1 != 'X') || (clear_preset_var2 && *clear_preset_var2 != 'X')`

This is pretty different from what you have, so I am assuming I'm misunderstanding something.

To be clear -- I also don't know what it means for both of `clear_preset_var*` to be unset on the Liberty side. It's surprising that specifying this behavior would be optional!

frontends/liberty/liberty.cc
```diff
@@ -248,36 +267,64 @@ static void create_ff(RTLIL::Module *module, const LibertyAst *node)
 		}
 	}
 
-	RTLIL::Cell *cell = module->addCell(NEW_ID, ID($_NOT_));
-	cell->setPort(ID::A, iq_sig);
-	cell->setPort(ID::Y, iqn_sig);
+	for (auto& [out_sig, cp_var, neg] : {tuple{iq_sig, clear_preset_var1, false}, {iqn_sig, clear_preset_var2, true}}) {
```
gussmith23, line 263: I understand what's happening here in the following way: it's no longer the case that you can always model Q and Qn as inverses, because `clear_preset_var*` may specify non-inverse behavior for the two. 

However, I suspect they'll still be inverses in the common case. Is that not true?

If that's the case, how much overhead (of any kind) are we incurring by _always_ duplicating these cells? Is it worth detecting when they really are inverses and deduplicating? (Is deduplication even possible?)

"Overhead" could be any number of things -- design size, verification runtime, synthesis QoR. I'm not actually sure which flows this is most likely to affect.

widlarizer:
> Also the [memory mapping docs provide verilog code for undefined collision behavior](https://yosyshq.readthedocs.io/projects/yosys/en/latest/using_yosys/synthesis/memory.html#synchronous-sdp-with-undefined-collision-behavior), is there a reason we can't do the same with our register models?

Yeah, the reason is that the collision with `$dffsr` is the collision of two asynchronous reset signals, which can't be modeled in synthesizable verilog, see [this comment](https://github.com/YosysHQ/yosys/issues/5194#issuecomment-3558237171)

frontends/liberty/liberty.cc
widlarizer: The warning makes no sense anymore and I must have left that in by accident. I'm removing it sorry

widlarizer: It was a forgotten copy of a warning that was gated like
```cpp
if (child->id == "clear_preset_var1" || child->id == "clear_preset_var2")
	if (child->value[0] != 'X' && !clear_preset_reported)
```

frontends/liberty/liberty.cc
widlarizer: While SKY130 has "Q inverse of Qn" true, asap7 doesn't (it has both variables set to L), and I think that is the most realistic PDK we have to the state of the art in modern tech based on how they [present themselves](https://asap.asu.edu/).

Duplicating the emitted cells is a question I posed myself as well while working on this and I don't believe we're incurring any overhead:

- in synthesis, you don't work with internal models for these cells, you just map to them and only have them available as blackboxes since they're lowest level hardware primitives. In short you never call read_liberty
- in hierarchical formal or simulation, you only add a cell per cell type
- in flattened formal or simulation, you throw out the polarity you aren't using which is the common case (cheap opt_clean case)

KrystalDelusion:
Ooh async edge events, yeah that's fun.

frontends/liberty/liberty.cc
gussmith23: This makes sense. Just a few clarifications

>in hierarchical formal or simulation, you only add a cell per cell type

Why is hierarchical different from flattened in this case? My guess is that, without flattening, you can't actually tell which polarity is used and which one isn't?

Also, in the hierarchical case, you're effectively doubling the number of DFF cells, right?

passes/sat/async2sync.cc
```diff
@@ -157,33 +157,49 @@ struct Async2syncPass : public Pass {
 
 						SigSpec sig_set = ff.sig_set;
 						SigSpec sig_clr = ff.sig_clr;
+						SigSpec sig_clr_inv = ff.sig_clr;
 
 						if (!ff.pol_set) {
-							if (!ff.is_fine)
+							if (!ff.is_fine || sig_set.size() > 1)
```
gussmith23: was this `|| sig_set.size() > 1` just missing? or did something about this PR necessitate its addition? 

I'm also wondering, does `sig_set.size() > 1` logically imply `!ff.is_fine`? Put another way, are fine cells always size <= 1?

passes/sat/async2sync.cc
```diff
@@ -157,33 +157,49 @@ struct Async2syncPass : public Pass {
...
 						} else {
 							SigSpec tmp = module->OrGate(NEW_ID, ff.sig_d, sig_set);
-							module->addAndGate(NEW_ID, tmp, sig_clr, new_d);
+							tmp = module->AndGate(NEW_ID, tmp, sig_clr_inv);
+							module->addMuxGate(NEW_ID, tmp, State::Sx, set_and_clr, new_d);
 
 							tmp = module->OrGate(NEW_ID, new_q, sig_set);
-							module->addAndGate(NEW_ID, tmp, sig_clr, ff.sig_q);
+							tmp = module->AndGate(NEW_ID, tmp, sig_clr_inv);
+							module->addMuxGate(NEW_ID, tmp, State::Sx, set_and_clr, ff.sig_q);
```
gussmith23, line 205: It took me a bit to reason through the polarities seemingly flipping here. It seems right to me now that I understand, but it's probably worth adding a comment somewhere (maybe on the definitions of sig_clr and sig_clr_inv) to state explicitly that both signals are normalized to being active-low.

frontends/liberty/liberty.cc
widlarizer: > without flattening, you can't actually tell which polarity is used and which one isn't

yep

> in the hierarchical case, you're effectively doubling the number of DFF cells, right?

I'm doubling the number of flops in the liberty models, but that's a fixed number as the actualy design grows. It doesn't scale with your design going from having 1 flop to 1 million. There is only one module per liberty flop type. So:

> Why is hierarchical different from flattened in this case?

Because flattening fundamentally replaces each instance of each flop type with its model. This makes the added flop redundancy in the models scale with the design size. But then you opt_clean the flops modeling unused outputs so the simulation or formal effort is unchanged

passes/sat/async2sync.cc
widlarizer: I must have been playing with some invalid test case, InternalCellChecker makes it clear that set and clr are the same size as d and q. Sorry bout that

passes/sat/async2sync.cc
widlarizer: You mean active-high, right? Added a comment commit

gussmith23: Yes! Thank you

passes/sat/async2sync.cc
gussmith23: No problem!

frontends/liberty/liberty.cc
gussmith23: Thank you, this makes sense to me!

docs/source/cell/gate_reg_ff.rst
```diff
@@ -154,6 +154,10 @@ to the following Verilog code template, where ``RST_EDGE`` is ``posedge`` if
 ``RST_LVL`` if ``1``, ``negedge`` otherwise, and ``SET_EDGE`` is ``posedge`` if
 ``SET_LVL`` if ``1``, ``negedge`` otherwise.
 
+When both set and reset are active, the state and output is undefined. The Verilog
+code model does not correspond to this due to limitations
+of synthesizable Verilog.
+
```
gussmith23, line 160: Is this referencing the verilog code model immediately below?

gussmith23: I'm also wondering if we should elaborate on "limitations of synthesizable Verilog" in the documentation somewhere. Maybe not here, but maybe. Even now I'm failing to completely reconstruct these "limitations" from memory based on your comment alone. IIRC it's something about sensitivity lists? 

IIRC, you actually traced this to lines in the spec definition of synthesizable Verilog?

tests/techmap/dfflibmap_dffsr_r.lib
```diff
@@ -7,8 +7,8 @@ library(test) {
       clear      : "CLEAR";
       preset     : "PRESET";
       clear_preset_var1 : L;
-      clear_preset_var2 : L;
```
gussmith23, line 10: was this a bug in the existing code? or are you changing the functionality? (having var1 and var2 both set to L seems odd, so i'm guessing it's a bug?)

tests/techmap/dfflibmap_proc_formal.ys
```diff
@@ -0,0 +1,100 @@
...
+sat -verify -prove-asserts -set-init-undef -show-public -seq 3 miter
+
+##################################################################
+
+delete top miter
+
+copy top_unmapped top
+dfflibmap -liberty dfflibmap_dffsr_mixedpol.lib top
+
+async2sync
```
gussmith23: I notice that this test and the test below are the only tests in this PR which use `async2sync` over `clk2fflogic`. Are they intentionally testing `async2sync`?

tests/techmap/dfflibmap_formal.ys
```diff
@@ -32,12 +151,14 @@ read_liberty dfflibmap_dffsr_not_next.lib
 copy top top_unmapped
 dfflibmap -liberty dfflibmap_dffn_dffe.lib -liberty dfflibmap_dffsr_not_next.lib top
 
-async2sync
+clk2fflogic
```
gussmith23, line 154: Why are you using clk2fflogic here (and in the new tests)? I'm not doubting it, just don't quite understand the nuances of when you want one over the other.

docs/source/cell/gate_reg_ff.rst
widlarizer: Yeah https://github.com/YosysHQ/yosys/issues/5194#issuecomment-3558237171 but also hardware side users (the kind to actually want async resets on thir stuff) tend to understand that this "unsynthesizable verilog" is lore of evil magic and simply carefully stay clear of it when invoked... the explanation wouldn't be brief and would have to be put elsewhere. And we'd be describing stuff beyond the yosys program itself, maybe like, educational content out of scope for our docs

tests/techmap/dfflibmap_dffsr_r.lib
widlarizer: Both var1 and var2 set to L is a real situation that real hardware does, as suggested by asap7 doing this exact thing. In this file, the suffix _r implies "reset takes priority over set". Before, it didn't have that suffix, as the priority wasn't the point of including this file in the tests and wasn't paid attention to at all

tests/techmap/dfflibmap_formal.ys
widlarizer: async2sync discards all behaviors in between clock edges. It's quite a violent pass and can reduce the scope of observable behavior by doing "one timestep = one clock cycle". It's why I filed #5654, I didn't expect this.  But this is cool and good as long as you're dealing with a well behaved synchronous design. clk2fflogic meanwhile sticks to stricly verilog event semantics and allows me to soundly prove or disprove equivalence with regards to how outputs change as inputs change in between clock edges. See [this comment](https://github.com/YosysHQ/yosys/pull/5658#issuecomment-3879714129)

tests/techmap/dfflibmap_proc_formal.ys
widlarizer: Probably accidentally left in... not sure yet

docs/source/cell/gate_reg_ff.rst
gussmith23: Got it!

tests/techmap/dfflibmap_proc_formal.ys
gussmith23: Ok, will wait on you to resolve this

tests/techmap/dfflibmap_formal.ys
gussmith23: Thank you!

This makes me wonder whether we should put more guardrails on async2sync. I've used both passes in the past when doing sat things, and I'm remembering now that I've always landed on clk2fflogic for the same reasons as what you say above and in the comments. But people likely still end up using async2sync without fully comprehending the consequences. Off the top of my head, we could remove the clock entirely during async2sync (to make it extra clear that the clock is meaningless). Alternatively, we could insert an assertion that the clock is always 0 (to ensure that the user doesn't accidentally toggle the clock).

gussmith23:
Did we discuss this as a group, by the way? One thing I don't have the wisdom to know is how this might affect existing Yosys users/flows.

widlarizer:
I brought it up a couple times and it didn't really turn into a discussion, but the gist is that we already were handling the S&R behavior as undefined in proc, at least. Other Verilog frontends don't make this change observable either. So just internally, we were doing unsound non-equivalent rewrites, and users weren't exposed to it, except for Liberty techmapping, which is now strictly more correct. The only other option to achieve consistency would be to increase the amount of cell types (set over reset dffsr, reset over set dffsr, undefined priority dffsr), which I don't think anybody wants. I also believe that authors and users of alternate HDLs don't care about async resets (based on knowing a whole bunch of them)

gussmith23:
> I brought it up a couple times and it didn't really turn into a discussion, but the gist is that we already were handling the S&R behavior as undefined in proc, at least. Other Verilog frontends don't make this change observable either. So just internally, we were doing unsound non-equivalent rewrites, and users weren't exposed to it, except for Liberty techmapping, which is now strictly more correct. The only other option to achieve consistency would be to increase the amount of cell types (set over reset dffsr, reset over set dffsr, undefined priority dffsr), which I don't think anybody wants. I also believe that authors and users of alternate HDLs don't care about async resets (based on knowing a whole bunch of them)

Got it! Thanks!

PR #5658, opened 2026-01-31
satgen: fix flip flop clock

### LINKED ISSUE #5654: sat: dff output changes without clock edge

widlarizer:
Fix #5654 by adding clock detection logic for clocked flops to `SatGen`

### DIFF
--- a/tests/aiger/run-test.sh
+++ b/tests/aiger/run-test.sh
@@ -32,6 +32,7 @@ prep
 design -stash gate
 design -import gold -as gold
 design -import gate -as gate
+formalff -clk2ff
 miter -equiv -flatten -make_assert -make_outputs gold gate miter
 sat -verify -prove-asserts -show-ports -seq 16 miter
 " -l ${aag}.log
@@ -49,6 +50,7 @@ prep
 design -stash gate
 design -import gold -as gold
 design -import gate -as gate
+formalff -clk2ff
 miter -equiv -flatten -make_assert -make_outputs gold gate miter
 sat -verify -prove-asserts -show-ports -seq 16 miter
 " -l ${aig}.log
--- a/tests/arch/ecp5/dpram.ys
+++ b/tests/arch/ecp5/dpram.ys
@@ -10,6 +10,7 @@ miter -equiv -flatten -make_assert -make_outputs gold gate miter
 
 #Blocked by issue #1358 (Missing ECP5 simulation models)
 #ERROR: Failed to import cell gate.mem.0.0.0 (type DP16KD) to SAT database.
+#formalff -clk2ff
 #sat -verify -prove-asserts -seq 3 -set-init-zero -show-inputs -show-outputs miter
 
 design -load postopt
--- a/tests/arch/ice40/dpram.ys
+++ b/tests/arch/ice40/dpram.ys
@@ -7,6 +7,7 @@ memory
 opt -full
 
 miter -equiv -flatten -make_assert -make_outputs gold gate miter
+formalff -clk2ff
 sat -verify -prove-asserts -seq 3 -set-init-zero -show-inputs -show-outputs miter
 
 design -load postopt
--- a/tests/fsm/generate.py
+++ b/tests/fsm/generate.py
@@ -116,5 +116,6 @@ for idx in range(args.count):
                 print('opt; wreduce; share%s; opt; fsm;;' % random.choice(['', ' -aggressive']))
                 print('cd ..')
             print('miter -equiv -flatten -ignore_gold_x -make_outputs -make_outcmp gold gate miter')
+            print('formalff -clk2ff')
             print('sat -verify-no-timeout -timeout 20 -seq 5 -set-at 1 %s_rst 1 -prove trigger 0 -prove-skip 1 -show-inputs -show-outputs miter' % ('gold' if rst2 else 'in'))
 
--- a/tests/liberty/read_liberty.ys
+++ b/tests/liberty/read_liberty.ys
@@ -4,6 +4,7 @@ read_verilog retention.lib.verilogsim
 proc
 rename retention_cell retention_cell_vlog
 async2sync
+formalff -clk2ff
 equiv_make retention_cell_lib retention_cell_vlog equiv
 equiv_induct equiv
 equiv_status -assert equiv
--- a/tests/opt/opt_dff_dffmux.ys
+++ b/tests/opt/opt_dff_dffmux.ys
@@ -6,7 +6,7 @@ endmodule
 EOT
 
 proc
-equiv_opt -assert opt
+equiv_opt -formalff -assert opt
 design -load postopt
 select -assert-count 1 t:$dffe r:WIDTH=2 %i
 select -assert-count 0 t:$dffe %% t:* %D
@@ -21,7 +21,7 @@ endmodule
 EOT
 
 proc
-equiv_opt -assert opt
+equiv_opt -formalff -assert opt
 design -load postopt
 wreduce
 select -assert-count 1 t:$dffe r:WIDTH=2 %i
@@ -37,7 +37,7 @@ endmodule
 EOT
 
 proc
-equiv_opt -assert opt
+equiv_opt -formalff -assert opt
 design -load postopt
 select -assert-count 1 t:$dffe r:WIDTH=2 %i
 select -assert-count 0 t:$dffe %% t:* %D
@@ -52,7 +52,7 @@ endmodule
 EOT
 
 proc
-equiv_opt -assert opt
+equiv_opt -formalff -assert opt
 design -load postopt
 select -assert-count 1 t:$dffe r:WIDTH=4 %i
 select -assert-count 0 t:$dffe %% t:* %D
@@ -67,7 +67,7 @@ endmodule
 EOT
 
 proc
-equiv_opt -assert opt
+equiv_opt -formalff -assert opt
 design -load postopt
 wreduce
 select -assert-count 1 t:$sdffe r:WIDTH=2 %i
@@ -86,7 +86,7 @@ endmodule
 EOT
 
 proc
-equiv_opt -assert opt
+equiv_opt -formalff -assert opt
 design -load postopt
 wreduce
 select -assert-count 1 t:$sdffe r:WIDTH=2 %i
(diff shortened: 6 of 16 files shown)


widlarizer:
FPGA addf tests are failing because of a really weird bug I'm exposing or creating with my changes:

image

```
module \equiv

  wire \q_gold
  wire \m_gold
  wire \q_gate
  wire \m_gate
  wire input 1 \clk
  wire input 2 \d
  wire output 4 \m
  wire input 3 \rst

  cell $sdff $sdffgold
    parameter \WIDTH 1
    parameter \SRST_VALUE 1'0
    parameter \SRST_POLARITY 1
    parameter \CLK_POLARITY 1
    connect \SRST \rst
    connect \Q \q_gold
    connect \D \d
    connect \CLK \clk
  end
  cell $mux $muxgold
    parameter \WIDTH 1
    connect \Y \m_gold
    connect \S \rst
    connect \B 1'0
    connect \A \q_gold
  end

  cell $sdff $sdffgate
    parameter \WIDTH 1
    parameter \SRST_VALUE 1'0
    parameter \SRST_POLARITY 1
    parameter \CLK_POLARITY 1
    connect \SRST \rst
    connect \Q \q_gate
    connect \D \d
    connect \CLK \clk
  end

  cell $mux $muxgate
    parameter \WIDTH 1
    connect \Y \m_gate
    connect \S \rst
    connect \B 1'0
    connect \A \q_gate
  end

  cell $equiv $checkcell
    connect \A \m_gold
    connect \B \m_gate
    connect \Y \m
  end
end


module \equiv_m

  wire \m_gold
  wire \m_gate
  wire input 1 \q
  wire input 2 \rst
  wire output 3 \m

  cell $mux $muxgold
    parameter \WIDTH 1
    connect \Y \m_gold
    connect \S \rst
    connect \B 1'0
    connect \A \q
  end

  cell $mux $muxgate
    parameter \WIDTH 1
    connect \Y \m_gate
    connect \S \rst
    connect \B 1'0
    connect \A \q
  end

  cell $equiv $checkcell
    connect \A \m_gold
    connect \B \m_gate
    connect \Y \m
  end
end


module \equiv_q

  wire \q_gold
  wire \q_gate
  wire input 1 \clk
  wire input 2 \d
  wire output 4 \q
  wire input 3 \rst

  cell $sdff $sdffgold
    parameter \WIDTH 1
    parameter \SRST_VALUE 1'0
    parameter \SRST_POLARITY 1
    parameter \CLK_POLARITY 1
    connect \SRST \rst
    connect \Q \q_gold
    connect \D \d
    connect \CLK \clk
  end

  cell $sdff $sdffgate
    parameter \WIDTH 1
    parameter \SRST_VALUE 1'0
    parameter \SRST_POLARITY 1
    parameter \CLK_POLARITY 1
    connect \SRST \rst
    connect \Q \q_gate
    connect \D \d
    connect \CLK \clk
  end

  cell $equiv $checkcell
    connect \A \q_gold
    connect \B \q_gate
    connect \Y \q
  end
end
```

Only in the big module does the equiv check fail when `equiv_induct equiv; equiv_status -assert equiv`. The equiv cells in the other modules are fine. And using the `sat` command instead actually works. Super confusing

widlarizer:
I fixed CLK input x propagation in satgen, but the problem I described persists and isn't specific to $sdff, reducing to $dff doesn't save it. A $dff feeding a $mux input isn't equivalent to the same thing repeated, somehow. @KrystalDelusion I think you've worked with `equiv_` commands recently, do you have any ideas? I tried printing out all the SAT clauses and that was no good. I could also replace the usage of equiv_opt with `miter; sat` in the affected test but we should know what's going on here

widlarizer:
Here's my theory: I've changed the model so that flops can "stall" when clock isn't clocking, which is desirable behavior. `equiv_induct` can no longer prove, as the help string says, "that the two circuits will not diverge after they produce equal outputs (observable points via $equiv) for at least <N> cycles" for any N. This is approximately making sense to me. What the muxes are doing is allowing the clear to hide the state of the flops indefinitely. Since `equiv_induct` doesn't do a real k-induction base case, it doesn't know that their states can't diverge due to those two flops having the same init, so it tries to prove that they can't diverge given any non-x hidden states. It prints `Undef modelling: force def on 2 initial reg values and 3 inputs.` when `-undef` is used. The init values never get used. As of this change, flop internal states aren't "forced to be replaced with D" since clock can stall. And they also can be hidden from the equiv cell with the muxes on their outputs.

Fortunately, this is satgen failing to prove something correct, which is the acceptable error direction, rather than proving something incorrect. I'm now happy with changing the tests to use sat instead

KrystalDelusion:
> Since `equiv_induct` doesn't do a real k-induction base case

Was `sat` passing because it does do the base case properly?

widlarizer:
I think that `sat`, at least the way I was using it, starts from the initial state, as set with the -set-init-* arguments. So it doesn't do k-induction, it... kind of does the base case, if you squint

widlarizer:
Here's all I've got so far.

## The problem

The current behavior of Yosys is to model $dff cells in a dubious way for SAT solvers. Specifically, it disregards its clk input, and models it as if on every timestep, there was a clock edge. There’s two ways these flip flops can be modeled:

* timestep is a clock cycle
  * convenient, can easily prove things like “in 10 clock cycles from X happening, Y happens” by looking at 10 timesteps
  * allows proving things in equiv in bounded number of timesteps, since it enforces a mandatory “forward progress” - for non-trivial sequential design
  * as of writing this this is how *all* flip flops are modeled for SAT
  * this completely ignores the clock input
  * if a flip flop has a clock wired to constant zero, or has an assume on the clock to be zero, this model shows it’s unsound - footgun
  * if two flip flops have different clock signals, this will model the clock as synchronous, which may not be true - footgun
* a timestep is a moment when inputs change
  * matches Verilog semantics
  * allows proving things for a stalled clock
  * allows proving behavior about asynchronous sets and resets
  * those are valid observable behaviors of synchronous circuits and I need to prove things about them in our tests

My original bug report has a test case that shows what I see as problematic:

https://github.com/YosysHQ/yosys/issues/5654

Yosys provides a $ff cell type that explicitly has a “global clock” which has no clock input. Satgen conflates $ff and $dff and all $dff derived to act the same. I think it’s a mistake to be corrected - if they were equivalent for formal, there would be no point to having $ff. `tests/techmap/dff2ff.ys` replaces all $dff with $ff and checks they’re equivalent, which shows they’re equivalent for formal

This satgen model is exposed with `sat` and `equiv_` commands. Related changes may impact `sby`. It seems that `formalff -clk2ff` is the modern solution to getting meaningful formal behavior.

## The solution

This PR comes with all of the following:

* don’t change $ff satgen modeling
* model $dff to respect clock signal
  * this mirrors `clk2fflogic` to the point that I have doubts whether I shouldn't just remove `$dff` support from satgen so that we don't duplicate this...
  * interestingly `clk2fflogic` doesn't seem to model CLK undef behavior as loading an undef state, but otherwise, it would be fine to use `clk2fflogic` everywhere I wanted to model `$dff`
* `async2sync` shouldn’t be changed, since it’s valid to turn async into sync without assuming that global clock is desired
  * I’m not 100% certain about this but it’s what I’m doing for now
* `equiv_opt -async2sync` automatically runs `formalff -clk2ff` after `async2sync`
  * at least I think so, because otherwise equiv_opt / equiv_induct would fail a lot due to proving only the induction case, as opposed to sat, which only proves the “base case”
* add `equiv_opt -formalff` for when `-async2sync` or `-multiclock` isn’t needed but $ff behavior is
* tests written with “$dff = $ff” expectations are fixed
  * `formalff -clk2ff` and `equiv_opt -formalff` added all over the place
  * removed `tests/techmap/dff2ff.ys` - I could keep it and use `equiv_opt -formalff` but then we’re comparing $ff vs $ff so it’s unnecessary

Maybe I should also:

* require opt-in to actually enable models for $dff for better behavior
  * Whether the user expects old ($ff-like) behavior from $dff, or correct new $dff behavior, they should take an action to pick, rather than silently having the behavior change

## The future

For me it’s paranoia and search for better testing methodology.

* miter; sat doesn’t prove equivalence, has false positives, and switches that can manipulate the design accidentally too much
* equiv_opt is not great
* do we need a new interface for the sake of our tests of sequential circuits?

widlarizer:
After a discussion we've decided to keep $dff behaving same as $ff since "timestep is a clock cycle" is expected behavior and that keeping the resulting internal footgun is okay since sby is not exposing it. SatGen is also used beyond these commands so we might be asking for trouble if we mess with it, particularly since there's undertested passes using it. I'll be working on improving the Yosys developer experience by improving the ergonomy of writing tests such that the $dff / $ff inconsistency isn't likely to lead to accidents in the future. I'll also solve my original problem (proving $dffsr rewrite equivalence) by relying on clk2fflogic and sat. Additionally equiv is sketchy in that it has both false positives and negatives and we may look into using sby/eqy to test rewrites within Yosys more thoroughly. Sby is already available for tests/sva

PR #5663, opened 2026-02-02
opt_expr: fix const lhs of $pow to $shl

### LINKED ISSUE #5661: Incorrect `$pow` transform by opt_expr
### Version

v0.61

macOS

### Reproduction Steps

```
read_rtlil <<EOF
module \top
  wire width 3 input 2 \binary

  wire width 32 output 3 \y

  cell $pow $0
    parameter \A_WIDTH 32
    parameter \B_WIDTH 3
    parameter \A_SIGNED 1
    parameter \B_SIGNED 0
    parameter \Y_WIDTH 32
    connect \A 2
    connect \B \binary
    connect \Y \y
  end
end
EOF

dump
opt_expr
dump
```

### Expected Behavior

The transformed netlist is equivalent to the original

### Actual Behavior

```
  cell $pow $0
    parameter \A_WIDTH 32
    parameter \B_WIDTH 3
    parameter \A_SIGNED 1
    parameter \B_SIGNED 0
    parameter \Y_WIDTH 32
    connect \A 2
    connect \B \binary
    connect \Y \y
  end
```
is replaced with
```
  cell $shl $0
    parameter \A_WIDTH 1
    parameter \B_WIDTH 3
    parameter \A_SIGNED 1
    parameter \B_SIGNED 0
    parameter \Y_WIDTH 32
    connect \A 1'1
    connect \B \binary
    connect \Y \y
  end
```

Note the A input of the new instance is tied to `-1`. The function of the two netlists differs, take e.g.
```
eval -set binary 0
```
which first gives the result
```
Eval result: \y = 1.
```
and then
```
Eval result: \y = 32'11111111111111111111111111111111.
```

widlarizer:
Set A input signedness to unsigned in $pow -> $shl rewrite in opt_expr, even if it's a single-bit $pow. Fixes #5661. Can't prove correct with equiv since `$pow` isn't modeled with `SatGen` for computational complexity reasons, but the meaning of the change seems pretty self-evident. It's not too surprising that `$pow` cells would be mishandled due to how rarely exponentiation operators get used in real designs

### DIFF
--- a/passes/opt/opt_expr.cc
+++ b/passes/opt/opt_expr.cc
@@ -1667,7 +1667,11 @@ skip_identity:
 			int bit_idx;
 			const auto onehot = sig_a.is_onehot(&bit_idx);
 
-			if (onehot) {
+			// Power of two
+			// A is unsigned or positive
+			if (onehot && (!cell->parameters[ID::A_SIGNED].as_bool() || bit_idx < sig_a.size() - 1)) {
+				cell->parameters[ID::A_SIGNED] = 0;
+				// 2^B = 1<<B
 				if (bit_idx == 1) {
 					log_debug("Replacing pow cell `%s' in module `%s' with left-shift\n",
 							cell->name.c_str(), module->name.c_str());
@@ -1679,7 +1683,6 @@ skip_identity:
 					log_debug("Replacing pow cell `%s' in module `%s' with multiply and left-shift\n",
 							cell->name.c_str(), module->name.c_str());
 					cell->type = ID($mul);
-					cell->parameters[ID::A_SIGNED] = 0;
 					cell->setPort(ID::A, Const(bit_idx, cell->parameters[ID::A_WIDTH].as_int()));
 
 					SigSpec y_wire = module->addWire(NEW_ID, y_size);
--- a/tests/opt/opt_expr.ys
+++ b/tests/opt/opt_expr.ys
@@ -319,3 +319,59 @@ check
 equiv_opt -assert opt_expr -keepdc
 design -load postopt
 select -assert-count 1 t:$mul r:A_WIDTH=4 %i r:B_WIDTH=4 %i r:Y_WIDTH=8 %i
+
+###########
+
+design -reset
+read_rtlil <<EOF
+module \top
+  wire width 3 input 2 \binary
+
+  wire width 32 output 3 \y
+
+  cell $pow $0
+    parameter \A_WIDTH 32
+    parameter \B_WIDTH 3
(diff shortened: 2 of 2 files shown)


povik:
I propose testing by taking the setup from #5661 and then
```
opt_expr
sat -verify -set binary 0 -prove y 1
sat -verify -set binary 1 -prove y 2
sat -verify -set binary 2 -prove y 4
sat -verify -set binary 3 -prove y 8
```

passes/opt/opt_expr.cc
```diff
@@ -1668,6 +1668,7 @@ void replace_const_cells(RTLIL::Design *design, RTLIL::Module *module, bool cons
 			const auto onehot = sig_a.is_onehot(&bit_idx);
 
 			if (onehot) {
```
povik: I'm wondering if this should be `onehot && (!cell->parameters[ID::A_SIGNED].as_bool() || bit_idx < sig_a.size() - 1)`.

While putting together a reproducer I may have stumbled on an issue with how $pow are evaluated for negative A when B_SIGNED=0:
```
read_rtlil <<EOF
module \top
  wire width 3 input 2 \binary

  wire width 32 output 3 \y

  cell $pow $0
    parameter \A_WIDTH 2
    parameter \B_WIDTH 3
    parameter \A_SIGNED 1
    parameter \B_SIGNED 0
    parameter \Y_WIDTH 32
    connect \A 2'10
    connect \B \binary
    connect \Y \y
  end
end
EOF

eval -set binary 0
eval -set binary 1
eval -set binary 2
```
produces
```
2. Executing EVAL pass (evaluate the circuit given an input).
Eval result: \y = 1.

3. Executing EVAL pass (evaluate the circuit given an input).
Eval result: \y = 2.

4. Executing EVAL pass (evaluate the circuit given an input).
Eval result: \y = 4.
```
RTLIL::const_pow itself has a branch for negative A, but signed1 may be overridden before it reaches that point. simlib.v and rtlil.cc suggest A_SIGNED!=B_SIGNED is explicitly allowed on $pow.

povik (approved):
There may be pending issues with the case A_SIGNED=1

passes/opt/opt_expr.cc
widlarizer: Very good point, I changed that condition and added a test that we don't opt on negative A

povik: > simlib.v and rtlil.cc suggest A_SIGNED=B_SIGNED is explicitly allowed on $pow.

Oops, this is meant to say A_SIGNED!=B_SIGNED

PR #5664, opened 2026-02-02
Parallelize `opt_clean` pass

rocallahan:

This is the next step after [parallelizing `opt_merge`](https://github.com/YosysHQ/yosys/pull/5530).

This PR depends on https://github.com/YosysHQ/yosys/pull/5621, https://github.com/YosysHQ/yosys/pull/5629 and https://github.com/YosysHQ/yosys/pull/5631. Once those are merged, I will clean up this PR. I'm putting it here now in case people want to see what I'm planning and so I can get CI clean.

`opt_clean` is much more complex than `opt_merge` and the changes here are correspondingly greater. In particular I need to introduce several more parallel abstractions. I have done my best to preserve the original code structure. I've fuzzed millions of testcases to detect any differences in results between this and the original pass, and as far as I know I've fixed them all.

The scalability on large flattened circuits is not quite as impressive as it was for `opt_merge` but it's still pretty good. For a circuit with 3.5M cells with `opt_merge` already applied, running `opt_clean` once removes about 10% of the wires. Then I run `opt_clean` again to measure the scalability when there is nothing to remove (a common case). The results:
opt_clean scalability
So with 40 cores we get a 6x speedup in the dirty case and a 9x speedup in the clean case. But the 1-core case is actually 1.6x faster than current yosys `main` (a68fee1115df955003ad803a7844a51d4f5023db) and the clean case is 2.1x faster, so the clean case is actually >20x current yosys `main`. (The dirty case doesn't parallelize as well because modifying RTLIL has to happen on the main thread and there are a lot of wires to remove in this case. This could be improved but it might not be worth the extra complexity --- removing this much of the design is probably rare.)

For smaller circuits there is a penalty for the extra complexity, but I've done my best to mitigate that and offset it with optimizations. For the jpeg synth testcase (`read_verilog -sv -I~/OpenROAD-flow-scripts/flow/designs/src/jpeg/include ~/OpenROAD-flow-scripts/flow/designs/src/jpeg/*.v; synth`), this is very slightly faster than current yosys `main` with or without multicore, on my system:
```
main YOSYS_MAX_THREADS=1:
  Time (mean ± σ):     17.094 s ±  0.707 s    [User: 16.256 s, System: 0.841 s]
  Range (min … max):   16.333 s … 17.816 s    10 runs
 
main:
  Time (mean ± σ):     16.888 s ±  0.421 s    [User: 16.038 s, System: 0.858 s]
  Range (min … max):   16.508 s … 17.952 s    10 runs

parallel-opt-clean YOSYS_MAX_THREADS=1:
  Time (mean ± σ):     16.924 s ±  0.495 s    [User: 15.606 s, System: 1.320 s]
  Range (min … max):   16.563 s … 18.245 s    10 runs
 
parallel-opt-clean:
  Time (mean ± σ):     16.778 s ±  0.150 s    [User: 15.634 s, System: 1.179 s]
  Range (min … max):   16.602 s … 17.101 s    10 runs
```

This is a big PR so let me know what I can do to make it easier to swallow!

### DIFF
--- a/passes/opt/opt_clean/opt_clean.cc
+++ b/passes/opt/opt_clean/opt_clean.cc
@@ -31,7 +31,7 @@ void rmunused_module(RTLIL::Module *module, bool rminit, CleanRunContext &clean_
 
 	// Use no more than one worker per thousand cells, rounded down, so
 	// we only start multithreading with at least 2000 cells.
-	int num_worker_threads = ThreadPool::work_pool_size(0, module->cells_size(), 1000);
+	int num_worker_threads = ThreadPool::work_pool_size(0, module->cells_size(), 10000);
 	ParallelDispatchThreadPool::Subpool subpool(clean_ctx.thread_pool, num_worker_threads);
 	remove_temporary_cells(module, subpool, clean_ctx.flags.verbose);
 	rmunused_module_cells(module, subpool, clean_ctx);
--- a/passes/opt/opt_clean/opt_clean.h
+++ b/passes/opt/opt_clean/opt_clean.h
@@ -64,7 +64,7 @@ private:
 		int thread_pool_size = 0;
 		for (auto module : selected_modules)
 			thread_pool_size = std::max(thread_pool_size,
-				ThreadPool::work_pool_size(0, module->cells_size(), 1000));
+				ThreadPool::work_pool_size(0, module->cells_size(), 10000));
 		return thread_pool_size;
 	}
 


rocallahan:
I should add some unit tests for the types in `threading.h`.

rocallahan:
The dependent CLs have been merged. There are two failures in CI. One issue is that this version of MSVC++ seems to be unable to instantiate `std::unordered_set` with move-only elements :-(. The other issue is a crash running `tests/svtypes/typedef_package.sv` with Verific. I'm not sure how this CL would affect that test specifically and there's no diagnostic information in the CI log.

rocallahan:
> The other issue is a crash running `tests/svtypes/typedef_package.sv` with Verific. I'm not sure how this CL would affect that test specifically and there's no diagnostic information in the CI log.

This was a bug in my code related to `init` wire attributes which apparently Verific generates but Yosys does not. Fuzzing didn't catch it because the fuzzing grammar didn't generate `init` attributes. I've updated the PR to extend the grammar with `init` attributes and verified that fuzzing now catches this case. I've also added an RTLIL testcase for the bug that will catch it when tests are run without Verific.

I'm still trying to find a workaround for the MSVC++ issue.

rocallahan:
The Verific build is failing because libgmock-dev isn't installed on that system. I don't know how to fix that.

rocallahan:
CI is clean except that the system that runs Verific tests needs gmock installed.

kernel/threading.h
```diff
@@ -154,6 +159,30 @@ class ThreadPool
 #endif
 };
 
+// A range of integers [start_, end_) that can be iterated over with a
+// C++ range-based for loop.
+struct IntRange {
```
widlarizer: Maybe move to util.h

kernel/threading.h
```diff
@@ -391,6 +391,177 @@ class ShardedVector {
...
+			// a lot and I never got around to figuring out why.
+			std::vector<std::vector<Value>> &buckets = all_buckets[thread.thread_num];
+			size_t bucket = static_cast<size_t>(v.hash) % buckets.size();
+			buckets[bucket].emplace_back(std::move(v));
+		}
+		// Then call `process` for each thread. All `process()`s must finish before using
+		// the `Builder` to construct a `ShardedHashSet`.
+		void process(const ThreadIndex &thread) {
+			int size = 0;
+			for (std::vector<std::vector<Value>> &buckets : all_buckets)
```
widlarizer, line 476: I thought this was wrong, but no, it's just that `all_buckets` has both its first and second dimension be thread count. Tricky to think about, consider adding a comment for the dimensions of `all_buckets`

kernel/threading.h
```diff
@@ -562,6 +562,112 @@ class ShardedHashSet {
...
+		: batch_size(batch_size), thread_states(num_threads) {}
+	int num_threads() const { return GetSize(thread_states); }
+	// Push some work to do. Pushes and pops with the same `thread` must
+	// not happen concurrently.
+	void push(const ThreadIndex &thread, T work) {
+		ThreadState &thread_state = thread_states[thread.thread_num];
+		thread_state.next_batch.emplace_back(std::move(work));
+		if (GetSize(thread_state.next_batch) < batch_size)
+			return;
+		bool was_empty;
```
widlarizer, line 578: Probably needs explicit initialization

widlarizer:
@mmicko The test-verific job is running in an unmanaged custom environment right? Please try adding gmock, it should fix the unit tests here since they use the convenient matchers like `UnorderedElementsAre`

mmicko:
@widlarizer Update CI docker, all green now

passes/opt/opt_clean.cc
```diff
@@ -123,117 +198,243 @@ void rmunused_module_cells(Module *module, bool verbose)
...
+	};
+	struct WireDriversCollisionHandler {
+		void operator()(WireDrivers &incumbent, WireDrivers &new_value) const {
+			log_assert(new_value.extra_driver_cells == nullptr);
+			if (!incumbent.extra_driver_cells)
+				incumbent.extra_driver_cells.reset(new pool<int>());
+			incumbent.extra_driver_cells->insert(new_value.driver_cell);
+		}
+	};
+	using Wire2Drivers = ShardedHashSet<WireDriver, WireDriversKeyEquality, WireDriversCollisionHandler>;
```
widlarizer: The fact that this is in fact still a map implemented with a hash set by hashing and comparing for equality only the bit and not the drivers took me a long time to figure out. Please make this explicit in a comment because at first N glances this looks like a parallel `std::unordered_set<WireDriver>` which it is not

kernel/threading.h
```diff
@@ -391,6 +391,177 @@ class ShardedVector {
 	Buckets buckets;
 };
 
+template <typename V>
+struct DefaultCollisionHandler {
+	void operator()(typename V::Accumulated &, typename V::Accumulated &) const {}
```
widlarizer, line 387: I would prefer for this to error out. It's not meeting the spec of "used to reduce two `V::Accumulated` values into a single value."

rocallahan:
> I've also put up a [PR](https://github.com/rocallahan/yosys/pull/2) into this PR branch with comments that make the data flow clearer

Would you prefer me to merge these changes into my commits (losing attribution) or keep your commit separate in my stack of commits?

kernel/threading.h
rocallahan: Done.

kernel/threading.h
rocallahan: Done.

kernel/threading.h
rocallahan: Does it? We initialize it unconditionally three lines later.

passes/opt/opt_clean.cc
rocallahan: Done. I've renamed it from `ShardedHashSet` to `ShardedHashtable`.

kernel/threading.h
rocallahan: Arguably it is :-). The default behavior is just "pick one of the two values" (and we pick the 'current' value because that's free). I'll add a comment to `DefaultCollisionHandler` and you can tell me if it's satisfying :-).

kernel/threading.h
widlarizer: Oh, I must have misread the anonymous namespace as a conditional, my bad

kernel/threading.h
widlarizer: That does seem like an edge case (something like mapping to a void type) even if it's already being used that way and I wouldn't want the default behavior of a map to be something that only makes sense for the edge case if I forget to override this

widlarizer:
With this test I see a significant increase in memory usage if I let it run parallelized on my 24 thread machine: `yosys -p "read_rtlil fft64_width_64.il; synth"` with [this file](https://github.com/widlarizer/yosys-perf/blob/main/artifacts/fft64_width_64.il)
384.73 MB -> 449.09 MB
I'm guessing this is expected? Is it because of the extra temporary space needed to collect stuff from shards?

widlarizer:
I've confirmed the original functionality is retained in code, meaning the changes to parallelization and precomputing things make sense to me. I have concerns about memory usage - I think it can (and as I have mentioned in my prev comment, does) increase overall synthesis memory usage while shaving off a significant but not drastic chunk of the overall run time for large designs. For a lot of users, memory is a fixed resource, while run time is a question of convenience. Is it really enough to have a single `YOSYS_MAX_THREADS` variable control all the parallelism here? Is it possible that this is a lot more likely to dominate memory usage unlike opt_merge? I think it is okay to see what feedback we get from users after the release though

kernel/threading.h
rocallahan: I think using a degenerate map as a set isn't really an edge case, it's a common case. But anyway, I'm happy to remove the default and call this "SetCollisionHandler" to make that clearer.

rocallahan:
I guess you missed this question so let me re-up it:
Would you prefer me to merge [these changes](https://github.com/rocallahan/yosys/pull/2) into my commits (losing attribution) or keep your commit separate in my stack of commits?

rocallahan:
Thank you again so very much for your time spent reviewing this.

I will look into what can be done to improve memory usage. There are a few places where we can proactively free memory, which should help a bit, and there may be other space optimization opportunities. But yes, this approach to parallelism, with strict phase boundaries, inherently increases memory usage.

> I think it can (and as I have mentioned in my prev comment, does) increase overall synthesis memory usage while shaving off a significant but not drastic chunk of the overall run time for large designs.

If we're eventually able to parallelize all the most common and expensive passes, then the overall run time *will* be drastically reduced, and the maximum memory usage should not increase much if we don't do anything stupid.

(Simple cost-benefit analysis of individual optimizations in a series can be misleading. Thought experiment: a task has two passes A and B that take equal amounts of time, and each pass can be optimized to make it 10x faster. But the first optimization PR will only offer a 1.8x speedup, which doesn't sound that compelling.)

> For a lot of users, memory is a fixed resource, while run time is a question of convenience.

True. The question is how much memory usage is a limiting factor for users in practice. The Yosys users I work with are much more time-sensitive than memory limited. And FWIW the parallel ABC pass that I did, and `opt_merge` too, also require more memory.

But let's see what can be done to reduce memory usage before stressing about this tradeoff.

widlarizer:
I prefer keeping no-functional-changes separate from functional ones, so please merge the refactoring PR after reviewing it, this should keep the refactor attribution to me (more for `git blame` than for green squares on my github profile), that was what I was trying to say [here](https://github.com/rocallahan/yosys/pull/2#issuecomment-3972971794) about the squash

> Thank you again so very much for your time spent reviewing this.

No problem, this is code that always runs and almost all of the actual review effort was the refactoring based on how it already worked beforehand. We've always had resurfacing issues and doubts about how it _should_ work so it's good we've got a better grip on it now

widlarizer:
> But let's see what can be done to reduce memory usage before stressing about this tradeoff.

To clarify, you meant in separate work or within this PR?

widlarizer:
One option would be to disable parallelism by default but advertise it in an info log message that prints how many threads are available

rocallahan:
> To clarify, you meant in separate work or within this PR?

Within this PR if possible, since we don't want regressions.

> With this test I see a significant increase in memory usage if I let it run parallelized on my 24 thread machine: yosys -p "read_rtlil fft64_width_64.il; synth" with [this file](https://github.com/widlarizer/yosys-perf/blob/main/artifacts/fft64_width_64.il)
384.73 MB -> 449.09 MB
I'm guessing this is expected? Is it because of the extra temporary space needed to collect stuff from shards?

This was a pain to track down but I figured it out. Here's what happens in the 2-thread case:
* We trundle along and eventually do an `opt_clean` on a module with 113K cells. This uses multiple threads.
** Each of those threads allocates and frees memory. I think the peak allocated memory is about 60MB, summed over all threads. So for the 2-thread case it's about 30MB per thread.
** The [glibc allocator](https://sourceware.org/glibc/wiki/MallocInternals) tries to reduce lock contention by creating multiple arenas, not necessarily one per thread, but at least in the 2-thread case it creates an arena for the extra thread as well as using the main thread's normal arena.
** So about 30MB of memory is allocated to the off-main-thread's arena, which becomes part of the process RSS.
** The off-main-thread exits.
** That thread's arena is not released! It's put on an [internal glibc free list](https://codebrowser.dev/glibc/glibc/malloc/arena.c.html#911).
* In my testing, we haven't actually reached the run's max RSS yet.
* We carry on and do an `opt_dff`.
** This pass allocates several tens of megabytes of memory (on the main thread, because it's not parallel yet).
** Unfortunately instead of using the free memory from that arena on the free list, glibc takes memory from the main thread's arena.
** This increases the process RSS because now the enlarged main thread arena and the other arena are both in the resident set.

With more threads the problem gets a bit worse because the fraction of `opt_clean` allocations that happen off the main thread increases. Of course that fraction must always be less than 1, so the RSS increase is bounded (by about 60MB in this case). And running more parallel passes, we won't keep leaking memory indefinitely, because glibc will reuse those arenas.

rocallahan:
AFAICT there's no easy fix for this, since it's really a property of the glibc allocator. We can call `mallopt(M_ARENA_MAX, 1)` to force only one arena, but that hurts parallelism. Interestingly, if you do use that, you can see that this PR does not increase max RSS, which proves we're not increasing the high-water-mark of actual memory usage apart from this glibc issue.

We could use a different allocator --- I suspect tcmalloc would work better here --- but that's obviously a big ask.

We could do something very hairy like preallocate a bunch of memory on the main thread from its arena to be used by the off-main-threads, but that would hurt performance and just be very unpleasant.

rocallahan:
I'm going to go out on a limb and suggest that increasing peak memory usage by 60-70MB here is not a big deal.

The memory is not lost and can be reused, and will be with bigger designs. glibc arenas default to 128MB in size, and IIUC when the main thread's arena fills up glibc will actually take arenas off the free list for use by the main thread. So the exact amount of excess RSS is hard to predict and depends on how arenas are mapped to threads and what the maximum heap sizes are of the various threads, but it shouldn't grow indefinitely.

I guess that your users typically aren't memory constrained. If they were, and their designs were pushing close to their memory limits, then you'd probably get a steady stream of bug reports about OOMs and requests to reduce Yosys' memory usage or discussions about how to work around it. (I've already fixed some bugs where Yosys was using memory [very inefficiently](https://github.com/YosysHQ/yosys/pull/5316) --- but I didn't fix them in response to OOM bug reports.) Do you have any more information about this?

rocallahan:
I have added a few commits to the tail end of the stack, you might want to take a look.

QuantamHD:
@widlarizer it seems like this speedup is worth it.

widlarizer:
Thanks for the deep dive, the extra RSS is well justified now and this is good to go without limiting parallelism by default. How intrusive would tcmalloc be in practice? It doesn't seem like a big ask at first glance to me. And thanks for the refactoring fixup

PR #5667, opened 2026-02-02
Guard vhdl_file::UNDEFINED behind VERIFIC_VHDL_SUPPORT.

Logikable:


mmicko:
Thanks for spotting this one

PR #5411, opened 2025-10-07
gowin: dsp: Add basic DSP block inferencing for various MULT cells

plaes:
Add basic MULT cell inferencing to gowin backend for gw1n and gw2a families.
It would be helpful if someone could verify this properly on real FPGA as well (haven't yet figured out how to properly build a real life "test" for it..).

This was motivated by missing MULT18X18 support for gw2a family which was discovered when attempting to synthesize this project: https://github.com/ryomuk/TangNanoDCJ11MEM/tree/main/hdl/unix

Tests have been added, but need some scrutiny.. ;)

PR #5670, opened 2026-02-03
Gowin: Add DSP inference for GW1N and GW2A

max-kudinov:
Adds `MULT9X9`, `MULT18X18` and `MULT36X36` inference. GW5A family has other primitives, and I don't have any hardware to test it, so it's left for later.

For now inference is off by default, it has to be enabled with `-dsp` option. There are issues in some complex designs, most likely related to PNR and/or packing, so this work is currently experimental.

Most of what was done was just copied from `synth_lattice` script. `techmap` has to be done before `share` pass, otherwise multiplying logic won't be merged. And according to this forum [answer](https://yosyshq.discourse.group/t/gowin-mapping-dsp-cells-mult36x36-mult18x18-and-mult9x9-during-synthesis/91/2), it also has to be before `alumacc`.

Because of that, `synth -run coarse` step had to be expanded on separate passes, basically the same way as in `synth_lattice` script. And also because of that, `flatten` label can be merged with `coarse` to avoid second `proc` run.

Currently `techlibs/gowin/cells_sim.v` is missing MULT models, but adding them had no effect, so I'm not sure if they are needed or not. If yes - let me know, I'll add them.

Tests are a bit modified version of  #5411.

### DIFF
--- a/tests/arch/gowin/mul_gw1n.ys
+++ b/tests/arch/gowin/mul_gw1n.ys
@@ -1,0 +1,53 @@
+read_verilog ../common/mul.v
+chparam -set X_WIDTH 8 -set Y_WIDTH 8 -set A_WIDTH 16
+hierarchy -top top
+proc
+# equivalence checking is somewhat slow (and missing simulation models)
+synth_gowin -family gw1n
+cd top # Constrain all select calls below inside the top module
+select -assert-count 1 t:MULT9X9
+
+
+# Make sure that DSPs are not inferred with -nodsp option
+design -reset
+read_verilog ../common/mul.v
+chparam -set X_WIDTH 8 -set Y_WIDTH 8 -set A_WIDTH 16
+hierarchy -top top
+proc
+synth_gowin -family gw1n -nodsp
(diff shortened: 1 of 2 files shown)


mmicko:
Adding here answer from @max-kudinov on other platform.

> Hi, all mentioned issues were resolved in gowin_pack and nextpnr (https://github.com/YosysHQ/apicula/pull/435 and https://github.com/YosysHQ/nextpnr/pull/1633). Currently, that larger design works without any issues and so far I haven't discovered anything related to DSP inference that doesn't work.
I would even suggest making DSP inference on by default and changing the flag to -nodsp to opt out like in lattice. If no one objects of course

Also do agree that it should be `-nodsp` in this case and got confirmation that it is fine from @yrabbit 
On technical side there are no objections.

yrabbit:
That's fine with me.

max-kudinov:
Done, changed the option to `-nodsp`, now DSP inference is on by default. Also updated tests to reflect that

mmicko:
Thanks a lot @max-kudinov. Looks good, just need to wait for CI to get green.

PR #5671, opened 2026-02-03
Only reuse ABC processes if we're using `yosys-abc` and it was built with `ENABLE_READLINE`

rocallahan:
See https://github.com/YosysHQ/yosys/pull/5593#issuecomment-3843677808.

Reusing ABC processes is fragile. Only enable it when using `yosys-abc` built with `ENABLE_READLINE`, which is required for ABC's prompts to be reliably flushed. Otherwise fall back to spawning a new ABC process for every ABC run, which is a lot slower when there are thousands of ABC runs to do, but more reliable.

The fallback code is the old `run_command` code so it should be reliable.

### DIFF
--- a/passes/techmap/abc.cc
+++ b/passes/techmap/abc.cc
@@ -143,6 +143,14 @@ struct AbcConfig
 	bool markgroups = false;
 	pool<std::string> enabled_gates;
 	bool cmos_cost = false;
+
+	bool is_yosys_abc() const {
+#ifdef ABCEXTERNAL
+		return false;
+#else
+		return exe_file == yosys_abc_executable;
+#endif
+	}
 };
 
 struct AbcSigVal {
@@ -155,7 +163,12 @@ struct AbcSigVal {
 	}
 };
 
-#if defined(__linux__) && !defined(YOSYS_DISABLE_SPAWN)
+// REUSE_YOSYS_ABC_PROCESSES only works when ABC is built with ENABLE_READLINE.
+#if defined(__linux__) && !defined(YOSYS_DISABLE_SPAWN) && defined(YOSYS_ENABLE_READLINE) && !defined(YOSYS_LINK_ABC)
+#define REUSE_YOSYS_ABC_PROCESSES
+#endif
+
+#ifdef REUSE_YOSYS_ABC_PROCESSES
 struct AbcProcess
 {
 	pid_t pid;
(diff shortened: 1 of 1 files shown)


passes/techmap/abc.cc
```diff
@@ -143,6 +143,14 @@ struct AbcConfig
...
 	pool<std::string> enabled_gates;
 	bool cmos_cost = false;
+
+	bool is_yosys_abc() const {
+#ifdef ABCEXTERNAL
+		return false;
+#else
+		return exe_file == yosys_abc_executable;
+#endif
+	}
```
KrystalDelusion, line 153: This is a nice way to do it.  It still misses the case where `yosys-abc` is built without readline, and then `yosys` is rebuilt without cleaning, but that's a build system problem and not something we need to address here.

passes/techmap/abc.cc
```diff
@@ -155,7 +163,12 @@ struct AbcSigVal {
 	}
 };
 
-#if defined(__linux__) && !defined(YOSYS_DISABLE_SPAWN)
+// REUSE_YOSYS_ABC_PROCESSES only works when ABC is built with ENABLE_READLINE.
+#if defined(__linux__) && !defined(YOSYS_DISABLE_SPAWN) && defined(YOSYS_ENABLE_READLINE)
+#define REUSE_YOSYS_ABC_PROCESSES
+#endif
```
KrystalDelusion: Minor difference that I did in https://github.com/YosysHQ/yosys/commit/0bee85f5c0ef6889a8323882b2bcba7eb91a6b8f was to put this in an `#elif` after `#ifdef YOSYS_LINK_ABC` since that will also disable the reuse.

KrystalDelusion:
Actually I can just merge this into the release candidate instead of my fix and add the `&& !defined(YOSYS_LINK_ABC)` myself (if you haven't already done it)

passes/techmap/abc.cc
rocallahan: Done.

KrystalDelusion:
Thanks!

mmicko:
Thanks. It is merged in https://github.com/YosysHQ/yosys/commit/7326bb7d6641500ecb285c291a54a662cb1e76cf but not recognized by github since it was on fork. Next time we are going just to merge PR to release branch in cases like this

PR #5673, opened 2026-02-05
Turn conflicting init values into a warning instead of a fatal error

rocallahan:
Fuzzing can generate test cases that trigger this.

An alternative approach would be to have some kind of global flag that we set while fuzzing to turn errors like this into warnings. That's a little more complicated. Let me know whether that is preferred.

### DIFF
--- a/kernel/ffinit.h
+++ b/kernel/ffinit.h
@@ -52,7 +52,7 @@ struct FfInitVals
 
 				if (initbits.count(bit)) {
 					if (initbits.at(bit).first != val)
-						log_error("Conflicting init values for signal %s (%s = %s != %s).\n",
+						log_warning("Conflicting init values for signal %s (%s = %s != %s).\n",
 								log_signal(bit), log_signal(SigBit(wire, i)),
 								log_signal(val), log_signal(initbits.at(bit).first));
 					continue;


widlarizer:
I think we shouldn't loosen this behavior for all users for the sake of fuzzers. Since FfInitVals is used all over the place, it seems that introducing a global variable set in a driver flag that renders in the developer section of the help. Probably best to add a general `--fuzz` switch or similar, and then if this switch accumulates other features that non-fuzzing users want, we can expose those later

whitequark (changes requested):
I am strongly against this.

`init` should have never been an attribute in first place; let's not make this mistake worse.

whitequark:
> Probably best to add a general `--fuzz` switch or similar, and then if this switch accumulates other features that non-fuzzing users want, we can expose those later

I don't think this is the way to go either; fuzzers can run a legalization pass before feeding the netlist in (such legalization pass might be fuzzer-specific or a part of Yosys, that is not very important).

rocallahan:
> fuzzers can run a legalization pass before feeding the netlist in (such legalization pass might be fuzzer-specific or a part of Yosys, that is not very important).

The problem is, `init` attribute conflicts can occur whenever new wire connections are created. So there is nothing you can do "up front" to detect or prevent `init` attribute conflicts that might occur later in some arbitrary flow.

I suppose we can solve this for limited cases by adding a legalization pass that removes conflicting `init` attributes based on the connections currently in the design, and then making sure we run that pass between anything that creates new connections and anything that creates `FfInitVals`. That's fragile though.

whitequark:
> So there is nothing you can do "up front" to detect or prevent `init` attribute conflicts that might occur later in some arbitrary flow.

I don't understand this argument: clearly HDL frontends are capable of doing so. Why are fuzzers different here?

rocallahan:
> I don't understand this argument: clearly HDL frontends are capable of doing so. Why are fuzzers different here?

I meant, you can't detect and prevent future `init` conflicts starting with arbitrary RTLIL. Unless you do something drastic like remove them all.

I assume the intention is for HDL frontends to only apply `init` to wires that are FF ouputs, so that conflicting `init` attributes would require a net driven by multiple FFs, which we're OK treating as a fatal error?

If so, then we could try to apply the same constraints during fuzzing. Applying `init` only to FF outputs is easy. The second constraint is a problem, because Yosys does support multiply-driven nets in some cases and we want fuzzing to test that.

povik:
> I assume the intention is for HDL frontends to only apply init to wires that are FF ouputs, so that conflicting init attributes would require a net driven by multiple FFs, which we're OK treating as a fatal error?

My working model is that `init` with a defined value can only occur on a net directly driven by a register, and that is an RTLIL invariant. Any pass is responsible for upholding that invariant.

I think the case of multiple flops driving the same net would be caught by `check -assert` or in the HDL frontend already.

widlarizer:
Agreed, a normalizing pass for fuzzing sounds smarter than a global var, and yes if from a valid design you can get a design with conflicting inits with a synthesis workflow that's a bug we'd like to hear about

PR #5675, opened 2026-02-09
kernel/celledges: cover more cell types

rowanG077:
I was experimenting with the `CellEdgesDatabase` and ran into missing cell types(`logic_` specifically). I took the time to add most of the missing ones marked in the `FIXME` notes where I knew what they did.

I can add a separate tests for this, but it seems none of the cell edges have tests. Please advise what you would like best.

### DIFF
--- a/kernel/celledges.cc
+++ b/kernel/celledges.cc
@@ -112,6 +112,41 @@ void reduce_op(AbstractCellEdgesDatabase *db, RTLIL::Cell *cell)
 		db->add_edge(cell, ID::A, i, ID::Y, 0, -1);
 }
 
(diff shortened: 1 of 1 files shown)


gussmith23:
This looks good to me, just waiting on tests to pass before I approve.

rowanG077:
The failing test seems like a flake to me.

gussmith23:
Agree, I will keep rerunning. I think GitHub actions might be degraded right now.

rowanG077:
CI passed, anything holding this one up?

gussmith23:
Nope, thanks for the bump! And thanks for the contribution!

Issue #5523, opened 2025-11-28
Unify info about make test and make unit-test


widlarizer:
### Link to page

https://yosyshq.readthedocs.io/projects/yosys/en/latest/yosys_internals/extending_yosys/test_suites.html

### Build number

#30469828

### Issue

We shouldn't consider make test and make unit-test as categorically different

### Expected

Whenever we instruct people to run `make test`, we should instruct them to run `make unit-test` too. Same goes for dependencies of `make test` (iverilog, gtkwave?) and `make unit-test` (gtest). Related to #5522

KrystalDelusion:
I don't think gtkwave is required for running the tests; the only place it gets called in the Yosys source repo is in one of the examples: https://github.com/YosysHQ/yosys/blob/e4044e1b4a28e4412a168d6ed3fe7c4192145814/examples/cmos/testbench_digital.sh#L8

Definitely an oversight that `iverilog` install instructions aren't included.

Issue #5522, opened 2025-11-28
Unify make test and make unit-test


widlarizer:
### Feature Description

I'd like the Makefile `test` target to be renamed to something (`regression-test`?) and for `make test` to run unit tests too

PR #5676, opened 2026-02-09
Run unit tests on make test

widlarizer:
Implements #5522 and how I think #5523 should be resolved in the docs. For what was `make test` I'm introducing the term "vanilla tests" because I have nothing smarter and I'm open to alternatives. Unit tests take a fraction of the runtime of vanilla tests so it's ok to always run them

### DIFF
--- a/.github/actions/setup-build-env/action.yml
+++ b/.github/actions/setup-build-env/action.yml
@@ -52,15 +52,6 @@ runs:
         packages: graphviz xdot
         version: ${{ inputs.runs-on }}-docsys
 
-    # if updating test dependencies, make sure to update
-    # docs/source/yosys_internals/extending_yosys/test_suites.rst to match.
-    # - name: Linux test dependencies
-    #   if: runner.os == 'Linux' && inputs.get-test-deps == 'true'
-    #   uses: awalsh128/cache-apt-pkgs-action@v1.6.0
-    #   with:
-    #     packages:
-    #     version: ${{ inputs.runs-on }}-testys
-
     - name: Install macOS Dependencies
       if: runner.os == 'macOS'
       shell: bash
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -132,7 +132,7 @@ jobs:
       - name: Run tests
         shell: bash
         run: |
-          make -j$procs vanilla-test OBJS= TARGETS= EXTRA_TARGETS= CONFIG=$CC
+          make -j$procs vanilla-test TARGETS= EXTRA_TARGETS= CONFIG=$CC
 
       - name: Report errors
         if: ${{ failure() }}


KrystalDelusion:
Maybe "execution tests"?  I think they're closest to functional tests in [this hierarchy of tests](https://www.atlassian.com/continuous-delivery/software-testing/types-of-software-testing), but we already use "functional tests" in a different meaning.

docs/source/yosys_internals/extending_yosys/test_suites.rst
```diff
@@ -8,7 +8,43 @@ Running the included test suite
...
+to generate a makefile appropriate for the given directory, so only
+afterwards when make is invoked do the tests actually run.
+
+Usually their structure looks something like this:
+you write a .ys file that gets automatically run,
+which runs a frontend like ``read_verilog`` or ``read_rtlil`` with
+a relative path or a heredoc, then runs some commands including the command
+under test, and then uses :doc:`/using_yosys/more_scripting/selections`
+with ``-assert-count``. Usually it's unnecessary to "register" the test anywhere
+as if it's similar to other tests it will be run together with the rest.
```
KrystalDelusion: I think it's important to clarify that this is directory specific and doesn't apply when adding a new test directory.  Maybe worth talking instead about how (e.g.) `generate_mk --yosys-scripts` automatically finds all .ys files in the current directory, and that provided a directory is already running .ys tests it is sufficient to just add a new .ys file.

.github/workflows/test-sanitizers.yml
```diff
@@ -76,4 +76,4 @@ jobs:
       - name: Run unit tests
         shell: bash
         run: |
-          make -j$procs unit-test ENABLE_LIBYOSYS=1
+          make -j$procs unit-test
```
KrystalDelusion: Has something changed to make `ENABLE_LIBYOSYS=1` unnecessary?

KrystalDelusion (commented):
I'm unclear how the changes in this PR would have introduced an error while building `kernel/yosys.o`, or why it's trying to build `kernel/yosys.o` in the middle of the tests (possibly an issue with the yosys build artifact caching?)

Oh!  `make unit-test` requires `libyosys.so`, but we didn't make it because we didn't have `ENABLE_LIBYOSYS=1`

.github/workflows/test-sanitizers.yml
KrystalDelusion: I figured it out; `ENABLE_LIBYOSYS=1` adds `libyosys.so` to the `TARGETS` list, which is a prerequisite for the assorted `*-tests/%` make targets, while `unit-test` requires `libyosys.so` specifically rather than `$(TARGETS)`, so it doesn't actually need it in the list. 

When calling `make test` this does mean that the `vanilla-test` can start running prior to `libyosys.so` being built, which is a plus for parallelism, but a potential minus for clarity of building everything and *then* testing.

KrystalDelusion:
@widlarizer do the unit tests require all of the build dependencies to be available?  If so it may be better to split the unit tests out from `test-yosys` (currently it only gets test-deps+iverilog and skips build-deps, and I would ideally like to keep it that way for making sure we know which dependencies are build-only and which are needed for running)

widlarizer:
About calling it execution tests: I don't think that works - unit tests build an executable each that gets executed. I thought about "script tests" but functional tests are python scripts and makefile tests aren't plain bash scripts

KrystalDelusion:
> unit tests build an executable each that gets executed

I meant as in "execution of `yosys`" rather than any executable 🤷

mmicko:
@widlarizer Building unit tests out of tree is broken, updating it so location of unit test binaries will be inside tests/unit for both in and out of tree builds

mmicko (approved):

mmicko:
Have run checks with sanitizers to see if that flow is working since yml was altered https://github.com/YosysHQ/yosys/actions/runs/21980191191 and that is green as well

PR #5679, opened 2026-02-11
abc9: remove -liberty

widlarizer:
There was an internal ticket for this, it is a bit misleading to just pass it on if our `abc9` command isn't set up to make use of Liberty files. The change does make `abc9 -liberty foo` error out. Should `-genlib` stay or go as well?


Ravenslofty:
Honestly a decent portion of that line looks like "arg compatibility with `abc` that never manifested"; you can drop `-genlib`, but you might as well also drop `-constr` (SDC was never really integrated into the FPGA flows) and `-dont_use` (ASIC command that makes no sense in FPGA flows).

PR #5686, opened 2026-02-13
Automatic version bump

mmicko:
To be able to enforce that all is going on main using PRs, created a change in Makefile that will update version number to proper one, and in case of tarball we will have +post but still there will be good git commit in version string.

Note that there is no more then need for github action for bumping version, and also during release procedure, we still need to change `YOSYS_VER` which is fine, but also to comment line `YOSYS_VER := $(YOSYS_VER)+post` so we can distinct release and non-release builds, for New cycle commit we are going to uncomment this line.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -802,9 +802,30 @@ endif
 	$(Q) mkdir -p $(dir $@)
 	$(P) $(CXX) -o $@ -c $(CPPFLAGS) $(CXXFLAGS) $<
 
+YOSYS_REPO :=
+ifneq (, $(shell command -v git 2>/dev/null))
+ifneq (, $(shell git rev-parse --git-dir 2>/dev/null))
+	GIT_REMOTE := $(strip $(shell git config --get remote.origin.url 2>/dev/null | $(AWK) '{print tolower($$0)}'))
+	ifneq ($(strip $(GIT_REMOTE)),)
+		YOSYS_REPO := $(strip $(shell echo $(GIT_REMOTE) | $(AWK) -F '[:/]' '{gsub(/\.git$$/, "", $$NF); printf "%s/%s", $$(NF-1), $$NF}'))
+	endif
+	ifeq ($(strip $(YOSYS_REPO)),yosyshq/yosys)
+		YOSYS_REPO :=
+	endif
+	GIT_BRANCH := $(shell git rev-parse --abbrev-ref HEAD 2>/dev/null)
+	ifeq ($(filter main HEAD release/v%,$(GIT_BRANCH)),)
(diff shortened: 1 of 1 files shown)


KrystalDelusion:
Possible problem here with divergent branches.  Currently `main` is the source of truth for version numbers, and any time we see a version number we know that was the last version of main.  With this change we could have two different (e.g.) `v0.64+17` depending on which branch you're on.  Say somebody reports an issue with the `v0.64+14` release version, I've got `v0.64+17` installed for testing but it actually branched from `main` after `v0.64` and doesn't reproduce.  With the current versioning system I would see my version is behind the reported version and checkout latest main.  But if my version was higher than theirs I might instead think that the issue had been resolved and ignore it.  If both versions were the same I would see a difference in the sha, but if the versions are different then there is no way to know at a glance that they are from different branches.

I don't know what the solution here is (other than changing my workflow to use `PROGRAM_PREFIX` and using aliases to choose between `release-yosys` and `dev-yosys` or something), but for me personally I see this change as potentially doing more harm than good.

mmicko:
Sorry I did not respond to this earlier, but even current state is giving you similar (wrong) version number.
Currently version will update daily, so if it is +14 it will be like that after every commit during that day, and also users may send feedback pointing to a version and it will still be wrong. Valuable info in version string is git commit sha, and that points to actual version user is executing. So at the end conlcusion could maybe even be that we do not need number of commits after release tag, but still worth of discussion.

mmicko:
Added that it is always relative to mentioned tag, and also making sure it does not fail when we are doing release until tag is set (since that will be case on commit of bump, since tag comes after)

KrystalDelusion:
> Currently version will update daily, so if it is +14 it will be like that after every commit during that day, and also users may send feedback pointing to a version and it will still be wrong.

Yeah I guess that's true, and I don't think I've had too much issue with a user referencing a specific version but the problem was actually a commit after that but before the next version bump, so it might not be as much of a problem as I think.  Could we maybe add something to the version if we have git tracking but it isn't the main branch?  I think that would also solve my problem with the same version number on divergent branches.

mmicko:
Added repository and branch on it to be displayed in version string. 
Note that if repository is `YosysHQ/yosys` that it will be removed, and also if on main or release branch, so we do not need to alter anything when doing release on that end (since it is done from branch). So please do take a look

ShinyKate:
[sc-new-story]

ShinyKate:
[sc-new-story]

ShinyKate:
[sc-new-story]

ShinyKate:
[sc-new-story]

Issue #351, opened 2017-06-16
No history on error exit


kleinreact:
Is it desired that the call history is not saved if yosys exits through an error in interactive mode? I find this very annoying when testing/developing a script, as I then have to restart from scratch, where I need to retype all previous commands. 

In general, I do not understand why the interactive mode is exited on the error at all. Wouldn't it be more useful if just the error message is printed an yosys then returns into its last valid state (just before the erroneous command was executed)?

AlexDaniel:
Yes, this is what I never understood also. Why exit on error? There could be a reason for this, but it is pretty damn inconvenient…

cliffordwolf:
Internally in Yosys there are two functions that can be used to raise an error log_error() and log_cmd_error(). When the latter is used the caller guarantees that the internal state is still a valid state for a design to be in, and in interactive mode yosys will return to the command prompt in response to this call. However, I'm a lazy guy and when I can avoid thinking about something I usually do so, so in most cases I simply call log_error() instead and don't need to bother to think about the state I'll leave the design in.

In many cases it should be simple to convert code that calls log_error() to call log_cmd_error() instead. If you provide a list of candidates that annoy you the most I'll be happy to look into those cases.

Yes, that the history is not saved on error is annoying. I'll look into that.

kleinreact:
For me, the most annoying ones are `abc` and `write_aiger`. 

I have not checked the other `write` commands so far, but I think they should have high priority too, as they usually are used to test and debug a script.

cliffordwolf:
Commit f6421c8 converts the `abc` "clock domain not found" message to log_cmd_error(). Afaics that's the only candiate for that in the `abc` command, all the other calls to log_error() are in response to things like filesystem IO errors.

Commits c0ca994 and 1f517d2 add writing of history file in case of an unrecoverable error.

kleinreact:
Thanks for fixes so far. They already help a lot.

Regarding the IO errors, why exactly are they excluded of using `log_cmd_error`? Lets say I write something incomplete like this
```
read_verilog <file>
synth 
abc -script write_aiger
```
This triggers an error and I have to start from scratch. I mean, in general, it may always happen that one misses some command or gets something wrong by accident. Still, I wouldn't expect that the program completely crashes in interactive mode.

cliffordwolf:
> abc -script write_aiger

I don't think that will be an IO error in Yosys. We simply pass the script file name to ABC, which will then fail. In yosys that is caught as this error:

```
[...]

PR #4929, opened 2025-03-04
Update contributing rst with command/cell help info

KrystalDelusion:
_What are the reasons/motivation for this change?_

Developer documentation is lacking and/or out of date.  This PR serves to address how commands and cells should provide the help text that is given when calling `help`, and how this is then used for generating the documentation.  

_Explain how this is achieved._

- [x] write bullet points
- [ ] write paragraphs
- [ ] consider documenting the internal `help` options that output rst/json

_If applicable, please suggest to reviewers how they can test the change._

[Preview build contributing page](https://yosyshq.readthedocs.io/projects/yosys/en/docs-preview-contributing_help/yosys_internals/extending_yosys/contributing.html).

PR #5687, opened 2026-02-15
Update help text for rename -witness and write_aiger -ywmap [sc-220]

Goubermouche:
Updates help strings to reflect changes introduced in #4222, which added per-property tracking support for SBY's `abc --keep-going` pdr mode (also documented on the SBY side in [SBY #351](https://github.com/YosysHQ/sby/pull/351)):

- `rename -witness`: document renaming of formal property cells (`$assert`, `$assume`, `$cover`, `$live`, `$fair`, `$check`), which enables individual properties to be identified by hierarchical path in verification flows
- `write_aiger -ywmap`: document `"asserts"` and `"assumes"` arrays in the generated JSON map, which allow downstream tools to map AIGER bad-state outputs back to specific formal properties

### DIFF
--- a/docs/source/using_yosys/more_scripting/model_checking.rst
+++ b/docs/source/using_yosys/more_scripting/model_checking.rst
@@ -121,12 +121,12 @@ Result with fixed :file:`axis_master.v`:
 Witness framework and per-property tracking
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-When using AIGER-based formal verification flows (such as the `abc` engine in
+When using AIGER-based formal verification flows (such as the ``abc`` engine in
 SBY), Yosys provides infrastructure for tracking individual formal
 properties through the verification pipeline.
 
-The `rename -witness` pass (run automatically by `prep`) assigns public
-names to all cells that participate in the witness framework:
+The `rename -witness` pass assigns public names to all cells that participate in
+the witness framework:
 
 - Witness signal cells: `$anyconst`, `$anyseq`, `$anyinit`,
   `$allconst`, `$allseq`
@@ -139,7 +139,7 @@ their hierarchical path rather than by anonymous internal identifiers.
 The `write_aiger -ywmap` option generates a map file for conversion to and from
 Yosys witness traces, and also allows for mapping AIGER bad-state properties and
 invariant constraints back to individual formal properties by name. This enables
-features like per-property pass/fail reporting (e.g. `abc pdr` with
+features like per-property pass/fail reporting (e.g. ``abc pdr`` with
 ``--keep-going`` mode).
 
 The `write_smt2` backend similarly uses the public witness names when emitting


docs/source/using_yosys/more_scripting/model_checking.rst
```diff
@@ -117,3 +117,33 @@ Result with fixed :file:`axis_master.v`:
 
     Solving problem with 159144 variables and 441626 clauses..
     SAT proof finished - no model found: SUCCESS!
+
+Witness framework and per-property tracking
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When using AIGER-based formal verification flows (such as the ``abc`` engine in
+SymbiYosys), Yosys provides infrastructure for tracking individual formal
```
KrystalDelusion: SymbiYosys as a name is deprecated, this should be SBY instead.

docs/source/using_yosys/more_scripting/model_checking.rst
```diff
@@ -117,3 +117,33 @@ Result with fixed :file:`axis_master.v`:
...
+properties through the verification pipeline.
+
+The ``rename -witness`` pass (run automatically by ``prep``) assigns public
```
KrystalDelusion: Cells and passes should use single backticks for automatic linking (passes with options won't link, but they will get syntax highlighting).  See [Working with Sphinx](https://yosyshq.readthedocs.io/projects/yosys/en/docs-preview-contributing_help/yosys_internals/extending_yosys/build_docs.html#working-with-sphinx) preview from #4929.

passes/cmds/rename.cc
```diff
@@ -263,9 +263,14 @@ struct RenamePass : public Pass {
 		log("    rename -witness\n");
 		log("\n");
 		log("Assigns auto-generated names to all $any*/$all* output wires and containing\n");
-		log("cells that do not have a public name. This ensures that, during formal\n");
-		log("verification, a solver-found trace can be fully specified using a public\n");
-		log("hierarchical names.\n");
+		log("cells that do not have a public name. Also renames formal property cells\n");
+		log("($assert, $assume, $cover, $live, $fair, $check) that have private names,\n");
+		log("giving them public witness-trackable names.\n");
```
KrystalDelusion: "Assigns auto-generated names ... that not have a public name" is the same thing as "renames ... that have private names".  I would recommend combining the lists instead of making a new one, e.g.

> Assigns auto-generated names to objects used in formal verification that do not have a public name. This applies to all formal property cells ($assert, $assume, $cover, $live, $fair, $check), $any*/$all* output wires, and their containing cells.

backends/aiger/aiger.cc
```diff
@@ -931,6 +931,11 @@ struct AigerBackend : public Backend {
 		log("\n");
 		log("    -ywmap <filename>\n");
 		log("        write a map file for conversion to and from yosys witness traces.\n");
+		log("        The generated JSON map includes \"asserts\" and \"assumes\" arrays\n");
+		log("        containing the hierarchical witness paths of the corresponding\n");
+		log("        $assert and $assume cells. This enables downstream tools to map\n");
+		log("        AIGER bad-state properties and invariant constraints back to\n");
+		log("        individual formal properties by name.\n");
```
KrystalDelusion: I'm not sure about this one.  There are other things in the map that we don't call out, so I'm not sure if we should be calling out the asserts/assumes.  But then I suppose the rest of the file is intended for the yosys witness tools (e.g. `aiw2yw`), which is covered by the "conversion to and from yosys witness traces."  Maybe just reduce to

> also allows for mapping AIGER bad-state properties and invariant constraints back to individual formal properties by name

and drop the mention of JSON arrays and $assert/$assume cells

docs/source/using_yosys/more_scripting/model_checking.rst
```diff
@@ -121,29 +121,29 @@ Result with fixed :file:`axis_master.v`:
 Witness framework and per-property tracking
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-When using AIGER-based formal verification flows (such as the ``abc`` engine in
-SymbiYosys), Yosys provides infrastructure for tracking individual formal
+When using AIGER-based formal verification flows (such as the `abc` engine in
```
KrystalDelusion: Ah, this one isn't referring to the command, so it should be in double quotes (or all caps).  Same for `abc pdr` below, that's an SBY option, not a Yosys command.

docs/source/using_yosys/more_scripting/model_checking.rst
```diff
@@ -121,29 +121,29 @@ Result with fixed :file:`axis_master.v`:
...
+SBY), Yosys provides infrastructure for tracking individual formal
 properties through the verification pipeline.
 
-The ``rename -witness`` pass (run automatically by ``prep``) assigns public
+The `rename -witness` pass (run automatically by `prep`) assigns public
```
KrystalDelusion: Unless I'm missing it, `rename -witness` is called by the prep stage of SBY, but *not* the `prep` pass.  I would probably drop the reference to it being run automatically entirely (since this is the Yosys docs).

docs/source/using_yosys/more_scripting/model_checking.rst
```diff
@@ -121,29 +121,29 @@ Result with fixed :file:`axis_master.v`:
...
-The ``write_aiger -ywmap`` option generates a JSON map file that includes, among
+The `write_aiger -ywmap` option generates a JSON map file that includes, among
 other things, ``"asserts"`` and ``"assumes"`` arrays.  Each entry contains the
-hierarchical witness path of the corresponding ``$assert`` or ``$assume`` cell.
-This lets tools such as SymbiYosys map AIGER bad-state properties and invariant
+hierarchical witness path of the corresponding `$assert` or `$assume` cell.
+This lets tools such as SBY map AIGER bad-state properties and invariant
 constraints back to individual formal properties, enabling features like
-per-property pass/fail reporting (e.g. ``abc pdr`` with ``--keep-going`` mode).
+per-property pass/fail reporting (e.g. `abc pdr` with ``--keep-going`` mode).
```
KrystalDelusion: See above

KrystalDelusion (changes requested):
A couple more minor things, the help text changes are good though :)

Issue #5690, opened 2026-02-17
wreduce reduces used, driven top bit


widlarizer:
### Version

8bbde80e

### On which OS did this happen?

Linux

### Reproduction Steps

Get `nerv_synth.v` from [here](https://github.com/YosysHQ/eqy/blob/main/examples/nerv/nerv_synth.v)
```
read_verilog -formal -icells nerv_synth.v
proc
hierarchy
opt_clean
dump w:*reg_q*
show -prefix pre w:*reg_q* %ci %co
wreduce
dump w:*reg_q*
show w:*reg_q* %ci %co
```

### Expected Behavior

No bits are removed from `mem_rd_reg_q`

### Actual Behavior

`Removed top 1 bits (of 5) from wire nerv.mem_rd_reg_q.`

This behavior is exposed since #5621 and [makes eqy CI fail](https://github.com/YosysHQ/eqy/pull/90) probably due to the connectivity change causing different partitioning (extra partition) together and the missing bit on a public wire can't be matched and isn't proven equivalent between gold and gate

widlarizer:
`wreduce` relies on the design being sorted, otherwise it breaks the `std::map` sortedness invariant of `ModIndex::database`. Horrifying

widlarizer:
Almost all synth passes sequence it right after opt so the design sort in opt_clean was the load-bearing one

PR #5691, opened 2026-02-17
modtools, wreduce: check_db sanity check

widlarizer:
Exposes #5690 `ModIndex` database insanity. Some tests fail this assert, which seems alarming


widlarizer:
No longer needed to demonstrate the problem since I added unit tests to #5692

PR #5692, opened 2026-02-18
modtools: fix database sanity

### LINKED ISSUE #5690: wreduce reduces used, driven top bit

widlarizer:
Fixes #5690. Erasing the sigbit entry entirely if no data about sigbit is required for the check to pass (for the index to track the current design as-is). Using pointer ordering for sigbits in `ModIndex` is required if swap_names is supposed to be legal while holding a `ModIndex`, otherwise `std::map` will fail to find out-of-order entries even though it can iterate over them. Both are covered with unit tests that fail on main

### DIFF
--- a/kernel/modtools.h
+++ b/kernel/modtools.h
@@ -151,11 +151,11 @@ struct ModIndex : public RTLIL::Monitor
 		}
 	}
 
-	void check()
+	bool ok()
 	{
 #ifndef NDEBUG
 		if (auto_reload_module)
-			return;
+			return true;
 
 		for (auto it : database)
 			log_assert(it.first == sigmap(it.first));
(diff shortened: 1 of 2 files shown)


mmicko (approved):
Fixes EQY test case that pointed to an issues

PR #5694, opened 2026-02-19
memory_libmap: Add -force-params [sc-273]

Ravenslofty:

(This is a patch that @KrystalDelusion wrote for `synth_analogdevices` that I thought should be pulled out of that patch series before it gets PR'd.)

It can be a bit unergonomic to configure memories when the intermediate cell generated by `memory_libmap` does not have certain size parameters.

Add a `-force-params` option to always instantiate these size parameters for more uniform mapping handling.

This is admittedly a little awkward, since these are used as part of `synth_analogdevices` and as such that flow would have the testcases in question.


KrystalDelusion:
Oop, thanks for splitting this and opening the PR 💜

PR #5696, opened 2026-02-19
gowin: Fix bram ADA byte enables

### LINKED ISSUE #5688: gowin: $_GOWIN_DP brams_map uses PORT_B_WR_BE for ADA byte enables instead of PORT_A_WR_BE, silently disabling   writes
### Version

Yosys 0.62 (git sha1 7326bb7d6641500ecb285c291a54a662cb1e76cf, clang++ 17.0.0 -fPIC -O3

macOS

### Reproduction Steps

Sample project:

```v
// Minimal reproduction of $__GOWIN_DP_ byte-enable bug in yosys brams_map.v
//
// Bug: In gowin/brams_map.v line 208, the $__GOWIN_DP_ template computes:
//   wire [13:0] ADA = `addrbe(PORT_A_WIDTH, PORT_A_ADDR, PORT_B_WR_BE);
// This should be PORT_A_WR_BE. Because PORT_B_WR_BE is used, the byte
// enables for the write port (Port A) are taken from the read port (Port B),
// which are always 0. This silently disables all writes to the BSRAM.
//
// To reproduce:
//   yosys -p "read_verilog yosys_dp_byte_enable_bug.v; synth_gowin -top top; write_verilog post_synth.v"
//
// Then inspect post_synth.v: the DPX9B instance will have ADA[1:0] tied to
// constant 0 (byte enables disabled), meaning writes never take effect.
//
// Expected: ADA[1:0] should be driven by the write-enable condition so that
// byte lanes are enabled when a write occurs.

`default_nettype none

module top (
    input  wire        clk,
    input  wire [9:0]  rd_addr,
    output reg  [15:0] rd_data,
    input  wire [9:0]  wr_addr,
    input  wire [15:0] wr_data,
    input  wire        wr_en
);

    (* ram_style = "block" *) reg [15:0] mem [0:1023];

    // Read port — separate always block
    always @(posedge clk) begin
        rd_data <= mem[rd_addr];
    end

    // Write port — separate always block
    always @(posedge clk) begin
        if (wr_en)
            mem[wr_addr] <= wr_data;
    end

endmodule
```

**Description:**
When using synth_gowin on a simple dual-port memory (separate read and write always blocks), yosys maps the memory to $__GOWIN_DP_ and then to the DPX9B primitive. In the $__GOWIN_DP_ template (gowin/brams_map.v line 208), the byte-enable bits for Port A's address (ADA[1:0]) are computed using PORT_B_WR_BE instead of PORT_A_WR_BE. Since Port B is the read port, its write byte-enables are always zero, causing ADA[1:0] to be tied to GND. This silently disables all byte lanes on the write port, so writes to the BSRAM never take effect. The memory appears to work for reads of initialized data but any runtime writes are silently discarded.

### Expected Behavior

ADA[1:0] (byte enables) should be driven by the write-enable condition, so that byte lanes are active when a write occurs on Port A. Specifically, when PORT_A_WIDTH == 18, ADA[1:0] should be PORT_A_WR_BE[1:0] (both bits set to 1 when writing), not PORT_B_WR_BE[1:0] (which is always 0 since Port B is the read port).

### Actual Behavior

ADA[1:0] (byte enables) are tied to GND (constant 0) because they are sourced from PORT_B_WR_BE — the read port's write byte-enables, which are always 0. This silently disables all write operations to the BSRAM. Data written to memory is discarded with no warning or error.

KrystalDelusion:
Close #5688.

Fix `PORT_B_WR_BE` in `ADA` to be `PORT_A_WR_BE`.

Test from issue included as `tests/arch/gowin/bug5688.ys`.  Fails without this change (so long as the `splitnets` is there, otherwise `wr_en` ends up in the input cone for some of the other bits), passes with.

### DIFF
--- a/tests/arch/gowin/bug5688.ys
+++ b/tests/arch/gowin/bug5688.ys
@@ -1,0 +1,31 @@
+read_verilog << EOT
+`default_nettype none
+
+module top (
+    input  wire        clk,
+    input  wire [9:0]  rd_addr,
+    output reg  [15:0] rd_data,
+    input  wire [9:0]  wr_addr,
+    input  wire [15:0] wr_data,
+    input  wire        wr_en
+);
+
+    (* ram_style = "block" *) reg [15:0] mem [0:1023];
+
+    // Read port — separate always block
+    always @(posedge clk) begin
+        rd_data <= mem[rd_addr];
+    end
+
+    // Write port — separate always block
+    always @(posedge clk) begin
+        if (wr_en)
+            mem[wr_addr] <= wr_data;
+    end
+
+endmodule
+
+EOT
+synth_gowin -top top
+splitnets
+select -assert-any top/mem.0.0 %ci*:+DPX9B[ADA]:+DFF:+IBUF i:wr_en %i


gatecat:
This clearly fixes a bug that I wasted some time on, making rare use of my commit powers and merging :)

PR #5699, opened 2026-02-20
Pre-generate celltype lookup tables to fix LTO compile time regression

Goubermouche:
#5512 implements constexpr lookups for celltype associated calls, leading to ~10% perf increase in some instances, but produces drastically longer compile times with LTO enabled. This PR aims to remedy the compile time issue while retaining the runtime performance increase. 

This implementation relies on a generator script (`misc/gen_celltypes.py`) to generate lookup tables once, instead of for every unit. Using this impl I achieve similar runtime characteristic as the original PR, with no visible compile time tradeoffs.

TODO:
- [ ] Unit tests  to verify new celltypes match the old implementation.

[sc-275]

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -568,7 +568,7 @@ $(subst //,/,$(1)/$(3)): $(2)
 	$$(Q) cp "$(YOSYS_SRC)"/$(2) $(subst //,/,$(1)/$(3))
 endef
 
-define add_gen_share_fila
+define add_gen_share_file
 EXTRA_TARGETS += $(subst //,/,$(1)/$(notdir $(2)))
 $(subst //,/,$(1)/$(notdir $(2))): $(2)
 	$$(P) mkdir -p $(1)


hexratcc:
Going to get started on unit testing...

widlarizer:
- [ ] try having all the tables in a header and seeing how that affects build time and run time
- [ ] prefer `pathlib` over `os`

kernel/celltypes.h
```diff
@@ -160,8 +160,11 @@ struct CellTypes
 		return cell_types.count(type) != 0;
 	}
 
-	bool cell_output(const RTLIL::IdString &type, const RTLIL::IdString &port) const
+bool cell_output(const RTLIL::IdString &type, const RTLIL::IdString &port) const
 	{
+		auto it = cell_types.find(type);
```
widlarizer, line 165: Doesn't moving the expensive dict lookup before the array access static builtin lookup completely negate the point of the PR?

hexratcc: Correct, I was just trying to see if the erroneous tests are caused by lookups or something else.

widlarizer: You could switch to the previous "newcelltypes" approach where the old system is initially completely retained, compared in a test to the new one, and then the new one is rolled out and the old ones removed. Also, are you using the CI to run tests? Is it not way slower than running them locally for you?

hexratcc:
Closing as this PR is no longer relevant (see #5512).

PR #5700, opened 2026-02-20
CI: WASI - Applying YoWASP changes to script 

mmicko:
It affects linking speed so it is overall improvement on build time for WASI

### DIFF
--- a/.github/workflows/extra-builds.yml
+++ b/.github/workflows/extra-builds.yml
@@ -103,6 +103,7 @@ jobs:
           ENABLE_ZLIB := 0
 
           CXXFLAGS += -I$(pwd)/flex-prefix/include
+          LINKFLAGS += -Wl,-z,stack-size=8388608 -Wl,--stack-first -Wl,--strip-all
           END
 
           make -C build -f ../Makefile CXX=clang -j$(nproc)


whitequark:
I would be very surprised if this change resulted in a speedup. What it *will* change is to ensure that the resulting binary does not crash on input files with high levels of AST nesting. (Only a few levels of nesting can overflow the stack with the default 64K stack size.)

Also `--stack-first` causes stack overflows to result in a crash rather than silent memory corruption.

mmicko:
@whitequark Noticed on test-action repo, where even for SMALL=1 build it took quite a lot of time for linking, this change halfed time spent on it. Would like to see if it performs same for full build. Even if not it is good to be synced with upstream, in that case would change description.

whitequark:
> Noticed on test-action repo, where even for SMALL=1 build it took quite a lot of time for linking, this change halfed time spent on it.

Oh, I guess LTO somehow makes relooper faster? Funky.

mmicko:
Looks like with LTO now there is 11 min for linking and previously it was instant. But on test-actions branch it was 14min link time without a change and then 15sec with, so very strange. Will wait for complete flow to finish, but it looks like I will close this one and leave as is.

whitequark:
> Will wait for complete flow to finish, but it looks like I will close this one and leave as is.

Please do leave the stack-related options in: they are extremely important for correctness. The rest is up to you. But yes, the 11 min link time is mostly due to a Wasm-specific pass (relooper) and I avoid LTO except for release builds.

Issue #5636, opened 2026-01-26
Possible `sim -r`/anyseq/VCD bug


gussmith23:
This is a tracking issue to document my rooting out of a possible bug when replaying traces (perhaps specifically VCDs) on designs with undriven outputs.

PR #5701, opened 2026-02-20
Add warnings and errors to `sim -r` with VCD code path

gussmith23:
We have been inspecting the sharp edges of the `sim` pass (see #5636). This change makes one of the sharp edges a little more visible: specifically, the behavior around undriven wires.

Currently, undriven wires will not be driven when replaying from a VCD, which is different from the behavior when replaying with a YW file. This sharp edge is often avoided entirely as flows will convert undriven wires to `$anyseq`, which _will_ be driven during VCD replay. However, if one forgets to do this, they will get unexpected simulation behavior.

This change adds a check + an  error (downgradeable to a warning, or entirely skippable) for undriven wires, during VCD simulation.

This check is implemented using DriverMap to check for undriven signals. Given that this may not be performant for large designs, we also add a flag to disable the check entirely.


passes/sat/sim.cc
```diff
@@ -426,7 +429,7 @@ struct SimInstance
 
 		Const value = builder.build();
 		if (shared->debug)
-			log("[%s] get %s: %s\n", hiername(), log_signal(sig), log_signal(value));
+			log("[%s] get %s: %s\n", hiername(), log_signal(sig, true), log_signal(value, true));
```
mmicko: is adding `true` as second parameter for `log_signal` affecting results log output at all ? since this should affect only constants, if so do you have any example for that ?

passes/sat/sim.cc
```diff
@@ -1192,6 +1195,54 @@ struct SimInstance
...
+					if (bit.wire != nullptr && drivermap(DriveBit(bit)).is_none())
+						undriven.append(bit);
+
+				undriven.sort_and_unify();
+				if (undriven.empty())
+					continue;
+
+				issue_count++;
+				std::string wire_name = scope + "." + RTLIL::unescape_id(wire->name);
+				log_warning("Input trace contains undriven signal `%s` (%s); values for this signal are not replayed from FST/VCD input.\n",
```
mmicko: Too long message and it will repeat if there are multiple signals, maybe add a flag for having undriven signals in file and then just display one more warning at the and with something like `Values for mentioned signal(s) are not replayed from FST/VCD input.\n`

passes/sat/sim.cc
hexratcc: The original includes` kernel/drivertools.h`, which expose `log_signal` overloads which weren't in the original, confusing the compiler. The added `true` gets rid of the ambiguity and the original `log_signal` is used.
```cpp
std::string log_signal(const RTLIL::SigSpec &sig, bool autoint = true);
```

gussmith23:
(thank you all for taking this over!)

PR #5702, opened 2026-02-23
Check verific configurations

mmicko:
This makes yosys to compile with various verific configurations, deleting built object files to prevent rebuilding that is not needed.


mmicko:
`config-clang` target also executes `clean` but since we are using `ccache` it does not affect built time. Most of time goes to linking since LTO is enabled. Better we make this separate workflow and make it execute only only on `workflow_dispatch` so manually when required.

PR #5704, opened 2026-02-24
abc9: preserve topological-loop asserts with targeted SCC fallback

apullin:

  This came up in a real synthesis run. After normal ABC9 SCC handling, a residual non-box combinational loop could still survive and trip the `no_loops` assertion in `prep_xaiger()`.

  The topology assertions are important (especially around boxes), so this PR keeps those assertions and fixes the residual-loop path instead of relaxing assertions.

  - `passes/techmap/abc9.cc`
  - use `scc -all_cell_types -specify -set_attr abc9_scc_id {}` in `map`
  - also run the same SCC tagging in `pre`, before `abc9_ops -break_scc`
  - `passes/techmap/abc9_ops.cc`
  - keep `log_assert(no_loops)` in `prep_xaiger()` and `reintegrate()`
  - if topo sort still reports loops in `prep_xaiger()`, insert extra `$__ABC9_SCC_BREAKER` cuts on non-box loop cells, rebuild topo sort, then re-check the same assertion
  - use a shared topo-build helper and conditional `TopoSort` reinit for the retry path (same logic path, no pointer-switching flow)

  This PR is scoped to ABC9 flow only (no XAIGER backend changes - had them initially, but now removed).

  Added regression test:
  - `tests/techmap/abc9-nonbox-loop-with-box.ys`

  The test builds a small non-box feedback loop in a design that also contains an `(* abc9_box *)` instance, and checks that `abc9 -lut 4` completes.

  Quick check:
  - `./yosys tests/techmap/abc9-nonbox-loop-with-box.ys`

### DIFF
--- a/passes/techmap/abc9_ops.cc
+++ b/passes/techmap/abc9_ops.cc
@@ -23,6 +23,7 @@
 #include "kernel/utils.h"
 #include "kernel/celltypes.h"
 #include "kernel/timinginfo.h"
+#include <optional>
 
 USING_YOSYS_NAMESPACE
 PRIVATE_NAMESPACE_BEGIN
@@ -587,6 +588,7 @@ void break_scc(RTLIL::Module *module)
 		auto id = it->second;
 		auto r = ids_seen.insert(id);
 		cell->attributes.erase(it);
+		// Cut exactly one representative cell per SCC id.
 		if (!r.second)
 			continue;
 		for (auto &c : cell->connections_) {
@@ -710,8 +712,6 @@ void prep_xaiger(RTLIL::Module *module, bool dff)
 
 	SigMap sigmap(module);
 
-	dict<SigBit, pool<IdString>> bit_drivers, bit_users;
-	TopoSort<IdString, RTLIL::sort_by_id_str> toposort;
 	dict<IdString, std::vector<IdString>> box_ports;
 
 	for (auto cell : module->cells()) {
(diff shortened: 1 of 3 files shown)


passes/techmap/abc9_ops.cc
```diff
@@ -792,7 +789,8 @@ void prep_xaiger(RTLIL::Module *module, bool dff)
 		}
 	}
 
-	log_assert(no_loops);
+	if (!no_loops)
+		log_warning("Module '%s' contains combinational loops after SCC breaking. Continuing with deterministic non-topological ordering.\n", log_id(module));
```
Ravenslofty: My understanding is that parts of the ABC9 flow require topological ordering, particularly around boxes. So I'm not comfortable with removing this assertion.

apullin: Meant to put this up as a draft, it is premature.

apullin:
Reworked to preserve the assertions.

The previous “remove/relax assert” approach is gone. The current version keeps the existing `no_loops` assertions in place in both `prep_xaiger()` and `reintegrate()`.

What changed instead:
- Keep normal SCC breaking behavior intact.
- Add a targeted fallback in `prep_xaiger()`: if loops still remain, insert additional `$__ABC9_SCC_BREAKER` cuts only on non-box loop cells, rebuild topo order, then run the existing assertion checks.
- Keep pre-ABC9 SCC tagging with `scc -all_cell_types`.

So the topology assumptions are still enforced; the fix is to handle residual non-box cycles that can survive the regular SCC path in real synthesis flows.

Also added a regression test:
- `tests/techmap/abc9-nonbox-loop-with-box.ys`

I re-ran the reproducer flow and this no longer reproduces.

apullin:
Follow-up after re-testing: I removed the `backends/aiger/xaiger.cc` change and the `tests/techmap/xaiger-comb-loop.ys` test from this PR.

This branch is now scoped to the ABC9-side fix only (`abc9.cc` / `abc9_ops.cc` + `abc9-nonbox-loop-with-box.ys`). I re-ran the oracle/reproducer flow and it still does not reproduce.

Title changed to "abc9: preserve topological-loop asserts with targeted SCC fallback".

I had initially suspected XAIGER export-loop handling was part of this same failure, but that was a misattribution for this case. A separate XAIGER combinational-export loop failure mode does still exist and can be addressed independently in a follow-up PR.

passes/techmap/abc9_ops.cc
```diff
@@ -585,8 +584,8 @@ void break_scc(RTLIL::Module *module)
 		auto it = cell->attributes.find(ID::abc9_scc_id);
 		log_assert(it != cell->attributes.end());
 		auto id = it->second;
-		auto r = ids_seen.insert(id);
 		cell->attributes.erase(it);
+		auto r = ids_seen.insert(id);
```
Ravenslofty, line 589: (Why was this changed? Seems kind of unnecessary.)

passes/techmap/abc9_ops.cc
```diff
@@ -750,39 +747,96 @@ void prep_xaiger(RTLIL::Module *module, bool dff)
...
+				for (auto driver_cell : bit_drivers.at(it.first))
+					for (auto user_cell : it.second)
+						toposort.edge(driver_cell, user_cell);
+		if (ys_debug(1))
+			toposort.analyze_loops = true;
+		return toposort.sort();
+	};
+
+	TopoSort<IdString, RTLIL::sort_by_id_str> toposort_initial;
+	TopoSort<IdString, RTLIL::sort_by_id_str> toposort_retry;
```
Ravenslofty: I don't think you need two toposorts here, do you? If the first fails, just replace it. Then you can also get rid of a few small bits of code churn caused by turning things into pointers.

passes/techmap/abc9_ops.cc
apullin: whoops, rebase thrash from pulling out the stuff touching xagier, will undo

passes/techmap/abc9_ops.cc
apullin: My reading was that TopoSort was stateful, so this was for correctness. Will rework to do conditional reinit.

apullin: actually, it looks like there is no clear reset/reinit for TopoSort. So we either need two, or conditionally build the second one. Rework uses std::optional and a builder to do this with a single location through the process.

Ravenslofty:
You'll have to re-bless the testcase numbers, then I'll take another look.

apullin:
Testcase counts updated. Also added a few comments.

Ravenslofty:
QoR change looks maybe okay? I don't want to drag the contribution process out too much further, but I suspect the QoR change implies this PR is being a bit more conservative than necessary? Something to think about, perhaps.

apullin:
whoops I didn't get all the test updates right, should be green now, I tested the workflows in my fork

PR #5710, opened 2026-02-26
Preserve \src attributes through ABC synthesis passes

robtaylor:
## Summary

- Propagate `\src` attributes through classic ABC and ABC9 synthesis passes using ABC's node retention feature
- Classic ABC pass: retained node IDs are emitted in BLIF; on read-back, `\src` attributes are restored from the retention map
- ABC9 pass: `&write_retention` exports GIA node mappings used to restore `\src` on cells after optimization
- Test suite validates attribute survival through both pipelines

## Dependencies

This PR depends on [YosysHQ/abc#41](https://github.com/YosysHQ/abc/pull/41) which adds the node retention infrastructure to ABC. The ABC submodule reference will need updating to point to the merged ABC commit once that PR lands.

## Test plan

- [ ] CI retention validation workflow passes
- [ ] Classic ABC: `\src` attributes survive synthesis round-trip
- [ ] ABC9: `\src` attributes survive synthesis round-trip
- [ ] Sideband invariant tests pass for full PnR-like flow

### DIFF
--- a/.github/workflows/test-src-retention.yml
+++ b/.github/workflows/test-src-retention.yml
@@ -1,0 +1,290 @@
+name: Test src retention through ABC
+
+on:
+  push:
+    branches:
+      - src_retention_abc
+      - src_retention_abc9
+  pull_request:
+    branches:
+      - master
+      - src_retention_abc
+  workflow_dispatch:
+
+jobs:
+  test-src-retention:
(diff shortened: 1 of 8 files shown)


robtaylor:
Closing pending plan re abc XAIGER alignment

Issue #487, opened 2018-01-17
“converting real value” error message can be more helpful


AlexDaniel:
In some cases yosys produces a warning like this:
```
Warning: converting real value 7.960000e+01 to binary 80 at foo.sv:19.
```
The warning is correct and I expected that it would do that without even warning me, although I don't mind. But what would be the most correct way to silence this warning? How can I tell it “yeah, just round the value to the closest integer and use that, I mean it”? And if there is an easy way to do that, can we change the warning to recommend that?

I got this warning when I was initializing block ram in the `initial` block where I had some `integers` which were then put into the block ram. I think that's a valid situation to do conversion like this.

cliffordwolf:
Use the standard Verilog `$rtoi()` system function to convert reals to ints without warning. (Note: This converts by "truncating" the real part, not by rounding to the next integer.)

AlexDaniel:
But the default behavior does the rounding to the next integer, how to get that behavior? Or am I misunderstanding something?

Also, the proposal was to change the error message. Something like ``Warning: converting real value 7.960000e+01 to binary 80 at foo.sv:19 (use $rtoi to hide the warning).``… and if that is rejected, why?

RossBencina:
[Apologies for length of comment, I want to spell everything out. Proposed workaround at end.]

I have the same problem. If I understand correctly, the requirement is to be able to write elaboration-time constant code like the following that performs an *intentional round-to-nearest operation without receiving a warning*. For me, ideally the code should be portable (since I want to keep using icecube2 as a cross-check). This is the code that causes a warning:

```
local real a = 2;
local real b = 3;
local integer c = a/b; // intentional round-to-nearest, expect c == 1
```

The above works, but Yosys gives a warning like:

```
Warning: converting real value 6.666667e-01 to binary 1.
```

Indeed Synplify Pro warns similarly, but not always: it didn't warn for a=2, b=3, but did warn for a=25000000, b=115200:

```
Rounding real from 217.013889 to 217 (simulation mismatch possible)
```

As I understand it, the "modern" way to do this in SystemVerilog is with an integer cast:

```
local integer c = int'(a/b); // integer cast produces intentional round-to-nearest
```

In Yosys the above code gives:

```
ERROR: Static cast is only supported in SystemVerilog mode.
```
and Synplify Pro fails with a parsing error.

[...]

PR #5712, opened 2026-02-27
Preserve \src attributes through ABC9 via XAIGER y extension

robtaylor:
## Summary

Preserve `\src` (source-location) attributes on LUT cells through the ABC9 synthesis flow using the XAIGER "y" extension.

- **XAIGER writer** (`backends/aiger/xaiger.cc`): emits "y" identity extension + `src` map entries in the map file
- **XAIGER readers** (`frontends/aiger/aigerparse.cc`, `frontends/aiger2/xaiger.cc`): parse "y" extension + src map, apply `\src` to mapped LUT cells
- **abc9_exe**: simplified to use `&verify` (without `-y`) for correctness checking only
- **aigmap**: propagates `\src` from replaced cells to new `$_AND_`/`$_NOT_`/`$_NAND_` gates

### How it works

Yosys writes an XAIGER file with a "y" extension containing an identity mapping (each object maps to itself) plus a map file with `src` lines recording which AIG objects have `\src` attributes. ABC reads this, propagates the origin mapping through all optimization passes, and writes the updated mapping back. Yosys then reads the output and uses the mapping to look up the original `\src` for each LUT cell.

### Dependency: ABC-side origin tracking

This PR depends on [berkeley-abc/abc#487](https://github.com/berkeley-abc/abc/pull/487), which adds per-object origin tracking (`vOrigins`) to ABC's `Gia_Man_t`.

Without that change, ABC has no mechanism to propagate the "y" mapping through its optimization passes. The identity mapping Yosys writes becomes meaningless after ABC restructures the AIG — objects are renumbered, merged, and eliminated by passes like `&dc2`, `&dch`, and the IF mapper. We initially prototyped a `&verify -y` approach that used combinational equivalence checking to reconstruct the mapping *after* optimization, but this only achieved 17-57% coverage on non-trivial designs and doesn't work for sequential circuits.

The ABC PR adds a lightweight `vOrigins` vector that tracks each object's input-side origin as it flows through dup operations, structural hashing, AIG round-trips, and technology mapping — achieving ~100% coverage with negligible overhead (~0.6% time, ~3% memory).

**This PR should not be merged until the ABC maintainer has reviewed and accepted the ABC-side changes.** The approach and feasibility depend on their feedback.

### Coverage

With the ABC origin tracking in place, `\src` retention on LUT cells achieves **100%** on all tested designs (simple combinational, Amaranth-style, and larger multi-output designs with 54 LUTs). Without it, coverage was only 17-57%.

### Context

Follows feedback from @Ravenslofty on YosysHQ/abc#41 suggesting the XAIGER "y" extension approach instead of custom node retention infrastructure in ABC.

cc @Ravenslofty — would appreciate your input on this approach. Note the ABC-side dependency at [berkeley-abc/abc#487](https://github.com/berkeley-abc/abc/pull/487).

## Test plan

- [x] Basic `.ys` test: `\src` survives simplemap → abc9 round-trip
- [x] Full validation: JSON export confirms `$lut` cells have valid `file:line.col` `\src` format
- [x] Existing `abc9.ys` test suite passes (no regressions)
- [x] `aigmap` `\src` propagation verified for AND/NOT/NAND paths
- [x] 100% `\src` coverage on simple, amaranth, and large (54 LUT) test designs
- [x] Negligible performance overhead (~0.6% time, ~3% memory on picorv32)

### DIFF
--- a/frontends/aiger/aigerparse.cc
+++ b/frontends/aiger/aigerparse.cc
@@ -429,6 +429,7 @@ void AigerReader::parse_xaiger()
 				log_assert(output_sig);
 				uint32_t nodeID;
 				RTLIL::SigSpec input_sig;
+				std::vector<int> input_node_ids;
 				for (unsigned j = 0; j < cutLeavesM; ++j) {
 					nodeID = parse_xaiger_literal(f);
 					log_debug2("\t%u\n", nodeID);
@@ -439,6 +440,7 @@ void AigerReader::parse_xaiger()
 					RTLIL::Wire *wire = module->wire(stringf("$aiger%d$%d", aiger_autoidx, nodeID));
 					log_assert(wire);
 					input_sig.append(wire);
+					input_node_ids.push_back(nodeID);
 				}
 				// Reverse input order as fastest input is returned first
 				input_sig.reverse();
@@ -461,6 +463,7 @@ void AigerReader::parse_xaiger()
 				auto *lut = module->addLut(stringf("$lut$aiger%d$%d", aiger_autoidx, rootNodeID), input_sig, output_sig, std::move(lut_mask));
 				// Track for \src application via "y" mapping
 				lut_by_obj[rootNodeID] = lut;
+				lut_input_objs[rootNodeID] = std::move(input_node_ids);
 			}
 		}
 		else if (c == 'r') {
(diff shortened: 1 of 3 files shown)


Ravenslofty:
Including bbdfd0f48e376ea4f558516fb88df264955ef4f2 in this PR seems premature to me without knowing what Alan thinks, and even if a better approach than `&verify -y` is developed, `&verify -y` is a thing that exists right now and still provides results that are better than nothing.

robtaylor:
> Including [bbdfd0f](https://github.com/YosysHQ/yosys/commit/bbdfd0f48e376ea4f558516fb88df264955ef4f2) in this PR seems premature to me without knowing what Alan thinks, and even if a better approach than `&verify -y` is developed, `&verify -y` is a thing that exists right now and still provides results that are better than nothing.

that's why it's draft =) 
Just keeping you in the loop. Let's see what Alan thinks.

Ravenslofty:
Well, we know what Alan thinks now.

I think you should revert https://github.com/YosysHQ/yosys/commit/bbdfd0f48e376ea4f558516fb88df264955ef4f2 because, as mentioned, `&verify -y` is an approach we can get into Yosys *now*, and it still provides an improvement over the present state. You can always make a follow-up PR to extend it further if a better approach is hashed out with Alan.

robtaylor:
Progress update on the ABC side (https://github.com/berkeley-abc/abc/pull/487):

**Origin propagation through all ABC9 engines is now implemented and reviewed.** Alan Mishchenko has reviewed the approach and confirmed the design is clean. The engines instrumented are: `&dc2`, `&dch`, `&if`, `&syn2`, `&synch2`, `&sweep`, `&scorr`, `&mfs`, `&nf` — covering the full abc9 pipeline.

Performance overhead is within measurement noise (<1%) across all engines.

On the Yosys side, I've also updated this branch to handle **multi-source attribution**: each mapped cell now collects `\src` from both its output object's origin and each input object's origin, merging unique values with `|`. This addresses the case where a LUT covers logic from multiple RTL source lines — ABC tracks one best origin per object, and Yosys merges them at the mapping readback stage.

robtaylor:
@Ravenslofty Coming back to this — the ABC PR (#487) now has full engine coverage for the abc9 pipeline and Alan has confirmed the approach. Two options:

1. **Split now**: Revert `bbdfd0f`, keep `&verify -y` as the mechanism, land this PR independently. Origin tracking via "y" extension becomes a follow-up once ABC #487 merges.
2. **Wait for ABC to land**: Keep the PR as-is, land both together.

Do you have a preference? I'm happy either way — splitting might be cleaner since it decouples the two repos.

whitequark:
@robtaylor Speaking with my formal @YosysHQ staff hat on for the duration of this comment, and representing the team as a whole: We have discussed this PR internally and decided not to merge it. This is due to both the manner in which it was produced as well as technical issues with it. Since I am far from being an expert on ABC, I will leave the discussion of the latter to @Ravenslofty, who has final say over this part of Yosys.

robtaylor:
*shrug*, well the underlying needs will land in abc. Up to you guys what
you do with that.

robtaylor:
IMHO Technical issues should be raised as technical issues. I’ve reviewed all code manually. Any errors are mine, not of any tool I'm using.

Closing issues without any clear issues raised and purely on political/ideological grounds without any clear reasoning is problematic at
the very least.

If the Yosys project has the policy that no tool that includes an LLM should be used to interact with the PR process, then that should be firmly and clearly explicated.

ShinyKate:
YosysHQ is not outright against use of LLMs. Our [review guidelines](https://yosyshq.readthedocs.io/projects/yosys/en/latest/yosys_internals/extending_yosys/contributing.html#how-to-review) state "We do not ban LLM code from the codebase, we ban bad code."

But our [contributing guidelines](https://yosyshq.readthedocs.io/projects/yosys/en/latest/yosys_internals/extending_yosys/contributing.html#code-that-matters) ask that people discuss their motivation and approach ahead of time on the Discourse forum. Especially with such a large change this is what we expect so that we can make sure our goals are aligned and no one wastes time.

In the case of the 2,000 line ABC PR, you did not do this. We at YosysHQ accept responsibility that our downstream ABC repo did not communicate this expectation because we did not expect anybody to file PRs against it and not against upstream ABC; as a result we have explicitly disabled PRs on that repo.

At the time Lofty pointed you to the Yosys Discourse, and [you said](https://yosyshq.discourse.group/t/src-attribute-preservation/126/2) that you would update it as things progress, but you haven't in two weeks. To underline this need for communication, [the Yosys PR template](https://github.com/YosysHQ/yosys/blob/main/.github/PULL_REQUEST_TEMPLATE.md) has as its top line "If your work is part of a larger effort, please discuss your general plans on Discourse first to align your vision with maintainers."

While Lofty did suggest that splitting up the PR would have been the preferred approach that alone will not be enough to approve. Between pushing back on feedback because the PR was in a draft and ignoring suggestions for multiple weeks we're losing faith that we will see enough alignment with our quality standards and Yosys style patterns.

We do think this functionality is valuable but we take issue with the implementation and the process in which this was presented. To move this forward I suggest restarting discussion in the discourse thread.

robtaylor:
I did discuss with Lotfy. The main discussion of this has been in [abc
#487](
https://github.com/berkeley-abc/abc/pull/487)
where Lotfy has been tagged and asked for his input.

Ravenslofty:
The focus of that discussion was on ABC internals. Other than my suggestion that the XAIGER y extension was used, what goes on inside ABC is up to Alan.

robtaylor:
Yes, but it is a dependency of any successful implementation of this functionality.

@Ravenslofty did you see the private email exchange with Alan following on your comment ~two weeks ago?

Also on the abc discussion, Alan is happy with the general approach in that issue, and as long as when the functionality is zero cost when not used, he's flexible. What's really needed here is for us to figure out what we want to do wrt node merging. 

My current plan is to make a test set so we can evaluate how much merging happens in common designs, but you will have a deeper understanding here than I.

PR #5717, opened 2026-03-02
gowin: add hardware latch support (DL/DLN/DLC/DLP variants)

zaun:
Add simulation models, techmap, and dfflegalize rules for Gowin DL-series latch primitives. Latches use the same physical BEL as DFFs with REGMODE set to LATCH. All 12 variants are supported: DL, DLE, DLN, DLNE, DLC, DLCE, DLNC, DLNCE, DLP, DLPE, DLNP, DLNPE.

**Depends on:** YosysHQ/apicula#476 YosysHQ/nextpnr#1652

### DIFF
--- a/techlibs/gowin/cells_sim.v
+++ b/techlibs/gowin/cells_sim.v
@@ -543,7 +543,6 @@ endmodule // DFFNCE (negative clock edge; asynchronous clear; clock enable)
 // Latch sim cells
 // Gate signal uses CLK port name to match the physical DFF BEL pin
 
-(* lib_whitebox *)
 module DL (output reg Q, input D, CLK);
   parameter [0:0] INIT = 1'b0;
   initial Q = INIT;
@@ -551,7 +550,6 @@ module DL (output reg Q, input D, CLK);
     if (CLK) Q <= D;
 endmodule
 
-(* lib_whitebox *)
 module DLN (output reg Q, input D, CLK);
   parameter [0:0] INIT = 1'b0;
   initial Q = INIT;
@@ -559,7 +557,6 @@ module DLN (output reg Q, input D, CLK);
     if (!CLK) Q <= D;
 endmodule
 
-(* lib_whitebox *)
 module DLE (output reg Q, input D, CLK, CE);
   parameter [0:0] INIT = 1'b0;
   initial Q = INIT;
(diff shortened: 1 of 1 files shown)


techlibs/gowin/cells_sim.v
```diff
@@ -540,6 +540,113 @@ module DFFNCE (output reg Q, input D, CLK, CE, CLEAR);
   end
 endmodule // DFFNCE (negative clock edge; asynchronous clear; clock enable)
 
+// Latch sim cells
+// Gate signal uses CLK port name to match the physical DFF BEL pin
+
+(* lib_whitebox *)
```
Ravenslofty: I don't think these need to be marked as `(* lib_whitebox *)`.

zaun:
Removed the `(* lib_whitebox *)`.

Ravenslofty (approved):

mmicko:
@zaun Just to notify you that I have rebased your PR just to retrigger CI and make it possible to check all and be ready when it needs to be merged, so be aware that if you need to do some changes it will require you first to pull the latest.

pepijndevos:
On the Apicula end this now appears to fail with
```
ABC: Timing manager is given but there is no GIA of boxes.
ABC: Error: Abc_FrameUpdateGia(): Transformation has failed.
```

https://github.com/YosysHQ/apicula/actions/runs/22755526431/job/66017400564?pr=476

Ravenslofty:
> On the Apicula end this now appears to fail with
> 
> ```
> ABC: Timing manager is given but there is no GIA of boxes.
> ABC: Error: Abc_FrameUpdateGia(): Transformation has failed.
> ```

ABC is complaining, yes, but it's not fatal to Yosys. Are you grepping the logs for "error" and giving up if you find it or something?

> https://github.com/YosysHQ/apicula/actions/runs/22755526431/job/66017400564?pr=476

Honestly I would love to help further, since this is an ABC9 issue, but you appear to be running Yosys on four different designs simultaneously and the logs are unhelpful.

pepijndevos:
Yea I agree a parallel make isn't very helpful for debugging and I didn't realise the ABC failure was nonfatal.

I'm trying to do this between things that pay the bills so I'll have to find some time to run the problematic example locally or run CI on a single thread for debugging. Actually that seems pretty quick compared to building three PR branches of different stuff. brb single threaded CI log incoming.

pepijndevos:
The Apicula side has been merged

p.s. the single threaded log revealed the actual error which was caused by the nextpnr PR lagging behind main

pepijndevos:
Could this be merged? Nextpnr side has also landed.

PR #5721, opened 2026-03-04
abc: new option to pass extra ABC read_lib args

tefantasy:
This pull request adds an option `-liberty_args` to the `abc` command when performing technology mapping with liberty libraries. The option allows user to pass custom arguments to ABC's read_lib command. 

This pull request can be regarded as a revival of #4343 but with minimal changes focusing on the `-liberty_args` feature and without other code restructurings (e.g., the introduction of user_script in #4343). 

As noted in #4343, the main purpose of adding this option is to control how ABC generates its internal delay model from timing data in liberty files. This can improve mapping quality by enabling users to tune ABC’s delay modeling. For example, we tested on 12 IWLS'05/OpenCores designs, mapped to NanGate45 library. By adding `-liberty_args "-G 250"`, the averaged (geometric mean) delay improvement (reported by post-synthesis STA) is 12.0%, at the cost of a 4.6% area increase. 

Although #4343 has been replaced by #4592 which introduces larger scaled customization of ABC flow, it involves many new features and code refactoring, and possibly requires adding a new command. In contrast, this pull request makes a small modification to the existing `abc` command to increase the flexibility for mapping quality improvement.


passes/techmap/abc.cc
```diff
@@ -1024,7 +1025,7 @@ void AbcModuleState::prepare_module(RTLIL::Design *design, RTLIL::Module *module
 		}
 		bool first_lib = true;
 		for (std::string liberty_file : config.liberty_files) {
-			abc_script += stringf("read_lib %s %s -w \"%s\" ; ", dont_use_args, first_lib ? "" : "-m", liberty_file);
+			abc_script += stringf("read_lib %s %s %s -w \"%s\" ; ", dont_use_args, first_lib ? "" : "-m", config.abc_liberty_args, liberty_file);
```
hexratcc: We've recently introduced `convert_liberty_files_to_merged_scl`, which attempts to merge the read files into an scl, please update that path so that it also includes the args.

hexratcc (changes requested):
Seems sound to me, but please apply the suggested changes.

passes/techmap/abc.cc
tefantasy: Hi, may I confirm whether you mean adding the extra args to the ABC read_lib invocations in `convert_liberty_files_to_merged_scl` as well?

hexratcc: Correct.

tefantasy: Updated. In particular, if `abc_liberty_args` (extra args to ABC read_lib) is provided, SCL merge will not be performed and the program falls back to the original read_lib mode. This is because passing gain and slew parameters to read_lib does not affect the generated SCL (they only affect the internal genlib of ABC). So, if gain and slew parameters is provided in `abc_liberty_args` and the SCL merge mode is executed, the real ABC execution with read_scl is unable to capture the gain and slew parameters.

passes/techmap/abc.cc
```diff
@@ -1030,14 +1031,17 @@ void AbcModuleState::prepare_module(RTLIL::Design *design, RTLIL::Module *module
...
-		std::string merged_scl = convert_liberty_files_to_merged_scl(config.liberty_files, run_abc.dont_use_args, config.exe_file);
+		std::string merged_scl;
+		if (config.abc_liberty_args.empty()) {
+			merged_scl = convert_liberty_files_to_merged_scl(config.liberty_files, run_abc.dont_use_args, config.exe_file);
+		}
 		if (!merged_scl.empty()) {
 			run_abc.abc_script += stringf("read_scl \"%s\" ; ", merged_scl.c_str());
 		} else if(!config.liberty_files.empty()) {
-			log_warning("ABC: Merged scl conversion failed, using liberty format\n");
+			log_warning("ABC: Merged scl conversion failed, or abc_liberty_args provided, using liberty format\n");
```
hexratcc: LGTM, but please split up the warning for the two paths - one for when the SCL merge actually fails and one for when it's just skipped due to read_lib args.

hexratcc (changes requested):

passes/techmap/abc.cc
tefantasy: Updated.

hexratcc (approved):

PR #5724, opened 2026-03-04
setundef: respect selection for cells, processes, and connections

### LINKED ISSUE #5624: Setundef doesn't respect selection
### Version

Yosys 0.61+21 (git sha1 967b47d98, clang++ 18.1.8 -fPIC -O3)

Linux

### Reproduction Steps

```Verilog
module test;
    wire a = 1'bx;
    wire b = 1'bx;
endmodule
```

`setundef -zero w:a`

### Expected Behavior

`setundef` should only target selection, so the value of `a` driver has to be changed to `0`, but `b` should remain `x`.

### Actual Behavior

`setundef` changes all wire values to `0`, ignoring selection.

abhinavputhran:
Fixes #5624

Previously, setundef would rewrite sigspecs in all cells, processes, and connections regardless of the active selection. Only modules and memories were correctly filtered by selection.

Fix by using module->selected_cells() for cells, adding a module->selected() check for processes, and checking wire selection on the lhs of each connection before rewriting.

Tested with the reproduction case from the issue:
- `setundef -zero w:a` now only changes `a`, leaving `b` as `x`
- `setundef -zero` with no selection still changes everything

### DIFF
--- a/tests/various/setundef_selection.ys
+++ b/tests/various/setundef_selection.ys
@@ -11,7 +11,7 @@ sat -enable_undef -prove b 0 -falsify
 design -reset
 
 # Test that setundef -undriven -zero respects wire selection
-read_rtlil setundef_selection_undriven.il
+read_verilog setundef_selection_undriven.v
 setundef -undriven -zero w:b
 sat -prove b 0
 sat -enable_undef -prove a 0 -falsify
--- a/tests/various/setundef_selection_undriven.il
+++ b/tests/various/setundef_selection_undriven.il
@@ -1,4 +1,0 @@
-module \test
-  wire output 1 \a
-  wire output 2 \b
-end
--- a/tests/various/setundef_selection_undriven.v
+++ b/tests/various/setundef_selection_undriven.v
@@ -1,0 +1,4 @@
+module test;
+    wire a;
+    wire b;
+endmodule


passes/cmds/setundef.cc
```diff
@@ -502,14 +502,22 @@ struct SetundefPass : public Pass {
...
 
-			for (auto &it : module->cells_)
-				if (!it.second->get_bool_attribute(ID::xprop_decoder))
-					it.second->rewrite_sigspecs(worker);
+			for (auto cell : module->selected_cells())
+				if (!cell->get_bool_attribute(ID::xprop_decoder))
+					cell->rewrite_sigspecs(worker);
 			for (auto &it : module->processes)
-				it.second->rewrite_sigspecs(worker);
+				if (module->selected(it.second))
```
widlarizer: We do have `selected_processes` as well

abhinavputhran: Switched to `selected_processes()`

KrystalDelusion:
While this is an improvement (and fixes the specific example in #5624), there are still a lot of places in `setundef` that still don't respect the selection (e.g. `setundef -undriven` and `setundef -init`), so while I wouldn't block this PR for it I do think #5624 should remain open.

passes/cmds/setundef.cc
```diff
@@ -502,14 +502,21 @@ struct SetundefPass : public Pass {
...
-			for (auto &it : module->cells_)
-				if (!it.second->get_bool_attribute(ID::xprop_decoder))
-					it.second->rewrite_sigspecs(worker);
-			for (auto &it : module->processes)
-				it.second->rewrite_sigspecs(worker);
+			for (auto cell : module->selected_cells())
+				if (!cell->get_bool_attribute(ID::xprop_decoder))
+					cell->rewrite_sigspecs(worker);
+			for (auto &it : module->selected_processes())
+				it->rewrite_sigspecs(worker);
```
KrystalDelusion: Suggest changing `&it` to (e.g.) `proc` for clarity.

passes/cmds/setundef.cc
```diff
@@ -502,14 +502,21 @@ struct SetundefPass : public Pass {
...
-				worker(it.second);
+				SigSpec lhs = it.first;
+				bool selected = false;
+				for (auto &chunk : lhs.chunks())
+					if (chunk.wire && module->design->selected(module, chunk.wire))
+						selected = true;
+				if (selected) {
+					worker(it.first);
+					worker(it.second);
+				}
```
KrystalDelusion, line 529: I'm not entirely clear on the distinction between `connections_` and `wires_`, but is there a way to refactor this to be able to use `Module::selected_wires()` instead of iterating through the wires in a connection to check if they are selected?

abhinavputhran: connections_ are SigSig pairs that represent direct assignments (for example `assign a = b` or `wire a = 1'bx`). They don't have a direct selected_wires() equivalent because the lhs is a SigSpec (which may span multiple wires or partial wires), not a single Wire*. selected_wires() returns Wire* objects, so we'd still need to check each chunk of the lhs SigSpec against the selected wire set.

We could build a pool of selected wires first and check against that, which might be  cleaner:

```
pool<Wire*> selected = {module->selected_wires().begin(), 
                                           module->selected_wires().end()};
for (auto &it : module->connections_) {
    bool sel = false;
    for (auto &chunk : it.first.chunks())
        if (chunk.wire && selected.count(chunk.wire))
            sel = true;
    if (sel) {
        worker(it.first);
        worker(it.second);
    }
}
```
I'll make that change if you think it's cleaner.

passes/cmds/setundef.cc
abhinavputhran: Done

abhinavputhran:
> While this is an improvement (and fixes the specific example in #5624), there are still a lot of places in `setundef` that still don't respect the selection (e.g. `setundef -undriven` and `setundef -init`), so while I wouldn't block this PR for it I do think #5624 should remain open.

Just added selection checks to `-init` mode.

For `-undriven`, I noticed both the `expose_mode` and non-expose branches iterate `module->wires_` and `module->cells_` without selection checks. I wanted to check if you select `w:a`, should only `a` be exposed/driven if undriven, or should all cells still be analyzed to determine what's driving what? 

edit: @widlarizer same question for you regarding -undriven. Should selection filter only the output (which wires get exposed/driven) while still analyzing all cells globally, or should it filter the analysis too?

passes/cmds/setundef.cc
widlarizer: I don't think that changes helps things. The problem is that RTLIL is disorienting, which isn't your fault, your solution to this bit is totally fine IMO

abhinavputhran: Understood, I'll leave it as is then

widlarizer:
Only output. The other option would be error prone

abhinavputhran:
Just added selection checks to -undriven mode as well. I tested with
```
module test(output wire a, output wire b);
    assign a = 1'b0;
endmodule
```

Without selection should add connect \b 1'0 
`./yosys -p "read_verilog /tmp/test_undriven.v; setundef -undriven -zero; dump"`

With selection on b should also add connect \b 1'0
`./yosys -p "read_verilog /tmp/test_undriven.v; setundef -undriven -zero w:b; dump"`

With selection on a (which is already driven) so nothing should change, b stays undriven
`./yosys -p "read_verilog /tmp/test_undriven.v; setundef -undriven -zero w:a; dump"`

I think all modes now respect selection

passes/cmds/setundef.cc
KrystalDelusion: Agreed

passes/cmds/setundef.cc
```diff
@@ -417,6 +421,9 @@ struct SetundefPass : public Pass {
 							if (wire->name[0] == (wire_types ? '\\' : '$'))
 								continue;
 
+							if (!design->selected(module, wire))
+									continue;
```
KrystalDelusion: There's a couple places like this where the `continue;` is indented an extra level (I'm guessing it was a copy paste that just didn't get reflowed properly).

passes/cmds/setundef.cc
```diff
@@ -467,6 +477,9 @@ struct SetundefPass : public Pass {
...
+
```
KrystalDelusion: This can be simplified to `for (auto wire : module->selected_wires())` (there are a couple of these).

passes/cmds/setundef.cc
KrystalDelusion, line 367: I think it might make more sense here to use `module->selected_cells()` and then _not_ check if the wire is selected.  If I call `setundef -init c:myff`, I would expect the `myff` cell to be processed, but the current approach requires that I instead do something like `setundef -init c:myff %co` to select the output wires (and then it doesn't care if the cell is even selected).  I think this deserves an additional test case or two to demonstrate intended usage.

KrystalDelusion: Can we change this to iterate over `module->selected_wires()` instead?  That way we don't need to check if the `sig.chunks` are selected because we only added ones that were.  I'm not sure if that would cause problems when we iterate over all the cell connections and delete the outputs, but if it does then the current approach is fine.

passes/cmds/setundef.cc
abhinavputhran: Fixed, my default tab was 4 spaces.

passes/cmds/setundef.cc
abhinavputhran: Got it

passes/cmds/setundef.cc
abhinavputhran: Switched to selected_cells() for building ffbits, and kept module->wires() for the wire search loops so we can find where those bits live regardless of wire selection. Added a test in tests/various/setundef_selection.ys using c:myff_a that confirms only the selected FF's output wire gets its init attribute set.

abhinavputhran: Switched to selected_wires() when building undriven_signals in the non-expose branch, and removed the chunk-level selection check. Added a test confirming setundef -undriven -zero w:b only drives the selected wire.

tests/various/setundef_selection.ys
```diff
@@ -0,0 +1,27 @@
+# Test that setundef -zero respects wire selection: only selected wire is changed
+read_verilog <<EOT
+module test;
+    wire a = 1'bx;
+    wire b = 1'bx;
+endmodule
+EOT
+setundef -zero w:a
+sat -prove a 0
```
KrystalDelusion, line 9: These verify that `a` is zeroed, but it doesn't very that `b` is not.
Adding below suggestion passes with these changes (because `b=1'x`) but fails with existing code, correctly showing that *only* the selected wire is changed.
```suggestion
sat -prove a 0
sat -enable_undef -prove b 0 -falsify
```

tests/various/setundef_selection.ys
```diff
@@ -0,0 +1,27 @@
...
+
+# Test that setundef -undriven -zero respects wire selection
+read_verilog <<EOT
+module test(output wire a, output wire b);
+    assign a = 1'b0;
+endmodule
+EOT
+setundef -undriven -zero w:b
+sat -prove b 0
+design -reset
```
KrystalDelusion: `b` is the only wire with an undef value so would be the only possible change regardless of selection.  As with the first test, this only shows that providing a selection doesn't break but it doesn't actually show that the selection or the `-undriven` flag matter.

abhinavputhran:
@KrystalDelusion could pls rerun the workflows? I think CI runs tests from within the tests/various/ directory so I changed the file path from tests/various/setundef_selection_ff.il to setundef_selection_ff.il

passes/cmds/setundef.cc
```diff
@@ -362,7 +364,7 @@ struct SetundefPass : public Pass {
 				pool<SigBit> ffbits;
 				pool<Wire*> initwires;
 
-				for (auto cell : module->cells())
+				for (auto cell : module->selected_cells())
```
KrystalDelusion: I'm actually second guessing this one on further inspection.  `init` attributes are typically assigned to the wire, not the flop itself, and for a `reg a` we generate an RTLIL wire named `a` driven by a `$dff` with a name like `$procdff$3`.  So we would also want to be able to select just the wire without having to select the ff that drives it, as opposed to the current:
```
read_verilog <<EOT
module test(input clk, q, output reg a = 1'bx);
	always @(posedge clk)
        a <= q;
endmodule
EOT
proc
setundef -init -zero w:a %ci
```

KrystalDelusion:
And yes, as you noticed `make test` runs tests in the test directories :)

(I had it as a review comment but you already fixed it before I posted the review)

tests/various/setundef_selection.ys
abhinavputhran: Okok, I added the sat line as suggested

tests/various/setundef_selection.ys
abhinavputhran: I see. I switched to design with both a and b undriven so selection actually matters. Added falsify check to verify a was left unchanged

passes/cmds/setundef.cc
abhinavputhran: Yeah you should be able to select wire also. The flipflop scan now checks if either the cell itself is selected or its Q output wire is selected, so both setundef -init -zero w:a and setundef -init -zero c:myff work correctly. Also added test cases for both selection methods with checks confirming the unselected wire is not affected.

PR #5728, opened 2026-03-06
Replace deprecated Tcl API to fix use-after-free

povik:
Under Tcl 9.0 the Tcl_SetResult utility is a macro:

	#define Tcl_SetResult(interp, result, freeProc) \
		do { \
		    const char *__result = result; \
		    Tcl_FreeProc *__freeProc = freeProc; \
		    Tcl_SetObjResult(interp, Tcl_NewStringObj(__result, -1)); \
		    if (__result != NULL && __freeProc != NULL && __freeProc != TCL_VOLATILE) { \
			if (__freeProc == TCL_DYNAMIC) { \
			    Tcl_Free((char *)__result); \
			} else { \
			    (*__freeProc)((char *)__result); \
			} \
		    } \
		} while(0)

Temporaries constructed as part of the 'result' expression will be dropped before the 'result' pointer is used. What was safe when Tcl_SetResult was a function isn't safe with the macro definition. Transition away from deprecated SetResult to calling SetObjResult/MewStringObj directly.

### DIFF
--- a/kernel/tclapi.cc
+++ b/kernel/tclapi.cc
@@ -279,7 +279,7 @@ static int tcl_get_attr(ClientData, Tcl_Interp *interp, int argc, const char *ar
 		ERROR("object not found")
 
 	if (string_flag) {
-		Tcl_SetResult(interp, (char *) obj->get_string_attribute(attr_id).c_str(), TCL_VOLATILE);
+		Tcl_SetObjResult(interp, Tcl_NewStringObj(obj->get_string_attribute(attr_id).c_str(), -1));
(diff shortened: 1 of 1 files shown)


widlarizer:
~~What's the intention with the -1? Doesn't `Tcl_SetObjResult` only take two arguments?
https://www.tcl-lang.org/man/tcl8.5/TclLib/SetResult.html
https://www.tcl-lang.org/man/tcl9.1/TclLib/SetResult.html~~ guess I can't read C++ without bracket colorizers anymore

PR #5729, opened 2026-03-06
dfflibmap: pass selection to dfflegalize

### LINKED ISSUE #5650: dfflibmap violates selection through dfflegalize
### Version

main

Linux

### Reproduction Steps

`read_verilog tests/sim/dffe.v; proc; techmap; stat; dfflibmap -liberty tests/techmap/dfflibmap.lib nothing; stat`

### Expected Behavior

`nothing` is an empty selection so the design should be unchanged

### Actual Behavior

`nothing` is an empty selection and you get a warning for it, but `dfflibmap` calls calls `dfflegalize` on the whole design despite `dfflegalize` supporting selections too. This threw me off when investigating test failures, because I was doing

```
copy top top_unmapped
dfflibmap -liberty dfflibmap_dffn_dffe.lib top
```

and it would modify `top_unmapped` which I saved to prevent modifying it

abhinavputhran:
Fixes #5650 

dfflibmap was calling dfflegalize on the whole design regardless of the active selection, causing unselected modules to be modified unexpectedly.

Fix by appending the selected module names to the dfflegalize command string.

Added a test to tests/techmap/dfflibmap.ys that verifies unselected modules are not modified after calling dfflibmap with a selection.

### DIFF
--- a/passes/techmap/dfflibmap.cc
+++ b/passes/techmap/dfflibmap.cc
@@ -694,6 +694,8 @@ struct DfflibmapPass : public Pass {
 			for (auto it : cell_mappings)
 				dfflegalize_cmd += stringf(" -cell %s 01", it.first);
 			dfflegalize_cmd += " t:$_DFF* t:$_SDFF*";
+			for (auto module : design->selected_modules())
+				dfflegalize_cmd += " " + module -> name.str();
 			if (info_mode) {
 				log("dfflegalize command line: %s\n", dfflegalize_cmd);
 			} else {
--- a/tests/techmap/dfflibmap.ys
+++ b/tests/techmap/dfflibmap.ys
@@ -97,3 +97,14 @@ select -assert-count 1 t:dffe
 # but we have to invert it because the CLEAR pin is active low.
 # ff2.CLEAR = !R
 select -assert-count 1 c:ff2 %x:+[CLEAR] %ci t:$_NOT_ %i
+
+# Test that dfflibmap respects selection and doesn't run dfflegalize on unselected modules
+design -load orig
+copy top top_unmapped
+dfflibmap -liberty dfflibmap.lib top
+# top should be mapped
+select -module top -assert-count 1 t:dffn
+# top_unmapped should have no mapped cells
+select -module top_unmapped -assert-none t:dffn
+select -module top_unmapped -assert-none t:dffe
+select -module top_unmapped -assert-none t:dffsr


KrystalDelusion:
The included tests pass even without this change (and that's not even getting into whether the change is the correct way to do it or not).  Will post more in the issue thread, but using `Pass::call()` is already passing the current design, *which includes the current selection*.

tests/techmap/dfflibmap.ys
```diff
@@ -97,3 +97,14 @@ select -assert-count 1 t:dffe
...
+# Test that dfflibmap respects selection and doesn't run dfflegalize on unselected modules
+design -load orig
+copy top top_unmapped
+dfflibmap -liberty dfflibmap.lib top
+# top should be mapped
+select -module top -assert-count 1 t:dffn
+# top_unmapped should have no mapped cells
+select -module top_unmapped -assert-none t:dffn
+select -module top_unmapped -assert-none t:dffe
+select -module top_unmapped -assert-none t:dffsr
```
KrystalDelusion, line 110: Test `dfflegalize` isn't running, not `dfflibmap`.
```suggestion
select -module top_unmapped -assert-count 1 t:$_DFFE_PP_
select -module top_unmapped -assert-none t:$_DFF_N_
```

passes/techmap/dfflibmap.cc
```diff
@@ -694,6 +694,8 @@ struct DfflibmapPass : public Pass {
 			for (auto it : cell_mappings)
 				dfflegalize_cmd += stringf(" -cell %s 01", it.first);
 			dfflegalize_cmd += " t:$_DFF* t:$_SDFF*";
+			for (auto module : design->selected_modules())
+				dfflegalize_cmd += " " + module -> name.str();
```
KrystalDelusion, line 698: This ain't it.  Not only does this expand partially selected modules into fully selected modules, but it is still selecting *all* `t:$_DFF* t:$_SDFF*`, making this actively worse and causing failures if the design contains any other types of flop (raising an error in `dfflegalize`).

My current thought is running through all the currently selected cells, pushing them to a new selection if they are the right type, and then using `Pass::call_on_selection()` with the new selection.  Or doing `% t:$_DFF* t:$_SDFF* %u %i` might work, but that's starting to get messy in terms of readability.

KrystalDelusion:
@abhinavputhran , when you are adding tests for a fix *they should fail without the fix*.  My usual approach/suggestion is to write the tests before writing the fix, to make sure you understand what isn't working.  This is the second PR where you've added tests that don't show anything.  I suggest you take another look at your third PR and make sure the test is actually useful (at a very brief glance, just calling `dump` probably isn't).

abhinavputhran:
Well noted, first time writing tests so I'm learning. I'll make sure to verify each test fails without the fix before submitting and that the tests test what they should

KrystalDelusion:
Thanks :)

PR #5730, opened 2026-03-06
async2sync: inherit src attributes on new_q and new_d wires

### LINKED ISSUE #5674: async2sync doesn't copy src attributes consistently
### Version

main

Linux

### Reproduction Steps

Stare at asyn2sync.cc and notice `Wire *new_q` and related multiplexers

### Expected Behavior

Said objects should inherit src from flop and maybe also derive name rather than NEW_ID

### Actual Behavior

Inverse of expected

abhinavputhran:
Fixes #5674 

async2sync created wires new_q and new_d without src attributes, making them hard to trace in debug output. Fixed by inheriting the src attribute from the original flop cell on all newly created wires across all four conversion paths.

Added a test in tests/sat/async2sync.ys verifying the generated wires carry the correct src attribute.

### DIFF
--- a/tests/sat/async2sync.ys
+++ b/tests/sat/async2sync.ys
@@ -1,3 +1,4 @@
+# has_arst path
 read_verilog << EOT
 module top(input clk, arst, d, output reg q);
   always @(posedge clk or posedge arst)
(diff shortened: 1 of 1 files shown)


abhinavputhran:
Added more comprehensive tests and confirmed they fail without fix

Issue #5708, opened 2026-02-25
Print wall time on exit


widlarizer:
### Feature Description

Now that yosys has parallelized passes, the user and sys time printed at the end of the program's run no longer adds up to the wall time the program ran for
[sc-new-story]

PR #5731, opened 2026-03-08
Implement wall clock time measurement

Goubermouche:
Implement wall clock time measurement as mentioned in #5708

**TODO:**
- [x] Use wall clock time in yosys-perf

### DIFF
--- a/kernel/driver.cc
+++ b/kernel/driver.cc
@@ -705,9 +705,9 @@ int main(int argc, char **argv)
 		double wall_seconds = std::chrono::duration<double>(
 				std::chrono::steady_clock::now() - wall_clock_start).count();
 
-		log("End of script. Logfile hash: %s%sCPU: user %.2fs system %.2fs, wall: %.2fs%s\n", hash,
-				stats_divider.c_str(), ru_buffer.ru_utime.tv_sec + 1e-6 * ru_buffer.ru_utime.tv_usec,
-				ru_buffer.ru_stime.tv_sec + 1e-6 * ru_buffer.ru_stime.tv_usec, wall_seconds, meminfo.c_str());
+		log("End of script. Logfile hash: %s%stime: %.2fs, user: %.2fs, system: %.2fs%s\n", hash,
+				stats_divider.c_str(), wall_seconds, ru_buffer.ru_utime.tv_sec + 1e-6 * ru_buffer.ru_utime.tv_usec,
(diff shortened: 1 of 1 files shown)


kernel/driver.cc
```diff
@@ -700,9 +702,12 @@ int main(int argc, char **argv)
 		meminfo = stringf(", MEM: %.2f MB peak",
 				ru_buffer.ru_maxrss / (1024.0 * 1024.0));
 #endif
-		log("End of script. Logfile hash: %s%sCPU: user %.2fs system %.2fs%s\n", hash,
+		double wall_seconds = std::chrono::duration<double>(
+				std::chrono::steady_clock::now() - wall_clock_start).count();
+
+		log("End of script. Logfile hash: %s%sCPU: user %.2fs system %.2fs, wall: %.2fs%s\n", hash,
```
widlarizer: Looking at it like this, I think that what we should do is present it like "real X.XXs, user X.XXs, system X.XXs" so that the numbers users are most interested in is shown first and with a user-friendly tag. "CPU:" suggests CPU time, which is specific to user and system, but developers can be expected to know what those terms conventionaly mean in the context of program run time, so I think we can replace "CPU:" with "time:" as well

widlarizer: or just "time: X.XXs, user X.XXs, system X.XXs" and remove the "CPU:"

hexratcc: Updated, new format is: 
```
End of script. Logfile hash: c65e6794c8, time: 0.01s, user: 0.01s, system: 0.00s, MEM: 17.62 MB peak
```

whitequark:
Worth checking what this does on WASI (if it doesn't break the executable it's fine).

mmicko:
Have manually triggered github action to create WASI build https://github.com/YosysHQ/yosys/actions/runs/22955594014 but that will just do compile test not checking binary. Please note that extra builds are checked but only when checking merge queue or when manually triggered to save some CI time.

hexratcc:
> Worth checking what this does on WASI (if it doesn't break the executable it's fine).

@whitequark  As far as I can tell this shouldn't break anything and should work just fine. I found [this repo](https://github.com/WebAssembly/wasi-clocks) which seems to confirm that monotonic clocks should work in WASI (specifically used [here](https://github.com/WebAssembly/wasi-libc/blob/main/libc-bottom-half/clocks/clock.c)).  Additionally, I noticed that `steady_clock` is already used in `abc`, which I assume is also compiled for WASI.

PR #5736, opened 2026-03-10
proc_mux: avoid redundant mux cells for full_case switches with a dominant arm value

likeamahoney:
**Problem**

  When `proc_rmdead` removes an exhausted implicit default branch it marks the switch `full_case`. `proc_mux` then masks all covered signal bits to `Sx`  before the mux-generation loop. If most arms assign the same value, every arm still produces a concrete value `≠ Sx` and gets its own `$eq+$mux` pair - even the identical ones.

While experimenting with the Verilog code of the open-source [tv80](https://github.com/hutch31/tv80/tree/master/rtl/core) processor, I noticed that this behavior can lead to a large number of redundant cells. After minimizing the example (see attached file - [tv80_reduced.txt](https://github.com/user-attachments/files/25869258/tv80_reduced.txt)), the generated netlist contained almost two orders of magnitude more cells than necessary (see comparison in files with and without the patch -  [tv80redstat_with_patch.txt](https://github.com/user-attachments/files/25869341/tv80redstat_with_patch.txt) and [tv80redstat_wo_patch.txt](https://github.com/user-attachments/files/25869339/tv80redstat_wo_patch.txt)). And both designs are semantically equivalent after evaluation. On the original tv80_mcode module the effect is smaller, but the optimization still reduces the cell count by around 5-8% (see comparison in files with and without the patch  [tv80stat_wo_patch.txt](https://github.com/user-attachments/files/25869284/tv80stat_wo_patch.txt) and [tv80stat_with_patch.txt](https://github.com/user-attachments/files/25869285/tv80stat_with_patch.txt)).

For cell stats comparison I used such pipeline:

```
read_verilog test.sv; proc; opt; stat -hierarchy -tech cmos;
```

  **Fix**

  After the `full_case_bits` to `Sx` mask, scan all arms' direct actions and find the majority value for each signal bit. If one value appears in more than half the arms, use it as the initial result seed instead of `Sx`. The existing `gen_mux` early-return (when == else - skip) then silently elides cells for every arm that matches the dominant value.

### DIFF
--- a/passes/proc/proc_mux.cc
+++ b/passes/proc/proc_mux.cc
@@ -393,6 +393,27 @@ RTLIL::SigSpec signal_to_mux_tree(RTLIL::Module *mod, SnippetSwCache &swcache, d
 			if (full_case_bits.count(sig[i]))
 				result[i] = State::Sx;
 
+		// For full_case switches, if a majority of arms assign the same value to a bit
+		// (via direct actions), use that as the else-branch seed instead of Sx. This
+		// lets proc_mux skip generating mux cells for arms that produce the dominant value.
+		if (!full_case_bits.empty() && !sw->cases.empty()) {
+			int n = GetSize(sw->cases);
+			std::vector<dict<SigBit, int>> counts(GetSize(sig));
+			for (auto cs2 : sw->cases) {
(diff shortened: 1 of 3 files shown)


whitequark:
This is a very interesting solution to a problem I've tried to fix before (without success).

PR #5743, opened 2026-03-11
gowin: wire output clock enable ports to enable inputs, not constants

### LINKED ISSUE #5693: Improper d-reg behavior with Yosys/Gowin
EDIT: This issue can also be reproduced without an actual Tang Nano 9K board, just by inspecting Yosys's output file `hardware.json`. As shown in [this post below](https://github.com/YosysHQ/yosys/issues/5693#issuecomment-3940277074), the problem shows up when Yosys doesn't connect the signal `probe_data_out` to the Q of the d-reg but to a LUT, which seems to be an incorrect synthesis.

----

### Version

Yosys 0.61+112 (git sha1 ac427a79b-dirty, aarch64-apple-darwin23.5-clang++ 18.1.8 -fPIC -O3)

macOS

### Reproduction Steps

Notes: 
* This issue is a continuation of https://github.com/YosysHQ/yosys/discussions/5689
* The issue was observed only with actual Gowin silicon (Sipeed Tang Nano 9k and 20K) and was not observed with ICE40 or simulation.

Prerequisites:
* A Sipeed Tang Nano 9K board.
* An oscilloscope or a logic analyzer.

Steps:
1. Connect the Tang Nano 9k board to your computer USB port.
2. Setup Yosys environment (`source environment`).
3. Uncompress the zip below to get the test project. (the zip file includes also my original build artifacts under `_build/tang9k`. ).
4. Run the script `run-tang9k.sh' to build and program your Tang Nano 9k. (The updated build artifacts can be found under`_build/tang9k`).
5. Using an oscilloscope or logic analyzer observe the following signals on the board's pads.

| SIGNAL | Pad | Comment | 
| :-- | :-- | :-- |
|  probe_sync |  38|  Sync on high to low transition |
|  probe_clk |  37 |  The D-Reg triggers on posedge of this clock |
|  probe_en | 36|  The D-Reg should change state only when this signal is high |
|  probe_data_out | 39 | The output of the D-Reg |

[gowin-bug1.zip](https://github.com/user-attachments/files/25405765/gowin-bug1.zip)

### Expected Behavior

The D-Reg should not change state on posege of its clock when probe_en is low.

The relevant code is shown below. All signals are synchronous on a posedge of a single clock. The clock speed is about 4Mhz which is way below the max reported speed of 97Mhz. 

https://github.com/zapta/published-snippets/blob/0a0f42e52ae5f6a096aacfd5b9419150cbd27eba/yosys/gowin-bug1/async_fifo.v#L92-L99

```
  // Test probes that are connected to external pads.
  assign probe_clk = rd_clk;
  assign prob_data_en = (rd_en && !empty);
  assign probe_data_out = dout[0];

  // Read data directly using CURRENT read pointer
  always @(posedge rd_clk) begin
    if (rd_en && !empty) dout <= mem[rd_bin[ADDR_WIDTH-1:0]];
  end
```

### Actual Behavior

The D-reg changes its state one clock after the trigger point, when its probe_en condition is clearly low. Shown below is a logic analyzer screenshot that demonstrate the the extra change while probe_en is low.

Image

widlarizer:
Intends to fix #5693, see https://github.com/YosysHQ/yosys/discussions/5689

Not sure this can be tested conveniently


widlarizer:
The need or validity of this is dependent on more specific chip data which we don't have in yosys, and there's workarounds in nextpnr to do it instead. See links in description

PR #5746, opened 2026-03-12
fabulous: add frame_config_mux BEls

IAmMarcelJung:
This adds support for the more recent naming of the [FABulous](https://github.com/FPGA-Research/FABulous) primitives `In/OutPass4_frame_config` which have a  `_mux` suffix (In/Outpass4_frame_config_mux`) so they can instantiated without using a custom prims file.


IAmMarcelJung:
@KelvinChung2000 and/or @EverythingElseWasAlreadyTaken might also want to have a look at this.

PR #5750, opened 2026-03-13
Fix missing return in NDEBUG case

calewis:
non void functions should return something.

### DIFF
--- a/kernel/modtools.h
+++ b/kernel/modtools.h
@@ -178,8 +178,8 @@ struct ModIndex : public RTLIL::Monitor
 
 			return false;
 		}
-		return true;
 #endif
+		return true;
 	}
 
 	void check()


whitequark:
@KrystalDelusion Shouldn't ubsan have caught this? I'm not up to date on what's going on with the integration of that but this is one of the easiest bits of UB to catch, so worth looking into why it wasn't.

KrystalDelusion:
UBSAN is run on the full test suite, but (at a guess) it doesn't use `NDEBUG` so didn't catch this.

whitequark:
We should probably add a run with ubsan and -DNDEBUG (or remove -NDEBUG if we can't guarantee that it's kept UB-free).

KrystalDelusion:
The problem with that is building/running the test suite with sanitizers is currently ~90 minutes.  Granted that is with both ubsan and asan, but even without any sanitizers it's ~30 minutes per configuration.  I could be wrong but I think this particular case could've been caught by static analysis (provided the tool supported `#if` controlled blocks and was properly configured), which I also think would be far better for runtime and coverage of all the different possible configurations.

whitequark:
> I could be wrong but I think this particular case could've been caught by static analysis (provided the tool supported `#if` controlled blocks and was properly configured), which I also think would be far better for runtime and coverage of all the different possible configurations.

It would even be caught with appropriately configured compiler warnings (in both gcc and clang, as far as I remember). Clang-analyzer might or might not have too many false positives for Yosys.

PR #5753, opened 2026-03-16
Add Carry-save adders

Goubermouche:
Adds a new pass that finds chains of `$add` and `$sub` cells and replaces them with Wallace-style carry-save adder trees built from `$fa` cells, followed by a single final `$add` for carry propagation, achieving a shorter critical path, while preserving the same results.

#### pre-csa:
image 

#### post-csa: 
image

I haven't wired the pass into synth yet (although it is available through `-arith_tree` or just `arith_tree`), but all tests seem to pass with the pass forcibly enabled without any major differences in test results.

**TODO:**
- [x] Consolidate test cases
- [ ] Benchmark

### DIFF
--- a/passes/techmap/arith_tree.cc
+++ b/passes/techmap/arith_tree.cc
@@ -337,27 +337,28 @@ struct Rewriter {
 		auto parent_of = find_parents(candidates);
 		auto [children_of, has_parent] = invert_parent_map(parent_of);
 
-		pool<Cell *> processed;
+		pool<Cell *> to_remove;
 		for (auto root : candidates) {
-			if (has_parent.count(root) || processed.count(root))
+			if (has_parent.count(root) || to_remove.count(root))
 				continue; // Not a tree root
 
 			pool<Cell *> chain = collect_chain(root, children_of);
 			if (chain.size() < 2)
 				continue;
 
-			for (auto c : chain)
-				processed.insert(c);
-
 			int neg_compensation;
 			auto operands = extract_chain_operands(chain, root, parent_of, neg_compensation);
 			if (operands.size() < 3)
 				continue;
 
+			for (auto c : chain)
+				to_remove.insert(c);
+
 			replace_with_carry_save_tree(operands, root->getPort(ID::Y), neg_compensation, "Replaced add/sub chain");
-			for (auto cell : chain)
-				module->remove(cell);
 		}
+
+		for (auto cell : to_remove)
+			module->remove(cell);
 	}
 
 	void process_maccs()


hexratcc:
@Ravenslofty Thanks for the suggestion, I'll take a look!

passes/techmap/csa_tree.cc
```diff
@@ -0,0 +1,517 @@
...
+		add->setPort(ID::B, b);
+		add->setPort(ID::Y, y);
+	}
+
+	struct DepthSig {
+		SigSpec sig;
+		int depth;
+	};
+
+	std::pair<SigSpec, SigSpec> reduce_wallace(std::vector<SigSpec>& sigs, int width, int& fa_count)
```
widlarizer: When looking at this I remembered we have `booth` and apprently it does CSAs internally. It totally warrants being a separate pass, but I think we could refactor out the overlap between `reduce_wallace` and `BoothPassWorker::WallaceSum` despite the code being short. In `booth`, there's no depth field for the summands, and I don't know how that affects things. So, maybe we could have followup PR work to refactor this bit out

hexratcc:
After some discussion with @widlarizer we decided to implement both the simple `$add`/`$sub` chains, and the `$macc` and `$alu` cells. This way the pass can be run both with and without the alumacc pass (possibly producing slightly worse results as the 'naive' path does not capture special cases that are captured by the alumacc pass).

passes/techmap/csa_tree.cc
```diff
@@ -0,0 +1,517 @@
...
+
+			if (consumer == nullptr)
+				consumer = c;
+			else if (consumer != c)
+				return nullptr;
+		}
+		return consumer;
+	}
+
+	dict<Cell*, Cell*> find_parents(const pool<Cell*>& candidates)
```
widlarizer: Needs a note defining the "parent" term

passes/techmap/csa_tree.cc
```diff
@@ -0,0 +1,517 @@
...
+		for (auto& pair : bit_consumers)
+			fanout[pair.first] = pair.second.size();
+
+		for (auto wire : module->wires())
+			if (wire->port_output)
+				for (auto bit : sigmap(SigSpec(wire)))
+					fanout[bit]++;
+	}
+
+	Cell* sole_chainable_consumer(SigSpec sig, const pool<Cell*>& candidates)
```
widlarizer: Needs a note defining the "chainable" / "chain" terms

passes/techmap/csa_tree.cc
widlarizer: Also, needs description of this code and reasoning for its correctness, likely referring to literature, using well-known consistently-used terms or defining the terms used

passes/techmap/csa_tree.cc
```diff
@@ -0,0 +1,517 @@
...
+				candidates.insert(cell);
+
+		if (candidates.empty())
+			return;
+
+		auto parent_of = find_parents(candidates);
+
+		dict<Cell*, pool<Cell*>> children_of;
+		pool<Cell*> has_parent;
+		for (auto& [child, parent] : parent_of) {
```
widlarizer: This further processing to construct further analysis/traversal indices could also be broken out of this, since it has nothing directly to do with the action of processing chains

passes/techmap/csa_tree.cc
```diff
@@ -0,0 +1,517 @@
...
+				continue;
+
+			pool<Cell*> chain = collect_chain(root, children_of);
+			if (chain.size() < 2)
+				continue;
+
+			for (auto c : chain)
+				processed.insert(c);
+
+			int correction;
```
widlarizer: What does this value express?

tests/csa_tree/csa_tree_add_chains.ys
```diff
@@ -0,0 +1,234 @@
...
+module alu_add3(
+	input [7:0] a, b, c,
+	output [7:0] y
+);
+	assign y = a + b + c;
+endmodule
+EOT
+hierarchy -auto-top
+proc
+alumacc
```
widlarizer: This actually gives you macc cells so you don't cover your alu logic here and `select -assert-none t:$alu` would fail even before csa_tree

tests/csa_tree/csa_tree_synth.ys
```diff
@@ -0,0 +1,65 @@
...
+);
+	assign y = a + b + c + d + e + f + g + h;
+endmodule
+EOT
+hierarchy -auto-top
+proc
+opt
+techmap
+abc -g AND,OR,XOR
+select -assert-min 236 t:$_AND_ t:$_OR_ t:$_XOR_ %u
```
widlarizer: I'm concerned about the stability of this test because of how abc tends to be chaotic

widlarizer: I think having tests that check we're doing equivalent, worthwhile rewrites elsewhere, is good enough, and the synth integration doesn't need to be covered

hexratcc: True, the purpose of this test was originally to be a regression test of sorts to make sure I'm not generating more gates somehow. Later I realized that CSA chains aren't really guaranteed to reduce gate/cell count, just the critical depth, so the test is completely useless.

passes/techmap/csa_tree.cc
hexratcc: > In booth, there's no `depth` field for the summands, and I don't know how that affects things. So, maybe we could have followup PR work to refactor this bit out

`booth` uses a Wallace tree tree reduction without depth tracking as (from what I can see from a cursory glance online) Booth multipliers don't benefit from the optimal depth, unlike CSA's, in other words the Wallace tree implementation in `booth` is greedy. I believe we can just use my depth-aware scheduling implementation, and just place all of the operands onto the first level.

hexratcc:
> In booth, there's no depth field for the summands, and I don't know how that affects things. So, maybe we could have followup PR work to refactor this bit out

I've added the change to this PR for now, since it's just extracting the Wallace tree implementation into a header and calling it from `booth` (mentioned in https://github.com/YosysHQ/yosys/pull/5753#discussion_r3020843907).

tests/csa_tree/csa_tree_add_chains.ys
hexratcc: Good catch, i jumbled it up, I've updated the test cases so that they use icells for alu declaration, as it makes testing easier and more dependable.

hexratcc:
@widlarizer Addressed your suggestions, and I've made some improvements to the overall code style as well.

kernel/wallace_tree.h
```diff
@@ -0,0 +1,112 @@
...
+
+	SigSpec carry;
+	carry.append(State::S0);
+	carry.append(cout.extract(0, width - 1));
+	return {sum, carry};
+}
+
+/**
+ * wallace_reduce_scheduled() - Reduce multiple operands to two using a Wallace tree
+ * @module: The Yosys module to which the compressors will be added
```
widlarizer, line 37: That'll make the second doxygen-style documented function in the codebase - not that we're running doxygen or anything. I guess it's as valid a function comment as any

kernel/wallace_tree.h
```diff
@@ -0,0 +1,112 @@
...
+ * @sigs: Vector of input signals (operands) to be reduced
+ * @width: Target bit-width to which all operands will be zero-extended
+ * @compressor_count: Optional pointer to return the number of $fa cells emitted
+ *
+ * Return: The final two reduced operands, that are to be fed into an adder
+ */
+inline std::pair<SigSpec, SigSpec> wallace_reduce_scheduled(Module *module, std::vector<SigSpec> &sigs, int width, int *compressor_count = nullptr)
```
widlarizer, line 44: I wonder if overall we have preference in the "header with inline" vs "proper header with only decls, and an impl .cc file" pattern tradeoff. My take would be that with LTO we should be inlining hard enough not to see a difference, and that smaller components that are okay to be header-only sometimes grow, so maybe I'd prefer having a .cc too. It's not big enough a difference to include in this PR, I mean just generally, I might note this in the guidelines later

techlibs/common/synth.cc
```diff
@@ -67,6 +67,10 @@ struct SynthPass : public ScriptPass {
 		log("    -booth\n");
 		log("        run the booth pass to map $mul to Booth encoded multipliers\n");
 		log("\n");
+		log("    -csa\n");
```
Ravenslofty: I think I'd prefer a longer option name here, to be more descriptive.

Ravenslofty:
This is perhaps reaching bikeshedding, but I prefer naming passes according to what they do, not how they do it. At its core, this is a pass to better handle arithmetic with more than two addends. "CSA" as an acronym could also refer to carry-select adders, and a different name would leave the door open to different approaches later on. So, I think this should be renamed "arith_tree" or something. But I don't think that should block the PR.

PR #5758, opened 2026-03-18
abc: remove -fast [sc-269]

Ravenslofty:
Similar reasoning to #5757.

Yeet.


Ravenslofty:
oops, the tests broke but somehow this still got merged.

PR #5763, opened 2026-03-18
genrtlil: fast memory initialization

### LINKED ISSUE #4601: Importing init of memory is extremely slow
### Version

Yosys 0.45+106

Linux

### Reproduction Steps

Complete Yosys version :
`Yosys 0.45+106 (git sha1 982fade0d, g++ 14.2.1 -march=x86-64 -mtune=generic -O2 -fno-plt -fexceptions -fstack-clash-protection -fcf-protection -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -ffile-prefix-map=/home/prostboa/Dev/AUR/yosys-git/src=/usr/src/debug/yosys-git -flto=auto -fPIC -O3)
`

Reproducer can't be attached due to size. Here is a public archive :
https://cloud.univ-grenoble-alpes.fr/s/f9ADZGf7sRpr9Ff

Init with zeros is done here for the sake of a simple reproducer. Actually for original design, init with arbitrary data is used.
What causes the issue if the init block with following template (generated by ghdl) :
```verilog
initial begin
    mem[65535] = 1'b0;
    mem[65534] = 1'b0;
    [...]
    mem[1] = 1'b0;
    mem[0] = 1'b0;
end
```

### Expected Behavior

Yosys should perform synthesis in very short time (for example less than 2 seconds)

### Actual Behavior

Yosys is taking at least several minutes to import the design (killed after 5 minutes)

### LINKED ISSUE #4882: `read_verilog` never completes and continues to grow system memory usage
### Version

Yosys 0.49+10 (git sha1 18a7c0038, g++ 11.4.0-1ubuntu1~22.04 -fPIC -O3)

Linux

### Reproduction Steps

[test.tar.gz](https://github.com/user-attachments/files/18626734/test.tar.gz)

```
tar xvf test.tar.gz
cd test
yosys -s run.ys
```

### Expected Behavior

Completes reading, ideally without a large memory footprint.

### Actual Behavior

```
-- Executing script file `run.ys' --

1. Executing Verilog-2005 frontend: fdivsqrtuslc4.v
Parsing SystemVerilog input from `fdivsqrtuslc4.v' to AST representation.
Generating RTLIL representation for module `\fdivsqrtuslc4'.
Warning: Replacing memory \USel4 with list of registers. See fdivsqrtuslc4.v:44
```

It never finishes replacing the memory for USel4 and continues to grow in system memory (I killed it at 120GB).

Looking at a stacktrace it appears to be stuck in: `mem2reg_as_needed_pass2` (maybe helpful, maybe not)

widlarizer:
Improves performance when handling initialized memories.
Fixes #4601
Fixes #4882

```
#4601 reduced reproducer with SIZE changed
8192: 2.63s -> 1.04s
      145.65 MB -> 141.10 MB
16384: 9s -> 3.6s
       273.77 MB -> 269.77 MB

fdivsqrtuslc4 from #4882 with upper iteration bound changed
1: 22.60s -> 2.54s
   286.59 MB -> 202.30 MB
2: 65.38s -> 5.30s
   551.50 MB -> 387.52 MB
```

Also speeds up `make test -j24` by about 5%

Notable changes:

- faster `SigSpec::remove2`
- only traverses all actions to remove assigned bits from case trees if necessary

### DIFF
--- a/frontends/ast/genrtlil.cc
+++ b/frontends/ast/genrtlil.cc
@@ -342,6 +342,9 @@ struct AST_INTERNAL::ProcessGenerator
 	// The most recently assigned $print or $check cell \PRIORITY.
 	int last_effect_priority;
 
+	// Track which signals have been assigned in current_case to avoid unnecessary removeSignalFromCaseTree calls
+	pool<RTLIL::SigBit> current_case_assigned_bits;
+
 	ProcessGenerator(std::unique_ptr<AstNode> a, RTLIL::SigSpec initSyncSignalsArg = RTLIL::SigSpec()) : always(std::move(a)), initSyncSignals(initSyncSignalsArg), last_effect_priority(0)
 	{
 		// rewrite lookahead references
@@ -430,6 +433,10 @@ struct AST_INTERNAL::ProcessGenerator
 			subst_rvalue_map = subst_lvalue_from.to_sigbit_dict(RTLIL::SigSpec(RTLIL::State::Sx, GetSize(subst_lvalue_from)));
 		} else {
 			addChunkActions(current_case->actions, subst_lvalue_to, subst_lvalue_from);
+			// Track initial assignments
+			for (auto &bit : subst_lvalue_to)
+				if (bit.wire != NULL)
+					current_case_assigned_bits.insert(bit);
 		}
 
 		// process the AST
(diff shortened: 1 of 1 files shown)


widlarizer:
CI failure is temporarily caused by failing tests on main

widlarizer:
VeryLargeScalingTest from the unit test takes 13s on main, 23ms on this branch

widlarizer:
@Ravenslofty sorry I haven't added this context before but I was still messing with this patch and figuring out how to formulate this:

This optimization was experimentally discovered and fixed with an LLM. I've approached the outputs with a lot of distrust. The unit test passes on main, so I'm confident in the remove2 change. As for the genrtlil change, we're changing code that overrides earlier assignment actions by remove2ing corresponding bits on both sides of the assignment if the lhs matches. This means the invariant is that "each bit is assigned to only once", and the bits refer to wires rather than variables, so this works fine with nesting since nested wires are separated. I still didn't feel 100% confident there isn't something I'm misunderstanding or overlooking, so I ran `diff` on RTLIL outputs of main vs this branch with verismith generated inputs for 2 hours on 24 threads and found no difference. So I'm fine with being responsible for this code change.

I was pinged by FOSS yosys users about these long-standing performance issues, and honestly I would not have worked on it if I had to stare at the frontend to figure out the fix, so I touched it with a digital 10 foot pole, and I'm happy with the result.

Let me know if this changes how you want to review this PR

Ravenslofty:
Honestly I don't think it meaningfully changes how I review the patch; I'm not nearly familiar enough with the relevant code to offer any useful insight either way.

PR #5765, opened 2026-03-18
muxpack: fix wide Y port handling

### LINKED ISSUE #5734: muxpack crash on wide Y port
### Version

main

Linux

### Reproduction Steps

```
module top(b);
  input b;
  wire ['h12 : 0] h = ~|b;
endmodule
```

`read_verilog garbage/muxpack-bug.v; proc; check -assert; muxpack`

see [here](https://github.com/YosysHQ/yosys/issues/5720#issuecomment-4024163983)

### Expected Behavior

muxpack gracefully handles whatever valid IR is thrown at it

### Actual Behavior

muxpack chokes on wide Y port on a `$logic_not` etc with `ERROR: Assert \`sig.size() == 1' failed in ./kernel/rtlil.h:2673.`

widlarizer:
Fixes #5734

### DIFF
--- a/tests/various/muxpack_wide_y.ys
+++ b/tests/various/muxpack_wide_y.ys
@@ -1,0 +1,9 @@
+# Regression test for issue #5734: muxpack crash when $logic_not / $eq / $reduce_or
+# has Y port width > 1 (e.g. boolean result assigned to a wide wire).
+read_verilog <<EOT
+module top(input b, output [18:0] h);
+  assign h = ~|b;
+endmodule
+EOT
+proc
+muxpack


passes/opt/muxpack.cc
```diff
@@ -38,19 +38,25 @@ struct ExclusiveDatabase
 		pool<Cell*> reduce_or;
 		for (auto cell : module->cells()) {
 			if (cell->type == ID($eq)) {
+				SigSpec y_sig = sigmap(cell->getPort(ID::Y));
+				if (GetSize(y_sig) == 0)
+					continue;
```
gussmith23, line 43: out of curiosity, does this actually occur in practice? or are you just being safe here?

widlarizer: Purely avoiding a bad crash when later getting the y[0] SigBit. Could be an assert instead, but an error handling fix doesn't deserve the mental effort of pondering "is it possible that in some theoretical usage pattern of yosys something will emit a mux with zero width Y port"

PR #5766, opened 2026-03-19
Upgrade CI actions

mmicko:
Upgrades CI actions to remove warnings. Some actions like microsoft/setup-msbuild@v2 still do not have modern upgrade. Anyway less warnings -> better

NOTE: Private runners still do not support this new version


mmicko:
For `awalsh128/cache-apt-pkgs-action`  there is  https://github.com/awalsh128/cache-apt-pkgs-action/pull/193 pending, and this is what appears at strange `actions/cache` warnings

PR #5768, opened 2026-03-20
Update tests and test infrastructure

mmicko:
- Moves test make rules to a new file
- Created gen_tests_makefile.py to make Makefile generators as 
- Use generate_mk.py in each sub directory
- All tests were moved except `functional` that are done as pytest
- Revived some of tests for `realmath` and `bind`
- Enabled ` arch/fabulous` that was never executed before
- Made wrapper hide all stdout/stderr outputs
- Made `blif/gatesi` test actually check output
- Disabled `svinterface/svinterface_at_top` since it was actually failing, need fixing that test


widlarizer:
Wait, this was failing for me locally :O

widlarizer:
[log.txt](https://github.com/user-attachments/files/26960980/log.txt)

Ravenslofty:
`make[2]: /bin/bash: No such file or directory`
something something `/usr/bin/env bash`

widlarizer:
Oh, that'll do it, yes. I was used to grepping for `make:` to find failures and that doesn't work anymore

mmicko:
@widlarizer can you please change in  `tests/aiger/generate_mk.py` bash to be used with `SHELL := /usr/bin/env bash` ? That is only place where it is enforced

widlarizer:
On another note though, was the runner supposed to catch this kind of error and report it as FAIL? That's what I would expect, that instead of grepping for "make:" I could grep for "FAIL" but the make process errored out without pointing me to the problem which is confusing

widlarizer:
Also, is there a way of running tets in a directory? `make -C tests/aiger` won't rebuild the makefiles. If I run it after `make clean` I get `make: *** No targets specified and no makefile found.  Stop.`. Previously, it was possible to do `make makefile-tests/tests/aiger`. Is there something like `make -C tests makefile-tests/tests/aiger`?

mmicko:
@widlarizer since makefiles are generated you would need first to run `make prep` in tests directory or `make -C tests prep` in root, and after that makefiles and generated tests are all there so you can run them, also possible to go in specific directory and just run `make`
Regarding catching errors, thing is it is not allmighty so it can catch errors running scripts, but here it failed earlier executing shell from makefile so that was not able to be processed

PR #5771, opened 2026-03-23
verilog frontend/backend: Fix multiple signedness issues

### LINKED ISSUE #5745: `read_verilog` problem with `signed` keyword and parameters
### Version

Yosys 0.63

Linux

### Reproduction Steps

Run the following script with/without verific, be confused about difference.
```
read -vlog2k << EOT
module mod #(parameter p2=3) (output k);
  wire [8:0] m = (1 << 5) - (1 << 8);
  assign k = $signed(m) < p2;
endmodule
EOT
hierarchy -top mod -chparam p2 11
sat -prove k 1 
```

### Expected Behavior

`verific` and `read_verilog` behaviour should match.  Substituting in values we get `m=9'b100100000`.
11 is a positive number, (when converted to signed) `m` is a negative number => `k` must be 1, which is what `verific` gives us.

### Actual Behavior

`read_verilog` says `k=0`.

Changing to `assign k = $signed(m) < $signed(p2);` gives 1 for both frontends, as does changing the default p2 directly instead of using `-chparam`.  Calling `-chparam p2 11` when the default value is 11 *still produces the failure*.

The first workaround made me think that this might be the same as #4402, but the second workaround makes me less sure.

If we call `dump t:$lt` with the default of `p2=11` and no `-chparam` we get the following:
```
  cell $lt $lt$<<EOT:3$1
    parameter \A_SIGNED 1
    parameter \B_SIGNED 1
    parameter \A_WIDTH 32
    parameter \B_WIDTH 32
    parameter \Y_WIDTH 1
    connect \A { \m [8] \m [8] \m [8] \m [8] \m [8] \m [8] \m [8] \m [8] \m [8] \m [8] \m [8] \m [8] \m [8] \m [8] \m [8] \m [8] \m [8] \m [8] \m [8] \m [8] \m [8] \m [8] \m [8] \m }
    connect \B 11
    connect \Y $lt$<<EOT:3$1_Y
  end
```

But *with* the `-chparam` we get this instead:
```
  cell $lt $lt$<<EOT:3$1
    parameter \A_SIGNED 0
    parameter \B_SIGNED 0
    parameter \A_WIDTH 32
    parameter \B_WIDTH 32
    parameter \Y_WIDTH 1
    connect \A { 23'00000000000000000000000 \m }
    connect \B 11
    connect \Y $lt$<<EOT:3$1_Y
  end
```

gussmith23:
Should close #5745  and #4402.

### DIFF
--- a/backends/verilog/verilog_backend.cc
+++ b/backends/verilog/verilog_backend.cc
@@ -456,21 +456,22 @@ void dump_wire(std::ostream &f, std::string indent, RTLIL::Wire *wire)
 		if (wire->attributes.count(ID::single_bit_vector))
 			range = stringf(" [%d:%d]", wire->start_offset, wire->start_offset);
 	}
+	std::string sign = wire->is_signed ? " signed" : "";
 	if (wire->port_input && !wire->port_output)
-		f << stringf("%s" "input%s %s;\n", indent, range, id(wire->name));
+		f << stringf("%s" "input%s%s %s;\n", indent, sign, range, id(wire->name));
 	if (!wire->port_input && wire->port_output)
-		f << stringf("%s" "output%s %s;\n", indent, range, id(wire->name));
+		f << stringf("%s" "output%s%s %s;\n", indent, sign, range, id(wire->name));
 	if (wire->port_input && wire->port_output)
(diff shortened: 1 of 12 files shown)


hexratcc:
Seeing as we've split the PR into two separate PR's I'm closing this one.

PR #5774, opened 2026-03-24
sim: cache sigmap in register_output_step_values

abtondapu:
register_output_step_values() calls get_state(wire) for every signal on every simulation step. get_state() runs sigmap() internally, which re-resolves the same wire-to-sigbit mapping every step even though the mapped SigSpec for a registered wire is fixed once the SimInstance is constructed. For designs with many signals and long VCD traces, this redundant work adds up.

This change caches the sigmap result per signal at initialization time. A signal_entry_t struct replaces the pair in signal_database, adding a mapped_sig field that stores the pre-mapped SigSpec. A new get_state_mapped() helper resolves state from a pre-mapped signal, and get_state() is refactored to call it internally so the bit-resolution logic is not duplicated.

At runtime, register_output_step_values() calls get_state_mapped(entry.mapped_sig) instead of get_state(wire), skipping sigmap entirely on the hot path. Benchmarked locally across multiple designs of different sizes, showing an average simulation speedup of ~12 percent.

No new tests are required. This is a performance-only refactor with identical functional behavior. get_state() delegates to get_state_mapped() so all existing callers produce the same results. The existing tests/sim/ suite passes with no regressions. To verify, run sim with VCD/FST output on any design and confirm identical output.

### DIFF
--- a/passes/sat/sim.cc
+++ b/passes/sat/sim.cc
@@ -216,7 +216,13 @@ struct SimInstance
 
 	std::vector<Mem> memories;
 
-	dict<Wire*, pair<int, Const>> signal_database;
+	struct signal_entry_t {
+		int id;
+		Const last_value;
+		SigSpec mapped_sig;
+	};
+
+	dict<Wire*, signal_entry_t> signal_database;
 	dict<IdString, std::map<int, pair<int, Const>>> trace_mem_database;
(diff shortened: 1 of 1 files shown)


abtondapu:
CI failure looks like a runner infra issue (lost communication), not related to this change. Could someone re-run the workflow?

mmicko:
@tondapusili Thanks a lot for contributions so far.

PR #5782, opened 2026-03-27
abc: remove -fast (again)

Ravenslofty:
https://github.com/YosysHQ/yosys/pull/5758, with test suite passing (for me at least). SBY's dependence on `-fast` for AIG mapping has also been fixed in the meantime.

### DIFF
--- a/passes/techmap/abc.cc
+++ b/passes/techmap/abc.cc
@@ -35,12 +35,6 @@
 #define ABC_COMMAND_SOP "strash; &get -n; &fraig -x; &put; scorr; dc2; dretime; strash; dch -f; cover {I} {P}"
 #define ABC_COMMAND_DFL "strash; &get -n; &fraig -x; &put; scorr; dc2; dretime; strash; &get -n; &dch -f; &nf {D}; &put"
 
-#define ABC_FAST_COMMAND_LIB "strash; dretime; map {D}"
-#define ABC_FAST_COMMAND_CTR "strash; dretime; map {D}; buffer; upsize {D}; dnsize {D}; stime -p"
-#define ABC_FAST_COMMAND_LUT "strash; dretime; if"
-#define ABC_FAST_COMMAND_SOP "strash; dretime; cover {I} {P}"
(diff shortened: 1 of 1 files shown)


widlarizer:
In #5758 I approved the removal of `-fast` since the diff showed it's not used anywhere, but it was used somewhere, in fact in `synth` which I occasionally run as the stock ASIC synthesis command for benchmarking purposes. The CI didn't catch that fact due to CI issues, those got fixed in #5761, but also in that PR, `synth.cc` got changed to no longer use `abc -fast`, which I didn't really get to check the impact of. The effect is `synth` on this multiplier: 
```v
module top #(parameter WIDTH = 64)(
    input signed [WIDTH-1:0] a, input signed [WIDTH-1:0] b,
    output signed [WIDTH*2-1:0] p);
assign p = a * b;
endmodule
```
Goes from
```
End of script. Logfile hash: f3fe394d7d, CPU: user 1.17s system 0.05s, MEM: 101.90 MB peak
Yosys 0.63 (git sha1 70a11c6bf, ccache clang++ 16.0.6 -fPIC -O3 -flto=thin)
```
to
```
End of script. Logfile hash: 5b24022561, time: 92.23s, user: 1.27s, system: 0.09s, MEM: 111.69 MB peak
Yosys 0.64 (git sha1 6d2c445ae, ccache clang++ 16.0.6 -fPIC -O3 -flto=thin)
```
Which might be a valid performance result of fraiging vs not fraiging a big ole multiplier. Don't mind the lack of a wall clock in 0.63, `time` reports it takes 1.91s. ORFS wasn't impacted since it provides its own abc script. Also, `synth` doesn't exactly promise the user anything, so I guess this is fine? It just really caught me off guard and I thought it was a hang. I'll have to adjust my benchmarking to match.

Ravenslofty:
In an ideal world such a multiplier would probably be a box, but yeah, a 64x64 multiplier is going to result in a lot of AIG to traverse.

widlarizer (approved):

PR #5803, opened 2026-04-14
Fix $specrule port naming inconsistency

### LINKED ISSUE #5800: $specrule has inconsistent enable port naming
### Version

all

Linux

### Reproduction Steps

see Actual Behavior

### Expected Behavior

internal port direction metadata (CellTypes), help messages, internal cell library models shouldn't be wrong about cell interfaces

### Actual Behavior

note the EN_SRC vs SRC_EN discrepancy:

```
yosys> help $specrule

    $specrule (EN_SRC, EN_DST, SRC, DST)
```
- `techlibs/common/simlib.v` says `module \$specrule (EN_SRC, EN_DST, SRC, DST);`
- `kernel/celltypes.h` says `setup_type(ID($specrule), {ID::EN_SRC, ID::EN_DST, ID::SRC, ID::DST}, pool<RTLIL::IdString>());`

but:

- `InternalCellChecker` in `kernel/rtlil.cc` says `port(ID::SRC_EN, 1);`
- `read_verilog` in `frontends/verilog/verilog_parser.y` says `cell->children.back()->str = "\\SRC_EN";`
- `write_verilog` in `backends/verilog/verilog_backend.cc` says `dump_sigspec(f, cell->getPort(ID::SRC_EN));`

widlarizer:
Fix #5800. The `InternalCellChecker` view of the cell is taken as canon and anything inconsistent is changed to match


hexratcc (approved):

KrystalDelusion:
This PR appears to have introduced a crash when synthesizing ibex (with Verific):
```
2.7. Executing OPT_CLEAN pass (remove unused cells and wires).
Finding unused cells or wires in module \ibex_core..
  what():  dict::at()
Aborted (core dumped)
```

Currently minimizing.

KrystalDelusion:
Looks like the problem was solved by rebuilding `cells_temp.o`, which smells like a missing dependency (potentially on `celltypes.h`?), odd that it didn't appear to impact anything else though (and I don't see anything that would make `cells_temp.o` different in that regard)

widlarizer:
Probably gets back to `constids.inc`. Sorry about that, I'm not super excited about fixing those dependencies, that file moves rarely. I'm a bit religious about `make clean` when switching branches, with ccache it's minimally painful

KrystalDelusion:
I should try to get in the habit too, it's not the first time I've had a problem from not doing the clean (and I already use `ccache`)

PR #5807, opened 2026-04-16
Adjusted dot call to avoid pdflatex rejecting input due to too high resolution.

petterreinholdtsen:
Set fairly randomly picked size when generating dot graphs based on example I came across, to avoid the following error from pdflatex when building documentation:

```
  [46 <./rdata_map_luts.pdf>]
  ! Dimension too large.
```

I am not sure which size make sense, nor how to decide which value are best.

This issue is a fix for a build problem reported in <URL: https://bugs.debian.org/1133986 >, which include a link to the build log of the failed build.

### DIFF
--- a/docs/source/_images/Makefile
+++ b/docs/source/_images/Makefile
@@ -33,7 +33,7 @@ TEX_FILES := $(shell find . -name *.tex)
 all_tex: $(TEX_FILES:.tex=.pdf) $(TEX_FILES:.tex=.svg)
 
 %.pdf: %.dot
-	$(FAKETIME) dot -Tpdf -o $@ $<
+	$(FAKETIME) dot -Tpdf -Gsize="9x15!" -o $@ $<
 
 %.pdf: %.tex
 	cd $(@D) && $(FAKETIME) pdflatex $(<F) --interaction=nonstopmode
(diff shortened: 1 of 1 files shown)


KrystalDelusion:
[The image in question](https://yosyshq.readthedocs.io/projects/yosys/en/latest/_images/rdata_map_luts.svg) is indeed quite large (and there is a bigger one right after it), but I can't reproduce the error you're seeing with a local `make docs DOC_TARGET=latexpdf`.  Do you know if there is anything in the way that the latexpdf build is being done that is modified from the standard Yosys repo?  Or any latexpdf settings?

In principle I don't have a problem with this change, though I will point out that the output pdf (at least when I build it with my local clone) has `Page Size: 8.5 × 11 in (Letter, portrait)`, which seems like a better (less arbitrary) maximum image size (I'm sure you could find the actual maximum size within the margins, but the page size seems good enough for me).

petterreinholdtsen:
[KrystalDelusion]
> Do you know if there is anything in the way that the latexpdf build is
> being done that is modified from the standard Yosys repo?  Or any
> latexpdf settings?

I do not know.

As far as I can tell from the debian patches listed in
<URL: https://tracker.debian.org/pkg/yosys >, only
0026-Quiet-write-rst-command-ref-manual.patch and
0031-Set-pythonpath-for-usage-docs.patch touch files under docs/, and
none of them seem relevant to adjust how pdflatex is used.

-- 
Happy hacking
Petter Reinholdtsen

PR #5809, opened 2026-04-17
opt: Remove O(n²) opt routines across the codebase for pmux

QuantamHD:
$pmux have bad runtime behavior with large cases and output bit widths. On one Google test case synthesis takes 57 hours with these changes it takes 350s.

* Replacing SigSpec::extract in a loop by taking slices from a SigBit vector
* In generate loop replace O(n^2) linear scan with hashmap, in combination with pmux O(n^2) loop in techmap.v this loop is O(n^4)

### DIFF
--- a/passes/opt/opt_muxtree.cc
+++ b/passes/opt/opt_muxtree.cc
@@ -21,11 +21,9 @@
 #include "kernel/sigtools.h"
 #include "kernel/log.h"
 #include "kernel/celltypes.h"
+#include <cstddef>
 #include <stdlib.h>
 #include <stdio.h>
-#include <unordered_map>
-#include <unordered_set>
-#include <set>
 
 USING_YOSYS_NAMESPACE
 PRIVATE_NAMESPACE_BEGIN
@@ -128,7 +126,7 @@ struct OptMuxtreeWorker
 		// In case of $pmux, port B is multiple slices, concatenated, one per bit of port S
 		for (int i = 0; i < GetSize(sig_s); i++) {
 			RTLIL::SigSpec sig = sig_b.extract(i*GetSize(sig_a), GetSize(sig_a));
-			RTLIL::SigSpec ctrl_sig = assign_map(sig_s.extract(i, 1));
+			RTLIL::SigSpec ctrl_sig = assign_map(SigSpec{sig_s[i]});
 			portinfo_t portinfo = used_port_bit(sig, this_mux_idx);
 			portinfo.ctrl_sig = sig2bits(ctrl_sig, false).front();
 			portinfo.const_activated = ctrl_sig.is_fully_const() && ctrl_sig.as_bool();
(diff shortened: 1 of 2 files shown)


widlarizer:
- opt_muxtree: The analysis here was probably originally written with the expectation that design tend to have modules with mux regions that don't dominate the module cell count, so the activity information is held in a sparse format. But then, `bit2info` marks all (sigmap representatives of) all port bits. So I'm on board with the densification improvement still makes sense. Happy that the refactor of this pass was used for something
- not using extract in a loop over all "ports" is definitely the way to go. I used to chalk all this inefficiency down to `$pmux` and other concatenation-based cells being horribly designed due to making them fit a Verilog techmap based mindset but this helps a lot. It would be best to have a wrapper like `kernel/macc.h` or `kernel/ff.h`. Might take a shot at that
- I have to play around with the AST genblock changes some more

QuantamHD:
@widlarizer I'd be totally fine to hand this one off to you if you have time. I tested this on our internal designs pass, but I'm not confident about the AST changes because we only use the AST to parse the techmap.v file, and I know it's used as the elaborator for the actual Yosys frontend which we don't test/use

widlarizer:
With #5832 we'll have fast `SigSpec::extract` again so I've removed the bits of the code that were avoiding extracting B subports

widlarizer (approved):

maliberty:
@QuantamHD  I asked AI for a review and it notes a potential QoR issue:

 2. opt_reduce.cc — ⚠️  dropped optimization (QoR regression, not a correctness 
  bug)

  The old opt_pmux seeded handled_sig.insert(sig_a) so that any $pmux B-slice
  equal to the default A input was dropped (its select bit removed). When such a
  port's select is hot, the output equals A anyway, so dropping it is
  value-preserving — and when all B slices equal A, the whole pmux collapsed to
  a wire (new_sig_s.size()==0 → connect Y to A, remove cell).

  The new dict<SigSpec, vector<SigBit>> grouping has no comparison against 
  sig_a, so B-slices equal to A are now kept.

  - Correctness: preserved. Keeping a redundant B_i == A port produces identical
  output; $pmux single-hot semantics are unchanged.
  - QoR: a real regression — redundant ports matching the default are no longer
  eliminated, and pmuxes that previously collapsed entirely now survive (won't
  be cleaned up by opt_clean/opt_merge either).

  Action: confirm with the author this is intentional (acceptable trade for the
  perf win) and ideally re-add cheaply inside the new loop, e.g. skip insertion
  when this_b == sig_a. Worth a regression test asserting a pmux with a B_i == A
  case still gets reduced. The PR description doesn't mention this behavior
  change.

  (Ordering is fine: Yosys dict preserves insertion order with no erases, so 
  new_sig_b/new_sig_s are built in first-appearance order, matching the old code
  — output is deterministic.)

PR #5813, opened 2026-04-17
abc/abc_new: cache parsed liberty files as .scl to avoid re-parsing across invocations

hexratcc:
When `abc` is invoked with `-liberty`, the liberty file paths are embedded into the generated ABC script as `read_lib` commands. ABC then re-parses every liberty file from scratch on each invocation. In hierarchical flows where `abc` runs once per module, this means the same liberty files are parsed N times for N modules. For large libraries this can drastically slow down the entire flow.

This PR aims to fix this issue by adding a cache for the liberty files. The cache is implemented as a merged `scl` file, which is generated on the first `abc` invocation, subsequent runs use the merged `scl`. From my research, `scl` should preserve everything ABC uses for tech mapping, while stripping the majority of the unneeded data, drastically reducing file size, and being much faster to process.

In order to test perf I've ran the following benchmark: `abc -liberty` invoked on a couple simple modules with a varying amount of liberty files, resulting in a ~6x speedup. Module contents are intentionally trivial because the bottleneck being addressed is parse time, not mapping time. See results below:

image

I've also tested an implementation which modified ABC so that it's capable of merging `scl` files via `read_scl`, the idea behind this being that we could cache separate files, not the entire list, which could prove beneficial when a flow invokes `abc` with different sets of liberty files, but the speedup went down to about ~4x. I believe that these flows are less common, so I chose to go with the above method. 

TODO:
- [x] abc9_exe support
- [x] check equivalence of formats
- [x] verify dont_use_cells

### DIFF
--- a/passes/techmap/liberty_cache.h
+++ b/passes/techmap/liberty_cache.h
@@ -11,17 +11,18 @@ YOSYS_NAMESPACE_BEGIN
 
 /*
  * convert_liberty_files_to_merged_scl() - Convert multiple Liberty files to a single merged SCL cache file.
- * @liberty_files - Vector of liberty file paths to merge
- * @abc_exe - Path to ABC executable for conversion
+ * @liberty_files: Vector of liberty file paths to merge
+ * @dont_use_args: Pre-built ABC -X flags string
+ * @abc_exe: Path to ABC executable for conversion
  *
  * Return: Path to merged SCL cache file, or empty string if conversion fails
  */
-inline std::string convert_liberty_files_to_merged_scl(const std::vector<std::string> &liberty_files, const std::string &abc_exe)
+inline std::string convert_liberty_files_to_merged_scl(const std::vector<std::string> &liberty_files, const std::string &dont_use_args, const std::string &abc_exe)
 {
 	if (liberty_files.empty())
 		return "";
 
-	// Sort filenames to ensure consistent hash regardless of order
+	// Sort to ensure consistent hash regardless of order
 	std::vector<std::string> sorted_files = liberty_files;
 	std::sort(sorted_files.begin(), sorted_files.end());
 	std::string hash_input;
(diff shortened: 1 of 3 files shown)


widlarizer:
cc @QuantamHD @mikesinouye

widlarizer:
I was wondering if SCL could be accepted by Yosys in place of Liberty so that Liberty is parsed only once across both programs, but unfortunately since abc-written SCL skips sequential cells, Yosys couldn't map flip flops with `dfflibmap`

mikesinouye:
This is excellent! Thank you for working on this feature.

> when a flow invokes abc with different sets of liberty files

This is not a current use case of ours so optimizing for it should not be necessary.

Ravenslofty (approved):

passes/techmap/liberty_cache.h
```diff
@@ -17,6 +17,9 @@ YOSYS_NAMESPACE_BEGIN
  *
  * Return: Path to merged SCL cache file, or empty string if conversion fails
  */
+#if defined(YOSYS_LINK_ABC)
+inline std::string convert_liberty_files_to_merged_scl(const std::vector<std::string> &, const std::string &, const std::string &) { return ""; }
```
widlarizer: Wait, is it impossible to make this function work with YOSYS_LINK_ABC as well? It's a legitimate option on other platforms, too

hexratcc: Ah, my thinking was that this is only related to WASI, where this optimization wouldn't really be needed and this would be a quick fix. Will fix using `Abc_RealMain`. Thanks for the headsup.

passes/techmap/liberty_cache.h
```diff
@@ -0,0 +1,111 @@
...
+
+	// SCL filename
+	std::string first_dir;
+	size_t last_slash = liberty_files[0].find_last_of("/\\");
+	unsigned int hash = 0;
+
+	if (last_slash == std::string::npos) {
+		first_dir = ".";
+	} else {
+		first_dir = liberty_files[0].substr(0, last_slash);
```
mikesinouye: In most of our flows, the liberty file location and the CWD are non-writeable as you're often inside a container, with some other dir for allocated scratch space. Would it be possible to store this file in the same (or similar) abc tmpdir that's created? This could be done in a follow-up change too to land this one.

https://github.com/YosysHQ/yosys/blob/main/kernel/io.cc#L149

PR #5816, opened 2026-04-19
Rewrote Linux edition of proc_self_dirname() to handle any symlink length.

petterreinholdtsen:
This make sure the method work also when the program is located in deep or long file paths, longer than both PATH_MAX and "getconf PATH_MAX .".  Use the same code on GNU Hurd, where it now work.

I am not sure how to test this in a platform independent way.

### DIFF
--- a/kernel/yosys.cc
+++ b/kernel/yosys.cc
@@ -471,17 +471,30 @@ struct TclPass : public Pass {
 
 #endif
 
-#if defined(__linux__) || defined(__CYGWIN__)
+#if defined(__linux__) || defined(__CYGWIN__) || defined(__gnu_hurd__)
 std::string proc_self_dirname()
 {
-	char path[PATH_MAX];
-	ssize_t buflen = readlink("/proc/self/exe", path, sizeof(path));
+	std::string path(4096, '\0');
+	ssize_t buflen = -1;
+	// Double until sucess, while avoiding endless loop.  Give up
+	// when symlink is longer than 4096*(2^30) = 4398046511104
+	// bytes.
+	for (int tries = 30; 0 < tries; tries--) {
+		buflen = readlink("/proc/self/exe", path.data(), path.size());
+		if (buflen < (ssize_t)path.size())
+			break;
+		else
+			path.resize(path.size() * 2);
+	}
 	if (buflen < 0) {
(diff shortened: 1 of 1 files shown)


whitequark (commented):
Could you not do this with just `readlink` by doubling the buffer size whenever it gets truncated? That avoids all the ifdefs and the testing issues (I'm not comfortable having ifdefs for Hurd given we do not test on this platform nor does any of the maintainers even run it.)

petterreinholdtsen:
[Catherine]
> Could you not do this with just `readlink` by doubling the buffer size
> whenever it gets truncated? That avoids all the ifdefs and the testing
> issues (I'm not comfortable having ifdefs for Hurd given we do not
> test on this platform nor does any of the maintainers even run it.)

Sure, if that is a more acceptable solution.  I replaced the commit in
the pull request with such approach.

-- 
Happy hacking
Petter Reinholdtsen

whitequark (commented):
I think you can create an `std::string` here, resize it to the expected buffer size, then resize down to the returned value (which makes the code quite a bit clearer imo)

petterreinholdtsen:
[Catherine]
> I think you can create an `std::string` here, resize it to the
> expected buffer size, then resize down to the returned value (which
> makes the code quite a bit clearer imo)

I do not understand your proposal.

-- 
Happy hacking
Petter Reinholdtsen

whitequark:
You're repeatedly creating and destroying a unique_ptr only to convert it to a std::string later. You could do something like

```
std::string result(256, '\0');
while (1) {
  int length = readlink(path, result.data(), result.size());
  if (length < result.size) {
    result.resize(length);
  } else {
    result.resize(result.size() * 2);
  }
}
return result;
```

(for illustration only, actual PR needs to handle errors as you're doing it currently)

petterreinholdtsen:
[Catherine]
> You're repeatedly creating and destroying a unique_ptr only to convert
> it to a std::string later. You could do something like

Ah, then I understand.  Just pushed third try to pull request using this
approach.

-- 
Happy hacking
Petter Reinholdtsen

whitequark (commented):
LGTM, thanks!

widlarizer (approved):
Looks good

PR #5817, opened 2026-04-20
clockgate: reject $sdffe to fix priority handling

widlarizer:
Alternative to #5812

- [x] TODO add test coverage


widlarizer:
The new formal test covering `$sdffe` fails if I undo the `clockgate` code change, as it should

hexratcc (approved):

PR #5819, opened 2026-04-21
xprop: ignore $scopeinfo cells

nakengelhardt:
Reduce some warning spam of the form `Warning: Unhandled cell gold ($scopeinfo) during maybe-x marking`. There's nothing `xprop` needs to do for these cells anyway.


nakengelhardt:
[sc-385]

widlarizer (approved):

PR #5821, opened 2026-04-21
hierarchy: refactor

widlarizer:
Chop up hierarchy.cc into digestible pieces. Required for improving hierarchy with signorm (#5804) later. If I wanted to be consistent with opt_clean refactoring, I would have to create a directory at `passes/hierarchy/hierarchy/`, so it's `passes/hierarchy/util/` for now. No public interfaces were harmed


widlarizer:
It's interesting to note that almost all of the complexity of `hierarchy` comes from the fact that it has to handle langauge features. It's a deferred phase of elaboration, but there is zero separation. Load-bearing attributes are just glued onto RTLIL instead of proper modeling. Another direct consequence of having a simple pass interface for scripting - frontends should call into hierarchy or the other way around in some way, rather than hierarchy etc uncontrollably accumulating this flotsam

whitequark:
I hope the RPC frontend could also be better integrated into the flow after the refactor.

widlarizer:
@whitequark What would making it better for RPC entail?

whitequark:
> What would making it better for RPC entail?

It's been a few years, but I remember that the RPC frontend had to plug into `hierarchy` and it was kind of a pain to adapt it for the elaboration hook intended for `read_verilog`. One issue was that it had to mangle names of everything returned by the frontend to avoid name collisions. Perhaps a cleaner implementation of `hierarchy` would make this less painful, seeing as this is morally something best done in `hierarchy` so all the metadata could be updated, etc.

widlarizer:
[sc-497]

widlarizer:
The CI went green with tests skipped, how? Usually I run tests locally since my machine outruns the runners, but I skipped that for a draft refactor and went on when I saw green... Weird

widlarizer:
The part that I'm interested in (checking port directionality correctness) is apparently intertwined with the systemverilog interface resolution code. I'm interested in it because I want to later add directional resolution for signorm $connect cells, and it's not a great idea to take more passes through the design needlessly, even if that would be a clean way to separate systemverilog from fundamental hierarchy logic

Issue #5805, opened 2026-04-16
Consider patches used by Debian in a future release?


petterreinholdtsen:
I notice from <URL: https://tracker.debian.org/pkg/yosys > using the link "debian patches" there are currently 17 patches applied to the Debian packages.  Are you aware of these patches?  Perhaps something to consider including in a future release?  I did not make any of the patches myself, just wanted to make sure you as upstream are aware of them.

whitequark:
The propensity of Debian packagers to apply questionably-conceived and written patches to YosysHQ software has historically been enough of a problem that I recommend never using Yosys tooling from Debian.

widlarizer:
I'd be happy to discuss patches if their authors or maintainers understand them and can champion them to us one by one. We've previously removed the need for any nixpkgs patches. At a brief glance, none of the Debian patches seem particularly worthwhile. Consider the patches considered and feel free to submit PRs with any of them you care about

petterreinholdtsen:
[Emil J]
> I'd be happy to discuss patches if their authors or maintainers
> understand them and can champion them to us one by one.

Good.  Where should patches for abc be submitted?
<URL: https://github.com/YosysHQ/abc > do not accept pull requests, and 
<URL: https://github.com/berkeley-abc/abc > do not contain the same code
base, making it very unclear to me what should be done with patches
related to the abc repository.

I've uploaded 0.64 to Debian experimental with a few patches to get 32
bit builds working, among other things.  The current status can be seen
on <URL: https://buildd.debian.org/status/package.php?p=yosys&suite=experimental >.
Unfortunately big endian machines and i386 have a failing test run.

-- 
Happy hacking
Petter Reinholdtsen

widlarizer:
AFAIK the YosysHQ abc fork is meant to be minimally different from Berkeley upstream, but I see we have more patches there than I expected. The reason we disabled PRs there is that people kept submitting functional improvements to abc to our fork which isn't the right approach, we aim to have only patches we need for yosys to work well. So the answer is "it depends". Patches specific to Yosys use cases of abc go on our fork (sometimes temporarily before upstreaming). Patches that improve or fix things in abc go directly upstream.

[...]

PR #5825, opened 2026-04-22
Added locking around compiling in autotest.sh to avoid text busy race.

petterreinholdtsen:
Added locking around compiling in autotest.sh to avoid text busy race.

Patch by Daniel Gröber via Debian.

See also issue #5805.

### DIFF
--- a/tests/tools/autotest.sh
+++ b/tests/tools/autotest.sh
@@ -25,9 +25,11 @@ firrtl2verilog=""
 xfirrtl="../xfirrtl"
 abcprog="$toolsdir/../../yosys-abc"
 
+exec {lock}<"$toolsdir"; flock "$lock" 1>&2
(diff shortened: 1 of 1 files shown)


widlarizer:
Do you have an example of what this solves?

petterreinholdtsen:
[Emil J]
> Do you have an example of what this solves?

Sorry, I do not.  I know it originate from this version of yosys in
Debian:

yosys (0.33-2) unstable; urgency=medium

  * Fix text file busy due to autotest.sh gcc race
  * Fix superfluous-file-pattern in d/copyright

 -- Daniel Gröber ***@***.***>  Wed, 06 Sep 2023 11:04:15 +0200

I suspect it trigger when running tests in parallell.

-- 
Happy hacking
Petter Reinholdtsen

widlarizer:
I see, kind of makes sense, I can't rule out autotest.sh might get run in parallel causing this race

PR #5827, opened 2026-04-22
Support positional assignment patterns for unpacked arrays

cdleary:
## What are the reasons/motivation for this change?

I am attempting to support unpacked array assignment patterns (XLS-emitted) in the Yosys SV frontend.

SystemVerilog can assign a whole unpacked array with a positional assignment pattern:

```systemverilog
wire lit[2];
assign lit = '{1'h1, 1'h0};
```

XLS can also emit multidimensional unpacked array literals as nested SystemVerilog assignment patterns, for example:

```systemverilog
wire lit[2][2];
assign lit = '{'{1'h1, 1'h0}, '{1'h0, 1'h1}};
```

SystemVerilog positional assignment patterns can initialize or assign whole unpacked arrays, including multidimensional unpacked arrays; see IEEE Std 1800-2017, Section 10.9, "Assignment patterns." The existing implementation handled simple flat patterns, but two legal cases needed support: runtime expressions in the first pattern element, and nested positional patterns for multidimensional arrays.

## Explain how this is achieved.

`AST_ASSIGN_PATTERN` now has its own hierarchy-flag handling so parameter-context state is propagated uniformly to all pattern children instead of treating the first child like a replication count.

Array assignment expansion now validates and walks nested assignment patterns dimension by dimension for multidimensional unpacked arrays, while preserving the existing direct array and ternary array assignment behavior.

For multidimensional patterns, each nested assignment-pattern level corresponds to one unpacked array dimension. During scalarization, the RHS element is selected by walking the pattern according to the current multidimensional array position.

## Make sure your change comes with tests. If not possible, share how a reviewer might evaluate it.

This change includes regression coverage in `tests/svtypes/array_assign.sv` for:
- a runtime function call in the first assignment-pattern element
- a nested positional assignment pattern assigned to a `[2][2]` unpacked array
- a multidimensional assignment pattern whose elements are compatible row arrays
- a procedural blocking assignment pattern whose RHS reads from the LHS array

It also adds `tests/svtypes/array_assign_flat_multidim_pattern.ys` to verify that a flattened positional pattern is rejected for a multidimensional unpacked array, since the top-level pattern items must match the outer unpacked dimension.

Verified with:

```sh
make -j4
cd tests/svtypes && bash run-test.sh && make -f run-test.mk svtypes-array_assign.sv svtypes-array_assign_flat_multidim_pattern.ys
./yosys -q -p "read_verilog -sv tests/svtypes/array_assign.sv; hierarchy -top top; proc; opt"
```


widlarizer (approved):
Looks good, the existing checks and simplifications are kept but extended with the new node type

PR #5828, opened 2026-04-22
Remove FHS dependency by replacing /bin/bash with /usr/bin/env bash

widlarizer:
Fixes #5768 on nixOS particularly due to `tests/aiger/generate_mk.py`, verified locally on nixOS

### DIFF
--- a/.github/workflows/wheels/cibw_before_all.sh
+++ b/.github/workflows/wheels/cibw_before_all.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -e -x
 
 # Build-time dependencies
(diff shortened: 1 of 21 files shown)


mmicko (commented):
While you are at it, there are actually 3 more files that needs updating `techlibs/sf2/NOTES.txt`, `libs/minisat/00_UPDATE.sh` and `libs/fst/00_UPDATE.sh`

widlarizer:
hmm, I'm pretty sure /bin/sh can be relied on as much as /usr/bin/env

widlarizer:
I've asked for feedback elsewhere and asking env is supposedly proper for `sh` as well since that can/is overridden

whitequark:
> hmm, I'm pretty sure /bin/sh can be relied on as much as /usr/bin/env

I think it might not be: https://github.com/NixOS/nix/issues/6081

PR #5830, opened 2026-04-22
 simplify: fast lookup for expand_genblock

widlarizer:
Speeds up ingesting cases that emit large `$pmux` cells. Split out from #5809

### DIFF
--- a/frontends/ast/ast.h
+++ b/frontends/ast/ast.h
@@ -262,7 +262,10 @@ namespace AST
 		bool simplify(bool const_fold, int stage, int width_hint, bool sign_hint);
 		void replace_result_wire_name_in_function(const std::string &from, const std::string &to);
 		std::unique_ptr<AstNode> readmem(bool is_readmemh, std::string mem_filename, AstNode *memory, int start_addr, int finish_addr, bool unconditional_init);
+		dict<std::string, AstNode*> generate_module_fast_lookup_for_genblock_expansion();
+		void add_to_module_fast_lookup(dict<std::string, AstNode*>& module_fast_lookup, AstNode* node);
 		void expand_genblock(const std::string &prefix);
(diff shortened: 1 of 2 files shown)


QuantamHD:
Thanks for merging in the other PRs. This is the last one blocking us. Is there anything we could do to assist with the testing of this one to help get it over the line?

widlarizer:
Since there's other proposed changes to the frontend I've thought about how we could increase the velocity there and started a refactor that lifts the entire AST layer into a more typed design by constraining field access based on ndoe type and gradually degrading the AstNode struct into an implementation detail that is never touched in the parser or simplify or genrtlil. While it's really promising it also needs more time, so I'm going to review this as-is before I put that up

widlarizer:
Apparently, the AST-based lookup in expand_genblock is dead code, at least as far as our test suite is concerned. It's being hit in `tests/verilog/prefix.sv` but not covered by the actual test.  Does reverting #4235 have the same perf impact you're looking for? The code that was copied from simplify into expand_genblock is just incorrect in the context of expand_genblock, failing to correctly assess when an autowire is needed due to the `// named blocks pick up the current prefix and will expanded later` skip, which is preventing the scope to be resolved before checking identifiers hierarchically referencing wires defined in those blocks.

Another consideration: this PR has an eagerness aspect that might hit some specific cases, particularly, well-formed verilog with labels or autowires. I don't have a clear idea of how much of an issue that might be but preparing the fast lookup whenever calling expand_block non-recursively means traversing the module level / top level AST even if you don't actually end up using the lookup. It would be safer if the dict was persistent as it already is in this PR, but on-demand, only constructed when first attempted to be accessed.

Ideally I'd just fix all the bugs I'm seeing here but realistically I should wait with more archaeology and surgery of the extremely concerning (from the point of both correctness and perf) scoping system to my broader rework of the AST layer.

PR #5831, opened 2026-04-23
Make sure co-simulation only uses integer arithmetic

### LINKED ISSUE #5822: Self testing tests/sim of yosys 0.64 fail on i386 (ERROR: Expected log pattern 'Co-simulating cycle 40' found 1 time(s), instead of 2 time(s) !)
### Version

Yosys 0.64 (git sha1 6d2c445aebd37261d0e33ed4d90d09fd1a7bf618) [electronics-team/yosys at debian/experimental]

Linux

### Reproduction Steps

The self test fail during build of i386 and hurd-i386 on Debian Testing.  Builds logs are available from https://buildd.debian.org/status/fetch.php?pkg=yosys&arch=i386&ver=0.64-4&stamp=1776785342&raw=0 (i386) and https://buildd.debian.org/status/fetch.php?pkg=yosys&arch=hurd-i386&ver=0.64-4&stamp=1776814061&raw=0 (hurd-i386).  The debian package contain a set of patches, where I suspect the most vital one is for abc to fix build on 32 bit architectures (tried upstreaming via email as pull request access is blocked, also available from https://salsa.debian.org/electronics-team/yosys/-/blob/debian/experimental/debian/patches/abc/1050-no-128-bit-archs.patch?ref_type=heads ).  Note, the self test work on other 32 bit architectures like armhf work, and also 64 bit architectures like arm64 and amd64.

### Expected Behavior

I expected the self testing to succeed.

### Actual Behavior

The self testing fail with this message:

```
Passed sim-sim_aldffe.ys
YOSYS_MAX_THREADS=4 "/home/pere/yosys-salsa/yosys" -ql sim_cycles.ys.err -w 'Yosys has only limited support for tri-state logic at the moment.' sim_cycles.ys && mv sim_cycles.ys.err sim_cycles.ys.log
ERROR: Expected log pattern 'Co-simulating cycle 40' found 1 time(s), instead of 2 time(s) !
make[3]: *** [run-test.mk:36: sim-sim_cycles.ys] Error 1
make[3]: Leaving directory '/home/pere/yosys-salsa/tests/sim'
make[2]: *** [Makefile:1013: makefile-tests/tests/sim] Error 2
```

Note, I edited run-test.mk to drop the @ and show the command being executed.

stashcroft:
This should fix #5822

Although other issues stop the i386 build working at the moment I think this is still worth doing.
The assumption that you can take an integer, multiply it by a double, divide by the same double and get the original integer back is flawed. Just because the tests only caught it only i386 doesn't mean it can't happen elsewhere.

### DIFF
--- a/kernel/fstdata.cc
+++ b/kernel/fstdata.cc
@@ -45,8 +45,7 @@ FstData::FstData(std::string filename) : ctx(nullptr)
 	ctx = (fstReaderContext *)fstReaderOpen(filename.c_str());
 	if (!ctx)
 		log_error("Error opening '%s' as FST file\n", filename);
-	int scale = (int)fstReaderGetTimescale(ctx);	
-	timescale = pow(10.0, scale);
+	scale = (int)fstReaderGetTimescale(ctx);
 	timescale_str = "";
 	int unit = 0;
 	int zeros = 0;
--- a/kernel/fstdata.h
+++ b/kernel/fstdata.h
@@ -55,7 +55,7 @@ class FstData
 	std::string valueOf(fstHandle signal);
 	fstHandle getHandle(std::string name);
 	dict<int,fstHandle> getMemoryHandles(std::string name);
-	double getTimescale() { return timescale; }
+	int getScale() { return scale; }
 	const char *getTimescaleString() { return timescale_str.c_str(); }
 private:
 	void extractVarNames();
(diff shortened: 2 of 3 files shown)


widlarizer (changes requested):
Looks logically sound. Please ensure it is obvious at the point of declaration of the scale variable that it is an exponent of ten, ideally along with an explicit example that "us corresponds to a scale of 6"

petterreinholdtsen:
This look like a much better solution than my #5834.

petterreinholdtsen:
I can confirm that this solved the i386 build problem on Debian as reported in #5822.

widlarizer (approved):

PR #5832, opened 2026-04-23
rtlil: rewrite SigSpec::extract for perf and packing

widlarizer:
There is a tradeoff in `SigSpec::extract`, which returns a slice of a `SigSpec`, between performance and maximal packing.

First, some context: Before thread-safety motivated changes (#5453), SigSpec could hold either a SigBit vector (unpacked), or a SigChunk vector (packed). A SigChunk references a Wire slice or holds a vector of Const bits. Since those changes, SigSpec can hold a SigBit vector or a single SigChunk. All usage of chunks was converted to read-only chunks() iterators including `SigSpec::extract` at e1169d03. However since `SigSpec::extract` was written for a speed/packing tradeoff in a multiple chunks world, this worked out to be pretty drastically slow on large repeated extractions when these temporary chunks would be reconstructed in a read-only way.

In this PR I've changed the implementation to be always unpacking. This helped performance with an example built to test #5809 almost as much as #5809 helped. I've additionally added the `opt_reduce` change from that PR to this one, and I rewrote `SigSpec::extract` again to be still bit-iteration-based but to one-shot pack if possible, and to fall back to bits once it fails to pack into a single Wire slice. This beats #5809 on this one benchmark:

[pmux-big.txt](https://github.com/user-attachments/files/27018282/pmux-big.txt)

Keeping packing where possible is helpful as it reduces memory overhead and potentially speeds up traversal and analysis in some places. The particular case of large `$pmux` points to a restriction in the RTLIL format in having something like port arrays, so that all the "ports" are actually concatenated into a single huge port, even while they may point each to a wire with no slicing or concatenation. Prior to the thread-safety changes the huge port would hold one chunk per such "port", and since those, it holds a vector of all their bits. But now when extracting one of these "ports" from the huge port, it cheaply coalesces into a SigChunk in a one-time pass over only the relevant bits.

`hyperfine "uut/main-lto/bin/yosys -d g/pmux-big.ys" "uut/pmux-opt/bin/yosys -d g/pmux-big.ys" "uut/fa
ster-extract-lto/bin/yosys -d g/pmux-big.ys" "uut/extract-packing-lto/bin/yosys -d g/pmux-big.ys"`
```
Summary
  uut/pmux-opt/bin/yosys -d g/pmux-big.ys ran
    1.00 ± 0.01 times faster than uut/extract-packing-lto/bin/yosys -d g/pmux-big.ys
    1.10 ± 0.02 times faster than uut/faster-extract-lto/bin/yosys -d g/pmux-big.ys
    1.80 ± 0.02 times faster than uut/main-lto/bin/yosys -d g/pmux-big.ys
```
`uut/extract-packing-lto/bin/yosys` = this PR
`uut/faster-extract-lto/bin/yosys` = only the first commit of this PR
`uut/pmux-opt` = #5809

No regression found on ibex and jpeg synthesis with read_verilog in the loop. Packing behavior is checked in the new unit test.

cc @QuantamHD @rocallahan


rocallahan:
Very nice!

QuantamHD:
@widlarizer Seems to work rather well. On my benchmark it ran in about 400s, which is technically slower than 350s I got a different time, but that's not really much of a concern and could be due to machine differences or noise.

QuantamHD:
The dense-ification of the opt_muxtree pass would also be good to get in

widlarizer:
I've moved the opt-side changes out to #5809 instead

hexratcc (approved):
lgtm

PR #5842, opened 2026-04-29
opt_dff: Eliminate equivalent bits

hexratcc:
Adds a new optimization to opt_dff -sat that finds groups of FF bits provably holding always-equal values across cycles and collapses each group to a single bit. As an example take [tests/opt/opt_dff_eqbits_large.sv](https://github.com/YosysHQ/yosys/blob/97cd01b9a17a6bcd9feb813b942fef98940c8015/tests/opt/opt_dff_eqbits_large.sv), base Yosys produces 420 FFs - this PR drops it down to 6 (same result as Genus). Main benefit would be reducing FF count for generated code (ie. from SystemVerilog generate statements). The idea is to check equivalence on classes of FF bits (separated into groups with the same signature (see code)) and merge them, thus reducing code size and FF count.


hexratcc:
[sc-517]

hexratcc:
I think there are a couple ways to reduce the perf hit from this optimization:
- Use the SAT counterexample to better split classes, potentially reducing the number of SAT calls needed
- The main loop currently re-evaluates every class when any class splits. I can probably speed this up by using a work list ordered by topology, so that we don't need to evaluate irrelevant classes.

hexratcc:
Implemented counterexample splitting and a basic worklist, seems to produce much better results. `yosys-perf` `ibex` takes the biggest hit - **with `run_eqbits` I go from 2.5s to more than 100s** on my system (other perf tests don't see performance hits this large, which is interesting). **With this optimization applied ibex takes about 7s**, which is still too expensive to run every time.

Investigating other ways to improve perf, as I havent't discovered any heuristic which could be used to determine whether to run the pass on a give module or not.

hexratcc:
I was able to add a simulation prepass which quickly disqualifies the majority of unoptimizable classes, drastically reducing the number of classes which reach the SAT, thus improving performance. On my machine, **`ibex` now takes about 0.2s more than base Yosys with `eqbits` enabled**. The hit across other tests is comparable, so I've opted to remove the flag and place `eqbits` behind `-sat`.

mikesinouye:
With the latest commits I am seeing about a 2x slowdown in opt_dff runtime, but it's negligible as it's about 1.2% -> 2.8% of our design's runtime. To me that is an acceptable runtime penalty for designs that are not eligible for the optimization. That is significantly better than before the disqualification pass. This looks good to me to enable globally internally, thank you.

image

kernel/bitsim.h
```diff
@@ -0,0 +1,79 @@
...
+	Module *module;
+	SigMap &sigmap;
+	ModWalker &modwalker;
+	dict<SigBit, uint64_t> sim_vals;
+	uint64_t rng_state;
+
+	BitSim(Module *m, SigMap &sm, ModWalker &mw) 
+		: module(m), sigmap(sm), modwalker(mw), rng_state(1337) {}
+
+	uint64_t xorshift64() {
```
widlarizer: There's `mkhash_xorshift`

mikesinouye:
Hi @widlarizer, can this be reviewed on the YosysHQ side?

passes/opt/opt_dff.cc
```diff
@@ -919,6 +919,286 @@ struct OptDffWorker
...
+			return h;
+		}
+	};
+
+	bool is_def(State s) {
+		return s == State::S0 || s == State::S1;
+	}
+
+	int sat_mux(QuickConeSat &qcsat, int s, int a, int b) {
+		return qcsat.ez->OR(qcsat.ez->AND(s, a), qcsat.ez->AND(qcsat.ez->NOT(s), b));
```
widlarizer: `qcsat.ez->ITE` ?

passes/opt/opt_dff.cc
```diff
@@ -919,6 +919,286 @@ struct OptDffWorker
...
+	}
+
+	int sat_const(QuickConeSat &qcsat, State v) {
```
widlarizer: `qcsat.ez->value(v == State::S1)` ?

passes/opt/opt_dff.cc
```diff
@@ -919,6 +919,286 @@ struct OptDffWorker
...
+				continue;
+
+			ff_for_cell.emplace(cell, ff);
+
+			for (int i = 0; i < ff.width; i++) {
+				// X value
+				if (ff.has_srst && !is_def(ff.val_srst[i])) continue;
+				if (ff.has_arst && !is_def(ff.val_arst[i])) continue;
+
+				// Missing anchor
```
widlarizer: Please define terms used in comments

passes/opt/opt_dff.cc
```diff
@@ -919,6 +919,286 @@ struct OptDffWorker
...
+				bool def_init = is_def(ff.val_init[i]);
+				if (!def_init && !ff.has_srst && !ff.has_arst)
```
widlarizer, line 1081: What does it mean when this condition is met? I guess the most basic DFFs (no reset, undef init) aren't a target of this opt, should be explicit if that's the case

kernel/bitsim.h
```diff
@@ -0,0 +1,79 @@
+#ifndef BITSIM_H
+#define BITSIM_H
+
+#include "kernel/modtools.h"
+
+YOSYS_NAMESPACE_BEGIN
+
+struct BitSim {
```
widlarizer: Before I've been in favor of pushing analysis into shared headers like `kernel/bitpattern.h` for code reuse but this analysis seems kind of quirky rather than general. Maybe more importantly there isn't an explanation for what it could be reused for or the principle of operation. The name kind of suggests actual values being simulated, while as far as I can tell this is a heuristic for filtering for signals or cells that might be found equal but have to still be checked if they're equal with SAT. It's really clever, was it inspired by something pre-existing? Would be good to add references if that is the case, otherwise it needs a pretty explicit description

kernel/bitsim.h
```diff
@@ -0,0 +1,79 @@
...
+		: module(m), sigmap(sm), modwalker(mw), rng_state(1337) {}
+
+	uint64_t xorshift64() {
+		rng_state ^= rng_state << 13;
+		rng_state ^= rng_state >> 7;
+		rng_state ^= rng_state << 17;
+		return rng_state;
+	}
+
+	uint64_t eval_bit(SigBit b) {
```
widlarizer: For extremely large designs it would be a good failsafe to limit the amount of effort this analysis can do. I think our SAT infrastructure (qcsat and ezsat) is missing this feature and we can't do anything there, but having some kind of limit to the depth and size of sim_vals could help. See opt_muxtree for recursions_left and glob_evals_left. We should also retain user control if we add that, maybe through scratchpad variables would be enough

passes/opt/opt_dff.cc
```diff
@@ -919,6 +920,386 @@ struct OptDffWorker
...
 		return did_something;
 	}
+
+	struct EqBit {
+		Cell *cell;
+		int idx;
+		SigBit q;
+	};
+
+	struct SigKey {
```
widlarizer, line 1016: This partial duplication `kernel/ff.h` structures is permissible here as it achieves a more efficient view of the cells for this particular analysis, but this should be noted in a comment justifying that decision

passes/opt/opt_dff.cc
```diff
@@ -919,6 +920,386 @@ struct OptDffWorker
...
+			}
+
+			// Split at counterexamples
+			int rep = cls[0];
+			for (int i = 1; i < GetSize(cls); i++) {
+				// Trivially equivalent
+				if (n_lit[rep] == n_lit[cls[i]])
+					continue;
+
+				int query = qcsat.ez->NOT(qcsat.ez->IFF(n_lit[rep], n_lit[cls[i]]));
```
widlarizer: So, XOR, right

passes/opt/opt_dff.cc
```diff
@@ -919,6 +920,386 @@ struct OptDffWorker
...
+				std::vector<int> modelExprs;
+				for (int b : cls)
+					modelExprs.push_back(n_lit[b]);
```
widlarizer, line 1292: Are we solving only one SAT query per class? Can you expand on what we're assuming and proving here?

widlarizer (changes requested):

mikesinouye:
Hi @widlarizer, could you review again when you get a chance?

passes/opt/opt_dff.cc
```diff
@@ -919,6 +1004,397 @@ struct OptDffWorker
...
+				continue;
+
+			FfData ff(&initvals, cell);
+			if (!ff.has_clk && !ff.has_gclk)
+				continue;
+
+			ff_for_cell.emplace(cell, ff);
+
+			for (int i = 0; i < ff.width; i++) {
+				// Skip bits whose reset drives an undefined (x) value
```
widlarizer: "is driven by an undefined (x) value"?

passes/opt/opt_dff.cc
```diff
@@ -919,6 +1004,397 @@ struct OptDffWorker
...
+			// Scan the class members against the representative and issue a query per pair,
+			// stopping early at the first counterexample, which is reused to split the entire
+			// class at once
+			int rep = cls[0];
+			for (int i = 1; i < GetSize(cls); i++) {
+				if (n_lit[rep] == n_lit[cls[i]])
+					continue;
+
+				// Can the next state of the rep and this member ever differ?
+				int query = qcsat.ez->XOR(n_lit[rep], n_lit[cls[i]]);
```
widlarizer, line 1287: This suggests to me that we're not handling undef-ness which can subtly break `-keepdc`. Consider looking into that, checking if you can construct a test case where undef and zero bits get confused here as equivalent, and accordingly making `-sat` conflict with `-keepdc` if that suggests it is required to do so to meet user expectations set by the option documentation. It's a bit concerning that we seem to have little pre-existing coverage of `-sat` and zero of `-sat -keepdc`

widlarizer (approved):
The reasoning can now be argued about, making the pass maintainable, with some notes that I wouldn't block the merge on

PR #5843, opened 2026-04-29
simplemap: Moves $pmux mapping from techmap.v to simple map

QuantamHD:
This Fixes the slow downs I observed in techmap.v, which we attempted to fix via the simplify ast.h route originally. This turned out to be rather complex though. https://github.com/YosysHQ/yosys/pull/5830

By moving $pmux to simplemap we can just avoid that code. My test case now runs in 310s which is 40s faster than the baseline change.

### DIFF
--- a/passes/techmap/simplemap.cc
+++ b/passes/techmap/simplemap.cc
@@ -438,6 +438,48 @@ void simplemap_ff(RTLIL::Module *, RTLIL::Cell *cell)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


QuantamHD:
@widlarizer This should allow us to close https://github.com/YosysHQ/yosys/pull/5830

widlarizer (approved):
Makes sense and matches the verilog implementation

PR #5844, opened 2026-04-30
abc_new: integration testing via synth_gatemate

Ravenslofty:
This kind of ended up as a stacked PR, apologies.

In the process of trying to refactor `abc_new` by unifying `read_aiger` and `read_xaiger2`, I discovered that `abc_new` has exactly one test in the test suite. My understanding is that @Goubermouche is going to write some unit tests for it more specifically, but I think it is also useful to test it in integration. One challenge is that Yosys does not really have any ASIC flows (because that would require bundling a PDK), but `synth_gatemate` is mapped as if it was an ASIC and so makes a reasonable test case for this.

The `synth_gatemate` test suite is ported to `abc_new`, and in the process tests are also added for the `-luttree` flow (through classic `abc`) that is used for targeting nextpnr.

By running the `synth_gatemate` flow with `abc_new`, we test both `abc_new` integration, as well as `synth_gatemate -luttree`, which did not have any explicit tests.


techlibs/gatemate/synth_gatemate.cc
```diff
@@ -69,14 +69,18 @@ struct SynthGateMatePass : public ScriptPass
 		log("        do not use CC_MX{8,4} multiplexer cells in output netlist.\n");
 		log("\n");
 		log("    -luttree\n");
-		log("        use new LUT tree mapping approach (EXPERIMENTAL).\n");
+		log("        use LUT tree mapping for output to nextpnr. Do not use this if targeting\n");
+		log("        legacy p_r.\n");
```
widlarizer, line 73: p&r?

widlarizer (approved):
Looks good as far as I can tell

techlibs/gatemate/synth_gatemate.cc
Ravenslofty: The legacy tool is literally named `p_r`.

mmicko (approved):

Ravenslofty:
Um, it appears the abc_new flow has a memory leak?

PR #5845, opened 2026-04-30
accident PR

stanminlee:
1. Validation inside [] in RTL names to ensure that stoi does not throw a value error.
2. Flattened signal scope paths in vcd_reg_widths to allow for VCD and RTL hierarchies to match exactly when parsing VCDs and managing RTL hierarchies (particularly important for structs or fork scopes)
3. edited splitcells to properly handle struct attributes in registers, again using the Q wire name as a reference point instead of the register name itself (for renaming purposes) because the register name can't capture enough information about the struct.


stanminlee:
mistake again, sorry

PR #5846, opened 2026-05-01
Upgrade to WASI SDK 33 and drop exception workaround

flumf:
The latest WASI SDK ships with support for C++ exceptions, so the previous workaround is no longer required.

Includes small fixes to fix compilation on newer WASI SDKs and disables LTO (as it is currently not compatible with exceptions).

### DIFF
--- a/.github/workflows/extra-builds.yml
+++ b/.github/workflows/extra-builds.yml
@@ -79,8 +79,8 @@ jobs:
           persist-credentials: false
       - name: Build
         run: |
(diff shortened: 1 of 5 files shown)


whitequark:
What gives? https://github.com/YosysHQ/yosys/actions/runs/25248844502/job/74037588387

```
/home/runner/llvm/bin/clang: error while loading shared libraries: libtinfo.so.5: cannot open shared object file: No such file or directory
```

PR #5848, opened 2026-05-04
Use latest brew LLVM on macOS

mmicko:
Check if latest LLVM brew package is now fine.


mmicko:
It is running sanitizer builds fine https://github.com/YosysHQ/yosys/actions/runs/25308962296 and also compilers at https://github.com/YosysHQ/yosys/actions/runs/25317050971, regular build is anyway part of PR CI https://github.com/YosysHQ/yosys/actions/runs/25309798954 and that is fine as well.

KrystalDelusion (approved):

PR #5850, opened 2026-05-04
Add check before flatten in synth - Fix #429

### LINKED ISSUE #429: Feature request: synth* to call check after hierarchy
The below `test` module is very obviously broken since the `test2` instance is wired wrongly. Running a `hierarchy`, and then `check` gives good feedback, but just a `hierarchy; flatten;` will make all of the warnings disappear and subtle bugs sneak in.

```verilog
module test(input x,
            output y);

  test2 instance(
    .y(x),
    .x(y)
  );

endmodule

module test2(input x,
             output y);

  assign x = y;

endmodule
```

Basically this is a request for having the inbuilt synth scripts use `check` after `hierarchy` to expose warnings that might otherwise be lost.

hexratcc:
This PR adds `check` immediately before the `flatten` step in the generic synth command to make it easier to notice potentially subtle bugs as described in #429.

### DIFF
--- a/docs/source/code_examples/macro_commands/synth_ice40.ys
+++ b/docs/source/code_examples/macro_commands/synth_ice40.ys
@@ -6,6 +6,7 @@ begin:
     proc
 
 flatten:
+    check
     flatten
     tribuf -logic
     deminout
--- a/techlibs/achronix/synth_achronix.cc
+++ b/techlibs/achronix/synth_achronix.cc
@@ -129,6 +129,7 @@ struct SynthAchronixPass : public ScriptPass {
     if (flatten && check_label("flatten", "(unless -noflatten)"))
       {
         run("proc");
+        run("check");
         run("flatten");
         run("tribuf -logic");
         run("deminout");
--- a/techlibs/analogdevices/synth_analogdevices.cc
+++ b/techlibs/analogdevices/synth_analogdevices.cc
@@ -277,8 +277,10 @@ struct SynthAnalogDevicesPass : public ScriptPass
 
 		if (check_label("prepare")) {
 			run("proc");
-			if (flatten || help_mode)
+			if (flatten || help_mode) {
+				run("check");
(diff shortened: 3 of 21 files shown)


hexratcc:
[sc-536]

widlarizer (approved):
since `proc` should be trusted not to create any garbage `check` would erroneously fail on, this sounds good

nakengelhardt:
I think this kind of change should be applied consistently to all the `synth_*` passes, not just the one, otherwise we end up with divergent design interpretation on different architectures.

But for the `synth` pass specifically, since this is used for ASIC where we genrerally see the largest designs: we've had problems with `check` performance in the past, are there concerns with the impact of calling it twice?

hexratcc:
@nakengelhardt the additional call to `synth` results in a ~0.5% (0.6s) increase in the `fft64_width_64` test case in yosys-perf, which appears to be the heaviest example in the suite, as far as performance hits from `check` go. IMO this shouldn't be an issue, but please feel free to correct me if I'm mistaken.

widlarizer:
I'm ok with that perf hit as long as it's not a sign of poor scaling with design size, which I wouldn't expect with check to be the case

nakengelhardt:
I looked up the past issues/PRs (#5282 #5284 #4495) and you're right that it was not correlated to design size, rather to single cells with wide ports. So that's probably not a concern, not sure what the check runtimes currently are on the designs in question (cva6 and black-parrot) but I assume not much after the fixes.

ETA: yep, found my local checkout of black-parrot and not a problem any longer `1x check (0 sec)`

Ravenslofty (approved):

KrystalDelusion:
We should probably update the synthesis starter to match.  The macro to check got updated (which is why I got tagged as a code-owner), but the actual `getting_started/example_synth.rst` still doesn't mention the call to `check` until the coarse-grain representation.

Issue #5839, opened 2026-04-27
Yosys runs out of memory on PROC_RMDEAD in case statement


avgsurfman:
### Version

Yosys 0.57+157 (git sha1 b9dc57841, g++ 13.3.0-6ubuntu2~24.04 -fPIC -O3)

### On which OS did this happen?

Linux

### Reproduction Steps

1) Try to synthesize the following module with any synth option
2) Stuck on PROC_RMDEAD

```module lza_so_case_32 (input logic [31:0] a, b,
                      input logic cin,
                      output logic cout, 
                      output logic [6:0] y);


///// (Following Schmookler's notation)
//// T = A ^ B
/// G = A & B
// Z = nA & nB == n(A + B)
// Caveat one: A is always bigger than B. If not, the operands must be swapped.

logic [31:0] nT, nZ, E;


//// Two layers for LZA
///  1. The LZA Array - delay of O(n) (Chained simple boolean gates)*
//  2. Tree Leading Zero Counter
 

genvar i, j;
//// Layer one
generate
    // E_0 is skipped and asserted as one
    assign E[0] = 1'b1;
    // partly based on the schmookler paper and the suzuki paper
    assign nZ[0] = a[0] | b[0];
    assign nT[0] = a[0] ~^ b[0];
    for (i = 1; i < 32; i++) begin
        assign nT[i] = a[i] ~^ b[i];
        assign nZ[i] = a[i] | b[i];
        assign E[i] = nT[i] & nZ[i-1];
    end 
endgenerate


//// Layer two
/// LZD instantiated by Yosys.

always_comb
        casez(E)
        32'b00000000_00000000_00000000_00000000: y='d32;
        32'b00000000_00000000_00000000_00000001: y='d31;
        32'b00000000_00000000_00000000_0000001?: y='d30;
        32'b00000000_00000000_00000000_000001??: y='d29;
        32'b00000000_00000000_00000000_00001???: y='d28;
        32'b00000000_00000000_00000000_0001????: y='d27;
        32'b00000000_00000000_00000000_001?????: y='d26;
        32'b00000000_00000000_00000000_01??????: y='d25;
        32'b00000000_00000000_00000000_1???????: y='d24;
        32'b00000000_00000000_00000001_????????: y='d23;
        32'b00000000_00000000_0000001?_????????: y='d22;
        32'b00000000_00000000_000001??_????????: y='d21;
        32'b00000000_00000000_00001???_????????: y='d20;
[...]

PR #5852, opened 2026-05-04
bitpattern: add analysis effort limit, use in proc_rmdead

widlarizer:
Aims to fix the proc_rmdead issue #5839 and comes with unit test, but BitPatternPool is used in proc_mux too where I don't know yet if I can nope out like this. I wouldn't be surprised if the "real" solution was a different representation rather than BitPatternPool

### DIFF
--- a/kernel/bitpattern.h
+++ b/kernel/bitpattern.h
@@ -25,6 +25,12 @@
 
 YOSYS_NAMESPACE_BEGIN
 
+enum TakeResult {
+	FALSE,
+	TRUE,
+	TOO_BIG
+};
+
 /**
  * This file implements BitPatternPool for efficiently storing and querying
  * sets of fixed-width 2-valued logic constants compressed as "bit patterns".
@@ -39,6 +45,7 @@ YOSYS_NAMESPACE_BEGIN
  */
 struct BitPatternPool
 {
+	size_t limit = SIZE_MAX;
 	int width;
 	struct bits_t {
 		std::vector<RTLIL::State> bitdata;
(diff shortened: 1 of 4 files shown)


hexratcc (approved):
I think the 'correct' way of handling the explosion is to use a [bdd ](https://en.wikipedia.org/wiki/Binary_decision_diagram), but that seems like a bit of an overkill for removing dead switch statements lol. 

The implementation as-is looks good to me, and for the purposes of this PR I'd just leave a note in proc_mux.

widlarizer:
I don't actually think it would be overkill. If we were to bail from this analysis, we can't ensure the switch is full, we'd be emitting dlatches that we don't have a way of proving they're never active (QoR problem), and erroring out on the input design, since `always_comb` requires no latch to be emitted. We would basically be only consistently accepting `always_comb` switches that have a default statement if we sometimes give up on the analysis of the compare patterns.

If I reduce the test case to be tractable:
```
module foo(input logic [7:0] in, output logic [6:0] y);
always_comb
    casez(in)
        32'b00000000: y='d0;
        32'b00000001: y='d1;
        32'b0000001?: y='d2;
        32'b000001??: y='d3;
        32'b00001???: y='d4;
        32'b0001????: y='d5;
        32'b001?????: y='d6;
        32'b01??????: y='d7;
        32'b1???????: y='d8;
    endcase
endmodule
```
proc runs on this just fine on main, as well as on this PR, but if I force the analysis to give up because it's hitting the limit by lowering the limit
```diff
-		pool.limit = 100000;
+		pool.limit = 1;
```
It errors out:
```
2.8. Executing PROC_DLATCH pass (convert process syncs to latches).
ERROR: Latch inferred for signal `\foo.\y' from always_comb process `\foo.$proc$g/proc_rmdead-oom-small.sv:2$1'.
```

PR #5857, opened 2026-05-06
help with hierarchy: refactor

hexratcc:
Builds on top of #5821. This PR adds port extraction, implements a wire direction API which should now be separate from the SV frontend, and resolves `$connect` cells.

### DIFF
--- a/passes/hierarchy/hierarchy.cc
+++ b/passes/hierarchy/hierarchy.cc
@@ -26,6 +26,7 @@
 #include "passes/hierarchy/util/positionals.h"
 #include "passes/hierarchy/util/verilog.h"
 #include "passes/hierarchy/util/generate.h"
+#include "passes/hierarchy/util/ports.h"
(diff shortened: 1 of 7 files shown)


widlarizer:
Since GitHub reviews are broken I have to put this in a comment
- resolve_connect_directionality is an added pass over all modules which will incur a performance penalty. This is the simpler way of separating that out from unrelated functions, but I would expect it's not the best way - ideally we traverse the hierarchy only once and maybe collect some analysis data that we act upon separately? That way we could get all of this done with as few passes as possible while keeping separation between different tasks the pass is doing
- dynports are another quirk of read_verilog and could be split out as cleanly as possible out of ports.cc. The dynports derive action is also duplicated with another snippet like that that's left in verilog.cc
- please switch to destructuring when iterating over pairs: `for (const auto& [port, sig] : cell->connections()) {`

PR #5859, opened 2026-05-07
Generate coverage for tests

mmicko:
Generate coverage reports visible at https://yosyshq.net/reports/coverage/main/

### DIFF
--- a/.github/workflows/test-verific.yml
+++ b/.github/workflows/test-verific.yml
@@ -44,6 +44,14 @@ jobs:
       - name: Runtime environment
         run: |
           echo "procs=$(nproc)" >> $GITHUB_ENV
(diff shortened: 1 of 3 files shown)


.github/workflows/test-verific.yml
```diff
@@ -47,7 +47,7 @@ jobs:
 
       - name: Build Yosys
         run: |
-          make config-clang
+          make config-gcov
```
widlarizer, line 58: This switches from clang to gcc for this task, so we'd be testing gcc built yosys rather than clang built, right? I think gcov should [work with clang too](https://clang.llvm.org/docs/SourceBasedCodeCoverage.html), it should be possible to modify config-gcov to make it work

widlarizer (changes requested):

widlarizer (approved):

PR #5862, opened 2026-05-09
Cleaning up log_id()

codexplorer-fish:
### Context

This change is an implementation of the issue highlighted in this [discussion](https://yosyshq.discourse.group/t/cleaning-up-log-id-and-rtlil-unescape-id/82). 

### Motivation

`log_id()` has two problems. First, its name implies that it logs something, which it doesn't do. Second, it returns a `const char *`, when `std::string` is generally preferred. 

It is cleaner to directly call `IdString`'s dedicated `.unescape()` function, which has been implemented since the discussion. However, relying solely on `.unescape()` would lead to a large amount of clutter, going from `log("%s", log_id(cell))` to `log("%s", cell.name.unescape())`. Because the natural interpretation of passing `cell` to a `"%s"` specifier is to use `cell`'s unescaped name, it makes sense to bake the functionality directly into the `log()` family. It simplifies the call into `log("%s", cell)`.

### Implementation

Let `T` be any class with a member `name` of type `IdString`.

First, using SFINAE, two extra branches are added to `CONVSPEC_CHAR_PTR` to capture `T` and `T*`. Then, `.unescape()` is called on the confirmed `T`'s `IdString name`. This replicates `log_id()`'s functionality on `T`.

Second, uses of `log_id()` are refactored. `T`s and `T*`s are directly passed into the `log()` family. Meanwhile, `IdString`s are unescaped via `.unescape()`.
- `log("%s", log_id(cell))` -> `log("%s", cell)`
- `log("%s", log_id(cell_ptr))` -> `log("%s", cell_ptr)`
- `log("%s", log_id(name))` -> `log("%s", name.unescape())`

And where the `log()` family isn't used:
- `log_id(cell)` -> `cell.name.unescape()`
- `log_id(cell_ptr)` -> `cell->name.unescape()`

Finally, here are the instances where `log_id()` was not removed:

- In the .pmg files. They are outside of the scope of this PR.

- In frontends/ast/simplify.cc, where `std::string` is being passed into `log_id()`.

- In the use of `log_id()` in the `id2cstr()` implementation. Refactoring `id2cstr` should probably be its own task.

- In the uses of `log_id()` with a second argument, where the second argument acts as a fallback for if the first argument is `nullptr`. `.unescape()` does not have such a function, so `log_id()` may have a genuine purpose here.

### Testing

All existing `make test` tests pass. Because of the refactor, expected usage of the new branches is tested. However, no new tests have been created to verify the edge cases.


codexplorer-fish:
Thanks to @rocallahan for helping me with my first open-source PR!

mmicko:
@codexplorer-fish Thank you very much for contribution and taking care about explaining details of implementation and referring to original discussion. Still looking into details, but manged to found two instances that were nor wright so updated code for those (hope you do not mind). 
Also can confirm that this will not affect any 3rd party plugin implementation currently using log_id since that one is still in place.
Continue to look into more details and will do some experiments, but so far so good.

mmicko:
@codexplorer-fish Managed to find couple more differences, mostly by comparing logs before and after change, could be still something is missed, but anyway not sure if we are consistent always consistent of displaying unescaped names.

codexplorer-fish:
@mmicko Thanks for the updated code. Apologies for the inaccurate refactor in the `fprintf()` context. The refactor does break up unescaping names into multiple different "displays", but as I understand it, it is constant based on data type and whether it is used in the `log()` family context:

<pre>
Inside of the `log()` family context:
IdString:    log("%s", name.unescape())
Class:       log("%s", cell)
Class*:      log("%s", cell)
std::string: log("%s", RTLIL::unescape_id(name))

Outside of the `log()` family context:
IdString:    name.unescape()
Class:       cell.name.unescape()
Class*:      cell->name.unescape()
std::string: RTLIL::unescape_id(name)
</pre>

At least, that is the goal. It's true as you say that something could have been missed. Do you know of a good way to resolve that uncertainty?

mmicko (approved):

PR #5863, opened 2026-05-09
memlib: fix documentation for `PORT_<name>_CLK_POL`

mole99:
Fix a minor but important typo in the memlib documentation.
The parameter for anyedge clocks is `PORT_<name>_CLK_POL`, not `PORT_<name>_CLKPOL`. Please note the additional underscore.

This can also be seen in the source code here: https://github.com/YosysHQ/yosys/blob/1f023432681c159885f0b834a2e0717e67c4c115/passes/memory/memory_libmap.cc#L1755


widlarizer (approved):
Thanks

PR #5875, opened 2026-05-12
threading: redirect locks to no-op on single-threaded builds

widlarizer:
Previously, `ENABLE_THREADS=0` builds got broken due to mutex usage without including `<mutex>`. We could easily:
- use more ifdefs to exclude acquiring the lock
- unconditionally include `<mutex>` in single-threaded builds, locking without contention

Or, instead we can redirect to no-op locks and make the code more compact which is what this PR does

@rocallahan do you have the time for a review?
cc @whitequark

### DIFF
--- a/kernel/threading.h
+++ b/kernel/threading.h
@@ -15,6 +15,33 @@
 
 YOSYS_NAMESPACE_BEGIN
 
+// Redirect to no-op to avoid dependence on <mutex>
+// and <condition_variable> in single-threaded builds
+#ifdef YOSYS_ENABLE_THREADS
+using Mutex = std::mutex;
+using CondVar = std::condition_variable;
+template <class M> using UniqueLock = std::unique_lock<M>;
+template <class M> using LockGuard = std::lock_guard<M>;
+#else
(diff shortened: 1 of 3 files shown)


whitequark:
Thank you!

kernel/threading.h
```diff
@@ -15,6 +15,33 @@
 
 YOSYS_NAMESPACE_BEGIN
 
+// Redirect to no-op to avoid dependence on <mutex>
+// and <condition_variable> in single-threaded builds
+#ifdef YOSYS_ENABLE_THREADS
+using Mutex = std::mutex;
+using CondVar = std::condition_variable;
+template <class M> using UniqueLock = std::unique_lock<M>;
+template <class M> using LockGuard = std::lock_guard<M>;
```
rocallahan: I think these don't need to be templates, we won't be using these with anything but Mutex in the forseeable future. But I guess it's fine to leave them as templates for consistency with `std`. I could go either way.

rocallahan (approved):

PR #5876, opened 2026-05-12
Makefile: error on unused variables, including hashlib containers

widlarizer:
This change may be somewhat controversial

Code with unused variables sucks to read and may point to logical errors. Dicts and pools shouldn't be exluced just because of how C++ models side effects. This PR turns unused variables into an error and allows the reporting of unused hashlib containers on clang/gcc builds. Plugin authors may be affected, as they should, when `yosys-config --cxxflags` includes `-Werror=unused`

Explicit casts to void on results of side-effecty operations may be needed from time to time, but those were causing warnings before already

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,7 +104,7 @@ VPATH := $(YOSYS_SRC)
 UNITESTPATH := $(YOSYS_SRC)/tests/unit
 
 export CXXSTD ?= c++17
-CXXFLAGS := $(CXXFLAGS) -Wall -Wextra -ggdb -I. -I"$(YOSYS_SRC)" -MD -MP -D_YOSYS_ -fPIC -I$(PREFIX)/include
+CXXFLAGS := $(CXXFLAGS) -Wall -Wextra -Werror=unused -ggdb -I. -I"$(YOSYS_SRC)" -MD -MP -D_YOSYS_ -fPIC -I$(PREFIX)/include
 LIBS := $(LIBS) -lstdc++ -lm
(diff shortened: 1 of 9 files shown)


Ravenslofty (approved):

whitequark (approved):
Seems fine. The code was improved in pretty much every case.

Biswa96:
This commit causes build failure with Android NDK r29, clang 21.

```
libs/fst/fstapi.cc:4321:32: error: ignoring return value of function declared with 'warn_unused_result' attribute [-Werror,-Wunused-result]
 4321 |                 /* vardir = */ fgetc(
      |                                ^~~~~
 4322 |                     xc->fh); /* unused in VCD reader, but need to advance read pointer */
      |                     ~~~~~~
```

widlarizer:
@Biswa96 as a workaround, you can demote it back to a warning by adding `-Wno-error=unused-result` into your `CXXFLAGS` in `Makefile.conf` or make invocation or [whatever CMake does](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_FLAGS.html#variable:CMAKE_%3CLANG%3E_FLAGS) after CMake gets merged probably this week. Preferrably, file it as an issue upstream at https://github.com/gtkwave/libfst and ping us to pull it in. Seems that Android's SDK uses [bionic libc](https://android.googlesource.com/platform/bionic/) that declares that the return value of `fgetc` shouldn't be ignored, unlike the rest of the libc's we're exposed to.

PR #5880, opened 2026-05-13
verilog backend: preserve `signed` on wire and port declarations

alaindargelas:
`dump_wire` had no code path that emits the `signed` keyword for wires/ports whose RTLIL `is_signed` flag is set.  Reading

  module top(input signed [9:0] in, output signed [31:0] o);
    assign o = in;
  endmodule

and writing it back via `write_verilog` produced

  input  [9:0]  in;
  output [31:0] o;

losing the declared signedness even though `wire->is_signed` was tracked correctly in RTLIL throughout the round trip.  The IEEE 1364-2001 grammar (Annex A.2.1.2 / A.2.1.3) allows `signed` after the direction / net-type keyword, which is the dialect `write_verilog` targets by default — so the fix is to emit ` signed` between the direction/net-type and the range when `wire->is_signed`.

Closes the half of chipsalliance/synlig#2425 that lives in Yosys: the SystemVerilog frontend correctly produces a signed wire for `output int`, but the Verilog backend dropped it on write.

Adds `tests/various/write_verilog_signed_port.ys`, which round-trips a module with `signed` inputs, outputs, and an internal wire and greps for the keyword on each declaration — fails without the fix, passes with it.

### DIFF
--- a/backends/verilog/verilog_backend.cc
+++ b/backends/verilog/verilog_backend.cc
@@ -456,21 +456,26 @@ void dump_wire(std::ostream &f, std::string indent, RTLIL::Wire *wire)
 		if (wire->attributes.count(ID::single_bit_vector))
 			range = stringf(" [%d:%d]", wire->start_offset, wire->start_offset);
 	}
+	// Emit `signed` for wires/ports whose RTLIL is_signed flag is set.
+	// Without this, an `output signed [N:0] o` was silently demoted to plain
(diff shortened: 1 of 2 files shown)


widlarizer:
@Goubermouche does #5771 cover this too?

hexratcc:
@widlarizer Seems like it, the provided test cases passes under #5771 too.

alaindargelas:
Fixed on main.

PR #5881, opened 2026-05-13
Remove unused variable for WIN32

mmicko:
As flag is recently added to convert unused variable to error, this is triggered on MINGW build on Windows.

### DIFF
--- a/kernel/driver.cc
+++ b/kernel/driver.cc
@@ -664,6 +664,7 @@ int main(int argc, char **argv)
 
 #ifdef _WIN32
 		log("End of script. Logfile hash: %s\n", hash);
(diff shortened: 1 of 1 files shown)


kernel/driver.cc
widlarizer, line 666: Instead, please cast the variable to void here, since it gives us fewer ifdefs, and avoiding to construct the variable isn't worth having an extra ifdef

widlarizer (changes requested):

kernel/driver.cc
mmicko: Updated.

widlarizer (approved):

PR #5882, opened 2026-05-13
Bump required standard to C++20

mmicko:
This one bumps required minimal standard to be C++20 and also test latest compilers with C++26 features

### DIFF
--- a/.github/workflows/test-compile.yml
+++ b/.github/workflows/test-compile.yml
@@ -85,7 +85,7 @@ jobs:
       - name: Fix clang-10 toolchain
         if: matrix.os == 'ubuntu-latest' && matrix.compiler == 'clang-10'
         run: |
-          echo "CXXFLAGS=--gcc-toolchain=/usr/lib/gcc/x86_64-linux-gnu/10 -stdlib=libstdc++" >> $GITHUB_ENV
+          echo "CXXFLAGS=--gcc-toolchain=/usr" >> $GITHUB_ENV
 
       # minimum standard
       - name: Build C++20
--- a/.github/workflows/test-compile.yml
+++ b/.github/workflows/test-compile.yml
@@ -44,8 +44,8 @@ jobs:
           - ubuntu-latest
         compiler:
           # oldest supported
-          - 'clang-10'
-          - 'gcc-10'
+          - 'clang-14'
+          - 'gcc-11'
           # newest, make sure to update maximum standard step to match
           - 'clang-19'
           - 'gcc-14'
@@ -74,7 +74,7 @@ jobs:
         uses: aminya/setup-cpp@v1
         with:
           compiler: ${{ matrix.compiler }}
-          gcc: ${{ (matrix.os == 'ubuntu-latest' && matrix.compiler == 'clang-10') && '10' || '' }}
+          gcc: ${{ (matrix.os == 'ubuntu-latest' && matrix.compiler == 'clang-14') && '12' || '' }}
 
       - name: Tool versions
         shell: bash
@@ -82,10 +82,10 @@ jobs:
           $CC --version
           $CXX --version
 
-      - name: Fix clang-10 toolchain
-        if: matrix.os == 'ubuntu-latest' && matrix.compiler == 'clang-10'
+      - name: Fix clang-14 toolchain
+        if: matrix.os == 'ubuntu-latest' && matrix.compiler == 'clang-14'
         run: |
-          echo "CXXFLAGS=--gcc-toolchain=/usr" >> $GITHUB_ENV
+          echo 'CXXFLAGS=--gcc-toolchain=/usr -isystem /usr/include/c++/12 -isystem /usr/include/x86_64-linux-gnu/c++/12' >> $GITHUB_ENV
 
       # minimum standard
       - name: Build C++20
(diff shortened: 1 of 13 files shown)


widlarizer (approved):

widlarizer:
Previous version bump: #4460
@whitequark does this affect CXXRTL?

widlarizer:
Also we checked that the oldest supported Ubuntu LTS, which is 22.04, should be ok with this. This Ubuntu version is also tested on CI at least partially it seems

widlarizer (changes requested):
We should also fix mentions of C++17 in the README and the docs

mmicko:
There are two major issues, one is MSVC (Visual Studio) not working due to zero sized arrays, but working on fix for that and also having issues making clang-10 to work on CI, since mixture of headers from system gcc-14 and old clang is not working.

mmicko:
Visual studio build passing now https://github.com/YosysHQ/yosys/actions/runs/25792378993

mmicko:
@widlarizer https://github.com/YosysHQ/yosys/actions/runs/25799630356 had to upgrade minimal Clang to 14 and GCC to 11 since before that check for `__cplusplus >= 202002L` would not work.
Need to do some configuring to make old clang work when there is already installed later GCC on system, but that works finally.

Visual Studio is also building fine as well as WASI at https://github.com/YosysHQ/yosys/actions/runs/25799757527

Documentation reference should be fine now.

whitequark:
> @whitequark does this affect CXXRTL?

Affect how? CXXRTL's generated code and runtime are totally separate from the rest of the codebase. It's still on C++11 I think, maybe C++14.

mmicko:
Forgot to link that all is fine for compilers after latest update ( https://github.com/YosysHQ/yosys/actions/runs/25851127189 )

.github/workflows/test-compile.yml
```diff
@@ -74,27 +74,33 @@ jobs:
...
         shell: bash
         run: |
           $CC --version
           $CXX --version
 
+      - name: Fix clang-14 toolchain
+        if: matrix.os == 'ubuntu-latest' && matrix.compiler == 'clang-14'
+        run: |
+          echo 'CXXFLAGS=--gcc-toolchain=/usr -isystem /usr/include/c++/12 -isystem /usr/include/x86_64-linux-gnu/c++/12' >> $GITHUB_ENV
+
```
KrystalDelusion, line 89: The gcc install here was added for clang-10 right?  Do we still need it for clang-14?

.github/workflows/test-compile.yml
```diff
@@ -74,27 +74,33 @@ jobs:
...
         run: |
           make config-$CC_SHORT
-          make -j$procs CXXSTD=c++17 compile-only
+          make -j$procs CXXSTD=c++20 compile-only
 
       # maximum standard, only on newest compilers
-      - name: Build C++20
-        if: ${{ matrix.compiler == 'clang-19' || matrix.compiler == 'gcc-14' }}
+      - name: Build C++26
+        if: ${{ matrix.compiler == 'clang-19' || matrix.compiler == 'clang-22' || matrix.compiler == 'gcc-15' }}
```
KrystalDelusion: I think the 'clang-19' here was just for the the newest supported compilers; so it should be okay to run the macos builds with just minimum standard?  Or maybe we bump the macos arm build to use clang-22 if we do want to test max standard on macos

kernel/io.h
KrystalDelusion, line 306: There's a warning on this line about `arg` potentially being uninitialized under gcc-11 which doesn't appear under gcc-10, gcc-14, or gcc-15.  The warning in question is also specifically for this `stringf` in `jny.cc` :
```cpp
for (auto con : port_cell->connections()) {
    // ...
    if (con.second.size() == 1)
        f << _indent << "    \"range\": [0, 0]\n";
    else
        f << stringf("    %s\"range\": [%d, %d]\n", _indent, con.second.size(), 0);
    // ...
}
```
for the case where `con.second.size()` returns `SigChunk::width`... I don't see how it could be uninitialized though (and there aren't any warnings on `SigChunk::size()`), so I think this is a false positive and can be ignored.

passes/cmds/timeest.cc
KrystalDelusion, line 156: Not sure if it would still resolve the problem with gcc-11, but it would be cleaner to put this whole loop inside of an `if (clk.has_value())` instead of skipping the individual ports, since this loop does nothing if there is no `clk` (I mean I guess it does raise an error on an async read port, but if there's no clock to sync to I'm not sure that matters, and memories aren't accessed anywhere else in the function)

.github/workflows/test-compile.yml
mmicko: Actually this is added for clang-14 to work on ubuntu-latest that have gcc-14 by default, with this change then we install older gcc-12 that this clang is compatible with and use those libstc++ libraries as well, making it able to compile as expected. So this is needed :)

.github/workflows/test-compile.yml
mmicko: I have left it for now, also was not sure which one is latest available with xcode (seems to be clang-21). But agree we should probably update this one

kernel/io.h
mmicko: To be honest not sure if using gcc-11 as minimum is good,  as seen in next comment.

passes/cmds/timeest.cc
mmicko: Issue here with gcc-11 is type resolving, but agree that it could be better if it can be done is some other way. skipping that unsupported aync port could lead to some other problems, but anyway problem is more this type resolving that is problematic in C++20 mode of gcc-11 specifically.

widlarizer (approved):

Issue #5877, opened 2026-05-12
the documentation for the `--dont_use` argument for the `abc` pass is just a copy/paste of another


jeras:
### Link to page

https://github.com/YosysHQ/yosys/blob/main/passes/techmap/abc.cc#L1937-L1939

### Issue

The description of the `--dont_use` argument for the `abc` pass is just a copy/paste of the description for the previous argument.

### Expected

I would like the description to state whether the argument can be used multiple times and whether wildcard `*` can be used. Also it would help if the ability to use multiple/wildcard could be clarified for other [techmap commands](https://yosyshq.readthedocs.io/projects/yosys/en/latest/cmd/index_passes_techmap.html).

NOTE: LibreLane is currently filtering liberty files to exclude a list of cells from being used during synthesis, while for `OpenRoad` it is passing the list as an argument (or maybe file, I am not sure right now). A clarification in the Yosys documentation would allow the use of unmodified liberty files during synthesis.

QuantamHD:
I'm the original author of this functionality, but I'm on vacation right now, and don't have access to a laptop right now.

You can pass multiple dont_use flags to the command, and the strings are acceptable in glob format. Would you mind making a PR to update the documentation while I'm out?

jeras:
Thanks for the clarification, I created a trivial PR.

jeras:
I created a pull request to fix the documentation and it was merged.

PR #5884, opened 2026-05-13
documentation: updated description of 'abc' argument '-dont_use'

jeras:
Simple documentation fix, as discussed with the author of the relevant feature here #5877


QuantamHD:
@widlarizer This seems like an easy one to merge. I verified that the documentation is correct.

PR #5888, opened 2026-05-13
opt_muxtree: reuse knowledge_t and pass by reference

mikesinouye:
Follow-up of https://github.com/YosysHQ/yosys/pull/5809. The optimizations there regressed certain other design types, this PR covers these other cases.

Reuse knowledge_t and pass by reference, to avoid the creation of large vectors repeatedly when the number of bits in the mux is high. For designs with high mux count with high bitwidth, these lines become hot: https://github.com/YosysHQ/yosys/blob/2046a23a2fe201118b27627dee50a742b86ddfba/passes/opt/opt_muxtree.cc#L354-L355

Metrics:
Current baseline:
5zHQ6yWkq93R6EK

With change:

4EuAx54KJsYwmhv

For our design (~3M instances), this reduces the total runtime from 11h17m to 4h37m. For smaller designs, I don't anticipate this to make a difference.

### DIFF
--- a/passes/opt/opt_muxtree.cc
+++ b/passes/opt/opt_muxtree.cc
@@ -200,6 +200,29 @@ struct OptMuxtreeWorker
 				root_muxes.at(driving_mux) = true;
 	}
 
+	struct knowledge_t
+	{
+		// Known inactive signals
+		// The payload is a reference counter used to manage the list
+		// When it is non-zero, the signal in known to be inactive
+		// When it reaches zero, the map element is removed
+		std::vector<int> known_inactive;
+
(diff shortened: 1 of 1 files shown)


QuantamHD:
@widlarizer Could you take a look at this when you get a chance?

widlarizer (approved):
Since each activation has a matching inactivation, this caching is safe. A special case is the second return in eval_mux_port based on glob_evals_left, which exits with stuff activated out of sync with traversal, but running out of glob_evals_left leads to total termination of the pass, so the cache isn't used anymore. Looks good

PR #5889, opened 2026-05-15
[WIP] RapidFlex eFPGA Plugin

tangxifan:


Ravenslofty:
Right now this is entirely unreviewable; please rebase your branch on Yosys main.

Ravenslofty (changes requested):
Why are you including a 12,000 line XML parser?

techlibs/rapidflex/Makefile.inc
```diff
@@ -0,0 +1,55 @@
...
+
+$(eval $(call add_share_file,share/rapidflex/common,techlibs/rapidflex/common/cells_sim.v))
+
+# --------------AlkaidC cell lib ------------------------
+$(eval $(call add_share_file,share/rapidflex/alkaidC,techlibs/rapidflex/alkaidC/arith_map.v))
+$(eval $(call add_share_file,share/rapidflex/alkaidC,techlibs/rapidflex/alkaidC/ccb_inst_code.v))
+$(eval $(call add_share_file,share/rapidflex/alkaidC,techlibs/rapidflex/alkaidC/cell_sim_arith.v))
+$(eval $(call add_share_file,share/rapidflex/alkaidC,techlibs/rapidflex/alkaidC/cell_sim_ccb.v))
+$(eval $(call add_share_file,share/rapidflex/alkaidC,techlibs/rapidflex/alkaidC/cell_sim_ff.v))
+$(eval $(call add_share_file,share/rapidflex/alkaidC,techlibs/rapidflex/alkaidC/cell_sim_pcnt.v))
```
mmicko, line 29: Remove this line since there is actually other rule to create this file

techlibs/rapidflex/util/pcnt_cell_sim_gen.py
```diff
@@ -0,0 +1,142 @@
...
+
+#####################################################################
+# Error codes
+#####################################################################
+error_codes = {"SUCCESS": 0, "ERROR": 1, "FILE_ERROR": 3}
+
+#####################################################################
+# Initialize logger
+#####################################################################
+logging.basicConfig(format="%(levelname)s: %(message)s", level=logging.INFO)
```
mmicko, line 19: change to logging.ERROR to make it not output info data while building

techlibs/rapidflex/Makefile.inc
mmicko: Also add .gitignore file to ignore this generated file

techlibs/rapidflex/Makefile.inc
```diff
@@ -0,0 +1,55 @@
+# cell lib generation
+techlibs/rapidflex/alkaidC/cell_sim_pcnt.v: techlibs/rapidflex/util/pcnt_cell_sim_gen.py
+	$(P) mkdir -p $(dir $@) && $(PYTHON_EXECUTABLE) $^ --file $@
+
+CXXFLAGS += -Itechlibs/rapidflex/src/
+OBJS += techlibs/rapidflex/src/pugixml.o
```
mmicko, line 6: remove pugixml related files and also from build

techlibs/rapidflex/src/clock_buffer_cmd.cc
```diff
@@ -0,0 +1,553 @@
...
+      RTLIL::Cell *ckbuf_cell =
+          module->addCell(stringf("$ckbuf$%s", ckbuf.c_str()), "\\ckbuf");
+      ckbuf_cell->setPort("\\in", module->wire("\\" + ckbuf));
+      ckbuf_cell->setPort("\\out", module->wire("\\" + ckbuf + "_ckbuf"));
+      ckbuf_cell->set_src_attribute("");
+    }
+  }
+
+  /*This function is for generating cell map file*/
+  void generate_cell_map(const char *fname,
```
mmicko, line 227: since this function is used to generate dubug data, please just remove function

techlibs/rapidflex/src/clock_buffer_cmd.cc
```diff
@@ -0,0 +1,553 @@
...
+            find_internal_clk_r_signal(module, ckbuf_type);
+        /*insert ckbuf and rewire dff */
+        insert_ckbuf(module, ckbuf_info);
+
+        module->fixup_ports();
+        /*rewire luts */
+        rewire_luts(module, ckbuf_info);
+
+        module->fixup_ports();
+        /* print out cells */
```
mmicko, line 540: remove this block so debug function is not called

techlibs/rapidflex/src/clock_buffer_cmd.cc
```diff
@@ -0,0 +1,553 @@
...
+#include <string>
+
+#include "backends/rtlil/rtlil_backend.h"
+#include "kernel/celltypes.h"
+#include "kernel/log.h"
+#include "kernel/register.h"
+#include "kernel/rtlil.h"
+#include "kernel/sigtools.h"
+#include "kernel/yosys.h"
+#include "pugixml.hpp"
```
mmicko, line 13: remove include of pugixml.hpp

mmicko (changes requested):
Please do check comments, with that some issues related to build and requirement for pugixml can be safely removed making it more reasonable to review

tangxifan:
> Why are you including a 12,000 line XML parser?

It is because an XML file is outputted to note critical details of buffer insertion, so that RapidFlex downstream engine can properly map map the inserted buffers. 
If Yosys provides a good XML writer, I will be glad to use it.

techlibs/rapidflex/src/clock_buffer_cmd.cc
tangxifan: We cannot as we rely on this function to drive our downstream flow. Either there is a good XML parser Yosys suggests or we suggest to use an existing XML lib.

tangxifan:
> Right now this is entirely unreviewable; please rebase your branch on Yosys main.

I will close this PR, clean up my branch and open a new one.

techlibs/rapidflex/util/pcnt_cell_sim_gen.py
tangxifan: Will do. Thanks for the suggestion.

techlibs/rapidflex/Makefile.inc
tangxifan: Thanks. Will do.

techlibs/rapidflex/src/clock_buffer_cmd.cc
tangxifan: This function is not to generate debugging information. If the function name is confusing, we can rename it

mmicko: @tangxifan in this case and since here your code is generating quite simple XML and you are only generating XML, I would suggest creating it with pure C/C++ (using fprintf for example). Should be quite simple and maybe just XML escaping for strings would be need as utility function. Also this should not affect usability of your code and should be also quite readable.

tangxifan: Agree. Will do.

Issue #5885, opened 2026-05-13
16 failing tests in tests/aiger when building Fedora RPM


gsomlo:
### Version

0.65
(to be completely honest, it's commit #90c26fe from the `main` branch, but I see no practical difference :)

### On which OS did this happen?

Linux

### Reproduction Steps

`make test ABCEXTERNAL=/bin/abc SEED=314159265359`

NOTE that /bin/abc is provided by `yosyshq-abc-0.65-1.20260512git5d51a5e.fc45.x86_64`, an RPM built based on github.com/yosyshq/abc release 0.65 !

### Expected Behavior

all tests succeed, return code 0

### Actual Behavior

16 tests in `tests/aiger` fail, non-zero return code:

```
<mock-chroot> sh-5.3# cd build/BUILD/yosys-0.65-build/yosys-9yosys-90c26fe6714b441363d2ddb47a6d646b8868ff5d/

<mock-chroot> sh-5.3# rpm -qa | grep abc
yosyshq-abc-libs-0.65-1.20260512git5d51a5e.fc45.x86_64
yosyshq-abc-0.65-1.20260512git5d51a5e.fc45.x86_64

<mock-chroot> sh-5.3# make test ABCEXTERNAL=/bin/abc SEED=314159265359
[Makefile.conf] CONFIG := gcc
make[1]: Entering directory '/builddir/build/BUILD/yosys-0.65-build/yosys-90c26fe6714b441363d2ddb47a6d646b8868ff5d/tests'
Generate FST for sim models
Test tb_adff
Test tb_adffe
Test tb_adlatch
Test tb_aldff
Test tb_aldffe
...
make[2]: Entering directory '/builddir/build/BUILD/yosys-0.65-build/yosys-90c26f
e6714b441363d2ddb47a6d646b8868ff5d/tests/aiger'
FAIL and_.aag
FAIL and_to_bad_out.aag
FAIL buffer.aag
FAIL cnt1.aag
FAIL cnt1e.aag
FAIL empty.aag
FAIL false.aag
FAIL halfadder.aag
FAIL inverter.aag
FAIL notcnt1.aag
FAIL notcnt1e.aag
FAIL or_.aag
FAIL symbols.aag
FAIL toggle-re.aag
FAIL toggle.aag
FAIL true.aag
PASS io.ys
PASS neg.ys
PASS aigmap
[...]

PR #5890, opened 2026-05-16
Fix aiger tests when ABCEXTERNAL is set

strophy:
16 aiger tests fail when ABCEXTERNAL is set as described in https://github.com/YosysHQ/yosys/issues/5885

`generate_mk.py` incorrectly assumes vendored abc. This PR adds a check for ABCEXTERNAL and uses it if set.

### DIFF
--- a/tests/aiger/generate_mk.py
+++ b/tests/aiger/generate_mk.py
@@ -54,6 +54,13 @@ def create_tests():
         "rm -f aigmap.err"
     ]))
 
-extra = [ f"ABC ?= {gen_tests_makefile.yosys_basedir}/yosys-abc", "SHELL := /usr/bin/env bash" ]
+extra = [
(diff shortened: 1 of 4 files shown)


KrystalDelusion:
@mmicko this was broken in #5768; previous behavior was to use `../../yosys-abc` as default with override by passing in the `-A <abc>` option.  The `abcopt-tests` used the `$(ABCOPT)` arg, which was (and still is) either set to nothing or `-A $(ABCEXTERNAL)`.  The `tests/Makefile` still gets `ABCOPT=$(ABCOPT)`, but doesn't do anything with it.  `abcopt-tests` was previously used for `tests/memories`, `tests/aiger` and `tests/alumacc`.  Probably worth checking all of them, not just aiger.

mmicko:
@strophy Thank you for contribution. @KrystalDelusion all clear, will check them all

PR #5891, opened 2026-05-17
smtbmc: drop obsolete bitwuzla SMT2 CLI options

gipsyh:
## Summary

Bitwuzla 0.9.0 no longer accepts the legacy `--smt2` and `-i` command line options. `yosys-smtbmc -s bitwuzla` currently passes those options unconditionally, which causes Bitwuzla to exit before reading the SMT2 stream and leaves `yosys-smtbmc` reporting a broken pipe.

This patch keeps the existing Boolector invocation unchanged and splits Bitwuzla into its own solver branch. Bitwuzla is now launched as `bitwuzla` plus any user-specified `-S` options.

## Testing

```
python3 -m py_compile backends/smt2/smtio.py tests/functional/smtio.py
python3 backends/smt2/smtbmc.py -s bitwuzla --presat --unroll --noprogress -t 20 --append 0 ../fifo/model/design_smt2.smt2
```

The local Bitwuzla 0.9.0 run reports `Status: PASSED`.

### DIFF
--- a/backends/smt2/smtio.py
+++ b/backends/smt2/smtio.py
@@ -226,11 +226,18 @@ class SmtIo:
                 print('timeout option is not supported for mathsat.')
                 sys.exit(1)
 
-        if self.solver in ["boolector", "bitwuzla"]:
+        if self.solver == "boolector":
             if self.noincr:
                 self.popen_vargs = [self.solver, '--smt2'] + self.solver_opts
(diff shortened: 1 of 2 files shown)


widlarizer:
see https://github.com/YosysHQ/oss-cad-suite-build/issues/87

PR #5894, opened 2026-05-18
Bump CI to gcc-16 and fix warnings

mmicko:
opt_merge was actually not performing as described in comment, it should do that now

### DIFF
--- a/passes/opt/opt_merge.cc
+++ b/passes/opt/opt_merge.cc
@@ -324,7 +324,7 @@ struct OptMergeThreadWorker
 			CellEqualOp> known_cells(0, CellHashOp(), CellEqualOp(*this));
 
 		std::vector<DuplicateCell> duplicates;
-		for (const std::vector<std::vector<CellHash>> &buckets : in.bucketed_cell_hashes) {
+		for (std::vector<std::vector<CellHash>> &buckets : in.bucketed_cell_hashes) {
 			// Clear out our buckets as we go. This keeps the work of deallocation
 			// off the main thread.
 			std::vector<CellHash> bucket = std::move(buckets[index]);
--- a/.github/workflows/test-compile.yml
+++ b/.github/workflows/test-compile.yml
@@ -47,7 +47,7 @@ jobs:
           - 'gcc-11'
           # newest, make sure to update maximum standard step to match
           - 'clang-22'
-          - 'gcc-15'
+          - 'gcc-16'
         include:
           # macOS x86
           - os: macos-15-intel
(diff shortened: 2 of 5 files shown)


widlarizer:
By "the comment" do you mean "Clear out our buckets as we go. This keeps the work of deallocation off the main thread." so we weren't actually moving buckets out of `bucketed_cell_hashes`?

mmicko:
@widlarizer In this case if we use `const` then `std::move` is not doing anything , so think we were not actually moving anything from `bucketed_cell_hashes`

mmicko:
@widlarizer this is something for after CMake merge, but please do check if opt_merge change is actually good

passes/opt/opt_merge.cc
```diff
@@ -324,7 +324,7 @@ struct OptMergeThreadWorker
 			CellEqualOp> known_cells(0, CellHashOp(), CellEqualOp(*this));
 
 		std::vector<DuplicateCell> duplicates;
-		for (const std::vector<std::vector<CellHash>> &buckets : in.bucketed_cell_hashes) {
+		for (std::vector<std::vector<CellHash>> &buckets : in.bucketed_cell_hashes) {
 			// Clear out our buckets as we go. This keeps the work of deallocation
 			// off the main thread.
 			std::vector<CellHash> bucket = std::move(buckets[index]);
```
widlarizer, line 330: Before your change, this assertion would have tripped, I tested locally:
```suggestion
			std::vector<CellHash> bucket = std::move(buckets[index]);
			log_assert(buckets[index].capacity() == 0);
```
We can add it to make sure, running it should be free, but it's not necessary now that we know we fixed it

widlarizer (approved):

jeras:
I have two issues with this merge:
1. the build documentation was not updated yet, but I managed to build and install by just following a CMake tutorial, so I am rather happy with the change.
2. Since the version was previously stored inside `Makefile`, this merge regressed the version number to:
```
$ yosys --version
Yosys 0.64+305 (git sha1 136d80e3d, GNU /usr/bin/c++ 15.2.0) [git@github.com:YosysHQ/yosys at main]
```

widlarizer:
@jeras Do you mean the merge of #5895? Sorry for the build docs inconvenience. What's wrong with the version number, that a clean build was required for it to be correct?

jeras:
Yes, sorry I meant to post this in #5895, but I refreshed the browser and did not see it was not the last commit anymore.

The problem is the version reported from a clean build should be `0.66+`, but is `0.64+305`, since this is where the the Cmake update was initially branched from. The `Makefile` previously updated to `0.66+` was deleted during this merge.

PR #5895, opened 2026-05-18
Migrate build system to CMake

whitequark:
This is a very large change that completely replaces the existing build system with a CMake one. Essentially all functionality is preserved, although the interface changes, and some features may by necessity work slightly differently.

This PR has been tested on Linux (glibc), Linux (musl), Windows (MSYS), Windows (MinGW cross), macOS, FreeBSD, OpenBSD, and on a number of different architectures. It is not feasible to test every combination of features, OSes, and architectures, but a great deal of effort went into making sure everything works.

Some aspects have been left to other team members to finish as per previous discussions.

---

not implemented:
- CI updates (left to Micko)
- doc updates (left to Krystine)

config migration:
- ENABLE_TCL - YOSYS_WITHOUT_TCL
- ENABLE_ABC - YOSYS_WITHOUT_ABC
- ENABLE_GLOB - automatic
- ENABLE_PLUGINS - YOSYS_WITHOUT_LIBFFI
- ENABLE_READLINE - YOSYS_WITHOUT_READLINE
- ENABLE_EDITLINE - YOSYS_WITHOUT_EDITLINE
- ENABLE_GHDL - unsuitable
- ENABLE_VERIFIC - YOSYS_VERIFIC_DIR=...
- ENABLE_VERIFIC_SYSTEMVERILOG - YOSYS_VERIFIC_FEATURES=systemverilog
- ENABLE_VERIFIC_VHDL -  YOSYS_VERIFIC_FEATURES=vhdl
- ENABLE_VERIFIC_HIER_TREE - YOSYS_VERIFIC_FEATURES=hier_tree
- ENABLE_VERIFIC_YOSYSHQ_EXTENSIONS - YOSYS_VERIFIC_FEATURES=extensions
- ENABLE_VERIFIC_EDIF - YOSYS_VERIFIC_FEATURES=edif
- ENABLE_VERIFIC_LIBERTY - YOSYS_VERIFIC_FEATURES=liberty
- ENABLE_LIBYOSYS - YOSYS_INSTALL_LIBRARY
- ENABLE_LIBYOSYS_STATIC - BUILD_SHARED_LIBS
- ENABLE_ZLIB - YOSYS_WITHOUT_ZLIB
- ENABLE_HELP_SOURCE - YOSYS_ENABLE_HELP_SOURCE
- ENABLE_PYOSYS - YOSYS_WITH_PYTHON
- PYOSYS_USE_UV - automatic
- ENABLE_GCOV - YOSYS_ENABLE_COVERAGE
- ENABLE_GPROF - YOSYS_ENABLE_PROFILING
- ENABLE_DEBUG - CMAKE_BUILD_TYPE=Debug
- ENABLE_LTO - CMAKE_INTERPROCEDURAL_OPTIMIZATION=ON
- ENABLE_CCACHE - YOSYS_COMPILER_LAUNCHER=ccache
- ENABLE_SCCACHE - YOSYS_COMPILER_LAUNCHER=sccache
- ENABLE_FUNCTIONAL_TESTS
- LINK_CURSES - automatic
- LINK_TERMCAP - automatic
- LINK_ABC - YOSYS_ABC_EXECUTABLE=INTEGRATED-NOTFOUND
- DISABLE_SPAWN - automatic
- ENABLE_THREADS - automatic
- DISABLE_ABC_THREADS - auto

testing strategy:
- `sh ../misc/cmake/check_missing_depends.sh` - OK
- `diff share/ build/share/` - OK
- `ctest --test-dir tests/unit/` - OK

moderate changes:
- positive options for 'default-on' features like `ENABLE_READLINE` are replaced with negative options like `WITHOUT_READLINE`
  - rationale: most users will want to build with every autodetected library without opting in, and most users who lack some obscure library for a feature they don't care about would not want to receive an error because of it. the minority which wants to avoid linking to usable and present libraries can opt out explicitly
  - alternatives: need a three-state readline switch: force on, force off, don't care. this is quite awkward to implement for every feature we need, but can be done with a custom module
  - drawbacks of not picking an alternative: it's hard to positively assert that a feature was turned on (but might be less bad ergonomically than the three-state switch..?)
- exact options returned by `yosys-config` have changed significantly and not in a way that can be compensated
- `yosys-config --cxxflags` no longer leaks `-std=...`, `-gsplit-dwarf`, `-O3`, `-Werror`, etc
  - this is potentially a breaking change, but i want to note that anybody with non-trivial plugin needs had to work around this honestly broken behavior for many years now
- `yosys-config --cxxflags` no longer has `-DYOSYS_ENABLE_*` definitions (they are in `kernel/yosys_config.h`)
  - also potentially a breaking change but those don't seem all that useful for the most part
  - could be solved with `-include` cxxopt to force pull in `kernel/yosys_config.h`?
- the method to make a "built-in" GHDL build has been removed because it will require GHDL plugin changes

small changes:
- compiler identification has slightly different format
- git origin information no longer privileges github by hiding the domain
- new `yosys_config.h.in` template, for configuration defines
- new `#define YOSYS_ENABLE_LIBFFI` (plugins just need dlopen, dpicall needs dlopen+libffi)
- explicit linking to curses/termcap dropped (readline/editline *should* pull this in via pkgconfig)
- `yosys-config --cxx` no longer remembers that you used ccache when building yosys
- stripping is not done unconditionally during installation but requires `cmake --install build --strip`
- pyosys generator now doesn't generate invalid code for Tcl+Python build
- `uv` is no longer a hard-dep the moment you turn on pyosys, but is used opportunistically
- replaced `ENABLE_VERIFIC_*` with `-DYOSYS_VERIFIC_FEATURES=...`
- sanitizer builds are done with `-DCMAKE_BUILD_TYPE=Sanitize -DSANITIZE=undefined,address`
- flake now requires Nix >=2.27

tiny changes:
- version.cc is generated in a different way
- cellhelp generation has been moved from passes/techmap/ to kernel/
- smt2 python files have slightly different substitutions
- gatemate lut tree library generator uses different output path
- `-DYOSYS_PROGRAM_PREFIX` always defined, `#ifdef`s removed
- `-DYOSYS_DATDIR` always defined, `#ifdef`s removed
- `-DYOSYS_SRC` for log.cc removed (unused)
- `-DYOSYS_VERSION`&c removed internally (unused), kept for export
- pyosys generator now produces a `*.d` depfile (+ a few more adjustments)
- ENABLE_LIBYOSYS_STATIC replaced with `-DBUILD_SHARED_LIBS=OFF`
- ENABLE_CCACHE replaced with `-DYOSYS_COMPILER_LAUNCHER=ccache`
- ENABLE_SCCACHE replaced with `-DYOSYS_COMPILER_LAUNCHER=sccache`
- enabling "linked-in ABC" is now `-DYOSYS_ABC_EXECUTABLE=INTEGRATED-NOTFOUND`
- ENABLE_LTO replaced with `-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON`
- replaced `-DYOSYS_DISABLE_SPAWN` with `#define YOSYS_ENABLE_SPAWN` for consistency


whitequark:
OK, this is done from my side, other team members have their aspects of this effort to complete. Please feel free to ask any questions you might have (this applies to non-YosysHQ folks as well), complain if the build breaks (it shouldn't! not even on FreeBSD or Haiku), etc.

dettus:
I like it! :+1:

cmake/YosysAbc.cmake
```diff
@@ -0,0 +1,96 @@
...
+			endif()
+		endforeach()
+	endforeach()
+endfunction()
+
+function(_yosys_abc_extract_makefile result vardecl filename)
+	# Parse a Makefile fragment and extracts the first matching variable assignment into
+	# a list of values.
+	file(READ ${filename} contents)
+	set_property(DIRECTORY APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS ${contents})
```
cr1901: (This is relevant to me because bypassing the Makefile may render most of [a patch](https://github.com/berkeley-abc/abc/pull/435) I want to submit to abc redundant)

Why isn't it `set_property(DIRECTORY APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS ${filename})`? It [looks like](https://cmake.org/cmake/help/latest/prop_dir/CMAKE_CONFIGURE_DEPENDS.html#prop_dir:CMAKE_CONFIGURE_DEPENDS) `CMAKE_CONFIGURE_DEPENDS` wants a path, not a string of read file contents.

whitequark: Oh that's definitely a bug

whitequark: Fixed, thanks

cr1901:
I tried looking for a solution myself, but got bored. As you [alluded to](https://mastodon.social/@whitequark@treehouse.systems/116583139183721591) last week, Windows builds are hitting some sort of export limit now:

```
William@DESKTOP-3H1DSBV MINGW64 ~/Projects/FPGA/yosys/build
$ ninja
[1/1] Linking CXX executable yosys.exe
FAILED: [code=1] yosys.exe yosys.exe.a
C:\Windows\system32\cmd.exe /C "cd . && C:\msys64\mingw64\bin\c++.exe  -Wall -Wextra -Werror=unused -Wl,--export-all-symbols @CMakeFile

s\yosys.rsp -o yosys.exe -Wl,--out-implib,yosys.exe.a -Wl,--major-image-version,0,--minor-image-version,0 && cd ."
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/16.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: error: export ordinal too large:

66447
collect2.exe: error: ld returned 1 exit status
ninja: build stopped: subcommand failed.
```

Is there a `cmake`-level workaround at the cost of "plugins don't work"? It's not clear to me why this became a problem w/ the `cmake` build, when MINGW has had `--out-implib` [for a while now](https://github.com/YosysHQ/yosys/commit/d1b3a250db111364a3356dbc9e0f715aff159b48).

I think you can [use a DEF file](https://sourceware.org/binutils/docs/ld/WIN32.html) to avoid the need for ordinals, but somehow, I don't think you want that :P.

CMakeLists.txt
```diff
@@ -0,0 +1,511 @@
...
+	set(libyosys_type SHARED)
+else()
+	set(libyosys_type STATIC)
+endif()
+yosys_cxx_library(libyosys ${libyosys_type}
+	OUTPUT_NAME libyosys
+	INCLUDE_IN_ALL_IF ${YOSYS_INSTALL_LIBRARY}
+)
+yosys_link_components(libyosys PRIVATE ${library_components})
+add_library(Yosys::libyosys ALIAS libyosys)
```
cr1901: What is `Yosys::libyosys` syntax?

mmicko:
@cr1901 
Thing is that Debug build is default now, but if you do Release build this will error will not appear in mingw builds

CMakeLists.txt
whitequark: It's just a name that cannot be a filesystem path.

whitequark:
> @cr1901 Thing is that Debug build is default now, but if you do Release build this will error will not appear in mingw builds

Oh, so *this* is the condition for the ordinal issue... I spent a while trying to figure out why it was broken & then got fixed seemingly on its own, but this makes sense. I'll have to look into it a bit more.

cr1901:
> Thing is that Debug build is default now,

Compiling now, but this would explain why the link took an unusual amount of memory, even without LTO.

Could we possibly show a string which build config is active during config, and maybe even warn "you don't want a Debug build on Windows, it will not reliably link" right now?

whitequark:
Well for one I'd like to understand more exactly why it has so many symbols in Debug builds only.

Ravenslofty:
(Perhaps also worth checking `RelWithDebInfo`?)

cr1901:
For me, when using `yosys-config` from the install directory, `--ldlibs` returns an unqualified path to `yosys.exe.a`:

```
$ /home/William/Projects/FPGA/yosys/build/yosys-config --ldflags --ldlibs
-LC:/msys64/home/William/Projects/FPGA/yosys/build yosys.exe.a
```

I think we either want a full path or to use `-l:yosys.exe.a` instead. I have not tested what happens when you install it b/c I'm not prepared to blow away my currently-installed `yosys` right now.

(Also, should `yosys-config` work for MSVC, or is this out of scope for now?)

whitequark:
> For me, when using `yosys-config` from the install directory, `--ldlibs` returns an unqualified path to `yosys.exe.a`:

This is intended. (Not asserting that it's necessarily correct.)

> I think we either want a full path or to use `-l:yosys.exe.a` instead. I have not tested what happens when you install it b/c I'm not prepared to blow away my currently-installed `yosys` right now.

You could use `-DCMAKE_INSTALL_PREFIX=c:\temporary-yosys` to test it. Please let me know how that goes if you do it.

whitequark:
> (Also, should yosys-config work for MSVC, or is this out of scope for now?)

I can't see how it would work without breaking the contract. The plan is to transition to a CMake configuration file in a follow-up PR (since this is strictly speaking beyond the scope of just build system migration.)

cr1901:
> You could use -DCMAKE_INSTALL_PREFIX=c:\temporary-yosys to test it. Please let me know how that goes if you do it.

I used ```-DCMAKE_INSTALL_PREFIX=`pwd`/../tmp-install``` as the install directory. Trying to compile a plugin still doesn't work, same error:

```
William@DESKTOP-3H1DSBV MINGW64 ~/Projects/FPGA/ghdl-yosys-plugin
$ /home/William/Projects/FPGA/yosys/tmp-install/bin/yosys-config --exec --cxx --cxxflags --ldflags -o ghdl.so ghdl.o --ldlibs -shared C:/msys64/mingw64/lib/libghdl-6_0_0_dev.dll -Wl,-rpath,C:/msys64/mingw64/lib
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/16.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find yosys.exe.a: No such

file or directory
collect2.exe: error: ld returned 1 exit status
```

The following manual command line with `-l:yosys.exe.a`, however, _does_ work.

```
William@DESKTOP-3H1DSBV MINGW64 ~/Projects/FPGA/ghdl-yosys-plugin
$ /home/William/Projects/FPGA/yosys/tmp-install/bin/yosys-config --exec --cxx --cxxflags --ldflags -o ghdl.so ghdl.o -l:yosys.exe.a -shared C:/msys64/mingw64/lib/libghdl-6_0_0_dev.dll -Wl,-rpath,C:/msys64/mingw64/lib
```

All of this is to say... it requires some changes on the `ghdl`/`ghdl-yosys-plugin` side. And I'm doing stuff I shouldn't<sup>1</sup>. But on Windows I can get the `ghdl` plugin to load. So the loss of `ENABLE_GHDL` won't change much _for me_. Hope this is a useful data point.

1. AFAICT, GHDL doesn't make an import library for `libghdl`. GNU `ld` can handle linking to a DLL directly, but MSVC can't. Also I'm masquerading `libgnat-16.dll` as `libgnat-15.dll`, because my GHDL is a bit out of date.

whitequark:
> The following manual command line with `-l:yosys.exe.a`, however, _does_ work.

Thanks, adjusted the flag.

whitequark:
> All of this is to say... it requires some changes on the `ghdl`/`ghdl-yosys-plugin` side.

Yes, it's expected that the GHDL Yosys plugin will need adjustment.

cr1901:
> (Perhaps also worth checking `RelWithDebInfo`?)

This also, worked fine, thanks :D! I forgot about it.

cr1901:
> Yes, it's expected that the GHDL Yosys plugin will need adjustment.

I got a bit nerdsniped, so I did a proof of concept of this: https://github.com/ghdl/ghdl/pull/3273

docs/Makefile
```diff
@@ -245,45 +247,40 @@ clean-examples: $(CLEAN_EXAMPLES)
...
+	@cp ../backends/functional/smtlib.cc source/generated/functional/
+	-@cd .. && diff -U 20 backends/functional/smtlib.cc backends/functional/smtlib_rosette.cc \
+		> docs/source/generated/functional/rosette.diff || true
+
+	@$(YOSYS) --help >source/generated/yosys
+	@$(YOSYS_SMTBMC) --help >source/generated/yosys-smtbmc
+	@$(YOSYS_WITNESS) --help >source/generated/yosys-witness
+	@$(YOSYS_CONFIG) --help >source/generated/yosys-config
+	-@$(YOSYS_FILTERLIB) --help 2>source/generated/yosys-filterlib || true
+	-@$(ABC) --help 2>source/generated/yosys-abc > /dev/null || true
```
KrystalDelusion: Two things here, I think both may have been added by @mmicko.

The `-` at the start of the line means that Make ignores the error (printing e.g.
`make[1]: [Makefile:271: extract] Error 1 (ignored)`) and returns 0.  If we want to silently ignore the error instead (the `|| true`) then we don't need the `-` and should remove it.

Second thing is that without the `ifeq ($(ABCEXTERNAL),)` the docs will output the help text for the external abc in a section of text that refers to it as `yosys-abc`... or so I thought, in practice it seems like the docs targets still generate `yosys-abc`.  It looks like `--target test-vanilla` is doing the same.  Is that intentional?

KrystalDelusion: This one might be for @whitequark, I'm also noticing that cmake doesn't like to configure without an included abc.  If I provide `YOSYS_ABC_EXECUTABLE` then I shouldn't need to have the submodule.  By default (with an empty/missing abc directory) it will raise an error asking that I initialize the submodule, if I provide `YOSYS_SKIP_ABC_SUBMODULE_CHECK` as well then it does skip that check, but then fails with the following:
```
CMake Error at cmake/YosysAbc.cmake:19 (file):
  file failed to open for reading (No such file or directory):

    /home/dawn/yosys/abc/Makefile
Call Stack (most recent call first):
  cmake/YosysAbc.cmake:35 (_yosys_abc_extract_makefile)
  CMakeLists.txt:387 (yosys_abc_target)
```

kernel/register.h
```diff
@@ -23,29 +23,15 @@
...
-};
+#include <version>
+#if __cpp_lib_source_location >= 201907L
+    #include <source_location>
+    using std::source_location;
+#elif defined(__has_include)
+    #if __has_include(<experimental/source_location>)
+        #include <experimental/source_location>
+        using std::experimental::source_location;
+    #endif
```
KrystalDelusion: This feels either incomplete or unnecessarily nested; either we do
```
#else
	#include <experimental/source_location>
	using std::experimental::source_location;
#endif
```
or we need to account for the possibility that we don't have a `source_location`.  (Also this section of code is using spaces)

CMakeLists.txt
```diff
@@ -0,0 +1,557 @@
...
+		URL "https://www.python.org/"
+		DESCRIPTION "Dynamic programming language (Embedding)"
+		PURPOSE "Binding Yosys API"
+	)
+
+	find_package(PyosysEnv REQUIRED)
+	set_package_properties(PyosysEnv PROPERTIES
+		DESCRIPTION "Pyosys wrapper generator environment"
+		PURPOSE "Either 'uv' or 'pybind11>3,<4 cxxheaderparser'"
+	)
```
KrystalDelusion, line 275: Does this make uv optional, and if so should we advertise it as such or let people figure it out themselves if they want to avoid it?

docs/Makefile
```diff
@@ -245,18 +247,40 @@ clean-examples: $(CLEAN_EXAMPLES)
...
 images:
-	$(MAKE) -C source/_images
-	$(MAKE) -C source/_images convert
+	@$(MAKE) -C source/_images
+	@$(MAKE) -C source/_images convert
+
+.PHONY: extract
+extract:
+	@mkdir -p source/generated
+	@$(YOSYS) -q -p 'help -dump-cmds-json source/generated/cmds.json'
```
KrystalDelusion, line 260: The switch to CMake seems to have changed all of the `source_location`s from relative paths to absolute paths, which causes problems for the `cmds.json` file in the command group detection.  Adding a `-DCMAKE_CXX_FLAGS="-fmacro-prefix-map=$(pwd)/="` when configuring resolves this, but if we make that default we probably want to drop it from the `yosys-config --cxxflags` output.  Alternatively it should be possible to fixup the paths in the `Pass` struct by comparing it to the location of `register.cc`.  @whitequark do you have a preference here?

docs/Makefile
mmicko: @KrystalDelusion Agree `-` is not needed in both cases so it should be removed.
Regarding ABCEXTERNAL, since we always need to provide location of ABC we can generate a file, but as I am aware documentation creation will fail if there is missing file so even if we skip creating some of files in `source/generated/`.
For testing we need to provide it anyway since it can be called during testing as separate binary, but for here we could also output ABCEXTERNAL in case it is defined as such.
But please do suggest your opinion.

kernel/register.h
mmicko: Agree your code should be enough.

docs/Makefile
mmicko: Also usage lines in generated files are altered, but we can use sed on them I guess

CMakeLists.txt
whitequark: Yes, this makes uv optional, and when the wheel is built via pyproject.toml, uv is not used (or should not be used unless I missed something).

docs/Makefile
whitequark: > Alternatively it should be possible to fixup the paths in the Pass struct by comparing it to the location of register.cc.

This is more portable since `-fmacro-prefix-map` may not be implemented in all compilers.

docs/Makefile
whitequark: Fixed the `-` and the abc submodule issue.

kernel/register.h
whitequark: Fixed

pyosys/build/local_backend.py
```diff
@@ -0,0 +1,146 @@
...
+PROJECT_VERSION = subprocess.check_output([
+	"cmake",
+	f"-DCMAKE_SOURCE_DIR={os.getcwd()}",
+	"-P", "cmake/GetPyosysVersion.cmake"
+], encoding="ascii").strip()
+DIST_NAME = f"{PROJECT_NAME}-{PROJECT_VERSION}"
+
+# https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/
+if sys.implementation.name == "cpython":
+	PYTHON_TAG = f"cp{sysconfig.get_config_var("py_version_nodot")}"
```
donn: f-strings need an alternate set of quotes inside the string before Python 3.12, see https://docs.python.org/3/whatsnew/3.12.html#whatsnew312-pep701

```suggestion
	PYTHON_TAG = f"cp{sysconfig.get_config_var('py_version_nodot')}"
```

pyosys/build/local_backend.py
```diff
@@ -0,0 +1,146 @@
...
+def build_wheel(wheel_dir, config_settings=None, metadata_directory=None):
+	wheel_filename = f"{DIST_NAME}-{COMPAT_TAG}.whl"
+
+	with WheelFile(pathlib.Path(wheel_dir) / wheel_filename, "w") as wheel:
+		for filename, contents in get_metadata_files().items():
+			wheel.writestr(f"{DIST_NAME}.dist-info/{filename}", contents)
+
+		cmake_options = []
+		if config_settings is not None:
+			if cmake_options := config_settings.get("cmake"):
```
donn: The walrus operator assigns to `cmake_options` even if the right-hand side is None, so if `config_settings` is not None but `config_settings` has no `cmake` key, `cmake_options` becomes `None` and the splat operator on line 48 throws a TypeError

This worked in my testing. `bool([])` is false but it doesn't actually matter in this case

```suggestion
			if cmake_options := config_settings.get("cmake") or cmake_options:
```

pyproject.toml
```diff
@@ -1,10 +1,11 @@
...
 	"pybind11>=3,<4",
-	"cxxheaderparser"
+	"cxxheaderparser",
 ]
-build-backend = "setuptools.build_meta"
+backend-path = ["pyosys/build"]
+build-backend = "local_backend"
 
 [tool.ruff]
 target-version = "py38"
```
donn: While we're at it, might be cool to drop 3.8/3.9 (both are EOL)

```suggestion
target-version = "py310"
```

donn (commented):
Still working on making `cibuildwheel` work again, but until I do I did some testing locally and have a couple small fixes

whitequark:
@donn All applied, thanks! The final version is slightly different from your suggestions but should be equivalent.

whitequark:
Rebased on master.

donn:
There's a complication I'm running into which is basically, technically to build the wheel you need the component `Development.Module` instead of `Development.Embed`. You need `Development.Embed` for `yosys -y` on the other hand. I had this pretty jank solution for the Makefile:

https://github.com/YosysHQ/yosys/blob/e56c6a954ce26b6263dedd8a9f853a468b925c83/Makefile#L225-L229

I'm trying to recreate this in the CMake file tree somehow

EDIT: Also, the manylinux Docker images (used to build wheels) intentionally do not provide `Development.Embed` which means I can't just link both

whitequark:
I'll take a look today or tomorrow

whitequark:
> There's a complication I'm running into which is basically, technically to build the wheel you need the component `Development.Module` instead of `Development.Embed`. You need `Development.Embed` for `yosys -y` on the other hand. I had this pretty jank solution for the Makefile:

Could you explain what the specific difference between the two is? Which step would fail exactly if you only had `Development.Module` and not `Development.Embed`?

donn:
Sorry for the late response:

- `Development.Module` exposes Python headers and links libraries needed by Python, similar to `python3-config --libs`, but does not link Python itself or expose the `Python3::Python` target in CMake, exposing `Python3::Module` instead
  - The resulting binary can be loaded by `libpython`
- `Development.Embed` exposes the same headers and links the same libraries, similar to `python3-config --embed --libs`, but also links Python itself and exposes the `Python3::Python` target in CMake
  - The resulting binary itself loads `libpython`

```
~% python3-config --include
-I/nix/store/h3l9h1vmbzvgs4x1ffl0pqbmzhlk89yl-python3-3.13.12/include/python3.13 -I/nix/store/h3l9h1vmbzvgs4x1ffl0pqbmzhlk89yl-python3-3.13.12/include/python3.13
~% python3-config --embed --include
-I/nix/store/h3l9h1vmbzvgs4x1ffl0pqbmzhlk89yl-python3-3.13.12/include/python3.13 -I/nix/store/h3l9h1vmbzvgs4x1ffl0pqbmzhlk89yl-python3-3.13.12/include/python3.13
~% python3-config --ldflags        
-ldl -framework CoreFoundation
~% python3-config --embed --ldflags
-lpython3.13 -ldl -framework CoreFoundation
```

Since wheels are supposed to be loaded by libpython and they don't want you linking the .so itself against libpython, they don't make `Development.Embed` available.

As the headers and libraries of `embed` are in theory a strict superset of `module`, you can list `Python3::Module` as a library for libyosys/pyosys.so and `Python3::Python` for the `yosys` executable, but the hard part is making the CMake logic tolerate not compiling the `yosys` executable if Python is enabled but `Python3::Python` is missing

EDIT: One way to do this is a separate `YOSYS_WITH_PYTHON_INTERPRETER` based the result of `Development.Embed` while keeping `YOSYS_WITH_PYTHON` for `Development.Module`

Ofc that would mean you can compile pyosys without `yosys -y` but them's the breaks

whitequark:
> EDIT: One way to do this is a separate `YOSYS_WITH_PYTHON_INTERPRETER` based the result of `Development.Embed` while keeping `YOSYS_WITH_PYTHON` for `Development.Module`

Yes I think this is the way to go here!

whitequark:
Would you like to try and implement it? If you have a rough working sketch I'm happy to review and polish it.

donn:
Yeah, I can tackle it.

donn:
Rough working sketch: https://github.com/YosysHQ/yosys/pull/5924

donn:
At @mmicko's request I tracked down why macOS wheels were failing to configure. Unfortunately, it is very silly.

Basically, on macOS at least, `Python3_EXECUTABLE` == `$VIRTUAL_ENV/bin/python` != the `python` wrapper in the "isolated build environment" created by `python -m build`.

Yep.

The solution is to just fall back to PATH:

https://github.com/donn/yosys/commit/63b0e6377928db164291b3d6abbdce7bb2a91cac.patch

I'm testing a CI run here:

https://github.com/donn/yosys/actions/runs/26822766681/job/79081664116

EDIT: P.S. I renamed the venv strategy to virtualenv to minimize my own confusion because apparently those are two different things: https://virtualenv.pypa.io/en/latest/explanation.html#virtualenv-vs-venv-vs-uv

Feel free to revert this rename though

EDIT 2: Okay, new failure during fixup, looking at that one as well

whitequark:
Thanks for tracking this down! I'm wondering if the way to fix it is to provide `Python3_EXECUTABLE` from the build backend (I'm trying this now)

whitequark:
I think that worked! https://github.com/YosysHQ/yosys/actions/runs/26828585237/job/79103085033

donn:
EDIT: DO NOT APPLY THIS FIX, give me a sec working on a better one

---

Looks like it! What's left is just the compat_tag issue, it really doesn't like the `.` in the macOS platform tag

```
>>> from packaging.utils import parse_wheel_filename
>>> list(parse_wheel_filename('pyosys-0.64.post303-cp310-none-macosx_11.0_arm64.whl'))
['pyosys', <Version('0.64.post303')>, (), frozenset({<cp310-none-macosx_11 @ 4435197184>, <cp310-none-0_arm64 @ 4435196544>})]
```

Plus there's a collision on all platforms bec local_backend.py doesn't account for both `cp314` and `cp314t`. I'm testing whether just `str(next(packaging.tags.cpython_tags()))` will always just get the correct compat_tag:

```diff
commit f416e797ce8f441e12acef265560a32cf686dd52
Author: Mohamed Gaber <me@donn.website>
Date:   Tue Jun 2 18:06:45 2026 +0300

    fix abi and platform tags

diff --git a/pyosys/build/local_backend.py b/pyosys/build/local_backend.py
index cf22215b5..58607bfef 100644
--- a/pyosys/build/local_backend.py
+++ b/pyosys/build/local_backend.py
@@ -9,7 +9,7 @@ import pathlib
 import tarfile
 import tempfile
 import subprocess
-import sysconfig
+import packaging.tags
 from email.policy import EmailPolicy
 from email.message import EmailMessage
 from wheel.wheelfile import WheelFile
@@ -23,13 +23,10 @@ PROJECT_VERSION = subprocess.check_output([
 ], encoding="ascii").strip()
 DIST_NAME = f"{PROJECT_NAME}-{PROJECT_VERSION}"
 
-# https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/
-if sys.implementation.name == "cpython":
-	PYTHON_TAG = f"cp{sysconfig.get_config_var('py_version_nodot')}"
-else:
+if sys.implementation.name != "cpython":
 	raise NotImplementedError("unsupported Python implementation")
-PLATFORM_TAG = sysconfig.get_platform().replace("-", "_")
-COMPAT_TAG = f"{PYTHON_TAG}-none-{PLATFORM_TAG}"
+# https://packaging.pypa.io/en/stable/tags.html#packaging.tags.cpython_tags
+COMPAT_TAG = str(next(packaging.tags.cpython_tags()))
 
 
 def compile_pyosys(cmake_options=[], parallel=os.cpu_count() or 1):
```

https://github.com/donn/yosys/actions/runs/26829924070/job/79108007524

donn:
Good news: wheels are getting past the delocate phase now

Bad news: tests are failing on macOS……………

…I do need to take a break for now

mmicko:
This solves issue on macOS and it is actually how it should be declared for python modules
```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 5633036b3..287631547 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -453,7 +453,7 @@ endif()
 
 # Python binary extension (for using Yosys as a Python library).
 if (YOSYS_ENABLE_PYTHON)
-       yosys_cxx_library(pyosys SHARED
+       yosys_cxx_library(pyosys MODULE
                OUTPUT_NAME pyosys
        )
        yosys_link_components(pyosys PRIVATE ${library_components})
@@ -473,7 +473,7 @@ if (YOSYS_ENABLE_PYTHON)
                        DESTINATION ${pyosys_install_dir}
                )
                install(FILES $<TARGET_FILE:pyosys>
-                       RENAME libyosys${CMAKE_SHARED_LIBRARY_SUFFIX}
+                       RENAME libyosys${CMAKE_SHARED_MODULE_SUFFIX}
                        DESTINATION ${pyosys_install_dir}
                )
                if (YOSYS_ENABLE_ABC AND YOSYS_ABC_EXECUTABLE STREQUAL "")
```

We just need to make google tests not to be compiled or to link with python libraries when compiling pyosys, that is reason why currently CI is failing

donn:
I think I have it now…

https://github.com/YosysHQ/yosys/pull/5928

mmicko:
@whitequark  Also got gtests to link when pyosys is used, but there can be cleaner solution

```
diff --git a/tests/unit/CMakeLists.txt b/tests/unit/CMakeLists.txt
index c52bf2e4e..633b6cf91 100644
--- a/tests/unit/CMakeLists.txt
+++ b/tests/unit/CMakeLists.txt
@@ -7,7 +7,7 @@ function(yosys_gtest arg_TARGET)
 
        set(target "gtest-${arg_TARGET}")
        add_executable(${target} ${arg_SOURCES})
-       target_link_libraries(${target} PRIVATE GTest::gmock_main yosys_common)
+       target_link_libraries(${target} PRIVATE GTest::gmock_main yosys_common $<${YOSYS_ENABLE_PYTHON}:Python3::Python>)
        yosys_expand_components(test_components essentials ${arg_COMPONENTS})
        yosys_link_components(${target} PRIVATE ${test_components})
 

```

whitequark:
> This solves issue on macOS and it is actually how it should be declared for python modules

Pulled this in

whitequark:
> @whitequark Also got gtests to link when pyosys is used, but there can be cleaner solution

Checking a solution I found yesterday, if that doesn't work I'll use yours

mmicko (approved):

PR #5897, opened 2026-05-19
Enable Compilation on OpenBSD 

dettus:

I wanted to be able to git pull the latest version from the repo and compile it without carrying my own patch around.

By adding one environment variable to the Makefile.

Ideally, the change is transparent to existing tests.


widlarizer:
Bad timing! We'll be switching to CMake imminently. #5895 is reported to work on OpenBSD. Please comment on that PR if not. Until then, we won't be reviewing this change since this patch won't last if merged anyway.

PR #5900, opened 2026-05-19
arith_tree improvements

hexratcc:
This PR aims to improve QoR for `airth_tree`, namely:
- [**FMA**](https://en.wikipedia.org/wiki/Multiply%E2%80%93accumulate_operation): For `$macc` terms with multiplications
- 4:2 compressor support (the original only supports 3:2)
- More optimal final adder selection based on datapath width
- Dadda tree scheduling

TODO:
- [x] parallel_prefix
- [x] Cleanup


hexratcc:
[sc-559]

kernel/compressor_tree.cc
```diff
@@ -0,0 +1,345 @@
...
+
+	SigSpec carry;
+	carry.append(State::S0);
+	if (width > 1)
+		carry.append(carry_full.extract(0, width - 1));
+
+	return {sum, carry};
+}
+
+SigSpec normalize_to_width(SigSpec sig, bool is_signed, int width)
```
widlarizer: I know this is pre-existing, but, `sig.extend_u0(width, is_signed);` would get the same job done as this function I think

tests/arith_tree/arith_tree_signed_fma.ys
```diff
@@ -0,0 +1,135 @@
...
+	output signed [7:0] y
+);
+	assign y = a * b + c;
+endmodule
+EOT
+hierarchy -auto-top
+proc
+alumacc
+opt
+equiv_opt arith_tree
```
widlarizer: These equiv_opt calls need the `-assert` flag added so that they actually fail the test if the circuits are not equivalent

kernel/compressor_tree.h
```diff
@@ -0,0 +1,118 @@
...
+std::pair<SigSpec, SigSpec> reduce_scheduled(Module *module, std::vector<DepthSig> operands, int width, Strategy strategy, int *out_compressor_count = nullptr, int *out_final_depth = nullptr);
+
+/**
+ * emit_kogge_stone() - Emit a Kogge-Stone parallel-prefix adder
+ * @module: The Yosys module to which the gates will be added
+ * @a: Signal A
+ * @b: Signal B
+ * @y: Signal Y = (A + B) mod 2^W
+ */
+void emit_kogge_stone(Module *module, SigSpec a, SigSpec b, SigSpec y);
```
widlarizer, line 97: If we have an implementation here, as well as one in `techlibs/common/choices/kogge-stone.v`, can we get a test that checks they are equivalent? It could go into a new small and simple test pass like `passes/tests/test_cell.cc` since it needs C++

passes/techmap/arith_tree.cc
```diff
@@ -69,59 +77,55 @@ struct Cells {
...
-
 		return true;
 	}
-};
-
-struct Rewriter {
-	Module *module;
-	Cells &cells;
-	Traversal traversal;
-	AluInfo alu_info;
```
widlarizer, line 111: Hmm I kind of liked this split between traversal/analyses/modification. Any particular reason to flatten these structs?

kernel/compressor_tree.h
```diff
@@ -0,0 +1,118 @@
...
+#include "kernel/sigtools.h"
+#include "kernel/yosys.h"
+
+YOSYS_NAMESPACE_BEGIN
+
+namespace CompressorTree
+{
+
+// Width and depth thresholds below which a ripple is preferred over parallel-prefix
+constexpr int RIPPLE_PREFIX_THRESHOLD = 16;
```
widlarizer: Was this chosen based on experience? Or literature?

widlarizer (changes requested):
Some style and testing coverage suggestions

kernel/compressor_tree.h
hexratcc: It's based on [Binary Adder Architectures for Cell-Based VLSI and their Synthesis](https://iis-people.ee.ethz.ch/~zimmi/publications/adder_arch.pdf), specifically tables 4.7 and 4.9 (post-layout area and AT comparisons - RCA/Kogge Stone). The threshold of `16b`  should be the point where Kogge Stone isn't strictly less efficient than RCA.

passes/techmap/arith_tree.cc
hexratcc: When I was working on the improvements the coupling felt a bit too rigid for me, and as the pass isn't *huge* or anything it felt alright to do so (as well as getting rid of the shared state which to me feels like noise, in this case).

kernel/compressor_tree.h
widlarizer: Consider adding this reference in a comment

mikesinouye:
This had a solid improvement on one of our accumulate heavy designs we were hopeful would largely benefit. Thanks for working on it!

widlarizer (approved):

PR #5904, opened 2026-05-20
[WIP] Rapidflex xt plugin

treelin611:
To address the code comments in PR: https://github.com/YosysHQ/yosys/pull/5889

1. Removed all pugixml-related files.
2. Changed logging level to logging.ERROR.
3. Removed the redundant line in Makefile.inc.


techlibs/rapidflex/alkaidC/cell_sim_arith.v
```diff
@@ -0,0 +1,15 @@
+//---------------------------------------
+// 1-bit adder 
+//---------------------------------------
+(* abc9_box, lib_whitebox *)
```
Ravenslofty, line 4: This is an `(* abc9_box *)` but it has no timings.

techlibs/rapidflex/alkaidC/cell_sim_ff.v
```diff
@@ -0,0 +1,586 @@
+//-----------------------------
+// Rising-edge D-type flip-flop
+//-----------------------------
+(* abc9_flop, lib_whitebox *)
```
Ravenslofty, line 4: This is an `(* abc9_flop *)` but it has no timings.

techlibs/rapidflex/alkaidC/cell_sim_ff.v
```diff
@@ -0,0 +1,586 @@
...
+          1'b1:
+            always @(negedge C)
+                Q <= D;
+    endcase
+endmodule
+
+//-----------------------------
+// Rising-edge D-type flip-flop with active-high asynchronous reset
+//-----------------------------
+(* abc9_flop, lib_whitebox *)
```
Ravenslofty, line 28: This flop has an asynchronous control, which cannot be `(* abc9_flop *)`

techlibs/rapidflex/alkaidC/cell_sim_ff.v
```diff
@@ -0,0 +1,586 @@
...
+                        Q <= 1'b0;
+                else
+                        Q <= D;
+    endcase
+endmodule
+
+//-----------------------------
+// Rising-edge D-type flip-flop with active-high asynchronous set
+//-----------------------------
+(* abc9_flop, lib_whitebox *)
```
Ravenslofty, line 59: This flop has an asynchronous control, which cannot be `(* abc9_flop *)`

techlibs/rapidflex/alkaidC/cell_sim_ff.v
```diff
@@ -0,0 +1,586 @@
...
+                        Q <= 1'b1;
+                else
+                        Q <= D;
+    endcase
+endmodule
+
+//-----------------------------
+// Rising-edge D-type flip-flop with active-low asynchronous reset
+//-----------------------------
+(* abc9_flop, lib_whitebox *)
```
Ravenslofty, line 90: This flop has an asynchronous control, which cannot be `(* abc9_flop *)`

techlibs/rapidflex/alkaidC/cell_sim_ff.v
```diff
@@ -0,0 +1,586 @@
...
+                        Q <= 1'b0;
+                else
+                        Q <= D;
+    endcase
+endmodule
+
+//-----------------------------
+// Rising-edge D-type flip-flop with active-low asynchronous set
+//-----------------------------
+(* abc9_flop, lib_whitebox *)
```
Ravenslofty, line 121: This flop has an asynchronous control, which cannot be `(* abc9_flop *)`

techlibs/rapidflex/alkaidC/cell_sim_ff.v
```diff
@@ -0,0 +1,586 @@
...
+          1'b1:
+            always @(negedge C)
+                Q <= D;
+    endcase
+endmodule
+
+//-----------------------------
+// Falling-edge D-type flip-flop with active-high asynchronous reset
+//-----------------------------
+(* abc9_flop, lib_whitebox *)
```
Ravenslofty, line 300: This flop has an asynchronous control, which cannot be `(* abc9_flop *)`

techlibs/rapidflex/alkaidC/cell_sim_ff.v
```diff
@@ -0,0 +1,586 @@
...
+                        Q <= 1'b0;
+                else
+                        Q <= D;
+    endcase
+endmodule
+
+//-----------------------------
+// Falling-edge D-type flip-flop with active-high asynchronous set
+//-----------------------------
+(* abc9_flop, lib_whitebox *)
```
Ravenslofty, line 331: This flop has an asynchronous control, which cannot be `(* abc9_flop *)`

techlibs/rapidflex/alkaidC/cell_sim_ff.v
```diff
@@ -0,0 +1,586 @@
...
+                        Q <= 1'b1;
+                else
+                        Q <= D;
+    endcase
+endmodule
+
+//-----------------------------
+// Falling-edge D-type flip-flop with active-low asynchronous reset
+//-----------------------------
+(* abc9_flop, lib_whitebox *)
```
Ravenslofty, line 362: This flop has an asynchronous control, which cannot be `(* abc9_flop *)`

techlibs/rapidflex/alkaidC/cell_sim_ff.v
```diff
@@ -0,0 +1,586 @@
...
+                        Q <= 1'b0;
+                else
+                        Q <= D;
+    endcase
+endmodule
+
+//-----------------------------
+// Falling-edge D-type flip-flop with active-low asynchronous set
+//-----------------------------
+(* abc9_flop, lib_whitebox *)
```
Ravenslofty, line 393: This flop has an asynchronous control, which cannot be `(* abc9_flop *)`

techlibs/rapidflex/src/synth_rapidflex.cc
```diff
@@ -0,0 +1,526 @@
...
+    /* Pre-check on family name and confirm on selection*/
+    if (family != ALKDL_DNAME && family != ALKDT_DNAME &&
+        family != ALKDC_DNAME) {
+      log_cmd_error("Invalid family specified: '%s'\n", family.c_str());
+    }
+    log("Selected device family: %s\n", family.c_str());
+    /* Force to enable/disable options upon device limits */
+    if (family == ALKDL_DNAME || family == ALKDC_DNAME) {
+      if (!nodsp) {
+        log_warning("Force to disable dsp inference as the selected device "
```
Ravenslofty:
```suggestion
        log_warning("Disabling DSP inference as the selected device "
```

techlibs/rapidflex/src/synth_rapidflex.cc
```diff
@@ -0,0 +1,526 @@
...
+                    "does not contain dedicated resources\n");
+        nodsp = true;
+      }
+      if (inferBram) {
+        log_warning("Force to disable RAM inference as the selected device "
```
Ravenslofty:
```suggestion
        log_warning("Disabling RAM inference as the selected device "
```

techlibs/rapidflex/src/synth_rapidflex.cc
```diff
@@ -0,0 +1,526 @@
...
+                    "does not contain dedicated resources\n");
+        inferBram = false;
+      }
+    }
+    /* By default, no opt should be enabled. Throw a warning if not */
+    if (no_opt) {
+      log_warning("Force to disable any optimization, which may cast an "
```
Ravenslofty:
```suggestion
      log_warning("Disabling optimization, which may have a "
```

techlibs/rapidflex/src/synth_rapidflex.cc
```diff
@@ -0,0 +1,526 @@
...
+            run("chtype -set $mul t:$__soft_mul", "  (for alkaidT)");
+          }
+          run("select a:mul2dsp", "                (for alkaidT)");
+          run("setattr -unset mul2dsp", "          (for alkaidT)");
+          if (!no_opt) {
+            run("opt_expr -fine", "                  (for alkaidT)");
+            run("wreduce", "                         (for alkaidT)");
+          }
+          run("select -clear", "                   (for alkaidT)");
+          // Comment out for further development
```
Ravenslofty, line 373: This should probably be cleaned up.

techlibs/rapidflex/src/synth_rapidflex.cc
```diff
@@ -0,0 +1,526 @@
...
+      run("memory -nomap");
+      if (!no_opt) {
+        run("opt_clean");
+      }
+    }
+
+    if (check_label("map_bram", "(skip if -no_bram)") &&
+        ((help_mode || family == ALKDT_DNAME) && inferBram)) {
+      if (help_mode || family == ALKDT_DNAME) {
+        run("memory_bram -rules " + lib_path + family + "/bram.txt");
```
Ravenslofty, line 401: `memory_bram` should be considered deprecated for new flows; you should use `memory_libmap` instead.

techlibs/rapidflex/src/synth_rapidflex.cc
```diff
@@ -0,0 +1,526 @@
...
+        run("opt_clean");
+        run("opt -nodffe");
+      }
+    }
+
+    if (check_label("map_ffs")) {
+      run("memory");
+      /* TODO: Support shift-register mapping */
+      /* Run 2 times dff mapping incase anything missing */
+      run("dfflegalize -cell $_DFF_?_ 01 -cell $_DFF_???_ 01 -cell $_SDFF_???_ "
```
Ravenslofty, line 439: If you're "leaking" flops there is an issue in your mapping; you should only have to run it once.

techlibs/rapidflex/src/synth_rapidflex.cc
```diff
@@ -0,0 +1,526 @@
...
+      if (!no_opt) {
+        run("opt_merge");
+        run("opt_dff -nodffe");
+        run("opt_clean");
+        run("opt -nodffe");
+      }
+    }
+
+    if (check_label("map_luts")) {
+      run("abc -lut " + std::to_string(max_lut_size));
```
Ravenslofty, line 457: You're...using `(* abc9_box *)` and `(* abc9_flop *)` but not using `abc9` itself? For LUT mapping purposes the `abc` command is [deprecated](https://yosyshq.discourse.group/t/upcoming-abc-flow-changes/133).

techlibs/rapidflex/src/synth_rapidflex.cc
```diff
@@ -0,0 +1,526 @@
...
+      /* Map dff and adder again since ABC may generate new gates */
```
Ravenslofty, line 458: Uh, no, this should not happen.

techlibs/rapidflex/src/clock_buffer_cmd.cc
```diff
@@ -0,0 +1,581 @@
...
+        } else {
+          process_cell_rewire_lut(module, cell, internal_signal_io_map,
+                                  internal_signal_lut, sigmap);
+          /*replace internal clk/reset with clk/reset_buf signal */
+        }
+      }
+    }
+  }
+
+  void execute(std::vector<std::string> args, RTLIL::Design *design) override {
```
Ravenslofty, line 521: Okay, I might come back and reread this later, but so far as I can tell, the purpose of this is to emit metadata about the clock buffering for...VPR?

How much of the actual buffer insertion can be replaced by calling `clkbufmap`?

techlibs/rapidflex/alkaidC/synth.ys
```diff
@@ -0,0 +1,128 @@
+# Yosys synthesis script for ${TOP_MODULE}
```
Ravenslofty: So far as I can tell this file isn't used anywhere.

techlibs/rapidflex/alkaidC/synth_no_adder.ys
```diff
@@ -0,0 +1,128 @@
+# Yosys synthesis script for ${TOP_MODULE}
```
Ravenslofty: So far as I can tell, this file isn't used anywhere.

techlibs/rapidflex/alkaidC/verilog_rewrite.ys
```diff
@@ -0,0 +1,41 @@
+# Yosys synthesis script for ${TOP_MODULE}
```
Ravenslofty: So far as I can tell, this file isn't used anywhere.

Ravenslofty (changes requested):
There's still a lot of clean-up to do here.

techlibs/rapidflex/alkaidC/cell_sim_arith.v
tangxifan: You are referring to these lines, right? 

https://github.com/YosysHQ/yosys/blob/9d0cdb8551d7262a3b41533b048295719eaf5142/techlibs/quicklogic/qlf_k6n10f/cells_sim.v#L137-L144

As RapidFlex has devices on multiple technology node and PVT corners, it is hard to set unified delays here. What is your suggestion on the numbers? Will Yosys take them and apply timing-driven optimization during synthesis flow?

techlibs/rapidflex/alkaidC/synth.ys
tangxifan: Thanks for pointing it out. We will remove all the redundant *.ys files

techlibs/rapidflex/src/clock_buffer_cmd.cc
tangxifan: Here is an example of the output XML file. The key information required is 

- clock net name
- clock buffer cell name
- type of the signal, could be ``clock`` or ``reset``

I have checked the [clkbufmap](https://yosyshq.readthedocs.io/projects/yosys/en/0.47/cmd/clkbufmap.html). It can save a lot of effort in this function.
However, not all the global signals require buffer insertion in RapidFlex architecture, only the internally generated ones (by LUTs, FFs). The job of the clock_buffer_cmd.cc is mainly to find out the internally generated ones. 

If Yosys can provide the feature already, we are glad to use it.

```
<?xml version="1.0"?>
<!-- Clock buffer cell map 
     Generated by Frontend, e.g., Yosys-plugin     
     Global signals which are not generated internally do not need a specification here. 
     VPR treats the output net of a subckt as its cell (atom) name 
  -->
<ckbuf_cell_map>
  <ckbuf input_net="int_clk" cell="int_clk_ckbuf" type="clock"/>
  <ckbuf input_net="int_rst" cell="int_rst_ckbuf" type="reset"/>
</ckbuf_cell_map>
```

techlibs/rapidflex/src/synth_rapidflex.cc
tangxifan: That is a reminder for future upgrades. We may enable it soon. 
If you do not mind, we convert this block to a code comment about the future upgrades.

techlibs/rapidflex/src/synth_rapidflex.cc
tangxifan: Do you have any detailed documentation on the two commands? I have checked the Yosys documentation but it lacks examples of the input files. 

If ``memory_bram`` is to be deprecated, is there a smooth way (i.e., application notes) to make the transition?

techlibs/rapidflex/src/synth_rapidflex.cc
tangxifan: This is indeed a workaround. I think it should be resolved by clean up mapping. However, it takes time and effort. If you have recommendations, we are glad to take it. Otherwise, this is currently the best we can do.

techlibs/rapidflex/src/synth_rapidflex.cc
tangxifan: Do you have an application note on the abc9 vs abc? 
I am glad to move to abc9 if provides better PPA.
This command is new to us. If you have any existing example, we will take it and convert.

techlibs/rapidflex/src/synth_rapidflex.cc
tangxifan: Agree. The current flow has passed full flow regression tests and validated on silicon. If Yosys has a good infra on regression tests which can sanitize the outputs and redflag mapping fails on cells, we are glad to go. 
Otherwise, I see a risk to change the recipe aggressively here without any regression tests

techlibs/rapidflex/alkaidC/cell_sim_arith.v
Ravenslofty: > What is your suggestion on the numbers?

Use worst-case timings in integer picoseconds.

> Will Yosys take them and apply timing-driven optimization during synthesis flow?

Yes.

techlibs/rapidflex/src/synth_rapidflex.cc
Ravenslofty: The way we normally do "regression testing" is to run synthesis with equivalence checking on a sample design and then assert the result has a specific number and kind of cells:
https://github.com/YosysHQ/yosys/blob/9d0cdb8551d7262a3b41533b048295719eaf5142/tests/arch/ice40/logic.ys#L1-L7

techlibs/rapidflex/src/clock_buffer_cmd.cc
Ravenslofty: > However, not all the global signals require buffer insertion in RapidFlex architecture, only the internally generated ones

I assume these can be inferred, but if Yosys produces some redundant explicit clock buffer cells does it really matter?

techlibs/rapidflex/src/synth_rapidflex.cc
Ravenslofty: https://github.com/YosysHQ/yosys/blob/main/passes/memory/memlib.md

(I will also point out that Yosys has integrated help for both commands with `help memory_bram`/`help memory_libmap`.)

techlibs/rapidflex/src/synth_rapidflex.cc
Ravenslofty: When this PR is cleaned up I will investigate myself, I suppose.

techlibs/rapidflex/src/synth_rapidflex.cc
Ravenslofty: https://yosyshq.readthedocs.io/projects/yosys/en/latest/using_yosys/synthesis/abc.html

techlibs/rapidflex/alkaidC/cell_sim_ff.v
tangxifan: Do you have a guideline on what flags should be enabled for abc9_flop? Maybe this is the reason why ff_map failed to map all the FFs.

techlibs/rapidflex/src/clock_buffer_cmd.cc
tangxifan: That matters. RapidFlex clock routing cannot handle the redundant buffers.

Ravenslofty: Hmm. With my nextpnr hat on, I would just detect and remove the redundant clock buffer, but I don't think that approach is viable in VPR? Perhaps an approach where you `clkbufmap` to insert buffers, and then have a custom pass to detect and remove top-level buffers would be viable?

(unrelated, but I thought I remembered you from somewhere; turns out we both worked under P-E at U of U)

tangxifan: VPR has an option to wipe out buffers. In our toolchain, we need to selectively wipe out buffers, which is different than VPR's feature. Either I choose to insert-then-remove or inspect-and-insert, a custom pass in the source file is needed. I prefer the inspect-and-insert, which is easy and straightforward to implement. 

Yes. I remember you were contributing to LSOracle.

tangxifan:
A todo list so far:

- Add timing to cells for abc9
 - Use abc9 for flops without async reset/set
 - Use latest memlib map commands
 - Add reg tests

Ravenslofty:
Ah, this reminds me: flops that are marked `(* abc9_flop *)` must initialise to zero; I think you have `dfflegalize` set to use `01` for zero or one.

I should probably also confirm: is `01` accurate to the hardware (i.e. you can have flops that initialise to zero but can be asynchronously set to one)?

tangxifan:
> Ah, this reminds me: flops that are marked `(* abc9_flop *)` must initialise to zero; I think you have `dfflegalize` set to use `01` for zero or one.
> 
> I should probably also confirm: is `01` accurate to the hardware (i.e. you can have flops that initialise to zero but can be asynchronously set to one)?

@Ravenslofty RapidFlex's hardware will initialize each flops to 0 or 1 upon users's needs. Once user selects the initial value of the flop, the asynchronous function is fixed to reset (0) or set (1).

Ravenslofty:
That's a bit confusingly explained. Is it possible to express a flop that initialises to zero but has an asynchronous set? Because that answer sounds like "no".

tangxifan:
> That's a bit confusingly explained. Is it possible to express a flop that initialises to zero but has an asynchronous set? Because that answer sounds like "no".

@Ravenslofty We do not have such feature for FFs in hardware. 

- If it is zero, it is initialized to zero and could be asynchronously set to zero. 
- If it is one, it is initialized to one and could be asynchronously set to one.

Ravenslofty:
Okay, in that case you need to specify `r` instead of `01` for "init equals reset" behaviour in dfflegalize.

techlibs/rapidflex/src/clock_buffer_cmd.cc
treelin611: `clkbufmap` is a good fit for the clock-buffer insertion part of our flow.  However, our current implementation handles clock and reset together, so adopting` clkbufmap` would require splitting that logic into separate clock and reset paths.  In addition, `clkbufmap` does not cover the LUT truth-table rewriting logic needed to preserve behavior after internal signal rewiring. Most of our complexity and long-term maintenance cost is in that LUT rewriting path, not in basic clock-buffer insertion. Therefore, even if we introduce `clkbufmap` locally for the clock-only part, the practical benefit is limited: it removes only a small and relatively simple section, while the core custom logic and maintenance burden remain largely unchanged.

techlibs/rapidflex/alkaidL/synth.ys
```diff
@@ -0,0 +1,128 @@
+# Yosys synthesis script for ${TOP_MODULE}
```
Ravenslofty: Don't forget to remove these Yosys script files too.

techlibs/rapidflex/alkaidL/cell_sim_ff.v
```diff
@@ -0,0 +1,586 @@
+//-----------------------------
```
Ravenslofty, line 1: This file seems identical to the one in alkaidC; would it be feasible to have a common `alkaid` directory and then have family-specific differences in their own directory?

techlibs/rapidflex/src/clock_buffer_cmd.cc
Ravenslofty: Okay, fine, but in that case I would like this pass renamed to something with a `rapidflex_` prefix, because `insert_clock_buffer` makes this pass seem more generic than it actually is and would cause confusion with `clkbufmap`.

techlibs/rapidflex/alkaidC/cell_sim_ff.v
```diff
@@ -0,0 +1,586 @@
...
+module dffr(
+    output reg Q,
+    input D,
+    input R,
+    (* clkbuf_sink *)
+    (* invertible_pin = "IS_C_INVERTED" *)
+    input C
+);
+    parameter [0:0] INIT = 1'b0;
```
Ravenslofty, line 37: Knowing that the hardware flops initialise to their reset value, I think this parameter should be removed and the below `initial Q = INIT;` line changed to `initial Q = 1'b0;`, because `INIT = 1` is not expressible.

Ravenslofty (changes requested):
This is definitely progress, and I think once more of the redundant files across architectures is resolved, this will be much easier to review.

techlibs/rapidflex/src/clock_buffer_cmd.cc
treelin611: Thanks for your comment. I’ve renamed the internal pass to` rapidflex_insert_clock_buffer` to avoid confusion with `clkbufmap`, while keeping the synth_rapidflex -insert_clock_buffer flow unchanged.

Ravenslofty:
Hi, could I have a progress update on this?

tangxifan:
> Hi, could I have a progress update on this?

@Ravenslofty We are on an high-priroty project these days. We are still fully motivated to complete this PR. We will continue in July 2026. Thank you for the heads up.

Issue #5871, opened 2026-05-11
always_latch incorrectly tries to latch address of an array


kinthall:
### Version

Yosys 0.64 (git sha1 d8dab5b32666564eca8e18f412973853ce006e61, g++ 15.2.1_p20260214 -fPIC -O3)

### On which OS did this happen?

Linux

### Reproduction Steps

Hello, I'm having an issue with using always latch on SystemVerilog similar to the following

```
module examplename(input[3:0] addr, input we, input[31:0] data);
logic[31:0] register[0:15];
always_latch
    if (we)
         register[addr] = data;
endmodule
```

It tries to treat addr as a latch as well, the only workaround is using a case statement with every variation of addr and statically addressing via the case.

The error is the following, after running `proc`

```
ERROR: No latch inferred for signal `\registers.$mem2reg_rd$\regs$registers.sv:25$1_ADDR' from always_latch process `\registers.$proc$registers.sv:4$5'.
```

### Expected Behavior

Correct processing of Verilog always_latch.

### Actual Behavior

```
ERROR: No latch inferred for signal `\registers.$mem2reg_rd$\regs$registers.sv:25$1_ADDR' from always_latch process `\registers.$proc$registers.sv:4$5'
```

PR #5907, opened 2026-05-25
proc: ignore nosync temporaries in always_latch checks

JesseDebuger:
fix this bug.https://github.com/YosysHQ/yosys/issues/5871
proc: ignore nosync temporaries in always_latch checks

mem2reg can generate nosync address/data temporaries for dynamic memory
writes inside always_latch blocks. These temporaries are combinational
implementation details and should not trigger the always_latch "no latch
inferred" error.

Add a SystemVerilog regression for dynamic memory writes in always_latch.


widlarizer (approved):
Looks good, thanks!

PR #5914, opened 2026-05-28
gatemate: add option to create 'scopename' attributes when flattening netlist

pu-cc:
Add `-scopename` option to `synth_gatemate` to pass the `-scopename` flag to the `flatten` pass, enabling generation of scopename attributes when flattening the netlist.

This minor change is required to (conveniently) support the floorplanning feature in nextpnr.

### DIFF
--- a/techlibs/gatemate/synth_gatemate.cc
+++ b/techlibs/gatemate/synth_gatemate.cc
@@ -56,6 +56,9 @@ struct SynthGateMatePass : public ScriptPass
 		log("    -noflatten\n");
 		log("        do not flatten design before synthesis.\n");
 		log("\n");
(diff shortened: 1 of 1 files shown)


gatecat:
nextpnr will happily ignore these attributes if you aren't using floorplanning so maybe we could just always enable this rather than having another option?

mmicko:
@gatecat we have added some support for it for gatemate specifically, and were using this approach in past to provide info to it, so it is arch (well uarch) specific

gatecat:
Oh that makes sense - I didn't mean always enabling it inside flatten, rather, always passing it to `flatten` in `synth_gatemate`

mmicko (approved):

PR #5918, opened 2026-05-29
verilog backend: runtime optimization for keyword pool

mikesinouye:
Runtime optimization, on some of our designs copying the keyword pool took about 5% of overall Yosys time (12 minutes on 4hr run).

Although `VERILOG_BACKEND::verilog_keywords()`'s keyword pool is defined as static it creates a copy of it at return.

Profile diff: At our sampling rate, `VERILOG_BACKEND::verilog_keywords()` disappears from the profile entirely (100% relative speedup)
image


widlarizer (approved):
Nice catch, thanks!

PR #5919, opened 2026-05-30
[codex] verilog: warn for empty always @* sensitivity

### LINKED ISSUE #5916: Empty `always @*` constant assignment changes RTL/netlist behavior through inversion
### Version

Yosys 0.63+173 (git sha1 66306a8ca, g++ 13.3.0-6ubuntu2~24.04.1 -O0 -fPIC -O3)

Linux

### Reproduction Steps

Save as `top.v`:

```verilog
module top(input c, output reg o);
  reg r;
  always @* r = 0;
  always @(c)
    if (~r)
      ;
    else
      o = 0;
endmodule
```

Save as `tb.v`:

```verilog
`timescale 1ns/1ps

module tb;
  reg c;
  wire o_rtl;
  wire o_synth;

  top U_RTL(.c(c), .o(o_rtl));
  top_synth U_SYNTH(.c(c), .o(o_synth));

  initial begin
    c = 1'b0;
    #1;
    $display("after driving c: c=%b original_o=%b synthesized_o=%b",
             c, o_rtl, o_synth);
    if (((o_rtl === 1'b0) || (o_rtl === 1'b1)) &&
        !((o_synth === 1'b0) || (o_synth === 1'b1)))
      $display("SYNTH_UNKNOWN_FOR_KNOWN_RTL");
    $finish;
  end
endmodule
```

Run:

```sh
yosys -p "read_verilog -sv top.v; hierarchy -check -top top; synth; rename top top_synth; write_verilog synth.v"
iverilog -g2012 -o sim.vvp tb.v top.v synth.v
vvp sim.vvp
```

## 

### Expected Behavior

Yosys should produce a netlist whose simulation behavior matches the RTL, or emit a diagnostic for an `always @*` block with an empty inferred sensitivity list.

Icarus Verilog warns:

```text
top.v:3: warning: @* found no sensitivities so it will never trigger.
```

## 

### Actual Behavior

The RTL output is known zero, but the synthesized netlist output is unknown:

```text
after driving c: c=0 original_o=0 synthesized_o=x
SYNTH_UNKNOWN_FOR_KNOWN_RTL
```

Yosys emits no equivalent empty-sensitivity warning and generates:

```verilog
module top_synth(c, o);
  input c;
  wire c;
  output o;
  wire o;
  wire r;
  assign r = 1'h0;
  assign o = 1'hx;
endmodule
```

Yosys log excerpt:

```text
Note: Assuming pure combinatorial block at top.v:4.3-8.13 in
compliance with IEC 62142(E):2005 / IEEE Std. 1364.1(E):2002. Recommending
use of @* instead of @(...) for better match of synthesis and simulation.
...
Latch inferred for signal `\top.\o' from process `\top.$proc$top.v:4$2'
No latch inferred for signal `\top.\r' from process `\top.$proc$top.v:3$1'
```

## Control Case

Changing only `if (~r)` to `if (r)` removes the mismatch:

```verilog
module top(input c, output reg o);
  reg r;
  always @* r = 0;
  always @(c)
    if (r)
      ;
    else
      o = 0;
endmodule
```

Result:

```text
after driving c: c=0 original_o=0 synthesized_o=0
```

Yosys generates:

```verilog
assign r = 1'h0;
assign o = 1'h0;
```

## Analysis

`always @* r = 0;` is syntactically valid, but it has an empty inferred sensitivity list.

In RTL simulation:

```text
r = x
~r = x
if (~r) goes to else
o = 0
```

In Yosys synthesis:

```text
r = 0
~r = 1
if (~r) takes the empty true branch
else is unreachable
o = x
```

Without the inversion:

```text
RTL:   r = x, if (r) goes to else, o = 0
Yosys: r = 0, if (r) goes to else, o = 0
```

So the root issue is the empty-sensitivity `always @*` constant assignment. The inversion `~r` makes the RTL/synth interpretation difference observable at output `o`.

Yosys' own equivalence flow does not catch this, because after `proc; opt` both sides are interpreted with Yosys synthesis semantics:

```text
Presumably equivalent wires: r_gold (1'0), r_gate (1'0) -> r
Presumably equivalent wires: o_gold (1'x), o_gate (1'x) -> o
Equivalence successfully proven!
```
We carefully checked and analyzed this bug case. Hope this bug report helps.

JesseDebuger:
Fixes #5916.


widlarizer:
Entirely wrong approach as far as I can tell. Read the [contributing guidelines](https://yosyshq.readthedocs.io/projects/yosys/en/latest/yosys_internals/extending_yosys/contributing.html)

widlarizer:
I'm starting to suspect the reason you didn't explain anything is not because you completely ignored the PR template but because your Codex automation doesn't know it should read it and submits PRs with some github CLI interface. Is that the case or did you ignore it willingly?

PR #5924, opened 2026-06-01
cmake: accomodate wheel-only builds

donn:

Currently, builds in the manylinux environment fail as `Development.Embed` (i.e. embedding the Python interpreter)
is not available.

Adds new option, YOSYS_WITH_PYTHON_INTERPRETER, and corresponding feature, YOSYS_ENABLE_PYTHON_INTERPRETER. Usually, YOSYS_WITH_PYTHON_INTERPRETER is true if and only if YOSYS_WITH_PYTHON is true.
However, the interpreter cannot be included in wheel-only environments (e.g. manylinux). This options allows it to be forced off in said environments, with the stipulation that only libyosys/pyosys will build successfully. It's only done this way because making the top-level yosys executable a conditional target would complicate the CMake files significantly.

`python3 ./.github/workflows/wheels/_run_cibw_linux.py` on a Linux machine with Docker installed.

---

This is a strawman/kind of jank but this is the best I could come up with without altering the CMake infrastructure too much.  The alternatives are:

- Make `yosys` a conditional target, disabling it if YOSYS_ENABLE_PYTHON is ON but YOSYS_ENABLE_PYTHON_INTERPRETER is OFF. I feel like this adds too much complexity for such a niche use-case.
- Leave the CMake infrastructure as-is and `sed` Development.Embed for Development.Module before wheel builds. This would be considerably more brittle.


donn:
Forgot to mention a crucial detail: the reason the yosys executable would not build successfully is that without libpython linked, all of the Python symbols referenced by the object files because of Development::Module go unresolved…

whitequark:
Thank you for doing this exploratory work. After looking at it, I concluded that making the binary/library targets conditional is more robust and less invasive than this approach, and implemented it in 87835c50e (I also pulled your CI related changes). Please let me know what you think and if I've missed anything.

donn:
Will check!

PR #5927, opened 2026-06-03
Update docs for CMake

KrystalDelusion:
- Update install instructions
- Update test instructions
- Update `register.cc` to make source paths relative to the root source directory where possible (thereby fixing the groups mismatch in `cmds.json`).
- Strip `$(BUILD_DIR)` from tool help text (with the exception of `yosys-filterlib`, this should also work with suffixes, though I'm not convinced that the wrapper is setting them)
- [Read the Docs preview build](https://yosyshq.readthedocs.io/projects/yosys/en/docs-preview-cmake/getting_started/installation.html#build-prerequisites)

pending TODOs (I don't think any of these need to block a merge though):
- Update instructions for adding new passes
- running the functional tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,26 +102,28 @@ Once generated, available build variables can be inspected and modified with
 	$ ccmake build              #..or..
 	$ vi build/CMakeCache.txt
 
-When setting one-off variables, CMake provides the `-D <var>=<value>` command line
-option. For example, disabling zlib support:
+When setting one-off variables, CMake provides the `-D <var>=<value>` command
+line option. For example, disabling zlib support:
 
 	$ cmake -B build . -DYOSYS_WITHOUT_ZLIB=ON
 
 For a more persistent configuration, we recommend creating and using a
-`CMakeUserPresets.json` file. Below is an example file which sets the default
-compiler to clang when calling `cmake --preset clang`:
+`CMakeUserPresets.json` file in the root `yosys` directory. Below is an example
+file which enables ccache and sets the default compiler to clang when calling
+`cmake --preset clang`:
 
 ```json
 {
 	"version": 1,
 	"configurePresets": [
 		{
-			"name": "clang",
+			"name": "default",
 			"binaryDir": "build",
 			"generator": "Unix Makefiles",
 			"cacheVariables": {
 				"CMAKE_C_COMPILER": "clang",
-				"CMAKE_CXX_COMPILER": "clang++"
+				"CMAKE_CXX_COMPILER": "clang++",
+				"YOSYS_COMPILER_LAUNCHER": "ccache"
 			}
 		}
 	]
--- a/README.md
+++ b/README.md
@@ -75,10 +75,11 @@ or
 	$ cd yosys
 	$ git submodule update --init --recursive
 
-You need a C++ compiler with C++17 support (up-to-date CLANG or GCC is
-recommended) and some standard tools such as GNU Flex, GNU Bison, and GNU Make.
-TCL, readline and libffi are optional (see ``ENABLE_*`` settings in Makefile).
-Xdot (graphviz) is used by the ``show`` command in yosys to display schematics.
+A C++ compiler with C++20 support is required as well as some standard tools
+such as GNU Flex, GNU Bison (>=3.8), CMake (>=3.27), Make, and Python (>=3.11).
+Some additional tools: readline, libffi, Tcl and zlib; will be used if available
+but are optional. Graphviz and Xdot are used by the `show` command to display
+schematics.
 
 For example on Ubuntu Linux 22.04 LTS the following commands will install all
 prerequisites for building yosys:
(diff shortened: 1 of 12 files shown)


whitequark:
> * Finalize Pyosys instructions (not sure if the changes for wheels impact the user side of things)

I don't think so or at least I worked hard to make sure this wouldn't be needed. A simple `pip install git+https://github.com/YosysHQ/yosys.git` should do it.

> Work out if we can export the auto-detected YOSYS_VERIFIC_FEATURES (they get printed during build config, but only the empty string gets kept in the cache)

Yes however I left all of the "export build features as CMake files" type tasks for later since this PR is already massive. I'm quite happy to address this later under the same terms and diligence as the main PR, as there is a lot of nuance there and I'll have to study a few things in detail plus test it with downstream consumers (GHDL, Slang mainly).

cmake/YosysAbcSubmodule.cmake
```diff
@@ -56,7 +56,7 @@ function(yosys_check_abc_submodule)
 	else() #
 		message(FATAL_ERROR
 			"${CMAKE_SOURCE_DIR} is not configured as a git repository, and 'abc' folder is missing.\n"
-			"If you already have ABC, set 'ABCEXTERNAL' make variable to point to ABC executable.\n"
+			"If you already have ABC, set 'YOSYS_ABC_EXECUTABLE' make variable to point to ABC executable.\n"
```
whitequark: CMake variable perhaps?

docs/source/getting_started/installation.rst
```diff
@@ -127,22 +135,26 @@ Installing all prerequisites:
    .. note:: On FreeBSD system use gmake instead of make. To run tests use:
       ``MAKE=gmake CXX=cxx CC=cc gmake test``
 
+   .. TODO:: CMAKE_TODO
+
+      Is this still required, and (how) does it work with CMake
```
whitequark: You can use `cmake --build <builddir>` regardless of OS or generator, as that will always work.

docs/source/getting_started/installation.rst
```diff
@@ -289,19 +304,28 @@ directories:
...
+
+   - ``yosys_<pass|test_pass|frontend|backend>(<name>)`` for each pass
+
+      - see :file:`cmake/YosysComponent.cmake`
+
+   - if using a sub folder, add it to the parent's ``CMakeLists.txt`` with
+     ``add_subdirectory(<name>)``
+
+   The top-level Makefile includes :file:`frontends/{*}/Makefile.inc`,
+   :file:`passes/{*}/Makefile.inc` and :file:`backends/{*}/Makefile.inc`. So when
```
whitequark: With cmake we don't use globs, so each new backend needs to be declared explicitly. There is a way to use globs where you don't end up with missing dependencies (we do that for ABC for example where we don't know which files are added) but it's costly since they have to be re-evaluated on each rebuild, even if nothing is actually changed.

docs/source/using_yosys/pyosys.rst
```diff
@@ -14,27 +16,35 @@ Though unlike these two, Pyosys goes a bit further, allowing you to use the
...
 
    Do note this requires some build-time dependencies to be available to Python,
-   namely, ``pybind11`` and ``cxxheaderparser``. By default, the required
-   ``uv`` package will be used to create an ephemeral environment with the
-   correct versions of the tools installed.
-
-   You can force use of your current Python environment by passing the Makefile
-   flag ``PYOSYS_USE_UV=0``.
+   namely, ``pybind11`` and ``cxxheaderparser``. If available, ``uv`` will be
+   used to create an ephemeral environment with the correct versions of the
```
whitequark: `uv` is only used if a suitable environment is not found otherwise

docs/source/yosys_internals/extending_yosys/build_verific.rst
```diff
@@ -39,36 +39,36 @@ incorrect results.
...
+extensions     extensions  YosysHQ specific extensions support
+edif           edif        EDIF support
+liberty        synlib      Liberty file support
+============== =========== ===================================
+
+.. TODO:: CMAKE_TODO
+
+   ``yosys-config --cxxflags`` no longer includes the verific features, and the
+   CMakeCache.txt doesn't report auto detected :makevar:`YOSYS_VERIFIC_FEATURES`
+   - can we export these somehow?
```
whitequark: We can add them to `yosys-config --cxxflags` if need be, yes

whitequark: Micko says nobody probably relies on this so I'm not going to rush it

docs/source/yosys_internals/extending_yosys/test_suites.rst
```diff
@@ -7,19 +7,32 @@ Running the included test suite
...
-afterwards when make is invoked do the tests actually run.
+.. TODO:: update for test infra changes
+
+These make up the majority of our testing coverage. They can be run with the
+``test-vanilla`` CMake target, or by calling ``make vanilla-test`` from the
+``tests`` directory, and are based on calls to make subcommands (``make
+makefile-tests``) and shell scripts (``make seed-tests`` and ``make
+abcopt-tests``). Both use ``run-test.sh`` files, but make-based tests only call
+``tests/gen-tests-makefile.sh`` to generate a makefile appropriate for the given
+directory, so only afterwards when make is invoked do the tests actually run.
```
whitequark: I think we should probably recommend `cmake --build <builddir> -t test-vanilla` since this automatically configures all the right environment variables. Leave `make` an implementation detail.

kernel/register.cc
```diff
@@ -29,6 +29,11 @@
 #include <stdio.h>
 #include <errno.h>
 
+#if !defined(__wasm)
+#include <filesystem>
+namespace fs = std::filesystem;
+#endif
```
whitequark: I think std::filesystem should be available on Wasm now, does it still fail?

docs/Makefile
```diff
@@ -265,15 +265,15 @@ extract:
 
 	@mkdir -p source/generated/functional
 	@cp ../backends/functional/smtlib.cc source/generated/functional/
-	-@cd .. && diff -U 20 backends/functional/smtlib.cc backends/functional/smtlib_rosette.cc \
+	@cd .. && diff -U 20 backends/functional/smtlib.cc backends/functional/smtlib_rosette.cc \
 		> docs/source/generated/functional/rosette.diff || true
 
-	@$(YOSYS) --help >source/generated/yosys
+	$(YOSYS) --help | sed "s%$(BUILD_DIR)/yosys%yosys%g" - >source/generated/yosys
```
whitequark: Perhaps use `sed s%$(YOSYS)%yosys%g`?

README.md
```diff
@@ -96,31 +96,45 @@ CMake is used for build configuration, and requires a separate build directory:
...
-	$ cmake -G Ninja -B build .
-
-INSTALL
+When setting one-off variables, CMake provides the `-D <var>=<value>` command line
+option. For example, disabling zlib support:
+
+	$ cmake -B build . -DYOSYS_WITHOUT_ZLIB=ON
+
+For a more persistent configuration, we recommend creating and using a
+`CMakeUserPresets.json` file. Below is an example file which sets the default
```
whitequark: Nice, didn't know about this one

docs/Makefile
whitequark: Ah you're doing that already.

docs/source/yosys_internals/extending_yosys/test_suites.rst
KrystalDelusion: I didn't change this one yet because it was already out of date.  @mmicko are you able to update this for the changes you made to the test infrastructure or do you want me to do it

mmicko: @KrystalDelusion feel free to do it in documentation. Guess best is to rebase this PR on main when merge is done, will be much easier to review then

docs/source/using_yosys/pyosys.rst
KrystalDelusion: Got it, thanks!

docs/source/yosys_internals/extending_yosys/build_verific.rst
KrystalDelusion: Yeah I don't think it's a priority, in the two years or so that it's been available I've only encountered one time where I thought it might be relevant (asking someone who reported a git issue which verific features they had enabled, though in the end it actually didn't matter)

kernel/register.cc
KrystalDelusion: Oh yeah thanks, I based it on `verilog_frontend.cc` which still has `<filesystem>` guarded, but `io.cc` doesn't.

mmicko (approved):
Let's have this in ASAP and work on TODOs

PR #5928, opened 2026-06-03
CMake: macOS platform fixes

donn:
- Apply fixes from @mmicko
  - for Darwin modules using incorrect suffix
    - fixes macOS test failure 
  - for GTest not liking python
    - fixes Verific linux build 
- pyosys/build/local_backend.py:
  - use ABI tag to handle freethreaded (`t`) builds (`none` was incorrect also)
    - fixes Ubuntu build failure
  - use substitution logic similar to pypa/wheels for platform names
    - fixes macOS relocation failure
  - add some logic to handle the macOS minimum deployment target
    - just labels the wheels correctly according to the version of macOS more or less
- misc: reconfigure .github/workflows/wheels/_run_cibw_linux.py as .github/workflows/wheels/_run_cibw_local.py
  - allows local testing on macOS as well 
  - use tabs instead of spaces

---

CI:
https://github.com/donn/yosys/actions/runs/26869193075


whitequark:
Integrated in #5895.

PR #5929, opened 2026-06-03
Add register absorption for MULT* for Nexus (fix #5917).

### LINKED ISSUE #5917: Follow up to #5906 Nexus DSP techmap is missing MULT* register absorption
As a follow up to https://github.com/YosysHQ/yosys/issues/5906

I was looking at the synthesis results for some compute pipelines targeting Lattice Nexus, and I noticed that we aren’t fully utilizing the hardware capabilities of the DSP blocks. Specifically, `synth_nexus` successfully maps to the `MULT18X18` and `MULT36X36` primitives, but it completely ignores the internal pipeline registers, defaulting them all to bypass.

If you write a properly pipelined multiplier in RTL, the current mapper leaves the input and output registers out in the general FPGA fabric. You end up paying the routing penalty to go from fabric FFs into the DSP slice and back out, which is exactly what the hardened internal registers are designed to avoid.

Here is a minimal reproducer:

```systemverilog
module nx_mul18_pipe(
    input  logic clk,
    input  logic [17:0] a,
    input  logic [17:0] b,
    output logic [35:0] y
);
    logic [17:0] a_r;
    logic [17:0] b_r;

        a_r <= a;
        b_r <= b;
        y <= a_r * b_r;
    end
endmodule

```

Synthesizing this with:

```sh
yosys -p 'read_verilog -sv nx_mul18_pipe.sv; synth_nexus -family lifcl -top nx_mul18_pipe -json nx_mul18_pipe.json'

```

If you dig into the generated JSON, the `MULT18X18` cell is there, but the register configurations are pessimistic:

```json
"REGINPUTA": "BYPASS",
"REGINPUTB": "BYPASS",
"REGOUTPUT": "BYPASS"

```

The surrounding pipeline registers just remain separate flip-flops.

I ran some local timing measurements on a `LIFCL-40-9BG400C` to quantify the real-world impact. Adding the explicit pipeline registers in the RTL *does* help—FMAX scales up reasonably well:

```text
SynthMul18x18U36:      DSP=1 FMAX=160.5 MHz
SynthMul18x18U36Pipe:  DSP=1 FMAX=209.5 MHz

```

However, looking at the mapped netlist for the pipelined version, we are still just paying for fabric registers:

```text
MULT18X18 REGINPUTA=BYPASS REGINPUTB=BYPASS REGOUTPUT=BYPASS

```

So the RTL pipeline helps timing by breaking up the path, but the DSP block's native pipeline is sitting completely idle. We are burning routing resources we don't need to burn.

The right thing to do here is to teach the Nexus DSP mapping pass to look for these `$dff -> $mul -> $dff` subgraphs and pull those adjacent flip-flops into the primitive.

Ideally, when the control sets match, Yosys should swallow the fabric FFs and emit:

```json
"REGINPUTA": "REGISTER",
"REGINPUTB": "REGISTER",
"REGOUTPUT": "REGISTER"

```

To do this robustly, we’ll need test cases to cover the functional matrix:

* Standard unsigned 18x18 with both input and output registered.
* Signed 18x18 variants.
* Wide multipliers (24x24 / 32x32) mapping cleanly to the `MULT36X36` block.
* The crucial rejection cases—falling back to the fabric FFs when the clock, reset, or enable semantics aren't strictly compatible with what the Nexus DSP registers actually support.

It’s a relatively straightforward structural matching problem, but absorbing these into the DSP slice should yield a nice, free FMAX bump and reduce routing congestion for anyone doing heavy math on these parts.

hexratcc:
This PR aims to fix #5917 - MULT* DSP blocks now attempt to utilize the internal DSP registers, if applicable.


agentdavo:
@Dolu1990 checking this now with https://github.com/SpinalHDL/VexiiRiscv and will report on DSP & xmax.

agentdavo:
@hexratcc two bugs, both produce a netlist that looks correct in `stat` and `show` and computes the wrong number.

I found them the same way: synthesize the design, then prove the result equal to the behavioral source with a bounded SAT miter against the `cells_sim_nexus.v` cell models.

```
read_verilog dut.v ; (gold)  proc
synth_nexus -noiopad ; (gate) read_verilog +/lattice/cells_sim_nexus.v ; flatten
miter -equiv -flatten gold gate m ; sat -seq N -set-init-zero -prove-asserts -verify
```

The `-set-init-zero` matters: without it the miter free-runs from an arbitrary initial state and the retimed registers diverge, so you get a failure that tells you nothing. A pure-fabric (`-nodsp`) run of the same design proves equal, which is the sanity check that the harness itself is sound.

A note on width: bit-level equivalence of two structurally different multiplier pipelines is SAT-hard, so the proofs below run at 9 bit operands. That still fuses to the same `MULT18X18` / `MULTADDSUB18X18` cells and exercises the identical control wiring; only the multiplier array is narrower. The passing cases are real UNSAT. The failing cases come back as a counterexample in seconds.

The fused MAC ignores its C input

The `a*b+c` fusion (`MULTADDSUB18X18`) does not compute `a*b+c`. 

With everything combinational and all registers bypassed, 

`a=1, b=0, c=0` comes out `0x800000` instead of `0`. 

Signed and unsigned both.

The accumulator inside the cell is

```
z = (LOADC ? C : Z) + CIN + (sub ? -(A*B) : A*B)
```

`LOADC` picks between loading the C operand and accumulating the cell's own output Z. 

A plain MAC is the load case — you want `z = C + A*B` every cycle. 

The techmap macro never connected `LOADC`, so it defaulted to 0, which is the accumulate case: Z feeds back into the adder. With `REGOUTPUT` bypassed there is no register in that path, so it is a combinational cycle. The solver is free to pick any consistent value around the loop, and `0x800000` is one of them. There is nothing special about that number; it is just what the equations allowed.

There is a second half to it. `MULTADDSUB18X18`'s control and pipeline registers

`REGADDSUB`, `REGLOADC`, `REGLOADC2`, `REGCIN`, `REGPIPELINE` 

— all default to `"REGISTER"`, and the macro only ever set the data registers. So even after you fix `LOADC`, the control bits and the product (which passes through `REGPIPELINE`) sit behind registers the macro never clocks. On a combinational MAC there is no clock, so they hold their reset value and the product never reaches the adder. They have to be `BYPASS` so the control follows the data instead of lagging it or freezing.

This is the usual way these cells go wrong. The primitive has a dozen mode pins; the fusion drove the obvious handful and left the rest at defaults that happen to mean "accumulate" and "register" rather than "load" and "pass through". It synthesizes, it reports one DSP and no logic, and it is wrong.

`techlibs/lattice/dsp_map_nexus.v`, the `$__NX_MAC18X18` techmap macro:

```diff
 	MULTADDSUB18X18 #(
 		.REGINPUTA("BYPASS"),
 		.REGINPUTB("BYPASS"),
 		.REGINPUTC("BYPASS"),
+		.REGADDSUB("BYPASS"),
+		.REGLOADC("BYPASS"),
+		.REGLOADC2("BYPASS"),
+		.REGCIN("BYPASS"),
+		.REGPIPELINE("BYPASS"),
 		.REGOUTPUT("BYPASS")
 	) _TECHMAP_REPLACE_ (
 		.A(A),
 		.B(B),
 		.C({6'b0, C}),
 		.SIGNED(A_SIGNED ? 1'b1 : 1'b0),
 		.ADDSUB(SUBTRACT ? 1'b1 : 1'b0),
+		.LOADC(1'b1),
+		.CIN(1'b0),
 		.Z(Y)
 	);
```

With `LOADC=1`, `CIN=0`, and the control/pipeline registers bypassed, the fused MAC matches behavioral `a*b+c` under the miter, signed and unsigned.

The register-folding pass feeds the DSP a deleted flip-flop's output

A pipelined multiply `reg -> a*b -> reg` folds into a `MULT18X18` with its input and output registers turned on and zero fabric flip-flops left over. That part is the point of the pass and it looks done. The multiply is still wrong.

The pass pulls the flip-flops into the MULT's hardened registers. For the inputs it set the new cell's A/B to the multiplier's A/B nets — which, by the match that found them, are the input flip-flops' Q nets — and then removed those flip-flops:

```cpp
SigSpec sigA = port(mul, \A);   // == ffA's Q net
SigSpec sigB = port(mul, \B);   // == ffB's Q net
...
if (ffA_use) autoremove(ffA_use);
```

pmgen's `autoremove` is a plain cell delete. It does not splice D through to Q. So once the flip-flop is gone its Q net has no driver, and the DSP — which now holds that pipeline register internally — clocks in a dangling signal.

It looks right because the flip-flop really is gone and the register count really is zero. The only way to see it is a `dump`: the input reads `\a_r` (the dead flip-flop's net, which `opt` then merges with other undriven nets into nonsense) instead of `\a`.

The DSP's input register is meant to *be* the flip-flop. So it has to sample what the flip-flop sampled — the flip-flop's D, the pre-register value — and clock it itself. Feed it the D and the Q net loses its last reader, so `autoremove` can delete the flip-flop cleanly:

`techlibs/lattice/lattice_dsp_nexus.pmg`, `nexus_mul_reg`:

```diff
-			SigSpec sigA = port(mul, \A);
-			SigSpec sigB = port(mul, \B);
+			// Drive the DSP from the FF's D, not its Q: the FF is about to be
+			// removed, so its Q net loses its driver. The DSP re-registers the
+			// pre-register value via REGINPUT*, reproducing the FF.
+			SigSpec sigA = fa ? port(fa, \D) : port(mul, \A);
+			SigSpec sigB = fb ? port(fb, \D) : port(mul, \B);
 			SigSpec sigY = fy ? port(fy, \Q) : port(mul, \Y);
```

(`sigY` was already taking the output flip-flop's Q, which is correct on that side — the output register drives the post-flip-flop net.) After the change the input reads `\a`, and the pipelined multiply proves equal to the behavioural source. It does not before.

agentdavo:
@gatecat @hexratcc I am in the middle of testing on actual hardware, will report back too.

hexratcc:
@agentdavo Thanks for the feedback, looking into it.

hexratcc:
The latest commits should remedy the above issues @agentdavo.

agentdavo:
Looking positive! Thanks @hexratcc for the effort made here this week.

agentdavo:
@hexratcc 

Can you take a look at this if you have the time ?
Or is this another issue?
Making good progress with AI compute and inferred yosys!

```

From 0a51ca525de198399bc46ce092240aaeb4ce9541 Mon Sep 17 00:00:00 2001
From: agentdavo <44442387+agentdavo@users.noreply.github.com>
Date: Mon, 29 Jun 2026 13:53:55 +0100
Subject: [PATCH] Infer two more Nexus DSP MAC shapes

The DSP block does the multiply and the add in one cell, and nextpnr can
already place every variant, but the inference pass only matched a few of
them -- so a multiply went into a DSP and the add behind it fell into a
fabric carry chain. Two more shapes the silicon gives you for free:

  - nexus_mac9_4lane_c: a 4-lane 9x9 dot with a trailing accumulate
    (dot4 + C) rides the MULTADDSUB9X9WIDE hardened ACC54 C input instead
    of a CCU2 reduction tree.
  - nexus_mac36: a wide multiply-add (A,B <=36, sum <=108) becomes
    MULTADDSUB36X36 instead of MULT36X36 plus a 33-cell ripple carry --
    the shape of an address calculation, base + index*stride.

Both run after the existing narrow patterns so the cheap mappings still
win, and both are SAT-equivalent to the source; the 18x18 and dot-product
patterns are unchanged.

---
 techlibs/lattice/dsp_map_nexus.v       |  37 ++++-
 techlibs/lattice/lattice_dsp_nexus.cc  |   2 +
 techlibs/lattice/lattice_dsp_nexus.pmg | 183 +++++++++++++++++++++++++
 3 files changed, 220 insertions(+), 2 deletions(-)

diff --git a/techlibs/lattice/dsp_map_nexus.v b/techlibs/lattice/dsp_map_nexus.v
index 0489345..16eca74 100644
--- a/techlibs/lattice/dsp_map_nexus.v
+++ b/techlibs/lattice/dsp_map_nexus.v
@@ -137,6 +137,38 @@ module \$__NX_MAC18X18 (input [17:0] A, input [17:0] B, input [47:0] C, output [
 	);
 endmodule
 
+module \$__NX_MAC36X36 (input [35:0] A, input [35:0] B, input [107:0] C, output [107:0] Y);
+
+	parameter A_WIDTH = 36;
+	parameter B_WIDTH = 36;
+	parameter C_WIDTH = 108;
+	parameter Y_WIDTH = 108;
+	parameter A_SIGNED = 0;
+	parameter B_SIGNED = 0;
+	parameter SUBTRACT = 0;
+
+	MULTADDSUB36X36 #(
+		.REGINPUTA("BYPASS"),
+		.REGINPUTB("BYPASS"),
+		.REGINPUTC("BYPASS"),
+		.REGADDSUB("BYPASS"),
+		.REGLOADC("BYPASS"),
+		.REGLOADC2("BYPASS"),
+		.REGCIN("BYPASS"),
+		.REGPIPELINE("BYPASS"),
+		.REGOUTPUT("BYPASS")
+	) _TECHMAP_REPLACE_ (
+		.A(A),
+		.B(B),
+		.C(C),
+		.SIGNED(A_SIGNED ? 1'b1 : 1'b0),
+		.ADDSUB(SUBTRACT ? 1'b1 : 1'b0),
+		.LOADC(1'b1),
+		.CIN(1'b0),
+		.Z(Y)
+	);
+endmodule
+
 module \$__NX_PREADD18X18 (input [17:0] A, input [17:0] B, input [17:0] C, input CLK, output [35:0] Y);
 
 	parameter PIPELINED = 0;
@@ -161,9 +193,10 @@ module \$__NX_PREADD18X18 (input [17:0] A, input [17:0] B, input [17:0] C, input
 	);
 endmodule
 
-module \$__NX_MAC9X9WIDE_4LANE (input [8:0] A0, B0, A1, B1, A2, B2, A3, B3, output [53:0] Y);
+module \$__NX_MAC9X9WIDE_4LANE (input [8:0] A0, B0, A1, B1, A2, B2, A3, B3, input [53:0] C, output [53:0] Y);
 
 	parameter SIGNED = 0;
+	parameter USE_C = 0;
 
 	MULTADDSUB9X9WIDE #(
 		.REGINPUTAB0("BYPASS"),
@@ -177,7 +210,7 @@ module \$__NX_MAC9X9WIDE_4LANE (input [8:0] A0, B0, A1, B1, A2, B2, A3, B3, outp
 		.A1(A1), .B1(B1),
 		.A2(A2), .B2(B2),
 		.A3(A3), .B3(B3),
-		.C(54'b0),
+		.C(USE_C ? C : 54'b0),
 		.SIGNED(SIGNED ? 1'b1 : 1'b0),
 		.ADDSUB(4'b0000),
 		.Z(Y)
diff --git a/techlibs/lattice/lattice_dsp_nexus.cc b/techlibs/lattice/lattice_dsp_nexus.cc
index 8ce7127..c41450c 100644
--- a/techlibs/lattice/lattice_dsp_nexus.cc
+++ b/techlibs/lattice/lattice_dsp_nexus.cc
@@ -28,8 +28,10 @@ struct LatticeDspNexusPass : public Pass {
 		for (auto module : design->selected_modules()) {
 			lattice_dsp_nexus_pm pm(module, module->cells());
 
+			pm.run_nexus_mac9_4lane_c();
 			pm.run_nexus_mac9_4lane();
 			pm.run_nexus_mac18();
+			pm.run_nexus_mac36();
 			pm.run_nexus_preadd18();
 			pm.run_nexus_mul_reg();
 		}
diff --git a/techlibs/lattice/lattice_dsp_nexus.pmg b/techlibs/lattice/lattice_dsp_nexus.pmg
index 24b99c1..c8c8545 100644
--- a/techlibs/lattice/lattice_dsp_nexus.pmg
+++ b/techlibs/lattice/lattice_dsp_nexus.pmg
@@ -59,6 +59,67 @@ code
 	accept;
 endcode
 
+pattern nexus_mac36
+
+match mul
+	select mul->type.in($mul)
+	select GetSize(port(mul, \A)) <= 36
+	select GetSize(port(mul, \B)) <= 36
+	select GetSize(port(mul, \Y)) <= 108
+endmatch
+
+match add
+	select add->type.in($add, $sub)
+	select GetSize(port(add, \Y)) <= 108
+	choice <IdString> AB {\A, \B}
+	index <SigBit> port(add, AB)[0] === port(mul, \Y)[0]
+endmatch
+
+code
+	if (mul->getParam(\A_SIGNED).as_bool() != mul->getParam(\B_SIGNED).as_bool()) {
+		reject;
+	}
+
+	{
+		SigSpec mul_out = port(mul, \Y);
+		IdString add_AB;
+
+		if (GetSize(port(add, \A)) >= GetSize(mul_out) && port(add, \A).extract(0, GetSize(mul_out)) == mul_out) {
+			add_AB = \A;
+		} else if (GetSize(port(add, \B)) >= GetSize(mul_out) && port(add, \B).extract(0, GetSize(mul_out)) == mul_out) {
+			add_AB = \B;
+		} else {
+			reject;
+		}
+
+		if (add->type == $sub && add_AB == \A)
+			reject;
+
+		Cell *mac = module->addCell(NEW_ID, "$__NX_MAC36X36");
+		IdString add_C = (add_AB == \A) ? \B : \A;
+
+		bool ab_signed = mul->getParam(\A_SIGNED).as_bool();
+		SigSpec sigA = port(mul, \A);
+		SigSpec sigB = port(mul, \B);
+		SigSpec sigC = port(add, add_C);
+		sigA.extend_u0(36, ab_signed);
+		sigB.extend_u0(36, ab_signed);
+		sigC.extend_u0(108, ab_signed);
+
+		mac->setPort(\A, sigA);
+		mac->setPort(\B, sigB);
+		mac->setPort(\C, sigC);
+		mac->setPort(\Y, port(add, \Y));
+		mac->setParam(\A_SIGNED, mul->getParam(\A_SIGNED));
+		mac->setParam(\SUBTRACT, add->type == $sub ? State::S1 : State::S0);
+
+		autoremove(mul);
+		autoremove(add);
+	}
+
+	accept;
+endcode
+
 pattern nexus_mul_reg
 
 state <IdString> ffY_mux_ab
@@ -443,3 +504,125 @@ code
 
 	accept;
 endcode
+
+// Same four-lane 9x9 dot product as nexus_mac9_4lane, but with a trailing
+// accumulate add: (((mul0 + mul1) + mul2) + mul3) + C. The MULTADDSUB9X9WIDE
+// macro has a hard 54-bit C / accumulate input wired straight to ACC54.CINPUT
+// by nextpnr-nexus, so the +C folds into the ACC54 instead of a fabric adder.
+// Must run BEFORE nexus_mac9_4lane so the trailing add is consumed first; a
+// bare dot4 (no +C) still falls through to nexus_mac9_4lane.
+pattern nexus_mac9_4lane_c
+
+match add_top
+	select add_top->type == $add
+endmatch
+
+match add_mid
+	select add_mid->type == $add
+	index <SigBit> port(add_mid, \Y)[0] === port(add_top, \A)[0]
+endmatch
+
+match add_bot
+	select add_bot->type == $add
+	index <SigBit> port(add_bot, \Y)[0] === port(add_mid, \A)[0]
+endmatch
+
+match mul3
+	select mul3->type == $mul
+	select GetSize(port(mul3, \A)) <= 9 && GetSize(port(mul3, \B)) <= 9
+	index <SigBit> port(mul3, \Y)[0] === port(add_top, \B)[0]
+endmatch
+
+match mul2
+	select mul2->type == $mul
+	select GetSize(port(mul2, \A)) <= 9 && GetSize(port(mul2, \B)) <= 9
+	index <SigBit> port(mul2, \Y)[0] === port(add_mid, \B)[0]
+endmatch
+
+match mul1
+	select mul1->type == $mul
+	select GetSize(port(mul1, \A)) <= 9 && GetSize(port(mul1, \B)) <= 9
+	index <SigBit> port(mul1, \Y)[0] === port(add_bot, \B)[0]
+endmatch
+
+match mul0
+	select mul0->type == $mul
+	select GetSize(port(mul0, \A)) <= 9 && GetSize(port(mul0, \B)) <= 9
+	index <SigBit> port(mul0, \Y)[0] === port(add_bot, \A)[0]
+endmatch
+
+match add_acc
+	select add_acc->type == $add
+	select GetSize(port(add_acc, \Y)) <= 54
+	choice <IdString> CC {\A, \B}
+	index <SigBit> port(add_acc, CC)[0] === port(add_top, \Y)[0]
+endmatch
+
+code
+	bool is_signed = mul0->getParam(\A_SIGNED).as_bool();
+
+	if (
+		mul0->getParam(\B_SIGNED).as_bool() != is_signed ||
+		mul1->getParam(\A_SIGNED).as_bool() != is_signed ||
+		mul1->getParam(\B_SIGNED).as_bool() != is_signed ||
+		mul2->getParam(\A_SIGNED).as_bool() != is_signed ||
+		mul2->getParam(\B_SIGNED).as_bool() != is_signed ||
+		mul3->getParam(\A_SIGNED).as_bool() != is_signed ||
+		mul3->getParam(\B_SIGNED).as_bool() != is_signed
+	) {
+		reject;
+	}
+
+	{
+		// Recompute which add_acc input carries the dot4 sum at full width;
+		// the other input is the C / accumulate operand. (The bit[0] index is
+		// only a coarse join, exactly as nexus_mac18 re-derives add_AB.)
+		SigSpec top_out = port(add_top, \Y);
+		IdString acc_AB;
+
+		if (GetSize(port(add_acc, \A)) >= GetSize(top_out) && port(add_acc, \A).extract(0, GetSize(top_out)) == top_out) {
+			acc_AB = \A;
+		} else if (GetSize(port(add_acc, \B)) >= GetSize(top_out) && port(add_acc, \B).extract(0, GetSize(top_out)) == top_out) {
+			acc_AB = \B;
+		} else {
+			reject;
+		}
+
+		IdString acc_C = (acc_AB == \A) ? \B : \A;
+
+		Cell *mac = module->addCell(NEW_ID, "$__NX_MAC9X9WIDE_4LANE");
+
+		auto ext9 = [&](SigSpec s) {
+			s.extend_u0(9, is_signed);
+			return s;
+		};
+
+		mac->setPort(\A0, ext9(port(mul0, \A)));
+		mac->setPort(\B0, ext9(port(mul0, \B)));
+		mac->setPort(\A1, ext9(port(mul1, \A)));
+		mac->setPort(\B1, ext9(port(mul1, \B)));
+		mac->setPort(\A2, ext9(port(mul2, \A)));
+		mac->setPort(\B2, ext9(port(mul2, \B)));
+		mac->setPort(\A3, ext9(port(mul3, \A)));
+		mac->setPort(\B3, ext9(port(mul3, \B)));
+
+		SigSpec sigC = port(add_acc, acc_C);
+		sigC.extend_u0(54, is_signed);
+		mac->setPort(\C, sigC);
+
+		mac->setPort(\Y, port(add_acc, \Y));
+		mac->setParam(\SIGNED, is_signed ? State::S1 : State::S0);
+		mac->setParam(\USE_C, State::S1);
+
+		autoremove(add_acc);
+		autoremove(add_top);
+		autoremove(add_mid);
+		autoremove(add_bot);
+		autoremove(mul0);
+		autoremove(mul1);
+		autoremove(mul2);
+		autoremove(mul3);
+	}
+
+	accept;
+endcode
-- 
2.43.0


```

hexratcc:
@agentdavo Would you mind creating a new issue for that so that we can get the simpler version of this into Yosys first? Feel free to @ me in there and I'll take a look :wink:.

agentdavo:
Sure, will do.

techlibs/lattice/dsp_map_nexus.v
```diff
@@ -148,4 +182,4 @@ module \$__NX_MAC9X9WIDE_4LANE (input [8:0] A0, B0, A1, B1, A2, B2, A3, B3, outp
 		.ADDSUB(4'b0000),
 		.Z(Y)
 	);
-endmodule
+endmodule
```
KrystalDelusion: A couple of the files are missing trailing end lines

tests/arch/nexus/dsp_equiv.ys
```diff
@@ -0,0 +1,154 @@
...
+hierarchy -top mac_wide_s
+flatten
+proc
+design -stash gate
+design -copy-from gold -as gold mac_wide_s
+design -copy-from gate -as gate mac_wide_s
+miter -equiv -make_assert -flatten gold gate equiv
+sat -seq 1 -set-init-zero -prove-asserts -verify equiv
+
+# y = a*b - c (minuend mul)
```
KrystalDelusion: > minuend

typo?

KrystalDelusion (changes requested):
I'll admit, I'm very much not familiar with pmgen, but the pattern here (separate `match ffA` and `ffB`) seems quite different from e.g. ice_40 or xilinx which have a singular `match ff` (or rather, one for inputs one for outputs) with generic arguments and `subpattern`.  Is there a reason for this difference?  Also I think this needs rebasing.

hexratcc:
@KrystalDelusion 
>  Is there a reason for this difference?

No real reason,  I'm also new to pmgen and this was just how it grew,  I've reworked it to match the other implementations.

KrystalDelusion (approved):
Seems reasonable and tests are passing; provided it works as expected on hardware (which it appears has been verified) then I think this is good to go.

PR #5933, opened 2026-06-05
Various cleanups

### LINKED ISSUE #4080: Use /usr/bin/env for bash on remaining places
### Feature Description

Check remaining places of `/bin/bash` use and replace with `/usr/bin/env` as in  https://github.com/YosysHQ/yosys/pull/3887

### LINKED ISSUE #4438: Clean up uses of #if(n)def EMSCRIPTEN
As we removed this from the supported builds in the Makefile, we should also clean up the remainders in the codebase.

mmicko:
- WASI supports filesystem, so lets use that
- did remaining /bin/bash conversions ( closes #4080 )
- removed VisualStudio helper script
- Removed broken EMSCRIPTEN and related files ( closes #4438 )
- Removed .dockerignore file left after Dockerfile removal


whitequark (dismissed):
I think the Emscripten removal ~~is~~ may be a mistake, I'll explain in detail once I'm no longer on mobile

whitequark:
Now that I'm at desktop: the WASI build does not replace Emscripten, as it is not well integrated with JS. WASI is a CLI platform, YosysJS is a library. If anybody is still using YosysJS they will find WASI build probably not well fit for their goals. Also it should not be too hard to keep it working with CMake.

However, that said, from a personal perspective as a maintainer removing it sounds good. I will dismiss my review. I just wanted to put this note here to make sure it's recorded and to avoid issues later if someone's workflow breaks.

PR #5934, opened 2026-06-05
read_aiger: import standard-cell mappings from ABC

Ravenslofty:

Yet another part of the ABC unification work.

`read_xaiger2` re-implements `read_aiger` and `abc9_ops -reintegrate` in one pass; while there are technical advantages to `write_xaiger2` bitblasting the design as late as possible, once that bitblasting has occured there's no point to having an experimental pass doing what existing code does.

This is step one: `read_aiger -xaiger` now parses the standard-cell mapping section of an input file. Step two is to move the cell renaming code to `abc9_ops -reintegrate` so it can correctly map cells for both `abc9` and `abc_new` flows.

### DIFF
--- a/frontends/aiger/aigerparse.cc
+++ b/frontends/aiger/aigerparse.cc
@@ -406,7 +406,7 @@ void AigerReader::parse_xaiger()
 		module->connect(n0, State::S0);
 
 	int c = f.get();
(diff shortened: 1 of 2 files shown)


widlarizer (approved):
Looks good

PR #5936, opened 2026-06-05
Let nested make calls parallelize

KrystalDelusion:
Bump minimum cmake version to 3.28 to use `JOB_SERVER_AWARE` when calling the `test-vanilla` target (and also `docs-prepare` and `test-docs`, though it's not as important there).

Sidenote that install instructions still specify 3.27 as the min because I used the web interface and could only edit the one file.  I'm not sure if there's a specific reason why we are targeting 3.27 as the min version, but if not I think this change might be nice to have (or I'm dumb and missing how to get the tests parallel without this, I'm also unclear if it's possible to do this when using Ninja).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,10 @@ include(YosysAbcSubmodule)
 include(YosysVerific)
 include(UseHomebrew)
 
+# Fix for using Clang from nixpkgs
+# see https://github.com/YosysHQ/yosys/pull/5936#issuecomment-4637319568
+set(CMAKE_CXX_SCAN_FOR_MODULES NO)
+
 # Build options.
 set(YOSYS_COMPILER_LAUNCHER "" CACHE STRING "Compiler launcher (ccache, sccache)")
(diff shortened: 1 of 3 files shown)


whitequark (approved):
LGTM, good catch!

KrystalDelusion:
I don't understand why this is failing CI, the error reported seems to only show up in the Nix flake build (which uses Ninja), and reports 
```
yosys> /build/source/kernel/yosys_common.h:23:10: fatal error: 'array' file not found
```
But I can't reproduce this locally, or see a reason why the changes here would make a difference when trying to build Yosys (and even while testing as far as I understand Ninja builds should ignore the `JOB_SERVER_AWARE` line).

flumf:
CMake 3.28 enables scanning C++ sources by default, which is broken when using Clang from nixpkgs (see NixOS/nixpkgs#452260).

Since Yosys isn't using C++ modules currently, adding `set(CMAKE_CXX_SCAN_FOR_MODULES NO)` to the toplevel CMakeLists.txt seems like a serviceable workaround.

KrystalDelusion:
Thanks @flumf ! That appears to have fixed it

PR #5938, opened 2026-06-07
Add UDP support

taktoa:
I wanted to use Yosys for verilog vs netlist equivalence checking with ASAP7, but the simulation models for ASAP7 use UDPs. I thought that it must be possible to elaborate some large subset of UDPs into regular logic (erroring on any that are nondeterministic), so I asked Claude to write a module that does that, and make sure that the ASAP7 cells have the same behavior under both `sim` and iverilog. It seems to work.

I'm not an expert on these dark corners of Verilog semantics or the Yosys codebase, so I've had to lean on Claude a decent amount, but I did review the code and check for things that obviously don't make sense. I'm happy to make any requested changes or abandon this entirely, if it's too scary to merge. I thought it was probably short enough and fixes a common enough problem that it might be an attractive target to spend maintainer review time on.

My understanding is that the `verilog_udp.cc` module elaborates a combinational UDP into some `eqx`s, and then for sequential UDPs it does something like

```
assign out = G(in, prev_in, state);
prev_in <= in;
state   <= out;
```

where G is the `eqx` function.

It comes with some cosim tests that include an (I think) reasonable set of UDPs, and then I've also manually run it on all the UDPs in the ASAP7 simulation model. Here's the command for running it on the ASAP7 UDPs, assuming you have [asap7sc7p5t_SEQ_RVT_TT_220101.v](https://github.com/The-OpenROAD-Project/asap7sc7p5t_28/blob/main/Verilog/asap7sc7p5t_SEQ_RVT_TT_220101.v) in the root of a yosys checkout:

```
  YOSYS=$PWD/build/yosys python3 - <<'PY'
  import re, subprocess, sys, tempfile
  src = open('asap7sc7p5t_SEQ_RVT_TT_220101.v').read()
  for name in dict.fromkeys(re.findall(r'primitive\s+(\w+)', src)):
      block = re.search(r'primitive\s+'+name+r'\b.*?endprimitive', src, re.S).group(0)
      f = tempfile.NamedTemporaryFile('w', suffix='.v', delete=False); f.write(block); f.close()
      subprocess.run([sys.executable, 'tests/verilog/udp_cosim.py', f.name], check=True)
  PY
```

### DIFF
--- a/frontends/verilog/CMakeLists.txt
+++ b/frontends/verilog/CMakeLists.txt
@@ -23,6 +23,8 @@ yosys_frontend(verilog
 	verilog_frontend.h
 	verilog_lexer.h
 	verilog_location.h
+        verilog_udp.cc
+        verilog_udp.h
 	${FLEX_verilog_lexer_OUTPUTS}
 	${BISON_verilog_parser_OUTPUTS}
 	INCLUDE_DIRS
--- a/frontends/verilog/verilog_lexer.l
+++ b/frontends/verilog/verilog_lexer.l
@@ -280,6 +280,7 @@ static parser::symbol_type process_str(char *str, int len, bool triple, parser::
 %x SYNOPSYS_FLAGS
 %x IMPORT_DPI
 %x BASED_CONST
+%x UDPTABLE
 
 UNSIGNED_NUMBER [0-9][0-9_]*
 FIXED_POINT_NUMBER_DEC [0-9][0-9_]*\.[0-9][0-9_]*([eE][-+]?[0-9_]+)?
@@ -355,6 +356,9 @@ TIME_SCALE_SUFFIX [munpf]?s
 
 "module"       { return parser::make_TOK_MODULE(out_loc); }
 "endmodule"    { return parser::make_TOK_ENDMODULE(out_loc); }
+"primitive"    { return parser::make_TOK_PRIMITIVE_DEF(out_loc); }
+"endprimitive" { return parser::make_TOK_ENDPRIMITIVE(out_loc); }
+"table"        { BEGIN(UDPTABLE); return parser::make_TOK_TABLE(out_loc); }
 "function"     { return parser::make_TOK_FUNCTION(out_loc); }
 "endfunction"  { return parser::make_TOK_ENDFUNCTION(out_loc); }
 "task"         { return parser::make_TOK_TASK(out_loc); }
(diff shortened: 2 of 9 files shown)


whitequark:
I would be opposed to merging anything written with Claude.

taktoa:
If that is the policy, can someone please update this part of the [contribution guide](https://yosyshq.readthedocs.io/projects/yosys/en/latest/yosys_internals/extending_yosys/contributing.html#how-to-review):

> If a PR author submits LLM outputs they haven’t understood themselves, they will not be able to implement feedback. Take this into consideration as well. We do not ban LLM code from the codebase, we ban bad code.

I don't believe this is bad code or that my understanding of it is too poor to implement feedback.

widlarizer:
That is still our current policy and we'll likely be altering it soon.

`read_verilog` is neglected but load-bearing at the moment. This makes it the worst part of Yosys to make big changes to, while it's also a part users also are directly exposed to a lot, with its lack of features and quality overall. Have you tried https://github.com/povik/yosys-slang ? It has support for importing UDPs as blackboxes. We would like to move Yosys to yosys-slang entirely.

While others may be fundamentally opposed to an LLM-heavy PR such as this one, I am not. On the other hand, it's fairly large, covers a niche use case that we explicitly declare is unsupported (only select systemverilog features are supported), and it's improving a part of the codebase we are planning to remove. This means I am unlikely to commit to a review.

povik:
Note yosys-slang won't import a model for UDPs, rather it imports them as opaque blackboxes. This read_verilog enhancement does model the UDPs

taktoa:
> Have you tried https://github.com/povik/yosys-slang ? It has support for importing UDPs as blackboxes. We would like to move Yosys to yosys-slang entirely.

That makes sense to me, and I would be willing to port this over to yosys-slang. Most of the change is `verilog_udp.cc` which should be agnostic to the parser used.

> On the other hand, it's fairly large, covers a niche use case that we explicitly declare is unsupported (only select systemverilog features are supported), and it's improving a part of the codebase we are planning to remove.

UDPs are a Verilog (IEEE 1364-2005 chapter 8) feature, not a SystemVerilog feature.

widlarizer:
Sorry for the confusion - it's not sv but UDPs are not in the standardized synthesizable subset of verilog though, which is what we declare to support in the docs.

taktoa:
I've added support for UDPs to `yosys-slang` in https://github.com/taktoa/yosys-slang/commit/2ceda1f838ab1736ae153d03cc486b97718d6695 (branch `taktoa-udp`). Given `yosys-slang`'s contribution policy around LLM-generated code, I won't be following up with a PR unless asked, and I will be closing this one.

I would be happy to put whatever legwork in to show that either of these patches are correct, and I encourage both the yosys and yosys-slang projects to accomodate LLM-generated submissions in the future by simply raising their standards for testing, fuzzing, and formal verification.

widlarizer:
Thanks for your input. IMO that "simply" is doing a lot of heavy lifting, particularly in the case of Yosys

PR #5943, opened 2026-06-08
move `abc9_ops -reintegrate` into its own pass

Ravenslofty:
Yet another part of the ABC unification work.

Both `abc9` and `abc_new` perform design "reintegration", in which the contents of an existing module in the design are replaced with the imported mapping.
- `abc9` renames signals in `read_aiger`, and reintegrates signals with the same name inside `abc9_ops -reintegrate`.
- `abc_new` performs both steps inside `read_xaiger2`, which I plan to remove.

`abc_new`'s signal map relies on knowing the contents of the target module to transfer blackbox cells across, so this can't be done inside `read_aiger`, but instead has to be done inside `abc9_ops -reintegrate`.

`abc9_ops` is already a kind of complex pass doing a lot of things, so I think it would be cleaner to put the `-reintegrate` code into its own pass.

No functional change intended, and the update to `abc9` to call the renamed pass means that it's covered by existing ABC9 tests.

### DIFF
--- a/passes/techmap/CMakeLists.txt
+++ b/passes/techmap/CMakeLists.txt
@@ -67,11 +67,15 @@ yosys_pass(abc9_ops
 	REQUIRES
 		proc
 )
+yosys_pass(abc_ops_reintegrate
+	abc_ops_reintegrate.cc
+)
 yosys_pass(abc9
 	abc9.cc
 	DEFINITIONS
 		${abc_definitions}
 	REQUIRES
+		abc_ops_reintegrate
 		abc9_exe
 		abc9_ops
 		aigmap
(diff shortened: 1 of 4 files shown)


widlarizer (changes requested):
I'm in favor of finer, more hierarchical organization of our code, which I expect is the main motivation here. But the change from a command-like option to a pass of its own seems orthogonal. I'm in favor of breaking down passes into components across various files (see opt_clean) regardless of where the actual pass declaration lives. Is there a specific reason to move this to a new pass?

Ravenslofty:
So, my current plan is that the `abc_ops_reintegrate` pass will take in `-map` as path to the signal remapping file. It feels...very clumsy to me that `-reintegrate` takes sub-options that are not applicable to other "modes" of `abc9_ops`. To me, `-reintegrate` is simply orthogonal code that can and should be put somewhere else, it's never used in combination with the other `abc9_ops` modes.

whitequark (approved):
CMake changes look good

widlarizer (approved):
I see, that kind of does warrant a separate interface

PR #5949, opened 2026-06-09
MSVC full build

mmicko:
Made CMake build do build ABC as well.
Note that ABC does not provide debug pthread library, so that build will compile/link but not work.


whitequark (approved):
CMake changes LGTM, haven't evaluated changes to abc

mmicko:
Changes to ABC are sent upstream and they do fix existing warnings, for our build those warnings are considered errors so they were failing, I could have just ignore those but it was cleaner to fix them.

Issue #964, opened 2019-04-27
Possible enhancements to muxcover


eddiehung:
## Steps to reproduce the issue

` yosys -p "proc; techmap; opt; muxcover; stat"`

```
module mux_index_8_3 #(parameter N=8, parameter W=3) (input [N*W-1:0] i, input [$clog2(N)-1:0] s, output [W-1:0] o);
assign o = i[s*W+:W];
endmodule
```

Works when `W=2` or `W=4`...

## Expected behavior

3 x `$_MUX8_`

## Actual behavior

```
   Number of cells:                 36
     $_AND_                          5
     $_MUX4_                         2
     $_MUX8_                         2
     $_MUX_                         21
     $_OR_                           1
     $_XOR_                          5
```

eddiehung:
Before `muxcover` it looks like this: [show.pdf](https://github.com/YosysHQ/yosys/files/3123444/show.pdf)

After: [show.pdf](https://github.com/YosysHQ/yosys/files/3123446/show.pdf)

hexratcc:
It appears that `yosys -p "proc; techmap; opt; muxcover; stat"` behaves as expected when W is power of two, but for non power of two W's `i[s*W+:W]` lowers to a barrel shifter, which `muxcover` can't cover. This can be fixed by using the `peepopt` pass like so: `yosys -p "proc; opt; peepopt; techmap; opt; muxcover; stat"`.

PR #5953, opened 2026-06-10
Add muxcover x peepopt regression test (#964).

hexratcc:
This PR adds regression tests related to #964.

### DIFF
--- a/tests/various/muxcover_index.ys
+++ b/tests/various/muxcover_index.ys
@@ -18,8 +18,8 @@ clean
 opt_expr -mux_bool
 
 select -assert-count 2 t:$_MUX8_
-select -assert-count 0 t:$_MUX_ t:$_MUX4_ t:$_MUX16_
-select -assert-count 0 t:$_AND_ t:$_OR_ t:$_XOR_ t:$_NAND_ t:$_NOR_ t:$_XNOR_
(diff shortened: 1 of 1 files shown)


hexratcc:
[sc-602]

tests/various/muxcover_index.ys
```diff
@@ -0,0 +1,87 @@
...
+opt
+peepopt
+techmap
+opt
+muxcover -mux8
+clean
+opt_expr -mux_bool
+
+select -assert-count 2 t:$_MUX8_
+select -assert-count 0 t:$_MUX_ t:$_MUX4_ t:$_MUX16_
```
widlarizer: There's a pattern used in the tests like "there's no cells that don't match this type" rather than listing some types that shouldn't be created

widlarizer: For example tests/arch/analogdevices/adffs.ys
```
select -assert-count 1 t:FFCE
select -assert-count 1 t:LUT1

select -assert-none t:FFCE t:LUT1 %% t:* %D
```

widlarizer (approved):

tests/various/muxcover_index.ys
Ravenslofty: Yeah, I would also say use that pattern.

PR #5955, opened 2026-06-12
simplify.cc: Fix crash when identifier matches module

### LINKED ISSUE #5954: Yosys crash when module output signal name same as module name
### Version

Yosys 0.66+70 (git sha1 8eb313307, Release, Clang /usr/bin/clang++ 18.1.8)

Linux

### Reproduction Steps

With source:
```
module pdm #(
    parameter PHY = "zero"
)(
    output wire pdm,
);
    if (PHY == "zero") begin
        assign pdm = 1'b0;
    end
endmodule
```
and command line of:
```
yosys -p 'synth_ice40 -top pdm ' pdm.v
```

### Expected Behavior

Yosys should run to completion without crashing.

### Actual Behavior

Yosys runs and then crashes with the following output:
```
2.3. Executing AST frontend in derive mode using pre-parsed AST for module `\pdm'.
Generating RTLIL representation for module `\pdm'.
  what():  basic_string::at: __n (which is 4) >= this->size() (which is 4)
Aborted (core dumped)
```
This has been observed on several versions of yosys going back some time (for example Yosys 0.61+129 (git sha1 153ddc0c8, clang++ 18.1.8 -fPIC -O3) has this crash), however it is *not* seen on much older versions. Yosys 0.42+15 (git sha1 2fd2b6538, clang++ 14.0.0-1ubuntu1.1 -fPIC -Os) works fine for example.

KrystalDelusion:

Fix #5954.  #4235 added auto identifiers such as `\a.x`, and then extracts them as `\x` during simplification if the current scope is `\a`.  If the identifier (`identifier_str`) starts with the scope name (`current_ast_mod->str`), it tries to evaluate `identifier_str.at(current_ast_mod->str.size()) == '.'`.  If the two strings are exactly the same, then it tries to access the next character (to check if it's a `\a.x` situation) which is out of bands and crashes.

Only check for the `.` if the identifier string is longer than the scope string, preventing the out of bounds access.  Not sure if it's worth including a test case here, but the minimal script is:

```
read_verilog << EOT
module pdm #(
    parameter PHY = "zero"
)(
    output wire pdm
);
    if (PHY == "zero") begin
        assign pdm = 1'b0;
    end
endmodule
EOT
```

### DIFF
--- a/frontends/ast/simplify.cc
+++ b/frontends/ast/simplify.cc
@@ -1057,7 +1057,7 @@ bool AstNode::simplify(bool const_fold, int stage, int width_hint, bool sign_hin
 
 #if 0
 	log("-------------\n");
-	log("AST simplify[%d] depth %d at %s:%d on %s %p:\n", stage, recursion_counter, location.begin.filename, location.begin.line, type2str(type), this);
+	log("AST simplify[%d] depth %d at %s on %s %p:\n", stage, recursion_counter, location.to_string(), type2str(type), this);
 	log("const_fold=%d, stage=%d, width_hint=%d, sign_hint=%d\n",
 			int(const_fold), int(stage), int(width_hint), int(sign_hint));
 	// dumpAst(nullptr, "> ");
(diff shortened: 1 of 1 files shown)


widlarizer (approved):
Thanks!

PR #5959, opened 2026-06-15
Improve test

mmicko:
This PR:
- force Ninja to give output while running tests
- not we can use V=1 or VERBOSE=1 when running individual tests to get log at least for some of tests
- Fix some documentation
- Add option to run functional tests separately


mmicko:
@whitequark can we somehow make sure that parallel parameter is somehow propagated to make  when we run using `cmake --build build --target test --parallel 8` ?

KrystalDelusion:
While it is possible for Ninja to use the jobserver provided by Make, there doesn't seem to be a way to do the reverse (which is what we have here).  For now it might be better to just give a warning that we currently recommend using the `Unix Makefiles` generator to run tests in parallel.  Or maybe add a `tests/Makefile.conf` build target which contains the exported `makefile_vars` and suggest `cmake --build build --target tests-prepare --parallel && make -C tests -j`.

whitequark:
What Krystine said, yeah.

whitequark (approved):

PR #5961, opened 2026-06-15
opt_muxtree: fix reconvergence

### LINKED ISSUE #5957: Self-assigning case corrupts ternary selector synthesis
### Version

Yosys 0.66+4 (git sha1 8125af88d, g++ 13.3.0-6ubuntu2~24.04.1 -fPIC -O3)

Linux

### Yosys synthesis command

```bash
```

### Verilog before synthesis

Input RTL (`rtl.v`):

```verilog
module top(input in_1, input in_4, output reg out);
  always @* begin
    out = in_1 ? in_4 : 1'b0;

    case (in_4)
      1'b0: out = out;
      1'b1: out = out;
    endcase
  end
endmodule
```

Important minimization note: the two concrete `case` arms must remain separate.Merging them into a multi-label arm does not reproduce the issue:

```verilog
case (in_4)
  1'b0, 1'b1: out = out;
endcase
```

### Verilog after synthesis

```verilog
/* Generated by Yosys 0.66+4 (git sha1 8125af88d, g++ 13.3.0-6ubuntu2~24.04.1 -fPIC -O3) */

(* top =  1  *)
(* src = "rtl.v:1.1-10.10" *)
module top_synth(in_1, in_4, out);
  (* src = "rtl.v:1.18-1.22" *)
  input in_1;
  wire in_1;
  (* src = "rtl.v:1.30-1.34" *)
  input in_4;
  wire in_4;
  (* src = "rtl.v:1.47-1.50" *)
  output out;
  wire out;
  assign out = in_1;
endmodule
```

The relevant change is that the RTL ternary:

```verilog
out = in_1 ? in_4 : 1'b0;
```

is synthesized as:

```verilog
assign out = in_1;
```

This drops the dependence on `in_4`.

### Step-by-step Analysis

The input RTL first assigns `out` with a blocking assignment:

```verilog
out = in_1 ? in_4 : 1'b0;
```

Immediately after that assignment, the `case (in_4)` statement only assigns
`out` back to itself:

```verilog
case (in_4)
  1'b0: out = out;
  1'b1: out = out;
endcase
```

Under Verilog blocking-assignment semantics, each `out = out` statement reads the value that was just assigned by the ternary expression. Therefore the case statement is a no-op and the whole process should be equivalent to:

```verilog
assign out = in_1 ? in_4 : 1'b0;
```

Equivalently, the intended Boolean function is:

```text
out = in_1 & in_4
```

The synthesized netlist instead contains:

```verilog
assign out = in_1;
```

This is equivalent to:

```verilog
assign out = in_1 ? 1'b1 : 1'b0;
```

The "replaced with constant `1'b1`" observation is also visible in intermediate RTLIL. After process conversion and basic cleanup, the ternary mux still has `in_4` as its high data input:

```text
cell $mux $ternary$rtl.v:3$2
  connect \A 1'0
  connect \B \in_4
  connect \S \in_1
end
```

After adding `opt_muxtree`, the same mux input has changed to constant `1'1`:

```text
cell $mux $ternary$rtl.v:3$2
  connect \A 1'0
  connect \B 1'1
  connect \S \in_1
end
```

The `case` selector is also `in_4`, while the previous ternary assignment uses `in_4` as the data input selected when `in_1` is high. The self-assigning case should not constrain or overwrite that data input. The intermediate RTLIL therefore suggests that `opt_muxtree` incorrectly uses information from the `case (in_4)` decision when simplifying the earlier ternary mux.

This changes behavior for:

```text
in_1=1, in_4=0
```

For these inputs, the RTL evaluates the ternary as `out = in_4 = 0`, while the synthesized netlist evaluates `out = in_1 = 1`.

### Expected Behavior

Both reachable `case (in_4)` items are self-assignments, so the case statement should be a no-op. The output should remain the ternary result:

```text
out = in_1 ? in_4 : 1'b0
```

For known values, this is equivalent to:

```text
out = in_1 & in_4
```

### Actual Behavior

```text
out = in_1
```

This is not equivalent to the RTL because it returns `1` when `in_1=1` and `in_4=0`, while the RTL returns `0`.

widlarizer:
Fixes #5957.

A mux is now a root mux even it has multiple unique mux *cell port* users, not just mux *cell* users. This inhibits the erroneous optimization which was wrecking the input cone that was also observed by another port of the same mux cell.

Test is included

### DIFF
--- a/tests/opt/opt_muxtree_port_reconverge.ys
+++ b/tests/opt/opt_muxtree_port_reconverge.ys
@@ -1,0 +1,55 @@
+read_rtlil << EOT
+module \top
+  wire input 1 \a
+  wire input 2 \b
+  wire output 3 \out
+  wire $0\out[0:0]
+  wire \y
+  wire $1\out[0:0]
+  wire \s
+  wire \q
+  wire \n_a
+
+  cell $mux \mux
+    parameter \WIDTH 1
+    connect \A 1'0
+    connect \B \a
+    connect \S \b
+    connect \Y \y
+  end
+  
+  attribute \full_case 1
+  cell $pmux \pmux
+    parameter \WIDTH 1
+    parameter \S_WIDTH 2
+    connect \A 1'x
+    connect \B { \y \y }
+    connect \S { \n_a \s }
+    connect \Y \q
+  end
+  
+  attribute \full_case 1
+  cell $not \not
+    parameter \A_SIGNED 0
+    parameter \Y_WIDTH 1
+    parameter \A_WIDTH 1
+    connect \A \a
+    connect \Y \n_a
+  end
+
+  connect $0\out[0:0] $1\out[0:0]
+  connect \s \a
+  connect $1\out[0:0] \q
+  connect \out \q
+end
+EOT
+# a drives out
+select -assert-any i:a %co* o:out %i
+
+opt_muxtree
+# a gets optimized out
+opt_reduce
+opt_expr
+
+# a no longer drives out
+select -assert-any o:out %ci* i:a %i
--- a/tests/opt/opt_muxtree_port_reconverge.ys
+++ b/tests/opt/opt_muxtree_port_reconverge.ys
@@ -47,9 +47,8 @@ EOT
 select -assert-any i:a %co* o:out %i
 
 opt_muxtree
-# a gets optimized out
 opt_reduce
 opt_expr
 
-# a no longer drives out
+# a still drives out
 select -assert-any o:out %ci* i:a %i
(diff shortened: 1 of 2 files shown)


tests/opt/opt_muxtree_port_reconverge.ys
```diff
@@ -0,0 +1,55 @@
...
+# a drives out
+select -assert-any i:a %co* o:out %i
+
+opt_muxtree
+# a gets optimized out
+opt_reduce
+opt_expr
+
+# a no longer drives out
+select -assert-any o:out %ci* i:a %i
```
KrystalDelusion: The comments here should be updated (or just removed), since the point is that `a` _shouldn't_ be optimized out.  Also this file is missing a newline at the end.

KrystalDelusion (changes requested):
Thanks for this, minor nit in the test

KrystalDelusion (approved):

Issue #5719, opened 2026-03-02
User control over latch inference


widlarizer:
### Feature Description

#5718 but for latches. 

Currently, inferred latches emit no warning, not even when they're not supported by dedicated FPGA hardware primitives and get synthesized into LUTs:

`yosys -p "read_verilog tests/sim/dlatch.v; synth_ice40"`

techhead-datnextcom:
hi this is industry important - any discussion on these ?

whitequark:
Should be resolved by https://github.com/YosysHQ/yosys/pull/5963.

PR #5963, opened 2026-06-17
User control over latch inference

hexratcc:
This PR adds a message severity option for latch inference (#5719). The default for now is `AUTO=WARN` (defaulting to an error breaks about 30 tests, but we may discuss if it's better to default to it instead). 
Additionally, this PR adds the `-nolatches` and `-latchonly` options to `check`, which emits a warning when a latch is encountered (error with `-assert`, standard with `check` behavior, for users who may want to ensure that no latches have been emitted. 
Additionally, for default synth flows targets which don't support latches, a new check pass is now invoked which ensures that no latches get to the `techmap` of these flows.


KrystalDelusion:
I wonder if it's worth adding something brief to [the map_luts stage of the synthesis starter docs](https://yosyshq.readthedocs.io/projects/yosys/en/latest/getting_started/example_synth.html#luts) about the latch check there.

Also am I interpreting this right, if the `check.latchonly` scratchpad variable is set then it forces the `-nolatches` option?  This pattern feels incorrect:
```cpp
	active_design->scratchpad_set_bool("check.latchonly", true);
	run("check -assert");
	active_design->scratchpad_unset("check.latchonly");
```
If we only ever set it, run `check -assert`, then unset it, and only during synth passes... why not just make it a `check` arg?  e.g. `check -assert-nolatches` or something?

hexratcc:
> If we only ever set it, run check -assert, then unset it, and only during synth passes... why not just make it a check arg? e.g. check -assert-nolatches or something?

This is mainly a workaround for testing, as a workaround (some tests emit `check` warnings, which, with `-assert`, get promoted into errors, which makes it difficult to keep the original behavior of those tests). I originally didn't see any need to expose this to the user via a separate option.

I think that if this were to be exposed as a separate option it could be interesting to expose the same functionality for other `check` options as well.

KrystalDelusion:
Could you potentially workaround by using `logger -werror` and dropping the `-assert`?  Or alternatively if it's just for testing and not really expected during real world synth calls, could we maybe move the scratchpad variable into the test scripts?

hexratcc:
I've  opted to add the `-nolatches` flag as a response to @KrystalDelusion, which replaces the previous scratchpad strategy I used in default synth flows. I think this is also useful beyond the synth flows: a user with a design that emits other warnings, but who only wants to confirm there are no latches, can run `check -latchonly` without having to reach for the logger.

hexratcc:
@KrystalDelusion could you help me determine the cause behind the failing test case? `various/check_nolatches.ys` passes for me locally and I have no idea what the reason could be, as I'm not very familiar with Yosys' CI.

techlibs/efinix/synth_efinix.cc
```diff
@@ -190,6 +203,10 @@ struct SynthEfinixPass : public ScriptPass
 		if (check_label("map_ffs"))
 		{
 			run("dfflegalize -cell $_DFFE_????_ 0 -cell $_SDFFE_????_ 0 -cell $_SDFFCE_????_ 0 -cell $_DLATCH_?_ x");
+			if (help_mode)
+				run("check -latchonly -assert", "(only if -latches error, the default)");
+			else if (latches == "error")
+				run("check -latchonly -assert");
```
KrystalDelusion: You should be able to collapse this to `if (latches == "error" || help_mode)` since the actual call is the same in both branches (and the info text is only used in help mode).

KrystalDelusion (approved):
LGTM.  Would be nice to have some way to re-use the latches arg code instead of reproducing it for each synth pass, but I think it would be better to do that separately, and in a way that allows for more than just this one arg to be deduplicated (meaning I don't think it's worth trying to address for this PR, just wanted to comment on it because I don't remember where we ended up re: unifying synth_* passes).

hexratcc:
@KrystalDelusion Agreed, I just wanted to get this through but there will be more work done related to latch inference later, but I wish to discuss that with @widlarizer first.

PR #5967, opened 2026-06-18
Slang frontend integration

mmicko:
This PR adds : 
- [x] Add slang as library
- [x] Add fmt as library (since required by slang)
- [x] Add yosys-slang submodule in frontend/slang
- [x] Add option to disable building with slang (if someone do have issue building it)
- [x] Bumps minimum required compiler to clang-16
- [x] Temporary fix for building documentation

TODO:
- [x] Fix warnings/errors on yosys-slang repo to be able to use default compiler settings 
- [x] Fix VisualStudio build ( alternative https://github.com/YosysHQ/yosys/pull/5972 )
- [x] Check what happens when we load yosys-slang plugin https://github.com/povik/yosys-slang/pull/348
- [x] Actually fix building documentation (requires slang API change https://github.com/MikePopoloski/slang/pull/1881 ) need https://github.com/povik/yosys-slang/pull/346
- [x] Nix build need fixing (submodules of submodule are not used/needed)


libs/CMakeLists.txt
```diff
@@ -7,3 +7,18 @@ add_subdirectory(json11)
 add_subdirectory(minisat)
 add_subdirectory(sha1)
 add_subdirectory(subcircuit)
+set(_old_BUILD_SHARED_LIBS "${BUILD_SHARED_LIBS}")
```
whitequark: You should be able to use `block()` to temporarily override a variable.

mmicko: Thanks. This works and looks cleaner

mmicko:
@whitequark any idea about how to fix nix build https://github.com/YosysHQ/yosys/actions/runs/27818376168/job/82325224549 ? please note that we actually do not care about submodules under `frontend/slang/lib` and that they are not used in build, can it be made to ignore that ?

whitequark:
Will check with someone

flumf:
`.?submodules=1` should be dropped from the Nix build in CI (`self.submodules = true;` is set in the flake instead, which correctly handles fetching submodules).

I have noticed that the Nix build results in an unknown slang version though (I'm not sure if this is a concern or of a good way to fix it):
```
yosys> slang_version
yosys-slang revision UNKNOWN
slang revision UNKNOWN
```

mmicko:
To catch all output to actually be processed by yosys logs we can use:
```
    auto guard = slang::OS::captureOutput([&](std::string_view text, bool) { log("%s", text); });
    slang::driver::Driver driver;
    driver.setTerminalColorsEnabled(false);
```
It also disable coloring of output since that put ANSI escape sequences into log files, so would need to work on that in future to display it on screen but not in logs, but it would be part of overall log improvement effort.

povik:
@mmicko please hold off merging this until I make a few more changes upstream

mmicko:
@povik sure, will do clean builds as well to check if there is anything need updating still.

povik:
I've done the changes I wanted. The upstream project is now named `sv-elab` and its home is at https://github.com/povik/sv-elab. Github makes the original yosys-slang path work for fetching but it would be good to use the new path, still.

PR #5973, opened 2026-06-19
Move rename logic to abc_ops_reintegrate

Ravenslofty:
Another part of the ABC reunification work, although this patch ended up getting away from me a bit.

In the `abc9` flow, the reintegration of an ABC-mapped module into the main design is split across `read_aiger` and `abc_ops_reintegrate`, while in the `abc_new` flow, both steps are done inside `read_xaiger2`. The `abc_new` flow has the quirk that its box handling requires access to the original module.

While `write_xaiger2` has technical advantages in that it bit-blasts at point of AIGER emission, `read_xaiger2` duplicates functionality in the more stable `read_aiger`. To remove `read_xaiger2`, however, we need to move the `read_aiger` rename logic to `abc_ops_reintegrate` to have access to the original module to handle `abc_new`'s box handling.

Move the code from `read_aiger` to `abc_ops_reintegrate`.

The test suite has a number of modified tests due to ordering changes, plus two invalid tests from `various/abc9.ys` were removed.


tests/arch/intel_alm/logic.ys
```diff
@@ -5,8 +5,8 @@ equiv_opt -assert -map +/intel_alm/common/alm_sim.v synth_intel_alm -family cycl
 design -load postopt # load the post-opt design (otherwise equiv_opt loads the pre-opt design)
 cd top # Constrain all select calls below inside the top module
 
-select -assert-count 1 t:MISTRAL_NOT
-select -assert-count 6 t:MISTRAL_ALUT2
+select -assert-count 4 t:MISTRAL_NOT
+select -assert-count 3 t:MISTRAL_ALUT2
```
hexratcc, line 9: Was this expected? It looks like the `$lut` cells get named in `post_process` while wires still have their `$aiger` names (since the rename now happens later in reintegrate), so the LUT lookup in the NOT handling searches  `$lut\B1` etc. and fails? Or is this something to fix up later?

hexratcc (changes requested):
lgtm just want clarification, please see comment.

tests/arch/intel_alm/logic.ys
Ravenslofty: Hmm. Maybe I could try moving that code into reintegrate but it's a little clumsy and this patch is already quite big. So perhaps it goes in the "fixup later" category.

hexratcc: Fair enough.

hexratcc (approved):

PR #5613, opened 2026-01-16
Make async flip flops easy to opt with new $priority cell

widlarizer:
This PR aims to fix #3442 and supersede #4781 and #5570 by introducing a `$priority` cell. This should allow me to avoid adding new flip flop types while respecting set-reset priority in the Liberty-specified PDK cells selected by `dfflibmap`

PR #5976, opened 2026-06-22
Update to `dfflibmap` to be able to map LATCHes

jeras:
I added code to `dfflibmap` so it can map LATCHes in addition to FFs to cells from a Liberty file.
The choice to edit `dfflibmap` instead of creating a new pass `dlatchlibmap` was made in accordance to a related change to `dfflegalize` done by @hexratcc, where latches are supported regardless of the pass name.

PROCESS: This is my first no trivial contribution to Yosys, I am not involved in discussions yet, for now I would like to ask @hexratcc for a bit of help with the process.

MOTIVATION: Open source PDK standard libraries now use `techmap -map latch_map.v` to map latches to SCL cells. This requires custom map files (they are partially available) and a `dfflegalize -cell <cell_type_pattern>` (the patterns are not part of the PDK support for LibreLane yet). Adding this feature would broadly improve LibreLane support for latches without additional custom code for each PDK/SCL.

HOW: Code in `dfflibmap` intended for FFs was copied and modified to support LATCHes. 

TODO:
- split code and documentation changes so they are easier to review,
- try to organize some duplicated code into functions (there were already 2 partial copies, now there are 4),
- update cell selection process to take something more than the cell size into account (power?), now for some `sky130` HD cells the `*_2` variant is selected instead of the `*_1` (the two variants have the exact same size but different power).
- get tests to pass.

TESTING: I have copied the tests for `dfflibmap` renamed them to `dlatchlibmap*` and then made all the necessary search/replace changes (well a bit more, but still rather trivial). For now the tests are still failing.

STATUS: The code compile without warnings. I was able to test the basic principle of finding and mapping lathes with the `sky130` HD SCL. It works, but I did not check every detail manually, this is what the unit tests are for. The unit tests are able to perform LATCH mapping, but formal checking is still failing, and I do not have enough experience with formal to quickly figure out the issue.


QuantamHD:
@widlarizer I believe we discussed the possibility of erroring on unintentional latches by default unless they were correctly annotated with always_latch. Should yosys allow latch mapping by default or do we need to disable it by default or add the requisite check?


@jeras This is a great change. I think we should think a bit about its unintended consequences namely, that it'll allow you to write latch based code that you perhaps didn't intend, and now it'll be even more automatic where as before you would have needed to provide the latch mapping manually.

widlarizer:
@QuantamHD we've got #5963 in the works

jeras:
@QuantamHD Years ago while working on ASIC designs I would grep the synthesis log for the keyword `latch`, and report the issue to the team member who wrote the RTL. I do not know how professional tools handle it now. For ASIC designs containing latches this is probably what the designer would still check manually. Disallowing latches by default if they are outside of `always_latch` would probably make sense, at least it seems a good compromise. In part of the test code I had to use the old Verilog style, since `equiv_opt -map dlatchlibmap-sim.v ...` does not have a SystemVerilog option.

Note that I only have ASIC designs in mind here, not FPGA. Actually I never used lathes in any design. I started working on this, since it would simplify `open_pdks` integration with LibreLane. Another reason was that after looking a bit at the code it looked like something simple enough for my first Yosys contribution, and the C++ code was simple, but formal verification in unit tests is still bothering me.

For now I got one (out of 3) test scripts to pass. The tests using `equiv_opt`, but the ones using `miter` are not.

I am also seeing some inconsistent behavior with FF/LATCH cells with complementary outputs (`Q`/`QN`). Sometimes `QN` is used, sometimes an inverter added by `fddlegalize` is used. I should investigate this later when I have automatic tests. The current `dfflibmap` always picks a single output cell if available (probably smaller area), I was thinking about how/when to map complementary output cells, but I do not know what are the best practices for using them (advantages/disadvantages, use cases).

jeras:
I had to update the outdated Liberty file parser for LATCHes, fortunately I was able to just copy and modify the more recent code for FF parsing. Noticed FF parsing is not handling FFs with `enable` signal.

I got another test to pass, and one is failing only partially (not detecting equivalence mismatch of `clear`/`preset` priority).

widlarizer:
> Noticed FF parsing is not handling FFs with enable signal

The ff group doesn't take an enable signal (see [reference manual](https://people.eecs.berkeley.edu/~alanmi/publications/other/liberty07_03.pdf)), we infer it from the expression: #4698

jeras:
The code should be ready for a review now. The tests are passing and I did a bit of cleanup/rebase of the commits.

The LATCH tests are a copy/modify or `dfflibmap` tests for FF, the copies use the prefix `dfflatchmap`. Therefore the tests should provide the same coverage for mapping LATCHes as existing tests for mapping FFs.

A few points for the reviewer:
- I added a recommendation to run `opt_merge` after `dfflibmap`. `dfflegalize` can add inverters for each cell, if the RTL signal (reset/enable, ....) is active high, while the cell signal is active low. `opt_merge` recombines those inverters into a single one, and `opt -full` would also be an option.
- I suspect I introduced some new C++ warnings while collecting copied code into a function. While I understand pointers well enough to write the code, I do not understand the C++ syntax well enough to avoid warnings. I would be glad if the reviewer could help with this.
- Except for the last commit, the existing functionality (mapping FFs) is not affected.
- Depending on the SCL the last commit can change the strength of the selected cell. The original process choose the cell with smallest area and fewest pins, but if two cells were the same under those conditions, the later (stronger) one was selected, now the first one is selected.

Old code picking last of the equal area cells (from sky130 HD SCL 2 cells with `_2` suffix are selected):
```
  cell sky130_fd_sc_hd__dfxtp_1 (noninv, pins=3, area=20.02) is a direct match for cell type $_DFF_P_.
  cell sky130_fd_sc_hd__dfrtn_1 (noninv, pins=4, area=25.02) is a direct match for cell type $_DFF_NN0_.
  cell sky130_fd_sc_hd__dfrtp_1 (noninv, pins=4, area=25.02) is a direct match for cell type $_DFF_PN0_.
  cell sky130_fd_sc_hd__dfstp_2 (noninv, pins=4, area=26.28) is a direct match for cell type $_DFF_PN1_.
  cell sky130_fd_sc_hd__edfxtp_1 (noninv, pins=4, area=30.03) is a direct match for cell type $_DFFE_PP_.
  cell sky130_fd_sc_hd__dfbbn_1 (noninv, pins=6, area=32.53) is a direct match for cell type $_DFFSR_NNN_.
  cell sky130_fd_sc_hd__dfbbp_1 (noninv, pins=6, area=32.53) is a direct match for cell type $_DFFSR_PNN_.
  cell sky130_fd_sc_hd__dlxtn_1 (noninv, pins=3, area=15.01) is a direct match for cell type $_DLATCH_N_.
  cell sky130_fd_sc_hd__dlxtp_1 (noninv, pins=3, area=15.01) is a direct match for cell type $_DLATCH_P_.
  cell sky130_fd_sc_hd__dlrtn_2 (noninv, pins=4, area=17.52) is a direct match for cell type $_DLATCH_NN0_.
  cell sky130_fd_sc_hd__dlrtp_1 (noninv, pins=4, area=16.27) is a direct match for cell type $_DLATCH_PN0_.
```
New code picking first of the equal area cells (all selected cells have `_1` suffix):
```
  cell sky130_fd_sc_hd__dfxtp_1 (noninv, pins=3, area=20.02) is a direct match for cell type $_DFF_P_.
  cell sky130_fd_sc_hd__dfrtn_1 (noninv, pins=4, area=25.02) is a direct match for cell type $_DFF_NN0_.
  cell sky130_fd_sc_hd__dfrtp_1 (noninv, pins=4, area=25.02) is a direct match for cell type $_DFF_PN0_.
  cell sky130_fd_sc_hd__dfstp_1 (noninv, pins=4, area=26.28) is a direct match for cell type $_DFF_PN1_.
  cell sky130_fd_sc_hd__edfxtp_1 (noninv, pins=4, area=30.03) is a direct match for cell type $_DFFE_PP_.
  cell sky130_fd_sc_hd__dfbbn_1 (noninv, pins=6, area=32.53) is a direct match for cell type $_DFFSR_NNN_.
  cell sky130_fd_sc_hd__dfbbp_1 (noninv, pins=6, area=32.53) is a direct match for cell type $_DFFSR_PNN_.
  cell sky130_fd_sc_hd__dlxtn_1 (noninv, pins=3, area=15.01) is a direct match for cell type $_DLATCH_N_.
  cell sky130_fd_sc_hd__dlxtp_1 (noninv, pins=3, area=15.01) is a direct match for cell type $_DLATCH_P_.
  cell sky130_fd_sc_hd__dlrtn_1 (noninv, pins=4, area=17.52) is a direct match for cell type $_DLATCH_NN0_.
  cell sky130_fd_sc_hd__dlrtp_1 (noninv, pins=4, area=16.27) is a direct match for cell type $_DLATCH_PN0_.
```

## Cell mapping tables

Since I created this potentially useful table during research for this PR, I am attaching it here (it probably has a few typos). LEGEND: the `C` in the `Q` column stands for complementary outputs.

| world    | tech           |`C`|`E`|`S`|`R`|`Q`| sky130 HD | sky130 HDLL | sky130 [HML]S | sky130 LP | sky130 HVL | gf180mcu [79]t | ihp-sg13g2 |
|----------|----------------|---|---|---|---|---|-----------|-------------|---------------|-----------|------------|----------------|------------|
| `$dff`   | `$_DFF_P_`     |`P`|   |   |   |`P`| `dfxtp`   |             | `dfxtp`       | `dfxtp`   | `dfxtp`    | `dffq`         |            |
| `$dff`   | `$_DFF_N_`     |`N`|   |   |   |`P`|           |             |               |           |            | `dffnq`        |            |
| `$dff`   | `$_DFF_P_`     |`P`|   |   |   |`C`| `dfxbp`   |             | `dfxbp`       | `dfxbp`   | `dfxbp`    |                |            |
| `$dffe`  | `$_DFFE_PP_`   |`P`|`P`|   |   |`P`| `edfxtp`  |             | `edfxtp`      |           |            |                |            |
| `$dffe`  | `$_DFFE_PP_`   |`P`|`P`|   |   |`C`| `edfxbp`  |             | `edfxbp`      | `edfxbp`  |            |                |            |
| `$dffsr` | `$_DFF_NN0_`   |`N`|   |   |`N`|`P`| `dfrtn`   |             | `dfrtn`       | `dfrtn`   |            |                |            |
| `$dffsr` | `$_DFF_PN0_`   |`P`|   |   |`N`|`P`| `dfrtp`   | `dfrtp`     | `dfrtp`       | `dfrtp`   | `dfrtp`    | `dffrnq`       | `dfrbpq`   |
| `$dffsr` |                |`P`|   |   |`N`|`C`| `dfrbp`   |             | `dfrbp`       | `dfrbp`   |            |                | `dfrbp`    |
| `$dffsr` | `$_DFF_PN1_`   |`P`|   |`N`|   |`P`| `dfstp`   | `dfstp`     | `dfstp`       | `dfstp`   | `dfstp`    | `dffsnq`       |            |
| `$dffsr` |                |`P`|   |`N`|   |`C`| `dfsbp`   |             | `dfsbp`       | `dfsbp`   | `dfsbp`    |                |            |
| `$dffsr` | `$_DFF_NN0_`   |`N`|   |   |`N`|`P`|           |             |               |           |            | `dffnrnq`      |            |
| `$dffsr` | `$_DFF_NN1_`   |`N`|   |`N`|   |`P`|           |             |               |           |            | `dffnsnq`      |            |
| `$dffsr` | `$_DFFSR_NNN_` |`N`|   |`N`|`N`|`C`| `dfbbn`   |             | `dfbbn`       | `dfbbn`   |            |                |            |
| `$dffsr` | `$_DFFSR_PNN_` |`P`|   |`N`|`N`|`C`| `dfbbp`   |             | `dfbbp`       | `dfbbp`   |            |                |            |
| `$dffsr` | `$_DFFSR_PNN_` |`P`|   |`N`|`N`|`P`|           |             |               |           |            | `dffrsnq`      |            |
| `$dffsr` | `$_DFFSR_NNN_` |`N`|   |`N`|`N`|`P`|           |             |               |           |            | `dffnrsnq`     |            |


| world       | tech              |`G`|`S`|`R`|`Q`| sky130 HD | sky130 HDLL | sky130 [HML]S | sky130 LP | sky130 HVL | gf180mcu [79]t | ihp-sg13g2 |
|-------------|-------------------|---|---|---|---|-----------|-------------|---------------|-----------|------------|----------------|------------|
| `$dlatch`   | `$_DLATCH_P_`     |`P`|   |   |`P`| `dlxtp`   |             | `dlxtp`       | `dlxtp`   | `dlxtp`    | `latq`         | `dlhq`     |
| `$dlatch`   | `$_DLATCH_N_`     |`N`|   |   |`P`| `dlxtn`   | `dlxtn`     | `dlxtn`       | `dlxtn`   |            |                |            |
| `$dlatch`   | `$_DLATCH_P_`     |`P`|   |   |`C`| `dlxbp`   |             | `dlxbp`       | `dlxbp`   |            |                |            |
| `$dlatch`   | `$_DLATCH_N_`     |`N`|   |   |`C`| `dlxbn`   |             | `dlxbn`       | `dlxbn`   |            |                |            |
| `$dlatchsr` | `$_DLATCH_PN0_`   |`P`|   |`N`|`P`| `dlrtp`   | `dlrtp`     | `dlrtp`       | `dlrtp`   |            | `latrnq`       | `dlhrq`    |
| `$dlatchsr` | `$_DLATCH_NN0_`   |`N`|   |`N`|`P`| `dlrtn`   | `dlrtn`     | `dlrtn`       | `dlrtn`   |            |                | `dllrq`    |
| `$dlatchsr` | `$_DLATCH_PN0_`   |`P`|   |`N`|`C`| `dlrbp`   |             | `dlrbp`       | `dlrbp`   |            |                | `dlhr`     |
| `$dlatchsr` | `$_DLATCH_NN0_`   |`N`|   |`N`|`C`| `dlrbn`   |             | `dlrbn`       | `dlrbn`   |            |                | `dllr`     |
| `$dlatchsr` | `$_DLATCH_PN1_`   |`P`|`N`|   |`P`|           |             |               |           |            | `latsnq`       |            |
| `$dlatchsr` | `$_DLATCHSR_PNN_` |`P`|`N`|`N`|`P`|           |             |               |           |            | `latrsnq`      |            |

tests/techmap/dfflibmap.ys
```diff
@@ -23,7 +23,7 @@ read_liberty -lib dfflibmap.lib
 equiv_opt -map dfflibmap-sim.v -assert -multiclock dfflibmap -liberty dfflibmap.lib
 equiv_opt -map dfflibmap-sim.v -assert -multiclock dfflibmap -prepare -liberty dfflibmap.lib
 
-dfflibmap -prepare -liberty dffl*bmap.lib
```
widlarizer, line 26: That was actually on purpose to cover wildcards in liberty file names. If you want a grep to catch the true file name, maybe you can use an equivalent but commented out line instead?

tests/techmap/dlatchlibmap-sim.v
```diff
@@ -0,0 +1,19 @@
+module dlatchn(input ENA, D, output reg Q, output QN);
```
widlarizer, line 1: Motivation for this file over using and extending `tests/sim/dlatch.v` etc ?

tests/techmap/dfflibmap.ys
jeras: @widlarizer I assume this is a typo, since the asterisk is placed in the middle of the abbreviation for LIBerty, which would make a very unusual file naming convention (I did not see any examples of this pattern used to name files). Also the testcase is designed to require this specific file and nothing else. So even if many tests expect multiple files matching a pattern, this test does not.

tests/techmap/dlatchlibmap-sim.v
jeras: @widlarizer The testcases for this LATCH functionality of the `dfflibmap` step are a modified copy of `tests/techmap/dfflibmap*` tests (FF functionality of `dfflibmap` step). As with the FF tests the LATCH tests expect specific contents to be present/absent in the models (not present in `tests/sim/dlatch.v`) to be able to achieve the desired usecase coverage.

An alternative would be to add LATCH tests to existing `tests/techmap/dfflibmap*` testcases. The reason I choose to make copies was that it made it simpler to use a graphical merge tool like `meld` to compare `dlatchlibmap*` to `dfflibmap*` to see what changes were necessary to apply the same tests to LATCHes instead of FFs.

Would you suggest I merge `dlatchlibmap*` into `dfflibmap*`? This would make sense since the `dfflibmap*` step implements both FFs and LATCHes (as does `dfflegalize`).

hexratcc (changes requested):
lgtm, but I'd prefer better testing coverage (H/X priority, mixedpol functional correctness, etc.)

tests/techmap/dfflibmap.ys
widlarizer: Ah, it's not supposed to match multiple files, just that the command is wired into the same wildcard supporting code path

jeras: @widlarizer while I would like to I do not really understand your explanation, but now I assume the wildcard is there for a good reason. So I plan to remove this change.

Should I also use the wildcard in `tests/techmap/dlatchlibmap.ys` which is based on `tests/techmap/dfflibmap.ys`?

jeras:
@hexratcc I unintentionally triggered another review request (this is a part of GitHub UI I am not femiliar with yet).

> lgtm, but I'd prefer better testing coverage (H/X priority, mixedpol functional correctness, etc.)

@hexratcc The LATCH tests are a copy of `dfflibmap` FF tests, so the coverage is the same. If you would like to increase the coverage for LATCHes, it would make sense to also update FF tests.

After checking again what the tests do, I think I understand, what are the extra tests you would like to see. I will make the changes, and write a report.

jeras:
@hexratcc I added a few test, but I also tried to reorganize the files a bit. The current code is still disorganized so I did not commit it yet. Before I decide how best to organize the files, I have questions.

When it comes to `clear`/`preset` priority, the current `dfflegalize` does not insert the logic for changing the priority if the RTL matches a `$_DLATCHSR_*` model (`preset` has priority over `clear`) but not the cell from the Liberty file (while it does insert inverters). Current tests for FF/LATCH mapping check for equivalence under the explicit assumption that `clear`/`preset` are not active at the same time, and without the assumption the tests with mismatched priority are expected to fail. Is this the intended behavior of `dfflegalize`? Will this continue to be the intended behavior in the future? Should this behavior be part of the `dfflegalize` documentation? Would it make sense to add an option to add priority mismatch logic?

hexratcc:
@jeras iirc the internal `$_DLATCHSR_` cells have `clear` prio (`simcells.v` checks `R` first), and the priority info is actually dropped by `dfflibmap`, not `dfflegalize`, as `find_cell_dffsr`/`find_cell_dlatchsr` never read `clear_preset_var`, so a set-dominant cell is silently accepted.

`dfflegalize` shouldn't change: its internal types are all clear-dominant, and it never sees liberty data. I think documenting this is a good idea, but in `dfflibmap` instead. I think an option is a good idea as well (the fix-up already exists in `create_ff`/`create_latch`), and `dfflibmap` could do the same. fwiw the shipped `equiv` tests don't assume mutual exclusion - they do exercise both asserted states and pass because `L` matches the internal priority, so I'd keep your mismatched priority tests as expected fail rather than adding the assumption.

tests/techmap/dlatchlibmap-sim.v
widlarizer: Sounds good, no need

tests/techmap/dfflibmap.ys
widlarizer: Thanks. No, this is fine

widlarizer:
> Is this the intended behavior of dfflegalize?

The intended behavior of dfflegalize is to preserve equivalence. For which we have to know what the semantics of cells are.

> iirc the internal `$_DLATCHSR_` cells have clear prio (simcells.v checks R first)

As with #5194 as long as the priority is unspecified it can't be one way or the other IMO. So it has to be undefined. simcells.v is only forced to pick a priority because verilog is evil. The question is, can we even produce $dlatchsr etc? As far as I can tell, the answer is "only with Verific". With `$dffsr` the model was "fixed" by undefining S&R behavior since `proc` already made that safe. With `$dlatchsr` I can't make that argument. But currently, I'd say those undefined S&R semantics are just missing, and we have to create it. Only then can we have meaningful formal tests for dlatchsr, and until then we can keep them commented out.

jeras:
@hexratcc I have updated the testcases, and ported the same changes to the FF portion of `dfflibmap` tests.

Regarding asynchronous `clear`/`preset` priority. For now I added a comment that if the priority for the Liberty file cell and the RTL do not match, it might result in an equivalence mismatch. This is the same behavior as before, just with an added warning.

I remembered where I have seen an example of handling the `clear`/`preset` priority, it is part of `frontends/liberty`. I could copy the the handling code to `dfflibmap`, but I would like to do it as a separate PR.

@widlarizer my change rewrites your PR #6038. To me it seems correct, but I am not sure what was the purpose of your PR. Could you check it please.

widlarizer:
@jeras #6038 fixes a pre-existing bug that I found in dfflibmap while I was looking at your code. You have to rebase such that the fix is included in your latch code, too

hexratcc (approved):

widlarizer (changes requested):
While reviewing this I noticed another thing that's broken on main. All calls to `dfflibmap` in `tests/techmap/dfflibmap_proc_formal.ys` do nothing. I'll fix that and get back to this. Otherwise, your code changes look good, except for the duplication of code, which I'm happy to refactor myself (it already wasn't great before either)

widlarizer:
Please rebase when #6069 is merged. Sorry for the inconvenience, this is on us. And by us, I mean me. I can take over this PR if you prefer.

jeras:
> Please rebase when #6069 is merged. Sorry for the inconvenience, this is on us. And by us, I mean me. I can take over this PR if you prefer.

I checked #6069 and I do not expect to have much trouble with merge conflicts.
I will also port the `dfflibmap` test changes to `dlatchmap` tests.

> While reviewing this I noticed another thing that's broken on main. All calls to dfflibmap in tests/techmap/dfflibmap_proc_formal.ys do nothing.

I will check, but for for now the tests seem correct to me.
For latches `tests/techmap/dlatchlibmap_proc_formal.ys`, the tests do not really map `DLATCHSR` latches, since as I understand `proc` is not inferring them (I would have to check again to be sure). For now this are just dimple `DLATCH` cells with some extra logic to emulate `SR`. The test should still work if `DLATCHSR` inference was added to `proc`, so I left them in.

> I'll fix that and get back to this. Otherwise, your code changes look good, except for the duplication of code, which I'm happy to refactor myself (it already wasn't great before either)

I have seen a branch with refactored code. My C++ is at a rather basic level, so I was conservative with my changes, and since this is my first code contribution to yosys, I tried to aim for a simple review. I would prefer to leave the refactoring to you.

I already combined the code handling output ports, since I was thinking about adding support for complementary output cells in a future PR. I was thinking about using `extract` (match FF/LATCH and inverter), but my naive approach would probably have bad performance (multiple `extract` calls).

jeras:
@widlarizer

I rebased this patches on `main` (including your recent changes). And I propagated the changes you did to FF code/tests to LATCH code/tests (for best test coverage).

In `liberty.cc` I again copied/modified FF code to LATCH code. I apologize for leaving refactoring the redundancy to you.

I updated the message in your code catching attempts to map _wide_ registers with terminology used in Yosys documentation (_word-level_ instead of _wide_). I was confused, i initially thought _wide_ cells were standard library cells combining multiple FF.

I now understand what was the missing `techmap` issue. I probably missed it while watching `dot` files, because Liberty cells names are so similar to Yosys internal cell names (the only difference is `$`.

I noticed some remaining issues causing overhead, but they are not regressions caused by this PR:
0. When viewing standard cells from a liberty file in `dot` format (using `show`), the outputs are placed on the same left side of the cell block as the inputs (the right side is empty). This is more of a cosmetic problem, but it is confusing during debugging.
1. `proc` adds logic handling preset/clear priority and the logic remains after all mapping steps for FF and LATCHES regardless if the FF standard cell matches the RTL preset/clear priority or not. This adds some redundancy into the netlist.
2. `proc` is indeed not inferring `$dlatchsr`, so if the Liberty file contains only a $_DLATCHSR_***` compatible cell, the `preset/clear` inputs are not connected.
3. The current `d*libmap_proc_formal` testcases create non-optimal netlists. An `opt` step after `techmap` would help. Maybe more `opt` steps would be useful in a synthesis flow, but I am not sure exactly where should I place them.

I might have a look at `proc-latch` to see if I can add inference for `$dlatchsr`.

widlarizer:
If it helps you can add a liberty_ prefix to cell types in `tests/liberty/dff.lib` and similar and adjust tests to match to make them less confusing and more searchable. 

> the outputs are placed on the same left side of the cell block as the inputs 

This is a legendary Yosys footgun - when dfflibmap is used, the types stay unknown to Yosys, and it doesn't know port directions. `read_liberty -lib filename.lib` will add missing blackboxes into the design. This is necessary to continue synthesis without bugs and `hierarchy -check` would report errors on such a design. [Librelane does this](https://github.com/librelane/librelane/blob/e39720fff016c354d822885b2d1acd5508010653/librelane/steps/yosys.py#L101), [ORFS as well](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/da8f092a02a8e75658cc3100691aabff05f35629/flow/scripts/synth_stdcells.tcl#L4).

> clear/preset priority logic

#5613 designed to address that. 

> non-optimal netlists

I'm a bit hesitant to add opt passes into tests like that. It's fine to have a bit of redundancy in the netlist if it means we focus on the pass we're testing and the minimal set of passes needed to test its behavior

Feel free to take a stab at `proc_latch` too!

widlarizer:
Thanks for taking the initiative, superseded by #6072

jeras:
Thanks for taking my first Yosys contribution seriously (I only contributed small documentation fixes before).

PR #5981, opened 2026-06-22
equiv_opt: Add ignore-unknown-cells

KrystalDelusion:

https://yosyshq.discourse.group/t/equiv-opt-on-hierarchical-designs/152 identified that #5666 added the `-ignore-unknown-cells` option to `equiv_induct` but not `equiv_opt`, breaking the ability to call `equiv_opt` on designs with submodules.

Add `-ignore-unknown-cells` to `equiv_opt`.  I wasn't sure if it made sense to use the `EquivBasicConfig` to parse _all_ the available options, given that `-seq` and `-set-assumes` are also not provided, or if that was simply an artifact of them also being on the newer side.  @widlarizer do you have a preference here?

It would still be nice to have some way of supporting recursive `equiv_opt` rather than requiring a flattened design (as pointed out by @donn), but in the interim this does at least maintain the previous behavior of ignoring submodules.

### DIFF
--- a/passes/equiv/equiv_opt.cc
+++ b/passes/equiv/equiv_opt.cc
@@ -69,7 +69,7 @@ struct EquivOptPass:public ScriptPass
 	}
 
 	std::string command, techmap_opts, make_opts;
-	bool assert, undef, multiclock, async2sync, nocheck;
+	bool assert, undef, ignore_unknown_cells, multiclock, async2sync, nocheck;
 
 	void clear_flags() override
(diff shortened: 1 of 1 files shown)


widlarizer (approved):
Since this wraps the commands that I deduplicated the interfaces of with EquivBasicConfig it's fine as-is, since the point was to consistently wire those interfaces into satgen etc (and more things were just convenient to just dunk in there). Thanks

PR #5982, opened 2026-06-23
File cleanup

mmicko:
This PR do lot of cleanups and setup a way for us to maintain this.
Libraries are excluded from checks, since those are synced with external sources from time to time.
What we cover in this PR:
- Remove executable flag from verilog files 
- Fix UTF-8 marker
- Mixed end of file fixes
- End of file fixes
- Trailing whitespaces

This PR also adds configuration for `pre-commit` tool that at least developers could install locally and configure in their pre-commit hook, so we can document that.
Impact on actual source files is not that big, but mostly touch rarely changed files as in examples/tests/docs


KrystalDelusion (approved):
I haven't looked at the changes in detail, but in principle I think this is good.

widlarizer (approved):

PR #5986, opened 2026-06-25
fix case item containing x/z treated as wildcard in proc_rmdead

### LINKED ISSUE #5979: Case item containing X is treated as wildcard and drops reachable branch
### Version

Yosys 0.66+4 (git sha1 8125af8, g++ 13.3.0-6ubuntu2~24.04.1 -fPIC -O3)

Linux

### Reproduction Steps

Yosys synthesis command:

```bash
```

### Verilog before synthesis

Input RTL (`rtl.v`):

```verilog
module top (
  input wire [1:0] sel,
  input wire [3:0] a,
  input wire [3:0] b,
  output reg [3:0] y
);
  always @* begin
    case (sel)
      2'b1x: begin
        y = a;
      end
      2'b10: begin
        y = b;
      end
      default: begin
        y = a;
      end
    endcase
  end
endmodule
```

Important minimization note: this must be an ordinary Verilog `case`.Changing it to `casez` or `casex` gives `x` wildcard semantics and does not represent the same issue. The concrete `2'b10` arm must also remain after the `2'b1x` arm and assign a distinguishable value, otherwise the reachable-branch drop is hidden.

### Verilog after synthesis

```verilog
/* Generated by Yosys 0.66+4 (git sha1 8125af88d, g++ 13.3.0-6ubuntu2~24.04.1 -fPIC -O3) */

(* top =  1  *)
(* src = "rtl.v:1.1-20.10" *)
module top_synth(sel, a, b, y);
  (* src = "rtl.v:2.20-2.23" *)
  input [1:0] sel;
  wire [1:0] sel;
  (* src = "rtl.v:3.20-3.21" *)
  input [3:0] a;
  wire [3:0] a;
  (* src = "rtl.v:4.20-4.21" *)
  input [3:0] b;
  wire [3:0] b;
  (* src = "rtl.v:5.20-5.21" *)
  output [3:0] y;
  wire [3:0] y;
  assign y = a;
endmodule
```

The relevant change is that the reachable RTL branch:

```verilog
2'b10: begin
  y = b;
end
```

is dropped entirely, leaving:

```verilog
assign y = a;
```

This drops the dependence on `b` when `sel == 2'b10`.

## Step-by-step Analysis

The source RTL uses an ordinary Verilog `case`, where `x` bits in case items are not wildcards. Therefore `2'b1x` should not cover the known selector value `2'b10`; the later concrete arm is still reachable and should assign `y = b`.

```verilog
case (sel)
  2'b1x: y = a;
  2'b10: y = b;
  default: y = a;
endcase
```

The `write_rtlil` result before dead-branch removal still contains both the `2'b1x` arm and the reachable `2'b10` arm:

```rtlil
switch \sel
  case 2'1x
    assign $1\y[3:0] \a
  case 2'10
    assign $1\y[3:0] \b
  case
    assign $1\y[3:0] \a
end
```

After `proc_rmdead`, `write_rtlil` shows that the `2'b10` arm has been removed and the switch has been marked as `full_case`:

```rtlil
attribute \full_case 1
switch \sel
  case 2'1x
    assign $1\y[3:0] \a
  case
    assign $1\y[3:0] \a
end
```

The Yosys log reports the same change:

```text
Executing PROC_RMDEAD pass (remove dead branches from decision trees).
Removed 1 dead cases from process $proc$rtl.v:7$1 in module top.
Marked 1 switch rules as full_case in process $proc$rtl.v:7$1 in module top.
```

This indicates that the behavior changes in `proc_rmdead`, before mux generation or later optimization. The likely cause is that the dead-branch analysis treats the `x` bit in `2'1x` as a don't-care pattern, so it incorrectly considers `2'b10` covered by the earlier arm. After that deletion, the only remaining assignments are `y = a`, so the design collapses to:

```verilog
assign y = a;
```

This changes the synthesized behavior for:

```text
sel = 2'b10, a = 4'd10, b = 4'd3
```

For these inputs, the source RTL selects the ordinary `case` arm that assigns `y = b`, while the synthesized netlist assigns `y = a` unconditionally.

### Expected Behavior

For an ordinary Verilog `case`, `2'b1x` is not a wildcard pattern. The known selector value `sel = 2'b10` should reach the concrete `2'b10` arm:

```verilog
y = b;
```

With `a = 4'd10` and `b = 4'd3`, the synthesized netlist should drive:

```text
sel=10 y=3
```

### Actual Behavior

```verilog
assign y = a;
```

This is not equivalent to the RTL because it returns `a` when `sel = 2'b10`, while the RTL returns `b`.

hexratcc:
This PR aims to fix #5979 - in a normal verilog `case`, x/z bits in a case label are literals using 4-state eq, instead of dc's. `BitPatternPool::sig2bits` collapsed all non 0/1 states (`Sx`, `Sz`, `Sa`) to the don't-care `Sa`, so `2'b1x` wrongly covered `2'b10`/`2'b11`, which made `proc_rmdead` delete reachable branches (collapsing the design to `assign y = a`). This PR makes it so that only `Sa` is treated as a wildcard (as `Sx`/`Sz` in a query pattern with them covers nothing).


hexratcc:
[sc-621]

kernel/bitpattern.h
```diff
@@ -102,11 +102,19 @@ struct BitPatternPool
 		bits_t bits;
 		bits.bitdata = sig.as_const().to_bits();
 		for (auto &b : bits.bitdata)
-			if (b > RTLIL::State::S1)
+			if (b > RTLIL::State::S1 && b != RTLIL::State::Sx && b != RTLIL::State::Sz)
 				b = RTLIL::State::Sa;
```
KrystalDelusion: Two things, one is that this could be reduced to `if (b == RTLIL::State::Sm)`. The second is that if we're not changing `Sx` or `Sz`, then the comment on the function (that we're changing to "three-valued logic") is incorrect. And while we don't appear to have any users of this function that require it to actually be three-value logic, I wonder if there might be downstream users that this affects. Since `covers_nothing` is always the first thing called, maybe it would be better to have it take the `RTLIL::SigSpec sig` instead?

KrystalDelusion (commented):
May be worth adding x/z cases to `tests/unit/kernel/bitpatternTest.cc`.  e.g. `pool({011}).has_any(01x) == false` and `pool({011}).has_all(01x) == true`.

tests/verilog/temp/issue4402_syn.v
KrystalDelusion, line 1: Is this supposed to be committed?

KrystalDelusion (commented):
Thanks for the changes! Looks good except for one file that I'm not sure is supposed to be there?

tests/verilog/temp/issue4402_syn.v
hexratcc: Nope, my bad :smile:.

KrystalDelusion (approved):

PR #5988, opened 2026-06-26
CI cleanups and improvements

mmicko:
- Added ccache for compiling, getting great speedup
- Moved to latest versions of actions
- Used forked version of skip-duplicate-action to prevent lot of warnings
- CodeQL now executed once per week

Note: Still waiting for solution for cpp-action but this already gives us a lot


KrystalDelusion:
FYI that https://github.com/fkirc/skip-duplicate-actions has now pushed a new release for nodejs 24

PR #5997, opened 2026-06-30
FABulous Updates

mole99:
This PR updates the FABulous techlib to remove all hardcoded mappings and primitives, allowing them to be specified via the options of `synth_fabulous`. This makes it much easier to fully customize your fabric, without the need to recompile Yosys. It also means the CAD setup for your fabric can stay next to your fabric database.

I tried to keep the changes separate and in small bite-sized commits for easier review.
The tests have been updated and show that the old functionality can still be matched.

The new [fabulous-tiles](https://github.com/FPGA-Research/fabulous-tiles), which will become the default tile repository after FABulous 2.0, already contains all of the necessary files.
Both the fabric in [HeiChips 2025](https://github.com/FPGA-Research/heichips25-tapeout) and the fabric in the upcoming HeiChips 2026 already make use of the new setup.

### Compatibility with older options

To match the functionality for older fabrics, pass the following options to `synth_fabulous`: `-noiopad -ff $_DFF_P_ x -ff $_DLATCH_?_ x -extra-plib prims.v -arith-map arith_map.v -extra-map ff_map.v -cells-map cells_map.v -extra-map latches_map.v -extra-mlibmap ram_regfile.txt -extra-map regfile_map.v`
This will disable I/O mapping and use a simple DFF and latch only, while adding the primitives and mapping files.
To match the legacy `-complex-dff` option, use: `-ff $_DFF_P_ x -ff $_DFFE_PP_ x -ff $_SDFF_PP?_ x -ff $_SDFFCE_PP?P_ x -ff $_DLATCH_?_ x`

This may seem like a handful, but keep in mind that each future fabric (i.e. from [fabulous-fabrics](https://github.com/FPGA-Research/fabulous-fabrics)) comes with a proper setup that can easily be adapted.

cc @KelvinChung2000


tests/arch/fabulous/carry.ys
```diff
@@ -1,8 +1,9 @@
 read_verilog ../common/add_sub.v
 hierarchy -top top
 proc
-equiv_opt -assert -map +/fabulous/prims.v synth_fabulous -carry ha # equivalency check
+equiv_opt -assert -map prims.v synth_fabulous -carry ha -noiopad -ff $_DFF_P_ x -ff $_DLATCH_?_ x -extra-plib prims.v -cells-map cells_map.v  -arith-map arith_map.v -extra-map ff_map.v -extra-map latches_map.v  # equivalency check
 design -load postopt # load the post-opt design (otherwise equiv_opt loads the pre-opt design)
+show -format dot top
```
KelvinChung2000: I assume this is for debug?

techlibs/fabulous/synth_fabulous.cc
```diff
@@ -378,27 +416,38 @@ struct SynthPass : public ScriptPass
...
 		if (check_label("check")) {
 			run("hierarchy -check");
 			run("stat");
 		}
 
-		if (check_label("blif"))
-		{
-			if (!blif_file.empty() || help_mode)
-			{
+		if (check_label("blif")) {
```
KelvinChung2000: we can remove blif support. More accurately speaking we are not using it any way.

techlibs/fabulous/synth_fabulous.cc
```diff
@@ -312,6 +316,27 @@ struct SynthPass : public ScriptPass
...
 				run(stringf("techmap -map +/cmp2lut.v -map +/cmp2lcu.v -D LUT_WIDTH=%d", lut));
+			if (help_mode || multiplier_map != "") {
+				run("wreduce t:$mul");
+				if (help_mode) {
+					run("techmap -map +/mul2dsp.v -map <multiplier_map> -D DSP_A_MAXWIDTH=<a_max> -D DSP_B_MAXWIDTH=<b_max> "
+					    "-D DSP_A_MINWIDTH=<a_min> -D DSP_B_MINWIDTH=<b_min> -D DSP_Y_MINWIDTH=<y_min> "
+					    "-D DSP_NAME=$__FABULOUS_MUL",
+					    "(if -multiplier-map)");
+				} else {
+					run(stringf("techmap -map +/mul2dsp.v -map %s -D DSP_A_MAXWIDTH=%d -D DSP_B_MAXWIDTH=%d "
```
KelvinChung2000, line 318: For full dsp mapping, I assume we need to use the wrap-extract-map pattern. If you are just looking for replacing `$MUL` would it be better just use the arith map?

techlibs/fabulous/synth_fabulous.cc
```diff
@@ -321,12 +346,13 @@ struct SynthPass : public ScriptPass
 			run("opt_clean");
 		}
 
-		if (check_label("map_ram", "(unless -noregfile)")) {
-			// RegFile extraction
-			if (!noregfile) {
-				run("memory_libmap -lib +/fabulous/ram_regfile.txt");
-				run("techmap -map +/fabulous/regfile_map.v");
-			}
+		if (check_label("map_memory")) {
```
KelvinChung2000, line 340: for the full memory mapping sequence should it be something like:
```
memory -nomap
memory_libmap -lib <lib-map>
techmap -map memory_map.v
memory_map
```

Which I assume can be merged with the `map_ffram` below?

tests/arch/fabulous/carry.ys
mole99: Good catch!

techlibs/fabulous/synth_fabulous.cc
mole99: Done.

techlibs/fabulous/synth_fabulous.cc
mole99: Yes, here I'm only looking to map to hardware multipliers of the specified size.
To do this, we first need to decompose the `$mul`s to a size that is supported by the hardware. `mul2dsp.v` does just that. In our case, we ignore the `$shl` and `$add` cells (if there are any). `synth_ice40` does it the same way, though there is a custom pass to further map `$mul/SB_MAC16 + $add`.

techlibs/fabulous/synth_fabulous.cc
mole99: Yes, we do all of these steps. In our case, the last two lines are swapped, which doesn't matter since `memory_libmap` ran just before.

I would keep it separate, so we have labels for `map_memory` and `map_ffram`.
`techmap` is run as part of `map_extra`.

mole99:
I would appreciate some help from the Yosys team to fix the verific fails.

mmicko:
@mole99 It is nothing verific specific. Thing is you have moved files from techlib and put them in tests, for you locally tests are working since in your build directory there is still `fabulous` sub directory with old files so tests are able to find them, just do delete that and do compile again (it should not be created), when you run tests again you will see them failing since they address files not existing.

mmicko:
To fix it just change `-map +/fabulous/prims.v` to `-map prims.v` in your tests

mole99:
@mmicko Thank you! I didn't think of that ^^

I've updated the tests and they are clean now.

mole99:
@KelvinChung2000 If there are no further comments, I suppose the PR is ready to merge from our side.

mmicko:
@mole99 Was thinking about approach used here and it seems quite complicated for a user to need to specify all given mappings. Not sure how if you have many different implementations, but if you have at least some standard (like `v1` and `v2`) there can be something like `-family` parameter to specify mappings and use those (you can form file used like `+/fabulous/%s/cells_map.v` and use family as parameter in forming path, but also can and should validate existence of family name at start (lattice, quicklogic and nanoxplore use this for reference). If it is possible that use have a need to specify some custom family you can always leave these options, or make them able to override existing family so then you would need to specify just those files that changed compared to original architecture.

mole99:
@mmiko thanks for the comments!

The only purpose of v1 and v2 is to distinguish between the legacy file formats of v1 (and should probably be removed in the near future). Please bear in mind that there are no families, as FABulous allows to fully customize your FPGA. The closest thing that comes to it are the tile libraries, but even those can be extended with custom tiles.

Therefore, I found that the most straightforward way, without imposing restrictions on the person developing the fabric, is to specify each of the files individually. This doesn't prevent you from using wildcards one level above, e.g. in your Makefile or other setup.

Anyone who simply wants to use a FABulous FPGA will never have to specify these files to Yosys, since the fabric already comes with its proper setup: https://github.com/FPGA-Research/fabulous-fabrics/tree/main/user_designs
You simply run `make all` to build all examples.

mmicko:
@mole99 I see, clear now. Was just  wondering if there was some cleaner approach.

KelvinChung2000:
I don't like having so many different maps and configs. It's just too much cognitive load. The "family" idea is something I have been thinking about as well. As I have been thinking, including all the customisation capability in Yosys might not be ideal. Since if the user is deeply customising beyond the standard tile library, they will need to engage with the synthesis process anyway; a generic template is too rigid and unlikely to capture all the user's needs. 

I think the best approach is for FABulous to gather all the maps and cells into a few files that an architecture is using and for the Yosys upstream script to account only for the "standard primitives" we provide. We might offer a "family" feature that performs ABC tuning for a specific routing graph we provide for our "template" architecture. But right now, FABulous cannot yet gather multiple maps from different places, nor does it have any "tuning needs". Setting a bunch of maps as parameters is a "fine solution" for now.

I have been thinking of letting FABulous "own" more of the synthesis responsibility, such as by generating an architecture-specific synthesis script with ABC tuning + all the "extraction" features. This is possible for us because the tool can access the BEL and the routing graph implementation. Another advantage is that our feature development can proceed independently with each Yosys release, giving us more flexibility.

KelvinChung2000 (approved):

mole99:
This may eventually be the right approach, though you need to a full set of "standard primitives". One of the issues this PR solves is that the current standard primitives for flip-flops only support enable before reset.

For now, this PR expands the capabilities of `synth_fabulous` for custom fabrics such as HeiChips or the Tiny Tapeout series so one can specify custom memory mappings and insert global buffers automatically.

Ravenslofty:
(I asked Leo to update CODEOWNERS, since the FABulous group is going to be testing downstream.)

Ravenslofty (approved):

PR #6001, opened 2026-07-04
opt_reduce: derive pmux helper names from source cells

es3298:
## Summary

Add `RTLIL::Module::derive_id()` for generating a unique name from an existing object name plus a descriptive suffix, and use it for the helper `$reduce_or` cell and wire created by `opt_reduce` when it merges duplicate `$pmux` select cases.

This keeps generated helper objects tied to the source `$pmux` name instead of assigning opaque `$auto$...` names. The helper is intentionally small so other passes can adopt the same pattern incrementally.

## Testing

Built locally on Windows with CMake/MSBuild, configured with ABC, Tcl, readline/editline, zlib, and Python disabled.

Ran:

- `build-es3298\Release\yosys.exe -q tests\opt\opt_reduce_pmux_names.ys`
- `build-es3298\Release\yosys.exe -q tests\opt\opt_reduce_bmux.ys`
- `build-es3298\Release\yosys.exe -q tests\opt\opt_reduce_andor.ys`

### DIFF
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -3053,6 +3053,15 @@ RTLIL::IdString RTLIL::Module::uniquify(RTLIL::IdString name)
 	return uniquify(name, index);
 }
 
+RTLIL::IdString RTLIL::Module::derive_id(RTLIL::IdString base, std::string_view suffix)
+{
+	log_assert(!suffix.empty());
+	std::string name = base.str();
+	name += "_";
+	name += suffix;
+	return uniquify(name);
+}
+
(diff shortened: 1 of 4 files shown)


whitequark (commented):
One issue here is that you're adding new public wires to a design, which will show up in debug tools like waveform viewers if you're debugging a post-synthesis netlist.

I think making automatically generated wire IDs more readable is a great goal, but I believe these IDs need to be internal names (so instead of `\\WIRE_SUFFIX` something like `$SUFFIX$WIRE`.

es3298:
Closing this draft. I’m going to rework this.

Issue #5394, opened 2025-09-26
autoname may be inefficient for large circuits


KrystalDelusion:
### Version

0.57

### On which OS did this happen?

Linux

### Reproduction Steps

Spinning out from #5277:

> The current approach (which I haven't changed) is to check against double the best score. If the best name is more than twice as good as the current proposed name, we delay checking until the next iteration. Since the best score will always be assigned, the next iteration can now use that newly named object for naming other cells/wires. In the worst case this can however mean that only 1 name gets applied in each iteration, and there is no logic to check if a given proposal could actually be affected by another proposal being applied (meaning that it can take extra iterations without producing a different result). I think the way to do it would be that each wire/cell checks all of the connections, recording the best connection for that wire/cell separated into public/private scores, and then will only skip if there is a better private score and it isn't the best score available.

I'm not sure how much performance impact this has in practice, but I thought I would at least open an issue so that there is some record.

### Expected Behavior

Wires/cells delay renaming only if they have a private connection which may be better than their best public connection.

### Actual Behavior

Wires/cells delay renaming if any private connection has a score less than half their best public connection.  This means that an object *with only public connections* may still delay renaming, which in turn delays any objects waiting for it to be renamed.

Issue #4509, opened 2024-07-25
Autoname seems to get stuck in a loop and consume all the memory on the system


QuantamHD:
### Version

557968567360026ae8dc75341564209868171296

### On which OS did this happen?

Linux

### Reproduction Steps

I've been having a hard time reducing a test case, the test case crashes my remote desktop connection. 

Reviewing the code a bit suggests that maybe this line is responsible
```c++
			int count = 0, iter = 0;
			while (1) {
				iter++;
				int n = autoname_worker(module, wire_score);
				if (!n) break;
				count += n;
			}
```

### Expected Behavior

Not get stuck in a loop

### Actual Behavior

Autoname gets stuck in a loop.

Issue #2816, opened 2021-06-07
`autoname` memory usage very high with CPU generated from `ghdl-yosys-plugin`.


cr1901:
## Steps to reproduce the issue

When specifically trying to synthesize a [Microwatt CPU](https://github.com/antonblanchard/microwatt) SoC using `yosys`, I've noticed that the `autoname` pass takes an unusual amount of memory that I can't seem to duplicate with other designs. Microwatt is a large CPU, and while I can't easily synthesize another SoC quite as big as that one at present, I've noticed that a SoC about half the size uses 1/4 of the memory, and `autoname` is not where the most time is spent.

Here is an example run of `yosys` trying to synthesize the Microwatt CPU using `ghdl-yosys-plugin`:

```
6.48. Executing AUTONAME pass.
Renamed 2309229 objects in module gsd_orangecrab (484 iterations).
<suppressed ~63121 debug messages>

6.49. Executing HIERARCHY pass (managing design hierarchy).

6.49.1. Analyzing design hierarchy..
Top module:  \gsd_orangecrab

6.49.2. Analyzing design hierarchy..
Top module:  \gsd_orangecrab
Removed 0 unused modules.

6.50. Printing statistics.

=== gsd_orangecrab ===

   Number of wires:              30195
   Number of wire bits:         176365
   Number of public wires:       30195
   Number of public wire bits:  176365
   Number of memories:               0
   Number of memory bits:            0
   Number of processes:              0
   Number of cells:              41188
     CCU2C                        1409
     CLKDIVF                         1
     DDRDLLA                         1
     DELAYG                         43
     DP16KD                         38
     DQSBUFM                         2
     ECLKBRIDGECS                    1
     ECLKSYNCB                       1
     EHXPLLL                         2
     IDDRX2DQA                      16
     L6MUX21                      1765
     LUT4                        23896
     MULT18X18D                     16
     ODDRX2DQA                      18
     ODDRX2DQSB                      2
     ODDRX2F                        27
     PDPW16KD                       16
     PFUMX                        5921
     TRELLIS_DPR16X4               331
     TRELLIS_FF                   7644
     TRELLIS_IO                     20
     TSHX2DQA                       16
     TSHX2DQSA                       2

6.51. Executing CHECK pass (checking for obvious problems).
Checking module gsd_orangecrab...
Found and reported 0 problems.

6.52. Executing JSON backend.

Warnings: 4 unique messages, 4 total
[...]

PR #6003, opened 2026-07-06
autoname: avoid O(iterations x module size) full rescan

gonsolo:
## Summary

`AutonamePass::execute()` calls `autoname_worker()` in a loop, and every call rescans every selected cell's every connection's every bit from scratch. Naming only propagates one hop per call (an object becomes nameable only once a directly connected neighbor already has a public name), so the number of iterations scales with the propagation depth through the design's connectivity graph. On a large, fully flattened netlist with long dependency chains this becomes `O(iterations x module_size)`.

We hit this synthesizing an RV64 CPU + 4-lane SIMT GPU SoC for ECP5 (`synth_ecp5`'s `check:` stage runs `autoname`): 40+ GB RSS and 30+ minutes with no end in sight, on a ~500k-wire post-techmap netlist. A `perf record` profile of the running process showed ~73% of CPU time in `autoname_worker` / `Module::selected_cells()` / `IdString::operator[]`, confirming the rescan is the bottleneck rather than anything design-specific.

This looks like the same underlying issue reported in #5394, #4509, and #2816.

## What this changes

Replaces the full-rescan loop with a persistent worklist: `pending_cells`/`pending_wires` hold the current best proposal only for `$`-named objects that are *currently* reachable from a public name (objects with no such path never enter these maps at all, unlike a naive "still unresolved" set would). Recomputing an object's proposal only happens when one of its direct neighbors was just renamed, via precomputed wire<->cell adjacency built once up front, instead of rescanning the whole module every round.

This preserves the original's round-by-round "accept everything within 2x of this round's best" batching behavior, so the final naming is unchanged — verified against `tests/various/autoname.ys`, including the "don't rename prematurely" ordering test that specifically exercises this scheduling behavior.

One subtlety found along the way: `Module::selected(cell)` looks a cell up by its *current* name, so a cell renamed earlier in the same `autoname` invocation can fall out of a name/type-based selection (e.g. `autoname t:$or`) that matched it before the rename. The original algorithm observes this naturally because it calls `module->selected_cells()` fresh every round; this patch re-checks `module->selected(cell)` at the point a wire proposal is (re)computed to reproduce the same behavior without a full rescan. (`tests/various/autoname.ys`'s "wires get autonames too" test exercises exactly this case.)

## Test plan

- [x] `tests/various/autoname.ys` passes unchanged (all expected log messages/rename counts match, including the selection-staleness edge case)
- [x] Re-ran the ECP5 synthesis flow that previously never completed (40+ GB RSS, 30+ min with no end in sight) — now proceeds through `autoname` with stable, modest (~150-200MB) memory use and completes
- [x] `hardware.borg`/CPU Chisel test suites for the design that triggered this (unrelated to yosys itself, just confirming the underlying RTL was unaffected)

Happy to add a regression test to `tests/various/` covering a large/deep design shape if that's useful — let me know what form would be most useful (e.g. a generated wide+deep chain of anonymous cells).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/passes/cmds/autoname.cc
+++ b/passes/cmds/autoname.cc
@@ -35,76 +35,69 @@ typedef struct name_proposal {
 	}
 } name_proposal;
 
-int autoname_worker(Module *module, const dict<Wire*, unsigned int>& wire_score)
+// Recompute the current best name proposal for a single $-named cell, using
+// only that cell's own connections (bounded by its arity, not module size).
+// Mirrors the per-cell inner loop of the original single-pass algorithm.
+bool recompute_cell_proposal(Cell *cell, const dict<Wire*, unsigned int> &wire_score, name_proposal &out)
 {
-	dict<Cell*, name_proposal> proposed_cell_names;
-	dict<Wire*, name_proposal> proposed_wire_names;
-	name_proposal best_name;
-
-	for (auto cell : module->selected_cells()) {
-		if (cell->name[0] == '$') {
-			for (auto &conn : cell->connections()) {
-				string suffix;
-				for (auto bit : conn.second)
-					if (bit.wire != nullptr && bit.wire->name[0] != '$') {
-						if (suffix.empty())
-							suffix = stringf("_%s_%s", cell->type.unescape(), conn.first.unescape());
-						name_proposal proposed_name(
-							bit.wire->name.str() + suffix,
-							cell->output(conn.first) ? 0 : wire_score.at(bit.wire)
-						);
-						if (!proposed_cell_names.count(cell) || proposed_name < proposed_cell_names.at(cell)) {
-							if (proposed_name < best_name)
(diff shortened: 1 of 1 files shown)


whitequark:
> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

These PRs are not welcome.

gonsolo:
> These PRs are not welcome.

Ok. @whitequark, are you willing to take a patch manually written by me? I'm willing to invest some time in this because it fixes are real bug in yosys for my (larghish) design; autoname blows memory up so yosys doesn't finish on my 64GB machine. With autoname fixed it stays benign under 2GB.

whitequark:
> @whitequark, are you willing to take a patch manually written by me?

Yep, me or another team member will happily review that.

PR #6004, opened 2026-07-07
Propagate src attributes in opt_share

es3298:
Propagates `src` attributes through helper cells created by `opt_share`.

`opt_share` creates helper mux/pmux/neg cells while merging mutually exclusive operators. Those helper cells now inherit the `src` attribute from the mux that drives the sharing opportunity.

Test added:

- `tests/opt/opt_share_src.ys`

Validation run:

- `cmake --build ~/yosys-src-prop-build --target yosys --parallel 2`
- all 15 `tests/opt/opt_share*.ys` scripts


es3298:
It looks like the GitHub Actions runs are waiting for maintainer approval (action_required). Could someone approve the workflows when convenient?

widlarizer (changes requested):
strpool src attributes are inefficient from a performance standpoint and we aren't planning on using of that mechanism anymore. Instead I've been working for some time on [entirely new infrastructure](https://yosyshq.discourse.group/t/consistent-source-location-transfer-with-patch/121) to address this. The opt_share change seems fine. Drop the muxpack part and I'll approve this.

By the way, am I talking to an LLM agent? Is there an actual human who wants this patch? Is this for a bounty?

es3298:
hi hi human here and yes this is upstreaming the feature as part of a bounty. removed the muxpack part btw thanks!

PR #6006, opened 2026-07-07
Add slang information in README

mmicko:
This PR adds links to `sv-elab` and `slang` repositories to promote their usage.

Also it removes tip to use `git clone` with `--recurse-submodules` since it would pull modules from sv-elab that are not used, and with just first level of submodules all is functioning fine.


povik (approved):
Looks good to me. If you wanted simpler wording "Yosys is using sv-elab and slang libraries to [...]" would be fine with me too.

nakengelhardt (approved):

PR #6007, opened 2026-07-07
Fix wide shift count folded as truncated/signed negative count

### LINKED ISSUE #5996: Wide shift count is folded as truncated/signed negative count
### Version

Yosys 0.65+51 (git sha1 5c6de0446, g++ 13.3.0-6ubuntu2~24.04.1 -fPIC -O3)

Linux

### Yosys synthesis command

```bash
```

### Verilog before synthesis

Input RTL (`rtl.v`):

```verilog
module top(input [7:0] s, output [7:0] y);
  assign y = s >> 33'h0ffffffff;
endmodule
```

The shift amount is 33 bits wide. For an 8-bit left operand, this shift amount is far larger than the operand width, so the logical right shift should deterministically produce zero for all known input values.

### Verilog after synthesis

```verilog
/* Generated by Yosys 0.65+51 (git sha1 5c6de0446, g++ 13.3.0-6ubuntu2~24.04.1 -fPIC -O3) */

(* top =  1  *)
(* src = "rtl.v:1.1-3.10" *)
module top_synth(s, y);
  (* src = "rtl.v:1.24-1.25" *)
  input [7:0] s;
  wire [7:0] s;
  (* src = "rtl.v:1.40-1.41" *)
  output [7:0] y;
  wire [7:0] y;
  assign y = { s[6:0], 1'h0 };
endmodule
```

The relevant change is that the deterministic RTL expression:

```verilog
assign y = s >> 33'h0ffffffff;
```

is synthesized as:

```verilog
assign y = { s[6:0], 1'h0 };
```

This turns a huge logical right shift that should produce zero into a left shift by one.

### Step-by-step Analysis

The RTL computes a logical right shift by a very large count:

```verilog
assign y = s >> 33'h0ffffffff;
```

The left operand `s` is 8 bits wide. The RHS is a 33-bit constant with numeric value `0xffffffff`, which is far larger than the operand width. For a logical right shift, shifting an 8-bit value by that amount should shift out all original bits and fill with zeros:

```text
s >> 33'h0ffffffff == 8'h00
```

So the whole design should be equivalent to:

```verilog
assign y = 8'h00;
```

Instead, Yosys synthesizes the design as:

```verilog
assign y = { s[6:0], 1'h0 };
```

That is a left shift by one, not a huge logical right shift. This is consistent with the 33-bit shift count being mishandled as if it were truncated to a signed 32-bit value, where `32'hffffffff` would behave like `-1`. The simulation mismatch is:

```text
MISMATCH s=01 rtl=00 synth=02
MISMATCH s=11 rtl=00 synth=22
```

The RTLIL milestone trace shows where the bad fold appears.

`after_hierarchy.il`:

```text
cell $shr $shr$rtl.v:2$1
  parameter \A_SIGNED 0
  parameter \B_SIGNED 0
  parameter \A_WIDTH 8
  parameter \B_WIDTH 33
  parameter \Y_WIDTH 8
  connect \A \s
  connect \B 33'011111111111111111111111111111111
  connect \Y $shr$rtl.v:2$1_Y
end
connect \y $shr$rtl.v:2$1_Y
```

`after_proc.il`:

```text
connect \y $shr$rtl.v:2$1_Y
connect $shr$rtl.v:2$1_Y { \s [6:0] 1'0 }
```

`after_techmap_opt.il`:

```text
connect \y { \s [6:0] 1'0 }
```

The key bad transition is visible by the first RTLIL milestone after `proc`: the huge logical right shift has already been folded into a left shift by one.

### Additional related reproducer

The same root cause also reproduces with `33'h0fffffffe`. This variant is useful because it shows that the bug is not limited to the all-ones RHS: the huge left shift is folded as if the count were truncated to signed `-2`, producing a right shift by two.

Input RTL:

```verilog
module bug19_min(input [7:0] s, output [7:0] y);
  assign y = s << 33'h0fffffffe;
endmodule
```

Key synthesized shape:

```verilog
assign y = { 2'h0, s[7:2] };
```

The left shift count is still far larger than the 8-bit operand width, so the result should be zero. Instead, Yosys turns it into a right shift by two.

Example mismatch:

```text
MISMATCH s=04 rtl=00 synth=01
```

## 

### Expected Behavior

A logical right shift by a count much larger than the input width should fully shift out the 8-bit input and produce zero:

```verilog
assign y = 8'h00;
```

The related left-shift variants should similarly produce zero when the shift count is larger than the left operand width. The arithmetic-right-shift variant should preserve sign-fill behavior.

### Actual Behavior

Yosys synthesizes the main reproducer as:

```verilog
assign y = { s[6:0], 1'h0 };
```

This replaces a huge logical right shift with a left shift by one. Related variants show the same 33+ bit shift count being treated like a truncated signed negative count, producing reverse small shifts or incorrect condition logic.

hexratcc:
This PR aims to fix #5996 - `Const::convertible_to_int` checked `back()` (full width MSB) instead of the minimal representation's MSB, so a 33-bit count `0x0ffffffff` was read as signed -1 and `s >> 33'h0ffffffff` folded to `{ s[6:0], 1'h0 }` instead of 0. This PR replaces the `back()` with `(*this)[size - 1]`.

### DIFF
--- a/kernel/rtlil.cc
+++ b/kernel/rtlil.cc
@@ -588,7 +588,7 @@ bool RTLIL::Const::convertible_to_int(bool is_signed) const
 	if (size == 32) {
 		if (is_signed)
 			return true;
-		return back() != State::S1;
+		return (*this)[size - 1] != State::S1;
 	}
 
 	return false;
--- a/tests/opt/opt_expr_shift.ys
+++ b/tests/opt/opt_expr_shift.ys
@@ -121,6 +121,36 @@ module top (
 	output wire        [7:0]  shr,
 	output wire        [7:0] sshl,
 	output wire        [7:0] sshr,
+);
+	assign  shl = in << 33'h0ffffffff;
+	assign  shr = in >> 33'h0ffffffff;
+	assign sshl = in <<< 33'h0ffffffff;
+	assign sshr = in >>> 33'h0ffffffff;
+endmodule
+EOT
+
+select -assert-count 1 t:$shl
+select -assert-count 1 t:$shr
+select -assert-count 1 t:$sshl
+select -assert-count 1 t:$sshr
+
+equiv_opt -assert opt_expr
+
+design -load postopt
+select -assert-none t:$shl
+select -assert-none t:$shr
+select -assert-none t:$sshl
+select -assert-none t:$sshr
+
+design -reset
+
+read_verilog <<EOT
+module top (
+	input  wire        [3:0]  in,
+	output wire        [7:0]  shl,
+	output wire        [7:0]  shr,
+	output wire        [7:0] sshl,
+	output wire        [7:0] sshr,
 	output wire        [7:0] shiftx,
 
 	output wire        [7:0]  shl_s,
--- a/tests/unit/kernel/rtlilTest.cc
+++ b/tests/unit/kernel/rtlilTest.cc
@@ -1,4 +1,5 @@
 #include <gtest/gtest.h>
+#include <limits>
 #include "kernel/rtlil.h"
 
 YOSYS_NAMESPACE_BEGIN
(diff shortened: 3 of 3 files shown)


hexratcc:
[sc-629]

Ravenslofty (approved):

PR #6008, opened 2026-07-07
Bump sv-elab to 2026-07-07

povik:

This does not bring in any code change but switches sv-elab to a tagged revision for the sake of packaging. I'm not sure if you plan to bundle sv-elab sources with yosys in the release archive, or expect downstream consumers to source the two independently. The tag can be useful in the latter case.


mmicko:
Thanks, will merge when CI gets green.

Have added all required submodules and referenced them directly and enforced that cmake is not fetching dependencies but using those distributed.
Also tar is created as in https://github.com/YosysHQ/yosys/actions/runs/28870729123?pr=6008 and you can always download vendored-sources, and I do upload that as artifact for Yosys release,  so yosys can be built without a need for internet access.

PR #6022, opened 2026-07-09
autoname: avoid O(iterations x module size) full rescan

gonsolo:
@whitequark: Note that I still used Claude to explain me things but I made sure I understand what's going on. I hope that's acceptable.

## Summary

`AutonamePass` rescanned every selected cell/wire from scratch on every
iteration, and naming only propagates one hop per iteration, so cost
scaled with `O(iterations x module_size)`. On a large, fully flattened
netlist with long dependency chains this meant multi-hour runtime and
40+ GB RSS. Likely the same issue as #5394, #4509 and #2816.

- Replace the full-rescan loop with a `ModuleAutonamer` that keeps a
  persistent worklist and only recomputes an object's proposal when a
  direct neighbor was just renamed, using precomputed adjacency. Same
  round-by-round batching as before, so the final naming is unchanged.
- Add `tests/various/autoname_scaling.sh`, which checks a 10000-cell
  propagation chain still autonames within a time budget the old
  algorithm could not meet.

## Test plan

- [x] `tests/various/autoname.ys` passes with unchanged output
- [x] `tests/various/autoname_scaling.sh` passes (~2s, vs. 20s+ with the
      old full-rescan algorithm at the same size)
- [x] Verified against the real-world synthesis flow that previously
      never completed

### DIFF
--- a/passes/cmds/autoname.cc
+++ b/passes/cmds/autoname.cc
@@ -17,13 +17,42 @@
  *
  */
 
+// -----------------------------------------------------------------------
+// Algorithm
+//
+// A cell or wire with a private ($-prefixed) name becomes nameable once it
+// has at least one directly connected neighbour with a public name: the
+// proposed name is that neighbour's name plus a short suffix (cell/port
+// derived). Renaming an object can make ITS neighbours nameable in turn,
+// so public names propagate outward one hop per round. Each round, every
+// currently-nameable object within 2x the round's best score is renamed
+// together (this batches together names that are equally good, instead of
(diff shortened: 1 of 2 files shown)


rowanG077:
A colleague tested this PR and it solved a pathological autoname runtime case for us as well.

widlarizer:
Thanks for bringing attention to this performance issue. This does help but it also regresses in other cases by constructing strings for even rejected names. Meaning, you're basically undoing the work in #3945. Even if I remove that, I still get better results with #6050.

PR #6023, opened 2026-07-09
Michaelbaier/sc 649/segmentation fault during proc init with

b-michi:
This is a proposed fix to the issue #1206, considering all the changes proposed in PR #3460

To solve the problems,I have added a constant to rtlil.h that is used to check if wires or constants get to big.

### DIFF
--- a/frontends/rtlil/rtlil_frontend.cc
+++ b/frontends/rtlil/rtlil_frontend.cc
@@ -31,10 +31,6 @@
 YOSYS_NAMESPACE_BEGIN
 
 struct RTLILFrontendWorker {
-	// Forbid constants of more than 1 Gb.
-	// This will help us not explode on malicious RTLIL.
-	static constexpr int MAX_CONST_WIDTH = 1024 * 1024 * 1024;
-
 	std::istream *f = nullptr;
 	RTLIL::Design *design;
 	bool flag_nooverwrite = false;
@@ -283,7 +279,7 @@ struct RTLILFrontendWorker {
 			++idx;
 
 		std::vector<RTLIL::State> bits;
-		if (width > MAX_CONST_WIDTH)
+		if (width >= RTLIL::WIDTH_LIMIT)
 			error("Constant width %lld out of range before `%s`.", width, error_token());
 		bits.reserve(width);
 		int start_idx = idx;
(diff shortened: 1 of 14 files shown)


frontends/rtlil/rtlil_frontend.cc
```diff
@@ -589,8 +589,12 @@ struct RTLILFrontendWorker {
...
-				width = parse_integer();
+			if (try_parse_keyword("width")){
+				long long width_val = parse_integer();
+				if (width_val < 0 || width_val >= RTLIL::WIDTH_LIMIT)
+					error("Memory width %lld out of range before `%s`.", width_val, error_token());
+				width = width_val;
+			}
 			else if (try_parse_keyword("size"))
 				size = parse_integer();
 			else if (try_parse_keyword("offset"))
```
hexratcc: I think we need the same check for `size` and `offset`, as there still exists a `long long` to `int` truncation.

kernel/rtlil.h
```diff
@@ -99,6 +99,9 @@ namespace RTLIL
 		PD_INOUT = 3
 	};
 
+	// Maximum width in bits of RTLIL::Wire or RTLIL::Const
+	constexpr int WIDTH_LIMIT = 1 << 30;
```
hexratcc, line 103: Is there a reason why the old `MAX_CONST_WIDTH` is `1 << 30` instead of `1 << 24 ` like `kWidthLimit` in the Verilog frontend?
https://github.com/YosysHQ/yosys/blob/b88b73a99a9db67030a952e00f2f111d00cb6013/frontends/ast/genrtlil.cc#L1408

hexratcc (changes requested):

kernel/rtlil.h
b-michi: The comment on the last PR (https://github.com/YosysHQ/yosys/pull/3460#issuecomment-3655882567) let me think that this is the supposed limit.

frontends/rtlil/rtlil_frontend.cc
b-michi: I have added the check in dca9234d8 I checked here for INT_MIN/INT_MAX truncation for size and offset

kernel/rtlil.h
hexratcc: Fair enough, missed that.

hexratcc (approved):
lgtm

PR #6024, opened 2026-07-10
Include build datetime in scl cache hash

hexratcc:
Mixes the build datetime into the merged `scl` cache filename hash so caches from older builds are not reused across yosys/abc builds, potentially preventing confusing behavior for developers.

### DIFF
--- a/kernel/version.cc.in
+++ b/kernel/version.cc.in
@@ -1,4 +1,5 @@
 namespace Yosys {
 	const char *yosys_version_str = "@YOSYS_BUILD_INFO@";
 	const char *yosys_git_hash_str = "@YOSYS_CHECKOUT_INFO@";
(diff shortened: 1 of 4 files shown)


hexratcc:
[sc-655]

KrystalDelusion (approved):
I think this is fine, and should resolve the problem I encountered, but is there a compelling argument to use the build datetime over just using the version/git hash?  Is it to avoid treating two different `+dirty` builds as the same in case something in Yosys has changed that would impact the SCL?  Is there anything in Yosys that _would_ impact the SCL, or is it just if ABC has changed?

hexratcc:
I think both work but mine just covers more potential (albeit unlikely) issues, but at virtually no cost.

PR #6029, opened 2026-07-10
Speed up opt_prienc fingerprinting and candidate discovery

akashlevy:
## Summary
- Cut `opt_prienc` runtime on the test suite from ~35s to ~2s (RR still on by default) via shared ConstEval, inverted wire indexing, budgeted large-cone rewalks, and sampled large-N PE/RR fingerprints.
- Tighten Stage 1 prefilters (interesting drivers / PE-shaped cones) and stop treating `$shl` as RR-like so big CLZ cones are not fingerprinted as round-robin.
- Add Group G tests locking the large-N PE one-hot and RR start sampling paths.

## Test plan
- [x] `./yosys tests/opt/opt_prienc.ys` (EXIT 0; ~2s in `opt_prienc`)
- [ ] CI opt / unit checks on this PR
- [ ] Greptile review triaged


akashlevy:
Opened against upstream by mistake; Silimate-internal change will land via Silimate/yosys.

PR #6030, opened 2026-07-11
opt_dff: add SAT effort limits

### LINKED ISSUE #6016: opt_dff -sat significantly slower for certain design patterns
### Version

Yosys 0.66+ (git sha1 b35b6706f, GNU /usr/bin/c++ 15.2.0)

Linux

### Reproduction Steps

`yosys -p "read_verilog -sv repro.v; hierarchy -top repro; prep; opt_dff -sat`

repo.v:
```
// yosys -p "read_verilog -sv repro.v; hierarchy -top repro; prep; opt_dff -sat"
module repro (
  input wire clk,
  input wire rst_n,
  input wire [43:0] in,
  output reg [99:0] q
);
  function automatic satisfied(input [43:0] in_val);
    integer i, j;
    reg [3:0] p[0:10];
    reg valid, no_coll;
    begin
      for (i = 0; i < 11; i = i + 1)
        p[i] = in_val[i*4 +: 4];
      valid = 1'b1;
      for (i = 0; i < 11; i = i + 1)
        if (p[i] >= 10) valid = 1'b0;
      no_coll = 1'b1;
      for (i = 0; i < 11; i = i + 1)
        for (j = i + 1; j < 11; j = j + 1)
          if (p[i] == p[j]) no_coll = 1'b0;
      satisfied = valid && no_coll;
    end
  endfunction
  genvar i;
  generate
    for (i = 0; i < 100; i = i + 1) begin : g_
      always @(posedge clk or negedge rst_n) begin
        if (!rst_n)
          q[i] <= 1'b0;
        else
          q[i] <= satisfied(in ^ i);
      end
    end
  endgenerate
endmodule

```

### Expected Behavior

The above design is significantly slower when using opt_dff with `-sat` than without `-sat`. We'd like to use `-sat` to make use of the recent eqbits elimination feature in -sat. However, this case will go from ~2m to over an hour with -sat.

### Actual Behavior

The SAT should have an exit criteria or maximum effort for challenging problems. It is understandable that it would be slower to some degree.

hexratcc:
This PR fixes #6016 by adding a SAT effort limit to `opt_dff -sat`, measured in minisat propagation steps. The limit bounds the total effort spent per module and is configurable via the `opt_dff.sat_effort` scratchpad variable (default `1e9`, 0 = unlimited). Once the budget is exhausted, remaining proofs are skipped with a warning, so the result is always a subset of what an unlimited run would prove.

The original issue only concerns `run_constbits`, but the same shared budget also covers `run_eqbits` to avoid the equivalent problem there. I've also considered using ezsat's solver timeout, but opted not to as it produces non-deterministic results and QoR would vary wildly across machines. Since the limit caps total rather than per-query effort, runtime also stays roughly flat as the number of FFs grows. The case below (repro from #6016) was reported as taking >1hr, and on my machine it takes about 350s. With this PR's default effort limit it takes about 90s, still folding 29 of the 100 FFs to constants.
```verilog
// yosys -p "read_verilog -sv repro.v; hierarchy -top repro; prep; opt_dff -sat"
module repro (
  input wire clk,
  input wire rst_n,
  input wire [43:0] in,
  output reg [99:0] q
);
  function automatic satisfied(input [43:0] in_val);
    integer i, j;
    reg [3:0] p[0:10];
    reg valid, no_coll;
    begin
      for (i = 0; i < 11; i = i + 1)
        p[i] = in_val[i*4 +: 4];
      valid = 1'b1;
      for (i = 0; i < 11; i = i + 1)
        if (p[i] >= 10) valid = 1'b0;
      no_coll = 1'b1;
      for (i = 0; i < 11; i = i + 1)
        for (j = i + 1; j < 11; j = j + 1)
          if (p[i] == p[j]) no_coll = 1'b0;
      satisfied = valid && no_coll;
    end
  endfunction
  genvar i;
  generate
    for (i = 0; i < 100; i = i + 1) begin : g_
      always @(posedge clk or negedge rst_n) begin
        if (!rst_n)
          q[i] <= 1'b0;
        else
          q[i] <= satisfied(in ^ i);
      end
    end
  endgenerate
endmodule
```

### DIFF
--- a/passes/opt/opt_dff.cc
+++ b/passes/opt/opt_dff.cc
@@ -1260,9 +1260,18 @@ struct OptDffWorker
 		// Build the next-state function n_lit[idx] of every candidate bit by
 		// folding the FF's control logic on top of the D input (-> next value)
 
+		// Verified merges can depend on the equality of classes that are still unproven, so an incomplete run cannot keep any of them
+		auto drop_all = [&]() -> std::vector<std::vector<int>> {
+			log("Dropping all equivalent-FF merge candidates in module %s (sat effort budget exceeded before all classes were proven).\n", log_id(module));
+			return {};
+		};
+		int64_t cells_charged = 0;
+
 		// Two bits are equivalent if their next states always agree whenever their
 		// current states (and those of every other candidate pair) agree
 		for (auto &cls : classes) {
+			if (sat_budget_spent())
+				return drop_all();
 			for (int idx : cls) {
 				const EqBit &eb = bits[idx];
 				const FfData &ff = ff_for_cell.at(eb.cell);
@@ -1294,6 +1303,8 @@ struct OptDffWorker
 
 				n_lit[idx] = n;
 			}
+			qcsat.prepare();
+			charge_import(qcsat, cells_charged);
 		}
 
 		qcsat.prepare();
--- a/passes/opt/opt_dff.cc
+++ b/passes/opt/opt_dff.cc
@@ -143,6 +143,57 @@ struct OptDffWorker
 
 	std::vector<Cell *> dff_cells;
 
+	// Solver effort budget shared by run_constbits and run_eqbits, per module
+	static constexpr int64_t sat_import_cell_cost = 100;
+	static constexpr int sat_batch_cells = 10000;
+	int64_t sat_effort = 0;
+	int64_t sat_effort_left = 0;
+	bool sat_budget_out = false;
+
+	void sat_budget_exceeded()
+	{
+		if (!sat_budget_out)
+			log_warning("opt_dff -sat: solver effort budget of %lld exceeded in module %s, "
+					"skipping remaining sat proofs (scratchpad option 'opt_dff.sat_effort').\n",
+					(long long)(sat_effort / 1000), log_id(module));
+		sat_budget_out = true;
+	}
+
+	bool sat_budget_spent()
+	{
+		if (sat_budget_out)
+			return true;
+		if (sat_effort > 0 && sat_effort_left <= 0) {
+			sat_budget_exceeded();
+			return true;
+		}
+		return false;
+	}
+
+	int solve_budgeted(QuickConeSat &qcsat, int64_t cap, const std::vector<int> &modelExprs,
+			std::vector<bool> &modelVals, const std::vector<int> &assumptions)
+	{
+		// SAT = 1
+		// UNSAT = 0
+		// effort limit reached = -1 (can't be treated as UNSAT)
+		if (sat_effort > 0)
+			cap = (cap > 0) ? min(cap, sat_effort_left) : sat_effort_left;
+		qcsat.ez->setSolverPropLimit(cap);
+		bool sat_res = qcsat.ez->solve(modelExprs, modelVals, assumptions);
+		if (sat_effort > 0)
+			sat_effort_left -= qcsat.ez->getSolverProps();
+		if (!sat_res && qcsat.ez->getSolverPropLimitStatus())
+			return -1;
+		return sat_res ? 1 : 0;
+	}
+
+	void charge_import(QuickConeSat &qcsat, int64_t &cells_charged)
+	{
+		if (sat_effort > 0)
+			sat_effort_left -= (GetSize(qcsat.imported_cells) - cells_charged) * sat_import_cell_cost;
+		cells_charged = GetSize(qcsat.imported_cells);
+	}
+
 	bool is_active(SigBit sig, bool pol) const {
 		return sig == (pol ? State::S1 : State::S0);
 	}
@@ -1283,6 +1334,9 @@ struct OptDffWorker
 				if (n_lit[rep] == n_lit[cls[i]])
 					continue;
 
+				if (sat_budget_spent())
+					return drop_all();
+
 				// Can the next state of the rep and this member ever differ?
 				int query = qcsat.ez->XOR(n_lit[rep], n_lit[cls[i]]);
 				// Capture every member's next-state value in that model so one counterexample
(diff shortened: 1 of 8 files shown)


hexratcc:
[sc-650]

mikesinouye:
Thanks for working on the fix for this! I wonder why your setup was able to complete that test case in 350s vs. 1hr. Either way, the test case was a shrunk-down version of the original design with the issue, so a lever to control SAT effort is a good solution.

KrystalDelusion:
@hexratcc maybe worth adding a small test with/without a very low `opt_dff.sat_effort` asserting that the two are equivalent, but that the very low effort one has (or may have) more FFs and that the warning is raised for the low effort but not the one without?  `opt_dff` should also respect selections, so maybe a second very low effort run but using a selection that results in a different optimization?

passes/opt/opt_dff.cc
```diff
@@ -143,6 +143,56 @@ struct OptDffWorker
 
 	std::vector<Cell *> dff_cells;
 
+	// Solver effort budget shared by run_constbits and run_eqbits, per module
```
widlarizer: Everything related to SAT budgeting should be encapsulated in some way, rather than just flat in this struct. For the sake of somebody reading the opt_dff sources who isn't interested in opt_dff -sat, which I think is very common - usually people look for the core ideas first rather than optional modes

passes/opt/opt_dff.cc
```diff
@@ -143,6 +143,56 @@ struct OptDffWorker
...
+			sat_budget_exceeded();
+			return true;
+		}
+		return false;
+	}
+
+	int solve_budgeted(QuickConeSat &qcsat, int64_t cap, const std::vector<int> &modelExprs,
+			std::vector<bool> &modelVals, const std::vector<int> &assumptions)
+	{
+		// SAT = 1
```
widlarizer: This strongly suggests an enum

passes/opt/opt_dff.cc
```diff
@@ -143,6 +143,56 @@ struct OptDffWorker
...
+		// UNSAT = 0
+		// effort limit reached = -1 (can't be treated as UNSAT)
+		if (sat_effort > 0)
```
widlarizer: This following code actually seems kind of unrelated to opt_dff - if passes that aren't opt_dff integrate SAT-based optimizations and want to use effort limits, would they have to copy it over? Could be tricky to separate from the logic of when effort is spent, but it could be worth it

passes/opt/opt_dff.cc
```diff
@@ -143,6 +143,56 @@ struct OptDffWorker
...
+		qcsat.ez->setSolverPropLimit(cap);
+		bool sat_res = qcsat.ez->solve(modelExprs, modelVals, assumptions);
+		if (sat_effort > 0)
+			sat_effort_left -= qcsat.ez->getSolverProps();
+		if (!sat_res && qcsat.ez->getSolverPropLimitStatus())
+			return -1;
+		return sat_res ? 1 : 0;
+	}
+
+	void charge_import(QuickConeSat &qcsat, int64_t &cells_charged)
```
widlarizer: I think "charging a cell" is a new term, so you should define it

passes/opt/opt_dff.cc
```diff
@@ -919,83 +953,234 @@ struct OptDffWorker
...
-				if (keep_bits.empty()) {
-					cells_to_remove.push_back(cell);
-				} else {
-					ff = ff.slice(keep_bits);
-					ff.cell = cell;
-					ffs_to_emit.push_back(ff);
+			while (!out_of_budget) {
+				bool all_resolved = true;
+
+				for (int obi = batch_begin; obi < batch_end; obi++) {
```
widlarizer, line 1101: While it's not too hard to understand this loop is getting lengthy. I would prefer if the body of `while (!out_of_budget)` already was quite short by turning that body into a function or method, for example, so that control flow isn't mixed too hard with data handling. Also, it could help to split SAT-based operations out of opt_dff.cc entirely

passes/opt/opt_dff.cc
```diff
@@ -1196,6 +1381,13 @@ struct OptDffWorker
 		return refined_classes;
 	}
 
+	std::vector<std::vector<int>> drop_all_classes()
+	{
+		// Verified merges can depend on the equality of classes that are still unproven, so an incomplete run cannot keep any of them
+		log("Dropping all equivalent-FF merge candidates in module %s (sat effort budget exceeded before all classes were proven).\n", log_id(module));
```
widlarizer: What should this communicate to a hardware designer reading the log file?

widlarizer (changes requested):
Some readability considerations for now. It's not quite clear to me how the batching affects operation of the pass and correctness, could you describe it a bit here / in the code?

passes/opt/opt_dff.cc
hexratcc: I've  modified the message to hopefully be a bit more clear, the idea is to highlight an insufficient sat effort value.

hexratcc:
I've made some changes, in particular encapsulating the SAT effort budget (`SatEffortBudget`), which should hopefully clear up the main const/eqbits parts.

passes/opt/opt_dff.cc
```diff
@@ -919,83 +921,242 @@ struct OptDffWorker
...
 		// Find FFs that are provably constant
 		ModWalker modwalker(module->design, module);
-		QuickConeSat qcsat(modwalker);
 
-		std::vector<RTLIL::Cell*> cells_to_remove;
-		std::vector<FfData> ffs_to_emit;
+		dict<Cell *, pool<int>> const_bits;
 		bool did_something = false;
 
+		// Fold constant D/AD inputs into the tested value directly bits whose remaining inputs are
```
widlarizer: Seems like two sentences concatenated

mikesinouye:
Hi @widlarizer , is the current status of this PR that it is under internal review?

passes/opt/opt_dff.cc
```diff
@@ -919,83 +921,242 @@ struct OptDffWorker
...
+				ob2->dropped = true;
+		}
+		ob.dropped = true;
+		return true;
+	}
+
 	bool run_constbits()
 	{
 		// Find FFs that are provably constant
 		ModWalker modwalker(module->design, module);
```
widlarizer: This modwalker is expensive to construct so it should be shared between eqbits and constbits if both are run

widlarizer (changes requested):
Small fixups needed and good to go

widlarizer (approved):

hexratcc:
Turns out we didn't have much test coverage for `constbits` -sat, I've added some tests to hopefully improve the situation (these also pass with an older version of Yosys).

mikesinouye:
Thank you @hexratcc  and @widlarizer!

Issue #6028, opened 2026-07-10
fedora packaging issues for 0.67+


gsomlo:
### Version

0.67

### On which OS did this happen?

Linux

### Reproduction Steps

I'm packaging yosys  for fedora, and currently trying to catch up with the switch to cmake, as well as the addition of systemverilog slang support.

While I try to figure out whether I need to separately package sv-slang as its own RPM (with a "BuildRequires" dependency on it added to yosys -- as opposed to the bundled subrepo), I tried building with:

```
cmake -DYOSYS_ABC_EXECUTABLE=/bin/abc -DYOSYS_WITHOUT_SLANG=ON
```

but I (still) get the following error:

```
...
-- Configuring Yosys 0.67+post
-- Git commit b88b73a99a9db67030a952e00f2f111d00cb6013
-- External ABC: /usr/bin/abc
CMake Error at frontends/slang/CMakeLists.txt:1 (include):
  include could not find requested file:

    lib/cmake/GitRevision.cmake


CMake Error at frontends/slang/CMakeLists.txt:2 (git_rev_parse):
  Unknown CMake command "git_rev_parse".


-- Configuring incomplete, errors occurred!
...
```

Should `frontends/slang` not also be skipped when `YOSYS_WITHOUT_SLANG == ON` ?

Is there a better way to disable slang (for now) that I missed?

I suppose I could always try to `rm -rf frontends/slang` before running cmake with `-DYOSYS_WITHOUT_SLANG=ON`, but that's still technically working around a bug in the build environment, I think... :)

I'm inevitably expecting a few more issues to pop up as I progress with rpm packaging -- would it be better to update this issue as I run into problems, or create new ones?

Many thanks,
-G

### Expected Behavior

cmake completes with slang support disabled

### Actual Behavior

cmake fails as shown above

wkazubski:
On openSUSE Tumbleweed I get the following errors:
```
CMake Error at libs/CMakeLists.txt:2 (add_subdirectory):
  The source directory

    /home/wk/rpmbuild/BUILD/yosys-0.67-build/yosys-0.67/libs/cxxopts

  does not contain a CMakeLists.txt file.


CMake Error at libs/CMakeLists.txt:40 (add_subdirectory):
  The source directory

    /home/wk/rpmbuild/BUILD/yosys-0.67-build/yosys-0.67/libs/slang

  does not contain a CMakeLists.txt file.


[...]

PR #6031, opened 2026-07-12
cmake: allow system fmt/tomlplusplus/boost for bundled deps

daeho-ro:

Hopefully use system installed libraries not vendored.

First use find_package and after then fetch library if it is missing.

FetchContent now tries find_package() first for fmt, tomlplusplus and Boost.Regex, only falling back to the vendored copies when no system package is found. This lets distribution and package-manager builds (e.g. Homebrew) link against system libraries instead of the bundles, without changing the default behaviour for a plain source build.

### DIFF
--- a/libs/CMakeLists.txt
+++ b/libs/CMakeLists.txt
@@ -9,33 +9,43 @@ add_subdirectory(sha1)
 add_subdirectory(subcircuit)
 block()
     set(BUILD_SHARED_LIBS OFF)
+    set(CMAKE_FIND_PACKAGE_TARGETS_GLOBAL TRUE)
     include(FetchContent)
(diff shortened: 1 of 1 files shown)


whitequark:
This seems like a duplicate of #6028.

daeho-ro:
Nice catch, I could change homebrew patch url to the new one if @gsomlo open a PR.
And nice to see you again here 👋 @whitequark

whitequark:
Ha, yes, I just noticed that you filed the git-pages issues too. (I'm working on those.)

mmicko:
Superseded by  https://github.com/YosysHQ/yosys/pull/6034

PR #6032, opened 2026-07-12
CMake: better support for distibution packaging of yosys

gsomlo:
This series improves upstream support for distribution packaging.

Many Linux distros (particularly fedora) strongly encourage building against existing packaged dependencies rather than "vendoring" or "bundling".

The first patch provides an option `YOSYS_WITH_PKG_DEPS` (off by default) that tells `cmake` to avoid using (and attempting to download) vendored dependencies.

The second patch prevents a config failure when `YOSYS_WITHOUT_SLANG` is turned on, but the vendored slang repo has  not (recursively) been added.

Finally, the third patch adds support for using distro-packaged `cxxopts` when `YOSYS_WITH_PKG_DEPS` is selected (again, preserving the current behavior by default).

This addresses https://github.com/YosysHQ/yosys/issues/6028 (and possibly also https://github.com/YosysHQ/yosys/pull/6031)


whitequark:
I disagree with the distro policies that concern unbundling of vendored dependencies. In case of self-contained libraries like cxxopts that are not exposed to untrusted data, this has no benefit to the end user (when it's a header-only C++ library you cannot claim that "dynamic linking" is a benefit, not that it's a very convincing argument in other cases), and a real drawback for maintainers: namely, it can introduce breakage that will not be seen by anybody building from upstream sources.

I am aware that a policy document may require you to blindly do this regardless of context. That is bad policy, and I see no reasons to encourage sticking to bad policies.

CMakeLists.txt
```diff
@@ -56,6 +56,7 @@ option(YOSYS_WITHOUT_EDITLINE "Disable editline integration" OFF)
 option(YOSYS_WITHOUT_SLANG "Disable Slang integration" OFF)
 option(YOSYS_WITHOUT_TCL "Disable Tcl integration" OFF)
 option(YOSYS_WITH_PYTHON "Enable Python integration" OFF)
+option(YOSYS_WITH_PKG_DEPS "Disable bundled/vendored dependencies" OFF)
```
whitequark, line 59: Please name this something different. `YOSYS_WITH_*` and `YOSYS_WITHOUT_*` prefixes are meant to be used with a package name, and `PKG_DEPS` is not a package name.

`-DYOSYS_USE_SYSTEM_SLANG=ON` would work.

gsomlo:
On Sun, Jul 12, 2026 at 08:51:24AM -0700, Catherine wrote:
> In case of self-contained libraries like cxxopts, this has no
> benefit to the end user (when it's a header-only C++ library you cannot claim
> that "dynamic linking" is a benefit, not that it's a very convincing argument
> in other cases), and a real drawback for maintainers: namely, it can introduce
> breakage that will not be seen by anybody building from upstream sources.

Well in that case you could always just copy the standalone
`cxxopts.hpp` file into the yosys sources and forego the
pomp and circumstance of making it a "vendored submodule"
(with all the additional CMake glue that currently entails),
which would then simplify everyone's existence... :)

whitequark:
> Well in that case you could always just copy the standalone
> `cxxopts.hpp` file into the yosys sources and forego the
> pomp and circumstance of making it a "vendored submodule"
> (with all the additional CMake glue that currently entails),
> which would then simplify everyone's existence... :)

I don't see how this will in any way simplify your existence. (Also, it doesn't actually stop people from trying to patch it out or complain about it. Nextpnr has a module that's copied into the source tree and we had this exact argument with a Debian guy about it.)

mmicko:
Superseded by  https://github.com/YosysHQ/yosys/pull/6034

PR #6033, opened 2026-07-12
gowin: infer DSP multipliers for the GW5A family

olby:
Adds GW5A DSP multiply inference to `synth_gowin -family gw5a`, using the same
mul2dsp + dsp_map pattern as gw1n/gw2a. Maps the GW5A's three multiplier
widths:

- `$__MUL27X18` -> `MULTALU27X18` (M0, 27x18)
- `$__MUL12X12` -> `MULT12X12` (M1, 12x12)
- `$__MUL27X36` -> `MULT27X36` (two cascaded M0s, 27x36)

<=12x12 multiplies route to MULT12X12 so they pack two-per-block via M0+M1;
B>18 multiplies go to MULT27X36 (two cascaded DSPs); the rest use
MULTALU27X18.

Both multiplier types are signed-only (no per-operand sign control), so mul2dsp
takes `DSP_SIGNEDONLY` on every pass without it a wide signed multiply
decomposes into a mixed-sign chunk the DSP evaluates wrong (32x16 -1*-1 ->
0x08000001 instead of 1).

Verified: Verilator (32x16 16/16, 12x12, 27x24), packing (20 <=12x12 -> 10
tiles, 2/block), and on silicon (Tang Primer 25K: self-check sweeps pass for
all three cells; removing DSP_SIGNEDONLY reproduces the -1*-1 corruption).

Known limitation: nextpnr-himbaechel packs MULT12X12 pairs two-per-block but
does not co-locate MULTALU27X18 with MULT12X12, so M1 sits idle next to any
MULTALU27X18 (whether standalone, in an A-overflow decomposition like 28x18,
or as a 27x36 cascade half MULT27X36 uses two DSP blocks' M0s). Whether M1 can
operate during an M0 cascade is unverified; the chip spec does not clearly
describe M1's state when M0 is in cascade mode. These are placer/chipdb/spec
questions, not the techmap's.

This is a bit out of my depth and I have been using GLM-5.2 to both understand
specifications, code and related issues.


techlibs/gowin/synth_gowin.cc
```diff
@@ -44,6 +44,12 @@ struct SynthGowinPass : public ScriptPass
 		{9, 9, 4, 4, "$__MUL9X9"},
 	};
 
+	const std::vector<DSPRule> gw5a_dsp_rules = {
+		{27, 36,  2, 19, "$__MUL27X36"},
+		{27, 18, 13, 13, "$__MUL27X18"},
+		{12, 12,  2,  2, "$__MUL12X12"},
+	};
```
vanhanit, line 51: Claude Fable 5 review:

The `{13, 13}` min-widths leave a gap for asymmetric products, and it can't be fixed with a gw1n-style min-width pair: mul2dsp checks `DSP_A/B_MINWIDTH` *before* it commutes a narrow-A/wide-B `$mul` (it normalises A to the wider operand) and re-checks the same mins after commuting, so a companion `{4, 13}` rule accepts the product, commutes it, then rejects the commuted cell. (The existing gw1n `{18, 18, 10, 4}` / `{18, 18, 4, 10}` pair has this exact problem — measured: signed 16x9 → 1 MULT18X18 but signed 9x16 → 2 MULT9X9 + 16 ALU — probably worth a follow-up.)

What does work is using that dead-end as a *normaliser* and repeating the wide rule after it, plus a floor of 4 (gw1n precedent) so trivial multiplies stay in LUTs:

```suggestion
	// mul2dsp checks DSP_A/B_MINWIDTH before it commutes a narrow-A/wide-B
	// $mul (it normalises A to the wider operand) and re-checks them after
	// commuting, so a single asymmetric rule only matches products that
	// already arrive wide-A. The {18, 27, ...} rule accepts a narrow-A
	// product, commutes it, then rejects the commuted cell -- leaving a
	// wide-A $mul behind for the repeated 27x18 rule to map. Without this
	// normalise-and-retry step, e.g. a signed 9x16 decomposes onto two
	// MULT12X12 halves plus a fabric adder where one MULTALU27X18 fits.
	const std::vector<DSPRule> gw5a_dsp_rules = {
		{27, 36,  4, 19, "$__MUL27X36"},
		{27, 18, 13,  4, "$__MUL27X18"},
		{18, 27,  4, 13, "$__MUL18X27"},
		{27, 18, 13,  4, "$__MUL27X18"},
		{12, 12,  4,  4, "$__MUL12X12"},
	};
```

Needs the `$__MUL18X27` wrapper in `dsp_map_gw5a.v` (suggested there). Measured deltas vs. the current table (yosys 0.66, same techmap sequence as this pass):

- signed 16x9, 9x16, 13x4, 4x13, 20x4: 2 MULT12X12 + fabric adder each → **1 MULTALU27X18 each**
- unsigned 12x12 (DSP_SIGNEDONLY makes it 13x13, which overflows the 12x12 rule): 4 MULT12X12 + fabric → 1 MULT12X12 + fabric
- signed 2x2 / 2x20: no longer burn DSP halves on trivial products (soft logic)
- unchanged: 12x12, 13x13, 18x18, 27x18, 27x19 → MULT27X36, u26x17
- 28x18 remainder now stays in fabric instead of 2 MULT12X12 (test update suggested in mul_gw5a.ys)

All mapped shapes verified bit-exact against behavioural `*` over 20k random vectors per shape in simulation. Remaining known wart: unsigned muls whose width is exactly 12 still split (the +1 signedness bit overflows the half-tile and the pre-commute min check hides them from the 27x18 rules) — any min tweak that fixes it regresses signed 12x12 from a half-tile to a full tile, so it seems better to pin it with a test.

techlibs/gowin/dsp_map_gw5a.v
```diff
@@ -0,0 +1,95 @@
...
+
+	wire [47:0] dout_w;
+
+	MULTALU27X18 #(
+		.MULT12X12_EN("FALSE"),
+		.MULT_RESET_MODE("SYNC"),
+		.AREG_CLK("BYPASS"),
+		.BREG_CLK("BYPASS"),
+		.PREG_CLK("BYPASS"),
+		.OREG_CLK("BYPASS")
```
vanhanit, line 18: Claude Fable 5 review:

`C_SEL` defaults to `1'b1` in the vendor primitive, which selects the `PRE_LOAD` constant (default 0) as the ALU addend rather than the C port this instantiation drives to zero. Both are functionally zero, but datapoint from an independent implementation of this same multiply-only configuration: the variant proven on GW5A-25 silicon (5009/5009 signed products correct via nextpnr-himbaechel + apicula, corners included) used `C_SEL(1'b0)`, i.e. the C-port path. Suggest pinning the hardware-verified configuration explicitly:

```suggestion
		.MULT12X12_EN("FALSE"),
		.MULT_RESET_MODE("SYNC"),
		.C_SEL(1'b0),
		.AREG_CLK("BYPASS"),
		.BREG_CLK("BYPASS"),
		.PREG_CLK("BYPASS"),
		.OREG_CLK("BYPASS")
```

(Nit: the four `*REG_CLK("BYPASS")` lines and `MULT12X12_EN("FALSE")` restate the primitive's defaults and could be dropped.)

techlibs/gowin/dsp_map_gw5a.v
```diff
@@ -0,0 +1,95 @@
...
+		.CASISEL(1'b0),
+		.ADDSUB(2'b00),
+		.CLK(2'b00),
+		.CE(2'b00),
+		.RESET(2'b00)
+	);
+
+	assign Y = dout_w[44:0];
+
+endmodule
```
vanhanit, line 43: Claude Fable 5 review:

The normalise-and-retry rule sequence suggested in `synth_gowin.cc` needs a wrapper for the commuted orientation (mul2dsp extends A to 18 and B to 27 bits for that rule, so the operands must be crossed onto the tile's 27-bit/18-bit ports — multiplication commutes). Verified bit-exact over 20k random vectors (signed 9x16, 4x13 among the shapes exercising it):

```suggestion
endmodule

module \$__MUL18X27 (input [17:0] A, input [26:0] B, output [44:0] Y);

    parameter A_WIDTH = 18;
    parameter B_WIDTH = 27;
    parameter Y_WIDTH = 45;
    parameter A_SIGNED = 0;
    parameter B_SIGNED = 0;

	wire [47:0] dout_w;

	MULTALU27X18 #(
		.MULT12X12_EN("FALSE"),
		.MULT_RESET_MODE("SYNC"),
		.C_SEL(1'b0)
	) __TECHMAP_REPLACE__ (
		.DOUT(dout_w),
		.CASO(),
		.SOA(),
		.A(B),
		.SIA(27'd0),
		.B(A),
		.C(48'd0),
		.D(26'd0),
		.CASI(48'd0),
		.ACCSEL(1'b0),
		.PSEL(1'b0),
		.ASEL(1'b0),
		.PADDSUB(1'b0),
		.CSEL(1'b0),
		.CASISEL(1'b0),
		.ADDSUB(2'b00),
		.CLK(2'b00),
		.CE(2'b00),
		.RESET(2'b00)
	);

	assign Y = dout_w[44:0];

endmodule
```

tests/arch/gowin/mul_gw5a.v
```diff
@@ -0,0 +1,11 @@
...
+// signed-only, so signed operands exercise the routing boundaries cleanly).
+module top
+#(parameter X_WIDTH=8, Y_WIDTH=8, A_WIDTH=16)
+(
+    input signed [X_WIDTH-1:0] x,
+    input signed [Y_WIDTH-1:0] y,
+    output signed [A_WIDTH-1:0] A
+);
+    assign A = x * y;
+endmodule
```
vanhanit, line 11: Claude Fable 5 review:

All test cases are signed; the DSP_SIGNEDONLY +1-bit path for unsigned multiplies has no coverage. An unsigned twin of `top` enables the cases suggested in `mul_gw5a.ys`:

```suggestion
endmodule

// Unsigned twin: exercises the DSP_SIGNEDONLY path (mul2dsp widens unsigned
// operands by one bit before mapping onto the signed-only GW5A multipliers).
module top_u
#(parameter X_WIDTH=8, Y_WIDTH=8, A_WIDTH=16)
(
    input [X_WIDTH-1:0] x,
    input [Y_WIDTH-1:0] y,
    output [A_WIDTH-1:0] A
);
    assign A = x * y;
endmodule
```

tests/arch/gowin/mul_gw5a.ys
```diff
@@ -0,0 +1,72 @@
...
+# 28x18 -> A overflows 27: 1 MULTALU27X18 + MULT12X12 partials
+design -reset
+read_verilog mul_gw5a.v
+chparam -set X_WIDTH 28 -set Y_WIDTH 18 -set A_WIDTH 46
+hierarchy -top top
+proc
+synth_gowin -family gw5a
+cd top
+select -assert-count 1 t:MULTALU27X18
+select -assert-count 2 t:MULT12X12
```
vanhanit, line 62: Claude Fable 5 review:

With the suggested rule table the 28x18 remainder (a 2-bit-wide slice after the 27-bit chunk) falls below the 12x12 rule's min-width floor and stays in fabric instead of occupying two MULT12X12 halves:

```suggestion
# 28x18 -> A overflows 27: 1 MULTALU27X18 + fabric remainder
design -reset
read_verilog mul_gw5a.v
chparam -set X_WIDTH 28 -set Y_WIDTH 18 -set A_WIDTH 46
hierarchy -top top
proc
synth_gowin -family gw5a
cd top
select -assert-count 1 t:MULTALU27X18
select -assert-none t:MULT12X12
```

tests/arch/gowin/mul_gw5a.ys
```diff
@@ -0,0 +1,72 @@
...
+# Make sure that DSPs are not inferred with -nodsp option
+design -reset
+read_verilog mul_gw5a.v
+chparam -set X_WIDTH 27 -set Y_WIDTH 18 -set A_WIDTH 45
+hierarchy -top top
+proc
+synth_gowin -family gw5a -nodsp
+cd top
+select -assert-none t:MULTALU27X18
+select -assert-none t:MULT12X12
```
vanhanit, line 85: Claude Fable 5 review:

Suggest also covering: both orientations of an asymmetric product (each fits a single MULTALU27X18 with the normalise-and-retry rules), the unsigned single-tile fit, the trivial-multiply floor, and pinning the known unsigned-12x12 decomposition:

```suggestion
select -assert-none t:MULT12X12


# 16x9 and 9x16 -> 1 MULTALU27X18 each (asymmetric fit, both orientations)
design -reset
read_verilog mul_gw5a.v
chparam -set X_WIDTH 16 -set Y_WIDTH 9 -set A_WIDTH 25
hierarchy -top top
proc
synth_gowin -family gw5a
cd top
select -assert-count 1 t:MULTALU27X18
select -assert-none t:MULT12X12

design -reset
read_verilog mul_gw5a.v
chparam -set X_WIDTH 9 -set Y_WIDTH 16 -set A_WIDTH 25
hierarchy -top top
proc
synth_gowin -family gw5a
cd top
select -assert-count 1 t:MULTALU27X18
select -assert-none t:MULT12X12


# unsigned 26x17 -> 1 MULTALU27X18 (DSP_SIGNEDONLY widens to signed 27x18)
design -reset
read_verilog mul_gw5a.v
chparam -set X_WIDTH 26 -set Y_WIDTH 17 -set A_WIDTH 43
hierarchy -top top_u
proc
synth_gowin -family gw5a -top top_u
cd top_u
select -assert-count 1 t:MULTALU27X18
select -assert-none t:MULT12X12


# unsigned 12x12: the +1 signedness bit overflows both the half-tile and the
# 27x18 rules' min-width checks, so it decomposes (known limitation -- fixing
# it would regress signed 12x12 from a half-tile to a full tile).
design -reset
read_verilog mul_gw5a.v
chparam -set X_WIDTH 12 -set Y_WIDTH 12 -set A_WIDTH 24
hierarchy -top top_u
proc
synth_gowin -family gw5a -top top_u
cd top_u
select -assert-none t:MULTALU27X18


# 2x2 -> trivial multiplies stay in LUTs (min-width floor of 4)
design -reset
read_verilog mul_gw5a.v
chparam -set X_WIDTH 2 -set Y_WIDTH 2 -set A_WIDTH 4
hierarchy -top top
proc
synth_gowin -family gw5a
cd top
select -assert-none t:MULTALU27X18
select -assert-none t:MULT12X12
```

vanhanit (commented):
I had Claude Fable 5 implement the same/similar fixes before I found you had opened a PR.
So instead of going forward with a colliding PR I thought that I could try and give a review on yours instead.

Thanks for this and hope the reviews help.

vanhanit:
Just a note, I'm trying to get everything verified as well. The comments were just from a review iteration.

tests/arch/gowin/mul_gw5a.v
olby: I would argue that the DSP_SIGNEDONLY + 1-bit logic shouldn't be verified by test in this PR. But if we would like to test it then "unsigned twin" isn't the correct approach. The + 1 bit logic will have different boundary conditions that should be targeted with other test cases not a twin.

techlibs/gowin/synth_gowin.cc
olby: I verified that this is an issue in my version. But I need to spend some time with the mul2dsp logic to understand if the repeated rule is a clever but hard to maintain hack or the proper solution.

tests/arch/gowin/mul_gw5a.v
vanhanit: I myself are not able to argue for or against, but Claude agrees with you 👍🏻 

> Fair point, and on reflection I agree with dropping it: `DSP_SIGNEDONLY` and its +1-bit widening aren't logic this PR introduces (that's existing `mul2dsp.v` behavior), so re-proving its general correctness isn't this PR's job. And even the two cases I had didn't fully address what you're describing -- they targeted the post-widen MAXWIDTH edge (unsigned 26x17 -> widens to 27x18, exact fit; unsigned 12x12 -> widens to 13x13, overflows and decomposes), but a "twin" module never touched the pre-widen MINWIDTH edge (original width sitting exactly at the floor vs. one under it), which is arguably the more interesting gw5a-rule-specific interaction. Rather than half-cover it, I'll drop the unsigned cases from this PR.

techlibs/gowin/synth_gowin.cc
vanhanit: Asked Claude about this as well, and it says there is a bug also in the other implementations and it thinks (and I agree) that the issue should be fixed at the root so the other families also get fixed.

> You're right to be wary -- the repeated rule is doing real work but it's disguised.
I instrumented it: the {18,27,...,$__MUL18X27} rule never actually maps a MUL18X27
tile (0 instantiations across every shape and a real ~20-multiplier design). mul2dsp
always commutes to A>=B, so A<=18 can never hold the wide operand -- the rule is a
pure orientation-normalizer whose only effect is the commute, feeding the repeated
{27,18,13,4} rule. The $__MUL18X27 wrapper is dead code.

> The reason it's needed at all: mul2dsp checks DSP_A/B_MINWIDTH on the *pre-commute*
orientation, and "max(operands) >= 13" isn't expressible as a pre-commute A/B floor on
an asymmetric tile. A swapped-min pair without the repeat doesn't fix it (signed 9x16 ->
2 MULT12X12 + fabric adder); notably gw1n's own {18,18,10,4}/{18,18,4,10} pair has the
exact same latent bug (9x16 -> 2 MULT9X9).

> A cleaner fix is to address it where it lives, with an opt-in mul2dsp flag that checks
the min-widths against the post-commute orientation:

    `ifdef DSP_MINWIDTH_AFTER_SWAP
        else if (`MAX(A_WIDTH, B_WIDTH) < `DSP_A_MINWIDTH) ...   // and MIN for B
    `else
        else if (A_WIDTH < `DSP_A_MINWIDTH) ...
    `endif

> With -D DSP_MINWIDTH_AFTER_SWAP the gw5a table collapses to three rules
({27,36,4,19}, {27,18,13,4}, {12,12,4,4}) -- no normalizer, no repeat, no dead wrapper.
I tested it: every boundary shape maps correctly (both orientations of 9x16/5x20/4x18 ->
one tile), a real design is unchanged (21 MULTALU27X18), the full tests/arch/gowin suite
passes, and gw1n/gw2a are byte-identical since it's opt-in. Happy to fold that in if you're
OK touching common/mul2dsp.v; otherwise I'll keep the self-contained version but rename the
rule + comment so the normalizer role is explicit.

vanhanit:
I think that Claude might have fixed the root cause in this patch file (should apply on the current latest commit you have).

[suggestion.patch](https://github.com/user-attachments/files/30046934/suggestion.patch)

### Root cause

`mul2dsp` always swaps a multiplier's operands so the wider one is on `A` (`A_WIDTH >= B_WIDTH`) before it maps to a DSP tile. But it checks the `DSP_A_MINWIDTH` / `DSP_B_MINWIDTH` limits **before** that swap, on the original operand positions.

That's fine for a **symmetric** tile (e.g. 18×18). It breaks on an **asymmetric** tile like the GW5A's 27×18: a product whose wide operand arrives on `B` — e.g. a signed `9×16` — is rejected, because the narrow `A` (9) fails the `A` min-width (13), even though after the swap it fits perfectly (16 on the 27-bit side, 9 on the 18-bit side). So instead of using one tile it falls back to splitting across smaller tiles + LUTs.

### The fix

Check the min-widths against `MAX(A_WIDTH, B_WIDTH)` and `MIN(A_WIDTH, B_WIDTH)` instead of `A` and `B` directly. That matches the orientation the cell will actually have after the swap, so the check no longer depends on which input happened to be wider.

For a **symmetric** min-width this is exactly the old behaviour (`A < k || B < k` is the same as `min(A,B) < k`), so nothing changes for existing users. Only **asymmetric**-tile rules are affected, and only for the better.

### What it fixes, and for which families

An asymmetric DSP tile now accepts a product regardless of which input is wider.

- **GW5A** (this PR): both `16×9` and `9×16` map to a single `MULTALU27X18`. This  also lets the family's rule table collapse to one 27×18 rule (no  normalise-and-retry, no dead `$__MUL18X27` wrapper).
- **gw1n / gw2a** (18×18 tile) and **Lattice Nexus** (36×18 tile) had the same  latent issue and are fixed for free — e.g. gw1n signed `9×16` was 2 `MULT9X9` +  a fabric adder, now 1 `MULT18X18`.

All `tests/arch` multiplier suites pass across every family (gowin, nexus, xilinx, ice40, ecp5, gatemate, intel_alm, microchip, analogdevices).

techlibs/gowin/synth_gowin.cc
olby: * Fix for mul2dsp landed in https://github.com/YosysHQ/yosys/pull/6047 so we no longer need the hack.
* I fixed the invalid B_MINWIDTH for 27x18 caught by Fable.

I didn't change the min width to 4 to not do narrow multiplication on the DSP. In most cases that would be the correct thing to do but I like the idea of having this only consider fit and not resource planning.

techlibs/gowin/dsp_map_gw5a.v
olby: No longer needed after https://github.com/YosysHQ/yosys/pull/6047

tests/arch/gowin/mul_gw5a.ys
olby: Leaving resource planning out of this PR with the hope of finding a better way to implement it.

tests/arch/gowin/mul_gw5a.ys
olby: No longer needed after https://github.com/YosysHQ/yosys/pull/6047 as order is now normalized.

olby:
> ### The fix
> 
> Check the min-widths against `MAX(A_WIDTH, B_WIDTH)` and `MIN(A_WIDTH, B_WIDTH)` instead of `A` and `B` directly. 

https://github.com/YosysHQ/yosys/pull/6047 solve this in a slightly different way. It moves the width order normalization before the width check.

tests/arch/gowin/mul_gw5a.v
```diff
@@ -0,0 +1,11 @@
+// Signed multiply for GW5A DSP boundary tests (the GW5A multipliers are
```
Ravenslofty, line 1: This file is just `tests/arch/common/mul.v`; can you use that instead?

Ravenslofty (changes requested):

PR #6034, opened 2026-07-13
cmake: Use system libs when available

mmicko:
This one should handle issues noted with  https://github.com/YosysHQ/yosys/issues/6028, https://github.com/YosysHQ/yosys/pull/6031 and  https://github.com/YosysHQ/yosys/pull/6032 but solving it using new parameters available for configuring slang build.


mmicko:
@whitequark wondering if we should have a way to force building with vendored libraries only (even if system lib exists), mostly to make sure those are always working correctly ?

whitequark (approved):
LGTM, I like your approach here!

whitequark:
> wondering if we should have a way to force building with vendored libraries only (even if system lib exists), mostly to make sure those are always working correctly ?

Yeah, I think this would be a useful addition.

mmicko:
@whitequark any good idea for option name  ?

Something like :
```option(YOSYS_USE_BUNDLED_LIBS "Use bundled third-party libraries" OFF)```

Not  sure if we should use YOSYS_WITH/WITHOUT for this as well

whitequark (approved):
Thanks!

PR #6035, opened 2026-07-13
gowin: infer DSP multiply-accumulate for the GW5A family

olby:
Maps $macc_v2 cells (one signed <=27x18 product + one <=48-bit addend) to MULTALU27X18 with the C addend enabled, so a*b + c maps to a single DSP block instead of a multiply + fabric adder. Uses the same macc_map_gw5a.v techmap pattern as the existing dsp_map.v.

The C addend requires .DYN_C_SEL("TRUE") + .CSEL(1'b1) -- gowin_pack reads the CSEL port, not the C_SEL parameter. Without this, the addend is  silently dropped (DOUT = AB, not AB + C).

alumacc + the macc techmap run before wreduce (gw5a only). wreduce narrows $mul Y (48->45) but not the connected $add A (stays 48), and alumacc cannot merge mismatched widths. Running alumacc first ensures the ports still match, so $mul is absorbed into $macc_v2 with the addend intact.

Verified: Verilator (512/512 signed-boundary vectors), and on silicon (Tang Primer 25K: 8-vector self-check OK, inferred from RTL a*b + c through the full flow with flatten). Boundary tests cover each _TECHMAP_FAIL_ condition (oversized A/B/C, negated addend, two products, unsigned).

### DIFF
--- a/techlibs/gowin/CMakeLists.txt
+++ b/techlibs/gowin/CMakeLists.txt
@@ -52,4 +52,5 @@ yosys_pass(synth_gowin
 		lutrams_map.v
 		lutrams.txt
 		dsp_map.v
+		macc_map_gw5a.v
 )
--- a/techlibs/gowin/macc_map_gw5a.v
+++ b/techlibs/gowin/macc_map_gw5a.v
@@ -1,0 +1,59 @@
+module \$macc_v2 (A, B, C, Y);
+
+	parameter NPRODUCTS = 0;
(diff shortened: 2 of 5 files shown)


Ravenslofty:
The fragility of `$macc` and friends is generally why we avoid trying to techmap it directly to a target primitive.

Typically, we would use `mul2dsp` (as in #6033) to break the multiply into blocks and then use pmgen to pattern match surrounding logic to merge it in (e.g. `ice40_dsp`).

PR #6036, opened 2026-07-14
`proc_dlatch`: error on implicit latches early [sc-645]

hexratcc:
This PR makes latches explicitly requested with `always_latch` exempt from the `proc -latches` policy and tags the generated cell with an `always_latch` attribute. `check -nolatches`/`check -latchonly` skip tagged cells.

Latch-less synth flows (`ice40`, `efinix`, `fabulous`, `nanoxplore`, `lattice`, `quicklogic-pp3`) now forward `-latches error` (the default) directly to `proc`, so an accidental latch fails right at elaboration with the signal and source location, while `always_latch` synthesizes cleanly.

I've left message severity levels in generic `synth` unchanged, as it currently fails in a bunch of tests, but the user may elevate them via `-latches error`.

followup to https://github.com/YosysHQ/yosys/pull/5963


Ravenslofty (changes requested):
`synth_analogdevices` and `synth_intel_alm` too, from recollection

passes/cmds/check.cc
```diff
@@ -142,10 +143,10 @@ struct CheckPass : public Pass {
 			// latch-only mode only flags latches, skipping the (potentially false-positive mid-flow) undriven/driver/loop checks below
 			if (latchonly) {
 				for (auto cell : module->cells())
-					if (
+					if ((
 						cell->type.in(ID($dlatch), ID($adlatch), ID($dlatchsr)) ||
 						cell->type.begins_with("$_DLATCH_") || cell->type.begins_with("$_DLATCHSR_")
-					) {
+					) && !cell->get_bool_attribute(ID::always_latch)) {
```
KrystalDelusion: I wonder if it's worth putting the attribute check first just to avoid the `if ((\n`

docs/source/using_yosys/verilog.rst
```diff
@@ -172,7 +172,10 @@ Verilog Attributes and non-standard features
 - The frontend sets attributes ``always_comb``, ``always_latch`` and
   ``always_ff`` on processes derived from SystemVerilog style always blocks
   according to the type of the always. These are checked for correctness in
-  ``proc_dlatch``.
+  ``proc_dlatch``. Latches inferred from ``always_latch`` processes are exempt
+  from the ``proc -latches <info|warn|error>`` reporting policy, and the
+  generated latch cells carry the ``always_latch`` attribute, which also
+  exempts them from `check` ``-nolatches``/``-latchonly``.
```
KrystalDelusion: `-latchonly` already references that it performs the `-nolatches` check in the help text, I think we can skip it in documentation and just focus on `-nolatches` as being the user-facing part.

techlibs/analogdevices/synth_analogdevices.cc
```diff
@@ -244,10 +252,17 @@ struct SynthAnalogDevicesPass : public ScriptPass
...
 			}
+			if (args[argidx] == "-latches" && argidx+1 < args.size()) {
+				latches = args[++argidx];
+				continue;
+			}
 			break;
 		}
 		extra_args(args, argidx, design);
 
+		if (latches != "info" && latches != "warn" && latches != "error")
```
widlarizer: This suggests to me that this should be an enum type that you cast into, especially as it's duplicated across these passes. Outside the scope of this PR more generally, anything duplicated across synth_* is inconvenient and should be refactored out so that it is clear what the differences between the passes are. For now let's figure out a place that shared parts of the interface handling of these passes can go and just use it for the latches option. I've done something like that with EquivBasicConfig

widlarizer (changes requested):

techlibs/analogdevices/synth_analogdevices.cc
```diff
@@ -439,6 +446,9 @@ struct SynthAnalogDevicesPass : public ScriptPass
 		}
 
 		if (check_label("map_ffs")) {
+			if (latches.policy == LatchPolicy::Error || help_mode)
+				run("check -latchonly -assert", "(only if -latches error, the default)");
+			run("techmap -map +/analogdevices/latches_map.v");
```
Ravenslofty: Um, latches are explicitly not supported under `synth_analogdevices`; latches *must* error, which happened through `dfflegalize` before.

hexratcc: Good catch, missed that, thanks.

widlarizer (approved):

Ravenslofty (approved):

PR #6037, opened 2026-07-14
README: cmake build and test instructions

widlarizer:
I'm struggling a bit to use CMake with Yosys so I'm adapting the README with what I learned, but I'm pretty unsure about if it's the right approach, so, draft for now.

IMO Ninja still can't be used since tests can't be run through cmake if ninja is used. If you configure for Ninja to build, and then you decide to run tests, requiring users to reconfigure to make is inconvenient, and so is requiring users to use undocumented make invocations. So I'm making it clear that GNU Make is required. An alternate approach would be to document how to run tests with make directly, would it be preferrable to forbid using CMake to run tests instead? Maybe making it an error?

I'm also fixing the preset file name, which prevented it from doing anything. Also, `-S .` can stay implicit, simplifying the release build cmake invocation.

Why is the release build specified with both `-DCMAKE_BUILD_TYPE=Release` at config time and `--config Release`? Does it have something to do with Ninja Multi-Config? I would far more prefer to configure once and choose a build config at build time. Should we instead turn asking cmake for tests into an error and document how to run tests without cmake to be able to benefit from Ninja Multi-Config?

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,8 +75,7 @@ sure to call e.g.
 	$ git submodule update --init
 
 A C++ compiler with C++20 support is required as well as some standard tools
-such as GNU Flex, GNU Bison (>=3.8), CMake (>=3.28), Make (or other CMake
-generator such as Ninja), and Python (>=3.11). Some additional tools: readline,
+such as GNU Flex, GNU Bison (>=3.8), CMake (>=3.28), GNU Make, and Python (>=3.11). Some additional tools: readline,
 libffi, Tcl and zlib; will be used if available but are optional. Graphviz and
 Xdot are used by the `show` command to display schematics.
 
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ file which enables ccache and sets the default compiler to clang when calling
 	"version": 1,
 	"configurePresets": [
 		{
-			"name": "default",
+			"name": "clang",
 			"binaryDir": "build",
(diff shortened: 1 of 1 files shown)


mmicko:
Running tests using cmake target makes sense if you are running all of them, or running just unit  tests, but in most cases you would go into `tests/` and run `make` from there or run failing tests or newly added tests separately.
If your build directory is `build` then you are all set with running make, otherwise setting  `BUILD_DIR` environment variable or setting it just for individual make run will do.

whitequark:
> Why is the release build specified with both `-DCMAKE_BUILD_TYPE=Release` at config time and `--config Release`?

The former works on with most generators, the latter is required for MSVC.

KrystalDelusion:
> tests can't be run through cmake if ninja is used

They definitely can, they just don't parallelize (and possibly don't report progress by default?  I don't remember if that was fixed in the `CMakeLists.txt`)

>  An alternate approach would be to document how to run tests with make directly

This is the case in the [rst docs](https://yosyshq.readthedocs.io/projects/yosys/en/latest/yosys_internals/extending_yosys/test_suites.html).

> I'm also fixing the preset file name, which prevented it from doing anything. 

I'm unclear what you mean here exactly, but if it's about the `CMakeUserPresets.json` file then that is what I've been using and it works fine for me.

PR #6042, opened 2026-07-15
Add `symfpu` pass

KrystalDelusion:

Using SAT solvers to prove floating point arithmetic is (computationally) difficult. Having a library of optimized models makes it more feasible. Doing this within Yosys means we can (theoretically) output these optimized models for any solver backends supported by Yosys.

Use [martin-cs/symfpu](https://github.com/martin-cs/symfpu) to generate SMT-LIB optimized verilog models of IEEE-754 operators. Forked at [YosysHQ/symfpu](https://github.com/YosysHQ/symfpu/tree/floatWithStatusFlags) in order to inject support for:
1. operators to return the exceptions encountered (status flags)
  a. with tininess determined after rounding, as in the RISC-V specification
  b. includes float/int conversions
2. optional alternative sqrt/division operators 
  a. useful for testing control logic with a simpler arithmetic operation (a rotate and a subtraction respectively)

Tests included at `tests/symfpu`, largely based on the concept of problem classes and satisfiability (e.g. two sufficiently large operands must overflow when multiplied, or that add/sub can never underflow) and reachability (e.g. for a pair of unconstrained inputs, an add operation must be able to overflow).  These are largely intended for checking the status flag behavior, rather than the arithmetic, since that should be unchanged from the underlying library (which has its own tests).  SymFPU also includes preconditions/postconditions/invariants which get turned into `assert` statements (which get checked at run time provided `NDEBUG` isn't set) and `$assert` cells (which get checked during testing for single precision floats across all operators/rounding modes, or can be checked by the user with `sat -prove-asserts -verify`).

I have locally tested the generated models for equivalence against HardFloat (which also helped with some of the trickier points around tininess), but haven't included it here because it uses SBY to do the checking.  It uses a wrapper to map the HardFloat ports to match the SymFPU interface, and then uses `assert (o_gate == o_gold);` for the signals.


KrystalDelusion:
[sc-666]

passes/cmds/symfpu.cc
```diff
@@ -0,0 +1,799 @@
...
+ *  WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ *  ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ *  OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *
+ */
+
+#include "kernel/log_help.h"
+#include "kernel/yosys.h"
+
+#include "libs/symfpu/baseTypes/shared.h"
```
mmicko: There is no need to have `libs/` prefix, so it should be removed from includes since library include is added in cmake file

KrystalDelusion: Thanks!

mmicko (approved):

PR #6045, opened 2026-07-15
const2ast: handle constant parsing from the command-line again

### LINKED ISSUE #6009: Multi-bit unsized constant used in pass invocation triggers segfault instead of error
### Version

0.66

macOS

### Reproduction Steps

Version: Yosys 0.66 (git sha1 7f8fdfd8d7bc08c749a2a969388d3425d4f369d5), clang++ 21.0.0, macOS (Apple silicon, Homebrew build).

Command (SIGSEGV, exit 139):
yosys -p "read_rtlil bug.il; memory; sat -prove out 8'00000101"

Minimal bug.il — a module whose only output is an asynchronous read of an uninitialized memory:
```
module \top
  wire width 4 input 1 \addr
  wire width 8 output 2 \out
  wire width 8 $r
  memory width 8 size 16 \mem
  cell $memrd $rd
    parameter \MEMID "\\mem"
    parameter \ABITS 4
    parameter \WIDTH 8
    parameter \CLK_ENABLE 0
    parameter \CLK_POLARITY 0
    parameter \TRANSPARENT 0
    connect \CLK 1'x
    connect \EN 1'x
    connect \ADDR \addr
    connect \DATA $r
  end
  connect \out $r
end
```

Isolation notes (all on the same bug.il):
- read_rtlil bug.il; memory → OK (exit 0). The crash is not in read_rtlil or memory.
- read_rtlil bug.il; memory; sat (find-a-model) → OK (exit 0).
- read_rtlil bug.il; memory; sat -prove out 8'00000101 → SIGSEGV.
- read_rtlil bug.il; memory_collect; memory_map; sat -prove … → also SIGSEGV.
- Adding/removing a $meminit, and the specific prove-value, make no difference — the crash is the sat -prove path over a $memrd-derived signal (present regardless of whether memory mapped it to logic).

So: sat succeeds but sat -prove crashes on the identical circuit, whenever the proven signal derives from a memory read. bug.il is at .../scratchpad/bug.il if you want the exact file to attach.

### Expected Behavior

No SIGSEGV

### Actual Behavior

SIGSEGV

widlarizer:
Fixes #6009. AST nodes produced by reading verilog always have locations, but that isn't the case for AST nodes created by parsing Verilog constants provided in the invocation of a pass. The new behavior is `INTERNAL:0: ERROR: Unsized constant must have width of 1 bit, but have 8 bits!` due to the use of `8'00000101` and not `8'b00000101`

### DIFF
--- a/frontends/verilog/const2ast.cc
+++ b/frontends/verilog/const2ast.cc
@@ -45,11 +45,13 @@ using namespace AST;
 using namespace VERILOG_FRONTEND;
 
 void ConstParser::log_maybe_loc_error(std::string msg) {
-	log_file_error(*loc.begin.filename, loc.begin.line, "%s", msg);
+	log_file_error(loc.begin.filename ? loc.begin.filename->c_str() : "INTERNAL",
+    loc.begin.line, "%s", msg);
 }
 
 void ConstParser::log_maybe_loc_warn(std::string msg) {
-	log_file_warning(*loc.begin.filename, loc.begin.line, "%s", msg);
+	log_file_warning(loc.begin.filename ? loc.begin.filename->c_str() : "INTERNAL",
+    loc.begin.line, "%s", msg);
 }
 
 // divide an arbitrary length decimal number by two and return the rest
--- a/frontends/verilog/const2ast.cc
+++ b/frontends/verilog/const2ast.cc
@@ -36,6 +36,7 @@
 
 #include "verilog_frontend.h"
 #include "kernel/log.h"
+#include "kernel/register.h"
 #include <string.h>
 #include <math.h>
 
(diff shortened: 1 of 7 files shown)


KrystalDelusion (changes requested):
Thanks!  I'm not sure the `sat` example is the best option, but could we get a test case in here to make sure this doesn't break again?

I think this is worth merging as-is, but I do wonder if for the example with `sat` being in a script, and the error is after a few lines of output where it appears to be working, the error message doesn't make the cause immediately obvious.  Is there any way to set the source location as the line in the script file?  Or even just including the string being parsed would probably help.

hexratcc:
@KrystalDelusion I've added the regression test, and constants parsed from the command line now report the script `file:line` plus the offending text.

KrystalDelusion (approved):
V cool.

Interestingly, if the `setattr -set foo 2'd7 t:*` (for example) is done manually after calling `shell` from a script, the error will point to the `shell` line, but if done entirely from the command line it gracefully falls back to not printing any source location.  At a quick glance it also appears to handle nested `script` calls correctly (or at least, calling `script const_cmdline.ys` from a yosys prompt passes the `logger` call and then reverts back to no source location).

May be worth adding a quick test file to make sure we don't accidentally get a regression on the nesting behavior, but otherwise I think this is good.

PR #6049, opened 2026-07-16
gatemate: fix CC_FIFO_40K sim for non 2^n width usages

mrcmry:
The fifo address calculation  is not correct for the the non power of 2  fifo WIDTHs. When using only power of 2 data, the behavior of the fifo is correct (for example 32 bits in 40 bits mode). When using the full bits, the MSB is incorrect (bit 39:32 for 40 bits mode for example).

This PR correct the address computation.

I don't have a working build of Yosys and I am not very familiar with the project. So I was not able to test the fix with unit tests in Yosys. Perhaps someone that know better the project can.

I have some fifo unit tests with a wrapper of CC_FIFO_40K in SpinalHDL. With power of 2 width my test are passing using the unpatched version, but not non power of 2. With this fix it works with various sizes. The test is done with verilator.

Also, perhaps this file come from somewhere else and the upstream should also be notified of the bug/fix.

### DIFF
--- a/techlibs/gatemate/cells_sim.v
+++ b/techlibs/gatemate/cells_sim.v
@@ -1683,8 +1683,8 @@ module CC_FIFO_40K (
 	 end
 
 	always @(*) begin
-		fifo_wraddr = {wr_pointer[tp-1:0], {(15-tp){1'b0}}};
-		fifo_rdaddr = {rd_pointer[tp-1:0], {(15-tp){1'b0}}};
+		fifo_wraddr = wr_pointer[tp-1:0] * B_WIDTH;
+		fifo_rdaddr = rd_pointer[tp-1:0] * A_WIDTH;
 
 		rdaddr = {rd_pointer[tp], rd_pointer_int[tp-1:0]};
 		wraddr = {{(15-tp){1'b0}}, wr_pointer[tp], wr_pointer_int[tp:0]};


KrystalDelusion:
Previously both addresses were set with `tp`, which is derived from `A_WIDTH`, but your change uses `A_WIDTH` for the write and `B_WIDTH` for the read.  Is that intentional and/or related?

> I have some fifo unit tests with a wrapper of CC_FIFO_40K in SpinalHDL. With power of 2 width my test are passing using the unpatched version, but not non power of 2. With this fix it works with various sizes. The test is done with verilator.

Can you share any of those tests?

mrcmry:
> Previously both addresses were set with `tp`, which is derived from `A_WIDTH`, but your change uses `A_WIDTH` for the write and `B_WIDTH` for the read. Is that intentional and/or related?

You mean `A_WIDTH` for the read and `B_WIDTH` for the write in your message, probably.

Conceptually, the fifo use B port for write and A for read (except for dual port implementation):

```verilog
	always @(posedge fifo_wrclk)
	begin
		for (i=0; i < B_WIDTH; i=i+1) begin
			if (ram_we && B_BM[i])
				memory[fifo_wraddr+i] <= B_DI[i];
		end
	end
```
```verilog
			always @(posedge fifo_rdclk)
			begin
				for (i=0; i < A_WIDTH; i=i+1) begin
					if (ram_en) begin
						A_DO_out[i] <= memory[fifo_rdaddr+i];
					end
				end
			end
```

But both with must be the same anyway:

```verilog
		if (A_WIDTH != B_WIDTH) begin
			$display("ERROR: The values of A_WIDTH and B_WIDTH must be equal.");
		end
```

It's not only a simulation limitation, in the datasheet DS1001 from August 2025 under section "2.4.9 FIFO Application" it is stated "Widths of the input and output buses must be equal."

So it doesn't really matter, but writing it like this reflect how conceptually the BRAM work in a better way (separate counter for separate clock domains).

mrcmry:
> Can you share any of those tests?

It would require me to setup a git repos with instructions to how to setup the SpinalHDL toolchain and run the tests. It is ultimately my goal, but there is still much cleanup to do and I intend to do this later when it works more or less on the hardware.

The fifo tests are very basic, only test TDP with constant fifo almost full flag config. It doesn't checks error flags, corner cases with fifo full, empty. The test is also with the two TDP clocks being the same clock. I only setup this because I wanted to be sure there was this non 2^n bug.

All in all, I think it's not worth my cleanup effort and your setup effort (and SpinalHDL learning curve) for a test what is very far from extensive. I hope there is somewhere a test suite for this sim component.

Ideally, comparison test of output of the synthesized CC_FIFO_40K sim and  the  real hardware fifo on the real FPGA  would be the ultimate test. I don't know if the sim version take care of metastability or does only work on the "perfect" FF of the simulators. But nextpnr does not allows cross domain constraints anyway, so this not be possible for dual clock anyway. Perhaps a test with the same clock domain would be something feasible.

Perhaps these tests already exist and we only need to find who has them/where they are.

KrystalDelusion:
> It would require me to setup a git repos with instructions to how to setup the SpinalHDL toolchain and run the tests.

You mentioned the tests are done in Verilator, are you able to share what Verilator runs and what gets input to Yosys?  If there are a series of tests, just one that fails without the fix here is fine.

KrystalDelusion:
> I was not able to test the fix with unit tests in Yosys

It seems like there are currently no tests for the CC_FIFO_40K, only CC_BRAM_40K, and even then those are mapping tests not simulation tests.  

@pu-cc are you able to take a look at this PR?

mrcmry:
I used the same modified model as this PR in my simulation in this project here:

https://github.com/BrenaySoc/BrenaySoc/blob/382a98995c3150cdfb19678cca84334f56632a0b/src/main/scala/brenay/lib/gatemate/SioSerializerCc.scala#L93

I realize that I hardcoded a value of 10, that is not very wise (the system use a 32 bit fifo for a 64 byte burst, so around 16 entry max, I got lucky in the randomized unit tests. It works also with 10 on the hardware, but there the worst case is a few entries.

Perhaps I will find time in the future to  do signal comparison of the real CC_FIFO_40K and the sim version in the real fpga to confirm the fix works.

mrcmry:
@ShinyKate I open sourced my whole code so you can find my test at this place:

https://github.com/BrenaySoc/BrenaySoc/blob/dev/src/test/scala/brenay/lib/gatemate/GmStreamFifoCCSim.scala

I use the CC_FIFO_40K as fifo for an interface from a vexiiriscv at 25 MHz to a PSRAM at 100 MHz on the gatemate olimex board. This this fix, it works fine with 40 bit width in simulation and it also work with the real fifo.

Tell me how you want to test this and if I can help in any way to do this. But perhaps @pu-cc has already a test upstream for this.

PR #6050, opened 2026-07-16
autoname: rewrite

widlarizer:
Significantly faster than main, better performance as #6022 on the perf test that PR provides and way better on hub-shaped designs as well. No more fixpoint outer loop or magic `2*score` that would force the outer loop to iterate senselessly, instead we use a priority queue based graph traversal. Each node records which node it's inheriting its name from and at what cost. Before names are applied to RTLIL (requiring churn in the name to wire and cell dicts, and IdString interning), they're collected in the graph representation as std::string, and only the ones that actually improve the cost are constructed as std::string.

@KrystalDelusion I had to "fix" how the test handles selections, but I'm still confused why it was the way it was, I think the test was pinning down a bug? The privately named $or cell matches the selection, so we should be able to propagate names from it to its neighbors once the cell becomes publicly named, right?

### DIFF
--- a/passes/cmds/autoname.cc
+++ b/passes/cmds/autoname.cc
@@ -17,95 +17,233 @@
  *
  */
 
+
 #include "kernel/yosys.h"
+#include <functional>
+#include <queue>
+#include <ranges>
 
 USING_YOSYS_NAMESPACE
 PRIVATE_NAMESPACE_BEGIN
 
-typedef struct name_proposal {
-	string name;
+// An "object" is a cell or a wire.
+// An object is "private" if its name starts with a '$'.
+//
+// The "autoname" pass renames private objects based on public object names,
+// with suffixes added to show the relationship to the public objects.
+// The pass chooses the "best" names based on minimizing their "cost".
+
+// The cost of a new name for a private object is the cost along a path
+// to a publicly named object.
+struct cost {
+	// score is a property of one "edge", not of the entire path.
+	// 0 if the neighbour drives the object, and the wire's fanout otherwise
+	unsigned int score = UINT_MAX;
+	// New name length
+	size_t length = SIZE_MAX;
+	// Earlier discovered connection wins
+	int edge_pos = 0;
+
+	auto operator<=>(const cost &) const = default;
+};
+
+// Endpoints of an "edge" are always a cell index and a wire index. Goofy, I know.
+// SigSig connections (module->connections()) don't count
+// as neighbors. This means that the equivalent of `assign $w1 = \w2;` won't lead
+// to $w1 being renamed.
+struct Edge {
+	// Cell port name at which the wire connects to the cell
+	IdString port;
+	int cell;
+	int wire;
+	bool cell_is_output;
 	unsigned int score;
-	name_proposal() : name(""), score(-1) { }
-	name_proposal(string name, unsigned int score) : name(name), score(score) { }
-	bool operator<(const name_proposal &other) const {
-		if (score != other.score)
-			return score < other.score;
-		else
-			return name.length() < other.name.length();
-	}
-} name_proposal;
+	// Edge's position in the edge list
+	int pos_in_cell;
+	int pos_in_wire;
+};
 
-int autoname_worker(Module *module, const dict<Wire*, unsigned int>& wire_score)
+// decide() finds shortest names for every object reachable from a public name,
+// cheapest-first. It doesn't apply the renames to the objects yet.
+//
+// commit() then renames in decide() order, which is topological, so a source
+// neighbour always already carries its final uniquify()d name.
+
+// A selected cell or wire
+struct node {
+	// Exactly one of cell/wire is set
+	Cell *cell = nullptr;
+	Wire *wire = nullptr;
+	vector<Edge> edges;
+	// Ignores module->connections() just as the rest of the code
+	unsigned int fanout = 0;
+	bool is_public = false;
+	bool renameable = false;
+
+	size_t name_length = 0;
+	// Node index from which we want to construct the rename
+	int from_node = -1;
+	// Cost for the edge from that node
+	cost c;
+	// Suffix to append to the name of that node
+	string suffix;
+	// Is this name final?
+	bool decided = false;
+};
+
+// Decides the order of exploring neighbors
+struct queue_item {
+	unsigned int score;
+	size_t length;
+	int index;
+
+	auto operator<=>(const queue_item &) const = default;
+};
+
+struct ModuleAutonamer
 {
-	dict<Cell*, name_proposal> proposed_cell_names;
-	dict<Wire*, name_proposal> proposed_wire_names;
-	name_proposal best_name;
+	Module *module;
 
-	for (auto cell : module->selected_cells()) {
-		if (cell->name[0] == '$') {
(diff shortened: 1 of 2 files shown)


gonsolo:
Cool, thanks for improving on my PR, but why are you dropping the perf test?

widlarizer:
Such a test would go against our demands for the reproducibility and hardware invariance of our test suite. Something that can fail based on a timeout can't be in our functional regression test suite. We keep some end to end synthesis perf tests that catch perf in hot, important parts like opt_expr and techmap, but other than that it's rarely worth more isolated perf regression testing since we have a lot of code that moves rarely like `autoname`, and perf regression testing has higher maintanance overhead over functional regression testing.

KrystalDelusion:
> we should be able to propagate names from it to its neighbors once the cell becomes publicly named, right?

Uh asterisk, it depends on how you treat the selection.  If you treat it as "only modify the selected parts of the design" then you shouldn't be able to apply names to anything that doesn't match the selection.  But if that's the case then it shouldn't have been able to rename the `$d` wire either.  The comment suggests that `$d` being renamed was expected, since it was the output of a selected cell... but then why wasn't `$e` renamed?

My inclination is that `autoname t:$or %co` should rename `$d`, `$name2`, and `$e`; but not `$c` (which would require `%x` instead).  Expanding the selection with `%co` doesn't affect the behavior of Yosys without this PR, and even extending it with `%x*` (which should be equivalent to the full design) only adds `$name2` to the list of things that get renamed; so I'm inclined to agree that there was a bug, though I don't know that I agree with the solution.

passes/cmds/autoname.cc
```diff
@@ -17,95 +17,238 @@
...
+#include <ranges>
 
 USING_YOSYS_NAMESPACE
 PRIVATE_NAMESPACE_BEGIN
 
-typedef struct name_proposal {
-	string name;
+// -----------------------------------------------------------------------
+// An "object" is a cell or a wire.
+// An object is "private" is its name starts with a '$'.
```
KrystalDelusion:
```suggestion
// An object is "private" if its name starts with a '$'.
```

passes/cmds/autoname.cc
```diff
@@ -17,95 +17,238 @@
...
+//
+// The "autoname" pass renames private objects based on public object names,
+// with suffixes added to show the relationship to the public objects.
+// The pass chooses the "best" names based minimizing their "cost".
```
KrystalDelusion:
```suggestion
// The pass chooses the "best" names based on minimizing their "cost".
```

passes/cmds/autoname.cc
```diff
@@ -17,95 +17,238 @@
...
+	int from_node = -1;
+	// Cost for the edge from that node
+	cost c;
+	// Suffix to append to the name of that node
+	string suffix;
+	// Is this name final?
+	bool decided = false;
+};
+
+// Decides the order of exploring neighbors.
```
KrystalDelusion: Minor nitpick that some comments end with a fullstop, but some don't.

passes/cmds/autoname.cc
```diff
@@ -17,95 +17,238 @@
...
 	}
 
-	for (auto &it : proposed_wire_names) {
-		if (best_name < it.second)
-			continue;
-		IdString n = module->uniquify(IdString(it.second.name));
-		log_debug("Rename wire %s in %s to %s.\n", it.first, module, n.unescape());
-		module->rename(it.first, n);
-		count++;
+	// Expand a public (or newly decided) node ever expands, so the name it lends
```
KrystalDelusion: phrasing?

passes/cmds/autoname.cc
```diff
@@ -17,95 +17,238 @@
...
+	void offer(int from, int n, const Edge &edge, int edge_pos)
+	{
+		node &nd = nodes[n];
+		if (!nd.renameable || nd.decided)
+			return;
+		string suffix = nd.cell
+			? stringf("_%s_%s", nd.cell->type.unescape(), edge.port.unescape())
+			: stringf("_%s", edge.port.unescape());
+		cost c{edge.score, nodes[from].name_length + suffix.length(), edge_pos};
+		if (!(c < nd.c))
```
KrystalDelusion: Why not `c >= nd.c`?

passes/cmds/autoname.cc
```diff
@@ -17,95 +17,238 @@
...
+					};
+					nodes[ci].edges.push_back(edge);
+					nodes[wi].edges.push_back(edge);
+				}
 			}
 		}
+
+		for (auto &nd : nodes) {
+			IdString name = nd.cell ? nd.cell->name : nd.wire->name;
+			nd.is_public = !is_private(name);
```
KrystalDelusion: `nd.is_public` is only set here, and `is_private` is only used here... why not make them consistent?

passes/cmds/autoname.cc
```diff
@@ -17,95 +17,238 @@
...
+// The pass chooses the "best" names based minimizing their "cost".
+
+
+
+// The cost of a new name for a private object is the cost along a path
+// to a publicly named object.
+struct cost {
+	// score is a property of one "edge", not of the entire path.
+	// 0 if the neighbour drives the object, and the wire's fanout otherwise
+	unsigned int score = ~0U;
```
KrystalDelusion:
```suggestion
	unsigned int score = UINT_MAX;
```

More readable equivalent

tests/various/autoname.ys
```diff
@@ -180,15 +180,15 @@ design -save order_test
...
 logger -expect log "Rename wire .d in top to or_Y" 1
 logger -expect log "Rename cell .name2 in top to or_Y_.or_B" 1
-logger -expect log "Renamed 2 objects" 1
+logger -expect log "Rename wire .e in top to or_Y_.or_B_Y." 1
+logger -expect log "Rename wire .c in top to or_Y_.or_B_A." 1
+logger -expect log "Renamed 4 objects" 1
 debug autoname t:$or
 logger -check-expected
 
-# $name gets shortest name (otherwise bcd_$__unknown_B)
```
KrystalDelusion, line 187: Is there a reason this comment was removed?

tests/various/autoname.ys
```diff
@@ -180,15 +180,15 @@ design -save order_test
...
-logger -expect log "Rename cell .name in top to a_.__unknown_A" 1
-# another output wire
-logger -expect log "Rename wire .e in top to or_Y_.or_B_Y" 1
-logger -expect log "Renamed 4 objects" 1
+logger -expect log "Rename cell .name in top to a_.__unknown_A." 1
```
KrystalDelusion: I probably wouldn't include the trailing `.`, at least not without escaping it (but if you do, then you should add it for all of them).

KrystalDelusion (changes requested):

passes/cmds/autoname.cc
widlarizer: Leftover from previous work, removing the function since it's not actually helping readability anymore

passes/cmds/autoname.cc
widlarizer: Should be better now

KrystalDelusion (approved):

PR #6055, opened 2026-07-20
transparent memory verilog pattern

### LINKED ISSUE #5082: write_verilog: Vivado transparent memory port regression
### Version

0.40

Linux

### Reproduction Steps

1. Emit Verilog code for memory with one write port and read port that is transparent to write port.

Code is emitted with https://github.com/YosysHQ/yosys/blob/aa30589c123844bc576fc3c9938157b5db35a7c1/backends/verilog/verilog_backend.cc#L732 creating registered address input on read port.

Example:
```verilog
  reg [20:0] tag_mem_0 [127:0];
  always @(posedge clk) begin
    if (tag_mem_wp__en)
      tag_mem_0[tag_mem_wp__addr] <= tag_mem_wp__data;
  end
  reg [6:0] _2_;
  always @(posedge clk) begin
    _2_ <= tag_mem_rp__addr;
  end
  assign tag_mem_rp__data = tag_mem_0[_2_];
```

2. Synthesize it with Vivado

Tested on Vivado 2024.2
This issue comes from Amaranth language: https://github.com/amaranth-lang/amaranth/issues/1589

### Expected Behavior

Vivado synthesizes it correctly to block ram with transparency.

I assume it worked in previous Vivado versions, because this issue was not reported earlier, with successful Xilinx designs made.

### Actual Behavior

Vivado synthesis emits warning `The block RAM may get collision error if read and write address collide` and leads to non-functional gateware.

Workaround is to add `(* rw_addr_collision = "yes" *)` attribute for each memory declaration. Without it, the default behaviour is `design writes to the same address it is reading from, the RAM output is unpredictable`. Warnings are not emitted for non-transparent ports.
This is problematic for HDL generated code, that depends on verilog backend for external synthesis.

qbojj:
closes #5082 

The pattern shown in the issue got implemented with minor change - the write port data is only latched when there was address address collision.
Additionally the pattern
```v
always @(...) begin
   _3_ <= RD_ADDR;
end
assign RD_DATA = mem[_3_];
```
got removed.

Some changes were done to make enable signals work:
1. the bypass logic is bitwise (is_bypassed, bypass_data) tuples rather than a single common is_bypassed bit. This should be simply optimised by the verilog consumer, as it should detect all bits from is_bypassed vector are assigned with the same values at the same times, allowing for collapsing them to a single wire.

### DIFF
--- a/backends/verilog/verilog_backend.cc
+++ b/backends/verilog/verilog_backend.cc
@@ -626,15 +626,13 @@ void dump_memory(std::ostream &f, std::string indent, Mem &mem)
 			lof_reg_declarations.push_back( stringf("reg [%d:0] %s;\n", port.data.size() - 1, temp_id) );
 
 			bool has_transparency = false;
-			int bypass_part_width = port.data.size() / port.en.size();
 			for (int i = 0; i < GetSize(mem.wr_ports); i++) {
-				auto &wport = mem.wr_ports[i];
 				if (!port.transparency_mask[i] && !port.collision_x_mask[i])
 					continue;
 				has_transparency = true;
-				bypass_part_width = std::min(bypass_part_width, wport.data.size() / wport.en.size());
 			}
-			int bypass_valid_width = port.data.size() / bypass_part_width;
+			int bypass_valid_width = port.data.size();
+			int bypass_part_width = 1;
 
 			std::string bypass_valid_id;
 			std::string bypass_valid_reset_str;
@@ -864,7 +862,7 @@ void dump_memory(std::ostream &f, std::string indent, Mem &mem)
 
 						std::ostringstream os;
 						os << "assign ";
-						dump_sigspec(os, port.data.extract(seg_beg, seg_end - seg_beg));
+						dump_sigspec(os, port.data.extract(seg_beg, bypass_part_width));
 						os << " = ";
 						os << stringf("%s[%d]", bypass_valid_id, sub);
 						os << stringf(" ? %s[%d:%d]", bypass_data_id, seg_end - 1, seg_beg);
--- a/backends/verilog/verilog_backend.cc
+++ b/backends/verilog/verilog_backend.cc
@@ -631,8 +631,6 @@ void dump_memory(std::ostream &f, std::string indent, Mem &mem)
 					continue;
 				has_transparency = true;
 			}
-			int bypass_part_width = 1;
-			int bypass_valid_width = port.data.size() / bypass_part_width;
 
 			std::string bypass_valid_id;
 			std::string bypass_valid_reset_str;
@@ -640,11 +638,11 @@ void dump_memory(std::ostream &f, std::string indent, Mem &mem)
 			std::string bypass_data_reset_str;
 			if (has_transparency) {
 				bypass_valid_id = next_auto_id();
-				lof_reg_declarations.push_back( stringf("reg [%d:0] %s;\n", bypass_valid_width - 1, bypass_valid_id) );
+				lof_reg_declarations.push_back( stringf("reg [%d:0] %s;\n", port.data.size() - 1, bypass_valid_id) );
 
 				std::ostringstream os;
 				os << stringf("%s <= ", bypass_valid_id);
-				dump_sigspec(os, Const(State::S0, bypass_valid_width));
+				dump_sigspec(os, Const(State::S0, port.data.size()));
 				os << ";\n";
 				bypass_valid_reset_str = os.str();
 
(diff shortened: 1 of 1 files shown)


whitequark:
Thank you for working on this! Do you have before/after comparisons on hand?

qbojj:
I am yet to fully implement this patch, but here is current status
Single transparent read and write port of the same width and with read/write enables.
```diff
  reg [7:0] mem [63:0];
  // initial mem block
  always @(posedge clk) begin
    if (wr_en)
      mem[wr_addr] <= wr_data;
  end
  reg [7:0] _3_;
+ reg _4_;
+ reg [7:0] _5_;
  always @(posedge clk) begin
    if (rd_en) begin
      _3_ <= mem[rd_addr];
-     if (wr_en && rd_addr == wr_addr)
-       _3_ <= wr_data;
+     _4_ <= 1'h0;
+     _5_ <= 8'hxx;
+     if (wr_en && rd_addr == wr_addr) begin
+       _4_ <= 1'h1;
+       _5_ <= wr_data;
+     end
    end
  end
- assign rd_data = _3_;
+ assign rd_data = _4_ ? _5_ : _3_;
```

The `assign r_data = mem[latched r_addr]` pattern was fully removed, so for the same circuit as above excluding the read enable, we have:
```diff
  reg [7:0] mem [7:0];
  // initial mem block
  always @(posedge clk) begin
    if (wr_en)
      mem[wr_addr] <= wr_data;
  end
  reg [2:0] _3_;
+ reg _4_;
+ reg [7:0] _5_;
  always @(posedge clk) begin
-   _3_ <= rd_addr;
+   _3_ <= mem[rd_addr];
+   _4_ <= 1'h0;
+   _5_ <= 8'hxx;
+   if (wr_en && rd_addr == wr_addr) begin
+     _4_ <= 1'h1;
+     _5_ <= wr_data;
+   end
  end
- assign rd_data = mem[_3_];
+ assign rd_data = _4_ ? _5_ : _3_;
```

whitequark:
Looks good so far!

qbojj:
The diff turned out to be messier than I would have liked it to be, as I had removed the `assign r_data = mem[latched r_addr];` all together - the only case for which it could be used (other than the case when we want LUTRAMs) would be when there are no write ports, but that case will behave well with `r_data <= mem[r_addr]` pattern either way.

@awariac: could you look at this?

qbojj:
The state after fixing the quicks (actually making per-bit enables for write ports work correctly).

Single transparent read and write port of the same width and with read/write enables.
```diff
  reg [7:0] mem [63:0];
  // initial mem block
  always @(posedge clk) begin
    if (wr_en)
      mem[wr_addr] <= wr_data;
  end
  reg [7:0] _3_;
+ reg [7:0] _4_;
+ reg [7:0] _5_;
  always @(posedge clk) begin
    if (rd_en) begin
      _3_ <= mem[rd_addr];
-     if (wr_en && rd_addr == wr_addr)
-       _3_ <= wr_data;
+     _4_ <= 8'h00;
+     _5_ <= 8'hxx;
+     if (wr_en && rd_addr == wr_addr) begin
+       _4_ <= 8'hff;
+       _5_ <= wr_data;
+     end
    end
  end
- assign rd_data = _3_;
+ assign rd_data = (_4_ & _5_) | (~_4_ & _3_);
```

whitequark:
> ```diff
> +     _5_ <= 8'hxx;
> ```

I'm concerned about this bit. This isn't synthesizable; as far as I'm aware, in most cases (and certainly in the case of Yosys/nextpnr) the synthesizer will ignore this statement. Does Vivado or another toolchain require it?

qbojj:
> > ```diff
> > +     _5_ <= 8'hxx;
> > ```
> 
> I'm concerned about this bit. This isn't synthesizable; as far as I'm aware, in most cases (and certainly in the case of Yosys/nextpnr) the synthesizer will ignore this statement. Does Vivado or another toolchain require it?

There is a precedent for this - the `collision_x_mask` on the read port. It assigned `{width{1'x}}` on the read port for the data that aliases the ongoing write operation. This assignment also semantically removes the dependence on the value in previous cycle, but I hadn't yet tested if it is actually needed.

whitequark:
Gotcha, that's fine then.

hexratcc:
I've run some verification locally and didn't find anything wrong and in general the change looks good to me. I can't actually check with Vivado as I don't have access to it - @awariac can you confirm this fixes the issue?

awariac:
thanks everyone! I will check next week

qbojj:
I have confirmed that this pattern generates exactly the same resource use in quartus as pattern B from the issue, having least resource usage from all suggested patterns (width 16, depth 128, single write port and transparent read port). It looks like `_4_ <= 8'hxx;` assignment is not required for quartus.

backends/verilog/verilog_backend.cc
```diff
@@ -602,211 +602,262 @@ void dump_memory(std::ostream &f, std::string indent, Mem &mem)
 				}
 			}
 
-			// Decide how to represent the transparency; same idea as Mem::extract_rdff.
-			bool trans_use_addr = true;
-			for (auto bit : port.transparency_mask)
-				if (!bit)
-					trans_use_addr = false;
+			// Due to some synthesis tools being incompetent in detecting address colission pattern
+			// from verilog with strictly defined semantics and not infering bram (see github issue #5082),
```
hexratcc: Typo in "collision", "inferring".

hexratcc (changes requested):
Minor changes, I'll merge after these are fixed :).

whitequark:
@qbojj @awariac Would you like an update of `amaranth-yosys` once this is merged?

awariac:
@whitequark yes, that would be very helpful for packaging, thanks!

whitequark:
@awariac OK, just ping me once you've thoroughly tested it and I'll make it happen!

awariac:
in vivado pattern looks good and results in the same circuit as proposed in #5082. It looks like `_4_ <= 8'hxx;` is required for vivado, without it more LUTs are used.

hexratcc (approved):

PR #6056, opened 2026-07-20
CI: back to using official action

mmicko:


mmicko:
Something is not yet right there, still giving warning. Let's wait

PR #6062, opened 2026-07-22
Add dependencies to gtests

mmicko:
Required to be able to run unit tests only, but also makes sure all is built in time in other use cases
```
make clean
make test-unit -j9
```


whitequark (changes requested):
That's really janky. You should make a custom target `yosys-gtest-all` or something, add gtests to it as they're created, then use `DEPENDS $<$<TARGET:yosys-gtest-all>:yosys-gtest-all>` or something like that.

mmicko:
Thanks. Yeah that was ugly, that is why I was asking looking for suggestion :)

whitequark (approved):
LGTM!

PR #6064, opened 2026-07-24
Fix merging comparisons with oversized unsigned subtraction in `alumacc`

### LINKED ISSUE #6063: Signed comparison with nested ternary subtraction is mis-synthesized
### Version

Yosys 0.67+ (git sha1 30fe16c7f, Release, Clang /usr/bin/clang++ 18.1.3)

Linux

#### Yosys synthesis command:

```bash
yosys -p 'read_verilog -sv rtl.v; hierarchy -check -top top; proc; check -assert; synth; rename top top_synth; write_verilog synth.v'
```

#### Verilog before synthesis

Input RTL (`rtl.v`):

```verilog
module top(input signed a,b,input s,output [1:0] y);
assign y=a>b?s?1'b0:a-b:0;
endmodule
```

Important minimization note: `a` and `b` must remain 1-bit signed inputs, the output must remain 2 bits, and the inner `s ? 1'b0 : a-b` ternary must remain. Changing `y` to a scalar or replacing the inner ternary with direct `a-b` removes the mismatch.

#### Verilog after synthesis

```verilog
/* Generated by Yosys 0.67+ (git sha1 30fe16c7f, Release, Clang /usr/bin/clang++ 18.1.3) */

(* top =  1  *)
(* src = "rtl.v:1.1-3.10" *)
module top_synth(a, b, s, y);
  input signed a;
  wire signed a;
  input signed b;
  wire signed b;
  input s;
  wire s;
  output [1:0] y;
  wire [1:0] y;
  wire _0_;
  assign _0_ = a & ~(s);
  assign y[0] = _0_ & ~(b);
  assign y[1] = 1'h0;
endmodule
```

The relevant change is that the signed comparison and signed subtraction are replaced by:

```verilog
assign y = {1'b0, a & ~s & ~b};
```

For `s=0`, this becomes `{1'b0, a & ~b}`.  This expression is not equivalent to the RTL.

### Expected Behavior

The synthesized design must implement the RTL expression directly:

If the signed comparison `a > b` is true and `s=0`, `y` is the signed subtraction `a-b`, sign-extended to 2 bits.If `a > b` is false, or `s=1`, `y` is `2'b00`.

For example:

```text
a=0, b=1, s=0: 0 > -1, so a-b = 1 and y = 2'b11
a=1, b=0, s=0: -1 > 0 is false, so y = 2'b00
```

### Actual Behavior

```verilog
assign y = {1'b0, a & ~s & ~b};
```

This drops the signed comparison and signed subtraction behavior.  For`a=0, b=1, s=0`, the RTL interprets `b` as `-1`, takes the comparison branch, and sign-extends the result to `y=2'b11`; the synthesized Boolean expression instead gives `2'b00`.  For `a=1, b=0, s=0`, the signed comparison is false so the RTL gives `2'b00`, while the synthesized expression incorrectly gives `2'b01`.

KrystalDelusion:

Close #6063 - Folding an `$lt` (or `$gt`) into a `$sub` gives incorrect results when they comparison is signed and the arithmetic is unsigned, and the `$sub` is oversized (Y wider than A and B).

As pointed out by @Ravenslofty, the `of` and `sf` (overflow- and signed-flags) should be determined by the width of the result, not the width of the output.  An extra flag is added for `$macc` derived `$alu` models where the output is wider than the inputs, in which case the max width of the inputs (which determines the width of the result) is used instead.

I also noticed that when the `$sub` is oversized and signed, that the `sf` is equivalent to the `$lt` result (if a signed subtraction has a negative result then `b` must be smaller than `a`, and if the output is wider than the result then it can't underflow).

I'm not sure if it can ever come up naturally, but I did noticed that `alumacc` results don't maintain equivalence if the arithmetic output is smaller than one of the inputs - this is now checked for and skips merging if true.

And finally, when the signedness doesn't match between the merging candidates, if the input widths differ then the merging breaks the comparator results.  This can occur naturally, though only after `wreduce` - without the `wreduce` the inputs are sign extended such that the widths do match.  This is also resolved in this PR by skipping merging in this case.

One test (reduced from the issue) is included (and the `tests/opt/alumacc.ys` was moved to `tests/alumacc/cmp_merge.ys`).  The `generate_mk.py` file has been modified to generate and test a broad mix of (`$alu` compatible) cells and widths and signedness.  I did also notice that `alumacc` results don't maintain equivalence if the arithmetic output is smaller than one of the inputs - I'm not sure if this can ever come up naturally, but fixing it would be possible by adding checks that `GetSize(node->y) >= max(GetSize(node->a), GetSize(node->b))` in the `extract_cmp_alu()` function before assigning `n = node`, or just not adding the alu to the `sig_alu` pool since it would never be used.


Ravenslofty:
I think it would be useful to include that `generate_mk.py`, since the bug wasn't caught because it only got a narrow test case; any future improvements to `alumacc` would probably benefit from being 'fuzzed' like that.

tests/alumacc/cmp_merge.ys
```diff
@@ -31,3 +31,15 @@ proc
 equiv_opt -assert alumacc
 alumacc
 select -assert-count 1 t:$alu
+
```
Ravenslofty, line 32: this should go in `tests/alumacc` instead

passes/techmap/alumacc.cc
```diff
@@ -410,6 +427,24 @@ struct AlumaccWorker
...
 
+		auto cmp_mergeable = [](const alunode_t *node, const SigSpec cmp_A, const SigSpec cmp_B, bool is_signed) {
+			if (!node->invert_b || node->c != State::S1) {
+				// comparators need subtraction
+				return false;
+			}
+			if (max(GetSize(cmp_A), GetSize(cmp_B)) > GetSize(node->y)) {
+				// must have at least as many output bits
+				return false;
+			}
```
KrystalDelusion: May be more efficient to instead skip line 383 (`sig_alu[RTLIL::SigSig(A, B)].insert(alunode);`) if `max(GetSize(A), GetSize(B)) > GetSize(n->Y))`, since an undersized alu will never be a valid merge candidate.

tests/alumacc/generate_mk.py
```diff
@@ -5,4 +5,100 @@
...
+                out_cells += f" t:${c}"
+
+        undersized = c2 in ARITH and y_width < max(a_width, b_width)
+        mixed_sign = c1_signed != c2_signed and a_width != b_width
+        if c2 != "add" and not undersized and not mixed_sign:
+            # cells correctly merged
+            postopt_assert = f"select -assert-count 1 {out_cells}"
+        else:
+            # postopt_assert = f"select -assert-count 2 {out_cells}"
+            # while currently correct, the above assertion would block potential improvements
```
Ravenslofty: `-assert-max` would leave wiggle room for it to improve while catching it getting worse.

Ravenslofty (approved):
Up to you if you want to act on the comment or not; otherwise seems good to me.

passes/techmap/alumacc.cc
Ravenslofty: Perhaps a comment in the code to that effect; I think that would get lost in GitHub otherwise.

Ravenslofty (approved):

passes/techmap/alumacc.cc
KrystalDelusion: Ah, I missed that `sig_alu` is also how the alu models get converted to `$alu` cells so that's not actually viable.

KrystalDelusion:
Reverted the breaking change (moving the undersize check) and rebased on latest main.

PR #6065, opened 2026-07-24
ast: guard against module name collision during reprocessing

### LINKED ISSUE #6002: Assertion in hierarchy when elaborating parameterized generate instances due to non-unique name
### Version

Yosys 0.66+183 (git sha1 b285b72e3, GNU /usr/bin/c++ 13.3.0)

Linux

### Reproduction Steps

In directory with CanonicalNBitBinaryAdd.v and CanonialNBitBinaryMultiply.v run the the command

yosys -p "read_verilog -sv ./*.v; hierarchy -top CanonicalNBitBinaryMultiply -chparam AWIDTH 2 -chparam BWIDTH 2 -chparam RWIDTH 4 -chparam COUNTGATES 0 -purge_lib; stat"

[MultiplyError.zip](https://github.com/user-attachments/files/29712034/MultiplyError.zip)

### Expected Behavior

The design elaborates successfully. I have a divide that uses similar parameterized modules that does elaborate successfully. I used the command bellow to run the divide.

yosys -p "read_verilog -sv ./*.v; hierarchy -top CanonicalNBitBinaryDivide -chparam AWIDTH 2 -chparam BWIDTH 2 -chparam RWIDTH 4 -chparam COUNTGATES 0 -purge_lib; stat"

[DivideExpected.zip](https://github.com/user-attachments/files/29712074/DivideExpected.zip)

### Actual Behavior

I expect the multiply to elaborate but it fails with the error below every time. I have tried changing my run commands but no permutation has had an effect. 
ERROR: Assert `modules_.count(module->name) == 0' failed in /home/jojomon/Palace/KAOS/PbitOpt/NewYosys/yosys/kernel/rtlil.cc:1229.

greeshma-manjegowda:
process_module() could hit the `modules_.count(module->name) == 0` assertion in `Design::add()` when a module gets reprocessed via `AstModule::reprocess_if_necessary()`. This happens for a parameterized module that instantiates a submodule inside a generate block, where the submodule only becomes available after the enclosing module's initial derivation. In that case, `process_module()` could attempt to register a module under a name that was already occupied.
Guard the `design->add()` call in `process_module()` by removing any stale module already registered under the target name before adding the newly generated one.
Verified against the reproduction case in issue #6002 (multiply elaborates successfully without crashing) and confirmed the previously working divide case produces identical output before and after this change.
I was not able to run the full local test suite (`make test`) due to an unrelated jobserver/parallelism issue on my machine, but the two targeted regression cases above cover the reported bug directly.
Fixes #6002

### DIFF
--- a/frontends/ast/ast.cc
+++ b/frontends/ast/ast.cc
@@ -1325,6 +1325,18 @@ static RTLIL::Module *process_module(RTLIL::Design *design, AstNode *ast, bool d
 		log("--- END OF RTLIL DUMP ---\n");
 	}
 
+	// Guard against a module with this name already being registered --
+	// this can happen when reprocessing/re-deriving a module (e.g. via
+	// AstModule::reprocess_if_necessary()) causes the same parameterized
+	// module to be (re)generated more than once before the earlier
+	// in-flight copy has been added/removed. Without this, the
+	// "modules_.count(module->name) == 0" assertion in Design::add()
+	// below can fail.
+	if (RTLIL::Module *existing = design->module(module->name)) {
+		if (existing != current_module)
+			design->remove(existing);
+	}
+
 	design->add(current_module);
 	return current_module;
 }


frontends/ast/ast.cc
```diff
@@ -1325,6 +1325,18 @@ static RTLIL::Module *process_module(RTLIL::Design *design, AstNode *ast, bool d
 		log("--- END OF RTLIL DUMP ---\n");
 	}
 
+	// Guard against a module with this name already being registered --
+	// this can happen when reprocessing/re-deriving a module (e.g. via
+	// AstModule::reprocess_if_necessary()) causes the same parameterized
```
KrystalDelusion, line 1330:
```
#X# New IdString '\CanonicalNBitBinaryMultiply_before_process_and_replace_module_0' with index 1095.
#X# New Module: CanonicalNBitBinaryMultiply_before_process_and_replace_module_0
Generating RTLIL representation for module `$paramod$6b432225b58daca3f9d57abcb8ecd78e52ad9fe8\CanonicalNBitBinaryMultiply'.
```

`process_and_replace_module()` has logic to rename the old module to avoid such collisions; why is it renaming the `CanonicalNBitBinaryMultiply` module rather than the paramod?

greeshma-manjegowda: Yes, that's right. The module named CanonicalNBitBinaryMultiply at that point is a clone of the paramod module, not the paramod module itself. hierarchy.cc derives the paramod-named module first, then clones it and renames the clone to the plain top-level name, so top_mod ends up pointing at the clone rather than the original. process_and_replace_module() is renaming the object it was actually given, which is correct - it's just that the clone no longer shares a name with the paramod module its own AST still targets, which is why the rename here doesn't touch the name that ends up colliding later.

KrystalDelusion: Is it possible to rename the paramod (and set the `to_delete` attribute on it) like we do with non-parametrized modules instead? (I imagine there is a reason we do that instead of deleting them at the time, so we should probably do the same here)

PR #6069, opened 2026-07-27
read_liberty: fix set&reset polarity

widlarizer:
The `read_liberty` code added in #5655 was causing incorrect polarity handling. The formal Verilog test for `$dffsr` mapping didn't catch it because I didn't run `techmap`, and `dfflibmap` ignores coarse grained cells. This PR
- adds checks that `dfflibmap` maps all cells in said test
- adds the `techmap` command to it, fulfilling those checks, but exposing the equivalence error
- removes the weird "optimization" from `read_liberty` that makes comprehension tricker at minimal cost (if it's to save on solver queries, run `opt_expr` instead...) and fixes the logic error discovered by the test
- adds an error to dfflibmap when wide registers are selected since they can't be mapped (this alone would have also discovered the test issue, you can verify by cherrypicking it on top of main and running `dfflibmap_proc_formal.ys`)


jeras:
I was not sure about what the `liberty` change was supposed to do, so I checked.
As I understand, a cell is now always translated to a basic `$_DFF` with all `P` ports instead of detecting inverters and using `$_DFF` cells with `N` ports. I agree this makes sense, the previous approach left the unused inverters dangling (were not cleaned).
Could you go a bit further and remove the code related to polarity variables (`bool clk_polarity = true, clear_polarity = true, preset_polarity = true;`)? It looks out of place now.

passes/techmap/dfflibmap.cc
```diff
@@ -503,6 +504,11 @@ static void dfflibmap(RTLIL::Design *design, RTLIL::Module *module)
 
 	std::vector<RTLIL::Cell*> cell_list;
 	for (auto cell : module->cells()) {
+		auto cats = StaticCellTypes::categories;
+		if (cats.is_ff(cell->type) && !cats.is_stdcell(cell->type))
+		log_error("Wide register cell type %s is not supported.\n"
```
hexratcc: `log_error` indent.

frontends/liberty/liberty.cc
```diff
@@ -273,21 +249,21 @@ static void create_ff(RTLIL::Module *module, const LibertyAst *node)
 		cell->setPort(ID::C, clk_sig);
 
 		if (clear_sig.size() == 0 && preset_sig.size() == 0) {
-			cell->type = stringf("$_DFF_%c_", clk_polarity ? 'P' : 'N');
+			cell->type = stringf("$_DFF_P_");
```
hexratcc: I think `ID($_DFF_P_)` would be cleaner (same for the other 4 instances below).

tests/techmap/dfflibmap_wide.ys
```diff
@@ -0,0 +1,5 @@
+read_verilog ../sim/dff.v
+proc
+logger -expect error "not supported" 1
+dfflibmap -liberty dfflibmap.lib
+logger -check-expected
```
hexratcc: Missing a trailing newline.

hexratcc (approved):
Just minor nit-picks.

PR #6073, opened 2026-07-31
opt_modadd_tree: match on the state recurrence, not addend provenance

akashlevy:
## Summary

A mod-M digit-sum cascade whose digits are sliced out of a barrel shifter was not rewritten at all. Two checks rejected it, and both keyed on where the addend bits came from rather than on the recurrence the pass owns:

- `is_cut` rejects a cut when a cell upstream of it feeds a downstream cell other than through the cut. The shared shifter sits upstream of every cut and feeds every step, so no chain was ever built.
- `region_digit` only accepted *undriven* cone leaves as a step's digit, because forcing a driven bit could collide with ConstEval's cell-output cache. With a shifter in place every digit bit is driven.

Cells that neither produce nor can reach a candidate state bus are now classified as addend logic and may fan out freely; driven digit bits are allowed once a cone walk proves the step closes on state+digit with no forced bit produced inside it, so soundness still comes from a proof rather than a looser fingerprint. Provenance no longer matters: a shifter, mux tree, register or arbitrary cell output all behave alike.

Widening the match exposed two ways the rewrite could lose, both fixed here:

- **Escaping step reads.** A step cell read from outside the cone survives the rewrite and keeps the serial chain alive next to its new tree. Node 0 now grows until it covers every such cell. This also fixes a pre-existing false economy: on ariane the old pass rewrote a 30-node cascade in the PLIC `SEQUENTIAL` arbiter, growing that module from 180 to 1079 cells while depth moved only 92 → 90, because `irq_id` still reads every intermediate `max_prio`.
- **Depth estimated by cell count.** That only tracks real depth where a step carries a ripple, and it counted node 0 and the addend logic, which are paid either way. The gain test now compares the state path alone and requires at least one ripple cell in the step. Without this the widened match rebalanced a purely boolean thermometer accumulate and made it worse.

## Results

| | before | after |
|---|---|---|
| shifter-fed mod-7 cascade | lol 40.5 / clk 605.7 | **22.25 / 269.5** |
| pass runtime, veer (29,118 cells) | 25.14s | **23.13s** |
| full flow, ariane | 726.58s | **658.76s** |

The ariane and veer speedups come from no longer emitting the PLIC rewrite that downstream ABC and STA had to process. Both designs now emit zero cells from this pass.

## Test plan

- [x] New regression: shifter-fed mod-7 cascade, asserts the rewrite substring and lol < 27
- [x] New negative: second-order cascade (each step also folds the accumulator from two steps back) must not rewrite, since rebalancing one accumulator would leave the serial chain alive to feed the other
- [x] `qor_mod7_digit_sum` 16.75, `qor_mod9_digit_sum` 34.5 (still the cheap associative tier), `qor_mod7_skew_nomatch` 15.5 — all unchanged
- [x] `qor_mod61_wide_nomatch` still refused by the state-width cost guard
- [x] `qor_vmw_slot_lane` 16.0, guarded against the new false match
- [x] All 83 `qor_*` regressions pass; full suite 299 passed / 8 xfailed
- [ ] `customer_veer_eh2` not re-run end to end on the final build; verified instead that the pass reports zero rewrites and zero emitted cells in both of veer's synthesis invocations, so the netlist handed downstream is identical to baseline

## Known limitation

The ripple requirement is a cost-model heuristic, not a proof: a genuine narrow-state recurrence spelled entirely in bitwise gates is now refused. That is deliberate — ABC restructures those anyway, and both boolean cascades measured were net losses — but it is a judgment call.


akashlevy:
Opened against the wrong repository by mistake; this belongs on the Silimate fork. Closing.

PR #6075, opened 2026-08-02
feat: add support for UDPs

rhanqtl:
**Motivation**

Provide support for user-defined primitives (UDP). See IEEE 1800-2023, Chapter 29.

**Changes**

- Added UDP lexer and parser in the verilog frontend
- Added a UDP importer from frontends to RTLIL `kernel/udp.{h,cc}`
- Added e2e tests

### DIFF
--- a/frontends/ast/ast.cc
+++ b/frontends/ast/ast.cc
@@ -28,6 +28,7 @@
  */
 
 #include "kernel/yosys.h"
+#include "kernel/udp.h"
 #include "libs/sha1/sha1.h"
 #include "ast.h"
 
@@ -1158,7 +1159,7 @@ static RTLIL::Module *process_module(RTLIL::Design *design, AstNode *ast, bool d
 
 		bool blackbox_module = flag_lib;
 
-		if (!blackbox_module && !flag_noblackbox) {
+		if (!blackbox_module && !flag_noblackbox && !ast->get_bool_attribute(ID(udp))) {
 			blackbox_module = true;
 			for (const auto& child : ast->children) {
 				if (child->type == AST_WIRE && (child->is_input || child->is_output))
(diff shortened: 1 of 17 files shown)


whitequark:
How did you implement this feature?

rhanqtl:
How should r/f be represented? For example, in below case:

```verilog
primitive dff_udp (q, clk, d);
  input clk, d;
  output reg q;

  table
// clk  d   q   q+
    r   0 : ? : 0 ;
    r   1 : ? : 1 ;
    f   ? : ? : - ;
    ?   * : ? : - ;
  endtable
endprimitive
```

Can I create a `SyncRule(type=STp, signal=clk, action=(q, next_q))`? Or, should I preserve the 0->1 transition precisely, by a more elaborated check like:

```
process xxx
  switch { clk_prev, clk }
    case 2'b01
       assign next_q d
```

If so, can Yosys infer DFFs as expected?

rhanqtl:
@whitequark @widlarizer As decribed in the summary, I added UDP lexer and parser, and a `UDPImporter` for bridging frontend IRs to RTLIL. Currently only combinational logics are supported in the importer, by constructing a SwitchRule according to UDP table entries. For sequential logics, I'm seeking some advice on representation of r/f, could you give me some advice?

widlarizer:
[This comment from related prior work still applies](https://github.com/YosysHQ/yosys/pull/5938#issuecomment-4648112112). We're working on making the slang integration in Yosys usable and productive and adding features to read_verilog, a tough to maintain and review, yet critical to current yosys requiring deep understanding in review, doesn't make sense for us at this point. So I realistically won't be merging this PR based on that simple principle. There's serious changes coming up that should resolve long-standing issues, changes that scale beyond the accumulation of individual features, and we would like to focus on that.

It would probably be helpful if you shared your specific use case at https://github.com/povik/sv-elab/issues/184

Furthermore, some feedback for your sake: the description of your implementation approach isn't very helpful. A description of a PR should go a lot deeper than that, since as-is, this requires reviewers to go and directly read the code to extract the ideas based on which the code was written. It is far more efficient to write out those ideas in the description! As I'm sure you've seen in the [contributing guidelines](https://yosyshq.readthedocs.io/projects/yosys/en/latest/yosys_internals/extending_yosys/contributing.html#contributing-code), we prioritize what we review based on descriptions first.

As I look at the code at speed, I see several issues

- the data modeling of UDP tables is entirely unexplained - answers to questions like "what does a UdpTableEntry exactly represent?" can once again only be answered by reading code that handle them, not by looking at the definition of struct UdpTableEntry, either just the naming or a comment that explains it
- code to convert a UDP model to RTLIL lives in kernel/ despite being specific to resolution into RTLIL in a single frontend rather than shared infrastructure to use across Yosys
- you're introducing undocumented attributes
- I'm not sure that frontends other than read_verilog could use that model, their integrations differ a lot

I don't have guidance for you to implement sequential UDPs since I don't understand UDPs well enough and would be kind of just guessing.

Overall I understand that this code might work to implement combinatorial UDPs for read_verilog but

1. that's not something we want
2. this approach to development isn't likely to lead to code that can be maintained in an open source project by anybody but its author

rhanqtl:
@widlarizer Got it. I’ll take a look at the projects you mentioned. I also realize that I missed some important context.

Regarding the coding style, I consider this PR a draft. My main goal was to ask for help and present a working implementation, so I haven’t polished the code yet.

PR #6076, opened 2026-08-03
driver: disable exec unless --enable-exec

widlarizer:
Security feature so that system commands can't be simply run by supplying an evil techmap file. Also disables undocumented `!` feature of `techmap`. Motivated by use cases of Yosys being run on a server exposed over the internet to anonymous users. Help still tells you about exec but the help message mentions it's not enabled unless the CLI argument is passed and you get an error telling yo about the CLI argument if you do try to call it

### DIFF
--- a/kernel/driver.cc
+++ b/kernel/driver.cc
@@ -165,6 +165,7 @@ int main(int argc, char **argv)
 			cxxopts::value<std::vector<std::string>>(), "<plugin>")
 		("D,define", "set the specified Verilog define to <value> if supplied via command \"read -define\"",
 			cxxopts::value<std::vector<std::string>>(), "<define>[=<value>]")
+		("enable-exec", "enable the 'exec' command, which runs arbitrary shell commands")
 		("S,synth", "shortcut for calling the \"synth\" command, a default script for transforming " \
 					"the Verilog input to a gate-level netlist. For example: " \
 					"yosys -o output.blif -S input.v " \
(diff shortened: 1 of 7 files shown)


KrystalDelusion:
> Also disables undocumented `!` feature of `techmap`.

Oof. What about script wires?

nakengelhardt:
This is by far not enough, if you want to actually introduce a security boundary you'll need to do a full rethink of the yosys architecture. What do you do about tcl? Custom abc scripts? Custom arguments for any other subprocess that some yosys command calls? I'm against adding an argument that would just give a false sense of security. If you want to actually prevent execution there's `YOSYS_ENABLE_SPAWN` (which is disabled by the WASM builds IIRC?).

Websites that run yosys on untrusted input should do whatever godbolt.org does to make it ok to let you run arbitrary C++ on their server. I haven't looked at what they do but it's clearly a solved problem.

widlarizer:
Sorry I said that wrong, `!` isn't a techmap feature. It's a kind of a command that gives you shell access as well. But somehow... different tty detection behavior? See:

```
yosys> ! ls /

4. Shell command:  ls /
bin   dev  home  lib64	     mnt  opt	root  srv  tmp	var
boot  etc  lib	 lost+found  nix  proc	run   sys  usr

yosys> exec -- ls /

5. Executing command "ls /".
bin
boot
dev
etc
...
```

` wire [1023:0] _TECHMAP_DO_` (that's what you were referring with script wires, right?) gives you access to any yosys commands, so I'm hiding the sketchy commands in this PR. I'm not actually doing anything to `techmap`

whitequark:
> Websites that run yosys on untrusted input should do whatever godbolt.org does to make it ok to let you run arbitrary C++ on their server. I haven't looked at what they do but it's clearly a solved problem.

Simply use yowasp-yosys. It is designed to provide a robust sandbox if you adjust the mountpoints. This is what ChipFlow did.

PR #6078, opened 2026-08-03
hierarchy: fix keep cache usage by caching false too

### LINKED ISSUE #1689: Recursive module instantations cause stack overflow
MCVE:
```verilog
module top(input x, output y);
top top_i(.x(x), .y(y));
endmodule
```
```
yosys -p hierarchy recurse.v
```

Produces
```
-- Parsing `recurse.v' using frontend `verilog' --

1. Executing Verilog-2005 frontend: recurse.v
Parsing Verilog input from `recurse.v' to AST representation.

-- Running command `hierarchy' --

2. Executing HIERARCHY pass (managing design hierarchy).
[1]    5707 segmentation fault (core dumped)  yosys -p hierarchy recurse.v
```

Backtrace:
```
0x00005555556ec6e1 in Yosys::RTLIL::Design::module (this=<optimized out>, name=...) at kernel/rtlil.cc:423
423		return modules_.count(name) ? modules_.at(name) : NULL;
(gdb) bt
#0  0x00005555556ec6e1 in Yosys::RTLIL::Design::module (this=<optimized out>, name=...) at kernel/rtlil.cc:423
#1  0x0000555555c643d0 in (anonymous namespace)::set_keep_assert (cache=std::map with 0 elements, mod=<optimized out>)
    at passes/hierarchy/hierarchy.cc:508
#2  0x0000555555c643eb in (anonymous namespace)::set_keep_assert (cache=std::map with 0 elements, mod=<optimized out>)
    at passes/hierarchy/hierarchy.cc:509
#3  0x0000555555c643eb in (anonymous namespace)::set_keep_assert (cache=std::map with 0 elements, mod=<optimized out>)
    at passes/hierarchy/hierarchy.cc:509
#4  0x0000555555c643eb in (anonymous namespace)::set_keep_assert (cache=std::map with 0 elements, mod=<optimized out>)
    at passes/hierarchy/hierarchy.cc:509
#5  0x0000555555c643eb in (anonymous namespace)::set_keep_assert (cache=std::map with 0 elements, mod=<optimized out>)
    at passes/hierarchy/hierarchy.cc:509
#6  0x0000555555c643eb in (anonymous namespace)::set_keep_assert (cache=std::map with 0 elements, mod=<optimized out>)
    at passes/hierarchy/hierarchy.cc:509
#7  0x0000555555c643eb in (anonymous namespace)::set_keep_assert (cache=std::map with 0 elements, mod=<optimized out>)
    at passes/hierarchy/hierarchy.cc:509
#8  0x0000555555c643eb in (anonymous namespace)::set_keep_assert (cache=std::map with 0 elements, mod=<optimized out>)
    at passes/hierarchy/hierarchy.cc:509
#9  0x0000555555c643eb in (anonymous namespace)::set_keep_assert (cache=std::map with 0 elements, mod=<optimized out>)
    at passes/hierarchy/hierarchy.cc:509
```
(and so on)

Reduced from a testcase from @cr1901 which involved a missing `-lib`, it would be nice to error out here more meaningfully.

widlarizer:
Fixes #1689. Previously, only the fact that a module is known to be "keep" was cached. With this change, it is also cached that a module is known to NOT be "keep"

This is ~~vibe coded with~~ _code generated by_ opus 5 ($1.48 if I was at API pricing) and the prompt "Fix ./y -p "read_verilog ../yosys/g/recurse.v; hierarchy -auto-top" segfault. Rebuild with cmake --workflow --preset d" with g/recurse.v the issue repro.

Correction: I was reminded that "vibe coding" refers to "not even checking the code". Here I checked the code, liked it well enough for the scope of the bug, and commited the code *without modifying it*. This differs from any of my other LLM usage, where I rewrite any LLM generated code entirely based on my own knowledge and taste, so that the LLM provides only a proof-of-concept based on a combination either my ideas or simple debugging/profiing jobs, and the code isn't pushed without a near total rewrite

### DIFF
--- a/passes/hierarchy/hierarchy.cc
+++ b/passes/hierarchy/hierarchy.cc
@@ -684,24 +684,30 @@ void hierarchy_clean(RTLIL::Design *design, RTLIL::Module *top, bool purge_lib)
 
 bool set_keep_print(std::map<RTLIL::Module*, bool> &cache, RTLIL::Module *mod)
 {
-	if (cache.count(mod) == 0)
-		for (auto c : mod->cells()) {
-			RTLIL::Module *m = mod->design->module(c->type);
-			if ((m != nullptr && set_keep_print(cache, m)) || c->type == ID($print))
-				return cache[mod] = true;
-		}
-	return cache[mod];
+	auto it = cache.find(mod);
+	if (it != cache.end())
+		return it->second;
+	cache[mod] = false;
+	for (auto c : mod->cells()) {
+		RTLIL::Module *m = mod->design->module(c->type);
+		if ((m != nullptr && set_keep_print(cache, m)) || c->type == ID($print))
+			return cache[mod] = true;
+	}
+	return false;
 }
 
 bool set_keep_assert(std::map<RTLIL::Module*, bool> &cache, RTLIL::Module *mod)
 {
-	if (cache.count(mod) == 0)
-		for (auto c : mod->cells()) {
-			RTLIL::Module *m = mod->design->module(c->type);
-			if ((m != nullptr && set_keep_assert(cache, m)) || c->type.in(ID($check), ID($assert), ID($assume), ID($live), ID($fair), ID($cover)))
-				return cache[mod] = true;
-		}
-	return cache[mod];
+	auto it = cache.find(mod);
+	if (it != cache.end())
+		return it->second;
+	cache[mod] = false;
+	for (auto c : mod->cells()) {
+		RTLIL::Module *m = mod->design->module(c->type);
+		if ((m != nullptr && set_keep_assert(cache, m)) || c->type.in(ID($check), ID($assert), ID($assume), ID($live), ID($fair), ID($cover)))
+			return cache[mod] = true;
+	}
+	return false;
 }
 
 int find_top_mod_score(Design *design, Module *module, dict<Module*, int> &db)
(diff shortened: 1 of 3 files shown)


hexratcc (approved):
I think it would be good to also add a regression test for this, other than that, lgtm

passes/hierarchy/hierarchy.cc
```diff
@@ -684,24 +684,30 @@ void hierarchy_clean(RTLIL::Design *design, RTLIL::Module *top, bool purge_lib)
...
-		}
-	return cache[mod];
+	auto it = cache.find(mod);
+	if (it != cache.end())
+		return it->second;
+	cache[mod] = false;
+	for (auto c : mod->cells()) {
+		RTLIL::Module *m = mod->design->module(c->type);
+		if ((m != nullptr && set_keep_print(cache, m)) || c->type == ID($print))
+			return cache[mod] = true;
```
Ravenslofty, line 694: Sorry, what? I misread this as an `==`, but no, this is an assignment used as an expression for `true`. This should perhaps be split up into statements for clarity.

widlarizer: Yeah, this was present already. But we have way, way bigger readability problems in hierarchy than this C-brained leftover stuff and I'll get back to redesigning that pass soonish #5821.

PR #6079, opened 2026-08-04
Make the Silimate matcher passes stop re-walking cones

akashlevy:
## Summary

The Silimate matcher passes had become the dominant cost of synthesis on large designs: the full chain took 170s on nvdla and 94s on veer. All of the blowup came from the same shape — work that depends only on a root's cone was being redone once per candidate the matcher tried against that root.

Two commits, each removing one instance of that pattern:

- **`opt_modadd_tree`, `opt_vps`**: `find_chain` re-walked the cone and re-ran `is_cut` for every candidate, and `opt_vps` re-scanned every cell per decoder. Replaced with a bitset cone index (O(1) ancestor/escape queries, fixed-point propagation so cyclic cones still index) and an O(1) decoder-bit map with `$pmux` bucketing.
- **`cut_region.h`, `opt_compact_prefix`, `opt_priority_onehot`**: each candidate cut re-ran a hashed `SigBit` BFS that re-`sigmap`ped every connection. Now a root's cone is interned once into dense integer vectors and cuts run as array BFS over it. Also memoizes `sigmap` of the repeatedly re-mapped candidate buses, indexes wires by bit so the bus collectors stop rescanning every module wire per root, and rejects cuts whose cone has more leaves than the cut has bits before walking anything.

No matcher heuristic, budget, or fingerprint was loosened — this is purely a change in how the same search is executed.

## Measurements

Per-pass, versus the pre-change binary:

| Pass | veer | nvdla |
|---|---|---|
| `opt_compact_prefix` | 28.1s → 5.5s | 43.5s → 13.4s |
| `opt_priority_onehot` | 21.8s → 5.4s | 61.8s → 23.2s |
| `opt_modadd_tree` | 23.3s → 1.5s | 55.7s → 4.5s |
| `opt_argmax` | 5.1s → 3.3s | — |

`opt_vps` went from quadratic to linear in decoder count on a synthetic ladder (7.9s → 0.29s at 256 lanes).

Full matcher chain: **veer 93.5s → 22.6s**, **nvdla 170.3s → 53.5s**, sargantana 37.5s → 13.1s.

## Test plan

- [x] Chain netlist byte-identical to the old binary on veer, veer_flat, sargantana, nvdla (after normalizing the version banner and the source line numbers embedded in autogenerated cell names)
- [x] Full chain log text identical on all four designs
- [x] Per-pass debug logs identical line-for-line — 1.77M lines for `opt_compact_prefix` and 1.63M for `opt_priority_onehot` on veer — apart from the coarser failure reason the new leaf-count precheck reports, confirming every candidate and cut verdict is unchanged. Same for `opt_argmax`, `opt_first_fit_alloc`, `opt_maxcmp`, `opt_xor_fold`.
- [x] All ten matcher pass unit tests pass
- [x] `gtest-opt` and `gtest-kernel` pass


akashlevy:
Opened against the wrong remote by mistake; this targets a fork's internal branch. Closing.

PR #6084, opened 2026-08-05
`tribuf`: infer tri-state nested behind a mux

### LINKED ISSUE #1595: Tribuf hierarchy on bidirectional IO buffer insertion

When I've compiled i4004 block from http://www.opencores.org/cores/mcs-4/
I've encountered a problem with IO buffer insertion for inout port.
The problem is, that z-state for this port is hidden in the second layer of conditional operator.
> assign data_pad = poc ? 4'b0000 : (n0700 ? 4'bzzzz : data_out);

In this case tri-buf isn't on the same level, as the assignment, so IO insertion logic failed.

To test this assumption I've used 32-bit mxe-build Yosys of commit dadaf7ed in Windows 7 OS.
The first design was:

```
module bidirtest (A, COND, DATA, EO, Q, BI);
inout BI;
output Q;
input A, COND, DATA, EO;

assign BI = COND ? 1'b0 : (A ? 1'bz : DATA);
assign Q = BI;

endmodule 
```

The synthesis script was:

```
read_verilog bidirtest.v
tribuf -logic
synth -top bidirtest
iopadmap -bits -inpad IBUF O:PAD -outpad OBUF I:PAD -tinoutpad IOBUF ENA:O:I:PAD bidirtest
write_verilog -noattr proj_iotemp.v
```
As expected, Yosys didn't detect z-state buffer, It didn't even isert any IO cell for the BI port:

```
/* Generated by Yosys 0.9+932 (git sha1 dadaf7ed, i686-w64-mingw32.static-g++ 5.5.0 -Os) */

module bidirtest(A, COND, DATA, EO, Q, BI);
  wire _00_;
  wire _01_;
  wire _02_;
  wire _03_;
  wire _04_;
  input A;
  inout BI;
  input COND;
  input DATA;
  input EO;
  output Q;
  assign BI = _02_ & ~(_03_);
  IBUF _06_ (
    .O(_04_),
    .PAD(A)
  );
  IBUF _07_ (
    .O(_03_),
    .PAD(COND)
  );
  IBUF _08_ (
    .O(_02_),
    .PAD(DATA)
  );
  IBUF _09_ (
    .O(_01_),
    .PAD(EO)
  );
  OBUF _10_ (
    .I(_00_),
    .PAD(Q)
  );
  assign _00_ = BI;
endmodule
```

## Expected behavior

When the same expression is changed to have z-state on the upper-most level, then everything will work well.
RTL Verilog:

```
module bidirtest (A, COND, DATA, EO, Q, BI);
inout BI;
output Q;
input A, COND, DATA, EO;

assign BI = !COND && A ? 1'bz : (COND ? 1'b0 : DATA);
assign Q = BI;

endmodule 
```

The output netlist:

```
/* Generated by Yosys 0.9+932 (git sha1 dadaf7ed, i686-w64-mingw32.static-g++ 5.5.0 -Os) */

module bidirtest(A, COND, DATA, EO, Q, BI);
  wire _00_;
  wire _01_;
  wire _02_;
  wire _03_;
  wire _04_;
  wire _05_;
  wire _06_;
  wire _07_;
  input A;
  inout BI;
  input COND;
  input DATA;
  input EO;
  output Q;
  assign _01_ = _04_ & ~(_05_);
 assign _00_ = _05_ | ~(_06_);
  IOBUF _10_ (
    .ENA(_00_),
    .I(_01_),
    .O(_07_),
    .PAD(BI)
  );
  IBUF _11_ (
    .O(_06_),
    .PAD(A)
  );
  IBUF _12_ (
    .O(_05_),
    .PAD(COND)
  );
  IBUF _13_ (
    .O(_04_),
    .PAD(DATA)
  );
  IBUF _14_ (
    .O(_03_),
    .PAD(EO)
  );
  OBUF _15_ (
    .I(_02_),
    .PAD(Q)
  );
  assign _02_ = _07_;
endmodule
```

## Actual behavior

It didn't insert IO cell and didn't detect tribuf.

hexratcc:
This PR aims to fix #1595 - z hidden in a nested conditional (`cond ? val : (en ? z : x)`) sits below the assignment level, so `tribuf` never inferred it and `iopadmap` skipped the `IOBUF` on the inout. `tribuf` now folds z up through `$mux`/`$_MUX_` and through the data input of another `tribuf`, to a fixpoint, before merge/-logic.

### DIFF
--- a/passes/techmap/tribuf.cc
+++ b/passes/techmap/tribuf.cc
@@ -52,9 +52,116 @@ struct TribufWorker {
 		return true;
 	}
 
+	static IdString tribuf_en_id(Cell *cell)
+	{
+		return cell->type == ID($tribuf) ? ID::EN : ID::E;
+	}
+
+	static SigSpec tribuf_en(Cell *cell)
+	{
+		return cell->getPort(tribuf_en_id(cell));
+	}
+
+	Cell *inner_tribuf(SigSpec sig, dict<SigBit, Cell*> &tribuf_driver, dict<SigBit, int> &fanout)
+	{
+		Cell *c = nullptr;
+		for (auto bit : sigmap(sig)) {
+			auto it = tribuf_driver.find(bit);
+			if (it == tribuf_driver.end() || fanout[bit] != 1)
+				return nullptr;
+			if (c && c != it->second)
+				return nullptr;
+			c = it->second;
+		}
+		if (c && sigmap(c->getPort(ID::Y)) == sigmap(sig))
+			return c;
+		return nullptr;
+	}
+
+	// Fold tri-state that is hidden behind an ordinary $mux (or behind the data input of another tribuf)
+	// into a single tribuf, so nested z like cond ? val : (en ? z : x) is not lost before iopadmap can see it
+	bool fold_tristate()
+	{
+		dict<SigBit, Cell*> tribuf_driver;
+		dict<SigBit, int> fanout;
+
+		for (auto cell : module->cells()) {
+			for (auto &conn : cell->connections())
+				if (!cell->output(conn.first))
+					for (auto bit : sigmap(conn.second))
+						fanout[bit]++;
+			if (cell->type.in(ID($tribuf), ID($_TBUF_)))
+				for (auto bit : sigmap(cell->getPort(ID::Y)))
+					tribuf_driver[bit] = cell;
+		}
+		// Ports keep their nets alive, so never fold them away
+		for (auto wire : module->wires())
+			if (wire->port_id)
+				for (auto bit : sigmap(wire))
+					fanout[bit]++;
+
+		struct MuxJob { Cell *mux, *ta, *tb; };
+		struct TriJob { Cell *outer, *inner; };
+		std::vector<MuxJob> mux_jobs;
+		std::vector<TriJob> tri_jobs;
+
+		for (auto cell : module->selected_cells()) {
+			if (cell->type.in(ID($mux), ID($_MUX_))) {
+				Cell *ta = inner_tribuf(cell->getPort(ID::A), tribuf_driver, fanout);
+				Cell *tb = inner_tribuf(cell->getPort(ID::B), tribuf_driver, fanout);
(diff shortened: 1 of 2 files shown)


hexratcc:
[sc-714]

widlarizer (changes requested):
At a quick glance, I think the newly added folding happens regardless of mode? This doesn't seem right. In an ASIC context, unconditionally restructuring tribufs driving a mux would be unexpected behaviour of this pass. `tribuf -logic` keeps tribufs that drive output pins, which makes sense for FPGA where it's valid to want to infer tristate IO pads, but it also silently eliminates other tribufs. Some FPGA scenarios:

1. the user has tristate / z-bits because of writing poor verilog or misunderstanding the IP cores they're using in their design. To prevent this, default behavior should be that the tribuf pass, the way it is used in `synth_*`, provides a convenient error
2. the user has tristate / z-bits on specifically the output pins of the top module, because they are synthesizing an entire design with Yosys, and want that behavior. The pass should enable this without even emitting a warning, because it's a reasonable thing to want, and it probably shouldn't even need a special flag. I think this is the case with the originally reported issue here
3. the user has tristate / z-bits accidentally on the top level output pins, but they're not synthesizing a whole design, they're synthesizing an inner "macro" they'll be instantiating manually in the actual design that gets passed to the vendor-provided physical design automation flow. They may want to opt into convenient errors that reflect the fact that the yosys synthesis top module isn't the final design top module
4. the user violently wants to synthesize something that has z-bits in the middle of everything and doesn't mind any discrepancies at all

Most problematically we have a distinct lack of tests on this pass on its own, so I don't really know what the point of -logic is without REing the sources. Extend / redesign the interface and behavior of this pass so that the scenarios, particularly 1 and 2, are fulfilled, while the ASIC (probably just the current default `tribuf` behavior) and formal scenarios (`tribuf -formal`) stay unaffected. I think something like `tribuf -iopad` that converts z-bit driven muxes driving top outputs to tribuf, and errors out on anything else, is the important part, but also maybe a `tribuf -none` would be convenient too?

PR #6086, opened 2026-08-05
Make `proc_self_dirname()` work for non-ASCII paths to the binary

whitequark:

Yosys crashes if launched from a directory containing Unicode characters on Windows, e.g. `C:\Program Files\Glasgow 可重构数字接口调试器`.

Since the installation directory is user-configurable this has to be fixed in Yosys itself.

This is achieved by using wide character functions to get the actual path (there is no workaround here that will do the job) and then changing the current directory so that no other part of Yosys has to care about Unicode.

This works because Yosys doesn't try to use absolute paths and does not change the current directory anywhere else.

### DIFF
--- a/kernel/yosys.cc
+++ b/kernel/yosys.cc
@@ -537,18 +537,16 @@ std::string proc_self_dirname()
 #elif defined(_WIN32)
 std::string proc_self_dirname()
 {
-	int i = 0;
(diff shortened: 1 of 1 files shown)


whitequark:
Nevermind, realized that this cannot work.

PR #6087, opened 2026-08-05
Accept UTF-8 filenames and terminal I/O on Windows.

whitequark:

Yosys on Windows fails to work if the installation directory or user name (the temporary directory is typically under `%LocalAppData%`) contain Unicode characters.

Switch the CRT to use UTF-8 for filenames and console I/O.

### DIFF
--- a/kernel/driver.cc
+++ b/kernel/driver.cc
@@ -48,6 +48,7 @@ namespace py = pybind11;
 #include <string.h>
 #include <limits.h>
 #include <errno.h>
+#include <fcntl.h>
 #ifndef __STDC_FORMAT_MACROS
 #  define __STDC_FORMAT_MACROS
 #endif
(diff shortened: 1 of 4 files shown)


whitequark:
Had to back out `_setmode(_fileno(stdin|stdout|stderr), _O_U8TEXT);` because it breaks readline. I *think* this only affects wide stdio functions so should not break anything, but haven't tested yet.

whitequark:
This has been tested now, everything works including console I/O and running from a directory with Unicode name:

image

whitequark:
Passing CLI arguments as well: 
image

kernel/driver.cc
```diff
@@ -112,8 +113,29 @@ namespace Yosys {
...
 
+#ifdef _WIN32
+int wmain(int argc, wchar_t **wargv)
+{
+	// Configure ASCII functions like fopen() to use UTF-8 filenames.
+	// See https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/setlocale-wsetlocale?view=msvc-170#utf-8-support
+	setlocale(LC_ALL, ".UTF-8");
+	// Configure the Win32 console to use UTF-8.
+	SetConsoleCP(CP_UTF8);
+	SetConsoleOutputCP(CP_UTF8);
```
mmicko, line 126: can we do next since it can still fail on MSVCRT ?

```
	// Configure the Win32 console to use UTF-8.
	SetConsoleCP(CP_UTF8);
	SetConsoleOutputCP(CP_UTF8);
#ifdef _UCRT
	setlocale(LC_ALL, ".UTF-8");
#else
    setlocale(LC_ALL, "");
#endif
```

mmicko (approved):
Did not notice issues on oss-cad-suite made build, there is also separate package for UCRT compiler, but there is issue of libraries missing, so can not switch easy to it. 
Think we can proceed with this as is, but think for safety that one ifdef could be added

kernel/driver.cc
whitequark: Will do!

whitequark: Added a guard around that call.

mmicko: Thanks. Have started extra flow build action manually to check once more

mmicko (approved):

PR #6092, opened 2026-08-06
opt_balance_tree: don't hang on combinational loops [sc-734]

### LINKED ISSUE #6091: opt_balance_tree gets stuck in BFS for designs with combo loops
### Version

Yosys 0.68+ (git sha1 72ce1b63d, GNU /usr/bin/c++ 15.2.0)

Linux

### Reproduction Steps

test.ys:
```
read_verilog <<EOT
module top(input a, output b);
  assign b = a | b;
endmodule
EOT
hierarchy -auto-top
proc
opt_balance_tree
```

`yosys -s test.ys`

### Expected Behavior

The pass will abandon the balance opportunity for that cascade as the logic is invalid.

### Actual Behavior

The pass gets unexpectedly [stuck here](https://github.com/YosysHQ/yosys/commit/b874d7de5bc2283caece98dc0f57c39d1d663ee9#diff-ed2d001e787d825958efdfd4410f14aefed63fd87d4be5dc74e593f69909f3abR193).

hexratcc:
This PR aims to fix #6091 - forward BFS now tracks visited cells and chains whose driver cone contains a cycle are abandoned before rebalancing. A plain visited set in the backward traversal would instead rewrite `b = a | b` to `b = a` and break reconvergent source multiplicity (`y = x + x`), so cyclic chains are skipped entirely.

### DIFF
--- a/passes/opt/opt_balance_tree.cc
+++ b/passes/opt/opt_balance_tree.cc
@@ -22,7 +22,6 @@
 #include "kernel/yosys.h"
 #include "kernel/sigtools.h"
 #include "kernel/utils.h"
-#include <deque>
 
 USING_YOSYS_NAMESPACE
 PRIVATE_NAMESPACE_BEGIN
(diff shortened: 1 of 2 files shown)


passes/opt/opt_balance_tree.cc
```diff
@@ -64,6 +69,43 @@ struct OptBalanceTreeWorker {
 		return y_width >= natural_width;
 	}
 
+	// Check if the driver graph reachable from head contains a cycle,
+	// following the same edges as the backward chain traversal
+	bool has_cycle(Cell *head, IdString cell_type) {
```
Ravenslofty: I don't think Yosys needs yet another implementation of a cycle-detection algorithm...

We have the `scc` pass to detect cycles, but it might be a bit clumsy to invoke here, so instead I think I would implement this by using a `TopoSort`; `sort()` will return false if it finds a loop (and therefore cannot be topologically sorted).

Ravenslofty (changes requested):

passes/opt/opt_balance_tree.cc
hexratcc: I've reworked the chain walk to now be a toposort.

Ravenslofty (approved):
This worked out even better than I was expecting.

widlarizer:
Well, while we're at it, it's not clear to me why we also have `kernel/topo_scc.h` added in 092214256, please investigate @hexratcc

Ravenslofty:
I know the `scc` pass has its own implementation as well, though that's probably a little complicated to convert because I had to make it aware of edges inside cells described by `specify` blocks.

hexratcc:
@widlarizer Looks like it's a generic tarjan that yields actual SCCs in topo order, which utils.h TopoSort can't do (it only flags loops). From what I can see the places where it's used need it, and not the alternative I've used here.

PR #6096, opened 2026-08-07
opt_muxtree: fix OOB read on self-conflicting mux Y port

hexratcc:
`wide_all.v` had a `rd[15:0]` vs `rd[15:8]` typo that multi-drove `rd[7:0]`, making both read ports collapse into a wide port with duplicated data. `opt_muxtree` then read `mux2info` out of bounds while reporting the multi-driver error (the offending mux isn't pushed yet when its own Y port repeats a bit), crashing instead of erroring. Push before the Y scan so the error reports cleanly. Additionally, generated test makefiles ignored autotest exit codes, which is why CI never caught this.

### DIFF
--- a/passes/opt/opt_muxtree.cc
+++ b/passes/opt/opt_muxtree.cc
@@ -144,6 +144,9 @@ struct OptMuxtreeWorker
 		// Analyze port A
 		muxinfo.ports.push_back(used_port_bit(sig_a, this_mux_idx, -1));
 
+		// push before the y port scan so self-conflicts can be reported too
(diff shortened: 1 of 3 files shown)


mmicko (approved):

widlarizer (approved):
context: #5958 introduced assuming single drivers to speed up the pass, but its check of whether a single driver is present wasn't perfect. So an already buggy`tests/memories/wide_all.v` started to error out, which we didn't notice because of test autogeneration step return code being ignored

PR #6077, opened 2026-08-03
abc_ops_reintegrate: xaiger2 compatibility

Ravenslofty:
_If your work is part of a larger effort, please discuss your general plans on [Discourse](https://yosyshq.discourse.group/) first to align your vision with maintainers._
Yet another part of the ABC reunification project.

_What are the reasons/motivation for this change?_
There is redundancy between `read_aiger -xaiger` and `read_xaiger2`, with the main difference being that the `abc_new` flow produces a slightly different symbol format that needed to be understood by `abc_ops_reintegrate`.

_Explain how this is achieved._
The difference in symbol files comes down to a different box format, where `abc9`-style boxes are "the entire box", and `abc_new` outputs per-box-port symbols (so-called "pseudo-primary-outputs", or "pseudopos"), which have to be mapped to the boxes in the original design.

_Make sure your change comes with tests. If not possible, share how a reviewer might evaluate it._
This should be covered by the `synth_gatemate` `abc_new` tests.

(please squash-merge; it's a bit awkward to preserve Emil's important commit while working around it)

PR #6101, opened 2026-08-10
read_xaiger2: remove

Ravenslofty:
More ABC reunification work.

With #6077, this pass isn't used by anything any more, and its functionality is handled by the more robust `read_aiger -xaiger` code, so it's no longer necessary.

Yeet.


hexratcc (approved):
lgtm, but I think we should update/remove the following 2 refs: 
- https://github.com/YosysHQ/yosys/blob/fd6367405eb50b252496087e3f12cfc1c06509a4/backends/aiger2/aiger.cc#L1502 
- https://github.com/YosysHQ/yosys/blob/fd6367405eb50b252496087e3f12cfc1c06509a4/docs/source/using_yosys/more_scripting/load_design.rst?plain=1#L131

KrystalDelusion (approved):

Issue #5639, opened 2026-01-27
ABC9 is replacing missing ports with ports connected to 0, breaking Nexus carry chains


gatecat:
### Version

Yosys 0.61+56 (git sha1 125609105-dirty, g++ 15.2.1 -fPIC -O3)

### On which OS did this happen?

Linux

### Reproduction Steps

The approach used for carry chain synthesis for nexus always uses first carry feed in cell that does not have a `CIN` port on it. This is because in practice, two carry chains might be placed next to each other so the `CIN` of the first (coming from hard routing) needs to be a "don't care": https://github.com/YosysHQ/yosys/blob/33e4b1d97f59eb6e3b1359b376ba2179b5d63ade/techlibs/lattice/arith_map_nexus.v#L67-L76 (nextpnr strictly enforces this).

When running an adder design like this through `synth_nexus -abc9`, this is replaced with a `CIN` connected to a constant zero, which cannot be placed with how nextpnr handles this currently. This definitely used to be left disconnected at some point in the past...
 
```verilog
module top(input [4:0] a, b, output [4:0] y);
assign y = a + b;
endmodule
```

```
 cell \VLO \y_OB_O_I_CCU2_S0_A1_VLO_Z
    connect \Z \y_OB_O_I_CCU2_S0_A1
  end

  attribute \module_not_derived 1
  attribute \src "add.v:2.12-2.17|/usr/local/bin/../share/yosys/nexus/arith_map.v:72.4-76.3"
  cell \CCU2 \y_OB_O_I_CCU2_S0_2_CIN_CCU2_COUT
    parameter \INIT0 "0xFFFF"
    parameter \INIT1 "0x00AA"
    parameter \INJECT "NO"
    connect \A0 \y_OB_O_I_CCU2_S0_D0
    connect \A1 \y_OB_O_I_CCU2_S0_A1
    connect \B0 \y_OB_O_I_CCU2_S0_D0
    connect \B1 \y_OB_O_I_CCU2_S0_D0
    connect \C0 \y_OB_O_I_CCU2_S0_D0
    connect \C1 \y_OB_O_I_CCU2_S0_D0
    connect \CIN \y_OB_O_I_CCU2_S0_A1
    connect \COUT \y_OB_O_I_CCU2_S0_2_CIN
    connect \D0 \y_OB_O_I_CCU2_S0_D0
    connect \D1 \y_OB_O_I_CCU2_S0_D0
[...]

PR #6103, opened 2026-08-10
synth_*: remove classic ABC mapping

Ravenslofty:
More ABC unification work.

I intend to remove `abc -lut`, so all users of it must be moved over to use `abc9 -lut` instead, and options enabling/disabling ABC9 must be removed.

I did my best to break the changes to each synthesis pass into their own commits to make it as easy as possible to review.

I have updated the synth tests where they exist.

cc @mole99 for the changes to `synth_fabulous`
cc @tony-min-1 / @Damon-Ma-MCHP for the changes to `synth_microchip`
cc @gatecat for the changes to `synth_nexus`

### DIFF
--- a/techlibs/achronix/synth_achronix.cc
+++ b/techlibs/achronix/synth_achronix.cc
@@ -51,9 +51,6 @@ struct SynthAchronixPass : public ScriptPass {
     log("    -noflatten\n");
     log("        do not flatten design before synthesis\n");
     log("\n");
-    log("    -retime\n");
-    log("        run 'abc' with '-dff -D 1' options\n");
-    log("\n");
     log("\n");
     log("The following commands are executed by this synthesis command:\n");
     help_script();
@@ -61,13 +58,12 @@ struct SynthAchronixPass : public ScriptPass {
   }
 
   string top_opt, family_opt, vout_file;
-  bool retime, flatten;
+  bool flatten;
 
   void clear_flags() override
   {
     top_opt = "-auto-top";
     vout_file = "";
-    retime = false;
     flatten = true;
   }
 
@@ -100,7 +96,7 @@ struct SynthAchronixPass : public ScriptPass {
           continue;
         }
         if (args[argidx] == "-retime") {
-          retime = true;
+          // Removed: ABC9 does not support retiming.
           continue;
         }
         break;
@@ -151,13 +147,11 @@ struct SynthAchronixPass : public ScriptPass {
         run("clean -purge");
         run("setundef -undriven -zero");
         run("dfflegalize -cell $_DFF_P_ x");
-        if (retime || help_mode)
-          run("abc -markgroups -dff -D 1", "(only if -retime)");
       }
 
     if (check_label("map_luts"))
       {
-        run("abc -lut 4" + string(retime ? " -dff -D 1" : ""));
+        run("abc9 -lut 4");
         run("clean");
       }
 
--- a/techlibs/common/synth.cc
+++ b/techlibs/common/synth.cc
@@ -89,9 +89,6 @@ struct SynthPass : public ScriptPass {
 		log("        from label is synonymous to 'begin', and empty to label is\n");
 		log("        synonymous to the end of the command list.\n");
 		log("\n");
-		log("    -abc9\n");
-		log("        use new ABC9 flow (EXPERIMENTAL)\n");
-		log("\n");
 		log("    -no-rw-check\n");
 		log("        marks all recognized read ports as \"return don't-care value on\n");
 		log("        read/write collision\" (same result as setting the no_rw_check\n");
@@ -174,6 +171,7 @@ struct SynthPass : public ScriptPass {
 			}
 			if (args[argidx] == "-lut" && argidx + 1 < args.size()) {
 				lut = atoi(args[++argidx].c_str());
+				abc = "abc9";
 				continue;
 			}
 			if (args[argidx] == "-nofsm") {
@@ -209,7 +207,7 @@ struct SynthPass : public ScriptPass {
 				continue;
 			}
 			if (args[argidx] == "-abc9") {
-				abc = "abc9";
+				// Removed: ABC9 is the default if -lut.
 				continue;
 			}
 			if (args[argidx] == "-no-rw-check") {
@@ -235,9 +233,6 @@ struct SynthPass : public ScriptPass {
 
 		if (!design->full_selection())
 			log_cmd_error("This command only operates on fully selected designs!\n");
-
-		if (abc == "abc9" && !lut)
-			log_cmd_error("ABC9 flow only supported for FPGA synthesis (using '-lut' option)\n");
 
 		log_header(design, "Executing SYNTH pass.\n");
 		log_push();
--- a/techlibs/ice40/synth_ice40.cc
+++ b/techlibs/ice40/synth_ice40.cc
@@ -45,8 +45,8 @@ struct SynthIce40Pass : public ScriptPass
 		log("This command runs synthesis for iCE40 FPGAs.\n");
 		log("\n");
 		log("    -device < hx | lp | u >\n");
-		log("        relevant only for '-abc9' flow, optimise timing for the specified\n");
-		log("        device. default: hx\n");
+		log("        optimise timing for the specified device. default: hx\n");
+		log("        (irrelevant if -noabc)\n");
 		log("\n");
 		log("    -top <module>\n");
 		log("        use the specified module as top module\n");
@@ -72,10 +72,7 @@ struct SynthIce40Pass : public ScriptPass
 		log("        do not flatten design before synthesis\n");
 		log("\n");
 		log("    -dff\n");
-		log("        run 'abc'/'abc9' with -dff option\n");
-		log("\n");
-		log("    -retime\n");
-		log("        run 'abc' with '-dff -D 1' options\n");
+		log("        run 'abc9' with -dff option\n");
 		log("\n");
 		log("    -nocarry\n");
 		log("        do not use SB_CARRY cells in output netlist\n");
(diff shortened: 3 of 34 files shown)


gatecat:
Unfortunately abc9 is currently broken on nexus (https://github.com/YosysHQ/yosys/issues/5639) - this will need a workaround either in Yosys and/or in nextpnr (doing it in the latter would break radiant compatibility, which is occasionally useful for debugging)

gatecat:
this is dealt with in nextpnr in https://github.com/YosysHQ/nextpnr/commit/a435792828e3114844e68fae09d7453a5f447ea7 so this is now good from my PoV

techlibs/achronix/synth_achronix.cc
```diff
@@ -100,7 +96,7 @@ struct SynthAchronixPass : public ScriptPass {
           continue;
         }
         if (args[argidx] == "-retime") {
-          retime = true;
+          // Removed: ABC9 does not support retiming.
```
hexratcc, line 99: `retime` is no longer initialized in `clear_flags()` but still read at line 150, where the old abc `retime` block also survived - should all of it go?

techlibs/xilinx/synth_xilinx.cc
```diff
@@ -645,9 +632,7 @@ struct SynthXilinxPass : public ScriptPass
 			}
 			if (help_mode)
 				run("abc -luts 2:2,3,6:5[,10,20] [-dff] [-D 1]", "(option for '-nowidelut', '-dff', '-retime')");
```
hexratcc: I think the help option should also be updated for abc9.

techlibs/xilinx/synth_xilinx.cc
```diff
@@ -645,9 +632,7 @@ struct SynthXilinxPass : public ScriptPass
...
-			else if (abc9) {
-				if (lut_size != 6)
-					log_error("'synth_xilinx -abc9' not currently supported for LUT4-based devices.\n");
+			else {
 				if (family != "xc7")
 					log_warning("'synth_xilinx -abc9' not currently supported for the '%s' family, "
 							"will use timing for 'xc7' instead.\n", family.c_str());
```
hexratcc: Warns about an option that no longer exists, and now fires for every non-xc7 family instead of only with `-abc9`

techlibs/xilinx/synth_xilinx.cc
```diff
@@ -660,37 +645,14 @@ struct SynthXilinxPass : public ScriptPass
 					k = stringf("synth_xilinx.abc9.%s.W", family);
 					abc9_opts += stringf(" -W %s", RTLIL::constpad.at(k, RTLIL::constpad.at("synth_xilinx.abc9.xc7.W")));
 				}
-				if (nowidelut)
+				if (nowidelut || lut_size != 6)
```
hexratcc: Wouldn't this mean that xc3s/xcv now lose `MUXF5`/`F6` - is this intentional?

techlibs/common/synth.cc
```diff
@@ -236,7 +237,7 @@ struct SynthPass : public ScriptPass {
 		if (!design->full_selection())
 			log_cmd_error("This command only operates on fully selected designs!\n");
 
-		if (abc == "abc9" && !lut)
+		if ((abc == "abc9" && !lut) || (abc != "abc9" && lut))
```
hexratcc: I think this is unreachable, as `-lut` sets `abc` to `"abc9"`.

techlibs/ice40/synth_ice40.cc
```diff
@@ -127,7 +118,7 @@ struct SynthIce40Pass : public ScriptPass
 	}
 
 	string top_opt, blif_file, edif_file, json_file, device_opt, latches;
-	bool nocarry, nodffe, nobram, spram, dsp, flatten, retime, noabc, abc2, vpr, abc9, dff, no_rw_check;
+	bool nocarry, nodffe, nobram, spram, dsp, flatten, retime, noabc, vpr, abc9, dff, no_rw_check;
```
hexratcc: `retime` and `abc9` are dead now (and `abc9` is no longer cleared).

techlibs/ice40/synth_ice40.cc
```diff
@@ -99,16 +96,10 @@ struct SynthIce40Pass : public ScriptPass
 		log("    -noabc\n");
 		log("        use built-in Yosys LUT techmapping instead of abc\n");
```
hexratcc: I think this should say abc9 instead.

techlibs/xilinx/synth_xilinx.cc
```diff
@@ -144,7 +137,7 @@ struct SynthXilinxPass : public ScriptPass
 
 	std::string top_opt, edif_file, blif_file, json_file, family;
 	bool flatten, retime, ise, noiopad, noclkbuf, nobram, nolutram, nosrl, nocarry, nowidelut, nodsp, uram;
```
hexratcc: `retime` is also dead here.

techlibs/lattice/synth_lattice.cc
```diff
@@ -170,7 +161,7 @@ struct SynthLatticePass : public ScriptPass
 	}
 
 	string top_opt, edif_file, json_file, family, latches;
-	bool noccu2, nodffe, nobram, nolutram, nowidelut, asyncprld, flatten, dff, retime, abc2, abc9, iopad, nodsp, no_rw_check, have_dsp;
+	bool noccu2, nodffe, nobram, nolutram, nowidelut, asyncprld, flatten, dff, iopad, nodsp, no_rw_check, have_dsp;
 	bool cmp2softlogic;
 	string postfix, arith_map, brams_map, dsp_map, cells_map, map_ram_default, widelut_abc;
```
hexratcc: `widelut_abc` is dead.

techlibs/lattice/synth_lattice.cc
```diff
@@ -544,38 +522,25 @@ struct SynthLatticePass : public ScriptPass
...
+			std::string abc9_opts;
+			if (nowidelut)
+				abc9_opts += " -maxlut 4";
+			std::string k = "synth_lattice.abc9.W";
+			if (active_design && active_design->scratchpad.count(k))
+				abc9_opts += stringf(" -W %s", active_design->scratchpad_get_string(k));
+			else
+				abc9_opts += stringf(" -W %s", RTLIL::constpad.at(k));
+			if (nowidelut)
+				abc9_opts += " -maxlut 4";
```
hexratcc, line 536: `-maxlut 4` is appended twice (here and on line 533), looks like this was already there, please fix.

techlibs/nanoxplore/cells_sim.v
```diff
@@ -67,7 +64,6 @@ assign O = data_inv ? O_reg : ~O_reg;
 endmodule
 
 
-(* abc9_box, lib_whitebox *)
 module NX_CY(input A1, A2, A3, A4, B1, B2, B3, B4, (* abc9_carry *) input CI, output S1, S2, S3, S4, (* abc9_carry *) output CO);
```
hexratcc: I think `(* abc9_carry *)` shouldn't be here anymore

tests/arch/ice40/bug1626.ys
```diff
@@ -214,4 +214,4 @@ module \ahb_async_sram_halfwidth
 end
 EOT
 
-synth_ice40 -abc2 -abc9
+synth_ice40 -abc9
```
hexratcc: See ice40 comment - I believe `-abc9` doesn't do anything here.

passes/techmap/nlutmap.cc
```diff
@@ -54,7 +54,7 @@ struct NlutmapWorker
 		Pass::call_on_selection(module->design, get_selection(), "lut2mux");
 
 		if (lut_size > 0)
-			Pass::call_on_selection(module->design, get_selection(), stringf("abc -lut 1:%d", lut_size));
+			Pass::call_on_selection(module->design, get_selection(), stringf("abc9 -lut 1:%d", lut_size));
 		else
 			Pass::call_on_selection(module->design, get_selection(), "abc");
```
hexratcc: This still runs classic abc - intended?

passes/techmap/nlutmap.cc
```diff
@@ -54,7 +54,7 @@ struct NlutmapWorker
 		Pass::call_on_selection(module->design, get_selection(), "lut2mux");
 
 		if (lut_size > 0)
-			Pass::call_on_selection(module->design, get_selection(), stringf("abc -lut 1:%d", lut_size));
+			Pass::call_on_selection(module->design, get_selection(), stringf("abc9 -lut 1:%d", lut_size));
```
hexratcc: nlutmap runs ABC on a partial selection after the first round, but abc9 only accepts full modules, so `synth_greenpak4` now errors out on any design with logic:
```sv
module top(input a, b, c, output y);
  assign y = a & b | c;
endmodule
```
```
yosys -p 'read_verilog top.v; synth_greenpak4' -> ERROR: Can't handle partially selected module top!
```

hexratcc (changes requested):
Mostly just minor nitpicks:
- I think the docs failure is caused by these stale commands:
https://github.com/YosysHQ/yosys/blob/488ba22092d121ba315632280bf2fc1b112000ba/docs/source/code_examples/macro_commands/synth_ice40.ys#L67-L68
- This option shouldn't be documented imo if it's a no-op now:
https://github.com/YosysHQ/yosys/blob/488ba22092d121ba315632280bf2fc1b112000ba/techlibs/common/synth.cc#L92-L93
- `widelut_size` is now never read as far as I can tell:
https://github.com/YosysHQ/yosys/blob/488ba22092d121ba315632280bf2fc1b112000ba/techlibs/xilinx/synth_xilinx.cc#L144

hexratcc:
Interesting that some of my code selection is broken, looks like a plugin issue, sorry.

passes/techmap/nlutmap.cc
Ravenslofty: in this context this is the classic ABC standard cell flow, rather than the LUT flow. it'll probably get removed in another PR, but out of scope for this one.

techlibs/xilinx/synth_xilinx.cc
Ravenslofty: ended up reusing some of the old code to handle this better.

passes/techmap/nlutmap.cc
Ravenslofty: https://github.com/YosysHQ/yosys/blob/dbe5b7c03f0608df3d87bcd2053512da943ff3fd/passes/techmap/abc9.cc#L396-L398
I guess it does something after all...

Ravenslofty:
The `nlutmap` issue needs some more investigation, but I've pushed the fixes for the first round of review anyway for further inspection.

ArcaneNibble:
I have discussed, in a private conversation with @Ravenslofty , that I am okay with `synth_coolrunner2` being deprecated and removed as a result of this migration.

I do want sum-of-products/PLD workflows to _theoretically_ be possible in Yosys (for retrodev, teaching, or other purposes), but the flow as currently implemented does not actually work all that well. As long as these architectures don't get _completely_ conceptually forgotten, I can accept needing to reimplement them to a higher standard.

azonenberg:
Same as above, if synth_greenpak is blocking work for more modern architectures I'm OK with it being removed. I can always port it over to the new abc9 flow if we find that it's necessary but I haven't touched a greenpak in quite a while

hexratcc (approved):
lgtm to me, some minor cleanups could be done, but I think it's in an ok state once all the PR's get merged.

PR #6106, opened 2026-08-11
opt_expr: respect -keepdc for muxes with a fully undef select

### LINKED ISSUE #4988: Inconsistent 'x handling in opt_expr for muxes
### Version

0.51+107

Linux

### Reproduction Steps

uut.il
```
module \uut
  wire output 1 \out
  wire input 1 \a
  wire input 1 \b
  cell $mux \mux
    parameter \WIDTH 1
    connect \A \a
    connect \B \b
    connect \S 1'x
    connect \Y \out
  end
end
```
Call `yosys -p 'opt_expr -mux_undef [-keepdc]; dump' uut.il`

For reference:
`yosys -p 'opt_expr; dump' uut.il` has `out` directly connected to `a`.
`yosys -p 'opt_expr -keepdc; dump' uut.il` is unchanged.

### Expected Behavior

`yosys -p 'opt_expr -mux_undef' uut.il` shouldn't connect `out` to `a`, I would have expected it to connect `out` to `1'x` (which is what happens when both `A` and `B` are undef), but I'm not entirely sure.
`yosys -p 'opt_expr -mux_undef -keepdc; dump' uut.il` should be unchanged.

### Actual Behavior

`opt_expr -mux_undef` with or without `-keepdc` gives
```
module \uut
  wire input 1 \b
  wire input 2 \a
  wire output 3 \out
  connect \out \a
end
```
i.e. the same behaviour as `opt_expr` without `-keepdc`, completely ignoring the x bits..

phverg:
## Summary

Fixes #4988.

`opt_expr -mux_undef` replaced a mux whose select `S` is fully undef with its `A` input **even under `-keepdc`**. `-keepdc` exists to preserve don't-care (`x`) semantics for formal/equivalence flows, so silently resolving the undef select to `A` is a miscompile: the output goes from a don't-care to a concrete value.

This affected three cell paths, all now handled consistently:

* **`$mux`/`$pmux` fast path** (`opt_expr.cc:1489`): the old code OR'd "both data inputs undef" with "select undef" and replaced with `A` unconditionally. These are split — `A&&B` fully undef still replaces with `A` unconditionally (the output is `x` regardless of the select, so no don't-care is resolved and it is fine under `-keepdc`), while a fully undef select resolves to `A` only when `!keepdc`. The trailing `goto next_cell` also stops the per-select-bit loop below from re-collapsing an all-undef select via its `mux_empty` path.
* **Gate-level `$_MUX_`** (`opt_expr.cc:1161`): the undef-select case (`"  *"`) is now gated on `!keepdc` as well, so a post-techmap `$_MUX_` behaves the same as `$mux` under `-keepdc`.

## Before / after (reporter's testcase)

```
$ yosys -p 'read_rtlil uut.il; opt_expr -mux_undef -keepdc; dump'
```

* **Before:** the mux is removed and `\out` is connected to `\a` (don't-care lost).
* **After:** the mux is left unchanged, as expected for `-keepdc`.

`opt_expr -mux_undef` without `-keepdc` still resolves to `\out = \a` as before.

## Testing

* Extended `tests/opt/opt_expr_mux_undef.ys` with fully-undef-select cases for `$mux`, `$pmux` (multi-bit data and select) and `$_MUX_`: under `-keepdc` the mux is asserted still present, without `-keepdc` it is asserted removed. The new assertions fail on the unpatched tree and pass with the fix.
* `tests/opt` (89 scripts) and `tests/various` all pass.

## Scope / notes for reviewers

* The non-`keepdc` behaviour (resolve a fully undef select to `A`) is intentionally left unchanged — picking `A` is a valid don't-care resolution and `-mux_undef` is documented as being allowed to choose values for undef inputs. The issue mentions `1'x` as a possible alternative for the non-keepdc result; happy to make that change separately if preferred, but it would be a behavioural change across existing flows.
* This PR is scoped to a **fully** undef select, matching the issue. Resolution of individual undef select bits / undef data inputs in partially-undef `$pmux` cells (the per-bit loop and the `$_MUX_` undef-data cases) is pre-existing and not guarded by `-keepdc`; that can be addressed as a follow-up if desired.

### DIFF
--- a/passes/opt/opt_expr.cc
+++ b/passes/opt/opt_expr.cc
@@ -1158,7 +1158,9 @@ skip_fine_alu:
 			if (mux_undef) {
 				if (input.match("*  ")) ACTION_DO(ID::Y, input.extract(1, 1));
 				if (input.match(" * ")) ACTION_DO(ID::Y, input.extract(2, 1));
-				if (input.match("  *")) ACTION_DO(ID::Y, input.extract(2, 1));
+				// An undef select resolves to the A input only when we may pick
+				// values for don't-cares; with -keepdc leave the cell untouched.
+				if (!keepdc && input.match("  *")) ACTION_DO(ID::Y, input.extract(2, 1));
 			}
 		}
 
@@ -1486,9 +1488,19 @@ skip_identity:
 		if (mux_undef && cell->type.in(ID($mux), ID($pmux))) {
 			RTLIL::SigSpec new_a, new_b, new_s;
 			int width = GetSize(cell->getPort(ID::A));
-			if ((cell->getPort(ID::A).is_fully_undef() && cell->getPort(ID::B).is_fully_undef()) ||
-					cell->getPort(ID::S).is_fully_undef()) {
+			// If both data inputs are fully undef the output is undef regardless
+			// of the select, so replacing the mux with A is always valid. This
+			// does not resolve any don't-care and is therefore fine with -keepdc.
+			if (cell->getPort(ID::A).is_fully_undef() && cell->getPort(ID::B).is_fully_undef()) {
 				replace_cell(assign_map, module, cell, "mux_undef", ID::Y, cell->getPort(ID::A));
+				goto next_cell;
+			}
+			// A fully undef select makes the output a don't-care. Resolving it to
+			// the A input is only valid when we are allowed to pick values for
+			// don't-cares; with -keepdc the cell must be left untouched.
+			if (cell->getPort(ID::S).is_fully_undef()) {
+				if (!keepdc)
+					replace_cell(assign_map, module, cell, "mux_undef", ID::Y, cell->getPort(ID::A));
 				goto next_cell;
 			}
 			for (int i = 0; i < cell->getPort(ID::S).size(); i++) {
--- a/tests/opt/opt_expr_mux_undef.ys
+++ b/tests/opt/opt_expr_mux_undef.ys
@@ -287,3 +287,76 @@ select -assert-none t:$_MUX16_
 select -assert-count 1 o:out %ci*
 
 ##########
+# fully undef select, defined data inputs
+# -keepdc: the output is a don't-care that must be preserved,
+#          so the mux is left unchanged
+# no -keepdc: the undef select is resolved to the A input
+# checked for $mux, $pmux and the gate-level $_MUX_
+
+# $mux
+design -reset
+read_rtlil <<EOT2
+module \uut
+  wire output 1 \out
+  wire input 1 \a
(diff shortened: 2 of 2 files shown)


widlarizer:
Did you actually run into this issue, or are you fixing a random issue you don't understand enough to provide a hand-written description of the fix?

phverg:
This fixes #4988, which was filed by @KrystalDelusion. I was going through the issue tracker to help out and thought this one looked like a small, self-contained fix.

To my understanding, `opt_expr -mux_undef` replaces a mux whose select is fully undef with its `A` input. `-keepdc` is supposed to disable optimizations that change don't-care behavior, but this path was not gated on it, so the don't-care select was silently resolved to a concrete value even under `-keepdc`. On the issue's testcase, `opt_expr -mux_undef -keepdc; dump` on master removes the cell and emits `connect \out \a`, while with the patch the `$mux` is left untouched, which is what `-keepdc` should do.

For `$mux`/`$pmux` the fix splits the original condition. The case where `A` and `B` are both fully undef still collapses to `A` unconditionally, which is don't-care safe because the output is `x` for every select value, so nothing is resolved. Only the fully-undef select case is gated on `!keepdc`. The `goto next_cell` matters too. Without it, the per-select-bit loop just below would re-collapse an all-undef select to `A` through its `mux_empty` path, so gating only the fast path would not be enough.

The same bug exists for the gate-level `$_MUX_` at line 1161 (`"  *"` maps to `Y = A`), so that is gated on `!keepdc` as well.

The non-`keepdc` behavior is intentionally unchanged. The added test asserts the mux survives under `-keepdc` and is removed without it. It fails on master and passes with the patch.

From your response I get the feeling this issue is not a priority. I am happy to close the PR if you do not like the fix or think the issue is not important.

widlarizer:
While that issue is pretty concerning, pointing an LLM harness to an issue isn't helping out, it's practically spam. I expect you're a Yosys user who decided to help out but there's no added value here beyond the single digit dollars at API pricing. I can prompt an LLM to fix an issue like that myself, the value of a code change comes from experience with the problem at hand, with the codebase, with software engineering expertise, and with the ability to concisely communicate the problem solving. The description and the code comments here are vague, verbose, use undefined terms. These are problems humans create too, but when I take the time to read and respond to their outputs, I'm investing into a contributor's skill level and rewarding effort. So your code is a proof of concept but writing it out isn't the hard part, understanding and sharing that understanding is, and I'm not convinced you've done a good job there. On different sorts of issue, an LLM-generated proof of concept can be super valuable actually, but again, if I wanted raw LLM outputs, I'd generate them myself. I can afford it

PR #6107, opened 2026-08-11
opt_clean: move unused_bits analysis out

widlarizer:
opt_clean is a critical part of yosys and should be as simple as possible. And as it's providing the most basic garbage collection that runs dozens of times per synth run, it shouldn't be bundled with code that sets "just in case" unused_bits attributes. These are only used by two cold passes that create implicit pass sequencing demands. I expect a free speedup from this or minimal impact

TODOs
- [ ] check performance impact


passes/fsm/fsm_opt.cc
```diff
@@ -294,7 +281,7 @@ struct FsmOpt
 		}
 	}
 
-	FsmOpt(RTLIL::Cell *cell, RTLIL::Module *module)
+	FsmOpt(RTLIL::Cell *cell, RTLIL::Module *module) : unused_bits(module)
```
hexratcc: Since port directions don't change here, couldn't you build it once per pass and use `UnusedBits(module, ct)`, like `wrapcell`?

hexratcc (approved):
lgtm, I've also run some quick benchmarks locally (flattened picorv32), and I'm seeing a ~7% decrease in `opt_clean` runtime and a ~3% peak memory usage decrease (16 threads).

hexratcc (approved):

PR #6109, opened 2026-08-12
gatemate_foldinv: count module output ports as consumers [sc-737]

hexratcc:
As discussed in https://github.com/YosysHQ/yosys/pull/6077#issuecomment-5238903257, `fold_output_inverters` only counted cell inputs as consumers, so inverters driving module outputs were left unfolded. This showed up as extra `CC_LUT1` cells with `-abc_new`. This PR modifies the fold so that it also these bits as used.


Ravenslofty (approved):
lgtm, @mmicko are you happy with it too?

mmicko (approved):

mmicko:
LGTM as well

PR #6110, opened 2026-08-12
opt_dff -sat: cleanup and refactor [sc-725]

hexratcc:
Followup to #6030, after some discussion with @widlarizer, we've come to the conclusion that in order to improve maintainability of this pass, a cleanup/rewrite will be needed, so here it is. This PR aims to improve the maintainability of -sat `constbits`/`eqbits`, as well as further improve the `opt_dff` test situation. `constbits` is now split into individual phases (gather/solve/commit) with more explicit batching and a single SAT query per obligation, and `eqbits` state is more clearly defined. This PR also doesn't intend to add any functional changes, and performance is just marginally faster (~3% from my testing).


passes/opt/dff/constbits.cc
```diff
@@ -0,0 +1,318 @@
...
+		log("Setting constant %d-bit at position %d on %s (%s) from module %s.\n",
+				ob.val == State::S1 ? 1 : 0, ob.idx, ob.cell, ob.cell->type.unescape(), worker.module);
+		worker.initvals.remove_init(ob.q);
+		worker.module->connect(ob.q, ob.val);
+		const_bits[ob.cell].insert(ob.idx);
+	}
+
+	bool add_const_target(ConstObligation &ob, SigBit sig)
+	{
+		if (!worker.opt.sat || (ob.val != State::S0 && ob.val != State::S1))
```
widlarizer: What puzzles me, is that opt.sat gets checked here. If I am reading this correctly, that makes all changes unfeasible, so `gather_const_obligations` does nothing to the design, but it does iterate over all cells and constructs FfData for all FFs. That sounds like even a performance risk. Then there's no obligations, and nothing else in this file does anything. The fact that opt.sat is checked this deep into the control flow of this flow suggests that there is valid usage of these functions without opt.sat being set, but without it, it does nothing, so the guard could be moved out of constbits.cc into opt_dff.cc execute. And so could eqbits.cc, though there it was at least at its own top level in run_eqbits

passes/opt/dff/constbits.cc
```diff
@@ -0,0 +1,318 @@
...
+			if (batch_end > batch_begin && GetSize(qcsat.imported_cells) >= sat_batch_cells)
+				break;
+			ob.q_lit = qcsat.importSigBit(ob.q);
+			int vlit = qcsat.ez->value(ob.val == State::S1);
+			std::vector<int> differ;
+			for (auto sig : ob.targets)
+				differ.push_back(qcsat.ez->NOT(qcsat.ez->IFF(qcsat.importSigBit(sig), vlit)));
+			ob.differ_lit = qcsat.ez->expression(ezSAT::OpOr, differ);
+			qcsat.prepare();
+			worker.sat_budget.charge_import(qcsat, cells_charged);
```
widlarizer: I find cells_charged being both an input and output parameter reference, storing budget usage state outside of the budget state pretty confusing. One light-weight way to do this differently would be to pass it in by value and return the new value, do `cells_charged = worker.sat_budget.charge_import(qcsat, cells_charged);` or return the additionally charged count and have it be `cells_charged += ...`

passes/opt/dff/constbits.cc
```diff
@@ -0,0 +1,318 @@
...
+					watches.watch(ob);
+			}
+
+			for (int obi = batch_begin; obi < batch_end; obi++) {
+				auto &ob = obligations[obi];
+				if (ob.status != ConstObligation::Pending)
+					continue;
+				if (worker.warn_if_budget_spent())
+					return;
+				if (!resolve_const_obligation(qcsat, cap, ob, watches))
```
widlarizer: The return value here can seem like it's conveying something that was found out about the design, rather than about the SAT effort limit being hit and no more optimization being impossible. Could be just a naming thing. Maybe invert the bool and name it `given_up` ?

widlarizer (changes requested):
Constbits is now a lot easier to understand, and covered with tests. Just some changes we want for extra clarity:

passes/opt/dff/constbits.cc
hexratcc: `constbits` still does work without `-sat`: all-const-input bits are folded trivially (`opt_dff_const.ys`), so its guard can't move to execute. Hoisted it to the gather loop as an explicit `needs_sat` check instead. `eqbits` only ever executes under `-sat`, so I moved it as suggested.

widlarizer (approved):

phsauter:
Does this impact #5548 as well?  
(I still have it as an open todo, I just had some more urgent things to knock-out this week)

widlarizer:
@phsauter Yes, things moved around in the refactor. If that's inconvenient for your patch, we can take over and rebase it if you'd like

phsauter:
> @phsauter Yes, things moved around in the refactor. If that's inconvenient for your patch, we can take over and rebase it if you'd like

No, should be fine then. I was just wondering if this may have already covered the other problem in some way.

Issue #4623, opened 2024-10-01
log_deprecated


widlarizer:
### Feature Description

When discussing the use case for `opt_demorgan` the idea came up that there may be unloved and unused parts of yosys that would best be removed. However, since that might break somebody's workflow, it should be communicated ahead of time to the users. For such cases we should add log_deprecated so that users know ahead of time their workflow may be affected in the future with a feature removal

PR #6115, opened 2026-08-13
log rewrite [sc-748]

mmicko:
This one is based on old PR https://github.com/YosysHQ/yosys/pull/5311 and continues work on that

- LogSeverity includes more options
- LogMessage preserves format and timestamp as well
- timestamp is now using chrono
- ~~code is using callback till sinks are added and existing code use them~~
- logs are matching previously generated

Task list:
- [x] remove log_last_error (this was added for YosysJS)
- [x] add LogSink
- [x] create ConsoleLogSink and StdErrorLogSink
- [x] create FileLogSink and use it in driver
- [x] create ScratchPadSink
- [x] create HelpMsgSink and use it in register.cc
- [x] make AST dump not using log_files
- [x] remove log_files, log_streams, log_scratchpads, log_errfile, log_error_stderr, log_quiet_warnings, log_verbose_level, log_time
- [x] get rid of static variables

Potential future updates improvements:
- [ ] add filtering to LogSink based on LogMessage
- [ ] differentiate fatal error and regular error
- [ ] get rid of macros

### DIFF
--- a/examples/cxx-api/demomain.cc
+++ b/examples/cxx-api/demomain.cc
@@ -5,7 +5,7 @@
 
 int main()
 {
-	Yosys::log_streams.push_back(&std::cout);
+	Yosys::log_sinks.push_back(std::make_unique<Yosys::ConsoleLogSink>());
 	Yosys::log_error_stderr = true;
 
 	Yosys::yosys_setup();
--- a/examples/cxx-api/demomain.cc
+++ b/examples/cxx-api/demomain.cc
@@ -5,8 +5,7 @@
 
 int main()
 {
-	Yosys::log_sinks.push_back(std::make_unique<Yosys::ConsoleLogSink>());
-	Yosys::log_error_stderr = true;
+	Yosys::logger().add_sink<Yosys::ConsoleLogSink>();
 
 	Yosys::yosys_setup();
 	Yosys::yosys_banner();
--- a/kernel/driver.cc
+++ b/kernel/driver.cc
@@ -608,9 +608,14 @@ int main(int argc, char **argv)
 		if (logger().get_warnings_total())
 			log("Warnings: %d unique messages, %d total\n", logger().get_warnings_unique(), logger().get_warnings_total());
 
-		if (logger().get_experimentals_num() != 0)
-			log("Warnings: %d experimental features used (not excluded with -x).\n", logger().get_experimentals_num());
+		if (GetSize(logger().get_experimental()) != 0)
+			log("Warnings: %d experimental features used (not excluded with -x).\n", GetSize(logger().get_experimental()));
 
+		if (GetSize(logger().get_deprecated()) != 0) {
+			log("Deprecated features used:\n");
+			for(auto &feature : logger().get_deprecated())
+				log("    %s\n", feature);
+		}
 #ifdef _WIN32
 		log("End of script. Logfile hash: %s\n", hash);
 		(void)wall_clock_start;
--- a/kernel/driver.cc
+++ b/kernel/driver.cc
@@ -604,7 +604,12 @@ int main(int argc, char **argv)
 		yosys_xtrace = 0;
 		log_spacer();
 
-		logger().report_warning_stats(mode_v && !mode_q);
+		logger().set_log_forced(mode_v && !mode_q);
+		if (logger().get_warnings_total())
+			log("Warnings: %d unique messages, %d total\n", logger().get_warnings_unique(), logger().get_warnings_total());
+
+		if (logger().get_experimentals_num() != 0)
+			log("Warnings: %d experimental features used (not excluded with -x).\n", logger().get_experimentals_num());
 
 #ifdef _WIN32
 		log("End of script. Logfile hash: %s\n", hash);
--- a/kernel/driver.cc
+++ b/kernel/driver.cc
@@ -161,6 +161,7 @@ int main(int argc, char **argv)
 	bool run_tcl_shell = false;
 	bool mode_v = false;
 	bool mode_q = false;
+	FILE *log_errfile = NULL;
 
 	cxxopts::Options options(argv[0], "Yosys Open SYnthesis Suite");
 	options.set_width(SIZE_MAX);
@@ -314,12 +315,11 @@ int main(int argc, char **argv)
 		for (const auto& key : {"l", "L"}) {
 			if (result.count(key)) {
 				for (const auto& filename : result[key].as<std::vector<std::string>>()) {
-					if (FILE* f = fopen(filename.c_str(), "wt")) {
-						log_files.push_back(f);
-						if (key[0] == 'L') setvbuf(f, NULL, _IOLBF, 0);
-					} else {
-						std::cerr << "Can't open log file `" << filename << "' for writing!\n";
-						exit(1);
+					try {
+						log_sinks.push_back(std::make_unique<FileLogSink>(filename, key[0] == 'L', false));
+					} catch (const std::runtime_error &e) {
+ 					   std::cerr << e.what();
+					   exit(1);
 					}
 				}
 			}
@@ -405,8 +405,9 @@ int main(int argc, char **argv)
 		}
 
 		if (log_errfile == NULL) {
-			log_files.push_back(stdout);
-			log_error_stderr = true;
+			log_sinks.push_back(std::make_unique<ConsoleLogSink>());
+		} else {
+			log_sinks.push_back(std::make_unique<StderrLogSink>());
 		}
 
 		if (print_banner)
@@ -609,7 +610,7 @@ int main(int argc, char **argv)
 		log_spacer();
 
 		if (mode_v && !mode_q)
-			log_files.push_back(stderr);
+			log_stderr_force = true;
 
 		if (log_warnings_count)
 			log("Warnings: %d unique messages, %d total\n", GetSize(log_warnings), log_warnings_count);
--- a/kernel/driver.cc
+++ b/kernel/driver.cc
@@ -161,7 +161,7 @@ int main(int argc, char **argv)
 	bool mode_v = false;
 	bool mode_q = false;
 	bool quiet_warnings = false;
-	FILE *log_errfile = NULL;
+	bool log_stderr = false;
 
 	cxxopts::Options options(argv[0], "Yosys Open SYnthesis Suite");
 	options.set_width(SIZE_MAX);
@@ -326,12 +326,12 @@ int main(int argc, char **argv)
 		}
 		if (result.count("q")) {
 			mode_q = true;
-			if (log_errfile == stderr) quiet_warnings = true;
-			log_errfile = stderr;
+			if (log_stderr) quiet_warnings = true;
+			log_stderr = true;
 		}
 		if (result.count("v")) {
 			mode_v = true;
-			log_errfile = stderr;
+			log_stderr = true;
 			logger().set_verbose_level(result["v"].as<int>());
 		}
 		if (result.count("t")) logger().set_log_time(true);
@@ -404,7 +404,7 @@ int main(int argc, char **argv)
 			Hasher::set_fudge((Hasher::hash_t)seed);
 		}
 
-		if (log_errfile == NULL) {
+		if (!log_stderr) {
 			logger().add_sink<ConsoleLogSink>();
 		} else {
 			logger().add_sink<StderrLogSink>(quiet_warnings);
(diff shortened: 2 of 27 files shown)


kernel/log.cc
```diff
@@ -38,79 +38,145 @@
...
+			time_str += stringf("[%05d.%06d] ", int(us.count() / 1'000'000),int(us.count() % 1'000'000));
+		}
+
+		if (!format.empty() && format.back() == '\n')
+			next_print_log = true;
+
+		// Special case to detect newlines in Python log output, since
+		// the binding always calls `log("%s", payload)` and the newline
+		// is then in the first formatted argument
+		if (format == "%s" && message.back() == '\n')
```
hexratcc: I think we should guard with `!message.empty()`, from what I understand pyosys binds `log_warning` with `"%s"`

passes/cmds/tee.cc
```diff
@@ -58,69 +58,42 @@ struct TeePass : public Pass {
...
-				yosys_input_files.insert(args[argidx]);
-				if (f == NULL) {
-					for (auto cf : files_to_close)
-						fclose(cf);
-					log_cmd_error("Can't create file %s.\n", args[argidx]);
+				try {
+					logger().add_sink<FileLogSink>(path.c_str(), false, is_append);
+				} catch (const std::runtime_error &e) {
+					std::cerr << e.what();
+					exit(1);
```
hexratcc: I think this disables tcl sessions, which previously wasn't the case - is this intended?

passes/cmds/tee.cc
```diff
@@ -58,69 +58,42 @@ struct TeePass : public Pass {
...
-		backup_log_scratchpads = log_scratchpads;
+		auto log_scope = logger().scoped();
 
 		size_t argidx;
 		for (argidx = 1; argidx < args.size(); argidx++)
 		{
-			if (args[argidx] == "-q" && files_to_close.empty()) {
-				log_files.clear();
-				log_streams.clear();
+			if (args[argidx] == "-q") {
```
hexratcc, line 66:
```cmd
$ yosys -q -p 'tee -o out.log -q help tee'; echo $?; wc -c out.log
main binary (f4bc771a1):
ERROR: No such command: -q (type 'help' for a command overview)
1
64 out.log
PR binary (ec80d539e):
0
0 out.log
```
On main, `out.log` contains that same ERROR line, but due to this change it's now empty.

kernel/log.h
```diff
@@ -85,162 +86,472 @@ YOSYS_NAMESPACE_BEGIN
...
-extern string log_last_error;
-extern void (*log_error_atexit)();
+enum LogSeverity {
+	LOG_DEBUG,
+	LOG_COMMENT,
+	LOG_INFO,
+	LOG_HEADER,
+	LOG_WARNING,
+	LOG_ERROR
+};
```
hexratcc: I'd expect this to clash with glibc's `<syslog.h>` macros - can this be made into a scoped enum?

pyosys/generator.py
```diff
@@ -119,19 +119,15 @@ def __post_init__(self):
...
-        "log_warn_regexes",
-        "log_nowarn_regexes",
-        "log_werror_regexes",
         ## function pointers
         "log_error_atexit",
         "log_verific_callback",
         # incomplete type
         "yosys_get_tcl_interp",
+        # manually added later
+        "logger",
```
hexratcc: Is this correct? I don't see a a manual logger binding.

kernel/driver.cc
```diff
@@ -161,6 +160,8 @@ int main(int argc, char **argv)
 	bool run_tcl_shell = false;
 	bool mode_v = false;
 	bool mode_q = false;
+	bool quiet_warnings = false;
+	FILE *log_errfile = NULL;
```
hexratcc: Since this is only ever used for checks, can we make this into a `bool`?

kernel/log.h
```diff
@@ -85,162 +86,472 @@ YOSYS_NAMESPACE_BEGIN
...
+private:
+	FILE *file;
+};
+
+class ConsoleLogSink : public LogSink
+{
+public:
+	void log(const LogMessage &msg) override;
+	void flush() override;
+	FILE *file_handle() override { return stdout; };
```
hexratcc: Stray `;`.

hexratcc (changes requested):
Mainly minor notes, I've also noticed that some files have both space and tab indents.

kernel/log.h
mmicko: Would go with next proposal, makes it scoped, but protects form usual macro defines as well
```
enum class LogSeverity {
	Debug,
	Comment,
	Info,
	Header,
	Warning,
	Error
};
```

passes/cmds/tee.cc
mmicko: ah, was copy pasting from driver. my mistake

passes/cmds/tee.cc
mmicko: Solving this one, but I think current main is wrong since it is ignoring -q in cases you have already specified files.  Fixing it by removing all sinks from original if -q options is specified

mmicko:
@hexratcc addressed all points, also did clear places where I mixed spaces/tabs (some were already there)

hexratcc (approved):
lgtm, but I'd wait for @widlarizer to see what he has to say.

mmicko:
Implemented feature from https://github.com/YosysHQ/yosys/issues/4623 and marked `opt_demorgan` as deprecated

examples/cxx-api/demomain.cc
```diff
@@ -5,7 +5,7 @@
 
 int main()
 {
-	Yosys::log_streams.push_back(&std::cout);
+	Yosys::log_sinks.push_back(std::make_unique<Yosys::ConsoleLogSink>());
```
widlarizer: `Yosys::log_sinks` no longer exists AFAICT - needs a final compilation check when the interface is finished

kernel/log.h
```diff
@@ -85,162 +86,483 @@ YOSYS_NAMESPACE_BEGIN
...
+}
+
+template <typename... Args>
+inline void log_comment(FmtString<TypeIdentity<Args>...> fmt, const Args &... args)
+{
+	logger().log_formatted_string({}, fmt.format_string(), fmt.format(args...), LogSeverity::Comment);
+}
+
+template <typename... Args>
+inline void log_debug(FmtString<TypeIdentity<Args>...> fmt, const Args &... args)
```
widlarizer: `log_debug` being a macro before would strongly prevent the arguments to be constructed, we have inline now and I haven't seen a performance difference, but I'm still interested in the reasoning about this change

kernel/log.h
```diff
@@ -85,162 +86,483 @@ YOSYS_NAMESPACE_BEGIN
...
+
+inline void log_experimental(const std::string &str)
+{
+	logger().log_experimental(str);
 }
 
-void log_experimental(const std::string &str);
+inline void log_deprecated(const std::string &str)
+{
+	logger().log_deprecated(str);
```
widlarizer: log_experimental would only print one callsite once, while log_deprecated prints it every timem it's called. Either behavior is good IMO, but it seems like it should be consistent. There's also a switch for allowing experimental features, should we have something similar for allowing deprecated features? Maybe not?

widlarizer: Also, if we need to get `logger().` then these methods shouldn't be prefixed with `log_`. We're wrapping the with `log_` prefixed already, and that's all we need to retain for low code damage

kernel/log.h
```diff
@@ -85,162 +86,483 @@ YOSYS_NAMESPACE_BEGIN
...
+public:
+	explicit StderrLogSink(bool quiet) : quiet_warnings(quiet) {}
+	bool should_log(const LogMessage &msg) const override;
+	void log(const LogMessage &msg) override;
+	void flush() override;
+private:
+	bool quiet_warnings;
+};
+
+class StreamLogSink : public LogSink
```
widlarizer, line 165: Is the intention with `StreamLogSink` to not print "Warning:" etc prefixes and timestamps? That seems to be the current behavior

kernel/log.h
widlarizer: Prior art - performance sensitive projects with debug log macros:
[spdlog](https://github.com/gabime/spdlog/blob/f5f173a1a57d0e2e0115f2ed71ee7ea316516853/include/spdlog/spdlog.h#L311)
[LLVM](https://github.com/llvm/llvm-project/blob/d9739a01045b14e2a9fcaaeb88d3a2bb7c333be1/llvm/include/llvm/Support/Debug.h#L22)
[abseil](https://github.com/abseil/abseil-cpp/blob/1c84143fb79e6041f3e0596df51c3ae7d7e46a8a/absl/log/log.h#L228)
[tracing crate](https://github.com/tugglecore/rust-tracing-primer#events)

widlarizer (changes requested):

kernel/log.h
```diff
@@ -85,162 +86,483 @@ YOSYS_NAMESPACE_BEGIN
...
+	void set_verbose_level(int level) { log_verbose_level = level; }
+	void add_verbose_level(int level) { log_verbose_level += level; }
+	void set_expect_no_warnings(bool value) { log_expect_no_warnings = value; }
+	void set_log_time(bool value) { log_time = value; }
+	void set_cmd_error_throw(bool value) { log_cmd_error_throw = value; }
+	void set_hdump_all(bool value) { log_hdump_all = value; }
+	void set_log_forced(bool value) { log_forced = value; }
+	int get_verbose_level() const { return log_verbose_level; }
+	bool get_log_time() const { return log_time; }
+	bool get_log_forced() const { return log_forced; }
```
widlarizer: This naming is confusing to me (what are we forcing would be the interesting part?), maybe reconsider

kernel/log.cc
```diff
@@ -246,12 +271,10 @@ void log_formatted_header(RTLIL::Design *design, std::string_view format, std::s
 			if (yosys_xtrace)
 				log("#X# -- end of dump --\n");
 		}
-
-	if (pop_errfile)
-		log_files.pop_back();
+	log_forced = false;
```
widlarizer: Could be just my misunderstanding - isn't it meant to be consistently scope-based?

kernel/log.cc
```diff
@@ -201,38 +230,34 @@ static void logv_string(std::string_view format, std::string str) {
 	}
 }
 
-void log_formatted_string(std::string_view format, std::string str)
+void LogManager::log_formatted_string(std::string_view prefix, std::string_view format, std::string str, LogSeverity severity)
```
widlarizer: Passing around both format and str is super confusing and probably inefficient

examples/cxx-api/demomain.cc
mmicko: Thanks for spotting this one. Examples are not built on CI, so one of the things we need to take care of

kernel/log.h
mmicko: Made it display warning once since that makes sense, but disabling deprecated warnings I think we should not add.

kernel/log.h
mmicko: Right, fixed this one with latest commit. Was not affecting usage since it is used internally for generating help but yeah was an error.

kernel/log.h
mmicko: log_forced is ugly one agree. Thing is that current driver displays all log messages after certain point in code, mostly while displaying headers when you are using -v option on driver. It mostly ends up being used to show header and then do 
`log("Dumping current design to '%s'.\n", filename);` (since xtrace_db_dump will anyway be removed)

kernel/log.cc
mmicko: since format is determining actual warning/error idea was to use this in future to extract messages, form their ids and so on. But we can discuss how to improve on this one

kernel/log.h
mmicko: since this is something used only by driver and it affects only one sink, I am not exposing it any more in logger class, but as variable set in driver.cc directly

kernel/log.cc
mmicko: passing std::string_view format should be quite cheap, also current implementation depends on value when not complete lines are logged. Guess we should move towards logging full lines only in future. Will try to see how many of those we actually have.

mmicko:
https://github.com/YosysHQ/yosys/pull/3352 This one should be addressed by using singleton

widlarizer (approved):
Looks good!

PR #6116, opened 2026-08-13
Fix ignored `hdlname` attribute in `dfflibmap` pass

magancarz:
This pull request fixes omitted `hdlname` attribute when running `dfflibmap` pass. In current implementation only `src` attribute was moved to new cells from the old ones. I've added a regression test ensuring all desired attributes to be present after the `dfflibmap` pass.

Is there a reason behind only `src` attribute being copied in the `dfflibmap` pass or should all attributes be passed to the newly created cell?


widlarizer (changes requested):
Sorry about that. Yes, I think all attributes should be copied there

magancarz:
I've changed the fix so now all attributes are moved and added a custom attribute in `dfflibmap_attributes.ys` to ensure all attributes are handled. Thanks for a quick response!

widlarizer (approved):

PR #6118, opened 2026-08-14
liberty_cache: key SCL cache on ABC executable mtime and size

Emin017:
The SCL format is written/read by ABC and can change between ABC builds, so a cache entry created with a different ABC binary must not be reused.

### DIFF
--- a/passes/techmap/liberty_cache.h
+++ b/passes/techmap/liberty_cache.h
@@ -37,6 +37,14 @@ inline std::string convert_liberty_files_to_merged_scl(const std::vector<std::st
 	std::string hash_input;
 	time_t newest_mtime = 0;
 
+	// The SCL format is written/read by ABC and can change between ABC builds,
+	// so the cache key must identify the ABC executable.
+	if (!abc_exe.empty()) {
+		struct stat abc_stat;
+		if (stat(abc_exe.c_str(), &abc_stat) == 0)
+			hash_input += stringf("|abc:%s:%lld:%lld", abc_exe.c_str(), (long long)abc_stat.st_mtime, (long long)abc_stat.st_size);
+	}
+
 	for (const std::string &liberty_file : sorted_files) {
 		struct stat liberty_stat;
 		if (stat(liberty_file.c_str(), &liberty_stat) != 0) {


Emin017:
## Reproduction

  Same yosys build, two different `yosys-abc` binaries (taken from the nixpkgs
  yosys 0.55 and 0.67 packages, 19663360 vs 28330336 bytes), using
  `tests/liberty/normal.lib` and a trivial AND-gate design:

```shell
yosys -p 'read_verilog test.v; proc; abc -liberty test.lib -exe <abc-from-yosys-0.55>'
yosys -p 'read_verilog test.v; proc; abc -liberty test.lib -exe <abc-from-yosys-0.67>'
```

  Before this change, the second run prints:

```log
ABC: using cached merged SCL: /tmp/yosys-liberty-scl-cache/yosys_merged_c8c3c4ea.scl (1 files)
```

  i.e. it reuses the SCL file written by the 0.55 ABC build with the 0.67 ABC
  binary, which will cause ABC error:

```log
4.1.1. Executed ABC.
Extracted 24 gates and 48 wires to a netlist network with 24 inputs and 8 outputs.
Running ABC script: <abc-temp-dir>/abc.script
ABC: ======== ABC command line "source <abc-temp-dir>/abc.script"
ABC: + read_blif <abc-temp-dir>/input.blif
ABC: + read_scl /tmp/yosys-liberty-scl-cache/yosys_merged_c8c3c4ea.scl
ABC: Reading SCL library from file "/tmp/yosys-liberty-scl-cache/yosys_merged_c8c3c4ea.scl" has failed.
ABC: ** cmd error: aborting 'source <abc-temp-dir>/abc.script'
ABC: Error: Wrong version of the SCL file.
ERROR: Can't open ABC output file `/tmp/yosys-abc-oxctwl/output.blif'.
```

  After this change, the second run produces a different cache file
  (`yosys_merged_b51b554a.scl`) and reconverts; repeating the same ABC again
  hits the cache as expected.

  ## Fix

  Include the ABC executable's path, mtime and size in the cache hash. Path and
  size matter in addition to mtime because reproducible-build environments
  (nix store) normalize mtimes to the epoch.

hexratcc (approved):
Seems reasonable.

PR #6123, opened 2026-08-16
Fix resetall not affecting default_nettype

sifferman:
Hello! Thank you for this tool.

Currently, `` `resetall `` fails to reset `` `default_nettype ``. The preprocessor consumes `` `resetall `` before the lexer ever sees it, so the existing solution (`default_nettype_wire = true`) never reaches the lexer.

Now on `` `resetall ``, the preprocessor emits `` `default_nettype wire `` into its output stream, so the lexer's own directive handler performs the reset. The dead preproc-side assignment and the now-unused `parse_state` parameter are removed.

Thanks!


widlarizer (approved):
Looks good, thanks

sifferman:
Thanks!

PR #6124, opened 2026-08-17
scl_cache: add pass to control merged SCL caching for ABC [sc-771]

hexratcc:
The SCL liberty cache runs unconditionally, while `libcache` is opt-in, with no way to turn the SCL cache off (#6118). This PR adds SCL cache toggles via: `libcache -scl {-enable|-disable}`, and `libcache -list` now also lists SCL caching state.

Also fixes `libcache -disable` acting as `-enable`.


Ravenslofty:
I'm not sure how intuitive it is for this to be a separate pass, rather than an option to the already existing `libcache`.

hexratcc:
@Ravenslofty Moved it and updated the PR description.

Ravenslofty (approved):

PR #6126, opened 2026-08-17
synth_coolrunner2: remove

Ravenslofty:
Sort-of-related to the ABC unification work.

Per [this comment](https://github.com/YosysHQ/yosys/pull/6103#issuecomment-5316971530) the author of `synth_coolrunner2` is okay with the pass being removed.

This pass is an awkward edge case because it's the one flow using `abc -sop`, which has no XAIGER equivalent. Originally the plan was to keep a minimal version of the `abc` pass around to support this, but with @ArcaneNibble's approval, I will be able to tear out the whole `abc` pass instead.

Yeet.


hexratcc (approved):
lgtm but I think we should also remove https://github.com/YosysHQ/yosys/blob/621d943ac87adc932bdabb53f4f51af8433611f0/docs/source/cmd/index_techlibs_coolrunner2.rst

hexratcc (approved):

PR #6127, opened 2026-08-17
abc: Add file lock on liberty scl cache to support parallelism

mikesinouye:
When launching Yosys with the same liberty libraries in multiple threads/processes in a wrapper application, this file has a race condition upon which multiple threads attempt to build the cache simultaneously. The file then becomes unreadable by abc.

For Linux builds, use flock() to ensure that the first creator of the cache is allowed to finish before other threads can read it. Also verify that none of the liberty files have changed since the cache was last created.

This necessitates a wrapper flow of Yosys to exercise, and it will likely take 50+ concurrent invocations to reliably trigger the issue.

### DIFF
--- a/passes/techmap/liberty_cache.h
+++ b/passes/techmap/liberty_cache.h
@@ -11,6 +11,8 @@ namespace abc {
 
 #if !defined(_WIN32)
 #include <sys/file.h>
+#include <fcntl.h>
+#include <unistd.h>
 #endif
 
 YOSYS_NAMESPACE_BEGIN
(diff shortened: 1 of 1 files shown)


mikesinouye:
Hi @hexratcc and @widlarizer , please take a look when you get a chance. If this specific solution is not viable we would need a path forward to avoid the race on this file (outside of skipping cache generation).

hexratcc (approved):
lgtm, verified with 80 concurrent instances of Yosys. I'll merge this as-is and create a PR  later with some cleanups + a similar guard for Windows.

hexratcc:
Moved this work over to #6129.

PR #6131, opened 2026-08-18
arith_tree: add regression tests for links narrower than their consumer (follow-up to #6130)

DanielMBouyou:
**What are the reasons/motivation for this change?**

Follow-up to #6130 (merged as 59a3826), which stops `arith_tree` from folding a chain link that is narrower than the cell consuming it. This PR adds tests only, there is no source change. #6130 shipped with one equivalence test; the three blocks below pin down the structure the pass must now produce, and check that it excludes nothing more than the truncating link.

**Explain how this is achieved.**

`tests/arith_tree/arith_tree_edge_cases.ys`, two new blocks:

1. `trunc_narrow_no_fold`: `wire [7:0] t = a + b; assign y = t + c;` with `y` 9 bits wide, built from raw `$add` cells. After `arith_tree`, no `$fa` cell may exist and both `$add` cells must remain (`select -assert-count 2 t:$add`).
2. `trunc_narrow_no_fold_alu`: the same module after `alumacc`, so the chain is made of `$alu` cells, with the same expectations on `$alu`.

`tests/arith_tree/arith_tree_equiv.ys`, one new block:

3. `equiv_trunc_mixed`: the truncated `t` enters the wider chain `y = t + c + d` as a leaf. `equiv_opt -assert arith_tree` must hold, and the wide part must still be folded (`select -assert-min 1 t:$fa` on the post-opt design). Only the truncating link is left out, the rest of the chain keeps its tree.

**If applicable, please suggest to reviewers how they can test the change.**

`yosys -q tests/arith_tree/arith_tree_edge_cases.ys` and `yosys -q tests/arith_tree/arith_tree_equiv.ys`.

### DIFF
--- a/tests/arith_tree/arith_tree_edge_cases.ys
+++ b/tests/arith_tree/arith_tree_edge_cases.ys
@@ -443,3 +443,41 @@ opt_clean
 select -assert-none t:$fa
 select -assert-min 1 t:$macc t:$macc_v2 %u
 design -reset
+
+# A link narrower than its consumer truncates its partial sum; folding it into
+# the wider tree would resurrect the discarded carry. Such a link must be left
+# out of the chain, both from raw $add cells and from alumacc $alu cells.
+read_verilog <<EOT
+module trunc_narrow_no_fold(
+	input [7:0] a, b, c,
+	output [8:0] y
+);
+	wire [7:0] t = a + b;
+	assign y = t + c;
+endmodule
+EOT
+hierarchy -auto-top
+proc
+arith_tree
+select -assert-none t:$fa
+select -assert-count 2 t:$add
+design -reset
+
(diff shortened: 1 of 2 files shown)


hexratcc:
Doesn't this solve the same issue as #6130? I think it would be enough to add the tests, no?

DanielMBouyou:
Right — #6130 landed the fix, so I've reduced this PR to the remaining regression tests, rebased on main: structural no-fold checks on the $add and $alu paths, plus a mixed case where the wide part of the chain still folds around the truncated link. Thanks for the review and the merge!

hexratcc (approved):
I'll merge this, as it covers a part of `arith_tree` that isn't well tested, even though it looks LLM generated/assisted, please disclose this fact next time.

PR #6132, opened 2026-08-18
synth_greenpak4, nlutmap: remove

Ravenslofty:

More ABC unification work.

Per [this comment](https://github.com/YosysHQ/yosys/pull/6103#issuecomment-5329170455) the code owner (or "point of contact", perhaps) for `synth_greenpak4` is okay with the pass being removed.

`synth_greenpak4` has `nlutmap` which iteratively maps logic to smaller LUTs to deal with the extremely limited resources of the chip, but `nlutmap` depends on behaviour of `abc -lut` which cannot easily be adapted to `abc9`. Removing both `nlutmap` and `synth_greenpak4` is the easiest of resolving this issue and unblocking #6103.

Yeet.


hexratcc (approved):
lgtm, but I think we should remove the remaining references in:
- https://github.com/YosysHQ/yosys/blob/9eb62484dc79e1fa0615bce2e00b023a18828686/docs/source/cmd/index_techlibs_greenpak4.rst#L3
- https://github.com/YosysHQ/yosys/blob/9eb62484dc79e1fa0615bce2e00b023a18828686/passes/opt/rmports.cc#L31

Ravenslofty:
Hmm, nothing seems to use `rmports` either, but I'll leave it be.

hexratcc (approved):

PR #6134, opened 2026-08-19
Retry read operations interrupted by EINTR

SimonEbner:
Interrupted system calls (`EINTR`) during read operations were previously
treated as fatal errors. Add retry logic to continue reading when interrupted.


QuantamHD:
@widlarizer Could you review this PR. It's a bug we're running into at Google

mmicko (approved):
LGTM, it is proper way to handle specific situation, and should be fine for all platforms, but CI will tell.

PR #6138, opened 2026-08-20
xilinx_dsp: do not absorb nonzero sync-load constants into RSTM/RSTP

0xa000:
*Authorship notice: investigation and patch developed with substantial AI assistance (Claude), reviewed and hardware-verified by me.*

**Symptom.** `synth_xilinx` silently miscompiles a register with synchronous-load-of-a-nonzero-constant semantics feeding from a multiply, e.g. `if (cond) q <= a*b; else q <= 50;`. The netlist reads 0 instead of 50 whenever the load branch is active. RTL sim is correct, gate-level sim and hardware are wrong. Found on real hardware: a MEGA65 core port (xc7a100t) where a terminal renderer's row-advance constant vanished, visibly shifting all on-screen text by one row.

**Root cause.** `opt_dff` turns the load-of-constant into an `$sdff` with `SRST_VALUE=50`. In `techlibs/xilinx/xilinx_dsp.pmg`, the `in_dffe` subpattern guards against this (`filter ff->type.in($dff, $dffe) || param(ff, \SRST_VALUE).is_fully_zero()`), but the `out_dffe` subpattern — MREG/PREG absorption, whose own header comment says the reset mux's "other input is fully zero" — matches `$sdff`/`$sdffe` with no `SRST_VALUE` check. The register is absorbed into MREG and its srst line wired to RSTM, which resets to zero; the nonzero constant is dropped from the design. `xilinx_dsp48a.pmg` has the same omission in its `out_dffe`; `microchip_dsp.pmg` checks `srstZero` correctly. abc9 is not involved (reproduces with and without).

**Fix.** Add the same `is_fully_zero()` filter to both `out_dffe` subpatterns (`xilinx_dsp.pmg`, `xilinx_dsp48a.pmg`). The multiply is still packed into the DSP; the register simply stays in fabric when its reset value is nonzero.

### DIFF
--- a/techlibs/microchip/microchip_dsp.pmg
+++ b/techlibs/microchip/microchip_dsp.pmg
@@ -408,6 +408,11 @@ match ff
 	// does not support clock inversion
 	select param(ff, \CLK_POLARITY).as_bool()
 
+	// Check that reset value, if present, is fully 0 (the DSP's output
+	// register resets to zero; absorbing a nonzero sync-load constant
+	// would corrupt it)
+	filter ff->type.in($dff, $dffe) || param(ff, \SRST_VALUE).is_fully_zero()
+
 	slice offset GetSize(port(ff, \D))
 	index <SigBit> port(ff, \D)[offset] === argD[0]
 
--- a/techlibs/xilinx/xilinx_dsp.pmg
+++ b/techlibs/xilinx/xilinx_dsp.pmg
@@ -488,6 +488,10 @@ match ff
 	// DSP48E1 does not support clock inversion
 	select param(ff, \CLK_POLARITY).as_bool()
 
+	// Check that reset value, if present, is fully 0 (RSTM/RSTP reset to
+	// zero; absorbing a nonzero sync-load constant would corrupt it)
+	filter ff->type.in($dff, $dffe) || param(ff, \SRST_VALUE).is_fully_zero()
+
 	slice offset GetSize(port(ff, \D))
 	index <SigBit> port(ff, \D)[offset] === argD[0]
 
--- a/techlibs/xilinx/xilinx_dsp48a.pmg
+++ b/techlibs/xilinx/xilinx_dsp48a.pmg
@@ -397,6 +397,11 @@ match ff
 	// DSP48E1 does not support clock inversion
 	select param(ff, \CLK_POLARITY).as_bool()
 
+	// Check that reset value, if present, is fully 0 (the DSP's output
+	// registers reset to zero; absorbing a nonzero sync-load constant
+	// would corrupt it)
+	filter ff->type.in($dff, $dffe) || param(ff, \SRST_VALUE).is_fully_zero()
+
 	slice offset GetSize(port(ff, \D))
 	index <SigBit> port(ff, \D)[offset] === argD[0]
 
--- a/tests/arch/microchip/dsp_nonzero_srst.ys
+++ b/tests/arch/microchip/dsp_nonzero_srst.ys
@@ -1,0 +1,34 @@
+# The MACC_PA output register resets to zero, so a register with a
+# synchronous load of a nonzero constant must not be absorbed into the DSP.
+# (No equivalence check here: cells_sim.v has no behavioural MACC_PA model.)
+# https://github.com/YosysHQ/yosys/pull/6138
+
+read_verilog <<EOT
+module top(input clk, cond, input signed [8:0] a, b, output reg signed [17:0] q);
+always @(posedge clk)
+	if (cond)
+		q <= a * b;
+	else
+		q <= 50;
+endmodule
+EOT
+synth_microchip -top top -family polarfire -noiopad
+select -assert-count 1 t:MACC_PA
+# the register must stay in fabric
+select -assert-min 1 t:SLE
+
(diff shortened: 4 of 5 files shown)


hexratcc:
Would you please provide steps to reproduce this issue? This seems like a real issue but I don't feel like deciphering an LLM-written PR.

0xa000:
Sure, see below. This repro only needs yosys and iverilog. It uses `xilinx/cells_sim.v` as the DSP48E1 model (no vendor files).

`dspbug.v`; A multiply whose output register has a synchronous load of a *nonzero* constant (50) in the else-branch:

```verilog
module dspbug(input clk, input cond, input advance, input [3:0] nrows,
              output reg [11:0] lsa = 0);
  reg [4:0] skip_rows = 0;
  reg [10:0] skip_bytes = 0;
  always @(posedge clk) begin
    if (cond) begin
      skip_rows <= 1 + nrows;
      skip_bytes <= skip_rows * 50;
    end else begin
      skip_rows <= 1;
      skip_bytes <= 50;      // nonzero constant load
    end
    if (advance) lsa <= lsa + skip_bytes;
  end
endmodule
```

`tb_dspbug.v`; holds `cond=0` (so `skip_bytes` should be 50) and pulses `advance` periodically, so `lsa` should grow by 50 per pulse.

```verilog
`timescale 1ps/1ps
module tb_dspbug;
  reg clk=0; reg advance=0;
  wire [11:0] lsa;
  dspbug dut(.clk(clk), .cond(1'b0), .advance(advance), .nrows(4'd0), .lsa(lsa));
  always #5000 clk=~clk;
  integer c=0;
  always @(posedge clk) begin
    c=c+1;
    advance <= (c%20==0);
    if (c%20==2 && c>100) $display("advance %0d: lsa=%0d", c/20, lsa);
    if (c==200) $finish;
  end
endmodule
```

Steps:

```sh
# behavioral reference
iverilog -g2005 -o beh.vvp tb_dspbug.v dspbug.v && vvp beh.vvp

# synthesize and simulate the netlist against yosys' own DSP48E1 model
yosys -p "read_verilog dspbug.v; synth_xilinx -arch xc7 -top dspbug; write_verilog dspbug_synth.v"
iverilog -g2005 -DUNISIM -o syn.vvp tb_dspbug.v dspbug_synth.v $(yosys-config --datdir)/xilinx/cells_sim.v && vvp syn.vvp
```

Behavioral output:

```
advance 5: lsa=250
advance 6: lsa=300
advance 7: lsa=350
```

Synthesized netlist output:

```
advance 5: lsa=0
advance 6: lsa=0
advance 7: lsa=0
```

What happens: `opt_dff` turns the load-of-50 into an `$sdff` with `SRST_VALUE=50`. `xilinx_dsp` then packs the multiply into a DSP48E1, absorbs the output register into MREG, and wires that `$sdff`'s reset line to RSTM — but RSTM resets M to zero, not to 50, so the constant disappears from the netlist. The `in_dffe` subpattern in `xilinx_dsp.pmg` already guards against this with `param(ff, \SRST_VALUE).is_fully_zero()`; `out_dffe` lacks the same check (as does `xilinx_dsp48a.pmg`; `microchip_dsp.pmg` checks correctly). The PR adds that filter to both `out_dffe` subpatterns, so the register stays outside the DSP as fabric FFs when the reset value is nonzero.

Reproduced on 0.66 and on current main as of today (47bbdab29, Yosys 0.68+122). With the PR applied the netlist matches the behavioral sim and the multiply still infers a DSP48E1. abc9 makes no difference (reproduces with and without).

hexratcc (changes requested):
Reproduced locally, and the fix looks correct, but please add a regression test for this behavior. Secondly: 
> microchip_dsp.pmg checks srstZero correctly

I think this is incorrect - it checks `srstZero` only in `in_dffe`, but https://github.com/YosysHQ/yosys/blob/dd83bbad28b61f244e1ae00639d85a07c2a5c95f/techlibs/microchip/microchip_dsp.pmg#L407 does not - would you mind adding this fix as well?

0xa000:
Both done, thanks for the catch on `microchip_dsp.pmg`.

Regression tests added for both architectures:

- `tests/arch/xilinx/dsp_nonzero_srst.ys` — a multiply whose output register has a synchronous load of a nonzero constant (50). Checks `equiv_opt -assert` against `cells_sim.v` and that the register stays in fabric, for both DSP48E1 and DSP48A1 (`-family xc6s`), plus a zero-constant control case that must still be absorbed into PREG. Verified it catches the bug: with the `xilinx_dsp`/`xilinx_dsp48a` fix reverted, `equiv_opt` reports 3 unproven `$equiv` cells.
- `tests/arch/microchip/dsp_nonzero_srst.ys` — same design through `synth_microchip -family polarfire`; structural checks only (`cells_sim.v` has no behavioural MACC_PA model): the multiply still packs into MACC_PA, the register stays in fabric as SLE cells, and the zero-constant control is absorbed (no SLEs). Verified it catches the bug: against a build without the `microchip_dsp` fix, the SLE assertion fails.

hexratcc:
As this PR is LLM generated and our [LLM policy](https://blog.yosyshq.com/p/interim-yosyshq-llm-policy/) is still under discussion I'll hold off on this until that is finalized.

PR #6140, opened 2026-08-21
verilog: fix `**` associativity to right per IEEE 1364-2005 / 1800-2017

john-kearney:
The power operator `**` is right-associative per IEEE 1364-2005 section 5.1 and IEEE 1800-2017 section 11.4.1. The grammar declared it `%left`, so unparenthesized chains were parsed left-associatively and silently produced wrong values:

```verilog
module m(output [15:0] y);
    assign y = 2**3**2;
endmodule
```

Parsed as `(2**3)**2` = 64. Should be `2**(3**2)` = 512 per the LRM. No warning, no error; the wrong constant propagates through ordinary optimization.

The fix is the declaration change (`%left` to `%right` in `frontends/verilog/verilog_parser.y`). The rest of the precedence table matches the standard; this was the only operator with the wrong associativity (every other binary operator is left-associative, and the ternary is handled structurally).

The regression test (`tests/opt/pow_assoc.ys`) uses `equiv_opt`-style checking against hardcoded expected constants: the unparenthesized form and the explicit right-associative form must both fold to 512, and the explicit left-associative form to 64. Verified it fails on the unpatched grammar and passes with the fix.

### DIFF
--- a/tests/opt/pow_assoc.ys
+++ b/tests/opt/pow_assoc.ys
@@ -1,6 +1,4 @@
-# The power operator `**` is right-associative per IEEE 1364-2005 sec 5.1 /
-# 1800-2017 sec 11.4.1: `2**3**2` is `2**(3**2)` = 512. A `%left` declaration
-# in the grammar parsed it as `(2**3)**2` = 64.
+# `**` is right-associative: 2**3**2 is 2**(3**2) = 512.
 
 read_verilog << EOT
 module top(output [15:0] unparen, output [15:0] right, output [15:0] left);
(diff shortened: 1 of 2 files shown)


tests/opt/pow_assoc.ys
```diff
@@ -0,0 +1,35 @@
+# The power operator `**` is right-associative per IEEE 1364-2005 sec 5.1 /
+# 1800-2017 sec 11.4.1: `2**3**2` is `2**(3**2)` = 512. A `%left` declaration
```
widlarizer: No need to declare removed incorrect behavior that no longer exists in a test that is meant to pin the correct behavior. Also no need to describe the cause in the test down to the exact feature of the parser generator. It can be pretty distracting from what the test is testing and how the test is implemented

widlarizer (changes requested):
Parser change seems fine, simplify or remove the test comment and it's good to go

widlarizer (approved):

PR #6142, opened 2026-08-21
chores: Fix mixed spaces vs tabs

mmicko:
Majority of files is using tabs for indent, but made a script to leave out those that are cleanly using spaces only.

Change is equivalent to use of  VS Code "Convert indentation to Tabs" command.

Let's discuss this on DevJF.

There are small number of files that is using spaces instead of tabs, that could be converted, but would like to check with the rest.

```
backends/functional/cxx_runtime/sim.h
backends/jny/jny.cc
docs/source/code_examples/extensions/my_cmd.cc
frontends/aiger/aigerparse.h
kernel/compute_graph.h
kernel/json.cc
kernel/json.h
kernel/sexpr.cc
kernel/topo_scc.h
passes/sat/recover_names.cc
techlibs/achronix/synth_achronix.cc
techlibs/gatemate/gatemate_foldinv.cc
techlibs/quicklogic/ql_dsp_macc.cc
tests/cxxrtl/test_value.cc
tests/fmt/fuzz/x_test.cc
tests/unit/kernel/modindexTest.cc
```

Note: Some files had already wrong formatting of text, fixed those manually where noticed (by removing/adding tabs)


whitequark:
No problem with me.

KrystalDelusion:
It looks like you're using 4 width tabs instead of 8?  IIRC we agreed to standardise on 8 for the `.cc` files.  I also think there are enough files being touched here that it's probably better to just convert the ones that are exclusively space only to also use tabs.

widlarizer (changes requested):
Yes, 8 character tabs are specified in the contributing guidelines, but with full knowledge of that I kept my editor at 4 because I prefer that. I expect others have done the same. Looking at this diff, some of the lines being changed were actually done for compatibility with that fact. A lot of files had used spaces for arbitrary indentation in multi line expressions. This PR doesn't fit the contributing guidelines by assuming tabs are 4 spaces wide, and redoing it to assume 8 would create misaligned indentation on multiline expressions in my editor if I don't set my editor to 8, which I can accept, but I'd prefer keeping those space-based. There's also lines here that are correct indentation fixes regardless, though.

PR #6143, opened 2026-08-22
proc_mux: avoid repeated work in mux tree construction

apullin:
## Summary

- skip process actions whose left-hand-side wires cannot affect the output snippet currently being lowered
- cache parallel-case grouping once per switch instead of rebuilding it for every output snippet
- avoid `BitPatternPool` argument copies in the grouping hot path
- retain `proc_mux -no-wire-prefilter` as a legacy reference path and add direct and SAT-equivalence regressions

These changes avoid work and cache invariant analysis; they do not change mux-selection policy or output quality.

## Correctness

- release build with bundled libraries passes
- all 17 `tests/proc/*.ys` scripts pass
- `BitpatternTest.*` passes
- optimized and legacy `proc_mux` paths produce byte-identical RTLIL on the differential workloads
- current `main` and this branch produce identical post-`proc_mux` RTLIL, excluding only the generated version header, on Hazard3, OpenSPARC, WF68K, PicoRV32, UDP, and an AXI-lite crossbar
- final Hazard3 synthesis JSON canonicalizes identically after removing creator/build-path metadata

## Complete-flow performance

Paired runs pin both revisions sequentially to the same physical core.

| Workload | Pairs | `main` median | Branch median | Result |
| --- | ---: | ---: | ---: | --- |
| Public Hazard3 Icebreaker | 8 | 32.560 s | 13.826 s | 57.524% less wall time; 2.354x throughput |
| OpenSPARC T2 | 10 | 724.731 s | 721.961 s | +0.633%; 95% CI -1.363% to +2.697% |
| WF68K | 10 | 1537.206 s | 1532.606 s | +0.262%; 95% CI -0.493% to +0.976% |

On public `yosys-bench`, 20 pairs per design produced an equal-weight geomean of +0.272% (95% CI -0.038% to +0.493%). UDP was individually positive at +0.553% (95% CI +0.190% to +0.707%); PicoRV32 and the AXI-lite crossbar were inconclusive.

The large Hazard3 effect is workload-specific: its process tree contains substantially more affected action/case work. On the regular large designs, the isolated pass falls from 2.457 s to 0.300 s on OpenSPARC and from 3.295 s to 0.326 s on WF68K, so the savings are real but diluted inside 12- and 26-minute synthesis flows.

Hazard3 was tested from the public `Wren6991/Hazard3` repository at `c15bffd3e5e80d7b6eb1f98a735f63c8a68da149`.

This is a draft while the change and benchmark presentation receive review.

### DIFF
--- a/kernel/bitpattern.h
+++ b/kernel/bitpattern.h
@@ -97,7 +97,7 @@ struct BitPatternPool
 	 * Convert a constant SigSpec to a pattern. Normalize Yosys many-valued
 	 * to three-valued logic.
 	 */
-	bits_t sig2bits(RTLIL::SigSpec sig)
+	bits_t sig2bits(const RTLIL::SigSpec &sig)
 	{
 		bits_t bits;
 		bits.bitdata = sig.as_const().to_bits();
@@ -111,7 +111,7 @@ struct BitPatternPool
 	 * A literal x/z bit can never match a 2-valued selector, so a pattern containing
 	 * one covers nothing.
 	 */
-	static bool covers_nothing(RTLIL::SigSpec sig)
+	static bool covers_nothing(const RTLIL::SigSpec &sig)
 	{
 		for (auto bit : sig)
 			if (bit.wire == NULL && (bit.data == RTLIL::State::Sx || bit.data == RTLIL::State::Sz))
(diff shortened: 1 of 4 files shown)


whitequark:
This PR looks AI-generated to me. Is that so?

apullin:
> This PR looks AI-generated to me. Is that so?

@whitequark Yes, largely. I actually thought drafts were only visible to me before I made them ready. Oops. Cleanup of the summary is intended.

The origin of this was working on a design with the Hazard core, where a lot of the speedup was found.

PR #6145, opened 2026-08-22
hdlname_recover (withdrawn)

colneech-dev:
Withdrawn — not ready for review.

This was an experiment in mapping synthesised cells back to their RTL hierarchy. The approach needs more work before it is worth reviewing.


whitequark:
This PR appears to be AI-generated, as such it won't be merged.

PR #6146, opened 2026-08-24
abc: remove -lut

Ravenslofty:
More ABC unification.

After #6103 and #6137, nothing inside Yosys uses `abc -lut` any more; the functionality is replaced by `abc9 -lut`, so this can be removed.

Yeet.


widlarizer (changes requested):
Until interface instability is explicit, let's just mark things like this deprecated, either in comments, or in user-facing messages with an implementation of #4623

Ravenslofty:
I think instead of removing this option, I'll turn the `abc` pass into something like `synth_ecp5`; a pass which delegates to other passes.

PR #6148, opened 2026-08-24
twines: rich interned strings

widlarizer:
See the Twine section of the [Discourse post](https://yosyshq.discourse.group/t/consistent-source-location-transfer-with-patch/121), and [LLVM Twine docs](https://llvm.org/doxygen/classllvm_1_1Twine.html#details) since this work is partly inspired by those.

# Code

A twine is either a Leaf node, carrying a string, or a Suffix node, carrying a suffix string and referring to a Twine to which it is appended. This deduplicates shared prefixes. See `kernel/twine.h`.

An `IdString` is a publicity-tagged index of a `TwineNode` within a per-`Design` `TwinePool`. Garbage collection collects dead twines, and a `TwinePool` releases memory as it's backed by `std::deque` for simplicity. `kernel/constids.inc` is supported with the same approach as before. Parameter and attribute keys are only ever Leaf nodes. Masquerade types are provided in `kernel/rtlil_twine_compat.h` to reduce the code damage with dark C++ magic. Twines hash and compare equal cheaply based on their structure. Utilities to hash, sort, and compare IdStrings based on the strings they represent is provided, it's just not the default behavior as that could get computationally expensive in hot loops and is not really needed.

# Merit

`flatten` had been reported as an expensive pass, dominating runtime and memory usage in both ASIC and FPGA use cases on some designs. On a synthetic benchmark, this work speeds up flatten by 50% and reduces its memory consumption by 30%. It also improves the parallelism potential of Yosys by replacing global string interning with per-Design pools. Name publicity checks are cheaper now.

But the real point is to make source locations cheaper to keep around later (not in this PR), so that we can keep more of them throughout opt and synth.

# Costs

The IdString is still 4 bytes in the default build and the same comptime dispatch works the same way, so performance isn't a problem. But there's a lot of required code changes involved. It breaks down something like this:

- new `twines` block in textual RTLIL is required to keep RTLIL replayable
- `Design*` has to be passed around for things like minting or printing an `IdString`, or using `PooledName` instead of `IdString` which packs the `Design*` alongside an `IdString`
- `std::string -> IdString` lookup is expensive, and equality comparison of two IdStrings that resolve to the same string can compare as not equal due to different twine structure. I believe this is acceptable for anything Yosys *should* do and assume, and I've resolved all of the impact uncovered by tests. Symbol files for abc don't use strings but IdString raw integers as a result, for example
- a lot of pre-existing code can stay unchanged through redirecting Masquerade types (`kernel/rtlil_twine_compat.h`)
- non-standard but functional use of `offsetof` and `no_unique_address` / `msvc::no_unique_address`
- annoying macros and redundant interfaces to keep the diffs small in the affected files - these can be reduced and deprecated later (`concept IsNameMasq`, `is_unpooled_name_v`)
- tricky to understand performance considerations - transparent hashing, or that `TwineNode` and `TwineSpec` are different types due to memory usage and cache locality vs convenience and safety tradeoffs
- plugins need code changes to live with this, but it's easy to explain what changes should be made in which situation
- passes that work with multiple `Design`s can have subtle bugs, but a provenance tagged 8-byte `IdString` build can be enabled with `YOSYS_ENABLE_TWINE_PROVENANCE` .
- pass changes may be unnecessarily verbose as the build of the easy-to-fix passes was fixed mechanically with a lot of copypasting

Outside of known improvements, performance seems to be at parity. I don't foresee showstopper performance regressions for any use case of Yosys as a result of this.

# About this PR

Because of the scale of the change, the description may be incomplete, and I'm open to discussing any part on this PR or on another communications channel. The patch hunks are categorized into commits by interestingness and affected directory, though this categorization could be imperfect, and I have touched the core of the change in additional commits, including adding or fixing comments.

# TODOs

- [ ] Is this commit and file structure even reviewable?
- [ ] Is this blocked by an explicit stability contract due to the scale of the changes required and the likelihood of further changes to the interfaces involved here?
- [ ] Should the leafness of keys be conveyed via some clever typing rather than asserts?
- [ ] Publish slang submodule changes
- [ ] Write internal docs about the interface including migration notes once we agree on the interface used

# Next steps

Source locations will be reworked with a new data structure implementing sets of source IdStrings, replacing the src attribute within Yosys, and more

cc @rocallahan due to prior work on strings and GC


passes/techmap/abc_new.cc
```diff
@@ -191,11 +192,11 @@ struct AbcNewPass : public ScriptPass {
 				}
 
 				run(stringf("  abc9_ops -write_box %s/input.box", tmpdir));
+
```
Ravenslofty, line 195: stray whitespace.

techlibs/coolrunner2/coolrunner2_fixup.cc
```diff
@@ -0,0 +1,525 @@
...
+ *
+ */
+
+#include "kernel/yosys.h"
+#include "kernel/sigtools.h"
+
+USING_YOSYS_NAMESPACE
+PRIVATE_NAMESPACE_BEGIN
+
+static IdString uniq(RTLIL::Module *module, std::string name)
```
Ravenslofty, line 26: I nuked `synth_coolrunner2`, so this folder can be deleted.

techlibs/greenpak4/greenpak4_dffinv.cc
```diff
@@ -0,0 +1,210 @@
...
+		cell->unsetPort(ID(nQ));
+		cell_type_i = false;
+	} else {
+		cell->setPort(ID(nQ), cell->getPort(ID::Q));
+		cell->unsetPort(ID::Q);
+		cell_type_i = true;
+	}
+
+	if(cell_type_latch)
+		cell->type = cell->module->design->twines.add(std::string{stringf("\\GP_DLATCH%s%s%s", cell_type_s ? "S" : "", cell_type_r ? "R" : "", cell_type_i ? "I" : "")});
```
Ravenslofty, line 84: I nuked `synth_greenpak4`, so this folder can be removed.

Ravenslofty (commented):
honestly I kind of glazed over while reading through the code but I tried my best. it seems okay in the bits I have code ownership over, at least.

widlarizer:
> glazed over while reading through the code

Sorry, this comes with the territory. Last time I'd done a full sweep of every single line of the diff, it took about 8 hours and I was only looking for leftovers from unrelated work, rather than logic errors or such

> the bits I have code ownership over

I think this is what I can realistically ask in review of work like this - check the core idea, maybe its implementation too, and its application in code you own or have recently worked with

whitequark:
My main conclusion (and I realize it's not very actionable in its current form, which is why I debated posting it at all) is that the new API seems... worse. There are some very common operations which require you to drag some object out of `desgin`, and I can't help but think that some of this could be simplified. But I don't have concrete suggestions, mainly because that'd require me to review all 8k lines.

widlarizer:
I think that IdString has always been an interned string and interning a string is something that the API shouldn't aim to make easy to do thoughtlessly. It had been done thoughtlessly in the past and especially with reference counting it lead to poor performance.

As for the API, with the Masq types you can still print like this since a cell knows what design it belongs to:

```cpp
log_debug("  removing unused `%s' cell `%s'.\n", cell->type, cell->name);
```

But not this:

```cpp
void foo(IdString cell_type, IdString cell_name) {
    log_debug("  removing unused `%s' cell `%s'.\n", cell_type, cell_name);
}
```

You can still explicitly haul the `Design*` around with `PooledName`:

```cpp
void foo(PooledName cell_type, PooledName, cell_name) {
    log_debug("  removing unused `%s' cell `%s'.\n", cell_type, cell_name);
}
```

I don't really have any extra convenience to add for interning a string. You really do need a `Design*` or `TwinePool*` for that. That's unavoidable:

```cpp
IdString cell_type = design->twines.add(RTLIL::escape_id(args[++argidx]));
```

Current global state isn't a better solution in my book. I think I could symetrically extend the Masq types to support an std::string being passed in. Also it can be convenient to have a redirect from `cell->twines.add(` to avoid the verbosity of `cell->module->design->twines.add(`.

Ultimately, I think that every single callsite perhaps seemingly arbitrarily affected by this is a callsite worth revisiting and reconsidering. Accidental thrashing in your string pool on a hot path is no good. So is constructing leaf nodes from stringf-produced strings when a suffix node would deduplicate a common prefix. It does feel like trying to unmix a soup to rework how strings work in a string-abuse-based codebase like Yosys, so there's going to be churn, but that's just what it takes, as far as I can tell

whitequark:
> Also it can be convenient to have a redirect from `cell->twines.add(` to avoid the verbosity of `cell->module->design->twines.add(`.

That's one of the things I had in mind, yeah.

mmicko:
Just few observations for now. 
- Does not pass compiling when Verific is enabled, this one affects hierarchy and frontend/verific sources additionally, but guess it will not take much time to do it
- Tried to compare logs differences to see if there are any regressions and noticed that some names are formed differently now, note that $proc is added in different place. But I think new placement (at the end) makes more sense

```diff
-Creating decoders for process `\toggle-re.$proc$toggle-re_ref.v:20$11'.
-Creating decoders for process `\toggle-re.$proc$toggle-re_ref.v:17$10'.
+Creating decoders for process `\toggle-re.$toggle-re_ref.v:20$proc$11'.
+Creating decoders for process `\toggle-re.$toggle-re_ref.v:17$proc$10'. 
```
Log difference is about 82MB so hard to review, but for alumacc/basic.ys there is difference in number of wires.
Note that I have run first tests with yosys based on your branch but without your commits so it should match.


```diff
@@ -148,7 +148,7 @@
         +----------Local Count, excluding submodules.
         | 
        12 wires
-       66 wire bits
+       38 wire bits
         8 public wires
        22 public wire bits
         7 ports                   
```
followed by  (note there are no more undriven)
```diff
 5.5. Executing EQUIV_MAKE pass (creating equiv checking module).
-Presumably equivalent wires: a1_gold (\a1_gold), a1_gate (\a1_gate) -> a1
-Presumably equivalent wires: a2_gold (\a2_gold), a2_gate (\a2_gate) -> a2
-Presumably equivalent wires: ab1_gold (\ab1_gold), ab1_gate (\ab1_gate) -> ab1
-Presumably equivalent wires: b1_gold (\b1_gold), b1_gate (\b1_gate) -> b1
-Presumably equivalent wires: b2_gold (\b2_gold), b2_gate (\b2_gate) -> b2
-Presumably equivalent wires: c_gold (\c_gold), c_gate (\c_gate) -> c
-Presumably equivalent wires: d_gold (\d_gold), d_gate (\d_gate) -> d
 Presumably equivalent wires: y_gold (\y_gold), y_gate ($auto$maccmap.cc:245:synth$45_gate) -> y
-Changing input A of cell $add$<<EOF:5$3_gate ($add): \ab1_gate -> \ab1
-Changing input A of cell $add$<<EOF:5$3_gold ($add): \ab1_gold -> \ab1
-Setting undriven nets to undef: $auto$wreduce.cc:514:run$10_gold [5:4]
-Setting undriven nets to undef: $auto$wreduce.cc:514:run$10_gate [5:4]
+Presumably equivalent wires: d_gold (\d_gold), d_gate (\d_gate) -> d
+Presumably equivalent wires: c_gold (\c_gold), c_gate (\c_gate) -> c
+Presumably equivalent wires: b2_gold (\b2_gold), b2_gate (\b2_gate) -> b2
+Presumably equivalent wires: b1_gold (\b1_gold), b1_gate (\b1_gate) -> b1
+Presumably equivalent wires: ab1_gold (\ab1_gold), ab1_gate (\ab1_gate) -> ab1
+Presumably equivalent wires: a2_gold (\a2_gold), a2_gate (\a2_gate) -> a2
+Presumably equivalent wires: a1_gold (\a1_gold), a1_gate (\a1_gate) -> a1
+Changing input A of cell $<<EOF:5$add$3_gate ($add): \ab1_gate -> \ab1
+Changing input A of cell $<<EOF:5$add$3_gold ($add): \ab1_gold -> \ab1      
```
This is not only test that have different number of wire bits. Since names of cells/wires are different it could be that abc is just producing different results due to that, but worth of checking. 


For sv-elab part, we need to check with Martin but so far in that code there were ifdefs depending of yosys version if API change, so it supported building with older versions.  Guess we can make this work and not look ugly in code base.

Tested using rtlil files generated by previous versions and handling those looks fine.

kernel/yosys_common.h
```diff
@@ -298,18 +313,17 @@ extern int yosys_xtrace;
...
 		return prefix.get(); \
-	}(__FUNCTION__))
+	}(__FUNCTION__), std::to_string(YOSYS_NAMESPACE_PREFIX autoidx++)}}
 #define NEW_ID_SUFFIX(suffix) \
-	YOSYS_NAMESPACE_PREFIX new_id_suffix(__FILE__, __LINE__, __FUNCTION__, suffix)
-
-namespace ID = RTLIL::ID;
-
+	YOSYS_NAMESPACE_PREFIX TwineSpec{YOSYS_NAMESPACE_PREFIX TwineSpec::AutoSuffix{[](std::string_view func) -> const std::string * { \
+		static std::unique_ptr<const std::string> prefix(YOSYS_NAMESPACE_PREFIX create_id_prefix(__FILE__, __LINE__, func)); \
```
mmicko, line 324: We can use `source_location` here instead of __FILE__ and __LINE__ . Include part from `register.h` can be copied in this .h, think it could be useful on other places as well.

kernel/yosys_common.h
```diff
@@ -267,6 +271,17 @@ inline void memhasher() { if (memhasher_active) memhasher_do(); }
...
+void log_assert_failure(const char *expr, const char *file, int line);
+#ifndef NDEBUG
+static inline void log_assert_worker(bool cond, const char *expr, const char *file, int line) {
+	if (!cond) log_assert_failure(expr, file, line);
+}
+#  define log_assert(_assert_expr_) YOSYS_NAMESPACE_PREFIX log_assert_worker(_assert_expr_, #_assert_expr_, __FILE__, __LINE__)
+#else
+#  define log_assert(_assert_expr_) do { if (0) { (void)(_assert_expr_); } } while(0)
+#endif
+
```
mmicko, line 284: in log rewrite i did not touch these, since you have been moving them around, but we anyway have log_yosys_abort_message in top of file. Maybe worth grouping these together in this file.

kernel/log.cc
```diff
@@ -240,8 +231,6 @@ void log_formatted_header(RTLIL::Design *design, std::string_view format, std::s
 	if (log_hdump.count(header_id) && design != nullptr)
 		for (auto &filename : log_hdump.at(header_id)) {
 			log("Dumping current design to '%s'.\n", filename);
-			if (yosys_xtrace)
-				IdString::xtrace_db_dump();
 			Pass::call(design, {"dump", "-o", filename});
 			if (yosys_xtrace)
```
mmicko, line 235: logging end of dump with previous xtrace_db_dump removed does not make sense, so think should also be removed

kernel/rtlil.h
```diff
@@ -2815,6 +2332,11 @@ void RTLIL::Process::rewrite_sigspecs2(T &functor)
 		it->rewrite_sigspecs2(functor);
 }
 
+namespace RTLIL {
```
mmicko, line 2335: Guess this empty namespace block can be removed.

kernel/register.cc
```diff
@@ -306,8 +306,8 @@ void Pass::call(RTLIL::Design *design, std::vector<std::string> args)
 
 void Pass::call_on_selection(RTLIL::Design *design, const RTLIL::Selection &selection, std::string command)
 {
-	std::string backup_selected_active_module = design->selected_active_module;
-	design->selected_active_module.clear();
+	IdString backup_selected_active_module = design->selected_active_module;
+	design->selected_active_module = IdString::Null;
```
mmicko, line 310: not sure how intuitive this is. Maybe it would be better to have `clear()` there doing the same thing. Also hiding implementation details that way.

kernel/rtlil.cc
```diff
@@ -1401,23 +1242,31 @@ void RTLIL::Design::optimize()
...
+	log_assert(dst->modules_.empty());
+	dst->twines = twines;
+	for (auto it = modules_.rbegin(); it != modules_.rend(); ++it)
+		it->second->clone(dst);
+}
+
+bool RTLIL::Design::selected_module(IdString mod_name) const
 {
-	if (!selected_active_module.empty() && mod_name != selected_active_module)
+	if (selected_active_module != IdString::Null && mod_name != selected_active_module)
```
mmicko, line 1255: Same as for clear, empty could be still used, less impact on source

rocallahan:
I haven't looked at the code, just the PR description.

> Utilities to hash, sort, and compare IdStrings based on the strings they represent is provided, it's just not the default behavior as that could get computationally expensive in hot loops and is not really needed.

I'd like to question this decision a little. Yosys's determinism guarantee has always been fragile; anything that causes IdStrings to be created differently breaks it, because that leads to different IdString orderings. For example, running a series of passes, clearing the Design, and then running some new passes does not give the same results in general as just running the new passes. This trips people up over and over again in our neck of the woods. I always thought it would be great if we could fix this by making IdString hashing and ordering a function of the IdString contents. I tried it but I couldn't totally eliminate the performance penalty, though I got close. (I used a special hash function where strings that look like autoidx IdStrings use the autoidx value as the hash value and other strings cached their hash value in the global intern table.)

I think if you did this change at this time it would fix some of the cons here:
> new twines block in textual RTLIL is required to keep RTLIL replayable
> equality comparison of two IdStrings that resolve to the same string can compare as not equal due to different twine structure.

This latter point seems like a real problem to me. It breaks the fundamental principle of string interning. I believe you that you can get away with it in practice, but it's highly counterintuitive.

Since I haven't looked at the code, I haven't judged how hard it would be to cache the hash value with this approach.

widlarizer:
@rocallahan Thanks, that's a valid expectation. If I reworked what's currently global state (hashidx_ and autoidx) to be captured in the design state or replaced with alternatives, wouldn't that use case be served too? Should get us all the determinism we'd like. As canonical ordering etc just doesn't seem like a necessary part of a minimal model of the problem domain, but as harm reduction for Yosys sins, and it can get pricey. Maybe it can be made cheap, but I'd like to take a stab at the sins themselves first

rocallahan:
> If I reworked what's currently global state (hashidx_ and autoidx) to be captured in the design state or replaced with alternatives, wouldn't that use case be served too?

Yes, that would certainly improve the situation.

kernel/yosys_common.h
widlarizer: I'd need to strip the rest of the signature from the function name, right?
https://godbolt.org/z/cTo5fsfME
I don't know of a way to get just the function name from std::source_location

mmicko: Right std::source_location provides same output as __PRETTY_FUNCTION__ and that one can be useful for log messages, but not in this case. It is fine to leave it as __FUNCTION__

mmicko:
@widlarizer just a reminder that it still needs fixing when Verific is enabled, all is in 3 files `passes/hierarchy/hierarchy.cc` (just trivial change) but  `frontends/verific/verificsva.cc` and `frontends/verific/verificsva.cc` are bit more demanding.

PR #6150, opened 2026-08-25
rmports: remove output ports unused by all instances [sc-799]

### LINKED ISSUE #1793: rmports doesn't remove unused module output.
    module TOP(a, b, y);
      input [3:0] a, b;
      output [2:0] y;
      reg [2:0] tmp;
      wire NC;
      sub p(a, b, {NC, tmp});
      assign y = tmp;
    endmodule
    
    module sub(as, bs, ys);
      input [3:0] as, bs;
      output [3:0] ys;
      assign ys = as & bs;
    endmodule

```
read_verilog test.v
prep
splitnets
rmports
delete w:NC
clean -purge
rmports
splitnets -ports
rmports
clean -purge
opt_clean -purge

```
## Expected behavior
module sub should have 3bit wide output ys instead of 4.

## Actual behavior
module sub ys[3] is not used by design and according to "help rmports"

> This pass identifies ports in the selected modules which are not used or
> driven and removes them.
>

hexratcc:
This PR aims to fix #1793 - `rmports` only examined connections inside a module, so an internally driven output was never removed even when no instance used it. This adds a design-wide scan so that an output port is now also removed when its connections on all instances lead nowhere (public wires, parent ports, keep wires and modules with processes count as uses). Top modules and never-instantiated modules are untouched.

### DIFF
--- a/passes/opt/rmports.cc
+++ b/passes/opt/rmports.cc
@@ -33,7 +33,10 @@ struct RmportsPassPass : public Pass {
 		log("    rmports [selection]\n");
 		log("\n");
 		log("This pass identifies ports in the selected modules which are not used or\n");
-		log("driven and removes them.\n");
+		log("driven and removes them. Output ports which are driven inside the module\n");
+		log("are also removed if their connections on all instances of the module lead\n");
+		log("nowhere. This does not apply to the top module and to modules which are\n");
+		log("never instantiated.\n");
 		log("\n");
 	}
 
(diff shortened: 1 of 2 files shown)


widlarizer (changes requested):
Even though I really wish this wasn't the case, it had kind of always been possible to use Yosys for module-scope operations on designs with multiple modules and no unique top module. A lot of passes, like this one, would work fine on non-parametric designs without running `hierarchy` that would enforce a single top module. The change (as described in the help message - I hadn't read the code yet) can be the default behavior of a top module exists, because that implies we are working with a nice regular design, but if it doesn't, we can't assume that we can remove a port on a non-top module if the port is unused. I hope this makes sense, we don't even have convenient terminology to describe these use cases I think

hexratcc:
Didn't think of that, thx. I made it so the driven-output rule now needs a clear top module (one with the `top` attribute, or the only module nothing instantiates, so a flat module library is left alone).

passes/opt/rmports.cc
```diff
@@ -30,33 +30,139 @@ struct RmportsPassPass : public Pass {
...
-		log("    rmports [selection]\n");
+		log("    rmports [options] [selection]\n");
 		log("\n");
 		log("This pass identifies ports in the selected modules which are not used or\n");
 		log("driven and removes them.\n");
 		log("\n");
+		log("An output which is driven inside the module is also removed, if no\n");
+		log("instance of the module uses it. The top module, a module which nothing\n");
+		log("instantiates, and a port with the keep attribute stay. This needs a clear\n");
+		log("top module: one with the top attribute, or the only module which nothing\n");
```
widlarizer, line 41: I don't think it's desirable to try and infer a top module outside of `hierarchy`. You should actually rely on the top attribute here

widlarizer (changes requested):

widlarizer (approved):

PR #6154, opened 2026-08-26
liberty: add filtering for things we need

widlarizer:
Filterlib gets all of the AST, possible external users of libparse get everything by default as well, but Yosys synthesis passes now filter for everything that they all collectively need, which is compatible with caching (caching is aware, and we still avoid multiple liberty parsing runs within a synthesis run). This skips timing information which dominates real world PDKs and saves a lot of runtime and memory. I'm relying on pre-existing testing coverage for all necessary parts of the AST being present. On the other hand, two static_asserts check that the expensive groups aren't included in the synth filter. All in all the filtering is shaped a bit like `kernel/newcelltypes.h`

### DIFF
--- a/passes/techmap/libparse.cc
+++ b/passes/techmap/libparse.cc
@@ -49,17 +49,20 @@ using namespace Yosys;
 
 LibertyAstCache LibertyAstCache::instance;
 
-std::shared_ptr<const LibertyAst> LibertyAstCache::cached_ast(const std::string &fname)
+std::shared_ptr<const LibertyAst> LibertyAstCache::cached_ast(const std::string &fname, const LibertyFilter &filter)
 {
 	auto it = cached.find(fname);
 	if (it == cached.end())
 		return nullptr;
+	if (!it->second.filter.covers(filter))
+		return nullptr;
 	if (verbose)
 		log("Using cached data for liberty file `%s'\n", fname);
-	return it->second;
+	return it->second.ast;
 }
 
-void LibertyAstCache::parsed_ast(const std::string &fname, const std::shared_ptr<const LibertyAst> &ast)
+void LibertyAstCache::parsed_ast(const std::string &fname, const LibertyFilter &filter,
+		const std::shared_ptr<const LibertyAst> &ast)
 {
 	auto it = cache_path.find(fname);
 	bool should_cache = it == cache_path.end() ? cache_by_default : it->second;
@@ -67,7 +70,7 @@ void LibertyAstCache::parsed_ast(const std::string &fname, const std::shared_ptr
 		return;
 	if (verbose)
 		log("Caching data for liberty file `%s'\n", fname);
-	cached.emplace(fname, ast);
+	cached.emplace(fname, CacheEntry{filter, ast});
 }
 
 #endif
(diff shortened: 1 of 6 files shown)


passes/techmap/libparse.cc
```diff
@@ -644,19 +648,23 @@ LibertyAst *LibertyParser::parse(bool top_level)
...
-		return NULL;
+		return ParseResult::closed();
 
 	if (tok != 'v') {
 		report_unexpected_token(tok);
 	}
 
-	LibertyAst *ast = new LibertyAst;
-	ast->id = str;
+	std::string id = std::move(str);
```
calewis, line 662: Is there a benefit to "renaming" this string? Otherwise you could just name the string at 637, `id` and not pay for the move (at runtime or at read time).

calewis: I see, `str` is used farther down.  Still seems like:
```
std::string id;
... do stuff with id ...
std::string value, str;
... enter the while(1) loop ...
```
would read a bit more cleanly and save you a move.

widlarizer: Usually what I'd see in a situation like this is a comment that says something like `// str is an identifier/group name here` above the usage of str, so I was renaming it with a move to express that. But in a simple situation like this it was mostly distracting. If I named it above it would not yet have been sure to be an "id". str is used in this out-parameter reference way just to avoid extra allocations so it's generally a weirdly shaped code for a decent reason

passes/techmap/libparse.cc
```diff
@@ -49,25 +49,28 @@ using namespace Yosys;
...
+		const std::shared_ptr<const LibertyAst> &ast)
 {
 	auto it = cache_path.find(fname);
 	bool should_cache = it == cache_path.end() ? cache_by_default : it->second;
 	if (!should_cache)
 		return;
 	if (verbose)
 		log("Caching data for liberty file `%s'\n", fname);
-	cached.emplace(fname, ast);
+	cached.emplace(fname, CacheEntry{filter, ast});
```
hexratcc, line 73: I think we want `cached[fname] = CacheEntry{filter, ast}` here. `dict::emplace` is a no-op when the key exists, so once a narrower AST is cached, any request with a broader filter misses in `cached_ast`, re-parses, and the result is dropped. Every later broad request re-parses again.

hexratcc (approved):
Just a minor nitpick, other than that lgtm.

hexratcc:
Thanks, lgtm :).

calewis:
Quick question.  What will be filtered by default if anything?

passes/techmap/libparse.h
```diff
@@ -33,11 +37,119 @@
...
+
+	inline constexpr auto stat_liberty_names = merge_names(
+		liberty_common_names,
+		liberty_names({
+			"ff", "port_names",
+			"single_area_parameterised", "double_area_parameterised",
+		}));
+
+	inline constexpr auto liberty_synthesis_names =
+			merge_names(read_liberty_names, dfflibmap_names, clockgate_names, stat_liberty_names);
```
widlarizer, line 136: @calewis see here and above, the filter used by all yosys passes is a superset of the groups etc needed by yosys for synthesis

PR #6155, opened 2026-08-26
ABCU: fix bufnorm edge cases

hexratcc:
While working on write_xaiger2 for ABCU I discovered some minor edge cases, list of my fixes is below. 
- Move the enqueue seeding from into `Module::bufNormalizeInit()`, guarded by a per-module flag. A module created after the design entered bufnorm mode was never seeded, so its cells and wires never got queued.
- Skip zero-width input ports when forcing an `$input_port` driver cell.
- Skip zero-width wires in the undriven scan, since they carry no bits and need no driver.

### DIFF
--- a/kernel/rtlil_bufnorm.cc
+++ b/kernel/rtlil_bufnorm.cc
@@ -44,38 +44,37 @@ void RTLIL::Design::bufNormalize(bool enable)
 				wire->driverPort_ = IdString();
 			}
 			module->buf_norm_connect_index.clear();
+			module->buf_norm_initialized = false;
 		}
 
 		flagBufferedNormalized = false;
 		return;
 	}
 
-	if (!flagBufferedNormalized)
-	{
-		for (auto module : modules())
-		{
-			// When entering buf normalized mode, we need the first module-level bufNormalize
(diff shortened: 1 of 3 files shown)


widlarizer (changes requested):
Zero width skips are fine, but here's how new modules get added with signorm instead: https://github.com/YosysHQ/yosys/pull/5804/changes#diff-d049f2b84b01cb3ae27d93df05b345b4122c7ec7f4acbdd7fd9567a4757d6aa8R1237
Making an existing RTLIL method bufnorm-aware and moving it to rtlil_bufnorm.cc seems preferable to complicating the RTLIL interface by exposing more bufnorm details

kernel/rtlil_bufnorm.cc
```diff
@@ -53,23 +66,7 @@ void RTLIL::Design::bufNormalize(bool enable)
 	if (!flagBufferedNormalized)
 	{
 		for (auto module : modules())
-		{
-			// When entering buf normalized mode, we need the first module-level bufNormalize
```
widlarizer, line 57: You dropped this handy comment

widlarizer (approved):

PR #6098, opened 2026-08-09
Enrich silimate pass help() with durable algorithm notes

akashlevy:
## Summary
- Expand `help()` on `obs_clean`, `splitnetlist`, `annotate_cell_fanout`, and `carvenetlist` with algorithms, limitations, and DESIGN-cell caveats previously kept in preqorsor `dev/` markdown.
- Companion to the preqorsor PR that deletes those developer docs in favor of code-local documentation.

## Test plan
- [ ] Spot-check `help obs_clean`, `help splitnetlist`, `help annotate_cell_fanout`, `help carvenetlist` in a local Yosys build
- [ ] No functional code changes expected (help text only)

Made with [Cursor](https://cursor.com)

PR #6156, opened 2026-08-27
Bison mirror: add to release too

gigeresk:


KrystalDelusion:
@akashlevy How did you get on with requesting the fork be unlinked? https://github.com/YosysHQ/yosys/pull/6098#issuecomment-5229165868

akashlevy:
@KrystalDelusion Unfortunately, it's not so simple to unlink our fork, since it itself has forks. We're looking at other options to prevent our PRs from flying to the wrong place (using git hooks and such).

In the meantime, if it's bothering you, feel free to put some automation in place to auto-delete PRs from Silimate branches ("from Silimate:*"). If we want to upstream something, we'll make sure to do so from our individual forks.

KrystalDelusion:
> In the meantime, if it's bothering you, feel free to put some automation in place to auto-delete PRs from Silimate branches ("from Silimate:*"). If we want to upstream something, we'll make sure to do so from our individual forks.

Y'all are good about closing them yourself, and I don't think  it's possible to delete a PR without going through GitHub support.  So automation could only do what you're already doing, and I doubt it would be fast enough to prevent the push notification of a PR requesting review by code owners.

akashlevy:
Thanks for your understanding and apologies about this recurrence.

akashlevy:
(Hope GitHub eventually adds a way to create a default PR target branch for forks...)

PR #6161, opened 2026-08-31
rtlil: remove dead verilog bindings

widlarizer:
#2875 as a component of #2752 which never made it in and never will. I'm removing write-only `RTLIL::Design::bindings_`, `RTLIL::Binding`, `AST::Binding`. `tests/bind/` only contained checks that the constructs can be parsed, not that they are implemented. In such a case, I think they should error as unsupported


mmicko:
Was looking if it is worth keeping tests and making them use `read_slang` instead, but these are just testing AST parsing since there is no underneath functionality. Due to all that  this is good as is.

PR #6166, opened 2026-09-02
Run ABC for each module in parallel

jmichalski-ant:
Currently parallelism in `AbcPass` is limited to parallelization over clock-domains within a single module and is limited to DFF mode, so not every design can take advantage of this kind of parallelization. The goal of this PR is to reduce runtime by extending parallelism support in `AbcPass`.

Current clock-domain only parallelism gathers ABC tasks for each clock-domain inside a single module and simultaneously executes them in parallel. This PR modifies it to gather ABC tasks over the whole module processing loop, allowing parallel execution of ABC jobs from multiple modules at the same time in both DFF and non-DFF mode.

It shows noticeable runtime improvement on tested designs. Presented results were obtained using designs from OpenROAD-flow-scripts, configured to not flatten the design and show runtime of the parallelized loop in `AbcPass`:
                    
| design name                    | main          | this PR
|---------------------------|-------------------|---
| aes                       | 1.260s              | 0.520s
| ethmac                    | 8.401s              | 5.127s
| jpeg                      | 4.624s              | 0.548s
| ariane136                 | 113.919s            | 25.371s
| ibex                      | 8.869s              | 1.795s
| black_parrot              | 18.038s             | 2.197s
| tinyRocket                | 16.114s             | 5.907s  

On one large proprietary design the runtime of `AbcPass` was reduced 8x with this PR.
  
Multiple existing tests go through this pass in both DFF and non-DFF mode, hitting both branches of `if (num_worker_threads > 0)`, so exercising both parallel and non-parallel execution.

### DIFF
--- a/passes/techmap/abc.cc
+++ b/passes/techmap/abc.cc
@@ -611,7 +611,7 @@ std::string AbcModuleState::remap_name(RTLIL::IdString abc_name, RTLIL::Wire **o
 				const auto &bit = signal_bits.at(sid);
 				if (bit.wire != nullptr)
 				{
-					std::string s = stringf("$abc$%d$%s", map_autoidx, bit.wire->name.c_str()+1);
+					std::string s = stringf("$abc$%d$%d$%s", map_autoidx, state_index, bit.wire->name.c_str()+1);
 					if (bit.wire->width != 1)
 						s += stringf("[%d]", bit.offset);
 					if (isnew)
@@ -624,7 +624,7 @@ std::string AbcModuleState::remap_name(RTLIL::IdString abc_name, RTLIL::Wire **o
 			}
 		}
 	}
-	return stringf("$abc$%d$%s", map_autoidx, abc_name.substr(1));
+	return stringf("$abc$%d$%d$%s", map_autoidx, state_index, abc_name.substr(1));
 }
 
 void AbcModuleState::dump_loop_graph(FILE *f, int &nr, dict<int, pool<int>> &edges, pool<int> &workpool, std::vector<int> &in_counts)
@@ -748,7 +748,7 @@ void AbcModuleState::handle_loops(AbcSigMap &assign_map, RTLIL::Module *module)
 
 			log_assert(signal_bits[id1].wire != nullptr);
 
-			RTLIL::Wire *wire = module->addWire(stringf("$abcloop$%d$%d", map_autoidx, loop_count++));
+			RTLIL::Wire *wire = module->addWire(stringf("$abcloop$%d$%d$%d", map_autoidx, state_index, loop_count++));
 
 			bool first_line = true;
 			for (int id2 : edges[id1]) {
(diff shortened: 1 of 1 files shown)


Ravenslofty:
If ever there was a demonstration for why "If your work is part of a larger effort, please discuss your general plans on [Discourse](https://yosyshq.discourse.group/) first to align your vision with maintainers." is the first line of the issue template, it would be this; I have already announced [the planned deprecation and removal](https://yosyshq.discourse.group/t/upcoming-abc-flow-changes/133) of `abc`.

To be clear, I do think module-level parallelism (and perhaps also clock-domain parallelism) is a useful addition to Yosys, but I think this needs to be implemented in `abc_new` (as it is currently named). That pass is presently one big loop of {prepare, map, reintegrate} over each module, and I think it would be possible to break it up into three {prepare, map, reintegrate} loops that could then be parallelised.

widlarizer:
What are the difference of your approach and #5653? How many threads did it take to get that speedup?

jmichalski-ant:
`What are the difference of your approach and https://github.com/YosysHQ/yosys/pull/5653?`

Feature-wise looks like the only significant difference is that https://github.com/YosysHQ/yosys/pull/5653 supports only non-DFF mode while this PR supports both DFF and non-DFF modes. I compared the performance (with a small modification to https://github.com/YosysHQ/yosys/pull/5653, which by default reserves 4 cores for the main thread - I changed it to 1 reserved core to make it in line with this PR) and it was very close, with this PR being slightly (up to ~10%) faster on a few designs. Looks like this difference can be attributed to this PR processing modules in sorted order from the largest to the smallest.

`How many threads did it take to get that speedup?`

16, but note that it was on intel cpu with different core types, so the scaling wouldn't be linear with core count on it. When I checked scaling on some synthetic design with evenly sized modules and pinned it to a core cluster with the same core type, the runtime scaling was close to linear with core count. In case of the proprietary design I unfortunately don't have information about core count.

QuantamHD:
@widlarizer It sounds like from our discussion this morning this can be reviewed and merged if it looks good. Appreciate your flexibility as this massively benefits some of our designs.

jmichalski-ant:
I had to rebase and adjust this PR to accommodate merge of https://github.com/YosysHQ/yosys/pull/6177

@widlarizer  I changed the way autoidx is handled in this pass, since keeping the current scheme (reserving idx for each ABC job ahead of time with `autoidx.ensure_at_least()`) would require knowledge of the sum of `GetSize(assigned_cells)`'s before the module processing loop and this value is not easily obtainable at that point, since `assigned_cells`s are computed by the module processing loop itself. Instead I modified it to reserve a single idx common for each ABC job during a single `AbcPass` and use a pair of this idx and `state_index` to create identifiers, e.g.  
`module->addWire(stringf("$abcloop$%d$%d", map_autoidx /*unique for each ABC job*/, loop_count++))`   
became  
`module->addWire(stringf("$abcloop$%d$%d$%d", map_autoidx /*unique for each AbcPass*/, state_index /*unique for each ABC job in a single AbcPass*/, loop_count++))`,  
still generating unique identifier for each element

I checked it with the reproducer from https://github.com/YosysHQ/yosys/issues/6170 (also extended it to contain multiple modules) and the results were deterministic

jmichalski-ant:
superseded by https://github.com/YosysHQ/yosys/pull/6194

Issue #6165, opened 2026-09-01
Fails to substitute module code  for instance in particular situations


pbreuer:
### Version

Yosys 0.66 (git sha1 86f2ddebce7e98ce7cacc27e8a5c14cb53b51b51)

### On which OS did this happen?

Linux

### Reproduction Steps

(Also yosys 0.67+102)

The issue is a wire initialized to some value, e.g. `logic foo = 0;`, and then driven in subsequent clocks but not the inital state by a module, e.g. `flop foodriver (clk, foo_in, foo);` where the "flop"  module merely does a delayed update of foo, copying foo_in, e.g. `module foo (input logic clk, input logic foo_in, output logic foo); always @(posedge clk) foo <= foo_in; endmodule`.

Substituting the module code for the instance gives `logic foo=0; always @(posedge clk) foo <= foo_in;`, which works fine. foo starts zero and thereafter follows foo_in, one clock behind.

But  yosys does not do the substitution as above. Instead it sees one wire, two drivers, and SILENTLY drops the module code entirely, leaving only the foo=0 initialization, and foo is stuck at zero forever. That is a different functional semantics altogether.

The code described above and detailed for reproducibility in the following part of the report is accepted by yosys, sv2v, verilator. It is rejected by icarus. 

This bug has been investigated in "collaboration" with the AI Claude, which analysed what yosys does with a degree of expert knowledge on yosys that I do not possess myself. I merely noted the problem originally, and Claude elaborated the mechanisms and generated various tests (carried out itself, all of that is to be found below), which allowed me to opine this is probably a yosys bug, because as a matter of good engineering practice I project that yosys ought to at least FLAG  a potential problem there in terms of its treatment.

Claude generates the following example code ("repro.sv") for reproducing the problem. Claude's ff1 module is what I called the "foodriver" module above.  It's just a foo <= foo_in, clocked. The "top" module tries both what I said goes wrong above, i.e. initializing foo (Claude calls it "q_inst") and also driving it via a module instance, with delay in the module, so it is not driven on clock zero. Then Claude also tries doing it all instead inline in the top module, with the module substituted through, on a foo that Claude calls "q_gold". 

Claude drives both q_inst and q_gold off the same mythical input "d". Claude makes an assert that q_inst and q_gold should be the same, and that fails, for the reasons connected with yosys treatment of that code that I outlined above. Except that Claude will go into detail about what yosys does and how to fix it .... but first the code. This is *repro.sv*:
 
```
[...]

PR #6169, opened 2026-09-02
check: detect process-driver conflicts

widlarizer:
Fixes the necessity to sequence `check` after `proc` to avoid #6165. In other words: `check` now reports an error in that scenario (see included test) even before `proc`. The way it works is that outside of initial blocks we track all bits written to by a process and count each such process to be a driver of those bits. We skip initial blocks to avoid warnings in the second case in the test which we already supported. In the first case, the initial value assignment isn't in an initial block, but as an unconditional assignment in a process like this:

```
process $proc$...
  assign $0\q_inst[0:0] 1'0
  sync always
    update \q_inst $0\q_inst[0:0]
end
```

### DIFF
--- a/passes/cmds/check.cc
+++ b/passes/cmds/check.cc
@@ -160,13 +160,17 @@ struct CheckPass : public Pass {
 			TopoSort<std::pair<RTLIL::IdString, int>> topo;
 			for (auto &proc_it : module->processes)
 			{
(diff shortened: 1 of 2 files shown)


whitequark (approved):
We should've had this long ago, but the second best time is now. Thank you!

PR #6171, opened 2026-09-02
abc: fixed preparing/extracting order for determinism

### LINKED ISSUE #6170: abc -dff: different netlist on every run with multiple threads (same OS, same input)
### Summary

With more than one thread, `abc -dff` produces a **different netlist on every run** from byte-identical input, on the same machine, same binary, same OS. Setting `YOSYS_MAX_THREADS=1` makes it fully reproducible.

This contradicts the assumption stated in #5205: *"Yosys is currently deterministic within the same OS: On any version of Linux, whether it's x86-64 or aarch64, for the same inputs, Yosys will always return the same result."*

### Reproducer

One file, no downloads. `retime_nondet.v`:

```verilog
module retime_nondet #(parameter AW = 6, parameter DW = 32) (
    input                  clk,
    input                  rst,
    input  [AW-1:0]        addr,
    input  [DW-1:0]        din,
    input  [DW/8-1:0]      wen,
    output reg [DW-1:0]    dout,
    output reg [DW-1:0]    acc
);
    reg [DW-1:0] mem [0:(1<<AW)-1];
    integer b;
    always @(posedge clk) begin
        for (b = 0; b < DW/8; b = b + 1)
            if (wen[b])
                mem[addr][b*8 +: 8] <= din[b*8 +: 8];
        dout <= mem[addr];
        acc  <= rst ? {DW{1'b0}} : (acc ^ {mem[addr][DW-2:0], mem[addr][DW-1]}) + din;
    end
endmodule
```

```bash
for i in $(seq 8); do
  yosys -q -p "read_verilog retime_nondet.v; hierarchy -top retime_nondet; proc; \
    opt -fast -full; memory_map; opt -full; techmap; \
    abc -dff -script +st;retime,-v,-o;retime,-M,5,-v,-o;retime,-M,5,-v,-o;retime,-M,5,-v,-o;map; \
    write_rtlil out.rtlil"
  sha1sum out.rtlil
done
```

The ABC script is the retiming script shipped by OpenROAD-flow-scripts (`flow/scripts/abc_retime.script`), inlined here so the reproducer needs no second file.

### Observed

8 runs, **6 distinct netlists**:

```
3x 09d4f4c41967
   c5309dfd9f22
   b1255a71ff27
   a9d05f25a8b9
   742ac1bc7e0c
   702bee6ebdc4
```

Sizes differ too (e.g. 2564324 vs 2565572 vs 2566509 bytes), so this is a structurally different mapping, not just renaming.

With `YOSYS_MAX_THREADS=1`, every run is identical.

### Why this design triggers it

21 lines of source, but `memory_map` expands the byte-enabled memory into a large FF array with per-word, per-bit write enables:

```
4909 cells
2048   $_DFFE_PP_
2048   $_MUX_
  32   $_DFF_P_
  32   $_SDFF_PP0_
```

`abc -dff` then splits the module into **258 clock/enable domains**, one ABC invocation each, which is what the thread pool parallelises. Designs that produce only a handful of domains (e.g. the `jpeg` design in OpenROAD-flow-scripts) are stable in both modes.

### Provenance

First observed on the public `ethmac` design in OpenROAD-flow-scripts (5 runs, 5 distinct netlists), then minimised to the synthetic testcase above.

### What it is not

- **Not ASLR** — `setarch -R` does not stabilise it.
- **Not CPU contention** — stable under heavy background load; varying CPU affinity does not fix it (note `hardware_concurrency()` ignores affinity, so `taskset` alone does not reduce the thread count).
- **Not the temp directory names** — single-threaded runs are bit-identical despite the random `/tmp/yosys-abc-XXXXXX` paths.
- **Not ABC process reuse** — passing `abc -exe <a copy of the abc binary>` takes the non-reuse branch and still diverges.

### What I could not determine from outside

Whether the parallel ABC invocations receive *different inputs* (i.e. the clock-domain grouping differs) or receive the same inputs and have their results *integrated* differently. The pass does re-order completed work by index (`work_finished_by_index`), so the leak is elsewhere in the concurrent path.

### Workaround

`YOSYS_MAX_THREADS=1` for reproducible synthesis. In our case the cost is nil, because the surrounding build system already parallelises independent synthesis jobs.

### Environment

- Yosys 0.68 (Bazel Central Registry `yosys 0.68.bcr.1`), Linux x86-64.
- This build carries one local patch, to the `clockgate` pass only. It touches neither `passes/techmap/abc.cc` nor `kernel/threading.*`, and the reproducer above never invokes `clockgate`.

*Reproducer minimised, and the eliminations above run, with AI assistance.*

widlarizer:
Aims to fix #6170, needs a bit more scrutiny

### DIFF
--- a/passes/techmap/abc.cc
+++ b/passes/techmap/abc.cc
@@ -2582,25 +2582,28 @@ struct AbcPass : public Pass {
 						work_finished_queue.push_back(std::move(*work));
 					}
 				});
-			int state_index = 0;
-			int next_state_index_to_process = 0;
 			std::vector<std::unique_ptr<AbcModuleState>> work_finished_by_index;
 			work_finished_by_index.resize(assigned_cells.size());
-			int work_finished_count = 0;
-			for (auto &it : assigned_cells) {
-				// Make sure we process the results in the order we expect. When we can
-				// process results before the next ABC run, do so, to keep memory usage low(er).
-				while (std::optional<std::unique_ptr<AbcModuleState>> work =
-						work_finished_queue.try_pop_front()) {
-					work_finished_by_index[(*work)->state_index] = std::move(*work);
-					++work_finished_count;
+			// Make sure we process the results in the order we expect. Process each result
+			// a fixed number of ABC runs after it was prepared, to keep memory usage low(er)
+			// without letting the order the runs finish in affect the result.
+			const int extract_lookahead = 256;
+			auto next_to_prepare = assigned_cells.begin();
+			for (int i = 0; i < GetSize(assigned_cells) + extract_lookahead; i++) {
+				if (i >= extract_lookahead) {
+					int extract_index = i - extract_lookahead;
+					while (work_finished_by_index[extract_index] == nullptr) {
+						std::unique_ptr<AbcModuleState> work = *work_finished_queue.pop_front();
+						work_finished_by_index[work->state_index] = std::move(work);
+					}
+					work_finished_by_index[extract_index]->extract(assign_map, design, mod);
+					work_finished_by_index[extract_index] = nullptr;
 				}
-				while (work_finished_by_index[next_state_index_to_process] != nullptr) {
-					work_finished_by_index[next_state_index_to_process]->extract(assign_map, design, mod);
-					work_finished_by_index[next_state_index_to_process] = nullptr;
-					++next_state_index_to_process;
-				}
-				std::unique_ptr<AbcModuleState> state = std::make_unique<AbcModuleState>(config, initvals, state_index++);
+				if (i >= GetSize(assigned_cells))
+					continue;
+				auto &it = *next_to_prepare;
+				++next_to_prepare;
+				std::unique_ptr<AbcModuleState> state = std::make_unique<AbcModuleState>(config, initvals, i);
 				state->clk_polarity = std::get<0>(it.first);
 				state->clk_sig = assign_map(std::get<1>(it.first));
 				state->en_polarity = std::get<2>(it.first);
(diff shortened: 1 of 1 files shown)


oharboe:
We found this because two builds of the same design came out ~200 ps apart in WNS. That made me wonder whether the nondeterminism is just noise, or whether some runs are systematically better — i.e. whether there is performance being left on the table here, for us and for others.

So I put this PR's `extract_lookahead` behind an env var and swept it. It doesn't only bound memory; it picks which netlist you get, and the values are not equally good.

Self-contained on top of this PR (`retime_nondet.v` from #6170, one public liberty, ABC's own `stime` so no external STA is needed):

```bash
curl -sSfLO https://raw.githubusercontent.com/The-OpenROAD-Project/OpenROAD-flow-scripts/master/flow/platforms/nangate45/lib/NangateOpenCellLibrary_typical.lib
for L in 1 2 4 8 16 32 64 128 256; do
  YOSYS_ABC_EXTRACT_LOOKAHEAD=$L yosys -p "
    read_verilog retime_nondet.v; hierarchy -top retime_nondet
    proc; opt -fast -full; memory_map; opt -full; techmap
    abc -dff -script +st;retime,-v,-o;retime,-M,5,-v,-o;retime,-M,5,-v,-o;retime,-M,5,-v,-o;map
    dfflibmap -liberty NangateOpenCellLibrary_typical.lib
    abc -liberty NangateOpenCellLibrary_typical.lib -script +strash;dch;map;topo;stime" 2>&1 |
  awk -v l=$L '/WireLoad/{for(i=1;i<=NF;i++)if($i=="Delay")d=$(i+2); print "lookahead="l"  "d" ps"}'
done
```

```
lookahead=1    4273.95 ps
lookahead=2    4908.74 ps      lookahead=32   4895.13 ps
lookahead=4    5013.82 ps      lookahead=64   4946.16 ps
lookahead=8    4889.77 ps      lookahead=128  4867.81 ps
lookahead=16   5012.53 ps      lookahead=256  4933.84 ps
```

It isn't a gradient — 1 is a cliff, and everything from 2 up is one noisy band ~600 ps worse. 1 is also the value that reproduces byte-for-byte what `YOSYS_MAX_THREADS=1` produces today, so it is the only value that moves nobody's netlist.

The catch, and why I am asking rather than proposing: lookahead 1 serializes the ABC runs — the next domain cannot be prepared until the previous result has been extracted, which is the parallelism this pass exists for (~25% more wall time on this testcase; much worse where individual ABC runs are long). So this looks like a QoR-vs-parallelism tradeoff rather than a free choice.

Two questions:

- Is the mechanism what it looks like — a domain prepared after more extracts sees a more up-to-date design and groups better? The domain grouping itself is computed once up front, so only `prepare_module` benefits from the freshness.
- If that is it, is there a way to get the freshness without the serialization?

I see the same direction on a large private design (small lookahead best, large worst, monotone on one partition), but everything above is from the public testcase.

PR #6173, opened 2026-09-03
log: Colored output for console

mmicko:
Only console output is colored and only in cases output is not redirected to files. Checks could have been done in log call, but since it is not possible to change it after no need to slow down the execution with additional checks. There is no performance impact when writing to file or redirecting output.
Since this is in driver library is not affected and therefor pyosys is not affected.


widlarizer (approved):
Works well with piping to file with `>` and with `-l`, terminal output looks good, code as well

PR #6174, opened 2026-09-03
abc9: migrate to write_xaiger2

Ravenslofty:
More ABC unification work.

The entire goal of the ABC unification work is to clean up the mess that is the Yosys ABC flows:
- `abc` mapping to standard cells and LUTs via eBLIF
- `abc9` mapping to LUTs via XAIGER using `write_xaiger` and `read_aiger -xaiger`.
- `abc_new` mapping to standard cells via XAIGER using `write_xaiger2` and `read_xaiger2`.

Both XAIGER flows [already](https://github.com/YosysHQ/yosys/pull/6077) use `read_aiger -xaiger`; with this PR both flows use `write_xaiger2`. That being said, because `abc9` is used by [every `synth_*` flow now ](https://github.com/YosysHQ/yosys/pull/6103) I have added a "chicken bit" in the form of `scratchpad -set abc9.xaiger 1` to use the old `read_aiger -xaiger` code to work around bugs.

Through the fixing of a number of bufnorm issues, improvement of `write_xaiger2`'s box handling and the odd test which itself was broken in some way.

(The test suite has been updated, and because `xaiger2` is the default, all `abc9` flows implicitly test xaiger2 import).

Co-authored by @hexratcc to whom I am very grateful.

### DIFF
--- a/backends/aiger2/aiger.cc
+++ b/backends/aiger2/aiger.cc
@@ -28,6 +28,7 @@
 #include "kernel/register.h"
 #include "kernel/newcelltypes.h"
 #include "kernel/rtlil.h"
+#include "kernel/utils.h"
 
 USING_YOSYS_NAMESPACE
 PRIVATE_NAMESPACE_BEGIN
@@ -43,9 +44,11 @@ PRIVATE_NAMESPACE_BEGIN
 				 ID($_XOR_), ID($_XNOR_), ID($_ANDNOT_), ID($_ORNOT_), ID($_MUX_), ID($_NMUX_), \
 				 ID($_AOI3_), ID($_OAI3_), ID($_AOI4_), ID($_OAI4_)
 
-#define CMP_OPS ID($eq), ID($ne), ID($lt), ID($le), ID($ge), ID($gt)
+#define CMP_OPS ID($eq), ID($ne), ID($eqx), ID($nex), ID($lt), ID($le), ID($ge), ID($gt)
 
 #define ARITH_OPS ID($add), ID($sub), ID($neg)
+
+#define SHIFT_OPS ID($shl), ID($shr), ID($sshl), ID($sshr), ID($shift), ID($shiftx)
 
 static constexpr auto known_ops = []() constexpr {
 	StaticCellTypes::Categories::Category c{};
@@ -61,7 +64,9 @@ static constexpr auto known_ops = []() constexpr {
 		c.set_id(id);
 	for (auto id : {ARITH_OPS})
 		c.set_id(id);
-	for (auto id : {ID($pos), ID($pmux), ID($bmux)})
+	for (auto id : {SHIFT_OPS})
+		c.set_id(id);
+	for (auto id : {ID($pos), ID($pmux), ID($bmux), ID($mul)})
 		c.set_id(id);
 	return c;
 }();
@@ -315,6 +320,9 @@ struct Index {
 	// Carry out of each bit of an arithmetic cell
 	dict<CellInstance, std::vector<Lit>> carry_chains;
 
+	// Product bits of a multiplier cell
+	dict<CellInstance, std::vector<Lit>> mul_products;
+
 	// Carry into obit
 	Lit arith_carry(HierCursor &cursor, Cell *cell, SigSpec &aport, SigSpec &bport,
 					bool complement, int obit)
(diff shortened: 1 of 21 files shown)


Ravenslofty:
(to avoid any confusion about the labels, the intended meaning is "don't merge for 0.69, must merge for 0.70")

backends/aiger2/aiger.cc
```diff
@@ -866,15 +1107,33 @@ struct AigerWriter : Index<AigerWriter, unsigned int, 0, 1> {
 	}
 };
 
+bool opaque_box_input(RTLIL::PortDir dir)
+{
+	return dir != RTLIL::PD_OUTPUT;
```
widlarizer, line 1112: This was done as specific compatibility for ambiguous port directions. I think those should be an assertion failure, as the user design should be free of them thanks to check or hierarchy -check, and every supported Yosys synthesis flow is responsible to maintain that invariant. When is this handling kicking in? How do unknown directions end up interpreted?

backends/aiger2/aiger.cc
```diff
@@ -936,10 +1195,19 @@ struct XAigerAnalysis : Index<XAigerAnalysis, int, 0, 0> {
...
+						// A constant carries no CI
+						if (!bit.wire || (ambiguous_dir(dir) && !box_drives(box, bit.wire)))
+							continue;
+
+						pi_literal(bit, &cursor) = 0;
+					}
+				}
 		}
 
 		for (auto w : top->wires()) {
```
widlarizer: I know this isn't changed here, but do we know why is this iterating over wires and filtering for output ports rather than iterating over ports again?

passes/techmap/abc9.cc
widlarizer, line 355: $abc9_holes is prepared and unused in xaiger2 mode (non-`-dff`), which seems potentially wasteful and we should factor it out such that when we switch over to xaiger2 fully we can get rid of the dead code easily

passes/techmap/abc9.cc
```diff
@@ -409,7 +416,10 @@ struct Abc9Pass : public ScriptPass
 						run_nocheck(stringf("abc9_ops -write_lut %s/input.lut", tempdir_name));
 					if (box_file.empty())
 						run_nocheck(stringf("abc9_ops -write_box %s/input.box", tempdir_name));
-					run_nocheck(stringf("write_xaiger -map %s/input.sym %s %s/input.xaig", tempdir_name, dff_mode ? "-dff" : "", tempdir_name));
+					if (xaiger == 1)
+						run_nocheck(stringf("write_xaiger -map %s/input.sym %s %s/input.xaig", tempdir_name, dff_mode ? "-dff" : "", tempdir_name));
+					else
+						run_nocheck(stringf("write_xaiger2 -mapping_prep -map2 %s/input.sym %s/input.xaig", tempdir_name, tempdir_name));
```
widlarizer, line 422: In #6077, did the symbol formats get unified? If not, this is my nit here: Is it a deliberate choice here that .sym is used rather than .map2? Ok if that is the case, but it could be convenient to continue to use the filename extension to differentiate, as long as we are keeping both implementations around

passes/techmap/abc_ops_reintegrate.cc
widlarizer, line 424: This assertion fails in abc_new on this:
```
read_verilog -specify <<EOT
(* abc9_box, whitebox *)
module boxmod(input A, output Y);
parameter P = 0;
assign Y = A;
specify (A => Y) = 1; endspecify
endmodule
module top(input a, output o);
boxmod #(.P(1)) foo (.A(a), .Y(o));
endmodule
EOT
abc_new -liberty tests/liberty/normal.lib top
```
This isn't exposed through a real flow. `abc9` resolves them in `-prep_hier` nicely. `abc_new` flows (gatemate, ORFS) don't have parametric whiteboxes, at least at this point. But, it seems like there's an attempt to derive them above, it's just that the parameters aren't cleared and I think the type isn't fixed up. It seems inconsistent and potentially fragile. The derive circuitry should either be exercised and relied on, or gutted and replaced with an error I think

widlarizer (changes requested):
First review pass, there might be more. Most of my requests here are about things unchanged in this PR, but more related to the abc_new crashes previously discovered, and to abc_new and write_xaiger2.

- `write_xaiger2` doesn't respect `keep` wires by exposing them as outputs as far as I can tell. Regression over `write_xaiger`? Best to fix here and now with test coverage through abc9, unless there's an alternate stance to take on the semantics of `keep` to what we've done before.
 
- The `i` section is written out in `write_xaiger` but not `write_xaiger2`. These arrival times get parsed in abc into `vInArrs` and then `&if` doesn't seem to use them because the code is commented due to memory corruption issues apparently. `&nf` could use them, but was always on `write_xaiger2` so it never tried it... worth looking into later, maybe? Should I open a new issue?

passes/techmap/abc9.cc
Ravenslofty: They are not *quite* unified; .map2 has `pseudopo` lines, .sym does not. It's mostly an artifact of me just copying the line over and I don't think it matters too much either way.

backends/aiger2/aiger.cc
hexratcc: This should kick in for box cells whose type has no module in the design, so `port_dir` returns `PD_UNKNOWN`. In practice I've seen it return `$__FABULOUS_IBUF`/`OBUF` cells synth_fabulous inserts with `iopadmap` before `abc9`, and I think its not caught by `-check` as it skips $ cells.
This version should function very similarly to the old one (when the direction is unknown we look at who drives the wire: if the box drives it, the bit is a box output, otherwise it's a box input. `write_xaiger` also accepted these cells, it just treated every port as both input and output), so this shouldn't allow anything that wasn't allowed before. I think an assert would break `synth_fabulous`, but I think a warning message for unknown box types could work here. EDIT: https://github.com/YosysHQ/yosys/pull/6201

widlarizer (approved):
Agreed on a call that it's best to get exposure to this on main right away, my complaints will be addressed with further refactoring later

PR #6175, opened 2026-09-03
[ENG-2597] - reg rename + sim fixes

stanminlee:
## Summary
- **sim:** bind FST inputs through one pipeline (`drive_vector` → tied-off → `path[k]` bit-selects → flattened unpacked words). A same-name handle whose width does not match the netlist wire is no longer treated as that word (so `din[1]` does not steal the 8-bit `din`).
- **activity_prop:** drop the `DEBUG_ACTIVITY_PROP` env var. Per-module / per-port spam is behind `-d`; the default log is a single end-of-pass summary (bits, wires, modules, missing activity/duty counts).
- **reg_rename:** `group_children` keys dump-order groups with a dict instead of a linear `std::find`, so large fork trees are not quadratic.


stanminlee:
accident sorry

PR #6177, opened 2026-09-04
abc: isolate state mutation in prepare and extract for multi-threaded determinism

### LINKED ISSUE #6170: abc -dff: different netlist on every run with multiple threads (same OS, same input)

widlarizer:
Fixes #6170.

While the ordering of all calls to prepare_module was the same as in single-threaded, and calls to extract, there was interaction left between extract and prepare_module, and the order of interleaved prepare_module and extract calls was left loose. This was exposed specifically in `abc -dff`.

Changes: The sigmap is not read in extract, so it doesn't have to be modified when adding connections. This avoids. Cell removals in prepare were affecting insertion ordering of the design. Autoidx churn also exposed ordering of prepare vs extract.

I have no idea how to test this reliably within our test suite.

`abc -dff` is now pessimized by design. Sorry! It should regress similar to [these numbers](https://github.com/YosysHQ/yosys/pull/6171#issuecomment-5525875909). This is because `-dff` is in conflict with clock domain parallelism as in that mode there's information that can traverse across clock domains as the flops are modeled and don't act as boundary blocks between independently synthesizable. Removing the unsafe interactions takes priority.

We can add a single-threaded mode to `-dff` later that does benefit from connectivity information by fully processing each clock domain at a time. The alternative is making `-dff` single-threaded, together with adding `-dff-parallel` or similar that does run in parallel and can give consistently worse results.

Please retest @oharboe
cc @rocallahan @jmichalski-ant

### DIFF
--- a/kernel/yosys.cc
+++ b/kernel/yosys.cc
@@ -125,6 +125,7 @@ Multithreading::~Multithreading() {
 }
 
 void Autoidx::ensure_at_least(int v) {
+	log_assert(!Multithreading::active());
 	value = std::max(value, v);
 }
 
--- a/passes/techmap/abc.cc
+++ b/passes/techmap/abc.cc
@@ -327,7 +327,7 @@ struct AbcModuleState {
 	void handle_loops(AbcSigMap &assign_map, RTLIL::Module *module);
 	void prepare_module(RTLIL::Design *design, RTLIL::Module *module, AbcSigMap &assign_map, const std::vector<RTLIL::Cell*> &cells,
 		bool dff_mode, std::string clk_str);
-	void extract(AbcSigMap &assign_map, RTLIL::Design *design, RTLIL::Module *module);
+	void extract(RTLIL::Design *design, RTLIL::Module *module);
 	void finish();
 };
 
@@ -787,6 +787,7 @@ void AbcModuleState::handle_loops(AbcSigMap &assign_map, RTLIL::Module *module)
 			}
 			edges[id1].swap(edges[id3]);
 
+			// TODO
 			connect(assign_map, module, RTLIL::SigSig(signal_bits[id3], signal_bits[id1]));
 			dump_loop_graph(dot_f, dot_nr, edges, workpool, in_edges_count);
 		}
@@ -1513,7 +1514,7 @@ void emit_global_input_files(const AbcConfig &config)
 	}
 }
 
-void AbcModuleState::extract(AbcSigMap &assign_map, RTLIL::Design *design, RTLIL::Module *module)
+void AbcModuleState::extract(RTLIL::Design *design, RTLIL::Module *module)
 {
 	log_push();
 	log_header(design, "Executed ABC.\n");
@@ -1563,7 +1564,7 @@ void AbcModuleState::extract(AbcSigMap &assign_map, RTLIL::Design *design, RTLIL
 				RTLIL::IdString name_y = remap_name(c->getPort(ID::Y).as_wire()->name);
 				conn.first = module->wire(name_y);
 				conn.second = RTLIL::SigSpec(c->type == ID(ZERO) ? 0 : 1, 1);
-				connect(assign_map, module, conn);
+				module->connect(conn);
 				continue;
 			}
 			if (c->type == ID(BUF)) {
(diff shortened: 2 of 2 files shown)


Ravenslofty (approved):

oharboe:
Thank you!

Issue #2030, opened 2020-05-05
iCE40 DSP inference: Inferred multiply-add not working if parentheses omitted


dan-rodrigues:
## Steps to reproduce the issue

1. [Test module to reproduce the issue](https://gist.githubusercontent.com/dan-rodrigues/2a4c6ab5fbf98357f4069ccf375a7f91/raw/30511b28e2258a0bde3abfe8ec6b4cd5fbdcb747/infer_mac_test.v).
2. `yosys -p "synth_ice40 -dsp" infer_mac_test.v`

## Expected behavior

`SB_MAC16` multiply-add is inferred. This can be seen if the commented block is used instead of the currently uncommented one, where the extra parentheses are added:

```
=== infer_mac_test ===

   Number of wires:                 16
   Number of wire bits:            151
   Number of public wires:          16
   Number of public wire bits:     151
   Number of memories:               0
   Number of memory bits:            0
   Number of processes:              0
   Number of cells:                  2
     SB_MAC16                        2
```

## Actual behavior

Only the multiply is inferred. The add seems to be implemented using extra cells:

```
=== infer_mac_test ===

   Number of wires:                 20
   Number of wire bits:            239
   Number of public wires:          20
   Number of public wire bits:     239
   Number of memories:               0
   Number of memory bits:            0
   Number of processes:              0
   Number of cells:                 89
     SB_CARRY                       23
     SB_DFF                         40
     SB_LUT4                        24
     SB_MAC16                        2
```

## Version

Latest `master` at time this issue was written:

`Yosys 0.9+2406 (git sha1 584780d7, clang 11.0.0 -fPIC -Os)`

eddiehung:
So the relevant lines of your testcase are:
```
        // this seems to infer multiply-add with the extra parens
        // out <= pretranslate_x * a + (pretranslate_y * b + translate_x_r);

        // this doesn't seem to infer multiply-add. only multiply with extra cells spent
        out <= pretranslate_x * a + pretranslate_y * b + translate_x_r;
```

In the first case, the result before the `coarse` label is this:
![first](https://user-images.githubusercontent.com/6740044/81431863-bdedae80-9116-11ea-8678-ea4ef14d5b73.png)

And for the second case:
![second](https://user-images.githubusercontent.com/6740044/81431880-c47c2600-9116-11ea-9e47-bb1e57c25d93.png)

[...]

PR #6178, opened 2026-09-04
peepopt: reassociate multiply-add chains for DSP inference [sc-814]

hexratcc:
`a*b + c*d + e` parses as `((a*b)+(c*d))+e`, so one multiplier has no `$add` to absorb as a DSP post-adder and only parentheses work around it (#2030). This PR adds a new `muladd` peepopt rule that moves the non-product operand to the bottom of the `$add` chain, moving each product one adder up.


widlarizer:
What functionality executes that fold? Could we teach it that addition is associative instead in a reusable way? Conditionally restructuring with a peepopt to avoid having good analysis might work but may not be the neatest solution. We also were throwing around the idea of a specialized arithmetic representation and rewriting in Yosys on [Galaxy brain](https://docs.google.com/document/d/1KF5GL_y8VFXjh2b5Vpf5efZnd5U4Yh2144_bBXIJMnM/edit?tab=t.u2pxhbkg7ed2#heading=h.7fqo0pgqcuug) but didn't go beyond the basic idea and it's not necessary for just associative addition. We can start with something simpler

hexratcc:
@widlarizer 
> What functionality executes that fold?

The fold is the vendor matcher (`xilinx_dsp.pmg` `postAdd`, `ice40_dsp.pmg` `add`, etc.).
> Could we teach it that addition is associative instead in a reusable way?

I've tried doing so using `$macc` like  we've discussed (I've handled ordering via `maccmap -unmap` before `mul2dsp`). I got it to work, but (at least my attempt) it would introduce a new step in 12 flows, and the QoR seems to have suffered too (2x the DSPs on keep-wire aliases). I think that with enough time a better solution would be supporting `$macc_v2` in matchers, but IMHO that should be done separately, and I'd be down to take a look at it.

> Conditionally restructuring with a peepopt to avoid having good analysis might work but may not be the neatest solution.

I believe the same approach is also used elsewhere, e.g. in `shiftadd`, but I'm open to suggestions

widlarizer:
> 2x the DSPs on keep-wire aliases

This seems worth digging into. Raising arithmetics into $macc and lowering back into binary arithmetic operations should lose no meaningful structure, and introducing $macc-based technology-specific lowering to FMA should achieve what we're trying to do here without breaking anything else. Introducing a new step in 12 flows is fine, they're really only 1 flow with minor unjustified variations, which is something we can address continually by understanding those variations. For example this extra $macc step would only be beneficial for flows that have some reason to benefit from it at that stage and this isn't too hard to figure out and to make newly explicit and neat in the code

hexratcc:
I've pushed my `$macc` fix and replaced the matcher fix with it, I was also able to fix the regression I was originally seeing.

KrystalDelusion:
```
    alumacc -macc-only
    maccmap -unmap
    opt_clean
```

This feels incorrect, or at the very least has bad code smell.  Maybe splitting out the multiply-add chain inference into a separate pass that is automatically called by `alumacc` by default?

widlarizer (approved):
Nice!

Issue #6179, opened 2026-09-04
abc_new crashes in wrapped operator flow


widlarizer:
### Version

Yosys 0.68+188 (git sha1 edb882e87, Debug, Clang /nix/store/...-clang-wrapper-21.1.8/bin/clang++ 21.1.8)

### On which OS did this happen?

Linux

### Reproduction Steps

Run ORFS gcd synthesis with the wrapped operator flow
```make
cd OpenROAD-flow-scripts/flow
make DESIGN_CONFIG=./designs/sky130hd/gcd/config.mk \
     YOSYS_EXE=/path/to/yosys \
     SWAP_ARITH_OPERATORS=1 \
     results/sky130hd/gcd/base/1_2_yosys.v
```
Or run this Yosys script in the root directory of a checkout of the Yosys repository:
```tcl
read_liberty -overwrite -setattr liberty_cell -lib tests/liberty/normal.lib
read_liberty -overwrite -setattr liberty_cell -unit_delay -wb -ignore_miss_func -ignore_buses tests/liberty/normal.lib
read_verilog <<EOV
module tiny(input clk, input [15:0] a, b, output reg [15:0] y);
  always @(posedge clk) y <= a + b;
endmodule
EOV
synth -flatten -top tiny -run :fine
techmap -dont_map $alu
wrapcell -setattr arithmetic_operator -name ALU_{A_WIDTH}_{B_WIDTH}_{Y_WIDTH}{%unused}_HAN_CARLSON t:$alu
techmap -map +/techmap.v -map +/choices/han-carlson.v A:arithmetic_operator
setattr -mod -set abc9_box 1 A:arithmetic_operator
techmap
abc_new -liberty tests/liberty/normal.lib
```
This is a reduced version of the mentioned ORFS flow. Maybe we could base test cases off it to get abc_new coverage

### Expected Behavior

No crash. ORFS master at 8c0616910 currently may be assuming something about the behavior of abc_new that has changed recently. Yosys 0.68 doesn't crash - `Yosys 0.68 (git sha1 38e001a6f, Debug, Clang /nix/store/...-clang-wrapper-21.1.8/bin/clang++ 21.1.8)`

### Actual Behavior

ERROR: Cannot find existing box cell with name '$box0' in original design.

Ravenslofty:
> may be assuming something about the behavior of abc_new that has changed recently.

[...]

PR #6182, opened 2026-09-07
Temporarily revert #6101 and #6077

Ravenslofty:
The current `write_xaiger2`/`read_xaiger` flow in `abc_new` causes issues such as #6179. That can be fixed by #6174, but I'd like that PR to have some time to soak and find bugs in main before reaching a release. With the September release imminent, I think it's best to temporarily revert #6101 and #6077 before that release, then they can be "unreverted" and #6174 merged.

Temporarily yeet.


mmicko:
@Ravenslofty gatemate tests are failing with `ABC: Error: Current library is not available.` and xaiger test with

```
3.5.5. Executing XAIGER2 frontend.
Read 1 instances with cell library of size 10.
<suppressed ~58 debug messages>
ERROR: Malformed design (2)
ERROR: Expected error pattern 'Cannot find existing box cell' not found !
```

KrystalDelusion (approved):
No complaints from me

PR #6192, opened 2026-09-10
opt_ffinv: only treat single-bit $not/$_NOT_ as inverters

### LINKED ISSUE #6189: opt_ffinv crashes with std::out_of_range when a multi-bit $not is connected to a 1-bit FF
**Yosys version:**

```
Yosys 0.69+
commit 01dc90937b59f38e9763e6218ed05d1331296d90
```

**Minimal testcase (`case.v`):**

```verilog
module bug(input clk, d, input [7:0] x, output [8:0] y);
  reg q;
  always @(posedge clk) q <= ~d;
  assign y = ~{x, q};
endmodule
```

**Reproduce with:**

```
yosys -p 'read_verilog case.v; hierarchy -top bug; proc; opt_ffinv'
```

**Output:**

```
4. Executing OPT_FFINV pass (push inverters through FFs).
  what():  SigSpec::operator[]
Aborted (core dumped)
```

Exit code: 134

zphuang0429:
`opt_ffinv` crashes with `std::out_of_range` (`SigSpec::operator[]`) when a multi-bit `$not` cell is connected to a 1-bit FF.

Minimal testcase:

```verilog
module bug(input clk, d, input [7:0] x, output [8:0] y);
  reg q;
  always @(posedge clk) q <= ~d;
  assign y = ~{x, q};
endmodule
```

Reproduce with:

```
yosys -p 'read_verilog case.v; hierarchy -top bug; proc; opt_ffinv'
```

On current `main` (01dc90937):

```
4. Executing OPT_FFINV pass (push inverters through FFs).
  what():  SigSpec::operator[]
Aborted (core dumped)
```

**Root cause:** the FF-LUT discovery treats any `$not`/`$_NOT_` cell connected to the FF as a single-bit inverter, without checking its width. With the testcase above the candidate is a 9-bit `$not`, so `module->connect(lut->Y, ff.sig_q)` constructs a 9-bit-to-1-bit connection, and the subsequent `ModIndex` notification runs past the end of the 1-bit sig_q.

**Fix:** only treat `$not`/`$_NOT_` cells as inverters when both their `A` and `Y` ports are 1 bit wide (new `is_single_bit_inverter()` helper used at all four places where these cell types are checked). Wider `$not` cells are left untouched, so the optimization is simply skipped for them.

Regression test in `tests/opt/bug6189.ys` (crashed before the fix, passes with `check -assert` after).

Fixes #6189


widlarizer:
The behavior you're fixing also might be more of an implicit expectation about a fine-grained netlist which we could make explicit instead. I'm not sure, since there's zero prior test coverage of this pass.

First a meta question, why are you using [these](https://github.com/openecos-projects/ecc/blob/d27dfe18b64b458174c5d799babf62e0ed17f54d/chipcompiler/tools/yosys/scripts/yosys_synthesis.tcl#L427) optimization passes? Are you benefiting from them in QoR beyond arbitrary restructuring noise or in runtime/memory consumption? You're using `opt_ffinv` for standard cell synthesis, but it was explicitly intended for LUT synthesis. 

I have similar doubts about how valuable is it to use some other passes, specifically onehot and opt_demorgan. Haven't seen those used in other ASIC synthesis flows.

KrystalDelusion:
> The behavior you're fixing also might be more of an implicit expectation about a fine-grained netlist 

I believe this is the case, given that this pass only works on single bit FFs and adding `simplemap` prior to `opt_ffinv` works fine.

PR #6194, opened 2026-09-10
abc: module and clock domain parallelism, always

widlarizer:
On main, `abc` is parallelized across clock domains, but only when `-dff` is used, which seems like an arbitrary limitation. It's also parallelized within each module for simplicity, rather than across all modules.

This PR implements parallelism across clock domains across all modules. This happens both with and without `-dff`. This differs from #6166 and #5653 where non-`-dff` doesn't benefit from clock domain parallelism at all. The code damage is very limited as there's minimal conceptual changes, I'm only hoisting and what we have and removing a conditional, with simple code damage as a result (similar to #6177, and reblessing ordering-sensitive tests).

I don't think I can easily include a test because thread sensitive testing isn't appropriate for our CI-oriented test suite. Some more "offline" test could be added.

@jmichalski-ant @QuantamHD let me know what gains you see with this

### DIFF
--- a/frontends/blif/blifparse.cc
+++ b/frontends/blif/blifparse.cc
@@ -84,7 +84,12 @@ failed:
 	return std::pair<RTLIL::IdString, int>(RTLIL::IdString(), 0);
 }
 
-void parse_blif(RTLIL::Design *design, std::istream &f, IdString dff_name, bool run_clean, bool sop_mode, bool wideports)
+static IdString blif_new_id(int autoidx_base, int &local_autoidx)
+{
+	return stringf("$blif$%d$%d", autoidx_base, local_autoidx++);
+}
+
+void parse_blif(RTLIL::Design *design, std::istream &f, IdString dff_name, int autoidx_base, bool run_clean, bool sop_mode, bool wideports)
 {
 	RTLIL::Module *module = nullptr;
 	RTLIL::Const *lutptr = NULL;
@@ -93,6 +98,7 @@ void parse_blif(RTLIL::Design *design, std::istream &f, IdString dff_name, bool 
 	RTLIL::State lut_default_state = RTLIL::State::Sx;
 	std::string err_reason;
 	int blif_maxnum = 0, sopmode = -1;
+	int local_autoidx = 0;
 
 	auto blif_wire = [&](const std::string &wire_name) -> Wire*
 	{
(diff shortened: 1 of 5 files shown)


Ravenslofty (approved):
The change to `arch/gatemate/fsm.ys` is a bit odd, but otherwise looks okay to me.

jmichalski-ant:
I re-checked this on ORFS designs and looks like those don't benefit from clock-domain parallelism. The results are close to https://github.com/YosysHQ/yosys/pull/6166, with https://github.com/YosysHQ/yosys/pull/6194 being slightly slower in some cases (this is the most noticeable in case of ariane136). Part of this difference seems to be caused by the order of module processing (https://github.com/YosysHQ/yosys/pull/6166 processes them in sorted order by size, see https://github.com/antmicro/yosys/blob/971b7b08987c923a4f1b6fac2407d2f080776507/passes/techmap/abc.cc#L2376). Applying this kind of sorting to https://github.com/YosysHQ/yosys/pull/6194 makes them equal in case of DFF mode, but there is still some difference in case of non-DFF mode (e.g ariane136 is then still ~2s slower). I guess this is caused by clock-domain processing overhead

non-DFF mode:
| design name               | main           | https://github.com/YosysHQ/yosys/pull/6166 | https://github.com/YosysHQ/yosys/pull/6194 |
|---------------------------|---------------|--------------------------------------------|--------------------------------------------|
| aes                 | 1.274s                  | 0.581s                                       | 0.581s                            
| ethmac              | 8.685s                  | 5.312s                                       | 5.540s                           
| jpeg                | 4.567s                  | 0.488s                                       | 0.572s                            
| ariane136           | 115.133s                | 25.669s                                      | 29.551s                          
| ibex                | 9.280s                  | 1.962s                                       | 2.048s                           
| black_parrot        | 18.158s                 | 3.056s                                       | 3.155s                           
| tinyRocket          | 16.377s                 | 6.159s                                       | 6.363s                  

DFF mode:
| design name               | main          | https://github.com/YosysHQ/yosys/pull/6166 | https://github.com/YosysHQ/yosys/pull/6194 |
|---------------------------|---------------|--------------------------------------------|--------------------------------------------|
| aes                 | 1.276s                          | 0.579s                         | 0.569s                                   
| ethmac              | 8.627s                          | 5.202s                         | 5.299s                                  
| jpeg                | 4.645                           | 0.509s                         | 0.568s                                   
| ariane136           | 120.779s                        | 29.037s                        | 31.390s                                 
| ibex                | 8.968s                          | 1.874s                         | 1.992s                                  
| black_parrot        | 1.9040s                         | 3.042s                         | 3.229s                                  
| tinyRocket          | 16.337s                         | 6.194s                         | 6.378s

widlarizer:
@Ravenslofty I checked - the gatemate QoR change is purely due to a change in the iteration order

PR #6200, opened 2026-09-14
Use <inlined> for heredoc input filename

mmicko:
Makes log messages more readable and in sync with what is done for read_slang.
Verific creates temporary files for this case so full path for those is displayed.


whitequark:
Nice!

Ravenslofty (approved):

PR #6201, opened 2026-09-14
check: reject unknown internal cell types [sc-915]

hexratcc:
Split out of #6174. `write_xaiger2` had to tolerate box cells with unknown port directions. Those came from `$__FABULOUS_*` pad cells that `synth_fabulous` inserts via `iopadmap` without a module definition. `hierarchy -check` skipped every `$`-typed cell, so nothing caught it. Check now reports a `$`-prefixed cell type that is neither a known cell type nor a module in the design.


widlarizer (approved):
Thanks!

PR #6202, opened 2026-09-14
multiple constant multiplication (mcm) pass

hfthra:
Separate pass(mcm) for multiple constant multiplication operations. Replaces $mul cells that have a constant operand into a shared adder graph of shifts, additions and subtractions. 

It factors out coefficient shifts, greedily constructs a bounded depth A-operation graph for the remaining odd coefficients and reuses intermediate terms across outputs. The pass prunes unused nodes, sizes arithmetic to demanded output widths and a cost estimate to avoid unprofitable rewrites.

This could be useful in blocks such as filters with constant coefficients (FIR etc.) and MAC with constant weights etc.

Simple test file is included with the PR.


nickrallison:
I'll elaborate on my reply to the Discourse post here on my specific thoughts regarding the code.

The implementation doesn't seem unreasonable, but it's a bit hard to parse what the part of the code corresponds to what part of the algorithm (for me at least). I think it would be good if it were made both more generic and explicit what part of the code does what.

```cpp
struct SearchParams {
	int max_depth, max_shift, limit;
	long long search_budget, search_work;
	std::vector<int> targets;
}

struct SearchState {
	bool budget_exhausted;
	dict<int, int> depth;          // value -> depth
	std::vector<AOp> ops;
	dict<std::pair<int, int>, std::vector<AOp>> op_cache;
};

// #################

class Search {
public:
    virtual void search(const SearchParams &params, const Scorer &scorer) = 0;
    virtual ~Search() = default;
};

class HcubSearch : public Search {
public:
    void search(const SearchParams &params, const Scorer &scorer) override {
        SearchState state;
        // ...
    }
};

// #################

class Scorer {
public:
    virtual void score(const SearchState &curr_state) = 0;
    virtual ~Scorer() = default;
};

class ScorerImpl : public Scorer {
public:
    void score(const SearchState &curr_state) override {
	    // ...
    }
};
```

This is pseudocode based on my understanding of the PR and and MCM / other heuristic search problems so do take it with a grain of salt. 

MCM is a pretty active research area, so changing parts of the algorithm should ideally be as simple as possible.

nickrallison:
I still think it would be best if you isolated the search into a concrete implementation of an abstract class.  

```cpp
class Search {
public:
    virtual void search(const SearchParams &params, const Scorer &scorer) = 0;
    virtual ~Search() = default;
};

class HcubSearch : public Search {
public:
    void search(const SearchParams &params, const Scorer &scorer) override {
        SearchState state;
        // ...
    }
};
```

PR #6204, opened 2026-09-15
logger: add LogSourceLocation

mmicko:
- Define and use LogSourceLocation where applicable
- now complete location info is available when message is displayed
- cleanup of log_file_* implementations
- added non-fatal errors
- converted verific errors to use more verbose format


frontends/verific/verific.cc
```diff
@@ -116,20 +116,20 @@ vector<string> verific_incdirs, verific_libdirs, verific_libexts;
 
 void msg_func(msg_type_t msg_type, const char *message_id, linefile_type linefile, const char *msg, va_list args)
 {
-	string message_prefix = stringf("VERIFIC-%s [%s] ",
+	string message_type = stringf("VERIFIC-%s",
```
widlarizer: Could be refactored to a function that produces a `const char*` without stringf

frontends/verific/verific.cc
```diff
@@ -116,20 +116,20 @@ vector<string> verific_incdirs, verific_libdirs, verific_libexts;
...
 			msg_type == VERIFIC_INFO ? "INFO" :
 			msg_type == VERIFIC_COMMENT ? "COMMENT" :
-			msg_type == VERIFIC_PROGRAM_ERROR ? "PROGRAM_ERROR" : "UNKNOWN", message_id ? message_id : "");
-
-	string message = linefile ? stringf("%s:%d: ", LineFile::GetFileName(linefile), LineFile::GetLineNo(linefile)) : "";
-	message += vstringf(msg, args);
+			msg_type == VERIFIC_PROGRAM_ERROR ? "PROGRAM_ERROR" : "UNKNOWN");
 
+	std::string id = message_id ? stringf("[%s] ",message_id) : "";
 	if (log_verific_callback) {
```
widlarizer, line 142: Do we own log_verific_callback? If we do, we could change its signature to accept a location, and then construct a location from a linefile in a function. This would move stuff out of both branches (callback or not)

kernel/log.cc
```diff
@@ -331,34 +336,22 @@ void LogManager::formatted_warning(std::string_view prefix, std::string_view for
...
 		constexpr const char* format = "<suppressed ~%d debug messages>\n";
-		logv_string(LogSeverity::Info, {}, format, stringf(format, debug_suppressed));
+		logv_string(LogSeverity::Info, LogSourceLocation{}, {}, format, stringf(format, debug_suppressed));
 		debug_suppressed = 0;
 	}
 }
 
 [[noreturn]]
-void LogManager::error_with_prefix(std::string_view prefix, std::string_view format, std::string message)
+void LogManager::formatted_error(LogSourceLocation src, std::string_view prefix, std::string_view format, std::string message)
```
widlarizer: shouldn't formatted_error be now conceptually just `formatted_nonfatal_error(); abort();`, deduplicating the shared bits?

frontends/ast/ast.cc
```diff
@@ -1460,7 +1460,7 @@ void AST::process(RTLIL::Design *design, AstNode *ast, bool nodisplay, bool dump
 			if (design->has(child->str)) {
 				RTLIL::Module *existing_mod = design->module(child->str);
 				if (!nooverwrite && !overwrite && !existing_mod->get_blackbox_attribute()) {
-					log_file_error(*child->location.begin.filename, child->location.begin.line, "Re-definition of module `%s'!\n", child->str);
```
widlarizer, line 1463: For future work, it's best to separate the interface change and its application across the codebase. When the two commits stack into the diff of the original commit, the rest will rebase on top without conflicts, but not necessary here, I already checked it as-is

frontends/verilog/verilog_location.h
```diff
@@ -88,6 +88,10 @@ struct Location {
 
 		return oss.str();
 	}
+
+	YOSYS_NAMESPACE_PREFIX LogSourceLocation to_src() const {
```
widlarizer: maybe to_loc() ? src sounds like src attribute to me

widlarizer (approved):

Issue #5735, opened 2026-03-09
opt_dff unsound init handling assumes opt_clean


widlarizer:
### Version

main

### On which OS did this happen?

Linux

### Reproduction Steps

```
module \mod
  attribute \init 4'00
  wire width 2 output 2 \q
  wire input 1 \clk
  wire $2
  wire $3

  cell $dff $ff1
    parameter \WIDTH 1
    parameter \CLK_POLARITY 1'1
    connect \Q $2
    connect \D 1'0
    connect \CLK \clk
  end

  connect $3 $2
  connect \q { $2 $2 }
end
```
`read_rtlil g/init-sensitivity.il; check -assert; opt_dff; opt_clean`

### Expected Behavior

When it removes a flop driving some bits, opt_dff removes init attributes on all connected bits to the previously flop driven bits, or sets the appropriate bits in those init attributes to x

### Actual Behavior

opt_dff keeps a 0 bit on a wire bit aliasing the flop driven wire, this suggests that opt_dff expects opt_clean to be run before it runs, which is a source of fragility

widlarizer:
[sc-343]

Issue #5556, opened 2025-12-19
Behavior of `Module::connect()` with constants is confusing


rocallahan:
The comment in the code says `// ignore all attempts to assign constants to other constants`, but the actual code does more than that. Anywhere there is a constant on the LHS, that bit connection is entirely ignored, even if there is a wire in the corresponding bit of the RHS.

I wrote a fix for that [here](https://github.com/rocallahan/yosys/commit/cd40c118d2a3989c7aa07125aae5ec34ae0f2359) but it fails tests. In particular `opt_expr_combined_assign.ys` binds a zero constant to the output of an OR gate, which then causes `check` to fail with `Warning: Drivers conflicting with a constant 1'0 driver`.

I'm not really sure how to proceed. What is the intended behavior of `Module::connect()` when the LHS is a constant and the RHS is a wire?

widlarizer:
The SigSig connections are in a superposition of being directed and being undirected I'm afraid, see #3426. The semantics are inconsistent and somewhat infected by the directionality of verilog assignments, and assigning to a constant doesn't mean anything. You could say that `SigMap` is the arbiter of reordering things properly but I don't know yet what that implies in this case. What opt_expr_combined_assign is doing is concerning and I'll try to take a look

widlarizer:
The problems in `opt_expr_combined_assign.sv` are that we're hitting #4792 and we've been getting away with it so far. PR for avoiding that is up

rocallahan:
Ok, it sounds like we should fix the comment to match the code, and probably put it somewhere more prominent to make it clear that where the LHS has constant bits, those connections are ignored.

PR #6205, opened 2026-09-15
select: alias wires don't count for cone depth

widlarizer:
For historical reasons, RTLIL, our custom hardware IR models Verilog, not hardware. Verilog has wire-wire connections (I'll call them wire aliases here for convenience), so RTLIL does as well. In hardware, two connected together bits don't exist as two distinct objects. That concept is present purely in the user mental model. In Yosys synthesis, we need to treat them as the same object to implement optimizations.

That's why `SigMap` exists, which maps bits to bits, such that all aliasing wire bits will map to one of them chosen as the "representative". `SigMap` is used in tons of places where this traversal across wire-wire connections is necessary. It's also _unused_ in some places where it is also necessary.

That's because there exists an alternative, a fake solution: `opt_clean` is a garbage collection pass, but it also rewires wires to reduce aliasing in mysterious ways that also try to improve readability of dumped netlists. `opt_clean` provides zero structural certainty, since it would have to be run before every pass to do so, and yet `opt_dff` assumes it is run right before (#5735). `opt_clean` contains underdefined features (#2165, #6107) that came together only because the only concept of sharing analysis information across multiple features was to combine them into a single "pass".

Before and after, `opt_clean` reduces aliasing (those dots are intermediate private wires) and restructures your design for convenience into a new shape (B now driven by A, but they were separate branches before): 

image

There are some use cases where the exact structure of the input is expected to be retained within Yosys (`cutpoint`, `dft_tag`). Whether what they want to achieve is achievable now is given by the exact design and chosen pass sequencing, so they're "what you get is what you get" (#3426, #5556) and there is no real way for me to take them into consideration as a result. This situation can be improved with #4763, we'll need that functionality for SDC in the medium-term future. If somebody wants to pick it up before then for their own sake, I'd be happy to help/review.

Until then, connected wires are aliasing wires, and those should always be treated as the same wire.

So, alias wires don't count for cone depth. Most directly because `tests/opt/opt_balance_tree.ys` couldn't be adjusted for signorm (#5804) if it did (because signorm adds more aliasing), but really because of everything above.

If you're reading this PR description because it broke your Yosys scripts, the fix is simple, or your use case is strange or hits edge cases. Either way you can reach out here or on [Discourse](https://yosyshq.discourse.group/)


passes/cmds/select.cc
```diff
@@ -486,6 +486,69 @@ static int parse_comma_list(std::set<RTLIL::IdString> &tokens, const std::string
...
+		pool<RTLIL::IdString> &selected_members, int &sel_objects, int &max_objects)
+{
+	if (!selected_members.insert(wire->name).second)
+		return false;
+
+	lhs.selected_members[mod->name].insert(wire->name);
+	sel_objects++;
+	max_objects--;
+	return true;
+}
```
hexratcc, line 499: I think there should be a check here so we don't go negative (elsewhere you just check `max_objects != 0`, this also makes it so that the warning doesn't go off). I've also noticed that `select -count o:y %ci*.2` on the new test design gives 6 objects, while main gives 3 + that warning, and I think this is the cause.

tests/select/ci_alias_cone.ys
```diff
@@ -0,0 +1,26 @@
...
+  connect \n1 \n0
+  connect \n2 \n1
+  connect \y \n2
+end
+RTLIL
+
+select -assert-count 5 o:y %ci1
+select -assert-count 1 o:y %ci1 t:$_NOT_ %i
+select -assert-count 0 o:y %ci1 i:a %i
+select -assert-count 6 o:y %ci2
```
hexratcc, line 22: I think a test for the issue above would be nice to have -`select -assert-count 3 o:y %ci*.2`.

passes/cmds/select.cc
```diff
@@ -486,6 +486,69 @@ static int parse_comma_list(std::set<RTLIL::IdString> &tokens, const std::string
...
+
+			if (mode != 'o' && selected_wires.count(lhs_wire))
+				expanded |= select_reach_wire(lhs, mod, rhs_wire, selected_wires, selected_members, limits, sel_objects, max_objects);
+		}
+	}
+
+	return expanded;
+}
+
+static void select_close_aliases(RTLIL::Selection &lhs, RTLIL::Module *mod, std::set<RTLIL::Wire*> &selected_wires,
```
hexratcc, line 543: I'm not sure how often we get deep enough cones for this to matter, but is it necessary for us to sweep all the connections twice?

hexratcc (changes requested):
Mostly minor issues, I also think it could be a good idea to update the help text with info about wire-wire connections being transparent.

PR #6206, opened 2026-09-16
xaiger2: add testcase from #4349

Ravenslofty:
#4349 is apparently fixed with `write_xaiger2`, so let's include it as a testcase to ensure we don't regress that again.


whitequark:
Thanks!

widlarizer (approved):
Checked this reproduces the issue prior to its fix

PR #6208, opened 2026-09-16
yosys: fix share path on MacOS when behind a symlink

widlarizer:
I was used to `./y` being a symlink to `build/yosys` on Linux, but on MacOS this was broke the `share/` path:

`ERROR: init_share_dirname: unable to determine share/ directory!`

The solution is to get the realpath of whatever `_NSGetExecutablePath` gives us.

You can repro the problem on MacOS on main with `ln -s build/yosys y && ./y -p "read_verilog tests/arch/common/add_sub.v; synth"`


mmicko (approved):
LGTM

PR #6210, opened 2026-09-17
Fix concatenation of short and init-only witness traces

maikjjeong:
`yosys-witness yw2yw` crashes on valid short witness traces: a one-step input leaves the progress-report variable `t` unbound, and concatenating an initialization-only trace slices with `[-0:]`, then tries to unpack initialization bits into an empty map.

Filter the decoded first step by each signal's `init_only` flag instead of inferring its layout from the next step's packed string. Report the number of copied steps directly from the input trace length.

The new CLI regression covers one- and two-step traces with and without per-step input bits, and verifies that initialization values occur only in the first output step. Before the fix, 3 of the 4 cases fail; after the fix, all 4 pass.

Validation: `python -m pytest -q tests/functional/test_witness_concat.py` (4 passed). This Python witness utility change does not alter synthesis or solver code. The neighboring SMT solver tests were skipped because z3 is unavailable locally.

Prepared with Codex assistance.


KrystalDelusion (changes requested):
I'm not sure where this test should be (since as far as I can tell there are no existing yosys-witness tests), but it definitely isn't in the functional backends tests, since that is, in fact, for the _functional backends_.  You should also be using the `YOSYS_WITNESS` envvar exported by the test framework, since that points to the actual `yosys-witness` executable, instead of the source python which may or may not be executable.

PR #6212, opened 2026-09-17
cmake: do not alter name in yosys_python_executable

mmicko:
No need to alter name in this case, on mingw (when not using msys) will add .exe to script only but with bash it could be used as well without it.
oss-cad-suite was converting python scripts later with it's own loader wrapper and that one would then have .exe.exe


whitequark:
I think this was added because of Wasm, could you check that it's not broken by the PR please? It might be an outdated workaround.

whitequark (approved):
Nevermind, I completely misread what this was actually doing.

mmicko:
@whitequark sorry for not answering, was at conference so unable to check. Thanks for confirming on your end.

PR #6213, opened 2026-09-17
Execute clean shutdown on windows as well

mmicko:
This looks like oversight to me, no reason not to do a proper shutdown on Windows as well.


widlarizer (approved):

whitequark (approved):

kernel/driver.cc
```diff
@@ -820,13 +820,13 @@ int main(int argc, char **argv)
 		abort();
 
 	log_flush();
-#if defined(_MSC_VER)
+
+	yosys_shutdown();
+
+	#if defined(_MSC_VER)
```
whitequark: (Extra indentation?)

mmicko: Fixed. Thanks

PR #6218, opened 2026-09-18
abc_new: clear exe options on each invocation

hexratcc:
`abc_exe_options` was appended to but never reset, unlike cleanup beside it. A second `abc_new` in one process passed every option twice, so a duplicated `-liberty` made abc unhappy.


povik (approved):

Ravenslofty (approved):

tests/abc_new/generate_mk.py
```diff
@@ -12,7 +12,6 @@
     "hier_check.ys",
     "keep_wire.ys",
     "scopeinfo.ys",
-    "twice.ys",
```
widlarizer, line 14: Unrelated?

widlarizer (changes requested):

tests/abc_new/generate_mk.py
hexratcc: twice.ys is the test file which caught this issue originally (I do the same thing elsewhere throughout this PR stack - the root PR adds these tests as skipped and I unskip them in the child PRs)

widlarizer (approved):

PR #6222, opened 2026-09-19
add `MUX32` cells

jeras:
I have opened this PR draft early, since I have trouble understanding the current build errors, probably related to cell/port identifiers `ID::*`.

Gowin FPGA support dedicated `MUX32` structures. Since other FPGA device families do not appear to support multiplexer sizes above `MUX16` there was no need for a 32-bit cell type before.

I am also wtiting MUX mapping code for `synth_gowin`. Except for tests, the MUX32 and Gowin synthesis code changes are mostly independent.

Most of the changes were done by searching for _MUX16_ and adding a _MUX32_ version of the same code.
- `MUX*` cells use letters from the alphabet as port names and there are only 26 letters in the English alphabet.
  For `MUX16` this means `A`, `B`, ... `P` are used as data inputs and `S`, `T`, `U`, `V` are used as select signals and `Y` is the output. For `MUX32` I decided to keep using `A:P` for inputs `[0:15]` and for inputs `[16:31]` add a prefix `Q` (the first letter after the 16-th input `P`) to get input port names `QA`, `QB`, ... `QP`. Since `Q` is usually reserved for FF/latch outputs, the letter `Z` might be a better choice. The letter `W` is used for the added select input.
- In a few files I broke down some long lines of code.
- I am unsure I wrote the [muxcover.cc](https://github.com/jeras/yosys/blob/jeras_mux32/passes/techmap/muxcover.cc#L503-L674) code right, I did try to follow the pattern. Also I still have to check some indexes in the [argument parsing code](https://github.com/jeras/yosys/blob/jeras_mux32/passes/techmap/muxcover.cc#L893-L947).
- I have not studied the tests yet. In addition to architecture agnostic tests I plan to add specific test for Gowin FPGA. But I have to get the code to compile first.


widlarizer:
Well, that arch/gowin/mux.ys failure doesn't even reproduce for me locally...

PR #6223, opened 2026-09-19
Keep keep wires in abc.

hexratcc:
The xaiger2 writer ignored `ID::keep`, so kept wires were dropped. Each bit is now a primary out, and reintegration drives the original wire from its mapped net.


Ravenslofty:
Not directly related to this PR, but I remembered the existence of `(* abc9_keep *)` from commits like 67fc0c3 as an attribute to mark where an SCC has been broken, though oddly it's not used much in the source any more (possibly from later changes to SCC handling?)

Ravenslofty (approved):

PR #6224, opened 2026-09-19
aiger2: remove $input_port cells when leaving bufnorm

hexratcc:
`bufNormalize(false)` leaves the `$input_port` cells behind, so after abc_new or abc9 every submodule input has one and check reports conflicting drivers.


Ravenslofty (approved):

widlarizer (approved):
We might have to rethink this later, but, for now, this gets the job done for our use cases

PR #6231, opened 2026-09-22
Fix signed FMA correction in wide compressor trees

meetbhadra701-cloud:
# Fix signed FMA correction in wide compressor trees

## Status

Technical description for upstream review.

## Problem

In clean Yosys current main commit
`e8db64c609c7ae4574e66bbd475f5f10f36d7a31`, `synth -arith_tree` can produce
incorrect logic for a signed multiply-add when the product is accumulated at a
width wider than the native product width.

Minimal packaged regression: `tests/arith_tree/arith_tree_signed_fma_wide.ys`.
It synthesizes a two-product signed FMA with `-arith_tree` and proves the
result against an independently elaborated RTL reference with `sat`.

Run it from the repository root with:

```sh
yosys -Q -s tests/arith_tree/arith_tree_signed_fma_wide.ys
```

Before this change, the SAT check reports a counterexample (`a=6`, `b=7`)
and exits nonzero. With this change, it reports
`SAT proof finished - no model found: SUCCESS!` and exits successfully.

On clean current main, the functional SAT proof reports a real counterexample;
on the patched current-main build, it reports SUCCESS. The original FIR4 W4
and W8 reproductions and the minimized W4/W8/W16 cases show the same result.

## Root cause

`CompressorTree::generate_partial_products()` emits Modified Baugh--Wooley
partial products. The old code emitted one final correction bit at
`width_a + width_b - 1`, even when rows had already been zero-extended to a
wider FMA accumulation width. The correction must be ones from that product
sign position through the target width.

## Change

Replace the single correction bit with the W-bit sign-extension mask. Use a
wider intermediate integer for the product-width sum before comparing it with
the target width. FMA remains enabled; no witness-specific behavior is added.

## Validation

* Current clean main: original FIR4 W4/W8 and minimized W4/W8/W16 fail under
  default `-arith_tree`.
* Patched current main: the same cases pass.
* Documented `fa`/`42` compressor and `ripple`/`prefix` final-adder variants
  pass with the patch; `-no-fma` remains a passing control.
* 44-case patched width/signedness matrix: 42 complete PASS proofs and two
  SAT timeouts.
* The packaged regression fails on clean current main with a real SAT
  counterexample (`a=6`, `b=7`, return code 1) and passes on the patched build
  with `SAT proof finished - no model found: SUCCESS!` (return code 0).
* All 14 scripts in the generated `tests/arith_tree` Makefile pass on the
  patched build, including the new regression. The pre-existing 13 scripts
  also pass on both clean and patched builds.
* Two additional alumacc/MACC tests pass on both builds.
* Generic synthesis measurements show no FMA-wide disablement; mapped cell
  counts changed in the bounded FIR4 run, so physical impact is not claimed.

The local contribution commit is `30b851e63ff071578cda4e1e49eccd367f7881d9`,
based on `e8db64c609c7ae4574e66bbd475f5f10f36d7a31`.

## Limitations

The full FIR4 W16 proof remains inconclusive at the 300-second limit. Mixed
RTL signedness was tested, but `Macc::from_cell()` requires equal signedness
for each `$macc_v2` product, so mixed terms that do not enter this path are
not evidence about the Baugh--Wooley correction. The full Yosys suite and
physical design were not run.

The broader bounded validation included the original FIR4 reproducer,
additional signed-FMA widths, compressor/final-adder variants, and existing
arithmetic-tree and alumacc tests. The full FIR4 W16 proof and two larger
matrix cases timed out; those results are not treated as proofs.


meetbhadra701-cloud:
Hi YosysHQ maintainers,

I wanted to clarify the development process behind this PR after reviewing your interim LLM policy.

This contribution originated from a series of experiments investigating Yosys synthesis behavior. During that investigation, we discovered the signed FMA correctness issue described here.

I directed the investigation and experimental requirements, while Codex assisted with execution, source analysis, implementation, formal verification, and preparation of the contribution. The current C++ correction and regression test were generated during that Codex-assisted workflow.

The underlying defect was reproduced against clean upstream Yosys. The PR includes a minimal functional SAT regression that fails on the unmodified version and passes with the proposed correction.

I understand that your interim policy generally asks contributors not to submit LLM-generated bug fixes, and I apologize for not checking that requirement before submitting the PR.

Would you prefer that I close this PR and submit the minimal reproducer as a regular bug report, or would you be willing to consider the existing contribution under your case-by-case exception process?

I'm happy to follow whichever approach is most appropriate for the project.

Thanks for your time and for maintaining Yosys.

kernel/compressor_tree.cc
```diff
@@ -110,8 +110,22 @@ std::vector<DepthSig> generate_partial_products(Module *module, SigSpec a, SigSp
...
-		push_one_at(width_a + width_b - 1);
+	if (a_signed || b_signed) {
+		// The Baugh-Wooley partial products above represent the signed
+		// product with an implicit correction at the product sign bit.
+		// When the product is accumulated at a width larger than
+		// width_a + width_b, that correction must be sign-extended through
+		// the accumulation width. A single bit at the product sign bit
+		// is sufficient at the native product width, but is incorrect
+		// when the product is embedded in a wider FMA tree.
+		long long sign_extension_start = static_cast<long long>(width_a) + width_b - 1;
```
hexratcc, line 121: Is the `static_cast` needed?

kernel/compressor_tree.cc
```diff
@@ -110,8 +110,22 @@ std::vector<DepthSig> generate_partial_products(Module *module, SigSpec a, SigSp
...
+		if (sign_extension_start >= 0 && sign_extension_start < width) {
```
hexratcc, line 122: `sign_extension_start >= 0` is not needed since we're unsigned anyway.

kernel/compressor_tree.cc
```diff
@@ -110,8 +110,22 @@ std::vector<DepthSig> generate_partial_products(Module *module, SigSpec a, SigSp
...
-	if (a_signed || b_signed)
-		push_one_at(width_a + width_b - 1);
+	if (a_signed || b_signed) {
+		// The Baugh-Wooley partial products above represent the signed
+		// product with an implicit correction at the product sign bit.
+		// When the product is accumulated at a width larger than
+		// width_a + width_b, that correction must be sign-extended through
+		// the accumulation width. A single bit at the product sign bit
+		// is sufficient at the native product width, but is incorrect
+		// when the product is embedded in a wider FMA tree.
```
hexratcc, line 120: This comment is too long, I appreciate the explanation but I think a less verbose comment would be better.

tests/arith_tree/arith_tree_signed_fma_wide.ys
```diff
@@ -0,0 +1,58 @@
...
+	assign y = p0 + p1;
+endmodule
+EOT
+
+hierarchy -top fma_impl
+synth -top fma_impl -arith_tree
+rename fma_impl fma_candidate
+
+read_verilog -sv <<EOT
+module fma_gold #(
```
hexratcc, line 28: I don't see a difference between `fma_impl` and `fma_gold`, I think it would be better to use `design -reset ` to avoid writing the design out twice (you can see this approach in other tests as well). Also I think it would be good to test the negate path by adding a `c - a * b` variant.

hexratcc (changes requested):
Thanks for disclosing your LLM usage, I'm open to taking these changes, as the change is relatively small and simple. I have some minor nits which can be seen below. This PR made me think of a possible followup to this that would merge the three constant rows (`2^(m-1)`, `2^(n-1)` and the mask) into one, which would remove compressor inputs, but that's not in the scope for this change, and I can take a look at it when I have the time.

